PR #3398, opened 2016-10-14
[Rx] Jump to top of page when going to Detail or Glossary pages. Smooth scroll to order history for 'Track package'.

### LINKED ISSUE #3344: Rx moving between pages should jump to top
Right now, if I go to the bottom of this Active prescriptions page and click a prescription to go to the detail page, it goes there, but instead of being at the top of the detail page, I am at the bottom. 

In the general case when someone moves between pages, they should enter the new page at the top. The exception is from the Active prescriptions page, if you click "track package" it should take you to the tracking section on the detail page, which is lower down the page (I don't know if we have any prescriptions that are in this state on the active page though...if we don't and this makes it difficult to test, let me know and I can get the data in the API changed as needed)

Going to leave this unassigned for either @U-DON or @webinista to grab as time allows.

U-DON:
Closes #3344.
### Changes
- Loading detail or glossary pages will jump to the top of the page.
- Smooth scroll instead of just skipping to order history after clicking 'Track package' button.
- Fix links that do a full page load instead of going to the route within the app.


src/js/rx/components/GlossaryModal.jsx
```diff
@@ -23,8 +24,12 @@ class GlossaryModal extends React.Component {
             <div className="va-modal-body rx-modal-body">
               <GlossaryList terms={this.props.content}/>
               <div className="va-modal-button-group cf">
-                <button type="button" onClick={this.handlerCloseModal}>Close</button>
-                <a href="/rx/glossary/">See all status definitions</a>
+                <button type="button" onClick={this.handleCloseModal}>Close</button>
+                <Link
+                    to="/rx/glossary"
+                    onClick={this.props.onCloseModal}>
```
webinista: Should this be `handleCloseModal` as well? Or should the name of this prop change?

webinista (dismissed):

src/js/rx/components/GlossaryModal.jsx
U-DON: I didn't pass `handleCloseModal` here because the method does a `preventDefaults`, which stops the link from going to the glossary.

Issue #3351, opened 2016-10-13
ROTC year of commission should not be required


cweber:
Currently is a required field, but shouldn't be.

akainic:
Closed by #3400

PR #3400, opened 2016-10-14
Make ROTC year optional

alexose:
Tiny lil' fix for #3351.


akainic (commented):
Looks good, but also will need to update the validation function for this page to not require the commission year.

alexose:
Whoops!  Fixed.

src/js/edu-benefits/utils/validations.js
```diff
@@ -322,8 +322,7 @@ function isValidRotcScholarshipAmount(data) {
 }
 
 function isValidRotcHistoryPage(data) {
-  return data.seniorRotcCommissioned.value !== 'Y' || (isNotBlank(data.seniorRotc.commissionYear.value)
-    && data.seniorRotc.rotcScholarshipAmounts.every(isValidRotcScholarshipAmount));
+  return data.seniorRotcCommissioned.value !== 'Y' && data.seniorRotc.rotcScholarshipAmounts.every(isValidRotcScholarshipAmount);
```
akainic: This should be `||` not `&&`. Either they weren't commissioned through ROTC or they were in which case the scholarship amounts need to be valid.

akainic (commented):
Little fix for validation logic

src/js/edu-benefits/utils/validations.js
alexose: Erp, right.  The `!==` threw me.   Fixed

src/js/edu-benefits/components/military-history/RotcHistoryFields.jsx
```diff
@@ -27,7 +27,7 @@ export default class RotcHistoryFields extends React.Component {
             onValueChange={(update) => {this.props.onStateChange('seniorRotcCommissioned', update);}}/>
         <div>
           <div className="input-section">
-            <ErrorableNumberInput required
+            <ErrorableNumberInput
                 additionalClass="usa-input-medium"
                 errorMessage={validateIfDirty(this.props.data.seniorRotc.commissionYear, isValidYear) ? undefined : 'Please enter a valid year'}
```
jbalboni: Do you get an invalid year message if you focus and then blur the field without entering anything? I wonder if this needs to be changed to skip isValidYear if the year is blank.

jbalboni (approved):
One question, but LGTM.

src/js/edu-benefits/components/military-history/RotcHistoryFields.jsx
alexose: Good find-- I added a new validation for this case.

jbalboni:
👍

PR #3409, opened 2016-10-17
[Rx] Added expiration note about order history.

### LINKED ISSUE #3342: Rx detail page doesn't load when no tracking information prescription
One current example is the detail page for prescription_id: 13651310 when using the staging API (so http://localhost:3001/rx/prescription/13651310 locally).

Page should load as much information as it can, and say something to the effect of "no tracking information available" instead of just leaving a blank space. Feel free to ask back in this issue if that's not specific enough and we can work through the look/wording/details!

U-DON:
Closes #3342.

Adds a note about expiration in order history so that it doesn't confuse users when we have no tracking info available.

This also includes a few styling tweaks for spacing and sizing for tables on the detail page.

**This is branched from #3398, so merge/review that before this.**
### Desktop

screen shot 2016-10-17 at 12 24 08 pm
### Mobile

screen shot 2016-10-17 at 12 24 21 pm


webinista:
Merged #3398 and now there's a merge conflict. Looked Ok to me otherwise.

PR #3425, opened 2016-10-17
Sign out close browser

crwallace:
This adds the security of signing a user out once they close their browser.

PS - @melwoodard, would love to know your thoughts on if we need to convey to the user that we are signing them out if they close their window. I think I already know the answer to this question ;)


src/js/login/containers/Main.jsx
```diff
@@ -20,17 +20,24 @@ class Main extends React.Component {
...
         this.props.onUpdateLoginUrl(result.authenticate_via_get);
       });
     }
 
     // TODO (crew): Change to just listen for localStorage update but currently known bug in Chrome prevents this from firing (https://bugs.chromium.org/p/chromium/issues/detail?id=136356).
     window.addEventListener('message', this.setMyToken);
+
+    // This removes a user token and signs a user out of vets.gov when they close their browser window.
+    window.onunload = () => {
+      localStorage.removeItem('userToken');
```
plusjeff: Questions from the eng leads channel:

- Is there a reason to store the userToken in localStorage, versus a cookie or sessionStorage?
- There should be a call to the API to close out the user session on the server also ( @robbiethegeek : Does an endpoint to do this exist?)

melwoodard:
@crwallace re: sign out warning - I'm guessing we'd trigger a browser alert when a user attempts to close the tab? I'm hesitant to do that since it would show up every time.

bshyong:
@crwallace is this ready to merge? If so, I can fix the merge conflicts and merge it in today

Issue #3306, opened 2016-10-07
Rx Design Tweaks


gnakm:
@webinista These are some design tweaks that doesn't need stakeholder approval. We have a meeting with stakeholders next Wednesday so after that there will be more adjustments ahead. 

- [x] Bump the Prescription Name size down from 1.6 rem to 1.4 rem
- [x] Flip the order in contact so that Message your provider is below facility and phone number
<img width="480" alt="screen shot 2016-10-07 at 10 19 43 am" src="https://cloud.githubusercontent.com/assets/13420618/19193415/ee7d9e86-8c77-11e6-8435-7cf1af9d3d30.png">
- [x] Relabel the Refill statuses so that it says "Refill status: **`status`**" 
<img width="278" alt="screen shot 2016-10-07 at 10 23 45 am" src="https://cloud.githubusercontent.com/assets/13420618/19193545/6f6fde78-8c78-11e6-8a20-1f3c471b2ff6.png">
- [x] Tabs should be labeled `Refill Prescriptions` and `View History`
- [x] Have the tabs left aligned with header
<img width="518" alt="screen shot 2016-10-07 at 10 28 12 am" src="https://cloud.githubusercontent.com/assets/13420618/19193641/c9f52100-8c78-11e6-9473-70fa6206b402.png">
- [x] Unbold Prescription #: ________ in active tab. 
- [x] The heights of the cards are going to have to increase because Track Package and Refill Prescription buttons may be stacked if both are available
![available for refill track](https://cloud.githubusercontent.com/assets/13420618/19193749/4dcdcc98-8c79-11e6-8d04-c9e74bc6c6c1.png)

U-DON:
@webinista I think this can be closed, right?

webinista:
Yep.

PR #3427, opened 2016-10-17
Set height of prescription cards. Closes #3382.

### LINKED ISSUE #3382: [Rx] Increase height of each prescription card
Blocks / See #3306 

> The heights of the cards are going to have to increase because Track Package and Refill Prescription buttons may be stacked if both are available. 

Cards do not have a specific `height` value. Need to add one, however to accommodate stacked buttons if it occurs.

available for refill track

webinista:
- Adjusted position of .rx-prescription-countaction so that it remains anchored to the bottom.
- Part of #3306

### DIFF
--- a/src/sass/rx/partials/_rx-prescriptions.scss
+++ b/src/sass/rx/partials/_rx-prescriptions.scss
@@ -71,7 +71,6 @@ button.rx-prescription-button,
 }
 
 .rx-prescription-count {
-  // position: absolute;
   bottom: 0;
   display: inline-block;
   &-zero {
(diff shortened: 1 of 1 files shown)


src/sass/rx/partials/_rx-prescriptions.scss
```diff
@@ -70,7 +71,7 @@ button.rx-prescription-button,
 }
 
 .rx-prescription-count {
-  position: absolute;
+  // position: absolute;
```
U-DON: Why not just remove the line?

PR #3442, opened 2016-10-18
Updating status mappings, glossary definitions. Closes #3068.

### LINKED ISSUE #3068: [Rx] "Refill in process" glossary definition
We don't currently have a glossary definition for the "Refill in process" status. We need one.

webinista:


U-DON:
LGTM.

PR #3443, opened 2016-10-18
[Messaging] Sort folder messages

U-DON:
### Changes
- Fixed bug that was causing infinite fetching for the first page of a folder when no page query param was defined.
- Pulled SortableTable component into common (leaving original in Rx to remove in later PR).
- Implemented sorting for folder messages.
  - Works along with pagination as query parameters in URL.
  - Can directly access a page with a sort order.
- Pulled logic for disabling next and prev into MessageNav.
  - Resulted in a single function for changing pages in Folder.
- First pass at date formatting in folder messages table. Will fine-tune the behavior in separate PR.
### TODO
- Remove SortableTable from Rx and update its usage there.
- Pull out common table styling.
- Customize date formatting to our needs.

### DIFF
--- a/src/js/common/components/SortableTable.jsx
+++ b/src/js/common/components/SortableTable.jsx
@@ -68,16 +68,30 @@ class SortableTable extends React.Component {
 
 SortableTable.propTypes = {
   className: React.PropTypes.string,
+
+  // Field value to sort by in either ascending or descending order.
+  // The `value` must be one of the values in the `fields` prop.
   currentSort: React.PropTypes.shape({
     value: React.PropTypes.string.isRequired,
     order: React.PropTypes.oneOf(['ASC', 'DESC'])
   }).isRequired,
-  // Mappings of each header label to the property on each data object.
+
+  // Mappings of header labels to properties on the objects in `data`.
   fields: React.PropTypes.arrayOf(React.PropTypes.shape({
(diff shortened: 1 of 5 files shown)


src/js/common/components/SortableTable.jsx
```diff
@@ -0,0 +1,98 @@
...
+        // Next click will be descending.
+        sortIcon = <i className="fa fa-caret-up"></i>;
+        nextSortOrder = 'DESC';
+      }
+    }
+
+    return (
+      <th key={field.value}>
+        <a onClick={this.handleSort(field.value, nextSortOrder)}>
+          {field.label}&nbsp;&nbsp;&nbsp;{sortIcon}
```
webinista: Using `&nbsp;&nbsp;&nbsp;`, which adds too much of an underline to the sort label. Add a `msg-sort` class to `sortIcon` and use CSS.

webinista (changes requested):
One small CSS-y thing. LGTM otherwise.

src/js/common/components/SortableTable.jsx
U-DON: Sure. Can I merge this and make that style change in the follow-up PR that moves the styles to a common class?

webinista: Yeah. That's fine.

PR #3446, opened 2016-10-19
Only allow numbers for hours on edu benefits

### LINKED ISSUE #3417: Hours completed can be not-a-number
image

Steps:

1. Get to the Education history page
1. Enter letters into the Hours completed field

Expected results:

Error saying the text entered is not a number

Actual results:

Only numbers can be entered, and an error message if letters are entered

jbalboni:
Fixes #3417.

This changes the hours field to use our number input. We should also update the schema to allow floats, since it's restricted to integers now.


akainic (commented):
Tested this locally and something funky is happening when you enter hours completed. If you enter `100.00` and then start deleting numbers, once you delete first the `0` right before the `.` it deletes the `.` and jumps to the beginning of the number. Are you seeing that too? Otherwise the code looks good!

jbalboni:
@akainic Yeah, I see that, too. The DOM event in the number input component sends back "72" when you delete the last zero. Since we always replace the input value with the value in the Redux store, it changes what's displayed. I'm not seeing a way around it after googling for a bit.

akainic:
@jbalboni gotcha. I don't think it's a major problem, as I doubt many people will be deleting things the way I was, but we can file a bug for this to return to later. For now I think we can merge as is. Thanks for looking into it!

jbalboni:
Sounds good. Created #3542.

PR #3452, opened 2016-10-19
Removing placeholders from almost all fields

goldenmeanie:
- I left the dollar sign in the ROTC scholarship field
- I dealt with the APO thing by appending it to the label instead of a placeholder


jbalboni (approved):
LGTM. We can merge it when the CI finishes.

akainic:
@goldenmeanie @jbalboni since this changes common components, make sure someone from the Kudos team reviews these changes as well, probably @webinista or @U-DON.

webinista:
You're using components that we aren't. Doesn't affect us. 👍

PR #3460, opened 2016-10-20
Sends file attachments.

webinista:
Closes department-of-veterans-affairs/messaging-fe#105
- Adding postFormData configuration item.
- Refactoring sendMessage to use FormData object.
- Removes input[type=file] reset from MessageSend.jsx.
- Making the message textarea shadow sit in front of send buttons.
- Correcting function name in reducers/messages.js
- Changing the textarea field name.


U-DON (changes requested):

src/js/messaging/config.js
```diff
@@ -17,6 +17,10 @@ module.exports = {
           'Content-Type': 'application/json'
         })
       },
+     // For sending binary data requests. Assumes use of FormData API.
+      postFormData: {
+        method: 'POST'
```
U-DON: This should have the header for camel casing too, so that the id from the JSON response can be used for the link displayed in the success alert. Maybe also rename the other setting to `postJson` for clear distinction.

U-DON (changes requested):

U-DON (approved):

PR #3462, opened 2016-10-20
[Messaging] Thread props, navigation, and style fixes.

U-DON:
### Changes
- Updated props sent to MessageNav from ThreadHeader to resolve errors about missing props.
- Generalized MessageNav prop names to fit both usages (navigation within folders and threads).
- Navigating previous and next within message threads will actually update the URL.
- Minor style fixes for message bodies in thread.


webinista:
LGTM

PR #3472, opened 2016-10-20
Removing update CharacterCount component, related code.

webinista:
Closes department-of-veterans-affairs/messaging-fe#110


U-DON:
RIP...

PR #3481, opened 2016-10-20
Change page title from waiver to relinquishment

### LINKED ISSUE #3429: Switch waiver back to relinquishment
@mollyblake and @goldenmeanie, I was not on whatever call that this came up, but it sounds like 'relinquishment' may be a word we need to use.

In which case, Alex or I can take care of it while we're making copy changes.

jbalboni:
Closes #3429.

screen shot 2016-10-20 at 5 34 51 pm
screen shot 2016-10-20 at 5 34 32 pm


alexose:
👍

PR #3487, opened 2016-10-21
Fix icon and text on claims list page

jbalboni:
There was an info icon on the callout for no claims on the list page that overlapped with the text. That's fixed now.

screen shot 2016-10-21 at 11 34 25 am


molindoz (approved):
LGTM

PR #3488, opened 2016-10-21
Fix date format on edu benefits confirmation page

jbalboni:
Date format is wrong on the confirmation page.


akainic (approved):
LGTM!

akainic:
Ah actually, I think we also need to update the validation function for the whole section.

PR #3493, opened 2016-10-21
removes main styles from login component

### LINKED ISSUE #3312: Login React app is doubling the size of our React app css files
Right now, the output css size for most of the React apps is about 1.2mb. This is really big and the culprit appears to be the addition of `require('../login/login-entry.jsx')` to the app entry files. If I remove that line, the css output for most of the apps goes down to around 650kb.

I think this is because the login entry page is requiring its own css file, which includes the style.sass file that all the other sass files also include. 

We should be able to fix this by including login.sass in style.sass, instead of importing it in the entry jsx files.

crwallace:
this closes #3312.


bshyong:
LGTM!

PR #3390, opened 2016-10-14
Staging environment

jkassemi:
**WARN**: This PR should be _very_ carefully reasoned about before merge since introducing a staging environment also introduces a divergence that makes it impossible to test the production release before deployment.

This is the result of a conversation in #vetsdotgov-ui (https://dsva.slack.com/archives/vetsdotgov-ui/p1476371861000033), describing difficulties testing the education benefits API from the UI on staging since those features are not yet available on production. As @akainic suggested, it is likely preferable to temporarily point the development environment to the staging API, run the test, and then revert the change once the functionality is confirmed. The same method can be used to test the production API from the staging environment.

PR #3502, opened 2016-10-22
Remove claims status from ignore list

jbalboni:
Removes claims status page from ignore list, so it builds on staging. It's not linked, so it won't be visible in production yet.


austin-martinez (approved):
🌴  LGTM

ayaleloehr:
Hang on, I don't think we want this to be ignored in production yet!  Please don't merge until @jkassemi reviews. I think we want something more similar to https://github.com/department-of-veterans-affairs/vets-website/pull/3390.

ayaleloehr (changes requested):
Sorry to put a block here, but I feel strongly that we don't want this to build in production yet so going to block until we get confirmation I am wrong.

jbalboni:
@ayaleloehr We were told we're supposed to do stakeholder testing on staging. Whatever the best way to get this app to build on staging works for me.

jbalboni:
This isn't useful even if it were merged, since it won't use the staging api. We need the PR Alex linked to be merged to use staging.

PR #3503, opened 2016-10-22
Remove claims status from staging ignore list

jbalboni:
This allows claims status to be used on staging.

### DIFF
--- a/script/build.js
+++ b/script/build.js
@@ -118,7 +118,6 @@ if (options.buildtype === 'production') {
   ignoreList.push('auth/*');
   ignoreList.push('education/apply-for-education-benefits-new');
 } else if (options.buildtype === 'staging') {
-  ignoreList.push('disability-benefits/track-claims/*');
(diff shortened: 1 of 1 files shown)


jbalboni:
cc: @jkassemi @ayaleloehr

ayaleloehr:
This seems right to me. If @jkassemi gives the 👍  here, I will do the corresponding changes for Rx and SM in a separate PR.  I also don't know if `profile/*` needs to be removed for login to show up or not (likely not, but not positive).

jkassemi:
Looks good to me - login components contain logic that limits their functionality outside of the development environment that may need to be modified (`src/js/login/containers/Main.jsx` contains `__BUILDTYPE__ === 'development'` conditionals) to get things working.

We can probably work through this more easily if we set up feature flags. I'll do some thinking on this, but have you guys worked with any decent feature flag libraries for js?

jkassemi (approved):

PR #3507, opened 2016-10-24
changes env variables for correct environments

crwallace:


jkassemi:
@crwallace - the `local` buildtype may not build the required features since the `__BUILDTYPE__` constant is checked for some components (login, hca, education benefits, disability benefits). Do you have accompanying changes to add the `local` buildtype, or is this just for future work?

Also, should we update the production endpoint instead of removing it?

crwallace:
I think I will hardcode the local endpoints for people to use when they are running this locally.

alexose:
LGTM!

jkassemi:
Looks good - how are you going to hardcode the endpoints?

crwallace:
I am going to put that in the login component and it will allow people to un comment what they need when working on their local environments.

PR #3508, opened 2016-10-24
Setting a class name for the default sort value. closes #3501

### LINKED ISSUE #3501: Make clear default sort for Active Rx is "Last requested"
The API will be changing (https://github.com/department-of-veterans-affairs/vets-api/issues/321) to have the default sort for Active prescriptions be "Last requested".  Can we have the active page visually show (bold and not underlined) that this is what is selected when you go to the active page without selecting a sort yourself? 

Current:
screen shot 2016-10-21 at 6 23 36 pm

Desired:
screen shot 2016-10-21 at 6 23 19 pm

webinista:
- Sets defaultCssClass when the option value === this.props.selected


webinista:
Attempted to do this by redirecting to `/rx?sort=lastRequested` on initial load, but that ended in a lot of looping. This works until we can figure out a different way.

U-DON:
LGTM. I think it's fine to not redirect to the URL with the query string for the default sorting.

PR #3509, opened 2016-10-24
Fix scroll handling on claims status

jbalboni:
We're not really handling scroll position in a lot of places. This adds:
- Scroll to the top of the page when changing pages, except when changing tabs (that feels surprising)
- Scroll to the error message when there's an upload error
- Scroll to the just added file when adding files on the upload page
- Scroll to field errors when attempting to submit files
- Scroll to the top of the page when changing pages on the claims list


jbalboni:
@austin-martinez or @cagedcrown, can you take a look at this?

molindoz (approved):
LGTM 💯

PR #3512, opened 2016-10-24
Updated URL routes in Rx and messaging.

### LINKED ISSUE #3477: URL routes for Prescriptions and Secure Messaging
Currently, the prescriptions application has the base URL of /rx, so if this was in production and someone was looking at their historic prescriptions, the URL would be https://www.vets.gov/rx/history and if someone was looking at the details of prescription 123, the URL would be https://www.vets.gov/rx/prescription/123.  

Secure messaging is at /messaging, so if this was in production and someone was looking at their inbox, the URL would be https://www.vets.gov/messaging/folder/0 and if someone was looking at a message thread it would be https://www.vets.gov/messaging/thread/634998

The `/rx` and `/messaging` as the roots for each of these were largely chosen arbitrarily. I do not have strong opinions on Rx, but I do believe messaging should probably be namespaced under `/health` or something similar to be clear that it's about messaging a healthcare provider and not general messaging (such as tech support, which we don't have now on vets.gov but hopefully will in the future). 

 @gnakm @angel @emilyville thoughts?

U-DON:
Closes #3477.

CC: @ayaleloehr 
### Changes
- Updated routes as specified in the issue.
- Switched to using custom history that designates a base path for each app. Updated URLs accordingly.


ayaleloehr:
Re last commit ^, why not push Rx and SM to staging yet?  I am happy if we want to do that as a follow up PR instead, but also am good if it goes into this one.

U-DON:
@ayaleloehr Yeah, either could work. Mainly just trying it to see if it would pass this particular build check that keeps failing, but no such luck.

ayaleloehr:
@U-DON Got tests passing by logging into travis, clearing the cache, then restarting the build. Must have been a bad setup when it first ran!  Would love to have Rx and SM removed from the staging ignoreList in here as well, as that's only two lines and would be awesome to test on staging!

webinista:
LGTM

PR #3517, opened 2016-10-24
Make date of birth field disallow anyone under 17.

### LINKED ISSUE #3413: Date of birth allows years less than 18 years (i.e. for children)
image

This is true in a few areas throughout the application. 

Environment:

- Chrome

jbalboni:
Closes #3413.

Based on discussion in the above issue, we're setting the minimum age to be 17 to make sure we don't exclude anyone.

(Note that today is 10/24):

screen shot 2016-10-24 at 4 40 36 pm
screen shot 2016-10-24 at 4 40 33 pm


akainic (approved):
LGTM!

PR #3520, opened 2016-10-24
Facility locator v3

bshyong:
Facility locator v3 design implementation https://github.com/department-of-veterans-affairs/kudos-team/issues/207
- updated detail page
- updated map page

screen shot 2016-10-24 at 2 39 33 pm


crwallace:
@bshyong -- Could you take a look at the detail page again (https://marvelapp.com/118e46i/screen/16080215)? There seems to be some missing information and structure.

Also, I am seeing some weird padding on the cemetery drop down only

screen shot 2016-10-25 at 10 10 31 am

Also, it might be good to have @gnakm click around on the PR link because I don't think the interactions between the pins and the list are rendering the way she imagined. When I click on the pin, the list doesn't scroll that location to the top and vice versa.

bshyong:
Thanks @crwallace. I'll sync up with Gina to make sure we have all the interactions we need

gnakm:
1. Cemetery icon is being silly. 
screen shot 2016-10-25 at 12 26 04 pm

2. Search Result copy should be left aligned along with search results. The search inserted should be in quotations, so this should look like Search results near "**20006**"
screen shot 2016-10-25 at 12 26 15 pm

3. @crwallace Would making the textbox, all and Select Service Type a super light grey be 508 concern if we're trying to show that it is disabled? I'm afraid this looks sort of like a button. 

screen shot 2016-10-25 at 2 56 34 pm

screen shot 2016-10-25 at 3 07 56 pm

bshyong:
**Update:** in addition to the style changes above, I've also hooked up the front-end interface to the API so that it is using real data.

Notes and questions:
- Benefits facilities are not in the API yet
- What types of services can the user filter by for the Benefits facilities?
- Do we have a loading state yet? Now that we have a real API, the requests will take some time to return.
- @gnakm do we have SVG icons for the map pins? I think we should use SVG pins for consistency instead of trying to overlay font-awesome/custom cemetery icons onto a colored pin in the dropdown & map
- is there anything special we need to do from a devops perspective to allow the locator to communicate with vets-api? Locally, I'm hitting `localhost:3000` to access the API. See (https://github.com/department-of-veterans-affairs/vets-website/blob/5b359c5ad304e84f24409d9a480ca339f0fb5c88/src/js/facility-locator/config.js)

crwallace:
@bshyong -- to communicate with the api you will probably need to use `/src/js/common/helpers/environment.js`

bshyong:
latest state:

We are now using all real data from the API

screen shot 2016-10-28 at 11 36 25 am

src/js/facility-locator/actions/index.js
```diff
@@ -1,14 +1,15 @@
-import sampleData from 'json!../sampleData/sampleData.json';
+import { api } from '../config';
+import { find, filter, compact } from 'lodash';
 import { mapboxClient } from '../components/MapboxClient';
-import { find, filter } from 'lodash';
+import sampleData from 'json!../sampleData/sampleData.json';
```
crwallace: Are we still using sample data?

src/js/facility-locator/components/ServicesAtFacility.jsx
```diff
@@ -131,15 +129,16 @@ class ServicesAtFacility extends Component {
     // TODO: clean up once we have real data
     return (
       <div>
-        {facility.type === 'va_health_facility' ? this.renderHealthServices() : null}
+        <p style={{ margin: '0 0 0.5em' }}>Services current as of <strong>{moment().format('MMMM D, YYYY')}</strong></p>
```
crwallace: won't `moment.().format('MMM D, YYYY')` give us the current date? Do we need to pull this date from the api for each location? Otherwise we probably don't need it on each detail page.

src/js/facility-locator/components/SearchControls.jsx
```diff
@@ -55,32 +55,59 @@ class SearchControls extends Component {
...
-        <option key="primary_care" value="primary_care">Primary Care</option>,
-        <option key="mental_health" value="mental_health">Mental Health</option>,
-        <option key="more_services" value="more_services" disabled>More services coming soon</option>,
-      ];
+    const { currentQuery: { facilityType } } = this.props;
+
+    switch (facilityType) {
+      case 'health':
+        return [
+          <option key="primary_care" value="primary_care">Primary Care</option>,
```
NatalieEMoore: Users can't search by service for health services. 
Users can see services: "mental health" and "primary care" on details page. 
Users can search by service for VBA services. 

@crwallace

src/js/facility-locator/actions/index.js
bshyong: We aren't. I left it in there in case we need it but I can remove it all

src/js/facility-locator/components/ServicesAtFacility.jsx
bshyong: I'm not sure where we want to pull the date from. It's showing the current day right now but we can change it if there is a canonical source for when service data was last updated

crwallace:
This LGTM for the most part. Just a couple of questions.

src/js/facility-locator/components/ServicesAtFacility.jsx
crwallace: ok. I don't think there is one now so this can be a future question.

src/js/facility-locator/components/SearchControls.jsx
crwallace: @bshyong - Lets remove the users ability to search by service for health per @NatalieEMoore's direction.

src/js/facility-locator/actions/index.js
crwallace: 👍

src/js/facility-locator/components/ServicesAtFacility.jsx
NatalieEMoore: the goal on the date is to have it so when services get updates and refreshed on details page it is reflected in a "current as of date".  

Dat specifically asked to make sure we have that.

src/js/facility-locator/components/SearchControls.jsx
bshyong: @crwallace this option is overridden by the service filters being disabled for everything except for Benefits but I will remove it on the next PR so that we don't hold up this merge

PR #3524, opened 2016-10-25
Updates profile api // Adds login gating component

crwallace:
This adds login gating component to `/profile` and `/rx`.

All that is needed is to pass you authRequirement and your component to the `<RequiredLoginView/>` component.

For example: `/profile` only requires an auth level of 1 so I would do the following:

`<RequiredLoginView authRequired={1} component={parentProfileComponent}/>`

@ayaleloehr -- I have already added this component to Rx.
@austin-martinez -- I was going to add this to disability-benefits but I wasn't able to clearly see which component was your parent component. Let me know if you need help implementing this.


config/.sass-lint.yml
```diff
@@ -0,0 +1,145 @@
...
+  trailing-semicolon: 0 # TODO - Change this back to 1 later.
+  url-quotes: 0 # TODO - Change this back to 1 later.
+  variable-for-property:
+    - 0
+    - properties: []
+  variable-name-format:
+    - 1
+    - allow-leading-underscore: true
+      convention: hyphenatedlowercase
+  zero-unit: 0 # TODO - Change this back to 1 later.
```
ayaleloehr: With all the TODOs in this file, what is the "later" that will change it back to 1?  Is there some event or what will happen to allow us to change those to 1?

src/js/auth/containers/LoginApp.jsx
```diff
@@ -24,6 +24,8 @@ class LoginApp extends React.Component {
     }).then(response => {
       return response.json();
     }).then(json => {
+      // console.log(json);
```
ayaleloehr: remove commented out line

src/js/auth/containers/LoginApp.jsx
```diff
@@ -24,6 +24,8 @@ class LoginApp extends React.Component {
...
+      window.opener.localStorage.clear();
```
ayaleloehr: what if something else is in localStorage, does this remove it? (and is that desired?)

src/js/common/components/RequiredLoginView.jsx
```diff
@@ -0,0 +1,152 @@
...
+  constructor(props) {
+    super(props);
+    this.handleLogin = this.handleLogin.bind(this);
+    this.handleVerify = this.handleVerify.bind(this);
+    this.setMyToken = this.setMyToken.bind(this);
+    this.getUserData = this.getUserData.bind(this);
+  }
+
+  componentDidMount() {
+    if (localStorage.length > 0) {
```
ayaleloehr: similar to above, what if localStorage has something else in it to make length > 0, but it's not the userToken. Would it be better to check for userToken instead?

src/js/common/components/RequiredLoginView.jsx
```diff
@@ -0,0 +1,152 @@
...
+  getUserData() {
+    fetch(`${environment.API_URL}/v0/user`, {
+      method: 'GET',
+      headers: new Headers({
+        Authorization: `Token token=${localStorage.userToken}`
+      })
+    }).then(response => {
+      return response.json();
+    }).then(json => {
+      // console.log(json);
```
ayaleloehr: remove commented out line

src/js/common/components/RequiredLoginView.jsx
```diff
@@ -0,0 +1,152 @@
...
+      this.props.onUpdateProfile('dob', userData.birth_date);
+      this.props.onUpdateLoggedInStatus(true);
+    });
+  }
+
+  handleLogin() {
+    const myStore = commonStore.getState();
+    const login = myStore.login;
+    const myLoginUrl = login.loginUrl.first;
+    const receiver = window.open(myLoginUrl, '_blank', 'resizable=yes,top=50,left=500,width=500,height=750');
```
ayaleloehr: Question not to fix in this PR, but does defining these as specific pixel sizes mess things up on mobile or different screens?  If yes, add a TODO? If no, ignore. :)

src/js/common/components/RequiredLoginView.jsx
```diff
@@ -0,0 +1,152 @@
...
+    receiver.focus();
+  }
+
+  handleVerify() {
+    const myStore = commonStore.getState();
+    const login = myStore.login;
+    const myVerifyUrl = login.loginUrl.third;
```
ayaleloehr: It's safe to assume this order will stay the same? (i.e. pulling third will always work?)

src/js/common/components/RequiredLoginView.jsx
```diff
@@ -0,0 +1,152 @@
...
+        <div className="medium-12 small-12 columns">
+          <h1>Sign in to your Vets.gov Account</h1>
+          <p>Vets.gov is a new website from the VA offering online services for Veterans</p>
+          <p>We must meet increased security standards to keep your information secure. To safely verify your identity to these standards, we are using <strong>ID.me</strong>, at third-party service.</p>
+          <p>When you create an account on vets.gov, you will need to verifty your identity through <strong>ID.me</strong> in order for the VA to identify you and locate your records.</p>
+          <p>
+            <button className="usa-button-primary va-button-primary usa-button-big" onClick={this.handleLogin}><strong>Sign In</strong></button>
+            <button className="usa-button-big" onClick={this.handleLogin}><strong>Create an account</strong></button>
+          </p>
+          <p>Having trouble signing in or creating an account? See <a href="#">Frequently Asked Questions</a></p>
```
ayaleloehr: Add TODO to replace the FAQ link so we don't forget later?

src/js/rx/containers/Main.jsx
```diff
@@ -35,6 +38,12 @@ class Main extends React.Component {
         </div>
       </div>
     );
+
+    return (
+      <div>
+        <RequiredLoginView authRequired={3} component={view}/>
+      </div>
+      );
```
ayaleloehr: one tab left

ayaleloehr (approved):
Several nits and questions, but overall really nice work here!.  I am absolutely not the best person to review this, so would like to get another set of eyes on it tomorrow before it gets stale in our heads. With that though, I'm good to give this a LGTM and think this can be merged once you feel you have addressed the questions from this review and that one test passes.  

If future reviewer has additional questions, I think addressing in a follow up PR is appropriate (rather than holding up this one).

config/.sass-lint.yml
crwallace: this file shouldn't even really be here. We aren't using it so I am going to remove it.

src/js/auth/containers/LoginApp.jsx
crwallace: yes it will. No that is not desired.

PR #3525, opened 2016-10-25
Add missed 17 year old validation for edu

jbalboni:
I missed adding the 17 year old validation check to the validation that stops users from moving to the next page. This adds it.


akainic (approved):
LGTM!

PR #3526, opened 2016-10-25
removes 1010AD

### LINKED ISSUE #261: Remove10-10AD
Remove this 10-10AD  form https://www.vets.gov/healthcare/forms/1010ez-AD.pdf  since it does not work in any browser @akainic has tested and confirmed.

we will look at adding this field as an enhancement post november, but should be removed prior to Veterans Day since it does not currently work.

i believe we can then remove those odd instructions on the page as well since no forms will require that?  can we just have the CG pdf form on the page?  @crwallace 

we should clean up this page prior to Veterans Day since comms plan will highlight HCA and drive even more users to this page.

crwallace:
closes department-of-veterans-affairs/healthcare-application#261


koconnor719:
@crwallace i don't know how to do pull requests--i don't want to screw this up:(

akainic (commented):
I think we also have to change the yellow bar at the top of the page, because that currently still says "Active Duty".

Also, I think we can remove the JO form as well. Perhaps @koconnor719 can confirm this?

koconnor719:
yes confirmed; remove JO form.  this is in another ticket w/ the email from HEC team (Stacey Echols confirming it's no longer used.)  @akainic

crwallace:
@akainic // @koconnor719 -- Can you give us the copy for the top bar?

PR #3532, opened 2016-10-25
Remove education application from ignoreList for staging

akainic:
Now that education is password protected on staging (https://github.com/department-of-veterans-affairs/devops/pull/885), we can expose the application on staging for testing!


jbalboni (approved):
LGTM!

PR #3539, opened 2016-10-26
Split military screens on edu benefits

### LINKED ISSUE #3251: Split Military Service form into 3 screens
- ~~Screen 1~~ should end after active duty question (including sub-questions)
- ~~Screen 2~~ should be the growable table. _Small copy change coming._
- ~~Screen 3~~ ROTC with sub-questions (All users will now see this page, just not all the questions)

**Update**: See Caitlin's comment below!

jbalboni:
Closes #3251.

Splits and renames the military pages.

screen shot 2016-10-25 at 10 29 53 pm
screen shot 2016-10-25 at 10 28 49 pm
screen shot 2016-10-25 at 10 28 38 pm


jbalboni:
@cweber Does this look correct to you?

cweber:
Yes! Thank you.

akainic (commented):
Looks great! Found a little bug: when adding a tour of duty, you're allowed to add it without an exit date, even though that's a required field, but then the validation function catches it when trying to proceed to the next page, which is confusing.

jbalboni:
@akainic Good catch! I forgot to change the path in GrowableTable props.

PR #3548, opened 2016-10-26
Fix responsiveness for button on claims document upload form

molindoz:
Closes: [Sunsets #125](https://github.com/department-of-veterans-affairs/sunsets-team/issues/125)
### After

screen shot 2016-10-26 at 12 49 33 pm

### DIFF
--- a/src/js/disability-benefits/components/AddFilesForm.jsx
+++ b/src/js/disability-benefits/components/AddFilesForm.jsx
@@ -105,12 +105,18 @@ class AddFilesForm extends React.Component {
           <div key={index} className="document-item-container">
             <Element name={`documentScroll${index}`}/>
             <div className="document-title-size">
-              <div className="document-title-header">
-                <h4 className="title">{file.name}</h4>
+              <div className="document-title-text-container">
+                <div className="document-title-header">
+                  <h4 className="title">{file.name}</h4>
+                </div>
+                <div className="document-size-text">
+                  <p className="size">{displayFileSize(file.size)}</p>
+                </div>
               </div>
-              <div className="document-size-text">
-                <p className="size">{displayFileSize(file.size)}</p>
+              <div className="remove-document-button">
+                <button className="usa-button-outline" onClick={() => this.props.onRemoveFile(index)}>Remove</button>
               </div>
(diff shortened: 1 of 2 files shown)


src/js/disability-benefits/components/AddFilesForm.jsx
```diff
@@ -105,12 +105,18 @@ class AddFilesForm extends React.Component {
...
             <Element name={`documentScroll${index}`}/>
             <div className="document-title-size">
-              <div className="document-title-header">
-                <h4 className="title">{file.name}</h4>
+              <div className="document-title-text-container">
+                <div className="document-title-header">
+                  <h4 className="title">{file.name}</h4>
+                </div>
+                <div className="document-size-text">
+                  <p className="size">{displayFileSize(file.size)}</p>
```
jbalboni: Do we need the `p` tag in here?

src/js/disability-benefits/components/AddFilesForm.jsx
```diff
@@ -105,12 +105,18 @@ class AddFilesForm extends React.Component {
           <div key={index} className="document-item-container">
             <Element name={`documentScroll${index}`}/>
             <div className="document-title-size">
-              <div className="document-title-header">
-                <h4 className="title">{file.name}</h4>
+              <div className="document-title-text-container">
+                <div className="document-title-header">
+                  <h4 className="title">{file.name}</h4>
```
jbalboni: What happens if the file name is long enough to overlap the button? Does it wrap and look ok?

jbalboni (approved):
LGTM, had a couple questions

src/js/disability-benefits/components/AddFilesForm.jsx
cagedcrown: Good call, I added a width and a word-break property so the text won't go outside of the div.
screen shot 2016-10-26 at 1 35 02 pm

src/js/disability-benefits/components/AddFilesForm.jsx
cagedcrown: I just checked and the text-size is exactly the same without the `p` tag, so I can remove this.

PR #3550, opened 2016-10-26
Updated Rx and SM configs to use environment API URL.

U-DON:
This should get the API requests going to the right URLs on deploy environments (i.e., staging-api.vets.gov as opposed to staging.vets.gov/api).

Also enclosed config defaults in getter functions so they don't run on every import from config.

Note that the API URL imported from the environments module is going to point to localhost:3000, which vets-api uses when run locally.

Now that the user token is required for API requests, I'm not sure how the proxy to staging is going to work if you want to continue using it, but maybe it'll work if you do the login on staging first?


webinista:
LGTM

U-DON:
Going to try some things out before this is merged. Might block local development if not running local vets-api.

PR #3552, opened 2016-10-26
Makes simple search request

webinista:
WIP for department-of-veterans-affairs/messaging-fe#119
- Added SET_SEARCH, SEARCH_SUCCESS, SEARCH_FAILURE constants.
- Changed messaging search input to use ErrorableTextInput.
- Passes current folder ID to MessageSearch component.
- Added SET_SEARCH constant
- Adding an onSearchChange handler.
- Updating search state structure to be {params, advanced}
- Updating PropTypes definitions for MessageSearch, MessageSearchAdvanced

Related to department-of-veterans-affairs/messaging-fe#93


webinista:
To be clear: this just builds the query URL. It's a work in progress, but I wanted to get feedback on what's here now.

src/js/messaging/components/MessageSearch.jsx
```diff
@@ -25,23 +60,41 @@ class MessageSearch extends React.Component {
...
+    }),
+    from: React.PropTypes.shape({
+      value: React.PropTypes.string,
+      dirty: React.PropTypes.bool
+    }),
+    subject: React.PropTypes.shape({
+      value: React.PropTypes.string,
+      dirty: React.PropTypes.bool
+    })
+  })
```
U-DON: Add these prop types to `MessageSearchAdvanced` too.

src/js/messaging/containers/Folder.jsx
```diff
@@ -139,6 +141,7 @@ export class Folder extends React.Component {
 
   render() {
     const folderName = _.get(this.props.attributes, 'name');
+    const folderId = _.get(this.props.params, 'id', 0);
```
U-DON: The id in the URL corresponds to the folder being requested, and the one in attributes corresponds to the id of the actual current folder. This can get out of sync when the folder fails to load. It might be more intuitive to use the attributes id since that's the folder that's currently displayed.

src/js/messaging/reducers/search.js
```diff
@@ -1,30 +1,47 @@
...
   switch (action.type) {
-    case SET_ADVSEARCH_START_DATE:
-      return set('advanced.params.dateRange.start', action.date, state);
     case SET_ADVSEARCH_END_DATE:
-      return set('advanced.params.dateRange.end', action.date, state);
+      return set('params.dateRange.end', action.date, state);
+    case SET_ADVSEARCH_START_DATE:
+      return set('params.dateRange.start', action.date, state);
+    case SET_SEARCH_PARAM:
+      return set(action.path, action.field, state);
```
U-DON: Since it's always only affecting `params`, maybe set it on `params.${action.path}` and cut the `params.` prefix from the path arguments when calling the action creator.

U-DON (commented):
LGTM with some minor comments.

PR #3558, opened 2016-10-26
Copy updates and a small CSS tweak

goldenmeanie:


src/sass/edu-benefits.scss
```diff
@@ -159,6 +168,7 @@ ul.claim-list{
 // Testing a slightly modified required asterisk
 .form-required-span {
   font-size: 1.7em;
+  list-style: 1;
```
jbalboni: Is this intentional?

jbalboni (approved):
LGTM, one question

src/sass/edu-benefits.scss
goldenmeanie: Yes, it fixes the extra space between that was getting added between inputs.

goldenmeanie: Oh. Nope, bad auto-complete! Should have been line-height: 1;

PR #3567, opened 2016-10-27
Added Health and Benefits icon

gnakm:
@bshyong As mentioned, we shall go the svg route instead of font awesome. Added health and benefits icon. Please check!


bshyong:
closing deprecated PR. New pin icons are being used in the map

PR #3573, opened 2016-10-27
Updates the FE to match the new auth url params

crwallace:


robbiethegeek:
Looks good to me @omgitsbillryan want to review?

src/js/login/containers/Main.jsx
```diff
@@ -27,11 +27,11 @@ class Main extends React.Component {
 
     // TODO: Remove this conditional statement when going to production.
     if (__BUILDTYPE__ !== 'production') {
-      this.serverRequest = $.get(`${environment.API_URL}/v0/sessions/new`, result => {
+      this.serverRequest = $.get(`${environment.API_URL}/v0/sessions/new?level=1&location=/auth/login/callback/`, result => {
```
omgitsbillryan: Per our conversation, the `?location` param is going to be removed since the only relay location we will ever use is `/auth/login/callback`

omgitsbillryan (changes requested):

src/js/login/containers/Main.jsx
omgitsbillryan: Granted, passing it wont actually do any harm, but probably best to just remove it.

omgitsbillryan (approved):

PR #3574, opened 2016-10-27
Updates folder view when search results received.

webinista:
Closes department-of-veterans-affairs/messaging-fe#119.
- Replaces SEARCH_SUCCESS, SEARCH_FAILURE constants with FOLDER_SEARCH_SUCCESS, FOLDER_SEARCH_FAILURE
- Adds FOLDER_SEARCH_SUCCESS action, reducer
- Removes SEARCH_SUCCESS from search reducer
- Updating props for MessageSearchAdvanced, making params props required.
- Refactors sendSearch action.
- Refactors the object path.
- Updated folders unit test model.
- Updates URL with message search query and adds a search query param for navigating search results.
- Updates page navigation for search results.
- Added getQueryParams function to containers/Folder.jsx
- Added second argument for createQueryString to make camelCase optional (utils/helpers.js).

Unblocks to department-of-veterans-affairs/messaging-fe#93.


webinista:
Found a bug in this, so hold off on merging.

src/js/messaging/reducers/folders.js
```diff
@@ -87,6 +86,16 @@ export default function folders(state = initialState, action) {
       return set('data.items', items, state);
     }
 
+    case FOLDER_SEARCH_SUCCESS: {
+      const messages = action.messages.data.map(message => message.attributes);
+      const pagination = action.messages.meta.pagination;
+
+      // Update with message, pagination results
+      const newState = set('data.currentItem.messages', messages, state);
```
U-DON: Should it keep sort as well?

U-DON (changes requested):

src/js/messaging/utils/helpers.js
```diff
@@ -1,14 +1,18 @@
...
-export function createQueryString(query) {
+export function createQueryString(query, camelCase = true) {
   const segments = [];
 
   for (const key of Object.keys(query)) {
     // Linter only accepts camelCase keys, but API only
     // recognizes snake_case for query string parameters.
-    const formattedKey = _.snakeCase(key);
-    segments.push(`${formattedKey}=${query[key]}`);
+    if (camelCase) {
```
U-DON: OTOH, it makes sense if it's saying that the query keys are typically camel cased, but the transformation is for anything you want to make snake case.

src/js/messaging/containers/Folder.jsx
```diff
@@ -80,6 +143,16 @@ export class Folder extends React.Component {
     });
   }
 
+  handleSearch(searchParams) {
+    this.props.resetPagination();
```
U-DON: It's kind of convoluted how pagination works in this way. At this point, it simply changes the display of what page you're currently on and what pages the pagination controls lead to. Then when we send the query with the search filters, it uses whatever pagination we just reset to, but it's kind of dependent on the timing of that change in relation to the push below. Rather than using an action creator and modifying the store to achieve that purpose, it would be better to just use the default page we want (I'm guessing the first page) when we make the query. In fact, not specifying any page at all for the API request should get us the first page.

And now thinking about it, upon submitting a search, we probably want to reset the sort as well, since submitting a search would intuitively clear the current filters (in terms of what page you're on and what you're sorted by).

So, ultimately, we wouldn't want to use `this.props.page` or `this.props.sort` when building the query for the request and can altogether drop `page` and `sort` from the URL being pushed to history.

U-DON (approved):
In addition to some thoughts on refactoring, I think that `sendSearch` is basically identical to `fetchFolder`, so we could just use that. And if we're just using `fetchFolder`, that would eliminate the need for the `search` flag in both the store and query string. After that, there would be a few small details around the differences between creating the query for page/sort and for search params. Overall, not that trivial of a refactor, so I want to get the currently working changes in now; I can take care of the refactoring later.

PR #3575, opened 2016-10-27
Hide basic search when Advanced Search is open.

webinista:
- Closes department-of-veterans-affairs/messaging-fe#126
- Closes department-of-veterans-affairs/messaging-fe#40

### DIFF
--- a/src/js/messaging/components/MessageSearch.jsx
+++ b/src/js/messaging/components/MessageSearch.jsx
@@ -56,7 +56,10 @@ class MessageSearch extends React.Component {
       </div>
     );
 
-    const showBasicSearch = this.props.isAdvancedVisible ? <div/> : basicSearch;
+    let showBasicSearch;
+    if (!this.props.isAdvancedVisible) {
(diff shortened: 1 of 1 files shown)


U-DON:
LGTM. Though, I prefer doing the `let basicSearch; if (...) { basicSearch = ... }` way so the JSX doesn't get created if it doesn't have to be.

src/js/messaging/components/MessageSearch.jsx
```diff
@@ -40,25 +40,33 @@ class MessageSearch extends React.Component {
...
+            className="msg-search-btn">
+          <i className="fa fa-search"></i>
+          <span className="msg-search-btn-text">Search</span>
+        </button>
+      </div>
+    );
+
+    let showBasicSearch;
+    if (!this.props.isAdvancedVisible) {
+      showBasicSearch = basicSearch;
```
U-DON: I meant to say that the JSX should be expressed only within the block, so this would be assigned the literal above.

PR #3576, opened 2016-10-27
Display 'No messages' when message list is empty.

webinista:
Affects folder view and search results view.

Closes department-of-veterans-affairs/messaging-fe#127


U-DON:
LGTM.

PR #3577, opened 2016-10-27
[Messaging] Replies and reply drafts.

### LINKED ISSUE #95: Save and send drafts of replies
Currently, the backend doesn't provide a way to associate drafts with the messages they're replying to. Discussion ongoing at department-of-veterans-affairs/vets-api#194.

Holding off on further development until we figure out the details of this feature.

U-DON:
Closes department-of-veterans-affairs/messaging-fe#95 and department-of-veterans-affairs/messaging-fe#113.
### Changes
- Saves drafts of replies such that they're associated with the thread they are supposed to reply to and that thread can be viewed in the draft.
- Merged `sendReply` into `sendMessage` since they share a lot of logic. Requests to send reply will now use form data as well.
- Fixed bug where draft body was persisting across threads.
### TODO
- Resolve issues from department-of-veterans-affairs/vets-api#375.
- Attachments in general are having problems, so will fix that when the cause is determined.

### DIFF
--- a/src/js/messaging/actions/messages.js
+++ b/src/js/messaging/actions/messages.js
@@ -172,9 +172,17 @@ export function saveDraft(message) {
 
 export function sendMessage(message) {
   const payload = new FormData();
-  payload.append('message[recipient_id]', message.recipientId);
-  payload.append('message[category]', message.category);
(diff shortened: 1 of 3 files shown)


webinista:
LGTM

PR #3586, opened 2016-10-28
Fix checkmark color in subway map for claims status and edu benefits

jbalboni:
Closes [127](https://github.com/department-of-veterans-affairs/sunsets-team/issues/127)

Newer versions of iOS use the emoji variant of the heavy check character and don't let you change its color. This adds a variation character to switch it back to the unicode one that we can style. 

Weird: http://stackoverflow.com/questions/32639694/ios-9-removed-the-possibility-to-change-certain-symbol-colors-using-css

screen shot 2016-10-28 at 11 55 23 am


jbalboni:
@akainic FYI

molindoz (approved):
Nice catch. LGTM

PR #3590, opened 2016-10-28
Adds profile to staging.

crwallace:
This allows `/profile` to be deployed to staging.


jcmeloni:
@crwallace Ha. LGTM. When the tests pass, go for it & I'll circle back when it's deployed and will check it out.

jcmeloni:
@crwallace Does this just need a rebase and then get in line for deploy?

PR #3592, opened 2016-10-28
Hide day field for high school graduation date

alexose:
Because of the way our date validations work, I'm assuming that the day always equals the first of the month.

This is pretty inelegant and will lead to problems if a veteran has graduated in the same month that they're filling out the form.

I had actually started down a [different path](https://github.com/department-of-veterans-affairs/vets-website/compare/edu-benefits-date-issue%233293?expand=1) for this, but went for this simple approach in the interest of time.


leannammiller:
@alexose - who is assignee on this?

src/js/common/components/form-elements/DateInput.jsx
```diff
@@ -61,6 +61,13 @@ class DateInput extends React.Component {
       daysForSelectedMonth = days[month.value];
     }
 
+    // Our validations require that there is always a "day" field, so
+    // we'll assume the first day of the month even if we're not displaying it.
+    if (this.props.hideDayField) {
+      day.dirty = true;
```
jbalboni: This seems like it might eventually cause problems, since we're modifying data from the Redux store that should be immutable. We could have this component ignore the day field and default it to this in the initial veteran object.

jbalboni (approved):
Pointed out one issue, but looks alright to me. Like you said, this is a simpler change given the time we have, but I think we should eventually have a separate month/year component.

alexose:
Yeah, I agree... it's not a good idea to mutate the field in the `render` method.  At the very least I should at least do a deep copy.  Fixed.

I started on making a separate component, but abandoned the idea after I realized how tightly coupled our date checking logic is.

My feeling is that we should merge this PR if the change is needed immediately.  Otherwise, I'll spend a couple of hours today finishing up my other branch.

alexose:
Also, I have no idea why this e2e test is failing... /me keeps looking

jbalboni:
@alexose I think it's because it's trying to fill in the day field for that date, but it's now hidden.

leannammiller:
@jbalboni - status on this?

jbalboni:
@leannammiller I'm through reviewing, @alexose can give you the status on fixing the failing checks and merging.

leannammiller:
@alexose can you look at this after HCA is wrapped?

alexose:
Thank you @jbalboni!  Looks great.

Issue #3595, opened 2016-10-28
Allow past relinquishment dates


akainic:
Up to 2 years in the past

akainic:
Closed by #3596

PR #3596, opened 2016-10-28
Allow dates up to two years in the past for relinquished benefits

alexose:
Per #3595.

Not sure about the language-- @cweber @goldenmeanie  any ideas how to word this better?

image


cweber:
@mollyblake This seems like your wheelhouse:

"Please provide a date within the last two year or in the future"

src/js/edu-benefits/utils/validations.js
```diff
@@ -245,11 +245,19 @@ function isValidBenefitsInformationPage(data) {
   return data.chapter33 || data.chapter30 || data.chapter32 || data.chapter1606;
 }
 
+function isValidRelinquishedDate(field) {
+  // Allow dates up to two years ago
+  const pastDate = moment().subtract(2, 'years');
+  const date = dateToMoment(field);
+
+  return !isBlankDateField(field) && date.isAfter(pastDate);
```
jbalboni: Will this have trouble with partially entered dates? You could check `date.isValid()`, too.

goldenmeanie:
I don't think future dates error, so maybe we can ignore that part:

"Date cannot be earlier than two years ago"

jbalboni (approved):
LGTM

src/js/edu-benefits/utils/validations.js
alexose: This is kind of a weird once since we prefill today's date here, so it's not blank unless the user deletes it.

Good call on .isValid().  I'll add that as well.

alexose:
Oh... right.  It must be friday or something ^_^

src/js/edu-benefits/components/benefits-eligibility/BenefitsRelinquishmentFields.jsx
```diff
@@ -30,8 +30,8 @@ export default class BenefitsRelinquishmentFields extends React.Component {
             <div>
               <DateInput required={showRelinquishedEffectiveDate(this.props.data.benefitsRelinquished.value)}
                   allowFutureDates
-                  errorMessage="Please provide a date that's the same as or later than today"
-                  validation={validateIfDirtyDateObj(this.props.data.benefitsRelinquishedDate, isValidFutureDateField)}
+                  errorMessage="Date cannot be earlier than two years ago"
```
akainic: @cweber and @goldenmeanie can you just make sure this language is good? thanks!

akainic (approved):
This looks great, @alexose! Thanks so much for jumping on this so fast! Just want @caitlin and @goldenmeanie to look at the content and then we should merge!

PR #3600, opened 2016-10-31
Remove from ignoreList

akainic:
Should be password protected here: https://github.com/department-of-veterans-affairs/devops/blob/master/ansible/roles/revproxy-configure/defaults/main.yml#L8


jbalboni (approved):
LGTM!

PR #3603, opened 2016-10-31
Update schema with month to number change

akainic:
@jbalboni mind taking a look?


jbalboni (approved):
LGTM! Not much you can really see with this, unless it breaks a test.

akainic:
True :) thanks for the review anyway!

PR #3635, opened 2016-11-01
Updated glossary definitions

webinista:
- Updated glossary configuration object.
- Updated glossary list, glossary page.


U-DON:
LGTM.

PR #3636, opened 2016-11-01
Changing 'Phone number' to 'Facility number'. closes #3623

### LINKED ISSUE #3623: Relabel facility number
Users may confuse contact number as provider's number so change **Phone number** to **Facility number**.

Should look like this:
screen shot 2016-10-31 at 7 03 14 pm

webinista:


U-DON:
LGTM.

PR #3638, opened 2016-11-01
Backfill claims status unit tests

jbalboni:
Related to [157](https://github.com/department-of-veterans-affairs/sunsets-team/issues/157)

This contains unit tests for the components with more than basic rendering. It also fills in the tests for the untested reducers and validations. Once we set up code coverage and see how far off we are, the other components could get some simple unit tests to check that they render at all.

It has one functional change, which is just to switch from a button to a Link in one place.


jbalboni:
@austin-martinez This is big and hard to review, but can you take a look?

cc: @cagedcrown

austin-martinez (approved):
:ice_hockey: LGTM

test/disability-benefits/components/DocumentRequestPage.unit.spec.jsx
```diff
@@ -11,5 +13,121 @@ describe('<DocumentRequestPage>', () => {
...
+    };
+    const claim = {
+      id: 1
+    };
+    const tree = SkinDeep.shallowRender(
+      <DocumentRequestPage
+          claim={claim}
+          trackedItem={trackedItem}/>
+    );
+    expect(tree.subTree('.optional-upload')).not.to.be.false;
```
jbalboni: I was trying to avoid checking content since it feels like that can/will change more often.

PR #3639, opened 2016-11-01
Fix due date less than a day bug on claim status

jbalboni:
Closes [145](https://github.com/department-of-veterans-affairs/sunsets-team/issues/145)

This fixes a bug where we weren't flipping the text color to red if an item was due less than a day ago.


molindoz (approved):
LGTM

Issue #3406, opened 2016-10-17
Google Analytics integration


akainic:

goldenmeanie:
What needs to happen here?

akainic:
I need to configure Google Tag Manager for education, might be able to ask for help from an analytics person for that

ayaleloehr:
Dan and Bob from Ad Hoc should be able to help you here.

leannammiller:
@akainic @goldenmeanie  - where do i find info about the questions we're asking of google analytics? bob wants me to work with someone named dan on this.

akainic:
@leannammiller do you mean what should we be tracking in google analytics? I think that's a question Dan Hoicowitz should be able to answer, but quick thoughts I have on it are:
1. General arrivals and page view counts for each major stage of the application (landing, intro, submit)
2. How many people click on the submit button
3. How many successful/failed submissions there are
4. Where we're losing people along the way (drop-off rate at each page)

leannammiller:
@akainic would you like me to take this off of your plate?

robertfairhead:
@akainic i setup basic education application stats into google analytics this morning. i'll review with Dan at 11a.

@leannammiller let's talk later today and I can catch you up. Dan has an analytics template you can work through.

robertfairhead:
@akainic from your list above items 1,2,4 we have setup via google tag manager. caveat on 2 is that any click on the submit button will be counted even if the form fails validation and doesn't send. onClick still fires on those cases.

for item 3, we can do it via tag manager but after investigating a little bit i think that's a bad idea. basically we'd have to monkey patch fetch() to add in the google datalayer pushes via tag manager. the better solutions are either:

A) add in the couple lines of code into the react app handlers (this is how HCA does it)
B) track the success/failure on the server side instead of via GA

if you want to go with option A, i put together the couple line PR for it or send the lines to Jeff for him to integrate.

leannammiller:
@robertfairhead @dsh32va we will go with option A. Bob, do you mean that you wrote the code already?

leannammiller:
if not, @akainic will do it.

akainic:
Closed by #3462

PR #3642, opened 2016-11-01
Add GA dataLayer events for form submission

robertfairhead:
Added in events for form submission, submission success, and submission failure for Edu Application.

I left commented out sections for whatever data you'd like to attach to the events (submission ID or error text or whatever).

Will setup the tag manager separately to pick up these events and add into appropriate google analytics application for tracking.

This refers to #3406

### DIFF
--- a/src/js/edu-benefits/actions/index.js
+++ b/src/js/edu-benefits/actions/index.js
@@ -109,6 +109,11 @@ export function submitForm(data) {
   return dispatch => {
     dispatch(updateCompletedStatus('/review-and-submit'));
     dispatch(updateSubmissionStatus('submitPending'));
+    
+    window.dataLayer.push({
+      event: 'edu-submission',
+    });
+    
     fetch(`${getApiUrl()}/v0/education_benefits_claims`, {
       method: 'POST',
       mode: 'cors',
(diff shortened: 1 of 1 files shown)


src/js/edu-benefits/actions/index.js
```diff
@@ -124,9 +129,20 @@ export function submitForm(data) {
     })
     .then(res => {
       if (res.ok) {
+        
+        window.dataLayer.push({
+          event: 'edu-submission-successful',
+          //submission-id: If you want to log anything here
```
akainic: I don't think we need to log the submission ids here, since we're storing that in AWS.

src/js/edu-benefits/actions/index.js
```diff
@@ -124,9 +129,20 @@ export function submitForm(data) {
...
+        });
+        
         return res.json();
       }
 
+      window.dataLayer.push({
+        event: 'edu-submission-failed',
+        //submission-error: If you want to log anything here
```
akainic: It could be helpful to log the response in case of an error, but I'm pretty sure we're already logging that elsewhere, so probably not necessary to include in GA. @markolson can verify, but I would say we don't need to include it here for now.

akainic (approved):
Looks great! Thanks so much for tackling this. I assume the code for first including the dataLayer is already being done elsewhere?

robertfairhead:
@akainic GTM will automatically create an empty dataLayer for you if it doesn't find one. Since we're not currently doing any static dataLayer items (since the header/footer are global) we don't have to add it in explicitly.

akainic:
@robertfairhead got it, didn't know that! thank you!

Issue #3605, opened 2016-10-31
Write a11y test for all react apps


crwallace:
Each reactJS app must have an a11y  test written for the 508 scanner to run over it. Pleas include all routes for your apps.

Feel free to use `/test/accessibility/hca-508.spec.js` as an example.

- [x]  Facility Locator
- [x]  Auth
- [x]  Disability Benefits
- [x]  Education Benefits
- [x]  Login
- [x]  Messaging
- [x]  Prescriptions
- [x]  User Profile

Please check your app off once it has been completed.

akainic:
Is everyone able to run the hca-508 test successfully locally? This test fails for me when I try to run it locally :(

jbalboni:
@akainic It was working for me. What are you seeing?

akainic:
This error @jbalboni:    
```
- Begin application (4.032s)
   Timed out while waiting for element <.form-panel> to be present for 1000 milliseconds.  - expected "visible" but got: "not found"
       at Object.Begin application (/Users/vacokainia/Desktop/coding/vets-website/test/accessibility/hca-508.spec.js:13:8)
```

akainic:
Running `npm run build` fixed the issue for me!

crwallace:
Thank you everyone we are all set here! I am closing this out.

PR #3643, opened 2016-11-01
Set up for e2e and 508 tests behind login component

jbalboni:
Related to #3605. 

This is the initial set up to write tests behind a login component for disability benefits.

One global change in here is the e2e test runner uses port 3333 so that we could use it to determine what environment we're running in and use the mock api appropriately.

Also, these are incomplete 508 and e2e tests. I just wanted to get the set up code out there and approved.


jbalboni:
I'm not sure exactly who should review this, but FYI: @crwallace @alexose @ayaleloehr @akainic.

alexose:
Looks great.  I like the idea of using `webstorage` to share the login information.

👍 👍 👍

test/accessibility/disability-benefits-508.spec.js
```diff
@@ -0,0 +1,20 @@
...
+const LoginHelpers = require('../util/login-helpers');
+
+module.exports = E2eHelpers.createE2eTest(
+  (client) => {
+    DisabilityHelpers.initClaimsListMock();
+
+    LoginHelpers.logIn(client, '/disability-benefits/track-claims', 3);
+
+    client
+      .url(`${E2eHelpers.baseUrl}/disability-benefits/track-claims`)
```
crwallace: @jbalboni -- Is this your only route for the entire app?

crwallace (approved):
This looks good to me with one question. Would love for @jkassemi and @bshyong to check this out for awareness.

bshyong:
LGTM thanks for the heads up on the port change

test/accessibility/disability-benefits-508.spec.js
jbalboni: No, there are others, I just wanted to get this set up code merged first.

crwallace: Ok. Let's make sure that we have all of the routes before we check this off as done on the list.

PR #3644, opened 2016-11-01
Fixed auto-scroll to unclosed alert box when its parent component updates

U-DON:
The issue could be reproduced when keeping an alert open (not closing it) and switching between different views (changing tabs in Rx, going to settings or compose pages in Messaging, etc.).

Defined AlertBox's `shouldComponentUpdate` lifecycle method to determine whether or not it needs to update. For some reason, switching between views under the parent container was triggering the update for AlertBox as well.


webinista:
LGTM

PR #3647, opened 2016-11-01
Create 508 test for education benefits

akainic:
As requested in #3605


akainic:
@bshyong @crwallace might need some help with this, getting errors when I run the test locally and I am completely stumped as to what's wrong.

Specifically, the errors I'm getting say that the fields for the "Contact Information" page can't be found. I tried searching for those fields locally, and I found them all. Also, the helper function to complete that page is used by the e2e test for education benefits, which runs without a problem, so that's very strange. If you're able to take a look at this that would be super awesome, thank you!!

bshyong:
@akainic The tests were failing because one of the sections was out of order.

https://github.com/department-of-veterans-affairs/vets-website/pull/3647/commits/ce529cc2bae8949bbb73d4fe3c14cee369d233e1

I just fixed it and merged in master—this LGTM! Take a look and merge if ready.

akainic:
thanks @bshyong!! looks great

crwallace (approved):
LGTM!

PR #3648, opened 2016-11-01
Update date and sorting on claims list page

jbalboni:
Closes [128](https://github.com/department-of-veterans-affairs/sunsets-team/issues/128)

This updates our sorting and last updated date to use the most recent event's date. If there are no events, we're hiding the last updated date on the list page and it will sort by id (descending).


src/js/disability-benefits/components/ClaimsListItem.jsx
```diff
@@ -20,7 +20,8 @@ export default function ClaimsListItem({ claim }) {
           ? <p><i className="fa fa-exclamation-triangle"></i>Items need your attention</p>
           : null}
       </div>
-      <p>Last update: {moment(claim.attributes.updatedAt).format('MMM D, YYYY')}</p>
+      {claim.attributes.phaseChangeDate &&
```
austin-martinez: Is this preferred over an `if` condition?

test/disability-benefits/reducers/claims-list.unit.spec.jsx
```diff
@@ -8,19 +7,27 @@ describe('Claims list reducer', () => {
   it('should sort and populate the claims list', () => {
     const claims = Array(12).fill({
       attributes: {
-        updatedAt: '2010-01-04'
+        phaseChangeDate: '2010-01-01'
       }
     });
-    claims[11].attributes.updatedAt = '2011-01-05';
+    claims[11] = {
```
austin-martinez: Can we add some more data to this test that mimics a scenario where some claims have dates and others only have ID?

src/js/disability-benefits/components/ClaimsListItem.jsx
jbalboni: Yeah. It's sometimes awkward to use `if` in JSX components because it's not an expression in Javascript. So there are two ways people do conditionals:

`{someCondition ? <Component/> : null}`

or 

`{someCondition && <Component/>}`

The later is more concise but has some pitfalls. I usually use the former, but seemed alright here.

test/disability-benefits/reducers/claims-list.unit.spec.jsx
jbalboni: Yep, good idea.

jbalboni:
@austin-martinez I updated the sort order to put claims without a date below claims with dates. Previously, they were at the top.

src/js/disability-benefits/reducers/claims-list.js
```diff
@@ -12,10 +12,17 @@ const initialState = {
...
+  return claim.attributes && claim.attributes.phaseChangeDate
+    ? claim.attributes.phaseChangeDate
+    : '0';
+}
+
 export default function claimsReducer(state = initialState, action) {
   switch (action.type) {
     case SET_CLAIMS: {
-      const sortedList = _.orderBy('attributes.updatedAt', 'desc', action.claims);
+      const sortedList = _.orderBy([phaseChangeDate, 'id'], 'desc', action.claims);
```
austin-martinez: Should we use the `evss_id` here? The vets.gov id will probably change alot as we have to sweep through and remove records every day. Might want to revisit which ID we use throughout the app so OK if we don't solve in this PR

austin-martinez (approved):
:robot: LGTM, minor question about IDs, we can address across the app later if we decide to change it

src/js/disability-benefits/reducers/claims-list.js
jbalboni: That's a good point. We probably shouldn't be using that id for urls now, either.

PR #3651, opened 2016-11-01
Decrease bloat

bshyong:
- removes wow.js
- removes deprecated css and source code

@crwallace I also investigated using PurifyCSS to remove unused css selectors but there seems to be some problems with CSS being removed when it shouldn't be (likely related to metalsmith files not being scanned). Tabling this for now


goldenmeanie:
Does this create a problem for edu benefits form animation @jbalboni ?

jbalboni:
I think wow.js would remove the subway map fade in, but that doesn't seem super important. The other animation we do is our own css.

webinista:
This is a _massive_ pull request, and will take awhile to review. Can you break it up into smaller chunks? Even if it's chunked according to content directory, that's easier to review than this.

crwallace:
I agree with @webinista about the size of this PR. I also think maybe we hold on this PR until after veterans day. @goldenmeanie I think we do need to discuss wow.js because it does create accessibility issues. I am testing with the 508 office later this week and the last time I did manual testing with them a few months ago they suggested not using animation because it was problematic for screenreaders.

goldenmeanie:
@crwallace totally get it. I think the fade in is low value in the face of accessibility problems. But there are definitely other places where I think animation can help.

bshyong:
Makes sense—I'm going to close this PR for now. Do we want to keep the `/old` directory in the repo? Deleting those files is one of the contributing factors to this PR being so large.

crwallace:
As long as we can verify that we don't need anything in the `/old` directory, I would love to get rid of it.

PR #3659, opened 2016-11-01
Changing "Last requested date" to "Last submit date" closes #3628

### LINKED ISSUE #3628: Change all "Last requested date" to "Last submit date"
We initially thought last requested date is more accurate as a label but since the current MHV has it as last submit date, it might be a better transition for users and it's what API is calling it. 

Change all "Last requested date" to "Last submit date"

related to: #3626

webinista:


U-DON:
LGTM.

PR #3658, opened 2016-11-01
Label for British IOT was too vague

goldenmeanie:
Closes #3655

PR #3664, opened 2016-11-01
Update schema

akainic:
Here's the schema change: https://github.com/department-of-veterans-affairs/vets-json-schema/pull/36


jbalboni:
In other words: this change in vets-website looks fine, but the commit from vets-json-schema seems weird.

akainic:
@jbalboni ah, good point. Not sure we ran build before merging this PR: https://github.com/department-of-veterans-affairs/vets-json-schema/pull/36. I tried to do that locally myself and no changes were made, however. So perhaps it's okay?

jbalboni:
@akainic Hmm. That schema change isn't going to affect anything in the front end. Ours is in utils/options-for-select.js. Should probably make the same change there.

akainic:
Yup! We did: https://github.com/department-of-veterans-affairs/vets-website/pull/3658

jbalboni:
@akainic Ah, gotcha. This looks good to me, then.

jbalboni (approved):

PR #3667, opened 2016-11-01
Auto verified level up in child window

### LINKED ISSUE #3606: Up-level to loa3 when highest === 3

crwallace:
This PR adds:
- Auto level up in child window
- logout of gating component from common login component 
- Changed name of `LoginApp` to `AuthApp`

This is by no means a perfect solution but the is the best UX solution for our situation. I think this is our main reason for thinking about going full web app.

closes #3606


src/js/common/helpers/login-helpers.js
```diff
@@ -0,0 +1,34 @@
...
+  fetch(`${environment.API_URL}/v0/user`, {
+    method: 'GET',
+    headers: new Headers({
+      Authorization: `Token token=${localStorage.userToken}`
+    })
+  }).then(response => {
+    return response.json();
+  }).then(json => {
+    const userData = json.data.attributes.profile;
+    commonStore.dispatch(updateProfileField('accountType', userData.loa.current));
```
bshyong: I wonder if it's worth it to have an action to update attributes en masse—probably not if we're only doing this type of updating in this component

bshyong (approved):
LGTM

PR #3670, opened 2016-11-01
Adds check for empty form, displays error message.

webinista:
Closes department-of-veterans-affairs/messaging-fe#93

- Also refactored handleSearch to use spread operator


src/js/messaging/containers/Folder.jsx
```diff
@@ -236,9 +243,11 @@ export class Folder extends React.Component {
             isAdvancedVisible={this.props.isAdvancedVisible}
             onAdvancedSearch={this.props.toggleAdvancedSearch}
             onDateChange={this.props.setDateRange}
-            params={this.props.searchParams}
+            onError={this.props.openAlert}
+            onErrorClose={this.props.closeAlert}
```
U-DON: Is this needed? Doesn't look like it's being used.

webinista: Sure. I'll remove it. Anything else I should change before I re-assign it for review?

src/js/messaging/containers/Folder.jsx
```diff
@@ -279,7 +286,9 @@ const mapStateToProps = (state) => {
 };
 
 const mapDispatchToProps = {
+  closeAlert,
```
U-DON: Also can remove this since it's not being used here.

src/js/messaging/containers/Folder.jsx
webinista: Wait: was the comment to remove the params or the onerrorclose?

webinista: nvm. fixed it.

U-DON:
LGTM.

PR #3682, opened 2016-11-02
Add code coverage tooling

jbalboni:
I thought this was useful information to know. It doesn't affect our existing build process; you have to run `npm run test:coverage` to see the results. There are ways to fail the build if coverage numbers dip below a certain level, but that's another discussion.

Reports look like this (HTML in `coverage/` folder and console output):

screen shot 2016-11-02 at 11 02 07 am
screen shot 2016-11-02 at 11 01 54 am


alexose:
Looks great!

.eslintignore
```diff
@@ -8,3 +8,5 @@ src/js/legacy/*
 src/hca-migrate/*
 src/rx-migrate/*
 schema/*
+coverage/*
+nyc_output/*
```
jbalboni: Yep!

bshyong:
LGTM!

PR #3684, opened 2016-11-02
adds 508 tests for rx

### LINKED ISSUE #3499: Rx Accessibility testing
We need to set up and run automated accessibility testing for Rx.  Talking to @crwallace, it seems the steps are:
- Create a file like [this example from HCA](https://github.com/department-of-veterans-affairs/vets-website/blob/master/test/accessibility/hca-508.spec.js) that goes through each page of Rx and runs `.axeCheck('.main')` on it. 
- Then to run the tests, first start selenium: `npm run selenium:bootstrap`
- Then run the accessibility tests: `npm run build && npm run test:accessibility`

Last time we were going through issues, @webinista and @U-DON seemed to prefer things just get assigned to one of you then reassigned later if needed, so going to assign to @webinista , as @U-DON has been focused more on the unit tests. Let me know if you want those switched though!

crwallace:
This will fail because some issues need to be addressed but the test is available now.

closes #3499


U-DON (commented):
LGTM, but I think the file was meant to be named `rx-508.spec.js`.

U-DON:
Rx is passing now. Merging.

PR #3687, opened 2016-11-02
Reverting Rx/Messaging workaround that used the proxy instead of local API.

U-DON:
You'll need this change to point API requests to the local vets-api running at localhost:3000. It appears the proxy doesn't work with logins, and you'd need the token from the login to authorize the API requests.


webinista:
LGTM once tests pass

PR #3689, opened 2016-11-02
Facility locator a11y test

bshyong:
Adds 508 test for facility locator route per https://github.com/department-of-veterans-affairs/vets-website/issues/3605


crwallace (approved):
LGTM!

PR #3690, opened 2016-11-02
Fix file upload error check

jbalboni:
Closes [190](https://github.com/department-of-veterans-affairs/sunsets-team/issues/190)

I was checking too specific of a response code in a hack I had to do to get the uploader to work correctly. This should work, but I haven't been able to test it locally.

No unit tests for this because we don't have a way to mock libraries right now.


austin-martinez (approved):
LGTM 🇲🇴

PR #3695, opened 2016-11-02
Use the correct checks for tracked item state

jbalboni:
Closes [192](https://github.com/department-of-veterans-affairs/sunsets-team/issues/192)

When files are uploaded, the status changes but the list doesn't. We were accounting for that on the status tab, but not on the files tab.

Claim 344 from Freeman test account now:

screen shot 2016-11-02 at 5 32 28 pm


austin-martinez (approved):
LGTM 🇱🇦

PR #3698, opened 2016-11-02
Use correct base paths in history for facility locator

### LINKED ISSUE #3692: Open facility detail page in new tab goes to 404 page
From map pane, right click on one of the facility names and open in new tab, it shows a URL like `https://dev.vets.gov/facilities/facility/vha_358` in the location bar, but displays the "Sorry - we can't find that page' 404 response.

bshyong:
This may resolve #3692 if the reverse proxies are already configured correctly


crwallace (approved):
LGTM!

PR #3703, opened 2016-11-03
[Messaging] Refactored search logic.

U-DON:
### Changes
* Used `fetchFolder` as API request for search updates.
* Keep filter data in store to track whether the folder needs to send a new request after URL is updated.
* Drop page and sort when submitting a search; search just starts on first page.
* Removed unnecessary snake-casing flag in helper function to create URL query.
* Removed now unused imports, functions, constants, etc.


webinista:
LGTM

PR #3704, opened 2016-11-03
Use the production backend in production

markolson:


akainic (approved):
LGTM! But perhaps @ayaleloehr can also take a quick peek?

ayaleloehr:
LGTM. 

You're going to have to be off the VA network for this to work today though. @ShawnArnwine is working on fixing that, but it's not resolved yet.

PR #3706, opened 2016-11-03
Show notification after uploading additional evidence

jbalboni:
Closes [194](https://github.com/department-of-veterans-affairs/sunsets-team/issues/194) and [196](https://github.com/department-of-veterans-affairs/sunsets-team/issues/196)

I refactoring how the file alerts work into more generic notifications. I think that's easier to follow. I also updated the success messages to go away when you switch pages and fixed some spacing issues.

screen shot 2016-11-03 at 11 32 33 am
screen shot 2016-11-03 at 11 32 02 am


src/js/disability-benefits/components/Notification.jsx
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+import classNames from 'classnames';
+
+export default function Notification({ title, body, onClose, type }) {
```
austin-martinez: There is a common component `AlertBox`, should we be using that? It looks pretty similar but has a few differences in element nesting

jbalboni: I forgot about that, let me take a look.

jbalboni: Looks like we have a title and that component doesn't. It's something to look into later, I think.

austin-martinez (approved):
LGTM 👾

PR #3707, opened 2016-11-03
Fix text when no contention list in claims status

jbalboni:
Closes [179](https://github.com/department-of-veterans-affairs/sunsets-team/issues/179)

Sets contention list to 'Not available' if the list is empty or missing.

screen shot 2016-11-03 at 12 03 09 pm


molindoz (approved):
LGTM

PR #3713, opened 2016-11-03
Hide action items on claims when complete and add date received

jbalboni:
Closes [170](https://github.com/department-of-veterans-affairs/sunsets-team/issues/170) and [169](https://github.com/department-of-veterans-affairs/sunsets-team/issues/169)

Adds another date and hides some items when a claim is the complete phase, instead of just having a letter sent.

Also changed the name of a function to be clearer (and not conflict with this idea of "complete").

screen shot 2016-11-03 at 1 57 08 pm


austin-martinez (commented):
LGTM :beach_umbrella:

Issue #3492, opened 2016-10-21
Update browser notice to say v11 is the minimum version of Internet Explorer (IE)


plusjeff:
Previously, the minimum required version of IE was v9. IE11 is the new minimum, largely because < 2% of users come with IE9 or 10 and supporting those browsers would require a disproportionate amount of development effort.

More details in:
https://github.com/department-of-veterans-affairs/vets.gov-team/issues/358

crwallace:
@bshyong -- Can you work with @emilyville to figure out what this banner/copy should look like?

emilyville:
@bshyong @crwallace Can you give me more context for what you have so far and what you still need?  I think this should likely be treated as an alert/warning in what @melwoodard is working on.

emilyville:
Also tagging https://github.com/department-of-veterans-affairs/vets.gov-team/issues/365 since this is similar. Might be worth adding it in and tagging Anitra to look at copy.

crwallace:
I know we had discussed something like this, however we thought the yellow was too harsh.

Can we get @bshyong a mockup to work from for this specific alert copy?

<img width="1548" alt="screen shot 2016-10-25 at 11 41 17 am" src="https://cloud.githubusercontent.com/assets/3453669/19927954/add2d706-a0d2-11e6-9fd0-9fe550b871ac.png">

CC: @melwoodard // @AnitraAppa

AnitraAppa:
Hey there! Recommending:
"Your browser is out of date. To use this website, please <update your browser> or use a different device. If you're using Internet Explorer, try turning off Compatibility View."

Notes:
- I took out the "We're sorry" here because it can make people nervous about there being some kind of serious error on our part or something they can't remedy, when ideally this should be a quicker fix. Want to get to the solution as quickly as possible.
- Changed "does not support all required functionality" to "to use this website" to simplify and make it clear that they'll really want to update their browser or use another device to use this website at all

Let me know if that works! I can add it to the larger  department-of-veterans-affairs/vets.gov-team#365 issue as well.

melwoodard:
I like your take @AnitraAppa. 

@crwallace what are the odds that a user could be in Compatibility View? Is it fairly common or rare enough that we can leave that out?

I think using our mellow yellow will help to soften this visually, and there should be an option to close/hide it too. Thoughts?

[...]

PR #3716, opened 2016-11-03
Browser warning

bshyong:
- Resolves https://github.com/department-of-veterans-affairs/vets-website/issues/3492


content/includes/header.html
```diff
@@ -141,4 +148,21 @@ <h1 role="banner" class="va-header-logo"><a href="/">Vets.gov</a></h1>
...
   document.getElementsByTagName("head")[0].appendChild(script);
+
+  (function() {
+    var domReady = function(callback) {
+      document.readyState === "interactive" || document.readyState === "complete" ? callback() : document.addEventListener("DOMContentLoaded", callback);
+    };
+
+    domReady(function() {
+      // check whether browser is IE10 and older
+      if (window.navigator.userAgent.indexOf('MSIE ') > 0) {
```
crwallace: @bshyong -- Won't this just check to see if the browser is Internet Explorer and not what version it is?

bshyong: MSIE is only present in versions 10 and below; this works because our minimum version is IE11

crwallace (approved):
LGTM!

PR #3718, opened 2016-11-03
Add loading state to facility locator

bshyong:
Adds loading indicator to search results pane


crwallace (approved):
This looks great. I am also very happy that we added, "no facilities found" when there are none in the bounds.

PR #3719, opened 2016-11-03
Adding error message when there are no prescriptions. Closes #3671.

### LINKED ISSUE #3671: Rx error to show user when API returns 500?
Currently on Rx when our API returns a 500 error (`"{"errors":[{"title":"Internal Server Error","detail":"Internal Server Error","code":"500","status":"500"}]}"`), the active prescriptions page looks like the below.  I think we should show some sort of text there in this case. 

@gnakm and @emilyville Do you agree? What message should we should we show to the user when this occurs? 

screen shot 2016-11-01 at 7 55 23 pm

webinista:


U-DON:
LGTM.

PR #3722, opened 2016-11-03
Adding Object.assign polyfill to messaging config file.

webinista:
Work on department-of-veterans-affairs/messaging-fe#146.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,9 @@
   "nyc": {
     "sourceMap": false,
     "instrument": false,
-    "include": ["src/**"]
+    "include": [
+      "src/**"
+    ]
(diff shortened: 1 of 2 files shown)


webinista:
Assuming this fixes department-of-veterans-affairs/messaging-fe#146, I'll update other places where we use `Object.assign` to use this polyfill.

U-DON:
I think lodash provides an assign function as well, if you wanted to cut down on dependencies, or if this doesn't work. But LGTM either way.

webinista:
Bleh. Changing the syntax didn't work. Going to switch to lodash's `_.assign`

PR #3723, opened 2016-11-03
Replacing Object.assign with assign from lodash.

webinista:
Work on department-of-veterans-affairs/messaging-fe#146

### DIFF
--- a/src/js/messaging/config.js
+++ b/src/js/messaging/config.js
@@ -1,4 +1,5 @@
 import environment from '../common/helpers/environment';
+import assign from 'lodash/fp/assign';
 
 function getHeaders(optionalHeaders = {}) {
   const defaultHeaders = {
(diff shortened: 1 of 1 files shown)


src/js/messaging/config.js
```diff
@@ -9,7 +10,7 @@ function getHeaders(optionalHeaders = {}) {
     defaultHeaders.Authorization = `Token token=${localStorage.userToken}`;
   }
 
-  return Object.assign({}, defaultHeaders, optionalHeaders);
+  return assign({}, defaultHeaders, optionalHeaders);
```
U-DON: Since the FP functions have no side effects, I think you can just do without the `{}` in the first parameter wherever we use this form of `assign`. Whatever goes in the first parameter shouldn't mutate.

U-DON:
LGTM.

PR #3726, opened 2016-11-04
[Rx] Smooth scroll to top of History content when changing pages.

### LINKED ISSUE #3680: Add Smooth scroll to History page
On mobile, when I try to change pages in the view history section, it does not scroll to the top of the page. This is the same jumping problem that we had with the confirmation messages.

U-DON:
Closes #3680.

### Changes
* Smooth scrolling to top of History content after page update.
* Renamed variables in `componentDidUpdate` in History.
* Added error message in History when there are no prescriptions.
    * Although, I don't think these really work right now... Might have to handle the error more explicitly in the action creator and will revisit when implementing the loading screen).


webinista:
LGTM

PR #3727, opened 2016-11-04
Changes "Back to detail page" to "Back" on GlossaryPage.closes #3341

### LINKED ISSUE #3341: Rx glossary back button
Currently the back link on the Glossary page always says "back to detail page", even if you didn't come from the detail page. The back functionality works in that it sends you back to whatever page you came from, it's just that the title of the link is static.

Best case: link should be dynamic, so it would be "back to detail page", "back to history page", or "back to active prescriptions" (or some similar text), as appropriate

Alternative if that proves difficult: change link to just say"back"

webinista:


U-DON:
Would've liked to get this saying "Back to [specific page]", but we have bigger fish to fry. LGTM.

ayaleloehr:
Nice scope decision as we narrow in on Veterans day!

ayaleloehr:
Merging while this is updated and passing.

PR #3729, opened 2016-11-04
Wrap new accessibility tests in buildtype guards

akainic:
@jkassemi is this what you were looking for?


akainic:
Not sure if this is the right fix for the problem, temporarily closing for now.

Issue #3730, opened 2016-11-04
Handle 'Record not found' prescriptions for detail page


U-DON:
When accessing the route `/healthcare/prescriptions/:id` with a random id that doesn't exist, we should display an error message rather than try to render the usual components. It also displays current dates in the date fields rather than showing null.

webinista:
Now it says: 

> We couldn't retrieve your prescription. Please refresh this page or try again later.

That seems to resolve things, unless we specifically want to say that no such prescription exists, rather than we couldn't retrieve it.

PR #3731, opened 2016-11-04
[Rx] Fixes to pass 508 tests.

U-DON:
Fixes the 508 tests that Rx was failing.
1. Missing label on the sort menu; it didn't accept the hidden label for the sort select, so I added in an `aria-label` attribute for that.
2. It didn't register the button to open the glossary modal on the detail page as having text. Turns out the test is accessing a record that doesn't exist.
    * I put in a conditional to render nothing if the status was null, but that's a temporary fix.
    * I filed #3730 to handle the detail page for invalid ids/records.
    * I also made a `GlossaryLink` component for all the buttons that open a glossary modal to keep the behavior and usage consistent.
3. Updated test URL for detail page in 508 test.


ayaleloehr:
Redirecting to @crwallace or @bshyong for much more 508 expertise than I have!

webinista:
> it didn't accept the hidden label for the sort select

This is a case where the label and form element _should_ be hidden on mobile. Maybe a better fix is to use the select menu for sorting in all cases?

U-DON:
I understand the label was supposed to be hidden on mobile. Isn't this exactly the use case for `aria-label` (when UX calls for the label to be hidden)?

I'm not sure I understand the suggested fixes; the select menu is already supposed to appear in mobile, and the component displays as an inline list of links on desktop.

webinista:
Revising the markup or changing the design is the way to go here. ARIA should be used as a last resort. In this case, specifically, I think it will make the accessibility worse, not better.

U-DON:
What if we have both for now? As in re-introducing the `htmlFor` and `id` pairing even though the corresponding elements aren't simultaneously visible.

U-DON:
@webinista Your change to add the `sr-only` label worked and passed the test. I also added a unique id assignment for SortMenu. Can you look it over again and see if it's good to merge now?

webinista:
LGTM. I'll hold off on merging until @bshyong gives the say-so just in case we've both missed something.

bshyong:
@webinista this LGTM

PR #3746, opened 2016-11-04
Changing Rx, SM instances of Object.assign to use lodash's assign.

webinista:
Should prevent either from breaking in IE11.


src/js/rx/reducers/alert.js
```diff
@@ -9,7 +10,7 @@ const initialState = {
 export default function alert(state = initialState, action) {
   switch (action.type) {
     case 'OPEN_ALERT':
-      return Object.assign({}, state, {
+      return assign(state, {
```
U-DON: Since we're overwriting everything in the state here, I guess we don't really need `assign` at all, and can just return the object directly as `{ content: action.content, status:action.status, visible: true }`.

And for that matter, closing the alert can just return the `initialState`.

src/js/messaging/actions/folders.js
```diff
@@ -1,5 +1,6 @@
 import { api } from '../config';
 import { createUrlWithQuery } from '../utils/helpers';
+import assign from 'lodash/fp/assign';
```
U-DON: Nitpick: Can these imports go at the top so that third-party library imports go first?

U-DON (commented):
LGTM. Just some minor comments.

PR #3748, opened 2016-11-04
Disable send / save draft when message is empty.

webinista:
Closes department-of-veterans-affairs/messaging-fe#128


U-DON (commented):
LGTM with a minor comment.

PR #3750, opened 2016-11-04
quick fix for OSX naming problems

bshyong:


U-DON:
This resolves the issue for me. As long as the deleted assets are actually no longer in use, LGTM.

PR #3751, opened 2016-11-04
Excludes Drafts, Deleted, and Sent folders from MoveTo menu.

webinista:
Affects department-of-veterans-affairs/messaging-fe#143

- Also changed 'Inbox' to use string set by systemFolder


U-DON:
LGTM.

PR #3754, opened 2016-11-04
Disables 'Create' button until field is dirty.

webinista:
- ModalCreateFolder submit button is disabled by default.
- Error message still triggered onBlur, regardless of whether user has changed data.

Closes department-of-veterans-affairs/messaging-fe#139.


src/js/messaging/components/ModalCreateFolder.jsx
```diff
@@ -50,7 +50,7 @@ class ModalCreateFolder extends React.Component {
 
         <div className="va-modal-button-group">
           <button
-              disabled={error.hasError}
+              disabled={error.hasError || this.props.newFolderName.dirty === false}
```
U-DON: `!this.props.newFolderName.dirty`

PR #3755, opened 2016-11-04
Hides MessageSearch when the folder is empty.

webinista:
Closes department-of-veterans-affairs/messaging-fe#137


src/js/messaging/containers/Folder.jsx
```diff
@@ -241,6 +241,18 @@ export class Folder extends React.Component {
     const messageNav = this.makeMessageNav();
     const folderMessages = this.makeMessagesTable();
 
+    let messageSearch;
+    if (!!this.props.messages.length) {
```
U-DON: Should check truthiness of `this.props.messages` before the length. Similar to the "No messages" condition.

`if (this.props.messages && this.props.messages.length > 0)`

U-DON:
LGTM.

PR #3760, opened 2016-11-06
Fix placement of buttons on Files tab

jbalboni:
Related to [138](https://github.com/department-of-veterans-affairs/sunsets-team/issues/138)

Need to make the same change we did on the status tab for the View Details buttons on the Files tab.

screen shot 2016-11-06 at 10 45 45 am
screen shot 2016-11-06 at 10 45 38 am
screen shot 2016-11-06 at 10 45 35 am


molindoz (approved):
LGTM

PR #3762, opened 2016-11-07
Better handle focus when changing pages

jbalboni:
Related to [200](https://github.com/department-of-veterans-affairs/sunsets-team/issues/200) and [205](https://github.com/department-of-veterans-affairs/sunsets-team/issues/205)

This fixes focus handling at some other smaller 508 issues.

The logic for focus handling in here is:

- When changing pages, set focus to breadcrumbs
- If the page is a tab and the previous page was a tab, leave the focus on the current tab link.

Other changes are:

- Use aria attributes for progress bar
- Set aria attributes so that file up load label is seen as a button
- Set focus on notifications after failed file upload and on newly added files


jbalboni:
FYI @crwallace. This is related to the discussion in [200](https://github.com/department-of-veterans-affairs/sunsets-team/issues/200).

molindoz (approved):
Awesome! LGTM

PR #3763, opened 2016-11-07
Modal updates for accessibility

jbalboni:
Related to [205](https://github.com/department-of-veterans-affairs/sunsets-team/issues/205)

This adds some updates to make our modal more accessible. Mostly, this means focus handling updates. The changes largely follow [this article](https://www.smashingmagazine.com/2014/09/making-modal-windows-better-for-everyone/).

- Focus on modal when it opens, so content is presented to users
- Save the last focused element before the modal, so a user doesn't lose their spot
- Constrain focus changes to be inside the modal while it's open
- Add aria attributes

These all seemed reasonable to me, but if they're actually unhelpful changes, please let me know.


ayaleloehr:
cc @webinista and @U-DON since this is a common component.

webinista:
LGTM.

U-DON:
LGTM.

src/js/common/components/Modal.jsx
```diff
@@ -1,10 +1,38 @@
 import React from 'react';
 import classNames from 'classnames';
 
+function focusListener() {
+  const listener = event => {
+    const modal = document.querySelector('.va-modal');
+    if (!modal.contains(event.target)) {
+      event.stopPropagation();
+      modal.focus();
```
jbalboni: @U-DON If there were multiple visible, yes, but that wouldn't work with how the modal is displaying right now either. The `va-modal` class only appears to be in the DOM when the modal is visible.

U-DON: Actually, never mind. When the other modals are hidden, they're just a plain `div` without the `va-modal` class. Should be good.

PR #3764, opened 2016-11-07
Delete old versions and rename new versions of pages

akainic:
@crwallace this is one of the PRs we will need deployed this afternoon in preparation for launching education benefits. Let me know if you have any questions or concerns about this one!


cweber:
LGTM

PR #3777, opened 2016-11-07
Add notification after submitting on Ask VA page

jbalboni:
Related to [206](https://github.com/department-of-veterans-affairs/sunsets-team/issues/206)

Adds success message after leaving Ask VA page.

screen shot 2016-11-07 at 12 03 16 pm


test/disability-benefits/components/StatusPage.unit.spec.jsx
```diff
@@ -101,4 +102,59 @@ describe('<StatusPage>', () => {
...
+  it('should render notification', () => {
+    const claim = {};
+
+    const tree = SkinDeep.shallowRender(
+      <StatusPage
+          loading
+          message={{ title: 'Test', body: 'Body' }}
+          claim={claim}/>
+    );
+    expect(tree.props.message).not.to.be.null;
```
austin-martinez: This looks like its testing that the props are set by being passed in, should it be testing that a notification component is rendered with that message (I could be totally off here, not too sure how `tree.props` works)

jbalboni: Yep, nice catch. I meant to test that the message is passed to the ClaimDetailLayout component. That component renders the actual notification, so I wasn't going to test that the proper info was being passed to it.

jbalboni: Actually, this is correct. ClaimDetailLayout is the root component, so tree.props is checking the props for that component, which is what I'm intending.

molindoz (approved):
LGTM

PR #3795, opened 2016-11-07
[Rx] Send refill request with authorization headers.

### LINKED ISSUE #3774: PATCH for Rx refill sent without authorization headers
@U-DON it looks like the PATCH for refilling a prescription is being sent without the token needed for authorization.  As you can see in the first image, the /active request has the token and gets a 200 but in the second image the /refill request does not have the token and gets a 401. Is it possible the token is not being added on the call from vets-website?

screen shot 2016-11-07 at 11 22 39 am

screen shot 2016-11-07 at 11 23 00 am

U-DON:
Closes #3774.


webinista:
LGTM

PR #3796, opened 2016-11-07
Curly apostrophes on home page

goldenmeanie:


goldenmeanie:
Got this merged with something else

PR #3804, opened 2016-11-07
Fix needed files alert

jbalboni:
Related to [210](https://github.com/department-of-veterans-affairs/sunsets-team/issues/210)

This updates the alert to only show if we need files from the vet. It also changes the wording on the list page to only say items are needed, since they may not be needed from the vet.

screen shot 2016-11-07 at 3 21 39 pm


src/js/disability-benefits/containers/StatusPage.jsx
```diff
@@ -40,10 +41,13 @@ class StatusPage extends React.Component {
...
         && !claim.attributes.waiverSubmitted;
+      const showDocsNeeded = !claim.attributes.decisionLetterSent &&
+        claim.attributes.documentsNeeded &&
+        itemsNeedingAttentionFromVet(claim.attributes.eventsTimeline) > 0;
 
       content = (
         <div>
-          {claim.attributes.documentsNeeded && !claim.attributes.decisionLetterSent
+          {showDocsNeeded
             ? <NeedFilesFromYou claimId={claim.id} events={claim.attributes.eventsTimeline}/>
```
austin-martinez: Could this component just take the number of files needed as a prop/argument instead of needing to filter the timeline again? Also, is there a corresponding spec we could add or modify to test that this component is only rendered when there are items needed from the vet?

jbalboni: Yeah, it probably should do that. There's already a test for this should cover this logic, but I'll make sure it's covering the new case, too.

austin-martinez (approved):
LGTM 💸

PR #3806, opened 2016-11-07
Add aria and focus management to loading indicator

jbalboni:
Related to [202](https://github.com/department-of-veterans-affairs/sunsets-team/issues/202)

In the above issue, we got some feedback that there was no audio indication for the loading spinner.

This updates the loading indicator to have indeterminate `progressbar` aria attributes, which seemed most appropriate (based on [this](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_progressbar_role)).

I put the focus changes behind a prop so it doesn't disrupt anyone else.


jbalboni:
@bshyong and @crwallace, is this a reasonable solution? If we want to revise it later or keep these changes out of the global component, I can do that, too.

molindoz (approved):
LGTM!

src/js/common/components/LoadingIndicator.jsx
```diff
@@ -1,10 +1,35 @@
...
+            aria-valuetext={message || screenReaderMessage}
+            tabIndex="0"></div>
+        {message}
+      </div>
+    );
+  }
 }
+
+LoadingIndicator.propTypes = {
+  message: React.PropTypes.string,
```
crwallace: @jbalboni -- Can the `message` and the `screenReaderMessage` say the same thing? It is important to keep the experience consistent for sighted and non-sighted users. and the we can may the `message` `propType` required.

jbalboni: @crwallace Works for me!

PR #3810, opened 2016-11-08
Adds pagination to Facility Locator

bshyong:
- Adds pagination
- Persists zoom level and page number upon page refresh


crwallace (approved):
LGTM!

PR #3814, opened 2016-11-08
[Rx] Display loading screen when fetching data.

### LINKED ISSUE #3728: Add loading state to Rx
I couldn't find an issue in this repo about Rx loading state, so creating one here

U-DON:
Closes #3728.

## Changes
* Loading animation when navigating to Active, History, and Detail pages.
    * Prevent repeated requests for each page if a request for that page is currently already in progress.
    * Only scroll to order history on Detail after component updates and data finished loading.
    * Scroll to top of History tab content once after clicking page or selecting sort and again when the data has loaded.
* Handle errors properly when getting data fails.
    * Null out the data objects and turn off the loading indicator.
    * TODO: Differentiate between failed fetch and a 404 on detail page. We can pass the appropriate error message through `Promise.reject` in the handling of the response in the action creator.


webinista:
LGTM

webinista:
There's one test that keeps timing out. I'm not sure whether there's anything we can do to fix that since it pulls data from the API.

U-DON:
Turns out the Rx E2E test is really outdated. Not sure how it was passing before... Anyway, the mock data was missing some important fields like sort and pagination, which we expect to get, or the page just fails to load. I updated the test data and the test a little bit. It's still basic but should now match the current state of the app.

U-DON:
Actually, it wasn't even getting the mock data before since the data was associated with the wrong URL for API requests. So it kept the empty array that was the default state for `state.prescriptions.items` and loaded an empty table. I guess that's why it was passing.

PR #3817, opened 2016-11-08
Remove items without a date from claims timeline

jbalboni:
Related to [214](https://github.com/department-of-veterans-affairs/sunsets-team/issues/214)

If an event doesn't have a date, don't show it in the timeline. These are normally documents, which should show up still in the turned in documents list:

screen shot 2016-11-08 at 9 37 50 am


src/js/disability-benefits/containers/FilesPage.jsx
```diff
@@ -119,12 +119,12 @@ class FilesPage extends React.Component {
...
                     <div>
                       <h6 className="reviewed-file"><i className="fa fa-check-circle"></i>Reviewed by VA</h6>
-                      <p className="submission-date reviewed-file">{moment(getSubmittedItemDate(item)).format('MMM D, YYYY')}</p>
+                      <p className="submission-date reviewed-file">{moment(getItemDate(item)).format('MMM D, YYYY')}</p>
                     </div>
                     :
                     <div>
                       <h6>Submitted</h6>
-                      <p className="submission-date">{moment(getSubmittedItemDate(item)).format('MMM D, YYYY')}{item.status && ' (pending)'}</p>
+                      {!!getItemDate(item) && <p className="submission-date">{moment(getItemDate(item)).format('MMM D, YYYY')}{item.status && ' (pending)'}</p>}
```
austin-martinez: Is there a reason a date is required here but not for a reviewed item? Should the status still be shown if we don't have a date?

austin-martinez (approved):
LGTM, minor question about pending items

src/js/disability-benefits/utils/helpers.js
```diff
@@ -59,37 +59,51 @@ export function getUserPhase(phase) {
...
 }
 
+export function getItemDate(item) {
+  if (item.receivedDate) {
+    return item.receivedDate;
+  } else if (item.documents && item.documents.length) {
+    return item.documents[item.documents.length - 1].uploadDate;
+  } else if (item.type === 'other_documents_list' && item.uploadDate) {
+    return item.uploadDate;
+  }
```
carlmjohnson: On the backend, the date attribute is set by going through closed_date received_date opened_date requested_date and suspense_date (in that order) and returning the first one that is non-nil. I dunno if your work is a bit redundant here. Or if it is different, we should change the backend to match the behavior of the front-end because we sort by the `date` field.

src/js/disability-benefits/containers/FilesPage.jsx
jbalboni: In this case, the date might be empty because some documents don't have a date, so we're checking that first. For reviewed items, there's always date.

src/js/disability-benefits/utils/helpers.js
jbalboni: This function is mainly to grab the `uploadDate` for files, which isn't in your list. We might be able to simplify this if the `uploadDate` was used to populate `date`, though.

src/js/disability-benefits/containers/FilesPage.jsx
austin-martinez: There is also a status here that won't be shown if the date is empty, is that the intention?

jbalboni: I did leave that intentionally, but it's worth asking about.

jbalboni: Should be there, according to @suzchap

PR #3818, opened 2016-11-08
Fix condition alignment in claim tab header

jbalboni:
Related to [209](https://github.com/department-of-veterans-affairs/sunsets-team/issues/209)

Fixes an alignment issue in the conditions list. I also add some code to trim spaces, since the conditions often have extra.

screen shot 2016-11-08 at 10 04 20 am


molindoz (approved):
LGTM. I also noticed that one of the test fails for `<ClaimDetailLayout>` which is causing the build failure. (https://github.com/department-of-veterans-affairs/vets-website/blob/28e559d755e80d440218a4b4bdfd78943aa53bee/test/disability-benefits/components/ClaimDetailLayout.unit.spec.jsx#L33)

jbalboni:
@cagedcrown Thanks, fixed now!

PR #3824, opened 2016-11-08
Edu benefits 22

### LINKED ISSUE #3825: Landing page feedback from Edu Services Team
Adds 22- to form number references.
Fixes reviewed/processed language.
Calls out 'fork' as a warning.

cweber:
Closes #3825

image


akainic:
@cweber can you include a screenshot please in the PR description?

akainic (commented):
LGTM from a code perspective. Not sure how I feel about the button now being 2 lines, especially on mobile. What about just "Start 22-1990" and remove "Form"?

cweber:
Agreed that it's not ideal, but this is direct and solves the problem for now. I think I'm a bit concerned that taking out 'form' takes us just over the brink from trying to be helpful into too much shorthand.

I think we should get this up and then think about how else we can handle it. One option would be to allow for titles for the sticky box, though that cascades into needing to deal with that on mobile, etc.
e.g. 

Form 22-1990 
[Begin Application]

akainic:
Sorry, I just don't think it looks good on 2 lines, especially on mobile. Can I play with this a little bit to see if I can fit this into one line?

cweber:
Yep. Go for it. We were not happy but were willing to sacrifice that to get this up now and look at it again later.

cweber:
Oh man. No wonder it fit in my comps and not in code. 10px would do it. Yay!

PR #3838, opened 2016-11-08
Fix grouping bug in claims timeline

jbalboni:
Related to [182](https://github.com/department-of-veterans-affairs/sunsets-team/issues/182)

Fixes the grouping issue by simplifying the partitioning logic (and looping through more times).

Looks like this with claim data from staging:
screen shot 2016-11-08 at 5 19 27 pm

### DIFF
--- a/src/js/disability-benefits/utils/helpers.js
+++ b/src/js/disability-benefits/utils/helpers.js
@@ -59,40 +59,60 @@ export function getUserPhase(phase) {
   return phase - 3;
 }
 
+export function getSubmittedItemDate(item) {
+  if (item.receivedDate) {
+    return item.receivedDate;
+  } else if (item.documents && item.documents.length) {
+    return item.documents[item.documents.length - 1].uploadDate;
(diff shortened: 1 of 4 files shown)


src/js/disability-benefits/utils/helpers.js
```diff
@@ -59,40 +59,60 @@ export function getUserPhase(phase) {
...
       }
+
+      return event;
+    })
+    .filter(isEventOrPrimaryPhase);
+
+  phaseEvents.forEach(event => {
+    if (event.type.startsWith('phase')) {
+      activity.push(event);
+      phases[getUserPhase(event.phase)] = activity;
```
austin-martinez: Events that happen in phase 4-6 will be lost here when the phase3 even is parsed, as they all map to the same user phase (3). I think you want something like you had before
`(phases[getUserPhase(event.phase)] || []).concat(activity);`

austin-martinez (approved):
LGTM :ok_man:

PR #3839, opened 2016-11-08
Assign claim details to state

austin-martinez:
Fixes https://github.com/department-of-veterans-affairs/sunsets-team/issues/149#issuecomment-259262250

`_.merge` was merging the nested arrays improperly


jbalboni:
@austin-martinez I think need to change this in all our reducers. We're never really merging data, just replacing it. Most of the time it doesn't matter. You can do that here or we can make a separate item for it.

jbalboni (approved):
LGTM

PR #3852, opened 2016-11-09
Facility locator bug fixes

bshyong:
- Add mental health phone numbers
- add explanation text for access to care data
- only show access to care on Health facility detail pages


crwallace (approved):
LGTM!

PR #3853, opened 2016-11-09
[Rx] Always get userToken for API requests.

### LINKED ISSUE #3834: Rx 401's right after login; wrong token?
When I login to prescriptions on staging, prescriptions often don't load and I'm just shown the page scaffolding with no data in it.  Looking at developer tools, the API request got a 401:

screen shot 2016-11-08 at 4 25 57 pm

I then refreshed the page, and when I did that the same call succeeded:

screen shot 2016-11-08 at 4 28 33 pm

#### **The most interesting thing here is that if you look at that second image, the authorization token is different than the first one.**

@saneshark can you test this out as well and see if you get the same thing?
@U-DON is it possible that this API call is being made before the userToken is updated with the new one?

U-DON:
Closes #3834.

This fixes a problem where requests immediately following login return 401 due to using the undefined or previous `userToken` that loaded at the login page prior to logging in.

This also adds a helper to make API requests that gets the `userToken` for each request.

Will probably have to adapt a similar change to Messaging, since I anticipate it having the same issue.


webinista:
LGTM

PR #3864, opened 2016-11-09
Edu benefits review copy tweaks ast

goldenmeanie:
Found some copy edits that hadn't made it into master yet, combined with some late-breaking updates.


src/js/edu-benefits/components/education-history/EducationPeriod.jsx
```diff
@@ -64,7 +64,7 @@ export default class EducationPeriod extends React.Component {
             options={hoursTypes}
             onValueChange={(update) => {onValueChange('hoursType', update);}}/>
         <ErrorableTextInput
-            label="Degree, diploma or certificate received"
+            label="Name of degree, diploma, or certificate"
```
akainic: Whenever the labels change, we also need to change them on the review page for these questions. There should be a corresponding review fields file for each section, so check there for the labels that need to change there as well.

akainic (commented):
This looks great! For all of the label changes we'll also need to change the review labels for those questions, so that is the only thing we also need to be aware of.

src/js/edu-benefits/components/education-history/EducationPeriod.jsx
goldenmeanie: That's odd. Those should have been updated already. Checking!

PR #3875, opened 2016-11-09
Add tab aria attributes and shortcuts

jbalboni:
Related to [201](https://github.com/department-of-veterans-affairs/sunsets-team/issues/201)

This adds the proper aria attributes/roles for our tabs and also adds keyboard shortcuts, based on some feedback from the 508 session. I'm not sure how discoverable they are, but we can cover that in the next session.

I tried unit testing the shortcut code without any luck.


molindoz (approved):
LGTM

PR #3876, opened 2016-11-09
Front-end Profile and Homepage bugs

### LINKED ISSUE #3836: Header bar displays profile link and greeting when session expires
My session expired and I refreshed the page, the header bar still displayed a greeting and profile/sign out links.

screen shot 2016-11-08 at 1 44 38 pm

### LINKED ISSUE #3843: Expired userToken in the middle of a session doesn't automatically redirect to login
When your session expires, and you continue to make API requests with a stale userToken, the app will stop working, and there's no redirect the user to login. Only upon refresh does the login component render again.

I discovered this while working on Rx, but this would be true for any app that uses the userToken to make API requests. Depending on how the apps handle failed requests, it can be a confusing scenario for a user. In Rx's case, there will be a general error message telling the user to refresh or try again when data retrieval fails for any reason, so the user would probably refresh in that case. But it would be a less confusing experience to automatically redirect the user back to the login page.

### LINKED ISSUE #3874: [Homepage] Copy changes
The copy on the homepage is a little different what I have in the final design, and is likely from an older version. 

In the hero it should read: 
"Get the VA services you've earned." (No "that")

And the line below the stars should read
"Explore Services and Benefits"

cc @angel and @maryannbrody 
vets gov-homepage-11-11-updated-copy2 2

### LINKED ISSUE #3887: Pop-up stuck at "Logging you in..." after entering username/password
Please see closed issue https://github.com/department-of-veterans-affairs/vets-api/issues/460 for description and @omgitsbillryan's comment that "I was able to root cause to a XHR request race condition. If the response from /v0/user beats the response from /sessions/new?level=3 then the state.verifyUrl is read before it is ever set"

### LINKED ISSUE #3905: Profile page link label for claim status
"Check your claim status" should be "Check your disability compensation claim status"

crwallace:
This PR: 
- resolves #3836; Header bar displays profile link and greeting when session expires
- adds loading indicator while `<RequiredLoginView/>` updates to prevent login flicker
- resolves #3887; Pop-up froze after entering username/password
- resolves #3905; Profile page link label for claim status
- resolves #3874; [Homepage] Copy changes


src/js/common/components/RequiredLoginView.jsx
```diff
@@ -129,7 +129,7 @@ class RequiredLoginView extends React.Component {
           view = <SystemDownView messageLine1="Sorry, our system is temporarily down while we fix a few things. Please try again later."/>;
         } else if (this.state.profileStatus === 'NOT_FOUND') {
           // If va_profile is "not found", we cannot find you in our system and we will show a, we can't find you message.
-          view = <SystemDownView messageLine1="We couldn't find your records with that information. Please call support at 1-855-574-7286."/>;
+          view = <SystemDownView messageLine1="We couldn't find your records with that information. Please call us for support." messageLine2="The vets.gov Help Desk 855-574-7286 is open Monday through Friday, 8:00 a.m. to 8:00 p.m. (ET)"/>;
```
crwallace: @AnitraAppa -- Can I get your 👍 on this copy?

src/js/login/containers/Main.jsx
```diff
@@ -69,13 +52,29 @@ class Main extends React.Component {
...
       const myLogoutUrl = json.logout_via_get;
       const receiver = window.open(myLogoutUrl, '_blank', 'resizable=yes,top=50,left=500,width=500,height=750');
       receiver.focus();
     });
   }
 
+  checkTokenStatus() {
+    if (localStorage.userToken) {
+      if (moment() > moment(localStorage.entryTime).add(45, 'm')) {
+        if (confirm('Your login time has expired. Would you like to login again?')) {
```
crwallace: @AnitraAppa // @melwoodard  -- Does this copy work?

src/js/common/components/RequiredLoginView.jsx
AnitraAppa: Small tweaks:
"We couldn't find your records with that information. Please call the Vets.gov Help Desk at 1-855-574-7286. We're open Monday‒Friday, 8:00 a.m.‒8:00 p.m. (ET)."

src/js/login/containers/Main.jsx
AnitraAppa: @crwallace They've been fully logged out right? Is this from inactivity within a certain window or will I get logged out after a certain amount of time even if I'm actively doing things?

bshyong:
LGTM besides copy changes

austin-martinez:
I don't think this solves #3843, the token is only checked `onload`. The user could still have their session expire while making XHR requests in a React app that doesn't reload the page. I also don't think the client should be responsible for tracking the duration/expiration of a session, but should handle a server response that indicates a session is invalid.

src/js/login/containers/Main.jsx
crwallace: @AnitraAppa -- The user will get logged out after a certain amount of time even if they are actively doing things.

AnitraAppa: @crwallace -- Roger that, small tweaks then:
"Your login time has expired. **Sign back in?**"

melwoodard: I like that idea - assuming it's complex to add that functionality (or may require some approvals), we should track the session extension in a separate issue to address after tomorrow's release.

AnitraAppa: Agreed, just a nice-to-have if we can get it in at a later date.

melwoodard: Definitely - tracking in #3897

crwallace:
@austin-martinez -- You are correct and I commented on #3834 which is likely the same for disability claims.

crwallace:
@bshyong -- Can I get a re-review since there have been changes?

bshyong:
LGTM

crwallace:
@jbalboni -- Can you take a quick look at the disability e2e tests? Not sure why it is having issues seeing `a.claim-list-item:first-child`.

PR #3880, opened 2016-11-09
Removing from ignorelists

ayaleloehr:
Sending facilities locator to staging and Rx and SM to prod. 

Basic auth is already on Rx and SM in prod so when these deploy they will not be exposed publicly.


ayaleloehr:
Sorry for the off cycle request (unless the evening push hasn't gone out yet, then maybe I'm just in time!). Can this get to prod tonight so we can demo Rx and SM in the morning?

ayaleloehr:
@U-DON should have a PR up soon that would be awesome to get into the prod push as well. But if it doesn't make it in time, it's not a blocker.

ayaleloehr:
@U-DON 's PR made it in in #3884 . This should be good to merge and get to prod before 8am testing.

patrickvinograd (approved):
LGTM.

bshyong:
LGTM!

PR #3883, opened 2016-11-10
Messaging: Compose screen accessibility tweaks

webinista:
- Adds focus styles for Create New Folder button.
- Adds focus state to attachments file upload label.
- Adds focus state to Advanced Search trigger


U-DON:
LGTM.

PR #3884, opened 2016-11-10
[Rx] Change refill status to 'Submitted' after successful refill.

### LINKED ISSUE #3772: Refill submitted after prescription was refilled
After a user clicks refill prescription, the refill status should be `submitted` but right now it goes straight to `in process`. In process is a later step when the pharmacy is reviewing the submission. 

screen shot 2016-11-07 at 11 14 45 am

U-DON:
### Changes
* Refactored refills to properly handle errors.
    * Action creator propagates success and error responses accordingly.
    * Display success alert only after success response from API.
    * Show error alert in case of error response.
* Change status to 'Submitted' after a successful refill. Closes #3772.
* Not decrementing refill count after a successful refill.
    * The data returned by the API doesn't have a decremented count either, so it would be confusing to see it jump back to the original number after refreshing the page or going to the detail page for the prescription.
* Couple of fixes for warnings about props and keys.


webinista:
LGTM

PR #3918, opened 2016-11-10
Fl bugfixes

### LINKED ISSUE #3900: Select facility type drop down menu on the facility locator site doesn't reflect my selection - IE only
Site: https://dev.vets.gov/facilities
Browser: IE only - it worked in Chrome and Firefox
Expected outcome: Ability to select an individual drop down option, have it stick, and then have the option reflected on both the map and the sidebar list
Actual outcome: The "select facility type" drop down menu doesn't reflect my selections in the dropdown itself, on the map or the sidebar. I'm unable to get an individual option to stay highlighted in the drop down menu (so can't take a screenshot of it).

### LINKED ISSUE #3903: Chrome - Facility Type gets stuck as Benefits, later selections/searches don't reflect new type.
I was clicking/panning around/paging through results with all types selected. I then tried selecting Benefits facilities and filtering by type. This worked fine. Then I tried selecting Cemetery or Health facility type but the map and result list were stuck on the benefits search. 

Sentry indicates the problem is that the "services" param is sticky and is not getting cleared out when you go back to select a different facility type. It is showing InvalidFieldValue exceptions where a benefits services filter is specified but type=health or cemetery:
```
services[]:	VocationalRehabilitationAndEmploymentAssistance
type:	        health
```

### LINKED ISSUE #3906: How to display VBA facilities that are appointment only.
Billy showed me some of our data, and we came across Scott Air Force Base (below) and noticed that the excel doc's comment column info/hours are not showing for this facility.  This location is by appointment only, which is why there are no hours of operation.  We also noticed that another facility is not displaying the comment hours, so we assume this is a global issue.  

airforce

### LINKED ISSUE #3909: Update Access to Care date [launch blocker]
Can you please change the date on Access to Care section to this: 

Feb 2016 - Jul 2016

We currently have the wrong range listed.

### LINKED ISSUE #3912: Services Current As Of Date is wrong
lindadate

### LINKED ISSUE #3915: missing vba data should not sure up as blank on site [launch blocker]
If we don't have this data can we make the fields blank?

### LINKED ISSUE #3919: Adjust zoom levels between searches
The transition of search from a very zoomed in location search (address) to a very zoomed out location (state). Related to: https://github.com/department-of-veterans-affairs/kudos-team/issues/218

bshyong:
- resolves #3919 zoom levels for searches
- resolves #3903; resolves #3900 facility/service dropdowns
- resolves #3906; VBA hours
- resolves #3909; update access to care date
- resolves #3912; VHA services current date
- resolves #3915; missing vba data should not show up as blank


crwallace:
@bshyong Could we take out the note section for #3906? We will continue to further investigate the rest of this issue.

crwallace (approved):
LGTM with one note.

bshyong:
@crwallace I think these are the final two commits that will go in today—you might want to repull this branch, but otherwise I think we've gotten all of the high priority things! and also most of the low priority ones

PR #3922, opened 2016-11-10
Refactor api calls into wrapper and set unauthorized view

jbalboni:
Related to [222](https://github.com/department-of-veterans-affairs/sunsets-team/issues/222)

Adds an unauthorized message to the app. Should probably have UX look at this more closely (or figure out a global solution), but I think this is ok for now.

I'm not super excited about the pattern I have for handling the errors, but it works and minimizes some code duplication.

screen shot 2016-11-10 at 5 12 59 pm


src/js/disability-benefits/utils/helpers.js
```diff
@@ -199,3 +202,27 @@ export function isClaimComplete(claim) {
 export function itemsNeedingAttentionFromVet(events) {
   return events.filter(event => event.status === 'NEEDED' && event.type === 'still_need_from_you_list').length;
 }
+
+export function makeRequest(url, userOptions = {}) {
```
austin-martinez: Could this be `makeAuthedRequest` and include the `handleError` function you have as part of the promise chain here? I'm not entirely sure how the `dispatch` function is getting passed around, but it would be nice to not have to include that error handling in each request

src/js/disability-benefits/actions/index.js
```diff
@@ -102,29 +88,16 @@ export function submitRequest(id) {
...
           title: 'Request received',
           body: 'Thank you. We have your claim request and will make a decision.'
         }));
-      })
-      .catch(error => dispatch({ type: SET_DECISION_REQUEST_ERROR, error }));
+      }, handleError(dispatch, error => {
+        dispatch({ type: SET_DECISION_REQUEST_ERROR, error });
+      }));
   };
 }
```
austin-martinez: the document upload request should also handle a 401

src/js/disability-benefits/utils/helpers.js
jbalboni: That's what I wanted to do initially. It doesn't really work as part of the promise chain because the auth error should stop the promise from executing the success path and the normal error path. Promises only have two paths, so you would need to pass an error callback or something, which is confusing alongside promises. Plus, you need to pass `dispatch` in as well, though that's less of an issue.

It may be worth it to ditch promises here and just use callbacks.

src/js/disability-benefits/utils/helpers.js
```diff
@@ -199,3 +203,37 @@ export function isClaimComplete(claim) {
 export function itemsNeedingAttentionFromVet(events) {
   return events.filter(event => event.status === 'NEEDED' && event.type === 'still_need_from_you_list').length;
 }
+
+export function makeRequest(url, userOptions, dispatch, onSuccess, onError) {
```
austin-martinez: can this be named `makeAuthedRequest` or similar?

austin-martinez (approved):
LGTM :unicorn:

PR #3929, opened 2016-11-11
Update HCA call center number

akainic:


koconnor719:
@akainic i don't know how to read this thing..it should be one number "vets.gov Help Desk 855-574-7286 Monday through Friday, 8:00 a.m. to 8:00 p.m."  you guys can figure out the best wording for Health Care application:)  but that's the info:)  we should remove the other number.

akainic:
Yep, @koconnor719 that's what I've done, removed the old number and replaced it with the new one. The changes in green are what's going to be on the page once this is merged.

koconnor719:
awesome!  then perfecto!  ty:)

PR #3930, opened 2016-11-11
add new mapbox token

bshyong:
- adds production Mapbox token 1mm views per month


crwallace:
LGTM! cc: @ayaleloehr

PR #3932, opened 2016-11-11
Remove extra lodash and select option code from bundles

jbalboni:
Using `mixin` and lodashDeep bloats our bundles because it keeps the lodash Webpack plugin from stripping out unused functions. I've refactored that code to remove their use.

I also made some changes so that the claims status bundle isn't unnecessarily pulling in options-for-select.js, which has country and state codes in it that are not used.

Together it makes the disability-benefits bundle go from 880k to 743kb. Other apps will also see similar decreases (from just the Lodash change).


jbalboni:
@crwallace Can you take a look at this when you have a chance? The Lodash changes I made are in the login and profile reducers.

crwallace (approved):
@jbalboni -- This looks good to me

PR #3951, opened 2016-11-14
Redirect user to claims list on 404 from detail api

jbalboni:
Related to [224](https://github.com/department-of-veterans-affairs/sunsets-team/issues/224)

This will redirect users to the claims list page when a 404 is returned from the detail api. This case likely means they're using a wrong our outdated id, and we should send them to the page where they can navigate to the right place.

I was struggling to get unit tests working with the mocking required for this, so leaving them out for now.


src/js/disability-benefits/actions/index.js
```diff
@@ -68,7 +68,13 @@ export function getClaimDetail(id) {
       null,
       dispatch,
       resp => dispatch({ type: SET_CLAIM_DETAIL, claim: resp.data, meta: resp.meta }),
-      () => dispatch({ type: SET_UNAVAILABLE })
+      resp => {
+        if (resp.status !== 404 || !router) {
+          dispatch({ type: SET_UNAVAILABLE });
+        } else {
+          router.replace('your-claims');
```
austin-martinez: It looks like this could be unit tested with a sinon spy/stub as `router` is passed in to this method

austin-martinez (approved):
LGTM :racing_car:

PR #3959, opened 2016-11-14
Clarifying sent message language.

### LINKED ISSUE #158: Flash message when sending a message is confusing
It can take up to 72 hours for a message to be seen? I don't know if it is our responsibility to inform that is it? It should be more clear that we mean a physician might not see the message for 72 hours, but even then, doesn't everyone have the expectation that just because you sent an email does not mean it will be seen right away? 

This to me almost seems like we are saying "our email system is really slow, we deliver messages by carrier pigeon, and it will take 72 hours before it reaches its recipient." And anything else we might be trying communicate is basically akin to, this is email, if you've never used email before you should know, that it takes time for recipient to see the message you sent and reply.

screen shot 2016-11-07 at 4 54 05 pm

webinista:
closes department-of-veterans-affairs/messaging-fe#158

### DIFF
--- a/src/js/messaging/reducers/alert.js
+++ b/src/js/messaging/reducers/alert.js
@@ -137,9 +137,7 @@ export default function alert(state = initialState, action) {
 
       const content = (
         <b>
-          Your message has been sent. {link} It can take
-          up to 72 hours for a message to be seen and/or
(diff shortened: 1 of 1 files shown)


U-DON:
LGTM.

PR #3961, opened 2016-11-14
[Messaging] Refactor API requests into helper function.

U-DON:
### Changes
* Created `apiRequest` helper function to make requests to the API.
    * Checks `localStorage` each time, so we don't use a stale `userToken` after logging in, resulting in 401 responses and requiring a refresh.
    * Takes success and error handlers as arguments and simplifies response handling in the action creators.
* Removed no longer used preset options for `fetch` operations.
* Added `draft_id` to payload when sending a message from a draft so that the draft can be deleted once the message is successfully sent.


jbalboni:
@U-DON For what it's worth, we have a very similar api helper for disability benefits. Main difference is we pass `dispatch` and handle auth errors in it.

U-DON:
@jbalboni I did see that, and actually modeled this function off of that. We had something similar in Rx, but without the success and error handling built in. Maybe some version of these can be pulled out into common at some point.

webinista:
LGTM

PR #3968, opened 2016-11-15
Adjusting layout of compose message screen field

webinista:
Closes department-of-veterans-affairs/messaging-fe#133

- Also refactors CSS for compose message and message replies


src/sass/messaging/partials/_message-compose.scss
```diff
@@ -472,37 +406,77 @@ button.msg-btn-save {
...
+      border-color: $color-gray;
+      border-style: solid;
+      border-width: 1px 1px 0;
     }
   }
+}
 
+.msg-compose-error .usa-input-error {
+  padding-bottom: 1rem;
+  padding-right: 1rem;
```
U-DON: On the compose page, this gives the wrapper around textarea some padding as well, detaching the textarea from the send group.

U-DON: Could be wrong, but I _think_ you intended for this to give the error area some padding in the reply form; it looks like you can accomplish this and resolve the unnecessary padding in the new message form by (1) removing this block and (2) giving 
`.msg-reply-message.msg-compose-error .msg-write` a padding of 1rem instead of just `padding-left: 1rem`.

As a side note, as far as I can tell, this selector only affects the textarea-related components, or more specifically, just the `.msg-write` class. There's this confusing hierarchy of `.msg-write-group.usa-input-error` > `.msg-write` > `.usa-input-error` when that section has an error, so in turn, it can be confusing to tell what the selector is affecting.

U-DON (changes requested):
It looks like there are some padding issues when there are errors (for the message write group) and when viewing the forms in mobile view (form elements are flush against the sides and message write errors push the send group over).

webinista:
Closing in favor of another approach.

PR #3969, opened 2016-11-15
Fix address and date schema validation issues

jbalboni:
This contains fixes for the two schema validation issues Lihan brought up in Slack:

```Form The property '#/school/address' did not contain a required property of 'street' in schema 6d3ac55d-0966-5868-be2d-0f1191776ce9#```

```Form The property '#/postHighSchoolTrainings/0/dateRange' did not contain a required property of 'from' in schema 6d3ac55d-0966-5868-be2d-0f1191776ce9#, Form The property '#/postHighSchoolTrainings/1/dateRange' did not contain a required property of 'from' in schema 6d3ac55d-0966-5868-be2d-0f1191776ce9#, Form The property '#/postHighSchoolTrainings/1/dateRange' did not contain a required property of 'to' in schema 6d3ac55d-0966-5868-be2d-0f1191776ce9#```

The changes are to filter out partial addresses more thoroughly, until we get real partial address checking in the Address component, and to check for partial dates in the date range validator function. We were previously letting moment handle invalid date checking, but it treats dates without a day as valid, which causes problems we we convert them to the json application format.

### DIFF
--- a/src/js/common/components/form-elements/DateInput.jsx
+++ b/src/js/common/components/form-elements/DateInput.jsx
@@ -71,7 +71,7 @@ class DateInput extends React.Component {
     if (this.props.required) {
       isValid = validateIfDirtyDate(day, month, year, dateValidator) && (this.props.validation !== undefined ? this.props.validation : true);
     } else {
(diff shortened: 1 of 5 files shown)


akainic (approved):
LGTM!

PR #3972, opened 2016-11-15
Update claim estimation tests and add action tests

jbalboni:
Related to [157](https://github.com/department-of-veterans-affairs/sunsets-team/issues/157)

This adds some extra e2e tests to throughly test the claim estimation logic. This is already tested in unit tests.

I also added action tests for every except file upload, which I'm holding off on until we decide to drop or keep the uploader lib we're using.


molindoz (approved):
LGTM 🔥

PR #3973, opened 2016-11-15
Facility locator bugfixes and additions

### LINKED ISSUE #3860: Search button in mobile is too long
screen shot 2016-11-09 at 12 52 23 pm

The search button is too long in mobile version. 
I believe the other dropdowns are 359.25 px in width?

### LINKED ISSUE #3900: Select facility type drop down menu on the facility locator site doesn't reflect my selection - IE only

### LINKED ISSUE #3910: Text in Facility Type off
Browser: IE11

image

### LINKED ISSUE #3942: facilities search for "53130" returns result in Brazil
https://www.vets.gov/facilities/?address=53130

53130 is a valid zip code in Wisconsin, but the map centers on a location in Brazil. This is an issue in MapBox's geocoding service, so we should file an issue with them, but if there's any way to hint the geocoding to be US-centric maybe we consider that as well.

### LINKED ISSUE #3956: Facility Locator should always show minimum of 10 results
Related to issue here: https://github.com/department-of-veterans-affairs/vets-website/issues/3948

There are a couple of cases where entering a search result, e.g. a rural area or a zip code does not result in any facilities being shown. The solution we discussed is to disable searching via panning/zooming, making the only input the search box. The map will be purely for displaying where the facilities are in relation to each other.

@patrickvinograd what do we need to do to always surface a minimum of 10 results? If a bounding box call results in no facilities, could we show the 10 closest facilities to the center of the box?  

tagging @NatalieEMoore @gnakm @crwallace for awareness

bshyong:
- disables user panning/zooming 
- resolves https://github.com/department-of-veterans-affairs/vets-website/issues/3956 ensures results for searches by zooming out if there are no results in the queried locations 
- resolves https://github.com/department-of-veterans-affairs/vets-website/issues/3942
- resolves #3956 
- resolves #3900 
- resolves #3910
- resolves #3860; search button on mobilelves 

The FL is finicky and has delay in response time because we need to query both Mapbox and Vets API to get a response. However, this change will allow the map to automatically zoom out if there are no facilities in the results. It will continue to zoom until there are results that show up. User panning and zooming is also disabled.


src/js/facility-locator/components/SearchControls.jsx
```diff
@@ -132,8 +132,10 @@ class SearchControls extends Component {
           </div>
           <div className="columns medium-3">
             <label htmlFor="facilityType">Select Facility Type</label>
-            <div tabIndex="1" className={`facility-dropdown-wrapper ${facilityDropdownActive ? 'active' : ''}`} onClick={this.toggleFacilityDropdown} onBlur={() => {this.setState({ facilityDropdownActive: false });}}>
-              {this.renderSelectOptionWithIcon(currentQuery.facilityType)}
+            <div tabIndex="1" className={`facility-dropdown-wrapper ${facilityDropdownActive ? 'active' : ''}`} onClick={this.toggleFacilityDropdown} onBlur={() => {setTimeout(() => {this.setState({ facilityDropdownActive: false });}, 200);}}>
+              <div className="flex-center">
```
crwallace: I think this `setTimeout()` is causing some weirdness. When I try to open the locator the map zooms out every 2ms.

src/js/facility-locator/components/SearchControls.jsx
```diff
@@ -132,8 +132,10 @@ class SearchControls extends Component {
           </div>
           <div className="columns medium-3">
             <label htmlFor="facilityType">Select Facility Type</label>
-            <div tabIndex="1" className={`facility-dropdown-wrapper ${facilityDropdownActive ? 'active' : ''}`} onClick={this.toggleFacilityDropdown} onBlur={() => {this.setState({ facilityDropdownActive: false });}}>
-              {this.renderSelectOptionWithIcon(currentQuery.facilityType)}
+            <div tabIndex="1" className={`facility-dropdown-wrapper ${facilityDropdownActive ? 'active' : ''}`} onClick={this.toggleFacilityDropdown} onBlur={() => {setTimeout(() => {this.setState({ facilityDropdownActive: false });}, 200);}}>
```
crwallace: The selection is working in IE now, however, the user needs to click twice if they want to change their selection.

src/js/facility-locator/components/SearchControls.jsx
bshyong: Every 2ms or 2 sec? This timeout only affects the facility type dropdown - I did see the zooming sporadically but only if there are no results in the searched region..

Do you have geolocation on or is it the initial request for Washington DC?

crwallace: Sorry this issue was because I was not connected to the api locally.

crwallace:
@bshyong -- looks there is an issue with loading results after I click on a location and click “back to list”, nothing loads.

bshyong:
@crwallace issue fixed—feel free to verify/review!

crwallace:
I do have one small nit that can be fixed in the next PR. When I click "back" and then click on another facility I see the last facilities information for a quick second. Looks like it stays until the component gets updated.

PR #3975, opened 2016-11-15
Hide need files from you when claim isn't open

jbalboni:
Related to [227](https://github.com/department-of-veterans-affairs/sunsets-team/issues/227)

Adds a check for the `open` flag to the logic for the need files from you banner.


molindoz (approved):
LGTM 👍

PR #3976, opened 2016-11-15
Fix file type case issue on claims status

jbalboni:
Related to [229](https://github.com/department-of-veterans-affairs/sunsets-team/issues/229)

This was dumb of me. We should allow uppercase file extensions.

screen shot 2016-11-15 at 2 56 59 pm


molindoz (approved):
LGTM

PR #3990, opened 2016-11-16
Adjusting layout of compose message screen field

webinista:
Closes department-of-veterans-affairs/messaging-fe#133

### DIFF
--- a/src/sass/messaging/partials/_message-compose.scss
+++ b/src/sass/messaging/partials/_message-compose.scss
@@ -48,8 +48,9 @@
 .msg-subject-group {
   margin-top: 0;
   padding: 1.5rem 0;
-
+  
   @media (min-width: $medium-screen) {
+    border-left: 4px solid orange;
     display: inline-block;
     width: 95%;
   }
@@ -61,12 +62,6 @@
 }
 
 .msg-subject-group-inner {
-  @media (min-width: $medium-screen) {
-    -webkit-align-items: center;
-    align-items: center;
-    display: flex;
-  }
-
   .usa-input-error-message {
     margin-left: 0rem;
   }
@@ -111,14 +106,6 @@
   }
 }
 
-.msg-recipient > div {
-  -webkit-align-items: center;
-  align-items: center;
-  display: -webkit-flex;
-  display: flex;
-  margin-bottom: 0;
-}
-
 .msg-recipient > div,
 .msg-subject-group {
   padding: 1.5rem;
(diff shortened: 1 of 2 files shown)


src/sass/messaging/partials/_message-compose.scss
```diff
@@ -48,8 +48,9 @@
 .msg-subject-group {
   margin-top: 0;
   padding: 1.5rem 0;
-
+
```
U-DON: Trailing space.

src/sass/messaging/partials/_message-compose.scss
```diff
@@ -48,8 +48,9 @@
 .msg-subject-group {
   margin-top: 0;
   padding: 1.5rem 0;
-
+  
   @media (min-width: $medium-screen) {
+    border-left: 4px solid orange;
```
U-DON: Are the fields' left borders supposed to have color by default (when there's no error)? If so, this seems to remain orange even when I pick 'Other' for the category and leave the subject empty.

src/sass/messaging/partials/_message-compose.scss
```diff
@@ -272,8 +253,15 @@ button.msg-btn-save {
...
+.msg-compose-write {
+  @media (min-width: $medium-screen) {
+    border-left-color: green;
+    padding-left: 1.5rem;
+  }
+}
+
 .msg-write {
-  padding: 1.5rem 1.5rem 2rem 1.5rem;
+  padding: 1.5rem 1rem 2rem;
```
U-DON: With this, the message field doesn't line up with the other fields in mobile. It's slightly left (.5rem) of them. If this should have 1rem padding on the sides, the other fields (currently at 1.5rem for padding on sides) should as well.

U-DON (changes requested):
Should also make `.msg-write-group textarea` have `margin-bottom: 0` instead of `margin: 1rem 0 0 0` in order to match the other fields with the default margin between its label and itself.

Overall, looks good. I'm not sure if you intended to keep the colored borders, or if they were there for testing. It's okay if they were called for in the design. The only issue there is that the orange border on the subject group stays orange instead of turning red when there's an error.

U-DON:
Another minor observation: `#msg-reply .msg-compose-error .msg-write` currently has `padding: 1.5rem`, but it looks a little too much. If you can add it, I'd like to see it have `padding: .25rem 1rem 1rem` to reduce the excess top padding and to line up the sides with the text in the reply details.

U-DON (approved):

U-DON:
LGTM. Noticed the error state for the message write part of the reply form on mobile (super specific...) doesn't quite have the right padding, but that can be fixed later.

PR #3994, opened 2016-11-16
Display never received docs on claim status files tab

jbalboni:
Related to [228](https://github.com/department-of-veterans-affairs/sunsets-team/issues/228) and [227](https://github.com/department-of-veterans-affairs/sunsets-team/issues/227)

We're showing docs in the `never_received_from_x_list` in the documents filed section as pending docs and we should describe them properly instead.

screen shot 2016-11-17 at 12 28 17 pm


molindoz (approved):
LGTM

PR #3996, opened 2016-11-16
Adding EST to message detail timestamp.

webinista:
- Closes department-of-veterans-affairs/messaging-fe#145


U-DON:
I think the other place to add this for now is in `utils/helpers.js` in the `formattedDate` function so that it can show EST wherever we use the time.

src/js/messaging/utils/helpers.js
```diff
@@ -85,7 +85,7 @@ export function formattedDate(date, options = {}) {
   let dateString;
 
   if (momentDate.isSame(now, 'd')) {
-    dateString = momentDate.format('HH:mm');
+    dateString = `${momentDate.format('HH:mm')} <abbr title="Eastern Standard Time">EST</abbr>`;
```
U-DON: I think this literally prints out the full string un-marked-up. To return it as markup, it should be:

`<span>{momentDate.format('HH:mm')} <abbr title="Eastern Standard Time">EST</abbr></span>`

And also import React in this module.

webinista: Ok, I'll undo this and let you handle that. I am still using a mock API and have no way to test this on my end.

U-DON: Sure, will do.

PR #3998, opened 2016-11-16
Switch from merge to assign in claim status reducers

jbalboni:
Related to [215](https://github.com/department-of-veterans-affairs/sunsets-team/issues/215)

We should use `assign` instead of `merge` to avoid inadvertent data merging.


molindoz (approved):
LGTM 👍

PR #3999, opened 2016-11-16
[Messaging] Loading state when loading folders and threads.

U-DON:
## Changes
* Show loading indicator when opening a folder or thread. Closes department-of-veterans-affairs/messaging-fe#73.
* Refactored redirects so that they don't occur within the reducer.
* Removed `history` module since its usage is isolated to `messaging-entry` now.


webinista:
Lgtm

PR #4003, opened 2016-11-16
Adjusting positioning of track package link

### LINKED ISSUE #3832: Refill Prescription button is too short on small < 481px screens
screen shot 2016-11-08 at 1 03 48 pm

webinista:
- Closes department-of-veterans-affairs/messaging-fe#172
- Closes #3832

- Prevents track package link from wrapping.
- Prevents status from wrapping.
- Refactored markup to provide more consistent layout across viewports.


U-DON:
LGTM. One minor thing I would add is `line-height: 1` to `rx-call-provider` so that it can line up with the "x refills left" text like `rx-prescription-status` does when it's the last row in the section.

PR #4004, opened 2016-11-16
Hide pages turned off by financial disclosure choice in subway map

jbalboni:
Related [498](https://github.com/department-of-veterans-affairs/healthcare-application-team/issues/498)

Updates HCA to use the common subway map component so that pages are hidden conditionally.

screen shot 2016-11-16 at 4 41 53 pm


akainic (commented):
Looks good! Only thing I see is that the spouse information section is not appearing in the nav under the household-information chapter. It should appear when they indicate that they are married earlier in the form. Perhaps we need to add additional conditional logic to the helper function to account for that.

jbalboni:
@akainic Ack, good catch. Was just looking for the group of them showing up. Should be fixed now.

akainic (approved):
Looks great! 👍

PR #4005, opened 2016-11-16
Remove claim status from ignore list

austin-martinez:
Related to devops changes here https://github.com/department-of-veterans-affairs/devops/pull/1107


jbalboni (approved):
LGTM

crwallace:
@austin-martinez -- Can we just comment out all of the lines instead of deleting them because other teams like content use the ignore list as well.

PR #4009, opened 2016-11-17
Update 10-10cg form

alexose:
Per [this issue](https://github.com/department-of-veterans-affairs/healthcare-application-team/issues/486), we are updating the form to the newest version.


akainic:
Tested this locally and the form still says its from Feb 2016, I think the one we're looking for is from April. Work with @leannammiller to make sure we're using the right version of the form. Thanks!

alexose:
I can confirm that my local branch is serving the right PDF:

image

My guess is that for some strange reason, `npm run watch` doesn't copy pdf files over correctly.  This shouldn't be an issue in production since we build from scratch for each deploy.  Anyway-- try running `npm run build` then restart `npm run watch`?

jbalboni:
@alexose It get the version with the correct date if I restart my watch task. However, I can only open it with Acrobat, not with the Chrome viewer. Is that expected still?

alexose:
@jbalboni [Yes](https://github.com/department-of-veterans-affairs/healthcare-application-team/issues/486#issuecomment-259037824), unfortunately :(

jbalboni (approved):
LGTM.

alexose:
@akainic Were you able to see the correct form?

akainic:
Hm I'm still seeing the Feb version :/ even after restarting the server and running npm run build and pulling the latest. Not sure what I'm doing wrong. But if @jbalboni saw the right version and you do too I trust this is fixed!

alexose:
Maybe Acrobat is caching it or something silly like that

WMACI (will merge after CI)

PR #4010, opened 2016-11-17
remove 10-10EZ-DEMOB.pdf and associated links

alexose:
Remove link to 10-10EZ-DEMOB.pdf as required per [this](https://github.com/department-of-veterans-affairs/healthcare-application-team/issues/506).


content/pages/healthcare/apply.md
```diff
@@ -66,7 +66,6 @@ To update your personal, insurance, or financial information from a previous app
 ### Additional forms for caregivers, National Guard, and Reserve Servicemembers
```
akainic: Since we're removing the form for National Guard and Reserve, we should probably change this header. Also the banner at the top of the page.

alexose: Good call, will fix

cweber: Are you talking about these lines?

I haven't been following this directly, though I know a link is being removed. Why aren't we replacing it with a different way for National Guard to access an application?

 - ARE YOU A CAREGIVER, NATIONAL GUARD, OR RESERVE SERVICEMEMBER? THIS FORM’S NOT QUITE READY FOR YOU YET. FIND THE FORM YOU NEED.

 - Additional forms for caregivers, National Guard, and Reserve Servicemembers

akainic: Because there doesn't exist a different way for National Guard to apply, unfortunately. It's something we're trying to figure out, but the current application is super broken and was never officially approved. So we should just remove reference to them in those two bullets and just include something about caregivers.

cweber: Ok. My only question is whether we still need to acknowledge National Guard so that they don't mistakenly try to apply through our form.

leannammiller: @cweber as per @akainic - let's remove reference to national guard for now and get this one finished up. thanks.

cweber: Sure. If you want removal without replacing how we're addressing National Guard needs, then let's not overthink the copy. @alexose I think you can just straight up remove the National Guard references and relevant commas.

ARE YOU A CAREGIVER ~, NATIONAL GUARD,~ OR RESERVE SERVICEMEMBER? THIS FORM’S NOT QUITE READY FOR YOU YET. FIND THE FORM YOU NEED.

Additional forms for caregivers ~, National Guard,~ and Reserve Servicemembers

alexose: Changes made:

image


image

cweber:
Argh. So I'm realizing that we were referring to this as a 'National Guard' problem, but perhaps we actually meant National Guard & Reserves? I note that there is also no link for Reserve members.

@akainic You may be a better resource to get this sentence worked out. Does it make sense to remove Reserve Servicemember too? In which case those two lines could read.

ARE YOU A CAREGIVER? THIS FORM’S NOT QUITE READY FOR YOU YET. FIND THE FORM YOU NEED.

and

Alternate form for caregivers

akainic:
@cweber Yes you're right, it's for both the National Guard and the Reserves. I think that's exactly the language change we need to make. I can take care of that.

alexose:
The above three commits look good to me.  Makes sense in context

PR #4022, opened 2016-11-17
Changing thickness of va-h-ruled to 4px for sitewide consistency.

### LINKED ISSUE #4013: Tweak line stroke
In the facility name filter and glossary page the border-bottom is too thick. They should match same height was border-bottom for contact in detail page.

webinista:
- Closes #4013


U-DON:
Should the `rx-heading` class be removed or modified along with this change? And consequently, should we add `va-h-ruled` to the headers that are currently using `rx-heading`?

U-DON:
Since `.rx-heading` was removed, the `va-h-ruled` class should replace it in ContactCard and PrescriptionGroup.

U-DON:
The spacing between the prescription group header and the prescriptions is now a little too narrow, but that can be addressed separately.

PR #4025, opened 2016-11-17
Fix mobile scrolling bug on edu benefits

### LINKED ISSUE #3765: Add scrolling for education benefits
Without it, it's hard to use on mobile

jbalboni:
Closes #3765.

The scroll to element was inside the div for the subway nav that's hidden on mobile. Now it's not!


akainic (approved):
LGTM!

PR #4032, opened 2016-11-18
HCA: Shorten max length on street address

markolson:
The spouse's street address can only be 30 characters, and since we allow an applicant to indicate that they live with their spouse, it was possible to craft an invalid entry by setting the vet's street address to > 30 chars.


akainic (approved):
LGTM! Do we need to change the schema as well?

PR #4043, opened 2016-11-18
Correcting onFolderChange action for FolderNav

webinista:
Closes department-of-veterans-affairs/messaging-fe#177

- Removed handleFolderChange method.


U-DON:
Doesn't this render obsolete the `setCurrentFolder` action creator and all the handling around `persistFolder`? I was thinking of ways to refactor that, but we should probably keep that in if we want to continue using it to persist the folder selection in the menu (I think that's what it was for, if I recall correctly...). We could keep that and close the menu by adding a call to `toggleFolderNav` in the `handleFolderChange` method.

webinista:
I don't think it does, because the function attached to the `onFolderChange` handler was `this.handleOnFolderChange`, not `this.handleFolderChange`. But `this.handleOnFolderChange`  doesn't exist.

U-DON:
True, I think I noticed that at some point too, but forgot to address it. Although this wasn't the change that broke it, not using `setCurrentFolder` defeats the purpose of using `persistFolder`, which was to make it so that it didn't look like two folders were selected at once. But the other thing is that we probably don't need it anymore, and can determine it with a combination of the route param and current item id. Anyway, will look into that refactor separately.

In short, LGTM.

U-DON:
Actually, the one change I would make is `toggleFolderNav` only if it's already opened, since on desktop, it would change the state and cause a update cycle in Main every time the user clicks on a different folder. I don't think it would really impact performance or anything, but it seems unnecessary in that case.

webinista:
Resolved a merge conflict between this and `master` and now this fix is not the right one.

PR #4045, opened 2016-11-19
Gov should be lowercase

ayaleloehr:
closes https://github.com/department-of-veterans-affairs/vets.gov-team/issues/439

@emilyville this is the only instance of a capital G that I could find on the site.


emilyville:
LGTM

PR #4050, opened 2016-11-19
Make DOL links HTTPS

ayaleloehr:
On the VA network, DOL links weren't redirecting from http to https for some reason so the DOL links looked broken. Making both links https should solve this. Also added a trailing slash to the second link for consistency and since that's what DOL redirects to anyways. 

Assigning to @crwallace but also open to reviews from anyone else given how small this PR is!


crwallace (approved):
LGTM!

Issue #3741, opened 2016-11-04
Add fork modal to /apply-for-education-benefits-new/


cweber:
This modal will be in place to help direct people to the correct form. We expect our form to handle about 50% of all Education Benefits application cases, but the other 50% need to know how to find their way to VONAPP(as an iframe in eBenefits) for the other forms.

The base page for this has additional language to reinforce the single purpose of our form. That's addressed in #3740.

(@staceylanger, this is the solution we arrived at.)

The modal should sit on top of /apply-for-education-benefits-new/ and should take advantage of existing modal styles elsewhere on the site.

# Which form do you need to use?

## Servicemembers and Veterans applying for education benefits

[Apply with Form 22-1990](url)

---

If you are:

- A student who wants to **change programs or schools**
- A spouse or dependent
- A survivor of a Veteran
- A candidate for **National Call to Service**

[Apply on eBenefits](url)

![edu-fork-modal](https://cloud.githubusercontent.com/assets/417657/20015367/97da83de-a291-11e6-9dbb-c301c40a168d.png)

cc @goldenmeanie

cweber:
@goldenmeanie, I just realized we probably need to mock up what this looks like in mobile. Stacked?

goldenmeanie:
![apply-for-benefits-mobile-modal-or-left](https://cloud.githubusercontent.com/assets/1969/20064733/b610f960-a4d9-11e6-83dc-a0eb74af5b56.png)

staceylanger:
Looks good.

alexose:
Quick call with @goldenmeanie revealed a few items:

* We should implement a non-react modal from scratch using the existing styles
* Future modals may need a close box, but this one won't
* The modal should open when there is a url param (e.g., `?modal` or maybe `#modal`)

alexose:
Work started in [add-modal#3741](https://github.com/department-of-veterans-affairs/vets-website/compare/add-modal%233741?expand=1).

The modal + content is scaffolded out but still needs styles.

![image](https://cloud.githubusercontent.com/assets/627264/20412883/7c227294-acf7-11e6-8ddb-039541910d0d.png)

alexose:
Desktop styles:

![image](https://cloud.githubusercontent.com/assets/627264/20452561/af7ddcf8-adda-11e6-9419-e62456c5ea7f.png)

There are minor differences with the mock, but it is in line with the other modals across the site.  We can definitely make tweaks.

[...]

PR #4054, opened 2016-11-21
Add modal#3741

alexose:
Adds the forked modal design as detailed in issue #3741.

The modal will appear if you visit /education/apply-for-education-benfits/#modal

This only includes the desktop styles-- mobile is next.


content/includes/modal.html
```diff
@@ -0,0 +1,64 @@
...
+        <div class="va-modal-split-right">
+          <a href="https://www.ebenefits.va.gov/" class="usa-button-primary usa-button-big usa-button-outline">Apply on eBenefits <span class="exit-icon">&nbsp;</span></a> 
+        </div>
+      </div>
+    </div>
+  </div>
+</section>
+
+<script>
+(function($){
```
jbalboni: In the React Modal there's some focus handling that would probably be useful here. I added it to help screenreaders navigate easier.

src/sass/modules/_m-modal.scss
```diff
@@ -11,6 +11,41 @@
...
+    }
+    &-container {
+      display: flex;
+    }
+    &-left,
+    &-right {
+      flex: 1;
+      padding: 0 20px;
+    }
+    p:first-child,
```
jbalboni: Could we style a class instead of `p`, `h5`, and `ul` elements? Or use `>` to make the rules narrower, if that makes more sense? 

I keep getting tripped up by this in other places, where someone has styled elements for a different context and I have to overwrite those styles with `!important` or more specific ones.

jbalboni (approved):
Desktop styles look good, LGTM.

goldenmeanie:
Source order will matter for small screen layout.

apply-for-benefits-mobile-modal-or

src/sass/modules/_m-modal.scss
alexose: I pulled this from the react modal, but agreed.  Specificity is a good idea as the modal takes on more roles...

alexose:
Mobile styles complete:

image

image

alexose:
@jbalboni @goldenmeanie The mobile styles are ready for review-- This issue should be ready to close pending your approval :)

goldenmeanie:
This is looking great!
Two things:
1. I think the modal should have less top margin on small screens.
2. I think you should have to click one of the two buttons, clicking outside the modal to dismiss feels wrong in this context (this might be largely fixed by no. 1)

PR #4055, opened 2016-11-21
Resolves profile page tweaks

### LINKED ISSUE #4047: Change how we refer to users in greeting
Currently when a user is logged into a verified account, we know their name, but we still say "hello email@address" in the header. Should we use the user's name when we know it instead to be more personal?  

Probably a question first for @emilyville but going to tag this with Vanguard as well so we can track with the other identity issues. 

screen shot 2016-11-18 at 8 20 12 pm

### LINKED ISSUE #4048: Users middle name shown as null when it doesn't exist
Right now vets.gov.user+264@gmail.com has a middle name of null as you can see below. This user does not have a middle name in ID.me or MVI.  I first created this issue in vets-api [here](https://github.com/department-of-veterans-affairs/vets-api/issues/523) but @omgitsbillryan thought this should be taken care of on the front end instead, so opening this issue now. 

1bb7b30a-ab1d-11e6-82c3-25aabbcf951f

crwallace:
resolves department-of-veterans-affairs/vets.gov-team#441
resolves #4048 
resolves #4047


melwoodard:
@crwallace can you address [this comment](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/441#issuecomment-261997076) from @AnitraAppa in this PR:

> For (6), let's change **"Sex"** to **"Birth Sex"** (as per updated VA language guidance on birth sex and gender identity).

src/js/user-profile/components/UserDataSection.jsx
```diff
@@ -8,28 +8,24 @@ import moment from 'moment';
 class UserDataSection extends React.Component {
   render() {
     let content;
-    const name = `${this.props.name.first} ${this.props.name.middle} ${this.props.name.last}`;
+    const name = `${this.props.name.first ? this.props.name.first : ''} ${this.props.name.middle ? this.props.name.middle : ''} ${this.props.name.last ? this.props.name.last : ''}`;
```
bshyong: @crwallace you can consider using `compact` here to avoid all the ternary statements. I had to deal with null/falsey values for facility addresses too.

something like this  `compact([first, middle, last]).join(' ')`

bshyong (approved):
LGTM

PR #4059, opened 2016-11-21
Fix date range conversion to avoid empty objects on edu form

### LINKED ISSUE #4058: Education benefits dateRange schema error
We received this schema error:

`Form The property '#/postHighSchoolTrainings/0/dateRange' did not contain a required property of 'from' in schema 6d3ac55d-0966-5868-be2d-0f1191776ce9#, Form The property '#/postHighSchoolTrainings/0/dateRange' did not contain a required property of 'to' in schema 6d3ac55d-0966-5868-be2d-0f1191776ce9#`

This means we sent a dateRange object without `to` and `from` fields, which is an error. We should update our date range conversion to avoid this.

jbalboni:
Closes #4058

I've updated the dateRange case to do the date check and conversion, rather than assume it's ok as long as there's a month and day filled out for both fields.


test/edu-benefits/utils/veteran.unit.spec.js
```diff
@@ -41,6 +41,49 @@ describe('veteranToApplication', () => {
...
+          year: makeField('1997')
+        }
+      }
+    });
+    const applicationData = JSON.parse(veteranToApplication(formData));
+
+    expect(applicationData.toursOfDuty[0].dateRange.from).to.equal('1996-09-06');
+    expect(applicationData.toursOfDuty[0].dateRange.to).to.equal('1997-09-06');
+  });
+  it('removes partial date ranges', () => {
```
akainic: The only thing I'm a little worried about is removing partial dates just for the sake of passing our schema validations. From an education services perspective, partial date information is still more helpful than no date information, right? Is there a reason why we can't submit partial date information?

akainic (commented):
LGTM from a code perspective, one question about general approach to the problem at hand.

test/edu-benefits/utils/veteran.unit.spec.js
alexose: I struggled with this a bit when I was working on the graduation date ticket (where the user only needs to enter a month and a year).

In order for something to be a `date`, it technically needs precision down to the millisecond.  If the input isn't precise, Javascript assumes "the first possible moment that this could be":

```
new Date('December 2014')
Mon Dec 01 2014 00:00:00 GMT-1000 (HST)
``` 

This leads to problems when trying to check the difference between dates, since the user didn't necessarily mean "the first of the month".  The right thing to do here is probably to create a new type of object for partial dates, or to rework the way our validations happen.

markolson: @jbalboni it looks like our template treats these dates as optional, so nothing should blow up if we allow partial dates. I'll gin up a quick test to verify today, though.

jbalboni: I was thinking just about ranges that are missing the to or from date, but partial dates are also something we should consider.

akainic: @jbalboni I'm good with that. I will create that other issue. Feel free to merge this once tests pass.

PR #4066, opened 2016-11-21
Adds MFA to every verified login

crwallace:
resolves department-of-veterans-affairs/vets.gov-team#433


src/js/auth/containers/AuthApp.jsx
```diff
@@ -40,9 +39,10 @@ class AuthApp extends React.Component {
       const userData = json.data.attributes.profile;
       if (userData.loa.highest === 3) {
         // This will require a user to MFA if they have not verified in the last 2 mins.
```
ayaleloehr: This comment above can be removed now

ayaleloehr (commented):
LGTM, with one nit, although I'm not totally sure I understand this entirely to be honest; I can't wrap by head around why we need to set the logoutUrl inside the setMyToken method.

crwallace:
Thanks @ayaleloehr -- We need to set it in two places because one function is triggered on login when the child window is closed and one is triggered only `onload`. So if a user logs in and immediately needs to logout we need to request the logout url as part of the `/users` request.

PR #4072, opened 2016-11-22
Changing flexbox values for simple search.

webinista:
Possible fix for department-of-veterans-affairs/messaging-fe#157


U-DON:
LGTM.

Issue #4067, opened 2016-11-21
Show course of study in Education review section


cweber:
This is currently being omitted from the review screen.

leannammiller:
@cweber who is working on this?

cweber:
It doesn't look like anyone is assigned to it. @jbalboni may have picked it up in his work on the review section though.

jbalboni:
This is done now.

Issue #3267, opened 2016-10-04
Update review page for better UX


cweber:
The current HCA implementation of the review screen has caused confusion in all of our user testing. 

This redesign addresses these existing issues:
- Not being able to see all of the information easily
- Not understanding the mechanics of the next button
- Users intuitively trying to select the title of each section to expand it
- Problems with inconsistent pattern for how the existing buttons function, dependent on what you have or have not already looked at or clicked on.
- Clustering by page creates too many sections to click through
- Section titles don't look like titles

This is one basic example of the redesign, additional views (expanded, editable, etc.) can be found [here](https://github.com/department-of-veterans-affairs/edu-benefits-team/blob/master/design-documents/edu-bennies-review.pdf).

![review-first-view](https://cloud.githubusercontent.com/assets/417657/19092384/5e1bad44-8a54-11e6-9e5c-c6880204c744.png)

leannammiller:
@cweber - who is responsible for this?

akainic:
If @jbalboni @alexose or myself have time to take this before launch we will, but not critical for launch

cweber:
@jbalboni Here's how we'll deal with Education/Employment sections.

1. cluster stray questions at top with no section title and one edit button. (It breaks some of our rules, but I think it follows the spirit of them.)
2. For employment, if they change the first question to 'no' we need to hide their other answer. (We're currently not handling this appropriately.)

Also, see issues #4067 and #4065. Found those two while I was working this out.

Let me know if this doesn't make sense.

![review-basic-edit-education](https://cloud.githubusercontent.com/assets/417657/20500986/caff4a5c-b005-11e6-8aa9-3d5b83fc006a.png)

PR #4082, opened 2016-11-22
Update edu benefits review page design

jbalboni:
Related to #3267 and #4067.

This involved changing a lot of code. The main challenge was that for the pages with growable tables the design doesn't have clearly separated review and edit states. We want to edit the growable table rows in place. This meant I needed to ditch the separate review/fields components used for most pages and handle both states in one component.

The high level changes are:

- Changed collapsible panel component to take a list of pages and render those inside the accordion
- Updated collapsible panel to pass review/edit states and hooks to the component its rendering if a review component was not passed. I'm doing this so that we can still use the standard Review and Edit layouts we had before, but do more custom ones for the three pages that need it.
- Added a bunch of options to GrowableTable to support the above.

screen shot 2016-11-22 at 1 54 11 pm
screen shot 2016-11-22 at 1 54 17 pm
screen shot 2016-11-22 at 1 55 28 pm
screen shot 2016-11-22 at 1 55 57 pm
screen shot 2016-11-22 at 1 56 13 pm
screen shot 2016-11-22 at 1 56 19 pm


cweber:
Yay! That feel so much better to use.

To add:
- Scroll to top of section on expand or after updating

To fix:
- updating a card is causing a seemingly unpredictable scroll pattern. I can't tell what it's trying to scroll to, but it's odd.

goldenmeanie:
apply_for_education_benefits__vets_gov

Clicking add another for a growable table with children scrolls half-way into form. Probably same/similar issue @cweber mentions for edit.

src/js/edu-benefits/components/education-history/EducationHistoryFields.jsx
```diff
@@ -38,33 +65,44 @@ export default class EducationHistoryFields extends React.Component {
...
+        {this.props.editing
+          ? formView
+          : <EducationHistoryReview data={this.props.data} onEdit={this.props.onEdit}/>}
+      </div>
+      <div className="form-review-panel-page">
+        <div className="form-review-panel-page-header-row edu-growable-review-header">
+          <h5 className="form-review-panel-page-header">Education after high school</h5>
+          <button
+              className="edit-btn primary-outline"
+              onClick={this.addAnotherPeriod}><i className="fa before-text fa-pencil"></i>Add Another</button>
```
akainic: The addAnother button appears in review mode?

src/js/edu-benefits/reducers/uiState/index.js
```diff
@@ -21,83 +20,67 @@ const ui = {
   },
   pages: {
     '/introduction': {
-      complete: false,
-      verified: false,
+      reviewEdit: false,
```
akainic: The name of this property is a little confusing to me. To me on the review page there are 2 states: `review` or `edit`. I think I know what you're trying to get at here, is the state in `edit` on the review page? Maybe call this `editOnReview` on something like that instead?

src/js/common/components/form-elements/GrowableTable.jsx
```diff
@@ -212,12 +248,22 @@ GrowableTable.propTypes = {
   isValidSection: React.PropTypes.func.isRequired,
   addNewMessage: React.PropTypes.string,
   minimumRows: React.PropTypes.number,
-  rowTitle: React.PropTypes.string
+  rowTitle: React.PropTypes.string,
```
akainic: Pretty sure HCA now uses the common `GrowableTable` component, so these changes may also affect HCA. Need to make sure everything works there as well.

akainic (commented):
This is huge! Changes look good overall. A little hard to review everything. In general I think some comments in the code to explain some of the new props and how they're used could be really helpful. Because this is adding so much more complexity I want to make sure it's understandable to future users of these components.

src/js/edu-benefits/components/education-history/EducationHistoryFields.jsx
jbalboni: Yeah, if you're on the review page it always appears.

src/js/edu-benefits/reducers/uiState/index.js
jbalboni: `editOnReview` makes sense to me. I'll use that.

src/js/common/components/form-elements/GrowableTable.jsx
jbalboni: Just checked and it seems fine. I made a minor tweak to the button style and fixed a validation issue on the review page that I noticed, though.

PR #4087, opened 2016-11-22
Updates to run certain entry files and use specific host name

jbalboni:
This has two changes:

- The first is the ability to pass a host name to the build script. The use case for this is to set the host name to `0.0.0.0`, so that the app is visible on your local network and you can pull it up on a phone. We could do this by default, but that seemed slightly risky.
- The second is the ability to pass a list of entry files you want to build. Our Webpack build is slow and most people are not looking at all the apps, they're working on one or two. I think this is especially noticeable for content people, who might just want to build `no-react`, and work on the static pages.


jbalboni:
@ayaleloehr @jkassemi Thoughts on these changes? Not sure who else to pull in.

ayaleloehr:
@alexose helped a bunch with setting up these things on Rx when we were getting going, so tagging him to see if he has thoughts!

alexose:
Seems like a good, developer-friendly set of changes.  I would add in a line or two to README.md with a few examples.

LGTM though!

ayaleloehr:
Agree with @alexose ; Please add a row in the [README table](https://github.com/department-of-veterans-affairs/vets-website#i-want-to) documenting this, then feel free to merge @jbalboni .

PR #4090, opened 2016-11-22
Resolves #3840; login and logout messaging

### LINKED ISSUE #3840: User messages for login and logout
@melwoodard // @AnitraAppa -- Can we get your once over on these popup screens?

Would love to know what content we should show the user.

I user will see this quickly when they are logging in before we check to see if they are verified and right before we fully log them in and close out the login screen:
screen shot 2016-11-08 at 5 49 06 pm

This is what the user sees quickly before they are fully logged out. This one is placeholder text as an FYI: 
screen shot 2016-11-08 at 5 48 53 pm

bshyong:
Note: because of the way our pages are structured and complied using markdown templates, I used a CSS overlay to eliminate the header, nav and footer on the logging in state. This retains the header, nav, and footer on the other states.


crwallace:
LGTM! Thanks @bshyong

PR #4091, opened 2016-11-23
[Messaging] Refactored update conditions for Folder and Thread to better handle errors.

U-DON:
Cache the last request for folders and messages to compare when checking whether to fetch the requested folder or message on component updates.

When fetch failed before, we had default data, which had a falsy id that didn't match the id of the request, so it would infinitely loop back to fetch again. Now, we just stop when we see that the request is the same as before. Instead of retrying, we display an error message that allows the user to reload the failed folder or message.


webinista:
LGTM

Issue #4095, opened 2016-11-23
Lazy load react for Login component


alexose:
The Login component is written in React, which means we're now pulling the library into our non-React pages.

I suspect this is why we're seeing longer build times for changes to `/content`.  Also, React is somewhere around ~150k minified, so it adds some heft to each page load.

It would be nicer to lazy load the necessary javascript when the user clicks the Login button.  @jbalboni and I have some thoughts on how to do this.

jbalboni:
There are two ways of doing this, I think: loading the login code when the user clicks something and lazily loading it on page load, but separate from the main bundle. The login code looks pretty small outside of the React dependency and it might be awkward to write just the sign in link without React. So, the latter may be the better option. It should at least allow the rest of the page to load without having to parse and run React.

You can do this with Webpack pretty easily:

```
require.ensure([], (require) => {
  require('./login/login-entry.jsx');
});
```

This will cause Webpack to generate a separate bundle that contains the code for login-entry.jsx. There's also some tweaking that will need to happen in init-react.js to make sure we still mount React components even if `DOMContentLoaded` has already fired. And possibly a change to styles, since there's a login.scss and I'm not sure how Webpack handles that.

It would also help to remove the use of jQuery from the login code.

jcmeloni:
Moved issue to vets.gov-team.

PR #4096, opened 2016-11-23
Fix sticky action box

alexose:
The javascript that keeps the box "sticky" wasn't being loaded properly for a couple of reasons:

* `no-react-entry.js` is [no longer](https://github.com/department-of-veterans-affairs/vets-website/issues/4095) being used
* `window.onload` was being overridden by `login/containers/Main.jsx`

This PR adds the javascript to the React bundle and switches the `window.onload` function to an event listener.


jbalboni (approved):
LGTM

PR #4101, opened 2016-11-23
Edu benefits add kicker language

alexose:
This adds some conditional language to the `military-history/contributions` section in the form of `ExpandingGroup`s.


cweber:
LGTM. 

Confirmed that the messages are appearing correctly based on active or reserve selected earlier.

akainic (approved):
LGTM! One question for @cweber: I'm coming into this conversation about the kicker language a little fresh, and reading it I don't understand what it's telling me. What does it mean that I can only transfer the kicker from a relinquished benefit? Transfer to what? And how does that relate to why I won't be receiving the Active Duty kicker?

cweber:
@akainic Kickers are associated with a particular benefits. So if you have a kicker attached to Montgomery GI Bill AD (Active Duty), but you are relinquishing your benefits for Montgomery GI Bill SR (Selected Reserve) then your Active Duty Kicker will stay attached to your Active Duty benefit.

However, if you transfer the Active Duty benefit, then you can transfer the Kicker that was associated to your newly elected Post-9/11 GI Bill benefit.

The same is true in the reverse.

This road bump is only relevant for someone that is eligible for all three of those benefits and is likely only a problem for people with very large kickers. Most kickers are small enough that it doesn't matter.

cc @goldenmeanie

akainic:
How do you transfer a benefit?

cweber:
Our understanding is that it happens automatically. We thought at one point that you had to fill out another form, but Shay confirmed that you don't.

This is also the case where document upload becomes important. Kickers are one of the most common documents that are asked for. It sounds like that document is potentially the only proof of the kicker.

cweber:
The most important thing we're trying to do is call out Active Duty vs Selected Reserve. I think we could emphasize that a little more with language instead of -SR or -AD and bold those terms. But I'd like to get this version up first.

akainic:
Got it, fine with this language going up if Shay approved it. I think it could be further clarified, but we can tackle that in a future PR (like adding the detail that the transfer happens automatically, the way it sounds now is like you need to do that).

Issue #4100, opened 2016-11-23
Stuck in education section on live form


cweber:
Testing out the live form for other reasons and I'm getting stuck on the education page.

I have filled in NO fields in education, but it keeps scrolling me to the top of the page when I hit continue.

jbalboni:
The issue we found is that if you touch the year field on the graduation date, you get stuck.

This is happening because of the way we default the day field for month/year dates. Touching the year input updates the date's value from completely blank to having a 1 as the day. That causes some of our validation to see the date is incomplete.

cweber:
This seems like it could be related to #4085

jbalboni:
This should be fixed with #4102.

PR #4102, opened 2016-11-23
Add special validation for month year field

alexose:
This PR adds a new validation rule for the `highSchoolOrGedCompletionDate` field that prevents the bug described [here](https://github.com/department-of-veterans-affairs/vets-website/issues/4100)


akainic (approved):
LGTM!

PR #4112, opened 2016-11-28
Set autocomplete correctly on fields on pages with autofillable fields

### LINKED ISSUE #4060: Autocomplete not working
Mainly happening on Address, perhaps elsewhere.

jbalboni:
Closes #4060 

Apparently Chrome refuses to autofill if there's a field that it doesn't understand that doesn't have `autocomplete` set to false. For edu benefits, the radio buttons on this page were the culprit, so I'm setting `autoComplete` to false in `ErrorableRadioButtons`.

This also fixes the way we're setting React keys.


jbalboni:
Updated some more fields so that the veteran information and secondary contact pages also work correctly with autofill.

akainic:
Hm I just tested this locally and it doesn't seem to have fixed the problem for me. When on the contact information page for education, once I start trying to enter my address, the dropdown with previous addresses I've entered appears as usual. If I select one of those addresses, the expected behavior is that the rest of the address fields will be prefilled with this information. But this is not happening for me.

screen shot 2016-11-29 at 10 19 58 am

jbalboni:
@akainic That's odd. Does it work for you on the other pages (veteran information and secondary contact)?

akainic (approved):
LGTM! Noticed a separate bug where the page validation is letting you proceed without entering email address. I will create a separate ticket for that.

PR #4117, opened 2016-11-28
Adds URL params to login url to direct a user to login or signup

### LINKED ISSUE #4044: Create an Account button should open sign up pop-up instead of sign in
There are different pop ups for "sign in" / "create an account" but our two buttons go to the same "sign in" window. Can we change the create an account to open the account creation pop up?

screen shot 2016-11-18 at 5 50 46 pm

screen shot 2016-11-18 at 5 52 16 pm
screen shot 2016-11-18 at 5 52 27 pm

crwallace:
resolves #4044


bshyong (approved):
LGTM!

PR #4120, opened 2016-11-28
Focus on first element instead of dialog in Modal component

jbalboni:
Related to [235](https://github.com/department-of-veterans-affairs/sunsets-team/issues/235)

Focusing on the dialog element itself instead of the first focusable element was causing screen readers to read the content twice. When you focus on a button (for example) in a dialog for the first time, screen readers will automatically read the content. So setting the focus to the dialog itself meant they would read the content, then read it again after tabbing.


jbalboni:
FYI @crwallace @webinista @U-DON.

src/js/common/components/Modal.jsx
```diff
@@ -1,12 +1,15 @@
...
 import classNames from 'classnames';
 
-function focusListener() {
+function focusListener(selector) {
   const listener = event => {
     const modal = document.querySelector('.va-modal');
     if (!modal.contains(event.target)) {
       event.stopPropagation();
-      modal.focus();
+      const focusableElement = document.querySelector(selector);
```
austin-martinez: Will the element always be inside the modal? Could you use `modal.querySelector(selector)` here to avoid searching the whole DOM?

src/js/common/components/Modal.jsx
```diff
@@ -31,7 +34,10 @@ class Modal extends React.Component {
 
   componentDidUpdate(prevProps) {
     if (!prevProps.visible && this.props.visible) {
-      document.querySelector('.va-modal').focus();
+      const focusableElement = document.querySelector(this.props.focusSelector);
```
austin-martinez: Similar question as above, should this be limited to the modal itself rather than the whole document?

austin-martinez (approved):
LGTM 🗾

U-DON:
LGTM.

PR #4127, opened 2016-11-28
[Messaging] 'Send' button is disabled when message is empty and sends message when enabled and clicked.

U-DON:
## Changes
* Disable 'Send' button until message body is populated. Closes department-of-veterans-affairs#163.
* Clicking 'Send' button will actually send the message. Closes department-of-veterans-affairs#167.


webinista:
LGTM

Issue #3935, opened 2016-11-11
Sticky Box: Edu benefits "apply" button bleeds off-canvas on mobile


melwoodard:
The action box container should be 100% width

<img src="https://cloud.githubusercontent.com/assets/4106356/20226962/1d69aabc-a7ff-11e6-9813-44f87d14505a.PNG" width="300"/>

<img src="https://cloud.githubusercontent.com/assets/4106356/20226965/1e855c66-a7ff-11e6-8e2a-058c79d3e3a2.PNG" width="300"/>

Issue #3828, opened 2016-11-08
Sticky box should be full width on mobile


cweber:
The sticky box should be full width on mobile at all times, not just after it becomes sticky.

PR #4131, opened 2016-11-29
Fix various sticky action box issues

### LINKED ISSUE #3827: Sticky box is not sticking at the right point
The sticky box should not stick until the box reaches the top of the scream on a normal scroll. Currently it's jumping on mobile and jumps on desktop after you scroll down and then back up,

jbalboni:
Closes #3827, #3828, #3935 

Fixes some positioning issues by grabbing the initial position of the box and using that for comparisons when scrolling. Also sets the box to full width on mobile and adds a resize handler so that it works properly when resizing between mobile and desktop.

screen shot 2016-11-29 at 12 25 30 pm
screen shot 2016-11-29 at 12 25 16 pm
screen shot 2016-11-29 at 12 25 12 pm

### DIFF
--- a/src/js/common/utils/sticky-action-box.js
+++ b/src/js/common/utils/sticky-action-box.js
@@ -31,17 +31,25 @@ function resetSmallActionBox(smallActionBox) {
 }
 
 function moveActionBoxOnScroll(actionBox, offset, positionFunction, resetFunction) {
-  const initialPosition = offset.top + document.body.scrollTop;
+  let initialPosition = offset.height > 0 ? offset.top + document.body.scrollTop : null;
   window.addEventListener('scroll', _.throttle(() => {
     if (offset.top < 0) {
       positionFunction(actionBox);
       offset = actionBox.getBoundingClientRect();
     }
 
-    if (document.body.scrollTop > initialPosition) {
+    if (initialPosition && document.body.scrollTop > initialPosition) {
       positionFunction(actionBox);
     } else {
       resetFunction(actionBox);
+    }
+  }, 100));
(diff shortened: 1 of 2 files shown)


alexose:
Looks good at first glance.  I'm going to take a bit to try it on different devices, etc.

src/js/common/utils/sticky-action-box.js
```diff
@@ -25,22 +26,32 @@ function resetSmallActionBox(smallActionBox) {
...
     }
 
-    if (document.body.scrollTop > offset.top) {
+    if (initialPosition && document.body.scrollTop > initialPosition) {
       positionFunction(actionBox);
     } else {
       resetFunction(actionBox);
     }
   }, 100));
+  window.addEventListener('resize', _.throttle(() => {
```
alexose: Never knew about `_.throttle`... cool!

alexose:
I'm noticing that there's a minor jog that happens on mobile.  When the box pops out from `position:initial` to `position:fixed`, it goes from this:

image

to this:

image

It's a pretty minor problem, but it might be worth looking at another way to handle the width.

akainic (approved):
LGTM!

jbalboni:
@alexose I see that problem when the screen width is smaller than the minimum page header width:

screen shot 2016-11-30 at 9 01 16 am

screen shot 2016-11-30 at 9 01 25 am

I think a better fix for that is to set a minimum width for the whole page or fix the page header so it collapses more, rather than changing the way the box width works. It seems pretty minor, though, as it only shows up for me on screen sizes below the iPhone 5.

I'll create a separate issue for it.

PR #4136, opened 2016-11-29
Use tel and email types in form fields

### LINKED ISSUE #4110: Use HTML5 input types for all form inputs
Currently when entering a phone number on  mobile, the default letters keyboard is displaying first. Switch to trigger numbers firsts or, ideally, the phone pad.

jbalboni:
Closes #4110.

Uses `tel` and `email` input types for phone and email fields.

screenshot_20161129-143507
screenshot_20161129-143408

### DIFF
--- a/src/js/common/components/form-elements/ErrorableTextInput.jsx
+++ b/src/js/common/components/form-elements/ErrorableTextInput.jsx
@@ -96,7 +96,7 @@ class ErrorableTextInput extends React.Component {
             name={this.props.name}
             tabIndex={this.props.tabIndex}
             autoComplete={this.props.autocomplete}
-            type="text"
+            type={this.props.type}
             maxLength={this.props.charMax}
             value={this.props.field.value}
             onChange={this.handleChange}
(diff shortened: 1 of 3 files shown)


src/js/common/components/form-elements/ErrorableTextInput.jsx
```diff
@@ -96,7 +96,7 @@ class ErrorableTextInput extends React.Component {
             name={this.props.name}
             tabIndex={this.props.tabIndex}
             autoComplete={this.props.autocomplete}
-            type="text"
+            type={this.props.type}
```
jbalboni: This seemed like the most reasonable change, rather than creating something like `ErrorableEmailInput` and `ErrorableTelInput`, which would result in some mostly duplicated components. We could rename ErrorableTextInput, but it didn't seem worth the widespread change it'd be.

akainic: I agree!

akainic (approved):
LGTM!

PR #4139, opened 2016-11-29
Adjusting position of MoveTo buttons in a table.

webinista:
Solves bug in which options within the MoveTo menu are cut off. Closes department-of-veterans-affairs/messaging-fe#181.

- Removed z-index property for .disclaimer. Superfluous.
- Added comment to _messaging-folder.scss.


U-DON:
LGTM.

Issue #3266, opened 2016-10-04
FontAwesome is being included twice


webinista:
Was wondering why `/generated/messaging.css` is an absurdly large _1.3MB_. Huge contributing factor: FontAwesome is being [bundled twice](https://www.vets.gov/generated/no-react.css) in our CSS. One of those instances inlines the SVG font as a data uri, adding about 400Kb. 

To fix this bug:
- [x] Figure out why FontAwesome's CSS is being bundled twice and address that.
- [x] Remove the SVG font altogether. WOFF and WOFF2 covers 98% of our audience. (We don't actually need to serve .eot files either.)

Possibly related to #3057.
Related to department-of-veterans-affairs/messaging-fe#94

bshyong:
Looks like this is related to the new login component requiring `styles`, which essentially doubles our file sizes as all common styles are duplicated.

This the following fix being tracked in this issue will bring the file size down to ~600kb.  Are we confirmed that the SVG font should be removed?

https://github.com/department-of-veterans-affairs/vets-website/issues/3312
https://github.com/department-of-veterans-affairs/vets-website/commit/3f9ce70d6a637306311c85d01486df9706ab7ba2

webinista:
Yes.

SVG font support [no longer exists](http://caniuse.com/#search=svg%20font) in most browsers. It's been replaced with WOFF in all but IE8, and we aren't supporting IE8.

bshyong:
@crwallace thoughts? ^

I can issue a PR for any necessary it once we have a decision

webinista:
Any movement on this?

jcmeloni:
@crwallace Still holding per comment here? https://github.com/department-of-veterans-affairs/vets-website/issues/3312#issuecomment-255442327

PR #4141, opened 2016-11-29
Making alert button background color transparent.

webinista:


U-DON:
LGTM. The close icon looks a little big compared to the alert icon, but not sure what sizes they should be.

webinista:
Yeah, I'm not clear on that either TBH. 

@melwoodard @gnakm How should these alert messages look relative to the warning messages?

gnakm:
@webinista It should follow USWDS. https://standards.usa.gov/accordions/
It looks like they use svg and png there (402x402). 
@melwoodard  and I just discussed this. Since IE does not read font awesome, we may have to start change icons to svg. 

Just put the files in the assets folder. Could somebody merge?
https://github.com/department-of-veterans-affairs/vets-website/pull/4151

webinista:
@gnakm I just want to double check that this alert should follow USWDS and not the [modified the disclaimer layout](https://github.com/department-of-veterans-affairs/kudos-team/issues/269). 

I think the problem with IE and FontAwesome is due to our use of a base64 encoded SVG font in our CSS file, not FontAwesome itself. See department-of-veterans-affairs/vets-website#3266.

melwoodard:
@webinista @gnakm this should follow the modified layout [here](https://github.com/department-of-veterans-affairs/vets.gov-designpattrns/issues/28), not the USWDS default. We want to begin using the modified style across the board.

PR #4144, opened 2016-11-29
Positions delete attachment button at top right.

webinista:
Closes department-of-veterans-affairs/messaging-fe#184


U-DON:
The button goes to the top right, but for long enough file names, it can overlay the text.

screen shot 2016-11-29 at 6 16 25 pm

webinista:
I think that's okay considering the screenshot attached in the original issue. 

I'll ask the question about how to handle extra long names there.

U-DON:
LGTM otherwise.

PR #4148, opened 2016-11-29
Last name min length

lihanli:
https://github.com/department-of-veterans-affairs/healthcare-application-team/issues/512

### DIFF
--- a/src/js/common/components/form-elements/ErrorableTextInput.jsx
+++ b/src/js/common/components/form-elements/ErrorableTextInput.jsx
@@ -26,6 +26,7 @@ import { makeField } from '../../model/fields.js';
 class ErrorableTextInput extends React.Component {
   constructor() {
     super();
+    this.state = {};
     this.handleChange = this.handleChange.bind(this);
     this.handleBlur = this.handleBlur.bind(this);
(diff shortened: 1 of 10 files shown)


akainic:
@lihanli Didn't notice that you had created a PR for this work. I had already started working on it, oops! Just pushed up what I have so far: https://github.com/department-of-veterans-affairs/vets-website/compare/hca-require-min-2-characters-for-name?expand=1.

The way I handled this was mainly through custom validation functions, and passing optional props to FullName to include those custom validation functions where necessary. What do you think about this approach?

jbalboni:
@lihanli @akainic I think I prefer the direction in Anne's PR, for what it's worth. It's closer to how we've been doing validation in edu and here, where we pass in validation functions/results and errors.

We do need a good solution for setting error messages based on the specific validation failure, though. That's a weakness in our current approach.

lihanli:
we can go with your branch anne, i worked on it because leanne assigned me to it, don't know the frontend conventions yet though

akainic:
@lihanli ah gotcha. Sorry about that confusion. I appreciate you giving this a shot in any case!

PR #4151, opened 2016-11-29
Add files via upload

gnakm:
Please review +/- icons and merge!


melwoodard:
@gnakm @webinista do you know what's up with these duplicates in the assets folder? Is this a result of how things get compiled or are there actually 3 sets of these icons?

svg

webinista:
No. This is some legacy duplication that should be cleaned-up. AFAIK we're using the same icons as USWDS. Those are already in our assets folder in `assets/img/` and duplicated in `assets/img/icons`.

The `.png` vs `.svg` issue is a browser support/legacy issue. I _think_ USWDS began with support for IE8, but now supports IE9+. PNG icons are a holdover from that IE8 support. 

I think we can close this pull request since the icons already exist.

PR #4157, opened 2016-11-30
Hca require min 2 characters for name

akainic:
Corresponding schema change: https://github.com/department-of-veterans-affairs/vets-json-schema/pull/43/files

### DIFF
--- a/src/js/common/components/questions/FullName.jsx
+++ b/src/js/common/components/questions/FullName.jsx
@@ -73,7 +73,7 @@ class FullName extends React.Component {
             onValueChange={(update) => {this.handleChange('middle', update);}}/>
 
         <ErrorableTextInput
-            errorMessage={this.validateRequiredFields(this.props.name.last, this.props.required, true) ? undefined : errorMessage}
+            errorMessage={this.validateRequiredFields(this.props.name.last, this.props.required, this.props.customValidation ? true : false) ? undefined : errorMessage}
             label="Last name"
             name="lname"
             autocomplete="family-name"
--- a/src/js/common/components/questions/FullName.jsx
+++ b/src/js/common/components/questions/FullName.jsx
@@ -39,9 +39,9 @@ class FullName extends React.Component {
     this.props.onUserInput(name);
   }
 
-  validateRequiredFields(field, isRequired, useCustomValidation) {
-    if (useCustomValidation) {
-      return validateIfDirty(field, this.props.customValidation);
+  validateRequiredFields(field, isRequired, customValidation) {
+    if (customValidation) {
+      return validateIfDirty(field, customValidation);
     } else if (isRequired) {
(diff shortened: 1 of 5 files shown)


akainic:
@lihanli I love the test you added in your PR for this fix, do you want to add here as well?

https://github.com/department-of-veterans-affairs/vets-website/pull/4148

lihanli:
i added it to that branch https://github.com/department-of-veterans-affairs/vets-website/blob/478ed8ee71d46c85f38caa1829717694f088bd1b/test/hca/utils/validations.unit.spec.js

akainic:
Okay, I can pull it into this branch, I'm guessing the easiest thing to do is to close that branch and just go with this one.

src/js/common/components/questions/FullName.jsx
```diff
@@ -111,6 +115,8 @@ FullName.propTypes = {
       dirty: React.PropTypes.bool,
     }),
   }).isRequired,
+  customValidation: React.PropTypes.func,
```
jbalboni: I was going to say we should leave off `custom` in these two properties, but it's probably better to leave it, since they work differently than when we pass `errorMessage` and `validation` to other components.

I think it would make sense to move towards validation functions that return the message and the status. The way we put ternary statements in these `errorMessage` props doesn't work well for components with a couple different validation errors. Not something for this PR, but should think about it in the future.

src/js/hca/utils/validations.js
```diff
@@ -88,7 +88,11 @@ function isValidDate(day, month, year) {
 }
 
 function isValidName(value) {
-  return /^[a-zA-Z][a-zA-Z '\-]*$/.test(value);
+  return /^[a-zA-Z '\-]*$/.test(value);
+}
+
+function isValidMinimumName(value) {
```
jbalboni: Is it possible to get rid of the other `isValidName` and call this function `isValidName`? These are HCA specific validation functions.

jbalboni (approved):
LGTM. One minor comment.

src/js/common/components/questions/FullName.jsx
akainic: I agree, in general we should rethink how we're handling error messages, it does get confusing.

src/js/hca/utils/validations.js
akainic: I was thinking about that, but the validation needs to be different if it's first name or last name. Wasn't sure how to handle that, since this function accepts a value and not the whole field, there's not a way to have conditional logic within it to check with field it is?

jbalboni: Oh, I see. Yeah, I'd just leave it then. Or maybe call it `isValidLastName`.

akainic: Will do!

PR #4158, opened 2016-11-30
Spouse address required fields

lihanli:
closes https://github.com/department-of-veterans-affairs/healthcare-application-team/issues/510

### DIFF
--- a/test/hca/utils/validations.unit.spec.js
+++ b/test/hca/utils/validations.unit.spec.js
@@ -2,9 +2,7 @@ import { expect } from 'chai';
 
 import _ from 'lodash';
 
-import {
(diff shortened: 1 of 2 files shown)


test/hca/utils/validations.unit.spec.js
```diff
@@ -0,0 +1,52 @@
+import { expect } from 'chai';
+
+import _ from 'lodash';
+
+import {
+  isValidSpouseInformation
```
akainic: Nit: probably can put this whole thing on a single line since we're only importing one function?

akainic (approved):
Looks great! And tests added too, awesome :) Thanks for tackling this one!

PR #4160, opened 2016-11-30
Improve hca zip validation

akainic:
Closes https://github.com/department-of-veterans-affairs/healthcare-application-team/issues/509


jbalboni (approved):
LGTM. Reminds me that we should probably delete the Address component that's in common. I don't think it's used anywhere.

PR #4161, opened 2016-11-30
[Messaging] Dismiss open alerts when route changes.

U-DON:
Closes department-of-veterans-affairs/messaging-fe#131.


webinista:
LGTM

PR #4165, opened 2016-11-30
School selection review fix

### LINKED ISSUE #4154: Validation on School Selection broken on review page
For some reason the validation on the review page for the school selection chapter is broken (but not on the original page itself). When adding a school, it does not let me update the chapter, even though all of the required fields are complete.

lihanli:
closes #4154


jbalboni:
@lihanli I don't see how this would fix #4154, though it does fix a bug with what the review page displays.

lihanli:
if you try to add a school in the review page it won't let you update because this section won't render

jbalboni:
Oh, I see. Looks good to me, then.

jbalboni (approved):
LGTM

PR #4173, opened 2016-11-30
Use validation function from common

akainic:
Currently we are using 2 different validations for phone numbers on HCA:

1. The common `Phone` component uses the common validation function here: https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/js/common/utils/validations.js#L125
2. The chapter validation for the ContactInformation section uses a different validation for phone number, here: https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/js/hca/utils/validations.js#L267, and here: https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/js/hca/utils/validations.js#L107

The result was that a user would be prevented from continuing onto the next page if they entered `617-555-5555`, but they wouldn't see an error message on the page (confusing!). This just makes sure that the section validation is using the same logic as the `Phone` component validation.

We should not merge this PR before this one is merged first: https://github.com/department-of-veterans-affairs/healthcare-application/pull/305. Otherwise we risk allowing through dashes and parentheses to ES, which will fail.


jbalboni (approved):
LGTM

PR #4176, opened 2016-12-01
Time individual operations in travis build

aub:
This wraps all of the individual calls made by Travis as part of the build with a `time` call, so it'll report how long they took. Travis will tell us how long the entire script execution takes but not the individual calls within it. With this, each call will be followed by something like:

```
real	6m4.824s
user	4m18.831s
sys	0m30.667s
```


ayaleloehr:
LGTM

PR #4189, opened 2016-12-01
Menu now closes when veteran changes the folder.

webinista:
- Closes department-of-veterans-affairs/messaging-fe#177
- Updated handleFolderChange in Main to add this.props.toggleFolderNav()


U-DON:
I think it should only modify that part of the state if the nav is already open. Also realizing now that I also put in `toggleFolderNav` for the click handler of the close button, but it would be clearer to just make another method `closeFolderNav` that only calls `toggleFolderNav` if the nav is opened.

webinista:
I think that's a minor optimization that we can fix after launch.

U-DON:
Sure. It's not a major addition but also not critical. LGTM.

Issue #4028, opened 2016-11-18
Improve visual integration of account links and global nav


melwoodard:
- Moves account links into blue header container
- Adds responsive behavior to account links position
  - By default, the links appear on one line between the logo and search bar
  - If the links can't fit on one line in this position, they float right below the search bar
  - If the links can't fit on one line below the search bar, they may wrap
- Makes logo smaller at smallest breakpoint
- Retains full search bar at medium breakpoint; only changes to the button at smallest breakpoint
- Displays the search icon instead of text button at smallest breakpoint

<img width="984" alt="account-responsive" src="https://cloud.githubusercontent.com/assets/4106356/20444529/3d0f0a60-ad86-11e6-8a57-11e0508d0554.png">

<img width="987" alt="account-specs" src="https://cloud.githubusercontent.com/assets/4106356/20444532/3fe117e2-ad86-11e6-9ef5-750daeeafc6c.png">

melwoodard:
@angel since we're still a ways out from adding a full site nav, this seems like a good interim solution based on what we were working with. Thoughts?

angel:
@melwoodard this looks good! My only concern is when we don't have people's names and just an email. I pretty sure that is a possible use case. In which case, it can sometimes be pretty long, I'm wondering how that would look in this design specifically for large and small. I think it'll be fine for medium.

melwoodard:
Thanks, good point! 

My thought is that the responsive behavior for the account greeting and links is not dictated by viewport width, it's dictated by wherever it can all fit on one line. 

If all the text fits on one line between the logo and search bar it will go there, otherwise it'll drop below. 

When it's dropped below and still can't fit on one line, then it would break up into multiple lines like so:

<img width="320" src="https://cloud.githubusercontent.com/assets/4106356/20496099/a40500f4-afd8-11e6-8fb1-0aceb2b5663e.png">

angel:
@melwoodard cool, that looks good to me! Thanks!

melwoodard:
Thanks!

@crwallace this isn't high priority, let me know where this goes and if we're able to address this before launch.

bshyong:
@melwoodard @angel is it possible to have the sign-in/register links sit on the line below the search and logo on all viewports? In the mockups, large is the only exception. Implementation-wise it's a little finicky to have the links hop down like that. If the design is completely set in stone, I can implement as-is in the mock up, LMK

melwoodard:
[...]

Issue #4027, opened 2016-11-18
Visual & responsive improvements for notice banner


melwoodard:
Visual updates:
- [x] Color changes for better visual integration w/ the header
- [x] Updated layout on Crisis Line callout

Responsive behavior:
- [x] Layout modified to stack vertically instead of side-by-side at small breakpoints
- [x] Consistent sizing for Crisis Line icon across breakpoints

<img width="978" alt="screen shot 2016-11-17 at 4 57 46 pm" src="https://cloud.githubusercontent.com/assets/4106356/20414361/e6577cf6-ace6-11e6-9a91-b99035f7da4c.png">

![elements](https://cloud.githubusercontent.com/assets/4106356/20414487/89559294-ace7-11e6-8522-3fb072370fc7.png)

![structure-large](https://cloud.githubusercontent.com/assets/4106356/20414497/91c3abfa-ace7-11e6-8e78-bf19409d8f44.png)

<img width ="480" src="https://cloud.githubusercontent.com/assets/4106356/20414500/94ac3878-ace7-11e6-9568-927b2eb839d4.png">

PR #4194, opened 2016-12-01
Vanguard style cleanup

### LINKED ISSUE #3808: [HOMEPAGE] Add hover state to popular buttons
Let's make the border color the same blue as the cards when a Veteran hovers over that button. Can we also increase it to 2px so that it's more visible?

### LINKED ISSUE #4033: LinkedIn Trademark symbol appears as ?
I noticed that on this page https://www.vets.gov/employment/job-seekers/ , the LinkedIn symbol isn't rendering correctly. A ? is showing up and it should be the TM symbol.  

It's on this page 3 times. 

Thanks! 
@bethpotts  FYI

bshyong:
- Resolves https://github.com/department-of-veterans-affairs/vets-website/issues/4027 move login bar to top nav
- Resolves https://github.com/department-of-veterans-affairs/vets-website/issues/4028 topnav styling improvements
- Resolves #3808; popular on vetsgov button borders
- Resolves #4033, linkedin trademark symbol


crwallace:
@bshyong -- Should the VCL have a border radius on it?
screen shot 2016-12-01 at 6 34 37 pm

crwallace:
Also @bshyong, do we want the search icon to match the large version?
screen shot 2016-12-01 at 6 37 19 pm

Would love to get @AnitraAppa and @maryannbrody's thoughts on if we can use and icon alone in the mobile view or if we need to keep the full "search" text.

bshyong:
@crwallace looks like the border radius on the div doesn't show; fixed it in a commit.  

For the search icon, Mel had a mockup here that I missed implementing—this will be pushed up shortly

maryannbrody:
@crwallace @bshyong Believe it or not, in usability testing the search icon isn't always recognizable to all users, especially those with lower literacy skills. I'd recommend sticking to "search" in a button if we can. 

cc/ @AnitraAppa

AnitraAppa:
@maryannbrody Good to know, would have just used icon. +1 to keeping "Search" then.

bshyong:
Added back the search button for mobile: for the small screen size, the textbox is too large to display, so instead there will be a button labeled "Search" which opens the full screen search input.

melwoodard:
@bshyong looking good! A couple comments on the nav work:

1. Is there a middle breakpoint we can take advantage of? I had in mind that the 100% width for the VCL and beta banner outlined [here](https://cloud.githubusercontent.com/assets/4106356/20414361/e6577cf6-ace6-11e6-9a91-b99035f7da4c.png) would happen at something closer to <600px wide, but right now it's hitting everything <1024px when there's still plenty of room for them to sit side by side:

   vets_gov

2. I see the updated smaller size for the logo, but that smaller size was only supposed to be at the smallest breakpoint (corresponding to the idea of breakpoints above) - at the larger breakpoint it should remain at the original size we had before.

bshyong:
@melwoodard just fixed the breakpoints and logo size this morning. The only issue that I can see now moving through the breakpoints is that the crisis line text will not fit on one line right before it breaks to the small breakpoint. Unless we shorten the text to Vets Crisis Line or reduce the font size I don't think it can fit in the current state. 

However, this issue only happens when you are resizing the window to a non-standard size

screen shot 2016-12-05 at 8 38 12 am

melwoodard:
@bshyong is there a way to fix the width of the crisis line container until it breaks to the small breakpoint? It'd be preferable for the beta text to wrap and the crisis line to stay fixed, like so:

screen shot 2016-12-05 at 9 14 01 am

bshyong:
Got it; that approach looks great

screen shot 2016-12-05 at 10 26 56 am

melwoodard:
Thanks, looks great!

crwallace (approved):
LGTM

PR #4006, opened 2016-11-16
Allow modals to have content children

typesend:
- for GIBCT cleaner syntax was desired as an alternative to `content={}`
- `props.content` continues to work and existing modals are unaffected by this enhancement
- modals make for naturally good parents and deserve happy children

*example usage*
```
<Modal onClose={} visible={}>
  <h3>Military Status</h3>
  <p>blah blah blah</p>
</Modal>
```

Issue #3960, opened 2016-11-14
GI Bill Comparison Tool Integration


typesend:
The overall plan is to replace the GIBCT Rails app by implementing its backend and frontend within vets-api and vets-website respectively. This will eliminate header/footer synchronization problems and the need to maintain a separate standalone rails app. GIDS will likely still publish data to an isolated database but this is still being discussed.

*Related issues*
department-of-veterans-affairs/gi-bill-comparison-tool#438
department-of-veterans-affairs/gibct-data-service#160

jcmeloni:
@plusjeff This seems like it should be labelled somehow, but not sure how you'd like it.

typesend:
_Recent discussions have generated a number of proposed enhancements. Comments welcome._

### Data Previewing
- A "preview mode" feature will be implemented in the React app.
  - A url param determines the version of the data to display.
  - There will be an obvious visual indication that preview mode is active and showing which version of data is being served.
  - Just like the Preview app, the intention is that preview/experimental data is only going to be reviewed by a few known people. However, viewing it will no longer require login/authorization.
- The Preview app instance and its database will be decommissioned.
- _Benefits: fewer apps and databases to operate, data display logic in one place_

### Data Publishing
- The Data Service app will cease injecting data directly into external databases.
- Instead, vets-api will rely on postgres to maintain a read-only connection to the data service database (using the postgres_fdw foreign data wrapper extension).
  - Specifically, the data_csv table and a new table containing constants used during benefit calculations will be loaded as **foreign tables**
  - vets-api won't serve data out of those tables directly, instead copying the data to its own local tables when a new release is created. (Because performance, versioning.)
- Publishing data will now be triggered by the Data Service app calling a new endpoint in vets-api
- _Goals: keep vets-api database isolated and secure, simplify how data gets shoved around_

### Data Versioning
- Inside the Data Service app, a `data_version` auto-incrementing column will be added to its `data_csv` table. New data sets will be appended to this table (instead of replacing historical data.) The same pattern will be applied to the constants table. @mphprogrammer 
- Vets API will be responsible for tracking when data and constants are actually published (when the endpoint for that purpose is called)
- The goal is to be able to answer two questions (with little additional development work): "What data was displayed by the website on [date] at [time]?" and "What constants were used as the basis for benefit calculations at that time?"

### Content and Design
- We're also taking this opportunity to improve content, look and feel. (See department-of-veterans-affairs/vets.gov-team#508)
[...]

PR #4196, opened 2016-12-01
[WIP] Initial GIBCT react app

typesend:
Sharing my #3960 progress so far for review!

This is my first react application so I have definitely experienced a fair amount of "two-steps-forward, one-step back", but I feel like this is finally evolving in the right direction.

image

FYI, this includes code from #4006

@emilyville @ayaleloehr @plusjeff @mphprogrammer @patrickvinograd


src/js/gi-bill-comparison-tool/components/AboutYourSchoolFields.jsx
```diff
@@ -0,0 +1,66 @@
+import React from 'react';
+import Dropdown from './Dropdown';
+import Modal from '../../common/components/Modal';
```
austin-martinez: Don't see this or Dropdown used here

src/js/gi-bill-comparison-tool/components/Dropdown.jsx
```diff
@@ -0,0 +1,79 @@
+import React from 'react';
+
+class Dropdown extends React.Component {
```
austin-martinez: I'm not too familiar with all of our common React components, but we do have one for selects, could that be used here, or extended to add in any missing functionality?

src/js/gi-bill-comparison-tool/containers/LandingPage.jsx
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  renderBreadcrumbs() {
+    return (
+      <nav className="va-nav-breadcrumbs">
+        <ul className="row va-nav-breadcrumbs-list" role="menubar" aria-label="Primary">
+          <li><a href="/">Home</a></li>
+          <li><a href="/education/">Education Benefits</a></li>
+          <li><a href="/education/gi-bill/">GI Bill</a></li>
+          <li className="active">GI Bill® Comparison Tool</li>
```
austin-martinez: Why the ® here after GI Bill but not in the preceding breadcrumb? Had no idea that was a registered trademark, do we need to denote that? http://benefits.va.gov/GIBILL/Trademark_Terms_of_Use.asp this looks like it would only be necessary for third parties referring to GI Bill, but not the VA itself

src/js/gi-bill-comparison-tool/components/Dropdown.jsx
typesend: There is `<ErrorableSelect/>` but I didn't want to bite off trying to reuse that before I knew what that css style implications of that would be. Certainly open to reusing/extending that if people agree it makes sense (as a future enhancement.)

src/js/gi-bill-comparison-tool/containers/LandingPage.jsx
typesend: I agree that it is unnecessary. Removing.

src/js/gi-bill-comparison-tool/components/Dropdown.jsx
jbalboni: All of our components should use the USWDS styles, so it doesn't look like it would change much, style-wise. But waiting to incorporate them until the rewrite is more settled seems reasonable to me.

src/js/gi-bill-comparison-tool/components/Dropdown.jsx
```diff
@@ -0,0 +1,79 @@
...
+    if (!this.props.visible) {
+      return null;
+    }
+    return (
+      <div id={this.props.identifier + '-form'} className={this.props.className}>
+        {this.renderLabel()}
+        <select id={this.props.identifier}
+            name={this.props.identifier.replace(/\-/g, '_')}
+            alt={this.props.alt}
+            onChange={(e) => { this.props.onChange(e) && this.setState({ value: event.target.value }); }}
```
jbalboni: We've been writing our form input components as [controlled components](https://facebook.github.io/react/docs/forms.html#controlled-components), where we pass the value down from a parent component (and changes back up with `onChange` hooks like you have here). You can check out how `ErrorableSelect` works to see it. It'll make it easier to migrate to our common components later if you use the same style here (or give `ErrorableSelect` a shot now).

src/js/gi-bill-comparison-tool/components/Dropdown.jsx
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  renderLabel() {
+    if (!this.props.showLabel) {
+      return null;
+    }
+    return this.props.children;
+  }
+
+  renderDropdownOptions(optionsArray) {
```
jbalboni: Is there a reason for passing the options array as an array of strings, instead of an array of objects? It seems like it's making this more complicated.

If you passed in options arrays like this:

```
[
  { value: 'yes', label: 'Eligilble for Post-9/11 Gi Bill' },
  { value: 'no', label: 'Not Eligilble for Post-9/11 Gi Bill'}
]
```

This function can be one line:

`return optionsArray.map(({ value, label }) => <option key={value} value={value}>{label}</option>);`

typesend: Although it does increase the complexity of this one clear-purpose function, it really simplifies the code where these Dropdowns are used.

src/js/gi/components/AboutYourselfFields.jsx
```diff
@@ -0,0 +1,358 @@
...
+              'no', 'None Online',
+              'both', 'Both online and on-campus'
+            ]}
+            defaultOption="no"
+            alt="Online? Yes/No"
+            visible={this.state.dropdowns.onlineClasses || !this.props.labels}
+            onChange={this.handleDropdownChange}
+            showLabel={this.props.labels}/>
+
+        <Modal onClose={() => {this.toggleModalDisplay('militaryStatus');}} visible={!!this.state.modals.militaryStatus}>
```
U-DON: Not sure if this is considered more idiomatic or not, but what I like to do is pass something like:

`this.toggleModalDisplay.bind(this, 'militaryStatus')`

Should end up being the same as creating the anonymous function.

src/js/gi/components/AdditionalResources.jsx
```diff
@@ -0,0 +1,59 @@
+import React from 'react';
+
+class AdditionalResources extends React.Component {
+
+  render() {
+    return (
+      <span>
```
U-DON: This should probably be a `div`.

src/js/gi/components/Breadcrumbs.jsx
```diff
@@ -0,0 +1,45 @@
+import React from 'react';
+
+class Breadcrumbs extends React.Component {
```
U-DON: We have a Breadcrumbs component in `rx` as well. Probably not for this PR, but it might be useful to pull both out into `common` at some point.

src/js/gi/components/Dropdown.jsx
```diff
@@ -0,0 +1,75 @@
...
+
+    this.state = {
+      value: props.defaultOption
+    };
+  }
+
+  // renders a list of options defined as an array
+  // in the form of [value1, label1, value2, label2...]
+  renderDropdownOptions(optionsArray) {
+    const options = Object.assign([], optionsArray);
```
U-DON: I don't think `Object.assign` is intended for cloning arrays. Instead, maybe:

`const options = optionsArray.slice();`

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice

src/js/gi/components/SearchResult.jsx
```diff
@@ -0,0 +1,153 @@
...
+        </div>
+      </div>
+    );
+  }
+
+  render() {
+    this.estimate();
+
+    return (
+      <div id={this.props.facility_code} className="school_summary large-4 medium-5-center small-12-center columns">
```
U-DON: The props are all in camelCase, so probably meant `this.props.facilityCode`. Same with `cautionFlag` below.

src/js/gi/components/SearchResult.jsx
```diff
@@ -0,0 +1,153 @@
...
+    this.renderTuition = this.renderTuition.bind(this);
+    this.renderHousing = this.renderHousing.bind(this);
+    this.renderBooks = this.renderBooks.bind(this);
+    this.estimate = this.estimate.bind(this);
+  }
+
+  // WIP
+  estimate() {
+    // get current values about the user
+    const e = this.props.estimator;
```
U-DON: I see that it's a WIP, but in case you didn't already know, it's not considered good practice to modify props.

src/js/gi/components/profile/GraduationRates.jsx
```diff
@@ -0,0 +1,60 @@
+import React from 'react';
+
+class GraduationRates extends React.Component {
+
+  render() {
+    const school = this.props.institution;
+    const isNumeric = (n) => { return !Number.isNaN(parseFloat(n)); };
+    const isaV = isNumeric(school.graduationRateVeteran);
+    const isaC = isNumeric(school.graduationRateAllStudents);
+    const heading = <h3>Graduation Rate<a onClick={() => {this.props.toggleModalDisplay('gradrates');}} className="info-icons"><i className="fa fa-info-circle info-icons outcomes-learnmore"></i></a></h3>;
```
U-DON: Would be nice to break this up into multiple lines for readability.

src/js/gi/containers/LandingPageForm.jsx
```diff
@@ -0,0 +1,55 @@
...
+import AboutYourselfFields from '../components/AboutYourselfFields';
+import AboutYourSchoolFields from '../components/AboutYourSchoolFields';
+import AdditionalResources from '../components/AdditionalResources';
+
+class LandingPageForm extends React.Component {
+
+  render() {
+    return (
+      <div className="row">
+        <form action="/gi-bill-comparison-tool/institutions/search">
```
U-DON: Does this end up doing a full page refresh? If it's hooked up in the routes, you should have the `onSubmit` handler do something like a `history.push` instead, so it can render in place.

src/js/gi/components/Breadcrumbs.jsx
```diff
@@ -0,0 +1,45 @@
...
+  renderDefault(props) {
+    return (
+      <li className="active">{props.currentLabel}</li>
+    );
+  }
+
+  renderNamed(props) {
+    return (
+      <span>
+        <li><a href="/gi-bill-comparison-tool/">GI Bill Comparison Tool</a></li>
```
U-DON: This should probably be a `Link` from `react-router`. Any of this app's routes should use `Link` rather than `a`, or else the links will load the full page.

Also, to avoid prefixing all the internal links with `/gi-bill-comparison-tool`, you can do something like this: https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/js/rx/rx-entry.jsx#L16

src/js/gi/components/AboutYourselfFields.jsx
```diff
@@ -0,0 +1,358 @@
...
+  }
+
+  toggleModalDisplay(name) {
+    this.setState({ modals: {
+      [name]: !(this.state.modals[name])
+    } });
+  }
+
+  renderVREBenefitsMessage() {
+    // onClick="track('Tool Tips', 'Voc Rehab Warning');"
```
typesend: This is a reminder for how click tracking was implemented in the rails app

src/js/gi/components/Breadcrumbs.jsx
typesend: Noted.

src/js/gi/components/Dropdown.jsx
typesend: Good catch

src/js/gi/components/AdditionalResources.jsx
typesend: I was using `<span>` just to denote this parent tag was needed by JSX but not in the original project code. Agree, though, and changing this across all components where it was done.

src/js/gi/components/SearchResult.jsx
typesend: Yep, fixing

src/js/gi/components/AboutYourselfFields.jsx
typesend: Nice! This is way better than the anonymous function garbage.

src/js/gi/components/AboutYourSchoolFields.jsx
```diff
@@ -0,0 +1,61 @@
...
+        <div id="institution-search-form" className="form-group">
+
+          <If condition={this.props.labels}>
+            <div id="online-classes-form" className="form-group">
+              <label>Will you be taking classes online? </label>
+
+              <input type="radio" id="online-classes-all"
+                  name="online_classes" value="yes"
+                  defaultChecked={this.props.onlineClasses && this.props.onlineClasses === 'all'}/>
+              <label className="radio" htmlFor="online-classes-all" style={{ display: 'inline-block' }}>All</label>
```
typesend: Turns out `inline-block` is needed so I moved it into a stylesheet. (These are old styles; I've spent as little time as possible cleaning up old css.)

script/build.js
```diff
@@ -116,6 +116,13 @@ smith.destination(`../build/${options.buildtype}`);
 // }
 // smith.use(ignore(ignoreList));
 
+const ignore = require('metalsmith-ignore');
+const ignoreList = [];
+if (options.buildtype === 'production') {
+  ignoreList.push('gi/*');
```
U-DON: I think this uses the relative path from the `content` folder, so it should probably be `'gi-bill-comparison-tool/*`.

typesend:
Thank you reviewers!

Issue #3330, opened 2016-10-12
Check scrolling on form submissions with errors


cweber:
In at least one case when an error is triggered on page submission, the page is scrolling to a point that does not correlate to the actual error.

This is probably worth checking for all errors. It currently is happening on Military history if someone chooses to apply the service period to a different benefit and does not select which benefit. In this case it sends you to the top of the tour section instead of the top of the error.

jbalboni:
This is also necessary for screen reader users, since there's no real indication of what happened when validation fails right now. We should scroll to and focus on the erroring field. We may need to add `alert` roles to some of the error messages in our input components.

PR #4198, opened 2016-12-01
Tab order updates for edu benefits

### LINKED ISSUE #3498: Tab order starts with the continue button inside the edu flow
Currently the page loads with focus on the continue button. It should default to the top of the page.    If practical, "Skip to content" could become "Continue with form".

X of y step name at the top of the form should use https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_progressbar_role

jbalboni:
Closes #3498 

This updates how we handle focus in edu benefits. On the intro page, breadcrumbs are focused after load. On the form pages, the legend or the navigation header is focused, depending on what's visible. We could focus on the first input on the page, but I think that will skip content on screen readers. We could also always focus on the breadcrumbs, but they never change, so it seems redundant. This is similar to what we're doing on claim status, which is based on some feedback from Courtney and the 508 office.

We can make these changes on HCA, too, but I wanted to get some feedback on it first.


jbalboni:
@akainic @cweber @goldenmeanie Can you try this out locally and see what you think?

cweber:
Looks good. Found two funny things.

1. If you have a validation error and try to submit the page it scrolls you up to the error but it doesn't change the focus. The next tab takes you into the footer.

2. Review section is a little funny. Probably best shared through screen share.

goldenmeanie:
Super exciting to see the tabs mostly working!

jbalboni:
@cweber For 1, I was planning on handling that in #3330.

src/js/edu-benefits/components/SubmitMessage.jsx
```diff
@@ -8,6 +8,13 @@ export default class SubmitMessage extends React.Component {
     this.state = { isExpanded: false };
     this.handleClick = this.handleClick.bind(this);
   }
+  componentDidMount() {
+    const title = document.querySelector('.edu-page-title');
```
akainic: Wondering if we want to extract out this functionality into something that can be shared across components?

src/js/edu-benefits/containers/IntroductionPage.jsx
```diff
@@ -1,6 +1,11 @@
 import React from 'react';
 
 class IntroductionPage extends React.Component {
+  componentDidMount() {
+    const breadcrumbs = document.querySelector('.va-nav-breadcrumbs-list');
+    breadcrumbs.setAttribute('tabindex', '-1');
```
akainic: Do we want to check that `breadcumbs` exists first, like you do elsewhere?

akainic (commented):
Looks great! Just a couple small comments.

src/js/edu-benefits/containers/IntroductionPage.jsx
jbalboni: We could. I was assuming they exist here because they always should, whereas the other header elements are different depending on the page.

src/js/edu-benefits/components/SubmitMessage.jsx
jbalboni: I don't think it would help much here, since the four places where this is happening are different. It will probably be useful for hca, though, since we'll have to make the same change in a bunch of places.

akainic (approved):
LGTM!

Issue #3675, opened 2016-11-02
Create month year component


jbalboni:
On Edu benefits we had a need for a month/year date input. We ended up adding a prop to DateInput to handle this, but it might make sense to create a separate component. The current DateInput component is complicated as it is.

cweber:
Did this get resolved now that we have Month/Year for high school graduation?

jbalboni:
This is still outstanding. We likely want to make a separate component for month/year, instead of using the normal date component.

alexose:
The big problem with my approach to altering `DateInput` is that it causes the date validations to not play especially well with each other.  @cweber noticed the following:

![date-junk](https://cloud.githubusercontent.com/assets/627264/20807559/849331aa-b7a3-11e6-9832-f8f4d54e0f78.gif)

We also need to update the review component:

![image](https://cloud.githubusercontent.com/assets/627264/20807629/cca8c84c-b7a3-11e6-921c-c223309bffb4.png)

leannammiller:
@alexose - could you give an update on this?

alexose:
We weren't able to recreate the bug that @cweber noticed, but the review component is fixed.  Closing!

PR #4199, opened 2016-12-01
Month year fix

alexose:
Fixes issue with review component described in #3675:

image


cweber:
I could never really replicate that bug so it's hard for me to do too much review, but I did realize one more thing about language.

For some reason we have a completely different statement for high school graduation date in form/review. I had just noticed the month/year business before, but we need to fully normalize that statement. And let's just remove the reference to date fields entirely. It should read:

When did you earn your high school diploma or equivalency certificate?

jbalboni:
@alexose, I think this fixes the date not showing up on the review page, but it seemed like #3675 described a validation bug, too? I couldn't replicate it, though.

jbalboni (approved):
LGTM as far as fixing the date displaying on the review page.

alexose:
Correct, this is just for the review page.  Will merge after CI

PR #4201, opened 2016-12-02
Display proper error message for users without access to Rx and SM

### LINKED ISSUE #4195: Displaying proper error when user doesn't have Rx or SM access
Alright, taking this from Github (https://github.com/department-of-veterans-affairs/kudos-team/issues/271). Will restate the problem here first:

- Currently a Veteran needs an MHV account to use Rx on vets.gov.  
- Right now, when a user without an MHV account goes to Rx, the error says "We couldn't retrieve your prescriptions. Please refresh this page or try again later." This error statement is false, as retrying later won't work either. 
- Instead, when a user goes to Rx, we should check their profile object to see if they have access to Rx.  If they do, we let them through normally, if they don't, we put up a message saying:

"To refill prescriptions, you need to be registered as a VA patient through MyHealtheVet.
To register, visit MyHealtheVet. If you're registered but you don't see your prescriptions here, please call the Vets.gov Help Desk at 1-855-574-7286, Monday‒Friday, 8:00 a.m.‒8:00 p.m. (ET)."

The "visit MyHealtheVet" section should link to: https://www.myhealth.va.gov/web/myhealthevet/user-registration

(note, all the above applies to SM as well, I'm just using Rx as the first example as I assume once we do it for Rx it will be trivial to add it for SM as well)

Jeff Balboni [3:13 PM]  
@ayaleloehr Yep. The RequiredLoginView component passes a prop down to its child component called `isDataAvailable` if a user fails the serviceRequired check.

[3:13]  
We handle that here: https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/js/disability-benefits/containers/DisabilityBenefitsApp.jsx

[3:14]  
It's a little odd because you have to make the child of RequiredLoginView a React component, not an element, and the prop is only passed when it's false.

U-DON:
Closes #4195.


webinista:
Seems correct to me. @ayaleloehr Do you know off hand which users do or do not have access to SM or Rx?

ayaleloehr:
In staging 261 to 264 have access to both. 87 does not have access to either.

jbalboni:
Looks good to me!

ayaleloehr:
LGTM

Issue #4015, opened 2016-11-17
Loading animation for user actions


U-DON:
How should the loading indicator be used when the user has performed an action that hits the API?

Right now, when transitioning between pages or tabs, we display the loading indicator in place of the section that will appear when we get the relevant data from the server.

But for actions like refilling a prescription or sending a message, we should also indicate that something is happening. I'm not sure if it makes sense to just replace the entire view with just the loader similar to how loading a page works. For example, if we just clicked to confirm a refill or send a message, it wouldn't look right to replace everything with the loader, and then show everything back again when the action is done. Some actions will redirect and transition to a different page when successful, so the loading indicator would work there, but when there's an error, the modal or page would remain.

So I'm wondering what the UX for these things should look like. My initial thought is to overlay the full page with a transparent background and have the loader fixed in the center. Let me know if my descriptions weren't clear.

U-DON:
Or maybe the loader isn't even the right element to use in these cases. Maybe the relevant button could be disabled temporarily and replaced with something like 'Sending...' in the case of a message being sent.

Some other cases where we need some type of wait indicator are moving a message, deleting a message, creating a folder, and deleting a folder. Might be missing some, but in general, SM has a lot more possible user actions, so those should be covered.

gnakm:
@U-DON I think we were thinking to show the loader but with different messages at the bottom depending on what it was: Here is the previous issue for this: https://github.com/department-of-veterans-affairs/kudos-team/issues/169

ex. 
For sending a secure message: 
[Loader on top]
**is sending your message...**

For refilling a prescription:
[Loader on top]
**is submitting your refill request...** 

I'm not really sure if we need to show a different kind of loader for other scenarios. What about a more generic message for the ones mentioned above:

For moving a message, deleting a message, creating a folder, and deleting a folder
[Loader on top]
**is submitting your request...**

@angel please chime in your thoughts!

angel:
@U-DON @gnakm 

I agree with showing different messages where we can with the same loader. I think it should just say "Sending your message..." without the "is". I don't think the "is" is necessary.

I think it makes sense have the loading take the entire view so then the Veteran can know that something is happening. If it's just happening in a small part of the page then it might be easy for them to miss it.

I think if the loading is needed after submitting from a modal then I think the loading should replace the modal box with the transparent background, but if it was say sending a message it should just take the entire view with no transparency and just a white background with the loading. The second one would still have the header and footer. What do you think @gnakm?

[...]

PR #4202, opened 2016-12-02
[Rx] Show loader after submitting refill.

U-DON:
As far as I can tell, this should be the only Rx-related change for #4015.

screen shot 2016-11-30 at 7 46 58 pm

### DIFF
--- a/src/js/rx/reducers/modals.js
+++ b/src/js/rx/reducers/modals.js
@@ -1,7 +1,9 @@
+import merge from 'lodash/fp/merge';
 import set from 'lodash/fp/set';
 
 const initialState = {
   refill: {
+    loading: false,
     visible: false,
     prescription: null
   },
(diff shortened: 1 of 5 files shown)


src/js/rx/reducers/modals.js
```diff
@@ -19,15 +21,20 @@ export default function modals(state = initialState, action) {
...
-      return initialState;
-
     case 'OPEN_GLOSSARY_MODAL':
       return set('glossary', {
         visible: true,
         content: action.content
       }, initialState);
 
+    case 'REFILL_SUBMITTED':
+      return merge(state, {
```
jbalboni: You can use `set` here: `set('refill.loading', true, state)`

jbalboni (approved):
LGTM

ayaleloehr:
Thanks for the review and helpful comment on `set` @jbalboni !  Since @U-DON is out today, will have him address in a follow up PR.

PR #4203, opened 2016-12-02
[Messaging] Show loader when fetching recipients.

U-DON:
Closes department-of-veterans-affairs/messaging-fe#191.
Part of the work for #4015.

## Changes
* Added action types for loading of all API requests.
* Moved recipients to its own reducer.
* Show loader when fetching recipients.
* Provide a link for user to retry loading recipients if the request failed.


src/js/messaging/actions/messages.js
```diff
@@ -8,17 +8,21 @@ import {
   DELETE_DRAFT_ATTACHMENT,
   DELETE_MESSAGE_FAILURE,
   DELETE_MESSAGE_SUCCESS,
+  DELETING_MESSAGE,
```
jbalboni: Are all the actions added in here being handled in a later PR?

src/js/messaging/components/forms/NewMessageForm.jsx
```diff
@@ -14,20 +14,42 @@ import MessageWriteGroup from '../compose/MessageWriteGroup';
...
+          <MessageRecipient
+              errorMessage={validations.isValidRecipient(message.recipient) ? '' : composeMessage.errors.recipient}
+              cssClass="msg-recipient msg-field"
+              onValueChange={this.props.onRecipientChange}
+              options={this.props.recipients}
+              recipient={message.recipient}/>
+        );
+      } else {
+        // The list of recipients somehow came back empty.
+        recipientsField = <p>No recipients could be found.</p>;
```
jbalboni: Is this case meaningfully different from `recipients` being falsy? The comment suggests it's not clear how you would get here, so is there a good reason to hide the retry link?

Also, may just be me, but I find the nested if else blocks here hard to read. This would be clearer to me:

```
  if (this.props.recipients && this.props.recipients.length) {
      recipientsField = (
        <MessageRecipient
            errorMessage={validations.isValidRecipient(message.recipient) ? '' : composeMessage.errors.recipient}
            cssClass="msg-recipient msg-field"
            onValueChange={this.props.onRecipientChange}
            options={this.props.recipients}
            recipient={message.recipient}/>
      );
    } else {
      // only show the retry option if recipients is falsy, which means the request failed
      recipientsField = (
        <p>
          No recipients could be found.&nbsp;
          {!!this.props.recipients && <a onClick={this.props.onFetchRecipients}>
            Click here to retry loading recipients.
          </a>}
        </p>
      );
    }
```

jbalboni (approved):
LGTM, made a couple comments. I don't know how you're handling tests for components, but unit tests might be helpful for some of the loading logic there.

ayaleloehr:
These comments seem substantial enough that I'm going to assign this back to @U-DON to decide how to act. I don't want to unilaterally merge this without giving him time to review. Thanks @jbalboni !

src/js/messaging/actions/messages.js
U-DON: Yes, I started adding them, and figured I should break the other loading-required actions into separate PRs.

src/js/messaging/components/forms/NewMessageForm.jsx
U-DON: I had `recipients` be null as the case when the request comes back with an error. If it comes back with an empty array for some reason, a retry wouldn't resolve the issue, and more action would be necessary (and an official copy should be used for the wording of that case). It's just hypothetically possible; I haven't seen it actually happen.

Your suggested change makes sense. I wasn't feeling the nested if-elses. I can go with that.

PR #4206, opened 2016-12-02
Add call center number to Rx error

### LINKED ISSUE #4204: Add Call center number for Rx error
We need to add call center number to error below. 
screen_shot_2016-11-30_at_9 27 52_am_720

New copy:
We couldn't retrieve your prescriptions. Please refresh this page or try again later. If this problem persists, please call the Vets.gov Help Desk at 1-855-574-7286, Monday‒Friday, 8:00 a.m.‒8:00 p.m. (ET). 

cc: @U-DON ✈️ and @webinista 😷

ayaleloehr:
closes #4204

### DIFF
--- a/src/js/rx/containers/Active.jsx
+++ b/src/js/rx/containers/Active.jsx
@@ -72,7 +72,9 @@ class Active extends React.Component {
       content = (
         <p className="rx-tab-explainer rx-loading-error">
           We couldn't retrieve your prescriptions.
-          Please refresh this page or try again later.
+          Please refresh this page or try again later. <br/>
+          If this problem persists, please call the Vets.gov Help Desk at 
+          1-855-574-7286, Monday‒Friday, 8:00 a.m.‒8:00 p.m. (ET).
         </p>
       );
     }
--- a/src/js/rx/containers/History.jsx
+++ b/src/js/rx/containers/History.jsx
@@ -166,7 +166,9 @@ class History extends React.Component {
       content = (
         <p className="rx-tab-explainer rx-loading-error">
           We couldn't retrieve your prescriptions.
-          Please refresh this page or try again later.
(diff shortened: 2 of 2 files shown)


src/js/rx/containers/Active.jsx
```diff
@@ -72,7 +72,9 @@ class Active extends React.Component {
       content = (
         <p className="rx-tab-explainer rx-loading-error">
           We couldn't retrieve your prescriptions.
-          Please refresh this page or try again later.
+          Please refresh this page or try again later. <br/>
```
akainic: Without running this locally and seeing what styles these classes are actually adding, it seems like what you're trying to create with the `<br/>` is essentially 2 separate paragraphs. Why not wrap each paragraph in its own `<p>` instead? It's slightly more semantic this way and using HTML tags they way they should be used. This means you'll have something like:
```
<div>
  <p className="rx-tab-explainer rx-loading-error">
    We couldn't retrieve your prescriptions. Please refresh this page or try again later.
  </p>
  <p className="rx-tab-explainer rx-loading-error">
    If this problem persists, please call the Vets.gov Help Desk at ....
  </p>
</div>
```
The wrapping `div` will be necessary because React doesn't like sibling tags without being wrapped. 

If this isn't possible or becomes tricky because of how things are styled no worries! Just a suggestion. Otherwise looks good to me!

akainic (approved):
One comment! Otherwise LGTM.

ayaleloehr:
This is why Alex doesn't play front end developer often. I will try this again and check back. thanks @akainic !

ayaleloehr:
Closing per new guidance in https://github.com/department-of-veterans-affairs/vets-website/issues/4204#issuecomment-264473632

PR #4207, opened 2016-12-02
Fix Rx error

### LINKED ISSUE #4204: Add Call center number for Rx error

ayaleloehr:
Second attempt to close #4204 

These are the only 3 places to use the rx-loading-error class, so this style change shouldn't impact anything else. Also putting screenshots this time:

Desktop:
screen shot 2016-12-02 at 10 41 56 am

Mobile:
screen shot 2016-12-02 at 10 43 40 am

@akainic you had helpful comments the first attempt, can you review this as well?

### DIFF
--- a/src/sass/rx/rx.scss
+++ b/src/sass/rx/rx.scss
@@ -90,6 +90,7 @@ button.usa-button-outline {
 
 .rx-loading-error {
   font-weight: bold;
+  width: 66%;
 }
 
 #rx-active, #rx-history, #rx-detail {
--- a/src/js/rx/containers/Active.jsx
+++ b/src/js/rx/containers/Active.jsx
@@ -70,12 +70,12 @@ class Active extends React.Component {
       );
     } else {
       content = (
-          <p className="rx-tab-explainer rx-loading-error">
-            We couldn't retrieve your prescriptions.
-            Please refresh this page or try again later.
-            If this problem persists, please call the Vets.gov Help Desk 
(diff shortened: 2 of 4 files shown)


src/sass/rx/rx.scss
```diff
@@ -90,6 +90,7 @@ button.usa-button-outline {
 
 .rx-loading-error {
   font-weight: bold;
+  width: 66%;
```
akainic: Is this something you want on mobile as well? It's possible to make it 100% on mobile and then 66% on everything else, if you want.

akainic (commented):
Just one question about the width on mobile! Seems like it would be preferable to make separate widths for different screens, but up to you guys (might also want to pull in a designer for that decision).

src/sass/rx/rx.scss
ayaleloehr: Great point! Opened #4209 to ask this. In the mean time, I'm going to merge this, as not having the call center number really isn't ideal, so I'm good coming back to the width question to get the call center number in today.

PR #4252, opened 2016-12-05
Clean phone numbers of non digit characters before submitting

akainic:
Replicating this change in vets-website: https://github.com/department-of-veterans-affairs/healthcare-application/pull/305/files

### DIFF
--- a/src/js/common/model/veteran.js
+++ b/src/js/common/model/veteran.js
@@ -786,7 +786,7 @@ function veteranToApplication(veteran) {
         if (value.value === '') {
           return undefined;
         }
-        return value.value.replace(/[- )(]/g, '');
+        return value.value.replace(/[^\d]/g, '');
 
       default:
         // fall through.
--- a/src/js/common/model/veteran.js
+++ b/src/js/common/model/veteran.js
@@ -775,13 +775,18 @@ function veteranToApplication(veteran) {
       case 'cityOfBirth':
       case 'stateOfBirth':
       case 'email':
+        if (value.value === '') {
+          return undefined;
+        }
+        break;
+
       case 'homePhone':
       case 'mobilePhone':
       case 'spousePhone':
         if (value.value === '') {
           return undefined;
         }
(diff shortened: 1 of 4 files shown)


src/js/common/model/veteran.js
```diff
@@ -775,13 +775,18 @@ function veteranToApplication(veteran) {
...
+        break;
+
       case 'homePhone':
       case 'mobilePhone':
       case 'spousePhone':
         if (value.value === '') {
           return undefined;
         }
-        break;
+        return value.value.replace(/[- )(]/g, '');
```
markolson: Maybe change this to `/[^\d]/g` so it strips out any non-numeric characters?

markolson (changes requested):

src/js/common/model/veteran.js
akainic: Hm, thinking through if this will create any issues because the front-end strips out the same characters: https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/js/common/utils/validations.js#L127.

But we're validating for `^\d{10} so hypothetically we shouldn't be allowing through any non-digit characters anyway so it should be fine?

markolson: Right, but someone could do like `[555] 555-0101` and the `[555]` would get through with this regex, causing the schema validation to fail. I think we're probably fine letting them enter the number how they want, but then stripping it down to what we actually need.

akainic: Yep, makes sense, just think we want to use the same validation function everywhere, so might need to make this change elsewhere too.

akainic:
@markolson changed the replace function to clean out all non-digit characters. Would love for @jbalboni to take a look as well as this affects the common `Phone` component.

jbalboni (approved):
LGTM

markolson (approved):
LGTM. The validation files don't impact what's actually sent, so this is fine for EDU and HCA at least.

PR #4255, opened 2016-12-05
Brings functions that opens popups closer to user click

### LINKED ISSUE #4247: Pop up blocked on "verify your account" link
Steps to reproduce:
- Go to https://www.vets.gov/profile/
- Click Sign in
- Click Google login (already logged into this google account in this browser, so nothing more needed).  This takes you to the profile page logged in as LOA1. 
- Click "verify your account"

Pop up is blocked, as seen below. 

screen shot 2016-12-05 at 2 56 56 pm

This is on Chrome Version 54.0.2840.98 (64-bit) and has occurred both on and off the VA network.

crwallace:
this resolves #4247

### DIFF
--- a/src/js/common/components/RequiredLoginView.jsx
+++ b/src/js/common/components/RequiredLoginView.jsx
@@ -2,7 +2,7 @@ import React from 'react';
 import $ from 'jquery';
 
 import environment from '../helpers/environment.js';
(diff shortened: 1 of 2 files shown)


bshyong (approved):
LGTM

PR #4261, opened 2016-12-06
[Rx] Unit tests for prescriptions reducer.

U-DON:


bshyong (approved):
The tests are not passing but it looks like it's caught on just a linting error involving whitespace. The unit tests here LGTM

PR #4271, opened 2016-12-06
Updated date components for partial dates

### LINKED ISSUE #4073: do not require RAD
see photo

jbalboni:
Closes #4073 

This is what I have so far for allowing partial dates. Instead of adding more conditional logic to `DateInput`, I decided to copy it and create a simplified version that we can build on. My hope is that we can start with the least restrictive date component and build more restrictive ones on top of it, like this:

`ErrorableDate` - date component that does the bare minimum of validation and allows partial dates (unless you mark the field as required)
`ErrorableCurrentOrPastDate` - built on `ErrorableDate`, but uses `isValidDate` like the default option for `DateInput`
`ErrorableFutureDate` (example, haven't built it) - built on `ErrorableDate`, but includes validation to only allow future dates.
`ErrorableFullDate` (example, haven't built it) - built on `ErrorableDate`, but includes validation to make users enter all the fields or none of them (i.e. only allow complete dates)

And so on. I also switched the validation to take an object, so that we could write validation functions that return if the field is valid along with a specific message. Not super useful here, but I thought it'd be helpful as an example of the direction we could go in.

If we're ok with this approach for the components and validation I'll write some tests and do some more cleanup before making this PR officially ready for review.

### DIFF
--- a/src/js/common/components/form-elements/ErrorableCurrentOrPastDate.jsx
+++ b/src/js/common/components/form-elements/ErrorableCurrentOrPastDate.jsx
@@ -1,6 +1,9 @@
 import React from 'react';
+import moment from 'moment';
+
 import ErrorableDate from './ErrorableDate';
-import { isValidDateField, isDirtyDate } from '../../utils/validations';
+import { isDirtyDate } from '../../utils/validations';
+import { dateToMoment } from '../../utils/helpers';
 
 export default function ErrorableCurrentOrPastDate(props) {
   let validation = {
@@ -12,7 +15,7 @@ export default function ErrorableCurrentOrPastDate(props) {
   // the past date check, because if you want to override that check
   // you should just be using ErrorableDate
   if (isDirtyDate(props.date)) {
-    if (!isValidDateField(props.date)) {
+    if (dateToMoment(props.date).isAfter(moment().startOf('day'))) {
       validation = {
         valid: false,
         message: props.invalidMessage
--- a/src/js/common/components/form-elements/ErrorableCurrentOrPastDate.jsx
+++ b/src/js/common/components/form-elements/ErrorableCurrentOrPastDate.jsx
@@ -1,0 +1,30 @@
+import React from 'react';
+import ErrorableDate from './ErrorableDate';
+import { isValidDateField, isDirtyDate } from '../../utils/validations';
+
+export default function ErrorableCurrentOrPastDate(props) {
+  let validation = {
+    valid: true,
+    message: null
+  };
+
+  if (isDirtyDate(props.date)) {
+    if (!isValidDateField(props.date)) {
+      validation = {
+        valid: false,
+        message: props.invalidMessage
+      };
+    } else if (props.validation) {
+      validation = props.validation;
+    }
+  }
+  return (
+    <ErrorableDate
+        {...props}
+        validation={validation}/>
+  );
+}
+
+ErrorableCurrentOrPastDate.defaultProps = {
+  invalidMessage: 'Please provide a valid date in the past'
+};
--- a/src/js/common/components/form-elements/ErrorableCurrentOrPastDate.jsx
+++ b/src/js/common/components/form-elements/ErrorableCurrentOrPastDate.jsx
@@ -8,6 +8,9 @@ export default function ErrorableCurrentOrPastDate(props) {
     message: null
   };
 
+  // we're assuming any custom validation goes after
+  // the past date check, because if you want to override that check
+  // you should just be using ErrorableDate
   if (isDirtyDate(props.date)) {
     if (!isValidDateField(props.date)) {
       validation = {
--- a/src/js/common/components/form-elements/ErrorableDate.jsx
+++ b/src/js/common/components/form-elements/ErrorableDate.jsx
@@ -1,0 +1,168 @@
+import React from 'react';
+import _ from 'lodash';
+import { set } from 'lodash/fp';
+import moment from 'moment';
+
+import ErrorableSelect from './ErrorableSelect';
+import ErrorableNumberInput from './ErrorableNumberInput';
+
+import ToolTip from './ToolTip';
+
+import { isDirtyDate, isValidPartialDate, isNotBlankDateField } from '../../utils/validations';
+import { months, days } from '../../utils/options-for-select.js';
+
+/**
+ * A form input with a label that can display error messages.
+ *
+ * Props:
+ * `required` - boolean. Render marker indicating field is required.
+ * `validation` - object. Result of custom validation. Should include a valid prop and a message prop
+ * `label` - string. Label for entire question.
+ * `name` - string. Used to create unique name attributes for each input.
+ * `toolTipText` - String with help text for user.
+ * `tabIndex` - Number for keyboard tab order.
+ * `date` - object. Date value. Should have month, day, and year props
+ * `onValueChange` - a function with this prototype: (newValue)
+ */
+class ErrorableDate extends React.Component {
+  constructor() {
+    super();
+    this.handleChange = this.handleChange.bind(this);
+  }
+
+  componentWillMount() {
+    this.id = _.uniqueId('date-input-');
+  }
+
+  handleChange(path, update) {
+    let date = set(path, update, this.props.date);
+    if (!date.month.value) {
+      date = set('day.value', '', date);
+    }
+
+    this.props.onValueChange(date);
+  }
+
+  render() {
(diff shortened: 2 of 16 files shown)


alexose:
> Instead of adding more conditional logic to `DateInput`

Definitely agree with this approach.  `DateInput` is pretty overloaded as is.  On the other hand, I worry about code reuse-- Is this intended to live side by side with `DateInput`?

Also, thanks for tackling `isValidPartialDate`!

jbalboni:
@alexose My intent was that we would migrate uses of `DateInput` to `ErrorableCurrentOrPastDate` (or whatever the appropriate component is).

alexose:
That seems like a good plan.  Added the above issue for a rainy day.

jbalboni:
Cool. Most of them should get moved when we're updating the date fields that need to allow partials, we can use that issue to clean up anything left over.

akainic:
Haven't had a chance to thoroughly look through this yet, but I love the approach @jbalboni!

jbalboni:
@akainic @alexose This is ready for a full review now.

src/js/edu-benefits/components/military-history/MilitaryServiceTour.jsx
```diff
@@ -3,12 +3,25 @@ import React from 'react';
 import ErrorableCheckbox from '../../../common/components/form-elements/ErrorableCheckbox';
 import ErrorableTextInput from '../../../common/components/form-elements/ErrorableTextInput';
 import ErrorableTextarea from '../../../common/components/form-elements/ErrorableTextarea';
+import ErrorableDate from '../../../common/components/form-elements/ErrorableDate';
+import ErrorableCurrentOrPastDate from '../../../common/components/form-elements/ErrorableCurrentOrPastDate';
 import ExpandingGroup from '../../../common/components/form-elements/ExpandingGroup';
-import DateInput from '../../../common/components/form-elements/DateInput';
 
-import { validateIfDirtyDateObj, validateIfDirty, isNotBlank, isValidDateField, isValidDateRange } from '../../utils/validations';
```
akainic: Since we're no longer using `validateIfDirtyDateObj`, is it something we can delete from the `validation.js` file? If it's also not being used anywhere else of course.

src/js/edu-benefits/components/military-history/MilitaryServiceTour.jsx
```diff
@@ -29,23 +42,16 @@ export default class MilitaryServiceTour extends React.Component {
               name="serviceStatus"
               field={tour.serviceStatus}
               onValueChange={(update) => {onValueChange('serviceStatus', update);}}/>
-          <DateInput required
-              errorMessage="Please provide a valid date"
-              validation={validateIfDirtyDateObj(tour.dateRange.from, isValidDateField)}
+          <ErrorableCurrentOrPastDate required
               label="Start of service period:"
```
akainic: Nit: this has a `:` after it and the next label for end date doesn't.

src/js/common/components/form-elements/ErrorableCurrentOrPastDate.jsx
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+import moment from 'moment';
+
+import ErrorableDate from './ErrorableDate';
+import { isDirtyDate } from '../../utils/validations';
+import { dateToMoment } from '../../utils/helpers';
+
+export default function ErrorableCurrentOrPastDate(props) {
```
akainic: How does this access the props? I'm assuming these are the props of `ErrorableDate`?

akainic (commented):
Wow so much work! Awesome job.

Only thing I found in testing is when you add a period of service without an end date, the review states `0000` for the year.

screen shot 2016-12-07 at 2 33 36 pm

src/js/edu-benefits/components/military-history/MilitaryServiceTour.jsx
```diff
@@ -3,12 +3,25 @@ import React from 'react';
...
+import ErrorableCurrentOrPastDate from '../../../common/components/form-elements/ErrorableCurrentOrPastDate';
 import ExpandingGroup from '../../../common/components/form-elements/ExpandingGroup';
-import DateInput from '../../../common/components/form-elements/DateInput';
 
-import { validateIfDirtyDateObj, validateIfDirty, isNotBlank, isValidDateField, isValidDateRange } from '../../utils/validations';
+import { validateIfDirty, isNotBlank, isValidDateRange } from '../../utils/validations';
+import { isDirtyDate } from '../../../common/utils/validations';
 import ServicePeriodsReview from './ServicePeriodsReview';
 
+function validEndDate(tour) {
```
akainic: The only thing I'm a little worried about this is making it more complicated to just drop in components that are needed as we will need to add additional configuration outside of the just adding the component, if that makes sense.

akainic (commented):
One more comment.

test/common/components/form-elements/ErrorableDate.unit.spec.jsx
```diff
@@ -0,0 +1,69 @@
...
+    date.month.dirty = true;
+    date.day.dirty = true;
+
+    const tree = SkinDeep.shallowRender(
+      <ErrorableDate date={date} validation={{ valid: false, message: 'Test' }} onValueChange={(_update) => {}}/>);
+
+    expect(tree.everySubTree('.usa-input-error')).not.to.be.empty;
+    expect(tree.subTree('.usa-input-error-message').text()).to.equal('Test');
+  });
+});
```
alexose: Seems like there might also be room for a month/year test as well

alexose (commented):
This is looking great, @jbalboni.  I reckon this was a huge amount of work, too.

src/js/common/components/form-elements/ErrorableDate.jsx
```diff
@@ -0,0 +1,173 @@
...
+      // make sure the user has entered a minimally valid date
+      } else if (!isValidPartialDate(day.value, month.value, year.value)) {
+        isValid = false;
+        errorMessage = this.props.invalidMessage;
+      // show any custom validation for this field once we know it's at least partially valid
+      } else if (this.props.validation && !this.props.validation.valid) {
+        isValid = this.props.validation.valid;
+        errorMessage = this.props.validation.message;
+      } else {
+        isValid = true;
```
jbalboni: Nope, looks I should take that out.

src/js/common/components/form-elements/ErrorableCurrentOrPastDate.jsx
jbalboni: These are the props that are passed to `ErrorableCurrentOrPastDate. If you write React components using their function syntax, you get the props passed as a parameter, instead of using `this.props`.

src/js/edu-benefits/components/military-history/MilitaryServiceTour.jsx
jbalboni: I think I can change this around so it it's less complicated. This instance is more complicated mainly because it's not using a validateIfDirty-like helper function, which I can add.

src/js/edu-benefits/components/military-history/MilitaryServiceTour.jsx
jbalboni: Actually looks like we're using it in a bunch of places.

src/js/edu-benefits/components/military-history/MilitaryServiceTour.jsx
```diff
@@ -29,23 +30,19 @@ export default class MilitaryServiceTour extends React.Component {
...
-              month={tour.dateRange.from.month}
-              year={tour.dateRange.from.year}
+              date={tour.dateRange.from}
               onValueChange={(update) => {onValueChange('dateRange.from', update);}}/>
-          <DateInput required
-              errorMessage={isValidDateRange(tour.dateRange.from, tour.dateRange.to) ? 'Please provide a valid date' : 'Date separated must be after date entered'}
-              validation={validateIfDirtyDateObj(tour.dateRange.to, date => isValidDateRange(tour.dateRange.from, date))}
-              label="End of service period"
+          <ErrorableDate
+              validation={{
```
jbalboni: @akainic @alexose I changed how the validation works here. I think this is pretty simple. I also made it so you could pass in an array, so we can do multiple validations with their own messages more easily.

akainic: looks more clear to me!

akainic (approved):
LGTM @jbalboni!

PR #4275, opened 2016-12-06
Styling

### LINKED ISSUE #4034: Current As Of Field/Using Time Stamp
cc: @crwallace 

VSSC’s terminal point of data delivery is:

Server: vhacdwsql04.vha.med.va.gov\dev
DB: GIS_ETL
Table: VSSC.FacilityLocatorExport

The two columns of interest on that table are:

FacilityDataDate- The date at which names/phone numbers/addresses/lat&lon/op hours are current as of.
OutpatientServicesDataDate- The date at which data related to outpatient services are current as of (this is what you probably want to display on the site).
They should actually always be identical since the data are merged and sent from VSSC on the same date.

If these columns are not in the API that GeoBISL provides you, they can easily be added, I’m sure.  In fact, I understand that GeoBISL utilizes these fields to extract the latest data into their API, since no data are truncated in the VSSC.FacilityLocatorExport table to which I deliver VSSC’s VAST information.

### LINKED ISSUE #4184: Mobile Tab style tweak
Tab style still not consistent with other tabs on Vets.gov. Should have 0 rounded corners. 

Should look like this:

### LINKED ISSUE #4240: Replace back link with breadcrumb
We're aiming to replace back links in vets.gov and adding detailed information on breadcrumbs.

For facility detail page, could we have it so its:
Home > Facility locator > [Name of facility] 
Make sure unicode is `›` `‹` carats, not `<` `>`characters

bshyong:
- Resolves #4240; replace back links with breadcrumbs
- Resolves #4184; mobile tab styles
- Resolves #4034; add current as of field data to detail page

### DIFF
--- a/src/js/facility-locator/components/ServicesAtFacility.jsx
+++ b/src/js/facility-locator/components/ServicesAtFacility.jsx
@@ -141,7 +141,7 @@ class ServicesAtFacility extends Component {
 
     return (
       <div>
-        <p style={{ margin: '0 0 0.5em' }}>Services current as of <strong>{moment().subtract(9, 'd').format('MMMM D, YYYY')}</strong></p>
+        <p style={{ margin: '0 0 0.5em' }}>Services current as of <strong>{moment(services.last_updated).format('MMMM D, YYYY')}</strong></p>
         <div className="call-out clearfix">
           <div className="columns small-1">
             <h3><i className="fa fa-exclamation-circle"></i></h3>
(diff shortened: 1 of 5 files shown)


src/js/facility-locator/components/ServicesAtFacility.jsx
```diff
@@ -141,7 +141,7 @@ class ServicesAtFacility extends Component {
 
     return (
       <div>
-        <p style={{ margin: '0 0 0.5em' }}>Services current as of <strong>{moment().subtract(9, 'd').format('MMMM D, YYYY')}</strong></p>
+        <p style={{ margin: '0 0 0.5em' }}>Services current as of <strong>{moment(services.last_updated).format('MMMM D, YYYY')}</strong></p>
```
crwallace: 👍

crwallace (approved):
LGTM pending the fix of linting errors.

PR #4280, opened 2016-12-07
Resolves #4278 external link icon

### LINKED ISSUE #4278: "visit va.gov" link in header needs white external link icon
external link icon isn't showing up which makes it look like there's a big space before the period

image

bshyong:
Previous selector was too specific and overrode `background-image: none` on certain elements


crwallace (approved):
LGTM!

PR #4281, opened 2016-12-07
Consolidate general validations

akainic:
Consolidating HCA validations to use common general validations (`isValidRequiredField`, `isNotBlank`, etc.).


jbalboni (approved):
This looks good to me, but I won't have time until later to pull it down and try it, if you're looking for a more thorough review.

src/js/hca/components/veteran-information/MothersMaidenName.jsx
```diff
@@ -2,7 +2,8 @@ import React from 'react';
 import _ from 'lodash';
 
 import ErrorableTextInput from '../../../common/components/form-elements/ErrorableTextInput';
-import { isValidName, isBlank, isValidRequiredField } from '../../utils/validations';
+import { isBlank, isValidRequiredField } from '../../../common/utils/validations';
+import { isValidName } from '../../utils/validations';
```
crwallace: is this part of the incremental change? We don't want to add this to the common validations file because other apps might have different name restrictions? Just wondering why this wasn't in the common validations file.

crwallace (approved):
This looks good to me with one question. I might suggest that @webinista also review this because it touches secure messaging.

src/js/hca/components/veteran-information/MothersMaidenName.jsx
akainic: Yep, part of the incremental change! The way I'm breaking up these changes is first doing general validation functions like `isBlank` and `isValidRequiredField`, and then doing more field specific ones, which I started in this PR: https://github.com/department-of-veterans-affairs/vets-website/pull/4292, so it will move to common.

akainic:
@ayaleloehr could you recommend someone to us that could test secure messaging on this branch to make sure everything still works correctly? I know @webinista is out this week and don't want this to be held up until next week. Thanks!

akainic:
@webinista the only changes here that might affect Rx and SM are in the common/utils/validations.js file, but even then I don't think it would affect anything. Wanted your review just in case. The rest of the files changed are education and HCA specific, so shouldn't need to review those.

webinista:
Looks good to me. Like you said, it doesn't appear to affect our stuff.

akainic:
@jbalboni are you good if we merge this PR now? Tested both HCA and education benefits again locally since we merged other PRs into this one and both are working as expected.

jbalboni:
@akainic Yep, go for it!

PR #4292, opened 2016-12-07
Consolidate field validations

akainic:


src/js/common/utils/validations.js
```diff
@@ -95,7 +95,7 @@ function isValidDateOver17(day, month, year) {
 }
 
 function isValidName(value) {
-  return /^[a-zA-Z][a-zA-Z '\-]*$/.test(value);
```
jbalboni: Why did this change?

jbalboni (approved):
Tried hca and edu locally and didn't run into any validation issues. Had one question below, but otherwise it looks good.

src/js/common/utils/validations.js
akainic: It looked like the extra `[a-zA-Z]` was redundant? HCA was using a validation for `isValidName` without it so I removed it remove the common.

jbalboni: I think the extra `[a-zA-Z]` means the string has to at least have one letter at the start. With the new regex, the name could start with ` '\-` or be empty.

akainic: Ah, I see. Oops! I'll add that back in.

akainic:
@jbalboni would you mind taking a second look at this? I noticed I had a legit failing test, so I added some code to `GrowableTable.unit.spec.jsx` in order for the test to pass. The test failed because of a change to `GrowableTable` that I would also love your review on: we were setting `isValidSection` as a default prop, and passing it the `isValidSection` validation function from `/common/utils/validation.js`. However, the `isValidSection` validation function in the common file is specific to HCA. It didn't seem like a good idea to set as a default prop on a common component a validation function that is specific to HCA, so I changed it to not being set as the default, and not requiring that prop anymore. Want to make sure this was the right thing to do.

Update: now noticing another failing test. Will have to tackle that next.

src/js/hca/components/ReviewCollapsiblePanel.jsx
```diff
@@ -4,7 +4,7 @@ import { connect } from 'react-redux';
 import _ from 'lodash';
 
 import { ensureFieldsInitialized, updateIncompleteStatus, updateVerifiedStatus, updateCompletedStatus } from '../actions';
-import * as validations from '../../common/utils/validations';
+import { isValidSection } from '../utils/validations';
```
jbalboni: ++ for changing this. I'm not a big fan of the `* as validations` syntax.

src/js/common/components/form-elements/GrowableTable.jsx
```diff
@@ -284,7 +283,7 @@ GrowableTable.propTypes = {
   onRowsUpdate: React.PropTypes.func.isRequired,
   path: React.PropTypes.string.isRequired,
   rows: React.PropTypes.array.isRequired,
-  isValidSection: React.PropTypes.func.isRequired,
+  isValidSection: React.PropTypes.func,
```
jbalboni: I think this prop should still be required. This is the main validation function for the GrowableTable component and I believe it's always being passed to the components. If you left it and isValidRow off a GrowableTable component, the validation would definitely not work correctly.

jbalboni:
@akainic Yes, I think that was the right change to make, though like I mentioned above I think isValidSection is still required for any GrowableTable. Making isValidSection default to the HCA version was a stopgap so we could use it in edu without changing HCA (since it was just importing the HCA version directly before).

akainic:
@jbalboni Added `isValidSection` prop back as required! If you wouldn't mind taking one more look, I can then merge this into the other PR, do thorough testing there, and then we can merge that one into master. In the meantime working on education benefits validations.

PR #4293, opened 2016-12-08
Capitalization; addresses #3285

### LINKED ISSUE #3285: Add ability to override breadcrumbs for specific path sections
In the Jekyll iteration of vets.gov, there was functionality that enabled custom overrides of the default capitalize the folder name rule. 

For examples, see config > site.js

Current use case: 
Home > Vocational Rehab and Employment (VR&E)

rather than folder name capitalization: 
Home > Vocational rebhab and employment

bshyong:
- resolves #3285

### DIFF
--- a/content/includes/breadcrumbs.html
+++ b/content/includes/breadcrumbs.html
@@ -3,17 +3,24 @@
     <li><a href="/">Home</a></li>
       {% assign my_variable = breadcrumb_path | last %}
       {% for crumb in breadcrumb_path %}
+        {% assign crumb_words = crumb.name | split: '-' %}
+        {% capture titlecase_crumb %}
+          {% for word in crumb_words %}
+            {{ word | capitalize }}
+          {% endfor %}
+        {% endcapture %}
+
         {% if my_variable.name == crumb.name %}
           <li class="active">
             {% if title %}
               {{ title }}
             {% else %}
-              {{ crumb.name | replace: '-', ' ' | capitalize }}
+              {{ titlecase_crumb }}
             {% endif %}
           </li>
         {% else %}
           <li><a href="/{{ crumb.path }}">
-            {{ crumb.name | replace: '-', ' ' | capitalize }}
+            {{ titlecase_crumb }}
           </a></li>
         {% endif %}
       {% endfor %}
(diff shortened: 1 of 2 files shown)


bshyong:
@crwallace after spending time to look at this, for the individual pages, the breadcrumbs can be overridden with the `title` attribute. However, the parent pages cannot: only the specific page that is being rendered is in scope; the parent directories are not in scope and therefore cannot be modified. 

This PR makes every word in the breadcrumb/directory name capitalized instead of only the first word, but does not let one override the name. We _may_ be able to override things by making a [custom filter](https://github.com/superwolff/metalsmith-layouts/issues/6#issuecomment-121742338) that checks for matches that we need to replace but I couldn't get it working with metalsmith + liquid.

crwallace:
I think we want to change the way that the breadcrumbs are generated here. Right now the bread crumbs are generated from the url and I think we want to change them to be generated by the title attribute all the way up the stream because some of the breadcrumbs will change downstream if they are just generated by the url.

**For example:**

Currently this is the landing page for education benefits:
`Home > Education Benefits` 
Then the next level in is:
`Home > Education > GI Bill`
Then the next level is:
`Home > Education > Gi bill > Post-9/11 GI Bill`

I think we want this to look this:
`Home > Education Benefits > GI Bill > Post-9/11 GI Bill`

_Note: This feedback is based on a conversation that I had with @suzchap. cc: @maryannbrody // @bethpotts_

bshyong:
@crwallace we are in agreement on the goal here. I'm saying that with our current pipeline, there is _no visibility_ into the parent title attributes. When the files are processed, the only data available are the current file's attributes and it's file directory path. `metalsmith-navigation` looks at directory paths to [generate the links](https://github.com/unstoppablecarl/metalsmith-navigation/blob/master/lib/index.js#L505)

I will keep doing some research on this but it seems we'll need a more extensive solution than I thought. We might even need to take out `metalsmith-navigation` and use something else (or write our own?). Let me know if you have any thoughts on this—I'll post more findings after some more digging.

bshyong:
@crwallace I did a bunch more investigating and trying out some tweaks in our pipeline today—I can't get a way to override parent files/directory names (upstream) with our current set up using Liquid templates. I think the best we can do in the meantime is to rename directories/files to match what we want displayed in the breadcrumbs as a quick fix. The child page titles can be overridden with metadata attributes, but parents cannot.  We can discuss technical details behind our breadcrumb/template setup if you are interested.

cc @rroueche due to potential content team tooling changes that could impact this issue

crwallace:
@bshyong -- My concerns with renaming are bookmarks and SEO. I think this might need to be something that gets addressed in our new content storage discovery.

bshyong:
^ Good catch on SEO/bookmarks. I will close this PR in the meantime. All it does is ensure that all words in the breadcrumbs are capitalized (not just the first word).

Issue #4226, opened 2016-12-05
Review page text treatment should use regular/bold instead of blue/grey


goldenmeanie:
- [X] The review page lists the labels and values for all questions in the accordions. The label should change to regular weight text-grey and the values should be bold weight text-grey. 

- [ ] (While you are in there, consider a min-width for the value column so that unanswered question labels break the same way as answered.)

cweber:
@goldenmeanie How much do you love tables? ....cause those tables need love. Min-width is the least of it.

![screen shot 2016-12-06 at 6 07 24 pm](https://cloud.githubusercontent.com/assets/417657/20947815/f29117d2-bbde-11e6-936c-0427045859ae.png)

PR #4296, opened 2016-12-08
Edu review page style changes

### LINKED ISSUE #4221: Edit buttons have inconsistent treatment
The edit buttons should be bold and without icons.

Currently:

cweber:
This closes #4221 and #4226.

Removed pencil icon from remaining edit buttons and changed label/**data** styling in tables to be more consistent with vets.gov design patterns.

screen shot 2016-12-08 at 9 29 58 am

### DIFF
--- a/src/sass/_shame.scss
+++ b/src/sass/_shame.scss
@@ -69,18 +69,6 @@ body .row {
   }
 }
 
-.review {
-  tbody {
-    tr {
-      td {
-        &:first-child {
-          color: $color-primary-darker;
-        }
-      }
-    }
-  }
-}
-
 .hidden {
   display: none !important;
 }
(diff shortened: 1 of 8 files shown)


src/sass/_shame.scss
```diff
@@ -69,18 +69,6 @@ body .row {
   }
 }
 
-.review {
```
jbalboni: Is this used in the static pages (the content folder) anywhere?

jbalboni (approved):
LGTM. Had one question.

src/sass/_shame.scss
cweber: Good question. I can't find it in any of the static pages. It looks like it was written by Courtney during the HCA buildout. (Not that it couldn't have been reused.)

jbalboni: Does this change affect HCA, then?

cweber: Uh. Yes. Did I merge too soon?

cweber: So yes. Deleting that line removed the blue color from the :firstchild in both HCA and Edu. It looks like I added the inverse fix to a slightly more specific place that is only affecting Edu. It SHOULD get fixed in both places. @goldenmeanie can take care of that in the branch he's currently working on.

Sorry I jumped the gun. The interim state is fine in the short term.

PR #4303, opened 2016-12-08
Fix validation issue for service graduation year

### LINKED ISSUE #4302: validation error in serviceAcademyGraduationYear
`Form The property '#/serviceAcademyGraduationYear' did not have a minimum value of 1900, inclusively in schema 6d3ac55d-0966-5868-be2d-0f1191776ce9#`
happened on Dec 8, 2016 5:22:28 AM UTC

jbalboni:
Closes #4302.

Validation was letting invalid years through on serviceGraduationYear field.

### DIFF
--- a/src/js/edu-benefits/utils/validations.js
+++ b/src/js/edu-benefits/utils/validations.js
@@ -268,7 +268,7 @@ function isValidTourOfDuty(tour) {
 }
 
 function isValidMilitaryServicePage(data) {
-  return !data.chapter33 || isNotBlank(data.benefitsRelinquished.value);
+  return isBlank(data.serviceAcademyGraduationYear.value) || isValidYear(data.serviceAcademyGraduationYear.value);
 }
 
 function isValidServicePeriodsPage(data) {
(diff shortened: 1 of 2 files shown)


src/js/edu-benefits/utils/validations.js
```diff
@@ -268,7 +268,7 @@ function isValidTourOfDuty(tour) {
 }
 
 function isValidMilitaryServicePage(data) {
-  return !data.chapter33 || isNotBlank(data.benefitsRelinquished.value);
+  return isBlank(data.serviceAcademyGraduationYear.value) || isValidYear(data.serviceAcademyGraduationYear.value);
```
akainic: The MilitaryServicePage is using `isValidCurrentorPastYear` for the validation on the `serviceAcademyGraduationYear` itself, so we should use the same validation here too so it's consistent.

akainic (commented):
One small fix

jbalboni:
@akainic Ready to be looked again

akainic (approved):
LGTM!

PR #4307, opened 2016-12-08
Add max future date check

### LINKED ISSUE #4239: Date validation
Schema validation error occurred when someone tried to enter `20016` for their year. We should prevent years past a certain point (+100 years from current date, perhaps?)

jbalboni:
Closes #4239

Adds a plus 100 year check to our dates that allow future dates.

screen shot 2016-12-08 at 3 51 51 pm

### DIFF
--- a/src/js/common/utils/validations.js
+++ b/src/js/common/utils/validations.js
@@ -106,6 +106,10 @@ function isValidDate(day, month, year) {
 }
 
 function isValidAnyDate(day, month, year) {
+  if (Number(year) < 1900 || Number(year) > moment().add(100, 'year').year()) {
+    return false;
+  }
+
   return moment({
     day,
     month: month ? parseInt(month, 10) - 1 : month,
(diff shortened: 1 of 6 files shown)


src/js/common/utils/validations.js
```diff
@@ -122,6 +126,10 @@ function isValidPartialDate(day, month, year) {
 }
 
 function isValidDateOver17(day, month, year) {
+  if (Number(year) < 1900) {
```
akainic: do we not want to also add the +100 check for this too?

akainic (commented):
Wondering if we should have a separate validation function called `isValidYear` that first checks that the year is not before 1900 or after today+100, since that seems to be relevant to every date validation function, and we can call that initially within each of the specific date validation functions you've changed here. Might just make things more consistent. Thoughts?

src/js/common/utils/validations.js
jbalboni: Since it also has to be 17 years in the past, the 100 check isn't necessary.

jbalboni:
@akainic Yeah, an `isValidYear` function sounds like a good idea. We might have that already, actually, and I should be using it.

jbalboni:
@akainic Made the updates, looks better now!

akainic (approved):
LGTM!

PR #4319, opened 2016-12-09
Allow partial dates where applicable on edu benefits form

### LINKED ISSUE #4085: Figure out how to handle partial dates for education
Because there are no specific data requirements for the spool file for non-required fields, should we allow partial dates to be submitted? Currently only complete dates (having a month, day, and year) can be submitted. Partial dates will be stripped out and nothing will be submitting.

Tagging @jbalboni for awareness and to add further context.

jbalboni:
Closes #4085

This updates the fields listed in the above issue to be allow partial dates as required. It also converts some dates to a new month/year component. Most of the work here is updating validation; the other changes are largely the copy and pasted month year component and tests.

screen shot 2016-12-09 at 12 28 38 pm
screen shot 2016-12-09 at 12 28 52 pm
screen shot 2016-12-09 at 12 24 17 pm
screen shot 2016-12-09 at 12 25 26 pm

### DIFF
--- a/src/js/common/components/form-elements/ErrorableMonthYear.jsx
+++ b/src/js/common/components/form-elements/ErrorableMonthYear.jsx
@@ -1,0 +1,170 @@
+import React from 'react';
+import _ from 'lodash';
+import { set } from 'lodash/fp';
+import moment from 'moment';
+
+import ErrorableSelect from './ErrorableSelect';
+import ErrorableNumberInput from './ErrorableNumberInput';
+
+import ToolTip from './ToolTip';
+
+import { isDirtyDate, isValidPartialDate, isNotBlankDateField, validateCustomFormComponent } from '../../utils/validations';
+import { months, days } from '../../utils/options-for-select.js';
+
+/**
+ * A form input with a label that can display error messages.
+ *
+ * Props:
+ * `required` - boolean. Render marker indicating field is required.
+ * `validation` - object. Result of custom validation. Should include a valid prop and a message prop
+ * `label` - string. Label for entire question.
+ * `name` - string. Used to create unique name attributes for each input.
+ * `toolTipText` - String with help text for user.
+ * `tabIndex` - Number for keyboard tab order.
+ * `date` - object. Date value. Should have month, day, and year props
+ * `onValueChange` - a function with this prototype: (newValue)
+ */
+class ErrorableDate extends React.Component {
+  constructor() {
+    super();
+    this.handleChange = this.handleChange.bind(this);
+  }
+
+  componentWillMount() {
+    this.id = _.uniqueId('date-input-');
+  }
+
+  handleChange(path, update) {
+    let date = set(path, update, this.props.date);
+    if (!date.month.value) {
+      date = set('day.value', '', date);
+    }
+
+    this.props.onValueChange(date);
+  }
+
+  render() {
+    const { day, month, year } = this.props.date;
+
+    let daysForSelectedMonth = [];
+    if (month.value) {
+      daysForSelectedMonth = days[month.value];
+    }
+
+    // we want to do validations in a specific order, so we show the message
+    // that makes the most sense to the user
+    let isValid = true;
+    let errorMessage;
+    if (isDirtyDate(this.props.date)) {
+      // make sure the user enters a full date first, if required
+      if (this.props.required && !isNotBlankDateField(this.props.date)) {
+        isValid = false;
+        errorMessage = this.props.requiredMessage;
+      // make sure the user has entered a minimally valid date
+      } else if (!isValidPartialDate(day.value, month.value, year.value)) {
+        isValid = false;
+        errorMessage = this.props.invalidMessage;
+      } else {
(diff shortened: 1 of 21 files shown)


akainic (commented):
Haven't tested it locally and gone through the list @cweber provided of the different dates, but wanted to at least add my code comments first! Testing and additional general commentary forthcoming.

src/js/common/components/form-elements/ErrorableMonthYear.jsx
```diff
@@ -0,0 +1,146 @@
...
+    if (this.props.toolTipText) {
+      toolTip = (
+        <ToolTip
+            tabIndex={this.props.tabIndex}
+            toolTipText={this.props.toolTipText}/>
+      );
+    }
+
+    return (
+      <div>
```
jbalboni: Yes, I should.

src/js/common/utils/validations.js
```diff
@@ -129,6 +131,26 @@ function isValidPartialDate(day, month, year) {
   return true;
 }
 
+function isValidMonthYear(month, year) {
+  if (typeof month === 'object') {
```
jbalboni: I kept passing a date field into this function (instead of a month and year) and having confusing validation output as a result. We don't have to have this, but seemed useful.

src/js/edu-benefits/components/military-history/ContributionsFields.jsx
```diff
@@ -65,23 +61,19 @@ export default class ContributionsFields extends React.Component {
               checked={this.props.data.activeDutyRepaying}
               onValueChange={(update) => {this.props.onStateChange('activeDutyRepaying', update);}}/>
           <div>
-            <DateInput required
-                errorMessage="Please provide a response"
-                validation={validateIfDirtyDateObj(this.props.data.activeDutyRepayingPeriod.from, isValidDateField)}
+            <ErrorableCurrentOrPastDate required
```
jbalboni: Correct.

src/js/edu-benefits/components/military-history/ContributionsFields.jsx
```diff
@@ -65,23 +61,19 @@ export default class ContributionsFields extends React.Component {
...
-                year={this.props.data.activeDutyRepayingPeriod.from.year}
+                date={this.props.data.activeDutyRepayingPeriod.from}
                 onValueChange={(update) => {this.props.onStateChange('activeDutyRepayingPeriod.from', update);}}/>
-            <DateInput required
-                errorMessage={isToDateValid ? 'End Date must be after Start Date' : 'Please provide a response'}
-                validation={validateIfDirtyDateObj(this.props.data.activeDutyRepayingPeriod.to, (date) => isValidDateRange(this.props.data.activeDutyRepayingPeriod.from, date))}
+            <ErrorableCurrentOrPastDate required
+                validation={{
+                  valid: isValidDateRange(this.props.data.activeDutyRepayingPeriod.from, this.props.data.activeDutyRepayingPeriod.to),
+                  message: 'End date must be after Start date'
```
jbalboni: I think I did that because it's a field name, but it seems generic enough that it's not necessary.

akainic:
Tested locally and everything works as expected based on the code. I'm still confused by allowing partial dates for MonthYear questions. To me it doesn't make sense to allow a month without a year (might as well not provide anything). But curious if @cweber has more information about this.

jbalboni:
@akainic I am ambivalent about allowing just a month. Similarly, we allow just day dates. I think the main benefit is just that we reduce the places where the user has to go back and fix info they put in fields we don't require them to fill out in the first place.

cweber:
Agreed, a month alone seems silly and I think I'm ok with treating it however we want when we submit it, but for the user i think it's important to create as little friction as possible. If we aren't requiring that field, we should accept whatever they leave in it.

For Education Benefits specifically, the data is reviewed manually, so even though month alone is not particularly useful, I don't think it hurts to include it in the spool file.

akainic:
@cweber Okay that's fair. We should probably at least let Shay know we're making these changes, as it will change the data the field might receive, and we don't want them to think something is broken once they start receiving partial date information.

akainic (approved):
@jbalboni the updates you made to the PR LGTM! A couple general thoughts I have that do not prevent merging:
1. Do these new date components mean we can replace DateInput? If so, we should probably move everything to using these new ones and remove DateInput. Future issue.
2. In terms of naming, wondering if it would be helpful to name the components with `ErrorableDate` in front of each of them (e.g., `ErrorableDateMonthYear`, `ErrorableDateCurrentOrPast`). Couple of reasons why: 1) indicates up front that these are variations on the essential `ErrorableDate` component, 2) makes it easier when searching through all of the components to see all of the date-related components grouped together. Doesn't have to be changed here, but curious to get your thoughts on this.

jbalboni:
@akainic For 1, we are completely off of DateInput in edu-benefits with this PR. For 2, yeah that reads much better and I'm not sure why I didn't do it that way initially. :)

Issue #4316, opened 2016-12-09
investigate building multiple forms in a single react app


leannammiller:
output of this will be PR

leannammiller:
PR to review - seen in Github 

@akainic and anyone who is interested in FE, @lihanli @markolson

leannammiller:
@aub

leannammiller:
@jbalboni - is this correct?

PR #4320, opened 2016-12-09
Multiple forms in one React app

jbalboni:
Related to #4316.

The diff might be hard to read, but this is one approach to writing separate forms in one React app. The current form is in the 1990 folder now with minimal changes. The entry jsx file now has one component and dynamically loads whichever form you choose when you go to the route for it (the current form is at `/1990/introduction` now). 

The changes are primarily to prefix routes and the Redux state object with a `1990` namespace so we can have multiple of them. Any of the code in the 1990 folder could be pulled into the main edu-benefits folder and reused.

The Redux part of this could be done differently. We could have each form have its own Redux store and only create it when you start the form (instead of when you see the questions).


aub:
Conceptually, this makes a lot of sense to me, and it looks well done. I wondered about the files like `1990/containers/EduBenefits.jsx`, where they explicitly look in the state object for the `form1990` section. Shouldn't those guys just have that section passed to them directly as the state so they don't have to know about this structure?

Not sure what else I can really add here... is this in support of a real feature, or meant to remain an experiment?

jbalboni:
@aub Yeah, that makes sense. I think we'd want to move some of the components in the 1990 folder into a place common to any form and we'd likely want to pass things like the state property to them at that point.

This is in support of building the next edu form (1995), but not sure if we'll want to merge this particular PR.

src/js/edu-benefits/1990/containers/EduBenefitsApp.jsx
```diff
@@ -95,10 +95,10 @@ class EduBenefitsApp extends React.Component {
 
 function mapStateToProps(state, ownProps) {
   return {
-    pageState: state.uiState.pages,
+    pageState: state.form1990.uiState.pages,
```
akainic: Is the idea that all education forms will have the same store, and within the store it will be divided between each form? This seems like another thing that would be nice if it were automatically configured.

akainic (commented):
This is an awesome start! In order to actually be able to merge this, I wonder if it would be helpful to keep all existing 1990 files as is, and then just copy them into a new directory under `/1990/`, instead of renaming them as you've done here. That way we can add the path `education/1990/` to `ignoreList` and incrementally work on this while being able to merge to master.

akainic:
@jbalboni is this PR still relevant or is it replaced by #4471?

PR #4321, opened 2016-12-09
Scroll to errors on edu benefits form

### LINKED ISSUE #3330: Check scrolling on form submissions with errors

jbalboni:
Closes #3330 

This scrolls to the first invalid input element and sets focus.

### DIFF
--- a/src/js/common/components/NavButtons.jsx
+++ b/src/js/common/components/NavButtons.jsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import Scroll from 'react-scroll';
 
-import { getActivePages } from '../utils/helpers';
+import { getActivePages, focusElement } from '../utils/helpers';
 
 import ProgressButton from '../../common/components/form-elements/ProgressButton';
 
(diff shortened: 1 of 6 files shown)


jbalboni:
@cweber @goldenmeanie Can one of you try this out and make sure it works the way you expect?

akainic (commented):
This looks great! One thing I noticed is that on date fields it scrolls past the error message. Maybe we can make an exception for if it's a date field and handle that differently?

screen shot 2016-12-09 at 3 26 32 pm

cweber:
LGTM. 

One thing I noticed, but I don't think needs to be fixed: 
When the error that triggers the scroll is a date component, the alignment with the top of the screen is weird. The label for that component gets cut off. I'm guessing it has something to do with being a more complex component.

Samesies Anne!

jbalboni:
I think we can some top margin to the scroll position to capture the error message. Should still look fine for other fields.

jbalboni:
@cweber @akainic Added code to handle dates specifically, so this should work better now.

akainic (approved):
LGTM! Thanks for adding the handling of date inputs differently. Looks awesome.

PR #4336, opened 2016-12-12
Fix two digit year validation on education range

### LINKED ISSUE #4334: daterange validation error
Form The property '#/postHighSchoolTrainings/0/dateRange/from' value "96-01-XX" did not match the regex '^(\d{4}|XXXX)-(0[1-9]|1[0-2]|XX)-(0[1-9]|[1-2][0-9]|3[0-1]|XX)$' in schema 2c174352-90a4-5edd-988b-b2d311f687d5#, Form The property '#/postHighSchoolTrainings/0/dateRange/to' of type String did not match any of the required schemas. The schema specific errors were:

- oneOf #0:
    - The property '#/postHighSchoolTrainings/0/dateRange/to' value "98-04-XX" did not match the regex '^(\d{4}|XXXX)-(0[1-9]|1[0-2]|XX)-(0[1-9]|[1-2][0-9]|3[0-1]|XX)$'
- oneOf #1:
    - The property '#/postHighSchoolTrainings/0/dateRange/to' value "98-04-XX" did not match one of the following values: present

happened on Dec 12, 2016 4:26:53 PM UTC

jbalboni:
Closes #4334.

There were two failures for the above validation issue. 1) We were not properly validating the education period and 2) the date formatting was passing two digit years through. This fixes both.

### DIFF
--- a/src/js/edu-benefits/utils/validations.js
+++ b/src/js/edu-benefits/utils/validations.js
@@ -291,7 +291,7 @@ function isValidEmploymentHistoryPage(data) {
 }
 
 function isValidEducationPeriod(data) {
-  return isValidDateRange(data.dateRange.from, data.dateRange.to);
+  return isValidDateField(data.dateRange.from) && isValidDateField(data.dateRange.to) && isValidDateRange(data.dateRange.from, data.dateRange.to);
 }
 
 function isValidEducationHistoryPage(data) {
@@ -445,5 +445,6 @@ export {
   isValidRelinquishedDate,
   isValidTourOfDuty,
   isValidEmploymentPeriod,
-  isValidRotcScholarshipAmount
+  isValidRotcScholarshipAmount,
+  isValidEducationPeriod
 };
--- a/src/js/edu-benefits/utils/validations.js
+++ b/src/js/edu-benefits/utils/validations.js
@@ -291,7 +291,9 @@ function isValidEmploymentHistoryPage(data) {
 }
 
 function isValidEducationPeriod(data) {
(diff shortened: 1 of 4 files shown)


lihanli (approved):

test/edu-benefits/utils/helpers.unit.spec.js
```diff
@@ -79,5 +79,14 @@ describe('edu helpers:', () => {
 
       expect(formatPartialDate(date)).to.be.undefined;
     });
+    it('should format a partial year', () => {
```
akainic: Is this really able to validate dates with 2 digit years? When I test this locally it doesn't seem to work for me. A year of '07' is invalid.

src/js/edu-benefits/utils/validations.js
```diff
@@ -291,7 +291,9 @@ function isValidEmploymentHistoryPage(data) {
 }
 
 function isValidEducationPeriod(data) {
-  return isValidDateRange(data.dateRange.from, data.dateRange.to);
+  return (isBlankDateField(data.dateRange.from) || isValidDateField(data.dateRange.from))
```
akainic: I'm a little confused about this fix. So we're allowing the from date and/or the to date to be totally blank? Is this required for the schema error fix?

akainic (commented):
Tested this locally and am not able to submit a 2 digit year for education period range. Must be missing something but not sure what!

screen shot 2016-12-12 at 3 50 41 pm

akainic (approved):
LGTM! Had my questions answered via Slack. Sounds like we don't want to allow 2 digit years (which is what can happen now) and the partial dates PR that is forthcoming will allow partial dates to be entered for the date range on educational period.

PR #4350, opened 2016-12-13
Removing flex value from message search button.

webinista:
Closes department-of-veterans-affairs/messaging-fe#157.


U-DON:
LGTM.

PR #4351, opened 2016-12-13
Removed messaging-modal-title class.

webinista:
- Changes color of modal title to match default text instead of link color
- Closes department-of-veterans-affairs/messaging-fe#197


U-DON:
LGTM.

U-DON:
Also, a little confused; where is the color coming from if it wasn't defined in `messaging-modal-title`?

webinista:
It's a mystery to me too. Inspector said it was in `messaging-modal-title`. Looking at the SCSS did not show the same. Removing the class name fixed it though.

PR #4352, opened 2016-12-13
Make message navigation accessible after an error occurs

bshyong:
Resolves https://github.com/department-of-veterans-affairs/messaging-fe/issues/196

### DIFF
--- a/src/js/messaging/containers/Folder.jsx
+++ b/src/js/messaging/containers/Folder.jsx
@@ -343,9 +343,9 @@ export class Folder extends React.Component {
             </p>
           </div>
         );
+      } else {
+        componentContent = <div className="columns"><p>Sorry, this folder does not exist.</p></div>;
       }
-
-      componentContent = <div className="columns"><p>Sorry, this folder does not exist.</p></div>;
     } else {
       const messageNav = this.makeMessageNav();
       const sortMenu = this.makeSortMenu();
--- a/src/js/messaging/containers/Folder.jsx
+++ b/src/js/messaging/containers/Folder.jsx
@@ -325,6 +325,7 @@ export class Folder extends React.Component {
     }
 
     const folderId = _.get(this.props.attributes, 'folderId', null);
+    let componentContent;
 
     if (folderId === null) {
       const lastRequest = loading.request;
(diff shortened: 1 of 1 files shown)


src/js/messaging/containers/Folder.jsx
```diff
@@ -334,33 +335,48 @@ export class Folder extends React.Component {
...
+            <p>
+              Could not retrieve the folder.&nbsp;
+              <a onClick={reloadFolder}>Click here to try again.</a>
+            </p>
+          </div>
         );
       }
 
-      return <p>Sorry, this folder does not exist.</p>;
+      componentContent = <div className="columns"><p>Sorry, this folder does not exist.</p></div>;
```
U-DON: Won't this always overwrite the content provided in the conditional above?

src/js/messaging/containers/Folder.jsx
```diff
@@ -334,33 +335,48 @@ export class Folder extends React.Component {
           this.props.fetchFolder(lastRequest.id, lastRequest.query);
         };
 
-        return (
-          <p>
-            Could not retrieve the folder.&nbsp;
-            <a onClick={reloadFolder}>Click here to try again.</a>
-          </p>
+        componentContent = (
+          <div className="columns">
```
U-DON: What's the `columns` class from?

bshyong: @U-DON it's a global class that adds side padding—usually used with column size classes like `small-12`, etc

Without this the content is up against the edges on mobile

src/js/messaging/containers/Folder.jsx
bshyong: good catch—should be in an else statement

U-DON:
LGTM.

src/js/messaging/containers/Folder.jsx
U-DON: I was just curious since I couldn't find it explicitly defined within the sass src. Is it from some package?

bshyong: I think it's from Foundation because the design team is using their grid. I usually use that if it's not worth defining our own container/div classes just for the left + right padding

PR #4353, opened 2016-12-13
Reduced padding for .messaging-notice-box.

webinista:
- Removed @media query in favor of same amount of padding for every breakpoint
- Closes department-of-veterans-affairs/messaging-fe#188


U-DON:
LGTM.

PR #4354, opened 2016-12-13
Increased padding-right value for va-body-alert. Closes #4170

### LINKED ISSUE #4170: Disclaimer / alert button overlaps text when viewport ≤ 375px
Adding `width: 90%` to `.va-alert h5, .va-title` will fix it 99% of cases.

screen shot 2016-11-30 at 2 26 04 pm

webinista:


src/sass/modules/_m-alert.scss
```diff
@@ -17,7 +17,7 @@
 }
 
 .va-alert-body {
-  padding-left: 2.5rem;
+  padding-left: 3.5rem;
```
U-DON: This looks like it's for `padding-left`, not `padding-right` as the issue indicates.

webinista: It does solve the issue though. `padding-right` would push the icon away from the edge of the box.

PR #4357, opened 2016-12-13
Basic fix to keep all text visible

### LINKED ISSUE #4297: <tds> aren't wrapping correctly in review tables
The second `<td>` in tables is not wrapping consistently in the review tables. In some cases it looks  'fine'. In others it's refusing to wrap and increasing the container size of whatever it's in.

Looks fine:
screen shot 2016-12-08 at 9 26 32 am

Does not look fine:
screen shot 2016-12-08 at 9 26 45 am

goldenmeanie:
This is a half-solution, since at small screen sizes the columns are very narrow. But a switch to Definition Lists is more complicated so let's get this fix out there and then look at the pattern site-wide.

Fixes #4297


akainic (approved):
LGTM! I agree this is a good temporary solution, until we figure out a better layout for some of the review sections.

PR #4360, opened 2016-12-13
Switch to h2 for better hierarchy/screenreading capabilities

molindoz:
Closes: [Sunsets #264](https://github.com/department-of-veterans-affairs/sunsets-team/issues/264)

I used the same `font-size`, `line-height`, and `font-weight` to keep the styling of the text the of same as an h4.

### DIFF
--- a/src/sass/disability-benefits.scss
+++ b/src/sass/disability-benefits.scss
@@ -662,6 +662,12 @@ h1:focus {
   outline: none;
 }
 
+h2.claim-h2 {
+  font-size: 1.35em;
+  line-height: 1.5;
+  font-weight: bold;
+}
+
 .va-nav-breadcrumbs:focus {
   outline: none
 }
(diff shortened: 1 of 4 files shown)


src/sass/disability-benefits.scss
```diff
@@ -662,6 +662,12 @@ h1:focus {
   outline: none;
 }
 
+h2.claim-h2 {
```
jbalboni: Does this need `h2` in the selector?

jbalboni (approved):
Looks good to me. Should we change the headers in the sidebar as well? They're the same underlined style as these.

src/sass/disability-benefits.scss
cagedcrown: Good call, I can remove the `h2` selector. I remember running into a problem with `h`tags being overwritten by the uswds defaults, but it works in this case. 👍

Issue #4142, opened 2016-11-29
Add breadcrumb to Rx


gnakm:
Follow up on conversation here: https://github.com/department-of-veterans-affairs/vets.gov-designpattrns/issues/29

Rx is currently missing a breadcrumb. We are wondering if the < Back to link is sustainable. Mel and I had a discussion on taking out the back link and adding a breadcrumb to all the pages. 

@melwoodard, Could you validate the following?
For Prescription refill and History pages: Home > Health Care > Prescription Refill 
For Detail page: Home > Health Care > Prescription Refill > [Prescription Name]
For Glossary page: Home > Health Care > Prescription Refill > Glossary

@webinista I don't know how crazy it would be to add all breadcrumbs but for launch, could we have 
For Prescription refill and History pages: Home > Health Care > Prescription Refill ?

Mocks below:
![rx refill page](https://cloud.githubusercontent.com/assets/13420618/20730807/7cad8606-b655-11e6-9f16-5141834a144b.png)

![view history](https://cloud.githubusercontent.com/assets/13420618/20730819/8d287e28-b655-11e6-86fa-3e2d0400f646.png)

![glossary page](https://cloud.githubusercontent.com/assets/13420618/20730787/67cdb2a6-b655-11e6-81d9-577a027e9462.png)

![rx detail page](https://cloud.githubusercontent.com/assets/13420618/20730828/962be924-b655-11e6-91b5-3e494aac9997.png)

melwoodard:
@gnakm logic looks good!

PR #4363, opened 2016-12-13
Resolve 4142 RX breadcrumbs

bshyong:
- Resolves https://github.com/department-of-veterans-affairs/vets-website/issues/4142


src/js/rx/components/Breadcrumbs.jsx
```diff
@@ -0,0 +1,45 @@
...
+      }
+    } else if (pathname.match(/\/glossary\/$/)) {
+      crumbs.push(<a href="/healthcare/prescriptions" key="prescriptions">Prescription Refills</a>);
+      crumbs.push(<span key="glossary"><strong>Glossary</strong></span>);
+    } else {
+      crumbs.push(<span key="prescriptions"><strong>Prescription Refills</strong></span>);
+    }
+
+    return (<div className="rx-breadcrumbs">
+      {[].concat(...crumbs.map(e => [' › ', e])).slice(1)}
```
U-DON: Could this be more cleanly expressed as `crumbs.join(' › ')`, or does that do some weird conversion?

src/js/rx/components/Breadcrumbs.jsx
```diff
@@ -0,0 +1,45 @@
...
+    const { pathname } = this.props.location;
+    const { prescription } = this.props;
+
+    const crumbs = [
+      <a href="/" key="home">Home</a>,
+      <a href="/healthcare" key="healthcare">Health Care</a>,
+    ];
+
+    if (pathname.match(/\/\d+$/)) {
+      crumbs.push(<a href="/healthcare/prescriptions" key="prescriptions">Prescription Refills</a>);
```
U-DON: This (and any route within the app) should probably be `Link` rather than `a`.

src/js/rx/components/Breadcrumbs.jsx
```diff
@@ -0,0 +1,45 @@
...
+  }
+}
+
+const mapStateToProps = (state) => {
+  return {
+    prescription: state.prescriptions.currentItem,
+  };
+};
+
+export default connect(mapStateToProps)(Breadcrumbs);
```
U-DON: I don't think `Breadcrumbs` needs to be a container (connected component). I see you've already pulled `prescription` from state in `RxRefillsApp`, so you could pass that prop into this component.

On a related note, we keep any containers within the `containers` directory.

src/js/rx/components/GlossaryPage.jsx
```diff
@@ -1,5 +1,4 @@
 import React from 'react';
-import BackLink from '../components/BackLink';
```
U-DON: If `BackLink` isn't being used anywhere anymore, let's remove the component too.

src/js/rx/containers/RxRefillsApp.jsx
```diff
@@ -1,13 +1,14 @@
...
-import RequiredLoginView from '../../common/components/RequiredLoginView';
 import { closeDisclaimer } from '../actions/disclaimer';
 import { closeRefillModal, closeGlossaryModal } from '../actions/modals';
 import { refillPrescription } from '../actions/prescriptions';
-import Disclaimer from '../components/Disclaimer';
+import Breadcrumbs from '../components/Breadcrumbs';
 import ConfirmRefillModal from '../components/ConfirmRefillModal';
+import Disclaimer from '../components/Disclaimer';
 import GlossaryModal from '../components/GlossaryModal';
+import RequiredLoginView from '../../common/components/RequiredLoginView';
```
U-DON: Nitpick: this import was originally higher because I wanted to put imports that had to go back more levels at the top.

U-DON (changes requested):

src/js/rx/components/Breadcrumbs.jsx
bshyong: `join()` converts the array into a string; we need an array of objects. it's difficult to read here but the code basically interpolates the elements with `' › '` and returns an array

src/js/rx/containers/RxRefillsApp.jsx
bshyong: Got it; I will reorder them. I usually sort them alphabetically so if there's another ordering that should take precedence let me know and I will keep it in mind

src/js/rx/components/Breadcrumbs.jsx
bshyong: @U-DON I thought about this and the only benefit that `Link` provides is that it is router-aware. In this case, our [base name](https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/js/rx/rx-entry.jsx#L17) is `/healthcare/prescriptions` so all of the paths are relative to that.

I'll change it to use `Link` but we will need to override the paths by directly setting the `href` attribute on `Link`. (see latest commit for details).

src/js/rx/containers/RxRefillsApp.jsx
U-DON: Generally it's been by backwards depth in relative paths and then alphabetical. Might not be consistent across every module right now, but that's at least what I've been trying to do.

src/js/rx/components/Breadcrumbs.jsx
```diff
@@ -0,0 +1,38 @@
+import _ from 'lodash';
+import { Link } from 'react-router';
+import React from 'react';
+
+class Breadcrumbs extends React.Component {
+  render() {
+    const { location: { pathname }, prescription } = this.props;
+
+    const crumbs = [
+      <Link href="/" key="home">Home</Link>,
```
U-DON: I guess the only links that should be `Link` are the 'Prescription refills' crumbs that go to the app's root.

U-DON (approved):

src/sass/rx/partials/_rx-breadcrumbs.scss
```diff
@@ -0,0 +1,10 @@
+// Styles for contact card
```
U-DON: Comment from another stylesheet.

src/js/rx/components/Breadcrumbs.jsx
```diff
@@ -0,0 +1,38 @@
...
+      }
+    } else if (pathname.match(/\/glossary\/$/)) {
+      crumbs.push(<Link to="/" key="prescriptions">Prescription Refills</Link>);
+      crumbs.push(<span key="glossary"><strong>Glossary</strong></span>);
+    } else {
+      crumbs.push(<span key="prescriptions"><strong>Prescription Refills</strong></span>);
+    }
+
+    return (<div className="rx-breadcrumbs">
+      {crumbs.reduce((content, e) => { return [...content, ' › ', e]; }, crumbs[0]).slice(1)}
```
U-DON: Is the slice still necessary here? As long as it doesn't cut off the 'Home' part of the breadcrumbs, should be okay...

U-DON (commented):
LGTM. One last concern about the slice on the list in the breadcrumbs, but ignore if it works fine.

src/js/rx/components/Breadcrumbs.jsx
bshyong: Still needed; the first iteration for the reduce is `[null, ' › ', Home]`

U-DON: Maybe it's because the initializer needed to be `[crumbs[0]]`, not just `crumbs[0]`?

src/js/rx/components/Breadcrumbs.jsx
```diff
@@ -30,7 +30,7 @@ class Breadcrumbs extends React.Component {
     }
 
     return (<div className="rx-breadcrumbs">
-      {crumbs.reduce((content, e) => { return [...content, ' › ', e]; }, crumbs[0]).slice(1)}
+      {crumbs.reduce((content, e) => { return [...content, ' › ', e]; }, []).slice(1)}
```
U-DON: Wouldn't the initial value be `[crumbs[0]]`? When it was `crumbs[0]` by itself, the `...` operator probably didn't work correctly on it.

bshyong: that causes the first element to appear twice; the following code works but I prefer to use `slice` to avoid the ternary. `slice` is necessary to avoid the hanging `›` in front of the first element

`crumbs.reduce((content, e) => { return content.length > 0 ? [...content, ' › ', e] : [e]; }, [])`

PR #4367, opened 2016-12-14
[Rx] Unit tests for Detail container.

U-DON:


bshyong (approved):
LGTM

these tests look straightforward but feel free to let me know if there is anything special I should know about writing these types of tests

PR #4376, opened 2016-12-14
Adding GA events to actions for secure messaging

robertfairhead:
Per analytics conversation. These are the ones in the spreadsheet that need to be done via custom events (from what I could see). Actions seemed most natural spot for them but welcome your thoughts.

@U-DON definitely take a look at the send message one. Alex wanted to track if there was an additional subject entered so i think the extra field added there would capture it, but definitely welcome your thoughts on that one.

### DIFF
--- a/src/js/messaging/actions/compose.js
+++ b/src/js/messaging/actions/compose.js
@@ -12,6 +12,9 @@ import {
 import { apiRequest } from '../utils/helpers';
 
 export function deleteComposeMessage() {
+  window.dataLayer.push({
+    event: 'sm-delete-compose',
+  });
   return { type: DELETE_COMPOSE_MESSAGE };
 }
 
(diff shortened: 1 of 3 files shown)


U-DON:
Putting them in actions should be okay. That makes them easier to find than putting them in the individual handlers, although less coupled with the actual interaction with the page elements. But if the handler isn't triggering its corresponding action, that would most likely be an error anyway (in which case, would we care about counting broken interactions?). LGTM with these caveats.

Also, if there's a need differentiate between actions done on different pages, like moving a message from the folder view as opposed the message view, then that should be handled as well.

robertfairhead:
Cool. I'll clean it up and get the build working a bit. doing Rx now and want to see wrap up my thoughts there before i lose them.

At least in the list, i don't think @eli-lewis provided it was just asking for the moving action, etc. so I think this is a good first step. we can revisit once alex is back to see if he's happy with the data coming in or not.

eli-lewis:
@U-DON @robertfairhead I don't believe we need know moving actions at this point.  I've added it to a list of questions for Alex to confirm when he returns.

PR #4378, opened 2016-12-14
Make ConfirmRefillModal use common/Modal. closes #3091

### LINKED ISSUE #3091: [Rx] Refactor ConfirmRefillModal to use same modal pattern as Secure Messaging
See #3088.

~~Then move to `src/js/common`~~ **Update:** Moving is already done. Just need to make `ConfirmRefillModal` use the common component.

### LINKED ISSUE #4349: Change prescriptions modal title color?
In light of department-of-veterans-affairs/messaging-fe#197, should the prescription name change colors as well?

screen shot 2016-12-12 at 4 29 49 pm

If the answer is "no," it's safe to just close this issue.

webinista:
- Added title property to Modal component.
- Changed prescription title color. Closes #4349


U-DON:
LGTM. Although, it seems that the handling of the visibility in `ConfirmRefillModal` is redundant when returning an empty div, since the Modal component itself takes the visibility prop.

webinista:
It seems redundant, but `prescription` didn't become defined until the user triggers the modal and `this.props.isVisible` is `true`. Rewrote it to check whether `this.props.prescription` is available before using it. Please review again.

src/js/rx/components/ConfirmRefillModal.jsx
```diff
@@ -21,58 +21,54 @@ class ConfirmRefillModal extends React.Component {
...
-          </div>
-        );
-      } else {
+    if (this.props.isLoading) {
+      innerElement = (
+        <LoadingIndicator
+            message="Submitting your refill request..."/>
+      );
+    } else {
+      if (this.props.prescription) {
```
U-DON: Can combine this into an `else if`.

src/js/rx/components/ConfirmRefillModal.jsx
```diff
@@ -21,58 +21,54 @@ class ConfirmRefillModal extends React.Component {
...
       }
+    }
 
-      element = (
-        <section className="va-modal rx-modal" id="rx-confirm-refill">
-          <form className="va-modal-inner" onSubmit={this.handleConfirmRefill}>
-            {innerElement}
-          </form>
-        </section>
+    const element = (
```
U-DON: Can just return now without assigning first to a variable.

U-DON (commented):
LGTM. Suggested some really minor tweaks.

PR #4380, opened 2016-12-14
Check length of contention list

austin-martinez:


jbalboni (approved):
LGTM. Do we have any unit tests around this?

austin-martinez:
@jbalboni We did not, we actually don't have any tests for the containers. I can try and add one, just don't have a starting point so may take a bit longer

jbalboni:
@austin-martinez We do have tests for the containers, they're just under components because we don't test the Redux part of the component (which is what makes it a container). Not a very clear decision in retrospect. 

But it does look like we lack any tests for the details page, since it didn't have any behavior before. You should be able to start with the tests in ClaimDetailLayout, since it has the same logic. You would just need to import DetailsPage like the StatusPage test does (i.e. with brackets around the name).

PR #4381, opened 2016-12-14
Add GA events to Rx actions

robertfairhead:
Adding in the necessary custom events to track the requested Rx actions.


U-DON (approved):
LGTM.

PR #4388, opened 2016-12-14
Fix edu form error with missing to date

jbalboni:
If you leave the to field empty on a date range, we're passing an undefined value into the json stringify function in veteranToApplication, which causes an error. This adds a check to avoid that.


markolson (approved):
Logic LGTM

PR #4389, opened 2016-12-14
Jenkins CI

jkassemi:
Runs CI within docker container on Jenkins.

Once this is in place we'll continue to use Travis while we verify functionality and performance. Afterward we'll disable Travis and enable the deployment routine from the Jenkinsfile.


robbiethegeek:
@aub @crwallace and @alexose seems like you folks should keep this in your purview

src/js/edu-benefits/1990/utils/validations.js
```diff
@@ -274,6 +274,5 @@ export {
   isValidTourOfDuty,
   isValidEmploymentPeriod,
   isValidRotcScholarshipAmount,
-  isValidEducationPeriod,
-  isValidAddressField
```
crwallace: Should this be removed? This seems random.

crwallace (approved):
LGTM with one question

aub:
This seems fine to me conceptually, though I'm with @crwallace on the random change to the validations. Is the idea that it'll be faster under Docker?

src/js/edu-benefits/1990/utils/validations.js
jkassemi: @crwallace - it's a duplicate member. npm lint from the Jenkins docker-based tests caught the issue, but I haven't compared versions or config to determine why the version on Travis didn't. 

Removed from this and will submit in a new PR with better research.

Issue #4146, opened 2016-11-29
Add ascending/descending options for mobile sort


gnakm:

bshyong:
@gnakm are there mock ups for what this looks like?

gnakm:
Closing because all mocks can be found here #4411

PR #4391, opened 2016-12-14
Resolve #3978; prescriptions list view

### LINKED ISSUE #3978: Rx list view for refill page
At our most recent usability testing session, we had some Veterans from the Paralyzed Veterans of America and they shared with us that the list view is much more helpful for those who have difficulties with mobility. This is for post-launch but still in scope for 2016. 

rx refill list view

bshyong:
Resolves https://github.com/department-of-veterans-affairs/vets-website/issues/3978 list view for prescriptions


U-DON:
Sorry, I should've mentioned within the issue before you started on this, but we already have a `SortableTable` class that you can use from the common components. You can check for its usage in History.jsx under the containers directory and Folder.jsx in the messaging containers.

bshyong:
@U-DON I spend some time trying to use that component but found it easier to create a custom one to mirror the behavior of how the card views are currently sorted; i.e. unidirectional. If you pull down the branch you can see how the sorting on the list behaves in the same way as the card view

Is the plan to have both desc/asc sort on all of the views? I also noticed that the `History` view uses a different sort key from the `Active` view, which is a string instead of `{ value, order }`

U-DON:
Based on the addition of #4146, although it only mentions mobile sort, I think the designers have bidirectional sorting in mind for the Active page now too, since both sections have a sort dropdown on mobile. Maybe @gnakm can chime in on the plan for sorting on specifically the 'Refill prescriptions' section.

Whatever the design choice is, feel free to also augment SortableTable to have unidirectional sorting or disable sorting on specific fields as well. It looks like only the left three fields are intended to be sortable.

Additionally, although Active's sort key is a string rather than the `{ value, order }` object, it should be equivalent to the `value` part of that pair. If it turns out we need to differentiate, we can do the special handling between the types within Active's `handleSort` method.

bshyong:
Okay. Based on this conversation:
- modify Active object shape to include both value and order
- use sortable table component in favor of custom component
- modify sortable table to selectively enable sorting on columns

@gnakm do you have UX and design thoughts on sorting for the prescription refills page? In particular the sorting UI for card and list view differs dramatically: card view is currently unidirectional sort and text links while the list view is a table view where the sort buttons are embedded as part of the table headers.

bshyong:
I just added the sortable table component and reworked the prescriptions list to use it, which allows for both ascending and descending sort on the data.  @gnakm if you are opening that new issue with design/UX thoughts for the card view to help with consistency I can make more changes based on that.

src/js/common/components/SortableTable.jsx
```diff
@@ -33,6 +33,14 @@ class SortableTable extends React.Component {
       }
     }
 
+    if (field.nonSortable) {
+      return (
+        <th key={field.value}>
+          {field.label}
+        </th>
+      );
+    }
```
U-DON: I'd like to see this at the top of this function so it's clearly separate and not in the midst of the logic that deals with sorting stuff.

src/js/rx/containers/Active.jsx
```diff
@@ -28,20 +33,35 @@ class Active extends React.Component {
...
-    const oldSort = this.props.sort;
-
-    if (newSort !== oldSort) {
-      this.props.sortPrescriptions(newSort);
-    }
-  }
-
-  handleSort(sort) {
+  handleSort(sortKey, order) {
+    const sortParam = order === 'DESC' ? sortKey : `-${sortKey}`;
```
U-DON: Is this reversed? I thought descending corresponds to the key with the minus sign.

src/js/rx/containers/Active.jsx
```diff
@@ -95,6 +129,7 @@ const mapStateToProps = (state) => {
   return {
     ...state.prescriptions.active,
     prescriptions: state.prescriptions.items,
+    history: state.prescriptions.history,
```
U-DON: Is this necessary in Active? I don't see it being used.

src/js/rx/reducers/prescriptions.js
```diff
@@ -7,7 +7,10 @@ const initialState = {
   items: [],
   active: {
     loading: false,
-    sort: 'prescriptionName'
+    sort: {
+      value: 'prescriptionName',
+      order: 'DESC',
```
U-DON: I think the default is intended to be 'ASC'.

src/js/rx/components/PrescriptionTable.jsx
```diff
@@ -0,0 +1,77 @@
...
+import SortableTable from '../../common/components/SortableTable';
+import { formatDate } from '../utils/helpers';
+import Prescription from './Prescription';
+
+class RefillStatus extends Prescription {
+  render() {
+    const { id } = this.props;
+
+    return (
+      <div key={id}>
```
U-DON: I don't think the key prop is supposed to be used here. Usually it's at the time you're creating a component as a part of a collection, so it would be a prop in the component created within the object in the map below.

https://facebook.github.io/react/docs/lists-and-keys.html#extracting-components-with-keys

U-DON (changes requested):
Generally looks good, but noticed a few things that might need to be resolved.

One common thing I saw was that descending and ascending may be mixed up in a few places. 'ASC' was the original default (so for prescription name, A -> Z).

src/js/rx/containers/Active.jsx
U-DON: Actually, I take that back on the edit. (Reading old code is hard...) It looks like the handler should be passing the order that is being requested, so 'DESC' should correspond to the negative key. So `prescriptionName` is 'ASC' while `-prescriptionName` is 'DESC'.

src/js/rx/actions/prescriptions.js
```diff
@@ -110,6 +110,6 @@ export function refillPrescription(prescription) {
   return dispatch => dispatch({ type: 'REFILL_FAILURE' });
 }
 
-export function sortPrescriptions(sort) {
-  return { type: 'SORT_PRESCRIPTIONS', sort };
+export function sortPrescriptions(sort, order = 'DESC') {
```
U-DON: Let's have the default here be 'ASC' too.

src/js/rx/reducers/prescriptions.js
```diff
@@ -111,17 +114,24 @@ export default function prescriptions(state = initialState, action) {
...
-          return set('items', _.sortBy(state.items, sortByFacilityName), newState);
+          return set('items', _.orderBy(state.items, sortByFacilityName, [order]), newState);
         case 'lastSubmitDate':
-          return set('items', _.sortBy(state.items, sortByLastSubmitDate), newState);
+          return set('items', _.orderBy(state.items, sortByLastSubmitDate, [order]), newState);
         default:
-          return set('active.sort', 'prescriptionName', state);
+          return set('active.sort', {
+            value: 'prescriptionName',
+            order: 'DESC',
```
U-DON: Also default to 'ASC'.

src/js/rx/reducers/prescriptions.js
```diff
@@ -91,7 +94,7 @@ export default function prescriptions(state = initialState, action) {
       if (action.active) {
         newState.active = {
           loading: false,
-          sort: sortValue
+          sort: { value: sortValue, order: 'DESC' },
```
U-DON: Since we're introducing sorting in both orders for Active now, order here should also get `sortOrder` like History does.

U-DON (changes requested):
Few more changes regarding sort order.

U-DON (approved):

U-DON:
LGTM.

PR #4394, opened 2016-12-15
Standardizing width of Compose form fields.

webinista:
Closes department-of-veterans-affairs/messaging-fe#198

- Made recipients element 100% wide.
- Made category, subject 50% wide and container 100% wide.
- Fixed bug in which errored subject line labels weren't bold.


src/sass/messaging/partials/_message-compose.scss
```diff
@@ -385,16 +377,19 @@ button.msg-btn-save {
...
-    border: 1px solid $color-gray;
-  }
+.msg-compose-error--subject select,
+.msg-compose-error--category input {
+  border: 1px solid $color-gray;
+}
+
+.msg-compose-error--category .msg-category label,
+.msg-compose-error--subject .msg-subject label {
+  font-weight: bold;
```
U-DON: I think this block is duplicated a few changes above.

U-DON (commented):
Aside from a duplicate block in the stylesheet, LGTM, though didn't have time to check how it actually looks.

PR #4399, opened 2016-12-15
Fix edu form error with missing to date (#4388)

akainic:
Update production with a fix for an education form bug.


crwallace:
Lgtm

PR #4410, opened 2016-12-16
[Messaging] Loading state for user actions.

### LINKED ISSUE #4015: Loading animation for user actions

U-DON:
Closes #4015.

The set of changes might look massive, but a chunk of it is the newly created loading reducer and its corresponding unit tests, which are straightforward.

Everything else is rearranging the props mapped from state after the refactoring to use the new reducer.


U-DON:
## Previews

Most user-initiated actions that have latency will replace all of the main content with the loader, leaving only the main app header (includes alerts).
> screen shot 2016-12-15 at 6 02 22 pm

Only creating a new folder is different. It's contained within the create folder modal.
> screen shot 2016-12-15 at 6 06 33 pm

src/js/messaging/containers/Thread.jsx
```diff
@@ -277,18 +268,18 @@ export class Thread extends React.Component {
   }
 
   render() {
-    if (this.props.isNewMessage && this.props.loadingRecipients) {
-      return <LoadingIndicator message="Loading the application..."/>;
-    }
-
     const loading = this.props.loading;
```
bshyong: Do you have preferences for using destructuring assignment up front for all of the first line variables coming from `this.props`, e.g. `isNewMessage, loading, message`, etc. instead of having to reference `this.props` everywhere below?

src/js/messaging/reducers/compose.js
```diff
@@ -21,12 +21,7 @@ const initialState = {
 };
 
 const resetMessage = (state) => {
-  let msg = set('message.category', initialState.message.category, state);
-  msg = set('message.recipient', initialState.message.recipient, msg);
-  msg = set('message.subject', initialState.message.subject, msg);
-  msg = set('message.attachments', initialState.message.attachments, msg);
-  msg = set('message.body', initialState.message.body, msg);
-  return msg;
+  return set('message', initialState.message, state);
```
bshyong: 💯

bshyong (approved):
LGTM

src/js/messaging/containers/Thread.jsx
U-DON: Hm, that would be cleaner. I often forget that's a thing. I can certainly change it to that.

PR #4414, opened 2016-12-16
Resolves #4149 pagination characters

### LINKED ISSUE #4149: Pagination should use unicode right & left pointing angle quotation marks instead of greater-than & less-than characters
Pagination and should use unicode `›` `‹` carats, not `<` `>`characters

bshyong:
- Resolves #4149 pagination characters


U-DON (approved):
LGTM.

PR #4416, opened 2016-12-16
[Messaging] Changed header for editing draft of new message to "Edit draft".

U-DON:
Closes department-of-veterans-affairs/messaging-fe#193.

Also included refactoring to destructure parts of Thread.jsx that were individually getting props.

screen shot 2016-12-16 at 1 50 30 pm

### DIFF
--- a/src/js/messaging/containers/Thread.jsx
+++ b/src/js/messaging/containers/Thread.jsx
@@ -52,30 +52,27 @@ export class Thread extends React.Component {
   }
 
   componentDidMount() {
-    if (this.props.redirect) {
-      this.context.router.replace(this.props.redirect);
-      return;
-    }
+    const { loading, redirect } = this.props;
 
-    if (!this.props.loading.thread) {
-      const id = +this.props.params.messageId;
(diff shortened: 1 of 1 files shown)


src/js/messaging/containers/Thread.jsx
```diff
@@ -147,19 +154,16 @@ export class Thread extends React.Component {
     // Exclude the current folder from the list of folders
     // that are passed down to the MoveTo component.
     const moveToFolders = [];
-    this.props.folders.forEach((folder) => {
+    folders.forEach((folder) => {
```
bshyong: You could do this loop in one line if you prefer
`const moveToFolders = _.reject(folders, ['folderId', currentFolder.folderId]);`

bshyong (approved):
LGTM

src/js/messaging/containers/Thread.jsx
U-DON: `folders` is a Map, so it would need to be transformed to an Array first. I would like a one liner, but only if it does it in one pass.

PR #4425, opened 2016-12-17
Fixed color of track package link in list view. closes #4423

### LINKED ISSUE #4423: Track Package link text is invisible when in list view
On healthcare/prescriptions, when in list view, the link text disappears. 
Caused by the `.rx-table a` selector. 

screen shot 2016-12-16 at 3 42 48 pm
 
## To fix: 
- [x] Remove `.rx-table a` selector
- [x] Update links in tables to have their own class name (e.g. `rx-prescription-link`)

Related to #4422

webinista:
- Removed .rx-table a ruleset.
- Added rx-prescription-name class.

### DIFF
--- a/src/js/rx/components/PrescriptionTable.jsx
+++ b/src/js/rx/components/PrescriptionTable.jsx
@@ -41,9 +41,9 @@ class PrescriptionTable extends React.Component {
         id: item.id,
         prescriptionName: (
           <div>
-            <Link to={`/${attrs.prescriptionId}`}>
+            <Link to={`/${attrs.prescriptionId}`} className="rx-prescription-name">
               {attrs.prescriptionName}
-            </Link><br/>
+            </Link>
             <span>Prescription #: {item.id}</span>
           </div>
         ),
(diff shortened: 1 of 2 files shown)


src/js/rx/components/PrescriptionTable.jsx
```diff
@@ -41,9 +41,9 @@ class PrescriptionTable extends React.Component {
         id: item.id,
         prescriptionName: (
           <div>
-            <Link to={`/${attrs.prescriptionId}`}>
+            <Link to={`/${attrs.prescriptionId}`} className="rx-prescription-name">
               {attrs.prescriptionName}
-            </Link><br/>
```
U-DON: The actual `TrackPackageLink` only uses the `usa-button` class. Would that suffice for this `Link`?

U-DON (commented):
LGTM, but wondering if the new class is necessary given how `TrackPackageLink` is used.

src/js/rx/components/PrescriptionTable.jsx
webinista: For the text color, yes. But there are some margin and display differences too.

PR #4428, opened 2016-12-17
Resolves sm#148 message date queries

bshyong:
- Resolves https://github.com/department-of-veterans-affairs/messaging-fe/issues/148

This PR makes it so that a user selecting the same date for start and end will have the end date incremented by 1 day. If this is confusing, we can also hide this from the user and do it in the action itself.

It seems like we need to always increment the end date by 1 day to have inclusive behavior? i.e. 12-14 to 12-15 returns messages sent on 12-14 AND 12-15

### DIFF
--- a/src/js/messaging/reducers/search.js
+++ b/src/js/messaging/reducers/search.js
@@ -1,4 +1,5 @@
 import set from 'lodash/fp/set';
+import moment from 'moment';
 
 import { makeField } from '../../common/model/fields';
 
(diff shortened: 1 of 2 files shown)


U-DON:
Yes, the intent is to **always** be inclusive of end date, not only if it's the same day. I think that's how I wrote it in the issue. Sorry if that wasn't clear.

bshyong:
@U-DON got it. sounds like you are saying to do it in the action so that it's invisible to the user, and do it ALWAYS for every query? I'll make that change

U-DON:
Yes, it should not be an option to the user. Every search should just be start and end date inclusive.

Also, you can leverage the `endOf` function from Moment rather than adding a day. http://momentjs.com/docs/#/manipulating/end-of/

bshyong:
Thanks for the suggestion to use `endOf`. I revisited this issue and it works cleanly in the reducer. The date picker only displays dates so the user is oblivious to the `endOf` modification—the end date displayed remains the same

bshyong:
screenshot here showing a search for April 11 messages

screen shot 2016-12-19 at 10 35 39 am

U-DON:
LGTM.

PR #4429, opened 2016-12-19
Disables List view on mobile and fixes it for small screens

### LINKED ISSUE #4427: Hide list / card toggler on small screens [Was: Prescription list view needs labels]
When toggling into list view, the table header labels disappear. Each cell loses its label.

screen shot 2016-12-16 at 6 03 00 pm

Related to #4411

bshyong:
Resolves #4427 by disabling the toggle to List view on mobile devices and dynamically switching to card view on small screens for desktop

### DIFF
--- a/src/js/rx/containers/Active.jsx
+++ b/src/js/rx/containers/Active.jsx
@@ -29,6 +29,8 @@ class Active extends React.Component {
       const viewToggleElement = this.refs.viewToggle;
       const toggleDisplayStyle = window.getComputedStyle(viewToggleElement, null).getPropertyValue('display');
 
+      // the viewToggle element is hidden with CSS on the $small breakpoint
+      // on small screens, the view toggle is hidden and list view disabled
       if (viewToggleElement && (toggleDisplayStyle === 'none')) {
         this.setState({
           view: 'card',
--- a/src/js/rx/containers/Active.jsx
+++ b/src/js/rx/containers/Active.jsx
@@ -2,6 +2,7 @@ import React from 'react';
 import { connect } from 'react-redux';
 import classnames from 'classnames';
 import isMobile from 'ismobilejs';
+import _ from 'lodash';
 
 import {
   loadPrescriptions,
@@ -23,6 +24,15 @@ class Active extends React.Component {
   constructor(props) {
     super(props);
     this.handleSort = this.handleSort.bind(this);
+
+    this.checkWindowSize = _.debounce(() => {
+      if (document.documentElement.clientWidth < 600) {
+        this.setState({
+          view: 'card',
+        });
+      }
+    }, 200);
+
     this.state = {
       view: 'card',
     };
@@ -32,6 +42,11 @@ class Active extends React.Component {
     if (!this.props.loading) {
       this.props.loadPrescriptions({ active: true });
     }
+    window.addEventListener('resize', this.checkWindowSize);
+  }
+
+  componentWillUnmount() {
+    window.removeEventListener('resize', this.checkWindowSize);
   }
 
   handleSort(sortKey, order) {
--- a/src/js/rx/containers/Active.jsx
+++ b/src/js/rx/containers/Active.jsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import { connect } from 'react-redux';
 import classnames from 'classnames';
+import isMobile from 'ismobilejs';
 
 import {
   loadPrescriptions,
@@ -114,7 +115,7 @@ class Active extends React.Component {
 
     return (
       <div id="rx-active" className="va-tab-content">
-        {this.renderViewSwitch()}
+        {isMobile.any ? null : this.renderViewSwitch()}
         {content}
       </div>
     );
--- a/src/js/rx/containers/Active.jsx
+++ b/src/js/rx/containers/Active.jsx
@@ -26,7 +26,10 @@ class Active extends React.Component {
     this.handleSort = this.handleSort.bind(this);
 
     this.checkWindowSize = _.debounce(() => {
-      if (document.documentElement.clientWidth < 600) {
+      const viewToggleElement = this.refs.viewToggle;
+      const toggleDisplayStyle = window.getComputedStyle(viewToggleElement, null).getPropertyValue('display');
+
+      if (viewToggleElement && (toggleDisplayStyle === 'none')) {
         this.setState({
           view: 'card',
         });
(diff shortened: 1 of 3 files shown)


U-DON:
I think the decision was to always use card view in mobile and that the toggle should only be available on desktop. I left a suggestion (that may or may not work) in the issue.

bshyong:
@U-DON The implementation in this PR works in the following way:
- View state is always set to 'card' by default
- On mobile devices, the toggle is not rendered so users cannot switch to list view on mobile
- On desktop, if a user toggles to list view and then resizes the screen to less than 600px in width, they will end up with the buggy view. To address this, we can either add the labels in css (implemented here), or add a [debounced] event listener on `windowresize` so automatically change the view state back to 'card' if the window is resized

We can only make a pure CSS solution if we render both views and selectively hide one by setting `display: none`—I think that doing this places unnecessary load on the client because they have to render duplicated information

bshyong:
Screenshot of labels for small screen desktop. Looking back at this, it might be cleaner from a UX/Design perspective to just listen to window resize events and change the state to card view. Unless you are opposed to that approach, I'll implement that so we can get rid of the list view for good on desktop screens < 600px wide

screen shot 2016-12-19 at 10 37 52 am

U-DON:
I agree on the drawback of a pure CSS solution (unnecessary operations to render card view just to hide it), so we don't have to do it that way. I'm down with the resize event implementation.

On a related note, I'm not sure if I'm remembering incorrectly, but I thought list view was intended to be the default so that users don't get confused by the two-column card view. Maybe @gnakm can give the final word on that.

bshyong:
@gnakm summarizing for clarity

1. Is the List view the default view on desktop?
2. List view will be completely disabled on mobile devices: only the card view will be shown
3. On desktop, it is possible for the user to resize their window to < 600 px which breaks the List view. In this case, we will automatically toggle to card view if resized, and disable the toggle switch.

Please add notes if you have thoughts on any of the above!

gnakm:
@bshyong @U-DON 

Here are my thoughts. 
I think default should be cards on desktop. It is very important for us to have a list _option_ for those with limited mobility and head trauma (some users we talked to and did testing with). I've been helping the core team with visual consistency throughout Vets.gov and a lot of things are defaulted as cards so we should stick to that. Also, reading from above, the responsiveness of switching defaults certainly doesn't sound ideal. With that said, I think we should be conscientious of Veteran feedback around both views and the option to toggle.

bshyong:
The above feedback has been implemented:
- Default view style remains as `card`.
- Mobile devices will only ever see the `card` view, with no toggle to switch to the list view
- Desktop browsers will see a toggle that allows switching to `list` view
- Desktop browsers that resize their screen to < 600 px will see the `card` view and the toggle will be hidden

U-DON:
LGTM.

src/js/rx/containers/Active.jsx
```diff
@@ -22,6 +24,15 @@ class Active extends React.Component {
   constructor(props) {
     super(props);
     this.handleSort = this.handleSort.bind(this);
+
+    this.checkWindowSize = _.debounce(() => {
+      if (document.documentElement.clientWidth < 600) {
```
U-DON: How did you determine it was 600? Should it be assigned to a constant so it's clear what this limit is for?

bshyong: Should we put a comment here? It is 600 because the `$small` breakpoint from foundation is set at `40em`, and our base font-size is 15px

bshyong: I just thought of another way to do this. We could assign an ID to the view toggle element and then instead of checking window width, check if that element is visible in the DOM. The view toggle element should be hidden automatically by CSS classes (`display:none`)

U-DON: Sure, I'm on board with that. Could potentially use a React ref as well.

src/js/rx/containers/Active.jsx
```diff
@@ -65,7 +68,7 @@ class Active extends React.Component {
     ];
 
     return (
-      <div className="rx-view-toggle">View:&nbsp;
+      <div className="rx-view-toggle" ref="viewToggle">View:&nbsp;
```
U-DON: Although this is easier to work with, I've generally been trying to follow React's guidelines of using the callback version of refs.

I guess I don't mind either way, but thoughts on following their recommendation in case they deprecate the old way of using refs?

> If you are currently using `this.refs.myRefName` to access refs, we recommend using this pattern instead.

https://facebook.github.io/react/docs/refs-and-the-dom.html

src/js/rx/containers/Active.jsx
```diff
@@ -26,7 +26,10 @@ class Active extends React.Component {
     this.handleSort = this.handleSort.bind(this);
 
     this.checkWindowSize = _.debounce(() => {
-      if (document.documentElement.clientWidth < 600) {
+      const viewToggleElement = this.refs.viewToggle;
```
U-DON: What's happening here might not be obvious to someone unfamiliar with the behavior, so might be helpful to leave a comment about how the stylesheet is hiding the button when the window's resized to a small screen.

bshyong:
Great I'll add the comment now and get it merged!

--
Ben Shyong
408-505-9891




On Tue, Dec 20, 2016 at 9:03 AM -0800, "Eugene Doan" <notifications@github.com> wrote:












@U-DON commented on this pull request.





In src/js/rx/containers/Active.jsx:
> @@ -65,7 +68,7 @@ class Active extends React.Component {
     ];
 
     return (
-      <div className="rx-view-toggle">View:&nbsp;
+      <div className="rx-view-toggle" ref="viewToggle">View:&nbsp;



Although this is easier to work with, I've generally been trying to follow React's guidelines of using the callback version of refs.


I guess I don't mind either way, but thoughts on following their recommendation in case they deprecate the old way of using refs?



If you are currently using this.refs.myRefName to access refs, we recommend using this pattern instead.



https://facebook.github.io/react/docs/refs-and-the-dom.html





In src/js/rx/containers/Active.jsx:
> @@ -26,7 +26,10 @@ class Active extends React.Component {
     this.handleSort = this.handleSort.bind(this);
 
     this.checkWindowSize = _.debounce(() => {
-      if (document.documentElement.clientWidth < 600) {
+      const viewToggleElement = this.refs.viewToggle;



What's happening here might not be obvious to someone unfamiliar with the behavior, so might be helpful to leave a comment about how the stylesheet is hiding the button when the window's resized to a small screen.

PR #4441, opened 2016-12-20
[Messaging] Update message counts after moving messages and saving drafts.

U-DON:
## Changes
* When moving a message, decrement the count of the original folder and increment the count of the destination folder.
    * Closes department-of-veterans-affairs/messaging-fe#189.
    * Moved the responsibility of filtering the folders to display to the MoveTo component itself. It was previously filtering in two stages; one within a parent component to remove the current folder, and the other within itself to only keep 'Inbox' and personal folders. Now that it's also receiving the data of the current folder, it might as well process that folders list in one place.
* When saving a new draft (not saving an existing one), increment the Draft count.
    * Closes department-of-veterans-affairs/messaging-fe#195.
* When fetching a folder, update its entry in the map of all folders in state. So if, for example, a draft was saved in a separate session, fetching the Drafts folder will update with the correct count.

### DIFF
--- a/src/js/messaging/actions/messages.js
+++ b/src/js/messaging/actions/messages.js
@@ -105,9 +105,8 @@ export function fetchThreadMessage(messageId) {
   };
 }
 
-export function moveMessageToFolder(messageId, folder) {
-  const folderId = folder.folderId;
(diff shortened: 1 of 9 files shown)


bshyong (approved):
LGTM!

PR #4442, opened 2016-12-20
[Messaging] Show loader when fetching list of folders.

U-DON:
Before these changes, we were briefly displaying "Sorry, this folder doesn't exist." or something to that effect before we received all of the folders. This should resolve that and prevent race conditions where children components that rely on the folders list could have failed.


bshyong (approved):
LGTM!

PR #4445, opened 2016-12-20
Consolidate all education validations

akainic:
Consolidation of all education validations, both general and field. This is going to be hard to review because I moved around the order of a lot of validation functions, so the git comparison is hard to read. Mostly what I did was:
- Remove common validations from edu-benefits/utils/validations
- Move validations that could be shared from edu-benefits/utils/validations into common/utils/validations
- Reorder validations in common/utils/validations with some comments to group them

### DIFF
--- a/src/js/common/utils/validations.js
+++ b/src/js/common/utils/validations.js
@@ -175,6 +175,10 @@ function isValidPartialMonthYearInPast(month, year) {
   return !year || isValidPartialMonthYear(month, year) && momentDate.isValid() && momentDate.isSameOrBefore(moment().startOf('month'));
 }
 
+function isBlankMonthYear(field) {
+  return isBlank(field.month.value) && isBlank(field.year.value);
+}
+
 function isValidDateOver17(day, month, year) {
   if (!isValidYear(year)) {
     return false;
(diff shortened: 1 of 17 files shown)


src/js/edu-benefits/utils/validations.js
```diff
@@ -136,52 +46,10 @@ function isValidRoutingNumber(value) {
...
-
-function isValidRequiredField(validator, field) {
-  return isNotBlank(field.value) && validator(field.value);
-}
-
-function isBlankDateField(field) {
-  return isBlank(field.day.value) && isBlank(field.month.value) && isBlank(field.year.value);
-}
-
 function isBlankMonthYear(field) {
```
jbalboni: I think this is for a common component, maybe it should be moved?

jbalboni (commented):
This looks good to me. Couple of questions. I ran through things locally and didn't see any errors.

src/js/common/utils/validations.js
```diff
@@ -156,10 +188,86 @@ function isValidDateOver17(day, month, year) {
   return momentDate.isBefore(moment().endOf('day').subtract(17, 'years'));
 }
 
+function isValidEntryDateField(date, dateOfBirth) {
```
akainic: Yes, you're right! Not sure how it got placed in this file, I will move it out.

src/js/edu-benefits/utils/validations.js
akainic: I wasn't sure about this one, it seems to only be used by education for now. It's possible it will be used for other applications in the future, so I can move it out.

src/js/common/utils/validations.js
akainic: Yes, you're right! Not sure how it got placed in this file, I will move it out.

akainic:
@jbalboni made the fixes you recommended, if you want to take another look.

jbalboni (approved):
LGTM

PR #4448, opened 2016-12-20
Adds analytics for view toggle

bshyong:
Also uses React recommended ref style to avoid future deprecation issues


U-DON (approved):
LGTM.

PR #4452, opened 2016-12-20
Disable tests under jenkins

jkassemi:
While working through resource utilization issues this will disable
the tests from running under Jenkins. An updated Jenkinsfile will be
added later when these issues are resolved.


dkhenry (approved):
This will in fact disable tests via jenkins :+1:

PR #4461, opened 2016-12-21
Rewrote menu.js to remove jQuery dependency. closes #4444

### LINKED ISSUE #4444: Rewrite menu.js to remove jQuery dependency.
Rewrite `src/js/legacy/menu.js` using standard DOM API methods.

Blocks #4230.

webinista:


src/js/legacy/menu.js
```diff
@@ -1,28 +1,32 @@
...
-		
-		} else {
-		
-		  $(overlay).toggleClass('va-overlay--closed', false);
-	    $(overlay).toggleClass('va-overlay--open', true);
-		
-		}
-	}
+  const toggleOverlay = (domEvent) => {
+    domEvent.preventDefault();
```
jbalboni: I'm not entirely sure it's related to your changes, but right now for the crisis line overlay on mobile, clicking on the options just closes the overlay, rather than going to the chat or opening the phone link.

src/js/legacy/menu.js
```diff
@@ -1,28 +1,32 @@
-$(document).ready(function() {
+document.addEventListener('DOMContentLoaded', () => {
```
jbalboni: If for some reason this script is run after the content loaded event, this listener won't run. This is different from the jQuery ready function, I believe, which will run the function if the doc is in the ready state already.

I don't think that matters right now, but it could in the future if we load bundles with defer or async.

jbalboni (approved):
Looks good to me. I did notice one bug that might be related to your changes.

src/js/legacy/menu.js
webinista: I believe the mobile (< 481px viewports) version is still controlled by `menu.js`. They'd be unaffected by these changes.

src/js/legacy/menu.js
webinista: jQuery appears to use `DOMContentLoaded` and fallback to `document.readyState === 'complete'` for older versions of IE. I think we're okay for now, and probably okay if we load bundles with `defer`.

webinista:
@jbalboni Are you okay with this being merged with a promise to follow up on `DOMContentLoaded` if / when we load bundles differently?

jbalboni:
@webinista Yep, works for me.

PR #4467, opened 2016-12-21
[Messaging] Fixed unauthorized downloads of attachments.

U-DON:
Closes department-of-veterans-affairs/messaging-fe#159.

We were previously trying to access the attached files through the API without an auth header. Even with that header, we still need to convert the retrieved data into a blob and use the URL for it. In this case, I just have it open in a new tab.

I also added a small spinner to indicate that the file is being downloaded.

### DIFF
--- a/src/js/messaging/components/MessageAttachmentsViewItem.jsx
+++ b/src/js/messaging/components/MessageAttachmentsViewItem.jsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import classNames from 'classnames';
 
 import { authFetch } from '../utils/helpers';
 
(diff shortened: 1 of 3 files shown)


src/js/messaging/components/MessageAttachmentsViewItem.jsx
```diff
@@ -1,11 +1,52 @@
...
+      window.open(this.downloadUrl, '_blank');
+      return;
+    }
+
+    this.setState({ downloading: true });
+    const requestUrl = this.props.url;
+    authFetch(requestUrl)
+    .then(response => response.blob())
+    .then(blob => {
+      const downloadUrl = URL.createObjectURL(blob);
```
bshyong: 👍  FYI max blob size is about 500MB in most browsers; I don't expect that we will ever come close to that limit

bshyong (approved):
LGTM

src/js/messaging/components/MessageAttachmentsViewItem.jsx
bshyong: I don't know of any—the best way to handle large files is to have the API return a signed URL with expiration. Given that the request needs to have an auth header it is impossible to use links, and using AJAX XMLHttpRequest won't trigger a browser file download.

Outside of using signed URLs like S3/Cloudfront do, your approach here is the best way to do it given these limitations

U-DON: Looking further into our config.js, the limit we've set on attachments is 3 MB anyway, so seems like there won't be a problem with the blob size limit.

PR #4471, opened 2016-12-21
Move current edu form into a sub folder so we can support multiple forms

jbalboni:
Closes [141](https://github.com/department-of-veterans-affairs/edu-benefits-team/issues/141)

This moves most of our code into a 1990 folder, so that we can build multiple forms in a single app. For now, I'm just redirecting to the 1990 form all the time, but in the future we'll likely have a component at the root that asks questions of users to direct them properly.

There are a lot of changes that are mainly just paths. The notable code changes are in routes and entry files in the edu-benefits and edu-benefits/1990 folders.

Big important thing: Going to `/education/apply-for-education-benefits/application` should always redirect you to `/education/apply-for-education-benefits/application/1990/introduction`. Otherwise, we break links and confuse users.


akainic (commented):
Mostly looks good! The only issue I encountered when testing this locally was on the review page, with the expanding and collapsing review panels. Looks like some problem in the entry file.

jbalboni:
@akainic This should be working better now, if you want to take another look.

akainic (approved):
LGTM! We can test more on staging once this is merged.

PR #4476, opened 2016-12-22
Prevent default event on draft deletion

### LINKED ISSUE #206: Deleting drafts sends user in loop
1. Go to compose a new message and save it as a draft
2. Go to the drafts folder and open the saved draft
3. Click delete
4. Click "Yes, delete forever"

Expected: Draft is deleted and user is taken back to the drafts folder
Actual: Draft is deleted but user is taken through many different loading screens, eventually ending up at something that looks like this (note the URL):

screen shot 2016-12-21 at 1 19 26 am

Note: this same behavior happens whether you hit the delete button at the top or the bottom of the draft (seen as both the red circles below) so needs to be fixed in both places. 

screen shot 2016-12-21 at 1 18 19 am

bshyong:
- potentially resolves https://github.com/department-of-veterans-affairs/messaging-fe/issues/206

This is difficult for me to debug because I still have some trouble with using SSH to hit the real server, but could it be that clicking the submit button triggers a full page reload? This is what seems to happen for me when I click it.

With this change, the full page refresh goes away, and I can see the "Failed to delete message" alert on top. (Expected since the Heroku mock endpoint has a 500 error

### DIFF
--- a/src/js/messaging/components/compose/ModalConfirmDelete.jsx
+++ b/src/js/messaging/components/compose/ModalConfirmDelete.jsx
@@ -3,15 +3,21 @@ import React from 'react';
 import Modal from '../../../common/components/Modal';
 
 class ModalConfirmDelete extends React.Component {
+
+  handleDelete = (e) => {
+    e.preventDefault();
+    this.props.onDelete();
+  }
+
   render() {
     const modalContents = (
-      <form onSubmit={this.props.onDelete}>
+      <form>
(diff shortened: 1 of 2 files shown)


U-DON (approved):
Strange, I don't get the full refresh when I click the button, but that seems like letting the submit event happen could cause it in some browsers. In general, we should be preventing submits anyway. LGTM.

PR #4477, opened 2016-12-22
[Messaging] Refactored folder state usage. Prevent immediate dismissal of alerts after message operations.

### LINKED ISSUE #203: Alerts are not displaying after successful send, save, delete
Introduced a regression after adding the code to close alerts after a page change. We have redirects after a successful save, send, or delete, so the automatic page change closes the alert before it can be seen.

U-DON:
## Changes
* Reverted the addition of `fromFolder` from move message actions, as the origin folder should be the same as the current folder in state.
* To ensure we always have the current folder properly populated in state, fetch the folder when opening a thread directly by URL. The condition to fetch checks the folder name from the URL.
    * As a result, redirects after message operations should go back to the appropriate folder.
* Removed `persistFolder` and its related actions in favor of the data in `lastRequestedFolder`.
* Redirects won't immediately close alerts, ~since they are done with `history.replace`, and we check now that the route change is not a 'REPLACE' before closing the alert~ now that we set a flag on the relevant router operations to keep the alerts alive.
    * Closes department-of-veterans-affairs/messaging-fe#203.
* Fix for Compose page to redirect after deleting the message.
* Refactored redirects to allow going back.
    * Specifically, this works in the case of saving an already existing draft, since it's the one case where the data from the previous view should be preserved.


src/js/messaging/reducers/folders.js
```diff
@@ -27,17 +26,11 @@ const initialState = {
...
       filter: {},
       messages: [],
-      pagination: {
-        currentPage: 0,
-        perPage: 0,
-        totalEntries: 0,
-        totalPages: 0
-      },
-      persistFolder: 0,
+      pagination: {},
```
bshyong: Is it okay for all of the values to be null initially? It looks like some of these attributes may flow through to a couple of components that use it. It only affects the initial render (before folders are fetched and the API populates the pagination attribute, but it still might be visible while the API request is coming back

https://github.com/department-of-veterans-affairs/vets-website/blob/d2bfd141baedaee8f1013ff3b0f9bea03f43e425/src/js/messaging/components/MessageNav.jsx#L39

bshyong (approved):
LGTM

src/js/messaging/reducers/folders.js
U-DON: That's true. I don't think showing something like "0 of 0" looks that great either, though. Maybe that components that need data that are undefined should be hidden until they're populated?

bshyong: Yep I agree; we should either hide them or implement a reasonable empty state.

U-DON: I'm going to revert to the initial 0 values for now and follow up with a PR to handle undefined/null states for the relevant components.

PR #4481, opened 2016-12-22
Parallelize nightwatch

jkassemi:
Mock server implementation allows mocking requests based on user, which permits running multiple tests in parallel. Worker count set to n=10, may adjust while tuning performance for Jenkins environment.


test/util/login-helpers.js
```diff
@@ -52,16 +51,29 @@ function initUserMock(level) {
 }
 /* eslint-enable camelcase */
 
-function logIn(client, url, level) {
-  initUserMock(level);
+let tokenCounter = 0;
```
jkassemi: none of the tests we're currently running require multiple signed in users. This little extra protection (as opposed to making the user unique to the pid) ensures that if we do find a way to run tests in parallel within a single process, this will still function as expected.

test/util/login-helpers.js
```diff
@@ -52,16 +51,29 @@ function initUserMock(level) {
...
+
+function getUserToken() {
```
jkassemi: I considered trying to use a selenium-based session id, but the session id is not sent in the request to the mock server, and using less magic around this felt more maintainable.

test/util/login-helpers.js
```diff
@@ -52,16 +51,29 @@ function initUserMock(level) {
...
+  return `token-${process.pid}-${tokenCounter++}`;
+}
+
+function logIn(token, client, url, level) {
+  initUserMock(token, level);
+
   client
     .url(`${E2eHelpers.baseUrl}${url}`)
     .waitForElementVisible('body', Timeouts.normal);
```
jkassemi: Unfortunately setting the users' session prior to waiting for the body to show on this page provided inconsistent results. I'd love if there were a way to set the session on the first page request so we didn't need to make two for every login. This may be possible by passing a session parameter in the URL or finding a way for the react application to read the users' session id from a method or the redux state instead of directly from local storage.

test/util/mock-helpers.js
```diff
@@ -0,0 +1,16 @@
...
+  } : json;
+
+  request({
+    uri: `${E2eHelpers.apiUrl}/mock`,
+    method: 'POST',
+    json: jsonWithToken
+  });
+}
+
+module.exports = mock;
```
jkassemi: This is somewhat out of convention - might consider changing this to export multiple methods or singularizing the name of the file.

config/nightwatch.js
```diff
@@ -47,7 +45,12 @@ module.exports = {
         log_path: './logs/selenium',
         host: '127.0.0.1',
         port: selenium_server_port,
-      }
+      },
+      cli_args: { },
+      test_workers: {
+        enabled: true,
+        workers: 10,
```
alexose: 10 seems excessive... I think "auto" is a better option here.  Worth testing, maybe.

test/disability-benefits/e2e/00.claims-list.e2e.spec.js
```diff
@@ -1,50 +1,46 @@
-if (!process.env.BUILDTYPE || process.env.BUILDTYPE === 'development') {
```
alexose: I'm guessing we are ditching these now that we check buildtype at the `/script/run-nightwatch.js` level.  👍

test/util/login-helpers.js
alexose: That seems fine to me.  (If I've learned anything about Selenium in the last few days, it's to not rely on its APIs...)

test/util/login-helpers.js
alexose: Is this because localStorage isn't set until the body is already loaded?  If that's the case, I'd guess that the URL paramater approach would be the way to go.  It would just take a couple of changes in `test-server.js`, I think. 

Something for a future PR, perhaps!

test/util/mock-helpers.js
alexose: Seems fine for now.  The helpers.js files aren't particularly idiomatic anyway...

alexose (approved):
These all seem like solid, straightforward changes.  👍

config/nightwatch.js
jkassemi: I saw better responses discarding auto, which I believe defaults to the number of cpu's on the system. I'm guessing that the waits, pauses, and polling for elements to appear prevents the process from becoming cpu-bound.

test/util/login-helpers.js
jkassemi: I _believe_ that's the case. I'll set up to do some more testing in the future if we find there's a good performance improvement to be had from this.

jkassemi:
Test runtimes (observed):

| parallelization | unit | e2e | accessibility |
| --- | --- | --- | --- |
| no | 43s | 4m59s | 3m13s |
| yes | 44s | 1m13s | 1m13s |

PR #4486, opened 2016-12-23
[Messaging] Folder container unit tests.

U-DON:
### Changes
* Fleshed out the unit tests for Folder.
* Refactored pagination state and how it's used in Folder.
* Fixed redirect usage in Folder.
* Hide MessageNav in Thread until the messages list is populated and enables navigation.
* Reset current item in folder state when loading so no residual data remains and prevents error messages from displaying.

### DIFF
--- a/src/sass/messaging/partials/_messaging-count-nav.scss
+++ b/src/sass/messaging/partials/_messaging-count-nav.scss
@@ -30,6 +30,7 @@
   }
 
   &[disabled] {
+    &,
     &:hover,
     &:active {
       color: $color-gray;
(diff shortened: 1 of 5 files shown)


src/sass/messaging/partials/_messaging-count-nav.scss
```diff
@@ -30,6 +30,7 @@
   }
 
   &[disabled] {
+    &,
```
bshyong: What does this selector mean?

bshyong (approved):
LGTM

src/sass/messaging/partials/_messaging-count-nav.scss
U-DON: Since `&` takes the form of the selector it's nested under, `&[disabled]` here is `.messaging-message-nav button[disabled]`, and `&` again under that is the same (`.messaging-message-nav button[disabled]`).

PR #4488, opened 2016-12-23
Text alignment on list view

bshyong:
Resolves https://github.com/department-of-veterans-affairs/messaging-fe/issues/205

### DIFF
--- a/src/sass/rx/partials/_rx-prescriptions.scss
+++ b/src/sass/rx/partials/_rx-prescriptions.scss
@@ -50,7 +50,7 @@
 
   a {
     text-decoration: none;
-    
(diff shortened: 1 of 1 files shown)


src/sass/rx/partials/_rx-prescriptions.scss
```diff
@@ -50,7 +50,7 @@
 
   a {
     text-decoration: none;
-    
+
```
U-DON: We have a bunch of trailing spaces among the stylesheets. Might be nice to add a check for that to the linter.

U-DON (approved):
LGTM.

PR #4511, opened 2016-12-27
Address screenreader support on modals

### LINKED ISSUE #4492: 508 issues around status definition pop up
At the 508 office, our tester used IE 11 and when she clicked on refill status: in process link, focus didn't return back and the screen reader did not alert the user that a pop-up came up. However, I just tried it with the mac screen reader on chrome and it worked. I wonder if this is a browser specific problem. 

How it looks and reads on mac:
screen shot 2016-12-23 at 10 06 15 am

Also, user was not able to close out of the pop-up because the screen reader picked up on the content behind the overlay so she struggled to close out of the box. 

screen shot 2016-12-23 at 10 06 35 am

cc: @eli-lewis

bshyong:
- Potentially resolves https://github.com/department-of-veterans-affairs/vets-website/issues/4492

Testing with IE11 and Windows screen reader should be done to see if issues are resolved after this PR is merged in


U-DON (approved):
LGTM, although I can't immediately test this.

Looks like the GlossaryModal hasn't been refactored to use the common Modal component yet, so I just created #4517 for that. The common Modal has some focusing logic as well. For reference, #4378 made similar changes for the ConfirmRefillModal.

bshyong:
Sure I can refactor this. I wasn't sure why this one wasn't using the common modal so I originally left it as is.

If the refill modal doesn't have the screen reader problems then the focus logic is probably what we needed for Glossary too. I'll merge this for the aria labels and then do the refactor to the common component.

PR #4513, opened 2016-12-27
removes /old directory

crwallace:
This removes everything that was in `/old` from the previous jekyll version of `vets-website`.


webinista:
Seems sensible to me at first glance.

PR #4520, opened 2016-12-28
[Messaging] Added test data. Updated unit tests for reducers.

U-DON:
### Changes
* Added more test data, now in the same format as the API responses.
    * ~Will probably swap out the previous test data at some point.~
    * Also removed the old test data.
    * Should also be useful for E2E tests.
* Added a couple of missing unit tests for loading reducer.
* Added unit tests for messages reducer.
    * Closes department-of-veterans-affairs/messaging-fe#101.
* Updated unit tests for search reducer.
    * Closes department-of-veterans-affairs/messaging-fe#120.


bshyong (approved):
LGTM! Test data looks great

ayaleloehr:
So many tests added recently @bshyong and @U-DON . Awesome work!!

PR #4522, opened 2016-12-29
Use common modal component for glossary modal

### LINKED ISSUE #4517: Refactor GlossaryModal to use common Modal component

bshyong:
- Resolves #4517

### DIFF
--- a/src/js/rx/components/GlossaryModal.jsx
+++ b/src/js/rx/components/GlossaryModal.jsx
@@ -1,17 +1,12 @@
 import React from 'react';
 import { Link } from 'react-router';
+import Modal from '../../common/components/Modal';
 import GlossaryList from './GlossaryList';
 
 class GlossaryModal extends React.Component {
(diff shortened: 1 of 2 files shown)


src/js/rx/components/GlossaryModal.jsx
```diff
@@ -23,28 +18,32 @@ class GlossaryModal extends React.Component {
     let element;
     if (this.props.isVisible) {
```
U-DON: This should probably be added to the propTypes, but I've also noticed before that it'd be redundant to do handling around the visibility of the modal when Modal already does that.

And looking at the propTypes of Modal, it doesn't seem like we should require the content, since most use cases have an initially hidden modal that doesn't have content.

Anyway, just a couple of thoughts on how the apps are using modals, and not something that needs to change for this PR.

U-DON (approved):
LGTM.

PR #4527, opened 2016-12-29
Update title in index.md

jcmeloni-zz:
Per @emilyville suggestion in https://github.com/department-of-veterans-affairs/vets.gov-team/issues/499 to solve issue of title displayed when sharing links.

### DIFF
--- a/content/pages/index.md
+++ b/content/pages/index.md
@@ -1,7 +1,7 @@
 ---
 layout: home.html
 body_class: home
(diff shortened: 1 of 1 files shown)


carlmjohnson:
This works for the homepage, but in general, we should have YAML front matter for "description" that is different than just the title.

jcmeloni:
ping @crwallace

PR #4532, opened 2016-12-30
Handle requests that fail from expired userToken in Rx and SM.

### LINKED ISSUE #3843: Expired userToken in the middle of a session doesn't automatically redirect to login

U-DON:
Closes #3843.

### Changes
* Merged `authFetch `and `apiRequest` in SM back together.
    * Updated usage in `MessageAttachmentsViewItem` to also revert to base state when request fails.
* Refactored `apiRequest` in Rx to be same as SM.
* Rx and SM do a page refresh upon getting 401 in order to go back to the login view.

### DIFF
--- a/src/js/messaging/components/MessageAttachmentsViewItem.jsx
+++ b/src/js/messaging/components/MessageAttachmentsViewItem.jsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import classNames from 'classnames';
 
-import { authFetch } from '../utils/helpers';
+import { apiRequest } from '../utils/helpers';
 
 class MessageAttachmentsViewItem extends React.Component {
   constructor(props) {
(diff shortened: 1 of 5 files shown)


bshyong (approved):
LGTM;  not sure why one of the RX tests is failing but if we can't get it to pass I'll pull down this branch later to see what's going on.

U-DON:
It was the path for the mock request that was failing the test. I guess it doesn't handle differences in trailing slashes.

PR #4541, opened 2017-01-03
end to end test for secure messaging

bshyong:
Resolves https://github.com/department-of-veterans-affairs/messaging-fe/issues/104

I spent a lot of time figuring out why polyfilled functionality wasn't working with the testing suite; namely `Map` and `Set` were not being recognized, but after trying various approaches I got it to work by explicitly importing `core-js` at the top of the entry file. I considered switching to babel runtime transform instead of requiring babel-polyfill globally but that was breaking our build pipeline (and it's a big change)


test/messaging/00-required.e2e.spec.js
```diff
@@ -0,0 +1,49 @@
...
+      client
+        .waitForElementVisible('#messaging-nav', Timeouts.slow)
+        .waitForElementVisible('#messaging-content-header', Timeouts.slow)
+        .waitForElementVisible('#messaging-folder-controls', Timeouts.normal)
+        // expect messages to show up
+        .expect.element('.msg-table-list td:nth-of-type(1) a:nth-of-type(1)').text.to.equal('Clinician');
+
+      // Compose message view
+      client
+        .click('#messaging-folder-controls .messaging-compose-button')
```
U-DON: On desktop, the way to access it would be from the FolderNav component, while this particular button would appear only on mobile. I'm not sure about the right way to do it, but it might make sense to do mobile E2E separately?

test/messaging/00-required.e2e.spec.js
```diff
@@ -0,0 +1,49 @@
...
+        .waitForElementVisible('textarea[name="messageText"]', Timeouts.normal);
+      // select a recipient in the compose form
+      client.click('select[name=\'messageRecipient\']')
+        .click('select option[value=\'0\']')
+        .keys(['\uE006']);
```
U-DON: What does this do?

U-DON (approved):
LGTM, but had some questions. Also, should there be a test here for viewing a thread?

test/messaging/00-required.e2e.spec.js
bshyong: The default screen size is mobile sized - we can make a different test for desktop and resize the window so that it's larger. :+1:

test/messaging/00-required.e2e.spec.js
bshyong: This sequence presses the Return key so that the option is selected and the select box closes - I think the programmatic selection might not trigger the select box to close when the option is clicked.

PR #4560, opened 2017-01-03
Fix issues with new education benefits url

jbalboni:
There were some missed paths around the review and submit process that are causing the form to go to the confirmation page without submitting. This should fix them.


akainic (approved):
LGTM!

Issue #4563, opened 2017-01-04
Desktop E2E test for Messaging


bshyong:
Also need to add e2e test for Thread view

PR #4565, opened 2017-01-04
More end to end tests for Secure Messaging

bshyong:
Resolves https://github.com/department-of-veterans-affairs/vets-website/issues/4563

Adds test for Desktop view
Adds test for Thread view for both Mobile and Desktop

### DIFF
--- a/test/messaging/00-required.e2e.spec.js
+++ b/test/messaging/00-required.e2e.spec.js
@@ -23,6 +23,20 @@ if (!process.env.BUILDTYPE || process.env.BUILDTYPE === 'development') {
         // expect messages to show up
         .expect.element('.msg-table-list td:nth-of-type(1) a:nth-of-type(1)').text.to.equal('Clinician');
 
+      client.click('.msg-table-list td:nth-of-type(1) a:nth-of-type(1)');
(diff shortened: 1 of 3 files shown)


U-DON (approved):
LGTM. There are some merge conflicts, though?

bshyong:
Hmm will take a look and fix

bshyong:
merge conflicts fixed

U-DON:
Parts of the E2E tests aren't passing.

```
test/messaging/01-required-desktop.e2e.spec ✖ TypeError: client.url is not a function
```

```
test/messaging/00-required.e2e.spec ✖ Expected element <.messaging-thread-messages .messaging-thread-message:last-of-type .messaging-message-body> text to equal: "Reply 3" - element was not found - expected "present" but got: "not present"
```

bshyong:
Found the bug—due to the use of auth token in mock requests. Should be working now

U-DON:
Still LGTM, but should the commented out blocks be removed, since we're using `mock` now instead of just `request`?

bshyong:
oof—forgot to remove that after debugging. Removed.

PR #4581, opened 2017-01-04
Refactored search in SM to allow clearing search

U-DON:
Closes department-of-veterans-affairs/messaging-fe#147.

### Changes
* Removed 'term' from search state in favor of sharing 'subject' query between basic and advanced search, since 'term' was just a non-exact search for 'subject'.
* Updated and added unit tests for some search-related modules (reducer and MessageSearch component).
* Allow a blank search to clear the search and go back to viewing the entire folder.


bshyong (approved):
LGTM!

PR #4586, opened 2017-01-05
Moves Rx and SM to single store

crwallace:
This moves Rx and Secure Messaging into the `commonStore`. This is the first phase of moving everything to a single store.

edit: This also moves `rx` and `messaging` reducers under the `health` reducer inside of the `commonStore`.

This relates to department-of-veterans-affairs/vets.gov-team#581


U-DON (approved):
LGTM. Only thing I'm not sure about is whether it was necessary to move each app's reducers into common if they were going to be namespaced there anyway. I think the common reducer could've imported them like:

```
import rx from '../../rx/reducers';
import messaging from '../../messaging/reducers';
...
health: combineReducers({ rx, messaging })
...
```

src/js/common/reducers/index.js
```diff
@@ -1,10 +1,45 @@
...
+import prescriptions from '../../common/reducers/rx/prescriptions';
+
+import alert from '../../common/reducers/messaging/alert';
+import compose from '../../common/reducers/messaging/compose';
+import folders from '../../common/reducers/messaging/folders';
+import loading from '../../common/reducers/messaging/loading';
+import messages from '../../common/reducers/messaging/messages';
+import modals from '../../common/reducers/messaging/modals';
+import recipients from '../../common/reducers/messaging/recipients';
+import search from '../../common/reducers/messaging/search';
```
U-DON: If we're keeping all the other apps' reducers under `common/reducers`, these paths won't need the `../../common/reducers/` in front.

crwallace:
I agree with the moving reducers comment. I thought about that once I put the PR up since I was not moving actions. I think I am going to move directories back as not to confuse people that are working on these apps.

src/js/common/reducers/index.js
```diff
@@ -1,10 +1,44 @@
...
 
-import login from '../../common/reducers/login';
-import profile from '../../common/reducers/profile';
+import login from '../../login/reducers/index';
+import profile from '../../user-profile/reducers/index';
+
+import alertRx from '../../rx/reducers/alert';
+import disclaimer from '../../rx/reducers/disclaimer';
+import modalsRx from '../../rx/reducers/modals';
+import prescriptions from '../../rx/reducers/prescriptions';
```
U-DON: Could we just import the combined reducer from each app? Something like:

```
import rxReducer from '../../rx/reducers';
import msgReducer from '../../messaging/reducers';
```

It would reduce the number of imports in this module, especially the multiple `alert*` and `modal*`. And any changes to a particular reducer would be localized to the app.

PR #4589, opened 2017-01-05
Refills left column should be red when 0 remaining

### LINKED ISSUE #4465: Refills left column- 0 should be red
Rx Refill->Refill prescriptions tab->List view
0 refills- color should change to red, currently still showing as black

zero should be red

bshyong:
Resolves #4465


U-DON (approved):
LGTM.

PR #4590, opened 2017-01-05
WIP: DO NOT MERGE proposed hca scheduled maintenance message

goldenmeanie:
Flag is flipped to

in_maintenance: **true**

### DIFF
--- a/content/layouts/page-react.html
+++ b/content/layouts/page-react.html
@@ -8,7 +8,7 @@
           <div style="padding: 2em 0;">
           <h3>{{ maintenance_line1 }}</h3>
           <h4>{{ maintenance_line2 }}</h4>
-          <a href="/"><button>Go back to Vets.gov</button></a>
+          <a href="/"><button>Go back to the Home page</button></a>
           </div>
         </div>
       </div>
--- a/content/pages/healthcare/apply/application.md
+++ b/content/pages/healthcare/apply/application.md
@@ -3,9 +3,10 @@ title: Apply for Health Care
 entryname: hca
 layout: page-react.html
 body_class: page-healthcare
-in_maintenance: false
-maintenance_line1: We're sorry. The healthcare application is currently down while we fix a few things. We will be back up as soon as we can.
-maintenance_line2: In the meantime, you can still call 1-877-222-VETS(8387) and press 2 to complete this application over the phone.
+in_maintenance: true
+maintenance_line1: Scheduled Maintenance
+maintenance_line2: Please try again after 12pm EST Saturday, January 7th.
+
 ---
 <div id="main">
   <nav class="va-nav-breadcrumbs">
@@ -27,4 +28,3 @@ maintenance_line2: In the meantime, you can still call 1-877-222-VETS(8387) and 
   </div>
   <!-- HCA Application End -->
 </div>
-
--- a/content/pages/healthcare/apply/application.md
+++ b/content/pages/healthcare/apply/application.md
@@ -5,7 +5,7 @@ layout: page-react.html
 body_class: page-healthcare
 in_maintenance: true
 maintenance_line1: Scheduled Maintenance
-maintenance_line2: Please try again after 12pm EST Saturday, January 7th.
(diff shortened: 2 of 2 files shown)


content/layouts/page-react.html
```diff
@@ -8,7 +8,7 @@
           <div style="padding: 2em 0;">
           <h3>{{ maintenance_line1 }}</h3>
           <h4>{{ maintenance_line2 }}</h4>
-          <a href="/"><button>Go back to Vets.gov</button></a>
+          <a href="/"><button>Go back to the Home page</button></a>
```
akainic: @emilyville @crwallace could you confirm this change is okay to make site-wide?

content/pages/healthcare/apply/application.md
```diff
@@ -3,9 +3,10 @@ title: Apply for Health Care
 entryname: hca
 layout: page-react.html
 body_class: page-healthcare
-in_maintenance: false
-maintenance_line1: We're sorry. The healthcare application is currently down while we fix a few things. We will be back up as soon as we can.
-maintenance_line2: In the meantime, you can still call 1-877-222-VETS(8387) and press 2 to complete this application over the phone.
+in_maintenance: true
+maintenance_line1: Scheduled Maintenance
```
akainic: Do we not want to use more friendly, warm language here? This feels a bit cold to me.

content/pages/healthcare/apply/application.md
```diff
@@ -3,9 +3,10 @@ title: Apply for Health Care
...
+maintenance_line2: Please try again after 12pm EST Saturday, January 7th.
```
akainic: Do we still want to provide the call center number as a back-up option?

akainic (commented):
Thanks for taking a stab at this! Added a few comments/questions.

content/pages/healthcare/apply/application.md
goldenmeanie: Is the call center open on Saturday?

akainic: Also, really sorry this is my fault, but I got the dates wrong. It's Sunday, January 8th they're doing maintenance, not Saturday. Sorry!

akainic: Good question! I don't think they are yet. So nevermind!

content/pages/healthcare/apply/application.md
goldenmeanie: Making it friendlier was adding a lot of words that reduced clarity.

akainic: We don't want to keep the original language and just replace "down while we fix a few things" with "under scheduled maintenance"? So:
```
We're sorry. The health care application is currently under scheduled maintenance.
Please try again after 12pm EST Sunday, January 8th.
```
Or something like that?

Also wondering if we want to commit to an exact time. Just in case maintenance runs long, is there another way to say when they should come back without naming a time? Like "Please try again in a few hours"? Or "Please try again in the afternoon on Sunday, January 8th"?

content/layouts/page-react.html
crwallace: I am going to defer to @emilyville and @maryannbrody on this.

content/pages/healthcare/apply/application.md
maryannbrody: Jumping in here to say 1) let's stick with conversational language for error messages -- @akainic's suggestion works for me, and 2) I would keep the timing as specific as possible, but don't undershoot it. If there's a chance that it will take an extra 6 hours, let's buy ourselves some time.

akainic: Great, perhaps let's us 2pm then? I think that gives us enough buffer room.

content/layouts/page-react.html
maryannbrody: @crwallace - LGTM, except "Home" does not need to be capitalized. Thanks for checking!

akainic:
Closing as no longer relevant.

Issue #4439, opened 2016-12-20
Footer content update


emilyville:
Updating the footer to include new things like call center, with some adjustments to accommodate. All the links match pre-existing things. 
![footer-contentchange](https://cloud.githubusercontent.com/assets/1307774/21335254/6c76f660-c612-11e6-9cee-dd809608ebc2.png)

![footer-contentchange-mobile](https://cloud.githubusercontent.com/assets/1307774/21335347/009e85a6-c613-11e6-9a28-85b881ef64dd.png)

[Will update with copy from the images into the issue tomorrow, I need th chat URL still]

@AnitraAppa let me know if you have any edits/thoughts, I believe this is all consistent with the homepage aside from the help desk being added!

The only change in functionality is that the crisis line (now in the header and the footer) should open the modal on desktop like it does on mobile, and remove the down carat from the top.

AnitraAppa:
@emilyville Thanks for flagging! Some thoughts and content recos:
- Use sentence case instead of title case. Easier to read, save title case for biggest header on a page. (Going to do a sentence case scrub across the site - thanks for filing, Gina!)
- Under "Services and Benefits" we have a bunch of links for main pages that prominently show cards that link to the same things under the "Popular on Vets.gov" section. Example: We have both "Heath care" and "Apply for Health Care", "Education Benefits" and "Education Benefits by School". To simplify the footer and avoid redundant looking links, could we just link to the main sections?
- I put "Health care" on the top of the "Resources" section because I believe it's the most used section - is that right?
- I moved "Find Nearby VA locations" under "Services" instead
- Changed "Careers and Employment" to "Employment help"

Took a pass below:

**BENEFITS**
Health care
Disability
Education

**SERVICES**
Nearby VA facilities
Employment help
Refill prescriptions

**CONTACT**
Call the Vets.gov Help Desk
1-855-574-7286 or use chat
Monday‒Friday, 8:00 a.m.‒8:00 p.m. (ET)
Get help from the Veterans Crisis Line

dannychapman:
I really like this! One thought on the horizontal links at the bottom - it might be worth upping the contrast / weight for accessibility.

crwallace:
@emilyville -- I would +1 @dannychapman's comment. Can you either attach the sketch file here or drop it in the core team channel so we can see all of the colors, fonts, font-weights, etc?

emilyville:
[...]

PR #4599, opened 2017-01-05
BUG: includes/footer.html: Add the same links as the /gi-bill-comparison-tool/ footer

westurner:


jcmeloni:
@crwallace @webinista @melwoodard Heads up on this PR as it relates to other work ongoing re: the footer and blog/vec links which I believe are going away. I think we'll actually end up removing the links from GIBCT?

melwoodard:
@jcmeloni that's correct - removal of those links was intentional, with the next iteration of footer content planned in #4439. Once GIBCT is migrated to vets-website it will use includes/footer.html

webinista:
@jcmeloni Related to #4594. Should we close this PR since the plan is to remove the links from GIBCT, not add them to `/includes/footer.html`?

jcmeloni:
@webinista That makes sense to me, yeah

PR #4605, opened 2017-01-05
Form scaffolding for 1995

### LINKED ISSUE #4462: Validate the use of react-jsonschema-form for building hca/edu style forms
- [x] Fork repo into va org
- [x] Add the ability to show validation messages based on touched state of fields (make this optional so we can potentially PR it) 

Try out the following validation approaches:

- [x] Create a hook to replace messages from jsonschema validation with our own.
- [x] Figure out how to adapt jsonschema required validation to our current pattern.
- [x] Figure out how to do custom field type validation (e.g. our date validation)
- [x] Determine if we need to replace jsonschema validation with our own validation.

Build some example forms:

- [x] Simple example form that uses our field styles
- [x] Built out custom field (date field)
- [x] Growable tables
- [x] Date range validation
- [x] Build an example with the ExandableGroup pattern (or conditional fields)
- [x] Test out performance on mobile

jbalboni:
Closes #4462 

This has some initial set up for using react-jsonschema-form to generate forms. I've tried to strip it down as much as possible, but it's still pretty big. There's more prototype code in [this gist](https://gist.github.com/jbalboni/52a9183490e1608f5881d7f72b03370c) (includes growable table, expandable fields, and a date field), but it can be reviewed when we add fields/pages that need that code.

You can go to `/1995/introduction` and there's that page and a vet info page with validation. There is not a normal date field right now and there's nowhere to go after you fill in all the information, but it should look mostly fine.

The code for the 1995 form is all in src/js/edu-benefits/1995, mainly in config/form.js. The code for generating the form is all in src/js/common/schemaform for now.

Before this is merged, I'm going to comment out the code in the entry file, since making this lazy load is a bigger task.

screen shot 2017-01-05 at 4 15 06 pm
screen shot 2017-01-05 at 4 14 59 pm
screen shot 2017-01-05 at 4 14 55 pm


lihanli:
what should the backend do for validation, are you moving the schema to vets-json-schema later?

jbalboni:
@lihanli I think that's an open question. I would like to share this schema with the backend, but it's a big drag on the speed of making changes to a form built this way to have it in a separate repo. The overall schema for the frontend is also going to be more complicated than what the backend needs.

lihanli:
we can put a data-only schema in vets-json-schema and have the frontend import and modify it?

jbalboni:
@lihanli We can, but that's what I'm worried about being a drag on building the front end. Even if it's just the json schema part of the config info here, it still includes things like title and description properties, plus we'll likely group some fields in objects in order to do certain ui customizations.

lihanli:
if you change vets-json-schema to be a local path it shouldn't be much slower https://docs.npmjs.com/files/package.json#local-paths
or we can wait until everything's almost done and move the schema out at the end?

jbalboni:
@lihanli Looks interesting, I'll check that out.

I think for now you should build a schema for the backend based on the 1990 one and we should try to sync up before we go to production.

src/js/common/schemaform/helpers.js
```diff
@@ -0,0 +1,75 @@
+import React from 'react';
+import _ from 'lodash/fp';
+import { Route, IndexRedirect } from 'react-router';
+import FormPage from './FormPage';
+
+function createFormPageList(formConfig) {
+  return Object.keys(formConfig.chapters)
+    .reduce((acc, chapter) => {
```
akainic: What is `acc`?

src/js/edu-benefits/1995/config/form.js
```diff
@@ -0,0 +1,66 @@
...
+const formConfig = {
+  urlPrefix: '/1995/',
+  introduction: IntroductionPage,
+  confirmation: null,
+  chapters: {
+    veteranInformation: {
+      title: 'Veteran Information',
+      pages: {
+        veteranInformation: {
+          path: 'applicant-information',
```
akainic: Is this what we want to call this path? Instead of `veteran-information`? I think that's what we're using now for the 1990?

src/js/edu-benefits/1995/config/form.js
```diff
@@ -0,0 +1,66 @@
...
+          initialData: {},
```
akainic: What will this be used for?

src/js/common/schemaform/FormPage.jsx
```diff
@@ -0,0 +1,158 @@
...
+    const { uiSchema } = this.props.route.pageConfig;
+    if (uiSchema) {
+      uiSchemaValidate(errors, uiSchema, formData);
+    }
+    return errors;
+  }
+  render() {
+    const { schema, uiSchema } = this.props.route.pageConfig;
+    return (
+      <Form
```
akainic: `Form` comes from the library?

src/js/common/schemaform/definitions.js
```diff
@@ -0,0 +1,43 @@
...
+        'IV'
+      ]
+    }
+  }
+};
+
+export const ssn = {
+  title: 'Social security number',
+  type: 'string',
+  pattern: '^([0-9]{3}-[0-9]{2}-[0-9]{4}|[0-9]{9})$'
```
akainic: How do these validations work? This differs from our front-end validation for SSN now, so wondering when/how this will be used vs. what lives in `validations.js`.

src/js/common/schemaform/FormPage.jsx
```diff
@@ -0,0 +1,158 @@
...
+    const pageIndex = _.findIndex(item => item.pageKey === pageConfig.pageKey, pageList);
+    this.props.router.push(pageList[pageIndex + 1].path);
+  }
+  getEmptyState(pageConfig) {
+    return { formData: pageConfig.initialData, formContext: { touched: {}, submitted: false } };
+  }
+  goBack() {
+    const { pageList, pageConfig } = this.props.route;
+    const pageIndex = _.findIndex(item => item.pageKey === pageConfig.pageKey, pageList);
+    this.props.router.push(pageList[pageIndex - 1].path);
```
akainic: Do we want `scrollToTop()` here? To make the transition in going back a page a little smoother.

akainic (commented):
This is so much awesome work! Thanks for getting us going on this. A couple of things:
1. Left a bunch of little comments, mostly questions for you to make sure I understand.
2. Let's just make sure we're preventing this from being deployed to production before merging, either by commenting out the Route or asking @crwallace if the ignoreList can be used.

src/js/common/schemaform/helpers.js
jbalboni: Accumulator. It's a sort of common generic name for that parameter in a reduce function, but I should find a more descriptive name.

src/js/edu-benefits/1995/config/form.js
jbalboni: I wasn't sure since the title of the first page is Applicant information in the GH issue for the first page.

src/js/edu-benefits/1995/config/form.js
jbalboni: If you want to set an initial value for any of the fields in the form. We don't need to initialize everything with empty strings like we do now in veteran.js, but we do sometimes set dates to the current date initially, for example.

src/js/common/schemaform/FormPage.jsx
jbalboni: If you're going back to another page in the form, you should get scrolled to the top in the `componentDidUpdate` method. We may want to add a scroll to the Intro page, though.

src/js/common/schemaform/FormPage.jsx
jbalboni: Yes.

src/js/common/schemaform/definitions.js
jbalboni: This is used in the JSON schema validation, which the library runs internally. The stuff in validations.js runs in addition to that validation, so that we can do checks that aren't possible in JSON schema.

I'm not sure what the best way to deal with the two separate sets of validations is. If we didn't care about being in sync with JSON schema validation on the backend, we might want to just turn off the JSON schema validation and use our own.

src/js/edu-benefits/1995/config/form.js
jbalboni: I'm going to change it back to Veteran information for now.

src/js/common/schemaform/FormPage.jsx
akainic: I think that's what I was seeing when testing things locally on the Intro page, makes sense

jbalboni:
@akainic I put the 1995 route behind a build type check. I'm still going to comment the routes out until we make the changes to lazy load the 1995 form, but this makes me more comfortable that we won't accidentally put something in prod.

cc: @crwallace

src/js/common/schemaform/definitions.js
akainic: Hm okay. I thought I noticed when I was running this locally that it only seemed to run the validations in the schema, but I could be wrong.

jbalboni: It looks like it's working for me. 111111111 is valid with just the schema validation, but not with the additional validation.

akainic (approved):
LGTM!

PR #4607, opened 2017-01-05
Move education Address component to common

akainic:
Moving the education Address component into the `common` directory and changing paths in education to reference the new common component. 

Have not touched HCA yet.

Potential impact: None for education benefits. Just want to make sure HCA can accept AA/AE/AP for states, as those are now options in the dropdown. It validates locally, and should pass schema as well (those states are listed in the schema constants).


jbalboni (approved):
LGTM. Works for me locally.

PR #4609, opened 2017-01-05
Move Playbook partial from style.scss to no-react.scss. Closes #3997

### LINKED ISSUE #3997: Remove playbook styles from non-playbook pages
Playbook styles are being loaded for every page and I don't see any reason they should be.

Also, maybe we should remove the playbook.

webinista:


crwallace (approved):
LGTM

PR #4610, opened 2017-01-06
[SM] Fill in search fields with query params after successfully loading folder.

U-DON:
Closes department-of-veterans-affairs/messaging-fe#149.

### DIFF
--- a/src/js/messaging/reducers/search.js
+++ b/src/js/messaging/reducers/search.js
@@ -62,8 +62,8 @@ export default function modals(state = initialState, action) {
 
       if (subject) {
         params.subject.field = !!subject.eq
-                          ? makeField(subject.eq, true)
-                          : makeField(subject.match, true);
+                             ? makeField(subject.eq, true)
+                             : makeField(subject.match, true);
         params.subject.exact = !!subject.eq;
       }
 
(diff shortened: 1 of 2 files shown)


U-DON:
I left a little code in the test for filtering on recipient, which isn't relevant yet, but will be after I get the code in for showing a 'To' field instead of 'From' for searching in Drafts and Sent.

src/js/messaging/reducers/search.js
```diff
@@ -34,6 +35,40 @@ const initialState = {
 
 export default function modals(state = initialState, action) {
   switch (action.type) {
+    case FETCH_FOLDER_SUCCESS: {
+      const { filter } = action.messages.meta;
+      if (!filter) { return state; }
+
+      const params = {
+        dateRange: { ...initialState.params.dateRange },
```
bshyong: Would this be the same as `dateRange: initialState.paras.dateRange`?

bshyong (approved):
LGTM /w a minor question

src/js/messaging/reducers/search.js
U-DON: I just wanted a deep enough clone of `initialState`. If it were `dateRange: initialState.params.dateRange`, I think setting `params.dateRange.start = newDate` would affect `initialState.dateRange`.

bshyong: :+1:

PR #4616, opened 2017-01-06
Restoring clickability to mobile VCL links.

webinista:


crwallace:
@webinista -- Looks like mobile is fixed but desktop is still not working.

webinista:
@crwallace Just pushed another change.

crwallace (approved):
LGTM

PR #4617, opened 2017-01-06
[TEST] - JK travis memory tests

jkassemi:


jkassemi:
Travis recommends using sudo container for better guarantees on memory. I'm not interested in developing this further at this point with the Jenkins work on the horizon, as the gains there are guaranteed.

Issue #4592, opened 2017-01-05
508 Testing- SM


eli-lewis:
User was able to successfully log in to vets.gov, messaging.  When she navigated to "compose message" the screen reader did not recognize any content.  There were no alerts and focus was not taken to any of the fields.  The same happened with all the folders.  As a result the user was unable to compose, read, send, save messages.

bshyong:
Was this on Internet Explorer and Windows?

eli-lewis:
yes.  IE11

bshyong:
@eli-lewis do you know the screen reader and operating system that was used for testing?

eli-lewis:
Jaws was the screen reader, she was using a PC, not sure version.

PR #4618, opened 2017-01-06
Add aria-live to watch for content changes

bshyong:
- Resolves https://github.com/department-of-veterans-affairs/vets-website/issues/4592

I think that the accessibility issues are caused by a content change on the page triggered by Javascript. The compose view and all of the folder view are loaded on the page in the absence of a full page refresh.

I tried a few different things but settled on this very minor change which is the cleanest way. It is supported by the latest versions of JAWS and NVDA on IE. Adding various focus events on `componentDidMount` was messy because of load events that occur in rapid succession after the mount, e.g. fetching recipients.

After this is merged, we should retest with a windows PC/IE 11 screen reader to confirm that `aria-live` is sufficient to alert the screen reader that content has changed, which should make all of the content accessible.

FYI references for more info:
- https://msdn.microsoft.com/en-us/library/dd433065(v=vs.85).aspx
- https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions
- screen reader/browser usage % http://webaim.org/projects/screenreadersurvey6/

### DIFF
--- a/src/js/messaging/containers/Main.jsx
+++ b/src/js/messaging/containers/Main.jsx
@@ -114,7 +114,7 @@ export class Main extends React.Component {
               onCreateNewFolder={this.props.openCreateFolderModal}
               onFolderChange={this.handleFolderChange}/>
         </div>
-        <div id="messaging-content">
+        <div id="messaging-content" aria-live="assertive">
           {this.props.children}
         </div>
         <ModalAttachments


U-DON:
I think when this happened, the user was able to access the folders in the side bar, but those get rendered after an API response as well. Maybe that content changed quickly or early enough in the page load that it wasn't an issue?

U-DON (approved):
LGTM. Don't know for sure if it will work, but worth trying out.

U-DON:
I also wonder if this changes the behavior experienced by other screen readers that were already able to access the main content.

bshyong:
This explicitly informs screenreaders that the content in the div is dynamic and they should watch for changes, announcing when a change occurs. I think what happened with the compose and folder views is that only a part of the content changes (the left hand nav stays the same)—maybe a reflow is not triggered? If the problem is still occuring we can try adding `tabIndex=-1` as well to the content.

PR #4620, opened 2017-01-06
Reduce nightwatch parallelization

jkassemi:
Working with Travis to determine memory constraint issues that may be
killing processes mid-test when a larger number of processes are used
to run the E2E tests.

This has potentially caused itermittent failures in the form of timeouts
waiting for elements to appear and disconnections from the browser client.

### DIFF
--- a/config/nightwatch.js
+++ b/config/nightwatch.js
@@ -46,7 +46,7 @@ module.exports = {
       },
       test_workers: {
         enabled: true,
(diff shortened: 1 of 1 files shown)


crwallace:
This LGTM but hilariously enough I think you are having some timeout failures. I think we need to change the github configs to allow admins to merge even if tests have not passed.

PR #4631, opened 2017-01-07
[SM] Filter by 'To' instead of 'From' in Drafts and Sent folders.

U-DON:
Closes department-of-veterans-affairs/messaging-fe#208.

### Changes
* Show 'To' field in advanced search for Drafts and Sent folders instead of 'From'.
* Clear search fields when changing folders.
* Added unit tests for MessageSearchAdvanced component and more tests for existing search modules.

### DIFF
--- a/src/js/messaging/containers/Folder.jsx
+++ b/src/js/messaging/containers/Folder.jsx
@@ -92,6 +92,8 @@ export class Folder extends React.Component {
     const queryParams = [
       'page',
       'sort',
+      'filter[[recipient_name][eq]]',
+      'filter[[recipient_name][match]]',
       'filter[[sent_date][gteq]]',
       'filter[[sent_date][lteq]]',
       'filter[[sender_name][eq]]',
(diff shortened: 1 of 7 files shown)


src/js/messaging/containers/Folder.jsx
```diff
@@ -112,23 +114,33 @@ export class Folder extends React.Component {
...
   buildSearchQuery(object) {
     const filters = {};
+    const fromValue = _.get(object, 'from.field.value');
+    const toValue = _.get(object, 'to.field.value');
+    const subjectValue = _.get(object, 'subject.field.value');
+    const startDate = _.get(object, 'dateRange.start');
+    const endDate = _.get(object, 'dateRange.end');
 
-    if (_.get(object, 'from.field.value')) {
+    if (fromValue) {
```
bshyong: 👍

src/js/messaging/containers/Folder.jsx
```diff
@@ -364,6 +376,7 @@ export class Folder extends React.Component {
       let messageSearch;
       if (this.props.messages && this.props.messages.length || this.props.filter) {
         messageSearch = (<MessageSearch
+            hasRecipientField={folderName === 'Sent' || folderName === 'Drafts'}
```
bshyong: we can use `indexOf` if there are more than 2 elements we need to match in the future

bshyong (approved):
LGTM

PR #4632, opened 2017-01-07
Unit test for Prescription card

### LINKED ISSUE #2940: [Rx] Write unit tests for Prescription card.

bshyong:
Resolves #2940


test/rx/components/Prescription.unit.spec.jsx
```diff
@@ -0,0 +1,87 @@
...
+      },
+    };
+
+    const tree = SkinDeep.shallowRender(<Prescription {...newProps}/>);
+    expect(tree.subTree('.rx-call-provider')).to.exist;
+  });
+
+  it('should not show Message Provider link if refills remaining', () => {
+    const tree = SkinDeep.shallowRender(<Prescription {...props}/>);
+    expect(tree.subTree(['.rx-call-provider'])).to.be.false;
```
U-DON: I don't think the brackets are needed for `subTree()`.

test/rx/components/Prescription.unit.spec.jsx
```diff
@@ -0,0 +1,87 @@
...
+      },
+    };
+
+    const tree = SkinDeep.shallowRender(<Prescription {...newProps}/>);
+    expect(tree.subTree('SubmitRefill')).to.exist;
+  });
+
+  it('should show refill status if not refillable', () => {
+    const tree = SkinDeep.shallowRender(<Prescription {...props}/>);
+    expect(tree.subTree(['.rx-prescription-status'])).to.exist;
```
U-DON: Brackets for `subTree()` probably not needed here either.

test/rx/components/Prescription.unit.spec.jsx
```diff
@@ -0,0 +1,87 @@
...
+      },
+    };
+
+    const tree = SkinDeep.shallowRender(<Prescription {...newProps}/>);
+    expect(tree.subTree('TrackPackageLink')).to.exist;
+  });
+
+  it('should not render tracking link if not applicable', () => {
+    const tree = SkinDeep.shallowRender(<Prescription {...props}/>);
+    expect(tree.subTree(['.rx-track-package-link'])).to.be.false;
```
U-DON: Should probably be `TrackPackageLink` to be consistent with the opposite test above.

test/rx/components/Prescription.unit.spec.jsx
```diff
@@ -0,0 +1,87 @@
...
+    const newProps = {
+      ...props,
+      attributes: {
+        ...props.attributes,
+        isTrackable: true,
+      },
+    };
+
+    const tree = SkinDeep.shallowRender(<Prescription {...newProps}/>);
+    expect(tree.subTree('TrackPackageLink')).to.exist;
```
U-DON: :+1: Haven't come across this one. Probably should've used this instead of `to.not.be.false` in other tests.

U-DON (approved):
Generally LGTM with some minor comments.

Also, for tests that only use the default props (like 'should not show Message Provider Link if refills remaining`), I wonder if they should explicitly define the props that are relevant to that case in case the test data changes in the future. Or maybe at least have a comment about what value the prop is set to.

test/rx/components/Prescription.unit.spec.jsx
bshyong: yep this is tricky because `undefined !== false` so one of my tests was passing even though it shouldn't have been

bshyong:
Good point on the test data—I moved it into the test to make it more explicit and easy to change in the future

PR #4638, opened 2017-01-09
Add save progress alert for education benefits

molindoz:
Closes: [#132](https://github.com/department-of-veterans-affairs/edu-benefits-team/issues/132)

## Screenshot
screen shot 2017-01-09 at 11 52 56 am


cagedcrown:
I'm currently in the process of testing the vets-api locally to make sure everything works after the 1990 form is submitted.

jbalboni (approved):
This looks good to me. Once you test that the message doesn't show up on the confirmation page, I think it can be merged.

austin-martinez:
Is there anyway to make this reusable? The same feature was implemented for HCA

jbalboni:
@austin-martinez We're working on an approach to form building that shares more code directly and we'd be able to share something like this across forms if we end up going that direction.

PR #4639, opened 2017-01-09
Rename conditions to contentions

austin-martinez:
Closes https://github.com/department-of-veterans-affairs/sunsets-team/issues/292
screen shot 2017-01-09 at 10 29 28 am


jbalboni (approved):
LGTM. Do you think it'd be worth changing the class names? It'd be clearer, but change them is error prone.

PR #4642, opened 2017-01-09
Don't show card/list toggle if an error occurs

### LINKED ISSUE #4533: Hide card/list toggle when there are no prescriptions
If there's an error where the prescriptions aren't loading, for example, the toggle for card/list view still appears, but it should probably be hidden.

bshyong:
Resolves #4533

### DIFF
--- a/src/js/rx/containers/Active.jsx
+++ b/src/js/rx/containers/Active.jsx
@@ -1,7 +1,6 @@
 import React from 'react';
 import { connect } from 'react-redux';
 import classnames from 'classnames';
-import isMobile from 'ismobilejs';
 import _ from 'lodash';
 
 import {
@@ -70,7 +69,7 @@ class Active extends React.Component {
   }
 
   renderViewSwitch() {
-    if (!this.props.prescriptions || isMobile.any) {
+    if (!this.props.prescriptions) {
       return null;
     }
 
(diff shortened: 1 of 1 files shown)


U-DON (approved):
LGTM. Although, most of Rx's and SM's "mobile" views are based on the screen/window size, so it could be confusing to mix the conditions of `isMobile`/screen size if the user is just shrinking the window. It should probably all be based on the same condition for consistency.

bshyong:
Hm sure, I can change this up. Now that I think of it, maybe adding a CSS class to it is the best way to go about it. We have classes on the site to hide elements on small screens. I'll make changes and then push again.

bshyong:
removed use of `isMobile`

src/js/rx/containers/Active.jsx
```diff
@@ -70,13 +69,17 @@ class Active extends React.Component {
...
+    }
+
     const toggles = [
       { key: 'card', value: 'Card' },
       { key: 'list', value: 'List' },
     ];
 
     return (
-      <div className="rx-view-toggle" ref={(elem) => { this.viewToggle = elem; }}>View:&nbsp;
+      <div className="rx-view-toggle show-for-small-only" ref={(elem) => { this.viewToggle = elem; }}>View:&nbsp;
```
U-DON: Doesn't this show up if it's in a mobile-sized window? Unless I'm confusing myself, I thought this toggle is supposed to appear only on desktop, since we wanted to always show the cards in mobile.

bshyong: you're right—that was the opposite of what was intended! Fixed now

U-DON (approved):
LGTM.

Issue #3965, opened 2016-11-15
Search form input needs a label


webinista:
Need to change:

 `<input autocomplete="off" type="search" name="query">`

to:

 `<input autocomplete="off" type="search" name="query" id="search-field-small">`

jcmeloni:
Addressed in PR https://github.com/department-of-veterans-affairs/vets-website/pull/4646  will close this issue after merge and not move this issue to vets.gov-team.

jcmeloni:
Merged PR, will close after release.

jcmeloni:
Released 17 Jan 2017

Issue #3780, opened 2016-11-07
[HOMEPAGE] heavier font weight in top banner


melwoodard:
The text in the top banner & crisis line dropdown is a little light and hard to read at font-weight 300. Recommend bumping up to 400. @angel thoughts?

<img width="1279" alt="vets_gov" src="https://cloud.githubusercontent.com/assets/4106356/20068784/d383071e-a4ce-11e6-80b7-5ff3ee631838.png">

dannychapman:
+1 on this.

jcmeloni:
Addressed in PR https://github.com/department-of-veterans-affairs/vets-website/pull/4646  will close this issue after merge and not move this issue to vets.gov-team.

jcmeloni:
Merged PR, will close after release.

jcmeloni:
Released 17 Jan 2017

Issue #2719, opened 2016-08-12
Header incorrect sets og:type to article


awong-dev:
```
{% if title %}
<meta content="article" property="og:type">
{% else %}
<meta content="website" property="og:type">
{% endif %}
```

in `content/includes/header.html` seems wrong. I think it should always be website or at least not keyed off the presence of title which is true always.

awong-dev:
Passing to Danny for triage. (tick-tocking danny and courtney now)

carlmjohnson:
http://ogp.me/#types has good info about the meaning of the tags. An article, unlike a website, has a `published_time` and `author`s, but those don't apply to vets.gov, so I think it's a website.

jcmeloni:
Addressed in PR https://github.com/department-of-veterans-affairs/vets-website/pull/4646  will close this issue after merge and not move this issue to vets.gov-team.

jcmeloni:
Merged PR, will close after release.

jcmeloni:
Released 17 Jan 2017

Issue #2419, opened 2016-05-27
Clean up **<br> to remove the <br>


webinista:
Removing most of the markup from our Markdown means that there's a lot of extraneous HTML that's causing minor layout issues. In particular, this `**Example 2**<br>` syntax is adding extra line breaks.
#### Affected pages
- [ ] _disability-benefits/apply-for-benefits/one-year.md
- [ ] _disability-benefits/claims-process/claim-types/reopened-claim.md 
- [ ] _disability-benefits/claims-process/claim-types/secondary-claim.md 
- [ ] _disability-benefits/claims-process/date.md 
- [ ] _disability-benefits/claims-process/evidence.md
- [ ] _disability-benefits/conditions/exposure-to-hazardous-materials/agent-orange/service-outside.md
- [ ] _disability-benefits/conditions/exposure-to-hazardous-materials/agent-orange/water-vietnam.md
- [ ] _disability-benefits/conditions/special-claims/convalescence.md 
- [ ] _disability-benefits/conditions/special-claims/hospitalization.md 
- [ ] _disability-benefits/conditions/special-claims/individual-unemployability.md 
- [ ] _disability-benefits/conditions/special-claims/prestabilization.md 
- [ ] _education/gi-bill/survivors-dependent-assistance/fry-scholarship.md
- [ ] _education/other-educational-assistance-programs/veap.md
- [ ] _employment/job-seekers/service-disabled.md

jcmeloni:
Addressed in PR https://github.com/department-of-veterans-affairs/vets-website/pull/4646  will close this issue after merge and not move this issue to vets.gov-team.

jcmeloni:
Merged PR, will close after release.

jcmeloni:
Released 17 Jan 2017

PR #4646, opened 2017-01-10
Minor fixes to (hopefully) close 5 issues

jcmeloni-zz:
Issues addressed in this PR (one commit per issue):
- #2411 
- #2419 
- #2719 
- #3780 
- #3965 

Tagging @webinista and @crwallace for review.

### DIFF
--- a/content/includes/header.html
+++ b/content/includes/header.html
@@ -20,11 +20,7 @@
 <meta content="https://www.vets.gov" property="og:title">
 {% endif %}
 
-{% if title %}
-<meta content="article" property="og:type">
-{% else %}
 <meta content="website" property="og:type">
-{% endif %}
 
 {% if path %}
 <meta content="https://www.vets.gov/{{ path }}" property="og:url">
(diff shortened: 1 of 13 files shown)


content/includes/header.html
```diff
@@ -20,11 +20,7 @@
 <meta content="https://www.vets.gov" property="og:title">
 {% endif %}
 
-{% if title %}
```
webinista: I think this needs to stay for SEO purposes. @crwallace can weigh in more. This is a conditional for when there's a `title` property in the front matter. Not sure how much of a difference it makes, but that's why it was added.

jcmeloni: Ok - whatever you think is right! I was reading the issue as "remove the conditional because it should always be x in all cases".

webinista: Looked at the issue ... I don't have a hard opinion. Seems ok to remove.

webinista:
LGTM.

PR #4651, opened 2017-01-10
replace some jquery calls with fetch

aub:
References department-of-veterans-affairs/vets.gov-team#648

Replaces a couple of AJAX calls that used jQuery to use fetch instead.


jkassemi:
Jenkins failures due to maintenance issues at the time. Updated with changes from master to pull in latest Jenkinsfile updates and trigger a new build. Part of investigation in https://github.com/department-of-veterans-affairs/devops/issues/1352

aub:
@jbalboni mind giving this a review so we can get it merged?

src/js/login/containers/Main.jsx
```diff
@@ -49,15 +52,15 @@ class Main extends React.Component {
...
+    fetch(`${environment.API_URL}/v0/sessions`, {
+      method: 'DELETE',
+      headers: new Headers({
         Authorization: `Token token=${sessionStorage.userToken}`
-      },
-      success: (result) => {
-        this.setState({ logoutUrl: result.logout_via_get });
-      }
+      })
+    }).then(response => {
```
jbalboni: Should we handle errors in these calls? I think if there's an http error now we're going to generate some JS errors, instead of just failing silently.

jbalboni (approved):
Looks alright to me, but Courtney is a better person to review these changes. I do have a question about how we should handle errors.

aub:
@crwallace could you give this a review as well?

@jbalboni On the error handling, maybe we make that a separate issue? I'd like to keep this issue as just a one-to-one port, which I *think* it is right now.

jbalboni:
@aub I don't think it's a one to one port, but I haven't tried it to be sure. I think that if you do `response.json()` when the response is a 404, for example, you'll get a JS error. Whereas in the jQuery version, you'll just get the 404 network error and the success handler will be skipped.

jcmeloni-usds:
@aub @crwallace Any reason not to merge at this point? It would clear the blocker to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/649

aub:
@jcmeloni I think we need to address @jbalboni's concern above before merging. It seems like some feedback from @crwallace would take care of it.

jcmeloni-usds:
Excellent.  @crwallace please feedback :)

src/js/login/containers/Main.jsx
crwallace: So this is a larger UX question that I asked when we were first building login. We ended up kicking the can on this in an effort to just get an MVP up on vets.gov. 

I think we need to bring @AnitraAppa into this convo as she has given much of our direction on error handling from a UX perspective which has allowed me to implement it from a technical perspective.

aub:
@crwallace my understanding from @jbalboni was that we have a polyfill for fetch so it works cross-browser.

jcmeloni-usds:
I have IE 11 natively (not emulated) on one or more machines, and if something gets to dev or staging that I can test to verify, I'm super happy to do it.

aub:
@crwallace given that we have a polyfill for fetch do you think we're good to go here?

crwallace:
Yeah we should be good to go

crwallace (approved):

PR #4655, opened 2017-01-11
[SM] Show message before 'Edit draft'/'Reply' button on mobile.

U-DON:
### Changes
* Display the preview of the draft like a regular message in the thread on mobile. Closes department-of-veterans-affairs/messaging-fe#207.
* Fixed broken mobile header bar when only the 'Back' button is showing.
* Hide redundant or missing fields in expanded message details. No sent date when not available and no 'From' or 'To' on mobile.
* Show message title above the message for new message drafts only on mobile.
* Added unit tests for Message and MessageDetails components.

### DIFF
--- a/src/js/messaging/components/Message.jsx
+++ b/src/js/messaging/components/Message.jsx
@@ -12,13 +12,15 @@ class Message extends React.Component {
   }
 
   componentDidUpdate() {
+    const { attrs, isCollapsed } = this.props;
+
     const shouldFetchMessage =
-      !this.props.isCollapsed &&
-      this.props.attrs.attachment &&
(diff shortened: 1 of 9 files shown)


src/js/messaging/components/Message.jsx
```diff
@@ -12,13 +12,15 @@ class Message extends React.Component {
   }
 
   componentDidUpdate() {
+    const { attrs, isCollapsed } = this.props;
```
bshyong: 👍

bshyong (approved):
LGTM!

Also completes some of the tests in https://github.com/department-of-veterans-affairs/messaging-fe/issues/103

PR #4658, opened 2017-01-11
Add tabs for claims list

austin-martinez:
Related to https://github.com/department-of-veterans-affairs/sunsets-team/issues/293

- Adds a feature flag for all claims
- Adds tabs for the claim list


src/js/disability-benefits/actions/index.js
```diff
@@ -46,6 +46,23 @@ export function getClaims() {
...
+  return (dispatch) => {
+    makeAuthRequest('/v0/disability_claims',
+      null,
+      dispatch,
+      claims => {
+        const filteredClaims = claims.data.filter(function(claim) {
+          return claim.attributes.open === open;
+        });
+        dispatch({ type: SET_CLAIMS, claims: filteredClaims, meta: claims.meta });
+        dispatch({ type: CHANGE_CLAIMS_PAGE, page: 1 });
```
austin-martinez: This resets the page when switching tabs, which is intended behavior. However, it also resets the page when going back to the list from an individual claim. Is there a better way to achieve the reset on tab switch but keep current page accurate when going back?

src/js/disability-benefits/containers/ClosedClaimsPage.jsx
```diff
@@ -0,0 +1,84 @@
...
+import { scrollToTop, setUpPage, setPageFocus } from '../utils/page';
+
+class ClosedClaimsPage extends React.Component {
+  constructor(props) {
+    super(props);
+    this.changePage = this.changePage.bind(this);
+  }
+  componentDidMount() {
+    if(this.props.allClaims) {
+      this.props.getFilteredClaims(false);
```
austin-martinez: This is the only difference from the `OpenClaimsPage` component, but not sure best way to reuse the component and change this option

jbalboni:
@austin-martinez I think what I would do is use the same component for both routes, but add an attribute to the closed route:

```
<Route
         component={ClaimsPage}
         key="/your-claims/closed"
         showClosed
         path="/your-claims/closed"/>
```

Then something like this in the mount method:

```
componentDidMount() {
     const open = !this.props.route.showClosed;
     if (this.props.allClaims) {
       this.props.getFilteredClaims(open);
     } else {
       this.props.getClaims();
     }
     if (this.props.loading) {
       scrollToTop();
     } else {
       setUpPage();
     }
   }
```

For your second question, if you use the same component for both routes, I don't think React will remount them on route changes between them. So you can use `componentWillReceiveProps`:

```
  if (newProps.route.showClosed !== this.props.route.showClosed) {
    this.props.changeClaimsPage(1);
  }
}
```

You'd have to add the `changeClaimsPage` action creator, but I think that would work. That method will run anytime the props for the component change, but not if the component is being mounted.

src/js/disability-benefits/containers/ClaimsList.jsx
```diff
@@ -0,0 +1,94 @@
+import React from 'react';
```
jbalboni: Why did you separate out this component from YourClaimsPage?

src/js/disability-benefits/actions/index.js
```diff
@@ -46,6 +46,22 @@ export function getClaims() {
   };
 }
 
+export function getFilteredClaims(open = true) {
+  return (dispatch) => {
+    makeAuthRequest('/v0/disability_claims',
```
jbalboni: Do we need to get new data from the server whenever they change tabs?

If we do, we should probably show a spinner underneath the tabs when while we're waiting.

src/js/disability-benefits/routes.jsx
```diff
@@ -15,7 +16,14 @@ const routes = [
...
       component={YourClaimsPage}
       key="/your-claims"
-      path="/your-claims"/>,
+      path="/your-claims">
+    <IndexRoute component={ClaimsList}/>
+    <Route
+        component={ClaimsList}
+        showClosedClaims
+        key="/your-claims/closed"
+        path="/your-claims/closed"/>
```
jbalboni: If you're going to nest like this, you can just do `path="closed"`.

jbalboni (commented):
Looks good, have a couple questions, though.

src/js/disability-benefits/actions/index.js
austin-martinez: Data is only fetched from the server once, the filtering is done client side in this action before being passed to the reducer

jbalboni: It looks like you're calling this function, which always hits the server, whenever the route changes (through `componentWillReceiveProps`).

src/js/disability-benefits/components/ClaimDetailLayout.jsx
```diff
@@ -15,13 +15,14 @@ export default class ClaimDetailLayout extends React.Component {
 
     let content;
     if (!loading) {
+      const claimsLink = `your-claims${claim.attributes.open ? '' : '/closed'}`;
```
austin-martinez: The following scenario was causing unexpected behavior
1. User navigates to page 2 of closed claims
2. User clicks on a claim
3. User clicks link to 'Your Claims'
4. User is on page 2 of open claims

I think preserving the pagination on back navigation is important UX, which requires maintaining the pagination state between mounts of YourClaimsPage, so this change mimics going back from a claim

src/js/disability-benefits/reducers/claims-list.js
```diff
@@ -22,19 +22,30 @@ function phaseChangeDate(claim) {
 export default function claimsReducer(state = initialState, action) {
   switch (action.type) {
     case SET_CLAIMS: {
-      const sortedList = _.orderBy([phaseChangeDate, 'id'], 'desc', action.claims);
+      return _.assign(state, {
+        list: action.claims
+      });
+    }
+    case FILTER_CLAIMS: {
+      let sortedList = _.orderBy([phaseChangeDate, 'id'], 'desc', state.list);
```
austin-martinez: In the next feature, custom sort options will be added, so it made sense to separate sorting from setting the claims list, I kept it with the filtering for now but may separate it in next PR

src/js/disability-benefits/components/ClaimDetailLayout.jsx
jbalboni: There are other places where you'd want to change this, too, like on the file upload pages. Any claim specific page with breadcrumbs.

src/js/disability-benefits/containers/YourClaimsPage.jsx
```diff
@@ -51,6 +65,14 @@ class YourClaimsPage extends React.Component {
       content = <NoClaims/>;
     }
 
+    if (this.props.allClaims) {
+      content = (
+        <div className="va-tab-content db-tab-content" role="tabpanel" id="tabPanelOpenClaims" aria-labelledby="tabOpenClaims">
```
jbalboni: I think the id/labelledby attributes need to use the open/closed state.

jbalboni (approved):
Looks good to me. I made a couple comments that I think will require changes.

src/js/disability-benefits/components/ClaimDetailLayout.jsx
austin-martinez: There were a few breadcrumbs I didn't update as they shouldn't be reachable for a closed claim and didn't contain the full claim object

Issue #4608, opened 2017-01-05
Remaining form scaffolding work for 1995


jbalboni:
This is the remaining support work for the 1995 form, i.e. the work that needs to be done to support all the fields and our normal form UI flow.

I have a gist with prototype work for some of this [here](https://gist.github.com/jbalboni/52a9183490e1608f5881d7f72b03370c).

- [x] Object field
 - We need to create a flexible template for blocks in the form so that we can add descriptions when necessary (and fix some spacing issues)
- [x] Lazy load form and exclude routes for production
 - We don't want to add a lot of bloat to the 1990 form, so before we allow this to be usable on staging, we need to convert the edu benefits app to lazy loadable routes.
- [x] Date field
 - This is mostly done in the gist, but needs some style work and proper validation
- [x] Growable Table
 - This is maybe half done in the gist linked below. Mostly styling work left.
- [x] Redux set up
 - We need to store the form data and validation state in Redux so we can check them across all pages at the end before submitting
- [x] Review page
 - This is pretty big. We should be able to generate this page from the schema, but it'll take some work.
- [x] Add all 1995 fields/pages in config
- [x] Investigate performance issue
- [x] Confirmation page (in progress - Jeff)
  - This one is to build the confirmation page that we show after submitting a form. I think it’s the same as the 1990 one, but not sure yet. The important thing for that one is to put in place the common code so that after a successful submit it shows the confirmation page that’s set in the form config.
- [ ] Address field (in progress - Anne)
  - This needs to be a custom field to support the switching of the state/zip fields
- [ ] Conditional pages (in progress - Eugene)
  - This one is to add the functionality to skip a page in the form depending on form data. There’s a page in the form that should only be shown if a user has answered a particular question. What we did on the 1990 form was have a `depends` property on each page’s route that was either a function or an object that we could use with Lodash’s match function. It’d be a similar thing here, except in the form config object.
- [ ] Expandable fields (in progress - Jeff)
  - We have some fields that are only shown when you answer yes to a question above them (there’s an ExpandableFields component for this now on the 1990 form). We need to port that over somehow to this way of building forms.
- [ ] Conditional drop down options
  - This is something we need to do at some point, but may not be necessary for 1995 specifically. In some places we change the dropdown options for a field when form data changes. This would work similarly to the conditional validation
- [ ] Move more stuff to Redux state
[...]

PR #4659, opened 2017-01-11
Add review page for 1995 form

jbalboni:
Related to #4608 

This adds the review page for the new form work. It will need some tweaking as we go (especially with growable tables), but this should work for most of the form. It also sets up Redux, since that was necessary to get the form data onto the review page.

The way the review page works is that it still uses the library to render each form page, but when in review mode it replaces the fields with read-only versions. You can see these in the schemaform/review folder.

I had to change the markup for the read view of each page to be a definition list instead of a table. You can't nest `tbody` elements so it doesn't mesh will with the way the form library recursively generates forms. Caitlin had mentioned wanting to do this in other issue anyway.

Screenshots were taken with random chapters/pages I added in the form config, then removed.

Submit process works (but we don't have an endpoint yet)
screen shot 2017-01-10 at 4 16 18 pm

Collapsed review page
screen shot 2017-01-10 at 3 59 14 pm

Privacy agreement set up done
screen shot 2017-01-10 at 3 59 04 pm

Review mode of expanded chapter
screen shot 2017-01-10 at 3 58 26 pm

Expanded chapter in edit mode
screen shot 2017-01-10 at 3 58 53 pm


src/js/common/schemaform/actions.js
```diff
@@ -0,0 +1,95 @@
...
+}
+
+export function setSubmitted(response) {
+  return {
+    type: SET_SUBMITTED,
+    response
+  };
+}
+
+export function submitForm(formConfig, form) {
```
jbalboni: This whole function should be generic, but there's some edu specific stuff in here for now.

src/js/common/schemaform/review/ReviewCollapsibleChapter.jsx
```diff
@@ -0,0 +1,146 @@
...
+            <div className="accordion-header clearfix" >
+              <button
+                  className="usa-button-unstyled"
+                  aria-expanded={this.state.open ? 'true' : 'false'}
+                  aria-controls={`collapsible-${this.id}`}
+                  onClick={this.toggleChapter}>
+                {this.props.chapter.title}
+              </button>
+            </div>
+            {this.state.open && pageContent}
```
akainic: Do you need to have this check for `this.state.open` here if you have it above before setting the value of `pageContent`?

akainic (commented):
Wow, so much great work! Just a few small comments/questions.

jbalboni:
@akainic All fixed now!

src/js/common/components/form-elements/ErrorableCheckbox.jsx
```diff
@@ -90,7 +90,7 @@ ErrorableCheckbox.propTypes = {
   checked: React.PropTypes.bool,
   errorMessage: React.PropTypes.string,
   name: React.PropTypes.string,
-  label: React.PropTypes.string.isRequired,
+  label: React.PropTypes.object.isRequired,
```
U-DON: Is this an intentional change?

jbalboni: Yes, though only tangentially related to this PR. We sometimes pass rendered components as checkbox labels (so we can use links) and hadn't ever changed this prop type.

U-DON: Then wouldn't labels that are still strings trigger a warning about the type? This should probably be:

```
label: React.PropTypes.oneOfType([
  React.PropTypes.string,
  React.PropTypes.object
]).isRequired,
```

src/js/common/schemaform/FormPage.jsx
```diff
@@ -118,31 +136,50 @@ class FormPage extends React.Component {
...
       </Form>
     );
   }
 }
 
+function mapStateToProps(state, ownProps) {
+  return {
+    form: state.form,
+    route: ownProps.route,
+    reviewMode: ownProps.reviewMode
```
U-DON: I'm confused why these have to be extracted from `ownProps`; if they were already props being passed in, is this necessary?

src/js/common/schemaform/helpers.js
```diff
@@ -107,13 +124,21 @@ export function formatISOPartialDate({ month, day, year }) {
...
+  return undefined;
+}
 export function parseISODate(dateString) {
   if (dateString) {
     const [year, month, day] = dateString.split('-', 3);
 
     return {
       month: month === 'XX' ? '' : Number(month),
-      day: day === 'XX' ? '' : Number(day),
+      day: day === 'XX' ? '' : Number(day).toString(),
```
U-DON: Why isn't it also `Number(month).toString()` above this?

src/js/common/schemaform/review/ObjectField.jsx
```diff
@@ -0,0 +1,108 @@
...
+
+class ObjectField extends React.Component {
+  static defaultProps = {
+    uiSchema: {},
+    errorSchema: {},
+    idSchema: {},
+    registry: getDefaultRegistry(),
+    required: false,
+    disabled: false,
+    readonly: false,
```
U-DON: It seems some of these props aren't being used at all, like `required`, `disabled`, `readonly`.

src/js/common/components/form-elements/ErrorableCheckbox.jsx
jbalboni: Yes, it should be `oneOf`.

src/js/common/schemaform/FormPage.jsx
jbalboni: I was under the impression that react-redux wouldn't pass through props you passed to the connect wrapped component to the inner component. I'm not sure I have tried that directly, so I could be wrong.

src/js/common/schemaform/helpers.js
jbalboni: The way our current date component works (which I copied), it requires the month to be a number and the day to be a string.

src/js/common/schemaform/review/ObjectField.jsx
jbalboni: Yeah, I need to pass those to `SchemaField` or remove them. This component is a copied and modified version of one in the `react-jsonschema-form` library, so we don't necessarily use all of the props that could be passed to it.

jbalboni: Actually, those are all being used now.

src/js/common/schemaform/FormPage.jsx
jbalboni: I am wrong! I'll remove those.

test/common/schemaform/review/ObjectField.unit.spec.jsx
```diff
@@ -0,0 +1,84 @@
...
+          schema={schema}
+          formContext={{}}
+          idSchema={{ $id: 'root' }}
+          formData={{}}
+          onChange={onChange}
+          onBlur={onBlur}/>
+    );
+
+    expect(tree.everySubTree('.form-review-panel-page-header-row')).not.to.be.empty;
+    expect(tree.subTree('.form-review-panel-page-header').text()).to.be.equal('Blah');
```
U-DON: Nitpick: `to.equal`?

src/js/common/schemaform/review/ReviewCollapsibleChapter.jsx
```diff
@@ -0,0 +1,146 @@
...
+        </div>
+      );
+    }
+
+    return (
+      <div id={`${this.id}-collapsiblePanel`} className="usa-accordion-bordered form-review-panel">
+        <Element name={`chapter${this.props.chapterKey}ScrollElement`}/>
+        <ul className="usa-unstyled-list">
+          <li>
+            <div className="accordion-header clearfix" >
```
U-DON: Nitpick: extra space.

test/common/schemaform/widgets/DateWidget.unit.spec.jsx
```diff
@@ -63,9 +63,11 @@ describe('Schemaform: DateWidget', () => {
     const instance = tree.getMountedInstance();
     instance.handleChange('year', 2001);
     instance.handleChange('month', '1');
-    instance.handleChange('day', '2');
```
U-DON: I thought you mentioned for this component, month is expected to be a number and day to be a string? This seems at odds with the above change in this set of tests too.

jbalboni: Went back and looked at this to refresh my memory. I'm not sure what I was running into before, but it seems to work fine if they're all strings. I'm converting to number first to remove leading zeroes, but the end result seems to be fine in either type.

In this test I'm not actually using these values, so it doesn't break anything, but I'll fix it in both places.

U-DON (approved):
LGTM.

PR #4662, opened 2017-01-11
Add lazy loading for 1995 form

jbalboni:
Related to #4608 

This converts the 1990 routes to the react router object format and loads the 1995 code lazily if the build type is `development`.


jbalboni:
FYI @crwallace: we're lazy loading some parts of the edu benefits app now.

src/js/common/schemaform/helpers.js
```diff
@@ -18,6 +16,19 @@ function createFormPageList(formConfig) {
     }, []);
 }
 
+export function createPageListByChapter(formConfig) {
```
U-DON: Was confused that this is used in another PR, but not in this one, even though it was added here.

U-DON (approved):
LGTM.

src/js/common/schemaform/helpers.js
jbalboni: Oh, oops. I'll remove it from here.

PR #4664, opened 2017-01-11
Adds Rx store to CommonStore

crwallace:


src/js/common/reducers/index.js
```diff
@@ -1,10 +1,25 @@
...
 
-import login from '../../common/reducers/login';
-import profile from '../../common/reducers/profile';
+import login from '../../login/reducers/login';
+import profile from '../../user-profile/reducers/profile';
+
+import alertRx from '../../rx/reducers/alert';
+import disclaimer from '../../rx/reducers/disclaimer';
+import modalsRx from '../../rx/reducers/modals';
+import prescriptions from '../../rx/reducers/prescriptions';
```
U-DON: Same comment as in the original PR:

Could we just import the combined reducer from each app? Rx and SM already have them bundled in their `reducers/index.js`. So something like:

```
import rxReducer from '../../rx/reducers';
import msgReducer from '../../messaging/reducers';
```

It would reduce the number of imports in this module, especially the multiple `alert*` and `modal*`. And any changes to a particular reducer would be localized to the app.

U-DON (approved):
LGTM.

PR #4670, opened 2017-01-11
Simplified MessageDetails toggle logic.

U-DON:
Refactored MessageDetails so that instead of using hidden inputs, it uses state and CSS classes to show and hide message details. Also added the corresponding unit test.

### DIFF
--- a/src/js/messaging/components/MessageDetails.jsx
+++ b/src/js/messaging/components/MessageDetails.jsx
@@ -13,18 +13,20 @@ class MessageDetails extends React.Component {
   }
 
   toggleCompactDetails() {
-    this.setState((prevState) => {
-      const newState = { expanded: { ...prevState.expanded } };
-      newState.expanded.compact = !prevState.expanded.compact;
-      return newState;
+    this.setState({
+      expanded: {
+        ...this.state.expanded,
+        compact: !this.state.expanded.compact
+      }
     });
   }
 
   toggleFullDetails() {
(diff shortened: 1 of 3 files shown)


src/js/messaging/components/MessageDetails.jsx
```diff
@@ -76,35 +76,34 @@ class MessageDetails extends React.Component {
...
     );
 
     return (
       <div
-          ref="detailsControl"
           className="messaging-message-details-control"
           tabIndex="-1"
-          onBlur={this.hideDetails}
+          onBlur={() => {
+            if (this.state.expandedFull) {
```
bshyong: if the purpose of this is to always hide the details on blur, would it be worth doing something like `this. toggleFullDetails(false)`?, i.e. we can override the toggle action by passing a param (true or false).

bshyong (approved):
LGTM w/ a minor comment

src/js/messaging/components/MessageDetails.jsx
U-DON: Hm, in this case, maybe it's better to just directly set the state than calling a method named `toggle*` if it's not really toggling...

bshyong: Yup I think I've run into this situation before: it's an action that is used as a toggle in one case, but also reused to solely open/close in another case..

Really minor issue thought, this PR LGTM as is

PR #4676, opened 2017-01-12
[SM] Show 'Delete' button when editing messages on mobile.

U-DON:
Closes department-of-veterans-affairs/messaging-fe#192.

### DIFF
--- a/src/sass/messaging/partials/_message-compose.scss
+++ b/src/sass/messaging/partials/_message-compose.scss
@@ -103,6 +103,8 @@
   position: relative;
 
   @media (max-width: $medium-screen) {
+    -webkit-align-items: center;
+    align-items: center;
     border-width: 0;
     display: -webkit-flex;
     display: flex;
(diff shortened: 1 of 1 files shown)


src/sass/messaging/partials/_message-compose.scss
```diff
@@ -288,6 +286,17 @@ button.msg-btn-save {
     top: 0;
     width: auto;
   }
+
+  @media (max-width: $medium-screen) {
+    -webkit-flex: 0 0;
```
bshyong: I think this might be more of a question for core, but I saw that `autoprefixer` and `postcss` were both in the package.json, though I didn't see it configured in our webpack config.

If they are set up, they should automatically apply vendor prefixes for us so that we don't miss any.

https://github.com/postcss/postcss-loader

bshyong (approved):
LGTM

src/sass/messaging/partials/_message-compose.scss
U-DON: Yeah, this has come up before, but I don't remember what the actual resolution was. Just added it for consistency, but would like to have auto-prefixing and remove the prefixes that were added manually.

PR #4679, opened 2017-01-12
Adding Other as an option for birth state

knkski:
Fixes https://github.com/department-of-veterans-affairs/vets.gov-team/issues/727

Also removing old schema that is no longer in use that was found while working on this

cc @jbalboni @akainic for ^^


src/js/common/utils/options-for-select.js
```diff
@@ -359,6 +359,7 @@ const states = {
     { label: 'Ohio', value: 'OH' },
     { label: 'Oklahoma', value: 'OK' },
     { label: 'Oregon', value: 'OR' },
+    { label: 'Other', value: 'Other' },
```
jbalboni: I think you'll want to do something similar to the schema change instead of this, since this will mean Other shows up in the state list for addresses and on edu.

You could create a birth place states list and use that in the BirthInformation component (I think you'll have to concat Other to this states list and then sort it).

jkassemi:
Merged master into this to trigger new Jenkins build with associated fixes. Part of investigation in https://github.com/department-of-veterans-affairs/devops/issues/1352

src/js/common/utils/options-for-select.js
markolson: @jbalboni @knkski if either of you get a chance today, can you tackle this? I'd like to get this into master after the production branch goes out so we can test on staging.

knkski: Sure, I'll take a look at it

src/js/hca/components/veteran-information/BirthInformationSection.jsx
```diff
@@ -59,7 +59,7 @@ class BirthInformationSection extends React.Component {
               onValueChange={(update) => {this.props.onStateChange('cityOfBirth', update);}}/>
           <ErrorableSelect label="State"
               name="stateOfBirth"
-              options={states.USA}
+              options={states.USA.concat([{label: 'Other', value: 'Other'}]).sort((a, b) => a.label.localeCompare(b.label))}
```
jbalboni: Can you put the output of this concat and sort in options-for-select.js instead? In here, we're going to end up running it a bunch of times, every time the component re-renders. Probably not noticeable on a desktop browser, but I've seen select components with lots of options cause some sluggishness on phones in one of our other forms.

markolson (approved):

markolson:
@knkski LGTM after the linter fixes

jbalboni (approved):
LGTM

markolson:
Backend PR for healthcare-application is https://github.com/department-of-veterans-affairs/healthcare-application/pull/334. Once that's deployed, I'd like to test this frontend change on staging a bit before we go to production with it.

PR #4684, opened 2017-01-12
[Rx] Fixed alignment of sort menu elements for mobile. Tweaked spacing for desktop.

U-DON:
### Changes
* Tweaked spacing of sort menu elements on desktop.
* Fixed misalignment of sort menu elements on mobile.
    * Closes department-of-veterans-affairs/vets.gov-team#636.

### DIFF
--- a/src/js/rx/components/SortMenu.jsx
+++ b/src/js/rx/components/SortMenu.jsx
@@ -35,8 +35,8 @@ class SortMenu extends React.Component {
       'lastSubmitDate',
       'refillDate',
     ].includes(value) ? {
-      'Newest to Oldest': 'ASC',
-      'Oldest to Newest': 'DESC'
+      'Newest to Oldest': 'DESC',
+      'Oldest to Newest': 'ASC'
     } : {
       'A-Z': 'ASC',
       'Z-A': 'DESC'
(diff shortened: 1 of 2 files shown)


U-DON:
Also fixed the order for the sort menu when sorting by date. "Newest to Oldest" should now go from more recent dates to earlier dates and "Oldest to Newest" the reverse.

U-DON:
Found a few more issues around the sort menu but will do a separate follow-up PR to address them.
* Now that date order in the sort menu has been corrected, selecting date-based sort params will default to sorting by 'Oldest to Newest', since ascending order is our default for all selections. I'll make it so that date-based sorts will default to 'Newest to Oldest' instead.
* Somewhat related, while testing the sort menu on the History page, I noticed that the prescription name is labeled with 'Fill Date', which is an artifact of the small-screen styles for the table on the Active page. Those styles weren't high enough specificity, so it applied to both tables, but either way, they aren't being used anymore for the Active page, so I'll be removing them.
* Going to add unit tests for SortMenu as well.

src/js/rx/components/SortMenu.jsx
```diff
@@ -6,6 +6,14 @@ import React from 'react';
 import _ from 'lodash';
 import classNames from 'classnames';
 
+const isDateAttribute = (option) => {
```
bshyong: 👍

bshyong (approved):
LGTM

PR #4685, opened 2017-01-12
Change feature flag constant to rely on buildtype option

jkassemi:
Instead of the environment variable BUILDTYPE, which isn't defined
explicitly when defining the build's type for both local development or
fro the CI/CD pipeline.


austin-martinez (approved):
LGTM 👌

PR #4686, opened 2017-01-13
Fix ALL_CLAIMS_ENABLED boolean flag

jkassemi:
This didn't require the JSON.stringify, since we're not looking to convert to a JS value for webpack define replacement.


jkassemi:
Confirmed that allClaims is `true` in the build output after this update.

austin-martinez (approved):
👍

PR #4689, opened 2017-01-13
moves claims store to commonStore

crwallace:
This moves the store for disability claims to the common store.

This relates to department-of-veterans-affairs/vets.gov-team#581

### DIFF
--- a/src/js/common/reducers/index.js
+++ b/src/js/common/reducers/index.js
@@ -5,6 +5,8 @@ import profile from '../../user-profile/reducers/profile';
 
 import rx from '../../rx/reducers';
 
+import claims from '../../disability-benefits/reducers';
+
 export default combineReducers({
   user: combineReducers({
     login,
@@ -12,6 +14,9 @@ export default combineReducers({
   }),
   health: combineReducers({
     rx
+  }),
+  disability: combineReducers({
+    claims
   }),
 });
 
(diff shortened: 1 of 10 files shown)


molindoz (commented):
LGTM... I'll check with @jbalboni and @austin-martinez   to see if they have more thoughts on this.

src/js/common/reducers/index.js
```diff
@@ -13,5 +15,8 @@ export default combineReducers({
   health: combineReducers({
     rx
   }),
+  disability: combineReducers({
```
jbalboni: It seems like this is going to mean our app has all the Redux reducer code for all the other apps in it. That seems bad from a file size perspective. And also possibly cause action name collisions, because we're not trying to make those unique across apps.

Would it be better to have something like a `createCommonReducer` function that took the app specific reducer as a parameter and returned a reducer with the login code added in?

jbalboni (commented):
👍  to having a single store, but I have a concern about how it's structured currently.

src/js/common/reducers/index.js
crwallace: @jbalboni -- I think thats a valid point. This is a first phase of trying to get everything into the commonStore and decrease the amount of calls that we are making to the api and then optimize. I think we are moving to a world where are are going to try to have notifications and that will require data from many apps on the same page. I think what we are trying to do is implement and then optimize.

PR #4691, opened 2017-01-13
login status should remain false if session has expired

kreek:
Fixes https://github.com/department-of-veterans-affairs/vets.gov-team/issues/623

`this.props.onUpdateLoggedInStatus(true)` was being called even thought the session had expired. login-helpers.js also tried to update profile fields when the server returned unauthorized.


jcmeloni:
LGTM

PR #4692, opened 2017-01-13
[Rx] Sorting fixes and unit tests.

U-DON:
### Changes
* Sort query parameters on Active page are applied when directly accessing URL with them. Closes department-of-veterans-affairs/vets.gov-team#743.
* Removed incorrect 'Fill Date' label next to prescription name in mobile version of History table. Closes department-of-veterans-affairs/vets.gov-team#747.
* Removed `href` from sortable table header links (no fragment in URL after selecting a column to sort on). Added `tabIndex` to keep it accessible.
* Added SortMenu unit tests.

### DIFF
--- a/src/js/common/components/SortableTable.jsx
+++ b/src/js/common/components/SortableTable.jsx
@@ -43,7 +43,7 @@ class SortableTable extends React.Component {
 
     return (
       <th key={field.value}>
-        <a href="#" role="button" onClick={this.handleSort(field.value, nextSortOrder)}>
+        <a role="button" tabIndex="0" onClick={this.handleSort(field.value, nextSortOrder)}>
           {field.label}
           {sortIcon}
         </a>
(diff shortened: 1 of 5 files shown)


U-DON:
Apparently, the test environment doesn't support `Array.prototype.includes`. But it turns out IE doesn't support it at all, so might as well use the Lodash version.

src/js/common/components/SortableTable.jsx
```diff
@@ -43,7 +43,7 @@ class SortableTable extends React.Component {
 
     return (
       <th key={field.value}>
-        <a href="#" role="button" onClick={this.handleSort(field.value, nextSortOrder)}>
+        <a role="button" tabIndex="0" onClick={this.handleSort(field.value, nextSortOrder)}>
```
bshyong: 👍

bshyong (approved):
LGTM

PR #4693, opened 2017-01-13
Consolidated single store for secure messaging

bshyong:
Consolidated single store work for secure messaging

for issue https://github.com/department-of-veterans-affairs/vets.gov-team/issues/581

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: node_js
+sudo: required
 
 env:
   global:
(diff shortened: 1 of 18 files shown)


crwallace (approved):
LGTM!

U-DON (commented):
LGTM, but test is tripping on MessagingApp's usage of `state.health.msg` for some reason.

bshyong:
Weird because it should work as is - tests pass locally when I just use `state`... inspecting it shows that it is state.health.msg

PR #4699, opened 2017-01-16
Show claim type in title, breadcrumbs, profile page

earthboundkid:
Refs. https://github.com/department-of-veterans-affairs/sunsets-team/issues/289

I think I've got them all? Looking through with grep, it looks like the remaining reference to disability are either part of content or a variable/file name, but let me know if you see any I missed.

### DIFF
--- a/src/js/disability-benefits/components/ClaimEstimationPage.jsx
+++ b/src/js/disability-benefits/components/ClaimEstimationPage.jsx
@@ -17,7 +17,7 @@ export default class ClaimEstimationPage extends React.Component {
             <nav className="va-nav-breadcrumbs">
               <ul className="row va-nav-breadcrumbs-list" role="menubar" aria-label="Primary">
                 <li><Link to="your-claims">Your claims</Link></li>
-                <li><Link to={`your-claims/${this.props.params.id}/status`}>Your {getClaimType(claim)} Claim</Link></li>
+                <li><Link to={`your-claims/${this.props.params.id}/status`}>Your {getClaimType(this.props)} Claim</Link></li>
                 <li className="active">How We Come Up with Your Estimated Decision Date</li>
               </ul>
             </nav>
--- a/src/js/disability-benefits/components/ClaimEstimationPage.jsx
+++ b/src/js/disability-benefits/components/ClaimEstimationPage.jsx
@@ -2,6 +2,7 @@ import React from 'react';
 import { Link } from 'react-router';
 import AskVAQuestions from './AskVAQuestions';
 import { setUpPage } from '../utils/page';
+import { getClaimType } from '../utils/helpers';
 
 export default class ClaimEstimationPage extends React.Component {
   componentDidMount() {
@@ -16,7 +17,7 @@ export default class ClaimEstimationPage extends React.Component {
             <nav className="va-nav-breadcrumbs">
               <ul className="row va-nav-breadcrumbs-list" role="menubar" aria-label="Primary">
                 <li><Link to="your-claims">Your claims</Link></li>
-                <li><Link to={`your-claims/${this.props.params.id}/status`}>Your Disability Compensation Claim</Link></li>
+                <li><Link to={`your-claims/${this.props.params.id}/status`}>Your {getClaimType(claim)} Claim</Link></li>
                 <li className="active">How We Come Up with Your Estimated Decision Date</li>
               </ul>
             </nav>
--- a/src/js/disability-benefits/containers/AdditionalEvidencePage.jsx
+++ b/src/js/disability-benefits/containers/AdditionalEvidencePage.jsx
@@ -71,8 +71,9 @@ class AdditionalEvidencePage extends React.Component {
     if (this.props.loading) {
       content = <LoadingIndicator setFocus message="Loading claim information"/>;
     } else {
-      const filesPath = `your-claims/${this.props.claim.id}/files`;
-      const claimsPath = `your-claims${this.props.claim.attributes.open ? '' : '/closed'}`;
+      const claim = this.props.claim;
+      const filesPath = `your-claims/${claim.id}/files`;
+      const claimsPath = `your-claims${claim.attributes.open ? '' : '/closed'}`;
       const message = this.props.message;
 
       content = (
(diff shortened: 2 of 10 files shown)


src/js/disability-benefits/components/ClaimEstimationPage.jsx
```diff
@@ -2,6 +2,7 @@ import React from 'react';
 import { Link } from 'react-router';
```
austin-martinez: Should the content on this page (L33-34) be changed?
https://github.com/department-of-veterans-affairs/vets-website/pull/4699/files#diff-fb9d863698fbadc576db2ebce1749658R34

src/js/disability-benefits/containers/AdditionalEvidencePage.jsx
```diff
@@ -81,7 +82,7 @@ class AdditionalEvidencePage extends React.Component {
               <nav className="va-nav-breadcrumbs">
                 <ul className="row va-nav-breadcrumbs-list" role="menubar" aria-label="Primary">
                   <li><Link to={claimsPath}>Your claims</Link></li>
-                  <li><Link to={filesPath}>Your Disability Compensation Claim</Link></li>
+                  <li><Link to={filesPath}>Your {getClaimType(claim)} Claim</Link></li>
```
austin-martinez: This should be `this.pops.claim`

src/js/disability-benefits/components/ClaimEstimationPage.jsx
```diff
@@ -16,7 +17,7 @@ export default class ClaimEstimationPage extends React.Component {
             <nav className="va-nav-breadcrumbs">
               <ul className="row va-nav-breadcrumbs-list" role="menubar" aria-label="Primary">
                 <li><Link to="your-claims">Your claims</Link></li>
-                <li><Link to={`your-claims/${this.props.params.id}/status`}>Your Disability Compensation Claim</Link></li>
+                <li><Link to={`your-claims/${this.props.params.id}/status`}>Your {getClaimType(this.props)} Claim</Link></li>
```
austin-martinez: This component doesn't currently receive the claim as a prop, I think you can get it from the store/state by adding
```
function mapStateToProps(state) {
  return {
    claim: state.disability.status.claimDetail.detail
  }
}

export default connect(mapStateToProps)(ClaimEstimationPage);
```
@jbalboni Can you confirm the above?

src/js/disability-benefits/containers/StatusPage.jsx
```diff
@@ -7,15 +7,15 @@ import AskVAToDecide from '../components/AskVAToDecide';
...
+import { itemsNeedingAttentionFromVet, getClaimType, getCompletedDate } from '../utils/helpers';
 
 import { clearNotification } from '../actions';
 
 const FIRST_GATHERING_EVIDENCE_PHASE = 3;
 
 class StatusPage extends React.Component {
   componentDidMount() {
-    document.title = 'Status - Your Disability Compensation Claim';
+    document.title = `Status - Your ${getClaimType(this.props.claim)} Claim`;
```
austin-martinez: This is throwing errors in the console, the claim may not be loaded when the component is mounted (this is also true on the details and files tabs, but those are typically accessed after this tab). I think a generic title here might be best, and possibly update it when the props change. @jbalboni Do you have any suggestions on handling this case?

src/js/disability-benefits/components/ClaimEstimationPage.jsx
jbalboni: That looks right to me. We should move this file to the `containers` directory, though, since our convention is that page components that use Redux go there and not in `components/`.

src/js/disability-benefits/containers/StatusPage.jsx
jbalboni: Yeah, this is somewhat awkward. You'll need to check `this.props.loading` in `componentDidMount()` and use a generic title if it's still loading. That will catch the case where someone has navigated directly to this page and data is still loading or if they navigated here from a page that already loaded the claim. Then you will need to use `componentDidUpdate()` to detect when loading finishes by comparing `this.props.loading` and `prevProps.loading`. You can see an example in the current code for that method in this component (though with an extra check you won't need for the title).

src/js/disability-benefits/components/ClaimEstimationPage.jsx
jbalboni: There's that and the ClaimsDecision content which I'm not sure about.

jbalboni (commented):
Looks like there are some changes that Austin pointed out, but otherwise it looks good. I think we need to check with Suzanne about how to handle the content for the ClaimsDecision alert and the ClaimEstimationPage.

src/js/disability-benefits/components/ClaimEstimationPage.jsx
carlmjohnson: Maybe? For now, let's consider that a separate issue, since this change can go out before the switch to all claims.

src/js/disability-benefits/containers/AdditionalEvidencePage.jsx
carlmjohnson: Resolved by adding `const claim = this.props.claim;` to the file.

carlmjohnson:
I think the Jenkins failure has something to do with the slash in the branch name? The other two tests are passing. Merge?

jkassemi:
Addressing in https://github.com/department-of-veterans-affairs/vets-website/pull/4705/files - I'll merge master into this as soon as it's online and we should have consistent results.

carlmjohnson:
Thanks @jkassemi! All passing now.

src/js/disability-benefits/containers/ClaimEstimationPage.jsx
```diff
@@ -16,7 +18,7 @@ export default class ClaimEstimationPage extends React.Component {
             <nav className="va-nav-breadcrumbs">
               <ul className="row va-nav-breadcrumbs-list" role="menubar" aria-label="Primary">
                 <li><Link to="your-claims">Your claims</Link></li>
-                <li><Link to={`your-claims/${this.props.params.id}/status`}>Your Disability Compensation Claim</Link></li>
+                <li><Link to={`your-claims/${this.props.params.id}/status`}>Your {getClaimType(this.props.claim)} Claim</Link></li>
```
jbalboni: There's an issue here, similar to what you fixed on other pages: if you go to this page and refresh, you'll get a console error and the page won't load. You can add a check that `this.props.claim` exists before calling `getClaimType`, or add `loading` to `mapStateToProps` and check that.

jbalboni (approved):
Found one issue, can be merged after that's fixed.

PR #4705, opened 2017-01-17
Use modified BUILD_TAG docker image tag instead of BUILD_NUMBER

jkassemi:
BUILD_NUMBER is sequential and unique per branch. We need an additional identifier to prevent multiple branches from sharing the same docker container.

A date-based BUILD_ID was deprecated in Jenkins v1.597, but would have made a good candidate for this.


knkski (approved):
👍

PR #4708, opened 2017-01-18
Unit tests for messaging buttons

bshyong:
resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/795

### DIFF
--- a/src/js/messaging/components/buttons/ButtonPrint.jsx
+++ b/src/js/messaging/components/buttons/ButtonPrint.jsx
@@ -7,6 +7,7 @@ class ButtonPrint extends React.Component {
   }
 
   handleClick() {
+    this.props.onClick();
   }
 
   render() {
@@ -23,6 +24,7 @@ class ButtonPrint extends React.Component {
 }
 
 ButtonPrint.propTypes = {
+  onClick: React.PropTypes.func,
 };
 
 export default ButtonPrint;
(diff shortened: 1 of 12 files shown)


test/messaging/components/buttons/ButtonDelete.unit.spec.jsx
```diff
@@ -0,0 +1,36 @@
...
+
+  it('should handle click action properly', () => {
+    const onClickHandler = sinon.spy();
+    const buttonDelete = ReactTestUtils.renderIntoDocument(
+      <ButtonDelete
+          onClickHandler={onClickHandler}/>
+    );
+
+    buttonDelete.handleClick();
+    expect(onClickHandler.called).to.be.true;
```
U-DON: The other buttons have click handlers too, but why is this (and print) the only one that's tested? Is it based on the test coverage report?

U-DON: Also, `onClickHandler` seems redundant... Probably can just rename to `onClick` like the others for consistency.

src/js/messaging/components/buttons/ButtonPrint.jsx
```diff
@@ -23,6 +24,7 @@ class ButtonPrint extends React.Component {
 }
 
 ButtonPrint.propTypes = {
+  onClick: React.PropTypes.func,
 };
 
 export default ButtonPrint;
```
U-DON: `ButtonPrint` actually isn't being used at all right now, but I don't know if a print feature is going to be added in the future.

U-DON (approved):
LGTM with some comments.

test/messaging/components/buttons/ButtonDelete.unit.spec.jsx
bshyong: The other buttons don't have an internal `handleClick` function; they just pass down the onClick from the parent directly into their child button components, so there is nothing really to test there

PR #4713, opened 2017-01-19
Add sort select to claims list

austin-martinez:
Related to https://github.com/department-of-veterans-affairs/sunsets-team/issues/295

screen shot 2017-01-19 at 9 30 50 am

### DIFF
--- a/src/js/common/components/form-elements/ErrorableSelect.jsx
+++ b/src/js/common/components/form-elements/ErrorableSelect.jsx
@@ -98,7 +98,7 @@ class ErrorableSelect extends React.Component {
             autoComplete={this.props.autocomplete}
             value={selectedValue}
             onChange={this.handleChange}>
-          <option value="">{this.props.emptyDescription}</option>
+          {this.props.includeBlankOption && <option value="">{this.props.emptyDescription}</option>}
           {optionElements}
         </select>
         {toolTip}
@@ -123,6 +123,7 @@ ErrorableSelect.propTypes = {
         value: React.PropTypes.string }),
     ])).isRequired,
   required: React.PropTypes.bool,
+  includeBlankOption: React.PropTypes.bool,
   value: React.PropTypes.shape({
     value: React.PropTypes.string,
     dirty: React.PropTypes.bool
@@ -130,6 +131,10 @@ ErrorableSelect.propTypes = {
   onValueChange: React.PropTypes.func.isRequired,
   additionalClass: React.PropTypes.string,
   emptyDescription: React.PropTypes.string
+};
+
+ErrorableSelect.defaultProps = {
+  includeBlankOption: true
 };
 
 export default ErrorableSelect;
--- a/src/js/disability-benefits/actions/index.js
+++ b/src/js/disability-benefits/actions/index.js
@@ -44,9 +44,7 @@ export function getClaims(filter) {
       null,
       dispatch,
       claims => {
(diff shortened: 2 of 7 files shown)


src/js/common/components/form-elements/ErrorableSelect.jsx
```diff
@@ -98,7 +98,7 @@ class ErrorableSelect extends React.Component {
             autoComplete={this.props.autocomplete}
             value={selectedValue}
             onChange={this.handleChange}>
-          <option value="">{this.props.emptyDescription}</option>
+          {this.props.includeBlankOption && <option value="">{this.props.emptyDescription}</option>}
```
austin-martinez: @crwallace This is a change to a common component but I made the default state the same as it was previously

src/js/disability-benefits/containers/YourClaimsPage.jsx
```diff
@@ -31,6 +48,7 @@ class YourClaimsPage extends React.Component {
   componentWillReceiveProps(newProps) {
     if (this.props.allClaims && this.props.route.showClosedClaims !== newProps.route.showClosedClaims) {
       this.props.filterClaims(this.getFilter(newProps));
+      this.props.sortClaims();
```
jbalboni: Should `sortClaims` just reset the page, too? It looks like we always do that.

src/js/disability-benefits/reducers/claims-list.js
```diff
@@ -28,15 +47,26 @@ export default function claimsReducer(state = initialState, action) {
       });
     }
     case FILTER_CLAIMS: {
-      let sortedList = _.orderBy([phaseChangeDate, 'id'], 'desc', state.list);
+      let visibleList = state.list;
```
jbalboni: Should add visibleList to the initial state object.

src/js/disability-benefits/actions/index.js
```diff
@@ -45,6 +46,7 @@ export function getClaims(filter) {
       claims => {
         dispatch({ type: SET_CLAIMS, claims: claims.data, meta: claims.meta });
         dispatch({ type: FILTER_CLAIMS, filter });
+        dispatch({ type: SORT_CLAIMS });
```
jbalboni: This kind of multiple dispatching is not ideal in Redux, since it causes the UI to re-render after each dispatch. It also leads to harder to debug code, since actions aren't as easily associated with external events. 

And in the React components it results in UI code that is more coupled to the application logic because it has to know the series of granular updates to the application state that needs to happen when a user takes some action.

Some of that's just preference, but I do think making actions corresponding to what the user is doing rather than specific state changes results in more cleanly separated UI and application logic.

src/js/disability-benefits/containers/YourClaimsPage.jsx
```diff
@@ -45,6 +63,10 @@ class YourClaimsPage extends React.Component {
     }
     return undefined;
   }
+  handleSort(sortObject) {
+    this.props.sortClaims(sortObject.value);
+    this.changePage(1);
+  }
   changePage(page) {
     this.props.changePage(page);
     scrollToTop();
```
jbalboni: This scrollToTop call feels annoying when you're just changing the sorting dropdown. We could also skip it when changing tabs.

jbalboni (approved):
Made some comments, but looks good. I would prefer we do some of the Redux stuff a different way, but this works.

jbalboni (approved):
LGTM

PR #4715, opened 2017-01-19
Add array field components that follow our growable table patterns to 1995 form

jbalboni:
Related to #4608

This adds array field components that follow our growable table patterns. It also fixes issues with how touched fields are being tracked, how conditionally required fields are handled, and removing jsx from the form config.

Main changes are in ArrayField.jsx, ReviewArrayField.jsx, and ReviewCollapsibleChapter.jsx. I'll annotate the other changes so the reasons for them is clearer.

screen shot 2017-01-19 at 2 42 13 pm
screen shot 2017-01-19 at 2 41 54 pm
screen shot 2017-01-19 at 2 41 50 pm
screen shot 2017-01-19 at 2 41 43 pm
screen shot 2017-01-19 at 2 41 20 pm
screen shot 2017-01-19 at 2 41 14 pm
screen shot 2017-01-19 at 2 40 59 pm


src/js/common/schemaform/FieldTemplate.jsx
```diff
@@ -5,8 +5,8 @@ import React from 'react';
  */
 
 export default function FieldTemplate(props) {
-  const { id, schema, help, required, rawErrors, children, formContext, uiSchema } = props;
-  const hasErrors = (formContext.submitted || formContext.touched[id]) && rawErrors && rawErrors.length;
+  const { id, schema, help, required, rawErrors, children, formContext, touchedSchema, uiSchema } = props;
```
jbalboni: I updated how we tracked touched fields in our fork of the schema library, this is to accommodate that

src/js/common/schemaform/ObjectField.jsx
```diff
@@ -61,7 +63,14 @@ class ObjectField extends React.Component {
   }
 
   shouldComponentUpdate(nextProps, nextState) {
-    return shouldRender(this, nextProps, nextState);
+    const propsWithoutDataUnchanged = deepEquals(_.omit('formData', this.props), _.omit('formData', nextProps));
```
jbalboni: This addresses some performance issues I ran into on older phones. We can skip updates when the state and props versions of formData are the same and everything else matches.

src/js/common/schemaform/reducers/index.js
```diff
@@ -47,10 +47,12 @@ export default function createSchemaFormReducer(formConfig) {
         return _.set(['submission', action.field], action.value, state);
       }
       case SET_SUBMITTED: {
-        return _.assign(state, {
+        const submission = _.assign(state.submission, {
```
jbalboni: This was a bug I found.

src/js/common/schemaform/validation.js
```diff
@@ -78,9 +78,9 @@ export function transformErrors(errors, uiSchema) {
  * should call addError to add the error.
  */
 
-export function uiSchemaValidate(errors, uiSchema, formData, otherData, path = '') {
+export function uiSchemaValidate(errors, uiSchema, formData, formContext, path = '') {
```
jbalboni: formContext seemed more reasonable to pass in here.

src/js/edu-benefits/1995/config/form.js
```diff
@@ -73,60 +76,18 @@ const formConfig = {
...
-                    type: 'boolean'
-                  },
-                  chapter1607: {
-                    type: 'boolean'
-                  },
-                  transferOfEntitlement: {
-                    type: 'boolean'
-                  }
-                }
+                type: 'string',
```
jbalboni: This is supposed to only allow one benefit to be selected, so updating that.

package.json
```diff
@@ -143,6 +143,7 @@
     "react-transform-catch-errors": "^1.0.2",
     "react-transform-hmr": "^1.0.4",
     "reactable": "^0.14.0",
+    "recompose": "^0.21.2",
```
jbalboni: I added this because it makes doing some performance optimizations easier and we can import specific functions from it, so it isn't too bloated.

U-DON (commented):
Generally LGTM. Had a few comments.

src/js/common/schemaform/review/ReviewPage.jsx
```diff
@@ -54,22 +57,28 @@ export class ReviewPage extends React.Component {
...
-  formConfig: React.PropTypes.object.isRequired
+  form: React.PropTypes.object.isRequired,
+  route: React.PropTypes.shape({
+    formConfig: React.PropTypes.object.isRequired
+  }).isRequired
 };
 
 export default connect(mapStateToProps, mapDispatchToProps)(ReviewPage);
+
+export { ReviewPage };
```
jbalboni: We've been doing this to test the component and bypass the react-redux wrapper component.

src/js/common/schemaform/review/ReviewArrayField.jsx
```diff
@@ -0,0 +1,229 @@
...
+const scroller = Scroll.scroller;
+
+/* Growable table (Array) field on the Review page
+ *
+ * The idea here is that, because our pattern for growable tables on the review
+ * page is that each item can be in review or edit mode, we will treat each item
+ * as its own form page and this component will handle the edit/review states and
+ * make sure data is properly updated in the Redux store
+ */
+class ReviewArrayField extends React.Component {
```
jbalboni: I wasn't sure exactly what to to call this. It's not exactly the equivalent of the other ArrayField, like the normal/review ObjectField components are.

src/js/common/schemaform/ArrayField.jsx
```diff
@@ -0,0 +1,321 @@
...
+     * We're keeping the editing state in local state because it's easier to manage and
+     * doesn't need to persist from page to page
+     */
+    this.state = {
+      items,
+      editing: items.map(() => false)
+    };
+
+    this.onItemChange = this.onItemChange.bind(this);
+    this.handleAdd = this.handleAdd.bind(this);
```
jbalboni: I'm curious what you've been doing for this. I've normally done these bind calls for all methods that use `this` in their bodies (though I missed them here). But it's only really necessary if you're passing the methods as props to another component.

Do you always call `bind` or only when necessary?

src/js/common/schemaform/review/ReviewArrayField.jsx
jbalboni: I'll change it to ArrayField. The Review prefix isn't adding anything anyway.

src/js/common/schemaform/review/ReviewPage.jsx
jbalboni: Nevermind, misunderstood what you meant. I'll remove it from the class declaration.

src/js/common/schemaform/ArrayField.jsx
U-DON: Always been binding so far, just in case. Feels like the one time I don't, I'll end up debugging for way too long.

src/js/common/schemaform/review/ReviewPage.jsx
U-DON: Typically I've been doing the export on the class declaration rather than the separate export statement, but either way works.

src/js/common/schemaform/ArrayField.jsx
```diff
@@ -0,0 +1,324 @@
...
+  getDefaultFormState,
+  deepEquals
+} from 'react-jsonschema-form/lib/utils';
+
+import { errorSchemaIsValid } from './helpers';
+
+const Element = Scroll.Element;
+const scroller = Scroll.scroller;
+
+const scrollToFirstError = () => {
```
annekainicUSDS: We seem to have this basic function in a lot of places. Would it be useful to export it into helpers?

src/js/common/schemaform/review/ReviewPage.jsx
jbalboni: The hca code we started with was doing separate export statements, so we've been following that.

src/js/common/schemaform/ArrayField.jsx
jbalboni: Probably so at this point

annekainicUSDS (approved):
Few comments, mostly about concern around updating and storing state and making sure it's super clear what's happening. The additional comments in the code helped a lot as I was reviewing, so thank you for adding those. 

Tested it locally as well and the growable table stuff worked perfectly. Looks great!

src/js/edu-benefits/1995/config/form.js
jbalboni: Yes

src/js/common/schemaform/validation.js
jbalboni: Sort of. `otherData` was originally meant to be the form data other than the current page. `formContext` is an object that contains any custom data we need to pass to the components the schema form library renders. It would likely contain `otherData` at some point. 

We haven't needed this yet, so it may change again. I'm not sure of the exact use case yet.

src/js/common/schemaform/ArrayField.jsx
```diff
@@ -0,0 +1,324 @@
...
+    const newState = _.set('items', this.state.items.map(
+      (current, index) => {
+        return index === indexToChange
+          ? value
+          : current;
+      }
+    ), this.state);
+    // Then, update the actual form data
+    this.setState(newState, () => {
+      this.props.onChange(newState.items);
```
jbalboni: Basically, yes. The `onChange` call will update the overall form data help by the overall `Form` component, which will end up triggering an update to the Redux state.

src/js/common/schemaform/ArrayField.jsx
```diff
@@ -0,0 +1,324 @@
...
+   * Based on the default react-jsonschema-form optimizations, except it handles state/props syncing
+   * with formData
+   */
+  shouldComponentUpdate(nextProps, nextState) {
+    const propsWithoutDataUnchanged = deepEquals(_.omit('formData', this.props), _.omit('formData', nextProps));
+    const stateUnchanged = deepEquals(this.state, nextState);
+
+    // Because there's local state for the form data, we sometimes see form data updated in props that is already
+    // in state and we don't need to update
+    if (propsWithoutDataUnchanged && stateUnchanged && deepEquals(nextProps.formData, nextState.items)) {
```
jbalboni: Yeah, there are two optimizations here. One is to use local state, which was in the library's implementation of ArrayField, so I carried it over. The second is what I added to make sure we don't unnecessarily update when the local state and the data from props are the same.

It's probably worth removing the local state and seeing if the performance impact is noticeable. I agree that it'd be better not to have it, I just figured the authors of the library added it for a reason.

src/js/common/schemaform/review/ReviewCollapsibleChapter.jsx
```diff
@@ -97,17 +70,44 @@ export default class ReviewCollapsibleChapter extends React.Component {
...
-                    route={{ pageConfig: page }}/>
+                {hasNonArrayFields &&
+                  <FormPage
+                      reviewPage
+                      hideTitle={this.props.pages.length === 1}
+                      onEdit={() => this.handleEdit(page.pageKey, !editing)}
+                      onSubmit={() => this.handleEdit(page.pageKey, false)}
+                      reviewMode={!editing}
+                      route={{ pageConfig: page }}>
+                    {!editing ? <div/> : <ProgressButton
```
jbalboni: The library displays its own submit button if you don't pass any children to the `Form` component. So this works around that.

jbalboni:
@annekainicUSDS There's an item for moving more stuff in Redux state and out of local state, so I think we can attempt to move as much as possible out of local state in that issue. I think the main issue there is just making sure it doesn't impact performance on slower phones.

PR #4719, opened 2017-01-20
EVSS: Add /track-claims/ as a valid path

earthboundkid:
Refs. https://github.com/department-of-veterans-affairs/sunsets-team/issues/281

The bit with window.location feels a little hacky to me. Is there a better way @jbalboni ?

### DIFF
--- a/content/pages/track-claims/index.md
+++ b/content/pages/track-claims/index.md
@@ -1,0 +1,16 @@
+---
+title: Track Claims
+entryname: disability-benefits
+layout: page-react.html
+---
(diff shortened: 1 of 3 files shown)


jbalboni (approved):
LGTM. I don't think there's a better way than checking pathname. We'll also have to have devops update the proxy config to forward all `/track-claims` urls to the index page.

carlmjohnson:
One step ahead: https://github.com/department-of-veterans-affairs/devops/pull/1369 😄

PR #4720, opened 2017-01-20
Validation clean up and tests for 1995 form

jbalboni:
Related to #4608 

This cleans up the validation to match the 1995 form as much as possible right now and adds some tests.

No visual changes, but overall the changes are:

- Remove isValid state from Redux store and instead re-run validation when needed (isValidForm func)
- Move some functions from helpers to validation.js
- Fix some errors in uiSchemaValidate
- Tweak the form config to remove unnecessary validations and add date range check
- Add a couple of text tweaks that were in the GH issues


src/js/common/schemaform/validation.js
```diff
@@ -79,51 +81,89 @@ export function transformErrors(errors, uiSchema) {
...
-      uiSchemaValidate(errors, uiSchema.items, formData, formContext, newPath);
-    });
-  } else {
-    Object.keys(uiSchema)
-      .filter(prop => !prop.startsWith('ui:'))
-      .forEach((item) => {
-        const nextPath = path !== '' ? `${path}.${item}` : item;
-        if (!_.get(nextPath, errors)) {
-          _.get(path, errors)[item] = {
+  if (uiSchema) {
```
jbalboni: The changes in here are hard to understand, I think. They are mainly defensive: make sure we skip everything when there's no uiSchema (meaning there's no custom validation to check) and handle the case where path is an empty string.

annekainicUSDS (approved):
LGTM!

src/js/common/schemaform/validation.js
jbalboni: It happens the very first time this function is called, when we're at the root of the form schema.

PR #4721, opened 2017-01-20
Use error message responses from the API

bshyong:
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/779
Addresses https://github.com/department-of-veterans-affairs/vets.gov-team/issues/819

This PR enables the use of API error messages for fetching and refilling prescriptions.

### DIFF
--- a/src/js/rx/components/ErrorMessages.jsx
+++ b/src/js/rx/components/ErrorMessages.jsx
@@ -2,7 +2,7 @@ import React from 'react';
 
 const ErrorMessages = ({ errors }) => {
   if (errors && errors.length > 0) {
-    return (<div>
+    return (<span>
       <br/>
       {
         errors.map(e => {
@@ -14,8 +14,7 @@ const ErrorMessages = ({ errors }) => {
           );
         })
       }
-      <br/>
-    </div>);
+    </span>);
   }
   return null;
 };
--- a/src/js/rx/components/ErrorMessages.jsx
+++ b/src/js/rx/components/ErrorMessages.jsx
@@ -2,18 +2,19 @@ import React from 'react';
 
 const ErrorMessages = ({ errors }) => {
   if (errors && errors.length > 0) {
-    return (<p>
+    return (<div className="rx-error-messages">
+      <ul>
       {
         errors.map(e => {
           return (
-            <span key={e.code}>
+            <li key={e.code}>
               Error {e.code}: {e.detail}
-              <br/>
-            </span>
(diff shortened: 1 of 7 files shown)


src/js/rx/containers/Active.jsx
```diff
@@ -159,7 +161,8 @@ class Active extends React.Component {
       content = (
         <p className="rx-tab-explainer rx-loading-error">
           We couldn't retrieve your prescriptions.
-          Please refresh this page or try again later.
+          Please refresh this page or try again later.&nbsp;
```
U-DON: There's an `nbsp;` here but not in the change for the History page. I'm not sure whether `ErrorMessages` should be an inline or block element, but if the space isn't necessary, it should be removed.

src/js/rx/components/ErrorMessages.jsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+
+const ErrorMessages = ({ errors }) => {
```
U-DON: If this is going to be used with Secure Messaging as while, it should probably be in `common`. Can be done when that becomes the case.

src/js/rx/components/ErrorMessages.jsx
```diff
@@ -0,0 +1,30 @@
...
+  if (errors && errors.length > 0) {
+    return (<span>
+      <br/>
+      {
+        errors.map(e => {
+          return (
+            <span key={e.code}>
+              Error {e.code}: {e.detail}
+              <br/>
```
bshyong: Makes sense.  I wanted to ensure that there was always space between content and this error message; I'll tweak this.

bshyong:
Changes made: These are screenshots for what the messages should look like 

screen shot 2017-01-20 at 10 17 03 am
screen shot 2017-01-20 at 9 57 23 am

src/js/rx/components/ErrorMessages.jsx
U-DON: Even these `span` elements can be converted to block-level elements without the `br` if there's going to be a line break between each one. Maybe the outer can be a `div` and the inner errors can be `p` elements. Or `ul` and `li` elements, which might be more semantic.

bshyong: ul and li are probably more semantic: I added styles so that we can use that for the error messages component

U-DON (approved):
LGTM.

PR #4723, opened 2017-01-20
Change api url for dev and staging environments

annekainicUSDS:
The new vets-api for HCA needs to be tested in staging before being released in production. In order to do so we want to post HCA data to the new endpoint only if in dev and staging.

Need back-end to review and test: @markolson let me know how this looks!

### DIFF
--- a/src/js/hca/components/HealthCareApp.jsx
+++ b/src/js/hca/components/HealthCareApp.jsx
@@ -148,8 +148,6 @@ class HealthCareApp extends React.Component {
     if (testBuild) {
       // Allow e2e tests to override API URL
       // TODO: Remove the need for a separate code path here
-      //
-
       apiUrl = window.VetsGov.api.url === ''
         ? `${environment.API_URL}/v0/health_care_applications`
         : `${window.VetsGov.api.url}/v0/health_care_applications`;
--- a/src/js/hca/components/HealthCareApp.jsx
+++ b/src/js/hca/components/HealthCareApp.jsx
@@ -149,7 +149,7 @@ class HealthCareApp extends React.Component {
     // endpoints to submit to the new API. Keeping the same endpoints for production.
     if (testBuild) {
       // Allow e2e tests to override API URL
-      // TODO: Remove the need for a separate code path here
+      // Remove the need for a separate code path here
       apiUrl = window.VetsGov.api.url === ''
         ? `${environment.API_URL}/v0/health_care_applications`
         : `${window.VetsGov.api.url}/v0/health_care_applications`;
--- a/src/js/hca/components/HealthCareApp.jsx
+++ b/src/js/hca/components/HealthCareApp.jsx
@@ -145,7 +145,7 @@ class HealthCareApp extends React.Component {
     window.dataLayer.push({ event: 'submit-button-clicked' });
     const formIsValid = validations.isValidForm(veteran);
 
-    // In order to test the new Rails API in staging, we are temporarily changing the 
+    // In order to test the new Rails API in staging, we are temporarily changing the
     // endpoints to submit to the new API. Keeping the same endpoints for production.
     if (testBuild) {
       // Allow e2e tests to override API URL
--- a/src/js/hca/components/HealthCareApp.jsx
+++ b/src/js/hca/components/HealthCareApp.jsx
@@ -146,7 +146,14 @@ class HealthCareApp extends React.Component {
     const formIsValid = validations.isValidForm(veteran);
 
     if (testBuild) {
-      apiUrl = `${environment.API_URL}/v0/health_care_applications`;
+      // Allow e2e tests to override API URL
+      // TODO: Remove the need for a separate code path here
+      //
+
+      apiUrl = window.VetsGov.api.url === ''
+        ? `${environment.API_URL}/v0/health_care_applications`
+        : `${window.VetsGov.api.url}/v0/health_care_applications`;
+
       submissionPost.body = JSON.stringify({ form: submissionPost.body });
     }
 
--- a/src/js/hca/components/HealthCareApp.jsx
+++ b/src/js/hca/components/HealthCareApp.jsx
@@ -5,6 +5,8 @@ import _ from 'lodash';
 import { connect } from 'react-redux';
 import fetch from 'isomorphic-fetch';
 
+import environment from '../../common/helpers/environment';
+
 import IntroductionSection from './IntroductionSection.jsx';
 import Nav from '../../common/components/Nav.jsx';
 import ProgressButton from '../../common/components/form-elements/ProgressButton';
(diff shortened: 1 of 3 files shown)


src/js/hca/components/HealthCareApp.jsx
```diff
@@ -129,15 +129,20 @@ class HealthCareApp extends React.Component {
     e.preventDefault();
     const veteran = this.props.data;
     const path = this.props.location.pathname;
+    let apiUrl = `${window.VetsGov.api.url}/api/hca/v1/application`;
 
     window.dataLayer.push({ event: 'submit-button-clicked' });
     const formIsValid = validations.isValidForm(veteran);
 
+    if (__BUILDTYPE__ === 'development' || __BUILDTYPE__ === 'staging') {
+      apiUrl = `${window.VetsGov.api.url}/api/v0/health_care_applications`;
```
jbalboni: Do we need to update the e2e tests with this url?

annekainicUSDS: Good point. Not sure why that didn't cause things to fail...

jbalboni (approved):
LGTM, not sure what the build failures are, though.

src/js/hca/components/HealthCareApp.jsx
```diff
@@ -129,23 +129,38 @@ class HealthCareApp extends React.Component {
...
     const formIsValid = validations.isValidForm(veteran);
 
-    if (formIsValid && veteran.privacyAgreementAccepted) {
-      this.props.onUpdateSubmissionStatus('submitPending');
-
-      // POST data to endpoint
-      fetch(`${window.VetsGov.api.url}/api/hca/v1/application`, {
+    if (__BUILDTYPE__ === 'development' || __BUILDTYPE__ === 'staging') {
+      apiUrl = `${window.VetsGov.api.url}/v0/health_care_applications`;
+      submissionPost = {
```
lihanli: i think this line should be `submissionPost.body = JSON.stringify({ form: submissionPost.body });`
i didn't see form in the fetch api docs, what the api wants in the request body is `{ form: [the form as a JSON string] }`, we want it to be JSON encoded twice because otherwise the middleware will change the casing of the form and that will break things

lihanli (changes requested):

src/js/hca/components/HealthCareApp.jsx
annekainicUSDS: Ah gotcha, thanks! I'll update this.

lihanli (approved):

src/js/hca/utils/validations.js
```diff
@@ -316,6 +316,15 @@ function isValidServiceInformation(data) {
 }
 
 function isValidForm(data) {
+  let isValidFinancialDisclosureSections = true;
```
markolson: what are these changes for?

annekainicUSDS: I was hitting a validation error when testing the form locally. I'm not sure if this has been failing all along or not. I had entered that I was married, but declined to provide financial info. Where this failed was when trying to validate the spouse's full name. Because the name was empty (since I skipped these fields), it failed validation when running `isValidForm` --> `isValidSpouseInformation` on submit. I added the check first in `isValidForm` that it should only validate those sections that conditionally appear in the financial disclosure section if the value of `discloseFinancialInformation` is true.

annekainicUSDS:
@markolson @lihanli when testing this locally I'm getting a 404 when trying to POST to http://localhost:3001/v0/health_care_applications. I'm running vets-api locally. Not sure what else might be wrong?

annekainicUSDS:
@markolson @lihanli in testing this locally, submitting to the new endpoint doesn't seem to be working. Getting a 500 error. Can you take a look at this and see if you can help me identify the issue? This is the reason the e2e tests are failing.

annekainicUSDS:
@markolson @lihanli seeing an issue now where 2 pieces of information we used to get in the response from the Node server (`formSubmissionId` and `timestamp`) are not being sent now. We will need this information for the confirmation page. Can this be added to the response?

markolson:
Working on that. Should have it up in a PR this afternoon

annekainicUSDS:
@markolson thanks for the fix! Made some additional updates and it is now working locally. E2e tests are failing though, so trying to debug that now.

test/util/hca-helpers.js
```diff
@@ -436,12 +436,25 @@ function completeVaInsuranceInformation(client, data, onlyRequiredFields) {
 }
 
 function initApplicationSubmitMock() {
+  // TODO: Fully migrate to v1
+  mock(null, {
```
annekainicUSDS: @alexose wondering why we just created 2 separate mocks?

test/util/hca-helpers.js
```diff
@@ -436,12 +436,25 @@ function completeVaInsuranceInformation(client, data, onlyRequiredFields) {
...
+      }
+    }
+  });
   mock(null, {
     path: '/api/hca/v1/application',
     verb: 'post',
     value: {
-      formSubmissionId: '123fake-submission-id-567',
-      timeStamp: '2016-05-16'
+      data: {
```
annekainicUSDS: I think for this old endpoint we don't want this information to be nested within `data`, at least it was working find that way before.

test/util/hca-helpers.js
alexose-gov: This is so we can support both endpoints.  It would be better to create each endpoint conditionally, but this gets the job done.

test/util/hca-helpers.js
alexose-gov: Ah, I may have gotten it backwards.  Sorry about that

annekainicUSDS:
@markolson @jbalboni This should finally be ready to go (tests passing!). Could probably use on more quick review as it's changed quite a bit from the last time you took a look.

src/js/hca/components/HealthCareApp.jsx
```diff
@@ -129,23 +131,38 @@ class HealthCareApp extends React.Component {
...
+        'Content-Type': 'application/json'
+      },
+      timeout: 10000, // 10 seconds
+      body: veteranToApplication(veteran)
+    };
 
     window.dataLayer.push({ event: 'submit-button-clicked' });
     const formIsValid = validations.isValidForm(veteran);
 
+    if (testBuild) {
```
jbalboni: I think I'd put a comment here about why this is necessary right now.

Also, and it may not be worth it at this point, we do have the ability to create "feature flags" now in the webpack config that can be used for this sort of thing.

jbalboni (approved):
Only had one comment, looks ok to me.

src/js/hca/components/HealthCareApp.jsx
annekainicUSDS: Will add comment, thanks! And good to know, since this is temporary I may not change things now but will definitely take advantage of that in the future, this was pretty complicated to hack together :)

PR #4733, opened 2017-01-23
Fix for renamed click handler

bshyong:
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/838


U-DON (approved):
D'oh. Forgot this had to be changed too. I think things like this would become more obvious in unit tests if we eliminate the warnings we currently have. This is a required prop already, but it would be more conspicuous if we only allow legitimate warnings to appear.

LGTM.

PR #4735, opened 2017-01-24
Redirect old routes for education to the root route

annekainicUSDS:
Solves an issue where the VONAPP team was linking to an old education route that no longer exists. This addition will redirect all routes that are not matched to the root route.


jbalboni (approved):
LGTM

PR #4737, opened 2017-01-24
Add ability to use our expanding group pattern on 1995 form

jbalboni:
Closes [869](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/869)

Adds the ability to have fields expand or collapse underneath others, using our ExpandingGroup component/pattern. To trigger this, you use `expandUnder` in `ui:options` for fields that are conditionally shown. The value of `expandUnder` should be the property that triggers hiding/showing. See the form config change for an example.

Veteran information:
screen shot 2017-01-24 at 11 46 31 am
screen shot 2017-01-24 at 11 46 35 am

Review:
screen shot 2017-01-24 at 11 46 12 am
screen shot 2017-01-24 at 11 46 20 am


src/js/common/schemaform/ObjectField.jsx
```diff
@@ -47,6 +49,17 @@ class ObjectField extends React.Component {
     this.onPropertyChange = this.onPropertyChange.bind(this);
     this.onPropertyBlur = this.onPropertyBlur.bind(this);
     this.isRequired = this.isRequired.bind(this);
+    // This runs a series of steps that order properties and then group them into
+    // expandable groups. If there are no expandable groups, then the end result of this
+    // will be an array of single item arrays
+    this.orderAndFilterProperties = _.flow(
```
annekainicUSDS: Just curious why this is defined on `this`?

annekainicUSDS (commented):
LGTM!

src/js/common/schemaform/ObjectField.jsx
jbalboni: `_.flow` creates a function to process the list of object properties and it doesn't need to be recreated every time we render or props change, just run again. So this seemed more efficient. 

This is the type of code I would normally write as a class method, but that wouldn't work well with `_.flow`, which I thought made the code clearer.

PR #4740, opened 2017-01-24
Switch from phantomjs to electron

alexose:
This is a big PR with a lot of changes:

* Added electron and chromedriver dependencies
* Pulled out a lot of the Selenium and PhantomJS dependencies and installation scripts
* Upgraded nightwatch
* Added dependencies that PhantomJS somehow was able to ignore
* Added some extra magic to our .travis.yml that will make electron work headlessly
* Introduced a way to disable smooth scrolling on a per-page basis (at some point, this needs to have much less code repetition.)
* Added a `selectDropdown` helper that sidesteps the problems with clicking on large `select` lists
* Fixed a gaggle of tests so that they pass consistently

Despite all of this, it's possible that I haven't actually improved overall build times.  There's still a bunch of work to be done with Travis to avoid having to do things like download a Java runtime.

Futurevision:
I tried not to impose my view on the existing test suite, although I would like to consider drastically reducing the amount of individual assertions.

Even without reducing the workload, there should be plenty of room for speed improvements.  These will be showing up in future PRs, I think.

While working on this PR, I created a docker image that mimics our Travis environment.  This is so we can avoid that annoying thing where tests pass locally and fail in CI.  I [published it over at Docker hub](https://hub.docker.com/r/alexose/vets-gov-e2e/), but it needs documentation.


jcmeloni-usds:
This is so exciting!

robbiethegeek:
@alexose can you fix the conflicts then @crwallace @aub @robbiethegeek-usds will all review

robbiethegeek-usds:
@alexose can you please resolve the conflicts and then @aub @crwallace and I can jump in and do a review.

alexose:
On it

package.json
```diff
@@ -81,10 +83,13 @@
     "foundation-sites": "^5.5.3",
     "get-env": "^0.4.0",
     "glob": "^7.0.5",
+    "history": "2.1.2",
     "http-server": "^0.9.0",
     "img-loader": "^1.3.1",
     "imports-loader": "^0.6.5",
     "ismobilejs": "^0.4.0",
+    "isomorphic-fetch": "http://registry.npmjs.org/isomorphic-fetch/-/isomorphic-fetch-2.2.1.tgz",
+    "jquery": "^3.1.1",
```
crwallace: @alexose -- Do we need jquery? I know we are trying to pull it out of the codebase if we can.

alexose-gov: I don't think we need jQuery anymore.  Will test and find out

jcmeloni-usds: It's ALMOST out, pending a PR that @aub has, I believe.

crwallace:
I feel like you taking the time to work on tests to make sure that they pass consistently, alothough might not speed up tests, will contribute to developer happiness and I feel like that is important to note. 

One other question I have is if we are okay with the popup that opens while these tests are running. I was unable to run these tests locally in the background as the popup kept...well...popping up. 

Would love others thoughts on this.

crwallace:
sorry that was an accidental close.

alexose-gov:
It's definitely annoying that the electron window pops up frequently and switches focus-- I plan to fix that by wrapping the call to the binary in another file (which is exactly how spectron does it).

test/util/e2e-helpers.js
```diff
@@ -1,13 +1,105 @@
...
+			/* eslint-disable */
+			var element = document.querySelector(clientSelect);
+			var event = new Event('change', { bubbles: true });
+			element.value = clientValue;
+			element.dispatchEvent(event);
+			return element.value;
+			/* eslint-disable */
+		},
+		[select, value]);
+}
```
aub: Funky spacing here.

aub (changes requested):
This looks good to me with a couple of small comments below.

PR #4742, opened 2017-01-24
Styling changes to improve RX sort menu

bshyong:
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/855

### DIFF
--- a/src/js/rx/components/SortMenu.jsx
+++ b/src/js/rx/components/SortMenu.jsx
@@ -81,7 +81,7 @@ class SortMenu extends React.Component {
     });
 
     return (
-      <div className="row">
+      <div>
         <form className="rx-sort va-dnp">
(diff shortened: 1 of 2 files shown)


U-DON (changes requested):
I'm seeing this with the branch checked out locally. Looks assigning `row` is pushing the row a little further out left.

> screen shot 2017-01-24 at 8 56 45 pm

I'm not sure `width: auto` with padding to make space for the `option` is the way to fix the `ul` getting pushed to the next line. Probably the cleanest fix would be to make the desktop version of the `rx-sort` be `display: flex` and giving `flex: 1 0` to the `select`.

bshyong:
Agreed; we can use flexbox for the desktop display. I added min-width to the select so that it can shrink without getting too small.

U-DON (approved):
LGTM.

PR #4746, opened 2017-01-25
Add paperclip icon to messages with attachments

bshyong:
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/853


src/js/messaging/containers/Folder.jsx
```diff
@@ -279,7 +280,7 @@ export class Folder extends React.Component {
         fields.pop();
       }
     } else {
-      fields.push({ label: '', value: 'moveToButton' });
+      fields.push({ label: '', value: 'moveToButton', nonSortable: true });
```
U-DON: 👍

src/js/messaging/containers/Folder.jsx
```diff
@@ -312,9 +313,12 @@ export class Folder extends React.Component {
             onToggleMoveTo={() => this.props.toggleFolderMoveTo(id)}/>
       );
 
+      const attachmentIcon = message.attachment ? (<span className="fa fa-paperclip" aria-label="Message has an attachment"></span>) : null;
```
U-DON: If it works, that's fine, but I usually see `fa` used with `i` elements. Don't know if there might be other styles that depend on it being an `i` element. Could also wrap it (if turning it into `i`) and have the `"Message has an attachment"` inside a `usa-sr-only` classed `span`.

U-DON (commented):
I would also check to see how mobile layout gets affected. There are some styles that depend on the positioning of the cells. Also, might want clarification on how we show the paperclip in mobile.

bshyong:
Good call on the mobile layout—I'll check with Gina on how those should be displayed. The mobile views have very limited real estate.

bshyong:
Mobile layout is implemented

screen shot 2017-01-26 at 10 00 58 am

src/sass/messaging/partials/_messaging-folder.scss
```diff
@@ -105,10 +105,24 @@
...
 
-      // Date
+      // Attachment icon
       &:nth-child(3) {
         font-weight: normal;
         position: absolute;
+        right: 25%;
+        text-align: right;
+        top: 0;
+        width: 25%;
```
U-DON: I was trying to make it so that the widths of the From/To field and the Date field would add up to 100%. If this is to take up some space, the width of From/To or Date should be decreased. I would probably at least decrease the width of the Date field a bit so that the icon isn't too far away from it, while ensuring a long From/To doesn't mess up the layout.

U-DON (changes requested):

bshyong:
Noted. paperclip icon is now 5%, letting everything add up to 100%. Date width has been decreased, while the From/To length has been kept the same at 75%.

The from/to field will wrap so longer strings are not a problem

screen shot 2017-01-26 at 11 47 43 am

U-DON (approved):

PR #4751, opened 2017-01-26
Add asset fingerprinting

aub:
This adds hashes to the names of generated JS and CSS assets through WebPack so that we can tell the browser to cache those assets and then do cache busting as they change. Adding the hashes was pretty easy, but the more complicated thing was replacing the references to these files in the HTML and CSS files with the hashed versions of them. That work is done, and heavily comment in, build.js.

This is working correctly for me locally both in development mode and when building the production assets and running them. It's also passing tests. Do reviewers have ideas about anything further we need to do to feel sure that it will definitely work when deployed to production?

Also, my run of `npm shrinkwrap --dev` added a bunch of spurious changes to the shrinkwrap file... is that a problem?

### DIFF
--- a/config/webpack.config.js
+++ b/config/webpack.config.js
@@ -1,6 +1,7 @@
 // Staging config. Also the default config that prod and dev are based off of.
 
 const ExtractTextPlugin = require('extract-text-webpack-plugin');
+const ManifestPlugin = require('webpack-manifest-plugin');
 const bourbon = require('bourbon').includePaths;
 const neat = require('bourbon-neat').includePaths;
 const path = require('path');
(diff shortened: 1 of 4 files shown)


robbiethegeek:
LGTM @crwallace @alexose any final review / comment?

crwallace:
This LGTM but I would like to know what changes got added to the shrinkwrap file just because we want to keep that file as accurate as possible.

aub:
@crwallace you can see the diff if you click the link to show it in the files changed section.

crwallace:
@aub -- This looks like it just changed some of the urls to the `http` version instead of the `https` version. So that seems fine to me.

crwallace (approved):

PR #4752, opened 2017-01-26
Breadcrumbs fix for RX

bshyong:
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/809

### DIFF
--- a/src/js/rx/components/Breadcrumbs.jsx
+++ b/src/js/rx/components/Breadcrumbs.jsx
@@ -28,7 +28,7 @@ class Breadcrumbs extends React.Component {
     if (pathname.match(/\/\d+$/)) {
       crumbs.push(<Link to="/" key="prescriptions">Prescription Refills</Link>);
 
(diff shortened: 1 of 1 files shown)


src/js/rx/components/Breadcrumbs.jsx
```diff
@@ -13,6 +27,11 @@ class Breadcrumbs extends React.Component {
 
     if (pathname.match(/\/\d+$/)) {
       crumbs.push(<Link to="/" key="prescriptions">Prescription Refills</Link>);
+
+      if (this.state.prevPath.includes('history')) {
```
U-DON: I think IE doesn't support `String.prototype.includes`.

bshyong: This stuff should ideally be polyfilled, but I'm switching to use Regex to avoid any unexpected issues

U-DON (approved):
LGTM.

PR #4759, opened 2017-01-27
Persist view preferences for RX 

bshyong:
This PR uses the session store to persist the view option that a user selects in the Prescription app 

Resolves https://github.com/department-of-veterans-affairs/kudos-team/issues/330


U-DON (approved):
LGTM. I would say keeping the state in the Redux store could also work, but the thing with that is that it wouldn't persist on a page refresh. Meanwhile, keeping it in `sessionStorage` will survive a page refresh, although I think opening it again in a new window or tab might not.

PR #4760, opened 2017-01-27
Move things out of local state for 1995

jbalboni:
Closes [871](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/871)

This removes our use of local state for the schema in FormPage and form data in the object and array field components. It also adds some performance optimizations, because I was seeing some sluggishness when typing into form fields on a slow Android phone. Seems pretty solid now, especially when built in prod mode.

### DIFF
--- a/src/js/common/schemaform/ArrayField.jsx
+++ b/src/js/common/schemaform/ArrayField.jsx
@@ -26,7 +26,7 @@ export default class ArrayField extends React.Component {
      */
 
     this.state = {
-      editing: this.props.formData ? this.props.formData.map(() => false) : [true]
+      editing: props.formData ? props.formData.map(() => false) : [true]
     };
 
     this.onItemChange = this.onItemChange.bind(this);
--- a/src/js/common/schemaform/ArrayField.jsx
+++ b/src/js/common/schemaform/ArrayField.jsx
@@ -21,25 +21,12 @@ export default class ArrayField extends React.Component {
     super(props);
 
     /*
-     * We always want to show one empty item in the array, so add that if necessary
-     *
-     * It will not get saved unless a user changes data in it
-     */
-    const formData = Array.isArray(props.formData) ? props.formData : null;
-    let items;
-    if (!formData || formData.length === 0) {
-      items = [getDefaultFormState(props.schema.items, null, props.registry.definitions) || {}];
-    } else {
-      items = getDefaultFormState(props.schema, formData, props.registry.definitions) || [];
-    }
-
-    /*
      * We're keeping the editing state in local state because it's easier to manage and
      * doesn't need to persist from page to page
      */
+
     this.state = {
-      items,
-      editing: items.map(() => false)
+      editing: this.props.formData ? this.props.formData.map(() => false) : [true]
     };
 
     this.onItemChange = this.onItemChange.bind(this);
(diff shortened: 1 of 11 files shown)


src/js/edu-benefits/1990/components/debug/PerfPanel.jsx
```diff
@@ -1,3 +1,4 @@
+/* eslint no-console: 0 */
```
jbalboni: I was using this component for testing, thought I would leave in the changes I made to it.

src/js/common/schemaform/validation.js
```diff
@@ -115,7 +115,7 @@ export function uiSchemaValidate(errors, uiSchema, formData, formContext, path =
     }
 
     const validations = uiSchema['ui:validations'];
-    if (validations) {
+    if (validations && currentData) {
```
jbalboni: We don't need to run our validation if there's no data, since that will be caught by the required validation.

src/js/common/schemaform/ObjectField.jsx
```diff
@@ -49,29 +36,13 @@ class ObjectField extends React.Component {
...
-    const state = this.getStateFromProps(nextProps);
-    const { formData } = nextProps;
-    if (formData && objectKeysHaveChanged(formData, this.state)) {
-      // We *need* to replace state entirely here has we have received formData
-      // holding different keys (so with some removed).
-      this.state = state;
-      this.forceUpdate();
-    } else {
-      this.setState(state);
+    if (this.props.schema !== nextProps.schema || this.props.uiSchema !== nextProps.uiSchema) {
```
jbalboni: We only need to recalculate the list of properties if the schema have changed, so we can store the list on the component instance and only update it when necessary.

src/js/common/schemaform/ObjectField.jsx
```diff
@@ -49,29 +36,13 @@ class ObjectField extends React.Component {
...
-    // will be an array of single item arrays
-    this.orderAndFilterProperties = _.flow(
-      properties => orderProperties(properties, _.get('ui:order', this.props.uiSchema)),
-      _.groupBy((item) => {
-        const expandUnderField = _.get([item, 'ui:options', 'expandUnder'], this.props.uiSchema);
-        return expandUnderField || item;
-      }),
-      _.values
-    );
+    this.SchemaField = pureWithDeepEquals(this.props.registry.fields.SchemaField);
```
jbalboni: This resulted in something like a 50% improvement in the length of the key press event when typing in a form field.

U-DON (approved):
LGTM.

src/js/common/schemaform/ArrayField.jsx
```diff
@@ -52,36 +39,19 @@ export default class ArrayField extends React.Component {
...
   shouldComponentUpdate(nextProps, nextState) {
-    const propsWithoutDataUnchanged = deepEquals(_.omit('formData', this.props), _.omit('formData', nextProps));
-    const stateUnchanged = deepEquals(this.state, nextState);
-
-    if (propsWithoutDataUnchanged && stateUnchanged && deepEquals(nextProps.formData, nextState.items)) {
-      return false;
-    }
-
-    return true;
+    return !deepEquals(this.props, nextProps) || nextState !== this.state;
```
jbalboni: That's true, but this is to avoid updates to this component when a parent component triggers an update but the props for this component aren't actually different.

PR #4764, opened 2017-01-27
Skip conditional pages when their dependencies haven't been satisfied.

U-DON:
Closes department-of-veterans-affairs/vets.gov-team#868.

#### Changes:
* Going back or forward will skip conditional pages that haven't passed their conditions.
* Conditional pages that haven't passed their conditions won't appear in expanded review chapters.
* Added `data` object in `depends` object in form config to reflect the structure of the form data that it's compared with.

### DIFF
--- a/src/js/common/schemaform/review/ReviewCollapsibleChapter.jsx
+++ b/src/js/common/schemaform/review/ReviewCollapsibleChapter.jsx
@@ -2,7 +2,7 @@ import React from 'react';
 import Scroll from 'react-scroll';
 import _ from 'lodash';
 
-import { focusElement } from '../../utils/helpers';
+import { focusElement, getActivePages } from '../../utils/helpers';
 import FormPage from '../FormPage';
 import { getArrayFields, hasFieldsOtherThanArray } from '../helpers';
 import ArrayField from './ArrayField';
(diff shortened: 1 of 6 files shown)


src/js/common/schemaform/review/ReviewCollapsibleChapter.jsx
```diff
@@ -65,11 +65,14 @@ export default class ReviewCollapsibleChapter extends React.Component {
...
     let pageContent = null;
+
     if (this.state.open) {
+      const { data, pages } = this.props;
+
       pageContent = (
         <div id={`collapsible-${this.id}`} className="usa-accordion-content">
-          {this.props.pages.map(page => {
-            const editing = this.props.data[page.pageKey].editMode;
+          {getActivePages(pages, data).map(page => {
```
jbalboni: Can you add a test for this component where a conditional page is hidden?

I haven't written any tests for FormPage yet, so I can cover the changes there in the unit test item I'm working on.

src/js/common/schemaform/review/ReviewCollapsibleChapter.jsx
```diff
@@ -84,7 +87,7 @@ export default class ReviewCollapsibleChapter extends React.Component {
                 {hasNonArrayFields &&
                   <FormPage
                       reviewPage
-                      hideTitle={this.props.pages.length === 1}
+                      hideTitle={pages.length === 1}
```
jbalboni: Since this is checking the unfiltered page list, if a chapter has two pages and one is conditional, this will show the title, even if the user only encountered one page. Shouldn't affect the 1995, but I think it's more consistent to check that the length of the active pages is 1.

jbalboni (approved):
LGTM.

jbalboni:
I realized when I was looking at this that we should probably check the conditional pages in the overall form validation and when we flatten data to send it to the server. But I don't know for sure what the server transformation will look like yet, so I think we can leave it to a separate item.

U-DON:
Addressed PR comments.

jbalboni:
@U-DON Awesome. Looks good to merge whenever you're ready.

PR #4769, opened 2017-01-30
Review app deploy configuration

jkassemi:
Trigger a review instance deployment during CI.

This provides a GitHub deployment and accompanying link to the instance that runs against the vets-api at `master` within the development environment. If vets-api should be run with a specific branch, use the Jenkins interface to trigger a Review Instance Deployment manually.

This PR adds the process to the Jenkinsfile, and permits configuring the custom api and site base URLs required to operate a review instance.

### DIFF
--- a/src/js/common/helpers/environment.js
+++ b/src/js/common/helpers/environment.js
@@ -1,7 +1,7 @@
 const _Environments = {
   production: { API_URL: 'https://api.vets.gov', BASE_URL: 'https://www.vets.gov' },
   staging: { API_URL: 'https://staging-api.vets.gov', BASE_URL: 'https://staging.vets.gov' },
-  development: { API_URL: 'https://dev-api.vets.gov', BASE_URL: 'https://dev.vets.gov' },
+  development: { API_URL: (process.env.API_URL || 'https://dev-api.vets.gov'), BASE_URL: (process.env.BASE_URL || 'https://dev.vets.gov') },
   local: { API_URL: `http://${location.hostname}:3000`, BASE_URL: `http://${location.hostname}:3001` },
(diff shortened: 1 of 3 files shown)


src/js/common/helpers/environment.js
```diff
@@ -1,7 +1,7 @@
 const _Environments = {
   production: { API_URL: 'https://api.vets.gov', BASE_URL: 'https://www.vets.gov' },
   staging: { API_URL: 'https://staging-api.vets.gov', BASE_URL: 'https://staging.vets.gov' },
-  development: { API_URL: 'https://dev-api.vets.gov', BASE_URL: 'https://dev.vets.gov' },
+  development: { API_URL: (process.env.API_URL || 'https://dev-api.vets.gov'), BASE_URL: (process.env.BASE_URL || 'https://dev.vets.gov') },
```
jkassemi: There's some cleanup that could happen here when we get some ideas for how to most cleanly set up environments and feature flags.

jkassemi:
Requires https://github.com/department-of-veterans-affairs/devops/pull/1410 for successful deployment.

jkassemi:
@knkski - I'll reassign to you for review after devops#1410 so I can confirm the deployment works as expected.

CyberKoz:
The contents of the PR look good to me, the Multiple messages by va-bot concern me

CyberKoz (approved):
The only thing that looked off to me was the API_URL and BASE_URL, but I confirmed that they are being set as part of the review-instance-configure role

:+1:

jkassemi:
Multiple messages are from the testing process - it took a little while to get things working as expected (Jenkins script security exceptions required running the job to get an exception string which was then added to `scriptApproval.xml`, and then re-running for the next one).

PR #4770, opened 2017-01-31
Pull out vendor JS code into a separate, cacheable file

aub:
department-of-veterans-affairs/vets.gov-team#928

The JS code we bring in from vendors doesn't change very often. For that reason, it makes sense to pull it out into its own javascript file so the browser can cache it and hold onto the cache for longer than it would the individual entry files that include our code. Doing so also has the benefit of making the entry files much smaller, so that if a user went to multiple entries then they would download much less code.

Annoyingly, this requires us to specify which libraries we want to include in the vendor package, so that if that list changes then we have to update it manually in the WebPack config. But other than that the change is pretty straightforward.

### DIFF
--- a/config/webpack.config.js
+++ b/config/webpack.config.js
@@ -27,7 +27,7 @@ const configGenerator = (options) => {
   if (options.entry) {
     filesToBuild = _.pick(entryFiles, options.entry.split(',').map(x => x.trim()));
   }
-  filesToBuild['vendor'] = [
+  filesToBuild.vendor = [
     'core-js',
     'history',
     'lodash',
--- a/config/webpack.config.js
+++ b/config/webpack.config.js
@@ -3,6 +3,7 @@
 const ExtractTextPlugin = require('extract-text-webpack-plugin');
 const ManifestPlugin = require('webpack-manifest-plugin');
 const ChunkManifestPlugin = require('chunk-manifest-webpack-plugin');
+const WebpackMd5Hash = require('webpack-md5-hash');
 const bourbon = require('bourbon').includePaths;
 const neat = require('bourbon-neat').includePaths;
 const path = require('path');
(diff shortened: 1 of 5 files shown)


robbiethegeek-usds:
LGTM 👍 once @crwallace reviews and approves feel free to merge.

config/webpack.config.js
```diff
@@ -27,6 +27,18 @@ const configGenerator = (options) => {
   if (options.entry) {
     filesToBuild = _.pick(entryFiles, options.entry.split(',').map(x => x.trim()));
   }
+  filesToBuild.vendor = [
+    'core-js',
+    'history',
+    'lodash',
```
jbalboni: We might want to skip lodash here. We currently have babel and webpack plugins that transform it so we're only including the specific functions in use, but this will include all of lodash for every app.

aub: All right, removed lodash from the list.

jbalboni:
@aub One other thing: because it's common to the other app bundles, the vendor bundle is considered the entry bundle by Webpack, which means it contains a manifest with a list of all the other bundles. This is fine in dev, but in production that manifest will have file names with hashes, which means the vendor bundle will change anytime any of our app bundles change, busting the cache.

I think we would need to use another plugin to pull out that manifest and inline it in the html, like described [here](https://medium.com/@okonetchnikov/long-term-caching-of-static-assets-with-webpack-1ecb139adb95#.ypid6zi4l).

aub:
@jbalboni Ugh, ok I was hoping I could get away without doing that. I had built the thing in production a few times and noticed that the hashes didn't change, but I guess I hadn't done it while changing our code in between. On it.

aub:
@jbalboni Ok, I added a couple more things to make it create consistent hashes and it's working now. I did a test locally by making a spurious change to one of the non-vendor files, rebuilt, and then tested that only the appropriate hashes changed.

aub:
@crwallace want to give final approval on this one?

crwallace (approved):
@aub this looks great!

PR #4773, opened 2017-01-31
Add more unit tests for schemaform

jbalboni:
Closes #[873](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/873)

This adds a bunch of unit tests for the schemaform common code. There are a couple minor code tweaks, too.


annekainicUSDS (approved):
Looks great, just added a couple of questions.

test/common/schemaform/ObjectField.unit.spec.jsx
```diff
@@ -151,4 +151,52 @@ describe('Schemaform: ObjectField', () => {
...
+      <ObjectField
+          schema={schema}
+          idSchema={{}}
+          onChange={onChange}
+          onBlur={onBlur}/>
+    );
+
+    tree.getMountedInstance().onPropertyChange('test')('value');
+
+    expect(onChange.firstCall.args[0]).to.eql({
```
jbalboni: That would work here as far as I know, but I did it this way because it's is a more precise test. The onChange function should be called with an object with just one property called `test`.

test/common/schemaform/ArrayField.unit.spec.jsx
```diff
@@ -166,5 +169,31 @@ describe('Schemaform ArrayField', () => {
...
       expect(tree.getMountedInstance().state.editing[2]).to.be.false;
     });
+    it('add when invalid', () => {
+      errorSchema[1] = { __errors: ['Test error'] };
+      tree.getMountedInstance().handleAdd();
+
+      expect(tree.getMountedInstance().state.touchedSchema[1]).to.be.true;
+    });
+    it('remove', () => {
+      expect(tree.everySubTree('SchemaField').length).to.equal(1);
```
jbalboni: In this case, no. Normally the onChange prop would end up triggering another render with new props, but we're using a spy for that, so nothing changes. We could mock up the onChange prop to force a re-render the component with updated formData, but that seemed like overkill.

PR #4777, opened 2017-01-31
Custom address field jsonschemform

annekainicUSDS:
Still in progress, but adding a custom widget for Address component.


src/js/common/schemaform/Address.jsx
```diff
@@ -0,0 +1,275 @@
...
+  isMilitaryCity(city) {
+    const lowerCity = city.toLowerCase().trim();
+
+    return lowerCity === 'apo' || lowerCity === 'fpo' || lowerCity === 'dpo';
+  }
+
+  render() {
+    let stateList = [];
+
+    const states = {
```
annekainicUSDS: For whatever reason, even though I'm importing `states` from `utils/options-for-select`, it wasn't registered as actually being imported. For a shortcut to get this working for now, just copied and pasted the states list here, but will address a more permanent fix later.

src/js/common/schemaform/Address.jsx
```diff
@@ -0,0 +1,275 @@
...
+        { label: 'Virgin Islands', value: 'VI' },
+        { label: 'Virginia', value: 'VA' },
+        { label: 'Washington', value: 'WA' },
+        { label: 'West Virginia', value: 'WV' },
+        { label: 'Wisconsin', value: 'WI' },
+        { label: 'Wyoming', value: 'WY' }
+      ]
+    };
+
+    const selectedCountry = this.props.formData.country.value || this.props.formData.country;
```
annekainicUSDS: Have to check if `this.props.formData.country.value` exists first because when the component first renders, `this.props.formData.country` is just `'USA'`, so checking for `.value` will fail. But once something is selected, it changes the value to `{value: 'USA', dirty: false}`.

src/js/common/schemaform/Address.jsx
```diff
@@ -0,0 +1,127 @@
...
+    this.handleChange = this.handleChange.bind(this);
+    this.handleBlur = this.handleBlur.bind(this);
+    this.state = { value: props.formData, touched: { street: false, city: false, state: false, country: false, postalCode: false } };
+  }
+
+  componentWillMount() {
+    this.id = _.uniqueId('address-input-');
+  }
+
+  handleBlur(field) {
```
jbalboni: This will probably need to change some. I think it can be one line:

`this.props.onBlur([field])`

src/js/edu-benefits/routes.jsx
```diff
@@ -39,6 +45,12 @@ export default function createRoutes(store) {
     );
   }
 
+  const routesFor1995 = __BUILDTYPE__ !== 'production' && __BUILDTYPE__ !== 'staging'
```
jbalboni: You can remove all the changes in this file now.

src/js/common/schemaform/Address.jsx
```diff
@@ -0,0 +1,127 @@
...
+
+  isMilitaryCity(city) {
+    const lowerCity = city.toLowerCase().trim();
+
+    return lowerCity === 'apo' || lowerCity === 'fpo' || lowerCity === 'dpo';
+  }
+
+  render() {
+    let stateList = [];
+    const selectedCountry = this.props.formData.country.value || this.props.formData.country;
```
jbalboni: One thing we can do here is set formData to the default value when it's undefined:

```
let formData = this.props.formData;
if (typeof formData === 'undefined') {
  formData = getDefaultFormState(this.props.schema, formData, this.props.schema.definitions);
}
```

Then formData.country should always be populated, because it's defaulted to USA in the address schema.

You'll need to import that `getDefaultFormState` function from the library:

`import { getDefaultFormState } from 'react-jsonschema-form/lib/utils';`

annekainicUSDS:
@jbalboni This should be ready for another review now

src/js/common/schemaform/Address.jsx
```diff
@@ -0,0 +1,175 @@
...
+    const lowerCity = city.toLowerCase().trim();
+
+    return lowerCity === 'apo' || lowerCity === 'fpo' || lowerCity === 'dpo';
+  }
+
+  render() {
+    let stateList = [];
+    const { errorSchema, formData, formContext, touchedSchema, schema, idSchema, uiSchema, registry } = this.props;
+    const SchemaField = registry.fields.SchemaField;
+    const selectedCountry = formData.country.value || formData.country;
```
jbalboni: I think we remove the `formData.country.value` part of this.

src/js/common/schemaform/Address.jsx
```diff
@@ -0,0 +1,175 @@
...
+
+    this.props.onChange(newState);
+  }
+
+  isRequired(name) {
+    const { schema } = this.props;
+    const schemaRequired = Array.isArray(schema.required) &&
+      schema.required.indexOf(name) !== -1;
+
+    if (schemaRequired) {
```
jbalboni: I think we can remove this if statement and just return `schemaRequired` at the bottom.

src/js/common/schemaform/Address.jsx
```diff
@@ -0,0 +1,175 @@
...
+ *
+ * No validation is provided through a currently stubbed isAddressValid function.
+ */
+class Address extends React.Component {
+  constructor(props) {
+    super();
+    this.handleChange = this.handleChange.bind(this);
+    this.onPropertyBlur = this.onPropertyBlur.bind(this);
+    this.isRequired = this.isRequired.bind(this);
+    this.state = { value: props.formData, touched: { street: false, city: false, state: false, country: false, postalCode: false } };
```
jbalboni: I think we can remove this line.

src/js/common/schemaform/Address.jsx
```diff
@@ -0,0 +1,175 @@
+import React from 'react';
+import _ from 'lodash';
+import { set, assign } from 'lodash/fp';
+
+import { states } from '../utils/options-for-select';
+
+/**
+ * Input component for an address.
+ *
+ * No validation is provided through a currently stubbed isAddressValid function.
```
jbalboni: This line can be removed, I think?

jbalboni (commented):
Added some comments. It would also be good to add some tests, though we could do that in a separate item.

src/js/common/schemaform/FormPage.jsx
```diff
@@ -19,12 +20,14 @@ import { setData } from './actions';
...
-  ArrayField
+  ArrayField,
+  address: Address
 };
 
 const reviewFields = {
   ObjectField: ReviewObjectField,
-  ArrayField
+  ArrayField,
+  address: Address
```
jbalboni: I think for now you could change this to `address: ReviewObjectField`. Otherwise, we're going to show form fields in "view" mode on the review page.

We may need to write a custom Address component for the review page, but the normal object field might be good enough.

src/js/common/schemaform/Address.jsx
```diff
@@ -0,0 +1,175 @@
...
+  render() {
+    let stateList = [];
+    const { errorSchema, formData, formContext, touchedSchema, schema, idSchema, uiSchema, registry } = this.props;
+    const SchemaField = registry.fields.SchemaField;
+    const selectedCountry = formData.country.value || formData.country;
+    let postalCodeUiSchema = uiSchema.postalCode;
+    let stateUiSchema = uiSchema.state;
+    let stateSchema = schema.properties.state;
+
+    if (formData.country === 'CAN') {
```
jbalboni: Could probably use `selectedCountry` if both of these if statements.

annekainicUSDS:
@jbalboni made the fixes you recommend. Started trying to write some tests as well but they were failing and it's hard for me to figure out why. I created a separate branch off of this one for the tests, so if we want to merge this in the meantime, I can continue working on the tests and try to figure out a better way to debug them.

jbalboni (approved):

jbalboni:
@annekainicUSDS Cool, let's go ahead and merge this, then.

PR #4779, opened 2017-01-31
EVSS route and service name change

earthboundkid:
As part of changing to doing all claims, we will change the URL and service name away from "disability-claims". This must not be merged until after that change is live, but I am opening it now to keep us from forgetting to merge later.

### DIFF
--- a/src/js/disability-benefits/containers/DisabilityBenefitsApp.jsx
+++ b/src/js/disability-benefits/containers/DisabilityBenefitsApp.jsx
@@ -45,7 +45,7 @@ class DisabilityBenefitsApp extends React.Component {
     const { available, synced, authorized } = this.props;
 
     return (
-      <RequiredLoginView authRequired={3} serviceRequired={"disability-benefits"}>
+      <RequiredLoginView authRequired={3} serviceRequired={"evss-claims"}>
         <AppContent
             authorized={authorized}
             available={available}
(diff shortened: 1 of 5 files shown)


austin-martinez:
@jbalboni The backend changes maintained both route and service names, updating front end here so we can remove the old routes and service name in API

src/js/disability-benefits/containers/DisabilityBenefitsApp.jsx
```diff
@@ -45,7 +45,7 @@ class DisabilityBenefitsApp extends React.Component {
     const { available, synced, authorized } = this.props;
 
     return (
-      <RequiredLoginView authRequired={3} serviceRequired={"disability-benefits"}>
+      <RequiredLoginView authRequired={3} serviceRequired={"evss-claims"}>
```
jbalboni: We made the backend change to change this service name, too? I ask because I don't know how that works and assume it's separate from changing the api routes.

jbalboni (approved):
Looks good to me. I have one question (mostly just out of caution).

src/js/disability-benefits/containers/DisabilityBenefitsApp.jsx
austin-martinez: We are sending both service names
https://github.com/department-of-veterans-affairs/vets-api/blob/master/app/serializers/user_serializer.rb#L36

We also had both routes active, I have another PR for the API that will remove the route and old service name after this is merged

PR #4780, opened 2017-02-01
Add preliminary list of blue button report mappings

bshyong:
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/963

List of ambiguities here that we will resolve later this week with MHV:
- missing MHV key for `Labs and Tests, Self-Reported`
- missing MHV key for `Vitals and Readings, Self-Reported`
- what is the difference between `prescription` and `medication`?
- what does `vachemlab` map to?
- what does `vahth` map to?

### DIFF
--- a/src/js/blue-button/config.js
+++ b/src/js/blue-button/config.js
@@ -1,162 +1,198 @@
 module.exports = {
   reportTypes: {
-    appointments: [
-      {
-        value: 'futureappointment',
-        label: 'Future VA Appointments',
-      },
-      {
-        value: 'pastappointment',
-        label: 'Past VA Appointments (limited to past 2 years)',
-      },
-    ],
-    medications: [
-      {
-        value: 'prescription',
-        label: 'VA Medication History',
-      },
-      {
-        value: 'medication',
-        label: 'Medications and Supplements, Self-Reported',
-      },
-    ],
-    labs: [
-      {
-        value: 'labsandtest',
-        label: 'VA Laboratory Results',
-        hold: 3,
-      },
-      {
-        value: 'vapatholog',
-        label: 'VA Pathology Reports',
-        hold: 14,
-      },
-      {
-        value: 'varadiolog',
-        label: 'VA Radiology Reports',
-        hold: 3,
-      },
-      {
-        value: 'vaek',
-        label: 'VA Electrocardiogram (EKG) History performed at VA Treating Facilities',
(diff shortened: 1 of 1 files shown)


src/js/blue-button/config.js
```diff
@@ -0,0 +1,162 @@
...
+        label: 'VA Electrocardiogram (EKG) History performed at VA Treating Facilities',
+      },
+      {
+        value: '',
+        label: 'Labs and Tests, Self-Reported',
+      },
+    ],
+    ehr: [
+      {
+        value: 'vaproblemlis',
```
U-DON: `vaproblemlist`

src/js/blue-button/config.js
```diff
@@ -0,0 +1,162 @@
...
+      },
+    ],
+    labs: [
+      {
+        value: 'labsandtest',
+        label: 'VA Laboratory Results',
+        hold: 3,
+      },
+      {
+        value: 'vapatholog',
```
U-DON: `vapathology`

U-DON (changes requested):
Aside from the ones I commented on, it looks like all of the values are missing their last characters.

U-DON:
Might also consider giving each report type a title too.

src/js/blue-button/config.js
```diff
@@ -0,0 +1,162 @@
...
+        label: 'VA Pathology Reports',
+        hold: 14,
+      },
+      {
+        value: 'varadiolog',
```
austin-martinez: `varadiology`
Looks like a lot of these are missing the last letter
`welnes` -> `wellness`, etc.

bshyong:
@U-DON @austin-martinez I believe the key names are intentional - they depend on what MHV will take; I pulled them from the blue button client in the API https://github.com/department-of-veterans-affairs/vets-api/blob/master/lib/bb/generate_report_request_form.rb#L6

U-DON:
@bshyong It seems like the keys in the code you linked have their last letters, though. Like the first one, `seiactivityjournal` has that last `l`, but the one in the config is `seiactivityjourna` without the `l`.

bshyong:
Adding titles is a good suggestion; we might as well have it all in the config file. Those are added in the latest commit. Key names are also fixed

U-DON (approved):
LGTM.

PR #4783, opened 2017-02-01
Add initial Blue Button report generation page

bshyong:
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/931

mocks: https://marvelapp.com/2210i4e/screen/24082732


src/js/blue-button/containers/BlueButtonApp.jsx
```diff
@@ -9,7 +9,7 @@ function AppContent({ children, isDataAvailable }) {
   const unregistered = isDataAvailable === false;
   let view;
 
-  if (unregistered) {
+  if (!unregistered) {
```
U-DON: I don't think this is correct. If the user is unregistered (when `isDataAvailable` is specifically `false`), it should show the error message.

If you were getting it after being logged in, it could be because the account doesn't have access to this service, or Blue Button hasn't been added to the list of services for the user on the API side yet. I believe it's the latter, since I don't see it within the list of services:

* https://github.com/department-of-veterans-affairs/vets-api/blob/master/lib/backend_services.rb
* https://github.com/department-of-veterans-affairs/vets-api/blob/master/docs/api/user.md#response

So we should probably get someone to add it there. For now, in order to test, all I was doing was removing the `RequiredLoginView` wrapper, since we haven't hooked up to the API yet.

U-DON (changes requested):

U-DON:
Other than the handling around the `RequiredLoginView`, LGTM.

src/js/blue-button/containers/BlueButtonApp.jsx
```diff
@@ -31,7 +31,13 @@ class BlueButtonApp extends React.Component {
     return (
       <RequiredLoginView authRequired={3} serviceRequired={"bluebutton"}>
         <AppContent>
-          {this.props.children}
+          <div className="bb-app">
+            <div className="row">
+              <div className="columns small-12">
```
U-DON: If we're wrapping all the pages with this, I think the `.bb-app .row` rule can be removed from the stylesheet and the `row` class can be removed from `DownloadPage`.

U-DON:
The `to` in the date range isn't vertically centered on Firefox. Looks like this for me:

> screen shot 2017-02-01 at 7 22 57 pm

src/js/blue-button/containers/Main.jsx
```diff
@@ -0,0 +1,123 @@
...
+  render() {
+    // TODO: clean this up and hook up to action/reducer
+    const radioButtonProps = {
+      label: '',
+      options: [
+        { label: '3 months', value: '3mo' },
+        { label: '6 months', value: '6mo' },
+        { label: '1 year', value: '1yr' },
+        {
+          label: (
```
U-DON: I think this label is failing the propType validation on `ErrorableRadioButtons` because it's an `element` instead of `string`. I would make the propType a `oneOfType` with `string` and `element`.

src/js/blue-button/components/Breadcrumbs.jsx
```diff
@@ -0,0 +1,26 @@
+import { Link } from 'react-router';
+import React from 'react';
+
+class Breadcrumbs extends React.Component {
```
bshyong: 👍  Moved it to the Main App component

src/js/blue-button/containers/BlueButtonApp.jsx
```diff
@@ -34,6 +35,7 @@ class BlueButtonApp extends React.Component {
           <div className="bb-app">
```
U-DON: `.bb-app` is already used in `AppContent` and can be removed here..

U-DON:
`.bb-app` is already used to wrap `children` in `AppContent` and can be removed from its children.

U-DON:
Also, not sure why, but clicking the radio buttons doesn't seem to change the state of the input to checked.

bshyong:
It looks like the radio buttons are a controlled component, so we need to change the value of the `value` prop that gets passed in whenever a radio button is clicked (in the `onValueChange` function). If you manually change it to something like `3mo`, that button will always be checked.

bshyong:
Screenshot:

screencapture-localhost-3001-healthcare-blue-button-1486009620783

bshyong:
@austin-martinez do you want to give this one more quick lookover before I merge it?

austin-martinez (approved):
LGTM 🍶

PR #4789, opened 2017-02-02
Share 1995 schema fields with backend

jbalboni:
Closes [872](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/872)

This imports the 1995 schema from vets-json-schema and uses the definitions for the form config. This meant renaming some fields and also a better way of organizing common fields.

Changes (roughly):

- Create files for common fields that contain the uiSchema and regular schema definitions (see src/js/common/schemaform/definitions/). The uiSchema parts of those files were moved from uiDefinitions.js
- Add a `defaultDefinitions` property to the form so that we can put common schema definitions in one place for all pages
- Rename some fields `view:fieldName`. Lihan suggested namespacing fields that are only on the front end, so we can strip them out easier. This is the start of that.
- Add routing number validation
- Change email matching to use a generic matching validator
- Replace address schema from backend with our temporary address schema, so we don't break anything for now.
- Fixed an error with the review page growable table


src/js/common/schemaform/definitions/bankAccount.js
```diff
@@ -0,0 +1,32 @@
+import _ from 'lodash/fp';
+import fullSchema1995 from 'vets-json-schema/dist/change-of-program-schema.json';
```
jbalboni: This will hopefully change to import a common definitions file, instead of the 1995 schema, since these are common fields and it seems weird for them to depend on a specific form's schema.

src/js/edu-benefits/1995/helpers.jsx
```diff
@@ -10,6 +10,16 @@ export const benefitsLabels = {
   transferOfEntitlement: 'Transfer of Entitlement Program (TOE)'
 };
 
+export const educationTypeLabels = {
+  college: 'College',
```
jbalboni: I think Alex has real copy for these that we can use.

annekainicUSDS (approved):
LGTM

PR #4794, opened 2017-02-03
Auto-hide #va-modal

cvalarida:
Adds CSS rule to hide `#va-modal` to be opened if '#modal' is found in the url (content/includes/modal.html).

I wasn't sure if calling `modal.hide()` would be a better solution if `#va-modal` is used elsewhere, so I kept that as a comment in modal.html.

I looked for other uses of both `#va-moda`l and `modal.html`, but couldn't find any. Still, at this point don't know if it's safe, so somebody more familiar with the codebase should probably do some more thorough testing.

Also, Atom likes to remove trailing spaces, so that's what the seemingly-unchanged lines are. Sorry for the noise.

Closes department-of-veterans-affairs/vets.gov-team#965

### DIFF
--- a/content/includes/modal.html
+++ b/content/includes/modal.html
@@ -50,9 +50,7 @@ content.
 (function($){
   // TODO: eliminate jQuery dependency
   var modal = $('#va-modal');
-  // Start hidden and only open if the hash is found
-  // Alternatively, use the css #va-modal { display: none }
-  // modal.hide();
+  // Note: Hidden by default by the css
   if (document.location.hash === '#modal'){
     open();
   }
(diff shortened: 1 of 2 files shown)


content/includes/modal.html
```diff
@@ -50,22 +50,25 @@ <h5 class="va-modal-split-title">Servicemembers and Veterans applying for educat
 (function($){
   // TODO: eliminate jQuery dependency
   var modal = $('#va-modal').focus();
```
jbalboni: Can you remove the `focus` call from this and add it inside the `open` function? I'm not sure why it was like this before, but it doesn't make sense to focus on the modal unless we've opened it.

jbalboni:
I don't think we use `va-modal` as an id anywhere else, so I think adding the css class is fine. It should also avoid any flashing that might occur if it's shown by default and then hidden.

jbalboni (approved):
Looks good, made one comment that needs to be addressed and then this can be merged.

jbalboni:
@cvalarida Despite the failed check above, you can merge squash this into master.

PR #4795, opened 2017-02-03
Changed .va-footer-linkgroup a to be an inline block.

ghost:
Closes department-of-veterans-affairs/vets.gov-team#967


jcmeloni-usds:
LGTM

PR #4797, opened 2017-02-03
content/wip/updating order of cards to match nav

rroueche:
I just moved Health Care Benefits to the first card instead of the last, so that when we launch the nav, the order will match.


jkassemi (approved):

jkassemi:
LGTM - may have jumped the gun on the merge if it needed an additional reviewer - let me know if I should revert and throw back in for another PR.

rroueche:
@emdashed as long as you're ok w/this then I think we're good!

emdashed:
All good - thanks for checking @rroueche !!

PR #4801, opened 2017-02-03
Removal of schema submodule

ben-USDS-damman:
- Removes remnant schema directory
- Removes submodule init setup instruction from readme

My understanding is the submodule was intended to be removed previously when we switched to npm packaging for schemas (but please @patrickvinograd or someone should check to make sure no scripts or what have you are still mistakenly expecting the `schema` directory.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ very secret.
 
 | I want to... | Then you should... |
 | ------------ | ------------------ |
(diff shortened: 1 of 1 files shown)


patrickvinograd:
I'm not aware of any CI/CD/anything that is expecting a submodule to be there.

ayaleloehr:
@annekainicUSDS wanted to check with you on this, as I feel like I see Rainbows using submodules sometimes. Any objections to this PR?

jbalboni:
We (Rainbows) stopped using the submodule in favor of an npm module, so this should be fine.

PR #4802, opened 2017-02-04
Form 22-1990 Progress Bar

cvalarida:
The original issue (department-of-veterans-affairs/vets.gov-team#906) was to just get rid of the subway and use the segmented progress bar. Without the subway, it doesn't make sense to have the border around the content, and it's a bit off-putting to not have the title (as shown below for the 22-1995):

image

I'm not sure what to do with the headers. Sometimes they make sense (Personal Information --> Direct deposit):

image

And sometimes they're redundant (Veteran Information --> Veteran information)

image

I didn't want to over-implement without getting the green light on design, so this is what it looks like now. Suggestions?


src/js/edu-benefits/1990/containers/EduBenefitsApp.jsx
```diff
@@ -75,6 +75,14 @@ class EduBenefitsApp extends React.Component {
       submitBenefitsForm(this.props.data);
     };
 
+    // There's got to be a better way to do this using schemaform, but for now...
+    // Find the index of the chapter...
+    const currentIndex = _.findIndex(chapters, (chapter) => {
```
jbalboni: There should be logic for this in the NavHeader component.

jbalboni:
@cvalarida You can remove the headers on each page. We can have @goldenmeanie go through them and see if there are any that are necessary for specific pages, but the default should be that we don't show them.

src/js/edu-benefits/1990/containers/EduBenefitsApp.jsx
```diff
@@ -93,16 +101,16 @@ class EduBenefitsApp extends React.Component {
       <div className="row">
         {devPanel}
         <Element name="topScrollElement"/>
-        <div className="medium-4 columns show-for-medium-up">
-          <Nav
-              data={data}
-              pages={pageState}
-              chapters={chapters}
-              currentUrl={currentLocation.pathname}/>
+        <div className="medium-8 columns">
```
jbalboni: Does this need to be in a separate row, or can it be in the same div as NavHeader?

src/js/edu-benefits/1990/containers/EduBenefitsApp.jsx
```diff
@@ -93,16 +101,16 @@ class EduBenefitsApp extends React.Component {
...
+          {
+            // Only render SegmentedProgressBar if we have a current chapter
+            // Could move this into SegmentedProgressBar
+            currentIndex ? <SegmentedProgressBar total={chapters.length} current={currentIndex}/> : null
```
jbalboni: Hiding this when we don't have a chapter will probably work, but the actual logic is that we don't want this to show up on the introduction and submit-message pages. You could check for that specifically, which would make what's happening clearer.

jbalboni:
@cvalarida It probably does make sense to include a title like the 1995 form. You can use the text from the introduction page (Apply for education benefits).

goldenmeanie:
Is origin/1990-progress-bar up to date. I'll pull it down and take a spin through to try to clear this up!

jbalboni:
@goldenmeanie Should be.

src/js/edu-benefits/1990/containers/EduBenefitsApp.jsx
cvalarida: It can be in the same div, but when I tried it, the NavHeader and progress bar just got squashed together:

image

I'll see what I can do to separate them while they're in the same div.

cvalarida: Look good?

image

jbalboni: Looks good to me!

jbalboni:
@cvalarida It looks like there's a linting error that's causing the build check to fail above.

cvalarida:
To sum up the changes:

---
image

---
image

---
image

cvalarida:
Removed the border and aligned the intro page with the title to match the 22-1995.

image

---
While still keeping the indented forms.

image

jbalboni (approved):
Made a couple comments, but this looks good. One other thing I noticed from trying the form locally, you can remove the "Review Application" header on the review page, too.

jbalboni:
@cvalarida Looks good to me. I think you can merge this once you fix the merge conflict (and the build passes).

PR #4803, opened 2017-02-06
Initial skeleton of GIBCT

ben-USDS-damman:
Since I initially worked on the frontend we've planned design and UX changes. After talking it over with @saneshark on Friday I created a new clean branch off of master, built a skeleton based on the blue button and facility location code, and began layering on my [previous work](https://github.com/department-of-veterans-affairs/vets-website/commit/06bf5cae11b3ef021718cb10ac48f4e4c35649d8).

As I do this I'm cleaning it up and making sure it comports to the work @melwoodard and others have done on the new design and content. It's mostly copy-paste and then making any necessary changes to form/function.

**Seeking feedback on the general code structure and also wondering how I can make this use the new site-wide commonStore** instead of its own store. @crwallace? I tried to do it based on blue button but couldn't get it to work. _#newb_

image

image


src/js/gi/components/PreviewBanner.jsx
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
+
+class PreviewBanner extends React.Component {
+  render() {
+    if (!this.props.show) {
+      return null;
+    }
+
+    const headerDisplacementCSS = `
```
bshyong: It is possible to use these styles by having them in a class? instead of inline?

src/js/gi/containers/GiBillApp.jsx
```diff
@@ -0,0 +1,66 @@
+import React from 'react';
+import { connect } from 'react-redux';
+import Modals from '../containers/Modals';
+import PreviewBanner from '../components/PreviewBanner';
+import Breadcrumbs from '../components/Breadcrumbs';
+import AboutThisTool from '../components/AboutThisTool';
+
+function AppContent({ children, isDataAvailable }) {
```
bshyong: If veterans do not need to be registered to use this tool, we can get rid of the `AppContent` wrapper. It is meant to be used in conjunction with a `RequiredLoginView` component that wraps the entire thing (and calls AppContent after the user logs in successfully).

src/js/gi/containers/Modals.jsx
```diff
@@ -0,0 +1,288 @@
...
+    this.renderProfileSummaryModals = this.renderProfileSummaryModals.bind(this);
+    this.renderProfileHistoryModals = this.renderProfileHistoryModals.bind(this);
+    this.renderProfileCautionFlagModals = this.renderProfileCautionFlagModals.bind(this);
+    this.renderProfileCalculatorModals = this.renderProfileCalculatorModals.bind(this);
+  }
+
+  renderProfilePageModals() {
+    return (
+      <span>
+        <Modal onClose={() => this.props.hideModal()} visible={!!this.props.giModals.retention}>
```
bshyong: for these onClose function with no parameters, I believe you can do `onClose={this.props.hideModal}`; no need to wrap it in a function

src/js/gi/store/index.js
```diff
@@ -0,0 +1,8 @@
+import { createStore, applyMiddleware, compose } from 'redux';
```
bshyong: We can use the common store for this instead of having a separate store for GIBCT

src/js/gi/gi-entry.jsx
```diff
@@ -0,0 +1,26 @@
+import 'core-js';
+require('../common');  // common javascript.
+import React from 'react';
+import ReactDOM from 'react-dom';
+import { Router, useRouterHistory } from 'react-router';
+import { Provider } from 'react-redux';
+import { createHistory } from 'history';
+import initReact from '../common/init-react';
+import routes from './routes';
+import { store } from './store';
```
bshyong: Use the common store `import { commonStore } from '../common/store';`

To allow it to work, you also need to go to `common/reducers/index.js` and import your GIBCT reducers. 
https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/js/common/reducers/index.js

Then, in your GIBCT app, in the components that use state you can access them like so (depends on where you put GIBCT in the common store reducer):
 
This is an example from blue button

```
  const bbState = state.health.bb;

  return {
    form: bbState.form,
  };
};
```

bshyong (dismissed):
This generally looks good! I left some notes on how to get the common store working; let me know if you have any questions or issues when trying to get it to work.

Also, `AppContent` should be removed if users aren't logging in to use GIBCT

src/js/gi/components/PreviewBanner.jsx
ben-USDS-damman: Possible, but I wanted to make it explicit that I'm affecting rendering outside the visible boundaries of the component without actually touching DOM nodes outside the component.

src/js/gi/gi-entry.jsx
ben-USDS-damman: Oh duh - that's what I was missing - updating common reducers. Thank you!

crwallace:
Hey @ben-USDS-damman -- Do we need to use `commonStore` for this? Right now we are only moving apps that need login into `commonStore`.

ben-USDS-damman:
@crwallace Was under the impression we wanted to move over to `commonStore` when it was practical and I'd like to complete it now vs. defer to "later." (What are we trying to avoid by only switching the apps that require login? Some downside I'm not aware of?)

typesend:
Nope. Thanks for clarifying. We won't use commonStore.

ben-USDS-damman:
Addresses previous comments by @bshyong. I'm in the middle of completing the profile page and will be pushing that tonight.

image

image

ben-USDS-damman:
Please take special note of what I'm doing with `reselect` as it represents the pattern I'm following for the calculator, outcome graph calculations, and looks like the best way to deal with query param manipulations.

src/js/gi/actions/index.js
```diff
@@ -15,3 +34,143 @@ export function closeModal() {
...
+
+export function exitPreviewMode() {
+  return {
+    type: EXIT_PREVIEW_MODE
+  };
+}
+
+export function fetchConstants() {
+  const previewVersion = ''; // TODO
+  const url = `${api.url}/calculator/constants?preview=${previewVersion}`;
```
bshyong: Adding a note to remember to remove the temporary hack in `config.js` later on

src/js/gi/components/Checkbox.jsx
```diff
@@ -0,0 +1,102 @@
+import React from 'react';
+import _ from 'lodash';
+
+import ToolTip from '../../common/components/form-elements/ToolTip';
+
+
+/**
+ * A form checkbox with a label that can display error messages.
```
bshyong: Is it possible to use the common component for Errorable Checkbox here instead of a specific one for GIBCT?

src/js/gi/components/If.jsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+
+class If extends React.Component {
```
U-DON: Rather than create an `If` component, I think it's more idiomatic to either (1) conditionally assign JSX to variables or (2) just inline the conditions. https://facebook.github.io/react/docs/conditional-rendering.html

```
// Conditionally assigning JSX.

let conditionalSection;

if (condition) {
  conditionalSection = <div>This section appears because of condition.</div>;
}

return (
  <div>
    <span>This always appears.</span>
    {conditionalSection}
  </div>
);
```

```
// Inlining conditional sections.

function renderConditionalSection() {
  return <div>This section appears because of condition.</div>;
}

function render() {
  return (
    <div>
      <span>This always appears.</span>
      {condition && renderConditionalSection()}
    </div>
  );
}
```

From [this SO answer](http://stackoverflow.com/a/25225016): "As for the If component you asked about, one problem it has is that in its current form it will evaluate its children even if the condition is false. This can lead to errors when the body of the If only makes sense if the condition is true."

src/js/gi/components/RadioButtons.jsx
```diff
@@ -0,0 +1,155 @@
+import React from 'react';
+import _ from 'lodash';
+
+import ToolTip from '../../common/components/form-elements/ToolTip';
+import ExpandingGroup from '../../common/components/form-elements/ExpandingGroup';
+
```
bshyong: Is it possible to use the common Radio Button component?

src/js/gi/components/search/InstitutionFilterForm.jsx
```diff
@@ -0,0 +1,189 @@
...
+
+export class InstitutionFilterForm extends React.Component {
+
+  constructor(props) {
+    super(props);
+    this.renderTypeFilter = this.renderTypeFilter.bind(this);
+  }
+
+  renderTypeFilter() {
+    const type_facets = {
```
U-DON: I think this would fail the linter since it expects camelCase. Same for all the other variables in the other `render*` functions in this class.

src/js/gi/components/search/InstitutionFilterForm.jsx
```diff
@@ -0,0 +1,189 @@
+import React from 'react';
+import { connect } from 'react-redux';
+import * as actions from '../../actions';
+
+import Checkbox from '../Checkbox';
+import RadioButtons from '../RadioButtons';
+import Dropdown from '../Dropdown';
+
+export class InstitutionFilterForm extends React.Component {
+
```
U-DON: I _think_ this extra space would fail the linter; there may be a rule about not having extraneous line breaks at the beginning or end of a block. Something to look out for the other modules as well.

src/js/gi/components/search/InstitutionFilterForm.jsx
```diff
@@ -0,0 +1,189 @@
...
+            label={label('eight_keys_to_veteran_success', '8 Keys to Vet Success')}
+            onChange={this.props.handleCheckboxChange}/>
+      </div>
+    );
+  }
+
+  renderSchoolTypeFilter() {
+    const school_type_facets = {
+      'ALL': this.props.search.count,
+      ...this.props.search.facets.type_name
```
U-DON: Object key names are also subject to the linter's rule for camelCasing.

src/js/gi/components/search/SearchResult.jsx
```diff
@@ -0,0 +1,111 @@
...
+        <i className="fa fa-warning"></i>
+        Caution
+      </div>
+    );
+  }
+
+  place() {
+    const domestic = <span>{this.props.city}, {this.props.state}</span>;
+    const foreign = <span>{this.props.city}, {this.props.country}</span>;
+    return (this.props.country === 'usa' ? domestic : foreign);
```
U-DON: You can avoid extra JSX processing here, though it's not a big deal in this case because it's very simple. Just as an example, with the following, you end up only processing one JSX fragment.

```
return (
  <span>
    {this.props.city}, {this.props.country === 'usa' ? this.props.state : this.props.country}
  </span>
);
```

src/js/gi/selectors/calculator.js
```diff
@@ -0,0 +1,23 @@
+import { createSelector } from 'reselect';
+
+const getConstants = (state) => state.constants.constants;
+
+const getEligibilityDetails = (state) => {
+  const details = Object.assign({}, state.eligibility);
+  delete details.dropdowns;
```
bshyong: why is the dropdowns attribute deleted? no change needed, just wondering

bshyong (approved):

src/js/gi/components/search/SearchResult.jsx
```diff
@@ -0,0 +1,111 @@
...
+
+  render() {
+    const tuition = this.renderEstimate(this.props.estimated.tuition);
+    const housing = this.renderEstimate(this.props.estimated.housing);
+    const books = this.renderEstimate(this.props.estimated.books);
+
+    return (
+      <div className="search-result">
+        <div className="outer">
+          {this.cautionFlag.bind(this)()}
```
U-DON: Usually we just bind every method in the constructor and don't have to deal with fixing references to `this` later.

```
constructor(props) {
  super(props);
  this.cautionFlag = this.cautionFlag.bind(this);
  this.place = this.place.bind(this);
  this.renderEstimate = this.renderEstimate.bind(this);
}

File: src/js/gi/config.jsx [hunk 16]
```diff
@@ -0,0 +1,21 @@
+import environment from '../common/helpers/environment';
```
U-DON: This file doesn't need the `jsx` suffix, since it doesn't use any React, so can it just be `config.js`.

src/js/gi/selectors/calculator.js
bshyong: The use of reselect looks fine to me. Because some of the attribute calculations are fairly complicated, reselect lets us only recalculate when needed.

src/js/gi/reducers/autocomplete.js
```diff
@@ -0,0 +1,41 @@
+import { AUTOCOMPLETE_TERM_CHANGED, AUTOCOMPLETE_STARTED, AUTOCOMPLETE_FAILED, AUTOCOMPLETE_SUCCEEDED } from '../actions';
+
+const INITIAL_STATE = {
+  inProgress: false,
+  previewVersion: null,
+  search_term: '',
+  facility_code: null,
```
U-DON: Why are these two keys snake_case while the first two are camelCase? In any case, linter will fail on the snake_cased keys, so they should be changed to camelCase.

If it's because the data received from the API would need to be transformed into the proper casing, you can actually just add the header `X-Key-Inflection: 'camel'` to the requests.

src/js/gi/selectors/calculator.js
ben-USDS-damman: `details.dropdowns` holds the shown/not-shown state of the eligibility interdependent dropdowns and isn't used in the calculation. IOW the code for the visible/hidden presentation of the eligibility dropdowns is separate from the code that determines what chosen values become inputs to the calculation.

src/js/gi/components/RadioButtons.jsx
ben-USDS-damman: (The Checkbox and RadioButtons components were copied here to make that possible in a future iteration, but my needs were simpler and significantly different such that I didn't want to tackle generalizing common components right away. (Aside from making it easy to diff to what the diffs are between the common and gibct components.) I thought about using inheritance but that seemed less practical and would've introduced a dependency on the common components that I'm not sure is a good idea.)

src/js/gi/components/Checkbox.jsx
ben-USDS-damman: The Checkbox and RadioButtons components were copied here to make that possible in a future iteration, but my needs were simpler and significantly different such that I didn't want to tackle generalizing common components right away. (Aside from making it easy to diff to what the diffs are between the common and gibct components.) I thought about using inheritance but that seemed less practical and would've introduced a dependency on the common components that I'm not sure is a good idea.

src/js/gi/selectors/calculator.js
ben-USDS-damman: My approach is to try to make inputs as explicit/obvious as possible - like I did for ` getRequiredAttributes` in the `estimator.js` selector.

src/js/gi/selectors/estimator.js
```diff
@@ -0,0 +1,203 @@
...
+const getEligibilityDetails = (state) => {
+  const details = Object.assign({}, state.eligibility);
+  delete details.dropdowns;
+  return details;
+};
+
+const getRequiredAttributes = (_state, props) => {
+  const { type, bah } = props;
+  return { type, bah };
+};
```
U-DON: If the two above `get*` functions are getting used in a few different modules, could they be in `utils/helpers.js` and imported to where they're needed?

src/js/gi/components/Checkbox.jsx
ben-USDS-damman: * the concepts of required fields and error messages do not apply to GIBCT
* click handlers on the common components also seemed like overkill and I frankly understand why they were implemented the way they are, i.e. `{value: x, dirty: true}`. I'm sure there's a good reason but it felt like the reason was maybe only well-suited to more complicated forms and schema forms elsewhere on the site.

src/js/gi/selectors/estimator.js
```diff
@@ -0,0 +1,203 @@
...
+  if (activeDutyThirtyOr1607 || correspondenceOrFlightUnderOldGiBill) {
+    onlyTuitionFees = true;
+  } else {
+    onlyTuitionFees = false;
+  }
+
+  // The monthly benefit rate for non-chapter 33 benefits
+  const isOJT = its.type === 'ojt';
+  const isFlight = its.type === 'flight';
+  const n = Number(your.number_of_dependents);
```
U-DON: Here's a neat little shorthand for Number conversions:

`const n = +your.number_of_dependents`.

U-DON (commented):
Just noticed a few places that are probably failing the lint and the use of certain components. Some minor things.

src/js/gi/components/If.jsx
ben-USDS-damman: I seriously hate this `If` component, but it was a necessity when porting over legacy code.

src/js/gi/components/heading/PreviewBanner.jsx
```diff
@@ -0,0 +1,60 @@
...
+  version: React.PropTypes.object.isRequired,
+  toolURL: React.PropTypes.string.isRequired,
+};
+
+PreviewBanner.defaultProps = {
+  toolURL: 'http://www.usds.gov', // TODO
+};
+
+const mapStateToProps = (state) => state;
+const mapDispatchToProps = (dispatch) => {
```
ben-USDS-damman: Wow, nice of redux to make that work!

src/js/gi/selectors/estimator.js
ben-USDS-damman: Nice!

src/js/gi/components/Checkbox.jsx
jbalboni: For what it's worth, the `value` and `dirty` pattern is there to support showing validation errors only after a field has been interacted with (i.e. it's dirty). Definitely not needed if you don't have required or error messages.

Those components were originally built for HCA and forms like what the Rainbows team is building, but we're moving away from using these particular components now (the schema form work). So there's probably some room in the future to have a simpler set of components that fit more cleanly in apps that just have a few input fields.

typesend:
Thanks for that info Jeff!

src/js/gi/components/AccordionItem.jsx
```diff
@@ -0,0 +1,42 @@
...
+
+  toggle() {
+    this.setState({expanded: !this.state.expanded});
+  }
+
+  render() {
+    const expanded = this.state.expanded;
+    return (
+      <li>
+        <button onClick={this.toggle.bind(this)} className="usa-accordion-button" aria-expanded={expanded} aria-controls="amendment-1">
```
U-DON: Should be bound in the constructor.

src/js/gi/components/profile/AdditionalInformation.jsx
```diff
@@ -0,0 +1,166 @@
...
+          {it.accreditation_type.toUpperCase()}
+        </p>
+      );
+    };
+    const VetTuitionPolicy = () => {
+      if (!it.vet_tuition_policy_url) return null;
+      return (
+        <p>
+          <strong>Veterans tuition policy:&nbsp;</strong>
+          <a href={`http://${it.vet_tuition_policy_url}`} target="_blank">View policy</a>
```
U-DON: Does it make sense to just always use https here, or is it just not available?

src/js/gi/components/profile/AdditionalInformation.jsx
```diff
@@ -0,0 +1,166 @@
...
+              </tbody>
+            </table>
+          </div>
+        </div>
+      </div>
+    );
+  }
+}
+
+const mapStateToProps = (state, props) => state;
```
U-DON: Typically if the component only needs a part of the state, you want to pare down the parts of the store you want to translate to the component's props. In this case, you're only going to be using `profile.attributes`.

src/js/gi/components/profile/Graph.jsx
```diff
@@ -0,0 +1,121 @@
...
+      if (isNaN(percent)) { return null }
+      const width = 134 * (percent / this.props.max);
+      return <path fill={color} d={`M78 ${top}h${width}v32H78z`}/>;
+    };
+    const ValueLabel = ({y, text}) => (
+      <text fontFamily="SourceSansPro-Bold, Source Sans Pro" fontSize="16" fontWeight="bold" fill="#323A45">
+        <tspan x="220" y={y}>{text}</tspan>
+      </text>
+    );
+    const AverageMark = ({percent, text}) => {
```
U-DON: Other than `ValueBar`, these components don't directly use `Graph`'s props. They could probably be defined outside of the class, so that they don't get initialized on each render. Then for `ValueBar`, `this.props.max` could just be passed in as an arg.

src/js/gi/reducers/profile.js
```diff
@@ -1,12 +1,23 @@
...
 const INITIAL_STATE = {
   attributes: {},
   version: {},
   inProgress: false,
 };
 
+function normalizedAttributes(attributes) {
+  const name = attributes.name ? attributes.name.toUpperCase() : attributes.name;
+  const city = attributes.city ? attributes.city.toUpperCase() : attributes.city;
+  const state = attributes.state ? attributes.state.toUpperCase() : attributes.state;
```
U-DON: For these, you could just do something like:

`const name = attributes.name && attributes.name.toUpperCase()`.

U-DON (commented):
Left a few comments, but otherwise looks good. I notice similar things as my last review for things that would trip the linter. Be sure to run `npm run lint` to capture all the errors to fix before the tests pass.

src/js/gi/selectors/estimator.js
ben-USDS-damman: These are truly private functions that will be different across selectors.

src/js/gi/components/AccordionItem.jsx
ben-USDS-damman: done

src/js/gi/components/search/SearchResult.jsx
ben-USDS-damman: done

src/js/gi/components/search/SearchResult.jsx
ben-USDS-damman: done

src/js/gi/components/search/InstitutionFilterForm.jsx
ben-USDS-damman: it doesn't

src/js/gi/gi-entry.jsx
ben-USDS-damman: As stated in other comments we are not going to be using commonStore.

U-DON (approved):
LGTM. The main things that can be fixed after are moving away from `commonStore` and usage of some components.

typesend:
This code does NOT use commonStore at all. To be clear.

saneshark:
Looks like the build is passing. @ben-USDS-damman do you still need to add profile updates or are those in? Since this unauthed can we triple check and make sure it is behind a feature toggle and not publicly visible unless you know the URL before we merge?

typesend:
It's in the ignoreList AND the rev proxy is sending traffic to the old rails app still anyway - something we need to change for dev/staging if we want to look at this there.

PR #4805, opened 2017-02-06
Remove some jquery shenanigans

aub:
This is a start for department-of-veterans-affairs/vets.gov-team#649

It removes some places where we reference jQuery where it is unnecessary. Unfortunately we can't remove it entirely because the version of usds we're using depends on it explicitly and the new versions conflict with foundation. Once work is completed for department-of-veterans-affairs/vets.gov-team#549 we should be able to get rid of it entirely.

The changes here are:

* Refactor the education benefits modal to use native JS.
* Remove the components.js file, which @crwallace and I found to have only legacy code
* Remove the jquery-accessible-simple-tooltip-aria package, which was only used in components.js


content/includes/modal.html
```diff
@@ -47,29 +47,32 @@ <h5 class="va-modal-split-title">Servicemembers and Veterans applying for educat
...
+
   if (document.location.hash === '#modal'){
     open();
   }
 
-  $('.va-modal-close-button').click(close);
-
-  function open(){
-    modal.show();
-    modal.focus();
```
jbalboni: Did we lose this?

content/includes/modal.html
```diff
@@ -47,29 +47,32 @@ <h5 class="va-modal-split-title">Servicemembers and Veterans applying for educat
...
+
+  function closeModal() {
     if(window.history.pushState) {
       window.history.pushState('', '/', window.location.pathname)
     } else {
       window.location.hash = '';
     }
-    modal.hide();
+    modal.style.display = 'none';
     $('#content').focus();
```
jbalboni: Should this be `querySelectorAll`?

jbalboni (approved):
Couple comments, but LGTM.

jcmeloni-usds:
@aub Is this good to go after a rebase or should it be considered a WIP?

aub:
@jcmeloni good catch, I rebased and am merging.

PR #4807, opened 2017-02-06
adds updated date to pages automagically

crwallace:
This removes the need for editors to update the "update date" for content pages manually.

addresses department-of-veterans-affairs/vets.gov-team#986


aub:
Should the .updated.json file be in .gitignore?

package.json
```diff
@@ -113,6 +113,7 @@
     "metalsmith-permalinks": "^0.5.0",
     "metalsmith-redirect": "^2.1.0",
     "metalsmith-sitemap": "^1.0.0",
+    "metalsmith-updated": "^0.1.6",
```
aub: Can we remove the date-in-filename and redirect packages?

jcmeloni-usds (approved):
I'm cool with this after @aub has the answers he needs and jenkins/travis is green.

rroueche:
Awesome! @crwallace so in terms of a 'how do we use this' perspective :) all we need to do is make sure that the "lastupdate:" element exists in the YAML, like this:
image

and that's it?

package.json
crwallace: These were already removed.

crwallace:
@aub -- we need the `.updated.json` file because different people will be updating the same files and we need to keep track of dates when everything is updated. Otherwise every page will have an updated date of today because `.updated.json` will be generated each time if it is not found.

crwallace:
@rroueche -- You do not need to add `lastupdate:` to the YAML file. There is a `.updated.json` file that keeps track of when files are changed. You can just continue to edit files as usual. No extra work.

rroueche:
@crwallace awesome thank you!! and to triple confirm: so we don't need to include that element in the YAML, but how do we surface the back-end info on the live page? There are some pages that do not need to display the 'last updated' info. Like tool pages, or landing pages. It's really just the detailed informational pages - especially with health-related content - that need the date displayed on them.

So is there a way for us to determine - either manually or dynamically - which pages display this piece of metadata?

crwallace:
@rroueche -- To confirm, no you do not need to add the `lastupdated` flag. Right now, whether or not the updated date is being shown is dictated by the templates that @dannychapman has included the html for updated date. For example: https://github.com/department-of-veterans-affairs/vets-website/blob/8a7a4ec2783e7bc1d2f15134bbfe86a983bdc8af/content/includes/1-topic-landing.html#L12-L15. It doesn't appear that we have included tools like the healthcare application because they use different templates.

package.json
aub: Hmm, but they're there in both this branch and in master.

aub:
@crwallace how does updated.json get updated then? If it needs to be updated each time the page changes, it seems like developers must have to run something to generate a new version of it before pushing changes.

crwallace:
@aub -- It is updated as a part of the metalsmith build.

dannychapman:
@crwallace thanks for your work on this. Question: Is there a way to manually set the creation date? I realize this might not be possible technically, but it's going to be a bit disingenuous if all content has the same last updated date to start with.

ayaleloehr:
pinging @crwallace, as it looks like the last question is to you.

jcmeloni-usds:
@ayaleloehr @crwallace is this good to go after a rebase or .... ?

dannychapman:
@jcmeloni @crwallace I'd like to hold on merging this until we figure out a strategy for manually backdating/not dating posts. If we can't accommodate that functionality, the content team should probably weigh in on how best to handle dates on existing content going forward. cc @rroueche

jcmeloni-usds:
@robbiethegeek-usds Just a note that this is going to get stale soon.

rroueche:
@robbiethegeek @jcmeloni-usds were you able to find out if there's a way to manually set the 'created date' so it essentially overrides the automated posting of today's date?

and @dannychapman can you confirm that this is the only template that will display the date on the page? <div class="small-12 columns usa-content">

jcmeloni-usds:
@robbiethegeek-usds Were you picking this up after Courtney or ... ?

aub:
@jcmeloni I'm working on it.

@rroueche we could manually set the created date, though it would be tedious. What I can do more easily is cause it to not show the date at all unless the file has been changed since it was created.

I'm trying to figure out exactly how this thing works now and if we can make it meet our needs. Currently it seems like there may be other issues so that we have to consider another approach but I'll keep you posted.

aub:
The biggest problem I see here is that the file containing the dates is kept in version control and is only updated when a build is performed. This means that in order for the dates to be updated the person changing the content would have to make their changes, then do a build, then commit the changes as well as the updated date file.

There is a further problem where the production build updates the date file in a different way, since it includes all of the hashed assets, so the file will continually change as people do different builds on their local systems.

I think given these issues this is not a viable approach, but I'm struggling to come up with one that is. I can think of solutions to the latter problem but they don't address the former one. I'm going to close this PR and circle back with @rroueche to make sure I understand the requirements. @robbiethegeek-usds  and @jcmeloni let me know if you want to spend more time trying to salvage this.

PR #4808, opened 2017-02-06
Add Reducers and actions to Blue Button form

bshyong:
Making a PR for this before it gets too big. This PR adds the data shape, reducers, actions for the form. Users are able to toggle report types and select a date option.

Validations and selection a custom date range are broken out into issue https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1024

Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/987


src/js/blue-button/reducers/index.js
```diff
@@ -0,0 +1,7 @@
+import { combineReducers } from 'redux';
+
+import form from './form';
+
+export default combineReducers({
+  form,
```
austin-martinez: ✂️  trailing comma

austin-martinez (approved):
LGTM 👌

src/js/blue-button/reducers/index.js
bshyong: @austin-martinez do you have a preference for these? I usually leave them so that there is less of a chance for errors, and it's easier to add/remove lines.

austin-martinez: I prefer to not have the trailing comma, but not a strong preference. If this is the pattern you've been using for the other rx products, consistency is better than my preference 😄

PR #4811, opened 2017-02-07
Working submit and confirmation page for 1995

jbalboni:
Related to [939](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/939) and [1095](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1095)

Adds the code to transform the form into the right format for the backend. Also updates confirmation page with correct info and design changes from @goldenmeanie.

screen shot 2017-02-07 at 11 40 39 am


goldenmeanie:
The plus sign in that accordion looks smaller than usual, right?

goldenmeanie:
"Benefit claimed" might be better as "Benefit to be transferred"

jbalboni:
@goldenmeanie Yeah, not sure why the accordion styles are weird on this page. Same issue on 1990.

goldenmeanie:
Is there some sort of nested 80% thing going on?

src/js/common/schemaform/helpers.js
```diff
@@ -156,13 +158,63 @@ export function parseISODate(dateString) {
...
+ * Normal transform for schemaform data
+ */
+export function transformForSubmit(formConfig, form) {
+  const activePages = getActivePages(createFormPageList(formConfig), form);
+  const flattened = flattenFormData(activePages, form);
+  const withoutViewFields = filterViewFields(flattened);
+
+  return JSON.stringify(withoutViewFields, (key, value) => {
+    // an object with country is an address
+    if (value && typeof value.country !== 'undefined' &&
```
annekainicUSDS: Is this because country is defaulted to "USA"?

annekainicUSDS (approved):
Looks great and tested great too!

src/js/common/schemaform/helpers.js
jbalboni: Yeah, otherwise it'd be an empty object by default and we could just remove it.

PR #4814, opened 2017-02-07
Add glossary links and modal to Blue Button report generation page

bshyong:
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/933

Screenshots:

screen shot 2017-02-07 at 1 40 03 pm

screen shot 2017-02-07 at 1 40 10 pm


src/js/blue-button/components/GlossaryModal.jsx
```diff
@@ -0,0 +1,37 @@
...
+    this.handleCloseModal = this.handleCloseModal.bind(this);
+  }
+
+  handleCloseModal(event) {
+    event.preventDefault();
+    this.props.onCloseModal();
+  }
+
+  render() {
+    if (this.props.isVisible) {
```
austin-martinez: Can this be removed? This seems redundant with the visible property of the modal

src/js/blue-button/components/GlossaryModal.jsx
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+import Modal from '../../common/components/Modal';
+
+class GlossaryModal extends React.Component {
```
austin-martinez: Is this component necessary? The isVisible check seems redundant with the Modal component's internal check and use of the visible property, and the close handler could be moved up into the parent component.

bshyong:
Done. A vanilla modal does make sense here. I removed that component in favor of using the common modal that we have.

austin-martinez (approved):
LGTM :1st_place_medal:

Issue #4623, opened 2017-01-06
Build front end 1995 CH 2 : Benefit Selection


cweber:
# 2. Benefit Selection
- Select the benefit that is the best match for you **[check]**
  - _All options from 1990_
  - Add: Reserve Educational Assistance Program (REAP, Chapter 1607)
  - Add: Transfer of Entitlement Program (TOE)

mollyblake:
@cweber 
Transfer of Entitlement Program (TOE) 
* please add the acronym

goldenmeanie:
Thanks @mollyblake!

annekainicUSDS:
Notes from meeting with Shay:
 
- Switch order to TOE, Post Vietnam, then REAP

PR #4816, opened 2017-02-07
Changes order of benefits

cvalarida:
References #4623

Just changes the order of benefits as mentioned [here](https://github.com/department-of-veterans-affairs/vets-website/issues/4623#issuecomment-278017061).


jbalboni:
@cvalarida The way we need to make this change is going to change after I merge a PR of mine soon, so I wanted to hold off on this one. This won't cause any harm if we merge it, though.

Sorry for the confusion!

jbalboni:
@cvalarida You can merge this or close it, the order was changed with a schema update I made yesterday.

PR #4829, opened 2017-02-08
Fix errors on 1995 service periods page

jbalboni:
There are some issues with the code I committed to master earlier today on the service periods page. This fixes them so that data is correctly saved in the form and validation works again.

### DIFF
--- a/src/js/common/schemaform/ArrayField.jsx
+++ b/src/js/common/schemaform/ArrayField.jsx
@@ -44,13 +44,7 @@ export default class ArrayField extends React.Component {
   }
 
   onItemChange(indexToChange, value) {
-    const newItems = this.props.formData.map(
-      (current, index) => {
-        return index === indexToChange
-          ? value
-          : current;
-      }
-    );
+    const newItems = _.set(indexToChange, value, this.props.formData || []);
     this.props.onChange(newItems);
   }
 
--- a/src/js/common/schemaform/FormPage.jsx
+++ b/src/js/common/schemaform/FormPage.jsx
@@ -159,7 +159,7 @@ class FormPage extends React.Component {
   validate(formData, errors) {
     const { schema, uiSchema } = this.props.form[this.props.route.pageConfig.pageKey];
     if (uiSchema) {
-      uiSchemaValidate(errors, uiSchema, schema, formData);
+      uiSchemaValidate(errors, uiSchema, schema, schema.definitions, formData);
     }
     return errors;
   }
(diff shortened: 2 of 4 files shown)


src/js/common/schemaform/ArrayField.jsx
```diff
@@ -44,13 +44,7 @@ export default class ArrayField extends React.Component {
...
 
   onItemChange(indexToChange, value) {
-    const newItems = this.props.formData.map(
-      (current, index) => {
-        return index === indexToChange
-          ? value
-          : current;
-      }
-    );
+    const newItems = _.set(indexToChange, value, this.props.formData || []);
```
jbalboni: We're defaulting the array, so sometimes it's undefined here. `set` handles this better than the old code.

src/js/common/schemaform/FormPage.jsx
```diff
@@ -159,7 +159,7 @@ class FormPage extends React.Component {
   validate(formData, errors) {
     const { schema, uiSchema } = this.props.form[this.props.route.pageConfig.pageKey];
     if (uiSchema) {
-      uiSchemaValidate(errors, uiSchema, schema, formData);
+      uiSchemaValidate(errors, uiSchema, schema, schema.definitions, formData);
```
jbalboni: We're passing the schema to validation functions now and we need to fill in any fields that use JSON Schema's common definitions.

src/js/common/schemaform/validation.js
```diff
@@ -95,9 +98,9 @@ export function uiSchemaValidate(errors, uiSchema, schema, formData, formContext
...
           };
         }
-        uiSchemaValidate(errors, uiSchema.items, schema.items, formData, formContext, newPath);
+        uiSchemaValidate(errors, uiSchema.items, schemaWithDefinitions.items, definitions, formData, formContext, newPath);
       });
-    } else {
+    } else if (!uiSchema.items) {
       Object.keys(uiSchema)
         .filter(prop => !prop.startsWith('ui:'))
         .forEach((item) => {
```
cvalarida: When submitting the application, I get the following error at `_.get(path, errors)[item] = {`:

```
Uncaught TypeError: Cannot set property 'veteranFullName' of undefined
```

`path` is an empty string and `errors` is an empty object, so it fails. Not sure what the root problem is, though.

jbalboni: If you mean submitting on the review page, that's something I expect to be broken right now. Submit is being addressed in #4811.

cvalarida: Okay, then it looks golden to me. That was the only error I found.

PR #4832, opened 2017-02-09
[WIP] HCA progress bar

cvalarida:
Removes the subway map and uses the progress bar per department-of-veterans-affairs/vets.gov-team#907

I'm not too keen on the how the title and subtitle look, however:

---

image

---

Also, I don't know what page-specific headers need to be there and which don't (now that we have the NavHeader component to display the section / chapter name up top)


patrickvinograd:
"Apply online for health care" seems grammatically suspect to me.

src/js/hca/components/HealthCareApp.jsx
```diff
@@ -329,19 +332,44 @@ class HealthCareApp extends React.Component {
...
             <Nav
                 data={this.props.data}
                 pages={this.props.uiState.sections}
                 chapters={chapters}
                 currentUrl={this.props.location.pathname}/>
           </div>
+          */}
           <div className="medium-8 columns">
-            <div className="progress-box">
+            <FormTitle title="Apply online for health care with the 10-10ez" subTitle="OMB No. 2900-0091"/>
```
jbalboni: Is this block exactly the same as on the 1990 form? We could pull it into a separate component if so.

src/sass/_schemaform.scss
```diff
@@ -0,0 +1,61 @@
+// Style for schemaform
```
jbalboni: Thanks for moving this. Was going to make a tech debt issue for it.

I think it makes sense to move it into modules and follow the naming convention for those files.

jbalboni (commented):
Looks good overall, put a few comments. There may be e2e tests to change, too.

jbalboni:
@cvalarida I think you can default to removing all the page headers, same as on the 1990 form. @annekainicUSDS or @goldenmeanie can go through and see there are any we want to keep.

Also, @goldenmeanie would be the one to answer any design questions on this form, too.

jbalboni:
@cvalarida I think you can completely remove the test/hca/05-nav-completion.e2e.spec.js test. That is testing that the subway map renders correctly at each step.

cvalarida:
@jbalboni Would it make more sense to remove the test completely or just replace the check for `.section-complete` on `.form-process li.step:nth-child(X)` with checking `progress-segment-complete` on `.progress-bar-segmented div.progress-segment`?

jbalboni:
@cvalarida I think it's overkill to have a whole end to end test just for the progress bar. You could add those checks to one of the other e2e tests, though.

cvalarida:
@goldenmeanie When you get a chance, let me know which headers to put back in; I've removed them all for now.

Should I merge this and open a new PR for adding the headers back in? I don't want confusion in production by doing that, so I'll defer.

goldenmeanie:
1. The h4s in the form are visually too strong. Current style h5 is a better match.
2. These URLs need their (or some) headlines
    - [ ] veteran-address 
    - [ ] contact-information
    - [ ] military-service "Service periods"
    - [ ] additional-information
    - [ ] basic-information "Current compensation"
    - [ ] financial-disclosure
    - [ ] spouse-information
    - [ ] child-information
    - [ ] annual-income
    - [ ] deductible-expenses
    - [ ] medicare "Medicaid or Medicare coverage"
    - [ ] general "Other coverage"
    - [ ] va-facility "VA facility"

PR #4836, opened 2017-02-09
Removing the job seeker folder and index page

rroueche:
It appears in on these two pages and links to a page we have since retired:
https://www.vets.gov/employment/job-seekers/family-members/
https://www.vets.gov/employment/job-seekers/start/


rroueche:
@bethpotts and @maryannbrody I started to remove Job Seeker from the breadcrumb path, however I realized that there are lots of other pages that live underneath Job Seeker, and thus all of those pages would die or would need to have URL redirects set up for them if I removed Job Seeker from the structure. There was Slack conversation on this (whether all of those pages are supposed to go away or not) but the conclusion was to put this on hold until further Careers discussions are underway.

@maryannbrody and @amybehrens let me know if you feel differently - and if we think that all of the pages under Job Seeker should die and just redirect to Careers landing page, and I'll go ahead and move fwd with this.

rroueche:
closing this and deleting the branch this since we don't have a decision on whether to remove those child pages or not.

PR #4837, opened 2017-02-09
Removes extra api calls in login compoment

crwallace:
This PR leverages `commonStore` to pass down user and profile props to the `RequiredLoginComponent`.

### DIFF
--- a/src/js/common/components/RequiredLoginView.jsx
+++ b/src/js/common/components/RequiredLoginView.jsx
@@ -14,13 +14,7 @@ class RequiredLoginView extends React.Component {
     this.handleSignup = this.handleSignup.bind(this);
     this.setUserLevel = this.setUserLevel.bind(this);
     this.setInitialLevel = this.setInitialLevel.bind(this);
-
     this.handleVerify = handleVerify;
-    this.state = {
-      accountType: 0,
-      services: null,
-      loading: true
-    };
   }
 
   componentDidMount() {
@@ -38,19 +32,9 @@ class RequiredLoginView extends React.Component {
       this.setState({ loginUrl: json.authenticate_via_get });
     });
 
-    addEvent(window, 'message', (evt) => {
-      this.setInitialLevel(evt);
-    });
-
     setTimeout(() => {
       this.setState({ loading: false });
     }, 2000);
-  }
-
-  setInitialLevel(event) {
-    if (event.data === sessionStorage.userToken) {
-      this.setUserLevel();
-    }
   }
 
   setUserLevel() {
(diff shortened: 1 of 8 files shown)


ayaleloehr:
Pinging @alexose-gov as this feels like something we want to get in.

jcmeloni-usds:
ping @alexose @robbiethegeek-usds

src/js/common/components/RequiredLoginView.jsx
```diff
@@ -141,17 +99,17 @@ class RequiredLoginView extends React.Component {
       view = <LoadingIndicator setFocus message="Loading your information"/>;
     } else {
       if (this.props.authRequired === 1) {
-        if (this.state.accountType >= 1) {
+        if (this.props.userProfile.accountType >= 1) {
```
omgitsbillryan: Unrelated to your PR, can we have enum constants for readability instead of magic numbers?  For example, I dont know what an `authRequired == 1` or `accountType >=1` means.

omgitsbillryan (commented):
LGTM but I would recommend that a more seasoned react developer review as well.

robbiethegeek-usds:
The code is clean and LGTM would like @jbalboni or another front end engineer to review and sign off.

src/js/common/components/RequiredLoginView.jsx
crwallace: `auhtRequired` is the LOA required to access that app `accountType` is the LOA that that user has. So an example is that a user that has an LOA of 3 can access any app that only requires LOA of 1 because they are greater than 1.

alexose-gov:
Looks good here.  Always love negative SLOCs :)

crwallace:
Would love for @jbalboni to take a look since I had to fix conflicts with claimsApp

src/js/common/helpers/login-helpers.js
```diff
@@ -46,6 +46,8 @@ export function getUserData() {
       commonStore.dispatch(updateProfileField('userFullName.last', userData.last_name));
       commonStore.dispatch(updateProfileField('gender', userData.gender));
       commonStore.dispatch(updateProfileField('dob', userData.birth_date));
+      commonStore.dispatch(updateProfileField('status', json.data.attributes.va_profile.status));
```
jbalboni: Probably not something to do right now, but seems like this should be refactored to use one action. This is probably triggering a bunch of unnecessary React renders.

src/js/common/components/RequiredLoginView.jsx
jbalboni: Yeah, it's a little weird because it feels like you'd be making a constant that is renaming what is just a level number, but using something like `LOA_1` might make it clearer that `authRequired` and `accountType` are related information.

jbalboni (approved):
LGTM

Issue #4544, opened 2017-01-03
build front end 1995 CH 4 - School Selection


leannammiller:
# 4.1 NEW school, university, or training facility
_^ Title needs copy help_

In what type of education or training do you plan to enroll?

 - Type of education or training **[select]** 
 - Name of school, university, or training facility **[text]** 
 - Address **[standard address component]**
 - Education or career goal (for example, “Get a bachelor’s degree in criminal justice” or “Get an HVAC technician certificate” or “Become a police officer.”) **[textarea]**
 - Are you getting, or do you expect to get any money (including, but not limited to, federal tuition assistance) from the Armed Forces or public health services for any part of your coursework or training? **[check]** _ONLY IF on active duty_

# 4.2 OLD school, university, or training facility
_^ Title needs copy help_

 - Name of school, university, or training facility **[text]**
 - Address **[standard address component]**
 - When did you stop training? **[date]**
Why did you stop training? **[text]**

akainic:
@cweber @goldenmeanie is the school information for NEW and OLD conditional based on anything? Are all these questions on the same page? I think a little clarity might be helpful here on exactly how this page should look (at least I'm confused :)).

cweber:
I'm running through these and providing a bit more context for fields, as well as page breaks. Once I'm done with that, we can do a pass on what's conditional as well as what's required.

annekainicUSDS:
Notes from meeting with Shay:

1. NEW school
- Can we make it so that the drop down options for "Type of education or training" reflect what is relevant based on benefit selected? (separate ticket for this already exists for 1990)
- "Type of education or training" and "name of school" should be required
- Address isn't needed for national test, tuition and cert, or top-up (make this conditional on answer to first question, selecting first 4 options would make it required) (but still not required)

2. OLD school
- Include examples for "Why did you stop taking classes" question (Shay to provide)

jbalboni:
Pre-requisite for these changes is [#870](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/870), which I'm working on.

goldenmeanie:
We should switch the order of "Type of training" and "Name of institution"

annekainicUSDS:
@goldenmeanie Is this a new requirement, to switch the order? Perhaps we should create a separate ticket for that so it can be tracked.

annekainicUSDS:
1. "Type of education or training" and "name of school" are already required (done)

goldenmeanie:
[...]

PR #4838, opened 2017-02-09
Add conditional and hidden field support to schema form

jbalboni:
Related to [870](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/870) and #4544 

This adds the ability to do conditionally displayed fields on the schemaform. This is done by adding two options to uiSchema: `updateSchema` and `hideIf`. Those are expected to be functions and both get the current page data and overall form data. The former can return a new schema (with updated `enum` values, for example) and the latter can return true or false.

Examples of these can be seen on the new school selection page.

screen shot 2017-02-10 at 12 42 17 pm
screen shot 2017-02-10 at 12 42 06 pm


cvalarida (approved):

src/js/common/schemaform/reducers/index.js
```diff
@@ -31,13 +41,43 @@ export default function createSchemaFormReducer(formConfig) {
...
     switch (action.type) {
       case SET_DATA: {
+        // on each data change, we need to do the following steps
+        // Recalculate any required fields, based on the new data
+        let schema = updateRequiredFields(state[action.page].schema, state[action.page].uiSchema, action.data, state);
+        // Update the schema with any fields that are now hidden because of the data change
+        schema = setHiddenFields(schema, state[action.page].uiSchema, action.data, state);
+        // Update the schema with any general updates based on the new data
+        schema = updateSchemaFromUiSchema(schema, state[action.page].uiSchema, action.data, state);
+        // Remove any data that's how hidden in the schema
```
annekainicUSDS: Minor type: I think you meant `now` instead of `how`

annekainicUSDS (commented):
Testing this locally and I can't seem to get the top-up option to ever appear, assuming this works for you in testing locally?

jbalboni:
@annekainicUSDS It should be working now.

jbalboni:
@annekainicUSDS I removed the conditional logic for the top-up option, so it always shows up now.

annekainicUSDS (approved):
LGTM!

PR #4839, opened 2017-02-09
Set and validate date range inputs

austin-martinez:
Related to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1024

react-datepicker uses moment objects, so leaving those date properties as moment objects for the time being, could convert back and forth if we want to have them in a separate format in the store

The common validations expect date fields objects but we are using the datepicker and moment objects


bshyong (approved):
LGTM! Moment objects are good. If a user selects 3mo, 6mo, etc. we probably can use moment as well to manipulate dates

PR #4841, opened 2017-02-09
Require confirmation to exit 1995 form

cvalarida:
References department-of-veterans-affairs/vets.gov-team#1038

There are no tests for it yet, since my local e2e testing situation is sadly broken.

Also, since the confirmation page is broken right now, I couldn't test navigating away from it, but it should be fine.


jbalboni (approved):
LGTM. There's another item for e2e tests. They require some set up since we haven't written any for the 1995 yet.

PR #4842, opened 2017-02-10
Update children question on dependents page of 1995

jbalboni:
Related to [1091](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1091)

screen shot 2017-02-10 at 11 59 17 am
screen shot 2017-02-10 at 10 36 27 am

I moved these questions into a common definition in the edu-benefits folder because it's common to edu forms, though we only use this code on 1995 right now.

Most of the code changes are to show a description in the FieldTemplate. We had this code for object fields in ObjectField, but not on regular string/boolean fields.

### DIFF
--- a/src/js/edu-benefits/definitions/serviceBefore1977.jsx
+++ b/src/js/edu-benefits/definitions/serviceBefore1977.jsx
@@ -1,0 +1,26 @@
+import React from 'react';
+import commonDefinitions from 'vets-json-schema/dist/definitions.json';
+
+export const schema = commonDefinitions.serviceBefore1977;
+
+export const uiSchema = {
+  married: {
+    'ui:title': 'Are you currently married?',
+    'ui:widget': 'yesNo'
+  },
+  haveDependents: {
+    'ui:title': 'Do you have any children who fall into these categories?',
+    'ui:description': () => (
+      <ul className="edu-benefits-dependents-desc">
+        <li>Under age 18</li>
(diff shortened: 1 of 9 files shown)


src/js/edu-benefits/definitions/serviceBefore1977.jsx
```diff
@@ -0,0 +1,26 @@
...
+export const schema = commonDefinitions.serviceBefore1977;
+
+export const uiSchema = {
+  married: {
+    'ui:title': 'Are you currently married?',
+    'ui:widget': 'yesNo'
+  },
+  haveDependents: {
+    'ui:title': 'Do you have any children who fall into these categories?',
+    'ui:description': () => (
```
annekainicUSDS: Just curious why you set this to be a function instead of just setting it to a string of the markup itself?

annekainicUSDS (approved):
Looks good, though generally wondering why `ui:description` has to be a function and not a string?

src/js/edu-benefits/definitions/serviceBefore1977.jsx
jbalboni: If it's just a string of the markup, React will escape it and you'll see text with the markup in it (e.g. `<div>description text</div>`. I just pushed a change to allow regular JSX elements, which seems more consistent with what works in other parts of the config. Before I was allowing either a string or a React component (which is a function).

PR #4846, opened 2017-02-10
Add error messages for secure messaging

bshyong:
Adds error messages when user is no longer part of triage teams with existing messages

Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1004

Screenshot:

screen shot 2017-02-10 at 1 14 47 pm

### DIFF
--- a/src/js/messaging/containers/Thread.jsx
+++ b/src/js/messaging/containers/Thread.jsx
@@ -346,7 +346,7 @@ export class Thread extends React.Component {
     }
 
     // check if any senderIds are valid recipients
-    const disabled = intersection(
+    const disabled = !loading.recipients && intersection(
       recipients.map(e => e.value),
       threadMessages.map(t => t.senderId).concat(
         [message.senderId, Number(draft.recipient.value)]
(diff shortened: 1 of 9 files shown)


src/js/messaging/containers/Thread.jsx
```diff
@@ -328,9 +345,17 @@ export class Thread extends React.Component {
       return <p>Sorry, this message does not exist.</p>;
     }
 
+    // check if any senderIds are valid recipients
+    const disabled = intersection(
+      recipients.map(e => e.value),
```
austin-martinez: Does this need to check that the recipients are loaded first? I see a check https://github.com/department-of-veterans-affairs/vets-website/pull/4846/files#diff-a5c007b1e7ed1bbf23e95311462babb4R321 but also requires the isNewMessage to be true, so it does seem possible to get to this statement if !isNewMessage && loading.recipients

austin-martinez (approved):
LGTM, one question on the loading state for recipients

src/js/messaging/containers/Thread.jsx
bshyong: Good catch. I guess it is theoretically possible. The loading now happens in the parent container `Main` so it is unlikely that somebody will be able to navigate into a message thread before the recipients have loaded, but I will add a check just in case.

PR #4847, opened 2017-02-11
1995 intro edit

cvalarida:
Reference department-of-veterans-affairs/vets.gov-team#1096

image

### DIFF
--- a/src/js/edu-benefits/1995/components/IntroductionPage.jsx
+++ b/src/js/edu-benefits/1995/components/IntroductionPage.jsx
@@ -36,7 +36,7 @@ class IntroductionPage extends React.Component {
               </ul>
               <div className="usa-alert usa-alert-info">
                 <div className="usa-alert-body">
-                  <span><strong>You won't be able to save your work or come back to finish.</strong> So before you start, it's a good idea to gather information about your military and education history, and the school you want to attend.</span>
+                  <span><strong>You won’t be able to save your work or come back to finish.</strong> So before you start, it’s a good idea to gather information about your military and education history, and the school you want to attend.</span>
                 </div>
               </div>
               <p><a href="http://www.va.gov/ogc/apps/accreditation/index.asp">An accredited representative</a> with a Veterans Service Organization (VSO) can help you pick the right program.</p>
--- a/src/js/edu-benefits/1995/components/IntroductionPage.jsx
+++ b/src/js/edu-benefits/1995/components/IntroductionPage.jsx
@@ -10,13 +10,6 @@ class IntroductionPage extends React.Component {
   goForward = () => {
     this.props.router.push(this.props.route.pageList[1].path);
   }
-  goBack = () => {
-    // Wasn't able to figure out how to get react-router to go to
-    //  /education/apply-for-education-benefits/ without having it prepend the
-    //  pathname with /education/apply-for-education-benefits/application/
-    // this.props.router.replace('/education/apply-for-education-benefits/');
-    window.location = "/education/apply-for-education-benefits/";
-  }
   render() {
     return (
       <div className="schemaform-intro">
@@ -67,11 +60,9 @@ class IntroductionPage extends React.Component {
         </div>
         <div className="row progress-box progress-box-schemaform form-progress-buttons schemaform-buttons">
           <div className="small-6 medium-5 columns">
-            <ProgressButton
-              onButtonClick={this.goBack}
-              buttonText="Back"
-              buttonClass="usa-button-outline"
-              beforeText="«"/>
+            <a href="/education/apply-for-education-benefits/">
+              <button className="usa-button-outline">« Back</button>
+            </a>
           </div>
           <div className="small-6 medium-5 end columns">
             <ProgressButton
--- a/src/js/edu-benefits/1995/components/IntroductionPage.jsx
+++ b/src/js/edu-benefits/1995/components/IntroductionPage.jsx
@@ -14,7 +14,7 @@ class IntroductionPage extends React.Component {
     return (
       <div className="schemaform-intro">
         <h1>Manage Your Education Benefits</h1>
-        <p>This application is equivalent to Forms 22-1995 or 22-5495.</p>
+        <p>This application is equivalent to Form 22-1995 (Request for Change of Program or Place of Training) and Form 22-5495 (Dependents' Request for Change of Program or Place of Training).</p>
         <div className="process schemaform-process">
           <ol>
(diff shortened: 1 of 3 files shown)


src/js/edu-benefits/1995/components/IntroductionPage.jsx
```diff
@@ -4,37 +4,79 @@ import { focusElement } from '../../../common/utils/helpers';
...
+            </li>
+            <li className="step four last">
+              <div><h5>Decision</h5></div>
+              <ul><li>We usually process claims within 14 days.</li></ul>
+              <ul><li>You'll get a COE or Award Letter in the mail if your application was approved. Bring this to the VA certifying official at your school.</li></ul>
+            </li>
+          </ol>
         </div>
-        <div className="row form-progress-buttons">
+        <div className="row progress-box progress-box-schemaform form-progress-buttons schemaform-buttons">
```
cvalarida: This seems a little ridiculous, but it was the easiest way to get the buttons to be `width: 100%` that I could see without adding a new class for it. Open to suggestions.

src/js/edu-benefits/1995/components/IntroductionPage.jsx
```diff
@@ -4,37 +4,79 @@ import { focusElement } from '../../../common/utils/helpers';
...
-  goForward() {
+  goForward = () => {
     this.props.router.push(this.props.route.pageList[1].path);
   }
+  goBack = () => {
+    // Wasn't able to figure out how to get react-router to go to
+    //  /education/apply-for-education-benefits/ without having it prepend the
+    //  pathname with /education/apply-for-education-benefits/application/
+    // this.props.router.replace('/education/apply-for-education-benefits/');
+    window.location = "/education/apply-for-education-benefits/";
```
cvalarida: If anybody can figure out how to get the router to stop prepending the path with `/education/apply-for-education-benefits/application/`, I'd rather use the router.

src/js/edu-benefits/1995/components/IntroductionPage.jsx
```diff
@@ -4,37 +4,79 @@ import { focusElement } from '../../../common/utils/helpers';
...
 class IntroductionPage extends React.Component {
-  constructor() {
-    super();
-    this.goForward = this.goForward.bind(this);
-  }
   componentDidMount() {
     focusElement('.va-nav-breadcrumbs-list');
   }
-  goForward() {
+  goForward = () => {
```
cvalarida: We can do this rather than `this.goForward = this.goForward.bind(this);` in the `super()` to save space. Huzzah for ES6!

If the other way was used for a reason, just let me know.

src/js/edu-benefits/1995/components/IntroductionPage.jsx
jbalboni: There's some styling that assumes a progress-box class. I'm not sure why it exists, but yeah, this is probably easier for now. We should see if there's an easy way to re-write the rules that use progress-box.

src/js/edu-benefits/1995/components/IntroductionPage.jsx
jbalboni: The router won't work for this, because the page we're going to is outside of our edu-benefits app.

This should work, but If possible, we should use a regular link for this back button. Something like:

```
<a href="/education/apply-for-education-benefits/application/">
  <button className="usa-button-outline">« Back</button>
</a>
```

src/sass/edu-benefits.scss
```diff
@@ -311,9 +311,17 @@ form.rjsf {
 .schemaform-back-buttons {
   margin-bottom: 7rem !important;
 }
+
 .schemaform-intro {
   padding-bottom: 1rem;
   margin-bottom: 7rem;
+
+  // To be moved into modules/_m-schemaform.scss when it's merged
+  .process {
```
jbalboni: Why is this necessary?

src/js/edu-benefits/1995/components/IntroductionPage.jsx
```diff
@@ -4,37 +4,79 @@ import { focusElement } from '../../../common/utils/helpers';
...
+                <li>Social Security number (required)</li>
+                <li>Military history (required)</li>
+                <li>Basic information about the school or training facility where you want to attend (required)</li>
+                <li>Bank account direct deposit information</li>
+                <li>Education history</li>
+              </ul>
               <div className="usa-alert usa-alert-info">
                 <div className="usa-alert-body">
-                  <span><strong>You won’t be able to save your work or come back to finish</strong>. So before you start, it’s a good idea to gather information about your military and education history, and the school you want to attend.</span>
+                  <span><strong>You won't be able to save your work or come back to finish.</strong> So before you start, it's a good idea to gather information about your military and education history, and the school you want to attend.</span>
```
jbalboni: Could switch these quote marks back to the characters used in the old page.

src/sass/edu-benefits.scss
cvalarida: Without shifting the `ol` over to the left again, it created an uncomfortably large space not present in the screenshot:

---
image

---
Alternatively, I could have moved the title, text under the title, and buttons over to the right to close the gap, and that would have matched the screenshot referenced, but I didn't think that was intended since it would unalign the content with the banner, breadcrumbs, et al.

src/js/edu-benefits/1995/components/IntroductionPage.jsx
cvalarida: Sure thing. I didn't want to do that initially because it'll have a slightly different hover behavior (the link location will appear at the bottom of the page), but I can certainly make that happen.

src/sass/edu-benefits.scss
jbalboni: Gotcha. Can we add a separate class for this, instead of nesting the styles underneath `.schemaform-intro`? `schemaform-intro-process` or something.

I want us to avoid nesting as much as possible. It leads to things like the issue with progress-box above.

jbalboni (approved):
LGTM. Should get Alex or Molly to ok it.

mollyblake:
A few changes:  
[1] This application is equivalent to Form 22-1995 (Request for Change of Program or Place of Training) and Form 22-5495 (Dependents' Request for Change of Program or Place of Training). 

[2] See what benefits you'll get at the school you want to attend. `[Use` the GI Bill Benefits Tool]/gi-bill-comparison-tool/).

cvalarida:
With the changes @mollyblake mentioned:

image

PR #4852, opened 2017-02-13
Make 1995 benefits selection "Learn more" links clickable

cvalarida:
References department-of-veterans-affairs/vets.gov-team#1127

If I'm honest, I'm not sure why exactly this works, but I was experimenting with z-index and found I didn't need it when `position: relative`.


jbalboni (approved):
That is super weird and I have no idea why that works. But it does appear to! LGTM.

PR #4854, opened 2017-02-13
Address 508 issues for Secure messaging

bshyong:
- Changes header titles from h2 to h3
- adds `aria-live` to alerts for better screen reader support

Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1056


src/js/common/components/AlertBox.jsx
```diff
@@ -33,7 +33,7 @@ class AlertBox extends React.Component {
 
   render() {
     if (!this.props.isVisible) {
-      return <div/>;
+      return <div aria-live="assertive"/>;
```
austin-martinez: Does this need `aria-live` if it's empty?

austin-martinez (approved):
LGTM once tests pass 💯  Minor question about aria-live on an empty div

src/js/common/components/AlertBox.jsx
bshyong: yep I added this intentionally so that screen readers can expect that the content inside the div is expected to update/change. When our alert appears, the outer div should be in the same place so I expect most screen readers to recognize it as the same div

PR #4855, opened 2017-02-14
Adds a yes / no question for service periods

cvalarida:
References department-of-veterans-affairs/vets.gov-team#1160

Selecting nothing doesn't expand:

---
image

---
Neither does "No":

---
image

---
With "Yes" selected, we get the old view:

---
image

---
Including multiple entries:

---
image

---
As far as I know, only the review page needs to be addressed.


src/js/edu-benefits/1995/config/form.js
```diff
@@ -138,13 +138,21 @@ const formConfig = {
           initialData: {
           },
           uiSchema: {
+            'view:newService': {
+              'ui:title': 'Do you have any new military service to record?',
+              'ui:options': {
+                hideOnReviewIfFalse: true,
```
jbalboni: I think you can remove this. I think it looks weird on the review page if the question is sometimes hidden.

jbalboni (approved):
LGTM, made one comment. I also think we might want to remove the description text for toursOfDuty, but @goldenmeanie can weigh in on that.

goldenmeanie:
I think you can remove the copy in the nested section. I would like @mollyblake to weigh in on the yes/no question.

mollyblake:
@goldenmeanie I think we should phrase like this: 

Do you have any new periods of service to record since you last applied for education benefits?

goldenmeanie:
LGTM @mollyblake

jbalboni:
@cvalarida Feel free to merge this.

PR #4859, opened 2017-02-14
Allow dashes on SSN for 1995 form

jbalboni:
Closes [1140](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1140)

This adds an SSN widget to allow dashes (and spaces) from the user, but remove them before saving the data to the form.

This also includes some changes to the review page code, because there isn't currently a way to tell the review page what widget to use if a custom widget is being used for the regular pages.

screen shot 2017-02-14 at 11 35 31 am
screen shot 2017-02-14 at 11 35 22 am


src/js/common/schemaform/review/StringField.jsx
```diff
@@ -0,0 +1,33 @@
...
+  getUiOptions,
+  getWidget,
+  optionsList,
+  defaultFieldValue
+} from 'react-jsonschema-form/lib/utils';
+
+/*
+ * This is a minimal string field implementation for the
+ * review page, so we can define pass custom review widgets
+ * in the config
```
cvalarida: > so we can define pass custom review widgets

Perhaps you mean
> so we can pass custom review widgets

Or
> so we can define and pass

It's fiddly, I know, but while I'm here I thought I'd ask.

src/js/common/schemaform/definitions/ssn.js
```diff
@@ -1,9 +1,13 @@
 import { validateSSN } from '../validation';
 import commonDefinitions from 'vets-json-schema/dist/definitions.json';
+import SSNWidget from '../widgets/SSNWidget';
+import { TextWidget } from '../widgets';
 
 export const schema = commonDefinitions.ssn;
 
 export const uiSchema = {
+  'ui:widget': SSNWidget,
+  'ui:reviewWidget': TextWidget,
```
cvalarida: Is `TextWidget` right? I'm getting the following on review:

---
image

---
The SSN is editable prematurely.

Also, I did enter dashes that got stripped away. I'm sure that was the intention, but what if we could display it here like `XXX-XX-XXXX` even if they didn't enter it that way, then send it to the back end numbers-only? That's probably more a question for @goldenmeanie, eh?

src/js/common/schemaform/widgets/SSNWidget.jsx
```diff
@@ -0,0 +1,22 @@
...
+    this.state = { val: props.value };
+  }
+  handleChange = (val) => {
+    const strippedSSN = val.replace(/[\- ]/g, '');
+    this.setState({ val }, () => {
+      this.props.onChange(strippedSSN);
+    });
+  }
+  render() {
+    return <TextWidget type="email" {...this.props} value={this.state.val} onChange={this.handleChange}/>;
```
cvalarida: Is the SSN input `type="email"` for any particular reason?

cvalarida (changes requested):
I found only one real bug but had a couple other questions.

src/js/common/schemaform/definitions/ssn.js
jbalboni: Right widget, wrong file.

That formatting is something we've wanted to do before, so might as well add it.

src/js/common/schemaform/widgets/SSNWidget.jsx
jbalboni: Because I copied and pasted this widget from EmailWidget. That's not a good reason, but it is a reason. :)

jbalboni:
Updated screenshot:
screen shot 2017-02-15 at 9 52 34 pm

src/js/common/schemaform/widgets/SSNWidget.jsx
cvalarida: Hehe, true story!

cvalarida (approved):

cvalarida:
Looks good!

PR #4860, opened 2017-02-14
Exit with failure if node version incompatible

jkassemi:
Resolves issues with CI and review instance deployments which ran an older version of node (4.4.4). These would not exit with a failure, which resulted in builds not containing any output. This would result in issues in with subsequent processes that required the output to exist.


robbiethegeek:
can you add the status code to npm check also.

robbiethegeek-usds:
can you also add the status code to the npm check

jkassemi:
@robbiethegeek-usds - 👍  ready for review

PR #4865, opened 2017-02-14
Submit form to API

austin-martinez:
Related to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1103


src/js/blue-button/utils/helpers.js
```diff
@@ -0,0 +1,34 @@
+import environment from '../../common/helpers/environment';
```
austin-martinez: This was copy pasted from rx, but probably worth consolidating this into a common helper at some point (also present in messaging and disability benefits)

bshyong (approved):
LGTM! I can continue on to do the download page

PR #4867, opened 2017-02-14
Fix scrolling on 1995 Review page

jbalboni:
Closes [1177](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1177)

This fixes the scrolling issue on the review page by doing some refactoring I wanted to do.

Previously, `FormPage` was tasked with wrapping the rjsf `Form` component, handling moving between pages, and displaying slightly different views of the form depending on if you were on a regular page, the review page, or a row of an array field on the review page. 

Now, `SchemaForm` wraps `Form` with our basic form pattern, `FormPage` just handles the logic for moving between pages, focus handling, and displaying the form on a regular page. The review page components now use `SchemaForm`, since they don't need the logic to move between pages and the other scroll/focus handling.

The actual scrolling bug for this issue was because the old `FormPage` always wanted to scroll to the top of the page, which is what we want on a regular page, but not on the review page. Now the review page is no longer using `FormPage` and we don't have that issue.


src/js/common/schemaform/FormPage.jsx
```diff
@@ -56,57 +24,29 @@ function focusForm() {
...
 
-  onError() {
-    const formContext = _.set('submitted', true, this.state.formContext);
-    this.setState({ formContext });
-    scrollToFirstError();
-  }
-
   onSubmit() {
-    if (this.props.reviewPage) {
-      this.props.onSubmit();
```
cvalarida: Is the `onSubmit` received by props used anymore? If not, it can probably be taken out of the `propTypes` (line 130).

src/js/common/schemaform/SchemaForm.jsx
```diff
@@ -0,0 +1,125 @@
...
+};
+
+/*
+ * Each page uses this component and passes in config. This is where most of the page level
+ * form logic should live.
+ */
+class SchemaForm extends React.Component {
+  constructor(props) {
+    super(props);
+    this.validate = this.validate.bind(this);
```
cvalarida: Is `.bind(this)` preferable to ES6 arrow functions?

jbalboni: Do you mean the auto binding you were using the other day for class methods?

```
validate = () => {
...
}
```

I don't have a strong preference either way. I didn't know about that syntax until I saw it a few weeks ago in the react-jsonschema-form code. I'm torn about it because it might reduce bugs caused by forgetting to bind in the constructor, but the syntax itself seems non-obvious and hard to google.

src/js/common/schemaform/FormPage.jsx
jbalboni: It should go away in propTypes. FormPage is now always an actual page that is routed to and wouldn't every have an onSubmit prop passed to it. ArrayField is using SchemaForm now, which does still call that.

src/js/common/schemaform/SchemaForm.jsx
cvalarida: Yeah, that's the one.

I don't know about the googlability of it, though; if I were to google it, I'd probably search for `javascript =>` which leads me right to the [MDN page on arrow functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions), and once it has a name, it's easy to find more about it.

That's my two cents anyhow. With ES6, it's becoming pretty standard.

jbalboni: I don't mean the syntax for arrow functions is hard to google, I mean the syntax for auto binding class methods used here is (i.e. setting `validate` to equal an arrow function). It's not valid ES2015, it's a Babel [transform](https://babeljs.io/docs/plugins/transform-class-properties/) that's bundled in one of the presets we use.

I don't have any problems with arrow functions; this is a different (though related) feature.

src/js/common/schemaform/SchemaForm.jsx
```diff
@@ -0,0 +1,127 @@
...
+        submitted: false,
+        onEdit,
+        hideTitle,
+        pageTitle: title
+      }
+    };
+  }
+
+  transformErrors(errors) {
+    return transformErrors(errors, this.props.uiSchema);
```
annekainicUSDS: I'm confused what this is doing

annekainicUSDS (approved):
LGTM! Just one question.

src/js/common/schemaform/SchemaForm.jsx
jbalboni: This is taking the list of errors from JSON Schema and doing two things:

1. It moves required errors to the field instead of the object level. This is a quirk of JSON Schema that missing field errors are associated with the containing object, instead of the field itself. So like if you're missing a last name in our `veteranFullName` field, JSON Schema marks `veteranFullName` as having an error instead of `veteranFullName.last`. We change that so the error message shows up in the right place.
2. It replaces the JSON Schema error messages with messages from uiSchema or default ones in validation.js. The default error messages are pretty bad.

PR #4869, opened 2017-02-15
Add chunkfilename so names match

jbalboni:
Webpack is creating a 1995 bundle using a different filename from what's in the chunk-manifest.json file. Adding a `chunkFilename` that matches our normal `filename` parameter makes these match up.

To test this run:

```
`NODE_ENV=production npm run build -- --buildtype staging`
cd build/staging
php -S localhost:3333
```

Open http://localhost:3333/education/apply-for-education-benefits/application/1995/introduction

### DIFF
--- a/config/webpack.config.js
+++ b/config/webpack.config.js
@@ -46,7 +46,8 @@ const configGenerator = (options) => {
     output: {
       path: path.join(__dirname, `../build/${options.buildtype}/generated`),
       publicPath: '/generated/',
-      filename: (options.buildtype === 'development') ? '[name].entry.js' : '[name].entry.[chunkhash].js'
+      filename: (options.buildtype === 'development') ? '[name].entry.js' : '[name].entry.[chunkhash].js',
+      chunkFilename: (options.buildtype === 'development') ? '[name].entry.js' : '[name].entry.[chunkhash].js'
     },
     module: {
       loaders: [


config/webpack.config.js
```diff
@@ -46,7 +46,8 @@ const configGenerator = (options) => {
     output: {
       path: path.join(__dirname, `../build/${options.buildtype}/generated`),
       publicPath: '/generated/',
-      filename: (options.buildtype === 'development') ? '[name].entry.js' : '[name].entry.[chunkhash].js'
+      filename: (options.buildtype === 'development') ? '[name].entry.js' : '[name].entry.[chunkhash].js',
+      chunkFilename: (options.buildtype === 'development') ? '[name].entry.js' : '[name].entry.[chunkhash].js'
```
annekainicUSDS: where is `chunkFileName` used?

annekainicUSDS (approved):
LGTM!

config/webpack.config.js
jbalboni: It's used by Webpack to generate the name of any bundles we create using `require.ensure`, which is just the 1995 bundle (see the edu-benefits routes.jsx file).

For some reason, if we don't set it, the bundle name that is pulled out of the Webpack config by chunk-manifest-plugin is not the same as Webpack's default chunkFilename.

jbalboni:
FYI @aub @crwallace

config/webpack.config.js
annekainicUSDS: Got it, thanks!

PR #4871, opened 2017-02-15
Fix date UI on 1995 old school information page

raquelromano-gov:
Connects to  https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1123

@jbalboni It looks like this uses the date validation function in `src/js/common/schemaform/validation.js` which uses some but not all of the date validation's in `src/js/common/utils/validations.js`  Seems like we should beef up the schemaform date validation to at least use `isValidCurrentOrPastYear()`, and probably something less coarse. Is that just work in progress as we flesh out the schemaform framework?

image

### DIFF
--- a/src/js/edu-benefits/1995/config/form.js
+++ b/src/js/edu-benefits/1995/config/form.js
@@ -251,7 +251,7 @@ const formConfig = {
               },
               address: address.uiSchema()
             },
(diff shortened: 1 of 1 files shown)


jbalboni:
@raquelromano-gov Yeah, we'll definitely need to have more date validation options than we do now. In the past we've had the designers go through and tell us what the valid ranges are for each date in the form. I'll create an issue for that.

jbalboni (approved):
LGTM. I'll create a separate issue for the validation changes.

PR #3256, opened 2016-10-03
Advanced search form (layout only)

webinista:
WIP on department-of-veterans-affairs/messaging-fe#4
- Adds react-datepicker component.
- Fixes checkbox sizing bug inherited from USWDS. Resolves #3239

PR #4876, opened 2017-02-15
Resize the checkbox

cvalarida:
Connects to department-of-veterans-affairs/vets.gov-team#1206

Resizes the `input` size according to the label's `::before` size. It leaves a little space at the bottom, but I didn't want to finagle it too much to keep cross-browser testing down.

Chrome:

---
image

---
Firefox:

---
image

### DIFF
--- a/src/sass/modules/_m-form-elements.scss
+++ b/src/sass/modules/_m-form-elements.scss
@@ -75,9 +75,9 @@ button.form-button-disabled {
   // See 18F/web-design-standards#1490
   // This fixes things in the meantime.
   [type="checkbox"] {
-    height: auto !important;
+    height: 1.8rem !important;
     margin: 0;
-    width: auto !important;
+    width: 1.8rem !important;
   }
   > input[type="checkbox"] + label::before {
     display: block;


src/sass/modules/_m-form-elements.scss
```diff
@@ -75,9 +75,9 @@ button.form-button-disabled {
   // See 18F/web-design-standards#1490
   // This fixes things in the meantime.
   [type="checkbox"] {
-    height: auto !important;
+    height: 1.8rem !important;
```
raquelromano-gov: Are we sure this won't mess up layouts on other pages? It looks like this was originally added to fix something with the hca messaging application, but I don't yet know how to fake logging in so I can play with it locally: https://github.com/department-of-veterans-affairs/vets-website/pull/3256/files

jbalboni: Yeah, it'd be good to check that app. @bshyong might know the best way to log in and see.

bshyong: @raquelromano-gov do you have vets api set up locally or are you trying to take a look at the hca app without the backend?

ayaleloehr: @raquelromano-gov [these instructions](https://github.com/department-of-veterans-affairs/vets.gov-team/blob/3503c5b9fd936d83364f28a9ffe5bed2a2a7ed28/Products/Rx%20Refills/TestRxSMOnStaging.md) may be helpful. They are about testing on staging, but if you have vets-api set up locally, the same user can be used locally as well.

cvalarida: @raquelromano-gov No, I'm not sure, but I don't _think_ it will simply because it now matches the `input[type="checkbox"] + label::before`'s size, and that's the bit that visually looks like the checkbox. The `::before` size comes from the imported [USWDS style](https://github.com/18F/web-design-standards/blob/staging/src/stylesheets/elements/_inputs.scss#L165).

I'll see if I can't follow the instructions @ayaleloehr linked to and test out the messages.

cvalarida: I can't currently get into prescriptions or messaging to test (I am logged in, though). Can anybody who can get it test just make sure it doesn't mess up the styling in there? Again, I think it's probably alright, but I would like to just make sure.

cvalarida: Got in! Thanks @ayaleloehr!

In the messaging, the checkbox in advanced search looks like this originally (with `auto`):

---
image

---
And with width and height to `1.8rem`:

---
image

---
I think we should be golden.

jbalboni (approved):

PR #4882, opened 2017-02-16
Handle VA900 error code in RX

bshyong:
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1235


src/js/rx/components/ErrorView.jsx
```diff
@@ -15,7 +15,7 @@ class ErrorView extends React.Component {
     let content;
     let alert;
 
-    if (some(errors, errorCodeIncludes(['RX135']))) {
+    if (some(errors, errorCodeIncludes(['RX135', 'VA900']))) {
```
austin-martinez: Could this array and the others used in the conditionals be given names and moved out of this function, eg. `unacceptedTermsErrors`? Will help with readability and editing

src/js/rx/components/ErrorView.jsx
```diff
@@ -81,6 +81,7 @@ class ErrorView extends React.Component {
       'RX99',
       'RX503',
       'RX500',
+      'VA900',
```
austin-martinez: Could define this array outside of this function as it never changes

austin-martinez (approved):
LGTM ♻️  Would consider moving the arrays to the top of the file and out of the functions so they are only defined once

PR #4888, opened 2017-02-16
Add schema tests for 1995

jbalboni:
Closes [1098](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1098)

This adds a couple tests to run against the backend schema. This is not as extensive as what we did on 1990 for a few reasons: 

1. Since we're directly sharing most of the schema, there is less risk in us sending bad data to the backend. 
2. None of the schema mismatch errors we saw in production could have been caught by what we were doing before. 
3. Since we're not actually using the form to generate the test cases, it's not actually testing a whole lot. We would get much better mileage out of true e2e tests like we've talked about.

So instead I've done something simpler that we can add test cases for if we need to.

### DIFF
--- a/test/edu-benefits/1995/schema/maximal-test.json
+++ b/test/edu-benefits/1995/schema/maximal-test.json
@@ -1,0 +1,120 @@
+{
+  "privacyAgreementAccepted": true,
+  "veteranInformation": {
+    "data": {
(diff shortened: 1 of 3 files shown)


annekainicUSDS (approved):
LGTM!

PR #4889, opened 2017-02-16
Edu benefits landing page

cvalarida:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1092

---
image

---
There are a lot of bullet points...Does that look alright, @goldenmeanie?

Also, when we go to use it, we'll have to switch over all the `-new` files, of course, and make the appropriate changes to the references.


content/layouts/page-breadcrumbs-new.html
```diff
@@ -0,0 +1,63 @@
...
+    {% when '3-panel-filters' %}
+       {% include "content/includes/3-panel-filters.html" %}
+
+    {% when '4-action-page' %}
+      {% include "content/includes/4-action-page.html" %}
+
+    {% when '4-action-page-pending' %}
+      {% include "content/includes/4-action-page-pending.html" %}
+
+    {% when '6-info-page' %}
```
jbalboni: Would it work to add another statement to the original page-breadcrumbs.html file, instead of copying it?

```
{% when '6-info-page-new' %}
  {% include "content/includes/6-info-page-new.html" %}
```

jbalboni (approved):
Should have @goldenmeanie ok this, but it looks good to me.

mollyblake:
Can you put the blue box around the "Other Ways to Apply" section instead of the Checking in on the status?  That might help break up all the bullet points. The other option would be to add the green Apply for Benefits Button beneath the set of corresponding bullets. The same for the Manage Benefit.  @goldenmeanie

content/layouts/page-breadcrumbs-new.html
cvalarida: Yeah, it definitely would work and that would make any edits / merges easier later, but I wasn't as much concerned with copying the code as I was making it as obvious as I could that this is not production-facing until we're ready for it to be.

I could have accomplished that with comments, but I didn't find a way to comment in liquid and I didn't want to put it in HTML comments for fear of it not being stripped away in the production build chain.

goldenmeanie:
@mollyblake and I took another swing at the form section and pushed a (simplification) update.

content/layouts/page-breadcrumbs-new.html
jbalboni: Ok. This is short term anyway, so it shouldn't matter much either way.

jbalboni:
@cvalarida @goldenmeanie Is there anything left to do for this ticket? I think it can be merged.

goldenmeanie:
LGTM

PR #4890, opened 2017-02-16
Add back functionality to ReviewPage

annekainicUSDS:
The back button wasn't working on the `ReviewPage` because the `goBack` functionality wasn't being passed to the `SubmitButtons` component. Copied the functionality from `FormPage`, with a slight tweak (comparing `item.pageKey === path` instead of `item.pageKey === pageConfig.pageKey` because `pageConfig` isn't available in `ReviewPage`).

Closes https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1164

### DIFF
--- a/src/js/common/schemaform/review/ReviewPage.jsx
+++ b/src/js/common/schemaform/review/ReviewPage.jsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import Scroll from 'react-scroll';
+import _ from 'lodash/fp';
 import { connect } from 'react-redux';
 import { withRouter } from 'react-router';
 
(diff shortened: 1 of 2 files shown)


src/js/common/schemaform/review/ReviewPage.jsx
```diff
@@ -42,6 +44,22 @@ class ReviewPage extends React.Component {
...
+   * their dependencies and therefore should be skipped.
+   */
+  getEligiblePages() {
+    const { form, route: { pageList, path } } = this.props;
+    const eligiblePageList = getActivePages(pageList, form);
+    const pageIndex = _.findIndex(item => item.pageKey === path, eligiblePageList);
+    return { eligiblePageList, pageIndex };
+  }
+
+  goBack() {
```
jbalboni: Can we add a test for this method? I think there's one in the FormPage tests that should be similar.

jbalboni (approved):
LGTM

src/js/common/schemaform/review/ReviewPage.jsx
annekainicUSDS: Yep, will add that

annekainicUSDS:
@jbalboni I added a test for the back button functionality.

jbalboni:
@annekainicUSDS Looks good to me!

PR #4891, opened 2017-02-16
Hook up download link on Blue Button download page

bshyong:
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1253

Note that the MHV key for Labs and Tests, Self-reported is unknown at this time, so the box should NOT be checked, or else vets-api will throw an error.


src/js/blue-button/containers/DownloadPage.jsx
```diff
@@ -9,10 +11,7 @@ export class DownloadPage extends React.Component {
       pdfFileSize: '124 KB',
```
austin-martinez: Will this number be available when the component mounts or only after the user clicks 'Download PDF'?

src/js/blue-button/components/DownloadLink.jsx
```diff
@@ -0,0 +1,64 @@
...
+    if (this.state.downloading) return;
+
+    if (this.downloadUrl) {
+      window.open(this.downloadUrl, '_blank');
+      return;
+    }
+
+    this.setState({ downloading: true });
+    const requestUrl = '/v0/health_records';
+    apiRequest(
```
austin-martinez: Should this request send the file type too (.txt/.pdf)? Not sure how the API handles that

src/js/blue-button/components/DownloadLink.jsx
```diff
@@ -0,0 +1,64 @@
...
+    const iconClass = classNames({
+      fa: true,
+      'fa-spinner': this.state.downloading,
+      'fa-pulse': this.state.downloading,
+    });
+
+    const loadingMessage = <span>Loading...</span>;
+
+    return (
+      <button onClick={this.downloadHealthRecord} href={this.props.url}>
```
austin-martinez: Don't see `this.props.url` passed in, where is it coming from?

src/js/blue-button/components/DownloadLink.jsx
```diff
@@ -0,0 +1,64 @@
...
+        <i className={iconClass}></i>
```
austin-martinez: This icon tag will have no classes if the state is not downloading, should it be conditionally rendered instead of adding/removing classes?

austin-martinez (approved):
LGTM, few small questions

src/js/blue-button/containers/DownloadPage.jsx
bshyong: This is only available after the user clicks download PDF. Another option is to try downloading immediately when the component mounts, and have the entire row display a loading state until the download is complete. I will be discussing with the designers if we want to use this approach.

src/js/blue-button/components/DownloadLink.jsx
bshyong: This has not been implemented in the API yet; it is currently hardcoded to PDF

src/js/blue-button/components/DownloadLink.jsx
bshyong: I'll remove this—vestige from when I was experimenting with the behavior of using an `a`

src/js/blue-button/components/DownloadLink.jsx
ayaleloehr: pinging @saneshark to make sure he is aware of ^

saneshark: @bshyong what specifically is the challenge to using the browsers native download capability? Is this issue more specific to mobile vs desktop?

PR #4893, opened 2017-02-17
Add validation for current-or-past dates

raquelromano-gov:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1233

This adds a validator and unit test,and updates 1 field in the UI (see below screenshot)

Still todo: use the current-or-past validator for the service period date range, but will open a separate PR for that.

Note: I need to commit some time to getting e2e tests running locally, so I won't merge this until that's done and they're passing.

Screenshots:
image
image

### DIFF
--- a/src/js/common/schemaform/validation.js
+++ b/src/js/common/schemaform/validation.js
@@ -191,7 +191,7 @@ export function validateCurrentOrPastDate(errors, dateString) {
   validateDate(errors, dateString);
   const { day, month, year } = parseISODate(dateString);
   if (!isValidCurrentOrPastDate(day, month, year)) {
-    errors.addError('Please provide a valid date in the past');
+    errors.addError('Please provide a valid current or past date');
   }
 }
 
--- a/src/js/common/schemaform/validation.js
+++ b/src/js/common/schemaform/validation.js
@@ -3,7 +3,16 @@ import { Validator } from 'jsonschema';
 
 import { retrieveSchema } from 'react-jsonschema-form/lib/utils';
 
-import { isValidSSN, isValidPartialDate, isValidCurrentOrPastDate, isValidDateRange, isValidRoutingNumber, isValidUSZipCode, isValidCanPostalCode } from '../utils/validations';
+import {
+  isValidSSN,
+  isValidPartialDate,
+  isValidCurrentOrPastDate,
+  isValidDateRange,
+  isValidRoutingNumber,
+  isValidUSZipCode,
+  isValidCanPostalCode
+} from '../utils/validations';
+
 import { parseISODate } from './helpers';
 import { isActivePage } from '../utils/helpers';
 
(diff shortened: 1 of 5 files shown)


jbalboni (commented):
Looks good, had a couple questions.

src/js/edu-benefits/1995/config/form.js
```diff
@@ -258,7 +259,9 @@ const formConfig = {
               },
               address: address.uiSchema()
             },
-            trainingEndDate: date.uiSchema('When did you stop taking classes or participating in the training program?'),
+            trainingEndDate: _.merge(date.uiSchema('When did you stop taking classes or participating in the training program?'), {
```
raquelromano-gov: Yeah, sounds like a good idea. Will do that.

test/common/utils/validations.unit.spec.js
```diff
@@ -195,16 +196,16 @@ describe('Validations unit tests', () => {
     it('should validate partial range', () => {
       const fromDate = {
         month: {
-          value: '2'
+          value: 2
```
raquelromano-gov: Just for consistency as there seemed to be places they were strings and places they were numbers. Perhaps they should all be sent as strings then?

raquelromano-gov: @jbalboni LMK if you think these should be strings or numbers.

src/js/edu-benefits/1995/config/form.js
raquelromano-gov: Done

test/common/utils/validations.unit.spec.js
raquelromano-gov: PTAL - this makes the test kind of messy (esp. the places where values generated by moment() are converted to strings), but at least it's consistent now and emulates the components more accurately.

src/js/common/schemaform/validation.js
```diff
@@ -3,7 +3,7 @@ import { Validator } from 'jsonschema';
 
 import { retrieveSchema } from 'react-jsonschema-form/lib/utils';
 
-import { isValidSSN, isValidPartialDate, isValidDateRange, isValidRoutingNumber, isValidUSZipCode, isValidCanPostalCode } from '../utils/validations';
+import { isValidSSN, isValidPartialDate, isValidCurrentOrPastDate, isValidDateRange, isValidRoutingNumber, isValidUSZipCode, isValidCanPostalCode } from '../utils/validations';
```
jbalboni: We should probably reformat this so each function is on its own line.

jbalboni (approved):
LGTM.

src/js/common/schemaform/validation.js
raquelromano-gov: Done

Issue #3433, opened 2016-10-18
Phone and SSN display with dashes


akainic:
On the review screen, phone number and SSN should format with the dashes whether entered that way or not initially.

cvalarida:
I also found that it accepts alphabet characters as well. Should that be addressed with this?

cvalarida:
https://github.com/department-of-veterans-affairs/vets-website/pull/4859 formats the SSN, so just the phone is left.

PR #4896, opened 2017-02-18
Formats the phone number on the review page

cvalarida:
Connects to https://github.com/department-of-veterans-affairs/vets-website/issues/3433

Formats the phone numbers on the review page:

---
image

---
**Note:** If the number has more than 10 digits (validation requires that it has at least 10), it doesn't get formatted. Instead, all the non-numeric characters are just stripped away (as part of the formatting process).

---
image

---
This means it doesn't format international numbers too kindly. I didn't worry about international formatting for now because there are just [so many different formats](https://en.wikipedia.org/wiki/National_conventions_for_writing_telephone_numbers), but if it's required, I can certainly look into a more comprehensive solution.

### DIFF
--- a/src/js/common/schemaform/review/PhoneNumberWidget.jsx
+++ b/src/js/common/schemaform/review/PhoneNumberWidget.jsx
@@ -1,0 +1,11 @@
+import React from 'react';
+
+// When coming from a PhoneNumberWidget (not review), value will be only numbers
+export default function PhoneNumberWidget({ value }) {
+  let formatted = value;
+  if (value && value.length === 10) {
+    formatted = `(${value.substr(0, 3)}) ${value.substr(3, 3)}-${value.substr(6)}`;
+  }
+
+  return <span>{formatted}</span>;
+}
--- a/src/js/common/schemaform/widgets/PhoneNumberWidget.jsx
+++ b/src/js/common/schemaform/widgets/PhoneNumberWidget.jsx
@@ -1,0 +1,22 @@
+import React from 'react';
+import TextWidget from './TextWidget';
+
+/*
+ * Handles removing dashes from SSNs by keeping the user input in local state
+ * and saving the transformed version instead
+ */
+export default class PhoneNumberWidget extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = { val: props.value };
+  }
+  handleChange = (val) => {
+    const stripped = val.replace(/[\D]/g, '');
+    this.setState({ val }, () => {
+      this.props.onChange(stripped);
+    });
+  }
+  render() {
+    return <TextWidget {...this.props} value={this.state.val} onChange={this.handleChange}/>;
+  }
+}
--- a/src/js/common/schemaform/widgets/PhoneNumberWidget.jsx
+++ b/src/js/common/schemaform/widgets/PhoneNumberWidget.jsx
@@ -11,7 +11,11 @@ export default class PhoneNumberWidget extends React.Component {
     this.state = { val: props.value };
   }
   handleChange = (val) => {
-    const stripped = val.replace(/[\D]/g, '');
+    let stripped;
+    if (val) {
+      stripped = val.replace(/[ \-\(\)x+]/g, '');
+    }
+
     this.setState({ val }, () => {
       this.props.onChange(stripped);
     });
--- a/test/common/schemaform/review/PhoneNumberWidget.unit.spec.jsx
+++ b/test/common/schemaform/review/PhoneNumberWidget.unit.spec.jsx
@@ -1,0 +1,22 @@
+import React from 'react';
+import { expect } from 'chai';
+import SkinDeep from 'skin-deep';
+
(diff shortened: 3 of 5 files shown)


src/js/common/schemaform/widgets/PhoneNumberWidget.jsx
```diff
@@ -0,0 +1,22 @@
...
+ * Handles removing dashes from SSNs by keeping the user input in local state
+ * and saving the transformed version instead
+ */
+export default class PhoneNumberWidget extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = { val: props.value };
+  }
+  handleChange = (val) => {
+    const stripped = val.replace(/[\D]/g, '');
```
cvalarida: This line (14) is the only difference aside from naming between this component and `SSNWidget`. Perhaps we could consider making a `StrippedText` widget or something that takes a string `value` and a replacement regex?

Again, I'm not sure how to use it in the schema, is the problem.

src/js/common/schemaform/review/PhoneNumberWidget.jsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+
+// When coming from a PhoneNumberWidget (not review), value will be only numbers
+export default function PhoneNumberWidget({ value }) {
+  let formatted = value;
+  if (value && value.length === 10) {
+    formatted = `(${value.substr(0, 3)}) ${value.substr(3, 3)}-${value.substr(6)}`;
+  }
```
cvalarida: I'd like to pull this out into a common module to be used for any formatted string. Perhaps also supply a `transform` function to modify the value? But then how would we use that (easily) in the schema?

cvalarida (commented):
Still working on writing tests, but I just wanted to get this up so I could comment on it before I forgot my comments.

src/js/common/schemaform/review/PhoneNumberWidget.jsx
jbalboni: There are two ways I can think of doing it. We could add a transform option to the uiSchema. The other way is a higher order component: a function that takes a transform function as a parameter and returns a component that uses it.

That said, we have two instances of this so far. I'm hesitant to add harder to understand code for a use case that I'm not sure goes beyond these two fields.

test/common/schemaform/review/PhoneNumberWidget.unit.spec.jsx
```diff
@@ -0,0 +1,22 @@
...
+    );
+
+    expect(tree.text()).to.equal('(123) 456-7890');
+  });
+  it('should render empty value', () => {
+    const tree = SkinDeep.shallowRender(
+      <PhoneNumberWidget/>
+    );
+
+    expect(tree.text()).to.equal('');
```
jbalboni: This is actually a bug. It should equal `undefined`, because JSON Schema interprets empty strings as non-empty fields. There's an issue to fix this with the SSN widget, too.

src/js/common/schemaform/widgets/PhoneNumberWidget.jsx
jbalboni: I think it'd be better to switch this to exclude spaces, dashes, and parens only. If we strip other non-numeric characters, we're allowing through characters that are likely typed by accident and not telling the user.

jbalboni (commented):
Looks good, made a couple comments.

jbalboni:
@cvalarida I think that's probably ok for the phone format, but @goldenmeanie could weigh in.

goldenmeanie:
I'd like to deal with the international number formatting, but it seems like it's complicated and something that we might get on Core’s backlog. Let’s do the simple thing for now:

I think your current approach is a reasonable place to be. If it’s ten digits (we have high confidence in how to format) then do so. If there are extra numbers just show them. It may be worth accommodating `+` and `x` since they may be used for international numbers or extensions.

I will check with stakeholders about what problems they've had with phone numbers.

src/js/common/schemaform/review/PhoneNumberWidget.jsx
cvalarida: That's fair; we can always refactor it later if we find we'll be doing this kind of thing for more than these two fields.

test/common/schemaform/review/PhoneNumberWidget.unit.spec.jsx
cvalarida: Good point. I made a note in that issue so we don't lose track of it.

cvalarida:
No turns:

---
image

---
Into:

---
image

test/common/schemaform/review/PhoneNumberWidget.unit.spec.jsx
jbalboni: Actually, I take that back, this is probably fine. But we need a unit test that shows `undefined` is passed through to `onChange` correctly.

jbalboni (approved):
This looks good to me, once there's a test for the undefined issue that we saw with SSN.

PR #4898, opened 2017-02-20
Set proper service type; resolves #1287

bshyong:
Sets the correct service type for blue button app
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1287


U-DON (approved):
LGTM.

PR #4900, opened 2017-02-20
Add tests for common definitions and a 1995 page

jbalboni:
Closes [1242](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1242)

This adds some tests for our common and page definitions, since this was a gap in our testing.

There's some overlap here with what we test in other unit tests, so what I was trying to do was test things that are specific to the uiSchema and schema being tested. That means checking things like:

- Validation error message text
- That a particular validation function is being run
  - But not all the logic in a validation function, since that will be tested by unit tests
- That specific widgets are being used
  - Same as above, don't need to hit all the logic in the widget, just that it's being used
- Field ordering or labels (labels are probably overkill for the most part)
- Uses of hideIf, schemaUpdate, expandUnder

Any feedback on the above list would be great, too.

This PR has tests for most of our common definitions and tests for one 1995 page. We should write tests for the other 1995 pages, but I wanted to keep this PR a little shorter.


src/js/common/schemaform/SchemaForm.jsx
```diff
@@ -90,11 +90,13 @@ class SchemaForm extends React.Component {
...
       children,
       onSubmit,
-      onChange
+      onChange,
+      safeRenderCompletion
     } = this.props;
     return (
       <div>
         <Form
+            safeRenderCompletion={safeRenderCompletion}
```
jbalboni: This is something built into rjsf's Form component that removes some of the async state setting it normally does, which makes it easier to test.

src/js/common/schemaform/definitions/dateRange.js
```diff
@@ -3,17 +3,23 @@ import * as date from './date';
...
 import commonDefinitions from 'vets-json-schema/dist/definitions.json';
 
-export const schema = commonDefinitions.dateRange;
+export const schema = {
+  type: 'object',
+  properties: {
+    from: commonDefinitions.date,
+    to: commonDefinitions.date
+  }
+};
```
raquelromano-gov: I'm guessing the purpose of this just to simplify the schema? I.e., replace
```
dateRange: {
  from: {
    date: <start-date>
  },
  to: {
    date: <end-date>
  }
}
```
with
```
dateRange: {
  from: <start-date>,
  to: <end-date>
}
```
Any way to do this in the dateRange definition in definitions.js?

test/common/schemaform/definitions/dateRange.unit.spec.jsx
```diff
@@ -0,0 +1,81 @@
...
+    const form = ReactTestUtils.renderIntoDocument(
+      <DefinitionTester
+          schema={schema}
+          uiSchema={uiSchema()}/>
+    );
+
+    const formDOM = findDOMNode(form);
+
+    expect(formDOM.querySelectorAll('label').length).to.equal(8);
+    expect(formDOM.querySelectorAll('input').length).to.equal(2);
```
raquelromano-gov: Also check that 4 select elements are rendered?

test/common/schemaform/definitions/phone.unit.spec.jsx
```diff
@@ -0,0 +1,35 @@
...
+          schema={schema}
+          uiSchema={uiSchema()}/>
+    );
+
+    const formDOM = findDOMNode(form);
+
+    const input = formDOM.querySelector('input');
+    expect(input.type).to.equal('tel');
+    expect(input.classList.contains('home-phone')).to.be.true;
+    expect(input.classList.contains('va-input-medium-large')).to.be.true;
```
raquelromano-gov: Do we need to check for classes here? Seems like style changes could make this brittle, but I don't know how/where you've been testing styling in the past. Same for the fullName select menu above.

test/common/schemaform/definitions/phone.unit.spec.jsx
```diff
@@ -0,0 +1,35 @@
...
+      <DefinitionTester
+          schema={schema}
+          uiSchema={uiSchema('My phone')}/>
+    );
+
+    const formDOM = findDOMNode(form);
+
+    expect(formDOM.querySelector('label').textContent).to.equal('My phone');
+  });
+});
```
raquelromano-gov: Also check for an invalid phone number?

test/edu-benefits/1995/config/veteranInformation.unit.spec.jsx
```diff
@@ -0,0 +1,93 @@
...
+
+    ReactTestUtils.findRenderedComponentWithType(form, Form).onSubmit({
+      preventDefault: f => f
+    });
+
+    // show error for ssn, file number not visible
+    expect(formDOM.querySelector('#root_vaFileNumber')).to.be.null;
+    expect(formDOM.querySelector('.usa-input-error #root_veteranSocialSecurityNumber')).not.to.be.null;
+
+    ReactTestUtils.Simulate.change(ReactTestUtils.scryRenderedDOMComponentsWithTag(form, 'input')[4],
```
raquelromano-gov: This could be hard to maintain--maybe add a comment that veteranInformation has a fullName (4 inputs), followed by an SSN input, or just search for it by name if possible.

test/edu-benefits/1995/config/veteranInformation.unit.spec.jsx
```diff
@@ -0,0 +1,93 @@
...
+      {
+        target: {
+          checked: true
+        }
+      });
+
+    // no error for ssn, file number is visible with error
```
raquelromano-gov: Add a check that the file number error is present as well (this only checks that the ssn error is gone and the file number input is shown)

src/js/common/schemaform/definitions/dateRange.js
jbalboni: I mainly made this change because the common dateRange definition use `$ref`, which is JSON Schema's way to share common schema definitions within a schema. It makes it more difficult to use these common definitions in places because we have to handle including the referenced schema in places. So I've been trying to avoid it in our common definitions.

test/common/schemaform/definitions/phone.unit.spec.jsx
jbalboni: I was doing this because there's a class set in the definition for this field. So I wanted to make sure it was set correctly. Otherwise, yeah, we try to avoid checking for classes too much.

test/common/schemaform/definitions/dateRange.unit.spec.jsx
jbalboni: Yeah, that would be probably be better. I was having trouble figuring out where to draw the line for these tests. I don't want to test every detail of all the elements rendered, but I want some assurance the right elements are present.

test/edu-benefits/1995/config/veteranInformation.unit.spec.jsx
jbalboni: The React test utils are pretty low level and limited, so the only search methods present are for tag name, css class, or component class. I can make this clearer, though.

src/js/common/schemaform/definitions/dateRange.js
raquelromano-gov: Ah, OK--I saw $ref used all over the place so figured we were generally for it.

Aside: it might be useful to start a vets-website FE style guide that starts to capture any opinions we as a team develop, esp. regarding rjsf but anything else we use heavily, esp. if there are no external style guides we're conforming to. It could start as just a few thoughts and evolve as things become stable.

test/common/schemaform/definitions/dateRange.unit.spec.jsx
raquelromano-gov: Your call on the level of detail to test--I just noticed that the other tests all verify that all of the input elements are present.

src/js/common/schemaform/definitions/dateRange.js
ayaleloehr: @raquelromano-gov similar conversation started [here](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/659) if you're interested in starting that!

jbalboni:
@raquelromano-gov Comments should all be addressed.

test/edu-benefits/1995/config/veteranInformation.unit.spec.jsx
raquelromano-gov: Just a comment would help

src/js/common/schemaform/definitions/dateRange.js
raquelromano-gov: @ayaleloehr Thanks, I chimed in there.

raquelromano-gov (approved):
LGTM

PR #4901, opened 2017-02-20
Only show triage team error if folders load

bshyong:
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1255

The triage team error should only show when recipients fail to load. If there is an application level error, i.e. no folders load, only the standard loading error should be shown.

screen shot 2017-02-20 at 11 24 45 am


U-DON (approved):
LGTM.

PR #4904, opened 2017-02-20
SSN returns undefined if blank

cvalarida:
Connected to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1266

Basically, this just turns `''` into `undefined` so the value doesn't "pop" back if deleted.

### DIFF
--- a/src/js/common/schemaform/widgets/SSNWidget.jsx
+++ b/src/js/common/schemaform/widgets/SSNWidget.jsx
@@ -11,6 +11,7 @@ export default class SSNWidget extends React.Component {
     this.state = { val: props.value };
   }
   handleChange = (val) => {
+    // If val is blank or undefined, pass undefined to onChange
     let strippedSSN;
     if (val) {
       strippedSSN = val.replace(/[\- ]/g, '');
--- a/src/js/common/schemaform/widgets/SSNWidget.jsx
+++ b/src/js/common/schemaform/widgets/SSNWidget.jsx
@@ -11,7 +11,11 @@ export default class SSNWidget extends React.Component {
     this.state = { val: props.value };
   }
   handleChange = (val) => {
-    const strippedSSN = val.replace(/[\- ]/g, '');
+    let strippedSSN;
+    if (val) {
+      strippedSSN = val.replace(/[\- ]/g, '');
+    }
+
     this.setState({ val }, () => {
       this.props.onChange(strippedSSN);
     });
--- a/src/js/common/schemaform/widgets/TextWidget.jsx
+++ b/src/js/common/schemaform/widgets/TextWidget.jsx
@@ -12,7 +12,7 @@ export default function TextWidget(props) {
         maxLength={props.schema.maxLength}
         autoComplete={props.options.autocomplete || false}
         className={props.options.widgetClassNames}
-        value={props.value || ''}
+        value={props.value}
         onBlur={() => props.onBlur(props.id)}
         onChange={(event) => props.onChange(event.target.value ? event.target.value : undefined)}/>
   );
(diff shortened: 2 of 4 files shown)


src/js/common/schemaform/widgets/SSNWidget.jsx
```diff
@@ -11,7 +11,11 @@ export default class SSNWidget extends React.Component {
     this.state = { val: props.value };
   }
   handleChange = (val) => {
-    const strippedSSN = val.replace(/[\- ]/g, '');
+    let strippedSSN;
+    if (val) {
+      strippedSSN = val.replace(/[\- ]/g, '');
+    }
```
cvalarida: Perhaps having this in `handleChange` isn't the best place, but is it needed elsewhere? I mean, when we make a new component with `val=''` is the only time that it can be `''` without getting translated to `undefined`.

However, it also means the unit test that renders an empty string still works, but is perhaps misleading.

src/js/common/schemaform/widgets/TextWidget.jsx
```diff
@@ -12,7 +12,7 @@ export default function TextWidget(props) {
         maxLength={props.schema.maxLength}
         autoComplete={props.options.autocomplete || false}
         className={props.options.widgetClassNames}
-        value={props.value || ''}
```
cvalarida: Could change this to `value={props.value || undefined}` but it seemed a little redundant.

cvalarida (commented):
A couple things that may or may not be worth noting.

src/js/common/schemaform/widgets/TextWidget.jsx
jbalboni: It doesn't look like this change is affecting anything. The empty string should be turned into undefined by the onChange function below.

src/js/common/schemaform/widgets/SSNWidget.jsx
jbalboni: I'm not sure what you mean here.

src/js/common/schemaform/widgets/SSNWidget.jsx
```diff
@@ -11,7 +11,11 @@ export default class SSNWidget extends React.Component {
     this.state = { val: props.value };
   }
   handleChange = (val) => {
-    const strippedSSN = val.replace(/[\- ]/g, '');
+    let strippedSSN;
```
jbalboni: Might put a comment in here about intentionally letting `undefined` through. It sort of looks like we're accidentally (by not setting strippedSSN when val is falsy) allowing onChange to be called with `undefined`.

src/js/common/schemaform/widgets/SSNWidget.jsx
```diff
@@ -11,7 +11,11 @@ export default class SSNWidget extends React.Component {
...
+    if (val) {
+      strippedSSN = val.replace(/[\- ]/g, '');
+    }
+
     this.setState({ val }, () => {
       this.props.onChange(strippedSSN);
```
jbalboni: Should add a unit test that shows calling `handleChange` with `undefined` results in `props.onChange` being called with `undefined` now.

jbalboni (commented):
Looks good, had a couple comments.

src/js/common/schemaform/widgets/SSNWidget.jsx
cvalarida: I just mean we can initialize a component with `val=''`, but once we modify the value, it can never get back to `''`. Might lead to some weird cases is all, and I didn't know if there was a better way to ensure `''` is always translated to `undefined`.

src/js/common/schemaform/widgets/TextWidget.jsx
cvalarida: Right, leaving the empty string in there would be fine, but a bit redundant is all. Either the input's `value` is whatever falsey value `props.value` was, or the specific falsey value of `''` before turning into `undefined`. I just figured having that `|| ''` looked an awful lot like we wanted an empty string, when what we really want is `undefined` in the end.

cvalarida:
ssn validation

jbalboni (approved):

PR #4905, opened 2017-02-21
1995 new school order

cvalarida:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1247

I'm not super fond of this method, but it seemed a better fit than changing the schema since it's a purely cosmetic change.

### DIFF
--- a/src/js/edu-benefits/1995/helpers.jsx
+++ b/src/js/edu-benefits/1995/helpers.jsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import _ from 'lodash/fp';
 import { transformForSubmit } from '../../common/schemaform/helpers';
 
 export const benefitsLabels = {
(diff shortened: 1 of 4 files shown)


src/js/edu-benefits/1995/helpers.jsx
```diff
@@ -23,7 +24,16 @@ export const preferredContactMethodLabels = {
 };
 
 export function transform(formConfig, form) {
-  const formData = transformForSubmit(formConfig, form);
+  // All the king's horses and all the king's men
+  //  Put newSchool back together again.
+  const repairedForm = _.set('newSchool.data.newSchool', {
```
jbalboni: Should add a test for this.

jbalboni (approved):
This looks fine to me, once we add a test. Like I said last night, I think I'll talk to Alex and Lihan and figure out the best long term solution for this.

src/js/edu-benefits/1995/helpers.jsx
cvalarida: I didn't look to see that tests had been written for this at all yet...

Just modified the data for the unit tests to mirror what's happening in the front end with `newSchool`, so the transform does have to piece it together in the unit test now.

PR #4906, opened 2017-02-21
Updated strategy for 1995 OMB, et al

cvalarida:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1256

Removes OMB Number from subtitle and adds it to the bottom of the intro page along with respondent burden, expiration date, and privacy act statement link.

Adds modal containing the privacy act notice and respondent burden explanation.

image

When the screen is small enough, the close button overlaps the modal body's scrollbar, making it a bit wonky:

image

But it does play nicely with smaller screens:

image

### DIFF
--- a/src/js/common/components/OMBInfo.jsx
+++ b/src/js/common/components/OMBInfo.jsx
@@ -8,7 +8,6 @@ const modalContents = (
   </div>
 );
 
-// export default function OMBInfo({ resBurden, ombNumber, expDate }) {
 class OMBInfo extends React.Component {
   constructor(props) {
     super(props);
--- a/src/js/common/components/OMBInfo.jsx
+++ b/src/js/common/components/OMBInfo.jsx
@@ -1,12 +1,53 @@
 import React from 'react';
+import Modal from './Modal';
 
-export default function OMBInfo({ resBurden, ombNumber, expDate }) {
-  return (
-    <div className="omb-info">
-      <div>Respondent burden: <strong>{resBurden} minutes</strong></div>
-      <div>OMB Control # <strong>{ombNumber}</strong></div>
-      <div>Expiration date: <strong>{expDate}</strong></div>
-      <div><a href="#">Privacy Act Statement</a></div>
-    </div>
-  );
+const modalContents = (
+  <div>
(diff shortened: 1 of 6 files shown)


jbalboni:
@cvalarida Can you include a screenshot of how this looks on the intro page without the modal open?

cvalarida:
Right, of course. I finished up with making the modal and forgot about this bit:

---
image

src/sass/edu-benefits.scss
```diff
@@ -106,12 +107,22 @@ ul.claim-list{
   }
 }
 
+// Align with navigation buttons
+.omb-info--container {
```
jbalboni: Are these edu only styles? Or something that would be necessary if we used this component elsewhere?

jbalboni (commented):
Have a couple questions, but looks good overall.

src/js/edu-benefits/1995/components/IntroductionPage.jsx
```diff
@@ -73,6 +74,9 @@ class IntroductionPage extends React.Component {
                 afterText="»"/>
           </div>
         </div>
+        <div className="omb-info--container">
+          <OMBInfo resBurden={20} ombNumber="2900-0074" expDate="05-31-2018"/>
```
goldenmeanie: That's copied from the paper form, but I think using our usual vets.gov style is better!

src/sass/edu-benefits.scss
goldenmeanie: Every form we put on vets.gov will need to display this type of metadata

cvalarida: I could move it to `styles.scss` if it would make more sense there.

src/js/edu-benefits/1995/components/IntroductionPage.jsx
cvalarida: What's the usual format for dates?

jbalboni: @cvalarida MM/DD/YYYY

src/sass/edu-benefits.scss
jbalboni: Even though it's pretty small, I think it would be fine as its own module.

src/js/common/components/OMBInfo.jsx
```diff
@@ -0,0 +1,53 @@
...
+  </div>
+);
+
+// export default function OMBInfo({ resBurden, ombNumber, expDate }) {
```
jbalboni: Can remove this comment

jbalboni (approved):
LGTM

PR #4907, opened 2017-02-21
Add 1995 end to end tests

jbalboni:
Closes [1097](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1097)

Adds e2e tests for 1995. Pretty much the same as for the 1990. I'm using the same test data as our schema tests.


test/edu-benefits/1995/00-required.e2e.spec.js
```diff
@@ -0,0 +1,102 @@
...
+      client
+        .waitForElementVisible('label[for="root_view:newService"]', Timeouts.slow);
+      EduHelpers.completeServicePeriods(client, EduHelpers.testData.servicePeriods.data, true);
+      client.click('.form-progress-buttons .usa-button-primary');
+      E2eHelpers.expectNavigateAwayFrom(client, '/military-history/service-periods');
+
+      // Military service page.
+      client
+        .waitForElementVisible('label[for="root_view:hasServiceBefore1978"]', Timeouts.slow);
+      // Another mysteriously required pause. If we don't wait here, then the click below will
```
cvalarida: I'm not sure why this is needed, but when profiling the navigation to this page, I found it has a lot of unnecessary DOM updates to the `YesNoWidget`, which might be the cause. It still renders super quickly (6 - 7 ms), though.

image

test/edu-benefits/1995/00-required.e2e.spec.js
```diff
@@ -0,0 +1,102 @@
...
+
+      // Review and Submit Page.
+      client
+        .expect.element('label[name="privacyAgreement-label"]').to.be.visible;
+      // When you try to click on the label in the normal way, it'll instead click on the link
+      // inside the label that shows the popup. So we have to do this disgusting hack.
+      client.pause(1000);
+      client.execute((selector) => {
+        document.querySelector(selector).click();
+      }, ['label[name="privacyAgreement-label"]']);
```
cvalarida: How odd...Is there a reason it has to be the label, or can you do the following?
```js
client.click('input[name="privacyAgreement"]');`
```

cvalarida (commented):
Looks good to me. I'd like to clear up all those `client.pause()` calls, but I'd need to look more deeply into why they're needed. For now, it works.

One thing I was wondering, though: Is it outside the scope of our e2e tests to check validation?

test/edu-benefits/1995/00-required.e2e.spec.js
jbalboni: It seems like that should work. I'm not sure where this style of writing it came from.

test/edu-benefits/1995/00-required.e2e.spec.js
jbalboni: That seems odd. I think this was carried over from the edu tests and the page didn't have a YesNo widget there. I'm not sure these pauses are actually necessary.

test/edu-benefits/1995/00-required.e2e.spec.js
```diff
@@ -0,0 +1,102 @@
...
+    (client) => {
+      EduHelpers.initApplicationSubmitMock();
+
+      // Ensure introduction page renders.
+      client
+        .url(`${E2eHelpers.baseUrl}/education/apply-for-education-benefits/application/1995`)
+        .waitForElementVisible('body', Timeouts.normal)
+        .assert.title('Apply for education benefits: Vets.gov')
+        .waitForElementVisible('div.form-progress-buttons', Timeouts.slow)
+        .click('.form-progress-buttons .usa-button-primary');
```
jbalboni: Could also be `button[type=submit]`, which is more clear, too, I think.

jbalboni:
@cvalarida I don't think it's necessarily out of scope to test validation, but I think I would prefer to test that in the form config tests as much as possible. 

I need to write up a document about what we should test where, I think. So we're all on the same page.

test/edu-benefits/1995/00-required.e2e.spec.js
jbalboni: That apparently does not work. Not sure why.

test/edu-benefits/1995/00-required.e2e.spec.js
jbalboni: I think I'm going to leave this for now. My suggestion doesn't really seem clearer and I don't think we should add a class just for testing if we can avoid id.

PR #4932, opened 2017-02-22
Better browser-agnostic scrolling

cvalarida:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1317

Should work in most browsers now, but can you test IE, @annekainicUSDS (or anybody else)?

Turns out Firefox always returns `0` for `document.body.scrollTop` (and my tests showed that it returned `0` for any element's `scrollTop`, which was odd).


jbalboni (approved):
LGTM. Tried it in FIrefox and IE11 (on BrowserStack).

PR #4934, opened 2017-02-22
Resolve 1288

bshyong:
Updates styles on download page and adds a link to the text version of the health records

Screenshot:
screen shot 2017-02-22 at 8 48 23 am

Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1332
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1288

### DIFF
--- a/src/js/blue-button/components/DownloadLink.jsx
+++ b/src/js/blue-button/components/DownloadLink.jsx
@@ -50,8 +50,8 @@ class DownloadLink extends React.Component {
 }
 
 DownloadLink.propTypes = {
-  name: React.PropTypes.string,
-  docType: React.PropTypes.string,
+  name: React.PropTypes.string.isRequired,
+  docType: React.PropTypes.oneOf(['pdf', 'txt']),
 };
 
 export default DownloadLink;
--- a/src/js/blue-button/components/DownloadLink.jsx
+++ b/src/js/blue-button/components/DownloadLink.jsx
@@ -21,7 +21,7 @@ class DownloadLink extends React.Component {
     }
 
     this.setState({ downloading: true });
-    const requestUrl = '/v0/health_records';
+    const requestUrl = `/v0/health_records?doc_type=${this.props.docType}`;
     apiRequest(
       requestUrl,
       null,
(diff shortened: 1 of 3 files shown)


src/js/blue-button/containers/BlueButtonApp.jsx
```diff
@@ -32,25 +32,23 @@ function AppContent({ children, isDataAvailable }) {
 export class BlueButtonApp extends React.Component {
   render() {
     return (
-      <RequiredLoginView authRequired={3} serviceRequired={"health-records"}>
```
U-DON: Was this removed just for testing?

src/js/blue-button/containers/DownloadPage.jsx
```diff
@@ -8,45 +9,22 @@ export class DownloadPage extends React.Component {
...
+        {data.status}
+        <p>
+          <strong>Request Date:</strong> {data.requestDate}
+        </p>
+        <div>
+          <DownloadLink name="PDF File" docType="pdf"/>
+          <DownloadLink name="Text File" docType="txt"/>
+        </div>
+        <p>
+          <Link to="/blue-button">Start a new request</Link>
```
U-DON: Isn't the router rooted at `/healthcare/blue-button`? This could just be `to="/"` in that case.

src/js/blue-button/components/DownloadLink.jsx
```diff
@@ -51,7 +51,7 @@ class DownloadLink extends React.Component {
 
 DownloadLink.propTypes = {
   name: React.PropTypes.string,
-  contentType: React.PropTypes.string,
+  docType: React.PropTypes.string,
```
U-DON: `docType` seems like it should be a required prop so that the `requestUrl` doesn't get something strange. And probably limit it to specific doc types with `React.PropTypes.oneOf`.

U-DON (changes requested):
Screenshot looks good, but had a couple comments.

U-DON (approved):

PR #4939, opened 2017-02-22
Restoring min-height for .home .splash. Simplifying selector.

ghost:


melwoodard (commented):
@tiffanybbrown-va MUCH better thank you!

There's still an awkward point between 600-768 where the image size has jumped down but the tagline box hasn't, and it's blocking a lot of the image. I'd rather see the smaller tagline box size all the way up to 768. If you think that's a more lengthy fix, happy to push it to a separate issue.

tiffanybbrown-va:
It's a lengthier fix. Let's push that to a separate issue so that this one can land.

PR #4941, opened 2017-02-22
Fix required field on 1995

jbalboni:
I missed this in the PR somehow. Have to use the required array to make this field required.


cvalarida (approved):
LGTM

PR #4944, opened 2017-02-22
Fix some sluggishness on pages with addresses

jbalboni:
Related to [1313](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1313)

The Address component was missing some optimizations that ObjectField has. I also updated it to calculate the different state schemas outside of the render method, so we can switch between them and take advantage of the SelectWidget skipping updates when the schema is shallowly equal.

Before (time spent in `render()` that resulted in no visual changes when typing 3 characters into the street address field and then removing them):
screen shot 2017-02-22 at 4 47 08 pm

After:
screen shot 2017-02-22 at 4 49 17 pm

This is on a slow Android phone. It's not really noticeable on desktop or new phones.

### DIFF
--- a/src/js/common/schemaform/Address.jsx
+++ b/src/js/common/schemaform/Address.jsx
@@ -4,17 +4,37 @@ import { set, assign } from 'lodash/fp';
 
 import { getDefaultFormState } from 'react-jsonschema-form/lib/utils';
 
+import * as address from './definitions/address';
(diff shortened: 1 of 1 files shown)


cvalarida:
I'm getting pretty similar results running remote debugging on my Android Galaxy S5:

---
image

cvalarida (approved):
LGTM

PR #4945, opened 2017-02-22
Enhancement and bugfix for common pagination for last pages

ben-USDS-damman:
* Fixes strange behavior of showing links to pages that do not exist (past the total number of pages)
* Displays links as `(1) (2) (3) ... (500) (next)` except for the last set of pages, e.g. `(prev) (498) (499) (500)`
* To be used by GI Bill Comparison Tool


ayaleloehr:
@ben-USDS-damman tests need to be passing first, then some subset of @annekainicUSDS @U-DON and @crwallace should take a look, as this impacts a common component that is likely used elsewhere across the site.

jcmeloni-usds:
This would probably close https://github.com/department-of-veterans-affairs/vets.gov-team/issues/823, so I'm a fan (after tests pass, etc)

ben-USDS-damman:
This PR is the lowest priority for me right now - if anyone has spare cycles to shore up the affected tests please feel free to push commits here. fyi @saneshark

U-DON:
Fixed the lint issue. I was also adding some unit tests for this component and found when the current page was within range of the last page, assuming `maxPageListLength = 5`, it would look something like:

(6) (7) (8) ... (10) (Next)

It seemed strange to skip the (9) when it would naturally slot in there, so I made a fix to accommodate that. Also, included a test for this case.

bshyong:
LGTM

ben-USDS-damman:
Thanks be to @U-DON and @bshyong for the merge!

PR #4948, opened 2017-02-23
Disable submission for health records report when no types are chosen. Error handling for custom date range.

U-DON:
Resolves department-of-veterans-affairs/vets.gov-team#1294.

### Changes:
* Disabled 'Submit' button when no types are chosen.

    > screen shot 2017-02-22 at 11 45 52 pm

* Made default date option of '3 months' (first choice).
* Simplified reportTypes map initialization in reducer (avoids disabling es-lint).
* Allow picking same date for start and end date in custom range. Accounting for the exact time, the latter is the very end of the selected date, and the former is the beginning of that date.
* Limit date selection in picker based so that start date can't be after end date.
* If the user manually types an incorrect date format for custom date range, attempt to convert to the expected format after the input gets unfocused. Display error if unable.

    > screen shot 2017-02-22 at 11 46 20 pm

* Revert to non-error state for custom date range when switching away to another option and back. Also when re-focusing the input that had the error.

    > screen shot 2017-02-22 at 11 46 59 pm

### TODO:
* Open questions of what to do about enabling/disabling 'Submit' button when there's a problem with the custom date range when it's chosen. CC: @altangel 
    1. When there's no start or end date, what does that mean? Should it be allowed?
    2. If there's a date format error, does it matter as long as there's at least either a start date or end date already entered?


bshyong (dismissed):

src/js/health-records/containers/Main.jsx
```diff
@@ -155,24 +193,34 @@ export class Main extends React.Component {
...
+            <div className={customDateOptionClass}>
+              {
+                customDateRangeError && <p className="date-range-error">
+                Enter dates in the MM/DD/YYYY date format</p>
+              }
+              <span>Custom date range</span>
               <div className="date-range-fields">
                 <DatePicker
                     id="custom-date-start"
+                    onBlur={handleFormattedDate()}
```
bshyong: 👍

src/js/health-records/containers/Main.jsx
```diff
@@ -155,24 +193,34 @@ export class Main extends React.Component {
...
                     onChange={this.handleStartDateChange}
+                    onFocus={() => this.setState({ invalidStartDateFormat: false })}
                     placeholderText="MM/DD/YYYY"
-                    selected={this.props.form.dateRange.start}
+                    selected={startDate}
```
bshyong: 👍

src/js/health-records/utils/validations.js
```diff
@@ -0,0 +1,6 @@
+export function isValidDateRange(startDate, endDate) {
+  if (!startDate || !endDate) {
+    return true;
```
bshyong: Is this function supposed to return `true` if there is a null startDate or endDate?

bshyong (approved):
LGTM with minor question

src/js/health-records/utils/validations.js
U-DON: I think you originally wrote this function, but it made sense to me. If either is null, there's no need to check if it's a valid range since there's no comparison to be made.

PR #4949, opened 2017-02-23
Remove unused react-jsonschema-form components from bundle

jbalboni:
Related to [1313](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1313)

We're overwriting a lot of the built in widgets and fields from react-jsonschema-form, so this excludes them from the bundle.


config/webpack.config.js
```diff
@@ -106,6 +106,14 @@ const configGenerator = (options) => {
         {
           test: /\.json$/,
           loader: 'json-loader'
+        },
+        {
+          test: /react-jsonschema-form\/lib\/components\/(widgets|fields\/ObjectField|fields\/ArrayField)/,
```
annekainicUSDS: I'm a little confused what this is doing with `fields/ObjectField` and `fields/ArrayField`. I assume we want to not include those because we have our own versions? Is this excluding them? Why are they not in the `exclude` array below?

annekainicUSDS (commented):
One question for you!

config/webpack.config.js
jbalboni: Yeah, it's a little weird. The loader here sets files to null. So the `test` regex is telling you what files you want the loader to set to null (i.e. exclude). The `exclude` array specifies what files to exclude from being processed by the loader (i.e. set to null and excluded). So the `exclude` array is really specifying what files we still want to keep, even though they match the `test` regex.

annekainicUSDS: Ooo I see. Not intuitive, but when explained that way it makes sense, thanks!

annekainicUSDS (approved):
LGTM!

PR #4953, opened 2017-02-23
Touches up some styles on 1995

cvalarida:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1349

---
> Space above branch of service label should match other fields

image

---
> On the review page, the page edit button is touching the bottom border in Firefox

image

---
> The line spacing in a multi line label doesn't match the space between the label and a text field/area.  This makes fields like our education objective textarea look a little weird. We should even that spacing up.

The change I made affects the spacing of non-multi-line labels as well, but only for labels before tel / email / text `input`s, `textarea`s, and `select` boxes:

image

---
> Accordion styling on confirmation page should match normal accordion styles

image

image

---
> Capitalize "back" in button on confirmation page

image

### DIFF
--- a/src/js/edu-benefits/1995/containers/ConfirmationPage.jsx
+++ b/src/js/edu-benefits/1995/containers/ConfirmationPage.jsx
@@ -83,7 +83,7 @@ class ConfirmationPage extends React.Component {
             </li>
           </ul>
         </div>
-        <div id="collapsiblePanel" className="usa-accordion-bordered form-review-panel">
+        <div id="collapsiblePanel" className="usa-accordion-bordered">
           <ul className="usa-unstyled-list">
             <li>
               <div className="accordion-header clearfix">
--- a/src/js/edu-benefits/1995/containers/ConfirmationPage.jsx
+++ b/src/js/edu-benefits/1995/containers/ConfirmationPage.jsx
@@ -3,7 +3,6 @@ import moment from 'moment';
 import { connect } from 'react-redux';
 import Scroll from 'react-scroll';
 
-import ExpandingGroup from '../../../common/components/form-elements/ExpandingGroup';
 import { focusElement } from '../../../common/utils/helpers';
 
 import { benefitsLabels } from '../helpers';
--- a/src/js/edu-benefits/1995/containers/ConfirmationPage.jsx
+++ b/src/js/edu-benefits/1995/containers/ConfirmationPage.jsx
@@ -40,6 +40,17 @@ class ConfirmationPage extends React.Component {
     const name = form.veteranInformation.data.veteranFullName;
     const benefit = form.benefitSelection.data.benefit;
 
+    const docExplanation = this.state.isExpanded
+      ? (<div className="usa-accordion-content">
+        <p>In the future, you might need:</p>
+        <ul>
+          <li>Your reserve kicker</li>
+          <li>Documentation of additional contributions that would increase your monthly benefits.</li>
+        </ul>
+        <p>Documents can be uploaded using the <a href="https://gibill.custhelp.com/app/utils/login_form/redirect/account%252">GI Bill site</a>.</p>
+      </div>)
+      : null;
+
     return (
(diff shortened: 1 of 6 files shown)


src/sass/_normalize.scss
```diff
@@ -0,0 +1,12 @@
+// Ideally all of this will go in uswds/src/stylesheets/lib/normalize, but until
+//  it does, normalizing css goes in here.
+
+dl {
```
cvalarida: This didn't really seem to fit anywhere else, so I put it in a normalize scss file. It should go in uswds probably, but I don't know how to go about doing that, so for now, it's in here.

src/js/edu-benefits/1995/containers/ConfirmationPage.jsx
```diff
@@ -71,26 +82,27 @@ class ConfirmationPage extends React.Component {
...
-            <div>
-              <p>In the future, you might need:</p>
-              <ul>
-                <li>Your reserve kicker</li>
-                <li>Documentation of additional contributions that would increase your monthly benefits.</li>
-              </ul>
-              <p>Documents can be uploaded using the <a href="https://gibill.custhelp.com/app/utils/login_form/redirect/account%252">GI Bill site</a>.</p>
-            </div>
-          </ExpandingGroup>
+        <div id="collapsiblePanel" className="usa-accordion-bordered form-review-panel">
```
cvalarida: This section is copied from `src/js/common/schemaform/review/ReviewCollapsibleChapter.jsx`. That particular component isn't able to be used from the confirmation page, but I can pull this chunk out into a common component if desired.

src/sass/_normalize.scss
jbalboni: I think this is ok, but you should check the other uses of `<dl>` to make sure it looks ok (looks like there are only a few).

src/js/edu-benefits/1995/containers/ConfirmationPage.jsx
jbalboni: I'd like to keep the `form-review-panel` styles from leaking onto this page. You should be able to do an accordion with just the uswds styles (https://standards.usa.gov/components/accordions/)

src/sass/modules/_m-form-elements.scss
```diff
@@ -186,3 +186,12 @@ button.form-button-disabled {
     }
   }
 }
+
+// Even up the spacing between multi-line labels and their respective elements
+label {
```
jbalboni: Assuming Alex signs off on how this looks, I think we will just want to change this for the same inputs that the uswds styles set a top margin for.

jbalboni (commented):
Made a couple comments, but looks alright to me. Alex should take a look at the input changes and make sure they're ok.

jbalboni:
@goldenmeanie Can you look at the input spacing changes?

goldenmeanie:
Really nice improvement overall!

Date input looks like it partially escaped

screen_shot_2017-02-24_at_2_35_09_pm

cvalarida:
Removed `.form-review-panel` from the confirmation accordion:

src/sass/_normalize.scss
cvalarida: The only other place I saw a `<dl>` was in the [Rx GlossaryList](https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/js/rx/components/GlossaryList.jsx#L25). It should be fine since it's just normalizing. Takes what Chrome was already doing and making sure other browsers do the same.

src/sass/modules/_m-form-elements.scss
jbalboni: We should do the above and also move these styles to the edu-benefits styles, since this would affect all pages on vets.gov. I think that's what we want to do eventually, but Alex needs to run it by other people first.

jbalboni (approved):
Looks good, can be merged after my comment has been addressed.

PR #4955, opened 2017-02-24
1995 confirmation print

cvalarida:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1350

Printing now looks like the following screenshot of Chrome's print preview:

image


jbalboni (approved):
That looks way better

jbalboni:
@cvalarida Does this fix the problem on the 1990 form, too?

cvalarida:
@jbalboni Yup! Without filling any information in, the 1990 confirmation looks like this:

image

PR #4956, opened 2017-02-24
More consistent heading spacing, use h5 for block headings

jbalboni:
Related to [1352](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1352)

Changes around some spacing for the headers.

screen shot 2017-02-24 at 3 20 23 pm
screen shot 2017-02-24 at 3 20 17 pm
screen shot 2017-02-24 at 3 20 05 pm


src/sass/modules/_m-schemaform.scss
```diff
@@ -77,3 +81,36 @@
...
+  > legend, > h5, > p {
+    padding-bottom: 0 !important;
+    margin-bottom: 0 !important;
+  }
+  > p {
+    margin-top: 0.5em;
+  }
+}
+
+// This is the worst
```
jbalboni: I could not think of a better way to deal with this nonsense. Open to ideas.

cvalarida:
```css
.schemaform-block-header + .schemaform-first-field,
.schemaform-first-field--root .schemaform-first-field {
  label, .usa-input-error {
    margin-top: 0;
  }
}
```

Seems solid, but I'm not familiar with all the uses of this.

If we go this way, we won't even need `.schemaform-block-header--titled` up in `ObjectField` (and `AddressField`, probably).

cvalarida (commented):
It's maybe a bit wonky, but it certainly works. I left a suggestion on how we might simplify the CSS you mentioned, but other than that, I think it's good. We may need to figure out a more deterministic way to give the first visible (depth-first) label a specific class to reduce that margin later, though.

src/sass/modules/_m-schemaform.scss
jbalboni: Looks like this leaves a gap above the first field of pages with no title.

jbalboni: I updated these styles to (mostly) use specific classes, which hopefully makes it clearer. These are at least really specific and unlikely to cause problems later, aside from missing more deeply nested fields, which we can fix as we hit them.

PR #4957, opened 2017-02-24
Open edu benefit forms in new tabs

cvalarida:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1361

Opens the links to 22-1990 in a new tab. Also does the same for when we get the 1995 and friends out.


jbalboni (approved):
LGTM

PR #4960, opened 2017-02-25
1990e scaffolding

raquelromano-gov:
Relates to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1358


src/js/edu-benefits/1990e/helpers.jsx
```diff
@@ -0,0 +1,10 @@
+import { transformForSubmit } from '../../common/schemaform/helpers';
+
+export function transform(formConfig, form) {
```
raquelromano-gov: There aren't any transforms, but I'm guessing this function has to be defined for the review-and-submit page to be generated

src/js/edu-benefits/1990e/reducer.js
```diff
@@ -0,0 +1,7 @@
+import { combineReducers } from 'redux';
```
raquelromano-gov: reducer.js and routes.jsx are identical for the 1995 and 1990e, so do we want to move these into src/js/common?

src/js/edu-benefits/1990e/config/form.js
```diff
@@ -0,0 +1,59 @@
...
+    applicantInformation: {
+      title: 'Applicant Information',
+      pages: {
+        applicantInformation: {
+          path: 'applicant-information',
+          title: 'Applicant information'
+        }
+      }
+    },
+    benefitEligibility: {
```
raquelromano-gov: The 1995 uses "Benefit selection" on a similar page so may want to follow suit.

src/js/edu-benefits/1990e/reducer.js
jbalboni: Yeah, I think that makes sense for reducer.js.

I think we may break Webpack's code splitting if we try to move routes.jsx into common, because it looks for paths in `require` calls inside `require.ensure`. If we pass that path as a parameter to a function, I don't think it'll work correctly. I'm not positive on that, though, so might be worth trying.

src/js/edu-benefits/1990e/helpers.jsx
jbalboni: Yeah, the purpose of this is to transform the data into the body for the POST call. There's a default one, but it won't have the `educationBenefitsClaims` structure that the backend expects.

The review page should generate either way, but we do need at least this code for all the edu forms to POST data correctly. I'm not sure about the non-edu forms we're going to build, so I didn't try to generalize it more.

jbalboni (approved):
LGTM!

raquelromano-gov:
OK, will merge this and then look into where we can reduce code duplication.

PR #4961, opened 2017-02-27
Add tests for personal info chapter

jbalboni:
Related to [1320](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1320)

Adds tests for the form config for the personal information chapter.

### DIFF
--- a/test/edu-benefits/1995/config/contactInformation.unit.spec.jsx
+++ b/test/edu-benefits/1995/config/contactInformation.unit.spec.jsx
@@ -1,0 +1,172 @@
+import React from 'react';
+import { findDOMNode } from 'react-dom';
+import { expect } from 'chai';
+import sinon from 'sinon';
+import ReactTestUtils from 'react-addons-test-utils';
+import Form from 'react-jsonschema-form';
+
+import { DefinitionTester } from '../../../util/schemaform-utils.jsx';
+import formConfig from '../../../../src/js/edu-benefits/1995/config/form';
+
+describe('Edu 1995 contactInformation', () => {
+  const { schema, uiSchema } = formConfig.chapters.personalInformation.pages.contactInformation;
+  it('should render', () => {
+    const form = ReactTestUtils.renderIntoDocument(
+      <DefinitionTester
+          schema={schema}
+          data={{}}
+          uiSchema={uiSchema}/>
+    );
+
+    const inputs = ReactTestUtils.scryRenderedDOMComponentsWithTag(form, 'input').concat(
+      ReactTestUtils.scryRenderedDOMComponentsWithTag(form, 'select')
+    );
+
+    expect(inputs.filter(input => input.id.startsWith('root_preferredContactMethod')).length)
+      .to.equal(3);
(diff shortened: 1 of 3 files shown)


test/edu-benefits/1995/config/contactInformation.unit.spec.jsx
```diff
@@ -0,0 +1,130 @@
...
+    );
+
+    expect(inputs.filter(input => input.id.startsWith('root_preferredContactMethod')).length)
+      .to.equal(3);
+    expect(inputs.filter(input => input.id.startsWith('root_veteranAddress')).length)
+      .to.equal(6);
+    expect(inputs.filter(input => input.id.startsWith('root_view:otherContactInfo')).length)
+      .to.equal(4);
+  });
+  it('should render required fields', () => {
```
raquelromano-gov: Maybe rename to 'should render validation errors for required fields'?

test/edu-benefits/1995/config/contactInformation.unit.spec.jsx
```diff
@@ -0,0 +1,130 @@
...
+      <DefinitionTester
+          formData={{}}
+          schema={schema}
+          data={{}}
+          uiSchema={uiSchema}/>
+    );
+
+    const formDOM = findDOMNode(form);
+    const inputs = ReactTestUtils.scryRenderedDOMComponentsWithTag(form, 'input');
+    const emailInput = inputs.find(input => input.id.endsWith('email'));
```
raquelromano-gov: Is this case-sensitive? i.e., is it guaranteed to find the email input and not the confirmationEmail input (may work even if case-insensitive since email is first on the page, but that would be kind of brittle)

test/edu-benefits/1995/config/contactInformation.unit.spec.jsx
```diff
@@ -0,0 +1,130 @@
...
+    );
+
+    const formDOM = findDOMNode(form);
+    ReactTestUtils.findRenderedComponentWithType(form, Form).onSubmit({
+      preventDefault: f => f
+    });
+
+    expect(formDOM.querySelectorAll('.usa-input-error').length).to.equal(6);
+  });
+  it('should not render confirm email on review page', () => {
```
raquelromano-gov: I hadn't thought explicitly about review page behavior when I was writing tests, but in general maybe we should group these at the end so they're easier to maintain.

raquelromano-gov (commented):
LGTM (only a couple of minor suggestions/questions)

test/edu-benefits/1995/config/contactInformation.unit.spec.jsx
jbalboni: That's a good suggestion. A lot of these pages won't have have review specific behavior to test, for what it's worth.

test/edu-benefits/1995/config/contactInformation.unit.spec.jsx
jbalboni: Yeah, I shouldn't rely on this being case sensitive.

PR #4967, opened 2017-02-27
Edu 1995 school unit tests

cvalarida:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1319

They pass. Yay. If we need more tests, I can certainly fill them out, but this seemed to be in line with the amount of testing in the other schemaform unit tests.

### DIFF
--- a/test/edu-benefits/1995/config/schoolSelection-newSchool.unit.spec.jsx
+++ b/test/edu-benefits/1995/config/schoolSelection-newSchool.unit.spec.jsx
@@ -23,7 +23,7 @@ describe('Edu 1995 schoolSelection', () => {
       .to.not.be.empty;
   });
 
-  it('should require school name and education type', () => {
+  it('should require school only name and education type', () => {
     const form = ReactTestUtils.renderIntoDocument(
       <DefinitionTester
           formData={{}}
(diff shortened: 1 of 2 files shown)


test/edu-benefits/1995/config/schoolSelection-newSchool.unit.spec.jsx
```diff
@@ -0,0 +1,79 @@
...
+    ReactTestUtils.findRenderedComponentWithType(form, Form).onSubmit({
+      preventDefault: f => f
+    });
+
+    // Expect these fields to require validation
+    expect(formDOM.querySelector('.usa-input-error #root_newSchoolName')).to.not.be.null;
+    expect(formDOM.querySelector('.usa-input-error #root_educationType')).to.not.be.null;
+  });
+
+  it('should have no errors with only required info filled in', () => {
```
jbalboni: I think we should flip this around and test that we get the right number of required errors if you submit without filling anything in. This scenario should be covered by the e2e tests.

test/edu-benefits/1995/config/schoolSelection-newSchool.unit.spec.jsx
```diff
@@ -0,0 +1,79 @@
...
+import { findDOMNode } from 'react-dom';
+import { expect } from 'chai';
+import sinon from 'sinon';
+import ReactTestUtils from 'react-addons-test-utils';
+import Form from 'react-jsonschema-form';
+
+import { DefinitionTester } from '../../../util/schemaform-utils.jsx';
+import formConfig from '../../../../src/js/edu-benefits/1995/config/form';
+
+describe('Edu 1995 schoolSelection', () => {
```
jbalboni: Should also have a test that checks that the Address fields are shown when you pick the right education type.

jbalboni (commented):
Looks good, added some test suggestions.

test/edu-benefits/1995/config/schoolSelection-newSchool.unit.spec.jsx
cvalarida: I thought about that one, but isn't that going away? Or was what I'm thinking the bit where we conditionally add an option to the education type?

jbalboni: You're thinking of the additional option in the dropdown, which has already been removed. This is staying.

jbalboni (approved):

PR #4969, opened 2017-02-27
End-of-training date validation

cvalarida:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1380

Now it only throws a validation error if the date is 100 years before or after the current date.

*Note:* This also means that incomplete dates won't throw a validation error. The following is perfectly valid as far as validation goes.

---
image


jbalboni (approved):
This looks good to me. I'd like to get the question text updated in here, too.

goldenmeanie:
> When did you stop taking classes or participating in the training program? (Future dates are ok)

PR #4971, opened 2017-02-28
Adds unit test for Address

cvalarida:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1093

### DIFF
--- a/test/common/schemaform/definitions/address.unit.spec.jsx
+++ b/test/common/schemaform/definitions/address.unit.spec.jsx
@@ -74,6 +74,15 @@ describe('Schemaform definition address', () => {
     // jsdom doesn't support namedItem; use options instead
     expect(Array.from(stateField.options).find(op => op.value === 'OR')).to.not.be.undefined;
 
+    // Entering a military city should result in different "state" options
+    const cityField = formDOM.querySelector('#root_city');
+    ReactTestUtils.Simulate.change(cityField, {
+      target: {
+        value: 'apo'
+      }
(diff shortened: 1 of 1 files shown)


test/common/schemaform/definitions/address.unit.spec.jsx
```diff
@@ -0,0 +1,94 @@
...
+    // By default, the Country is USA, so the postalCode label should be 'ZIP Code'
+    //  and the State label should be 'State' and the field should be a dropdown
+    const labels = ReactTestUtils.scryRenderedDOMComponentsWithTag(form, 'label');
+    const postalCodeLabel = labels.find(label => label.htmlFor === 'root_postalCode');
+    const stateLabel = labels.find(label => label.htmlFor === 'root_state');
+    const stateField = formDOM.querySelectorAll('#root_state');
+
+    expect(postalCodeLabel.textContent === 'ZIP Code');
+    expect(stateLabel.textContent === 'State');
+    expect(stateField.tagName === 'select');
```
jbalboni: It would also be good to test that the state select box options are constrained correctly, depending on country. You could probably just check for particular states for CAN, USA, and MEX.

Also should test that setting a military city (APO/FPO) results in the military state codes being shown in the select box.

jbalboni (approved):
Looks good, had a couple suggestions for more checks. This item was originally to test the Address component itself, but I think testing the definition overall will hit all of the logic there.

PR #4977, opened 2017-02-28
Add military service tests for 1995, fix array field issue

jbalboni:
Related to [1318](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1318)

Adds military service page tests and fixes a bug on the service periods page. The bug is that if you click "Add Another" without touching any fields, there's an error. I've disabled the button for now when that's the case, because it doesn't really make much sense to add any empty period.


test/common/schemaform/ArrayField.unit.spec.jsx
```diff
@@ -196,4 +196,42 @@ describe('Schemaform <ArrayField>', () => {
...
+    const onBlur = sinon.spy();
+    const tree = SkinDeep.shallowRender(
+      <ArrayField
+          schema={schema}
+          errorSchema={errorSchema}
+          uiSchema={uiSchema}
+          idSchema={idSchema}
+          registry={registry}
+          onChange={onChange}
+          onBlur={onBlur}
```
cvalarida: I was wondering why these were necessary in this test, so looked in `ArrayField`'s props to see if they were required. `onChange` is, but `onBlur` never made it into `ArrayField.propTypes` at all. Just so you're aware.

test/edu-benefits/1995/config/servicePeriods.unit.spec.jsx
```diff
@@ -0,0 +1,104 @@
...
+  it('should render service fields', () => {
+    const form = ReactTestUtils.renderIntoDocument(
+      <DefinitionTester
+          schema={schema}
+          data={{}}
+          uiSchema={uiSchema}
+          definitions={definitions}/>
+    );
+    const formDOM = findDOMNode(form);
+    const newServiceInput = Array.from(formDOM.querySelectorAll('input[type=radio]'))[0];
```
cvalarida: Wouldn't `formDOM.querySelector('input[type=radio')` do the same? [querySelector](https://developer.mozilla.org/en-US/docs/Web/API/Element/querySelector) just gets the first element that matches the query.

test/edu-benefits/1995/config/servicePeriods.unit.spec.jsx
```diff
@@ -0,0 +1,104 @@
...
+          uiSchema={uiSchema}
+          definitions={definitions}/>
+    );
+    const formDOM = findDOMNode(form);
+    ReactTestUtils.findRenderedComponentWithType(form, Form).onSubmit({
+      preventDefault: f => f
+    });
+    expect(Array.from(formDOM.querySelectorAll('.usa-input-error'))).to.be.empty;
+
+    ReactTestUtils.findRenderedComponentWithType(form, Form).onSubmit({
```
cvalarida: Could extract the `Form` finding out since we do it multiple times.

```js
const formElement = ReactTestUtils.findRenderedComponentWithType(form, Form);
formElement.onSubmit({ preventDefault: f => f });
```

cvalarida (approved):
Just a couple things to make mention of, but it looks good overall...as soon as the Jenkins build goes through.

test/edu-benefits/1995/config/servicePeriods.unit.spec.jsx
jbalboni: Yeah, should be the same. I thought this might be clearer, but probably not by much.

PR #4979, opened 2017-03-01
Add PHR refresh flow in Health Records app

bshyong:
Putting this PR up so that we can review and fix any structural/approach issues. This is mergeable. However, there are some potential error cases that we may want to catch at some point. Noting them here for reference.

There is also an open question around what constitutes a successful refresh. Currently, the definition is no ERROR statuses, but we presumably need to check the last_updated dates too.

- calling the GET health-records/refresh endpoint fails. currently not handling this because it will be called every 10 seconds
- calling the POST /health_records endpoint to generate a report fails. this is currently unhandled. If a refresh is successful, put the generation of the record fails, what should the UX look like? Should the user be allowed to retry submitting the form, or go to the start of the flow to fill out the form again?
- calling the GET /health_records?doc_type= to download a report fails. this is currently unhandled. Potentially need to handle by implementing an error state/messaging so that a user can retry the download.

Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1102


src/js/health-records/containers/UpdatePage.jsx
```diff
@@ -0,0 +1,83 @@
...
+
+  handleSkipToDownload(e) {
+    e.preventDefault();
+    this.props.submitForm(JSON.parse(sessionStorage.getItem('hr-form')));
+    this.context.router.push('/download');
+  }
+
+  render() {
+    const statuses = this.props.refresh.statuses;
+    const completionPercentage = statuses.OK.length / (statuses.ERROR.length + statuses.OK.length) * 100;
```
U-DON: Are these the proper metrics for the progress bar? Would it be more accurate to track the percentage of data with the last updated date being after the submission?

src/js/health-records/containers/UpdatePage.jsx
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  componentDidMount() {
+    this.props.checkRefreshStatus();
+    this.pollRefresh();
+  }
+
+  componentWillReceiveProps(nextProps) {
+    const erroredUpdates = nextProps.refresh.statuses.ERROR;
+    if (erroredUpdates.length === 0) {
```
U-DON: Is it possible for the first refresh check to have no errors in the returned data? What if the first response is not up-to-date, but has status 'OK' on all of the data?

src/js/health-records/reducers/refresh.js
```diff
@@ -0,0 +1,22 @@
...
+    OK: [],
+    ERROR: [],
+  },
+};
+
+export default function refresh(state = initialState, action) {
+  switch (action.type) {
+    case 'REFRESH_POLL_SUCCESS': {
+      // returns group in form {'OK': [], 'Error': []}
+      const statuses = _.groupBy(action.data, 'attributes.status');
```
U-DON: Since it seems we're only looking at the lengths of the groups of statuses, would it make more sense to use `_.countBy`? Is the data going to be used for something else, maybe error handling?

U-DON (commented):
Looks good as an intermediate step, but certainly there are quite a few open questions about error cases and how they should be handled.

src/js/health-records/containers/UpdatePage.jsx
bshyong: Yes, this is a good point. I will follow up with @saneshark on what the success state looks like. Presumably the status should be OK, AND all of the dates should change to the date of the request. Right now this only checks for the statuses.

src/js/health-records/containers/UpdatePage.jsx
bshyong: similar situation as above

src/js/health-records/reducers/refresh.js
bshyong: I left this as `groupBy` in case we wanted to show which refreshes failed out. It is only possible to refresh once per day, so if one of the updates fails, there is no way to refresh that data again until the next day.

Given this limitation, we can potentially add improved messaging down the line that specifies which part of the update failed, and let the veteran make a choice on whether to proceed.

bshyong:
Made some changes in the reducer to check for success or failure of the updates.

PR #4982, opened 2017-03-01
Edu 1995 dev resubmit

cvalarida:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1314

This doesn't add a control to reset the submission status, but it does make the "Send Failed" button clickable to resubmit the application on dev only.

If the submission is successful in a subsequent attempt, we'll get redirected to the confirmation page like normal. On a failed attempt, it will seem like nothing happened, but if you check the network log, you'll see it did try.


jbalboni (approved):
LGTM

PR #4986, opened 2017-03-01
Add OMB info to 1990

cvalarida:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1257

It's a little...smushed against the button, but it seemed a bit overkill to make a whole new class to space it out a bit more (not sure if there's something already written I could use offhand) and there's rather a lot of space around as it is.

image


jbalboni (approved):
LGTM, assuming Alex is good with the spacing. I deleted my other PR with the `g` fix.

goldenmeanie:
I don't know where on the stack, converting the 1990 to the new FML should even out the spacing the rest of the way

PR #4989, opened 2017-03-01
Separate out uploader library

jbalboni:
Related to [1377](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1377)

This puts the require call for fine-uploader inside a `require.ensure` call, which will cause Webpack to put it in a separate bundle. This keeps it out of the bundles for the rest of our apps, since we use a common store and the action files for many of the apps are pulled in for every application.

### DIFF
--- a/src/js/disability-benefits/actions/index.js
+++ b/src/js/disability-benefits/actions/index.js
@@ -1,4 +1,3 @@
-import { FineUploaderBasic } from 'fine-uploader/lib/core';
 import environment from '../../common/helpers/environment';
 import { makeAuthRequest } from '../utils/helpers';
 
@@ -18,6 +17,7 @@ export const ADD_FILE = 'ADD_FILE';
 export const REMOVE_FILE = 'REMOVE_FILE';
 export const SUBMIT_FILES = 'SUBMIT_FILES';
 export const SET_UPLOADING = 'SET_UPLOADING';
+export const SET_UPLOADER = 'SET_UPLOADER';
 export const DONE_UPLOADING = 'DONE_UPLOADING';
 export const SET_PROGRESS = 'SET_PROGRESS';
 export const SET_UPLOAD_ERROR = 'SET_UPLOAD_ERROR';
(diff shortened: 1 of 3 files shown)


src/js/disability-benefits/actions/index.js
```diff
@@ -163,94 +163,104 @@ export function submitFiles(claimId, trackedItem, files) {
...
-      uploader
     });
     dispatch({
       type: SET_PROGRESS,
-      progress: filesComplete / files.length
+      progress: 0
     });
+    require.ensure([], (require) => {
+      const { FineUploaderBasic } = require('fine-uploader/lib/core');
+      const uploader = new FineUploaderBasic({
```
jbalboni: None of the code in here has changed, it's just inside a another function now.

src/js/disability-benefits/actions/index.js
```diff
@@ -163,94 +163,104 @@ export function submitFiles(claimId, trackedItem, files) {
...
     dispatch(clearNotification());
     dispatch({
       type: SET_UPLOADING,
       uploading: true,
-      uploader
     });
     dispatch({
       type: SET_PROGRESS,
-      progress: filesComplete / files.length
+      progress: 0
```
cvalarida: This is just for the initial state, correct? Why wasn't it set to 0 before? Looks good, but this is the one part I can't test (the actual file upload and by extension, the progress).

cvalarida (approved):
Looks good.

src/js/disability-benefits/actions/index.js
jbalboni: It would have effectively been 0 before (since `filesComplete` would be 0).

jbalboni:
I'm working on setting up my VA laptop to actually test this.

rnicholus:
Fine Uploader developer here - if the VA ever has any questions or concerns re my library, or anything related to uploading files as well, I'm more than happy to help out. Just throwing that out there.

jbalboni:
@rnicholus Thanks! This will likely the be the last bit of work around file uploading on this part of vets.gov for a while, but I will keep that in mind!

PR #1245, opened 2016-02-26
Create validations javascript.

awong-dev:
Stub out the file so our group can collaboratively hack.

PR #4992, opened 2017-03-02
Avoid npm run build requirement for e2e tests

alexose-gov:
Fixes [#1245](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1245).

This PR allows developers to run the e2e suite on top of `webpack-dev-server` (aka `npm run watch`).

The neat part about this is that you no longer have to run a full build each time you want to run e2e tests!  As long as your dev server is running, you will be testing against the most recent code under `/src`.

One thing to note is that I've changed the mock API to use port 3000.  This simplifies things a bit at the expense of developers who want to run `vets-api` locally along with the e2e tests.  If anyone finds this objectionable, please make a note of it-- I can definitely approach the issue in another way if needed.

### DIFF
--- a/script/run-nightwatch.sh
+++ b/script/run-nightwatch.sh
@@ -9,8 +9,14 @@ trap 'kill $(jobs -p)' EXIT
 
 BUILDTYPE=${BUILDTYPE:-development}
 
-# Run the api server
-node src/test-support/mockapi.js &
+# Check to see if we already have an API server running on port 3000
+if [ `nc -z localhost 3000; echo $?` -ne 0 ]; then
+  echo "Starting mockapi.js..."
+  node src/test-support/mockapi.js &
+else
+  echo "Error: Port 3000 is already in use.  If you're sure that's OK, tests will continue in 5 seconds..."
+  sleep 5;
+fi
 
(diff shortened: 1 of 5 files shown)


script/run-nightwatch.sh
```diff
@@ -9,13 +9,28 @@ trap 'kill $(jobs -p)' EXIT
 
 BUILDTYPE=${BUILDTYPE:-development}
 
-# Run the api server and the webserver.
-node src/test-support/mockapi.js &
-node src/test-support/test-server.js --buildtype ${BUILDTYPE} &
+# Check to see if we already have an API server running on port 3000
+if [ `nc -z localhost 3000; echo $?` -ne 0 ]; then
```
robbiethegeek-usds: this is hard coded to port 3000, and the rest of this process can be overwritten by an ENV variable.

robbiethegeek-usds (commented):
What if someone uses a different port for the API via environment variable.

jbalboni (approved):
LGTM. Worked locally without even a watch task restart!

robbiethegeek-usds:
Thanks @alexose-gov @alexose

PR #4993, opened 2017-03-02
GIBCT Calculator component [WIP]

ben-USDS-damman:
No reviews yet please; just for visibility into the work.


src/js/gi/components/AccordionItem.jsx
```diff
@@ -18,7 +18,7 @@ class AccordionItem extends React.Component {
     const expanded = this.state.expanded;
     return (
       <li>
-        <button onClick={this.toggle()} className="usa-accordion-button" aria-expanded={expanded} aria-controls="amendment-1">
+        <button onClick={this.toggle} className="usa-accordion-button" aria-expanded={expanded} aria-controls="amendment-1">
```
saneshark: Not sure what best practice is around this, but this looks almost like inline javascript. Is it intended? Wouldn't it be better to do some kind of live binding for a button having this class name?

src/js/gi/selectors/calculator.js
```diff
@@ -9,15 +9,764 @@ const getEligibilityDetails = (state) => {
...
+    numberOfTerms,
+    tuitionNetPrice,
+    housingStipdendTerm1,
+    housingStipdendTerm2,
+    housingStipdendTerm3,
+    housingStipdendTotal,
+    tuitionAllowTerm1,
+    tuitionAllowTerm2,
+    tuitionAllowTerm3,
+    tuitionAllowTotal,
```
saneshark: are these liable to change? I'm wondering if maybe these should be returned as a constant or something from the backend. Just a thought -- obviously for a future enhancement.

src/js/gi/components/AccordionItem.jsx
typesend: Nope. Short answer: This is the React way.

src/js/gi/selectors/calculator.js
typesend: They are not constants. They are computed based on calculator form inputs and data from the backend (school attrs and constants). Basically what you are suggesting is akin to making the calculator a server-side activity requiring a round trip every time the user changes a form input.

U-DON:
Closing this to open an intermediate PR.

PR #4994, opened 2017-03-02
Move off of react-jsonschema-form library fork

jbalboni:
Related to [983](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/983)

This moves us back the normal version of react-jsonschema-form. The main change for this is to remove the `touchedSchema` code I had added to handle showing validation errors after fields have been touched. I've changed it back to keeping that information in formContext. The downside to this is that blurring a field will force a re-render of most of the page, but this shouldn't be noticeable even on slower phones.

There are also some new unit tests in here because I noticed some gaps when looking at code coverage.


src/js/edu-benefits/1995/config/form.js
```diff
@@ -152,7 +152,8 @@ const formConfig = {
               },
               items: {
                 serviceBranch: {
-                  'ui:title': 'Branch of service'
+                  'ui:title': 'Branch of service',
+                  'ui:required': () => true
```
cvalarida: Curiosity: When did this become required?

cvalarida (commented):
I'm not super familiar with the `touchedSchema` and how that was working before, but it still works, so I think it's good to go as long as that was the only significant change in the fork.

I did confirm that it re-renders the entire form page `onBlur`, but I think that's fine. If we find it performs poorly, we can probably spam `shouldComponentUpdate` or something to improve that.

src/js/edu-benefits/1995/config/form.js
jbalboni: When I was testing I forgot to remove it?

jbalboni:
The other changes in the fork were ones that I submitted PRs to the original library for, so that should be it.

PR #4995, opened 2017-03-02
Edu 5490 scaffolding

cvalarida:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1367

Creates an empty form for the 5490.


src/js/edu-benefits/5490/helpers.jsx
```diff
@@ -0,0 +1,48 @@
+import React from 'react';
```
jbalboni: Can we delete everything but the transform in here? If we need to reuse any of the other data, it should be in a file that's shared between the forms.

src/js/edu-benefits/5490/components/ServicePeriodView.jsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
```
jbalboni: If this is the same as the 1990, we'll probably want to share it. So maybe remove it for now.

jbalboni (approved):
LGTM. Made a couple comments.

src/js/edu-benefits/5490/helpers.jsx
cvalarida: Ah, yes. I didn't even think about these. Thanks!

PR #4996, opened 2017-03-02
Fix click target for radio

jbalboni:
Related to [1437](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1437)

The margin change we made the other day is pushing down the clickable area for radio buttons. This removes the margin for radio buttons (which doesn't affect anything visually).

You can tests this by trying to click on the upper half of the radio button circle.


cvalarida (approved):
LGTM

PR #5002, opened 2017-03-02
Adds GA code and conditional to prevent child window from opening unless url is present

crwallace:


robbiethegeek-usds (approved):
:shipit: YOU ARE THE 💣

PR #5005, opened 2017-03-02
Add a spinner to forms that are lazily loading

jbalboni:
Related to [1110](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1110)

This adds a spinner while the lazily loaded form bundles are loading. This isn't noticeable in dev unless you add a timeout, but it does show up in staging some.

This will become more useful when we do multiform, since this spinner will be the feedback a user gets when they continue to a specific form.

screen shot 2017-03-02 at 4 39 43 pm

### DIFF
--- a/src/js/common/components/asyncLoader.jsx
+++ b/src/js/common/components/asyncLoader.jsx
@@ -1,3 +1,4 @@
+// Adapted from https://gist.github.com/acdlite/a68433004f9d6b4cbc83b5cc3990c194
 import React from 'react';
 import LoadingIndicator from '../../common/components/LoadingIndicator';
 
--- a/src/js/edu-benefits/components/asyncLoader.jsx
+++ b/src/js/edu-benefits/components/asyncLoader.jsx
@@ -1,0 +1,29 @@
+import React from 'react';
+import LoadingIndicator from '../../common/components/LoadingIndicator';
+
+export default function asyncLoader(getComponent, message) {
+  return class AsyncComponent extends React.Component {
+    static Component = null;
+    state = { Component: AsyncComponent.Component };
+
+    componentWillMount() {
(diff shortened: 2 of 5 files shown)


jbalboni:
FYI @goldenmeanie

src/js/common/components/asyncLoader.jsx
```diff
@@ -0,0 +1,30 @@
+// Adapted from https://gist.github.com/acdlite/a68433004f9d6b4cbc83b5cc3990c194
+import React from 'react';
+import LoadingIndicator from '../../common/components/LoadingIndicator';
+
+export default function asyncLoader(getComponent, message) {
+  return class AsyncComponent extends React.Component {
+    static Component = null;
```
cvalarida: Why static? I feel like this could come back to bite us later...If we're lazy-loading sibling or parent-child components, perhaps.

Then again, it doesn't look like the static variable ever gets changed, but if not, why not just set the initial
```js
state = { Component: null };
```
?

cvalarida (commented):
Very nice! I only have one concern (which may or may not be valid) noted below.

src/js/common/components/asyncLoader.jsx
jbalboni: It won't cause any problems because we're creating a new class each time this is used, so two uses of asyncLoader would never share the same class definition.

That said, I'm not entirely sure what the point of that is. I'll remove it.

jbalboni: Actually, I think the use case is for re-visiting a route. It's caching the component that the promise returns so that if you come back to a route it doesn't have to render twice.

That doesn't really matter for us, but I'm going to leave it so this stays generic enough for other uses.

PR #5008, opened 2017-03-03
Incremental builds for Jenkins

alexose-gov:
This PR allows for incremental builds in our CI environment.  It's the continuation of a whole bunch o' work I did a couple months ago, but never finished.

It relies on the [hard-source-plugin](https://github.com/mzgoddard/hard-source-webpack-plugin) to write the build cache to disk.  Normally, webpack can't do this because it stores a bunch of circular references in memory that can't be easily serialized.

You can test it locally by running `npm run build` twice in a row.  The second build should only take a few seconds, as it is actually grabbing the previous build from disk.

While everything works locally, it remains to be seen how this works with deployments and such.  It seems like the most likely thing is that the cache doesn't invalidate itself properly and we end up with a stale build.


ayaleloehr:
Is this something that Core will plan to complete @robbiethegeek ?

robbiethegeek-usds:
@ayaleloehr let's chat about this today in our 1 on 1.

robbiethegeek-usds:
Closing this down and opened a new ticket:
https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1844

PR #5010, opened 2017-03-03
GIBCT search displays loading indicator during fetch and updates URL query params.

U-DON:
## Changes
* Update URL accordingly with form interactions. Directly going to URL with the query params will pre-fill form inputs.
* `InstitutionFilterForm` downgraded from container to regular component. Receives props from parent `SearchPage`. DOM events are now captured at component level, and relevant data is passed up to the parent.
* Refactored action for updating filter state to update multiple params at once based on the query params.
* Fixed bug where no results for program filters shows `NaN` for the count in the label.
* Loading indicator when search is in progress.


src/js/gi/containers/SearchPage.jsx
```diff
@@ -10,22 +18,97 @@ import Pagination from '../../common/components/Pagination';
...
+
+    const shouldRemoveFilter =
+      (queryKey !== 'caution' && !queryValue) ||
+      (queryKey === 'caution' && queryValue) ||
+      ((queryKey === 'country' ||
+        queryKey === 'state' ||
+        queryKey === 'typeName') && queryValue === 'ALL');
+
+    if (shouldRemoveFilter) { delete query[queryKey]; }
+    this.props.router.push({ ...this.props.location, query });
```
U-DON: In this handler, form input events trigger a URL query update, which invokes `updateSearchResults` from `componentDidUpdate` to actually update the state of the form and also `fetchSearchResults`.

ben-USDS-damman: 👍

ben-USDS-damman (approved):
LGTM. I assume adding the keyword search param is up next.

ben-USDS-damman:
Also want to add that the loading graphic should really appear where the search results do and the filter controls should remain. Currently the cycle of everything-disappears-then-reappears is quite jarring.

ben-USDS-damman:
#testingGithub Ohai @department-of-veterans-affairs/vets-gov-engineers

PR #5013, opened 2017-03-03
Update schema and add benefits page to 5490

jbalboni:
Related to [693](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/693)

Update the schema package so we have the 5490 schema and adds the benefit selection page.

screen shot 2017-03-03 at 2 48 26 pm

### DIFF
--- a/src/js/edu-benefits/5490/helpers.jsx
+++ b/src/js/edu-benefits/5490/helpers.jsx
@@ -8,3 +8,8 @@ export function transform(formConfig, form) {
     }
   });
 }
+
+export const benefitsLabels = {
+  chapter35: 'DEA (Chapter 35)',
+  chapter33: 'Fry scholarship (Chapter 33)'
+};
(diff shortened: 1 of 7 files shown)


src/js/edu-benefits/5490/helpers.jsx
```diff
@@ -8,3 +8,8 @@ export function transform(formConfig, form) {
     }
   });
 }
+
+export const benefitsLabels = {
+  chapter35: 'DEA (Chapter 35)',
+  chapter33: 'Fry scholarship (Chapter 33)'
```
cvalarida: Are we going to add links to "Learn More" like we did with the others? I assume we don't have the content for those right now, but is that the plan?

cvalarida (approved):
Looks good to me!

src/js/edu-benefits/5490/helpers.jsx
jbalboni: I would guess so, but yeah, we don't have the content for them as far as I know.

PR #5015, opened 2017-03-03
Add benefit selection to 1990e

cvalarida:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/631

---
image


jbalboni:
@cvalarida Needs a screenshot, too.

src/js/edu-benefits/1990e/config/form.js
```diff
@@ -25,6 +39,31 @@ const formConfig = {
     benefitEligibility: {
       title: 'Benefit Eligibility',
       pages: {
+        benefitEligibility: {
+          path: 'benefit-eligibility',
+          title: 'Benefit Eligibility',
+          uiSchema: {
+            'view:benefitDescription': {
+              'ui:description': eligibilityDescription
```
jbalboni: You could try putting this just on the uiSchema object:

```js
uiSchema: {
  'ui:description': eligibilityDescription,
  benefit: {
    ...
  }
}
```

cvalarida:
Right! Updated with the screenshot. Thanks, @jbalboni.

src/js/edu-benefits/1990e/helpers.jsx
```diff
@@ -8,3 +10,24 @@ export function transform(formConfig, form) {
...
+          <li>You can only get payments from 1 program at a time.</li>
+          <li>You can’t get more than 48 months of benefits under any combination of VA education programs.</li>
+        </ul>
+      </div>
+    </div>
+  );
+}
+
+export const benefitsLabels = {
+  chapter33: <p>Post-9/11 GI Bill (Chapter 33)<br/><a href="/education/gi-bill/post-9-11/" target="_blank">Learn more</a></p>,
```
jbalboni: Are these the same as the 1995? Maybe we should put them in a helpers file in edu-benefits/utils that both forms could share?

jbalboni (commented):
Looks good. I made two small comments. You should also add a test for this page.

jbalboni:
@cvalarida You can do the schema update in this PR if you'd like. You'll have to merge from master first, since I had to fix something on the 1995 for it to work with the most recent schema change.

cvalarida:
With master merged, the schema updated, benefits labels shared, and tests added, can you take another look, @jbalboni? Should be good, but I like the assurance of double-checking.

src/js/edu-benefits/utils/helpers.js
```diff
@@ -14,6 +15,15 @@ export const chapterNames = {
   review: 'Review'
 };
 
+export const benefitsLabels = {
+  chapter33: <p>Post-9/11 GI Bill (Chapter 33)<br/><a href="/education/gi-bill/post-9-11/" target="_blank">Learn more</a></p>,
```
raquelromano-gov: I'm surprised these don't need to be surrounded by quotation marks, although it seems to have worked before. What happens if you add quotes?

src/js/edu-benefits/1990e/config/form.js
jbalboni: @cvalarida Did this not work?

src/js/edu-benefits/utils/helpers.js
jbalboni: This file should have a jsx extension now.

test/edu-benefits/1990e/config/benefitSelection.unit.spec.jsx
```diff
@@ -0,0 +1,41 @@
...
+          schema={schema}
+          onSubmit={onSubmit}
+          data={{}}
+          uiSchema={uiSchema}/>
+    );
+    const formDOM = findDOMNode(form);
+    submitForm(form);
+    expect(Array.from(formDOM.querySelectorAll('.usa-input-error'))).to.be.empty;
+
+    submitForm(form);
```
jbalboni: I don't think you need to call this twice.

jbalboni (approved):
Couple comments, but looks good.

src/js/edu-benefits/utils/helpers.js
cvalarida: I tried giving it `.jsx`, but it broke in the 1990 routes, so I changed the import statement there and it broke in two more places that...I couldn't even trace. Funky stuff was going on...I'll try it again, though.

cvalarida: Hmm. Now it works just fine. Guess I'll go with it.

cvalarida: Turns out I just need to restart `npm run watch` after files are renamed...which is odd, but that showed me the proper files that specifically referred to `helpers.js` (with the extension).

jbalboni: @raquelromano-gov The lack of quotes means that value is interpreted as jsx, which is compiled into function calls by Babel, like Chris mentioned.

By convention, we name files `.jsx` if they contain any jsx, but we're transpiling both `.js` and `.jsx` files.

PR #5017, opened 2017-03-03
Add personal info chapter for 5490

jbalboni:
Related to [723](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/723)

This adds the three pages for the personal information chapter to the form. The contact information page is the same as the 1995, so I've moved that up and shared it between the two forms. The direct deposit page is sharing the bankAccount definition, and the secondary contact page is new.

Contact info:
screen shot 2017-03-03 at 5 33 56 pm
screen shot 2017-03-03 at 5 34 01 pm

Secondary contact:
screen shot 2017-03-03 at 4 48 01 pm
screen shot 2017-03-03 at 4 49 31 pm

Direct deposit:
screen shot 2017-03-03 at 5 33 03 pm

### DIFF
--- a/src/js/edu-benefits/1995/config/form.js
+++ b/src/js/edu-benefits/1995/config/form.js
@@ -2,11 +2,9 @@ import _ from 'lodash/fp';
 
 import fullSchema1995 from 'vets-json-schema/dist/change-of-program-schema.json';
 
-import { validateMatch } from '../../../common/schemaform/validation';
 import {
   benefitsLabels,
   bankAccountChangeLabels,
-  preferredContactMethodLabels,
   transform,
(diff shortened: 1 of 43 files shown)


cvalarida (approved):
LGTM

src/js/edu-benefits/1995/config/form.js
```diff
@@ -34,7 +32,6 @@ const {
   educationObjective,
   nonVaAssistance,
   reasonForChange,
-  email,
```
jbalboni: It's still being used in the personal info definition. The schema definition for it is just a string with a format of `email`, so it's not a shared definition in vets-json-schema.

PR #5021, opened 2017-03-06
Show Login Failure Message if Login Attempt Fails

omgitsbillryan:
This little bug was _executing_ `window.close()` instead of passing the function as an object to `onClick`

### DIFF
--- a/src/js/auth/containers/AuthApp.jsx
+++ b/src/js/auth/containers/AuthApp.jsx
@@ -84,7 +84,7 @@ class AuthApp extends React.Component {
         <div>
           <h3>We are sorry that we could not successfully log you in.</h3>
           <h3>Please call the Vets.gov Help Desk at 1-855-574-7286. We're open Monday‒Friday, 8:00 a.m.‒8:00 p.m. (ET).</h3>
-          <button onClick={window.close()}>Close</button>
+          <button onClick={window.close}>Close</button>
         </div>
       );
     }


jcmeloni-usds:
@robbiethegeek-usds Who can review this for Bill?

robbiethegeek-usds:
Hey @jbalboni can you review and confirm and LGTM.

src/js/auth/containers/AuthApp.jsx
```diff
@@ -84,7 +84,7 @@ class AuthApp extends React.Component {
         <div>
           <h3>We are sorry that we could not successfully log you in.</h3>
           <h3>Please call the Vets.gov Help Desk at 1-855-574-7286. We're open Monday‒Friday, 8:00 a.m.‒8:00 p.m. (ET).</h3>
-          <button onClick={window.close()}>Close</button>
+          <button onClick={window.close}>Close</button>
```
jbalboni: This is fine, but it makes me a little antsy because you would normally want to bind this function so it has the right `this` value when executed (e.g. `window.close.bind(window)`). It doesn't cause any problems here because `window` is the global context, I think.

jbalboni (approved):

src/js/auth/containers/AuthApp.jsx
omgitsbillryan: Sorry for the delay - I did some quick manual testing to see if I could get `window.close` to call on the wrong `window` and couldn't do it.  I think both `window.close` & `window.close.bind(window)` are equivalent here.  I'll yield to whatever you think is better.

jbalboni: I think we can leave it the way it is.

PR #5024, opened 2017-03-06
Fix address country schema issue

jbalboni:
Related to [1487](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1487)

You can submit an address without a country, which will cause backend validation issues.

This changes it so that there's no blank option in the country dropdown and also cleans up some of the way we handle empty address objects.

screen shot 2017-03-06 at 12 54 32 pm


src/js/common/schemaform/Address.jsx
```diff
@@ -166,7 +176,7 @@ class Address extends React.Component {
             onBlur={onBlur}/>
         <SchemaField
             name="state"
-            required={_.includes(['USA', 'CAN', 'MEX'], formData.country) && schema.required}
+            required={_.includes(['USA', 'CAN', 'MEX'], formData.country) && !!schema.required}
```
jbalboni: This fixes a prop type warning

src/js/common/schemaform/widgets/SelectWidget.jsx
```diff
@@ -49,7 +49,7 @@ function SelectWidget({
           const newValue = getValue(event, multiple);
           onChange(processValue(schema, newValue));
         }}>
-      <option value="">{placeholder}</option>
+      {!schema.default && <option value="">{placeholder}</option>}
```
jbalboni: This is the primary change. This is what the non-customized SelectWidget does in rjsf.

test/common/schemaform/widgets/SelectWidget.unit.spec.jsx
```diff
@@ -122,4 +122,24 @@ describe('Schemaform <SelectWidget>', () => {
     });
     expect(onBlur.calledWith('testing', 'test')).to.be.true;
   });
+  it('should not render blank option when default exists', () => {
```
cvalarida: Should there be the inverse test? Should it render a blank option when no default exists? Is that a bit redundant?

cvalarida (approved):
Looks good with the addition of maybe one more test (though that may be overkill).

test/common/schemaform/widgets/SelectWidget.unit.spec.jsx
jbalboni: I think it's covered ok by the first test, which requires there be two options for an enum with a single value.

src/js/common/schemaform/Address.jsx
```diff
@@ -42,14 +42,24 @@ class Address extends React.Component {
...
+    let newState = set(path, update, formData);
 
     // if country is changing we should clear the state
     if (path === 'country') {
       newState = set('state', undefined, newState);
     }
 
-    this.props.onChange(newState);
+    const fields = Object.keys(newState);
+    if (fields.length === 0 || fields.every(field => field === 'country' || newState[field] === undefined)) {
```
annekainicUSDS: For the second condition, `fields.every(field => field === 'country' || newState[field] === undefined)`, we don't care if `country` has an undefined value because we've prevented the ability to select the blank placeholder option? Shouldn't we just check `newState[field] === undefined` for all fields?

src/js/common/schemaform/Address.jsx
```diff
@@ -42,14 +42,24 @@ class Address extends React.Component {
...
+      // send undefined so that the object is removed
```
annekainicUSDS: Could you explain this piece a little bit more?

annekainicUSDS (commented):
I tested this locally, and when I try to change the country to anything other than "United States" the field doesn't update. Maybe because we're not sending the `newState`?

jbalboni:
@annekainicUSDS That should be fixed now.

src/js/common/schemaform/Address.jsx
jbalboni: This is related to your above comment, too. I'm doing this to make sure we return `undefined` whenever we're in a state where the user has entered no form data. This keeps us from sending empty objects to the backend unnecessarily which can cause validation problems.

This particular code is catching the case where a user makes a change to an address field, then removes it. If the form data has only `undefined` fields, except for a country that matches the default, then return `undefined` instead of the object, because we know there's no address info in it.

annekainicUSDS (approved):
Fixes LGTM!

PR #5026, opened 2017-03-06
Explain defaultDefinitions

cvalarida:
I was unclear what this `defaultDefinitions` was for, so I added an example of why it's needed. It's probably not worded very well as-is, but I at least wanted to note that the `"$ref"`s in the `schema.properties` were referring to what goes into `defaultDefinitions`.

### DIFF
--- a/docs/schemaform/form-config.md
+++ b/docs/schemaform/form-config.md
@@ -25,9 +25,7 @@ Forms are created by creating a page that uses FormApp from the schemaform folde
   // Schema definitions that can be referenced on any page. These are added to each page's schema
   // in the reducer code, so that you don't have to put all of the common fields in the definitions
   // property in each page schema.
-  // In a schema's properties, there are references: "homePhone": { "$ref": "#/definitions/phone" }
-  // In the config file, the definition for phone would then need to be added into
-  // defaultDefinitions for it to be parsed correctly and added to homePhone
+  // For more information on definitions, see schema.definitions below.
   defaultDefinitions: {}, 
 
   // Object containing the configuration for each chapter. Each property is the key for a chapter
@@ -54,6 +52,11 @@ Forms are created by creating a page that uses FormApp from the schemaform folde
       // JSON schema object for the page. Follows the JSON Schema format.
       schema: {
         type: 'object',
+        // In a schema's properties, there are references to definitions. For example:
+        //   "homePhone": { "$ref": "#/definitions/phone" }
+        // In the config file, the definition for phone would then need to be added into definitions
+        // for it to be parsed correctly and added to homePhone.
+        definitions: {},
         properties: {
           field1: {
             type: 'string'
--- a/docs/schemaform/form-config.md
+++ b/docs/schemaform/form-config.md
@@ -25,6 +25,9 @@ Forms are created by creating a page that uses FormApp from the schemaform folde
   // Schema definitions that can be referenced on any page. These are added to each page's schema
   // in the reducer code, so that you don't have to put all of the common fields in the definitions
   // property in each page schema.
+  // In a schema's properties, there are references: "homePhone": { "$ref": "#/definitions/phone" }
+  // In the config file, the definition for phone would then need to be added into
+  // defaultDefinitions for it to be parsed correctly and added to homePhone
   defaultDefinitions: {}, 
 
   // Object containing the configuration for each chapter. Each property is the key for a chapter


docs/schemaform/form-config.md
```diff
@@ -25,6 +25,9 @@ Forms are created by creating a page that uses FormApp from the schemaform folde
   // Schema definitions that can be referenced on any page. These are added to each page's schema
   // in the reducer code, so that you don't have to put all of the common fields in the definitions
   // property in each page schema.
+  // In a schema's properties, there are references: "homePhone": { "$ref": "#/definitions/phone" }
+  // In the config file, the definition for phone would then need to be added into
+  // defaultDefinitions for it to be parsed correctly and added to homePhone
```
jbalboni: It might make sense to add a `definitions` property in the `schema` object in here and put this description there. In JSON Schema, the usual way of specifying definitions for fields using `$ref` is in a `definitions` property at the root of the schema. The point of `defaultDefinitions` is that we will add the definition in it to each page's schema automatically, instead of forcing a developer to add them to each page's schema individually.

docs/schemaform/form-config.md
```diff
@@ -51,6 +52,11 @@ Forms are created by creating a page that uses FormApp from the schemaform folde
       // JSON schema object for the page. Follows the JSON Schema format.
       schema: {
         type: 'object',
+        // In a schema's properties, there are references to definitions. For example:
+        //   "homePhone": { "$ref": "#/definitions/phone" }
+        // In the config file, the definition for phone would then need to be added into definitions
+        // for it to be parsed correctly and added to homePhone.
+        definitions: {},
```
raquelromano-gov: Thanks for adding to this document, though actually this still doesn't quite make sense to me. It looks like our vets-json-schema schema files typically have a definitions field (some subset of common definitions), and the corresponding form config file will have a defaultDefinitions field, which is yet another subset.

Guess I'm saying maybe we should make this example correspond more accurately to the way we're organizing our code.

cvalarida: @raquelromano-gov How would you suggest modifying it?

From what I understand, the schema we're importing from `vets-json-schema` is what's used to share schema between the front and back ends, but unless we put it in the config in `config/form.js`, it's not used. The trouble with piping some of the `properties` in the schema from `vets-json-schema` directly into the config is they have `$ref`s to `"#/definitions/something"`, so when `react-json-schema` (I think?) finds that ref, it looks in this `definitions` object (, not in the schema from `vets-json-schema`, and if we don't have the definition in _here_ (a page's individual schema in `config/form.js`), it'll error out.

The way @jbalboni explained it to me is every page is a separate schema as far as `react-json-schema` is concerned. What that means is we need to fill in everything that it needs to make it a complete json schema per the standard.

I really don't know how to explain it very well, and I feel like I may be missing your point entirely! I'm open to suggestions.

jbalboni: I think we can talk about this at the meeting today.

raquelromano-gov:
Actually, your explanation above makes perfect sense! My issue is that this page (and the whole code block) demonstrates how we write form configs, but the `definitions` field isn't used in form configs, but rather in the shared schemas in `vets-json-schema`:
- Shared schema has a `definitions` for those properties that use `$ref`: https://github.com/department-of-veterans-affairs/vets-json-schema/blob/master/src/schemas/change-of-program/schema.js 
- Form config doesn't have a `definitions` field; when there is no `$ref`, it pulls in a schema directly from the shared schema's `properties`; otherwise it pulls it in from `definitions`: https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/js/edu-benefits/1995/config/form.js 

So, my suggestion is we actually use this doc to show a snippet of both the shared schema and the form config, and include both a property that is explicitly defined and one that uses `$ref`. Then we can show how the form config uses those. I'm glad to do that if you want--today's sync with Jeff will be an opportunity to ask a couple of other questions I have, so I can also beef up this doc with what we learn then.

docs/schemaform/form-config.md
raquelromano-gov: We forgot to discuss this yesterday, but one suggestion is to go ahead and merge, and then later I can take a pass at revising this page to separate out the config snippet and a backend schema snippet to show how we're sharing definitions across the FE and BE.

cvalarida:
What was the conclusion we came to on this after the meeting yesterday, @raquelromano-gov @jbalboni?

PR #5027, opened 2017-03-06
Add employment history for 5490

jbalboni:
Related to [721](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/721)

screen shot 2017-03-06 at 4 04 04 pm
screen shot 2017-03-06 at 4 04 18 pm

### DIFF
--- a/src/js/edu-benefits/5490/config/form.js
+++ b/src/js/edu-benefits/5490/config/form.js
@@ -11,6 +11,7 @@ import * as phone from '../../../common/schemaform/definitions/phone';
 import contactInformation from '../../definitions/contactInformation';
 
 import IntroductionPage from '../components/IntroductionPage';
+import EmploymentPeriodView from '../components/EmploymentPeriodView';
 import ConfirmationPage from '../containers/ConfirmationPage';
 
 const {
@@ -18,7 +19,8 @@ const {
 } = fullSchema5490.properties;
 
 const {
-  secondaryContact
+  secondaryContact,
+  nonMilitaryJobs
 } = fullSchema5490.definitions;
 
 const formConfig = {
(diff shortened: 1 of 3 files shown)


src/js/edu-benefits/5490/config/form.js
```diff
@@ -69,9 +71,40 @@ const formConfig = {
...
+                itemName: 'Employment Period',
+                viewField: EmploymentPeriodView,
+                hideTitle: true
+              }
+            }
+          },
+          schema: {
+            type: 'object',
+            properties: {
+              nonMilitaryJobs: _.unset('items.properties.postMilitaryJob', nonMilitaryJobs)
```
jbalboni: Going to remove this from the schema, I think.

cvalarida: Makes sense if you can't add `postMilitaryJob` anywhere. I take it it's not being used elsewhere? Where does it come from?

test/edu-benefits/5490/config/employmentHistory.unit.spec.jsx
```diff
@@ -0,0 +1,65 @@
...
+          onSubmit={onSubmit}
+          data={{}}
+          uiSchema={uiSchema}/>
+    );
+    const formDOM = findDOMNode(form);
+    submitForm(form);
+
+    expect(Array.from(formDOM.querySelectorAll('.usa-input-error'))).to.be.empty;
+
+    submitForm(form);
```
cvalarida: This second call probably isn't necessary if there are no errors.

cvalarida (approved):
I had a question and comment, but it looks good.

src/js/edu-benefits/5490/config/form.js
jbalboni: After talking with Alex, we realized it's from the 1990 form. This is using the same definition. It looks awkward to remove it on the backend, so I may leave it for now.

PR #5028, opened 2017-03-06
GIBCT autosuggest styling.

U-DON:
Resolves department-of-veterans-affairs/vets.gov-team#1239.


ben-USDS-damman (approved):
LGTM

U-DON:
Jenkins build isn't happening here for some reason. Closing in order to get the build running on a new PR.

saintsoup52:
What happens after the school is selected? 

In the old version the user was redirected to the next page... here what they select will have to stay on the page... will it just be in the box?

U-DON:
@saintsoup52 It should update the search results like a search term. Since the selection is a specific name, it would list just the one result that matches it. This still needs to be implemented, however.

PR #5029, opened 2017-03-07
Fixed GIBCT filtering on values that should not be converted to camelCase.

U-DON:
So found an edge case here… With the `/v0/gi/institutions/search` resource, we get a list of countries in the `meta.facets.country` object, and `bosnia_herzegovina` is one of them. However, trying to filter with it doesn’t return any results: `/v0/institutions?country=BOSNIA_HERZEGOVINA&version=1`.

Furthermore, we convert to camelCase on the front end, so it actually removes the underscore and would get us `bosniaHerzegovina`. It seems that the correct filter would be `country=bosnia-herzegovina` with the hyphen.

Trying to standardize on a case messes up searching on the other countries that have spaces (and parentheses) instead of hyphens, so the solution here seems to be to keep the values from the original payload's facets (without case conversion). For this specific case, we also need to replace the underscore with a hyphen, but assuming here that the normalized facets are generally expected to have hyphens in place of underscores.

### DIFF
--- a/src/js/gi/reducers/search.js
+++ b/src/js/gi/reducers/search.js
@@ -1,6 +1,7 @@
 /* eslint-disable no-case-declarations */
-import { SEARCH_STARTED, SEARCH_FAILED, SEARCH_SUCCEEDED } from '../actions';
+import { camelCase } from 'lodash';
 import camelCaseKeysRecursive from 'camelcase-keys-recursive';
+import { SEARCH_STARTED, SEARCH_FAILED, SEARCH_SUCCEEDED } from '../actions';
 
 const INITIAL_STATE = {
   facets: {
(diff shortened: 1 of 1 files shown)


ben-USDS-damman (approved):
This solution is fine. Ideally this would be fixed in GIDS.

U-DON:
@ben-USDS-damman Who currently owns GIDS? Is it likely to get fixed there, or would it be better to merge this for now?

ayaleloehr:
@mphprogrammer and @saneshark know the GIDS API well so hopefully can answer @U-DON 's question?

saneshark:
Agree, this should be fixed in GIDS. Can we open ticket in #gibct-data-service and assign ticket to @mphprogrammer

U-DON:
Created an issue at department-of-veterans-affairs/gibct-data-service#224. Closing this PR.

PR #5034, opened 2017-03-07
Fix uncontrolled input warning

jbalboni:
Related to [1385](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1385)

TextWidget should be setting value to a string always, so React doesn't switch between modes.

### DIFF
--- a/src/js/common/schemaform/widgets/TextWidget.jsx
+++ b/src/js/common/schemaform/widgets/TextWidget.jsx
@@ -13,7 +13,7 @@ export default function TextWidget(props) {
         maxLength={props.schema.maxLength}
         autoComplete={props.options.autocomplete || false}
         className={props.options.widgetClassNames}
-        value={props.value}
+        value={props.value || ''}
         onBlur={() => props.onBlur(props.id)}
         onChange={(event) => props.onChange(event.target.value ? event.target.value : undefined)}/>
   );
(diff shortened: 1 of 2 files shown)


src/js/common/schemaform/widgets/TextWidget.jsx
```diff
@@ -13,7 +13,7 @@ export default function TextWidget(props) {
         maxLength={props.schema.maxLength}
         autoComplete={props.options.autocomplete || false}
         className={props.options.widgetClassNames}
-        value={props.value}
+        value={props.value || ''}
```
cvalarida: The original change from `value={props.value || ''}` to `value={props.value}` was for testing, correct? Was there any other reason to allow `undefined` instead of a blank string? Will it cause blank strings to be sent back to the back end when `undefined` (or nothing at all) should be?

cvalarida (approved):
Barring anything too strange with the blank string vs. undefined debate, I'd say it looks good. It was working like this before, yes?

src/js/common/schemaform/widgets/TextWidget.jsx
jbalboni: Blank strings are converted to `undefined` in the onChange function, so this shouldn't cause any problems (and was how it worked before the recent change). Otherwise, it should have no effect.

PR #5036, opened 2017-03-07
1990e applicant info form config and unit test

raquelromano-gov:
Closes https://github.com/department-of-veterans-affairs/vets.gov-team/issues/625

Not ready for merging, but I have a testing question: the test harness doesn't seem to be picking up the 'relationship' definition, even though I'm exporting it through defaultDefinitions...wondering if it has something to do with the fact that the 1990e backend schema pulls this in with $ref rather than defining it explicitly (only difference I can see between this and the similar issue I had with 1995 tests).

Also have a general questions about rules of thumb for adding common widgets and definitions to src/js/common/schemaform/


raquelromano-gov:
Unit test failure error:

```
/Users/romano/dsva/vets-website/src/js/edu-benefits/utils/helpers.jsx:236
  return enumValues.map(function (item) {
                   ^

TypeError: Cannot read property 'map' of undefined
    at enumToNames (/Users/romano/dsva/vets-website/src/js/edu-benefits/utils/helpers.jsx:147:10)
    at Object.<anonymous> (/Users/romano/dsva/vets-website/src/js/edu-benefits/1990e/config/form.js:64:28)

```

so looks like relationship isn't defined

src/js/common/schemaform/widgets/GenderWidget.jsx
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
```
jbalboni: I don't think we need a widget for gender. Should be able to work the same as the relationship field.

raquelromano-gov:
This is ready for review.

Screenshots:
image
image

src/js/edu-benefits/1990e/config/form.js
```diff
@@ -31,7 +48,36 @@ const formConfig = {
...
-          title: 'Applicant information'
+          title: 'Applicant information',
+          initialData: {},
+          uiSchema: {
+            applicantFullName: fullName.uiSchema,
+            applicantSocialSecurityNumber: ssn.uiSchema,
+            applicantDateOfBirth: currentOrPastDate.uiSchema('Applicant date of birth'),
+            applicantGender: {
+              'ui:widget': 'radio',
+              'ui:title': 'Applicant gender'
```
jbalboni: Same as above. I think this can just be Gender.

src/js/edu-benefits/1990e/config/form.js
```diff
@@ -31,7 +48,36 @@ const formConfig = {
...
+            relationship: {
+              'ui:widget': 'radio',
+              'ui:title': 'What is your relationship to the service member whose benefit is being transferred to you?'
+            }
+          },
+          schema: {
+            type: 'object',
+            required: ['applicantFullName'],
+            properties: {
+              applicantFullName: fullName.schema,
```
jbalboni: These property names should match the names in the backend schema. So, for example, `applicantFullName` should be `veteranFullName` ([here](https://github.com/department-of-veterans-affairs/vets-json-schema/blob/master/dist/transfer-benefits-schema.json#L524) in the schema).

src/js/edu-benefits/1990e/config/form.js
```diff
@@ -31,7 +48,36 @@ const formConfig = {
       pages: {
         applicantInformation: {
           path: 'applicant-information',
-          title: 'Applicant information'
+          title: 'Applicant information',
+          initialData: {},
+          uiSchema: {
+            applicantFullName: fullName.uiSchema,
+            applicantSocialSecurityNumber: ssn.uiSchema,
+            applicantDateOfBirth: currentOrPastDate.uiSchema('Applicant date of birth'),
```
jbalboni: This label can be 'Date of birth' (unless Alex asked for it to be this way).

test/edu-benefits/1990e/config/benefitSelection.unit.spec.jsx
```diff
@@ -16,9 +16,8 @@ describe('Edu 1990e benefitEligibility', () => {
           data={{}}
           uiSchema={uiSchema}/>
     );
-    const fields = ReactTestUtils.scryRenderedDOMComponentsWithTag(form, 'input');
-
-    expect(fields.length).to.equal(2);
+    expect(ReactTestUtils.scryRenderedDOMComponentsWithTag(form, 'input'))
```
jbalboni: Why did you make this change? If it was causing a test failure, I think you may need to run `npm install`.

src/js/edu-benefits/1990e/config/form.js
jbalboni: We've also in the past validated that applicants are at least 17. Should we do that here @goldenmeanie?

jbalboni (commented):
I made a few comments, but it looks good.

src/js/edu-benefits/1990e/config/form.js
raquelromano-gov: The outline I'm using is pretty rough and looks like the language will need to be refined: https://github.com/department-of-veterans-affairs/vets.gov-team/blob/master/Products/Education%20Benefits/1990E/1990E-outline-draft.md
E.g., I'm guessing DOB is too informal, but I'm just guessing on that and other fields.
Let me know if there's a better source for exact language.

src/js/edu-benefits/1990e/config/form.js
raquelromano-gov: Actually, it's not the veteran in this case, but I'll change it to `relativeFullName` to match the backend schema

test/edu-benefits/1990e/config/benefitSelection.unit.spec.jsx
raquelromano-gov: Originally yes, I changed it because of a failure, but after running npm install and getting updated, I noticed that in some tests we just check that inputs are present and in others we test for exact counts. Glad to go with exact counts if you don't think these will be changing often.

src/js/edu-benefits/1990e/config/form.js
goldenmeanie: I think that is true here as well. It won't be on the 5490.

raquelromano-gov: @goldenmeanie Can you check over https://github.com/department-of-veterans-affairs/vets.gov-team/blob/master/Products/Education%20Benefits/1990E/1990E-outline-draft.md and change anything that isn't exactly as it should appear on the form?

raquelromano-gov: Just revisited https://github.com/department-of-veterans-affairs/vets.gov-team/issues/625 and realized I also need to add this line of text: "You aren’t required to fill in all fields, but VA can evaluate your claim faster if you provide more information."

jbalboni: I've been using the language in the GH issues for each page, or what we've used for these fields in other forms (mainly the 1990). 

It's not a huge deal either way, since Molly and Alex will make a pass through the form and make copy changes (or tell us what changes to make).

src/js/edu-benefits/1990e/config/form.js
jbalboni: Oh, gotcha! Yeah, the `relative*` fields look like the right ones to match.

test/edu-benefits/1990e/config/benefitSelection.unit.spec.jsx
jbalboni: I'm not super happy with testing counts, but I think that's better than testing just the presence of inputs. We don't change input types that often and it's more likely to catch schema issues (which sometimes show up as an error message on the form, in lieu of the right input fields).

src/js/edu-benefits/1990e/config/form.js
jbalboni: @raquelromano-gov There should be a validation function used by the 1990 for this. I think we could make a dateOfBirth definition in `edu-benefits/definitions` that includes that validation and the current or past date one. We'll use it on a couple forms, potentially.

goldenmeanie: Latest copy is in the `ed-22-1990E` tagged issues.

raquelromano-gov: @goldenmeanie Another question: is it true that relatives of veterans applying for veterans must be 17 years of age? I poked around in our docs and (many) online resources, including vonapp and paper form and can't see that restriction. However, I did see somewhere that once the relative reaches 26 years of age, s/he is no longer eligible. Need to know so we can set up the validations properly.

raquelromano-gov: @goldenmeanie Just bumping this question...the PR is blocked on this. Thanks!

goldenmeanie: 1990E shouldn't have an upper limit, but I believe the 17 floor is true.

goldenmeanie: Ok, checked with Shay. No date validation at all!

src/js/edu-benefits/1990e/config/form.js
```diff
@@ -23,6 +37,9 @@ const formConfig = {
   transformForSubmit: transform,
   introduction: IntroductionPage,
   confirmation: ConfirmationPage,
+  defaultDefinitions: {
```
annekainicUSDS: What is this for?

annekainicUSDS (approved):
This looks great!

src/js/edu-benefits/1990e/config/form.js
raquelromano-gov: This is actually what prompted our chat yesterday (see https://github.com/department-of-veterans-affairs/vets-website/pull/5026/files?short_path=c48d4c5#diff-c48d4c5da3172b6c94db9bc6f8c70ad0), though I'm still lacking complete clarity.   We are going to make a 2nd pass at that doc to make it clear why/when/which defaultDefinitions are needed

PR #5038, opened 2017-03-07
Rewrote GIBCT max-width queries to use media-maxwidth mixin.

ghost:
Related to department-of-veterans-affairs/vets.gov-team#818


U-DON (approved):
LGTM.

PR #5040, opened 2017-03-07
Create school selection page for 5490

jbalboni:
Related to [722](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/722)

This adds the school selection page. Still getting info on what fields are required.

screen shot 2017-03-07 at 5 15 03 pm
screen shot 2017-03-07 at 5 15 09 pm
screen shot 2017-03-07 at 5 15 17 pm


src/js/edu-benefits/5490/config/form.js
```diff
@@ -34,6 +47,10 @@ const formConfig = {
   confirmation: ConfirmationPage,
   title: 'Update your Education Benefits',
   subTitle: 'Form 22-5490',
+  defaultDefinitions: {
+    educationType: educationType.schema,
```
cvalarida: Would this make more sense in the individual page schema, or will it show up in another page as well?

jbalboni: I'm tempted to suggest that we always use `defaultDefinitions`, actually. Might be simpler. We can talk about this in the meeting today.

jbalboni: @cvalarida I'm going to leave this, pending how the change to set all the $ref fields goes (https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1578)

PR #5045, opened 2017-03-08
Use better key in schemaform ObjectField

jbalboni:
Related to [1530](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1530)

This is good cautionary tale about using index keys. Since each property's SchemaField component is inside a div with an index key, those divs are not changed when the set of properties being rendered changes. The SchemaField component DOM just moves up or down to the next div. On the page mentioned in the issue, this meant that some new fields were being added inside a div that had the first field class (which wasn't removed because it's not set with React and is set on the containing div).

The solution to this is to use `propName` as a key, which should re-use divs properly.

screen shot 2017-03-08 at 2 11 48 pm

### DIFF
--- a/src/js/common/schemaform/ObjectField.jsx
+++ b/src/js/common/schemaform/ObjectField.jsx
@@ -165,9 +165,9 @@ class ObjectField extends React.Component {
       'schemaform-block': title && !isRoot
     });
 
-    const renderProp = (propName, index) => {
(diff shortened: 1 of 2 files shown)


cvalarida (approved):
Huh. I can't say as I quite understand the mechanics behind it yet, but very cleverly done! It works like a charm (if charms worked well, that is).

PR #5047, opened 2017-03-08
Add ability to specify enum labels in uiSchema

jbalboni:
This came out of conversation I had with Raquel yesterday. I think we can pull enum labels from `ui:options` in uiSchema and makes things a little easier.

I haven't removed the ability to do it with `enumNames`, but I've converted all the current uses of that.


src/js/edu-benefits/utils/helpers.jsx
```diff
@@ -142,3 +142,7 @@ export const preferredContactMethodLabels = {
   phone: 'Phone'
 };
 
+export const relationshipLabels = {
+  spouse: 'Spouse',
+  child: 'Child'
```
cvalarida: If these are the ones in the 5490, there was a copy change; they should be:

```js
export const relationshipLabels = {
  child: 'Child/Stepchild/Adopted Child',
  spouse: 'Spouse/Surviving Spouse',
};
```

I don't know if it would be better to do that here or to do that when resolving the merge conflict, but it might be less likely to be lost if done here.

src/js/common/schemaform/widgets/RadioWidget.jsx
```diff
@@ -23,7 +23,7 @@ export default function RadioWidget({
                 disabled={disabled}
                 onChange={_ => onChange(option.value)}/>
             <label htmlFor={`${id}_${i}`}>
-              {option.label}
+              {labels[option.value] || option.label}
```
jbalboni: The built in StringField component creates an `enumOptions` array that zips up `enum` and `enumNames` as an array of objects. That's what we've been using and what this PR should override, so I'm having the labels object passed in through `uiSchema` take precedent.

src/js/edu-benefits/utils/helpers.jsx
jbalboni: Good to know. I'll make sure it gets merged correctly.

src/js/common/schemaform/widgets/RadioWidget.jsx
cvalarida: Okay, I had a suspicion that was the case. Makes sense. Thanks!

cvalarida:
Looks good!

src/js/edu-benefits/utils/helpers.jsx
raquelromano-gov: Oh, the 1990e just has 'Child' and 'Spouse' :(
I'll check with Alex to see if 'Child' should be more expansive, but 1990e definitely won't be using 'Surviving Spouse' so maybe this set of labels shouldn't be shared after all?

test/common/schemaform/widgets/SelectWidget.unit.spec.jsx
```diff
@@ -24,6 +24,27 @@ describe('Schemaform <SelectWidget>', () => {
...
+    const onChange = sinon.spy();
+    const tree = SkinDeep.shallowRender(
+      <SelectWidget
+          schema={{}}
+          id="testing"
+          onChange={onChange}
+          options={{ enumOptions, labels }}/>
+    );
+
+    expect(tree.everySubTree('option')[1].text()).to.equal(labels.test);
```
raquelromano-gov: Test for 'Other' explicitly instead of `labels.test` as above?

raquelromano-gov (commented):
Nice! This makes the configs much cleaner.

src/js/edu-benefits/utils/helpers.jsx
cvalarida: Got another copy edit on this if we want to track it in here:

```js
export const relationshipLabels = {
  child: 'Child, stepchild, adopted child',
  spouse: 'Spouse or Surviving Spouse',
};
```

PR #5052, opened 2017-03-09
Change PHR polling frequency to once every 30 seconds

bshyong:
resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1587


U-DON (approved):
LGTM.

PR #5056, opened 2017-03-09
1990e Chapter 3

annekainicUSDS:
Closes https://github.com/department-of-veterans-affairs/vets.gov-team/issues/635.

After talking to @goldenmeanie we're not including the `relationship` question here (that is being included in Chapter 1 instead).

A few questions for @goldenmeanie:
1. Even though veteran first and last name aren't indicated as being required in the issue, do you want to make them required? Seems like that is useful and obvious information that should be in there?
2. Is it clear that these fields are in reference to the veteran/sponsor and not the applicant? Wonder if it would help to change the labels to be more specific (e.g., "Veteran first name" instead of "First name").

And thanks @raquelromano-gov for the unit tests! Copied and amended from your work on Chapter 1 :)

Screenshot:
screencapture-localhost-3001-education-apply-for-education-benefits-application-1990e-sponsor-veteran-1489087333150

### DIFF
--- a/src/js/edu-benefits/1990e/config/form.js
+++ b/src/js/edu-benefits/1990e/config/form.js
@@ -168,6 +168,7 @@ const formConfig = {
             properties: {
               sponsorVeteran: {
                 type: 'object',
+                required: ['veteranFullName', 'veteranSocialSecurityNumber'],
                 properties: {
                   veteranFullName: fullName,
                   veteranSocialSecurityNumber: ssn,
--- a/src/js/edu-benefits/1990e/config/form.js
+++ b/src/js/edu-benefits/1990e/config/form.js
@@ -10,8 +10,9 @@ import directDeposit from '../../pages/directDeposit';
 import createSchoolSelectionPage from '../../pages/schoolSelection';
 
 import * as currentOrPastDate from '../../../common/schemaform/definitions/currentOrPastDate';
-import * as fullName from '../../../common/schemaform/definitions/fullName';
-import * as ssn from '../../../common/schemaform/definitions/ssn';
+import * as fullNameCommon from '../../../common/schemaform/definitions/fullName';
+import * as ssnCommon from '../../../common/schemaform/definitions/ssn';
+import * as address from '../../../common/schemaform/definitions/address';
 
 import IntroductionPage from '../components/IntroductionPage';
 import ConfirmationPage from '../containers/ConfirmationPage';
@@ -24,11 +25,15 @@ import {
 
 const {
   gender,
-  relationship
+  relationship,
+  fullName,
+  ssn
 } = fullSchema1990e.definitions;
 
 const {
   benefit,
+  serviceBranch,
+  civilianBenefitsAssistance
 } = fullSchema1990e.properties;
 
 const {
@@ -58,8 +63,8 @@ const formConfig = {
           title: 'Applicant information',
           initialData: {},
           uiSchema: {
-            relativeFullName: fullName.uiSchema,
-            relativeSocialSecurityNumber: ssn.uiSchema,
+            relativeFullName: fullNameCommon.uiSchema,
+            relativeSocialSecurityNumber: ssnCommon.uiSchema,
             relativeDateOfBirth: currentOrPastDate.uiSchema('Date of birth'),
             gender: {
               'ui:widget': 'radio',
@@ -80,8 +85,8 @@ const formConfig = {
             type: 'object',
             required: ['relativeFullName'],
             properties: {
-              relativeFullName: fullName.schema,
-              relativeSocialSecurityNumber: ssn.schema,
+              relativeFullName: fullNameCommon.schema,
+              relativeSocialSecurityNumber: ssnCommon.schema,
               relativeDateOfBirth: currentOrPastDate.schema,
               gender,
               relationship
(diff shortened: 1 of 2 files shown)


src/js/edu-benefits/1990e/config/form.js
```diff
@@ -6,8 +6,9 @@ import {
 import fullSchema1990e from 'vets-json-schema/dist/transfer-benefits-schema.json';
 
 import * as currentOrPastDate from '../../../common/schemaform/definitions/currentOrPastDate';
-import * as fullName from '../../../common/schemaform/definitions/fullName';
-import * as ssn from '../../../common/schemaform/definitions/ssn';
+import * as fullNameCommon from '../../../common/schemaform/definitions/fullName';
```
raquelromano-gov: Hm...I know we talked about this in the schemaform sync yesterday (pulling the schema definition directly from the backend, like you do below for fullName, and here separately pulling out the front-end shared uiSchema definition). If the definitions in schemaform/ are only going to export uiSchema, then we could do something like
```
import uiSchema as fullNameUiSchema from '../../../common/schemaform/definitions/fullName';
```
and then below refer to `fullNameUiSchema` instead of `fullNameCommon.uiSchema`

It's a little cumbersome, but maybe more clear. @jbalboni, wondering what you think

src/js/edu-benefits/1990e/config/form.js
```diff
@@ -111,6 +116,38 @@ const formConfig = {
     sponsorVeteran: {
       title: 'Sponsor Veteran',
       pages: {
+        sponsorVeteran: {
+          title: 'Sponsor Veteran',
+          path: 'sponsor-veteran',
+          uiSchema: {
+            sponsorVeteran: {
+              veteranFullName: fullNameCommon.uiSchema,
+              veteranSocialSecurityNumber: ssnCommon.uiSchema,
```
raquelromano-gov: Same thing here as the suggestion for fullName's uiSchema

src/js/edu-benefits/1990e/config/form.js
```diff
@@ -111,6 +116,38 @@ const formConfig = {
...
+          },
+          schema: {
+            type: 'object',
+            properties: {
+              sponsorVeteran: {
+                type: 'object',
+                properties: {
+                  veteranFullName: fullName,
+                  veteranSocialSecurityNumber: ssn,
+                  veteranAddress: address.schema(false),
```
raquelromano-gov: This defaults to false, so address.schema() would be enough

test/edu-benefits/1990e/config/sponsorVeteran.unit.spec.jsx
```diff
@@ -0,0 +1,66 @@
...
+  it('should render', () => {
+    const form = ReactTestUtils.renderIntoDocument(
+      <DefinitionTester
+          schema={schema}
+          data={{}}
+          uiSchema={uiSchema}
+          definitions={definitions}/>
+    );
+
+    expect(ReactTestUtils.scryRenderedDOMComponentsWithTag(form, 'input'))
```
raquelromano-gov: I need to update the applicant-info test to actually test for the right number of input fields. Was discusisng this with @jbalboni in PR #5036 and we agreed to be more specific. Take a look at Chris's test in benefitSelection.unit.spec.jsx

src/js/edu-benefits/1990e/config/form.js
jbalboni: Yeah, I think that makes sense for cases where the front end does not need to modify the schema. But I don't think we should modify all the common components this PR uses now, though. I think that's better as a separate issue.

The fullName schema is a little tricky because we might need to modify it depending on if the name is required or not.

For now I think we could import it like this to make it easy to change to the above style in the future:

`import { uiSchema as fullNameUiSchema } from '../../../common/schemaform/definitions/fullName';`

jbalboni (approved):
I agree with the comments Raquel made and don't see any other issues, so LGTM, pending those fixes and whatever changes Alex wants.

goldenmeanie:
1. I went through both forms with Shay this afternoon and am updating issues with REQUIRED labels
2. I think that's consistent with feedback we've gotten and if it doesn't complicate our component situation I'm down.

src/js/edu-benefits/1990e/config/form.js
annekainicUSDS: I actually thought about that, but I wasn't sure how to handle it because I'm actually using the back-end schema for the regular schema (https://github.com/department-of-veterans-affairs/vets-website/blob/0056e270a153e62ade920213441b4eea4164b346/src/js/edu-benefits/1990e/config/form.js#L141) and the common schema for the uiSchema (https://github.com/department-of-veterans-affairs/vets-website/blob/0056e270a153e62ade920213441b4eea4164b346/src/js/edu-benefits/1990e/config/form.js#L124). I thought it would be important to use the same schema that the back-end uses to actually validate the fields. But I wasn't sure where we came out on that during the call yesterday. If you think I should use the common schema for the regular schema I can make that change!

annekainicUSDS:
@raquelromano-gov @goldenmeanie @jbalboni Made the changes from the review (minus the common schema thing, which I had a question about above). Also @goldenmeanie made the label changes and required field changes you requested, see new screenshot below.

screencapture-localhost-3001-education-apply-for-education-benefits-application-1990e-sponsor-veteran-1489158321111

src/js/edu-benefits/1990e/config/form.js
```diff
@@ -111,6 +118,55 @@ const formConfig = {
...
+                  'ui:title': 'Veteran middle name'
+                },
+                suffix: {
+                  'ui:title': 'Veteran suffix',
+                  'ui:options': {
+                    widgetClassNames: 'form-select-medium'
+                  }
+                }
+              },
+              veteranSocialSecurityNumber: _.set(ssnCommon.uiSchema, ['ui:title'], 'Veteran Social Security number'),
```
annekainicUSDS: @jbalboni @raquelromano-gov Realized after I pushed this change that this is going to change the label everywhere... going to have to come up with a different solution for changing the label for this field specifically!

src/js/edu-benefits/1990e/config/form.js
jbalboni: @annekainicUSDS I think we're on the same page as far as what schema you should be using (what you're doing now is what I think we agreed on), Raquel's comment was more just about how to import the uiSchema for fullName. You can use the import statement in my comment to replace `fullNameCommon.uiSchema` with `fullNameUiSchema`, which I think is clearer and it makes it easier to remove the common schema from that file later.

src/js/edu-benefits/1990e/config/form.js
jbalboni: Yeah, we can probably make the common definition take a parameter to prefix the labels. But this is good for now.

jbalboni: @annekainicUSDS I think you should switch this to import `lodash/fp` at the top of the file. That's what's causing that problem, I think. The regular lodash set modifies the common ssn uiSchema.

You'll have to move the `ssnCommon.uiSchema` argument to the end of the function when you do that.

src/js/edu-benefits/1990e/config/form.js
annekainicUSDS: Perfect, thanks, I'll make that change!

src/js/edu-benefits/1990e/config/form.js
annekainicUSDS: Great, thanks!

annekainicUSDS:
@raquelromano-gov @jbalboni This should be updated with all of the requested changes now

jbalboni (approved):
LGTM!

PR #5059, opened 2017-03-09
Add school selection page to 1990e

jbalboni:
Related to [642](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/642)

This makes the educationProgram definition shared between the two forms, but they differ in the other fields, so I'm not sharing the page.

screen shot 2017-03-09 at 4 18 23 pm
screen shot 2017-03-09 at 4 18 30 pm


src/js/edu-benefits/1990e/config/form.js
```diff
@@ -126,6 +139,29 @@ const formConfig = {
...
+              'ui:title': 'Are you getting, or do you expect to get any money (including, but not limited to, federal tuition assistance) from the Armed Forces or public health services for any part of your coursework or training?',
+              'ui:widget': 'yesNo'
+            }
+          },
+          schema: {
+            type: 'object',
+            properties: {
+              educationProgram: educationProgram.schema(),
+              educationObjective,
+              nonVaAssistance
```
cvalarida: `educationObjective` and `nonVaAssitance` are also used in 1995; Could we make the `educationProgram.schema()` take a list of fields to include like `toursOfDuty.schema()`? The second parameter could be the required fields (which I almost did for the `toursOfDuty`). We could then put the `uiSchema` for these in that definitions file and not repeat it between forms.

cvalarida (approved):
Looks good. I had one thought on adding some additional versatility, but it's not imperative.

src/js/edu-benefits/1990e/config/form.js
jbalboni: Yeah, I was on the fence about that. But we are going to have at least three forms with this page, so maybe it's worth it.

src/js/edu-benefits/pages/schoolSelection.js
```diff
@@ -0,0 +1,72 @@
...
+import { uiSchema as uiSchemaDate } from '../../common/schemaform/definitions/date';
+import { states } from '../../common/utils/options-for-select';
+
+const stateLabels = states.USA.reduce((current, { label, value }) => {
+  return _.merge(current, { [value]: label });
+}, {});
+
+export default function createSchoolSelectionPage(
+  schema,
+  fields = ['educationProgram', 'educationObjective', 'nonVaAssistance'],
```
raquelromano-gov: Getting this error on master, which is probably related to this change:
```
Uncaught Error: uiSchema order list contains extraneous property 'educationProgram'
```

jbalboni: @raquelromano-gov you probably need to run npm install and restart your watch task.

cvalarida: @raquelromano-gov Whenever the `vets-json-schema` changes, I've found this happens to me. Running `npm install` and restarting the watch task like @jbalboni says usually does the trick.

raquelromano-gov: ah, yes--thanks!

PR #5065, opened 2017-03-10
Edu applicant information page

cvalarida:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1603

Since the relationship question is in the 5490 in a different place, I made the page a function which takes the field names as parameters and returns the `schema` and `uiSchema` with the desired fields.

## 1990e
Same as above plus relationship question:
> image


src/js/edu-benefits/5490/config/form.js
```diff
@@ -381,7 +390,21 @@ const formConfig = {
             }
           }
         },
-        directDeposit
+        directDeposit: {
```
jbalboni: I think this is an incorrect merge change

src/js/edu-benefits/5490/config/form.js
```diff
@@ -22,7 +23,8 @@ import * as toursOfDuty from '../../definitions/toursOfDuty';
 import * as educationType from '../../definitions/educationType';
 
 import contactInformation from '../../pages/contactInformation';
-import directDeposit from '../../pages/directDeposit';
+
+import applicantInformation from '../../pages/applicantInformation';
```
jbalboni: I think this is an incorrect merge change

src/js/edu-benefits/5490/config/form.js
```diff
@@ -73,7 +75,14 @@ const formConfig = {
   chapters: {
     applicantInformation: {
       title: 'Applicant Information',
-      pages: {}
+      pages: {
+        applicantInformation: applicantInformation([
+          'relativeFullName',
```
jbalboni: Are these the right fields to use? There are also veteranFullName, veteranSocialSecurityNumber, and veteranDateOfBirth.

src/js/edu-benefits/pages/applicantInformation.js
```diff
@@ -0,0 +1,57 @@
...
+    relativeDateOfBirth: currentOrPastDate.schema,
+    gender: definitions.gender,
+    relationship: definitions.relationship
+  };
+
+  return {
+    path: 'applicant-information',
+    title: 'Applicant information',
+    initialData: {},
+    uiSchema: {
```
jbalboni: I think it'd be good to put `'ui:order': propNames` in here to be safe. I'm not positive the order is based on the list of fields in `_.pick`, and if not it could change depending on the schema

src/js/edu-benefits/pages/applicantInformation.js
```diff
@@ -0,0 +1,57 @@
...
+const defaultProps = [
+  'relativeFullName',
+  'relativeSocialSecurityNumber',
+  'relativeDateOfBirth',
+  'gender',
+  'relationship'
+];
+
+export default function applicantInformation(propNames = defaultProps) {
+  const possibleProperties = {
```
jbalboni: The way I did this for the school selection page was to pass in a schema and pick the properties from that. I think that avoids potentially getting out of sync with one of the form schemas, because if you include a field that isn't there, it won't show up, instead of getting included and causing a backend validation errors. And if you include `ui:order`, it'll throw a helpful error in the console, even.

jbalboni (commented):
A couple comments

src/js/edu-benefits/5490/config/form.js
jbalboni: @goldenmeanie might be the person to ask for this. Is the 5490 applicant information page collecting the names of a veteran or the names of a relative?

cvalarida: I'm pretty sure these are the right ones, but yeah @goldenmeanie can verify.

My understanding is this form is for the dependent to apply for the transfer of benefits because the sponsor veteran is MIA, POW, or deceased and can't transfer the benefit. The applicant in this case is the relative, not the veteran.

src/js/edu-benefits/pages/applicantInformation.js
cvalarida: Ooh, that's the next progression. I like that.

src/js/edu-benefits/5490/config/form.js
cvalarida: You're absolutely right. I fixed that.

src/js/edu-benefits/5490/config/form.js
cvalarida: Right again.

src/js/edu-benefits/5490/config/form.js
goldenmeanie: In both cases (5490 and 1990E) the applicant is the student, not the qualifying Veteran. The Veteran info is collected later in the form. :)

cvalarida: I take that back. It's called `relativeFullName` (and friends) in 1990e.

Is that right, @goldenmeanie?

raquelromano-gov (commented):
This is great--thanks for consolidating these

src/js/edu-benefits/5490/config/form.js
goldenmeanie: if relative == dependent? Then yes that makes sense. But shouldn't it be `relativeFirstName` etc?

cvalarida: I didn't refresh between my first and second comments, so I didn't see your comment, @goldenmeanie. It was a bit redundant, sorry.

Yeah, it is split up like that. `relativeFullName` is a sort of container for all the pieces.

src/js/edu-benefits/pages/applicantInformation.js
```diff
@@ -0,0 +1,61 @@
...
+const defaultProps = [
+  'relativeFullName',
+  'relativeSocialSecurityNumber',
+  'relativeDateOfBirth',
+  'gender',
+  'relationship'
+];
+
+export default function applicantInformation(
+  shema,
```
jbalboni: Looks like this isn't used (and has a typo)? I think you can use it instead of that definitions.json file.

src/js/edu-benefits/pages/applicantInformation.js
```diff
@@ -0,0 +1,59 @@
...
+];
+
+export default function applicantInformation(
+  schema,
+  fields = defaultProps,
+  required = ['relativeFullName']) {
+  const possibleProperties = {
+    relativeFullName: fullName.schema,
+    relativeSocialSecurityNumber: ssn.schema,
+    relativeDateOfBirth: currentOrPastDate.schema,
```
jbalboni: I think you could pull these three fields from the schema, too.

cvalarida: Not the `currentOrPastDate`, right? I was thinking about doing that for the `fullName` and `ssn`, but I figured in those files, they're just being pulled from `definitions.json` which is what the schemas are constructed from, but they may also have additional things (validations) tacked on them.

Also, I think `fullName` is defined in the `properties` of one of the schemas and the `definitions` of another, so I thought it would be inconsistent to have to include the definition in the config only sometimes when I'm doing the exact same thing on two different schemas.

Should I still go ahead and pull these from the `schema` param?

jbalboni: Based on what we talked about before, I think we want to move towards getting rid of the schema being included in our common definition files, if it's not necessary (which it isn't for those three; they just have specific uiSchema objects). So I've been pushing to stop using them in PRs, but haven't updated the files themselves yet.

That includes currentOrPastDate; all the dates have the same schema (the same one as on the backend), only the uiSchemas are different.

I'm not sure what you mean about the properties vs definitions thing. I think both schemas have a common `fullName` definition and several specific fields that reference it. I think you can just use the version from `definitions` here.

cvalarida: Oh, I remember. It was `toursOfDuty` that was inconsistent. In 1995, it's fleshed out in the property, but in 5490, the property is a `$ref` to the definition...which is exactly the same.

At any rate, I'll pull them out!

jbalboni (approved):
LGTM

PR #5066, opened 2017-03-10
5490 education history page

jbalboni:
Related to [713](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/713)

The is the 5490 education history page, with the postHighSchoolTrainings piece shared with 1990e.

screen shot 2017-03-10 at 5 00 35 pm
screen shot 2017-03-10 at 5 00 18 pm
screen shot 2017-03-10 at 5 00 26 pm
screen shot 2017-03-10 at 5 00 05 pm
screen shot 2017-03-10 at 5 00 11 pm


src/js/edu-benefits/5490/config/form.js
```diff
@@ -231,7 +239,128 @@ const formConfig = {
...
+          },
+          schema: {
+            type: 'object',
+            properties: {
+              highSchool: {
+                type: 'object',
+                properties: {
+                  status: highSchool.properties.status,
+                  highSchoolOrGedCompletionDate: date.schema,
+                  'view:hasHighSchool': {
```
cvalarida: Does this need to be flattened before sending to the api?

cvalarida (commented):
So far so good. I just had one question.

src/js/edu-benefits/5490/config/form.js
jbalboni: view: objects are automatically flattened

cvalarida: Good to know! I'll have to check on Monday to see if that's in the docs.

src/js/edu-benefits/utils/helpers.jsx
```diff
@@ -157,3 +158,8 @@ export function showYesNo(field) {
 export function enumToNames(enumValues, names) {
   return enumValues.map(item => names[item]);
 }
+
+export const stateLabels = states.USA.reduce((current, { label, value }) => {
```
cvalarida: Could this be replaced with `_.mapKeys()` or `_.mapValues()`?

cvalarida (approved):
Other than making a wisecrack at
```js
import {
  benefitsLabels,
  relationshipLabels,
  highSchoolStatusLabels,
  transform
} from '../helpers';
```
I don't have much to say; it looks pretty good. I only had one suggestion for clarity.

src/js/edu-benefits/utils/helpers.jsx
jbalboni: I don't think so. Those two transform object -> object, this is going from array -> object.

cvalarida: Ah, so it is. Alright.

PR #5043, opened 2017-03-08
content/wip/stub out new content templates

melwoodard:
Related: https://github.com/department-of-veterans-affairs/vets.gov-content/issues/206

https://github.com/department-of-veterans-affairs/vets.gov-team/issues/938

Add shell templates with placeholder content for the 3 new page templates:

- eligibility
- application-process
- after-you-apply

Templates copied into 3 sections:

- disability-benefits
- education
- healthcare

**Eligibility**
screenshot-localhost-3001-2017-03-07-17-29-21

**Application process**

screenshot-localhost-3001-2017-03-07-17-29-27

**After you apply**

screenshot-localhost-3001-2017-03-07-17-29-31

PR #5068, opened 2017-03-10
Content/wip/edu apply buttons

melwoodard:
This ties into the work in https://github.com/department-of-veterans-affairs/vets-website/pull/5043

Proposed layout update on `/education/apply` to make the use cases easier to scan.

@goldenmeanie @mollyblake notwithstanding further content finessing by Molly, thoughts on the approach? Is this helpful? Is it worth the extra vertical space to lay it out this way?

Current layout as seen at https://vetsgov-pr-5043.herokuapp.com/education/apply/:

image

Proposed change:

image

### DIFF
--- a/content/pages/disability-benefits/apply.md
+++ b/content/pages/disability-benefits/apply.md
@@ -28,10 +28,9 @@ For the first disability claim you file, please provide:
 
 - Discharge papers (DD214 or other separation documents) 
 - Service treatment records
+  - [Order service records through the National Archives](https://www.archives.gov/veterans/military-service-records).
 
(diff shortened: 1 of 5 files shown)


goldenmeanie:
I's be curious to see how it interacts with initial viewport, but in principle I like it.

One note about copy (that Molly is probably all over already)
Apply is apparently more complicated than first-time. If you are switching benefits (e.g. from Ch30 to Ch33) you need to fill out a 1990 again. Something like **To apply for a new benefit** is more accurate.

melwoodard:
Thanks @goldenmeanie! I'm going to merge this and copy over your comments for @mollyblake to address in the main `content-apply-templates` branch

PR #5069, opened 2017-03-11
Make sure the button is a button

cvalarida:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1618

It was defaulting to submit the form, so making sure type="button" ensures that it has no default behavior (i.e. submitting the form).

I have no idea why it was only acting like a submit when it was the first or last item in the array, but there it is.


jbalboni (approved):
LGTM

PR #5071, opened 2017-03-11
Fix label-rendering in review pages

raquelromano-gov:
Closes https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1615

This doesn't address the column width issues--just fixes the problem with values being rendered instead of labels. Screenshot:

image

### DIFF
--- a/src/js/common/schemaform/review/StringField.jsx
+++ b/src/js/common/schemaform/review/StringField.jsx
@@ -14,13 +14,15 @@ import {
  */
 export default function StringField(props) {
   const { registry, schema, uiSchema, formData } = props;
-  const enumOptions = Array.isArray(schema.enum) && optionsList(schema);
+  const uiOptions = getUiOptions(uiSchema);
+  const labels = uiOptions.labels || {};
+  const enumOptions = Array.isArray(schema.enum) && optionsList(
+    _.assign(schema, { enumNames: schema.enum.map(item => labels[item]) }));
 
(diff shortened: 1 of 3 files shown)


src/js/common/schemaform/review/StringField.jsx
```diff
@@ -14,13 +14,15 @@ import {
  */
 export default function StringField(props) {
   const { registry, schema, uiSchema, formData } = props;
-  const enumOptions = Array.isArray(schema.enum) && optionsList(schema);
+  const uiOptions = getUiOptions(uiSchema);
+  const labels = uiOptions.labels || {};
+  const enumOptions = Array.isArray(schema.enum) && optionsList(
```
jbalboni: I think if we're going to override `enumOptions` in here we should remove the code in review/widgets.jsx for `SelectWidget` that checks for `labels` since it won't ever be different from what's in `enumOptions`.

Alternatively, it looks like the bug in the original code is that the `options` prop passed to `Widget` needs to include `labels` as well as `enumOptions`. So I think passing that would also fix this (instead of overriding `enumOptions`).

jbalboni (commented):
Looks good, just had one comment.

src/js/common/schemaform/review/StringField.jsx
raquelromano-gov: Yeah, I realized after sending this that it was a bit silly--changed to follow your suggestion.

raquelromano-gov:
This should be good to merge:
image

PR #5073, opened 2017-03-13
1990e Chapter 5 Employment History

annekainicUSDS:
Closes: https://github.com/department-of-veterans-affairs/vets.gov-team/issues/641

Created a shared schema definition for the employment history fields. Need to set the `expandUnder` property, though. Changed 5490 as well to use the shared definition.

screencapture-localhost-3001-education-apply-for-education-benefits-application-1990e-employment-history-1489414853101

### DIFF
--- a/src/js/common/schemaform/definitions/nonMilitaryJobs.js
+++ b/src/js/common/schemaform/definitions/nonMilitaryJobs.js
@@ -19,7 +19,6 @@ export const uiSchema = {
   'ui:options': {
     itemName: 'Employment Period',
     viewField: EmploymentPeriodView,
-    hideTitle: true,
-    expandUnder: 'view:hasNonMilitaryJobs'
+    hideTitle: true
   }
 };
--- a/src/js/common/schemaform/definitions/nonMilitaryJobs.js
+++ b/src/js/common/schemaform/definitions/nonMilitaryJobs.js
@@ -1,0 +1,25 @@
+import commonDefinitions from 'vets-json-schema/dist/definitions.json';
+
+import EmploymentPeriodView from '../../../edu-benefits/components/EmploymentPeriodView';
+
+export const schema = commonDefinitions.nonMilitaryJobs;
+
+export const uiSchema = {
+  items: {
+    name: {
+      'ui:title': 'Main job'
+    },
+    months: {
+      'ui:title': 'Number of months worked'
+    },
+    licenseOrRating: {
+      'ui:title': 'Licenses or rating'
+    }
+  },
+  'ui:options': {
+    itemName: 'Employment Period',
+    viewField: EmploymentPeriodView,
+    hideTitle: true,
+    expandUnder: 'view:hasNonMilitaryJobs'
+  }
+};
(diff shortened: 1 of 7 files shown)


src/js/common/schemaform/definitions/nonMilitaryJobs.js
```diff
@@ -0,0 +1,24 @@
+import commonDefinitions from 'vets-json-schema/dist/definitions.json';
+
+import EmploymentPeriodView from '../../../edu-benefits/components/EmploymentPeriodView';
+
+export const schema = commonDefinitions.nonMilitaryJobs;
```
jbalboni: I think you can leave this out and make uiSchema the default export.

test/common/schemaform/definitions/nonMilitaryJobs.unit.spec.jsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import { findDOMNode } from 'react-dom';
+import { expect } from 'chai';
+import ReactTestUtils from 'react-addons-test-utils';
+
+import { DefinitionTester } from '../../../util/schemaform-utils.jsx';
+import { schema, uiSchema } from '../../../../src/js/common/schemaform/definitions/nonMilitaryJobs';
+
+describe('Schemaform definition nonMilitaryJobs', () => {
```
jbalboni: The 5490 page has a test for adding another employment history row. Can you move that here?

src/js/edu-benefits/1990e/config/form.js
```diff
@@ -190,6 +192,32 @@ const formConfig = {
...
+          schema: {
+            type: 'object',
+            properties: {
+              employmentHistory: {
+                type: 'object',
+                properties: {
+                  'view:hasNonMilitaryJobs': {
+                    type: 'boolean'
+                  },
+                  nonMilitaryJobs: _.unset('items.properties.postMilitaryJob', nonMilitaryJobs)
```
jbalboni: I need to make an item to get this out of the schema for these two forms, I think.

jbalboni (approved):
Looks good. I made a couple minor comments.

src/js/common/schemaform/definitions/nonMilitaryJobs.js
annekainicUSDS: If I use the default export, do I still name the object `uiSchema`? A little confused on how to use default exports from reading the docs.

annekainicUSDS: Actually removing this import causes the tests to fail. Should I import the schema in the tests alone so that passes, are just keep it here?

test/common/schemaform/definitions/nonMilitaryJobs.unit.spec.jsx
annekainicUSDS: Tried bringing it in and got some test errors, working through them now, but haven't figured it out yet...

src/js/common/schemaform/definitions/nonMilitaryJobs.js
jbalboni: Yeah, you can do the import in the test.

jbalboni: Also, yeah, I think you can still name it uiSchema in this file if it's the default export. Whatever file that's importing it can choose whatever name it wants.

test/common/schemaform/definitions/nonMilitaryJobs.unit.spec.jsx
annekainicUSDS: Think I figured it out! Just pushed a change with the test

PR #5074, opened 2017-03-13
Expound on the view: fields

cvalarida:
Gives another example of what the `view:` fields can be used for.

### DIFF
--- a/docs/schemaform/form-config.md
+++ b/docs/schemaform/form-config.md
@@ -366,3 +366,61 @@ You can use 'ui:description' to show text or a custom component before the field
   }
 }
 ```
+### I want to conditionally hide a group of fields
+
+We may have some fields that are siblings to others, but need to be hidden conditionally. Take the following schema snippet (from the education benefits form 22-5490):
+
+```json
+"previousBenefits": {
+  "type": "object",
+  "properties": {
+    "disability": { "type": "boolean" },
+    "dic": { "type": "boolean" },
+    "chapter31": { "type": "boolean" },
+    "ownServiceBenefits": { "type": "string" },
+    "chapter35": { "type": "boolean" },
+    "chapter33": { "type": "boolean" },
+    "transferOfEntitlement": { "type": "boolean" },
+    "other": { "type": "string" },
+    "veteranFullName": { "$ref": "#/definitions/fullName" },
+    "veteranSocialSecurityNumber": { "$ref": "#/definitions/ssn" }
+  }
+}
+```
+
+Only `chapter35`, `chapter33`, `transferOfEntitlement`, `veteranFullName`, and `veteranSocialSecurityNumber` need to be hidden conditionally, so we can make the `schema` and `uiSchema` like so:
+
+```js
+// schema
+{
+  disability: { ... },
+  dic: { ... },
+  chapter31: { ... },
+  ownServiceBenefits: { ... },
+  'view:sponsorServiceOptions': {
+    chapter35: { ... },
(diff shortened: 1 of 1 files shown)


docs/schemaform/form-config.md
```diff
@@ -66,9 +66,23 @@ Forms are created by creating a page that uses FormApp from the schemaform folde
           // Objects that start with view: will not be sent, but their children will be merged
           // into the parent object and will be sent
           // These can be used to remove fields from what is sent to the server, like if we have
-          // a question that is only used to reveal other questions
+          // a question that is only used to reveal other questions or group related questions
+          // together to be conditionally revealed that aren't in an object in the schema
```
cvalarida: Perhaps this could be explained better? I'm thinking for situations like the 5490's `previousBenefits` page where some options (`chapter35`, `transferOfEntitlement`, etc.`) were revealed under `view:claimedSponsorService`, but siblings of the other options (`chapter31`, `disability`, etc.) in the schema. In that case, I was able to `expandUnder` for each of them, but if `hideIf` is required, these could be grouped in a `view:someName` and the `hideIf` applied to `view:someName` only without affecting the end result sent to the api.

Does this explanation make the above use case clear?

docs/schemaform/form-config.md
```diff
@@ -66,9 +66,23 @@ Forms are created by creating a page that uses FormApp from the schemaform folde
...
           'view:field2': {
             type: 'string'
+          },
+          'view:artificialGroup'{
+            type: 'object',
+            properties: {
+              // view:artificialGroup will be flattened; subField1 and subField2 will be
+              // siblings of field1 when sent to the api
```
cvalarida: If the description above doesn't make it clear, I hope this does the trick.

docs/schemaform/form-config.md
jbalboni: We could add use cases to the cookbook section. Like, "I want to conditionally hide a group of fields" and "I want to group fields together to validate them together" (used for email on the contact info page).

docs/schemaform/form-config.md
```diff
@@ -352,3 +366,61 @@ You can use 'ui:description' to show text or a custom component before the field
...
+    hideIf: () => /* Some condition here */,
+    chapter35: { ... },
+    chapter33: { ... },
+    transferOfEntitlement: { ... },
+    veteranFullName: { ... },
+    veteranSocialSecurityNumber: { ... }
+  },
+  other: { ... }
+}
+```
```
jbalboni: Should add a line here about the backend not seeing the `view:` object.

"When this form is sent to the backend, the fields in the `view:sponsorServiceOptions` object will be moved up one level, i.e. they'll be sent alongside `dic` and `chapter31`. The backend will never see objects with names that start with `view:`, but it will get all the fields inside of them."

Or something similar. It's maybe redundant, but it can't hurt.

jbalboni (approved):
Made one suggestion, but LGTM.

docs/schemaform/form-config.md
```diff
@@ -352,3 +366,61 @@ You can use 'ui:description' to show text or a custom component before the field
   }
 }
 ```
+### I want to conditionally hide a group of fields
```
raquelromano-gov: Super helpful to have the below example in addition to the explanation above!

docs/schemaform/form-config.md
cvalarida: I like that.

Done!

PR #5075, opened 2017-03-13
Use right address field on contact information page for 1990e and 5490

jbalboni:
We're using the wrong address field on 1990e and 5490 for the contact information page. It should be relativeAddress. The 1995 should be veteranAddress.


cvalarida (approved):
Looks good. Good catch.

PR #5077, opened 2017-03-13
Add schema tests for 5490

jbalboni:
Related to [1612](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1612)

Adds schema tests for 5490, following what we did for the 1995. There are some schema updates, too, which I found in creating the tests.


test/edu-benefits/5490/schema/minimal-test.json
```diff
@@ -0,0 +1,96 @@
+{
+  "privacyAgreementAccepted": true,
+  "submission": {
```
raquelromano-gov: Why does this file have a "submission" field but the maximal example doesn't? Trying to follow suit for the 1990e schema tests in https://github.com/department-of-veterans-affairs/vets-website/tree/1990e-schema-test-rr

jbalboni: I just forgot to remove in here. It's not necessary.

raquelromano-gov:
LGTM

test/edu-benefits/5490/schema/highschool-test.json
```diff
@@ -0,0 +1,203 @@
...
+          "dateRange": {
+            "from": "2011-11-01",
+            "to": "2010-12-12"
+          },
+          "name": "Test high school",
+          "city": "Northampton",
+          "state": "MA",
+          "hours": 23,
+          "hoursType": "semester",
+					"degreeReceived": "Diploma"
```
cvalarida: Weird indentation.

test/common/schemaform/helpers.unit.spec.js
```diff
@@ -741,6 +741,33 @@ describe('Schemaform helpers:', () => {
...
+            pages: {
+              page1: {}
+            }
+          }
+        }
+      };
+      const formData = {
+        page1: {
+          data: {
+            someField: {
```
raquelromano-gov: If the form config is empty, does it matter whether these fields are empty, undefined, or defined? Just wondering what scenario this is testing as opposed to
```
formData = {};
```
or 
```
formData = {
  page1: {
    data: {}
  }
};

File: test/edu-benefits/5490/schema/maximal-test.json [hunk 4]
```diff
@@ -0,0 +1,221 @@
...
+          },
+          "serviceStatus": "Great"
+        }
+      ]
+    }
+  },
+  "contributions": {
+    "data": {
+      "civilianBenefitsAssistance": true,
+      "civilianBenefitsSource": "The govt?"
```
cvalarida: Ha! Brilliant!

cvalarida (commented):
Looks good. Lots of data.

test/common/schemaform/helpers.unit.spec.js
jbalboni: There will always be a data object underneath each page's object which has the form data. But inside that it's partly determined by how you filled out the form whether there are empty objects or undefined fields in spots. 

There's a change in this PR to replace empty objects with undefined before sending to the server, so it shouldn't matter in practice, though we could update ObjectField to clean up empty objects, too.

test/edu-benefits/5490/schema/minimal-test.json
```diff
@@ -0,0 +1,89 @@
...
+  "previousBenefits": {
+    "data": {
+      "previousBenefits": {
+        "veteranFullName": {}
+      }
+    }
+  },
+  "sponsorVeteran": {
+    "data": {
+      "relativeFullName": {
```
raquelromano-gov: Just noticed this--shouldn't this be veteranFullName?

cvalarida: Ah, I think you're right, @raquelromano-gov. It's currently `relativeFullName` in the config right now, so I'll fix that up when I'm making the tweaks to the 5490.

PR #1439, opened 2016-03-16
Update service-disabled.md

mollyblake:
Deleted pivotal numbers on links

PR #5079, opened 2017-03-13
Set the environment as a tagmanager variable

aub:
We shouldn't need to pass the environment with every event we record in tagmanager because it's a variable and so only needs to be set once. This follows Google's suggestions about how to preset such a variable. In addition, it stops loading GA at all in development mode.

Closes #1439

Google's documentation on the subject is [here](https://developers.google.com/tag-manager/devguide). On the subject of setting variables, it says: "Each of the variables declared within the data layer object will persist as long as the visitor remains on the current page."

### DIFF
--- a/content/includes/footer.html
+++ b/content/includes/footer.html
@@ -16,7 +16,7 @@
       <dd><a href="https://www.vets.gov/disability-benefits/track-claims/your-claims/">Check disability claim status</a></dd>
       <dd><a href="https://www.vets.gov/gi-bill-comparison-tool/">Education benefits by school</a></dd>
       <dd><a href="https://www.vets.gov/healthcare/prescriptions/">Refill prescriptions</a></dd>
-    </dl>   
+    </dl>
     <ul class="va-footer-linkgroup va-list--plain" id="footer-contact">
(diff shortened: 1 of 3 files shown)


content/includes/footer.html
```diff
@@ -96,11 +96,22 @@ <h4 class="va-crisis-line">Get help from Veterans Crisis Line</h4>
...
 
 <!-- htmllint attr-bans="[]" -->
-<!-- Google Tag Manager -->
-<noscript><iframe src="https://www.googletagmanager.com/ns.html?id=GTM-WFJWBD"
-height="0" width="0" class="csp-inline-patch-footer"></iframe></noscript>
-<script src="/js/google-analytics.js"></script>
-<!-- End Google Tag Manager -->
+{% if buildtype === 'development' %}
+  <script>
+    window.dataLayer = [];
```
robbiethegeek-usds: any reason that this doesn't need the environment: {{ buldtype }}?

aub: In development, this just mocks out the dataLayer so that later calls to `window.dataLayer.push()` don't raise JS errors. We're turning off GA in development, so there's no need to send real data.

robbiethegeek-usds:
LGTM!

PR #5080, opened 2017-03-14
GIBCT Calculator Form

U-DON:
Continuation of the GIBCT calculator WIP from #4993.

Filled in missing inputs and refactored code to display the proper form fields and values depending on the institution attributes and other form values.

Next thing to tackle is to actually compute the estimated benefits and display them. This PR includes a chunk of progress on that as I went through what's already been done to confirm and fix the logic that was copied over from the original jQuery-based GIBCT.

There are many things I would like to refactor, like using the common form elements or downgrading containers to components where full state access isn't necessary. However, those are going to have to remain on the back burner for now. There is also a bunch of styling issues, and those should be addressed after the estimator changes.

> screen shot 2017-03-13 at 11 17 49 pm


src/js/gi/selectors/calculator.js
```diff
@@ -8,16 +10,896 @@ const getEligibilityDetails = (state) => {
...
+    if (isOJT || inputs.calendar === 'quarters') {
+      numberOfTerms = 3;
+    } else if (inputs.calendar === 'semesters') {
+      numberOfTerms = 2;
+    } else if (inputs.calendar === 'nontraditional') {
+      numberOfTerms = +inputs.numberNontradTerms;
+    }
+
+    // Set the net price (Payer of Last Resort) - getTuitionNetPrice
+    const tuitionNetPrice = Math.max(0, Math.min(
```
bshyong: What is this `Math.min` being used for here?

bshyong (approved):
LGTM. This is a huge effort with all of logic for the various calculations. Good work!

src/js/gi/selectors/calculator.js
U-DON: I'm confused why it's being used here too. There's only one arg being passed in to it, so it doesn't look necessary, but leaving in here for now, since I want to leave the original logic untouched aside from obvious naming or operational errors.

saintsoup52: The math-min here is to avoid having the tuition netprice go below a certain threshold... it looks like this logic is missing the other part of the min which would the tuition costs.

U-DON: @saintsoup52 Pinging again in case you missed this.

saintsoup52 (commented):
So now I remember... the other half of the Math.min was just the original tuition/fees.  This was to combat the weird instance someone or something might input a negative number in the scholarship or TA fields. 

This was more of a problem when I originally built this thing in excel... so we could do without the Math.min in this instance.

PR #5081, opened 2017-03-14
Remove unnecessary sponsorVeteran object

jbalboni:
Related to [1667](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1667)

Removes an unnecessary object in one of our page schemas, which will cause the form to fail backend validation.


raquelromano-gov:
LGTM

PR #5083, opened 2017-03-14
WIP: Set lastupdate on markdown files based on git history

aub:
Don't merge this, it's a WIP

Closes department-of-veterans-affairs/vets.gov-team#986


script/build.js
```diff
@@ -136,6 +137,28 @@ smith.use(ignore(ignoreList));
...
+          runningCount -= 1;
+          if (runningCount == 0) {
+            done();
+          }
+        });
+      }
+    }
+  });
+});
+
```
aub: Few notes:

* It's annoying that the file paths don't include the full path, so I have to make the assumption that all markdown content we want to deal with is in `content/pages` and that seems crappy.
* The log listing is async, and I feel like the thing with the running count is super cheesy. Is there a better way to do this?

script/build.js
```diff
@@ -9,6 +9,7 @@ const commandLineArgs = require('command-line-args');
 const dateInFilename = require('metalsmith-date-in-filename');
 const define = require('metalsmith-define');
 const filenames = require('metalsmith-filenames');
+const git = require('simple-git')('.');
```
robbiethegeek-usds: this feels weird...should use __dirname rather then . I think

script/build.js
```diff
@@ -136,6 +137,28 @@ smith.use(ignore(ignoreList));
 // plugin chain.
 smith.use(filenames());
 
+smith.use((files, metalsmith, done) => {
+  runningCount = 0;
+  Object.keys(files).forEach((filename) => {
+    if (filename.match(/\.md$/) !== null) {
+      const fullPath = "content/pages/" + filename;
+      if (fs.existsSync(fullPath)) {
```
robbiethegeek-usds: this feels like an expensive check cause fs is going to check the disk

script/build.js
robbiethegeek-usds: if these were all promises you could do https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all but unless you are going to instantiate a promise and return for each then stuff then into the array of promises and wrap in all... I think this is the simplest.

script/build.js
aub: Turns out it's also totally unnecessary ;)

aub:
@robbiethegeek-usds I think this is done now. Want to give it another look? If good, still don't merge it... I need to figure out the best way to coordinate with the content folks in testing it.

robbiethegeek-usds:
This looks good, would like to open up to anyone else for a 2nd LGTM.  Will link in vets.gov-engineers for another reviewer.

jcmeloni-usds:
@robbiethegeek-usds Did we ever get another reviewer? 17 days later we should either merge or abandon (sadface)

aub:
I need to get it up on a preview system and show it to Danny. Will take care of that this week.

aub:
This PR contains everything we need to complete the requested feature. The problem is that on almost every piece of content the file has been changed recently in order to make updates to the markup rather than the content. In order to get some degree of accuracy around the update dates, the content team will need to make a pass through the various files and set an override date, and that will take some time. I'm adding @rroueche as an assignee and moving this back to the backlog until we can get that process rolling.

rroueche:
@aub want to confirm a few things before moving forward:

1) will that date only display on the public-facing webpage if `template: 4-action-page-pending`?

aub:
@rroueche to answer your questions:

1) I'm attaching an image showing where the last-update thing is included in a handful of the templates. I'm not sure if this maps back to more than `4-action-page-pending` or not.

2) All of the pages that include the `last-update` template would display it, unless the content page has never been changed in git since it was created. In practice almost all of them have been changed.

3) There is currently no way to suppress it, but I think we could add one pretty easily.

rroueche:
thanks @aub !! i don't know if the image came through - could you try again?

rroueche:
@aub pinging for screenshot (or lmk if i'm missing it)

aub:
Whoops, @rroueche I've added the image now

rroueche:
@aub thanks! how about this:

could we have the display of the date suppressed by default, and then use the YAML (or something else, css class for a div? whatever you think) to display the last-updated metadata on the page?

rroueche:
@aub pinging bc we'd ideally like to roll this to prod soon so we have it ready to implement in some new health content we're writing. but we only want the date to show on some select pages, which is why i'm thinking it may be best to default to suppressing the display, w/the ability to indicate in the yaml or an on-page css class or something, that we DO want the data displayed on the page, and also maintaining the ability to override the dynamically supplied data w/a manual override date.

aub:
@rroueche sorry for the delays. If we want to specify them all manually, then you can already do that without the code in this PR. Just add the `lastupdate` field to the front matter on pages you want to show it for.

rroueche:
oh! wow i did not realize that. so is that true of this update as well? that even if we roll this code to prod, the date will not display on any page until it has ```lastupdate``` in the front matter?

aub:
Not quite. If we merged this PR then it would start showing up on almost every page automatically based on the git history of the files, but you'd be able to override it if you wanted. I think that's not what you want.

rroueche:
ok cool! that is correct. i'll document what we want like this:

- want the dynamic last updated date to be captured for every page in the site
- want to be able to manually indicate pages on which the last updated date should display
- want the ability to override the date that displays with a custom date

aub:
Oh, I see. So the difference is that while in this PR it always shows the automatically-generated last update date you want it to only show it when you specifically ask it to. In that case, I'd add an option to the front matter like:

`show_auto_lastupdate: true`

That thing would default to false. In addition, you could still override the automatic date with something like `lastupdate: 2017-05-22`.

Does that sound right?

rroueche:
yes!! that sounds perfect

ayaleloehr:
Pinging on this @aub and @rroueche

aub:
Blargh, this got dealt with in #5614 and should have been closed after that.

PR #5084, opened 2017-03-14
Edu required fields

cvalarida:
Connected to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1640

This updates the required fields in 5490. The only 1990e fields this PR updates is the ones shared with 5490 (I believe they've been broken into another ticket). To get there, I had to do some modifications to get cross-page data passed to the `hideIf` and `'ui:required'` callbacks.

I also modified a couple of the shared pages to accept an `options` object as the second parameter since there was no optimal parameter ordering; in just the two or three forms that would use them, there was no way to avoid at least one form having to reiterate what was already a default parameter just to specify a change in a subsequent parameter. Now, options are just passed in an object and merged with a `defaults` object so we can specify whatever we like in whatever order we need.


src/js/common/schemaform/helpers.js
```diff
@@ -309,7 +309,7 @@ export function getNonArraySchema(schema) {
  * If no required fields are changing, it makes sure to not mutate the existing schema,
  * so we can still take advantage of any shouldComponentUpdate optimizations
  */
-export function updateRequiredFields(schema, uiSchema, formData) {
+export function updateRequiredFields(schema, uiSchema, pageData, formData) {
```
cvalarida: Methinks this naming convention makes more sense; `pageData` for the data on the current page, `formData` for the data in the whole form (whether broken up by page or not, though we should probably have a convention for that).

src/js/common/schemaform/helpers.js
```diff
@@ -318,7 +318,7 @@ export function updateRequiredFields(schema, uiSchema, formData) {
     const newRequired = Object.keys(schema.properties).reduce((requiredArray, nextProp) => {
       const field = uiSchema[nextProp];
       if (field && field['ui:required']) {
-        const isRequired = field['ui:required'](formData);
+        const isRequired = field['ui:required'](pageData, formData);
```
cvalarida: When passing the data to the callback, should we separate them out by `pageData` and `formData`? Can we just send `formData`? Currently, this `formData` isn't broken up by page, so it can be accessed the same way as `pageData`.

Theoretically, this shouldn't cause an issue since all the property names should be unique through all the pages, so when we flatten all page data into a single object, we shouldn't get any overrides, but I went with providing two variables for now in case there was a pressing need to make sure the data we're looking at was really in **this** page.

Perhaps if we keep it two variables, we switch the order?

src/js/edu-benefits/5490/config/form.js
```diff
@@ -241,7 +239,7 @@ const formConfig = {
                 }
               }),
               'ui:options': {
-                hideIf: (fieldData) => fieldData.relationship !== 'spouse'
+                hideIf: (pageData, allData) => _.get('applicantInformation.relationship', allData) !== 'spouse'
```
cvalarida: Just as long as we're talking about naming conventions, should it be `formData` or `allData`? It should be consistent either way.

src/js/edu-benefits/5490/config/form.js
```diff
@@ -312,6 +314,9 @@ const formConfig = {
           },
           schema: {
             type: 'object',
+            // If answered 'Yes' without entering information, it's the same as
+            //  answering 'No' as far as the back end is concerned.
+            required: ['view:applicantServed'],
```
cvalarida: Is this going to be a problem for any reason? Is there a field in the schema I overlooked that this corresponds to? If not, should it really be required?

src/js/edu-benefits/pages/applicantInformation.js
```diff
@@ -6,18 +6,35 @@ import * as ssn from '../../common/schemaform/definitions/ssn';
...
+};
+
+
+/**
+ * Returns an applicantInformation page based on the options passed.
+ *
+ * @param {Object} schema   The full schema for the form
+ * @param {Object} options  Options to override the defaults above
+ */
+export default function applicantInformation(schema, options) {
```
cvalarida: I think we should use this as the preferred way of passing options to the pages because it mitigates need to copy a default parameter if a later parameter needs to be overridden and it's cleaner than specifying a bunch of default parameters in-line.

Thoughts?

cvalarida (commented):
I still have a few tests to fix up, but I wanted to get this out sooner rather than later.

src/js/common/schemaform/reducers/index.js
```diff
@@ -21,12 +21,19 @@ function recalculateSchemaAndData(initialState) {
     .reduce((state, pageKey) => {
       const page = state[pageKey];
       // on each data change, we need to do the following steps
+
+      // Flatten the data from all the pages
+      const formData = _.mapValues('data', state);
```
jbalboni: The reason for the pageData/formData separation is mostly because I wasn't sure about doing something like this. All the processing in here happens every time a user changes data in the form. So it makes me nervous. 

I also sort of feel like we should avoid making a lot of dependencies between the different pages, because it creates forms that are hard to understand. This makes it really easy to do that, and makes it harder to tell when you've done so if you're coming back to the code, which also makes me nervous.

I don't have a firm opinion right now, so I'm open to other thoughts on this.

src/js/edu-benefits/5490/config/form.js
```diff
@@ -176,7 +172,7 @@ const formConfig = {
                 'ui:options': {
                   expandUnder: 'view:claimedSponsorService',
                   updateSchema: (data, form) => {
-                    if (_.get(['previousBenefits', 'data', 'previousBenefits', 'view:claimedSponsorService'], form)) {
+                    if (_.get('previousBenefits.data.previousBenefits.view:claimedSponsorService', form)) {
```
jbalboni: I wasn't sure if the `:` would break this path syntax, but looks like it works!

src/js/edu-benefits/5490/config/form.js
```diff
@@ -201,6 +198,9 @@ const formConfig = {
               previousBenefits: _.merge(
                 _.unset('properties.veteranFullName', previousBenefits),
                 {
+                  // Doesn't let us pass even if view:claimedSponsorService is false
+                  //  (which makes the ssn box is hidden)
+                  // required: ['veteranSocialSecurityNumber'],
```
jbalboni: This is sort of weird, but probably for the best. The data rjsf sends to the json schema validator has all object fields set to empty objects if they have no data (even though they might get stored in our Redux store as undefined). So the validation will always get triggered. This normally works fine, but our conditional stuff makes it a little complicated.

src/js/edu-benefits/5490/config/form.js
jbalboni: formData, I think.

src/js/edu-benefits/5490/config/form.js
jbalboni: It does seem a little weird, but I think the idea is just to force that you read the question and make a choice.

src/js/edu-benefits/pages/applicantInformation.js
jbalboni: Sounds good to me.

src/js/common/schemaform/helpers.js
jbalboni: If we keep the two variables, I think pageData should be first, since it should be the more common case. We shouldn't have a bunch of instances were data from one page is affecting fields in another.

jbalboni (commented):
The changes look good, I just have some uncertainty around making it so easy to create dependencies between pages.

src/js/common/schemaform/helpers.js
```diff
@@ -370,15 +370,15 @@ export const pureWithDeepEquals = shouldUpdate((props, nextProps) => {
  * hideIf function from uiSchema and the current page data. Sets 'ui:hidden'
  * which is a non-standard JSON Schema property
  */
-export function setHiddenFields(schema, uiSchema, data) {
+export function setHiddenFields(schema, uiSchema, data, allData) {
```
raquelromano-gov: Agree that pageData/formData is better than data/allData

src/js/edu-benefits/5490/config/form.js
```diff
@@ -186,8 +182,9 @@ const formConfig = {
               }),
               veteranSocialSecurityNumber: _.merge(ssn.uiSchema, {
                 'ui:title': 'Sponsor SSN',
+                'ui:required': (form) => _.get('previousBenefits.view:claimedSponsorService', form),
                 'ui:options': {
-                  expandUnder: 'view:claimedSponsorService'
+                  expandUnder: 'view:claimedSponsorService',
```
raquelromano-gov: nit: don't need this extra comma

src/js/edu-benefits/5490/config/form.js
raquelromano-gov: Would something like this work here?
https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/js/edu-benefits/1995/config/form.js#L71
I.e., only requiring it if `view:claimedSponsorService` is true

src/js/edu-benefits/5490/config/form.js
raquelromano-gov: I'm not following this line: why does this function need to take both pageData and allData when it only seems to use allData?

src/js/common/schemaform/reducers/index.js
jbalboni: I tried this branch out on my Moto G4 and I don't see any noticeable slow down compared to my testing on our other forms. I didn't really expect this one change to be noticeable, but I hadn't tried out either of the new forms, which are bigger than the 1995, at all yet. 

As far as encouraging dependencies between pages, probably doesn't make sense to make it more awkward for the developer when how the validation works is more of a UX decision. 

I think we should just be cautious when building forms if we start to add lots of complicated dependencies and conditional logic.

jbalboni: So, I think based on this let's just pass one variable, formData, that's a flattened version of the form. We should also update `updateSchema` to be consistent, but that should probably be a separate item because we need to update some existing configs.

raquelromano-gov:
When this gets merged, would you be able to check off the boxes in https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1646 that this fixes?

src/js/common/schemaform/helpers.js
cvalarida: On the other hand, all the data in `pageData` will also be in `formData`; we may not _need_ all the rest of it, but if we do, we wouldn't need to grab two variables when we only need one.

Either way, it's not a big deal.

src/js/edu-benefits/5490/config/form.js
cvalarida: Because the callback sends them in that order, so I can't just grab `allData` without first grabbing `pageData`. Hence my question above.

src/js/common/schemaform/helpers.js
cvalarida: Just read your review below; passing in just `formData` sounds good!

src/js/edu-benefits/5490/config/form.js
cvalarida: Yeah, I thought that might be the case too, so I tested it out and it looks like it's alright!

src/js/common/schemaform/reducers/index.js
cvalarida: Mmm, yeah I didn't want to do that to `updateSchema` because for that, the idea was to actually update the `schema` / `uiSchema` objects inside each page, so it didn't make as much sense.

src/js/edu-benefits/5490/config/form.js
cvalarida: Oh, yeah this works now (just confirmed again), I just forgot to remove the comment after I found `'ui:required'`.

src/js/common/schemaform/reducers/index.js
cvalarida: Just to be certain, I wanted to check the performance of this, so I wrote a little jsPerf using sample data from the form (captured from the redux store with `schema` and `uiSchema` still intact for added realism). This is what I found (note that I wasn't trying to compare the two, so I didn't optimize the native code at all):
image
So while I definitely don't want to keep adding to the reducer code that runs every single keystroke (ack!), I think this addition isn't too bad.

src/js/edu-benefits/5490/config/form.js
raquelromano-gov: Thanks for the explanation!

src/js/common/schemaform/reducers/index.js
```diff
@@ -17,16 +17,30 @@ import { SET_DATA,
...
 
 function recalculateSchemaAndData(initialState) {
-  return Object.keys(_.omit(['privacyAgreementAccepted', 'submission'], initialState))
+  const retVal = Object.keys(_.omit(['privacyAgreementAccepted', 'submission'], initialState))
     .reduce((state, pageKey) => {
       const page = state[pageKey];
       // on each data change, we need to do the following steps
+
+      // Flatten the data from all the pages
+      const formData = Object.keys(state).reduce((carry, pageName) => {
```
cvalarida: This actually flattens it properly; with `_.mapValues()`, it was still broken up by page. Woops. Fortunately, this also clocks in either faster than `_.mapValues()` or insignificantly slower. Either way, it was `console.time()`ing on my machine at ~0.1ms, so it's pretty insignificant.

jbalboni: I wouldn't worry too much about it. I think you need to remove the `submission` object from state, though.

I am sort of curious how this compares (or if it even works):

`Object.assign({}, ...Object.keys(state).filter(key => key === 'submission').map(key => state[key].data))`

cvalarida: `submission` never makes it to `formData` from my tests; it doesn't have `data`. Will there be `state.submission.data` at some point?

jbalboni (approved):
LGTM

src/js/common/schemaform/reducers/index.js
jbalboni: Ah, gotcha. No, we shouldn't.

cvalarida: As far as speed goes, they're close enough (and fast enough) it doesn't really make a difference, but not using the object spread is quicker. Taking out the `.filter()` speeds it up, but only by about 0.02ms.

It just comes down to readability. Which is clearer? Does it really matter with the explanatory comment?

jbalboni: Mine's really just code golfing, no need to change what you have.

PR #5085, opened 2017-03-14
Update confirmation page for 5490 and 1990e

jbalboni:
Related to [724](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/724)

Updates both confirmation pages so they work and display data. Copy may need to change in the future, but they work for now.

I also tweaked the review page styling to have less variation in the column size between rows.

screen shot 2017-03-14 at 4 28 26 pm
screen shot 2017-03-14 at 4 27 21 pm

Review page example:

screen shot 2017-03-14 at 4 49 23 pm


src/js/edu-benefits/1990e/containers/ConfirmationPage.jsx
```diff
@@ -31,8 +36,79 @@ class ConfirmationPage extends React.Component {
...
+          <span>for {name.first} {name.middle} {name.last} {name.suffix}</span>
+
+          <ul className="claim-list">
+            <li>
+              <strong>Selected benefit</strong><br/>
+              {benefitsLabels[benefit]}
+            </li>
+            <li>
+              <strong>Confirmation number</strong><br/>
+              <span>{response.confirmationNumber}</span>
```
raquelromano-gov: This looks like it's not rendering properly in your screenshot, but maybe that's an artifact of the dev environment?

raquelromano-gov (commented):
LGTM

src/js/edu-benefits/1990e/containers/ConfirmationPage.jsx
jbalboni: Yeah, that's the number in dev/staging, it should be a more appropriate looking string in prod. Though I think Alex was opening an issue about making sure that is the case.

PR #5089, opened 2017-03-15
Add a target for the linter to only work on changed files

aub:
This adds the option for `npm lint:js:changed`, which only operates on files that have updates in git. It will make it a lot faster to lint your changes before pushing them up.

Closes department-of-veterans-affairs/vets.gov-team#1244

### DIFF
--- a/script/build.js
+++ b/script/build.js
@@ -32,8 +32,8 @@ const minimumNpmVersion = '3.8.9';
 const minimumNodeVersion = '4.4.7';
 // Make sure git pre-commit hooks are installed
 ['pre-commit'].forEach(hook => {
-  const src = `../hooks/${hook}`;
-  const dest = `../.git/hooks/${hook}`;
+  const src = path.join(__dirname, `../hooks/${hook}`);
+  const dest = path.join(__dirname, `../.git/hooks/${hook}`);
   if (fs.existsSync(src)) {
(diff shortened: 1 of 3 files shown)


jbalboni:
@aub I think it would make sense to change the pre-commit hook to use this (hooks/pre-commit). I think you'd need to make another npm script that includes the sass linter, though.

Looks like the build script links that hook, so everyone should get the update the next time they do a build or restart the watch task.

aub:
@jbalboni good call. How's it look now?

script/build.js
```diff
@@ -32,8 +32,8 @@ const minimumNpmVersion = '3.8.9';
 const minimumNodeVersion = '4.4.7';
 // Make sure git pre-commit hooks are installed
 ['pre-commit'].forEach(hook => {
-  const src = `../hooks/${hook}`;
-  const dest = `../.git/hooks/${hook}`;
+  const src = path.join(__dirname, `../hooks/${hook}`);
```
jbalboni: @aub Apparently the linking code was broken at some point. I removed my git hook and it wasn't replacing it, because it was looking for the file in a place that didn't exist. This fixes it.

jbalboni (approved):
LGTM. I pushed one change to fix how where we're looking for the hook files.

PR #5090, opened 2017-03-15
Fix onBlur error message rendering

raquelromano-gov:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1646

@jbalboni I spent some time yesterday stepping through code to understand why our dateRange validations in the 1995, 1990e, and 5490 aren't triggered until submit (unlike the other validations, e.g., ssn, date, which are rendered onChange). Made some superficial changes here, but did not get to the bottom of what's happening--any ideas?


jbalboni:
@raquelromano-gov Do you have an example of where it's not working? The only way I've found where it doesn't work the way you'd expect is if you fill in the year first on a required date.

jbalboni:
@raquelromano-gov It looks like it was a problem with our ArrayFields, rather than date ranges specifically. When I moved us off the forked version of rjsf, I missed updating part of the ArrayField component. So it was passing bad id values up to the the SchemaForm component and was never marking the fields as touched.

raquelromano-gov:
Examples:
  * 1995: https://staging.vets.gov/education/apply-for-education-benefits/application/1995/military-history/service-periods
  * 1990e: https://staging.vets.gov/education/apply-for-education-benefits/application/1990e/education-history
  * 5490: https://staging.vets.gov/education/apply-for-education-benefits/application/5490/military-service/sponsor-veteran

Screenshots:
image
---
image
---
image

raquelromano-gov:
Yeah, saw that--thanks! Would love to know at some point how you tracked that down.

Just updated the tests, so ready for another look.

src/js/common/schemaform/definitions/dateRange.js
```diff
@@ -17,7 +17,7 @@ export function uiSchema(from = 'From', to = 'To', rangeError = 'To date must be
       validateDateRange
     ],
     'ui:errorMessages': {
-      dateRange: rangeError,
+      pattern: rangeError,
```
jbalboni: I'm curious about the reason for this change. This is the error key used by JSON Schema when a field fails a regex pattern match. It won't cause any problems, but it seems less clear to use it here?

jbalboni (approved):
LGTM, had one question.

src/js/common/schemaform/definitions/dateRange.js
raquelromano-gov: Just following suit with all the rest of the 'ui:errorMessages' properties in our ui schema definitions (date, ssn, bankAccount, currentOrPastDate). No reason other than consistency.

PR #5093, opened 2017-03-15
Replace ref schemas on start up

jbalboni:
Related to [1578](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1578)

This updates the reducer start up code to replace all instances of $ref with the definition from defaultDefinitions (or a page's definitions object). I think this will let us use more of the backend schemas, since one of the reasons we avoid them sometimes is that we need to conditionally modify a field's schema, which doesn't really work if it's a $ref field.

The performance penalty on a slow device on startup looks like about 15ms, which is about the same amount of time it takes to do the hidden fields or schema update processing (which makes sense, because they're very similar code). Since this is only happening on startup, and not each time form data is changed (as with the other processing), I think that's a reasonable tradeoff.


raquelromano-gov:
I looked over this and looks OK though I honestly need to better understand src/js/common/schemaform/reducers/index.js and how it and our schemaform components work together.

Will this allow us to simplify some config files as a follow-up?

jbalboni:
@raquelromano-gov It should, but mostly around fields that have conditional logic. So maybe not a huge impact.

raquelromano-gov:
LGTM

cvalarida (approved):

cvalarida:
I started that review yesterday, but apparently I never submitted it. Sorry.

src/js/common/schemaform/helpers.js
```diff
@@ -539,3 +539,41 @@ export function setItemTouched(prefix, index, idSchema) {
     return _.merge(idObj, setItemTouched(prefix, index, idSchema[field]));
   }, {});
 }
+
+export function replaceRefSchemas(schema, definitions, path = '') {
+  if (schema.$ref) {
```
jbalboni: If we start building schemas from classes and/or with prototypes, I will complain loudly. :-)

cvalarida: As well you should. =)

PR #5094, opened 2017-03-15
GIBCT keyword search

bshyong:
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1536

Users can press enter or select a result from the autocomplete suggestions to trigger a new search. Additionally, the search terms are reflected in the URL and will be persisted upon a page refresh to the same URL.


src/js/gi/components/search/KeywordSearch.jsx
```diff
@@ -15,6 +15,33 @@ export class KeywordSearch extends React.Component {
...
+  componentDidMount() {
+    const searchQuery = this.props.location && this.props.location.query;
+    if (searchQuery) {
+      this.handleChange(null, { newValue: searchQuery.name, method: 'enter' });
+    }
+  }
+
+  onKeyUp(e) {
+    const { onFilterChange, autocomplete } = this.props;
+    if ((e.which || e.keyCode || 0) === 13) {
```
U-DON: Is the `|| 0` here necessary?

src/js/gi/components/search/KeywordSearch.jsx
```diff
@@ -15,6 +15,33 @@ export class KeywordSearch extends React.Component {
...
+      onFilterChange('name', autocomplete.searchTerm);
+    }
+  }
+
+  handleChange(event, data) {
+    this.props.updateAutocompleteSearchTerm(data.newValue);
+  }
+
+  handleSuggestionSelected(event, data) {
+    this.props.onFilterChange('name', data.suggestionValue);
```
U-DON: Any issues when `onFilterChange` isn't provided, like on the `LandingPage`? Presumably not, but just wanted to check.

U-DON (approved):
LGTM. Had some minor comments.

src/js/gi/components/search/KeywordSearch.jsx
bshyong: Good catch; I set a no-op as a default prop for this to prevent any errors on the LandingPage.

PR #5097, opened 2017-03-16
Updates for chapter 2 on 5490

jbalboni:
Related to [1663](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1663) and [1660](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1660)

screen shot 2017-03-16 at 2 26 13 pm

### DIFF
--- a/docs/schemaform/form-config.md
+++ b/docs/schemaform/form-config.md
@@ -196,6 +196,13 @@ In addition to the uiSchema options listed in the library [https://github.com/mo
     labels: {
       chapter30: 'A readable description (Chapter 30)'
     },
+
+    // This is a map of values to a component, some text, or some jsx. If your field
+    // is a radio widget, the content here will be shown underneath the radio button
+    // for that value when it is selected.
+    nestedContent: {
+      'value': <p>Some text</p>
+    },
     
     // This is a string of class names that will be added to the widget for the current
     // field. Similar to the default `classNames` property, but will put the class names
--- a/src/js/common/schemaform/widgets/RadioWidget.jsx
+++ b/src/js/common/schemaform/widgets/RadioWidget.jsx
@@ -1,5 +1,7 @@
 import React from 'react';
 
+import ExpandingGroup from '../../components/form-elements/ExpandingGroup';
+
 export default function RadioWidget({
   options,
   value,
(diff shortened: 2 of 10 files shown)


goldenmeanie:
LGTM!

goldenmeanie:
Oops! Should have the conditional group format.

src/js/common/schemaform/widgets/RadioWidget.jsx
```diff
@@ -1,19 +1,29 @@
...
   disabled,
   onChange,
   id
 }) {
-  const { enumOptions, labels = {} } = options;
+  const { enumOptions, labels = {}, nestedContent = {} } = options;
+
+  // nested content could be a component or just jsx/text
+  let content = nestedContent[value];
+  if (typeof content === 'function') {
```
cvalarida: Is this the best way to test if a variable is a React component? I haven't found a better one... :confused:

jbalboni: As far as I know, it is.

docs/schemaform/form-config.md
```diff
@@ -196,6 +196,13 @@ In addition to the uiSchema options listed in the library [https://github.com/mo
     labels: {
       chapter30: 'A readable description (Chapter 30)'
     },
+
+    // This is a map of values to a component, some text, or some jsx. If your field
+    // is a radio widget, the content here will be shown underneath the radio button
+    // for that value when it is selected.
+    nestedContent: {
```
jbalboni: This is only for radio buttons. They're a single field/widget, so the only way to put content in between the options is some kind of option the widget can read.

jbalboni: I'm open to suggestions on how to unify those different approaches. The `view:descriptionBox` with an empty schema thing we've been doing feels hackish, so I'd love to get rid of it, but I don't have a better approach off hand for what we're using it for off hand.

cvalarida: Makes sense. I think we should have a blurb in the docs defining the differences between the different ways to expand / hide content, but that could be in a different PR.

cvalarida:
@jbalboni When this gets merged, could you check off the appropriate item in https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1683?

PR #5099, opened 2017-03-16
Add more changes and fix failing test

annekainicUSDS:
Closes: https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1657 (Add VA file number is no SSN is going to be done in a separate ticket)


annekainicUSDS:
Adding some more tests for this...

src/js/edu-benefits/pages/schoolSelection.js
```diff
@@ -6,15 +6,16 @@ import { uiSchema as uiSchemaDate } from '../../common/schemaform/definitions/da
 export default function createSchoolSelectionPage(
   schema,
   fields = ['educationProgram', 'educationObjective', 'nonVaAssistance'],
-  programRequired = false) {
+  programRequired = false,
```
raquelromano-gov: Maybe there should be a 1:1 correspondence between these parameters and the fields they affect (see below).

src/js/edu-benefits/pages/schoolSelection.js
```diff
@@ -41,11 +42,19 @@ export default function createSchoolSelectionPage(
   const pickFields = _.pick(fields);
 
   const schemaProperties = pickFields(schema.properties);
+
+  let requiredFields = [];
+  if (programRequired) {
```
raquelromano-gov: E.g. just like educationTypeRequired will cause the 'educationType' field to be required, we can have a function argument programNameRequired that will make the 'name' field required.

annekainicUSDS: Hm, not sure I fully understand because it's not always a 1:1 relationship. `programRequired` requires 2 fields. I'm thinking perhaps we should instead pass an array of required fields as a parameter instead of booleans to require specific fields?

annekainicUSDS:
@raquelromano-gov made updates to the PR!

test/edu-benefits/pages/schoolSelection.unit.spec.jsx
```diff
@@ -52,6 +52,40 @@ describe('Edu 1990e schoolSelection', () => {
...
+          schema={schemaIfEducationType}
+          onSubmit={onSubmit}
+          data={{}}
+          definitions={formConfig.defaultDefinitions}
+          uiSchema={uiSchemaIfEducationType}/>
+    );
+    const formDOM = findDOMNode(form);
+    const find = formDOM.querySelector.bind(formDOM);
+
+    ReactTestUtils.Simulate.change(find('#root_educationProgram_educationType'), {
```
raquelromano-gov: How about verifying that
  - an error is rendered
  - onSubmit isn't called
and then providing the education type

raquelromano-gov (approved):
Just one suggestion for beefing up the unit test.

test/edu-benefits/pages/schoolSelection.unit.spec.jsx
annekainicUSDS: Yes! Totally reasonable. Just added those changes. Let me know if this is good to go then.

raquelromano-gov: LGTM

PR #5100, opened 2017-03-16
Fix facility locator styling issues in IE11

bshyong:
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1644

- Fixed issue with overly broad css selector for external links
- Fixed issue where IE11 has unexpected behavior for flex styling on elements with a minimum height


U-DON (approved):
LGTM.

PR #5101, opened 2017-03-16
Use map view bounding box instead of mapbox bounding box for FL

bshyong:
resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1675


U-DON (approved):
LGTM.

PR #5104, opened 2017-03-17
Allow VA file number as alternate to SSN in 1995 and 1990e

raquelromano-gov:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1687

I'm getting some failing tests (1995 and 1990e config tests are complaining that veteranId is undefined), but wanted to get feedback on this in the meantime.

This relies on a fix to vets-json-schema (@annekainicUSDS I added a comment to https://github.com/department-of-veterans-affairs/vets-json-schema/pull/71 after it was merged if you can take a look).

And this schema and UI should also be use din the 5490, but I left that out for now since that backend schema also needs an update.

1990e
image

image

1995
image

image


src/js/edu-benefits/1990e/config/form.js
```diff
@@ -117,18 +116,20 @@ const formConfig = {
                 }
               }
             },
-            veteranSocialSecurityNumber: _.set(['ui:title'], 'Veteran Social Security number', ssnUiSchema),
+            // There must be a more succinct way for lodash to set 2 properties
+            veteranId: _.set(['vaFileNumber', 'ui:title'], 'Veteran File number',
```
jbalboni:
```js
_.merge(veteranId.uiSchema, {
  vaFileNumber: {
    'ui:title': 'Veteran file number',
  },
  veteranSocialSecurityNumber: {
    'ui:title': 'Veteran Social Security number'
  }
})
```

src/js/edu-benefits/1990e/config/form.js
```diff
@@ -117,18 +116,20 @@ const formConfig = {
...
             }
           },
           schema: {
             type: 'object',
-            required: ['veteranFullName', 'veteranSocialSecurityNumber'],
+            required: ['veteranFullName'],
             properties: {
               veteranFullName: fullName,
-              veteranSocialSecurityNumber: ssn,
+              veteranId: veteranId.schema,
```
jbalboni: This should be `view:veteranId`, I think, so that we don't send a veteranId object to the backend.

src/js/edu-benefits/definitions/veteranId.js
```diff
@@ -0,0 +1,44 @@
...
+      expandUnder: 'view:noSSN'
+    }
+  }
+};
+
+// Placeholder in case we need to pass in customizations for various edu forms
+export const schema = {
+  type: 'object',
+  properties: {
+    veteranSocialSecurityNumber: ssn.schema,
```
jbalboni: On the 5490, this is going to be relativeSocialSecurityNumber, I think. That's the field being used on the first page. If that's the case, we'll have to turn these two into functions, probably. I don't think we have to do that now, though.

src/js/edu-benefits/definitions/veteranId.js
```diff
@@ -0,0 +1,44 @@
+import _ from 'lodash/fp';
+
+import * as ssn from '../../common/schemaform/definitions/ssn';
+
+// TODO: move vaFileNumber out of 1995 and into a shared backend definition
```
jbalboni: In other pages/definitions we've been passing in the schema to a function, so that's a possibility, too.

src/js/edu-benefits/definitions/veteranId.js
```diff
@@ -0,0 +1,44 @@
...
+import fullSchema1995 from 'vets-json-schema/dist/change-of-program-schema.json';
+const {
+  vaFileNumber
+} = fullSchema1995.definitions;
+
+
+export const uiSchema = {
+  veteranSocialSecurityNumber: _.assign(ssn.uiSchema, {
+    'ui:required': (form) => !form.veteranId['view:noSSN']
```
cvalarida: When changing from `veteranId` to `view:veteranId`, this (and line 23 below) needs to change too.
```js
!form['view:veteranId']['view:noSSN']
```

raquelromano-gov: @jbalboni I'm tracking down why this change breaks a bunch of unit tests, and I think it's this line. Seems to work fine in the app (I inspected the code [here](https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/js/common/schemaform/reducers/index.js#L36) and [here](https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/js/common/schemaform/helpers.js#L329) and all looks fine)

I'm wondering why `form.veteranId` appears to be undefined in [DefinitionTester](https://github.com/department-of-veterans-affairs/vets-website/blob/master/test/util/schemaform-utils.jsx#L22) and if you have any ideas for what to investigate.

Also curious how you debug test failures, other than intuition :)

Here's the stack trace for one of the failing tests:
```
     TypeError: Cannot read property 'view:noSSN' of undefined
      at Object.uiRequired [as ui:required] (src/js/edu-benefits/definitions/veteranId.js:14:31)
      at src/js/common/schemaform/helpers.js:329:28
      at Array.reduce (native)
      at updateRequiredFields (src/js/common/schemaform/helpers.js:326:56)
      at src/js/common/schemaform/helpers.js:346:28
      at Array.reduce (native)
      at updateRequiredFields (src/js/common/schemaform/helpers.js:344:54)
      at new DefinitionTester (test/util/schemaform-utils.jsx:22:18)
```

jbalboni: Most of the time we use `_.get` here for nested properties, since nested objects are sometimes undefined (it depends on how you interact with the form).

`_.get('veteranId.view:noSSN', form)` will return undefined if the veteranId object doesn't exist.

src/js/edu-benefits/definitions/veteranId.js
raquelromano-gov: We could, but in this case the schemas will be identical across all 3 forms, so it seems we should just use a common vaFileNumber definition from the backend.

test/edu-benefits/1990e/config/sponsorVeteran.unit.spec.jsx
```diff
@@ -60,8 +60,7 @@ describe('Edu 1990e sponsorVeteran', () => {
         value: 'Veteran'
       }
     });
-
-    ReactTestUtils.Simulate.change(find('#root_veteranSocialSecurityNumber'), {
+    ReactTestUtils.Simulate.change(find('#root_view:veteranId_veteranSocialSecurityNumber'), {
```
raquelromano-gov: These paths aren't working...trying to fix.

src/js/edu-benefits/1990e/config/form.js
raquelromano-gov: Thanks!

src/js/edu-benefits/1990e/config/form.js
raquelromano-gov: Done, here and elsewhere.

src/js/edu-benefits/definitions/veteranId.js
raquelromano-gov: Yeah, will wait for the dust to settle on the great sponsor/veteran/applicant/relative SSN/file# debate before worrying about that.

src/js/edu-benefits/definitions/veteranId.js
raquelromano-gov: Thanks!! That did the trick.

test/edu-benefits/1990e/config/sponsorVeteran.unit.spec.jsx
cvalarida: If I recall correctly, it's the `:` in the id. To get around this, we've had to do something like:
```js
Array.from(find('input')).find((input) => input.id === 'root_view:veteranId_veteranSocialSecurityNumber')
```

raquelromano-gov: I was just now spelunking to find how other view: type elements were tested.  THanks! Will try this.

test/edu-benefits/definitions/veteranId.unit.spec.jsx
```diff
@@ -0,0 +1,87 @@
...
+      {
+        target: {
+          checked: true
+        }
+      });
+
+    // TODO: figure out why the below assertions fail; checking the box should make SSN error disappear
+    // and VA file number error appear
+    // No error is shown for empty SSN input; error is shown for empty file number input
+    // expect(formDOM.querySelector('.usa-input-error #root_veteranSocialSecurityNumber')).to.be.null;
```
raquelromano-gov: I have no idea why these 2 assertions fail, since once submit is called and the SSN box is checked, the error message should disappear from SSN and appear for FN.

raquelromano-gov:
These tests were super finicky due to the "view:" in the node ids. Still have one problem (see comment in veteranId test), but I think this should be merged and I can come back to that later.

test/edu-benefits/definitions/veteranId.unit.spec.jsx
jbalboni: Yeah, I don't see any reason, either. Can you create an item to come back and figure that out?

src/js/edu-benefits/1990e/config/form.js
```diff
@@ -51,7 +50,8 @@ const formConfig = {
   defaultDefinitions: {
     date,
     dateRange,
-    educationType
+    educationType,
+    veteranId
```
jbalboni: You shouldn't need this.

jbalboni (approved):
LGTM

src/js/edu-benefits/1990e/config/form.js
raquelromano-gov: You're right--removed it from here, and removed including these in unit tests.

PR #5107, opened 2017-03-17
Estimated benefits for GIBCT calculator

U-DON:
Resolves department-of-veterans-affairs/vets.gov-team#1064.

Basically picks out the corresponding values from `getDerivedValues` and translates them into displayed rows.

> screen shot 2017-03-16 at 11 30 12 pm


src/js/gi/components/profile/Calculator.jsx
```diff
@@ -16,34 +20,39 @@ const EligibilityDetails = ({ expanded, toggle }) => (
...
+const CalculatorInputs = ({ expanded, toggle, inputs, displayedInputs, onInputChange, onShowModal }) => (
   <div className="calculator-inputs">
     <button onClick={toggle} className="usa-button-outline">
       {expanded ? 'Hide' : 'Edit'} calculator fields
     </button>
     <div className="form-expanding-group-open">
-      {expanded ? <CalculatorForm/> : null}
+      {expanded ? <CalculatorForm
+          inputs={inputs}
+          displayedInputs={displayedInputs}
```
U-DON: Will combine these at some point similar to how `calculatedBenefits.outputs` is formatted.

src/js/gi/selectors/calculator.js
```diff
@@ -183,15 +177,18 @@ const getDerivedValues = createSelector(
...
       tuitionFeesCap = constant.CORRESPONDTFCAP;
     } else if (isPublic && institution.country === 'usa') {
-      tuitionFeesCap = inputs.inState
+      tuitionFeesCap = inputs.inState === 'yes'
                      ? institution.tuitionInState
                      : institution.tuitionOutOfState;
     } else if (isPrivate || isForeign) {
       tuitionFeesCap = constant.TFCAP;
+    } else {
+      // TODO: What should default cap be?
```
U-DON: Not sure what the `tuitionFeesCap` should be by default. If it's not assigned, it remains `undefined`, and there are `Math.min` operations further down that include it, which results in `NaN` for values we need to output in the estimated benefits section. For now, setting it to `constant.TFCAP`.

src/js/gi/utils/helpers.js
```diff
@@ -7,3 +7,7 @@ export function getCurrency(value) {
 
   return isNumeric(amount) ? +amount : 0;
 }
+
+export function formatCurrency(value) {
+  return `$${Math.round(value)}`;
```
U-DON: Wanted to use `Intl.NumberFormat` here, but couldn't get it to work. And didn't want to pull in a full `react-intl` package just for this.

src/js/gi/selectors/calculator.js
U-DON: @saintsoup52: Thoughts on this?

src/js/gi/utils/helpers.js
bshyong: this should be ok. I think we miss out on the automatic comma separators but it's not a big deal. I also think it lacks support in a few mobile browsers, notably Android.

src/js/gi/selectors/calculator.js
```diff
@@ -321,9 +327,12 @@ const getDerivedValues = createSelector(
     }
 
     // getTuitionFeesTerm3
-    if (isOJT || oldGiBill || (giBillChapter === 31 && isFlightOrCorrespondence)) {
-      tuitionFeesTerm3 = 0;
-    } else if (inputs.calendar === 'semesters' || inputs.calendar === 'nontraditional' && numberOfTerms < 3) {
+    const shouldHaveNoTuitionFeesTerm3 =
```
bshyong: 👍

bshyong (approved):
LGTM

src/js/gi/selectors/calculator.js
saintsoup52: I think the thousand commas are really important to the readability of the tool... Does this affect affect those? I just saw a screen shot where they were missing and it was hard to read the values.

https://github.com/department-of-veterans-affairs/vets-website/pull/5107

src/js/gi/selectors/calculator.js
saintsoup52: TFCAP is one of the most important constants in the tool... it is the private school tuition cap for the Post-9/11 GI Bill.

TFCAP =21970.46

It can be found in the current version of the tool in two places:

**Estimator (LINE 31)**
https://github.com/department-of-veterans-affairs/gi-bill-comparison-tool/blob/master/app/assets/js/estimator.js

image

**Calculator (LINE 54)**
https://github.com/department-of-veterans-affairs/gi-bill-comparison-tool/blob/production/app/assets/js/calculator.js

image

U-DON: @saintsoup52 Sure, that constant gets applied for private or foreign school types here, just like the original GIBCT. But how we should handle cases that don't apply to any of the conditions here is what I was wondering. If it's not flight, correspondence, public, private, or foreign, there ends up being an undefined cap. In particular, if it's a for-profit.

saintsoup52:
The TFCAP for "For Profit" schools is the same as private and foreign schools.

That should be covered in the isprivate function.

src/js/gi/selectors/calculator.js
U-DON: Will be addressed by #5123.

src/js/gi/selectors/calculator.js
saintsoup52: The TFCAP for "For Profit" schools is the same as private and foreign schools. 

That should be covered in the isprivate function.

PR #5112, opened 2017-03-17
Timestamp assets

aub:
This adds a timestamp to the end of asset filenames. There is a bug where the contents are being updated but not the hashes, and this will fix that by allowing caching within a build but busting the cache between builds.


jkassemi (approved):
Looks good to me. We'll pass on https://github.com/department-of-veterans-affairs/devops/pull/1593 since this addresses the issue.

We need to get a production deployment out containing this fix.

PR #5116, opened 2017-03-17
5490 Chapter 3 updates

jbalboni:
Related to [1664](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1664)

Add "Veteran" to field labels (and fix DOB case):
screen shot 2017-03-17 at 5 00 27 pm

Make tour of duty fields required:
screen shot 2017-03-17 at 4 27 56 pm
screen shot 2017-03-17 at 4 28 03 pm

Bold text:
screen shot 2017-03-17 at 2 32 56 pm

### DIFF
--- a/src/js/common/schemaform/ArrayField.jsx
+++ b/src/js/common/schemaform/ArrayField.jsx
@@ -40,6 +40,10 @@ export default class ArrayField extends React.Component {
     this.scrollToRow = this.scrollToRow.bind(this);
   }
 
+  // This fills in an empty item in the array if it has minItems set
+  // so that schema validation runs against the fields in the first item
+  // in the array. This shouldn't be necessary, but there's a fix in rjsf
+  // that has not been released yet
   componentDidMount() {
     const { schema, formData = [], registry } = this.props;
     if (schema.minItems > 0 && formData.length === 0) {
--- a/src/js/common/schemaform/ArrayField.jsx
+++ b/src/js/common/schemaform/ArrayField.jsx
@@ -40,6 +40,15 @@ export default class ArrayField extends React.Component {
     this.scrollToRow = this.scrollToRow.bind(this);
   }
 
+  componentDidMount() {
+    const { schema, formData = [], registry } = this.props;
+    if (schema.minItems > 0 && formData.length === 0) {
+      this.props.onChange([
+        getDefaultFormState(schema.items, undefined, registry.definitions)
+      ]);
+    }
+  }
+
   shouldComponentUpdate(nextProps, nextState) {
     return !deepEquals(this.props, nextProps) || nextState !== this.state;
   }
--- a/src/js/common/schemaform/ArrayField.jsx
+++ b/src/js/common/schemaform/ArrayField.jsx
@@ -47,9 +47,9 @@ export default class ArrayField extends React.Component {
   componentDidMount() {
     const { schema, formData = [], registry } = this.props;
     if (schema.minItems > 0 && formData.length === 0) {
-      this.props.onChange([
+      this.props.onChange(Array(schema.minItems).fill(
         getDefaultFormState(schema.items, undefined, registry.definitions)
-      ]);
+      ));
     }
   }
 
--- a/src/js/edu-benefits/5490/config/form.js
+++ b/src/js/edu-benefits/5490/config/form.js
@@ -275,8 +275,8 @@ const formConfig = {
             veteranSocialSecurityNumber: _.assign(ssn.uiSchema, {
               'ui:title': 'Veteran Social Security number'
             }),
-            veteranDateOfBirth: currentOrPastDate.uiSchema('Date of birth'),
-            veteranDateOfDeath: currentOrPastDate.uiSchema('Date of death or date listed as MIA or POW'),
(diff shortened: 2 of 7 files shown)


src/js/edu-benefits/5490/config/form.js
```diff
@@ -272,7 +288,7 @@ const formConfig = {
             ],
             properties: {
               spouseInfo,
-              relativeFullName: fullName,
+              veteranFullName: fullName,
```
jbalboni: I think this was a bug.

src/js/edu-benefits/5490/config/form.js
```diff
@@ -320,6 +336,7 @@ const formConfig = {
               'ui:options': {
                 expandUnder: 'view:applicantServed'
               },
+              'ui:required': _.get('view:applicantServed'),
```
jbalboni: I don't know how anyone feels about the clarity of this, but since lodash/fp auto-curries and undefined is falsy, it works here. Open to changing it.

src/js/common/schemaform/ArrayField.jsx
```diff
@@ -40,6 +40,19 @@ export default class ArrayField extends React.Component {
...
 
+  // This fills in an empty item in the array if it has minItems set
+  // so that schema validation runs against the fields in the first item
+  // in the array. This shouldn't be necessary, but there's a fix in rjsf
+  // that has not been released yet
+  componentDidMount() {
+    const { schema, formData = [], registry } = this.props;
+    if (schema.minItems > 0 && formData.length === 0) {
+      this.props.onChange([
+        getDefaultFormState(schema.items, undefined, registry.definitions)
```
cvalarida: Maybe it's a moot point if this isn't necessary, but if `schema.minItems == 2` and this only sends in one default item (from what I can tell), will that be a problem that should be avoided?

src/js/edu-benefits/5490/config/form.js
```diff
@@ -248,19 +249,34 @@ const formConfig = {
               },
               remarriageDate: _.assign(date.uiSchema('Date you got remarried'), {
                 'ui:options': {
-                  hideIf: (formData) => !_.get('spouseInfo.remarried', formData)
-                }
+                  expandUnder: 'remarried',
```
cvalarida: Ah, that's better.

src/js/edu-benefits/5490/config/form.js
```diff
@@ -248,19 +249,34 @@ const formConfig = {
...
               'ui:options': {
                 hideIf: (formData) => _.get('relationship', formData) !== 'spouse'
               }
             },
-            relativeFullName: _.assign(fullNameUISchema, {
-              'ui:title': 'Name of Sponsor'
+            veteranFullName: _.merge(fullNameUISchema, {
+              'ui:title': 'Name of Sponsor',
+              first: {
+                'ui:title': 'Veteran first name'
```
cvalarida: May want to relabel these using "Sponsor" per https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1696

I think to avoid confusion if the applicant is also a veteran. "Wait, which veteran? My sponsor or me?"

src/js/edu-benefits/5490/config/form.js
cvalarida: Ohhh, I see. Because it would "normally" be something like:
```js
(formData) => _.get('view:applicantServed', formData)
```
but because of the auto-currying, `_.get()` returns a function which `formData` is then passed into. It's tricky and I like it, but it's...not terribly clear, methinks.

On the other hand, what else would it be doing? I can see us going either way.

src/js/edu-benefits/pages/schoolSelection.js
```diff
@@ -45,9 +46,9 @@ export default function createSchoolSelectionPage(schema, options) {
       'ui:title': 'Would you like to receive vocational and educational counseling?',
       'ui:widget': 'yesNo'
     },
-    // May want to turn this into a yes/no
     civilianBenefitsAssistance: {
-      'ui:title': 'I am receiving benefits from the U.S. Government as a civilian employee during the same time as I am seeking benefits from VA.'
+      'ui:title': civilianBenefitsLabel,
+      'ui:widget': 'yesNo'
```
cvalarida: I agree this should be a `yesNo`, but should it be signed off by @goldenmeanie?

cvalarida (approved):
After the relabeling 'Veteran' to 'Sponsor', I think it's golden.

src/js/edu-benefits/pages/schoolSelection.js
goldenmeanie: Should be a yesNo. Copy update will happen as a separate issue.

src/js/edu-benefits/5490/config/form.js
goldenmeanie: Should be labeled Sponsor first name, etc. but there is a separate issue for that

src/js/common/schemaform/ArrayField.jsx
jbalboni: Yeah, I skipped it because our array pattern wouldn't make much sense with more than one empty row, but it wouldn't hurt to handle that.

PR #5118, opened 2017-03-17
5490 Content Changes

cvalarida:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1683
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1787 (for tracking purposes; this one is in the above)

> image

> image

> image

> image

> image

> image

> image

> image

> image

> image

> image

> image

> image

> image

> image


src/js/edu-benefits/pages/applicantInformation.js
```diff
@@ -51,7 +51,14 @@ export default function applicantInformation(schema, options) {
       'ui:order': fields,
       relativeFullName: fullName.uiSchema,
       relativeSocialSecurityNumber: ssn.uiSchema,
-      relativeDateOfBirth: currentOrPastDate.uiSchema('Date of birth'),
+      relativeDateOfBirth: _.assign(currentOrPastDate.uiSchema('Date of birth'),
+        {
+          'ui:errorMessages': {
+            pattern: 'Please provide a valid date',
+            futureDate: 'Please provide a valid date'
```
cvalarida: I'm not too fond of this, but should I override the `pattern` and use that in the validation? `futureDate` is specifically for if the date is not in the future, whereas `pattern` is for if it's not a valid date.

src/js/edu-benefits/5490/config/form.js
```diff
@@ -498,6 +500,15 @@ const formConfig = {
             'educationalCounseling'
           ],
           required: ['educationType']
+        }), {
+          // Rephrase the question for facility name in educationProgram
```
jbalboni: This is probably fine as a global change.

src/js/edu-benefits/5490/config/form.js
```diff
@@ -498,6 +500,15 @@ const formConfig = {
...
+          uiSchema: {
+            educationProgram: {
+              name: {
+                'ui:title': 'Name of school, university, or training facility you want to attend'
```
cvalarida: This bit is kind of gross. I realize I could use `_.set()` instead to compress this, but the bit that I don't like is having to modify the `uiSchema` outside of the page creation function. Other than giving it a very specific parameter (which I was trying to avoid to keep it as consistent as I could with the other page creation functions), this was the best solution I've dreamed up so far.

I'm open to suggestions.

src/js/edu-benefits/pages/applicantInformation.js
```diff
@@ -51,7 +51,14 @@ export default function applicantInformation(schema, options) {
       'ui:order': fields,
       relativeFullName: fullName.uiSchema,
       relativeSocialSecurityNumber: ssn.uiSchema,
-      relativeDateOfBirth: currentOrPastDate.uiSchema('Date of birth'),
+      relativeDateOfBirth: _.assign(currentOrPastDate.uiSchema('Date of birth'),
+        {
+          'ui:errorMessages': {
+            pattern: 'Please provide a valid date',
```
jbalboni: Why does this particular date need these validation messages, vs what's there by default? Should we make the changes for all current or past dates?

jbalboni (approved):
Have a couple questions, but looks good otherwise.

src/js/edu-benefits/pages/applicantInformation.js
cvalarida: cc @mollyblake

src/js/edu-benefits/5490/config/form.js
cvalarida: I thought `educationProgram` might also be used for previous education, so relabeling it with the "you want to attend" didn't seem to make sense.

jbalboni: Ah, gotcha. I think when we switch the 1995 to use educationProgram, we should make this one the default, since I think it's more common that previous education.

src/js/edu-benefits/pages/applicantInformation.js
jbalboni: Having two separate ones makes sense to me, but I don't know why they're the same message or why this date field is special. If there's better generic copy, we should make that the default.

If we actually just want them to be the same message, let's just do that in the common date definition's uiSchema.

I think this depends on what @mollyblake is looking for here.

cvalarida:
Opened a new issue for the date question so I can merge this: https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1807

PR #5120, opened 2017-03-20
Update 5490 to use re-aligned organization

jbalboni:
Related to [1696](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1696)

Updates the 5490 to follow the image in the linked issue. Also updated some urls in both 1990e and 5490 to be a little simpler,

screen shot 2017-03-20 at 11 46 51 am
screen shot 2017-03-20 at 11 46 41 am
screen shot 2017-03-20 at 11 46 16 am


raquelromano-gov (commented):
I like these simplified paths--is it too late to do the same for the 1995? I think we were following suit, but there are lots of hyphenated paths there as well.

jbalboni:
@raquelromano-gov I think we could do it there, too. Just have to change e2e tests as well for that one.

jbalboni:
@raquelromano-gov I updated some of them, but I'm not sure how much sense making all the same changes would make without the re-alignment here. I'm going to create a tech debt issue to clean up the urls.

src/js/edu-benefits/5490/config/form.js
```diff
@@ -300,8 +341,8 @@ const formConfig = {
           }
         },
         sponsorService: {
-          title: 'Sponsor Service',
-          path: 'military-service/sponsor-service',
+          title: 'Sponsor service',
```
cvalarida: The latest victim in _The Case of the Missing Title_: `Sponsor Service`.

image

src/js/edu-benefits/5490/config/form.js
```diff
@@ -124,9 +165,9 @@ const formConfig = {
             }
           }
         },
-        previousBenefits: {
-          title: 'Previous Benefits',
-          path: 'benefits-eligibility/previous-benefits',
+        benefitHistory: {
+          title: 'Benefit history',
```
cvalarida: Same trouble with the missing title here.

image

cvalarida (approved):
Only weird things I could find were the missing page titles (which I think is ubiquitous and as such not a part of this PR, strictly speaking).

src/js/edu-benefits/5490/config/form.js
```diff
@@ -91,15 +89,58 @@ const formConfig = {
       pages: {
         applicantInformation: applicantInformation(fullSchema5490, {
           labels: { relationship: relationshipLabels }
-        })
+        }),
+        additionalBenefits: additionalBenefits(fullSchema5490, {
+          fields: ['civilianBenefitsAssistance', 'civilianBenefitsSource']
+        }),
+        applicantService: {
+          title: 'Applicant service',
```
jbalboni: The only clear answer I've gotten is that chapter titles should be title case and everything else should be sentence case.

For this particular line, that's a page title and you'll only see it on the review page.

src/js/edu-benefits/5490/config/form.js
jbalboni: Same as above, it should just show up on the review page.

src/js/edu-benefits/5490/config/form.js
jbalboni: I didn't see anything suggesting we need to show titles where we weren't before, but @goldenmeanie could say for sure.

cvalarida: I think the idea was to add the subheadings. From the original issue:

>Red text is a change from current copy.
>Parentheticals are subheads within a chapter

But yeah, what say ye, @goldenmeanie?

goldenmeanie: Yeah, the parentheticals should be added as sub-head if they don't yet exist.

cvalarida:
Just merged `1990e-realign` into `master`, so I changed the base branch.

src/js/edu-benefits/5490/config/form.js
jbalboni: @goldenmeanie @cvalarida Page titles are updated now.

PR #5122, opened 2017-03-20
Return empty object when GIBCT constants are not yet populated.

U-DON:
Fixes error in calculating estimated benefits when constants haven't been fetched and are undefined, which results in displaying loading indicator forever.


bshyong (approved):
LGTM

PR #5131, opened 2017-03-21
Top entry in GIBCT autocomplete suggestions doesn't change to current selection.

U-DON:
The first entry should be the current search term the user has typed, but pressing 'Up' and 'Down' in the list was making it change to the selected entry instead. This is because it was changing the `searchTerm` in state, which was getting prepended to the list on render. So instead, we're now prepending the `searchTerm` upon receiving a new suggestion list and storing the result of that in state.

Also added a check to avoid duplicate entries when a suggestion is selected and becomes the `searchTerm` as the content of the input changes to it. The list won't include the institution name twice (as both the `searchTerm` and the result from the autocomplete API response).


bshyong (approved):
LGTM

PR #5134, opened 2017-03-21
Add copy changes for blue button

bshyong:
Adds error messaging copy for Blue button

resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/936


U-DON (approved):
LGTM.

Issue #4964, opened 2017-02-27
Update health records content - Alerts


altangel:
Update alerts on the download page with updated content:

**Successful update:**
![hr v4 download side by side](https://cloud.githubusercontent.com/assets/25392602/23374538/9555c586-fcd9-11e6-8ab1-9fd7312e892f.png)

**Errors with update:**
![hr download no table error 2](https://cloud.githubusercontent.com/assets/25392602/23374525/85123466-fcd9-11e6-98b5-c59d825e05e7.png)

**Not up-to-date:**
![hr download no table not up to date](https://cloud.githubusercontent.com/assets/25392602/23374526/85232c3a-fcd9-11e6-9780-4b8b2898c9ae.png)

/cc @AnitraAppa @mikeusdsva for awareness and in case I missed something.

mikeusdsva:
Instead of "PDF File" can we just say "PDF"

PR #5135, opened 2017-03-21
Update health records error states

bshyong:
Updates content for health records error states

Notes:
The assumption is that refresh statuses behave as follows:
- The initial response will yield statuses with a past date, with a status OK or Error (indicating the last time that a refresh occurred)
- Statuses will show the current date if the request succeeded; the status will be OK or Error depending on the outcome of the request
---
- All statuses are set to current date = update completed successfully
- Some statuses do not have the current date = update still in progress
- All statuses have current date, but some have status Error = some update processes failed downstream

Resolves https://github.com/department-of-veterans-affairs/vets-website/issues/4964


U-DON (commented):
Changes LGTM, but I think you need to update one of the unit tests that's now failing.

PR #5136, opened 2017-03-21
Add Raven to send errors to Sentry

jbalboni:
Related to [1674](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1674)

Adds Raven to send errors to our Sentry url.

I don't believe the devops changes for this are live, so this should not be merged until then.


src/js/common/sentry.js
```diff
@@ -0,0 +1,22 @@
+import Raven from 'raven-js';
+
+import environment from './helpers/environment';
+
+// url check is necessary for e2e tests
+const trackErrors = (__BUILDTYPE__ === 'production' || __BUILDTYPE__ === 'staging')
```
jbalboni: I think this is the logic we'd want, but need some guidance. My goal was to have this enabled only in staging and production, and not on dev, locally, or in tests.

jkassemi: This looks right in combination with the BASE_URL check. Any idea on how best to pull this out into a configuration file or set it up as a feature flag or option?

jkassemi (approved):

src/js/common/sentry.js
jkassemi: I think I'd still be in favor of keeping this active on dev.vets.gov, too. Can we just hold off on the __BUILDTYPE__ check and focus only on the BASE_URL?

jkassemi:
Changes from devops#1589 have been deployed to all environments, so this is just held up by the configuration discussion. /cc @ayaleloehr

ayaleloehr:
Core team said they were going to look at this today, if anyone on the West coast is still around?

cc @robbiethegeek for Monday otherwise.

jbalboni:
@jkassemi I updated the check in sentry.js to just use the base url. Let me know if there are other changes needed for this.

jcmeloni-usds:
@robbiethegeek-usds what does core need to do here?

jkassemi:
Tested in dev. We'll get this out and monitor sentry for exception reports.

PR #5138, opened 2017-03-21
SSN/FN fixes to 1990e and 5490

raquelromano-gov:
Closes https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1744

image

image

### DIFF
--- a/src/js/edu-benefits/1990e/config/form.js
+++ b/src/js/edu-benefits/1990e/config/form.js
@@ -121,7 +121,7 @@ const formConfig = {
                 'ui:title': 'Sponsor Social Security number'
               },
               'view:noSSN': {
-                'ui:title': 'Sponsor doesn\'t have a Social Security number',
+                'ui:title': 'I don\'t know my sponsor’s Social Security number',
               },
               vaFileNumber: {
                 'ui:title': 'Sponsor file number',
--- a/src/js/edu-benefits/1990e/config/form.js
+++ b/src/js/edu-benefits/1990e/config/form.js
@@ -117,11 +117,14 @@ const formConfig = {
               }
             }),
             'view:veteranId': _.merge(veteranId.uiSchema, {
-              vaFileNumber: {
-                'ui:title': 'Sponsor file number',
-              },
               veteranSocialSecurityNumber: {
                 'ui:title': 'Sponsor Social Security number'
+              },
+              'view:noSSN': {
+                'ui:title': 'Sponsor doesn\'t have a Social Security number',
+              },
+              vaFileNumber: {
+                'ui:title': 'Sponsor file number',
               }
             }),
             veteranAddress: address.uiSchema('Sponsor address'),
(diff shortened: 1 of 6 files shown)


raquelromano-gov:
Note that there is a test in 
`test/edu-benefits/definitions/veteranId.unit.spec.jsx` that sets the no-SSN box and verifies the FN input is shown and the SSN is no longer required.

That test is sort of duplicated in `test/edu-benefits/1990e/config/sponsorVeteran.unit.spec.jsx` and in `test/edu-benefits/1995/config/veteranInformation.unit.spec.jsx`

We can also add it to `test/edu-benefits/5490/config/sponsorInformation.unit.spec.jsx` or we can remove it from the above 2 tests and rely on the definitions test. Thoughts?

src/js/edu-benefits/1990e/config/form.js
```diff
@@ -117,11 +117,14 @@ const formConfig = {
...
             }),
             'view:veteranId': _.merge(veteranId.uiSchema, {
-              vaFileNumber: {
-                'ui:title': 'Sponsor file number',
-              },
               veteranSocialSecurityNumber: {
                 'ui:title': 'Sponsor Social Security number'
+              },
+              'view:noSSN': {
+                'ui:title': 'Sponsor doesn\'t have a Social Security number',
```
cvalarida: Should this be "I don’t know my sponsor’s Social Security number" (from your comment [here](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1744#issuecomment-288191320))?

src/js/edu-benefits/5490/config/form.js
```diff
@@ -321,22 +322,29 @@ const formConfig = {
...
               }
             }),
-            veteranSocialSecurityNumber: _.assign(ssn.uiSchema, {
-              'ui:title': 'Sponsor Social Security number'
+            'view:veteranId': _.merge(veteranId.uiSchema, {
+              veteranSocialSecurityNumber: {
+                'ui:title': 'Sponsor Social Security number'
+              },
+              'view:noSSN': {
+                'ui:title': 'Sponsor doesn\'t have a Social Security number',
```
cvalarida: Same here.

cvalarida (approved):

src/js/edu-benefits/1990e/config/form.js
raquelromano-gov: Yes! Done.

raquelromano-gov:
Thoughts on unit test coverage/duplication? (see comment above)

cvalarida:
Ah, sorry. Didn't see that when I did my review; must have skipped right over it.

Yeah...I think we can take it out of those two. I think that's what we've been doing. Of course, I can't remember where right now, but I vaguely remember @jbalboni telling me that when I duplicated a test much like this.

As long as it uses the same chunk of code in both places, I don't think we need to add the time to running unit tests.

raquelromano-gov:
OK, SG! Merging this, and will separately pull out the other tests.

PR #5144, opened 2017-03-21
GIBCT ROP calculation refactor

U-DON:
### Changes
* Fixes a bug that would result in `NaN` in multiple places if you selected an old GI Bill option (anything other than Ch. 33) in the eligibility form.
* Consistently formatted enrolled options and convert them to calculator values similarly.
* Determine the rendering of one or the other type of enrolled input in a single function.


saintsoup52:
I always get anxious when we open the calculator engine... but this looks good to me. When we get done with all this is would be helpful for someone to walk me through how the new naming conventions work. For example, I see inputs now have a "+" before the name and other things like that.

U-DON:
@saintsoup52 The "+" is a shorthand for wrapping a value in `Number()` for conversion to a number. Some of them inputs are stored at strings instead of actual numbers right now, and I would like to change that where possible.

Other than that, the naming convention was just changing from snake-case to camel-case. I would like to tidy up the `getDerivedValues` function a little more since it just has a bunch of variable declarations. They are scoped to the function at least, but there are so many of them that it seems messy.

src/js/gi/components/profile/CalculatorForm.jsx
```diff
@@ -16,7 +16,6 @@ class CalculatorForm extends React.Component {
     this.renderScholarships = this.renderScholarships.bind(this);
     this.renderTuitionAssist = this.renderTuitionAssist.bind(this);
     this.renderEnrolled = this.renderEnrolled.bind(this);
```
bshyong: Is it necessary to bind these render functions?

bshyong (approved):
LGTM

bshyong:
@saintsoup52 https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Arithmetic_Operators#Unary_plus_()

src/js/gi/components/profile/CalculatorForm.jsx
U-DON: Probably not, but it's just something I do anyway just in case. The one time I forget, I'll probably end up with some confusing bug related to `this`. If anything, it's like a quick table of contents for the functions in the class.

bshyong: 👍   

The one thing I would bind is `this.props.onShowModal`; if you set up that function in the constructor, you can avoid the performance impact of binding within render functions.

U-DON: `onShowModal` is being bound with different arguments for each of those render functions, since it's showing a different modal depending on the section. Each of those would have be set up in the constructor. Will have to revisit along with many other things that need to be refactored.

PR #5145, opened 2017-03-22
Resolves issue with overly broad .primary p:first-of-type selector

ghost:
**IMPORTANT** Review of this PR should focus on src/sass/modules/_m-alert.scss, and src/sass/base/_va.scss.
 
- Added .va-introtext class. closes department-of-veterans-affairs/vets.gov-team#619
- Make .call-out p, .feature p inherit font, color property values.
- Replaced secondary class with va-alert for alert boxes.
  - Closes department-of-veterans-affairs/vets.gov-team#1470

Affects:

- content/pages/education/apply-for-education-benefits.md
- content/pages/education/apply.md

### DIFF
--- a/content/pages/disability-benefits/after-you-apply.md
+++ b/content/pages/disability-benefits/after-you-apply.md
@@ -13,7 +13,11 @@ relatedlinks:
    description: "Find out what happens after you get your decision notice with your rating."
 ---
 
+<div class="va-introtext">
+
 Find out what happens to your claim after you apply for disability benefits. 
+
+</div>
 
 ### How long does it take VA to make a decision?
 
--- a/content/pages/disability-benefits/apply-for-benefits/help.md
+++ b/content/pages/disability-benefits/apply-for-benefits/help.md
@@ -17,9 +17,13 @@ relatedlinks:
     title: How does VA decide when my disability compensation starts?
 ---
 
+<div class="va-introtext">
+
 If you need help filing a disability claim, you may want to work with an accredited representative. We trust these professionals because they're trained in the claims process and can help you with VA-related needs.
 
 Most accredited representatives work for Veterans Service Organizations (VSOs). These private nonprofit groups advocate on behalf of Veterans and Servicemembers as well as their dependents and survivors. 
+
+</div>
 
 <div class="call-out" markdown="0">
 
--- a/content/pages/disability-benefits/apply-for-benefits/one-year.md
+++ b/content/pages/disability-benefits/apply-for-benefits/one-year.md
@@ -22,9 +22,14 @@ relatedlinks:
     description: ""
 ---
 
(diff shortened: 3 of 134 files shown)


tiffanybbrown-va:
Examples of affected content:

screen shot 2017-03-21 at 5 06 39 pm

screen shot 2017-03-21 at 5 20 05 pm

jcmeloni-usds:
LGTM - @melwoodard you ok w/ it?

melwoodard:
@tiffanybbrown-va the intention of https://github.com/department-of-veterans-affairs/vets.gov-team/issues/619 was to remove the `first-of-type` rule entirely and allow the `.va-introtext` class you've created to be manually applied as the content team sees fit, since there are a lot of cases where a page doesn't start out with an intro paragraph. 

It does looks like there's no more weirdness with this style showing up in the first paragraphs of callouts or alerts, but I see other instances where the page doesn't begin with a paragraph, and so `first-of-type` is being applied to the first paragraph that shows up:

https://vetsgov-pr-5145.herokuapp.com/disability-benefits/conditions/ptsd/

posttraumatic_stress_disorder__ptsd___vets_gov

@bethpotts @rroueche can you confirm? You'd rather have to apply this intro paragraph style manually than have it applied automatically, correct? In which case the content team will probably need to help run through all pages and identify where these styles are needed.

tiffanybbrown-va:
@melwoodard I understand the intention of department-of-veterans-affairs/vets.gov-team#619.  This PR is an interim step to resolve this issue for call-out / feature boxes, and give the content team time to update the markup.

Happy to remove the rule entirely, however, since `p:nth-child(1)` should work for first paragraphs without introducing the side effects that `:first-of-type` does.

tiffanybbrown-va:
@melwoodard Please review when you have a chance.

melwoodard (approved):
LGTM! I've left @bethpotts a list of URLs in the original issue to review that may need some style tweaks, but that doesn't need to block this PR.

rroueche:
Also, confirming -- yes that's correct that we'd like to apply the class manually to text we feel needs the special styling, instead of having the first-of-type rule apply the styles automatically.

rroueche:
although, @melwoodard and @tiffanybbrown-va here's an example where all of the text on the page has the special styling, whereas I'd think w/the rule removed, it would all have the normal p styles:

image

compare:
vets.gov/disability-benefits/claims-process/
https://www.vets.gov/disability-benefits/claims-process/

rroueche:
ah nvm, I see why:
image


so @melwoodard is that an example of a page that you've manually updated w/the class to have that style?

melwoodard:
@rroueche I applied the class to all content pages as part of this PR. I made a list for Beth [here](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/619#issuecomment-288892042) of a few pages that weren't super straightforward on where the intro/body break should happen that we can address separately from this PR.

rroueche:
ok awesome, yeah looks good to me too then!

PR #5147, opened 2017-03-22
Updated callback param names in updateSchema

cvalarida:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1519

Matches the naming convention

### DIFF
--- a/docs/schemaform/form-config.md
+++ b/docs/schemaform/form-config.md
@@ -236,7 +236,11 @@ In addition to the uiSchema options listed in the library [https://github.com/mo
     }
 
     // Function that conditionally replaces the current field's schema
-    updateSchema: function (pageData, formData, pageSchema) {
+    updateSchema: function (pageData, form, pageSchema) {
+      // form is organized like: form.pageName.data.fieldName
+      // This means it's not like formData, where all the data for all the forms is mashed
+      // together into one object. It's a bit of a pain, but this means updateSchema can also
+      // update the schema / uiSchema of other pages (but do so sparingly!).
       return {};
     }
   }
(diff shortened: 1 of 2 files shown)


raquelromano-gov (approved):

cvalarida:
@raquelromano-gov I just noticed the naming convention was weird in `updateSchemaFromUiSchema`, so I changed it in there as well. Could you give it another quick look?

raquelromano-gov:
Sorry, just saw that you had updated this! Reviewing now.

raquelromano-gov:
Thanks for fixing this--these names were making my head spin when I was writing that function to update the ui based on data (which I later threw away) to keep straight the various data objects we're passing around between `reducers/index.js`, `helpers.js` and the form config files:

It sounds like we have 3 things:
1.  `pageData`: structured object with all data from a specific page, not the whole form
2.  `formData`: flattened (?) object with all data from all pages
3.  `form`: structured object with all data from all pages

Can we rename (2) and (3)?  E.g., `formData` -> `flattenedData` and `form` -> `formData` (open to suggestions).

cvalarida:
`flattenedData` sounds good, but `formData` is misleading, methinks. If I saw that, I would think it would be structured either like `flattenedData` (the current `formData`), or just the data separated by page name like
```js
{
  pageName: {
    field1: 'value',
    field2: 'value'
  }
}
```
when it's really set up with `schema` and `uiSchema` as well as `data`.
```js
{
  pageName: {
    schema: { ... },
    uiSchema: { ... },
    data: {
      field1: 'value',
      field2: 'value'
    }
  }
}
```

raquelromano-gov:
Oh right...we use `formConfig` in a bunch of places. Can we use that here instead of plain old `form`?

cvalarida:
Ah, good ole `formConfig`. Sounds great! I'll get right on it. :+1:

cvalarida:
Actually...I think that might be misleading as well because the config doesn't have `data`. Hmm.

raquelromano-gov:
Tragic.
OK, well this is certainly an improvement so go for it.

raquelromano-gov (approved):

cvalarida:
If we come up with something better, we can definitely revisit this.

PR #5148, opened 2017-03-22
Run docker processes as Jenkins user

jkassemi:


Dockerfile
```diff
@@ -2,8 +2,8 @@
 
 FROM buildpack-deps:jessie
 
-RUN groupadd --gid 1000 node \
-  && useradd --uid 1000 --gid node --shell /bin/bash --create-home node
+RUN groupadd --gid 504 jenkins \
```
knkski: Does 504 have to match up with the host's `jenkins` user's id? Would be nice to get a comment as to why 504 specifically

knkski (approved):
👍 other than minor comment

Dockerfile
jkassemi: @knkski - yes, added a comment explaining this.

PR #5150, opened 2017-03-22
1990e e2e tests

raquelromano-gov:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1614

This is just the first few pages--will add others in a separate PR.


test/util/edu-helpers.js
```diff
@@ -16,6 +18,35 @@ function completeVeteranInformation(client, data, onlyRequiredFields) {
   }
 }
 
+function completeRelativeInformation(client, data, onlyRequiredFields) {
+  const dobFields = data.relativeDateOfBirth.split('-');
+  client
+    .clearValue('input[name="root_relativeFullName_first"]')
+    .setValue('input[name="root_relativeFullName_first"]', data.relativeFullName.first)
```
annekainicUSDS: Perhaps this is unnecessary, but would it be valuable to create a function that both clears and sets the value, taking as arguments the field and the new value? Only benefits I suppose would be that it reduce all these lines by half and would ensure you're always clearing the field before trying to set the value?

test/util/edu-helpers.js
```diff
@@ -16,6 +18,35 @@ function completeVeteranInformation(client, data, onlyRequiredFields) {
...
+      .setValue('input[name="root_relativeFullName_middle"]', data.relativeFullName.middle)
+      .click(data.gender === 'M' ? 'input[name=root_gender_0' : 'input[name=root_gender_1');
+    selectDropdown(client, 'root_relativeFullName_suffix', data.relativeFullName.suffix);
+  }
+}
+
+function completeAdditionalBenefits(client, data) {
+  client
+    .click(data.nonVaAssistance ? 'input[name="root_nonVaAssistanceYes"]' : 'input[name="root_nonVaAssistanceNo"]')
+    .click(data.civilianBenefitsAssistance ? 'input[name="root_civilianBenefitsAssistanceNo"]' : 'input[name="root_civilianBenefitsAssistanceNo"]');
```
annekainicUSDS: Do you mean for the first field to be `root_civilianBenefitsAssistanceYes` instead?

test/util/edu-1990e-helpers.js
```diff
@@ -0,0 +1,22 @@
+const mock = require('./mock-helpers');
+
+// Create API routes
+function initApplicationSubmitMock() {
```
annekainicUSDS: Can this also be shared with the other education forms? Looks like they're exactly the same except for the end of the endpoint, which corresponds with the form number.

annekainicUSDS (commented):
Looks good so far! Made a few suggestions that could be done in a future refactor if you think they would be good to do :). Only one change probably should be done in this PR.

test/util/edu-1990e-helpers.js
raquelromano-gov: Yes! This is actually the first refactoring I tried when I first poked around here, but I ran into a host of other issues with my local e2e test. Just put it back in, and using it for the 1995 and 1990e, but not the 1990, since the 1990 has its on special version of the shared helpers we have in `test/util/edu-helpers.js` which we can keep until we write the 1990 to use rjsf

test/util/edu-helpers.js
raquelromano-gov: I like this idea, though we'll have to stop chaining them as client methods, e.g. will look something like
```
clearAndSetValue(client, 'input[name="root_relativeFullName_first"]', data.relativeFullName.first);
clearAndSetValue(client, 'input[name="root_relativeFullName_last"]', data.relativeFullName.last);
```
Definitely prefer to do this in a separate PR though, since it's been a struggle to get these working. (So much flakiness locally, and then different results when running locally vs. on jenkins 😭 )

test/util/edu-helpers.js
raquelromano-gov: Oops, thanks! Will push that change once I see how jenkins behaves on this last commit.

raquelromano-gov:
I have no idea why this passes locally but fails on jenkins. Looking into it.

test/util/edu-helpers.js
```diff
@@ -16,11 +36,43 @@ function completeVeteranInformation(client, data, onlyRequiredFields) {
...
+    .setValue('input[name="root_relativeSocialSecurityNumber"]', data.relativeSocialSecurityNumber)
+    .clearValue('input[name="root_relativeDateOfBirthYear"]')
+    .setValue('input[name="root_relativeDateOfBirthYear"]', parseInt(dobFields[0], 10).toString())
+    .click('input[name="root_relationship_1"]');
+  selectDropdown(client, 'root_relativeDateOfBirthMonth', parseInt(dobFields[1], 10).toString());
+  selectDropdown(client, 'root_relativeDateOfBirthDay', parseInt(dobFields[2], 10).toString());
+
   if (!onlyRequiredFields) {
     client
-      .click('.form-radio-buttons:first-child input');
```
raquelromano-gov: This is a weird diff...this is from `completeBenefitsSelection()` below

raquelromano-gov:
OK, wondering if you have any ideas: on [jenkins](http://jenkins.vetsgov-internal/blue/organizations/jenkins/testing%2Fvets-website/detail/rr-1990e-e2e/9/pipeline), both of these new 1990e e2e tests fail with
```
Testing if the page title equals "Apply for education benefits: Vets.gov". 
expected "Apply for education benefits: Vets.gov" but got: "" at /application/test/edu-benefits/1990e/00-required.e2e.spec.js:13:15
at Object.BeginApplication [as Begin application] (/application/test/util/e2e-helpers.js:100:7)
```

But both tests pass locally, and the title looks fine.

I think it has something to do with the initialization (and nothing to do with the title), but I don't see why these new tests would initialize fine locally but not on jenkins.

test/util/edu-helpers.js
cvalarida: Unless we can bootstrap a little helper function onto `client` which just calls those two functions in order.

```js
function clearAndSetValue(selector, value) {
  this.clearValue(selector);
  return this.setValue(selector, value);
}
```
Or however it works out.

Just a thought.

test/edu-benefits/1995/00-required.e2e.spec.js
```diff
@@ -26,12 +26,12 @@ module.exports = E2eHelpers.createE2eTest(
...
 
-    // Benefits eligibility
+    // Benefits eligibility page.
     client
       .waitForElementVisible('label[for="root_benefit"]', Timeouts.slow);
-    EduHelpers.completeBenefitsSelection(client, testData.benefitSelection.data);
+    EduHelpers.completeBenefitsSelection(client);
     client.click('.form-progress-buttons .usa-button-primary');
-    E2eHelpers.expectNavigateAwayFrom(client, '/benefits/elibility');
+    E2eHelpers.expectNavigateAwayFrom(client, '/benefits/eligibility');
```
cvalarida: How was this working before??

cvalarida (approved):

cvalarida:
@raquelromano-gov I think these e2e tests fail because the `_BUILDTYPE_` is set to production on Jenkins when running e2e tests, and when the `_BUILDTYPE_ === 'production'`, the 1990e route isn't added, so electron can't get to the form.

raquelromano-gov:
For the record, the plan is to change how we're guarding unlaunched forms (see https://github.com/department-of-veterans-affairs/vets-website/pull/4497 and "Continuous Integration" section of https://github.com/department-of-veterans-affairs/vets-website/blob/master/README.md), and then re-enable these.

test/util/edu-helpers.js
raquelromano-gov: Oops, just saw this suggestion. Like this. Will try it in the next e2e PR

test/edu-benefits/1995/00-required.e2e.spec.js
raquelromano-gov: Ha...well`expectNavigationAwayFrom` just checks that the current URL does not equal the one passed in!

PR #5151, opened 2017-03-22
Add additional date validations and error messaging to health records

bshyong:
Adds additional date validation to health records form. Also sets default end date to current day.

Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1366

screen shot 2017-03-22 at 3 22 02 pm

### DIFF
--- a/src/js/health-records/containers/Main.jsx
+++ b/src/js/health-records/containers/Main.jsx
@@ -139,13 +139,16 @@ export class Main extends React.Component {
     const handleFormattedDate = (start = true) => {
       let handleDateChange;
       let setInvalidDateState;
+      let setMissingDateState;
 
       if (start) {
         handleDateChange = this.handleStartDateChange;
         setInvalidDateState = () => this.setState({ invalidStartDateFormat: true });
+        setMissingDateState = () => this.setState({ missingStartDate: true });
       } else {
         handleDateChange = this.handleEndDateChange;
         setInvalidDateState = () => this.setState({ invalidEndDateFormat: true });
+        setMissingDateState = () => this.setState({ missingEndDate: true });
       }
 
       return (e) => {
@@ -163,6 +166,8 @@ export class Main extends React.Component {
           handleDateChange(null);
           if (dateString) {
             setInvalidDateState();
+          } else {
+            setMissingDateState();
           }
         }
       };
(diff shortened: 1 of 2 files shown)


U-DON:
It's not that both are required, but that at least one is required. So even if the end date is today by default, if the user clears it and leaves both start and end empty upon submit, it should show the error message. And the error message should be "Enter a date range."

bshyong:
Wouldn't requiring a date range imply that both start and end dates are required?

U-DON:
My understanding is that if you left start date empty but filled in end date, it should cover everything before the end date. And if the start date is filled in, but the end date is empty, it should cover everything after the start date. Unless that's not how it is supposed to work at all.

bshyong:
I see. I think the API requires both a `fromDate` and `toDate`—we could plug it in behind the scenes if a user leaves one of the dates blank, but I think it might be better UX to have the user be intentional about the range that they want.

src/js/health-records/containers/Main.jsx
```diff
@@ -232,6 +232,7 @@ export class Main extends React.Component {
                     selected={endDate}
                     disabled={datePickerDisabled}
                     minDate={startDate}
+                    maxDate={moment()}
```
U-DON: 👍

src/js/health-records/containers/Main.jsx
```diff
@@ -173,12 +178,20 @@ export class Main extends React.Component {
...
+      this.state.missingStartDate ||
+      this.state.missingEndDate;
 
     const customDateOptionClass = classNames({
       'custom-date-option': true,
       'date-range-error': customDateRangeError
     });
 
+    const validationErrorMessages = {
+      rangeError: 'Enter dates in the MM/DD/YYYY date format',
```
U-DON: `customDateRangeError` and `validationErrorMessages.rangeError` seem less descriptive now that there are two types of errors that could occur with date range. Maybe something like `invalidFormatError` would be better now.

src/js/health-records/containers/Main.jsx
```diff
@@ -173,12 +178,20 @@ export class Main extends React.Component {
     const customDateRangeError =
       this.state.invalidStartDateFormat ||
       this.state.invalidEndDateFormat;
+    const missingDateError =
+      this.state.missingStartDate ||
+      this.state.missingEndDate;
 
     const customDateOptionClass = classNames({
       'custom-date-option': true,
       'date-range-error': customDateRangeError
```
U-DON: This should probably include both error types. `'date-range-error': invalidFormatError || missingDateError`.

U-DON (commented):
If the API doesn't allow having one of the two dates in the range be empty, this should be okay. Had some minor comments.

bshyong:
Sounds good. I did some light refactoring to make the logic a little cleaner

U-DON (approved):
LGTM.

PR #5152, opened 2017-03-22
Only check query object before updating URL and refreshing GIBCT search results.

U-DON:
Fixes issue where the search results would refresh even if the user clicked an already selected radio button. Before, the update logic was incorrectly looking at the entire `location` object, which includes keys like unique id and operation types that threw the whole conditional off. So now we're only looking for changes to the query params.


bshyong (approved):
LGTM

PR #5156, opened 2017-03-23
Updated handling of country facets in GIBCT after fix for filtering with hyphens.

U-DON:
Related to department-of-veterans-affairs/gibct-data-service#224.

This fix is to be combined with this change in GIDS: department-of-veterans-affairs/gibct-data-service#234.

`meta.facets.country` is becoming a list of `{ name, count }` objects instead of just being one object with country names as keys and counts as values. This avoids any messiness with camel-casing everything. Will apply camel-casing in all API responses separately from this PR at some point.


bshyong (approved):
LGTM

PR #5157, opened 2017-03-23
5490 Education type dropdown (changed reducer)

cvalarida:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1794

Benefit: Chapter 35
Relationship: Spouse
Effect: None
image

Benefit: Chapter 35
Relationship: Child
Effect: Removes `Correspondence`
> image

Benefit: Chapter 33
Relationship: Child
Effect: Adds `Vocational flight training`
> image

### DIFF
--- a/src/js/common/schemaform/reducers/index.js
+++ b/src/js/common/schemaform/reducers/index.js
@@ -91,7 +91,22 @@ export default function createSchemaFormReducer(formConfig) {
   return (state = initialState, action) => {
     switch (action.type) {
       case SET_DATA: {
-        const newState = _.set([action.page, 'data'], action.data, state);
+        // TODO: Write the documentation that says this is a thing (persistent schema
+        //  to be modified every pass)
+
+        // Use new data, but the initial schema
+        const newState = {};
+        Object.keys(state).forEach((pageName) => {
+          // Add data from state and schema / uiSchema from firstPassInitialState
+          //  for each page.
+          newState[pageName] = {
+            data: state[pageName].data,
+            ..._.omit('data', firstPassInitialState[pageName])
+          };
+        });
+
+        // Set the data just entered
+        newState[action.page].data = action.data;
 
         return recalculateSchemaAndData(newState);
       }
--- a/src/js/edu-benefits/5490/config/form.js
+++ b/src/js/edu-benefits/5490/config/form.js
@@ -526,38 +526,26 @@ const formConfig = {
               educationType: {
                 'ui:options': {
                   updateSchema: (pageData, form, schema) => {
-                    // updateSchema mutates the schema, so after an option is
-                    //  removed from the enum, either I can add it back in manually
-                    //  or take the base as the educationType definition in
-                    //  vets-json-schema instead of the schema passed in.
+                    // TODO: Reorder fields...unless it's a universal change...
+                    // https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1794
+
                     const newSchema = _.cloneDeep(schema);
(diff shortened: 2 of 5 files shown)


src/js/edu-benefits/pages/schoolSelection.js
```diff
@@ -55,6 +55,7 @@ export default function createSchoolSelectionPage(schema, options) {
 
   const schemaProperties = pickFields(schema.properties);
 
+  // educationProgram is a function, so pull out the schema
```
raquelromano-gov: I think you meant educationProgram.schema() is a function

src/js/edu-benefits/5490/config/form.js
```diff
@@ -522,6 +522,33 @@ const formConfig = {
...
+
+                    // Remove tuition top-up
+                    const filterOut = ['tuitionTopUp'];
+                    // Correspondence not available to chapter35 children
+                    if (benefitData === 'chapter35' && relationshipData === 'child') {
+                      filterOut.push('correspondence');
+                    }
+                    // Flight training available to fry scholarships only
+                    if (benefitData !== 'chapter33') {
+                      filterOut.push('flightTraining');
```
raquelromano-gov: From the issue, it sounds like they also want the flight training option to include the text "Fry Scholarship only" in parentheses. I am hoping this is also true for the 1990e so we don't need separate enums.

src/js/common/schemaform/reducers/index.js
```diff
@@ -91,7 +91,22 @@ export default function createSchemaFormReducer(formConfig) {
   return (state = initialState, action) => {
     switch (action.type) {
       case SET_DATA: {
-        const newState = _.set([action.page, 'data'], action.data, state);
+        // TODO: Write the documentation that says this is a thing (persistent schema
```
raquelromano-gov: While this makes sense, I'm a little hesitant to push it just before a launch, but maybe some testing would help. Is there a test that could be added to `test/common/schemaform/reducers/index.unit.spec.js` that exercises `recalculateSchemaAndData()` in the way that requires this fix to only update the data and not the schema or uiSchema?

J-Quag:
@raquelromano-gov     / @cvalarida  As of early afternoon I am still seeing the Top up. Is this just not merged?  (that is what im guessing since its still "in progress?"

raquelromano-gov:
@j-quag yes, still in review

src/js/edu-benefits/pages/schoolSelection.js
cvalarida: Er, yeah. That's the one!

src/js/edu-benefits/5490/config/form.js
cvalarida: I think it was an either or, but @J-Quag, could you verify? If we only add the option in the correct scenario, do we need to edit the label?

src/js/common/schemaform/reducers/index.js
cvalarida: Hmm...probably, but given your PR (https://github.com/department-of-veterans-affairs/vets-website/pull/5172), I'm thinking this one should be changed to address only the schemaform reducer modification. I don't have any ideas for that test off-hand, though. When it becomes more pertinent, we can loop back and touch on it again.

cvalarida:
@raquelromano-gov's https://github.com/department-of-veterans-affairs/vets-website/pull/5172 accomplishes the same thing without the change to the reducer, so I'm closing this for now. We should probably still look into modifying that reducer code, though, so we always start from a known state.

PR #5161, opened 2017-03-24
Fixed unintended GIBCT calculator elements from appearing with NaN values.

U-DON:
Resolves department-of-veterans-affairs/vets.gov-team#1819.

Some of the `NaN` issues started occurring when some profile attributes from API responses changed to uppercase, since we weren't doing case-insensitive matching on some calculations. As far as I can tell, we only have string matching with two institution attributes, which are `country` and `type`, so I made the changes for those.

There are also other cases when the institution has null values for things that are still relevant for calculation, in which case we will set default options in the form (set upon successful fetch in the reducer).


bshyong (approved):
LGTM

PR #5162, opened 2017-03-24
Separated GIBCT filter for institution type name into "category" and "type".

U-DON:
Resolves department-of-veterans-affairs/vets.gov-team#1814.

The related change in GIDS can be found here: department-of-veterans-affairs/gibct-data-service#235.


bshyong (approved):
LGTM

PR #5167, opened 2017-03-24
1990e remove learn more links

annekainicUSDS:
Closes https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1785

Creates custom benefit options for the 1990e in order to remove the learn more links. Seemed like the cleanest way of doing this as opposed to altering the common list. Also added language to the warning box.

One question for @raquelromano-gov: the alert box looks kind of far down. It has some top margin applied to it right now that I would like to get rid of. I tried adding a custom `className` to the `uiSchema` on the view field `view:benefitDescription`, but that did not work. Not sure if you happen to know of a way to do this easily?

screen shot 2017-03-24 at 2 42 07 pm

### DIFF
--- a/src/js/edu-benefits/1990e/config/form.js
+++ b/src/js/edu-benefits/1990e/config/form.js
@@ -20,12 +20,13 @@ import ConfirmationPage from '../containers/ConfirmationPage';
 
 import {
   transform,
-  eligibilityDescription
+  eligibilityDescription,
+  benefitsLabels
 } from '../helpers';
 
-import {
-  benefitsLabels
-} from '../../utils/helpers';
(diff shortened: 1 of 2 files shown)


raquelromano-gov:
Yeah, I see what you mean about the top margin. I haven't yet needed to do much styling, so not sure why the custom class didn't work. I'd say go ahead and merge this change and we can revisit/polish later.

raquelromano-gov (approved):
LGTM

annekainicUSDS:
@raquelromano-gov sounds good, I'll create a ticket to revisit at another time. Thanks!

PR #5172, opened 2017-03-25
5490: Filter education type menu options depending on other form data

raquelromano-gov:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1794

This picks up @cvalarida 's https://github.com/department-of-veterans-affairs/vets-website/pull/5157 but gets around making changes to the reducer. We should still revisit whether to go that route in the future, but this change touches fewer things, and allows us to decouple the reducer changes from the edu form launches.

### DIFF
--- a/src/js/edu-benefits/5490/config/form.js
+++ b/src/js/edu-benefits/5490/config/form.js
@@ -541,38 +541,26 @@ const formConfig = {
               educationType: {
                 'ui:options': {
                   updateSchema: (pageData, form, schema) => {
-                    // updateSchema mutates the schema, so after an option is
-                    //  removed from the enum, either I can add it back in manually
-                    //  or take the base as the educationType definition in
-                    //  vets-json-schema instead of the schema passed in.
+                    // TODO: Reorder fields...unless it's a universal change...
+                    // https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1794
+
                     const newSchema = _.cloneDeep(schema);
-                    // const benefitData = _.get('benefitSelection.data.benefit', form);
-                    // const relationshipData = _.get('applicantInformation.data.relationship', form);
-
-                    // console.log('-------------');
-                    // console.log(`relationship: ${relationshipData}, benefit: ${benefitData}`);
-
-                    // When these option are removed, they're removed permanently
+                    const benefitData = _.get('benefitSelection.data.benefit', form);
+                    const relationshipData = _.get('applicantInformation.data.relationship', form);
 
                     // Remove tuition top-up
                     const filterOut = ['tuitionTopUp'];
                     // Correspondence not available to chapter35 children
-                    // if (benefitData === 'chapter35' && relationshipData === 'child') {
-                    //   filterOut.push('correspondence');
-                    // }
+                    if (benefitData === 'chapter35' && relationshipData === 'child') {
+                      filterOut.push('correspondence');
+                    }
                     // Flight training available to fry scholarships only
-                    // if (benefitData !== 'chapter33') {
-                    //   filterOut.push('flightTraining');
-                    // }
+                    if (benefitData !== 'chapter33') {
+                      filterOut.push('flightTraining');
+                    }
 
-                    // console.log('removing:', filterOut);
-                    // console.log('original options (schema):', schema.enum);
-                    // console.log('original options (newSchema):', newSchema.enum);
                     newSchema.enum = _.without(filterOut)(newSchema.enum);
-                    // console.log('modified options (schema):', schema.enum);
-                    // console.log('modified options (newSchema):', newSchema.enum);
 
-                    // console.log('newSchema:', newSchema);
                     return newSchema;
                   }
                 }
--- a/src/js/edu-benefits/5490/config/form.js
+++ b/src/js/edu-benefits/5490/config/form.js
@@ -537,6 +537,45 @@ const formConfig = {
             educationProgram: {
               name: {
                 'ui:title': 'Name of school, university, or training facility you want to attend'
+              },
+              educationType: {
+                'ui:options': {
+                  updateSchema: (pageData, form, schema) => {
+                    // updateSchema mutates the schema, so after an option is
+                    //  removed from the enum, either I can add it back in manually
+                    //  or take the base as the educationType definition in
+                    //  vets-json-schema instead of the schema passed in.
+                    const newSchema = _.cloneDeep(schema);
+                    // const benefitData = _.get('benefitSelection.data.benefit', form);
+                    // const relationshipData = _.get('applicantInformation.data.relationship', form);
+
+                    // console.log('-------------');
+                    // console.log(`relationship: ${relationshipData}, benefit: ${benefitData}`);
+
+                    // When these option are removed, they're removed permanently
+
+                    // Remove tuition top-up
+                    const filterOut = ['tuitionTopUp'];
+                    // Correspondence not available to chapter35 children
+                    // if (benefitData === 'chapter35' && relationshipData === 'child') {
+                    //   filterOut.push('correspondence');
+                    // }
+                    // Flight training available to fry scholarships only
+                    // if (benefitData !== 'chapter33') {
+                    //   filterOut.push('flightTraining');
+                    // }
+
+                    // console.log('removing:', filterOut);
+                    // console.log('original options (schema):', schema.enum);
+                    // console.log('original options (newSchema):', newSchema.enum);
+                    newSchema.enum = _.without(filterOut)(newSchema.enum);
+                    // console.log('modified options (schema):', schema.enum);
+                    // console.log('modified options (newSchema):', newSchema.enum);
+
+                    // console.log('newSchema:', newSchema);
+                    return newSchema;
+                  }
+                }
               }
             }
           }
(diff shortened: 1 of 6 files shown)


src/js/edu-benefits/5490/config/form.js
```diff
@@ -537,6 +537,31 @@ const formConfig = {
...
+                    // Correspondence not available to chapter35 children
+                    if (benefitData === 'chapter35' && relationshipData === 'child') {
+                      filterOut.push('correspondence');
+                    }
+                    // Flight training available to fry scholarships only
+                    if (benefitData !== 'chapter33') {
+                      filterOut.push('flightTraining');
+                    }
+
+                    newSchema.enum = _.without(filterOut)(labels);
```
annekainicUSDS: I'm just a little confused what this syntax is doing. Is `labels` also being passed to `_.without`? Doesn't `_.without` need to know the original array it is removing the `filterOut` values from?

src/js/edu-benefits/5490/config/form.js
```diff
@@ -537,6 +537,31 @@ const formConfig = {
             educationProgram: {
               name: {
                 'ui:title': 'Name of school, university, or training facility you want to attend'
+              },
+              educationType: {
+                'ui:options': {
+                  updateSchema: (pageData, form, schema) => {
```
annekainicUSDS: Where does `updateSchema` come from? Can't find this in the rjsf docs.

src/js/edu-benefits/definitions/educationProgram.js
```diff
@@ -2,31 +2,18 @@ import _ from 'lodash/fp';
 
 import { showSchoolAddress } from '../utils/helpers';
 import * as address from '../../common/schemaform/definitions/address';
-import educationTypeUISchema from '../definitions/educationType';
+import educationTypeUISchema from './educationType';
 
 export const uiSchema = {
   'ui:order': ['name', 'educationType', 'address'],
-  address: address.uiSchema(),
+  address: _.merge(address.uiSchema(), {
```
annekainicUSDS: Is the only thing this change is doing is simplifying the code? And looks like the 1990e is also using this file, so just want to make sure the changes here don't negatively impact the 1990e.

annekainicUSDS (commented):
Looks good! Just had a few questions for you.

src/js/edu-benefits/5490/config/form.js
raquelromano-gov: Yes, that's right: `labels` is the original array and is passed to `_.without()` to filter out the given items. I found this documentation useful since we are using lodash/fp, which has slightly different syntax from lodash: https://gist.github.com/jfmengels/6b973b69c491375117dc#_withoutvalues-array

src/js/edu-benefits/5490/config/form.js
raquelromano-gov: `updateSchema` is a callback that's used by the reducer: https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/js/common/schemaform/helpers.js#L513

We only use it here and maybe one other place. What I've done below is a bit of a hack (stealing the keys from the UI object to get the original labels list). @cvalarida has  a more elegant solution in https://github.com/department-of-veterans-affairs/vets-website/pull/5157, but I thought we should discuss before making this change (especially just before launch). @cvalarida feel free to chime in here if you feel strongly we should go with your reducer changes instead!

src/js/edu-benefits/definitions/educationProgram.js
raquelromano-gov: This is branched from https://github.com/department-of-veterans-affairs/vets-website/pull/5157 so give credit to @cvalarida, but yes, looks like we now have the `showSchoolAddress()` helper so don't need to write a custom callback.

@cvalarida I'm not sure what issue this connects to  (only ask for address for certain education types)--can you add to the PR description if there is one? And are we sure it applies to both the 5490 and 1990e? I would assume/hope so, but you never know.

raquelromano-gov:
@cvalarida @annekainicUSDS  I spent some time trying to figure out why there's a unit test failing inside the `updateSchema()` callback (`schoolSelection.unit.spec.jsx`). It looks like when `updateSchema(data, formData, currentSchema`) is called in the test environment, `formData` isn't defined. All is fine in my local demo (and verified with Redux chrome dev tool that it's defined)), but not in the test.

Tracing back where `formData` is defined in the test environment, it looks like it's the app `state` passed in by `DefinitionTester` ([here](https://github.com/department-of-veterans-affairs/vets-website/blob/master/test/util/schemaform-utils.jsx#L29)), which makes me wonder if/whether/why the `state` wouldn't be well-defined.

(Aside: I also don't understand why both in `DefinitionTester` and in `recalculateSchemaAndData()` ([here](https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/js/common/schemaform/reducers/index.js#L43)), we pass in `state` instead of `formData` unless we're using `formData` to mean something different).

@cvalarida I think you're the most familiar with this part of the code, so LMK if you have any thoughts (or if you think we should move back to Plan A (https://github.com/department-of-veterans-affairs/vets-website/pull/5157).

raquelromano-gov:
And to complicate things, the jenkins build failure reported here is a remnant of some permissions changes last week that made it impossible to run test coverage. That's now fixed, but somehow still affects this branch. I can ask devops folks, but want to resolve the known unit test failures first.

src/js/edu-benefits/5490/config/form.js
cvalarida: @annekainicUSDS An example of `updateSchema` can be found in the schemaform docs [here](https://github.com/department-of-veterans-affairs/vets-website/blob/master/docs/schemaform/form-config.md#i-want-to-change-the-options-of-a-dropdown-based-on-some-other-field-data) (I've a PR out changing the variable names to be more descriptive, though).

@raquelromano-gov I'm all for using this PR instead! I wasn't too comfortable changing the reducer without @jbalboni's review, but I'm pretty sure it'll be alright; there are only two places that use `updateSchema` right now, and they're both here in the 5490, and the change will only slow down the `GET_DATA` action _(insert proper term here)_ by < 3 ms (on my machine, it was usually < 1 ms).

That said, I'm more than happy to hold off on it for now in case I missed something. Taking the the keys from the UI object is a clever workaround :smile:

src/js/edu-benefits/definitions/educationProgram.js
cvalarida: The short story is, "It does the same thing as before, but uses `hideIf` instead of `updateSchema` because it made more sense."

The long story goes like this:

This was a little cleanup from before the advent of `hideIf`. @jbalboni was basically doing a `hideIf` manually in `updateSchema`, but now that we have `hideIf`, I figured we ought to use it. I ran across this while looking for instances of `updateSchema` callbacks before I modified the reducer.

Also, I couldn't have conditionally updated the `educationType` dropdown _and_ kept the conditionally-hidden address calling the original `educationProgram`'s `updateSchema` in the new 5490 version's `updateSchema` (which is gross to explain...):

```js
const schoolSelectionPage = createSchoolSelectionPage(fullSchema5490, { ... });
...
schoolSelection: _.merge(schoolSelectionPage, {
  uiSchema: {
    educationProgram: {
      educationType: {
        'ui:options': {
          updateSchema: (pageData, form, schema) => {
            // Conditionally hide address
           let newSchema = schoolSelectionPage.uiSchema.educationProgram.educationType['ui:options'].updateSchema(pageData, form, schema);
            // Remove the options from newSchema
            ...
            return newSchema;
          }
        }
      }
    }
  }
})
```

As far as which forms it affects, anywhere `educationProgram` is used, this was making the address hide conditionally via `updateSchema`. In theory, `updateSchema` could have been overridden in a given form, but there wasn't a place that happened.

cvalarida (approved):

cvalarida:
@raquelromano-gov If I recall correctly, `state` was for the state of the rest of the pages, whereas `formData` is for the current page (again, before the name standardization when a page was a form, and a form was...a bunch of forms, hence the misleading name).

@jbalboni and I talked about it and figured the best course ahead was to just add a `state` (or perhaps there was already one there?) prop that could be used to outline the data on the rest of the pages (as it would be needed here).

That said, I'm looking into it now in more detail and will get back to you when I figure out anything more. That's just what I remember so far.

cvalarida:
Okay, I see what's going on here. The easiest solution would be to render `DefinitionTester` with something like:

```js
import educationType from '../../../../src/js/edu-benefits/definitions/educationType';
...
const form = ReactTestUtils.renderIntoDocument(
  <DefinitionTester
      schema={schema}
      data={{}}
      state={{
        schoolSelection: {
          uiSchema: {
            educationProgram: {
              educationType
            }
          }
        }
      }}
      definitions={formConfig.defaultDefinitions}
      uiSchema={uiSchema}/>
);
```

It sort of decouples the `educationType` from the school selection page, but I don't think that's too much of a problem if we import it from the same place as the education selection page does.

Still annoying...There's probably a better solution, but this gets it up an running, anyhow.

src/js/edu-benefits/definitions/educationProgram.js
raquelromano-gov: Thanks for the thorough explanation.

Tragically, the elegant `hideIf()` solution makes the unit test run ever so slightly more slowly (could not get the address inputs to show up after changing type to `college` without a timeout 😢 ). I'll comment out the timeout to avoid flakiness, but we may want to discuss solutions at the next FE architecture sync.

raquelromano-gov:
Thank you!! Stuffed in the state prop and that did the trick.

PR #5173, opened 2017-03-25
Addressed GIBCT design review items related to search results.

U-DON:
Related to search results section of department-of-veterans-affairs/vets.gov-team#1507.

### Changes
* Filter changes also reset page number to 1.
* Choosing same filter options don't update route and refresh screen.
* Filter button on mobile brings up form that overlays entire screen.
* Changed "Your estimated benefits" on cards to "You may be eligible for up to:".
* Spacing between GI Bill student count and "You may be eligible for up to:".


saintsoup52:
This might not have anything to do with this pull request... but the filters country and state should be reversed.  I would say 99% of schools are in a "state" and therefore that filter is much more important to the user. 

In the current tool "State" comes first and then "Country".

**Current State**
image

bshyong (approved):
LGTM

U-DON:
@saintsoup52: That's a design choice by @melwoodard, though I see "Country" being before "State" making more sense hierarchically.

PR #5176, opened 2017-03-27
Pass the logout function to SignInProfileMenu component

omgitsbillryan:
The `handleLogout` function is not being called when clicked because the function is not being passed as a property to the `SignInProfileMenu` component.

See https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1872

### DIFF
--- a/src/js/login/components/SearchHelpSignIn.jsx
+++ b/src/js/login/components/SearchHelpSignIn.jsx
@@ -27,7 +27,8 @@ class SearchHelpSignIn extends React.Component {
             this.props.onClickSearchHelpSignIn('account', !login.utilitiesMenuIsOpen.account);
           }}
           greeting={greeting}
-          isOpen={login.utilitiesMenuIsOpen.account}/>);
+          isOpen={login.utilitiesMenuIsOpen.account}
+          onUserLogout={this.props.onUserLogout}/>);
     } else {
       content = (<div>
         <a href="#" onClick={this.props.onUserLogin}>Sign In</a><span className="signin-spacer">|</span><a href="#" onClick={this.props.onUserSignup}>Register</a>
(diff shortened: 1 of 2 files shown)


tiffanybbrown-va:
LGTM, but please get a second set of eyes on this.

omgitsbillryan:
There should be a test for this.  My vote would be to merge this without a test for now since it's in `prod` and create a separate PR with a test.

src/js/login/components/SearchHelpSignIn.jsx
```diff
@@ -27,7 +27,8 @@ class SearchHelpSignIn extends React.Component {
             this.props.onClickSearchHelpSignIn('account', !login.utilitiesMenuIsOpen.account);
           }}
           greeting={greeting}
-          isOpen={login.utilitiesMenuIsOpen.account}/>);
+          isOpen={login.utilitiesMenuIsOpen.account}
+          onUserLogout={this.props.onUserLogout}/>);
```
cvalarida: `onUserLogout` should probably be added to the `propTypes` of both this (though this has no `propTypes` specified...) and `SignInProfileMenu` (`src/js/login/components/SignInProfileMenu.jsx`).

cvalarida (approved):
Looks golden with one suggestion for clarity later. As a newcomer here, I always find self-documenting `propTypes` helpful.

omgitsbillryan:
As a first time FE test-writer, I'm having some trouble determining if this should be an e2e test or unit test; I'm leaning e2e.  

```javascript
// set initial state as 'logged in'
// click profile menu drop down
// click 'Sign Out'
// expect new window to open with `logout_via_get` url
```

@tiffanybbrown-va @cvalarida am I on the right track?  Any pointers?

cvalarida:
@omgitsbillryan Yeah, I'm leaning toward E2E as well since the problem was with clicking the button. With a unit test, you'd be more checking that the logout function works, not that you can log out from the interface.

Your commented code block looks like it's on the right track, but I'm not familiar with the login code at all, so that's about all the help I can be there, I'm afraid!

PR #5177, opened 2017-03-27
Remove employmentHistory object

cvalarida:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1863

Not much to say here. Flattens out the `employmentHistory` object in the `employmentHistory` page in the `employmentHistory` chapter since `employmentHistory` isn't a property in the schema.

How many times can I say `employmentHistory`? I think I'll stop there.


annekainicUSDS (approved):
LGTM

markolson:
Testing locally real quick, then will 👍

PR #5064, opened 2017-03-10
Add appointment stats to Facility Locator detail pages

bshyong:
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1588

screenshot:
screen shot 2017-03-10 at 11 07 23 am

PR #5178, opened 2017-03-28
Revert "Add appointment stats to Facility Locator detail pages (#5064)"

bshyong:
This reverts commit 18804db3719f918a0dc09e457151190abcb9148a.

Reverts PR relating to issue https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1588


ayaleloehr:
For those watching at home, we needed to quickly get this code into staging, so it was merged into `master` this afternoon but is not yet ready for production, so it was reverted here. Ideally we would have just deployed the unmerged branch to dev, but with the short timeline and need to demo to leadership, we were concerned about making it and then it getting overwritten with that approach, so we just went with the `master` merge + revert.

PR #5191, opened 2017-03-29
Fixed BAH estimations for GIBCT search results.

U-DON:
With Georgetown University as a test and setting classes to in-person for eligibility, it was showing a default $1,611 housing allowance (average BAH, which is the default estimate for foreign schools) when it had $2,301 for its `bah` attribute.

Two issues are addressed with this fix: (1) `country` attribute wasn't being selected from the institution, so it came up as `undefined` in conditionals and (2) checking for `type` (and now `country`) were case sensitive.


bshyong (approved):
LGTM

PR #5192, opened 2017-03-29
Scaffolds the 1990n

cvalarida:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1609
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1261

Doesn't do anything yet since the chapters are all gutted, but the route is up and it's ready to be worked on.

### DIFF
--- a/src/js/edu-benefits/1990n/config/form.js
+++ b/src/js/edu-benefits/1990n/config/form.js
@@ -1,0 +1,36 @@
+// import _ from 'lodash/fp';
+
+// import fullSchema1990n from 'vets-json-schema/dist/ncs-benefits-schema.json';
+
+// import pages from '../../pages/';
+
+// import definitions from '../../definitions/';
+
+import IntroductionPage from '../components/IntroductionPage';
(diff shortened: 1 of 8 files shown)


src/js/edu-benefits/1990n/config/form.js
```diff
@@ -0,0 +1,36 @@
+// import _ from 'lodash/fp';
+
+// import fullSchema1990n from 'vets-json-schema/dist/ncs-benefits-schema.json';
```
raquelromano-gov: I really hate these names...wish it were something like `vets-json-schema/dist/form-22-1990n-schema.json`

raquelromano-gov (approved):
Looks splendid

src/js/edu-benefits/1990n/config/form.js
cvalarida: Yeah, me too. Fortunately we only reference them the once here.

PR #5193, opened 2017-03-29
Scaffolds the 5495

cvalarida:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1610

Doesn't do anything, but the route is up and the form is ready for work.


src/js/edu-benefits/5495/config/form.js
```diff
@@ -0,0 +1,56 @@
+// import _ from 'lodash/fp';
+
+// import fullSchema5495 from 'vets-json-schema/dist/dependents-change-of-program-schema.json';
```
raquelromano-gov: Help. `change-of-program-schema.json`, `dependents-benefits-schema.json`, and now `dependents-change-of-program-schema.json`

@lihanli is it unthinkable to change the names to `form-22-5490-schema.json` before we launch the 1990e and 5490? It's *so confusing* (or maybe my brain is just too small to hold all of these variations)

src/js/edu-benefits/5495/helpers.jsx
```diff
@@ -0,0 +1,33 @@
...
+import { transformForSubmit } from '../../common/schemaform/helpers';
+
+export const bankAccountChangeLabels = {
+  startUpdate: 'Start or update direct deposit',
+  stop: 'Stop direct deposit',
+  noChange: 'No change to payment method'
+};
+
+export function transform(formConfig, form) {
+  // All the king's horses and all the king's men
```
raquelromano-gov: Do we know we'll need this stuff for these forms? I know one of the forms needed it, but not all. if not sure, would just stringify the json and return for now.

raquelromano-gov (approved):
LGTM (just one suggestion, plus a plea to Lihan)

src/js/edu-benefits/5495/config/form.js
lihanli: i have a pr open for the name change https://github.com/department-of-veterans-affairs/vets-json-schema/pull/78

raquelromano-gov: Yay! Thanks, will look now.

src/js/edu-benefits/5495/helpers.jsx
cvalarida: An excellent point. Cleaned up!

If we do find we need it, this form was duplicated from the 1995, so we'll find this code in there.

PR #5200, opened 2017-03-29
Addressed GIBCT design review items related to profile pages.

U-DON:
Related to profile section of department-of-veterans-affairs/vets.gov-team#1507.

### Changes
* Scroll to top of profile page when loading and after loading finishes.
* Hide "No Data" if there's no national average line in outcome graphs.
* Fixed funky padding issues.
* Other style details pictured below.

### DIFF
--- a/src/js/gi/components/profile/Outcomes.jsx
+++ b/src/js/gi/components/profile/Outcomes.jsx
@@ -36,6 +36,13 @@ class Outcomes extends React.Component {
         </div>
 
         <div className="medium-12 column">
+          <svg>
+            <rect x="0" y="0" width="40" height="40" fill="#F1F1F1"/>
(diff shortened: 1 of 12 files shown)


src/js/gi/components/profile/Outcomes.jsx
```diff
@@ -36,6 +36,15 @@ class Outcomes extends React.Component {
         </div>
 
         <div className="medium-12 column">
+          <svg className="graph-key" viewBox="0 0 650 50">
```
bshyong: Can we potentially extract this SVG into a component instead of having it inline like this?

bshyong (approved):
LGTM; minor comment on how we are implementing SVGs

src/js/gi/components/profile/Outcomes.jsx
U-DON: Sure. I think I'll keep the component within this module though, since it's only used here.

PR #5205, opened 2017-03-30
Notes about troubleshooting e2e tests

omgitsbillryan:
As someone who was finally able to get e2e tests working locally, I'm adding the notes that I found helpful.

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,6 +222,26 @@ TODO(awong): Figure out sauce labs integrations. Not all e2e tests should always
 run on all browsers. That's wasteful. How do we determine what should be run on
 multiple browsers as opposed to on PhantomJS in Jenkins?
 
+#### E2E Troubleshooting
+##### Is `selenium` Installed?
+1. Navigate to `vets-website` dir
+1. Create new file find-selenium.js with the contents of:
+```
+const selenium = require('selenium-server');
+console.log(selenium.path);
+```
+1. Run `$ node find-selenium.js` and you should see:
+```
+<YOUR_HOME>/Github/vets-website/node_modules/selenium-server/lib/runner/selenium-server-standalone-3.1.0.jar
+```
+
+##### Is Java 8 Installed?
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -222,6 +222,26 @@ TODO(awong): Figure out sauce labs integrations. Not all e2e tests should always
 run on all browsers. That's wasteful. How do we determine what should be run on
 multiple browsers as opposed to on PhantomJS in Jenkins?
 
+#### E2E Troubleshooting
+##### Is `selenium` Installed?
```
jkassemi: Since selenium is part of the package, I don't think we'll run into a case where it's not installed, just issues running it. I think the `java -jar <selenium-jar-path>` test might be worth including here.

jkassemi (approved):

omgitsbillryan:
@jkassemi good point.  I think that `java` test actually accomplishes both by proving you have the right java version and in my example it points to where `selenium` should be installed.

ayaleloehr:
LGTM. thank you for documenting your struggles to help the next person!

README.md
```diff
@@ -222,6 +222,17 @@ TODO(awong): Figure out sauce labs integrations. Not all e2e tests should always
 run on all browsers. That's wasteful. How do we determine what should be run on
 multiple browsers as opposed to on PhantomJS in Jenkins?
 
+#### E2E Troubleshooting
+Try running your `selenium` server manually:
+```
+$ java -jar <YOUR_HOME>/Github/vets-website/node_modules/selenium-server/lib/runner/selenium-server-standalone-3.1.0.jar
```
cvalarida: It's nit-picky, but maybe the path would be better written as `<path to vets-website>/node_modules/...` or something. I only say that because I don't store my copy in `~/Github/` is all.

cvalarida (approved):
Looks good. If there's any other trouble, we at least have a place to document it now.

PR #5219, opened 2017-03-30
Refactor breadcrumbs to match common VA breadcrumb style

bshyong:
resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1911

refactors GIBCT breadcrumbs so that html structure and styles match common VOA breadcrumb style. This will improve accessibility and contribute to more consistency across apps

### DIFF
--- a/src/js/gi/components/heading/Breadcrumbs.jsx
+++ b/src/js/gi/components/heading/Breadcrumbs.jsx
@@ -28,9 +28,15 @@ class Breadcrumbs extends React.Component {
       crumbs.push(<span key="profile"><strong>{this.props.profileName || 'Profile'}</strong></span>);
     }
 
-    return (<div className="gi-breadcrumbs">
-      {crumbs.reduce((content, e) => { return [...content, ' › ', e]; }, []).slice(1)}
-    </div>);
+    return (
+      <nav className="va-nav-breadcrumbs">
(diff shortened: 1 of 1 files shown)


src/js/gi/components/heading/Breadcrumbs.jsx
```diff
@@ -28,9 +28,15 @@ class Breadcrumbs extends React.Component {
...
     }
 
-    return (<div className="gi-breadcrumbs">
-      {crumbs.reduce((content, e) => { return [...content, ' › ', e]; }, []).slice(1)}
-    </div>);
+    return (
+      <nav className="va-nav-breadcrumbs">
+        <ul className="row va-nav-breadcrumbs-list gi-breadcrumbs" role="menubar" aria-label="Primary">
+          {crumbs.reduce((content, e, i) => {
+            return [...content, ' ', <li key={i} className={i === crumbs.length - 1 ? 'active' : ''}>{e}</li>];
```
U-DON: Why is the `' '` necessary?

U-DON: When I tested this out, the extra space seems to push each item further right, making the arrows look off center. If we don't need it, maybe we could just do `crumbs.map(...)` instead of `reduce`.

bshyong: Agreed—I'll map over these elements since the arrow insertions are now done via CSS

U-DON (approved):
LGTM.

PR #5220, opened 2017-03-31
Fixed loading GIBCT profile pages for accredited institutions with null accreditation type.

U-DON:
Resolves department-of-veterans-affairs/vets.gov-team#1843.

Loading the pages took multiple clicks because of an error from calling `toUpperCase()` on a `null` `accreditationType` when `accredited` is `true`. First click triggers error, second click does a full page load of the linked URL.

Also did some light refactoring on the way.


bshyong (approved):
LGTM

PR #5227, opened 2017-03-31
Relabel the name fields for previous benefit sponsor

cvalarida:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1925

image

Having the possessive on the SSN label is inconsistent. Should we change that?


goldenmeanie:
Yeah, can just be Sponsor Social Security number

goldenmeanie:
Otherwise LGTM

PR #5228, opened 2017-03-31
Update file names from json schema

annekainicUSDS:
After this PR was merged to update the file names in vets-json-schema (https://github.com/department-of-veterans-affairs/vets-json-schema/pull/78), wanted to make sure those file names were updated here as well.

Also includes a change to `bankAccountChange`: moving it to be imported from `properties` instead of `definitions` because the change in this PR (https://github.com/department-of-veterans-affairs/vets-json-schema/pull/75/files#diff-dd7516d6db2bea9ee53d8ead45def186L422) to move it to `definitions` causes it not to be found.

### DIFF
--- a/test/edu-benefits/1990/schema/schema.unit.spec.js
+++ b/test/edu-benefits/1990/schema/schema.unit.spec.js
@@ -179,8 +179,7 @@ function createTestVeteran() {
 }
 
 describe('Edu benefits json schema', () => {
-  it.only('should have matching test veteran and blank veteran shape', () => {
-    console.log(schema); // eslint-disable-line
+  it('should have matching test veteran and blank veteran shape', () => {
     const testForm = qc.objectLike(createTestVeteran())(1);
     const blankForm = createVeteran();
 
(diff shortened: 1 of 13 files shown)


lihanli (approved):

test/edu-benefits/1990/schema/schema.unit.spec.js
```diff
@@ -1,6 +1,6 @@
 import { expect } from 'chai';
 import Ajv from 'ajv';
-import { eduBenefits as schema } from 'vets-json-schema';
```
raquelromano-gov: The failure seems to be here, but I'm not clear on how this worked before, and why it wouldn't work with your change.

annekainicUSDS: I'm not sure how it was passing before either, or why this would cause an issue. Tests pass also pass locally for me. Going to try to restart the build and see if it was just a fluke (which seems to have happened before).

lihanli: try changing to
```javascript
import schemas from 'vets-json-schema';
const schema = schemas['22-1990'];
```
if the next build fails

aub (approved):
This LGTM, although I'm no expert on the schema-related changes.

PR #5229, opened 2017-03-31
E2E Logout Test

omgitsbillryan:
**Context**
The "Sign Out" button broke as a result of some refactoring.  A [PR was quickly merged to fix issue in production](https://github.com/department-of-veterans-affairs/vets-website/commit/442c1c08351a911d639a2f3a27b805f6140eb986).

This PR is to add the necessary e2e test that we didn't have time for in the aforementioned PR.

**Relevant Issue**
https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1893

### DIFF
--- a/test/login/login.e2e.spec.js
+++ b/test/login/login.e2e.spec.js
@@ -6,6 +6,7 @@ if (!process.env.BUILDTYPE || process.env.BUILDTYPE === 'development') {
   module.exports = E2eHelpers.createE2eTest(
     (client) => {
       const token = LoginHelpers.getUserToken();
+      const logoutUrl = LoginHelpers.getLogoutUrl();
 
       // log in and wait for little person icon to appear next to the username
       LoginHelpers.logIn(token, client, '/', 3)
@@ -21,11 +22,11 @@ if (!process.env.BUILDTYPE || process.env.BUILDTYPE === 'development') {
       // click Sign Out & verify new window has correct logout url
       client
         .click('#accountMenu > ul > li:nth-child(2) > a')
-        .windowHandles(function foo(result) {
+        .windowHandles(function windowHandlesCallback(result) {
           this.verify.equal(result.value.length, 2, 'There should be 2 windows open');
           const newWindow = result.value[1];
           this.switchWindow(newWindow);
-          this.verify.urlContains(LoginHelpers.getLogoutUrl());
+          this.verify.urlContains(logoutUrl);
         });
 
       client.end();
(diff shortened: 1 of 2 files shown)


test/util/login-helpers.js
```diff
@@ -16,6 +16,10 @@ function setUserToken(token, client) {
   });
 }
 
+function getLogoutUrl() {
+  return 'http://fake_logout_url';
```
omgitsbillryan: I couldn't find a better place to declare constants for fake test data.

test/login/login.e2e.spec.js
```diff
@@ -0,0 +1,35 @@
...
+if (!process.env.BUILDTYPE || process.env.BUILDTYPE === 'development') {
+  module.exports = E2eHelpers.createE2eTest(
+    (client) => {
+      const token = LoginHelpers.getUserToken();
+      const logoutUrl = LoginHelpers.getLogoutUrl();
+
+      // log in and wait for little person icon to appear next to the username
+      LoginHelpers.logIn(token, client, '/', 3)
+        .assert.title('Vets.gov')
+        .waitForElementVisible('#login-root > div > div.sign-in-link > div > button > span > svg', Timeouts.slow);
```
omgitsbillryan: These selectors were all auto-generated using chrome developer tools.

cvalarida (approved):
I'm not sure what the deal is with the new window bit, but the test itself succeeds, so I'll just assume you're testing for the right thing there :wink: 

The rest looks good!

omgitsbillryan:
@cvalarida When a user clicks the "Sign Out" button in Vets.gov, it opens a popout window with the ID.me logout url.

raquelromano-gov:
Just noticed that the CSS selectors are very specific (e.g. `#accountMenu > ul > li:nth-child(2) > a'` and wondering if there are simpler (and less fragile) ways to find these elements, so that slight changes to the layout don't break the test. I am terrible at defining CSS selectors and always have to look at documentation, else would suggest something.

omgitsbillryan:
@raquelromano-gov I had the same thought - I don't have a lot of expertise in the matter.  I'll see what I can find.

omgitsbillryan:
I couldn't figure out a better way to get the "Sign Out" button.  

I could introduce an `id="va-signout"` on the anchor, but that html is defined within a component which may get reused which would result in a duplicate `id` in the DOM (bad) - so I decided against that.

PR #5232, opened 2017-03-31
Make effective date only required when relationship is child

annekainicUSDS:
When relationship was `spouse`, the benefit selection page wouldn't let you advance because the `benefitRelinquishedDate` was still required, even though it was not expanded. Fix is to change the required status of the `benefitRelinquishedDate` to only be true if the relationship is `child`.


raquelromano-gov (approved):
Thanks for the fix!

PR #5233, opened 2017-03-31
5490 sponsor information carry

cvalarida:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1924

## No Previous Benefits
image

## Previous Benefits
image

image

### DIFF
--- a/src/js/edu-benefits/5490/config/form.js
+++ b/src/js/edu-benefits/5490/config/form.js
@@ -361,32 +361,43 @@ const formConfig = {
                 hideIf: (formData) => _.get('relationship', formData) !== 'spouse'
               }
             },
-            veteranFullName: _.merge(fullNameUi, {
-              'ui:title': 'Name of Sponsor',
-              first: {
-                'ui:title': 'Sponsor first name'
+            'view:currentSameAsPrevious': {
+              'ui:options': {
+                hideIf: (formData) => !formData.previousBenefits['view:claimedSponsorService']
               },
-              middle: {
-                'ui:title': 'Sponsor middle name'
+              'ui:title': 'Same sponsor as previously claimed benefits'
+            },
+            'view:currentSponsorInformation': {
+              'ui:options': {
+                hideIf: (formData) => formData['view:currentSameAsPrevious']
               },
-              last: {
-                'ui:title': 'Sponsor last name'
-              },
-              suffix: {
-                'ui:title': 'Sponsor suffix'
-              }
-            }),
-            'view:veteranId': _.merge(veteranId.uiSchema, {
-              veteranSocialSecurityNumber: {
-                'ui:title': 'Sponsor Social Security number'
-              },
-              'view:noSSN': {
-                'ui:title': 'I don\'t know my sponsor’s Social Security number',
-              },
-              vaFileNumber: {
-                'ui:title': 'Sponsor file number',
-              }
-            }),
+              veteranFullName: _.merge(fullNameUi, {
+                'ui:title': 'Name of Sponsor',
+                first: {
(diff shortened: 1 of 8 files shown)


src/js/edu-benefits/5490/config/form.js
```diff
@@ -361,32 +361,52 @@ const formConfig = {
...
+                hideIf: (formData) => formData['view:currentSameAsPrevious']
               },
-              vaFileNumber: {
-                'ui:title': 'Sponsor file number',
-              }
-            }),
+              veteranFullName: _.merge(fullNameUi, {
+                'ui:options': {
+                  updateSchema: (data, form) => {
+                    if (!_.get('sponsorInformation.data.view:currentSameAsPrevious', form)) {
```
raquelromano-gov: I know this was already here, but I find it a bit more readable to test
```
if (_.get('sponsorInformation.data.view:currentSameAsPrevious', form)) {
  return nonRequiredFullName;
}
return fullName;
```

src/js/edu-benefits/5490/helpers.jsx
```diff
@@ -1,8 +1,23 @@
...
-  const formData = transformForSubmit(formConfig, form);
+  // Clone the form in so we don't modify the original...because of reasons FP
+  const newForm = _.cloneDeep(form);
+
+  // Copy the data if necessary
+  if (form.sponsorInformation.data['view:currentSameAsPrevious']) {
+    newForm.sponsorInformation.data['view:currentSponsorInformation'] = {
+      veteranFullName: form.benefitHistory.data.previousBenefits.veteranFullName,
+      'view:veteranId': {
+        veteranSocialSecurityNumber: form.benefitHistory.data.previousBenefits.veteranSocialSecurityNumber
```
raquelromano-gov: I'm now surprised that for previouesBenefits, we're not allowing either the SSN or VA File Number, but that's another story for another day/PR. (I think we should ask stakeholders about that--would be a better experience to be consistent about what info is required about any sponsor.)

raquelromano-gov (approved):

goldenmeanie:
Is the backend duplicating the info from the previous benefits in the appropriate place in the spool file?

raquelromano-gov:
It should be--Chris updated the function that transforms the data before stringifying the JSON to send to the api.

cvalarida:
@goldenmeanie Yes it is; I did test it manually to confirm that if that checkbox is checked, the information is carried over from the previous benefits sponsor to the current benefits sponsor.

@raquelromano-gov An excellent suggestion! I think we definitely should. I'll add it in here.

cvalarida:
@raquelromano-gov Can you review again? I just added the test for verifying the data is what we expect it to be (thereby adding an additional bit of functionality to this particular schema test file).

src/js/edu-benefits/5490/helpers.jsx
raquelromano-gov: Verified this is a good idea--separate ticket filed and added to current sprint: https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1929

raquelromano-gov (approved):
LGTM

test/edu-benefits/5490/schema/schema.unit.spec.js
```diff
@@ -26,5 +27,13 @@ describe('5490 schema tests', () => {
         }
         expect(result.valid).to.be.true;
       });
+      // Expect the data to be what's found in expected-data/${file}
+      if (expectedDataFiles.indexOf(file) !== -1) {
```
raquelromano-gov: Should probably add a comment here to replace this with includes() once the version of Node.js on Jenkins is upgraded

PR #5237, opened 2017-04-01
E2e helper code reorg

raquelromano-gov:
Closes https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1742
This is the last bit of the cleaning up our e2e helper code. I found there were a couple of unit tests that rely on the helper code (which I had previously thought was *only* used in e2e tests), so now not as sure the reorg is as much of an improvement. Feel free to add any other reviewers.


annekainicUSDS (approved):
LGTM! The only minor thing is that now the unit tests are importing helpers that live in a `e2e` file path name, but I don't think it's that big of a deal.

PR #5239, opened 2017-04-02
More 1990e e2e tests

raquelromano-gov:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1614

Finishes off the rest of 1990e e2e tests...hopefully some of this will be usable by the 5490 tests as well.

Also, this still doesn't run them on Jenkins, but at least one can run them locally and possibly catch any breakage. (Though it will take some dev discipline to routinely run e2e tests locally when making feature changes.)


config/nightwatch.js
```diff
@@ -19,6 +19,9 @@ module.exports = {
   test_workers: false,
   test_settings: {
     'default': {
+      globals: {
+        buildtype: process.env.BUILDTYPE
```
cvalarida: @annekainicUSDS Who would be the right person / people from core to verify this is golden?

test/edu-benefits/1990e/00-required.e2e.spec.js
```diff
@@ -1,51 +1,104 @@
...
-
-    // Sponsor information page
-    client
-      .waitForElementVisible('input[name="root_veteranFullName_first"]', Timeouts.slow);
-    EduHelpers.completeVeteranInformation(client, testData.sponsorVeteran.data, true);
-    client.click('.form-progress-buttons .usa-button-primary');
-    E2eHelpers.expectNavigateAwayFrom(client, '/sponsor/information');
-    */
-    client.end();
+    if (client.globals.buildtype !== 'production') {
```
cvalarida: It's nitpicky, but maybe add a comment explaining this is necessary because the path isn't open when the buildtype is `'production'`. Just on the off-chance we forget this pain. I'd just hate for this to be copied over to another test that it shouldn't be, or forgotten about entirely, y'know?

raquelromano-gov: Oh yeah, good point--will do that.

config/nightwatch.js
raquelromano-gov: @jkassemi  Can you look at this change to our nightwatch config and comment? Alternatively, I can define a global to guard this feature, but in that case, where would be the best place to set that global so that it is available to both our e2e tests and to https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/js/edu-benefits/routes.jsx (and so that it enables the app in development but not in production)?

jkassemi: This makes sense; 👍  if it's clear and works for everyone.

There's an existing pattern (https://github.com/department-of-veterans-affairs/vets-website/blob/master/test/messaging/01-required-desktop.e2e.spec.js#L7) that checks the env's buildtype before it registers the tests with nightwatch. That version targets the dev environment explicitly, but the general pattern is similar. 

Given that the e2e's are run against a build, it might be nice to have a single place (maybe a `config/settings.js`) to set that global to apply it to both and allow overrides:

```
// Set defaults based on buildtype

defaults = {
    development: {
       feature_1990e: true,
       ...
    },
    ...
}[process.env.BUILDTYPE]

// Override defaults if specified by environment
export {
   ...defaults,
   ...removeEmpty({
     feature_1990e: process.env.FEATURE_1990E,
     ...
   })
}
```

though it may be unnecessarily complex. We could also have the build expose information about what features it has enabled so the e2e's can determine dynamically which features to test based on the build. We've had a case or two where we forgot to remove the guards from the e2e's when the feature was launched this approach might help with.

cc @robbiethegeek-usds

raquelromano-gov: OK, I removed the `nightwatch.js` edits and am checking the buildtype outside of the test creation, so the tests themselves don't need to know the buildtype.

I do like the idea of tracking individual unlaunched features, because then we can grep for that key when launching the feature and make sure there are no lingering disabled tests. I'll move this conversation into a separate issue so it doesn't get lost when this PR is merged.

cvalarida (approved):
Looks good!

Issue #5211, opened 2017-03-30
GI Bill CT: Outcomes should be rounded to the whole number or to tenths of a percent


saintsoup52:
The WIP version of the tool is showing hundreds of a percent in the outcome graphs which is not how the current tool works and is way too much detail for a user to comprehend.

The tool should round the round that are displayed to the whole number (or at the most tenths of a percent).

**WIP Version**
![image](https://cloud.githubusercontent.com/assets/4397508/24523552/67aa2bf4-1561-11e7-9fa5-b5c73e43bc00.png)

@mphprogrammer @saneshark

saneshark:
Lets address this as time allows, but we will drop this down to 10ths instead of 100ths. Low priority.

PR #5243, opened 2017-04-03
Round GIBCT outcome data to tenths instead of 100ths

bshyong:
Resolves https://github.com/department-of-veterans-affairs/vets-website/issues/5211


U-DON (approved):
LGTM.

Issue #5214, opened 2017-03-30
GI Bill CT: Resizing the screen makes data disappear


saintsoup52:
When the screen shrinks to mobile size certain data disappears to include:
* Types of complaints
* Post-9/11 GI Bill # & $$ 

![image](https://cloud.githubusercontent.com/assets/4397508/24524038/3719397e-1563-11e7-8206-c2ff6a3b0270.png)

@mphprogrammer @saneshark

saneshark:
@bshyong to verify if this is incorrectly marked up or if it is just a design nuance that @saintsoup52 identified as an issue incorrectly.

bshyong:
This is due to an error with the CSS breakpoints. There is a gap between 481 and 784px where nothing will show up.

saintsoup52:
Sorry to revisit this issue... but I just checked the staging sit using my iphone (chrome) and I was not able to see the types of complaints.

![image](https://cloud.githubusercontent.com/assets/4397508/24794492/de69c32e-1b53-11e7-9a63-d32b9efd4ca3.png)

@bshyong @saneshark

U-DON:
I think this is expected when there are 0 total complaints. It should only display the complaints that have a non-zero count. At least that's what the code is doing right now, so I thought that was as designed. @melwoodard Is that the expected behavior?

saintsoup52:
Thank you for flagging that change... It does appear to work when I search for schools with no complaints. 

Nice new feature.

melwoodard:
@U-DON that's correct

PR #5244, opened 2017-04-03
Fix error in GIBCT mobile views

bshyong:
Resolves https://github.com/department-of-veterans-affairs/vets-website/issues/5214

There was a gap in the breakpoints where nothing will show up between 481px and 784px screen sizes


U-DON (approved):
LGTM.

PR #5248, opened 2017-04-03
Move relinquishment to separate page, dependent on relationship

raquelromano-gov:
Closes https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1923

image

image

### DIFF
--- a/src/js/edu-benefits/5490/config/form.js
+++ b/src/js/edu-benefits/5490/config/form.js
@@ -148,6 +148,7 @@ const formConfig = {
           path: 'benefits/eligibility',
           initialData: {},
           uiSchema: {
+            'ui:title': 'Benefit selection',
             'view:benefitsDisclaimerChild': {
               'ui:description': benefitsDisclaimerChild,
               'ui:options': {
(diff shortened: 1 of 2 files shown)


src/js/edu-benefits/5490/config/form.js
```diff
@@ -208,7 +191,35 @@ const formConfig = {
...
-              benefit,
+              benefit
+            }
+          }
+        },
+        benefitRelinquishment: {
+          title: 'Benefits relinquishment',
+          path: 'benefits/relinquishment',
+          initialData: {},
+          depends: {
```
cvalarida: Clever!

test/edu-benefits/5490/config/benefitRelinquishment.unit.spec.jsx
```diff
@@ -0,0 +1,105 @@
...
+            benefitSelection: {
+              uiSchema: {
+                benefit: {
+                  'ui:options': {
+                  }
+                }
+              }
+            }
+          }}
+          formData={{ relationship: 'child' }}
```
cvalarida: I notice this isn't present in the other tests. Is it not needed? Perhaps more importantly, should it be needed?

cvalarida (approved):
The tests may behave a bit unexpectedly; it doesn't look like the `DefinitionTester` takes into consideration the `depends` option on the page. Other than that, it looks good. I'm approving it so it can go through; we can discuss whether it's a bug or a feature outside this PR.

test/edu-benefits/5490/config/benefitRelinquishment.unit.spec.jsx
raquelromano-gov: No, it isn't and shouldn't be needed, since the logic for whether or not to show this page is elsewhere...removed.

raquelromano-gov:
Yeah, this unit test doesn't check anything outside the page level, so I was thinking the e2e tests should verify that the page is skipped or not depending on the data.

Issue #5209, opened 2017-03-30
GI Bill CT: Buy-Up Program


saintsoup52:
In the current WIP version a user should not be able to input a value higher than $600 for their buy-in. 

The typical buy-in is $600 but it could be less. If a user were to input a value above $600 into the tool they would be told they were eligible to receive wildly higher monthly benefits under MGIB.

In the production version we use a drop down menu to enforce that behavior.

**Current version**
![image](https://cloud.githubusercontent.com/assets/4397508/24523129/0116495a-1560-11e7-8bb7-557f13515a1d.png)

**WIP version**
![image](https://cloud.githubusercontent.com/assets/4397508/24523221/5336b44a-1560-11e7-8b02-e00af2b60af2.png)

@mphprogrammer @saneshark

saneshark:
explain via text that max should be 600, but allow the user to input values exceeding this, but the calculation will reflect 600 regardless of value exceeding 600.

melwoodard:
Append the label for the buy-up amount field to be:

How much did you pay toward buy-up (up to $600)?

PR #5252, opened 2017-04-04
Added in-state tuition fees input to GIBCT calculator, modified buy up amount label, and corrected calculations accordingly.

### LINKED ISSUE #5208: GI Bill CT: Calculator Questions In-State & Out-of State not appearing
In the WIP version when a user selects a public school and says they are an out-of-state student the tuition/fees questions are not changing accordingly.

An out of state student should be asked two tuition questions not just one.

**Current production version:**
image

**WIP version:**
image

@mphprogrammer @saneshark

U-DON:
Resolves #5208, #5209.

If the user selects "no" for "in-state student", an input for in-state tuition fees will appear pre-populated with the value we have for the given school. It will default to this when it appears, even if it has been modified previously and then hidden.

> screen shot 2017-04-04 at 2 28 54 pm

Modal for the learn more link on the in-state tuition fees input is taken from [here](http://www.benefits.va.gov/gibill/comparison_tool/about_this_tool.asp#tuition_fees_instate_input). The modal in production seems incorrect, as it talks about retention rates instead, so hoping this is the right one.

> screen shot 2017-04-04 at 2 29 20 pm

If the user enters buy-up amount that exceeds 600, calculations will appear as if it were capped at 600. So a higher amount won't be reflected in the calculations.

> screen shot 2017-04-04 at 2 51 04 pm

### DIFF
--- a/src/js/gi/components/profile/CalculatorForm.jsx
+++ b/src/js/gi/components/profile/CalculatorForm.jsx
@@ -46,6 +46,24 @@ class CalculatorForm extends React.Component {
 
   renderTuition() {
     if (!this.props.displayedInputs.tuition) return null;
+
+    const inStateTuitionInput =
+      this.props.inputs.inState === 'no' && (
+        <div>
+          <label htmlFor="inStateTuitionFees">
+            {this.renderLearnMoreLabel({
+              text: 'In-state tuition and fees per year',
+              modal: 'calcInStateTuition'
+            })}
+          </label>
+          <input
+              type="text"
+              name="inStateTuitionFees"
+              value={formatCurrency(this.props.inputs.inStateTuitionFees)}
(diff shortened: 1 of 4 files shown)


melwoodard (commented):
Thanks @U-DON. I'm not able to test it through the heroku link, but the screenshots are spot-on and I think you found the correct modal content too - thanks for noticing that.

Where does the buy-up field appear? I'm having trouble finding it.

U-DON:
@melwoodard The buy-up field should appear when you choose the Montgomery GI Bill (Ch. 30) in the eligibility form.

melwoodard:
Got it, thanks!

melwoodard (approved):

bshyong (approved):
LGTM

PR #5254, opened 2017-04-04
Edu apply page

cvalarida:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1892

With 5490
edu-apply-animation-2

Without 5490
edu-apply-animation-3

image

The animation isn't quite the same as in the issue, but without tweaking the cubic-bezier functions manually and slowing them all down to 200+ ms, it's close...ish.

@goldenmeanie How close should I try to get it?


content/pages/education/apply.md
```diff
@@ -95,3 +113,29 @@ You must apply for education benefits using eBenefits if you're:
...
+    });
+
+  // Make the go button go to the right place
+  document.getElementById('apply-go-button')
+    .addEventListener('click', function () {
+      var selectedForm = document.querySelector('input[name="form-selection"]:checked');
+
+      console.log('selectedForm:', selectedForm);
+      if (selectedForm) {
+        location.assign('/education/apply-for-education-benefits/application/' + selectedForm.value + '/introduction');
```
cvalarida: This seems pretty contrived, so I'm open to suggestions here too. I did this because it was short, sweet, and to the point.

src/sass/_shame.scss
```diff
@@ -138,3 +138,54 @@ body .row {
...
+    background-image: url(/img/arrow-down-white.svg);
+    background-position: right bottom;
+    background-repeat: no-repeat;
+    background-size: 0.6em auto;
+    display: inline-block;
+    margin-left: 0.5em;
+    margin-right: -0.3em;
+    width: 1em;
+    height: 1em;
+    content: ''; // Add content for accessibility purposes?
```
annekainicUSDS: This raises a good question: I wonder what the accessibility of this interaction is?

content/pages/education/apply.md
```diff
@@ -95,3 +113,28 @@ You must apply for education benefits using eBenefits if you're:
 </div>
 
 <div markdown="0"><br></div>
+
+<script type="text/javascript">
+  // I'm open to suggestions on how to not do this here
```
annekainicUSDS: Create a separate js file and just import it here? That way it can also be used by other pages if this becomes an interaction we want to use elsewhere.

src/sass/_shame.scss
```diff
@@ -138,3 +138,54 @@ body .row {
...
+    text-indent: -9999em;
+  }
+}
+
+// Yuck...
+#apply-expander-content {
```
annekainicUSDS: I'm assuming eventually we'll also have the same button for "Manage", so not sure if we want to make this a bit more extensible (doesn't have to be for now, but might be something we keep in mind for a future refactor).

src/sass/_shame.scss
```diff
@@ -138,3 +138,54 @@ body .row {
...
+  max-height: 19em;
```
annekainicUSDS: Is this needed? When playing locally in DevTools on your branch, I just unchecked this style and things seemed to work fine without it.

annekainicUSDS (commented):
Looks great! A few minor comments, and it might be good at some point to have a test for this to ensure it expands and takes you to the right forms.

src/sass/_shame.scss
```diff
@@ -138,3 +138,54 @@ body .row {
 .csp-inline-patch-logout {
   padding: 2em 0;
 }
+
+// /education/apply/ uses this expander button for navigation
+// Where else can I put this? Also, it could use a better name.
```
cvalarida: edu-benefits.scss isn't used on `/education/apply`; that was the first thing I tried. The entry point for this page is `no-react.css`. I'm not sure how to load another css file from this markdown...

src/sass/_shame.scss
cvalarida: An excellent question! I'm not familiar with testing for accessibility if I'm honest. Do we just point a screen reader at it and...see what it's like?

src/sass/_shame.scss
cvalarida: I talked to @goldenmeanie about that and the hope is we'll have multiform up and running by the time we'd need to use this on the Manage button.

src/sass/_shame.scss
cvalarida: Yeah, it has trouble animating from `max-height: 0em;` to `max-height: auto` or `initial` or anything not a number. I've noticed trouble with animation in the dev tools. It'd work without that line, but it won't animate.

Unless I'm missing something. I did try it without that in there first, though.

content/pages/education/apply.md
cvalarida: I wanted to do that, but where do I put that file? I'd love to take it out!

src/sass/_shame.scss
annekainicUSDS: Ah gotcha. Might be necessary then.

src/sass/_shame.scss
annekainicUSDS: I think it might be a combination of that and running an automated axe test on it. We're talking more about how to run our 508 tests on Friday so I think we can address this then.

src/sass/_shame.scss
annekainicUSDS: Ah gotcha. What if you created a separate module and then loaded it in no-react.scss?

cvalarida: Not a bad idea. How can I get that to be loaded by this file specifically? Just make another case in the parent layout, `page-breadcrumbs.html` to include an `include` (from `content/includes` which has a `<link>` to the css? (And actually, I guess that might make it the other way around; it'll do no-react.css by default and include this other little css file.)

The other consideration is the build process. Do we need to build another file and include it in only this? Seems a bit...excessive to me.

At any rate, this is only a temporary solution until we get multiform up and running, so maybe we do just leave it here in `_shame.scss` and if this is used elsewhere in the future, it can be addressed then?

annekainicUSDS:
Just noticed one more thing you'll want to change: the yellow box below the Apply benefits with the link to eBenefits, we'll want to take out the 1990e and 5490 from that list.

annekainicUSDS:
One more note: It seems likely that we will not be launching the 5490 today, so if we could comment out the 5490 radio buttons for now and keep it in the yellow box to redirect to eBenefits, that'd be great :).

content/pages/education/apply.md
```diff
@@ -56,10 +76,9 @@ If you need to make a change (for example, you’re moving to a new school), man
 <div class="usa-alert usa-alert-warning usa-content va-alert" markdown="1">
 	<div class="usa-alert-body">
 
-##### Forms 22-1990E, 22-1990N, 22-5490, and 22-5495
+##### Forms 22-1990N, 22-5490, and 22-5495
 
 You must apply for education benefits using eBenefits if you're:
-- A spouse or dependent
```
cvalarida: @J-Quag Is this right to get rid of? The 1990e covers this option, so they don't have to go to eBenefits, right?

src/sass/_shame.scss
annekainicUSDS: If this is something other teams might use at some point, we don't want to limit it to just this page anyway, right? But I agree with the last point, that we can probably leave it in _shame.scss for now. But I would suggest creating a ticket for us to share this UI pattern with the design team and ask if they would find this useful in other places. If so, we can then figure out how to make this extensible at that point. Just so we don't forget to come back to that :)

content/pages/education/apply.md
annekainicUSDS: The only thing I think we want to be careful about is that even though the 5490 is for "survivors", which is the bullet below, they may see themselves as "spouses and dependents" still and be confused about where to go if we remove that line. So maybe this is something both @J-Quag and @goldenmeanie can weigh in on; perhaps we need more detail here for now?

src/sass/_shame.scss
cvalarida: The issue has been created at https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1988

content/pages/education/apply.md
goldenmeanie: - A dependent whose sponsor is permanently and totally disabled, making a new application or moving a benefit to a new location
- A dependent whose sponsor is deceased, MIA, or a POW, making a new application or moving a benefit to a new location
- A candidate for the National Call to Service program

J-Quag: The third bullet doesn't make sense. 
    > Dependents applying where your sponsor is permanently and totally disabled their benefits to you...

annekainicUSDS:
One other thing @cvalarida: in the top level nav, when looking at Explore Benefits > Education & training, the "Apply" button still goes to the 1990. Should probably go to the /education/apply page now, right?

content/pages/education/apply.md
cvalarida: @J-Quag Oi gevalt...That's what I get for just replacing part of the copy. 😖  I'll fix it.
@goldenmeanie I'll Use that copy for the bullets.

Thanks!

J-Quag: I had to screenshot it quickly with the animation hah. I kept trying to read it while it was changing. 
What are the bullets for?

cvalarida: I updated the initial comment to reflect the changes. The bullet points are for the yellow box (at least I hope so).

content/includes/veteran-navigation.html
```diff
@@ -70,7 +70,7 @@
                 <li><a href="/education/apply/">Application process</a></li>
                 <li><a href="/education/gi-bill/">GI Bill programs</a></li>
                 <li><a href="/gi-bill-comparison-tool/">Compare GI Bill benefits</a></li>
-                <li><a href="/education/apply-for-education-benefits/application/1990" class="usa-button va-button-primary">Apply Now</a></li>
+                <li><a href="/education/apply/" class="usa-button va-button-primary">Apply Now</a></li>
```
cvalarida: Now both the `Application process` link and the `Apply Now` button go to apply page. If that's what we want for clarity, great, otherwise we can get rid of one.

@goldenmeanie Is that a question for you?

annekainicUSDS: I think that's what we want for now. This will change once we have multi-form.

annekainicUSDS (approved):
LGTM! :shipit:

PR #5256, opened 2017-04-04
Remove caution flag filters from GIBCT

bshyong:
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1919 for front-end

### DIFF
--- a/src/js/gi/containers/SearchPage.jsx
+++ b/src/js/gi/containers/SearchPage.jsx
@@ -60,7 +60,6 @@ export class SearchPage extends React.Component {
 
   updateSearchResults() {
     const programFilters = [
-      'caution',
       'studentVeteranGroup',
       'yellowRibbonScholarship',
       'principlesOfExcellence',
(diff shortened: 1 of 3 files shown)


src/js/gi/containers/SearchPage.jsx
```diff
@@ -111,8 +109,6 @@ export class SearchPage extends React.Component {
     delete query.page;
 
     const shouldRemoveFilter =
-      (field !== 'caution' && !queryValue) ||
-      (field === 'caution' && queryValue) ||
```
U-DON: There should still be a condition here for the checkbox true/false values, I think.

`const shouldRemoveFilter = !value || (...)`

src/js/gi/containers/SearchPage.jsx
```diff
@@ -101,7 +99,7 @@ export class SearchPage extends React.Component {
 
   handleFilterChange(field, value) {
     // Translate form selections to query params.
-    const queryValue = field === 'caution' ? !value : value;
+    const queryValue = value;
```
U-DON: I think we can just use `value` instead of assigning to `queryValue`; the only reason for reassigning was to conditionally set a value.

U-DON (commented):
Generally LGTM. Had a couple of comments. I think unchecking a filter should remove it from the URL instead of keeping it in with false.

Unrelated, but it seems there's a build failure around a missing dependency?

U-DON (approved):

PR #5257, opened 2017-04-04
Use Yarn as a dependency management tool

markolson:
This is a proposal to switch from npm to [Yarn](https://yarnpkg.com) as our dependency management tool, with a few other necessary fixes thrown in.

I spent some time this week wrangling with shrinkwrap, trying to get a package version bumped with minimal changes. Besides taking far too long (I ended up making the change by hand), I found a few other issues through the process: The big one for me is that it seems like shrinkwrapping produces unexpected or inconsistent results, and allows situations like where the USWDS package was pinned to `^1.0.0` in the `package.json` file, but was `0.10.0` in the `npm-shrinkwrap.json` file, even after multiple installs and shrinkwraps from multiple developers since the bump in `package.json`.  I think that 3 people ran `npm shrinkwrap` this Tuesday and got 3 different results, _all_ of which had the "wrong" version of USWDS in them. 

Yarn produces consistent install from it's lock file, and lays the packages down in the same `node_modules` paths that `npm` does, so all the `npm run` commands we use will stay the same; only `yarn` would be used to add (`yarn add pkg_name --dev`), upgrade (`yarn upgrade pkg_name@version`), and resolve (`yarn install`) packages. It also apparently does more checks than `npm install`, as it forced me to change the node version to be >= 6 because of a minimum version one of the libraries we use declares.

I understand there would need to be further changes needed around the deploy process and version bumps, but I wanted to do a quick spike and see how much effort it would be to change vets-website over to yarn.


README.md
```diff
@@ -14,7 +14,7 @@ very secret.
 
 | I want to... | Then you should... |
 | ------------ | ------------------ |
-| clone the site | `git clone https://github.com/department-of-veterans-affairs/vets-website.git` followed by `cd vets-website`, `npm install`. Run `npm install` anytime `package.json` changes. |
+| clone the site | `git clone https://github.com/department-of-veterans-affairs/vets-website.git` followed by `cd vets-website`, then follow the insturctions below to install node, npm and yarn if needed. Finally, run `yarn install` to fetch all the dependencies. Run `yarn install` anytime `package.json` changes. |
```
aub: instructions misspelled

aub:
This seems great conceptually. Did any issues arise from the update of node and npm versions? Do you have any concept of the LOE for getting that to work or making the required deploy-related changes?

markolson:
So all the the tests (save the currently failing 508 ones) pass, which ideally that means that things work just fine with the new node/npm versions if our tests are good. There's an open devops ticket to bump to those versions in the dev/staging environments as well, so this is something the core team is working towards anyways. For LOE, I have no idea! Hopefully with that devops change to change the version in dev/staging, it's easy to change in prod, but ultimately I couldn't say.

aub:
Ok, that sounds fine. @markolson do you have a handle on coordinating all of the things that need to happen in order to make this real? Do you want to manage that process or pass it off to somebody else to see it through? I'd be happy to help or take that over if you want.

@robbiethegeek I know you were looking at something similar so want to make sure we're not duplicating effort here. It looks like @markolson has both the node/npm upgrade and the port to Yarn all set in this PR.

ayaleloehr:
just here to tag the other @robbiethegeek-usds account so he is sure to see the question directed at him ^.

markolson:
@aub I'm happy to shepherd this through, though it's going to be a lower priority for me. Right now there are some accessibility tests failing on Jenkins and locally with `Capabilitie undefined undefined`, which ... is not a terribly helpful error message, and I'm not sure how much progress I'll be able to make.

robbiethegeek-usds:
I will work with @markolson & @CyberKoz to coordinate if we move forward with this.  This is also a lower priority for me, so will keep checking in when I have the time.

Dockerfile
```diff
@@ -72,8 +74,9 @@ RUN mkdir -p logs/selenium
 # Install required npm dependencies
 
 COPY package.json .
-COPY npm-shrinkwrap.json .
-RUN npm install
+COPY yarn.lock .
+# skips all dev dependencies if NODE_ENV=production.. so..
+RUN yarn install --production=false
```
jbalboni: Is this actually what we want? We don't really make a distinction between dev and prod dependencies in package.json right now (the three things in dependencies should really be in devDependencies). They're all necessary to build.

markolson: As a stopgap, we need this in order for yarn to install all the `devDependencies`, which is most of them. Ideally we'd sort out what needs to be in `devDependencies` vs `dependencies`.

bshyong:
@markolson I ran into the errors you saw with `Capabilitie undefined undefined`.

There were two things I did to resolve the issues:
1. Upgrade ChromeDriver to 2.29 (any version after 2.28 would work)
2. add `client.timeoutsAsyncScript(2000)`. This is only needed if after updating ChromeDriver, you encounter errors about timeouts

you can tail `logs/selenium/selenium-debug.log` for more context around the errors

jbalboni:
@bshyong Where did you add `client.timeoutsAsyncScript(2000)`? I get the same error even after updating chromedriver.

jbalboni:
I think I fixed with the last commit. There were pages in rx/messaging that weren't loading, because the were routes controlled by react-router and they weren't in `src/test-support/test-server.js`.

markolson:
Awesome! With the tests passing, I think we can move onto making sure the documentation in the README is acceptable, and then ensuring that we have a working deploy process for when this is merged.

jbalboni:
I think we should switch the commands in package.json to use yarn instead of npm. The yarn output is better (i.e. there's less nonsense printed out when a command fails) and it should will still work if someone uses npm instead.

bshyong:
@jbalboni GIBCT had a timeout issue that was resolved when I added the `timeoutsAsyncScript` to the top of the test

jbalboni:
Feedback from engineering meeting was to try to separate out any `npm run` and node version changes and just merge the dependency management changes for the first pass.

aub:
Nice, big win here! Thanks @markolson and @jbalboni!

PR #5260, opened 2017-04-05
Remove route from not production list to general list

annekainicUSDS:
Unblocks the 1990e for launch. Should only be merged once we are prepared to launch!


raquelromano-gov (commented):
Can you also enable the 1990e e2e tests? Just take out the guards for buildtype.

cvalarida (approved):

raquelromano-gov (approved):

PR #5261, opened 2017-04-05
Use sessionStorage for first_name to prevent flicker

omgitsbillryan:
**Issue**
https://github.com/department-of-veterans-affairs/vets.gov-team/issues/622
image

**Root Cause**
The response to `GET /v0/user` updates this field with `response.user.first_name`.  

**Solution**

1. Use `sessionStorage` to store the `first_name`
1. When rendering, use the following preference order: `jsonResponse.user.first_name` -> `sessionStorage.userFirstName` -> `jsonResponse.user.email`

### DIFF
--- a/src/js/login/components/SearchHelpSignIn.jsx
+++ b/src/js/login/components/SearchHelpSignIn.jsx
@@ -11,14 +11,13 @@ import { updateLoggedInStatus, toggleSearchHelpUserMenu } from '../actions';
 class SearchHelpSignIn extends React.Component {
   render() {
     let content;
-    let greeting;
     const login = this.props.login;
 
     if (login.currentlyLoggedIn) {
       const firstName = _.startCase(_.toLower(
         this.props.profile.userFullName.first || sessionStorage.userFirstName
       ));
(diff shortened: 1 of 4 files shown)


src/js/login/components/SearchHelpSignIn.jsx
```diff
@@ -15,12 +15,10 @@ class SearchHelpSignIn extends React.Component {
...
-      if (this.props.profile.userFullName.first) {
-        const firstName = _.startCase(_.toLower(this.props.profile.userFullName.first));
-        greeting = firstName;
-      } else {
-        greeting = this.props.profile.email;
-      }
+      const firstName = _.startCase(_.toLower(
+        this.props.profile.userFullName.first || sessionStorage.userFirstName
+      ));
+      greeting = firstName || this.props.profile.email;
```
U-DON: Since `greeting` is only used in the scope of this conditional, it can just be declared here as `const greeting = ...`.

U-DON (approved):
LGTM with minor comment.

PR #5262, opened 2017-04-05
Add visual focus indicator for IE on FL

bshyong:
resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1958


src/sass/facility-locator.scss
```diff
@@ -71,6 +71,12 @@
       color: $color-link-default;
       text-decoration: none;
 
+      &:focus {
+        h5 {
+          border: $color-link-default solid thin;
```
U-DON: It's usually the `outline` property that's used for focus, right? Does that not work for IE?

U-DON (approved):
LGTM.

ayaleloehr:
@bshyong can you check these tests then this should be good to get in it looks like?

PR #5267, opened 2017-04-05
Added heading-2 through heading-6 mixins.

ghost:
Closes department-of-veterans-affairs/vets.gov-team#760

### DIFF
--- a/src/sass/base/_b-variables.scss
+++ b/src/sass/base/_b-variables.scss
@@ -1,4 +1,6 @@
+//===================================
 // Typography
+//===================================
 
 // These variables override values set in either
 // ~uswds/src/stylesheets/core/defaults" or 
@@ -7,7 +9,42 @@
 $base-font-size:      rem(16px);
 $font-serif:          "Roboto Slab", serif;
 
+// $heading-1:
+@mixin heading-2 {
+  font-size: 1.2em;
+  line-height: 1.2;
+  font-weight: bold;
+}
+
+@mixin heading-3 {
+  font-size: 1.8em;
+  line-height: 1.5;
+  font-weight: bold;
+}
+
+@mixin heading-4 {
+  font-size: 1.35em;
+  line-height: 1.5;
+  font-weight: bold;
+}
+
+@mixin heading-5 {
+  font-size: 1.15em;
+  line-height: 1.5;
+  font-weight: bold;
+}
+
+@mixin heading-6 {
+  font-size: 0.9em;
+  line-height: 1.5;
+  font-weight: bold;
+}
+
+
+//===================================
 // Magic Numbers
+//===================================
+
 // Note: USWDS value is 53rem;
 $text-max-width:      70rem;
 $usa-form-width:      32em;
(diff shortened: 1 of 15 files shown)


jcmeloni-usds:
LGTM

src/sass/base/_va.scss
```diff
@@ -289,45 +289,38 @@ hr {
...
   padding: 0 0 .5em 0;
 }
 
 h2 {
-  line-height: 1.2em;
-}
-
-h1:first-child {
-  margin-top: 0;
+  @include heading-2;
```
omgitsbillryan: Shouldn't this be `heading-1`?

src/sass/base/_b-variables.scss
```diff
@@ -7,7 +9,42 @@
...
+@mixin heading-5 {
+  font-size: 1.15em;
+  line-height: 1.5;
+  font-weight: bold;
+}
+
+@mixin heading-6 {
+  font-size: 0.9em;
+  line-height: 1.5;
+  font-weight: bold;
```
omgitsbillryan: Can `font-weight: bold;` be shared to avoid repetition?

src/sass/base/_b-variables.scss
```diff
@@ -7,7 +9,42 @@
 $base-font-size:      rem(16px);
 $font-serif:          'Roboto Slab', serif;
 
+// $heading-1:
+@mixin heading-2 {
+  font-size: 1.2em;
```
omgitsbillryan: Excuse my ignorance - shouldn't `heading-2` have a larger font size than 3?

jcmeloni-usds:
@tiffanybbrown-va and @melwoodard are there shenanigans here re: presentation vs semantics or is @omgitsbillryan on to something?

melwoodard:
I'd need to look closer at specific places we're using H2s, but there is some weirdness between H1/H2/H3 due to the fact that H2s have been used really sporadically. All of our content pages go right from H1 to H3, skipping H2 (not exactly sure why), so H2 probably only shows up in the apps and is being treated inconsistently.

jcmeloni-usds:
Sounds like shenanigans to me :)

tiffanybbrown-va:
Some additional context:

`h1` was originally reserved for the logo. At some point, this changed, and we started using `h1` throughout the content. Our `h1` and `h2` elements are styled the same. That's why there's a `heading-2` mixin, but no `heading-1` mixin. 

The mixin names and styles are pulled directly from whatever element name they were originally attached to. This is why the font size for `heading-3` is larger than that for `heading-2`. Made sense to me at the time. It apparently does not make sense to anyone else.

jcmeloni-usds:
I think it makes sense to me enough that I recognized it immediately as probably shenanigans. :)  If this all _works_ I'm apt to merge it in with a comment somewhere that mixin names and styles are pulled directly from whatever element name they were originally attached to, and someone should maybe take a look at that bigger issue someday...

melwoodard:
Agreed - we should put a pin in the fact that we're skipping a heading level (weird experience for screenreaders) and revisit it separately to get our headings right across the board.

tiffanybbrown-va:
Not exactly sure why this test is failing, since I don't think it has anything to do with my changes to this PR 🤔

jcmeloni-usds:
All green now! Let's merge it.

PR #5271, opened 2017-04-06
Fix modal issues in GIBCT

bshyong:
Resolves Modal issues in https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1507

Modals that exceed the height of the viewport now scrollable. 
When modals are too long for the viewport:
- The modal starts at the top of the viewport instead of being vertically centered
- when modal is scrolled, the page and the dark overlay should remain fixed


src/sass/gi/partials/_gi-modal.scss
```diff
@@ -1,12 +1,21 @@
 .gi-app {
-  
+
   .va-modal {
     background-color: rgba($color-gray-dark, .8);
     padding: 10px;
 
+    @include media-maxwidth($small-screen) {
+      overflow-y: scroll;
+    }
```
U-DON: Should these be specific to GIBCT? I would think this issue can affect modals in general.

U-DON (approved):
LGTM with one comment. I would loop in perhaps @tiffanybbrown-va and @jbalboni to review as it affects modal behavior in general.

src/sass/gi/partials/_gi-modal.scss
bshyong: @U-DON I only enabled this for GIBCT as to not affect behavior in other apps, however I don't foresee any problems with applying this patch to all modals across vets-website

@jbalboni @tiffanybbrown-va thoughts on this PR? This PR will fix background-scrolling issues on all modals across vets-website

src/js/common/components/Modal.jsx
```diff
@@ -26,9 +26,11 @@ class Modal extends React.Component {
   componentWillReceiveProps(newProps) {
     if (newProps.visible && !this.props.visible) {
       this.setState({ lastFocus: document.activeElement, focusListener: focusListener(newProps.focusSelector) });
+      document.body.classList.add('modal-open');
```
jbalboni: I wonder if these should be in a different lifecycle method. Generally I've seen that people keep DOM operations in componentWillUpdate/componentDidUpdate/componentDidMount (though I bent that rule for the event listener code).

This method also doesn't run when a component is initially rendered, so if you render a page that has a modal open to start with, I don't think `modal-open` will be added to the body.

src/sass/gi/partials/_gi-modal.scss
jbalboni: I don't see a reason why we wouldn't want this everywhere.

jbalboni (approved):
Seems like a good change to me.

src/js/common/components/Modal.jsx
bshyong: Good call—`componentDidMount` would be a good place for adding the `modal-open` class. Removing the class could still work in this lifecycle method

src/sass/gi/partials/_gi-modal.scss
```diff
@@ -6,6 +6,11 @@
     .va-modal-inner {
       max-width: 600px;
 
+      @include media-maxwidth($small-screen) {
+        transform: none;
+        top: 0;
```
U-DON: Wouldn't you also want this in the main modal styles?

PR #5272, opened 2017-04-06
Allow e2e tests to run in prod

annekainicUSDS:


raquelromano-gov:
I think we should ask someone from devops to help out here. Previously, I had assumed they passed locally and failed on Jenkins because of the production buildtype guards in our route defintions, but now that they're live, I don't understand why the form can't be loaded.

raquelromano-gov:
See @jabalboni's comments in https://github.com/department-of-veterans-affairs/vets-website/pull/5280

> We need to make an update to src/test-support/test-server.js. Line 34 should be:
> 
> `app.use('/education/apply-for-education-benefits/application', fallback('/education/apply-for-education-benefits/application/index.html', { root }));`
> 
> That removes the 1995 from the url so all urls are routed correctly

PR #5275, opened 2017-04-06
Add thousands comma for GIBCT search results

bshyong:
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2002


U-DON (approved):
LGTM. You can also refactor the `formatCurrency` helper to use `formatNumber`. It can just be:

```
return `$${formatNumber(Math.round(+value)}`;
```

bshyong:
Good call; I will do that to make it cleaner

PR #5276, opened 2017-04-06
Hack attempt to fix login e2e test in jenkins

omgitsbillryan:


ayaleloehr:
@omgitsbillryan who should review this? and there is a merge conflict.

omgitsbillryan:
@ayaleloehr thanks for the reminder.  I'm going to close this PR as unmerged.  This test had failed unexpectedly for a few ppl and it was suspected that this test had a timing issue that caused it to fail intermittently.  Since then I have not been able to reproduce this failure nor have I seen anyone else's build fail on this test.  I don't think there was anything wrong with the original test and something else must have fixed the issue.

PR #5277, opened 2017-04-07
Fix hidden data removal in schemaform

jbalboni:
There are two underlying problems here:

One is that `removeHiddenData` doesn’t actually remove hidden data if it’s nested, because it’s not checking to see if the returned object has changed after each recursive call.

The second issue is that in `setHiddenFields`, `ui:collapsed` is not correctly reset for nested data because the `expandUnder` field is relative to the object it’s in, but the `formData` object passed in is for the whole form.

This should fix both of those.

It also revealed another issue, which is that for arrays, we have a single `items` schema for all the items in the array, but each item could have different data that shows/hides fields. If we modify the items schema, it applies to all rows, which is probably wrong.


src/js/common/schemaform/helpers.js
```diff
@@ -378,11 +378,13 @@ export const pureWithDeepEquals = shouldUpdate((props, nextProps) => {
  * hideIf function from uiSchema and the current page data. Sets 'ui:hidden'
  * which is a non-standard JSON Schema property
```
raquelromano-gov: Add a comment explaining what the optional path argument is?

src/js/common/schemaform/helpers.js
```diff
@@ -420,7 +422,9 @@ export function setHiddenFields(schema, uiSchema, formData) {
   }
 
   if (updatedSchema.type === 'array') {
-    const newSchema = setHiddenFields(updatedSchema.items, uiSchema.items, formData);
+    // this isn't really correct; since all array items share a schema, we can't hide and remove data specific
```
raquelromano-gov: This comment is pretty cryptic...if it's not correct, what can go wrong?

raquelromano-gov (commented):
Can you also re-order the checkboxes in `previousBenefits` to put `view:noPreviousBenefits` at the end, just before `other`? (https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1808)

raquelromano-gov:
Then this can close both https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1808 and  department-of-veterans-affairs/vets.gov-team#1926

src/js/common/schemaform/helpers.js
```diff
@@ -445,25 +449,25 @@ export function removeHiddenData(schema, data) {
       if (typeof data[next] !== 'undefined') {
         const nextData = removeHiddenData(schema.properties[next], data[next]);
 
+        // if the data was removed, then just unset it
         if (typeof nextData === 'undefined') {
           return _.unset(next, current);
         }
+
+        // if data was updated (like a nested prop was removed), update it
+        if (nextData !== data[next]) {
```
raquelromano-gov: Trying to figure out why this isn't also needed above if the schema is an object with nested properties...I think it's because you're passing along the path, right?

jbalboni: @raquelromano-gov Above where?

cvalarida (approved):
Yeah, we'll probably need to figure out a long-term solution for arrays, but the rest of it looks good.

src/js/common/schemaform/helpers.js
```diff
@@ -420,7 +427,11 @@ export function setHiddenFields(schema, uiSchema, formData) {
   }
 
   if (updatedSchema.type === 'array') {
-    const newSchema = setHiddenFields(updatedSchema.items, uiSchema.items, formData);
+    // Our current approach is to sometimes modify the schema based on form data
+    // This breaks down with arrays, because all items share the same schema
+    // So, this is technically not correct, but the last row is probably what's visible
+    // for a user, so it should be the least bad for now
+    const newSchema = setHiddenFields(updatedSchema.items, uiSchema.items, formData, path.concat(formData.length - 1));
```
annekainicUSDS: So based on your comment, this will only set hidden fields for the last item in an array? An arrays are present for things like schools or employment periods where they could one more than one?

src/js/common/schemaform/helpers.js
```diff
@@ -405,7 +412,7 @@ export function setHiddenFields(schema, uiSchema, formData) {
 
   if (updatedSchema.type === 'object') {
     const newProperties = Object.keys(updatedSchema.properties).reduce((current, next) => {
-      const newSchema = setHiddenFields(updatedSchema.properties[next], uiSchema[next], formData);
+      const newSchema = setHiddenFields(updatedSchema.properties[next], uiSchema[next], formData, path.concat(next));
```
annekainicUSDS: What is "next" in this case?

annekainicUSDS (approved):
Tried to understand this as best as I could; added a couple of questions to try to understand better what's happening. Also ran this locally and used React DevTools to see what data gets stored when the conditional question those `hidden` fields hinge on is changed, and this change does remove the data. The only consideration is something @cvalarida talked about before, which is: is there value is maintaining that data until the end and clearing it just before submit? If we imagine a situation where the veteran accidentally collapsed some fields, or just decided that they wanted to expand them again, they would have to re-enter their data. Not the most pressing concern, and something we could solve for in a future PR so that we can at least get this first fix out, but just something to keep in mind. Otherwise this seems to solve the problem, so LGTM!

src/js/common/schemaform/helpers.js
raquelromano-gov: In the case when `updatedSchema.type === 'object'. I think it's because the recursive call, along with the path argument, ensures that the traversal makes it down to the leaves of any nested properties, so for objects we can just check whether `nextData === 'undefined'` and un-set it if so. LMK if that's accurate.

src/js/common/schemaform/helpers.js
jbalboni: Correct. I don't think we have hidden fields inside of arrays right now, so I don't think it was causing any problems. But we do need to fix it at some point.

src/js/common/schemaform/helpers.js
jbalboni: The next property in the schema we're going to update.

src/js/common/schemaform/helpers.js
jbalboni: This code is in the `updatedSchema.type === 'object'` case. The diff is making it look like it isn't, I think.

annekainicUSDS (approved):

PR #5279, opened 2017-04-07
5490 add expect to graduate field

annekainicUSDS:
Currently, for the question on high school status, if they enter "graduated", a date field appears for the date they earned their diploma. The edu team requested that we also reveal a date field if they enter "expect to graduated" that will collect their future graduation date. This field needs to be validated for only being in the future.

The way I handled the validation and changing the title seems pretty hacky to me, and not particularly extensible, but not sure if there is a better way to do this.

Closes https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1944

screen shot 2017-04-07 at 2 51 27 pm


src/js/edu-benefits/5490/config/form.js
```diff
@@ -498,11 +502,25 @@ const formConfig = {
...
-                    expandUnder: 'status'
-                  }
-                }
+                    expandUnder: 'status',
+                    updateSchema: (pageData, form, schema) => {
+                      const status = _.get('educationHistory.data.highSchool.status', form);
+                      const field = _.get('educationHistory.uiSchema.highSchool.view:highSchoolOrGedCompletionDate', form);
+
+                      if (status === 'graduationExpected') {
+                        field['ui:title'] = 'When do you expect to earn your high school diploma?';
```
jbalboni: I think you can do this instead for updateSchema:

```js
  const status = _.get('educationHistory.data.highSchool.status', form);

  if (status === 'graduationExpected') {
    return { title: 'When do you expect to earn your high school diploma?' };
  }
  
  return { title: 'When did you earn your high school diploma?' };
```

You'll have to set the first argument to date.uiSchema to `null`.

src/js/edu-benefits/5490/config/form.js
```diff
@@ -498,11 +502,25 @@ const formConfig = {
                   'ui:options': {
                     hideIf: form => {
                       const status = _.get('highSchool.status', form);
-                      return status !== 'graduated';
+                      return status !== 'graduated' && status !== 'graduationExpected';
                     },
-                    expandUnder: 'status'
-                  }
-                }
+                    expandUnder: 'status',
```
jbalboni: We probably need to add a way to use expandUnder for specific values, instead of just checking if it's truthy. That would make this work better.

jbalboni (approved):
Made some comments, but mostly looks good.

src/js/edu-benefits/5490/config/form.js
cvalarida: Could check to see if `expandUnder` is a string (do the truthy thing like we are now) or a function (pass params to it like `hideIf`...or `updateSchema`, maybe...and then check if the result is truthy).

src/js/common/utils/validations.js
```diff
@@ -179,6 +179,11 @@ function isValidCurrentOrPastDate(day, month, year) {
   return momentDate.isSameOrBefore(moment().endOf('day'), 'day');
 }
 
+function isValidFutureDate(day, month, year) {
```
annekainicUSDS: Yeah, I looked everywhere, but couldn't find anything.

src/js/edu-benefits/5490/config/form.js
annekainicUSDS: Not sure I understand what the issue is here? Aren't we currently checking if `expandUnder` is a string? What about it now isn't working for what we're trying to do?

annekainicUSDS:
@jbalboni made fixes from your comments

src/js/edu-benefits/5490/config/form.js
jbalboni: The way it's written right now, there's no animation when the fields are shown, I think because of how `hideIf` and `expandUnder` interact. Being able to be more specific about when fields expand underneath the status might help (e.g. have something like `expandUnderValue: 'graduationExpected'`).

I think it's fine for this PR, but we should create an issue to work on it later.

jbalboni (approved):

src/js/edu-benefits/5490/config/form.js
annekainicUSDS: Ah okay, I'll create a ticket for us to work on that separately.

PR #5280, opened 2017-04-07
5490 e2e tests

cvalarida:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1637

I'm not convinced the post high school trainings are being added properly to the education history page, but everything runs, so I'm throwing this up in a PR for review!


test/e2e/edu-helpers.js
```diff
@@ -188,19 +215,13 @@ function completeSchoolSelection(client, data, onlyRequiredFields) {
   selectDropdown(client, 'root_educationProgram_educationType', data.educationProgram.educationType);
   if (!onlyRequiredFields) {
     client
-      .clearValue('input[name="root_educationProgram_name"]')
-      .clearValue('input[name="root_educationProgram_name"]', data.educationProgram.name)
```
cvalarida: This was erroring out, so I replaced all the clears and sets with `fill`.

test/edu-benefits/5490/e2e-test.js
```diff
@@ -0,0 +1,162 @@
...
+const EduHelpers = require('../../e2e/edu-helpers');
+const Edu5490Helpers = require('../../e2e/edu-5490-helpers');
+const testData = require('./schema/maximal-test.json');
+
+
+module.exports = (client, onlyRequiredFields) => {
+  // Cut out early if we're in production.
+  // Remove this once the 5490 is launched.
+  if (global.__BUILDTYPE__ === 'production') {
+    return;
```
cvalarida: This doesn't seem to be working.

@raquelromano-gov Is there anything else I need to do for this safeguard to function?

test/edu-benefits/5490/schema/maximal-test.json
```diff
@@ -15,7 +15,11 @@
   },
   "benefitSelection": {
     "data": {
-      "benefit": "chapter35",
+      "benefit": "chapter35"
+    }
+  },
+  "benefitRelinquishment": {
```
cvalarida: These pages are now split up.

test/edu-benefits/5490/00-required.e2e.spec.js
```diff
@@ -0,0 +1,6 @@
+const E2eHelpers = require('../../e2e/e2e-helpers');
+const runTest = require('./e2e-test.js');
+
+module.exports = E2eHelpers.createE2eTest(
+  (client) => runTest(client, true)
```
cvalarida: Doing it this way to keep it DRY.

test/e2e/edu-5490-helpers.js
```diff
@@ -0,0 +1,150 @@
...
+    client.selectDropdown('root_highSchool_status', data.highSchool.status);
+    if (completionDate) {
+      client.fillDate('root_highSchool_view:highSchoolOrGedCompletionDate', completionDate);
+    }
+
+    if (!_.isEmpty(data.postHighSchoolTrainings)) {
+      // Open up the trainings section if there are trainings in the data
+      client
+        .click('input[name="root_view:hasTrainingsYes"]')
+        .waitForElementVisible('input[name="root_postHighSchoolTrainings_0_name"]', Timeouts.normal);
```
cvalarida: The expandable fields seem to have this problem, but I don't know that it's consistent...Perhaps they don't need this sometimes because in some places nightwatch fills out other information while this is expanding?

test/e2e/e2e-helpers.js
```diff
@@ -87,6 +87,81 @@ function overrideAnimations(client) {
   });
 }
 
+/**
+ * Set up some helper functions to make writing tests nicer
+ */
+function bootstrapHelpers(client) {
```
cvalarida: Ideally, these boostrapped helpers would be in the `client`'s `prototype`, I think, but at this point, I think it's neither here nor there with the level of control we have.

test/e2e/edu-5490-helpers.js
```diff
@@ -0,0 +1,150 @@
+import _ from 'lodash/fp';
+const Timeouts = require('./timeouts.js');
+
+
+// Same as 1990e
+// TODO: relocate to keep it DRY
```
raquelromano-gov: Any reason not to move the identical function in `test/e2e/edu-1990e-helpers.js` into `test/e2e/edu-helpers.js` now instead of later?

test/e2e/edu-5490-helpers.js
```diff
@@ -0,0 +1,150 @@
...
+function completeEmploymentHistory(client, data, onlyRequiredFields) {
+  if (!onlyRequiredFields) {
+    const nonMilitaryJobs = data.nonMilitaryJobs[0];
+    client
+      .click('input[name="root_view:hasNonMilitaryJobsYes"]')
+      .clearValue('input[name="root_nonMilitaryJobs_0_name"]')
```
raquelromano-gov: Wondering why you're not using the new bootstrapped `client.fill()` method you wrote.

test/e2e/edu-5490-helpers.js
```diff
@@ -0,0 +1,150 @@
...
+      .clearValue('input[name="root_nonMilitaryJobs_0_months"]')
+      .setValue('input[name="root_nonMilitaryJobs_0_months"]', nonMilitaryJobs.months)
+      .clearValue('input[name="root_nonMilitaryJobs_0_licenseOrRating"]')
+      .setValue('input[name="root_nonMilitaryJobs_0_licenseOrRating"]', nonMilitaryJobs.licenseOrRating);
+  }
+}
+
+// This section is required, unlike the regular EduHelpers
+// Also actually uses the data
+function completeBenefitSelection(client, data) {
```
raquelromano-gov: You can still call the `completeBenefitsSelection()` function in `edu-helpers.js` (add a data parameter to it if you like), and set `onlyRequiredFields` to `false`

This seems to be how most of these shared functions were set up, and I think it's cleaner than having lots of form-specific helper functions that duplicate that code.

test/e2e/edu-5490-helpers.js
```diff
@@ -0,0 +1,150 @@
...
+    const activeDuty = `root_currentlyActiveDuty${data.currentlyActiveDuty ? 'Yes' : 'No'}`;
+    const felony = `root_outstandingFelony${data.outstandingFelony ? 'Yes' : 'No'}`;
+    client
+      .fill('input[name="root_serviceBranch"]', data.serviceBranch)
+      .click(`input[name="${activeDuty}"]`)
+      .click(`input[name="${felony}"]`);
+  }
+}
+
+function completeSecondaryContact(client, data, onlyRequiredFields) {
```
raquelromano-gov: I'm guessing other forms will need this as well, but I guess it's OK to leave it here and pull it into shared code when needed (assuming we don't forget about it).

test/e2e/edu-5490-helpers.js
raquelromano-gov: Weird...so if you skip this and start filling in data, you get an error that it can't find that element? I would definitely like to hear at some point what you discovered last week about what nightwatch is doing...I definitely do not understand the nuances of what gets executed when.

test/e2e/edu-helpers.js
```diff
@@ -18,26 +18,47 @@ function initApplicationSubmitMock(form) {
   });
 }
 
-function completeVeteranInformation(client, data, onlyRequiredFields) {
+function completeVeteranInformation(client, data, onlyRequiredFields, root = 'root') {
+  // Use relativeFullname if necessary
```
raquelromano-gov: I don't get when we would want to use the relative data to fill the veteran info.

test/edu-benefits/5490/e2e-test.js
raquelromano-gov: There's no global defined in the nightwatch config...you can do this instead:
```
if (!process.env.BUILDTYPE || process.env.BUILDTYPE === 'development') {
  module.exports = ...
```

test/edu-benefits/5490/00-required.e2e.spec.js
raquelromano-gov: Nice--will do this for the 1990e tests as well.

test/e2e/e2e-helpers.js
jbalboni: You should be able to do that: http://nightwatchjs.org/guide#extending

jbalboni:
@cvalarida We need to make an update to src/test-support/test-server.js. Line 34 should be:

```js
app.use('/education/apply-for-education-benefits/application', fallback('/education/apply-for-education-benefits/application/index.html', { root }));
```

That removes the 1995 from the url so all urls are routed correctly

jbalboni (approved):
This looks great. Definitely an improvement on what we've been doing.

test/edu-benefits/5490/e2e-test.js
cvalarida: Ohh, I thought this was just carried over from yours. Got it. 👍

test/e2e/edu-5490-helpers.js
cvalarida: @jbalboni Good to know! I think it's inconsistent just because of the way we were filling the forms out. Inconsistent between forms, not between runs. That's what I meant.

@raquelromano-gov What ^^ means: I think what we've been doing is opening up an expandable element, skipping it to fill in some data _not_ in the expanded section, then by the time we got back to it, it would be expanded.

I'd need to check to make sure, but because the commands we had for `client` were pretty limited, we had to break chaining to select dropdown items, which caused things to go a little funky with nightwatch execution order. Mind you I'm just guessing at this point...Also with the required vs unrequired fields, there are some cases where we fill out all the required fields, then check to see if we should fill out unrequired fields and do so if necessary, possibly going back up to the top of the form and filling in the blanks. We're then not just going from top to bottom like we would when filling out the form by hand (and then if we don't wait for the expanded elements, it may error).

test/e2e/e2e-helpers.js
cvalarida: @jbalboni Bloody brilliant! I will definitely look into these...Friday, probably.......

test/e2e/edu-5490-helpers.js
cvalarida: None at all. Just didn't get around to it...

test/e2e/edu-5490-helpers.js
cvalarida: Probably because I wrote the helper after I wrote this. Good catch!

test/e2e/edu-5490-helpers.js
cvalarida: I'll double-check, but I think I tried that first and it would have taken a lot of modification or something...

test/e2e/edu-helpers.js
cvalarida: You make an excellent point.

The json has `view:currentSponsorInformation` -> `relativeFullName` but the `formConfig` has `view:currentSponsorInformation` -> `veteranFullName` which makes more sense. I'll update the json to match the config.

test/e2e/e2e-helpers.js
```diff
@@ -87,6 +87,81 @@ function overrideAnimations(client) {
...
+   */
+  client.selectDropdown = function (name, value) {
+    selectDropdown(this, name, value);
+    return this;
+  };
+
+  /**
+   * Clicks the input specified if the condition evaluates to true.
+   */
+  client.clickIf = function (selector, condition, ...params) {
```
cvalarida: Oh, good point. I think something is in the works to upgrade that to 6.0, but for now, that's an excellent point. It's not really necessary for now, so I can take it out.

test/e2e/edu-5490-helpers.js
cvalarida: Oh, actually I didn't write that at all, that's why. I just copied from the 1990e. I'll change it when I move it over to edu-helpers.

test/e2e/edu-5490-helpers.js
cvalarida: Yeah, guess not. I think the only reason I did that was to use the actual data (easily modified, but then there's the refactoring which is fine) and because it's always required (easily worked around by always passing `false` for `onlyRequiredFields` like you suggest).

Alright! I'm on it!

PR #5282, opened 2017-04-07
Remove cooperative training from 5490 education type dropdown

jbalboni:
Connects to [1946](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1946)

This removes cooperativeTraining from the educationType dropdown. The actual change for that is to use the enum property from the educationType definition for 5490, instead of the labels object. I also changed the code to use a reselect selector, since that will avoid running the filter for every change and avoid unnecessary re-renders on the school selection page.

screen shot 2017-04-07 at 4 07 17 pm

### DIFF
--- a/src/js/edu-benefits/5490/config/form.js
+++ b/src/js/edu-benefits/5490/config/form.js
@@ -613,7 +613,8 @@ const formConfig = {
               },
               educationType: {
                 'ui:options': {
-                  updateSchema: ((edTypes) => {
+                  updateSchema: (() => {
+                    const edTypes = educationType.enum;
                     // Using reselect here avoids running the filter code
                     // and creating a new object unless either benefit or
                     // relationship has changed
@@ -636,7 +637,7 @@ const formConfig = {
                       });
 
                     return (pageData, form) => filterEducationType(form);
-                  })(educationType.enum)
+                  })()
                 }
               }
             }
--- a/src/js/edu-benefits/5490/config/form.js
+++ b/src/js/edu-benefits/5490/config/form.js
@@ -1,4 +1,5 @@
 import _ from 'lodash/fp';
+import { createSelector } from 'reselect';
 
 import fullSchema5490 from 'vets-json-schema/dist/22-5490-schema.json';
 
@@ -69,7 +70,6 @@ const ssnSchema = fullSchema5490.definitions.ssn;
 const nonRequiredFullName = _.assign(fullName, {
   required: []
 });
-
 
 const formConfig = {
   urlPrefix: '/5490/',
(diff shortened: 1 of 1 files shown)


cvalarida (approved):

src/js/edu-benefits/5490/config/form.js
```diff
@@ -613,26 +613,30 @@ const formConfig = {
...
-                    const newSchema = _.cloneDeep(schema);
-                    const benefitData = _.get('benefitSelection.data.benefit', form);
-                    const relationshipData = _.get('applicantInformation.data.relationship', form);
-                    const edTypeLabels = Object.keys(_.get('schoolSelection.uiSchema.educationProgram.educationType.ui:options.labels', form));
+                  updateSchema: ((edTypes) => {
+                    // Using reselect here avoids running the filter code
+                    // and creating a new object unless either benefit or
+                    // relationship has changed
+                    const filterEducationType = createSelector(
+                      _.get('benefitSelection.data.benefit'),
```
annekainicUSDS: Looking at the docs, do these arguments not need to be in an array?

src/js/edu-benefits/5490/config/form.js
```diff
@@ -613,26 +613,30 @@ const formConfig = {
...
-                      filterOut.push('flightTraining');
-                    }
+                        return { 'enum': _.without(filterOut, edTypes) };
+                      });
 
-                    newSchema.enum = _.without(filterOut)(edTypeLabels);
-                    return newSchema;
-                  }
+                    return (pageData, form) => filterEducationType(form);
+                  })(educationType.enum)
```
annekainicUSDS: What does this do?

annekainicUSDS (commented):
Couple of questions, but LGTM!

annekainicUSDS:
Latest I heard from @goldenmeanie is that the team wants to go with `cooperativeTraining` instead of `farmCoop`, but I believe this will require a change to the schema. This PR will at least fix the submission failure by aligning the front-end and schema, and then we can create a separate PR for the schema change.

src/js/edu-benefits/5490/config/form.js
jbalboni: I think you can pass them in an array or as individual arguments and have the same result.

src/js/edu-benefits/5490/config/form.js
jbalboni: It's a way of passing in a value to the function that's being immediately invoked here. I could just use `educationType.enum` directly in the function instead, which might be less surprising.

This pattern was more common before const and ES modules, but now it's probably not useful to pass in an argument like this.

src/js/edu-benefits/5490/config/form.js
annekainicUSDS: Gotcha

src/js/edu-benefits/5490/config/form.js
annekainicUSDS: Thanks for the explanation!

annekainicUSDS (approved):

PR #5289, opened 2017-04-10
Adding custom events for GIBCT

### LINKED ISSUE #5217: GI Bill CT: Google Analytics not firing correctly
In the WIP version of the tool it appears that Google Analytics is no longer firing for "EO Principles of Excellence" account. The account number is UA-46533988-2.

Additionally a number of the events seem to be disabled. For example:
* What "learn more"s are folks clicking on
* What schools are users selecting (by name versus facility code)
* What search terms are users using to find schools 

This is how VA tracks historical data on GI Bill CT usage and the way we can have great articles like these:
http://www.blogs.va.gov/VAntage/35078/gi-bill-comparison-tool-breaking-records-helping-veterans/

@mphprogrammer @saneshark

robertfairhead:
Fixes #5217

### DIFF
--- a/src/js/gi/actions/index.js
+++ b/src/js/gi/actions/index.js
@@ -27,8 +27,8 @@ export const FILTER_TOGGLED = 'FILTER_TOGGLED';
 
 export function showModal(modal) {
   window.dataLayer.push({
-    'event': 'gibct-learn-more',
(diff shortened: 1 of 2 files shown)


U-DON (approved):
LGTM.

PR #5291, opened 2017-04-10
5490 previous benefits sponsor file number

cvalarida:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1929
Requires https://github.com/department-of-veterans-affairs/vets-json-schema/pull/81 to be merged first

image

### DIFF
--- a/test/edu-benefits/5490/config/benefitHistory.unit.spec.jsx
+++ b/test/edu-benefits/5490/config/benefitHistory.unit.spec.jsx
@@ -57,7 +57,7 @@ describe('Edu 5490 benefitHistory', () => {
 
     // Should expand to 16
     expect(Array.from(formDOM.querySelectorAll('input,select')).length)
-      .to.equal(16);
+      .to.equal(17);
 
     expect(claimed.checked).to.be.true;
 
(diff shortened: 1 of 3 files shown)


test/edu-benefits/5490/config/benefitHistory.unit.spec.jsx
```diff
@@ -57,7 +57,7 @@ describe('Edu 5490 benefitHistory', () => {
 
     // Should expand to 16
     expect(Array.from(formDOM.querySelectorAll('input,select')).length)
-      .to.equal(16);
+      .to.equal(17);
```
jbalboni: I think we should also have a test that checks the required states for those two fields change appropriately.

jbalboni (approved):
LGTM, just had one request for another unit test.

test/edu-benefits/5490/config/benefitHistory.unit.spec.jsx
cvalarida: Isn't that a test somewhere else?

jbalboni: There might be a similar test somewhere else, but lines 330 and 337 in the config above are new and wouldn't be covered by a test yet.

PR #5293, opened 2017-04-11
Fixes for GIBCT landing page. Updated subheader and added missing eligibility form detail.

U-DON:
### Changes
* Added missing line under dropdown when "VR&E" is picked from the GI Bill benefit dropdown in eligibility form. Resolves department-of-veterans-affairs/vets.gov-team#1937.
* Updated the subheader text on the landing page. Resolves department-of-veterans-affairs/vets.gov-team#1997.
* Removed usage of `If` component on `LandingPage`. Will remove the one remaining instance of its usage and the component altogether in a separate PR.

### DIFF
--- a/src/js/gi/components/search/EligibilityForm.jsx
+++ b/src/js/gi/components/search/EligibilityForm.jsx
@@ -3,7 +3,6 @@ import { connect } from 'react-redux';
 import { showModal, hideModal, eligibilityChange } from '../../actions';
 
 import Dropdown from '../Dropdown';
-import If from '../If';
 import RadioButtons from '../RadioButtons';
 
 export class EligibilityForm extends React.Component {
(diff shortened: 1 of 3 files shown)


U-DON:
@melwoodard With the addition of the VR&E line, it looks like this on the sidebar once you enter search:

> screen shot 2017-04-10 at 10 49 14 pm

It doesn't distinguish itself well from the questions text and looks a little cluttered to me. Is there a particular way you want to deal with it, if anything needs to change? Maybe smaller font size only in the sidebar? Or maybe have it be in an alert box similar to what appears when you select "Active Duty" and "Ch 33", but have it be informational instead of a warning?

src/js/gi/components/search/EligibilityForm.jsx
```diff
@@ -72,19 +71,27 @@ export class EligibilityForm extends React.Component {
...
             eligible to receive a monthly housing allowance.
-          </div>
-        </If>
+          </div>)
+        }
+
+        {
+          this.props.eligibility.giBillChapter === '31' && (<div>
+            To apply for VR&E benefits, please <a
+                href="http://vabenefits.vba.va.gov/vonapp/main.asp"
```
annekainicUSDS: Is this the only place where this form is posted online? As we're trying to shut off VONAPP, not sure if we want to direct more traffic to it.

annekainicUSDS (approved):
LGTM!

melwoodard:
@U-DON the informational alert style works great, thanks!

src/js/gi/components/search/EligibilityForm.jsx
U-DON: The existing version of GIBCT has the same link, so we're just copying over existing functionality. If there's an updated link to replace this with either now or once VONAPP is shut down, I can make that change.

PR #5295, opened 2017-04-11
Add visibility so inputs don't block manage button on edu benefits page

jbalboni:
Connects to [2079](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2079)

Makes the entire button clickable again. Animation still works.


annekainicUSDS (approved):
This is in conjunction with @cvalarida's change, correct? Just wanted to make sure both PRs get in because this one still has the issue on mobile with the button being slightly hidden. Otherwise LGTM!

jbalboni:
@annekainicUSDS I tested it with his fix to be sure it didn't conflict, but this PR doesn't include it.

PR #5296, opened 2017-04-11
Refresh the Login URL Each Time "Sign In" is Clicked

omgitsbillryan:
Related Issue
https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1345

### DIFF
--- a/src/js/login/containers/Main.jsx
+++ b/src/js/login/containers/Main.jsx
@@ -13,6 +13,7 @@ class Main extends React.Component {
     super(props);
     this.setMyToken = this.setMyToken.bind(this);
     this.getLogoutUrl = this.getLogoutUrl.bind(this);
+    this.getLoginUrl = this.getLoginUrl.bind(this);
     this.handleLogin = this.handleLogin.bind(this);
     this.handleLogout = this.handleLogout.bind(this);
     this.handleSignup = this.handleSignup.bind(this);
@@ -25,13 +26,7 @@ class Main extends React.Component {
       this.getLogoutUrl();
     }
 
-    this.serverRequest = fetch(`${environment.API_URL}/v0/sessions/new?level=1`, {
-      method: 'GET',
-    }).then(response => {
-      return response.json();
-    }).then(json => {
-      this.props.onUpdateLoginUrl('first', json.authenticate_via_get);
-    });
+    this.getLoginUrl();
 
     addEvent(window, 'message', (evt) => {
       this.setMyToken(evt);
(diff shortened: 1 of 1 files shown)


src/js/login/containers/Main.jsx
```diff
@@ -75,6 +80,7 @@ class Main extends React.Component {
       });
       const receiver = window.open(`${myLoginUrl}&op=signin`, '_blank', 'resizable=yes,scrollbars=1,top=50,left=500,width=500,height=750');
       receiver.focus();
+      this.getLoginUrl();
```
cvalarida: Mmm...maybe the following are questions for @tiffanybbrown-va:

So if I'm reading this right, `componentDidMount` fetches the login url from...wherever `environment.API_URL` points to (vets-api? ID.me's api?). It passes that to `this.props.onUpdateLoginUrl()` which (I assume) updates `this.props.loginUrl.first`. We then use that to log in via `handleLogin()`.

Line 76 reads:
```js
const myLoginUrl = this.props.login.loginUrl.first;
```
That's the line that gets the url to actually log in on line...81? After we open that window, we go fetch another url.

Knowing as little as I do about this, I'm wondering why fetching a new url afterward would be necessary. I mean, doesn't it get a new url every time the component mounts? So are all these resubmissions people who login, do their thing, log out, but never leave the site, refresh, close the tab / window, or anything like that?

I'm probably missing something, but this seems like it'll fetch a login url before login and after login, but the second one probably won't be used most of the time. Would it make more sense to fetch the url at the beginning of `handleLogin()`, or is that in `componentDidMount` because it was taking a while and optimized because of that? Or was it because we need it in the redux store (I assume that's where the prop with the url comes from) before we log in for some reason?

Sorry for being so vague, but I'm trying to wrap my head around the whole problem before I can say this is "the right" solution.

cvalarida (commented):
There may be a better way, but without knowing all the moving parts, I can't say yet...

I also don't know what the side effects are of this change, so neither can I say that if it works, do it!

src/js/login/containers/Main.jsx
tiffanybbrown-va: I don't quite understand how login works, tbh. But it looks this PR just moves the `this.serverRequest` into a wrapper function.

tiffanybbrown-va:
LGTM.

src/js/login/containers/Main.jsx
cvalarida: Seems to me like it's creating two new sessions every time just in case the second one is going to be used in the future. Is that going to cause any trouble anywhere?

omgitsbillryan: @cvalarida I also noticed the refresh happening in `componentDidMount()`, which means that the [AuthN URL Re-Use](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1345) must be occurring by way of someone clicking "Sign In" repeatedly w/o refreshing the page - this PR addresses that use-case.

I have a feeling that we may not have heard the last of this and that the URLs are being re-used in some other unknown way, especially when you consider that one URL was used over the course of 30+ days!  I will followup with ID.me.

PR #5297, opened 2017-04-11
Build front end 5495 CH 1 : Applicant Information

jbalboni:
Connects to [683](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/683)

Builds the first page. I had to tweak the applicantInformation shared page, since it doesn't have vaFileNumber. We could use veteranId, but it's tied to veteranSocialSecurityNumber (vs relativeSocialSecurityNumber), so I left it alone.

screen shot 2017-04-11 at 1 43 50 pm
screen shot 2017-04-11 at 1 43 56 pm
screen shot 2017-04-11 at 1 44 00 pm

### DIFF
--- a/src/js/edu-benefits/pages/applicantInformation.js
+++ b/src/js/edu-benefits/pages/applicantInformation.js
@@ -41,6 +41,7 @@ export default function applicantInformation(schema, options) {
     'view:noSSN': {
       type: 'boolean'
     },
+    vaFileNumber: schema.definitions.vaFileNumber,
     relativeDateOfBirth: schema.definitions.date,
     gender: schema.definitions.gender,
     relationship: schema.definitions.relationship
@@ -58,6 +59,16 @@ export default function applicantInformation(schema, options) {
       }),
       'view:noSSN': {
         'ui:title': 'I don’t have a Social Security number',
+      },
+      vaFileNumber: {
+        'ui:required': (formData) => !!_.get('view:noSSN', formData),
+        'ui:title': 'File number',
+        'ui:errorMessages': {
+          pattern: 'File number must be 8 digits'
+        },
+        'ui:options': {
+          expandUnder: 'view:noSSN'
+        }
       },
       relativeDateOfBirth: _.assign(currentOrPastDate.uiSchema('Date of birth'),
         {
--- a/test/edu-benefits/pages/applicantInformation.unit.spec.jsx
+++ b/test/edu-benefits/pages/applicantInformation.unit.spec.jsx
@@ -10,8 +10,6 @@ import { DefinitionTester, submitForm } from '../../util/schemaform-utils.jsx';
 import applicantInformation from '../../../src/js/edu-benefits/pages/applicantInformation.js';
 
 import fullSchema1990e from 'vets-json-schema/dist/22-1990E-schema.json';
-import fullSchema5490 from 'vets-json-schema/dist/22-5490-schema.json';
-import fullSchema5495 from 'vets-json-schema/dist/22-5495-schema.json';
 
 function fillInformation(find) {
   ReactTestUtils.Simulate.change(find('#root_relativeFullName_first'), {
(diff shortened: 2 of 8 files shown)


test/edu-benefits/pages/applicantInformation.unit.spec.jsx
```diff
@@ -142,12 +148,16 @@ describe('Edu 1990e applicantInformation', () => {
 
 describe('Edu 5490 applicantInformation', () => {
   // Make sure these match the fields listed in the 5490
-  const { schema, uiSchema } = applicantInformation(fullSchema5490, [
-    'relativeFullName',
-    'relativeSocialSecurityNumber',
-    'relativeDateOfBirth',
-    'gender'
-  ]);
+  const { schema, uiSchema } = applicantInformation(fullSchema5490, {
```
jbalboni: I dislike this pattern for these unit tests. We're only coincidentally testing the actual forms and this test in particular had code that wasn't being used (the field list.

raquelromano-gov: Yeah, I think we need to overhaul how we test shared pages (make sure they are actually unit test, not moving toward integration tests, but test the functionality we know is used by those forms)

raquelromano-gov:
I'm surprised that we are now allowing VA File Number for an applicant/relative who is not necessarily a veteran. In past conversations, we settled on allowing the applicant to check "I don't have a Social Security number" but don't ask for anything else, under the assumption that the sponsor/veteran would still be identifiable later by either SSN or VA FN.

jbalboni:
@raquelromano-gov Me, too, but I asked @goldenmeanie about it specifically.

goldenmeanie:
This is because applicants who didn't have a SSN when they originally applied, are now making a change to an existing benefit. They now have a file number!

raquelromano-gov:
Ah...got it! Good to know.

src/js/edu-benefits/pages/applicantInformation.js
```diff
@@ -59,6 +60,16 @@ export default function applicantInformation(schema, options) {
       'view:noSSN': {
         'ui:title': 'I don’t have a Social Security number',
       },
+      vaFileNumber: {
+        'ui:required': (formData) => !!_.get('view:noSSN', formData),
```
cvalarida: Since this is such a common pattern, I think we really need a way to move as much logic out of the configs as possible and abstract it into its own file (which really means modifying `veteranId`'s `uiSchema` to allow for other field names). I have ideas for this...We could use default params and computed property names to allow for veteran vs relative information. Also passing in a `rootPath` object to preface the path in `ui:required` and `expandUnder`.

Should we do that here, or make another ticket for that?

test/edu-benefits/pages/applicantInformation.unit.spec.jsx
cvalarida: Could export the fields for each form in the page file.

```js
export const applicantInformationFields5490 = [ ... ];
```
or even
```js
const defaultFields = [ ... ];
export const formFields = {
  '1990e': defaultFields,
  '5490': [ ... ],
  '5495': [ ... ]
}
```

cvalarida (approved):

src/js/edu-benefits/pages/applicantInformation.js
raquelromano-gov: I like thus idea and vote for a separate issue & PR

test/edu-benefits/pages/applicantInformation.unit.spec.jsx
jbalboni: I rather we just pull the schemas from the form config object, I think.

src/js/edu-benefits/pages/applicantInformation.js
jbalboni: Sounds good to me. I'll create a ticket

jbalboni:
@cvalarida I ended up refactoring those tests some. The 5490/1990e forms are just using the defaults, so I gave them each a test that's really just to make sure something renders. The 5495 has that and a test to check the conditional file number. The common definition tests are mostly the same (except minus the form specific tests).

I think that's essentially the same coverage of config/logic with less duplication and less opportunity for tests and code getting out of sync.

PR #5300, opened 2017-04-11
Add references to commonly used rjsf functionality

cvalarida:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1519

If there are more you can think of that I should add, please let me know.

### DIFF
--- a/docs/schemaform/form-config.md
+++ b/docs/schemaform/form-config.md
@@ -140,7 +140,19 @@ This does not apply to array fields; for those, you still need to specify an `it
 ```
 
 ### uiSchema configuration
-In addition to the uiSchema options listed in the library [https://github.com/mozilla-services/react-jsonschema-form#the-uischema-object](docs), we have some additional options that are supported for all forms:
+If you're not already familiar with the rjsf uiSchema options, check out the [https://github.com/mozilla-services/react-jsonschema-form#the-uischema-object](library docs). Some commonly used options include:
+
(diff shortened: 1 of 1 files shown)


docs/schemaform/form-config.md
```diff
@@ -140,7 +140,19 @@ This does not apply to array fields; for those, you still need to specify an `it
 ```
 
 ### uiSchema configuration
-In addition to the uiSchema options listed in the library [https://github.com/mozilla-services/react-jsonschema-form#the-uischema-object](docs), we have some additional options that are supported for all forms:
+If you're not already familiar with the rjsf uiSchema options, check out the [https://github.com/mozilla-services/react-jsonschema-form#the-uischema-object](library docs). Some commonly used options include:
```
cvalarida: Why aren't any of these [links](https://guides.github.com/features/mastering-markdown/) showing up properly?

ayaleloehr: Swap what's in () to be in front of what's in [], instead of after.  I beleive the format should be `(anchor text)[https://url]`.

jbalboni (approved):
LGTM, once the links are fixed.

docs/schemaform/form-config.md
cvalarida: Blasted...I was only looking at the `[]` and `()`--which are in the right order--but it's the content inside. The text goes first and link goes second. There it is.

Issue #4232, opened 2016-12-05
Replace babel-polyfill with ES2015 only core-js polyfills


jbalboni:
We're importing babel-polyfill to polyfill ES2015 features for all browsers, but that polyfill includes polyfills for ES.next features (like observables) as well, which we don't need. It may also include ES5 and ES3 polyfills that the browsers we support do not need.

We should look into replacing babel-polyfill with the just the polyfills from core-js (which babel-polyfill uses under the hood) that we need and see if it reduces the size of app bundles notably.

Related to #4115.

annekainicUSDS:
Change needed: investigate if ES6 polyfill can be used in place of the babel-polyfill. Work with the Core team on this as this would be a global change. Cc: @robbiethegeek-usds

hzoo:
FYI you can use https://github.com/babel/babel-preset-env with the useBuiltIns option to only load polyfills required by the environments supported (and babel-preset-env in general for syntax) This would replace "es2015" or "latest". Let me know if you have questions or join http://slack.babeljs.io/

jbalboni:
Thanks @hzoo! We'll check that out.

PR #5302, opened 2017-04-11
Update to use babel env preset and only necessary polyfills

jbalboni:
Connects to #4232 

This updates us to the newer env preset from Babel, which includes es2015, 2016, and 2017. We were already transpiling those features due to the stage-2 preset, which is a superset of 2017 and lower. So I've left it alone because there are a handful of plugins in there that aren't in env (pretty sure we don't use them, but just to be safe).

The bigger change for this is around polyfills. The `useBuiltIns` option in the env preset does an autoprefixer-like check to only include polyfills you need for the browsers you support. Enabling this gets rid of some older polyfills, plus some stage-1/0 ones that we don't need, and reduces the size of core-js by 25kb (minified).

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,7 +1,13 @@
 {
   // Javascript settings.
   "presets": [
-    "es2015",
+    ["env", {
+      "targets": {
+        "uglify": true,
+        "browsers": ["last 2 versions", "ie 11", "not ie 10"]
+      },
+      "useBuiltIns": true
+    }],
     "react",
     "stage-2",
   ],
--- a/config/webpack.config.js
+++ b/config/webpack.config.js
@@ -34,7 +34,7 @@ const configGenerator = (options) => {
     filesToBuild = _.pick(entryFiles, options.entry.split(',').map(x => x.trim()));
   }
   filesToBuild.vendor = [
-    'core-js',
+    './src/js/common/polyfills',
     'history',
     'jquery',
     'react',
--- a/src/js/common/polyfills.js
+++ b/src/js/common/polyfills.js
@@ -1,0 +1,26 @@
+// Common Javascript environment setup to be used by all entrypoints.
+//
+// Keep this short. It should mostly be polyfills, global site style, and any
+// truly pervasive libraries. Be careful if you are tempted to put jquery or
+// other such libraries in here. Most of the site does not use these legacy
+// frameworks and it belongs in a lower-level module.
+
+require('babel-polyfill');
+
+// Basic polyfills.
+// TODO(awong): These do NOT correctly conditionally load the polyfill.
+// The polyfill is always loaded. require.ensure() should be used instead but
+// then load ordering needs to be worked out. Fix later.
+const Modernizr = require('modernizr');
+if (!Modernizr.classlist) {
+  require('classlist-polyfill'); // DOM element classList support.
+}
+if (!Modernizr.dataset) {
+  require('dataset');  // dataSet accessor support.
+}
+if (!Modernizr.fetch) {
+  require('whatwg-fetch');  // dataSet accessor support.
+}
+
+// This polyfill has its own test logic so no need to conditionally require.
+require('polyfill-function-prototype-bind');
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -1771,6 +1771,170 @@
         }
       }
     },
+    "babel-preset-env": {
+      "version": "1.3.3",
+      "from": "babel-preset-env@>=1.3.3 <2.0.0",
+      "resolved": "https://registry.npmjs.org/babel-preset-env/-/babel-preset-env-1.3.3.tgz",
+      "dependencies": {
+        "babel-code-frame": {
+          "version": "6.22.0",
+          "from": "babel-code-frame@>=6.22.0 <7.0.0",
+          "resolved": "http://registry.npmjs.org/babel-code-frame/-/babel-code-frame-6.22.0.tgz"
+        },
+        "babel-helper-call-delegate": {
+          "version": "6.24.1",
+          "from": "babel-helper-call-delegate@>=6.24.1 <7.0.0",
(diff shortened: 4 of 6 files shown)


.babelrc
```diff
@@ -1,7 +1,13 @@
 {
   // Javascript settings.
   "presets": [
-    "es2015",
+    ["env", {
+      "targets": {
+        "uglify": true,
```
jbalboni: We have to set this because uglify doesn't support ES6. Though it doesn't really matter until we drop IE 11 support.

config/webpack.config.js
```diff
@@ -34,7 +34,7 @@ const configGenerator = (options) => {
     filesToBuild = _.pick(entryFiles, options.entry.split(',').map(x => x.trim()));
   }
   filesToBuild.vendor = [
-    'core-js',
+    './src/js/common/polyfills',
```
jbalboni: If we leave core-js here, then we'll always bundle all of core-js in the vendor chunk, which defeats the purpose of having the Babel transform only include the polyfills we need.

jbalboni:
@ayaleloehr @robbiethegeek-usds Not sure who all should review this.

ayaleloehr:
maybe another Rainbows developer? @annekainicUSDS @raquelromano-gov or @cvalarida able to take a look?

src/js/common/polyfills.js
```diff
@@ -0,0 +1,26 @@
+// Common Javascript environment setup to be used by all entrypoints.
+//
+// Keep this short. It should mostly be polyfills, global site style, and any
+// truly pervasive libraries. Be careful if you are tempted to put jquery or
+// other such libraries in here. Most of the site does not use these legacy
+// frameworks and it belongs in a lower-level module.
+
+require('babel-polyfill');
```
annekainicUSDS: So we still need to use `babel-polyfill`, but now that we're using `babel-preset-env` it's only loading the polyfills we actually need based on the environments we support?

annekainicUSDS (approved):
LGTM! Had one question to make sure I understood these changes. The only thing I would want to make sure of is that we've tested this in various browsers before deploying. We could do that in staging before it hits production if that's easiest.

src/js/common/polyfills.js
jbalboni: Yeah, `babel-preset-env` has a transform that takes `require('babel-polyfill')` and turns into a list of require statements for the specific polyfills, sort of like this:

```js
require('core-js/es6/array/find');
require('core-js/es6/array/from');
require('core-js/es6/array/fill');
```

etc, based on what our supported browsers need.

jbalboni:
@annekainicUSDS I tested some of the edu forms in IE11, but it would be good to test non-rainbows apps.

.babelrc
```diff
@@ -1,7 +1,13 @@
...
+        "browsers": ["last 2 versions", "ie 11", "not ie 10"]
```
cvalarida: Are the last two versions our officially supported browsers? (Do we keep that information anywhere?)

annekainicUSDS:
@ayaleloehr perhaps we could ask Kudos to help test some of their apps as well on this branch?

src/js/common/polyfills.js
cvalarida: By what I'm reading [here](http://2ality.com/2017/02/babel-preset-env.html#usebuiltins-boolean-default-false), it looks like that's true for both `babel-polyfill` and `core-js`.

The only difference I see is this:
> `babel-polyfill` polyfills `core-js` and the regenerator runtime (to emulate generators on ES5).

Is there any other reason to choose one over the other?

config/webpack.config.js
cvalarida: I think perhaps I'm misunderstanding what this `filesToBuild.vendor` is for. Does it include these in the vendor bundle, or does it just transpile these (or some such)?

If it includes them, do we need to `require('./polyfills')` from `src/js/common/index.js`, or does that line just pull them into working memory from the vendor bundle where they reside?

src/js/common/polyfills.js
jbalboni: We've been using babel-polyfill, I think that's the main reason.

.babelrc
jbalboni: I think so, but we should verify that.

config/webpack.config.js
jbalboni: It includes this file and anything it depends on in the vendor bundle. It should not actually cause the require calls in the polyfills.js file to execute, though, which is why I left the `require('./polyfills')` statement in there. Similar to how React is in the vendor bundle, but it won't ever be executed unless you actually import/require it.

However, they're polyfills and it's possible some of them have code that is run before the actual `require` statement is called. I don't think that's the case, but it's possible.

ayaleloehr:
Kudos is neck deep in GIBCT testing. I will pull this down tomorrow and test out Rx, SM, and Facility Locator.

config/webpack.config.js
cvalarida: Gotcha. So these make it into the vendor bundle so we can use them in the other bundles. 👍

Thanks!

.babelrc
ayaleloehr: For whatever it's worth, [this](https://github.com/department-of-veterans-affairs/vets.gov-team/blob/4eaa95c7f30f8f4b78bdb8ba2ebe0d7d27dd7763/Decisons/Browser%20Support.md) is what is currently documented as our browser support policy, but I believe that needs updating. 

[This](https://github.com/department-of-veterans-affairs/vets-website/pull/3716/files) PR is what implemented telling users with out of date browsers to update if you want to see the code. 

And [this](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/358) issue is when the decision to make the minimum supported version IE11 was made. 

Phew, that's entirely disorganized and should be combined somewhere (likely in the first link?)

ayaleloehr:
Tested facility locator, prescription refills, and secure messaging locally and all look good. 👍  from me!

jbalboni:
Thanks @ayaleloehr!

PR #5306, opened 2017-04-11
5495 Chapter 3 -- Sponsor Information

cvalarida:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/687

## Page 1
@goldenmeanie Should these say "Sponsor first name" et al?
image

---
image

Note: I changed the wording on the felony question from the original ticket to use "sponsor."

## Page 2
Per our restructure, this is actually in chapter 1, not 3.
image

---

image

### DIFF
--- a/src/js/edu-benefits/5490/config/form.js
+++ b/src/js/edu-benefits/5490/config/form.js
@@ -27,7 +27,6 @@ import * as address from '../../../common/schemaform/definitions/address';
 import * as currentOrPastDate from '../../../common/schemaform/definitions/currentOrPastDate';
 import * as date from '../../../common/schemaform/definitions/date';
 import * as phone from '../../../common/schemaform/definitions/phone';
-import * as toursOfDuty from '../../definitions/toursOfDuty';
 import * as veteranId from '../../definitions/veteranId';
 
 import { uiSchema as dateRangeUi } from '../../../common/schemaform/definitions/dateRange';
@@ -37,9 +36,10 @@ import postHighSchoolTrainingsUi from '../../definitions/postHighSchoolTrainings
 
 import createContactInformationPage from '../../pages/contactInformation';
 import directDeposit from '../../pages/directDeposit';
-import applicantInformation from '../../pages/applicantInformation';
+import applicantInformationPage from '../../pages/applicantInformation';
+import applicantServicePage from '../../pages/applicantService';
 import createSchoolSelectionPage from '../../pages/schoolSelection';
-import additionalBenefits from '../../pages/additionalBenefits';
+import additionalBenefitsPage from '../../pages/additionalBenefits';
 
 import IntroductionPage from '../components/IntroductionPage';
 import ConfirmationPage from '../containers/ConfirmationPage';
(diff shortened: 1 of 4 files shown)


src/js/edu-benefits/5490/config/form.js
```diff
@@ -93,53 +93,13 @@ const formConfig = {
     applicantInformation: {
       title: 'Applicant Information',
       pages: {
-        applicantInformation: applicantInformation(fullSchema5490, {
+        applicantInformation: applicantInformationPage(fullSchema5490, {
```
cvalarida: Just to keep the naming consistent.

src/js/edu-benefits/5495/config/form.js
```diff
@@ -78,6 +82,35 @@ const formConfig = {
...
       }
+    },
+    sponsorInformation: {
+      title: 'Sponsor information',
+      pages: {
+        sponsorInformation: {
+          path: 'sponsor/information',
+          title: 'Sponsor information',
+          uiSchema: {
+            veteranFullName: fullName.uiSchema,
```
cvalarida: I may change this to be a function which takes a string param and tacks it on the beginning of the name field labels. It could also be used for both the sponsor (veteran) and applicant (dependent) with some additional modifications (passing another value and using it as a computed property name, probably).

test/edu-benefits/pages/applicantService.unit.spec.jsx
```diff
@@ -4,10 +4,10 @@ import { expect } from 'chai';
 import sinon from 'sinon';
 import ReactTestUtils from 'react-addons-test-utils';
 
-import { DefinitionTester, submitForm } from '../../../util/schemaform-utils.jsx';
-import formConfig from '../../../../src/js/edu-benefits/5490/config/form';
+import { DefinitionTester, submitForm } from '../../util/schemaform-utils.jsx';
+import formConfig from '../../../src/js/edu-benefits/5490/config/form';
 
-describe('Edu 5490 applicantService', () => {
+describe('Edu 5490 (default) applicantService', () => {
```
cvalarida: Do I need to do anything more than this here? The coverage is the same for both the 5490 and 5495.

src/js/edu-benefits/5490/config/form.js
jbalboni: We seem to be inconsistent across several pages. Some have page suffixes, some have create prefixes, or both/neither.

src/js/edu-benefits/pages/applicantService.js
```diff
@@ -0,0 +1,53 @@
...
+
+import * as toursOfDuty from '../definitions/toursOfDuty';
+
+/**
+ * Returns the applicantService page. This is for dependent applicants.
+ *
+ * Note: This is a function only to keep it consistent with the other pages, not
+ *  because it really needs to be.
+ */
+export default function applicantService() {
```
jbalboni: Probably should have this end with `Page`?

test/edu-benefits/pages/applicantService.unit.spec.jsx
jbalboni: I think in here, since it's meant to test the code in edu-benefits/pages/applicantService.js, I'd import from that file and call the function to get the uiSchema. And use the schema from either 5490 or 5495.

It's the same result, but if we change something in applicantService.js, this test may continue to work and make us think we're covering it when we aren't.

jbalboni (approved):
LGTM, had some minor comments.

src/js/edu-benefits/5490/config/form.js
cvalarida: Yeah, I'm not sure how much it really matters, but I figured I'd touch up as I went along.

src/js/edu-benefits/pages/applicantService.js
cvalarida: Good call.

PR #5307, opened 2017-04-12
GIBCT preview mode

U-DON:
Resolves department-of-veterans-affairs/vets.gov-team#1986.

There was some existing work for `PreviewBanner`, but I made a bunch of fixes and changes to make it appear based on the existence of a `?version=<uuid>` query parameter. Any API requests will also provide the same UUID. The idea is that admins can access a version of the app that uses preview data.

There is also a bunch of added logic to handle carrying over the query param (or dropping it when exiting preview mode) while navigating through the app.

> screen shot 2017-04-11 at 8 05 39 pm

### DIFF
--- a/src/js/gi/actions/index.js
+++ b/src/js/gi/actions/index.js
@@ -67,8 +67,9 @@ function withPreview(dispatch, action) {
   dispatch(action);
 }
 
-export function fetchConstants(preview = false) {
-  const url = `${api.url}/calculator_constants${preview ? 'version=preview' : ''}`;
+export function fetchConstants(version) {
+  const queryString = version ? `?version=${version}` : '';
+  const url = `${api.url}/calculator_constants${queryString}`;
 
   return dispatch => {
     dispatch({ type: FETCH_CONSTANTS_STARTED });
@@ -89,12 +90,13 @@ export function updateAutocompleteSearchTerm(searchTerm) {
   };
 }
 
-export function fetchAutocompleteSuggestions(text) {
-  const previewVersion = '1'; // TODO
-  const url = [
-    `${api.url}/institutions/autocomplete?preview=${previewVersion}`,
-    `term=${text}`
-  ].join('&');
+export function fetchAutocompleteSuggestions(text, version) {
+  const queryString = [
+    `term=${text}`,
+    (version ? `version=${version}` : '')
+  ].filter(q => q).join('&');
+
+  const url = `${api.url}/institutions/autocomplete?${queryString}`;
 
   return dispatch => {
     dispatch({ type: AUTOCOMPLETE_STARTED });
@@ -127,11 +129,10 @@ export function institutionFilterChange(filter) {
 }
 
 export function fetchSearchResults(query = {}) {
-  const fullQuery = { ...query, version: 1 };
   const queryString =
-    Object.keys(fullQuery).reduce((str, key) => {
+    Object.keys(query).reduce((str, key) => {
       const sep = str ? '&' : '';
-      return `${str}${sep}${snakeCase(key)}=${fullQuery[key]}`;
+      return `${str}${sep}${snakeCase(key)}=${query[key]}`;
     }, '');
 
   const url = `${api.url}/institutions/search?${queryString}`;
@@ -148,8 +149,9 @@ export function fetchSearchResults(query = {}) {
   };
 }
 
-export function fetchProfile(facilityCode, version = 1) {
-  const url = `${api.url}/institutions/${facilityCode}?version=${version}`;
+export function fetchProfile(facilityCode, version) {
+  const queryString = version ? `?version=${version}` : '';
+  const url = `${api.url}/institutions/${facilityCode}${queryString}`;
 
   return dispatch => {
     dispatch({ type: FETCH_PROFILE_STARTED });
--- a/src/js/gi/actions/index.js
+++ b/src/js/gi/actions/index.js
@@ -67,9 +67,8 @@ function withPreview(dispatch, action) {
   dispatch(action);
 }
 
-export function fetchConstants() {
-  const previewVersion = ''; // TODO
-  const url = `${api.url}/calculator_constants?preview=${previewVersion}`;
+export function fetchConstants(preview = false) {
+  const url = `${api.url}/calculator_constants${preview ? 'version=preview' : ''}`;
 
   return dispatch => {
     dispatch({ type: FETCH_CONSTANTS_STARTED });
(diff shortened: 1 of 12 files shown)


src/js/gi/config.js
```diff
@@ -6,8 +6,7 @@ module.exports = {
     url: `${environment.API_URL}/v0/gi`,
     settings: {
       headers: {
-        'Content-Type': 'application/json',
-        // 'X-Key-Inflection': 'camel',
+        'X-Key-Inflection': 'camel'
```
U-DON: I was going to add this in a separate PR that would remove any usage of the `camelcase-keys-recursive` package, but wanted to add this here in order to get the newly added `SET_VERION` action to read `version.createdAt` instead of `version.created_at`. Will still get around to the removal separately.

src/js/gi/containers/GiBillApp.jsx
```diff
@@ -16,18 +17,60 @@ const Disclaimer = () => {
...
+    } else if (shouldShowPreviewMode) {
+      this.props.enterPreviewMode();
+    }
+
+    const shouldRefetchConstants = prevProps.location.query.version !== uuid;
 
-  componentWillMount() {
-    this.props.updateConstants();
+    if (shouldRefetchConstants) {
+      this.props.fetchConstants(uuid);
```
U-DON: Constants don't really have versioning right now. The current purpose of this is to reset the `preview.version` object in state, in case the user goes back to preview mode right after exiting (going back in history so the query param gets tacked on again). But if versioning gets introduced later, there won't need to be a change here, so that's a bonus.

src/js/gi/actions/index.js
```diff
@@ -63,21 +64,27 @@ function withPreview(dispatch, action) {
...
+  const url = `${api.url}/calculator_constants${queryString}`;
 
   return dispatch => {
     dispatch({ type: FETCH_CONSTANTS_STARTED });
 
     return fetch(url, api.settings)
       .then(res => res.json())
       .then(
-        payload => dispatch({ type: FETCH_CONSTANTS_SUCCEEDED, payload }),
+        payload => withPreview(dispatch, { type: FETCH_CONSTANTS_SUCCEEDED, payload }),
```
U-DON: Currently, requests for constants will always return `version.preview` as `false`, since there's no versioning on constants at the moment. This is mainly to set the (otherwise null) date on the preview banner, since the landing page doesn't make any other API request, and we need to show the banner on every GIBCT page as long as there is a `version` query param.

src/sass/gi/partials/_gi-preview-banner.scss
```diff
@@ -27,8 +25,11 @@
       margin: 0;
 
       .actions {
-        white-space: nowrap;
-        margin: auto 2em;
+        @media screen and (min-width: $medium-screen) {
+          bottom: 0;
+          position: absolute;
```
cvalarida: At a specific width, the banner gets garbled. A little larger or a little smaller looks fine, though. image

src/js/gi/components/heading/PreviewBanner.jsx
```diff
@@ -1,58 +1,38 @@
...
       </div>
     );
   }
 }
 
 PreviewBanner.propTypes = {
-  show: React.PropTypes.bool.isRequired,
   version: React.PropTypes.object.isRequired,
-  toolURL: React.PropTypes.string.isRequired,
+  toolUrl: React.PropTypes.string.isRequired,
```
cvalarida: This `.isRequired`, but not passed in GiBillApp.jsx (line 67-69). Should it be required if it has a default (line 35)? This may just be convention, I'm not sure.

src/js/gi/containers/LandingPage.jsx
```diff
@@ -55,12 +72,15 @@ export class LandingPage extends React.Component {
...
-  return {
-    autocomplete: state.autocomplete,
-  };
+  const { autocomplete } = state;
+  return { autocomplete };
 };
+
 const mapDispatchToProps = {
-  setPageTitle
+  clearAutocompleteSuggestions,
```
cvalarida: Was this intentionally alphabetized? I like that. 😃

src/js/gi/containers/GiBillApp.jsx
```diff
@@ -16,18 +17,60 @@ const Disclaimer = () => {
...
+      preview,
+      location: { query: { version: uuid } }
+    } = this.props;
+
+    const shouldHidePreviewMode = preview.display && !uuid;
+    const shouldShowPreviewMode = !preview.display && uuid && preview.version.createdAt;
+
+    if (shouldHidePreviewMode) {
+      this.props.exitPreviewMode();
+    } else if (shouldShowPreviewMode) {
```
cvalarida: I did a double take here. I thought, "Wait, if it shouldn't show and it shouldn't hide, what should it do?" It wasn't until I went back to read the definitions of these variables that I figured out the purpose.

You might change these to `shouldExitPreviewMode` and `shouldEnterPreviewMode` for clarity. Then it's easy to figure out that if we shouldn't exit and we shouldn't enter, we're just staying where we're at.

cvalarida (approved):
I feel like I'm not able to give as good of a review as this deserves, coming in cold like this, but it looks good. I tested that it uses the version id and makes its queries using the right version.

I just had a few questions and comments, but I think it's golden ( :crossed_fingers: ).

src/js/gi/actions/index.js
```diff
@@ -90,12 +97,13 @@ export function updateAutocompleteSearchTerm(searchTerm) {
...
-  const previewVersion = '1'; // TODO
-  const url = [
-    `${api.url}/institutions/autocomplete?preview=${previewVersion}`,
-    `term=${text}`
-  ].join('&');
+export function fetchAutocompleteSuggestions(text, version) {
+  const queryString = [
+    `term=${text}`,
+    (version ? `version=${version}` : '')
+  ].filter(q => q).join('&');
```
U-DON: Removes the empty string if there's no version. Filters out anything falsy.

src/js/gi/components/heading/PreviewBanner.jsx
U-DON: Not sure what the convention is. It was like this (required with default) when I started working with the code, so I just left it alone.

src/js/gi/containers/LandingPage.jsx
U-DON: Yup, just something I like to do when cleaning up.

src/js/gi/containers/GiBillApp.jsx
U-DON: Yeah, good catch. They should match the names of the actions anyway.

PR #5310, opened 2017-04-12
Fix margin on 1990e benefits alert box

jbalboni:
Connects to [1847](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1847)

screen shot 2017-04-12 at 11 41 28 am

### DIFF
--- a/src/js/edu-benefits/1990e/config/form.js
+++ b/src/js/edu-benefits/1990e/config/form.js
@@ -68,9 +68,7 @@ const formConfig = {
           path: 'benefits/eligibility',
           title: 'Benefits eligibility',
           uiSchema: {
-            'view:benefitDescription': {
-              'ui:description': eligibilityDescription
-            },
+            'ui:description': eligibilityDescription,
             benefit: {
               'ui:widget': 'radio',
               'ui:title': 'Select the benefit that has been transferred to you.',
@@ -82,10 +80,6 @@ const formConfig = {
           schema: {
             type: 'object',
             properties: {
-              'view:benefitDescription': {
-                type: 'object',
-                properties: {}
-              },
               benefit
             }
           },
--- a/src/sass/edu-benefits.scss
+++ b/src/sass/edu-benefits.scss
@@ -329,8 +329,13 @@ label {
   margin-bottom: 7rem;
 }
 
-.usa-alert ul:first-child, .edu-alert-top {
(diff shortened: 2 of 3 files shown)


src/sass/edu-benefits.scss
```diff
@@ -329,7 +329,7 @@ label {
   margin-bottom: 7rem;
 }
 
-.usa-alert ul:first-child {
+.usa-alert ul:first-child, .edu-alert-top {
```
annekainicUSDS: Why is it that sometimes we want to apply the 0 margin to the `usa-alert` level div and sometimes to the `ul` within it?

src/js/edu-benefits/1990e/config/form.js
```diff
@@ -68,9 +68,7 @@ const formConfig = {
           path: 'benefits/eligibility',
           title: 'Benefits eligibility',
           uiSchema: {
-            'view:benefitDescription': {
-              'ui:description': eligibilityDescription
-            },
+            'ui:description': eligibilityDescription,
```
annekainicUSDS: Good to know we don't have to wrap this in a `view:` field. Are `view:` fields only needed for actual form fields, or when other fields are nested within them?

annekainicUSDS (commented):
LGTM! A couple of questions. Do you think the alert box would also benefit from a little bit of margin bottom to space out the questions the follow it?

src/js/edu-benefits/1990e/config/form.js
jbalboni: We've used them to conditionally show description text; I think that's the only use for empty ones like this one had.

src/sass/edu-benefits.scss
jbalboni: I think they're two separate cases that I put together because it was convenient. I could make separate rules for them.

The alert already has inner padding, so it doesn't need the `ul` margin. That's the reason for that class, but it doesn't have anything to do with the change I made.

jbalboni:
@goldenmeanie What do you think about the bottom margin for this alert box?

goldenmeanie:
I'd say it's a little close :)

src/sass/edu-benefits.scss
annekainicUSDS: Ah gotcha, it's fine to keep together, just wondering!

Issue #5317, opened 2017-04-12
Military tuition assistance field appears on OJT profiles


melwoodard:
@saintsoup52 - @U-DON noticed that OJT profiles show a field in the calculator for Military Tuition Assistance - which seems like something specific to school tuition and not on the job training. 

This exists in production as well as the new version we're building out - and the field doesn't seem to affect anything in the calculator results. Is this field supposed to be in OJT or can it be removed?

saintsoup52:
Good catch... TA only affects the tuition/fees benefit and therefore that question should be hidden when an OJT program is selected.

melwoodard:
Thanks, that makes sense!

PR #5319, opened 2017-04-12
Hide certain fields in GIBCT calculator that should not be appearing for OJT institutions.

U-DON:
### Notes
The logic is largely copied from an existing app that we're replacing, but I tried to minimize code spaghetti-ness in the process, so some bugs such as these arose from that refactoring.

These changes are to hide certain fields that are irrelevant to calculating benefits when viewing the profile page of an employer instead of a school. Variables such as tuition fees aren't taken into account, so those inputs and calculated values shouldn't be shown. All of the calculation is done in this `reselect` selector, so the changes were limited to one file here.

### Changes
* Hide tuition and fees per term section for OJT schools. Resolves department-of-veterans-affairs/vets.gov-team#2103.
* Append '/mo' to the book stipend value (which is monthly) for OJT schools. Resolves department-of-veterans-affairs/vets.gov-team#2104.
* Hide books and supplies field for OJT schools. Resolves department-of-veterans-affairs/vets.gov-team#2105.

### DIFF
--- a/src/js/gi/selectors/calculator.js
+++ b/src/js/gi/selectors/calculator.js
@@ -787,6 +787,10 @@ export const getCalculatedBenefits = createSelector(
       return calculatedBenefits;
     }
 
+    const { militaryStatus } = eligibility;
+    const giBillChapter = +eligibility.giBillChapter;
+    const institutionType = institution.type.toLowerCase();
+
     calculatedBenefits.inputs = {
       inState: false,
       tuition: true,
@@ -821,7 +825,7 @@ export const getCalculatedBenefits = createSelector(
       },
       bookStipend: {
         visible: true,
-        value: formatCurrency(derived.bookStipendTotal)
+        value: `${formatCurrency(derived.bookStipendTotal)}${institutionType === 'ojt' ? '/mo' : ''}`
       },
       outOfPocketTuition: {
         visible: true,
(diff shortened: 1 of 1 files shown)


src/js/gi/selectors/calculator.js
```diff
@@ -1083,6 +1061,29 @@ export const getCalculatedBenefits = createSelector(
...
+        calendar: false,
+        working: true,
+      };
+
+      calculatedBenefits.outputs.tuitionAndFeesCharged.visible = false;
+      calculatedBenefits.outputs.giBillPaysToSchool.visible = false;
+      calculatedBenefits.outputs.yourScholarships.visible = false;
+      calculatedBenefits.outputs.outOfPocketTuition.visible = false;
+      calculatedBenefits.outputs.totalPaidToYou.visible = false;
+      calculatedBenefits.outputs.perTerm.tuitionAndFees.visible = false;
```
U-DON: Here, I'm mainly just moving the OJT "show or hide" logic block around so it takes precedence. Also added this missing line to hide the tuition fees section of the per-term breakdown.

src/js/gi/selectors/calculator.js
```diff
@@ -1083,6 +1061,29 @@ export const getCalculatedBenefits = createSelector(
...
     }
 
+    if (institutionType === 'ojt') {
+      calculatedBenefits.inputs = {
+        ...calculatedBenefits.inputs,
+        tuition: false,
+        books: false,
+        yellowRibbon: false,
+        scholarships: false,
+        tuitionAssist: false,
```
U-DON: Because `tuitionAssist` could previously be set to `true`, this will also now hide it entirely as long as the profile page is for OJT. It seems it should be correct behavior, as tuition shouldn't factor into the benefits for an employer, but awaiting confirmation from #5317.

melwoodard: @U-DON we're a-go on hiding this field on OJT.

cvalarida (approved):
Without knowing all the details of the GIBCT, I'm hard-pressed to come up with any improvements. Looks great!

PR #5325, opened 2017-04-13
Shot in the dark to fix intermittent test failure

omgitsbillryan:


omgitsbillryan:
I'm not sure how - but my build magically fixed itself without actually re-running :man_shrugging:

PR #5327, opened 2017-04-13
1990n Chapter 3 -- School Selection

cvalarida:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/633

The schema had education objective as well, so I put that in there even though it wasn't in the above ticket.

---
image


src/js/edu-benefits/pages/schoolSelection.js
```diff
@@ -4,21 +4,38 @@ import * as educationProgram from '../definitions/educationProgram';
...
-    'educationObjective',
-    'nonVaAssistance'
-  ],
-  required: [
-    // 'educationType',
-    // 'name'
-  ]
+
+// Exported like this so we can share the fields between the formConfig and the tests.
+export const schoolSelectionOptionsFor = {
```
cvalarida: I just liked the idea of having these here so we can use them in our tests. Not that we _really_ need them in there, I don't think, but...It's super convenient.

test/edu-benefits/pages/schoolSelection.unit.spec.jsx
```diff
@@ -1,141 +1,76 @@
...
+// Actually use the configs from the forms
+const configs = {
+  '1990n': formConfig1990n,
+  '1990e': formConfig1990e,
+  '5490': formConfig5490 // eslint-disable-line
+};
+
+// Describe the tests for each form
+Object.keys(optionsFor).forEach((formName) => {
+  describe(`Edu ${formName} schoolSelection`, () => {
```
cvalarida: Yeesh, this is painful to read as a git diff! Basically I got rid of the individual form tests in favor of looping over the `optionsFor` object to build the tests for each form in a loop.

test/edu-benefits/pages/schoolSelection.unit.spec.jsx
```diff
@@ -1,141 +1,76 @@
+import _ from 'lodash/fp';
+
 import React from 'react';
 import { findDOMNode } from 'react-dom';
 import { expect } from 'chai';
 import sinon from 'sinon';
 import ReactTestUtils from 'react-addons-test-utils';
 
 import { DefinitionTester, submitForm } from '../../util/schemaform-utils.jsx';
-import formConfig from '../../../src/js/edu-benefits/pages/schoolSelection.js';
```
cvalarida: I'm using the config from the form instead of the config from the page because 1) it seemed like we were testing the forms more than the page anyhow, and 2) I ran across an instance where, after refactoring, I called the page building function with a bad parameter. The test ran fine, so I thought it was golden, but when I went to take a screenshot for this PR, the page was completely blank. No JS errors, just blank.

cvalarida (commented):
I refactored some things because I'm a glutton for punishment.

src/js/edu-benefits/pages/schoolSelection.js
raquelromano-gov: This is way cleaner, in my opinion. I also like that defaults is gone.

test/edu-benefits/pages/schoolSelection.unit.spec.jsx
```diff
@@ -1,141 +1,76 @@
...
+            data={{}}
+            definitions={configs[formName].defaultDefinitions}
+            uiSchema={uiSchema}/>
+      );
+
+      const formDOM = findDOMNode(form);
+
+      const inputs = Array.from(formDOM.querySelectorAll('input, select, textarea'));
+
+      // Should have a better test than this...
```
raquelromano-gov: You could define an object above and hard-code the expected number of inputs for each form, and use that here.

test/edu-benefits/pages/schoolSelection.unit.spec.jsx
```diff
@@ -1,141 +1,76 @@
...
-    expect(Array.from(formDOM.querySelectorAll('.usa-input-error'))).to.have.lengthOf(1);
-    expect(onSubmit.called).to.be.false;
-
-    ReactTestUtils.Simulate.change(find('#root_educationProgram_educationType'), {
-      target: {
-        value: 'college'
+    // Either test for required inputs or not, depending on the required option
+    const requiredInputs = !_.isEmpty(optionsFor[formName].required);
+
+    it(`should have${requiredInputs ? '' : ' no'} required inputs`, () => {
```
raquelromano-gov: Lol. Sure, why not.

test/edu-benefits/pages/schoolSelection.unit.spec.jsx
```diff
@@ -1,141 +1,76 @@
...
-          data={{}}
-          definitions={formConfig.defaultDefinitions}
-          uiSchema={uiSchema}/>
-    );
-    const formDOM = findDOMNode(form);
-    submitForm(form);
-
-    expect(Array.from(formDOM.querySelectorAll('.usa-input-error'))).to.not.be.empty;
-    expect(onSubmit.called).to.not.be.true;
+    // TODO: Test filling in the required data
```
raquelromano-gov: Run out of steam? :)

raquelromano-gov (approved):

src/js/edu-benefits/pages/schoolSelection.js
jbalboni: I think defaults being gone is an improvement at this point, but it bugs me that the configuration is in this file instead of somewhere in the folder for the specific form. It forces you to come here, in our common edu code, to see form specific details. Those details don't seem like they should be the concern of the common page code.

test/edu-benefits/pages/schoolSelection.unit.spec.jsx
jbalboni: That makes sense to me given how you've organized the config for this page for the different forms (though like I said above, I'm not sure how I feel about that).

src/js/edu-benefits/pages/schoolSelection.js
raquelromano-gov: Hm...good point about form-specific details living outside of the form's folder.

test/edu-benefits/pages/schoolSelection.unit.spec.jsx
```diff
@@ -1,141 +1,76 @@
...
+  '1990n': formConfig1990n,
+  '1990e': formConfig1990e,
+  '5490': formConfig5490 // eslint-disable-line
+};
+
+// Describe the tests for each form
+Object.keys(optionsFor).forEach((formName) => {
+  describe(`Edu ${formName} schoolSelection`, () => {
+    // This is a bit fragile; requires the chapter and page name to be schoolSelection
+    const { schema, uiSchema } = configs[formName].chapters.schoolSelection.pages.schoolSelection;
```
raquelromano-gov: Same here as below--maybe we can keep this less brittle by defining an object above that describes the config details of each form to be tested (chapter and page name corresponding to this shared page code, number of expected inputs, etc.) so that when things change, we just update them up at the top.

test/edu-benefits/pages/schoolSelection.unit.spec.jsx
cvalarida: Hehe, yes. I didn't really want to write the code to fill in _every_ field (if required, of course). Probably should at some point if we keep using this, though.

PR #5330, opened 2017-04-14
Resolved missing data in header summary of GIBCT profiles.

U-DON:
Resolves department-of-veterans-affairs/vets.gov-team#2100.

Two issues being fixed here. First is that `highestDegree` values that weren't `2` or `4` weren't being handled, and it could have a value of `'Certificate'`. I generalized it to handle any non-empty values.

Second is that we need to show "Unknown size" if the institution has a `null` value for `undergradEnrollment`. It wasn't showing anything before if the attribute was unavailable.

> screen shot 2017-04-13 at 10 16 48 pm


melwoodard (approved):
LGTM!

cvalarida (approved):

PR #5333, opened 2017-04-14
Add accordions to FSGLI tables

melwoodard:
Embeds FSGLI insurance coverage tables within accordions


melwoodard:
This is probably me being a noob, but I realized the issue here is that the content pages don't import javascript. I tried importing .js from USWDS as a test, and the accordions function now: https://vetsgov-pr-5333.herokuapp.com/life-insurance/options-and-eligibility/fsgli/

@tiffanybbrown-va is there a more eloquent way to handle this?

tiffanybbrown-va:
I think this is fine for now. Ideally we'd include it in the header, but I think the accordion is currently the only USWDS pattern we're using (that will change soon, and we can revisit it then.)

melwoodard:
Got it, that makes sense. I'll add a TODO to revisit later.

PR #5334, opened 2017-04-14
Avoid double prorating kicker benefit in GIBCT calculation of monthly rate.

U-DON:
Resolves department-of-veterans-affairs/vets.gov-team#2154.

While kicker benefit gets prorated with different rates of pursuit (`ropOjt`, `ropOld`, `rop`) depending on various factors, `monthlyRate` and `buyUpRate` are always going to be prorated with `ropOld`.

I'm not sure if those two values should also apply a different rate of pursuit, but at least this change will ensure that `kickerBenefit` doesn't factor in `ropOld` on top of the rate of pursuit it was prorated by in the earlier calculation.

### DIFF
--- a/src/js/gi/selectors/calculator.js
+++ b/src/js/gi/selectors/calculator.js
@@ -274,7 +274,7 @@ const getDerivedValues = createSelector(
     }
 
     // Calculate Housing Allowance Rate Final - getMonthlyRateFinal
-    const monthlyRateFinal = ropOld * (monthlyRate + buyUpRate + kickerBenefit); // TODO: double check order of operations
+    const monthlyRateFinal = ropOld * (monthlyRate + buyUpRate) + kickerBenefit;
 
     // Calculate the names of Terms 1-4
     if (isOJT) {


patrickvinograd (approved):
LGTM.

saintsoup52:
Thank you for catching this...  my question is why are we prorating the kicker amount before we get to this point?  Shouldn't it be the other way around. 

Seems to me that we should be keeping the living allowance rate formulas in tack (like the original code) and just be passing the inputted kicker value and not a prorated value. For example we don't prorate the living allowance rate, buy-up or books in this way.

U-DON:
@saintsoup52 I don't know, that's how it was calculated in the old version of the tool.

https://github.com/department-of-veterans-affairs/gi-bill-comparison-tool/blob/master/app/assets/js/calculator.js#L1191

saintsoup52:
Now that I look at this closer... This actually breaks the living allowance calculations throughout because we are double counting kickerbenefit in each of the living allowance calculations.

I will go through and make proposed changes to fix this.

saintsoup52:
See #5387

PR #5339, opened 2017-04-14
Add 508 checks to 1990e e2e test

raquelromano-gov:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2160
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1839

@jkassemi 
We're trying to reduce redundancy in our front-end tests. This change just puts the aXe checks directly into our e2e tests, and it seems to pass on Jenkins, even though not using the `accessibility` settings in `config/nightwatch.js`

It looks like the significant difference between the `accessibility` and default settings (other than the directory filter) is the global `asyncHookTimeout` property, presumably because the axe checks are slow, but I'm not seeing any failures so maybe not always necessary?

Can you check this over and let us know if there are any other reasons not to go this direction and begin deleting some of our 508 tests in tests/accessibility ?


jkassemi:
@raquelromano-gov -the `asyncHookTimeout` property was necessary for the sitemap test. Within the context of a single test it will iterate over many urls and run an axe check against each. One test originally ran through every url in the sitemap, but it's since been (manually) split into 4 test files to allow nightwatch to run with multiple workers. 

The only accessibility tests required will likely be those that perform this iteration over the sitemap - since there's no attached react application with any associated tests for much of our content.

raquelromano-gov:
OK, so it sounds like there's no reason for any of our React apps with e2e tests to be using this environment, assuming they are relatively well-contained, e.g. run through one form, which probably has O(10) URLs. (Though I'm still not 100% clear on why running through many URLs would slow them down and require the special timeout global.)

In any case, would it make sense for us to move everything except `sitemap-*spec.js` out of `test/accessibility`? If so, I will check with those who maintain the various other React apps (rx, messaging, disability-benefits, etc.) and propose we all follow suit and when possible just run our aXe checks as part of our e2e tests.

jkassemi:
@raquelromano-gov - yes, I think moving everything that's not sitemap-*spec.js out of `test/accessibility` makes sense. 

As for running through the URLs, the time that it took to perform the iteration in https://github.com/department-of-veterans-affairs/vets-website/blob/master/test/accessibility/sitemap-1.spec.js#L15 was greater than the default `asyncHookTimeout` (10s).

jkassemi (approved):

PR #5340, opened 2017-04-14
Clean up unit/e2e test output

jbalboni:
Connects to [2166](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2166).

This cleans up color coding and removes some of the debug output in the test out. I also started cleaning up warnings in the Mocha tests.

I could not figure out how to change the reporter for Nightwatch, so that is still the same.


test/e2e/e2e-helpers.js
```diff
@@ -32,7 +32,7 @@ function overrideVetsGovApi(client) {
   [`http://localhost:${process.env.API_PORT || 3000}`],
   (val) => {
     // eslint-disable-next-line no-console
-    console.log(`Result of overriding VetsGov.api.url${JSON.stringify(val)}`);
+    // console.log(`Result of overriding VetsGov.api.url${JSON.stringify(val)}`);
```
raquelromano-gov: Can we just remove these if unwanted instead of commenting out?

src/test-support/mockapi.js
```diff
@@ -40,7 +40,7 @@ function makeMockApiRouter(opts) {
     const verb = (req.body.verb || 'get').toLowerCase();
     const path = stripTrailingSlash(req.body.path);
 
-    opts.logger.info(`mock: ${auth} ${verb} ${path}`);
+    opts.logger.verbose(`mock: ${auth} ${verb} ${path}`);
```
raquelromano-gov: We want higher logging than info?

raquelromano-gov (commented):
Thanks for tracking all of these down.

config/nightwatch.js
```diff
@@ -15,7 +15,7 @@ module.exports = {
   custom_commands_path: './test/util/nightwatch-commands',
   live_output: true,
   parallel_process_delay: 10,
-  disable_colors: false,
+  disable_colors: true,
```
cvalarida: It's probably asking too much for this to only be true on Jenkins, eh? So...like if the buildtype is production or some such.

cvalarida (approved):
Wow, that was a lot to hunt down. Thanks!

config/nightwatch.js
jbalboni: No, that's definitely possible. I didn't do that because I was going to ask if people were attached to color output locally.

src/test-support/mockapi.js
jbalboni: The level that's currently being logged is `info`, so moving these up a level removes them from the default output. But if someone wants to turn the messages back on (which can be genuinely helpful for debugging), they can change the level.

config/nightwatch.js
cvalarida: I'm fond of it locally because it makes for a nice visual breakup when we have _so much_ output. It's just a pain on Jenkins because it doesn't work.

src/test-support/mockapi.js
raquelromano-gov: Oh, right! :)  Thanks.

PR #5341, opened 2017-04-14
Remove greater than slow timeouts

annekainicUSDS:
Closes https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2164

I tested all of the e2e tests using just the `Timeouts.slow` for the submissions and they all passed fine. 

Also fixed a few failing 5490 e2e tests.


test/edu-benefits/5490/e2e-test.js
```diff
@@ -146,7 +146,7 @@ const runTest = (client, onlyRequiredFields) => {
     .click('input[type="checkbox"]')
     .click('.form-progress-buttons .usa-button-primary');
   client.expect.element('.js-test-location').attribute('data-location')
-    .to.not.contain('/review-and-submit').before(Timeouts.submission);
+    .to.not.contain('/review-and-submit').before(Timeouts.slow);
```
raquelromano-gov: Are we OK to remove the BUILDTYPE condition below now that the 5490 is live?

annekainicUSDS: Yes I think so! I will make that update

test/edu-benefits/5490/e2e-test.js
```diff
@@ -155,11 +155,4 @@ const runTest = (client, onlyRequiredFields) => {
...
 
-// Prevent the test from being run on production.
-// TODO: Remove when the form is actually launched.
-if (!process.env.BUILDTYPE || process.env.BUILDTYPE === 'development') {
-  module.exports = runTest;
-} else {
-  // Empty function to skip the tests
-  module.exports = (client) => client.end();
-}
+module.exports = runTest;
```
annekainicUSDS: @raquelromano-gov let me know if this looks correct to you.

cvalarida (commented):
So the idea behind this is to fail more quickly if it's going to fail, right? Because those wait times are a `wait up to X ms` kind of thing, this doesn't speed up successes as much as it speeds up the fallback on failures.

Looks good for that! If we find we have mysterious failures on Jenkins where some components don't render but they should, we may have to bump up the time on these.

test/edu-benefits/5490/e2e-test.js
raquelromano-gov: Looks right, though just realized this file will go away with https://github.com/department-of-veterans-affairs/vets-website/pull/5342/files since we're eliminating the 'minimal' e2e tests.

raquelromano-gov (approved):

raquelromano-gov:
Does this also close https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1878?
Looks like a duplicate issue

cvalarida:
@raquelromano-gov That ticket has a broader scope of just speeding up the tests. This is one solution for sure, but I don't know that it closes it necessary.

annekainicUSDS:
Yeah @cvalarida and @raquelromano-gov I think this PR starts with at least cleaning up the very slowest timeouts we are using, but I agree there's probably more we can do there! Going to merge this for now, then.

PR #5343, opened 2017-04-17
Added Analytics Events for Login and Verify

omgitsbillryan:
**Related Issue**
https://github.com/department-of-veterans-affairs/vets.gov-team/issues/595

Added `login-prompt-displayed` & `verify-prompt-displayed` so these events may be tracked in google analytics.  

The logic for determining what to display inside `RequiredLoginView` is written into the `render()` function.  The problem with firing the event from within `render()` is that it's called repeatedly throughout the React lifecycle.  As a result, I decided to create two new components: `LoginPrompt` and `VerifyPrompt` and fire the events from within the respective `componentDidMount()` functions.

### DIFF
--- a/src/js/common/components/RequiredLoginView.jsx
+++ b/src/js/common/components/RequiredLoginView.jsx
@@ -1,17 +1,14 @@
 import React from 'react';
 
-import { handleVerify } from '../helpers/login-helpers.js';
-
 import SystemDownView from './SystemDownView';
+import LoginPrompt from './authentication/LoginPrompt';
+import VerifyPrompt from './authentication/VerifyPrompt';
 import LoadingIndicator from '../../common/components/LoadingIndicator';
 
 class RequiredLoginView extends React.Component {
   constructor(props) {
     super(props);
-    this.handleLogin = this.handleLogin.bind(this);
-    this.handleSignup = this.handleSignup.bind(this);
     this.setUserLevel = this.setUserLevel.bind(this);
-    this.handleVerify = handleVerify;
     this.state = {
       loading: true
     };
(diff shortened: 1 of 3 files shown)


src/js/common/components/RequiredLoginView.jsx
```diff
@@ -39,61 +36,11 @@ class RequiredLoginView extends React.Component {
...
-  handleSignup() {
-    const myLoginUrl = this.props.loginUrl;
-    const receiver = window.open(`${myLoginUrl}&op=signup`, '_blank', 'resizable=yes,scrollbars=1,top=50,left=500,width=500,height=750');
-    receiver.focus();
-  }
-
   render() {
     let view;
 
-    const loginComponent = (
```
omgitsbillryan: This moved to LoginPrompt.jsx

src/js/common/components/RequiredLoginView.jsx
```diff
@@ -39,61 +36,11 @@ class RequiredLoginView extends React.Component {
...
-            <p>
-              <button className="usa-button-primary va-button-primary usa-button-big" onClick={this.handleLogin}><strong>Sign In</strong></button>
-              <button className="va-button-secondary usa-button-big" onClick={this.handleSignup}><strong>Create Account</strong></button>
-            </p>
-          </div>
-        </div>
-      </div>
-    );
-
-    const verifyComponent = (
```
omgitsbillryan: This moved to VerifyPrompt.jsx

jbalboni (approved):
This looks good to me. I think this makes RequiredLoginView more understandable, too. I tested it briefly with claim status, might be good to have someone test the Kudos apps, too (if you haven't).

If we end doing this a lot, there are ways to make higher order components that can wrap any component to make analytics calls. But that's not necessary here, I don't think.

omgitsbillryan:
@jbalboni thank you for the review.  The need for manual tests from Kudos makes me think that test coverage is not adequate.  Is there merit to adding e2e tests for `RequiredLoginView`?

jbalboni:
@omgitsbillryan It looks like there's one e2e test for login. It definitely wouldn't hurt to have more. Could also unit test the analytics events.

I'm pretty paranoid about login, since it seems very complicated, but I haven't really worked on it. So if there are people on the core team who are more confident we don't need more tests, I'm happy to defer to that.

omgitsbillryan:
Adding @robertfairhead to see the two new event tags: `login-prompt-displayed` & `verify-prompt-displayed`.  `login-prompt-displayed` gets fired when this is shown:

image

`verify-prompt-displayed` gets fired when the "Verify Your Account" page which I'm having trouble making show up right now.

robertfairhead:
@omgitsbillryan Tag Manager is configured to catch these events and send on to the All Vets.gov Site Data property. I also enabled it for the testing property (which will allow events from staging, dev, etc.) Once this is merged I'll check on the testing property to be sure they're showing up.

jcmeloni-usds:
@omgitsbillryan to make the Verify Your Account page show up, you have to login as an LOA1 only and then hit a gated endpoint.

omgitsbillryan:
Alright, @bshyong was kind enough to help me out with some manual testing and all looks good!

robertfairhead:
screen shot 2017-04-18 at 7 04 12 pm

Verified events are getting picked up by GA in the test environment so we should be good for production deploy as well.

PR #5351, opened 2017-04-18
1990N CH 5: Review/Submit

jbalboni:
Connects to [673](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/673) and [2240](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2240)

This updates the confirmation page for 1990n. It also fixes a bug on the review page where required array fields prevent saving.

screen shot 2017-04-18 at 10 32 52 am

### DIFF
--- a/src/js/common/schemaform/helpers.js
+++ b/src/js/common/schemaform/helpers.js
@@ -305,7 +305,7 @@ export function getNonArraySchema(schema) {
     if (newProperties !== schema.properties) {
       let newSchema = _.set('properties', newProperties, schema);
       if (newSchema.required) {
-        const newRequired = _.intersection(newSchema.required, newSchema.properties);
+        const newRequired = _.intersection(Object.keys(newSchema.properties), newSchema.required);
         if (newRequired.length !== newSchema.required.length) {
           newSchema = _.set('required', newRequired, newSchema);
         }
--- a/src/js/common/schemaform/helpers.js
+++ b/src/js/common/schemaform/helpers.js
@@ -303,7 +303,15 @@ export function getNonArraySchema(schema) {
     }
 
     if (newProperties !== schema.properties) {
-      return _.set('properties', newProperties, schema);
+      let newSchema = _.set('properties', newProperties, schema);
+      if (newSchema.required) {
+        const newRequired = _.intersection(newSchema.required, newSchema.properties);
+        if (newRequired.length !== newSchema.required.length) {
+          newSchema = _.set('required', newRequired, newSchema);
+        }
+      }
+
+      return newSchema;
     }
   }
 
--- a/src/js/edu-benefits/1990n/containers/ConfirmationPage.jsx
+++ b/src/js/edu-benefits/1990n/containers/ConfirmationPage.jsx
@@ -5,10 +5,6 @@ import Scroll from 'react-scroll';
 
 import { focusElement } from '../../../common/utils/helpers';
 
-import {
-  benefitsLabels
-} from '../../utils/helpers';
-
 const scroller = Scroll.scroller;
 const scrollToTop = () => {
   scroller.scrollTo('topScrollElement', {
@@ -19,20 +15,9 @@ const scrollToTop = () => {
 };
 
 class ConfirmationPage extends React.Component {
-  constructor(props) {
-    super(props);
-    this.state = { isExpanded: false };
-    this.handleClick = this.handleClick.bind(this);
-  }
-
   componentDidMount() {
     focusElement('.edu-page-title');
     scrollToTop();
-  }
-
-  handleClick(e) {
-    e.preventDefault();
-    this.setState({ isExpanded: !this.state.isExpanded });
   }
 
   render() {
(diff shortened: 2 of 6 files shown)


test/edu-benefits/5495/containers/ConfirmationPage.unit.spec.jsx
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
```
jbalboni: Forgot to check this file in in my last PR.

src/js/common/schemaform/helpers.js
```diff
@@ -303,7 +303,15 @@ export function getNonArraySchema(schema) {
     }
 
     if (newProperties !== schema.properties) {
-      return _.set('properties', newProperties, schema);
+      let newSchema = _.set('properties', newProperties, schema);
+      if (newSchema.required) {
+        const newRequired = _.intersection(Object.keys(newSchema.properties), newSchema.required);
+        if (newRequired.length !== newSchema.required.length) {
+          newSchema = _.set('required', newRequired, newSchema);
```
cvalarida: So...if these fields were required, but somehow they have no information in them, stop requiring them? Will this cause submission errors to the api?

src/js/edu-benefits/1990n/containers/ConfirmationPage.jsx
```diff
@@ -52,15 +36,11 @@ class ConfirmationPage extends React.Component {
...
         </p>
         <div className="inset">
-          <h4>Education Benefit Claim <span className="additional">(Form 22-1990e)</span></h4>
+          <h4>Education Benefit Claim <span className="additional">(Form 22-1990n)</span></h4>
           <span>for {name.first} {name.middle} {name.last} {name.suffix}</span>
 
           <ul className="claim-list">
             <li>
-              <strong>Selected benefit</strong><br/>
-              {benefitsLabels[benefit]}
```
cvalarida: Why is this removed?

src/js/edu-benefits/1990n/containers/ConfirmationPage.jsx
```diff
@@ -52,15 +36,11 @@ class ConfirmationPage extends React.Component {
           <i>Please print this page for your records.</i>
         </p>
         <div className="inset">
-          <h4>Education Benefit Claim <span className="additional">(Form 22-1990e)</span></h4>
+          <h4>Education Benefit Claim <span className="additional">(Form 22-1990n)</span></h4>
```
cvalarida: Heh, good catch!

test/edu-benefits/1990n/containers/ConfirmationPage.unit.spec.jsx
```diff
@@ -0,0 +1,37 @@
...
+};
+
+describe('Edu 1990n <ConfirmationPage>', () => {
+  it('should render', () => {
+    const tree = SkinDeep.shallowRender(
+      <ConfirmationPage form={form}/>
+    );
+
+    expect(tree.subTree('.edu-page-title').text()).to.equal('Claim received');
+    expect(tree.everySubTree('span')[1].text().trim()).to.equal('for Jane Doe');
```
cvalarida: I don't suppose there's a better way to find this element, is there?

cvalarida (commented):
It looks good overall, but I'm concerned about removing the requirement on required fields with no data.

src/js/common/schemaform/helpers.js
jbalboni: This is due to how we deal with array fields on the review page. We pull them out of the page they're in and display them separately. Before, we were doing that but not removing the array fields from the required array, which caused the phantom validation error you saw. I won't cause any submission errors because we still validate everything before sending when you click submit. This change doesn't get saved anywhere, it's a temporary schema that's only used in the components on the review page.

It is another edge case, though, when you have an array field that requires at least one item. It's not really clear that the user has to enter an item if they come to the review page without one already. I opened an issue for us to think about simplifying how we handle array fields on the review page.

src/js/edu-benefits/1990n/containers/ConfirmationPage.jsx
jbalboni: You don't select an actual benefit on this form, just check a box saying pick the best one for me.

test/edu-benefits/1990n/containers/ConfirmationPage.unit.spec.jsx
jbalboni: We could probably put a class on it, but otherwise, I don't think so. Though I just copied this from an existing test and haven't take another look at it.

src/js/common/schemaform/helpers.js
cvalarida: So this will just let us update the page when reviewing, but will still fail validation when submitting?

jbalboni: @cvalarida Correct.

PR #5358, opened 2017-04-18
Fix search results error

bshyong:
Ran into this error while working on https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2098

Search results need to wait until constants are loaded before they can render

Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2259


U-DON:
Since constants are being fetched at the highest level (`GiBillApp` container), should there be a loading indicator in the render at that level instead?

bshyong:
@U-DON hmm that would blank out the entire UI right? how do you think about moving it up to the `GiBillApp` vs.  the SearchPage component using the existing indicator? 

```
    if (search.inProgress) {
      searchResults = (
        <div className={resultsClass}>
          {filterButton}
          <LoadingIndicator message="Loading search results..."/>
        </div>
      );
```

U-DON:
I think that would work too. But another case to consider is the profile page, since it also depends on the constants for calculations.

bshyong:
^ moved to top-level GIBillApp component

src/js/gi/components/search/SearchResult.jsx
```diff
@@ -92,7 +92,6 @@ export class SearchResult extends React.Component {
       </div>
     );
   }
-
```
U-DON: Super nitpick: if removing this newline at the end of the class, can also remove the one at the top.

U-DON (approved):
LGTM.

PR #5359, opened 2017-04-18
5495 intro change

cvalarida:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2245

image

I replaced the civilian life transition text with what was in the 1990n per @jbalboni's suggestion. Does that look right, @J-Quag?
image


jbalboni (approved):
Looks good, I made a comment about what we should do for that third point on the GH issue.

J-Quag:
LGTM 👍

jbalboni (approved):

PR #5360, opened 2017-04-18
Scaffolding for hca rjsf form

raquelromano-gov:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2220

### DIFF
--- a/src/js/hca-rjsf/hca-rjsf-entry.jsx
+++ b/src/js/hca-rjsf/hca-rjsf-entry.jsx
@@ -6,7 +6,6 @@ import { Router, useRouterHistory } from 'react-router';
 import { Provider } from 'react-redux';
 import { createStore } from 'redux';
 
-// import HealthCareApp from './components/HealthCareApp.jsx';
 import initReact from '../common/init-react';
 import createRoutes from './routes';
 import reducer from './reducer';
@@ -19,7 +18,7 @@ require('../login/login-entry.jsx');
 const store = createStore(reducer);
 
 const browserHistory = useRouterHistory(createHistory)({
-  basename: '/healthcare/rjsf/apply/application'
+  basename: '/healthcare/rjsf'
 });
 
 function init() {
--- a/src/js/hca-rjsf/hca-rjsf-entry.jsx
+++ b/src/js/hca-rjsf/hca-rjsf-entry.jsx
@@ -9,7 +9,7 @@ import { createStore } from 'redux';
 // import HealthCareApp from './components/HealthCareApp.jsx';
 import initReact from '../common/init-react';
 import createRoutes from './routes';
-import reducer from './reducers';
+import reducer from './reducer';
 
 require('../common');
 require('../../sass/hca.scss');
--- a/src/js/hca-rjsf/hca-rjsf-entry.jsx
+++ b/src/js/hca-rjsf/hca-rjsf-entry.jsx
@@ -1,0 +1,36 @@
+import 'core-js';
+import React from 'react';
+import ReactDOM from 'react-dom';
+import { createHistory } from 'history';
+import { Router, useRouterHistory } from 'react-router';
+import { Provider } from 'react-redux';
+import { createStore } from 'redux';
+
+// import HealthCareApp from './components/HealthCareApp.jsx';
+import initReact from '../common/init-react';
+import createRoutes from './routes';
+import reducer from './reducers';
+
+require('../common');
+require('../../sass/hca.scss');
+
+require('../login/login-entry.jsx');
+
+const store = createStore(reducer);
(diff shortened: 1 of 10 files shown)


src/js/hca-rjsf/hca-rjsf-entry.jsx
```diff
@@ -0,0 +1,35 @@
...
+import initReact from '../common/init-react';
+import createRoutes from './routes';
+import reducer from './reducer';
+
+require('../common');
+require('../../sass/hca.scss');
+
+require('../login/login-entry.jsx');
+
+const store = createStore(reducer);
```
cvalarida: May want to apply the middleware like the edu forms do:

```js
let store;
if (__BUILDTYPE__ === 'development' && window.devToolsExtension) {
  store = createStore(reducer, compose(applyMiddleware(thunk), window.devToolsExtension()));
} else {
  store = createStore(reducer, compose(applyMiddleware(thunk)));
}
```

I don't know if `redux-thunk` is necessary, strictly speaking, with the schemaform, but maybe @jbalboni can weigh in on that. At any rate, the react dev tools might be nice.

src/js/hca-rjsf/helpers.jsx
```diff
@@ -0,0 +1,10 @@
+import { transformForSubmit } from '../common/schemaform/helpers';
+
+export function transform(formConfig, form) {
+  const formData = transformForSubmit(formConfig, form);
+  return JSON.stringify({
+    educationBenefitsClaim: {
```
cvalarida: Mmm....not sure what this should be, but I'm guessing probably not `educationBenefitsClaim`.

cvalarida (commented):
Assuming we don't _need_ `redux-thunk`, it looks good except for the transform function.

src/js/hca-rjsf/hca-rjsf-entry.jsx
raquelromano-gov: Oh yeah--don't know if hca uses thunk. @annekainicUSDS ? Added it for now

src/js/hca-rjsf/helpers.jsx
raquelromano-gov: Taking a wild guess that it has something to do with what the backend will expect, though I'm not seeing it in https://github.com/department-of-veterans-affairs/vets-api/tree/master/docs/api . @annekainicUSDS, is that because we're not yet using the vets-api backend?

src/js/hca-rjsf/hca-rjsf-entry.jsx
```diff
@@ -0,0 +1,42 @@
+import 'core-js';
```
jbalboni: Oh, weird, I didn't realize this was in the entry files. We shouldn't need it.

src/js/hca-rjsf/routes.jsx
```diff
@@ -0,0 +1,18 @@
+import { createRoutes as createFormRoutes } from '../common/schemaform/helpers';
+import reducer from './reducer';
+import formConfig from './config/form';
+import HealthCareApp from './HealthCareApp.jsx';
+
+export default function createRoutes(store) {
```
jbalboni: We can simplify this file. The `createRoutes` function was created so that we could replace the Redux store whenever a different form was loaded. Since this app has just one form, it's not necessary. You can do:

```js
const route = {
  path: '/',
  component: HealthCareApp,
  indexRoute: { onEnter: (nextState, replace) => replace('/introduction') },
  childRoutes: createFormRoutes(formConfig)
};

export default route;
```

src/js/hca-rjsf/hca-rjsf-entry.jsx
jbalboni: We use redux-thunk whenever we make an api call, so it will use it eventually.

src/js/hca-rjsf/helpers.jsx
jbalboni: I think it's just called `form` for hca: https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/js/hca/components/HealthCareApp.jsx#L147

jbalboni (approved):
Made some comments, but looks good to me.

src/js/hca-rjsf/helpers.jsx
cvalarida: Ah, I looked for that, @jbalboni but didn't find it. Thanks!

src/js/hca-rjsf/routes.jsx
raquelromano-gov: Oh thanks--meant to ask you about this bit. Fixed

PR #5362, opened 2017-04-19
Add schema tests to 1990n

cvalarida:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2256

### DIFF
--- a/test/edu-benefits/1990n/schema/schema.unit.spec.js
+++ b/test/edu-benefits/1990n/schema/schema.unit.spec.js
@@ -1,0 +1,46 @@
+import { expect } from 'chai';
+import fs from 'fs';
+import path from 'path';
+import { Validator } from 'jsonschema';
+
+import { transform } from '../../../../src/js/edu-benefits/1990n/helpers';
+import formConfig from '../../../../src/js/edu-benefits/1990n/config/form';
+import fullSchema1990n from 'vets-json-schema/dist/22-1990N-schema.json';
+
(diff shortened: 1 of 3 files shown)


test/edu-benefits/1990n/schema/schema.unit.spec.js
```diff
@@ -0,0 +1,46 @@
...
+import { transform } from '../../../../src/js/edu-benefits/1990n/helpers';
+import formConfig from '../../../../src/js/edu-benefits/1990n/config/form';
+import fullSchema1990n from 'vets-json-schema/dist/22-1990N-schema.json';
+
+describe('1990n schema tests', () => {
+  const v = new Validator();
+  const files = fs.readdirSync(__dirname);
+  let expectedDataFiles;
+  try {
+    expectedDataFiles = fs.readdirSync(path.join(__dirname, 'expected-data'));
```
jbalboni: I just deleted this code from the 5495 test. This works, too.

jbalboni (approved):

test/edu-benefits/1990n/schema/schema.unit.spec.js
cvalarida: Yeah, I thought about deleting it. We probably won't need it, but I was just thinking if we ever copied the test over to somewhere that needed it...

Would you recommend deleting it? I'm on the fence.

PR #5364, opened 2017-04-19
Update scroll handling on schemaform pages

jbalboni:
Connects to [1778](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1778)

Adds back scrolling on FormPage and sets scroll restoration for manual to keep Chrome/FF from jumping to top of page and scrolling down.


cvalarida (approved):
Running OS X, I wasn't able to test it on IE, but it works great on Chrome and Firefox.

cvalarida:
Any idea what's up with that failed e2e test?

src/js/common/schemaform/FormPage.jsx
```diff
@@ -17,6 +18,15 @@ function focusForm() {
   }
 }
 
+const scroller = Scroll.scroller;
+const scrollToTop = () => {
+  scroller.scrollTo('topScrollElement', window.VetsGov.scroll || {
```
cvalarida: So...documentation isn't a silver bullet or anything--especially when you amass too much of it--but I wonder if this is the type of thing that we should have written somewhere.

"`window.VetsGov` is a thing! This is what it does."

jbalboni: Yeah. I didn't realize it was still used for things. It used to be used to swap urls out for e2e tests, but I think we replaced that with the environment.js file.

PR #5365, opened 2017-04-19
Rx refills 508 fix and test cleanup

raquelromano-gov:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2203

See the issue description for motivation behind removing tests from `test/accessibility` and adding aXe checks into e2e tests. Verified with James that these are OK to run under the default nightwatch configuration on jenkins.

This uncovered a minor 508 issue (not surprising since e2e tests generally seem more detailed and up-to-date across all our apps).


src/js/common/components/AlertBox.jsx
```diff
@@ -45,8 +45,8 @@ class AlertBox extends React.Component {
     let closeButton;
     if (this.props.onCloseAlert) {
       closeButton = (
-        <button className="va-alert-close" onClick={this.props.onCloseAlert}>
-          <i className="fa fa-close"></i>
+        <button className="va-alert-close" aria-label="Close notification" onClick={this.props.onCloseAlert}>
```
raquelromano-gov: There are several places in vets.gov where we create our own "va-alert-close" buttons instead of using this component for some reason (maybe that should change), but took this label from one of them: https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/js/disability-benefits/components/Notification.jsx

Oddly, the 508 checks don't complain about this button even though it's missing an aria-label:
https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/js/rx/components/Disclaimer.jsx#L28
so I left it alone.

bshyong (approved):
LGTM thanks for helping to clean these up!

PR #5367, opened 2017-04-19
Fix loading error on GIBCT profile page

bshyong:
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2263

In the initial render of the `ProfilePage` component, the profile attributes are empty, causing there to be an error. This PR adds a catch for the initial render.


U-DON (commented):
We should probably show an error message or something if it has attempted to load but got nothing. Good enough for now to get the tests passing.

PR #5370, opened 2017-04-19
Fix search term bug in GIBCT when pressing browser back button

bshyong:
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2098


U-DON (approved):
LGTM.

PR #5369, opened 2017-04-19
Add e2e and accessibility checks for 5495

jbalboni:
Connects to [2246](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2246)

This adds e2e and 508 tests. It also updates the 5490 test to fill in all fields, instead of just required fields.

PR #5371, opened 2017-04-19
1990n e2e test

cvalarida:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2248

Unfortunately, it doesn't finish. I'm not sure why, but it has trouble going from the review-and-submit page to the confirmation page.


test/edu-benefits/1990n/01-all-fields.e2e.spec.js
```diff
@@ -0,0 +1,95 @@
...
+    E2eHelpers.expectNavigateAwayFrom(client, '/personal-information/direct-deposit');
+
+    // Review and submit page
+    client
+      .waitForElementVisible('label[name="privacyAgreement-label"]', Timeouts.slow)
+      .pause(1000)
+      .click('input[type="checkbox"]')
+      .axeCheck('.main')
+      .click('.form-progress-buttons .usa-button-primary')
+      .expect.element('.js-test-location').attribute('data-location')
```
raquelromano-gov: Maybe add a timeout parameter to `E2eHelpers.expectNavigateAwayFrom()` that defaults to `Timeouts.normal`?

test/edu-benefits/1990n/01-all-fields.e2e.spec.js
```diff
@@ -0,0 +1,95 @@
...
+    client
+      .axeCheck('.main')
+      .click('.form-progress-buttons .usa-button-primary');
+    E2eHelpers.expectNavigateAwayFrom(client, '/personal-information/direct-deposit');
+
+    // Review and submit page
+    client
+      .waitForElementVisible('label[name="privacyAgreement-label"]', Timeouts.slow)
+      .pause(1000)
+      .click('input[type="checkbox"]')
```
raquelromano-gov: Weird that it's not progressing to the confirmation page. One thing to verify is that the privacy checkbox is indeed checked since it's the only thing being validated here that the other pages haven't already validated.

test/util/nightwatch-commands/fill.js
```diff
@@ -2,10 +2,15 @@
  * Clears the current value and if |value| is specified,
  *  enters the |value| in.
  */
-exports.command = function fill(selector, value) {
+exports.command = function fill(selector, value, callback) {
   this.clearValue(selector);
   if (typeof value !== 'undefined') {
-    this.setValue(selector, value);
+    // When callback is passed as undefined, setValue assumes selector is the locating strategy
```
raquelromano-gov: The comment is a little cryptic--was looking for an example of what sort of callback is passed in here in order to suggest something else, but I don't see it used anywhere.

raquelromano-gov (commented):
See below for one possible reason the submit isn't working, but if that's not it, I'm stumped as well.

jbalboni:
@cvalarida I updated this with the latest from master, it completes locally for me now. I think you were missing the last change I made to update the confirmation page.

test/e2e/edu-1990n-helpers.js
```diff
@@ -0,0 +1,16 @@
+// I really don't want to duplicate this, but the lack of view:veteranId really
```
jbalboni: I updated it in my PR to work with 5495, I think we can make it work here: https://github.com/department-of-veterans-affairs/vets-website/pull/5369/files#diff-74108ba937cb81f8e4d99f4fbb4803fc

jbalboni (approved):
LGTM

test/util/nightwatch-commands/fill.js
cvalarida: Yeah, it should work just fine according to the docs: http://nightwatchjs.org/api/#setValue
Unfortunately, it doesn't 🙁 It was throwing an error telling me that the location strategy of `input[name="whatever"]` is invalid and it must be [a whole list of things including `css selector`].

I would have thought that not passing the third param would act the same as passing the third param as undefined, but...I guess not so much.

The callback isn't used, but I wrote this bit while testing a way to fill either one element or another, but...that was just awful, so I abandoned it.

test/e2e/edu-1990n-helpers.js
cvalarida: I was trying to refactor the shared one first, but after much difficulty, I broke down and just copied it out. I tried making an `exists()` function for the client to fill in information if an element exists. I tried filling in the `view:veteranId` version first and if that fails, the field sans `view:veteranId`, but that was just all kinds of nasty (and broken, and left errors in the output).

I think @jbalboni's solution is the way to go. Because this didn't have the reference to `data['view:veteranId']`, I was assuming it was flattened...though in retrospect that wasn't a very astute assumption. 0.o That certainly makes life a lot less complicated!

cvalarida:
@jbalboni  I think what I'll do is wait for your tests for the 5495 to get merged into `master` and I'll pull them in here so I can reuse what you did for `completeVeteranInformation()`.

test/util/nightwatch-commands/fill.js
raquelromano-gov: That nightwatch documentation is confusing:
```
  browser.setValue('input[type=text]', ['nightwatch', browser.Keys.ENTER]);
```
I thought the callback was a 3rd parameter? Weird.

But anyway, if you're not using the callback, can this comment go away?

PR #5372, opened 2017-04-19
Remove unnecessary first-of-type rules

melwoodard:
First-of-type rules for lists are no longer needed now that we have the `.va-introtext` class to manually apply the style.


tiffanybbrown-va:
LGTM. Merging.

PR #5373, opened 2017-04-20
Additional Resources section in GIBCT profile pages

U-DON:
Resolves department-of-veterans-affairs/vets.gov-team#2253.

In addition to the new section in the header, I also tweaked some of the margins between the icon-with-info rows to be more consistent.

### DIFF
--- a/src/js/gi/components/content/VideoSidebar.jsx
+++ b/src/js/gi/components/content/VideoSidebar.jsx
@@ -12,7 +12,7 @@ class VideoSidebar extends React.Component {
         <iframe width="100%" src="https://www.youtube.com/embed/Z1ttkv9oRI4"
             title="Know Before You Go" frameBorder="0" allowFullScreen></iframe>
         <h5>Ready to apply?</h5>
-        <a href="https://www.vets.gov/education/apply/" target="_blank">
+        <a href="/education/apply/" target="_blank">
           Apply for education benefits
         </a>
       </div>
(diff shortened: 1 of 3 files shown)


melwoodard (approved):
LGTM

saintsoup52:
1. Are these additional resources showing up on the GI Bill CT landing page? That was the original intent with this change.

2. In the current version the additional resources disappear on a mobile device. I think we would be OK with that happening here too. Otherwise the additional resources delays users getting to the meat of the tool.

3. Is the video also showing up on the school profile page? I was not there at the Coy meeting but I think I heard that was his request.

melwoodard:
Answers to your questions @saintsoup52:

1. We already show the Apply link on the landing page, and can add the other two there
2. We'll make this change
3. The request to add the video back was only for the landing page. To Stephanie's point from our Tuesday meeting, the main reason for the links to exist on the profile page is to provide next steps for a Veteran is interested in that school, and the video is less critical to that - but I'll point out that it also exists on the page linked to from "Choosing a school". Also, if we were to add the video back it would cause the calculator section to be pushed down and a large amount of empty space between that and the section above it.

@U-DON can you address 1 and 2 above?

U-DON:
Done. No additional resources in mobile, so it will just be the icons with info rows and then the estimate section. Landing page sidebar will look like this:

> screen shot 2017-04-20 at 2 22 47 pm

melwoodard (approved):
Looks good, thanks!

bshyong (approved):
LGTM

PR #5376, opened 2017-04-20
Close GIBCT modals when navigating between pages.

U-DON:
Resolves department-of-veterans-affairs/vets.gov-team#2107.


bshyong (approved):
LGTM

PR #5378, opened 2017-04-20
[WIP] Disallow same sponsor and applicant SSN

cvalarida:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2255


docs/schemaform/form-config.md
```diff
@@ -180,22 +180,29 @@ We've also been adding some additional uiSchema functionality not found in the r
...
   'ui:validations': [
-    function (errors, fieldData, pageData, fieldSchema, errorMessages) {
+    /**
+     * Note the difference between the three data parameters:
+     *
+     * @param {any} fieldData The data for the current field being validated
+     * @param {object} pageData The data for all the fields on the current page
+     * @param {object} formData The data for all the fields in every page
+     */
+    function (errors, fieldData, pageData, formData fieldSchema, errorMessages) {
```
cvalarida: This is the only section that I actually modified in this file; the rest is Atom removing the whitespace at the end of lines.

test/edu-benefits/pages/contactInformation.unit.spec.jsx
```diff
@@ -53,7 +53,7 @@ describe('Edu pages contactInformation', () => {
     const formDOM = findDOMNode(form);
     submitForm(form);
 
-    expect(formDOM.querySelectorAll('.usa-input-error').length).to.equal(6);
+    expect(formDOM.querySelectorAll('.usa-input-error').length).to.equal(5);
```
cvalarida: I'll be honest with you; I have no idea why this worked before. The country is also a required field, but it's pre-populated with 'United States,' so that doesn't make much sense to me. There are 6 required fields, but 1 is filled in by default.

src/js/common/schemaform/review/ArrayField.jsx
```diff
@@ -164,7 +164,7 @@ class ArrayField extends React.Component {
                           ? <h5>New {uiSchema['ui:options'].itemName}</h5>
                           : null}
                       <SchemaForm
-                          data={item}
+                          pageData={item}
```
cvalarida: These may end up needing `formData` as well, but I'd rather not squash all data to get `formData` here as well...

docs/schemaform/form-config.md
raquelromano-gov: Ha, yeah took me a second to realize that you weren't obsessively going over documentation and removing unneeded whitespace :)

Isn't the formData also flattened whereas the pageData maintains the form config structure? May want to explain that if so.

Looks like you're missing a comma after formData

docs/schemaform/form-config.md
```diff
@@ -180,22 +180,29 @@ We've also been adding some additional uiSchema functionality not found in the r
...
       errors.addError('My error');
     },
     {
-      validator: (errors, fieldData, pageData, fieldSchema, errorMessages, options) => {
+      validator: (errors, fieldData, pageData, formData fieldSchema, errorMessages, options) => {
```
raquelromano-gov: comma missing here too

package.json
```diff
@@ -17,6 +17,7 @@
     "test:accessibility": "./script/run-nightwatch.sh --env accessibility",
     "test:unit": "mocha --recursive 'test/**/*.unit.spec.js?(x)' ",
     "test:coverage": "NODE_ENV=test nyc --all --reporter=lcov --reporter=text --reporter=html mocha --no-color --recursive 'test/**/*.unit.spec.js?(x)' ",
+    "test:unit:edu": "mocha --recursive 'test/edu-benefits/**/*.unit.spec.js?(x)' ",
```
raquelromano-gov: 👍

src/js/common/schemaform/validation.js
```diff
@@ -98,7 +98,7 @@ export function transformErrors(errors, uiSchema) {
  * should call addError to add the error.
  */
 
-export function uiSchemaValidate(errors, uiSchema, schema, formData, formContext, path = '') {
+export function uiSchemaValidate(errors, uiSchema, schema, pageData, formData, formContext, path = '') {
```
raquelromano-gov: Any interest in adding some parameter documentation here?

src/js/common/schemaform/SchemaForm.jsx
```diff
@@ -92,17 +92,18 @@ class SchemaForm extends React.Component {
...
-      uiSchemaValidate(errors, uiSchema, schema, formData);
+      uiSchemaValidate(errors, uiSchema, schema, pageData, formData);
     }
     return errors;
   }
 
   render() {
     const {
-      data,
+      pageData, // Unfortunately, Form calls this formData, but don't be confused,
```
raquelromano-gov: ugh, why? Can we change it to call it pageData?

src/js/common/schemaform/FormPage.jsx
```diff
@@ -79,17 +79,34 @@ class FormPage extends React.Component {
...
       uiSchema
-    } = this.props.form[route.pageConfig.pageKey];
+    } = form[route.pageConfig.pageKey];
+
+    // Flatten the data from every page to pass to SchemaForm to eventually be
+    //  used in uiSchemaValidate()
+    // I'd rather not have this here, but console.time() tells me it's executing
+    //  locally in 1.49e+12ms on the edu 5495, which is admittedly a relatively
+    //  small form, but it shouldn't be a significant hit at any rate.
+    const formData = Object.keys(form).reduce((carry, pageName) => {
```
raquelromano-gov: How about renaming `data` above to `pageData` so that there is no `data` variable.

Also, should we add a flatten helper and use it both here and in `validation.js`?

src/js/common/schemaform/SchemaForm.jsx
raquelromano-gov: Wait, I'm not seeing where that's true--what exactly are you referring to as Form here?

test/edu-benefits/pages/contactInformation.unit.spec.jsx
raquelromano-gov: This passed with 6 and now passes with 5?

docs/schemaform/form-config.md
cvalarida: OCD!
MUST REMOVE ALL TRAILING WHITESPACE!

The `formData` is basically all the pages' `data` contents slapped together in one object. That's all the more it's flattened. `pageData` is just the `data` on the current page, so they both have essentially the same structure. `formData` just has more`.

....bloody commas....should lint these .md files. >_>

test/edu-benefits/pages/contactInformation.unit.spec.jsx
cvalarida: Yup. No idea why.

src/js/common/schemaform/validation.js
cvalarida: Not even a little! But at this point, we probably should have more than just the schemaform documentation...Trouble is I don't really know what type `errors` is, what `formContext` is, nor what `path` is really used for, so... 😕

src/js/common/schemaform/SchemaForm.jsx
cvalarida: The Form from react-jsonschema-form, which means no, we can't rename the prop 😞

src/js/common/schemaform/FormPage.jsx
cvalarida: Ah, a flatten helper might be good. We may end up having to do this elsewhere as well.

src/js/common/schemaform/validation.js
jbalboni: I'll add some

src/js/edu-benefits/5495/config/form.js
```diff
@@ -104,6 +104,15 @@ const formConfig = {
               'view:noSSN': {
                 'ui:title': 'I don’t know my sponsor’s Social Security number',
               },
+              veteranSocialSecurityNumber: {
+                'ui:validations': [
+                  (errors, fieldData, pageData, formData) => {
+                    if (fieldData === formData.relativeSocialSecurityNumber) {
+                      errors.addError("Your sponsor's SSN cannot be the same as yours.");
```
jbalboni: Single quotes

src/js/common/schemaform/validation.js
```diff
@@ -96,9 +96,19 @@ export function transformErrors(errors, uiSchema) {
...
+ * @param {Object} pageData The data for the page the field is in
+ * @param {Object} formData The (flattened) data for the entire form
+ * @param {Object} formContext The formContext object passed to rjsf's Form component.
+ *   Should be removed at some point
+ * @param {String} path The path to the current field relative to the root of the page.
+ *   Used to select the correct field data to validate against
  */
 
-export function uiSchemaValidate(errors, uiSchema, schema, formData, formContext, path = '') {
+export function uiSchemaValidate(errors, uiSchema, schema, pageData, formData, formContext, path = '') {
```
cvalarida: I tried using just `formData` at first, but ran into issues...I don't remember what they were exactly, and it could be that they were unrelated, now that I think about it. I'll give it another shot without `pageData` after I get the tests all fixed up.

src/js/edu-benefits/5495/config/form.js
cvalarida: Lint didn't yell at me...Probably because I've got the single quote as an apostrophe for "sponsor's".

Fair enough.

src/js/common/schemaform/validation.js
```diff
@@ -96,6 +96,16 @@ export function transformErrors(errors, uiSchema) {
  *
  * The function is passed errors, fieldData, formData, and otherData and
```
cvalarida: The function is passed errors, fieldData, pageData, formData, and otherData (wha..?)

src/js/common/schemaform/FormPage.jsx
cvalarida: Schemaform helpers.js has `flattenFormData()` which may work, but I'll want to look into the differences before I commit to that. I'm just writing this here so I don't forget...

test/edu-benefits/pages/contactInformation.unit.spec.jsx
cvalarida: I haven't investigated it thoroughly yet, but this is because it was being validated with previous information. I just fixed up a bug where the validation was run with the data before it was changed. That made this go back to 6.

I'm guessing the `data={{}}` sets the country to blank, overriding the default..? That'd cause there to be 6 errors instead of 5.

src/js/common/schemaform/helpers.js
```diff
@@ -141,7 +141,7 @@ export function formatReviewDate(dateString) {
   return undefined;
 }
 export function parseISODate(dateString) {
-  if (dateString) {
+  if (typeof dateString === 'string') {
```
jbalboni: What's the reason for this change (it's fine, just curious)?

jbalboni (approved):

src/js/common/schemaform/helpers.js
cvalarida: Oh, this was back when the DefinitionTester passed non-objects as data. An object was passed to this function which, of course, passed this statement, but failed when we tried to `dateString.split()`.

Shouldn't be a problem now, but it just stays there as a sort of safeguard.

PR #5381, opened 2017-04-20
Capitalize 1990N when displayed

cvalarida:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2321

image
(In other news, we should have a ticket to add that form title instead of `(Form Title to be Added Here)`)

image

Don't mind the blank spaces...
image


jbalboni:
@cvalarida The paper form title is "APPLICATION FOR VA EDUCATION BENEFITS UNDER THE
 NATIONAL CALL TO SERVICE (NCS) PROGRAM", so I think we can go with that (but not all caps).

jbalboni (approved):

J-Quag:
LGTM, thanks guys!

cvalarida:
Yeah, I just didn't know if we wanted that whole thing. I wanna say we shortened some of the others...but I'll put it in there and let content figure out what the final version should say.

PR #5383, opened 2017-04-21
Intro and chapter 4 changes on 1990n

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2315
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2318

Makes direct deposit fields required and updates the intro page.

screen shot 2017-04-20 at 10 29 40 pm
screen shot 2017-04-20 at 10 28 51 pm
screen shot 2017-04-20 at 10 28 56 pm


J-Quag:
Looks good to me

src/js/edu-benefits/1990n/components/IntroductionPage.jsx
```diff
@@ -34,7 +33,7 @@ class IntroductionPage extends React.Component {
                   <span>You won’t be able to save your work or come back to finish. So before you start, it’s a good idea to gather information about your education history and the school you want to attend.</span>
                 </div>
               </div>
-              <p>We strongly encourage you to work with a trained professional who can help you pick the right program. <a href="/disability-benefits/apply/help/index.html">Find an accredited representative.</a> Or call 888-442-4551 (888-GI-BILL-1) from 8:00 a.m. to 7:00 p.m. (ET) to talk with an education specialist.</p>
```
cvalarida: In the original issue, we had the two lines `Remove "Find a representative section"` and `Remove VSO sentence`. Does removing this line take care of both of those?

test/edu-benefits/pages/directDeposit.unit.spec.jsx
```diff
@@ -0,0 +1,63 @@
...
+
+      submitForm(form);
+
+      expect(onSubmit.called).to.be.true;
+    });
+  }
+};
+
+describe('Edu directDepositChangePage', () => {
+  describe('1990e', () => pageTests(formConfig1990e.chapters.personalInformation.pages.directDeposit));
```
cvalarida: Ah, I like that. 👍

test/edu-benefits/pages/directDeposit.unit.spec.jsx
```diff
@@ -0,0 +1,63 @@
...
+          data={{}}
+          uiSchema={uiSchema}/>
+    );
+
+    const formDOM = findDOMNode(form);
+
+    expect(formDOM.querySelectorAll('input').length).to.equal(4);
+  });
+
+  if (isRequired) {
```
cvalarida: I can see how this is cleaner, but it leaves room for programmer error. What if we make a change later to require these fields in another form but forget we have to require them in the test explicitly?

Can we instead check to see if `schema.properties.bankAccount.required` is not empty? Better yet, can we recursively check to see if there are any `required` arrays that aren't empty in the `schema`?

test/edu-benefits/pages/directDeposit.unit.spec.jsx
```diff
@@ -0,0 +1,63 @@
...
+        <DefinitionTester
+            schema={schema}
+            data={{}}
+            uiSchema={uiSchema}/>
+      );
+
+      const formDOM = findDOMNode(form);
+      submitForm(form);
+
+      expect(formDOM.querySelectorAll('.usa-input-error').length).to.equal(3);
```
cvalarida: This could then change to be however many required fields there are.

cvalarida (approved):
I'd like to not pass a parameter for `isRequired` to the `pageTest` function, but other than that, it looks golden.

test/edu-benefits/pages/directDeposit.unit.spec.jsx
jbalboni: Maybe it would make sense to make this the number of required fields, and default it to 0 (or whatever the default count is). And then instead of conditionally adding a test, just test that the right number of required fields are flagged. I think that would catch the case you're thinking about.

I don't think we should check the required property, because the goal of the test is to catch any inadvertent changes to the required fields in the form config. Unintentional changes might correctly set the required property and still be wrong.

src/js/edu-benefits/1990n/components/IntroductionPage.jsx
jbalboni: I believe so.

Issue #3542, opened 2016-10-26
ErrorableNumberInput removing decimal unexpectedly


jbalboni:
On the hours input on the education history page in education benefits:
1. Enter 70.00
2. Backspace to delete both 0s
3. Decimal place is removed and cursor is moved to the beginning of the input

This seems to happen because the DOM event we get in the React component is "70" instead of "70.", so we when we put it in the Redux store and re-render, the value in the input field changes.

This is a little odd and unexpected.

jcmeloni:
Confirmed still an issue.

jbalboni:
This is a React bug: https://github.com/facebook/react/pull/7359, which is fixed in 15.5.

PR #5384, opened 2017-04-21
Update to React 15.5

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2042
Connects to https://github.com/department-of-veterans-affairs/vets-website/issues/3542

This updates us to React 15.5. It also makes some minor updates to other packages, and major version updates to react-router and skin-deep (major in the semantic versioning sense, not the impact sense).

This change creates some warnings in the console and in the unit test output. Some of those will go away when we update our prop types. I'm going to leave that to a separate issue, because it touches every component.


test/common/components/SampleFeature.unit.spec.jsx
```diff
@@ -7,21 +7,17 @@ import SampleFeature from '../../../src/js/common/components/SampleFeature.jsx';
 describe('<SampleFeature>', () => {
   it('should be empty when not enabled', () => {
     const tree = SkinDeep.shallowRender(
-      <div>
-        <SampleFeature/>
-      </div>
+      <SampleFeature/>
```
jbalboni: skin-deep 1.0 seems more strict about shallow rendering, and wasn't rendering `<SampleFeature/>` as it was written before.

test/edu-benefits/1990/components/SubmitMessage.unit.spec.jsx
```diff
@@ -5,7 +5,7 @@ import { expect } from 'chai';
 import SubmitMessage from '../../../../src/js/edu-benefits/1990/components/SubmitMessage';
 import { createVeteran } from '../../../../src/js/edu-benefits/1990/utils/veteran';
 
-describe('<BenefitsWaiverFields>', () => {
+describe('<SubmitMessage>', () => {
```
cvalarida: This seems like an odd change for this PR. What prompted it? Something with skin-deep 1.0, or should this have been `<SubmitMessage>` already?

test/common/components/SampleFeature.unit.spec.jsx
cvalarida: This is perhaps a result of their introducing the [dive](https://www.npmjs.com/package/skin-deep#going-deeper) command?

package.json
```diff
@@ -134,20 +134,22 @@
     "polyfill-function-prototype-bind": "0.0.1",
     "quick_check": "^0.7.0",
     "raven-js": "^3.12.1",
-    "react": "^15.2.1",
+    "react": "^15.5.4",
     "react-addons-css-transition-group": "^15.3.2",
     "react-addons-perf": "^15.3.0",
     "react-addons-test-utils": "^15.3.0",
```
cvalarida: Looks like this is [being depreciated](https://facebook.github.io/react/blog/2017/04/07/react-v15.5.0.html#react-test-utils) in favor of `react-dom/test-utils`. We'll have to change...I think all our unit tests to import that instead.

cvalarida (approved):
Looks good. Just had a question (mostly for my curiosity). The depreciation of `react-addons-test-utils` isn't a deal-breaker, but I think we should migrate while we're already upgrading.

test/edu-benefits/1990/components/SubmitMessage.unit.spec.jsx
jbalboni: It should have been that already.

test/common/components/SampleFeature.unit.spec.jsx
jbalboni: That was there before, I think.

package.json
jbalboni: Yep, I meant to look for a codemod to update that, but forgot.

jbalboni: I think I'm going to leave this to another PR (along with the prop types change). It'll touch so many components and I wanted to keep this PR as simple as possible.

bshyong (approved):
LGTM let's get this one merged before I merge in the webpack 2 changes

PR #5385, opened 2017-04-21
Build updates for Webpack 2

bshyong:
https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2262

Note that two of our dependencies, namely `metalsmith-webpack` and `metalsmith-webpack-dev-server` have outdated webpack dependencies. I have a fork of those libraries that sets the correct dependencies; we can consider copying those to repos under the `department-of-veterans-affairs` org if desired.


bshyong:
`npm run watch` and `npm run build` work on development but there is a failure when building with the `production` buildtype

jbalboni:
@bshyong I see a warning when building, any idea what it is?

```
loaderUtils.parseQuery() received a non-string value which can be problematic, see https://github.com/webpack/loader-utils/issues/56
parseQuery() will be replaced with getOptions() in the next major version of loader-utils.
```

config/webpack.config.js
```diff
@@ -183,12 +232,13 @@ const configGenerator = (options) => {
       filename: 'chunk-manifest.json',
       manifestVariable: 'webpackManifest'
     }));
-    baseConfig.plugins.push(new webpack.optimize.DedupePlugin());
     baseConfig.plugins.push(new webpack.optimize.OccurrenceOrderPlugin(true));
```
jbalboni: I don't think we need OccurenceOrderPlugin anymore.

jbalboni:
@bshyong I think the error is because chunk-manifest-webpack-plugin needs to be updated. That fixes it for me.

jbalboni:
Also, I think we should put those two forked modules in the GH VA repo. It doesn't look like they're getting updated anytime soon, so this might be a long-term thing.

bshyong:
@jbalboni thanks for the suggestions! I will do another pass and get this PR cleaned up. 

I traced down the warning yesterday and believe it's due to Babel v6. If we upgrade to v7 it should go away but 7 is currently still in alpha/beta

bshyong:
@ayaleloehr or somebody else with org admin access: can you create two repos for our metalsmith dependencies?

`metalsmith-webpack` and `metalsmith-webpack-dev-server`

These are currently on my personal github account but it is better to clone them into the VA org as they are a dependency for vets-website

ayaleloehr:
@bshyong I created:

https://github.com/department-of-veterans-affairs/metalsmith-webpack
https://github.com/department-of-veterans-affairs/metalsmith-webpack-dev-server

Let me know if you need anything else or there are issues!

bshyong:
@jbalboni do you have any ideas on this last remaining error? looks like babel is no longer transpiling the `mocha-setup` file that we have which is in es6

I noticed that we are requiring the compilers in `mocha.opts`, which seems to be correct

```

/application/test/util/mocha-setup.js:1
(function (exports, require, module, __filename, __dirname) { import chai from 'chai';
                                                              ^^^^^^
SyntaxError: Unexpected reserved word
```

jbalboni:
@bshyong Ah, I didn't think about that. We probably need to tell Babel to transpile modules when we're running unit tests, since that's in node. There's a test environment in .babelrc, where you could change the `modules` option to true, but you'll have to add the `NODE_ENV=test` prefix to the other test commands in package.json (like it has for `test:coverage`).

bshyong:
^ fixed! Figured it out just now

bshyong:
@jbalboni Can I merge this PR?

jbalboni:
@bshyong Do we still need to go through and look for files with `import` and `module.exports`? I haven't done that.

bshyong:
@jbalboni I went through and fixed all the ones I saw. The e2e tests are passing, and I went through the various apps to do a sanity check but it would be great to get more eyes on this.

U-DON (approved):
LGTM. Seems to build and pass sanity checks.

package.json
```diff
@@ -168,7 +168,8 @@
     "tinyliquid": "^0.2.33",
     "url-loader": "^0.5.7",
     "uswds": "0.10.0",
-    "webpack": "^1.13.1",
+    "vets-json-schema": "https://github.com/department-of-veterans-affairs/vets-json-schema#3d7c8e8b47443dfd2b494a3a4151aa2afaa63118",
```
jbalboni: This seems out of date? The commit should be ee444b33dd221524c71b3b0201beda250a0ffd65, based on what I committed this morning.

jbalboni: Oh, I see, it was added to dependencies. You can remove this line.

PR #5386, opened 2017-04-21
Handle empty keyword searches in GIBCT properly.

U-DON:
Quick fix for empty keyword searches in GIBCT. It was getting set to `undefined` when the query didn't have a name, causing errors in string-related operations in `KeywordSearch` for empty searches.


bshyong (approved):
LGTM

PR #5411, opened 2017-04-24
Requiredloginview Unit Test

omgitsbillryan:
No coverage exists for [some hairy logic inside RequiredLoginView](https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/js/common/components/RequiredLoginView.jsx#L45).  This PR intends to fix that.

This is my first FE unit test, so please use extra scrutiny.

**Relevant Issue**
https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2314

### DIFF
--- a/test/common/components/RequiredLoginView.unit.spec.jsx
+++ b/test/common/components/RequiredLoginView.unit.spec.jsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import SkinDeep from 'skin-deep';
 import { expect } from 'chai';
+import _ from 'lodash';
 
 import RequiredLoginView from '../../../src/js/common/components/RequiredLoginView.jsx';
 
@@ -54,7 +55,7 @@ describe('<RequiredLoginView>', () => {
   };
 
   function setup(props = {}) {
-    const mergedProps = Object.assign({}, defaultProps, props);
+    const mergedProps = _.merge({}, defaultProps, props);
     const tree = SkinDeep.shallowRender(
       <RequiredLoginView {...mergedProps}>
         <div>Test Child</div>
@@ -71,6 +72,13 @@ describe('<RequiredLoginView>', () => {
   it('should render', () => {
     const { vdom } = setup();
     expect(vdom).to.not.be.undefined;
+  });
+
+  it('should render a loading graphic while loading', () => {
+    const { tree, instance } = setup();
+    instance.setState({ loading: true });
+    const loadingIndicatorElement = tree.dive(['LoadingIndicator']);
+    expect(loadingIndicatorElement.text()).to.contain('Loading your information');
   });
 
   describe('LOA.current=1', () => {
@@ -94,14 +102,14 @@ describe('<RequiredLoginView>', () => {
     });
     describe('userProfile.status=SERVER_ERROR', () => {
       it('should display server error message', () => {
-        const serverErrorProfile = Object.assign({}, loa3User, { status: 'SERVER_ERROR' });
+        const serverErrorProfile = _.merge({}, loa3User, { status: 'SERVER_ERROR' });
         const { tree } = setup({ userProfile: serverErrorProfile });
         expect(tree.toString()).to.contain('Sorry, our system is temporarily down while we fix a few things');
       });
     });
     describe('userProfile.status=NOT_FOUND', () => {
       it('should display not found message', () => {
-        const notFoundProfile = Object.assign({}, loa3User, { status: 'NOT_FOUND' });
+        const notFoundProfile = _.merge({}, loa3User, { status: 'NOT_FOUND' });
         const { tree } = setup({ userProfile: notFoundProfile });
         expect(tree.toString()).to.contain('We couldn&#x27;t find your records with that information.');
       });
--- a/test/common/components/RequiredLoginView.unit.spec.jsx
+++ b/test/common/components/RequiredLoginView.unit.spec.jsx
@@ -54,6 +54,12 @@ describe('<RequiredLoginView>', () => {
     loginUrl: 'http://fake-login-url'
   };
 
+  class TestChildComponent extends React.Component {
+    render() {
+      return (<div>Child Component {this.props.name}</div>);
+    }
+  }
+
   function setup(props = {}) {
     const mergedProps = _.merge({}, defaultProps, props);
     const tree = SkinDeep.shallowRender(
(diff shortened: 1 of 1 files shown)


phiden:
I have no experience assessing unit tests -- I'll leave this one to @akainic if that's ok!

omgitsbillryan:
@akainic or @annekainicUSDS ?

test/common/components/RequiredLoginView.unit.spec.jsx
```diff
@@ -0,0 +1,116 @@
...
+
+    return { tree, instance, mergedProps, vdom };
+  }
+
+  it('should render', () => {
+    const { vdom } = setup();
+    expect(vdom).to.not.be.undefined;
+  });
+
+  describe('LOA.current=1', () => {
```
raquelromano-gov: This isn't super important, but these could be a little more descriptive, e.g. 'user is LOA1' and 'app requires LOA3'

test/common/components/RequiredLoginView.unit.spec.jsx
```diff
@@ -0,0 +1,116 @@
...
+        expect(tree.toString()).to.equal('<div><div>Test Child</div></div>');
+      });
+    });
+  });
+  describe('LOA.current=3', () => {
+    it('should display children elements', () => {
+      const { tree } = setup({ userProfile: loa3User });
+      expect(tree.toString()).to.equal('<div><div>Test Child</div></div>');
+    });
+    describe('userProfile.status=SERVER_ERROR', () => {
```
raquelromano-gov: This is more of a question about `RequiredLoginView`: why do we not have these same status checks (`SERVER_ERROR` and `NOT_FOUND`) for LOA1 users?

test/common/components/RequiredLoginView.unit.spec.jsx
```diff
@@ -0,0 +1,116 @@
...
+    describe('userProfile.status=NOT_FOUND', () => {
+      it('should display not found message', () => {
+        const notFoundProfile = Object.assign({}, loa3User, { status: 'NOT_FOUND' });
+        const { tree } = setup({ userProfile: notFoundProfile });
+        expect(tree.toString()).to.contain('We couldn&#x27;t find your records with that information.');
+      });
+    });
+  });
+  describe('when not logged in', () => {
+    it('should prompt for login', () => {
```
raquelromano-gov: How about testing that a not logged-in user can access a service with no minimum authorization level?

raquelromano-gov (commented):
Yikes, glad you are adding this! The logic in `RequiredLoginView` kind of scares me (nested ifs and sometimes checking account types and authorization levels with `>=` and other times with `==` but I think this test should be a good way to make sure that we don't break anything if we clean it up a bit.

test/common/components/RequiredLoginView.unit.spec.jsx
jcmeloni-usds: If a user is only LOA1, we do not check MVI for correlation IDs because we don't have authenticated info (fn, ln, dob, ssn, gender) to do so.  We only check MVI if a user is LOA3, so we would only get `SERVER_ERROR` and `NOT_FOUND` responses from MVI if a user is LOA3.

test/common/components/RequiredLoginView.unit.spec.jsx
```diff
@@ -0,0 +1,116 @@
...
+  function setup(props = {}) {
+    const mergedProps = Object.assign({}, defaultProps, props);
+    const tree = SkinDeep.shallowRender(
+      <RequiredLoginView {...mergedProps}>
+        <div>Test Child</div>
+      </RequiredLoginView>
+    );
+
+    const instance = tree.getMountedInstance();
+    instance.setState({ loading: false });
```
omgitsbillryan: Donezo. 👍

test/common/components/RequiredLoginView.unit.spec.jsx
```diff
@@ -0,0 +1,116 @@
...
+    describe('authRequired=3', () => {
+      it('should prompt for verification', () => {
+        const { tree } = setup({ userProfile: loa1User });
+        expect(tree.toString()).to.contain('Verify your Identity with ID.me');
+      });
+    });
+    describe('authRequired=1', () => {
+      it('should display children elements', () => {
+        const { tree } = setup({ authRequired: 1, serviceRequired: 'user-profile' });
+        expect(tree.toString()).to.equal('<div><div>Test Child</div></div>');
```
omgitsbillryan: I do agree that is a bit verbose and doesn't read well.  However, testing the entire html has the added bonus of testing that children components are nested inside a plain root `div`, whereas 

```jsx
expect(tree.toString()).to.contain('Test Child');
```
doesnt test the wrapping `<div>` and the fact that it dosnt have any attributes on it.  Is there another way to test?

raquelromano-gov: 👍  didn't realize you were intentionally testing the div attributes

test/common/components/RequiredLoginView.unit.spec.jsx
raquelromano-gov: Thanks, makes sense

test/common/components/RequiredLoginView.unit.spec.jsx
raquelromano-gov: Another thing you might test: add `instance.setState({ isServiceAvailableForUse: true });` to `setup()` and add a test to verify that children render if it's false

raquelromano-gov (approved):
Suggested one more path that you might want to test, but otherwise LGTM

test/common/components/RequiredLoginView.unit.spec.jsx
```diff
@@ -0,0 +1,154 @@
...
+
+  function setup(props = {}) {
+    const mergedProps = _.merge({}, defaultProps, props);
+    const tree = SkinDeep.shallowRender(
+      <RequiredLoginView {...mergedProps}>
+        <div>Test Child</div>
+      </RequiredLoginView>
+    );
+
+    const instance = tree.getMountedInstance();
```
raquelromano-gov: If `isServiceAvailableForUse` isn't true by default, I'd set it here.

raquelromano-gov (approved):
LGTM

Issue #5403, opened 2017-04-24
GI Bill CT: Calculator Results for OJT missing "/month"


saintsoup52:
The calculator results for OJT programs is missing key descriptive information.

Unlike how school results are displayed the calculated results for OJT/Apprenticeship programs are not totaled per term but rather they keep displaying a per month benefit. We need to clarify that these are monthly benefits not total benefits.

Additionally the OJT programs don't have "Terms" like school programs so we should probably modify that language as well.

**Old Version (Production)**
![image](https://cloud.githubusercontent.com/assets/4397508/25330462/bf2766b2-28ad-11e7-8f08-72e331cf47a6.png)

**Staging Version**
![image](https://cloud.githubusercontent.com/assets/4397508/25330506/dea1fd86-28ad-11e7-8efc-df067bfdf80d.png)

@leannammiller @saneshark @mphprogrammer

U-DON:
@melwoodard Any suggestions for the 'per term' language?

melwoodard:
@U-DON @bshyong let's change that heading to "Estimated benefits per month" since it's outlining monthly amounts

PR #5413, opened 2017-04-24
Copy change on calculator results for OJT

bshyong:
Resolves https://github.com/department-of-veterans-affairs/vets-website/issues/5403

### DIFF
--- a/src/js/gi/components/profile/Calculator.jsx
+++ b/src/js/gi/components/profile/Calculator.jsx
@@ -100,7 +100,7 @@ export class Calculator extends React.Component {
 
     return (
       <div>
-        <h3>Estimated benefits per month</h3>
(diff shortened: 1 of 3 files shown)


U-DON (commented):
I don't think we want to show the '/mo' on every `CalculatorResultRow`. And we should only show it if the institution is OJT.

I would make the change in selectors/calculator.js to add the '/mo' if it's OJT, similar to how it's being handled for `bookStipend`.

bshyong:
looks like conditionally adding `/mo` will cause the total calculation to fail somewhere

screen shot 2017-04-24 at 4 25 07 pm

U-DON:
In that screenshot, the terms are fall and spring, which means this is appearing in a non-OJT institution. It shouldn't be a problem if you also check for the institution type in the condition.

U-DON (commented):
After some debugging, the errors have been resolved. LGTM.

PR #5416, opened 2017-04-25
Corrected GIBCT housing allowance calculation for MGIB.

U-DON:
Resolves #5388.

Compared with the existing calculator's code to find that the constants being used were different depending on the length of enlistment service, so that's the fix.

PR #5418, opened 2017-04-25
Corrected GIBCT housing allowance calculation for MGIB.

### LINKED ISSUE #5388: GI Bill CT: Monthly living allowance rate for MGIB incorrect
The benefits estimator is not correctly calculating the MGIB (CH 30) benefits when the user served less than 3 years. The correct rate for a 2 year enlistment under CH 30 is $1,509/month.

http://www.benefits.va.gov/GIBILL/resources/benefits_resources/rates/ch30/ch30rates100116.asp

**3 years**
image

**2 years**
image

U-DON:
Resolves #5388.

Duplicate of #5416 using a branch name that will allow Jenkins to build it.


bshyong (approved):
LGTM!

PR #5419, opened 2017-04-25
Hca military service

raquelromano-gov:
Closes https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2232

Service information:
image

Additional information:
image


src/js/hca-rjsf/config/form.js
```diff
@@ -24,6 +46,91 @@ const formConfig = {
...
+          uiSchema: {
+            lastServiceBranch: {
+              'ui:title': 'Last branch of service',
+              'ui:options': {
+                labels: lastServiceBranchLabels
+              }
+            },
+            // TODO: this should really be a dateRange, but that requires a backend schema change. For now
+            // leaving them as dates, but should change these to get the proper dateRange validation
+            lastEntryDate: currentOrPastDate.uiSchema('Start of service period'),
```
raquelromano-gov: I think this should be changed to a `dateRange` in the backend, but not sure how to ensure nothing breaks with the live application. I neglected to set up time to discuss that, but will bring it up on #rainbows-team to see who wants to join a quick chat on how to orchestrate any backend changes.

src/js/hca-rjsf/config/form.js
```diff
@@ -24,6 +46,91 @@ const formConfig = {
...
+              'ui:title': 'Character of discharge',
+              'ui:options': {
+                labels: dischargeTypeLabels
+              }
+            }
+          },
+          schema: {
+            type: 'object',
+            properties: {
+              lastServiceBranch: _.assign(lastServiceBranch, { type: 'string' }),
```
raquelromano-gov: Maybe these types should be defined in the backend schema as well

jbalboni: They should, we can update that.

src/js/hca-rjsf/config/form.js
jbalboni: Yeah. We have to be careful with date on hca. On edu, we allow partial dates, but on hca they do not.

raquelromano-gov: Does "they do not" mean they *should* not, or they just didn't previously, but it would be fine if they did? @annekainicUSDS maybe you have more bg information on hca requirements

cvalarida:
I noticed that too, either here or somewhere else...I don't remember...I'm guessing it does that focus because of accessibility reasons (focus on the bit the screenreader should start reading or something).

cvalarida (approved):
Looks good to me. Once we get the schema changes, we can come back and fix up the date range and last service branch type.

jbalboni:
Yeah, there are some css changes that are in edu-benefits and not in the hca file. They should get migrated to the schemaform file.

jbalboni (approved):

PR #5421, opened 2017-04-25
Update GA "Learn More" tracking

robertfairhead:
Only fire the event for showing a modal (a "learn more" event) when a modal is being shown. Ignore the null case when a modal is being closed.


U-DON (approved):
LGTM.

PR #5423, opened 2017-04-25
Merge SM 508 and E2E.

U-DON:
Related to messaging task of department-of-veterans-affairs/vets.gov-team#2203.


ayaleloehr:
Tagging @raquelromano-gov as well since she started the other 508 tests, so just for awareness!

bshyong (approved):
LGTM Looks like the latest change should pass tests, if the HCA tests pass too. HCA failure may just be a flakey test since this PR doesn't touch the HCA code

src/js/common/components/form-elements/ErrorableTextarea.jsx
```diff
@@ -100,6 +101,7 @@ class ErrorableTextarea extends React.Component {
             disabled={this.props.disabled}
             className={classes}
             aria-describedby={errorSpanId}
+            aria-labelledby={`${this.inputId}-label`}
```
U-DON: Originally added `aria-label`, but it seems redundant and wrong to have it if the preceding `label` is visible. Instead of setting `#msg-reply .msg-write label { display: none; }`, I set it to `left: -9999em` so that it can still be read by screen readers, and appropriately set `aria-labelledby` on the textarea.

raquelromano-gov:
Thanks for doing this! LGTM

src/js/common/components/form-elements/ErrorableTextarea.jsx
bshyong: Approach looks good to me

PR #5424, opened 2017-04-25
Allow conditional fields in arrays on schemaform

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2077

This updates schemaform to allow conditional fields in arrays. The issue before was that we modify the schema for each page when there is a conditional field change, but there was only one schema for all items in an array, so different items couldn't have different conditional fields.

Now, we've switched to converting the `items` property to an array in the schema and putting the original schema in the `additionalItems` property. This is current JSON Schema functionality, where items in an array will use the corresponding schema in the `items` array, and if they're out of sync, it will use the schema from `additionalItems`.

There are very few developer facing changes. The main one is that some of the functions in the uiSchema like `ui:required`, `updateSchema`, and `hideIf` are now sent an index, so you can access the right item if you need to use form data.

I still need to update documentation.


src/js/edu-benefits/pages/applicantService.js
```diff
@@ -26,7 +26,18 @@ export default function applicantServicePage() {
         },
         'ui:required': form => _.get('view:applicantServed', form),
         items: {
-          serviceStatus: { 'ui:title': 'Type of separation or discharge' }
+          serviceStatus: {
+            'ui:title': 'Type of separation or discharge',
+            'ui:options': {
+              updateSchema: (pageData, formData, currentSchema, index) => {
+                return {
+                  minLength: _.get(`applicantService.data.toursOfDuty[${index}].serviceBranch`, formData) === 'Army'
```
cvalarida: Isn't `serviceBranch` a text field? We should probably `.toLowerCase()` it and compare it to `'army'`, aye?

test/util/schemaform-utils.jsx
```diff
@@ -39,8 +43,11 @@ export class DefinitionTester extends React.Component {
   handleChange = (data) => {
     const state = this.props.state;
     const uiSchema = this.state.uiSchema;
-    const formData = _.merge(this.props.formData, data);
-    let schema = updateRequiredFields(this.state.schema, uiSchema, formData);
+    const formData = this.state.schema.type === 'array'
+      ? data
+      : _.merge(this.props.formData, data);
+    let schema = updateItemsSchema(this.state.schema, formData);
+    schema = updateRequiredFields(schema, uiSchema, formData);
```
cvalarida: Is there an argument against making a little wrapper function for calling `updateItemsSchema() and `updateRequiredFields`? The latter seems to always be preceded by the former, so I'd hate for some modification down the road to only call one or the other, y'know? Debugging that would be such a pain!

Heh, on the other hand, we seem to be running out of names for these functions and still have them make sense!

cvalarida (approved):
Looks good, all the tests check out. 👍 from me!

I did have a couple questions / suggestions, though.

src/js/edu-benefits/pages/applicantService.js
jbalboni: This is going to go away completely, it's made up test code.

test/util/schemaform-utils.jsx
jbalboni: It might be good to wrap this whole sequence of function calls in a function that we could reuse more easily.

jbalboni:
@cvalarida I ended up splitting out the code that updates the schema each time data changes into its own file and creating a wrapper function that's used by the reducer and the schema form utils. I think that'll be a little cleaner.

src/js/common/schemaform/formState.js
```diff
@@ -0,0 +1,378 @@
...
+ * is updated to remove newly hidden data
+ *
+ * @param {Object} schema The current JSON Schema
+ * @param {Object} uiSchema The current UI Schema (does not change)
+ * @param {Object} formData Flattened data for the entire form
+ * @param {Object} pageData The data for the current page
+ * @param {Object} state The Redux state object, which contains schema and data for
+ *   all pages
+ */
+export function updateSchemaData(schema, uiSchema, formData, pageData, state) {
```
cvalarida: I don't have a better name, but this one is kind of misleading, isn't it? Mostly it updates the schema...but also updates the data as well.

`updateSchemaAndData` seems a bit much, but it's accurate enough...

cvalarida (approved):
The refactor looks good!

src/js/common/schemaform/formState.js
jbalboni: Yeah, I went back and forth. updateSchemaAndData seems more accurate.

PR #5427, opened 2017-04-25
Default GIBCT profile accordion items to open.

### LINKED ISSUE #5402: GI Bill CT: Default the accordions to open
The accordions/drawers to the GI Bill Comparison tool should be defaulted to open. 

image

The information provided in those drawers is critical for users making tough choices about where they should go to school. However many users will not know they should open these drawers and therefore miss out on that valuable information. Think of the benefit rates as desert and the key indicators of value as the veggies. Who goes looking for broccoli after you had an ice cream sandwich?

When we launched the original Paying for College tool one of the biggest lessons we learned was that many users will not open a drawer or an accordion unless they have to. For example... only 37% of CFPB's users opened our accordions/drawers even though that was an essential step for users using our tool.

image

@leannammiller @saneshark @mphprogrammer

U-DON:
Resolves #5402.

### DIFF
--- a/src/js/gi/components/AccordionItem.jsx
+++ b/src/js/gi/components/AccordionItem.jsx
@@ -37,7 +37,7 @@ AccordionItem.propTypes = {
 };
 
 AccordionItem.defaultProps = {
-  expanded: false
+  expanded: true
 };
 
 export default AccordionItem;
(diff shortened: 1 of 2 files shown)


bshyong (approved):
LGTM

PR #5428, opened 2017-04-25
Ensure that site navigation is always on top.

ghost:
Related to department-of-veterans-affairs/vets.gov-team#2273

Also adjusted scroll bar appearance for menu.


ayaleloehr (approved):
I added a comment in https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2273 about a different way to resolve this, so I will leave that open for now, but this LGTM and does visually fix the issue, which is great!

PR #5429, opened 2017-04-26
Remove conditional field for military history from applicant service page on 1990n

jbalboni:
https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2316

Removes the conditional field and adds service status.

screen shot 2017-04-25 at 8 15 07 pm
screen shot 2017-04-25 at 8 15 13 pm

### DIFF
--- a/test/e2e/edu-1990n-helpers.js
+++ b/test/e2e/edu-1990n-helpers.js
@@ -1,0 +1,14 @@
+function completeServicePeriods(client, data) {
+  client
+    .fill('input[name="root_toursOfDuty_0_serviceBranch"]', data.toursOfDuty[0].serviceBranch)
+    .fillDate('root_toursOfDuty_0_dateRange_from', data.toursOfDuty[0].dateRange.from)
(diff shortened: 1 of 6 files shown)


test/e2e/edu-1990n-helpers.js
```diff
@@ -0,0 +1,14 @@
+function completeServicePeriods(client, data) {
```
cvalarida: Could we put this in EduHelpers and make it a generic `completeToursOfDuty()` which the `completeServicePeriods()` in EduHelpers could call? Could loop over `data.toursOfDuty` and fill in all the data it has (so if we don't have `serviceStatus`, we pass it by).

cvalarida (approved):
Looks good. Just one suggestion about the versatility of the test.

test/e2e/edu-1990n-helpers.js
jbalboni: Yeah, though there's some variation in the tours of duty fields now that might be hard to accomodate. I think I'll make a separate item for it.

PR #5432, opened 2017-04-26
Fix HCA maritalStatus

cvalarida:
Closes https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2456

Just had to validate that field as well when Continue was hit.


jbalboni (approved):
LGTM

PR #5434, opened 2017-04-26
Add HCA Chapter 3

cvalarida:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2234

The new is on the left, the old is on the right.
image

Note: The page title doesn't show up...not sure why, but this is a ubiquitous issue, so I'm not worrying about it in this PR.

### DIFF
--- a/src/js/hca-rjsf/config/form.js
+++ b/src/js/hca-rjsf/config/form.js
@@ -39,7 +39,10 @@ const {
   isCoveredByHealthInsurance,
   vaMedicalFacility,
   isEssentialAcaCoverage,
-  wantsInitialVaContact
+  wantsInitialVaContact,
+  isVaServiceConnected,
+  compensableVaServiceConnected,
+  receivesVaPension
 } = fullSchemaHca.properties;
 
 const {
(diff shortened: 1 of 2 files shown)


jbalboni:
@cvalarida If you want the page title to show up there, it should go in `ui:title` in the root of `uiSchema`. The other title is used on the review page, but doesn't automatically show up on the regular pages because we don't always want a title to show up.

src/js/hca-rjsf/config/form.js
```diff
@@ -233,8 +236,40 @@ const formConfig = {
         }
       }
     },
+    vaBenefits: {
+      title: 'VA Benefits',
+      pages: {
+        vaBenefits: {
+          path: 'va-benefits/basic-information',
+          title: 'Current Compensation',
```
jbalboni: Page titles should be sentence case. Chapter titles are title case, so thank you for updating those.

jbalboni (approved):
LGTM, but needs some tests. Also, should be able to add the page title based on my comment.

src/js/hca-rjsf/config/form.js
cvalarida: Oh blast. You're right. I just copied and pasted this from the old one without thinking. Good catch!

PR #5437, opened 2017-04-26
Breadcrumbs consistency

U-DON:
Resolves department-of-veterans-affairs/vets.gov-team#2463.

Made styling of breadcrumbs consistent with pages that automatically include it through content with `gatePage: true`. Also went through Rx after this refactor to ensure the margins and padding of the content match with the breadcrumbs and are consistent with the other apps (at least GIBCT).

> screen shot 2017-04-26 at 5 54 40 pm


src/js/rx/components/ErrorView.jsx
```diff
@@ -58,9 +58,10 @@ class ErrorView extends React.Component {
             status="warning"/>
         );
     }
+
     return (
       <div className="call-out rx-app-error rx-tab-explainer">
-        // needed due to call-out class css rule targeitng
+        // needed due to call-out class css rule targeting
```
U-DON: I'm not sure what this is about. As far as I can tell, `call-out` is a class only used in Facility Locator. I can remove that from the class name and this hack (if it is one) if you don't think it's needed.

src/js/common/components/AlertBox.jsx
```diff
@@ -33,7 +33,7 @@ class AlertBox extends React.Component {
 
   render() {
     if (!this.props.isVisible) {
-      return <div aria-live="assertive"/>;
+      return null;
```
U-DON: Hopefully doesn't affect other apps. Just seems more proper that it should not display something that shouldn't be displayed rather than insert an empty `div`. Even with this, I think if an alert appears, it should still alert the screen reader, right?

src/sass/rx/partials/_rx-detail.scss
```diff
@@ -1,40 +1,45 @@
...
+    td:last-child {
+      width: 99%;
+    }
   }
 
-  td:last-child {
-    width: 99%;
+  .rx-contact-card {
+    margin-bottom: 3.5rem;
+    width: 100%;
```
U-DON: This is just to nest rules under one `#rx-detail` instead of prepending it individually to all of them.

src/sass/rx/rx.scss
```diff
@@ -33,10 +33,12 @@
...
 .rx-app {
-  background: $color-white;
-  padding: 3rem 1.5rem;
-  margin: auto;
-  max-width: $site-max-width;
+  margin-bottom: 3rem;
+
+  // TODO: Move to core?
+  .va-alert:first-child {
+    margin-top: 0;
```
U-DON: I see that there is a similar rule for `h1` in `_va.scss`, but not sure if we want to add it directly just yet.

src/js/common/components/AlertBox.jsx
bshyong: So the reasoning behind this is that there will always be a DOM element rendering, and if the AlertBox has content, it will re-render in the same place but with children. 

Because screenreaders are inconsistent in the way they handle content randomly being inserted into the DOM, it is best to always have a wrapper div with `aria-live` so that the reader knows that the content inside may change (and will alert the user accordingly).

src/js/rx/components/ErrorView.jsx
bshyong: I think there was some selector that caused the first `p` element to have special styling, hence that class and the extra `<p></p>` at the top. If that class is only in FL now, we can probably remove it. Maybe it will no longer cause a visual issue but I haven't checked

bshyong (approved):
LGTM; some minor comments

src/js/common/components/AlertBox.jsx
U-DON: Since this isn't really in the scope of these set of changes, I'll just revert it for now. I think having a wrapper `div` with `aria-live` specifically for an alerts section in the header of an app would be a good solution, but that should be a separate PR.

U-DON: Meanwhile, there will be a bit of extra top margin in Rx, but it's not too noticeable.

PR #5335, opened 2017-04-14
Adding USWDS replacement class names as needed.

ghost:
Closes department-of-veterans-affairs/vets.gov-team#830

Kinda sorta un-blocks department-of-veterans-affairs/vets.gov-team#549

PR #5438, opened 2017-04-27
Fixed column widths of header summary in GIBCT profile pages.

U-DON:
Resolves department-of-veterans-affairs/vets.gov-team#2443.

Foundation grid class names were removed accidentally during the intermediate set of changes to transition to USWDS grid classes (#5335). This should bring the icons with info section back to two columns.


bshyong (approved):
LGTM

PR #5440, opened 2017-04-27
Prevent 'Null' from being displayed as the users first name

omgitsbillryan:
**GH Issue**
https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2414

### DIFF
--- a/test/login/components/SearchHelpSignIn.unit.spec.jsx
+++ b/test/login/components/SearchHelpSignIn.unit.spec.jsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import SkinDeep from 'skin-deep';
 import { expect } from 'chai';
+import _ from 'lodash';
 
 import { SearchHelpSignIn } from '../../../src/js/login/components/SearchHelpSignIn.jsx';
 
@@ -22,7 +23,7 @@ const defaultProfileProps = {
 };
 
 function buildProps(defaultProps, props = {}) {
-  return Object.assign({}, defaultProps, props);
+  return _.merge({}, defaultProps, props);
 }
 
 describe('<SearchHelpSignIn>', () => {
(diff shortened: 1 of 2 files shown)


test/login/components/SearchHelpSignIn.unit.spec.jsx
```diff
@@ -4,18 +4,29 @@ import { expect } from 'chai';
 
 import { SearchHelpSignIn } from '../../../src/js/login/components/SearchHelpSignIn.jsx';
 
+const defaultLoginProps = {
```
omgitsbillryan: I had to refactor this unit test in order to keep it DRY while adding the one test to cover the `null` greeting fix.

raquelromano-gov (approved):

test/login/components/SearchHelpSignIn.unit.spec.jsx
```diff
@@ -1,21 +1,33 @@
...
+
+const defaultProfileProps = {
+  email: 'fake@aol.com',
+  userFullName: {
+    first: 'Sharon'
+  }
+};
+
+function buildProps(defaultProps, props = {}) {
+  return _.merge({}, defaultProps, props);
```
U-DON: I would suggest using `merge` from `lodash/fp` if you don't want the extra arg to ensure both objects get merged into a new object. The FP version will just create a new object by default.

omgitsbillryan: Is there doc that outlines this fact somewhere?  I'm not familiar with the `fp` flavor of lodash.

U-DON: It's not a very good doc, but here's the [official FP guide](https://github.com/lodash/lodash/wiki/FP-Guide).

raquelromano-gov: @omgitsbillryan @U-DON I believe this doc is auto-generated from the lodash API docs and might be a bit more comprehensive: https://gist.github.com/jfmengels/6b973b69c491375117dc

PR #5443, opened 2017-04-27
Add `depends` usage to cookbook

cvalarida:
Looking at `src/js/common/utils/helpers.js` -> `isActivePage()`, I think I got those options for `depends` right, but correct me if I'm wrong.

Also, should we think about passing in the flattened data so we don't have to access `data.pageName.data.fieldName`?


jbalboni (approved):
LGTM. Once my flatten PR is merged, we won't need the page name and `data` in the depends objects.

cvalarida:
@jbalboni Awesome. Sounds like a plan! Ping me when that's done and I'll fix it up here too.

cvalarida:
Does that look right, @jbalboni?

jbalboni:
@cvalarida Looks good to me!

PR #5445, opened 2017-04-27
Fix RequiredLoginView tests

jbalboni:
Fixes the broken login tests that are preventing us from releasing to production.


raquelromano-gov (approved):
Any idea why this passed on jenkins and was allowed to be merged? I have not looked at the failure logs

PR #5449, opened 2017-04-27
Add unit tests for GIBCT containers

bshyong:
resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2466


test/gibct/containers/LandingPage.unit.spec.js
```diff
@@ -0,0 +1,30 @@
+import { createStore } from 'redux';
+import { expect } from 'chai';
+import React from 'react';
+import sinon from 'sinon';
+import SkinDeep from 'skin-deep';
+
+import { LandingPage } from '../../../src/js/gi/containers/LandingPage';
+import reducer from '../../../src/js/gi/reducers';
+
+const store = createStore(reducer);
```
U-DON: In the test for `GiBillApp`, you set `defaultProps` in one line instead referring to it later as `store.getState()`. Probably should be consistent here (I prefer the `defaultProps`).

test/gibct/containers/Modals.unit.spec.js
```diff
@@ -0,0 +1,17 @@
...
+import SkinDeep from 'skin-deep';
+
+import Modals from '../../../src/js/gi/containers/Modals';
+import reducer from '../../../src/js/gi/reducers';
+
+const store = createStore(reducer);
+
+describe('<Modals>', () => {
+  it('should render', () => {
+    const tree = SkinDeep.shallowRender(<Modals store={store}/>);
```
U-DON: Why doesn't this require `getState()` like the others?

U-DON (approved):
LGTM. Couple of minor comments.

test/gibct/containers/LandingPage.unit.spec.js
bshyong: Will do

test/gibct/containers/Modals.unit.spec.js
bshyong: This container is only bringing in Modal components which are already covered by common component tests. So here we are testing the Modals connected component directly (passing in store) and doing a simple render check.

In the other container tests, I am testing the undecorated component, so we need to manually pass in the props from the reducers. That said, I can refactor this test to follow the same pattern as the others for consistency.

U-DON: Oh, didn't notice till now that this is the decorated component. Sure, would like to see it tested as a regular component like the others.

PR #5451, opened 2017-04-28
Add code climate config to reduce duplication issues

jbalboni:
We get a lot of codeclimate duplication errors, this updates the paths and increases the thresholds for those issues.

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,6 +1,10 @@
 engines:
   duplication:
-      enabled: false
+    enabled: true
+    config:
+      languages:
+        javascript:
+          mass_threshold: 100
 ratings:
   paths:
   - src/js/**
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -3,6 +3,9 @@ engines:
       enabled: false
 ratings:
   paths:
-  - src/**
+  - src/js/**
 exclude_paths:
 - test/**/*
+- config/**/*
+- vendor/**/*
+- script/**/*
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -8,8 +8,3 @@ engines:
 ratings:
   paths:
   - src/js/**
-exclude_paths:
-- test/**/*
-- config/**/*
-- vendor/**/*
-- script/**/*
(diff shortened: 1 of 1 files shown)


jbalboni:
FYI @raquelromano-gov

.codeclimate.yml
```diff
@@ -0,0 +1,15 @@
+engines:
+  duplication:
+    enabled: true
+    config:
+      languages:
+        javascript:
+          mass_threshold: 100
```
raquelromano-gov: I'm wondering how mass 100 compares to the default we have now. It's not clear to me what value we want, esp. given how mass is defined, but if I'm guessing this will dial down the sensitivity a bit?

From https://docs.codeclimate.com/v1.0/docs/duplication#section-understand-the-engine: "Mass" refers to the size of the duplicated code. Specifically, mass is determined by the size of a code block's s-expression, after it has been parsed into a node of an Abstract Syntax Tree (AST).

jbalboni: Yeah, I was trying to see what effect it would have. The default is 40. Higher produces fewer warnings.

That definition is pretty hard for me to intuitively apply to code, so I'm not sure how to set that value other than trial and error.

.codeclimate.yml
```diff
@@ -0,0 +1,15 @@
...
+ratings:
+  paths:
+  - src/js/**
+exclude_paths:
```
raquelromano-gov: Curious how `paths` above and `exclude_paths` interact.  If we only specify `src/js/**`, why do we also need to exclude these paths (and why don't we need to exclude other paths in `src`, like `src/test-support` that have javascript files?

Are these only directories that might contain javascript files?

raquelromano-gov (approved):

.codeclimate.yml
jbalboni: I think the exclude paths remove files from any of the `path` options used (there's only one in here, but there can be more depending on the engines you use). Might not be useful in our case, actually.

jbalboni: @raquelromano-gov It's doing something. I removed those excluded paths and CC updated to say there where 156 open issues (instead of none).

raquelromano-gov: Ha, OK let's keep them then.!

PR #5452, opened 2017-04-28
Refactor verification (leveling up from LOA1 to LOA3)

raquelromano-gov:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2413
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2474

---

What started as small fix to avoid a popup from being blocked ended up being a refactor of how a logged-in LOA1 user gets verified and bumped up to LOA3. 

**Original problem:** the user action to verify identity was triggering both the URL fetch from ID.me and the opening of the ID.me window in which the verification happens. Because the URL fetch is asynchronous, the window open was in a callback, which most browsers will block since it's not a direct result of a user action. This has been the source of many login-related help center calls. Here are screenshots of the 2 places this happens for a signed-in LOA1 user:

1) 'GetStarted' button on www.vets.gov/healthcare/messaging, www.vets.gov/healthcare/prescriptions, www.vets.gov/track-claims/your-claims (and healthcare/health-records, though that's not live); corresponds to `src/js/common/components/authentication/VerifyPrompt.jsx`

image

2) 'verify your account' link toward the bottom of www.vets.gov/profile; corresponds to `src/js/user-profile/components/AuthApplicationSection.jsx`

image

**Fix:** this fix treats leveling up (verifying) the same way logging works, i.e., fetches the  verification URL when the main login component loads and stores it in props passed down to child compoents so that it's available if/when a user needs to level up. The click handlers for the above button/link can immediately open the ID.me popup. This will also resolve [numerous JS errors](http://sentry.vetsgov-internal/vets-gov/website-production/?query=is%3Aunresolved+login) reported by Sentry because the previous handler was a static helper function that for some reason stored the `fetch` promise in `this.serverRequest_` (possibly remnants of a former refactor).

- Pros:
  - login and verification now behave the same
  - click handler can just pop up the ID.me page and not do an asynchronous fetch
- Cons:
  - 2 URL fetches when loading vets.gov (no change to page load latency since fetches are asynchronous)

**Other cleanups/fixes that got rolled into this fix:**

1. In `src/js/common/components/RequiredLoginView.jsx`:  In order to guard whether a user is authorized to use an LOA3-protected service, a container (e.g. `src/js/rx/containers/RxRefillsApp.jsx` will add a child `AppContent` to `RequiredLoginView` and `RequiredLoginView` will control the hiding/showing of `AppContent`. Previously, that prop got its value by a state variable of `RequiredLoginView` but a bug made the value passed in always `undefined`. This change makes it a local variable instead of a state variable.

2. In `src/js/login/containers/Main.jsx `: in addition to the verify URL issues, the logout URL was being updated with `setState()` instead of by dispatching an action.  It wasn't clear why this was done differently (and if I recall, there was a past call center issue where users could not log out, though it should be working in production now). For sanity, this change makes the logout URL fetching and updating identical to how the login and verify URLs are handled.

3. The login code was previously storing both the login URL and the verify URL as fields (named `first` and `third`) inside an object property called `loginUrl`, which didn't seem too helpful, but was definitely confusing, so these have been split into their own properties. In addition, wherever possible, we use `loginUrl` and not `signinUrl` which was sometimes being used.

4. `VerifyPrompt` and `AuthApplicationSection.jsx ` both have their own `verifyUrl` prop, just as `LoginPrompt` has a `loginUrl` prop, so that it doesn't need to be fetched on the fly.

5. Removed some dead logout-related code with bad imports that were causing webpack warnings 

-----

Things not fixed here that will get their own tickets (more to come):

- [ ] Add e2e test for verify flows: https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2843
- [ ] Improve unit test coverage: https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2517
- [ ] Re-evaluate architecture for guarding content from LOA3-protected services (the `AppComponent`/`RequiredLoginView`/`isDataAvailable` stuff)
- [ ] Revisit/understand our criteria for whether a components should map `login` and `profile` state to its props (e.g., `src/js/login/components/SearchHelpSignIn.jsx`) vs. inheriting them from parent component (e.g. `src/js/common/components/RequiredLoginView.jsx`): related to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1577
- [ ] Minor design work for what should happen when url fetches fail for cases when the user gets no indication that something is wrong. Currently an uncaught Promise will log a JS error and the login/verify/logout click just does nothing.

-----
Original comment:

This seemingly tiny change opened up a host of questions/issues for me:
1.  Why does `handleLogin()` not suffer from the same problem? Because it's not used anywhere as far as I can tell (we use the `handleLogin()` in https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/js/login/containers/Main.jsx#L72
2. Uh-oh, `login-helpers.js` has no tests; spent half of yesterday trying to get a very simple test to work and realized I don't understand what causes the `fetch()` to happen (help welcome here)
3. Maybe the test for  `src/js/login/containers/Main.jsx` will be enlightening? Nope, not really tested much: https://github.com/department-of-veterans-affairs/vets-website/blob/master/test/login/containers/Main.unit.spec.jsx
4. We should probably be tracking verify clicks with analytics; will add that once these other issues are resolved
5.  None of the `fetch()` calls (either here or in `Main.jsx`) seem to have fallbacks for when the ID.me server is unavailable. Is that by design or should we be providing a message to the user?


jcmeloni-usds:
on 4, agree.  We track showing the text (https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/js/common/components/authentication/VerifyPrompt.jsx#L13) but don't track click actions like we do for login (login-prompt-displayed, login-link-clicked, login-link-opened).  

on all the rest, i'll just reiterate how happy I am to have F/E engineers again....

test/common/helpers/login-helpers.unit.spec.js
```diff
@@ -0,0 +1,47 @@
+import { expect } from 'chai';
+import sinon from 'sinon';
+
+import { handleVerify } from '../../../src/js/common/helpers/login-helpers.js';
+
+// Does this need to be a React class in order for something to magically call this.serverRequest()?
```
jbalboni: I don't think `this.serverRequest` is a function, it looks like it's a fetch promise that sometimes has `abort` called on it in RequiredLoginView. Though I can't tell if that would every apply to the verify call. I also didn't know `abort` was something you could call on a promise?

raquelromano-gov: Ah, right...it's just a handle. There seems to be a lot of buzz about how to cancel/abort a promise chain which I will not attempt to understand right now. Would be glad to just get some functioning unit tests on this.

test/common/helpers/login-helpers.unit.spec.js
```diff
@@ -0,0 +1,47 @@
...
+      });
+
+      // All of these assertions fail... :(
+      expect(openCallback).to.have.been.called(); // trying something simple first
+      // expect(openCallback.calledWith(url, '_blank', 'params')).to.be.true;
+      expect(fetchStub.firstCall.args[0]).to.contain('/v0/sessions/new');
+      expect(fetchStub.firstCall.args[1].method).to.equal('GET');
+      expect(fetchStub.called).to.be.true;
+
+      dummy.handleVerify();
```
annekainicUSDS: To copy conversation from Slack, I think this can be moved to above the assertions.

test/common/helpers/login-helpers.unit.spec.js
```diff
@@ -0,0 +1,47 @@
...
+    });
+
+    it('should make a fetch request', () => {
+      const dummy = new Dummy;
+      fetchStub.returns({
+        then: (fn) => fn({ ok: true, json: () => Promise.resolve() })
+      });
+
+      // All of these assertions fail... :(
+      expect(openCallback).to.have.been.called(); // trying something simple first
```
annekainicUSDS: And this works if we instead use `sinon.assert.calledOnce(openCallback);`

raquelromano-gov:
@annekainicUSDS ready for another look.  I made some minor changes and added 2 analytics events. 

Testing should be improved: I filed  https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2517 to improve test coverage for both this and `src/js/login/containers/Main.jsx`

I will also file a ticket to handle failures to retrieve the login/verify/signup/logout urls from ID.me both here and in other places where we fetch it. @jcmeloni I didn't stuff a plain html placeholder in because 1) super ugly, and 2) want to understand when/if this happens here and elsewhere and if there's a template we should use for all these cases.

Note: there is a seemingly unrelated test failure, so rerunning CI but will look into it if it persists.

jcmeloni-usds:
@raquelromano-gov Of course. no worries.

src/js/common/helpers/login-helpers.js
```diff
@@ -17,14 +17,30 @@ export function handleLogin() {
...
-    const myVerifyUrl = json.authenticate_via_get;
-    const receiver = window.open(myVerifyUrl, '_blank', 'resizable=yes,scrollbars=1,top=50,left=500,width=500,height=750');
-    receiver.focus();
+    return json.authenticate_via_get;
+  });
+
+  // Open the window outside of the promise callback to avoid the browser blocking it.
+  // In case the url fetch fails, we should fix this to behave more gracefully than closing the window.
+  const receiver = window.open('', '_blank', 'resizable=yes,scrollbars=1,top=50,left=500,width=500,height=750');
+  urlPromise.then(myVerifyUrl => {
```
annekainicUSDS: I'm a little confused what `myVerifyUrl` is in this case. A new function? It looks like we removed declaring the variable `myVerifyUrl` and setting it to `json.authenticate_via_get` from the old code, so it's no longer defined up until this point, right?

annekainicUSDS (commented):
Looks good! One question though.

src/js/common/helpers/login-helpers.js
raquelromano-gov: Yeah, but it's still the value returned by the line `return json.authenticate_via_get;` in the promise chain above. Since that is returned in a promise (`urlPromise`), it's only available asynchronously, so we need to a `.then()` to get a handle on it.

At least that's my understanding from having read up on promises last week, as they were new to me. Just dug up this box analogy which may be helpful (see 1st answer to the question http://stackoverflow.com/questions/22536385/setting-a-variable-to-get-return-from-call-back-function-using-promise)

src/js/common/actions/index.js
```diff
@@ -0,0 +1,7 @@
+export const LOG_OUT = 'LOG_OUT';
```
raquelromano-gov: This is used by both `login` and `user-profile`. The import was failing in the `user-profile` reducer so I moved it back into `common/actions` (where it used to be until recently)

src/js/common/components/RequiredLoginView.jsx
```diff
@@ -103,7 +103,7 @@ RequiredLoginView.propTypes = {
   authRequired: PropTypes.number.isRequired,
   serviceRequired: PropTypes.string.isRequired,
   userProfile: PropTypes.object.isRequired,
-  loginUrl: PropTypes.string,
+  loginUrl: PropTypes.object.isRequired
```
raquelromano-gov: I am actually unclear about when we do/don't require props, but this seems like something we'd want to require either here or in `LoginPrompt` and `VerifyPrompt`. On second thought maybe better there?

src/js/common/components/authentication/VerifyPrompt.jsx
```diff
@@ -1,31 +1,40 @@
...
   render() {
     return (
       <div className="row primary">
         <div className="medium-12 small-12 columns">
           <div className="react-container">
             <h1>Verify your Identity with ID.me</h1>
             <p>You need to verify your identity to access this part of Vets.gov.</p>
-            <p>To access Vets.gov services, you'll need to verify your identity through ID.me, a third party service that provides strong identity verification. We have added this protection to increase security for your information. Here's what you'll need:
+            <p></p>
+            <div>To access Vets.gov services, you'll need to verify your identity through ID.me, a third party service that provides strong identity verification. We have added this protection to increase security for your information. Here's what you'll need:
```
raquelromano-gov: There was a warning/error (can't remember) about `<div>` inside a `<p>`

src/js/common/components/authentication/VerifyPrompt.jsx
```diff
@@ -36,4 +45,8 @@ class VerifyPrompt extends React.Component {
   }
 }
 
+VerifyPrompt.propTypes = {
+  verifyUrl: PropTypes.string
```
raquelromano-gov: Should this be required, and same for `LoginPrompt`'s `loginUrl`? (instead of requiring them in`RequiredLoginView`?)

src/js/disability-benefits/containers/DisabilityBenefitsApp.jsx
```diff
@@ -43,7 +43,7 @@ class DisabilityBenefitsApp extends React.Component {
     const { available, authorized } = this.props;
 
     return (
-      <RequiredLoginView authRequired={3} serviceRequired={"evss-claims"} userProfile={this.props.profile} loginUrl={this.props.signInUrl}>
+      <RequiredLoginView authRequired={3} serviceRequired={"evss-claims"} userProfile={this.props.profile} loginUrl={this.props.loginUrl}>
```
raquelromano-gov: Just renaming for clarity

src/js/user-profile/components/AuthApplicationSection.jsx
```diff
@@ -1,18 +1,26 @@
...
-import { handleVerify } from '../../common/helpers/login-helpers.js';
 
 class AuthApplicationSection extends React.Component {
   constructor(props) {
     super(props);
-    this.handleVerify = handleVerify;
+    this.handleVerify = this.handleVerify.bind(this);
+  }
+
+  handleVerify() {
```
annekainicUSDS: This looks identical to the `handleVerify` function in `VerifyPrompt`, save for the prop with the url being slightly different. Is there a reason you removed this from the common helpers and put it in both files directly instead?

annekainicUSDS (commented):
From reviewing all the changes, looks like these are main ones:
1. Moving `logOut` back to common actions
2. Renaming all props with the `loginUrl` to be `loginUrl` for consistency (some apps were calling it `signupUrl`)
3. Storing two urls under `loginUrl`: `first` (used for login) and `third` (used for verify)
4. Passing the `verifyUrl` directly to the `VerifyPrompt` component
5. Moving `handleVerify` from common helpers to the two components it is used in: `VerifyPrompt` and `AuthApplicationSection`
6. The original change of only opening a new window for the id.me flow if a `loginUrl` exists

src/js/common/components/RequiredLoginView.jsx
```diff
@@ -30,18 +30,16 @@ class RequiredLoginView extends React.Component {
...
     }
   }
 
   render() {
     let view;
 
-    const loginComponent = <LoginPrompt loginUrl={this.props.loginUrl}/>;
-    const verifyComponent = <VerifyPrompt/>;
+    const loginComponent = <LoginPrompt loginUrl={this.props.loginUrl.first}/>;
+    const verifyComponent = <VerifyPrompt verifyUrl={this.props.loginUrl.third}/>;
```
raquelromano-gov: Yeah--those fields were already there, though I don't think they're that helpful vs. just having 2 separate props. Unless...maybe there's a way to fetch both URLs at once, in which case it would be helpful to be able to stuff them into the same object all at once. I will poke around the ID.me logs and/or ask on slack if anyone knows

src/js/common/components/RequiredLoginView.jsx
omgitsbillryan: `RequiredLoginView` shouldn't require a `loginUrl` because of the times when the user is already logged in at the required LOA.  In these cases, the `RequiredLoginView` will simply render the child components - no `loginUrl` necessary.

raquelromano-gov: I was thinking the component that needs the props should be responsible for requiring them, and the parent shouldn't need to require the prop that will be passed down. E.g., if the child component were conditionally rendered, the parent wouldn't need to have a defined value for the prop.

src/js/common/components/authentication/VerifyPrompt.jsx
```diff
@@ -1,31 +1,40 @@
...
-    this.handleVerify = handleVerify;
+    this.handleVerify = this.handleVerify.bind(this);
   }
 
   componentDidMount() {
-		// event for google analytics that we prompted to verify account
     window.dataLayer.push({ event: 'verify-prompt-displayed' });
   }
 
+  handleVerify() {
```
omgitsbillryan: Why the move `handleVerify()` out of `login-helpers.js` and into `VerifyPrompt.jsx`?  This function is also used by the `AuthApplicationSelection` component.  The move makes things less DRY.

omgitsbillryan (commented):
Love the addition spec coverage.  Have a few comments/concerns.

src/js/common/components/RequiredLoginView.jsx
raquelromano-gov: Looking over https://github.com/department-of-veterans-affairs/vets.gov-team/blob/master/Products/Identity/reference_documents/authentication_and_authorization.md, I think this has to be 2 separate fetches

src/js/common/components/authentication/LoginPrompt.jsx
```diff
@@ -9,20 +9,27 @@ class LoginPrompt extends React.Component {
...
-    // event for google analytics that we prompted for login
     window.dataLayer.push({ event: 'login-prompt-displayed' });
   }
 
   handleLogin() {
+    window.dataLayer.push({ event: 'login-link-clicked' });
     const myLoginUrl = this.props.loginUrl;
-    const receiver = window.open(`${myLoginUrl}&op=signin`, '_blank', 'resizable=yes,scrollbars=1,top=50,left=500,width=500,height=750');
-    receiver.focus();
+    if (myLoginUrl) {
```
raquelromano-gov: I think we still need it because the url could be null, e.g. if the component renders before the URL is returned. If it's null, we don't want to open up an empty window.

I am going to open a separate issue for deciding what to do if the request fails (e.g., if the user loses connectivity).

src/js/login/containers/Main.jsx
```diff
@@ -141,7 +145,7 @@ class Main extends React.Component {
 const mapStateToProps = (state) => {
   const userState = state.user;
   return {
-    login: userState.login,
+    loginUrl: userState.loginUrl,
```
raquelromano-gov: I think naming of these fields is confusing.  I'm going to either
1)  rename `login.loginUrl` to `login.loginUrls` so it's clear when we're talking about the object (which holds 2 URLS, one for login and one for verification) and the actual loginURL (the 'first' property),
or
2) just make `verifyUrl` its own prop and get rid of the `loginUrl object; this will require a separate action to update the `verifyUrl` but might be clearer, since we have a separate `logoutUrl` prop already

LMK if you have opinions about one or the other

src/js/login/containers/Main.jsx
```diff
@@ -161,5 +165,7 @@ const mapDispatchToProps = (dispatch) => {
   };
 };
 
-export default connect(mapStateToProps, mapDispatchToProps, undefined, { pure: false })(Main);
+// TODO(romano): safe to remove { pure: false } now?
```
raquelromano-gov: OK, thanks--I had not read it carefully. Yes, sounds like as long as we're using the store, we should keep this here.

src/js/user-profile/components/AuthApplicationSection.jsx
raquelromano-gov: We chatted about this by phone, but replying here for the logs. I believe the old `handleVerify` was in the shared `login-helpers.js` code precisely for that reason (same code needed 2 places), but that handler was doing 2 things: fetching the url and opening the popup window.

I guess we could keep it here but we would still need to update the component prop with the value returned by the fetch. I guess the shared code could fire an action and any component that needed the url could update itself in its reducer. OK, will try this.

test/user-profile/containers/UserProfileApp.unit.spec.jsx
```diff
@@ -0,0 +1,39 @@
...
+  });
+
+  it('should render login view', () => {
+    tree = SkinDeep.shallowRender(
+      <UserProfileApp
+          store={store}
+          loginUrl={'http://fake-login'}
+          profile={{}}/>
+    );
+    // TODO(romano): figure out why this doesn't render the login view
```
raquelromano-gov: @omgitsbillryan I need to go read the SkinDeep docs to understand what's going on here, but if you have any ideas, LMK!

src/js/login/containers/Main.jsx
annekainicUSDS: The second option might be slightly clearer, but for the sake of getting this PR in, I think the first option still works well and we can always refactor that later :).

src/js/user-profile/components/AuthApplicationSection.jsx
annekainicUSDS: Yep, sounds like what existed in `login-helpers.js` for `handleVerify` isn't how we want to handle this going forward, but the new `handleVerify` functions seem very similar in both places, so we could possibly put that back into `login-helpers.js`.

src/js/common/components/RequiredLoginView.jsx
raquelromano-gov: I'd like for  `LoginPrompt`, `VerifyPrompt` and `AuthApplicationSection` require their url props (either `loginUrl` or `verifyUrl`, but because of the async fetch, they are not always available by the time the component is created. So, I've removed all of the requirements and left the check in the handlers to make sure these are well-defined before popping open the ID.me window.

src/js/common/actions/index.js
raquelromano-gov: The reason I'd moved it was because it seemed odd to include an action from one react component into another. But yes, it was sort of lonely there (which is probably why it was moved out of common and into login in the first place). In any case, I've moved it back into the login code and willing to live with importing an action from the login code into the user-profile code.

src/js/common/components/authentication/VerifyPrompt.jsx
raquelromano-gov: I don't remember where I saw this warning, so reverted for now. Will see if it returns.

src/js/common/components/authentication/LoginPrompt.jsx
raquelromano-gov: See above comment about component rendering before the url is available.

src/js/common/components/authentication/VerifyPrompt.jsx
raquelromano-gov: Moved it back!

src/js/common/actions/index.js
omgitsbillryan: Sounds good.  Keep in mind I'm fairly new to React, I'm not sure how faux pas it is for a component to import the actions of another component :smile:.  I yield to your best judgement.

src/js/common/components/RequiredLoginView.jsx
```diff
@@ -55,27 +53,29 @@ class RequiredLoginView extends React.Component {
...
+              // no data. (Only add this prop to React components (functions), and not ordinary
+              // DOM elements.)
               view = React.Children.map(this.props.children,
                 (child) => {
                   let props = null;
                   if (typeof child.type === 'function') {
-                    props = {
-                      isDataAvailable: this.state.isServiceAvailableForUse,
-                    };
+                    props = { isDataAvailable: false };
```
omgitsbillryan: Is this right? Defaulting to `false` instead of `this.state.isServiceAvailableForUse` basically tells child components that the user is LOA3 but does not have the necessary MVI attributes to use said resource.

raquelromano-gov: Yeah, I have no idea why I did that, and why things still work on a dev demo (which they shouldn't!)

raquelromano-gov: Ah, because if we're in this clause, then `this.state.isServiceAvailableForUse` should be false

omgitsbillryan: You are correct, however, we are within the `map()` function context, so `this` is not `RequiredLoginView`.  In it's previous form, I think that `this.state.isServiceAvailableForUse` within the `map()` context would evaluate to `undefined` - which in itself is a bug.

I think this code only worked before because a bug allowed it to :man_facepalming: which is generally a good indication that it needs refactoring.

omgitsbillryan: What's more troubling is that the correction is to make it read:
```jsx
props = { isDataAvailable: undefined };
```

:man_facepalming: :woman_facepalming:

test/common/components/authentication/VerifyPrompt.unit.spec.jsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import SkinDeep from 'skin-deep';
+import { expect } from 'chai';
+
+import VerifyPrompt from '../../../../src/js/common/components/authentication/VerifyPrompt.jsx';
+
+describe('<LoginPrompt>', () => {
```
annekainicUSDS: I think you mean `<VerifyPrompt>` here?

test/common/components/RequiredLoginView.unit.spec.jsx
```diff
@@ -98,10 +104,28 @@ describe('<RequiredLoginView>', () => {
     );
     tree.getMountedInstance().setState({
       loading: false,
-      isServiceAvailableForUse: false
     });
 
-    // each direct child should get their props.isDataAvailable set
+    // Child components should not be passed an isDataAvailable prop
+    tree.props.children.forEach((child) => {
+      expect(child.props.isDataAvailable).to.be.undefined;
```
omgitsbillryan: Not your job since this behavior existed before you refactored, but I've logged an issue to address the fact that the possible values for `isDataAvailable` is `undefined` & `false`.  I think ideally, it should be `true` or `false`.

https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2752

test/user-profile/containers/UserProfileApp.unit.spec.jsx
annekainicUSDS: I'm spending some time trying to debug this, made harder by our testing tools... will report if I figure this out, but after reading the docs it should be working...

annekainicUSDS (commented):
This is looking so much cleaner and more organized! Really excited about these changes coming in :)

robertfairhead:
@raquelromano-gov I'm guessing this is the PR you mentioned on Slack. We already had a config for these from when courtney put them in originally. I updated it a bit so that now any custom events matching this regex ^(login|register|logout|verify).* will get reported under the "Login" category in Vets.gov all.

I think that should take care of the items in here.

src/js/common/components/RequiredLoginView.jsx
raquelromano-gov: See above--fixed this by not using `setState()` after all since this value is only used once, which seems to make both the behavior and the tests work.

src/js/common/components/RequiredLoginView.jsx
```diff
@@ -9,39 +9,30 @@ import LoadingIndicator from '../../common/components/LoadingIndicator';
...
     const requiredApp = this.props.serviceRequired;
     const userServices = this.props.userProfile.services;
-    if (userServices) {
-      this.setState(
-        {
-          isServiceAvailableForUse: userServices.includes(requiredApp),
-        }
-      );
-    }
+    // TODO: replace indexOf() once NodeJS versions in all environments support includes()
```
raquelromano-gov: Yes, just upgraded locally. I'll try this after everything else is approved, and make sure everything passes both locally and on Jenkins. I can't bear to do it now and possibly see new failures.  😩

test/common/components/RequiredLoginView.unit.spec.jsx
raquelromano-gov: Thanks

src/js/common/components/RequiredLoginView.jsx
```diff
@@ -9,39 +9,30 @@ import LoadingIndicator from '../../common/components/LoadingIndicator';
...
+    return sessionStorage.userToken && userServices &&
```
raquelromano-gov: Done!

raquelromano-gov:
@omgitsbillryan @annekainicUSDS Please take another look.  I've added the "front matter" to the top of the PR description summarizing all that's here, so hopefully that helps navigate it a bit.

src/js/common/helpers/login-helpers.js
```diff
@@ -4,28 +4,13 @@ import environment from './environment.js';
...
-  this.serverRequest = fetch(`${environment.API_URL}/v0/sessions/new?level=3`, {
-    method: 'GET',
-  }).then(response => {
-    return response.json();
-  }).then(json => {
-    const myVerifyUrl = json.authenticate_via_get;
-    const receiver = window.open(myVerifyUrl, '_blank', 'resizable=yes,scrollbars=1,top=50,left=500,width=500,height=750');
+export function handleVerify(verifyUrl) {
+  window.dataLayer.push({ event: 'verify-link-clicked' });
+  if (verifyUrl) {
```
omgitsbillryan: If `verifyUrl` is passed as `null`, this will be a no-op.  Can we instead have an `else` condition that fetches the verify URL (a.k.a. LOA3 url) then opens it in a new window for the unlikely event that `verifyUrl` is `null`?

omgitsbillryan (approved):
LGTM with one small comment you can take or leave. :santa:

src/js/common/helpers/login-helpers.js
annekainicUSDS: Since we've now separated the actions for fetching the url and opening the window, and this `handleVerify` should only be responsible for opening the window; not sure we want to add an `else` condition that would fetch the url. I think Raquel had added a TODO to figure out a way to handle the case where the fetch of the url fails in a future PR; this probably needs more investigation before adding anything here.

annekainicUSDS (approved):
Reviewed again and looks great! I would only want to make sure that all the apps affected by this are working as expected. Once the sprint teams sign off on their apps, this looks good to go. I can help with trying to dig into some of the TODOs in the tests, but I don't think that should hold up this PR any more.

src/js/common/helpers/login-helpers.js
omgitsbillryan: Gotcha, fewer responsibilities is better 👍.  Perhaps a `Sentry` error log would be appropriate.

raquelromano-gov: Yeah, what Anne said. Basically, we need to decouple the user action (button/link click) from the URL fetch since the fetch callback can't open the window once the URL response is returned since most browsers will block it.  This is on the list of issues to file as follow-up to this PR.

jbalboni (approved):
I'm able to log in locally to claim status without any issues. I'd like to try it on staging, too, but seems fine to me.

U-DON (approved):
LGTM. Was able to log in on Rx, Secure Messaging, and Health Records.

PR #5456, opened 2017-04-29
HCA Chapter 4 - Household Information

cvalarida:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2235

As I was taking the screenshots, I noticed I missed scrolling down on the annual income section, so...I guess I'm not quite done with this, but I'm putting up the PR anyhow. I'll finish it on Monday.

## Financial Disclosure
image
image

## Spouse Information
image
image
image
image

## Child Information
image
image
image
image

## Annual Income
image
More to come...I missed scrolling down on this 😒 

## Deductible Expenses
image


src/js/hca-rjsf/config/form.js
```diff
@@ -265,7 +296,199 @@ const formConfig = {
...
+            cohabitedLastYear: {
+              'ui:title': 'Did your spouse live with you last year?',
+              'ui:widget': 'yesNo'
+            },
+            provideSupportLastYear: {
+              'ui:title': 'If your spouse did not live with you last year, did you provide financial support?',
+              'ui:widget': 'yesNo',
+              'ui:options': {
+                // Only show if 'No' is selected for cohabitedLastYear
+                hideIf: (formData) => formData.cohabitedLastYear !== false
```
jbalboni: It would be better to use expandUnder here, but I guess that's tricky because it shows up on "No"?

src/js/hca-rjsf/config/form.js
```diff
@@ -265,7 +296,199 @@ const formConfig = {
...
+              }
+            },
+            'view:spouseContactInformation': {
+              'ui:title': 'Spouse’s address and telephone number',
+              'ui:options': {
+                // Only show if 'No' is selected for sameAddress
+                hideIf: (formData) => formData.sameAddress !== false
```
jbalboni: Same as above. It'd be better, I think, to move this question below the same address question and use expandUnder. But maybe we should save that for a separate item if we have to tweak how expandUnder works.

jbalboni (commented):
Looks good so far. This is a giant chapter!

src/js/hca-rjsf/config/form.js
cvalarida: Yeah, that's exactly it. I need it to expand only if "No" is selected, and as far as I know, `expandUnder` will only work if "Yes" is selected.

src/js/hca-rjsf/definitions/child.js
```diff
@@ -0,0 +1,87 @@
...
+  'view:childSupportDescription': {
+    'ui:description': 'Count child support contributions even if not paid in regular set amounts. Contributions can include tuition payments or payments of medical bills.'
+  },
+  childReceivedSupportLastYear: {
+    'ui:title': 'If your dependent child did not live with you last year, did you provide support?',
+    'ui:widget': 'yesNo',
+    'ui:options': {
+      // Not being invoked until the data is changed...which means this is open
+      //  by default
+      hideIf: (formData, index) => formData.children[index].childCohabitedLastYear !== false
```
jbalboni: Yeah, it looks like the bug here is that the ArrayField component uses `additionalItems` to show fields before a user has entered any data, but `additionalItems` hasn't ever had the update functions run against it. I don't want to try to make that change in this PR, since it could have other effects I'm not thinking about.

The quick hack fix for this is to set `'ui:hidden': true` in this field's schema above.

jbalboni (approved):
Looks good to me. If you haven't yet, please create an issue for the child array changes that you're skipping in this PR.

src/js/hca-rjsf/config/form.js
```diff
@@ -10,23 +10,28 @@ import {
...
 } from '../helpers';
 
 import IntroductionPage from '../components/IntroductionPage';
 import InsuranceProviderView from '../components/InsuranceProviderView';
+import ChildView from '../components/ChildView';
 
 import { uiSchema as dateUI } from '../../common/schemaform/definitions/currentOrPastDate';
-import { uiSchema as fullNameUISchema } from '../../common/schemaform/definitions/fullName';
-import * as ssn from '../../common/schemaform/definitions/ssn';
+import { uiSchema as fullNameUI } from '../../common/schemaform/definitions/fullName';
```
cvalarida: Oh, I thought you already did so I changed this up a bit. Blast!

src/js/hca-rjsf/config/form.js
```diff
@@ -481,7 +511,232 @@ const formConfig = {
...
+              veteranOtherIncome,
+              'view:spouseIncome': {
+                type: 'object',
+                properties: {
+                  spouseGrossIncome,
+                  spouseNetIncome,
+                  spouseOtherIncome
+                }
+              },
+              'view:childrenIncome': {}
```
cvalarida: Yup!

src/js/hca-rjsf/definitions/child.js
```diff
@@ -0,0 +1,87 @@
+import _ from 'lodash/fp';
+import fullSchemaHca from 'vets-json-schema/dist/10-10EZ-schema.json';
+import { uiSchema as fullNameUI } from '../../common/schemaform/definitions/fullName';
+import { uiSchema as ssnUI } from '../../common/schemaform/definitions/ssn';
+
+
+export const schema = _.merge(_.omit(
```
cvalarida: Yeah, I was thinking it'd be silly to do that unless VA comes out with another health care application (please no!), but I can do that for consistency.

src/js/hca-rjsf/definitions/child.js
cvalarida: Oh, and then `hideIf` will override `'ui:hidden'`? Good to know.

jbalboni: Yeah, that's what `hideIf` is used to set in all that reducer code, then the components look for that property in order to filter what's being shown.

cvalarida:
For the remaining work that needs to be done here: https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2736

(Please feel free to edit the description to make it clearer; turns out I'm having a bad day for stringing words together coherently. 😞 )

PR #5379, opened 2017-04-20
Adding cleaned up process list CSS.

ghost:
- Closes department-of-veterans-affairs/vets.gov-team#657
- Replaces PR department-of-veterans-affairs/vets-website#5266

Replaced step with process-step, list number classes with list-*

PR #5459, opened 2017-05-01
Update edu form intro pages for new process list styles

jbalboni:
Fixes some edu issues that resulted from process list style clean up. The subway map styling disappeared from the edu intro pages.

Before:
screen shot 2017-05-01 at 3 15 32 pm

After:
screen shot 2017-05-01 at 3 15 20 pm


cvalarida (approved):
When did the "breaking" changes occur, and how can we prevent this from happening again?

tiffanybbrown-va:
> When did the "breaking" changes occur, and how can we prevent this from happening again?

They occurred when #5379 was merged. I typically do a find and replace for all instances to prevent things like this. But it's highly likely that I limited that search and replace to the `content` directory instead of `content` and `src`.

cvalarida:
@tiffanybbrown-va Ahh, I gotcha. Thanks!

PR #5462, opened 2017-05-01
Update common pages and definitions to used passed in schemas

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1574

This PR updates all (I think) of our common pages and definitions to remove schemas when we aren't modifying them, in favor of passing the form specific schema as a parameter.


src/js/edu-benefits/1995/config/form.js
```diff
@@ -123,7 +124,7 @@ const formConfig = {
               'view:newService': {
                 type: 'boolean'
               },
-              toursOfDuty
+              toursOfDuty: fullSchema1995.properties.toursOfDuty
```
cvalarida: Is there a reason we're not doing this with all of the `properties`?

test/common/schemaform/definitions/bankAccount.unit.spec.jsx
```diff
@@ -5,7 +5,10 @@ import ReactTestUtils from 'react-dom/test-utils';
 import Form from 'react-jsonschema-form';
 
 import { DefinitionTester } from '../../../util/schemaform-utils.jsx';
-import { schema, uiSchema } from '../../../../src/js/common/schemaform/definitions/bankAccount';
+import uiSchema from '../../../../src/js/common/schemaform/definitions/bankAccount';
+import commonDefinitions from 'vets-json-schema/dist/definitions.json';
```
cvalarida: Can we make make this
```js
import { bankAccount as schema } from 'vets-json-schema/dist/definitions.json';
```
like you do with some of the others in other tests?

test/edu-benefits/pages/applicantService.unit.spec.jsx
```diff
@@ -134,4 +140,9 @@ describe('Edu default applicantService page', () => {
     expect(formDOM.querySelector('.va-growable-background').textContent)
       .to.contain('Army');
   });
+};
+
+describe('Edu applicantServicePage', () => {
```
cvalarida: I like this model.

cvalarida (approved):
Full disclosure: I skimmed through the middle bit because there was so much here. I figure if it passes the tests, it's probably good to go, right? I did have a couple questions, though.

cvalarida:
When the build passes, I'll look again, but it looks good so far.

src/js/common/schemaform/Address.jsx
```diff
@@ -4,12 +4,13 @@ import classNames from 'classnames';
...
 
 import { getDefaultFormState } from 'react-jsonschema-form/lib/utils';
+import commonDefinitions from 'vets-json-schema/dist/definitions.json';
 
 import * as address from './definitions/address';
 import { states } from '../utils/options-for-select';
 import { pureWithDeepEquals } from './helpers';
 
-const defaultStateSchema = address.schema().properties.state;
+const defaultStateSchema = address.schema({ definitions: commonDefinitions }).properties.state;
```
jbalboni: We've already been pulling it in in many of the common definitions, so it's not much of a change from before. It's here because I want to be able to generate the different state lists for any instance of the component to switch between, since that's more efficient. It's tricky to do that based on the schema that's passed in to a particular instance of this component.

I would prefer not to have to do this, though, and be able to remove the common definitions file from our bundle. I think the best option for doing that is converting the address definition to use updateSchema instead of a custom field component (which I think is possible now).

src/js/edu-benefits/1995/config/form.js
jbalboni: I'm not sure what you mean.

jbalboni:
@cvalarida This is ready for another look.

src/js/edu-benefits/1995/config/form.js
cvalarida: This just seems less verbose than what we're doing for most of the properties we're pulling out of the form schemas and using once.

```js
const {
  someProp,
  anotherProp,
  moreProp,
  etc
} = fullSchema1234.properties;
...
```

cvalarida (approved):
Yup! Looks good!

PR #5463, opened 2017-05-01
Hca rjsf chapter 1 remaining pages

annekainicUSDS:
Remaining 3 pages of Chapter 1 (merging it into the original PR will, I think, make it easier to review this one while the other PR is still open. Connected to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2230.

#### Demographic Information
screen shot 2017-05-01 at 4 53 15 pm

#### Veteran Address
screen shot 2017-05-01 at 4 53 31 pm

#### Contact Information
screen shot 2017-05-01 at 4 53 45 pm

### DIFF
--- a/src/js/common/schemaform/Address.jsx
+++ b/src/js/common/schemaform/Address.jsx
@@ -171,6 +171,20 @@ class Address extends React.Component {
             formContext={formContext}
             onChange={(update) => {this.handleChange('street2', update);}}
             onBlur={onBlur}/>
+
+        {schema.properties.street3 ? <SchemaField
+            name="street3"
+            required={this.isRequired('street3')}
+            schema={schema.properties.street3}
+            uiSchema={uiSchema.street3}
+            idSchema={idSchema.street3}
+            formData={formData.street3}
+            errorSchema={errorSchema.street3}
+            registry={registry}
+            formContext={formContext}
+            onChange={(update) => {this.handleChange('street3', update);}}
+            onBlur={onBlur}/> : null}
+
         <SchemaField
             name="city"
             required={this.isRequired('city')}
--- a/src/js/common/schemaform/definitions/address.js
+++ b/src/js/common/schemaform/definitions/address.js
@@ -48,6 +48,9 @@ export function uiSchema(label = 'Address') {
     street2: {
       'ui:title': 'Line 2'
     },
+    street3: {
+      'ui:title': 'Line 3'
+    },
     city: {
       'ui:title': 'City'
     },
--- a/src/js/hca-rjsf/config/form.js
+++ b/src/js/hca-rjsf/config/form.js
@@ -225,11 +225,7 @@ const formConfig = {
           title: 'Permanent address',
           initialData: {},
           uiSchema: {
-            address: _.merge(address.uiSchema('Permanent address'), {
-              street3: {
-                'ui:title': 'Line 3'
-              }
-            })
+            address: address.uiSchema('Permanent address')
           },
           schema: {
             type: 'object',
--- a/src/js/hca-rjsf/config/form.js
+++ b/src/js/hca-rjsf/config/form.js
@@ -2,7 +2,12 @@ import _ from 'lodash/fp';
 
 import fullSchemaHca from 'vets-json-schema/dist/10-10EZ-schema.json';
 
-import { states } from '../../common/utils/options-for-select';
+import {
+  states,
+  genders,
+  maritalStatuses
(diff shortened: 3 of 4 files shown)


src/js/hca-rjsf/config/form.js
```diff
@@ -147,22 +163,146 @@ const formConfig = {
...
+          },
+          schema: {
+            type: 'object',
+            properties: {
+              address: _.merge(address.schema(true), {
+                properties: {
+                  street: {
+                    type: 'string',
+                    minLength: 1,
+                    maxLength: 30
```
annekainicUSDS: Current HCA has these maxLenghths for certain address fields due to ES requirements.

test/hca-rjsf/config/veteranInformation.unit.spec.jsx
```diff
@@ -34,4 +34,46 @@ describe('HCA veteranInformation', () => {
     expect(formDOM.querySelectorAll('input, select').length).to.equal(6);
     expect(formDOM.querySelector('#root_veteranSocialSecurityNumber')).not.to.be.null;
   });
+
+  it('should render demographicInformation page', () => {
```
jbalboni: We should have some tests for making sure you get the right number of required field errors when submitting on these pages.

jbalboni (approved):
Looks good, just needs a couple more tests, I think.

test/hca-rjsf/config/veteranInformation.unit.spec.jsx
annekainicUSDS: Good point, will add that!

annekainicUSDS:
Added tests for required fields for all pages in this chapter.

src/js/common/schemaform/Address.jsx
```diff
@@ -171,6 +171,20 @@ class Address extends React.Component {
             formContext={formContext}
             onChange={(update) => {this.handleChange('street2', update);}}
             onBlur={onBlur}/>
+
+        {schema.properties.street3 ? <SchemaField
+            name="street3"
+            required={this.isRequired('street3')}
+            schema={schema.properties.street3}
+            uiSchema={uiSchema.street3}
```
cvalarida: I'm not sure that it'll _really_ matter, but in `schemaform/definitions/address.js`, could we add a default `uiSchema.street3` like the other fields have? If nothing else, when we look at the definition file it'll act as an indication that we can, in fact, have 3 street lines if necessary.

test/hca-rjsf/config/veteranInformation.unit.spec.jsx
```diff
@@ -2,36 +2,107 @@ import React from 'react';
...
+      <DefinitionTester
+          schema={schema}
+          data={{}}
+          onSubmit={onSubmit}
+          uiSchema={uiSchema}/>
+    );
+    const formDOM = findDOMNode(form);
+
+    expect(formDOM.querySelectorAll('input, select').length).to.equal(8);
+    expect(formDOM.querySelector('#root_gender')).not.to.be.null;
```
cvalarida: Why is this here? Same with `#root_veteranFullName_first` above and `#root_address_country` below.

cvalarida (approved):
Looks good. Just had a couple questions.

test/hca-rjsf/config/veteranInformation.unit.spec.jsx
annekainicUSDS: This is a pattern we were using in the 1990n tests, which is where I think I started from in creating the HCA tests. It just checks that the first field for that page is actually rendered. It's a bit more specific to this page than just the number of inputs and selects. Open to feedback if there is a better way to test this.

cvalarida: Nah, I think it's fine. Just something I hadn't seen on any of the edu forms and was wondering if we should be doing that there.

src/js/common/schemaform/Address.jsx
annekainicUSDS: Yep, I can add that

src/js/common/schemaform/definitions/address.js
```diff
@@ -48,6 +48,9 @@ export function uiSchema(label = 'Address') {
     street2: {
       'ui:title': 'Line 2'
     },
+    street3: {
```
annekainicUSDS: @cvalarida per your suggestion added this here :)

PR #5476, opened 2017-05-03
Enable NavButtons to handle endpoints with slash (#2278)

cehsu:
Connects to department-of-veterans-affairs/vets.gov-team#2278
screen shot 2017-05-03 at 10 50 46 am

### DIFF
--- a/src/js/common/components/NavButtons.jsx
+++ b/src/js/common/components/NavButtons.jsx
@@ -81,7 +81,7 @@ export default class NavButtons extends React.Component {
     );
 
     let buttons;
-    if (path.endsWith('review-and-submit') || path.endsWith('review-and-submit/')) {
+    if (path.match(/review-and-submit(\/)?$/)) {
       let submitButton;
       let submitMessage;
 
@@ -147,7 +147,7 @@ export default class NavButtons extends React.Component {
           </div>
         </div>
       </div>);
-    } else if (path.endsWith('submit-message') || path.endsWith('submit-message/')) {
+    } else if (path.match(/submit-message(\/)?$/)) {
       buttons = (
         <div className="row form-progress-buttons">
           <div className="small-6 medium-5 columns">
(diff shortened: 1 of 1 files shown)


src/js/common/components/NavButtons.jsx
```diff
@@ -81,7 +81,7 @@ export default class NavButtons extends React.Component {
     );
 
     let buttons;
-    if (path.endsWith('review-and-submit')) {
+    if (path.endsWith('review-and-submit') || path.endsWith('review-and-submit/')) {
```
jbalboni: We could remove the `/` from path first, that would save some duplication.

jbalboni (approved):
Looks good to me. Had one nitpicky comment.

src/js/common/components/NavButtons.jsx
cehsu: I've gone ahead and implemented using regex instead.

src/js/common/components/NavButtons.jsx
```diff
@@ -81,7 +81,7 @@ export default class NavButtons extends React.Component {
     );
 
     let buttons;
-    if (path.endsWith('review-and-submit')) {
+    if (path.match(/review-and-submit(\/)?$/)) {
```
cvalarida: Fun fact, `.match()` is much slower than `.endsWith()`:
image
It's still so fast it doesn't really matter, but I was curious, so I wrote a little jsperf to see what the comparison is.

cvalarida (approved):

src/js/common/components/NavButtons.jsx
cehsu: Thanks @cvalarida, I hadn't compared the performance, but would it make more sense to include all of the potential code if doing so? Let me know if you have more ideas for speeding things up. And thanks for the code review.
screen shot 2017-05-08 at 4 18 29 pm

cvalarida: Nah, I'd just leave it with the `.match()`. 😄 This was mostly just my idle curiosity since the difference will be a fraction of a millisecond most of the time, methinks.

cehsu: Sounds good. In case you're interested in more like this: https://stackoverflow.com/questions/280634/endswith-in-javascript

iantopper:
@J-Quag please validate

J-Quag:
I dont know if this is for me to validate.

iantopper:
In that case, Jeff and Chris reviewed. moving to done.

PR #5477, opened 2017-05-03
Disability deprecate old claim status url

cehsu:
Connects to department-of-veterans-affairs/vets.gov-team#1465

### DIFF
--- a/src/js/disability-benefits/routes.jsx
+++ b/src/js/disability-benefits/routes.jsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { Route, IndexRedirect } from 'react-router';
+import { Route, IndexRedirect, Redirect } from 'react-router';
 
 import YourClaimsPage from './containers/YourClaimsPage.jsx';
 import ClaimPage from './containers/ClaimPage.jsx';
@@ -12,6 +12,10 @@ import AdditionalEvidencePage from './containers/AdditionalEvidencePage.jsx';
 import ClaimEstimationPage from './containers/ClaimEstimationPage.jsx';
 
 const routes = [
+  <Redirect
+      key="/track-claims/your-claims"
+      from="/disability-benefits/track-claims"
+      to="/your-claims"/>,
   <Route
       component={YourClaimsPage}
       key="/your-claims"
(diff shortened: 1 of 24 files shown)


jbalboni:
@cehsu Searching the repo for `disability-claims/track-claims`, I came up with more links that we should change. Can you do the same and update the ones you find?

cehsu:
Thanks @jbalboni. I've gone through and replaced all instances, as well as implemented a redirect. Your thoughts on these changes would be welcome when convenient.

src/js/disability-benefits/routes.jsx
```diff
@@ -12,6 +12,10 @@ import AdditionalEvidencePage from './containers/AdditionalEvidencePage.jsx';
 import ClaimEstimationPage from './containers/ClaimEstimationPage.jsx';
 
 const routes = [
+  <Redirect
```
jbalboni: Did you test this to see that it works? It's a bit surprising to me that it would work, since the base url is set to `track-claims` and is prepended to urls in the route config here.

jbalboni:
@cehsu It seems like there are some strange changes in the diff. Did you start with master when you created this branch?

src/js/disability-benefits/routes.jsx
cehsu: Thanks for your question @jbalboni. Navigating to /disability-benefits/track-claims results in the following:
screen shot 2017-05-09 at 10 06 20 am
---------------------------------------------------------------------------------------------
screen shot 2017-05-09 at 10 06 39 am
---------------------------------------------------------------------------------------------
screen shot 2017-05-09 at 10 06 43 am

cehsu:
Thanks @jbalboni, extraneous changes tidied up.

src/js/disability-benefits/routes.jsx
jbalboni: Nope, that all makes sense, I just wasn't sure that the redirect would actually work given that we use a base url. But it looks like it does, though I think we should change the `from` to:

`from="/disability-benefits/track-claims*"`

So that we catch any of the old routes, since those won't work anymore.

jbalboni (approved):
This looks good to me. I made one comment about adjusting the redirect path.

Can you wait until after 4pm EST to merge this? That's when our daily deploy cut off is, so doing it after that will give us a day to make sure everything is working in staging before it goes to production.

src/js/disability-benefits/routes.jsx
cehsu: Thanks for the helpful feedback @jbalboni, implemented in commit f412932.

PR #5478, opened 2017-05-03
WIP: Add metalsmith-linkcheck plugin

raquelromano-gov:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1691

I ran this locally and need to chat with someone on devops about whether anything special is needed to get this to be useful for jenkins builds. Locally, building with this plugin generates 2 files (`content/pages/.links_checked.json` and `content/pages/links_failed.json`) but still passes, so we need to fail if the failed links array isn't empty.

Currently, `links_failed.json` shows 16 external broken links, pasted below, and (unhelpfully) 744 internal broken links, so I'm looking into how to configure it to suppress validation of internal links since that should be taken care of by `metalsmith-broken-link-checker` According to the `metalsmith-linkcheck` README, "...relative and root-relative local links are checked by looking for them in the metalsmith files array, and so this may not work if you [are] using local links to things not included in your Metalsmith build."

Failed external links:
```
"https://www.dmdc.osd.mil/mfh/getLinks.do?tab=Services",
"https://www.virtualjobscout.org/",
"http://veteranrecruiting.com/",
"https://myseco.militaryonesource.mil/Portal/",
"https://msepjobs.militaryonesource.mil/msep/",
"https://myseco.militaryonesource.mil/Portal/Media/Default/Collaterals_Catalog/Program_Overview/MyCAA-Helping-Spouses-Reach-Career-Goals.pdf",
"https://jkodirect.jten.mil",
"https://www.score.org/topics/veteran_guards_reservists",
"https://myseco.militaryonesource.mil/Portal/Content/View/2622",
"https://dodtap.mil/index.html",
"https://jst.doded.mil/JST_SPEC.pdf",
"http://vaforvets.va.gov/",
"http://vaforvets.va.gov/hr/RVECS/pages/rvecs-map.asp",
"https://www.dmdc.osd.mil/milconnect",
"https://www.dmdc.osd.mil/milconnect/faces/faqs?_adf.ctrl-state=c4t1chkk8_4",
"https://ermc.amedd.army.mil/landstuhl/services.cfm?MTFinfo_id=733"
```

### DIFF
--- a/script/build.js
+++ b/script/build.js
@@ -418,7 +418,8 @@ smith.use(redirect({
 // generated? metalsmith-linkcheck only checks HTML files
 // https://github.com/gchallen/code.metalsmith-linkcheck
 smith.use(linkcheck({
-  failMissing: true
+  verbose: true,
+  timeout: '30s'
 }));
 
 /* eslint-disable no-console */
--- a/script/build.js
+++ b/script/build.js
@@ -322,7 +322,6 @@ smith.use(sitemap({
   hostname: 'http://www.vets.gov',
   omitIndex: true
 }));
-// TODO(awong): Does anything even use the results of this plugin?
 
 if (!options.watch) {
   smith.use(blc({
@@ -415,10 +414,12 @@ smith.use(redirect({
   '/education/apply-for-education-benefits/': '/education/apply/'
 }));
 
-// TODO: what is the earliest place in this script where we can guarantee all HTML has been built?
-// metalsmith-linkcheck only checks HTML files
+// TODO: what is the earliest place in this script where we can guarantee all HTML files have been
+// generated? metalsmith-linkcheck only checks HTML files
 // https://github.com/gchallen/code.metalsmith-linkcheck
-smith.use(linkcheck());
+smith.use(linkcheck({
+  failMissing: true
+}));
 
 /* eslint-disable no-console */
 smith.build((err) => {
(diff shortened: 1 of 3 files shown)


raquelromano-gov:
@tiffanybbrown-va Interested in your thoughts on this if you have time:

1) There doesn't seem to be a way to configure `metalsmith-linkcheck` to suppress checking relative links, which makes it redundant with `metalsmith-broken-link-checker`.  Would be easy to add this as a config option and submit a patch, but I'm not sure it's worth going down that path. Do you know of any other metalsmith plugins that check absolute links?

2) According to the `metalsmith-linkcheck` docs, this should be run after all HTML files are built. I'm assuming the `use()` calls are executed sequentially, so I've put this last, just before `smith.build()` is called. Still, it seems to complain about valid relative links. :( Back to (1) above, would be nice to suppress these.

Any advice/suggestions?

script/build.js
```diff
@@ -424,6 +414,13 @@ smith.use(redirect({
   '/education/apply-for-education-benefits/': '/education/apply/'
 }));
 
+// TODO: what is the earliest place in this script where we can guarantee all HTML files have been
+// generated? metalsmith-linkcheck only checks HTML files
+// https://github.com/gchallen/code.metalsmith-linkcheck
+smith.use(linkcheck({
+  failMissing: true
```
raquelromano-gov: Argh--just looked at the plugin code, and it turns out this is the only way to configure it to fail, but it *only* checks internal links and fails if they're broken. If external links are broken, it writes them to file and passes. 😠  I'm ready to give up on this library, but will file some issues against it and see if the developer has any interest in adding some config options

raquelromano-gov:
@robbiethegeek in case you have opinions on where to take this

robbiethegeek-usds:
I have looked at this and not sure how much value it will add with the amount of false positives it will trigger based on invalid SSL certs

PR #5474, opened 2017-05-03
Add divider to .va-h-ruled--stars

ghost:
Related to department-of-veterans-affairs/vets.gov-team#2555
Blocks #5479

PR #5479, opened 2017-05-03
Update home page layout to use navigation-links-lists.

ghost:
Closes department-of-veterans-affairs/vets.gov-team#2555
Closes https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2554
Closes https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2553

Blocked by #5474


tiffanybbrown-va:
Need to update the current hero image before merging. See department-of-veterans-affairs/vets.gov-team#2553

melwoodard (changes requested):
1. The "Get the VA services you've earned" tagline should still be two lines on the smallest breakpoint, with the line break occurring between services / you've
2. Per feedback from @suzchap and @amybehrens, let's make the navigation links list **one** column instead of two. We'll get user feedback on the two column variation post-launch.
3. Along with that, please give the navigation links list the `.usa-content` class to keep optimal text line length.
4. Remove the 3rem left/right padding around the links list - it should be the same width as the divider above
5. Increase the margin **above** the stars divider by 8px
6. Decrease the margin **below** the stars divider by 24px

melwoodard (changes requested):
Getting there! A few more bugs popped up with these changes:

1. The divider is off-center on desktop/tablet (but looks good on mobile)

vets_gov


2. The gutters on the left & right sides of the content are inconsistent. All content should have the same amount of spacing on the sides as the divider:

vets_gov

3. On mobile, the white background on the hero tagline should have space on either side of it (~24px), not span the entire width of the hero.

vets_gov

4. You can find the new background image asset [here](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2553#issuecomment-299978855)

tiffanybbrown-va:
With the change from cards to link + title, should everything be clickable, or just the title?

melwoodard:
Everything clickable would be ideal! Same would go for the links lists on the new templates

melwoodard (changes requested):
@tiffanybbrown-va ok, a few last nitpicky changes:

1. The tagline should always be centered - I'm seeing it aligned left at ~445-767px viewport width

vets_gov

2. At the ~481-768px viewport width, there's a larger margin on the right side of the 3rd button, and the buttons are not equal width. There are also large margins between the buttons at the higher end of that range.

vets_gov

3. I forgot this in the original issue - the "check claim status" button should be swapped for "In crisis? Get help now". This button should open the crisis line overlay. I'll add this to the issue too for historical reference.

After these changes I think we're good to put a hold on this and share a link with the wider team to preview. Heroku's saying it can't create a review app because the name is already taken. Does that mean we should just open a fresh PR to trigger an app to build?

tiffanybbrown-va:
Replaced by #5527

PR #5482, opened 2017-05-04
Change failed submission language

annekainicUSDS:
Connected to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2636.

Changing the failed submission language in the `NavButtons` component. Currently this is only used by the 1990. If we want to create a way to customize the language, we can do that in a separate PR.

The call center number we changed this to is the Vets.gov technical help line, so it should be generic enough to work for any app.

screen shot 2017-05-04 at 11 24 07 am


jbalboni (approved):

J-Quag (approved):
Approve! of the message. We can look at it again later to see if maybe we should put in the phone number links and such.

PR #5483, opened 2017-05-04
Add in step to clean working directory after build

CyberKoz:
This will prevent the Jenkins hosts from getting filled with old working directories

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -193,10 +193,10 @@ node('vets-website-linting') {
 
     parallel builds
   }
-}
-
-post {
+  
+  post {
         always {
             deleteDir() /* clean up our workspace */
         }
+  }
 }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -194,3 +194,9 @@ node('vets-website-linting') {
     parallel builds
   }
 }
+
+post {
+        always {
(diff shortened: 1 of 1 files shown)


ayaleloehr:
Pinging @CyberKoz to see if this should still be worked on or closed out?

Jenkinsfile
```diff
@@ -193,4 +193,10 @@ node('vets-website-linting') {
 
     parallel builds
   }
+  
+  post {
+        always {
+            deleteDir() /* clean up our workspace */
```
jkassemi: Does the scope of this reset any previous calls to `dir`?

jkassemi (approved):
Approving (if `deleteDir` can't get into a directory other than the jenkins working directory).

ayaleloehr:
Pinging on whether this should be worked on or closed?

CyberKoz:
I need to follow up on James comment and make sure this can't delete outside its work space

ayaleloehr:
putting this back on @CyberKoz 's radar to do or delegate.

wyattwalter:
I'm going to close this out as it's been sitting for several months without action and doesn't appear to be causing any issues (at least not anymore).

PR #5492, opened 2017-05-05
Footer design updates

ghost:
Closes department-of-veterans-affairs/vets.gov-team#1821
Closes department-of-veterans-affairs/vets.gov-team#1801

- Moved footer styles to a partial.
- Made footer link groups 33% wide
- Fixed an order rendering bug.
- Adjusted footer title font size.
- Changed padding around help desk text.
- Changed placement of playbook link.
- Veterans Crisis Line link in footer is now a button that triggers the modal.
- Refactored header partial to consolidate instances of VCL button.
- Removed toggle-veterans-crisis-line.js since it's no longer needed.


melwoodard (approved):
LGTM design-wise. @tiffanybbrown-va can you find a second set of engineer eyes for a technical review?

src/sass/modules/_m-crisis-line.scss
```diff
@@ -1,37 +1,46 @@
...
   }
 }
 
 .va-crisis-panel {
   color: $color-primary-darkest;
   list-style: none;
   margin: 0;
-  padding: 0;
+  top: 50%;
+  -webkit-transform: translateY(-50%);
```
jbalboni: Do we need the prefix?

src/sass/base/_va.scss
```diff
@@ -208,13 +208,13 @@ hr {
     position: relative;
   }
 
- .va-notice--banner {
+ /*.va-notice--banner {
```
jbalboni: Is there a reason to keep the comment around?

jbalboni (approved):
Looks good to me. I did notice that the feedback icon bumps up against the left edge of the screen on mobile. Is that something that should be addressed in here?

screen shot 2017-05-09 at 12 49 54 pm

src/sass/base/_va.scss
tiffanybbrown-va: No. In the process of removing it now.

tiffanybbrown-va:
The feedback icon looks the same in production. We should fix it, but I don't think it needs to hold up this PR.

PR #5496, opened 2017-05-08
Fix PropType warnings from messaging unit tests

raquelromano-gov:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2670

These are kind of a pain to track down. Usually the warning comes from a test for a parent component of the component with the required prop. There are many more throughout vets-website code (I only fixed these b/c I was trying to figure out where they come from and how easy/painful they are to fix).


jbalboni (approved):
Looks pretty much like what I was doing to fix them in edu.

PR #5500, opened 2017-05-08
Move schemaform styles into separate files for hca

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2445

Despite this being all sass changes, screenshot doesn't make sense.

I went through the entire edu-benefits.scss file and moved the parts relevant to rjsf forms into the the schemaform file. I also moved styles related to the confirmation pages into their own file.

I tried to annotate some of the edu file so that we can remove styles that are used on the old style 1990 form when we're ready.

### DIFF
--- a/src/sass/edu-benefits.scss
+++ b/src/sass/edu-benefits.scss
@@ -7,88 +7,20 @@
 @import "modules/m-schemaform";
 @import "modules/m-modal";
 @import "modules/m-omb-info";
+@import "modules/m-form-confirmation";
 
-$button-stroke: inset 0 0 0 1px;
-
-@media screen and (min-width: 600px) {
-  form {
-    max-width: 32em;
-  }
+// Also in 1990 review
+.edu-benefits-pre {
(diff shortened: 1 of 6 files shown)


src/sass/edu-benefits.scss
```diff
@@ -7,88 +7,20 @@
 @import "modules/m-schemaform";
 @import "modules/m-modal";
 @import "modules/m-omb-info";
+@import "modules/m-form-confirmation";
 
-$button-stroke: inset 0 0 0 1px;
```
cehsu: Great work @jbalboni, just want to make sure that some styles were deliberately removed rather than replaced, as indicated by a quick static comparison.

cehsu (approved):

src/sass/edu-benefits.scss
jbalboni: Yeah, this one was intentionally removed, because that variable is never referenced.

Issue #5508, opened 2017-05-09
Provide trimmed pathname to NavButtons


cehsu:
PR #5476 introduces a bug for pathnames that contain a trailing slash in that clicking the "Get Started" button simply scrolls up rather than navigating forward.

PR #5509, opened 2017-05-09
Nav buttons trimmed path

cehsu:
Connects to department-of-veterans-affairs/vets-website#5508

This change makes it unnecessary to handle trailing slashes in the NavButtons component (since it is now being provided with the trimmed pathname).

Also, although the original issue (department-of-veterans-affairs/vets.gov-team#2278) specified the "Get Started" button on the 1990 form in particular, it also indicated that other forms should be checked, and although other forms (e.g. 1990e, 5490) do not use the NavButton component, they do not handle the slash properly, and have been updated as well.

### DIFF
--- a/src/js/common/schemaform/FormApp.jsx
+++ b/src/js/common/schemaform/FormApp.jsx
@@ -25,10 +25,11 @@ export default class FormApp extends React.Component {
   onbeforeunload = e => {
     const { currentLocation } = this.props;
     const trimmedPathname = currentLocation.pathname.replace(/\/$/, '');
-    const isIntro = trimmedPathname.endsWith('introduction');
-    const isConfirmation = trimmedPathname.endsWith('confirmation');
+    const isIntroductionPage = trimmedPathname.endsWith('introduction');
+    const isConfirmationPage = trimmedPathname.endsWith('confirmation');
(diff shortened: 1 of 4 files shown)


src/js/common/schemaform/FormApp.jsx
```diff
@@ -25,7 +25,9 @@ export default class FormApp extends React.Component {
   }
 
   onbeforeunload = e => {
-    const endpoint = this.props.currentLocation.pathname.split('/').pop();
+    const { currentLocation } = this.props;
+    const trimmedPathname = currentLocation.pathname.replace(/\/$/, '');
+    const endpoint = trimmedPathname.split('/').pop();
```
jbalboni: Should we just change things to do the `endsWith` check, instead of splitting on `/`? Looks like it would mean an update in two places and then we'd be consistent across these two files.

cvalarida (approved):

src/js/common/schemaform/FormApp.jsx
cehsu: Good idea @jbalboni, I've made this more consistent, and extracted this logic into a helper to appease CodeClimate.

PR #5510, opened 2017-05-09
Add unit tests for GIBCT

bshyong:
Adds unit tests for all GIBCT reducers

- [x] autocomplete
- [x] calculator
- [x] constants
- [x] eligibility
- [x] filter
- [x] modals
- [x] preview
- [x] profile
- [x] search
- [x] title

Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2516


U-DON (commented):
Generally LGTM, but had a couple of questions. In particular, not sure if `eq` is a thing that should work.

test/gibct/reducers/autocomplete.unit.spec.js
```diff
@@ -0,0 +1,100 @@
...
+  it('should start autocomplete correctly', () => {
+    const state = autocompleteReducer(
+      initialState,
+      {
+        type: 'AUTOCOMPLETE_STARTED'
+      }
+    );
+
+    expect(state.inProgress).to.be.eq(true);
+    expect(state.suggestions).to.eql([]);
```
bshyong: That's strange that the tests pass—yes, I think it should be `.eql`

test/gibct/reducers/autocomplete.unit.spec.js
```diff
@@ -0,0 +1,100 @@
...
+  it('should change autocomplete term', () => {
+    const state = autocompleteReducer(
+      { ...initialState, facilityCode: 'code' },
+      {
+        type: 'AUTOCOMPLETE_TERM_CHANGED',
+        searchTerm: 'newSearchTerm',
+      }
+    );
+
+    expect(state.facilityCode).to.be.eql(null);
```
U-DON: Nitpick: I actually meant it would read better as `to.eql` than `to.be.eql`. As in, "this value should equal [value]".

PR #5517, opened 2017-05-10
Fix va file number collisions on 5495

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2733

This updates the file number in the applicant section to be relativeVaFileNumber, instead of just vaFileNumber, which is used on the sponsor page.

### DIFF
--- a/src/js/edu-benefits/pages/applicantInformation.js
+++ b/src/js/edu-benefits/pages/applicantInformation.js
@@ -38,6 +38,7 @@ export default function applicantInformation(schema, options) {
   const prefix = (options && options.isVeteran) ? 'veteran' : 'relative';
   const mergedOptions = _.assign(defaults(prefix), options);
   const { fields, required, labels } = mergedOptions;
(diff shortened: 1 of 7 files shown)


src/js/edu-benefits/pages/applicantInformation.js
```diff
@@ -38,6 +38,7 @@ export default function applicantInformation(schema, options) {
   const prefix = (options && options.isVeteran) ? 'veteran' : 'relative';
   const mergedOptions = _.assign(defaults(prefix), options);
   const { fields, required, labels } = mergedOptions;
+  const fileNumberProp = prefix === 'relative' ? 'relativeVaFileNumber' : 'vaFileNumber';
```
jbalboni: We have an issue to refactor this page to use the veteranId field, so this will probably be changed shortly.

The 5495 was the only form I found that was using a file number and this shared page definition.

markolson (approved):
LGTM

cvalarida (commented):
Looks good. I can't find anything else either.

cvalarida (approved):
That last was supposed to be approved. 😒

PR #5518, opened 2017-05-10
Fix styling issue with cancel button for Health Records

bshyong:
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2694


U-DON (approved):
LGTM. It'd be great if you could also include screenshots of visual changes in the future.

PR #5520, opened 2017-05-10
Relocate labels to own file

cehsu:
Connects to department-of-veterans-affairs/vets.gov-team#1576

Would appreciate feedback on whether additional labels should be relocated. Thanks @cvalarida for clarifying this.


cvalarida:
Ah yes, this is a legacy bit of code before @jbalboni refactored the way we handle labels in schemaform. Now the magic happens [here](https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/js/common/schemaform/widgets/SelectWidget.jsx#L53), if I'm not mistaken, when we pass `'ui:options': { labels: labelObject }` (documentation for that can be found [here](https://github.com/department-of-veterans-affairs/vets-website/blob/master/docs/schemaform/form-config.md#uischema-configuration)).

jbalboni (approved):
LGTM

PR #5522, opened 2017-05-10
Email notifications for Secure Messaging

U-DON:
Resolves department-of-veterans-affairs/vets.gov-team#2538.


src/js/common/components/Modal.jsx
```diff
@@ -49,6 +49,10 @@ class Modal extends React.Component {
     }
   }
 
+  componentWillUnmount() {
+    document.removeEventListener('focus', this.state.focusListener, true);
```
U-DON: Unbinds the event listener when the component unmounts. I think this is an issue when a modal renders and then no longer renders in an update of its parent component. I was getting some type errors about the modal being null when I made a parent component of a modal update without containing it.

Tagging @jbalboni for awareness.

src/js/messaging/containers/EmailNotifications.jsx
```diff
@@ -0,0 +1,179 @@
...
+
+import ModalDiscardChanges from '../components/ModalDiscardChanges';
+
+export class EmailNotifications extends React.Component {
+  constructor(props) {
+    super(props);
+    this.handleSubmit = this.handleSubmit.bind(this);
+    this.renderSaveButtons = this.renderSaveButtons.bind(this);
+
+    this.state = { discardChanges: false };
```
U-DON: Most of the modal states are in the Redux store, but wanted to keep this in the component's state. My thinking was that if the modal's usage is limited to a particular parent component, we don't need to track it in the more global Redux state. This limits the number of actions (open and close for each and every modal; yuck) we need to manage as well.

src/js/messaging/containers/ManageFolders.jsx
```diff
@@ -0,0 +1,75 @@
...
+import {
+  deleteFolder,
+  openCreateFolderModal,
+} from '../actions';
+
+import ButtonCreateFolder from '../components/buttons/ButtonCreateFolder';
+import ButtonDelete from '../components/buttons/ButtonDelete';
+import { folderUrl } from '../utils/helpers';
+
+export class ManageFolders extends React.Component {
```
U-DON: This new file just splits the original Settings.jsx into the "Manage folders" tab content.

src/sass/messaging/partials/_messaging-app.scss
```diff
@@ -9,9 +9,15 @@
 
 
 #messaging-app-header {
-  @media (max-width: $medium-screen) {
+  @include media-maxwidth($medium-screen) {
```
U-DON: Plenty of places where we should replace `@media` queries with the mixin, but that should be tackled separately.

src/js/common/components/Modal.jsx
bshyong: Looks good

src/js/messaging/components/FolderNav.jsx
```diff
@@ -132,10 +128,6 @@ class FolderNav extends React.Component {
   }
 }
 
-FolderNav.contextTypes = {
```
bshyong: 👍

src/js/messaging/containers/EmailNotifications.jsx
bshyong: Agreed.

bshyong (approved):
LGTM; do we need to hold off on this merge as well for Monday along with RX notifications?

bshyong:
Something happened with the Jenkins build, so adding an empty commit or merging in master should fix it

bshyong:
^ build failing due to a unit test issue
```

  1) Settings should show the correct folders:
     Error: table not found in tree
      at node_modules/skin-deep/skin-deep.js:77:29
      at Array.forEach (native)
      at Object.SkinDeep.api.dive (node_modules/skin-deep/skin-deep.js:75:13)
      at Context.<anonymous> (test/messaging/containers/Settings.unit.spec.jsx:26:23)
```

bshyong:
LGTM

PR #5525, opened 2017-05-11
Fix child and invalid SSN schema errors on hca

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2457

We're getting a lot of SSN related schema errors in Sentry. They're in two forms:

- Blank SSNs in the child fields
- SSNs with 1 or 2 dashes in any SSN field

The blank SSNs in child fields appear to be because on the review page, the child page isn't hidden based on the financial disclosure question. To fix that, I'm removing the children array before sending to the backend, similar to what happens for spouse info. For the other SSN errors, I've added another test to the validation function.

Ideally, we would also hide the children page correctly on the review page, but that should be fixed in the rjsf rewrite, so I'm going to leave it for now.


src/js/common/model/veteran.js
```diff
@@ -717,6 +717,10 @@ function veteranToApplication(veteran) {
     /* eslint-enable no-param-reassign */
   }
 
+  if (veteran.discloseFinancialInformation.value === 'N') {
+    delete veteran.children; // eslint-disable-line no-param-reassign
```
markolson: Can we set this to an empty array instead? That's how it's done in the conformance tests we have in the backend, and even though they work with the element missing I'd prefer to not chance anything..

jbalboni: @markolson Updated

markolson (approved):

markolson:
Once this gets deployed I'll bunker down and clear out any Sentry errors for it.

src/js/common/utils/validations.js
```diff
@@ -242,7 +242,7 @@ function isValidSSN(value) {
     return false;
   }
 
-  return /^\d{3}-?\d{2}-?\d{4}$/.test(value);
+  return /^\d{9}$/.test(value) || /^\d{3}-\d{2}-\d{4}$/.test(value);
```
cvalarida: How is this any different? I guess this way, if they have one `-`, they have to have the second, but...was that the problem?

Alternatively, would it be out of scope to strip non-numeric characters and force-format it with dashes?

markolson: XXX-XXXXXX was valid with the old regex

PR #5528, opened 2017-05-11
Refactor applicantInformation page to use veteranId

cehsu:
Connects to department-of-veterans-affairs/vets.gov-team#2108, which indicated that 

> We'd like to reuse veteranId, so the logic for it is in one place, in the applicant information page definition. To do that we need to update veteranId to be a function and an option that indicates whether to use veteran or relative for the field name.

Reduces duplicate logic by transforming `veteranId.uiSchema` into a function that takes `prefix` (an option that indicates whether to use veteran or relative for the field name) as a parameter, and calling it in `applicantInformation` (with additional references to `veteranId.uiSchema` updated as well).


cvalarida (approved):
It looks good, but I'm wondering if there's a better name to call it than `veteranId` now.`ssnOrFileNumber` seems a bit heavy-handed, but...well, that's what it is. `personId` doesn't seem like a very good option either, though 😛 

At any rate, `veteranId` is now misleading and we should probably change it.

PR #5504, opened 2017-05-09
Updating the build script and docker file to Node 6.10.3

robbiethegeek-usds:

PR #5530, opened 2017-05-11
Fix node version for Heroku builds

robertfairhead:
Heroku builds are failing after the merge of https://github.com/department-of-veterans-affairs/vets-website/pull/5504 . It's pulling node version from package.json which is now out of date.

       Node.js version (mininum): v6.10.3
       Node.js version (installed): v4.4.7

Bumping this up to be in sync. Also added in yarn version to stay in sync as well.

### DIFF
--- a/package.json
+++ b/package.json
@@ -183,8 +183,8 @@
     "fsevents": "^1.0.7"
   },
   "engines": {
-    "node": "4.4.7",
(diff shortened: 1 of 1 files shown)


annekainicUSDS (approved):
LGTM!

PR #5531, opened 2017-05-11
Convert schemaform address to use updateSchema

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2691

This updates our address definition to use the `updateSchema` option instead of a whole custom component. This should cut down on bundle size and make things more consistent.

Should be no visual changes.


cehsu (approved):

src/js/common/schemaform/definitions/address.js
```diff
@@ -30,14 +48,69 @@ export function schema(currentSchema, isRequired = false) {
...
+        if (country === 'USA') {
+          stateList = usaStates;
+        } else if (country === 'CAN') {
+          stateList = canProvinces;
+        } else if (country === 'MEX') {
+          stateList = mexStates;
+        }
+
+        if (stateList) {
+          // We have a list and it's different, so we need to make schema updates
```
cvalarida: Could this / should this be done in a selector? As long as we're already using reselect... :man_shrugging:

cvalarida (approved):

src/js/common/schemaform/definitions/address.js
```diff
@@ -30,14 +48,69 @@ export function schema(currentSchema, isRequired = false) {
...
+export function uiSchema(label = 'Address', useStreet3 = false) {
+  let fieldOrder = ['country', 'street', 'street2', 'street3', 'city', 'state', 'postalCode'];
+  if (!useStreet3) {
+    fieldOrder = fieldOrder.filter(field => field !== 'street3');
+  }
+
   return {
     'ui:title': label,
-    'ui:field': 'address',
+    // 'ui:field': 'address',
```
cehsu: @jbalboni could you please explain the comment here?

src/js/common/schemaform/definitions/address.js
jbalboni: Ha, good point, I completely forgot about reselect. I will give that a shot.

src/js/common/schemaform/definitions/address.js
jbalboni: I just need to remove it entirely. I commented out when I first started this work.

cvalarida (approved):
Excellent! Yay, selectors!

PR #5534, opened 2017-05-12
Fix IE error on education apply page

jbalboni:
Sentry: http://sentry.vetsgov-internal/vets-gov/website-production/issues/5450/

IE/Edge are throwing errors because I was calling `forEach` on a NodeList, which is wrong (but works in Chrome).


cehsu (approved):
LGTM

PR #5535, opened 2017-05-12
UX improvements to Health Records app

bshyong:
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2753
Also adds some UX improvements to the Health Records app during the report generation process

### DIFF
--- a/src/js/health-records/containers/DownloadPage.jsx
+++ b/src/js/health-records/containers/DownloadPage.jsx
@@ -5,6 +5,7 @@ import moment from 'moment';
 import isMobile from 'ismobilejs';
 
 import AlertBox from '../../common/components/AlertBox';
+import LoadingIndicator from '../../common/components/LoadingIndicator';
 
 import DownloadLink from '../components/DownloadLink';
 import { openModal, closeModal } from '../actions/modal';
@@ -103,6 +104,10 @@ export class DownloadPage extends React.Component {
   }
 
   render() {
+    if (this.props.form.inProgress) {
+      return <LoadingIndicator message="Generating Health Record"/>;
+    }
+
     return (
       <div>
         <h1>Access Your Health Records</h1>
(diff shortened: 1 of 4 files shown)


src/js/health-records/containers/UpdatePage.jsx
```diff
@@ -21,28 +21,31 @@ export class UpdatePage extends React.Component {
...
   }
 
   componentWillUnmount() {
     clearInterval(this.pollRefresh);
+    clearTimeout(this.refreshTimeout);
   }
 
-  handleSkipToDownload(e) {
-    e.preventDefault();
+  submitAndDownload() {
```
U-DON: Should be fine in this case, but noticed this didn't get bound in the constructor. Also, you needed to do it in the `setTimeout` anyway.

src/js/health-records/containers/DownloadPage.jsx
```diff
@@ -102,18 +104,44 @@ export class DownloadPage extends React.Component {
     return <DownloadLink ref={e => { this.pdfDownloadButton = e; }} onClick={linkOnClick} name="PDF File" docType="pdf"/>;
   }
 
+  renderRequestDate() {
+    const { form } = this.props;
+
+    if (!form.requestDate) {
```
U-DON: Nitpick: Noticed this is split into multiple lines while a similar check below is all in one.

src/js/health-records/reducers/form.js
```diff
@@ -33,13 +34,19 @@ export default function form(state = initialState, action) {
...
         ...state,
+        inProgress: false,
         requestDate: new Date().toISOString(),
       });
     case 'FORM_FAILURE':
-      return set('ready', false, state);
+      return set('ready', false, {
+        ...state,
+        inProgress: false,
+      });
```
U-DON: Although I didn't catch it in previous reviews, this and the change above both seem to me like a strange hybrid use of `set`. If we're using the rest operator anyway, the return value might as well be `{ ...state, ready: false, inProgress: false }`.

U-DON (approved):
LGTM. Just some nitpicks.

PR #5537, opened 2017-05-12
Disable broken link check on Heroku review app build

jkassemi:
Link check will still be performed in CI environment. This provides functionality provided for the fast-track option defined in https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2695.

PR #5538, opened 2017-05-12
Strip down builds for Heroku

robertfairhead:
Closes https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2762

Creates a build-unsafe script which strips down the metalsmith build steps and a webpack.heroku.js config that skips building all the React modules.

On my machine the striped down build runs in 44 seconds. The full build runs in about 4 minutes showing a roughly 5x speed improvement.

I always tweaked the procfile to run `npm rebuild node-sass` after every push to try to resolve the missing file descriptor errors we've been having since running once on post-install didn't seem to lick it.

Merging this will await discussion at the next Engineering meeting to make sure skipping the React builds (to focus on content and styling) won't wreak any workflows. /cc @ayaleloehr @rroueche


ayaleloehr:
I didn't realize that this would make Heroku links not work for any React applications.  That's a lot of Vets.gov that won't work anymore on the Heroku links (all our forms, for example).  

I also am concerned about having to now keep webpack and build files in sync between multiple versions.  

Sorry I didn't realize these things earlier; am interested in feedback from others as well or if people think I'm off!

robertfairhead:
I can add the React builds back in. I figured this would make what I meant by "fast path" more clear. Basically only build the static content. The heroku builds don't connect to any backends. So it's basically the same as running off localhost for development purposes. The API is mocked out.

That's why i'd suggest discussing at engineering meeting before doing anything globally to the Heroku builds.

We can determine WIP content PRs and only send those down the fast path to give the content team the quicker turnaround they want.

With respect to duplication, we can either factor the shared sections into a common JS file or I could just add in selective buildtype wrappers in the original file to skip the unneeded parts. I figured both are refinements we could ship after an initial solution (whatever that ends up being). I wanted to avoid messing with the standard build configs at this point while this is in flux to avoid borking any Jenkins builds.

jkassemi:
@CyberKoz - is this still good for inclusion based on your most recent conversations with Heroku support?

robertfairhead:
We're testing this config change on PR-5240 right now to see how it goes. If successful, I'll refactor this so it doesn't split out into different config files, run another quick test on a content PR and then we can merge.

I'm aiming for this to be in master Mon/Tue next week and then we'll update all the content PRs to use this config.

If https://github.com/department-of-veterans-affairs/vets-website/pull/5537 and/or https://github.com/department-of-veterans-affairs/vets-website/pull/5561 get merged to master first, I'll make sure those get incorporated in here.

CyberKoz:
@jkassemi, this changeset ( at least as of right now ) contains the most recent recommendation we have from Heroku

robertfairhead:
@CyberKoz @jkassemi @ayaleloehr Ready for a final review.

In the end the change is very minor. Just switching the heroku build to only use the no-react entry for webpack. Additionally, I incorporated the updated "postinstall" hook from https://github.com/department-of-veterans-affairs/vets-website/pull/5561 into this PR.

I tried a whole bunch of other stuff, but in the end the only thing that really made a substantial difference was skipping the React builds. It wasn't worth forking the configs to save a couple seconds.

Heroku does not properly persist the yarn cache between builds. I looked through some old commits on their nodejs buildpack. They had it enabled for awhile and then disabled it due to some issues encountered. I played with it a bit, but it would not recognize the cache properly for me. If that does work, it'd shave about 20-30s off the ~2 minutes the build currently takes.

jkassemi (approved):
So to clarify for the record - the Heroku review apps will be used for primarily for previewing content changes while the review instances will be used for application review (including VA integrations and react applications)?

robertfairhead:
Yes, that's my understanding. 

Since Heroku does not have access to an API layer and the reverse proxy setup isn't right to allow logins, it's not really suitable for testing the React apps. But it does allow access to those without SOCKS access (e.g., VA stakeholders for content review).

The review instances have a working API so they'd be the primary for anything that needs to be "live." For VA stakeholder review, they'd have to be deployed to Dev or Staging to get beyond the need for SOCKS proxy.

PR #5544, opened 2017-05-14
Edu helpers deprecate only required fields

cehsu:
Connects to department-of-veterans-affairs/vets.gov-team#2311.


test/edu-benefits/1995/01-all-fields.e2e.spec.js
```diff
@@ -32,7 +32,7 @@ module.exports = E2eHelpers.createE2eTest(
     // Benefits eligibility page.
     client
       .waitForElementVisible('label[for="root_benefit"]', Timeouts.slow);
-    EduHelpers.completeBenefitsSelection(client, testData.benefitSeletion, true);
+    EduHelpers.completeBenefitsSelection(client, testData.data);
```
jbalboni: Oh, interesting. Was this sending bad data before?

jbalboni (approved):
This looks good to me. Thank you for doing this, I'm sure it was a pain!

test/edu-benefits/1995/01-all-fields.e2e.spec.js
cehsu: Yes, in this case testData has no `benefitSeletion` property, and in fact, `completeBenefitsSelection` accesses the `benefit` property on the `data` parameter, and `data` is the only property on `testData` with a `benefit` property.

PR #5545, opened 2017-05-15
Fix back error on dependent pages

jbalboni:
Connected to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2479

To fix this, I'm sending users back to the beginning of the form when this happens. I think this is the best option because they shouldn't be on this page in the first place without entering other data. If they are, it's probably because they refreshed, in which case they need to start from the beginning anyway.

### DIFF
--- a/src/js/common/schemaform/FormPage.jsx
+++ b/src/js/common/schemaform/FormPage.jsx
@@ -75,7 +75,10 @@ class FormPage extends React.Component {
 
   goBack() {
     const { eligiblePageList, pageIndex } = this.getEligiblePages();
-    this.props.router.push(eligiblePageList[pageIndex - 1].path);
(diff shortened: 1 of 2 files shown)


cehsu (approved):
LGTM

cvalarida (approved):
Yeah, this looks good, but I'm wondering if we should make a ticket for @goldenmeanie to come up with something to show / tell the user why they're going back to the beginning of the form.

It should be pretty rare, but in these instances, it might be good to tell the user, "We're throwing you back to the beginning because you somehow got around our defenses!" rather than just throwing them back with no warning, y'know?

jbalboni:
@cvalarida I think if this keeps happening, that might be helpful. So far there's been one instance of this error that I can find.

PR #5546, opened 2017-05-15
SM notifications bugfixes

bshyong:
Because we are using a common store, constant names need to be unique; otherwise all reducers sharing the same constant name will update their respective stores.

Fetching preferences for SM was failing because the RX reducer shares the same constant name and was looking for a `flag` attribute which is nonexistent in the SM fetch preferences response.

Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2781


U-DON (approved):
LGTM. Thanks for catching this.

PR #5550, opened 2017-05-15
Adding GA events for setting email preferences

robertfairhead:
Closes https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2573

I am updating the GTM config to match these so will be ready when this PR ships.


U-DON (approved):
These are fine where they are, but for consistency with the other GA events, it might be better to have them in the `savePreferences` function in each app's `actions/preferences.js` file.

robertfairhead:
i put them where they were b/c the extra value for each (true/false or frequency) seemed clearer in those spots than in save preferences where it's just treated as a blob to be JSON'd.

I don't have strong feelings about it either. I'll merge as is so it's ready for launch, but I'm definitely open to moving it if you all have preferences since you'll interact with it more.

PR #5551, opened 2017-05-15
VA letters scaffolding

annekainicUSDS:
Closes https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2793

PR for the VA letters scaffolding. Added basic files, like entry, routes, and containers. Added path to ignoreList and entry file list.

Need to have a better sense of what the data store looks like before adding more detail here.

### DIFF
--- a/src/js/va-letters/va-letters-entry.jsx
+++ b/src/js/va-letters/va-letters-entry.jsx
@@ -1,0 +1,36 @@
+import 'core-js';
+import React from 'react';
+import ReactDOM from 'react-dom';
+import { createHistory } from 'history';
+import { IndexRedirect, Route, Router, useRouterHistory } from 'react-router';
+import { Provider } from 'react-redux';
+
+import VALettersApp from './containers/VALettersApp.jsx';
+import initReact from '../common/init-react';
+import { commonStore } from '../common/store';
+import routes from './routes.jsx';
+
+require('../common');  // Bring in the common javascript.
+require('../../sass/va-letters.scss');
+require('../login/login-entry.jsx');
+
+
+const history = useRouterHistory(createHistory)({
+  basename: '/va-letters'
+});
+
+function init() {
+  ReactDOM.render((
+    <Provider store={commonStore}>
+      <Router history={history}>
+        <Route path="/" component={VALettersApp}>
+          <IndexRedirect to="/confirm-address"/>
+          {routes}
+        </Route>
+      </Router>
+    </Provider>
+    ), document.getElementById('react-root'));
+}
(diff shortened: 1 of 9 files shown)


src/js/va-letters/va-letters-entry.jsx
```diff
@@ -0,0 +1,36 @@
...
+  basename: '/va-letters'
+});
+
+function init() {
+  ReactDOM.render((
+    <Provider store={commonStore}>
+      <Router history={history}>
+        <Route path="/" component={VALettersApp}>
+          <IndexRedirect to="/confirm-address"/>
+          {routes}
```
raquelromano-gov: Not sure why but it looks like the other components (`ConfirmAddress` and `DownloadLetters`) aren't getting rendered the `confirm-address` and `download-letters` endpoints  are loaded. I'm glad to look at it more this afternoon.

src/js/va-letters/va-letters-entry.jsx
```diff
@@ -0,0 +1,36 @@
...
+const history = useRouterHistory(createHistory)({
+  basename: '/va-letters'
+});
+
+function init() {
+  ReactDOM.render((
+    <Provider store={commonStore}>
+      <Router history={history}>
+        <Route path="/" component={VALettersApp}>
+          <IndexRedirect to="/confirm-address"/>
```
annekainicUSDS: Looks at the mockups, I don't think there is an intro page. I think you just go from your profile page to the first page where you confirm your address. I could be missing something though!

src/js/va-letters/va-letters-entry.jsx
annekainicUSDS: Hm that's odd, it was working for me locally. Maybe try again after running `npm run build` and restarting the `watch` task?

src/js/va-letters/va-letters-entry.jsx
jcmeloni-usds: I wouldn't stress too hard about this right now -- I haven't even approved a mockup/flow version yet.  Just fyi.

src/js/va-letters/va-letters-entry.jsx
raquelromano-gov: Oh, I was missing your latest commit. Looks good!

raquelromano-gov (approved):

annekainicUSDS:
I don't have the right permissions to resolve the Code Climate issue. Going to merge for now without it being resolved, even though merging a PR with a failed check gives me existential angst.

PR #5552, opened 2017-05-15
Clean up personId related tests

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1833

I didn't find any duplication, but I did find some missing tests and was able to fix a TODO in the personId tests.

### DIFF
--- a/test/edu-benefits/definitions/personId.unit.spec.jsx
+++ b/test/edu-benefits/definitions/personId.unit.spec.jsx
@@ -9,7 +9,7 @@ import * as personId from '../../../src/js/edu-benefits/definitions/personId';
 import commonDefinitions from 'vets-json-schema/dist/definitions.json';
 
 describe('Edu personId', () => {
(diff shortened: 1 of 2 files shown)


test/edu-benefits/definitions/personId.unit.spec.jsx
```diff
@@ -9,7 +9,7 @@ import * as personId from '../../../src/js/edu-benefits/definitions/personId';
 import commonDefinitions from 'vets-json-schema/dist/definitions.json';
 
 describe('Edu personId', () => {
-  const uiSchema = personId.uiSchema();
+  const uiSchema = personId.uiSchema('veteran', 'view:noSSN');
```
cehsu: It looks like the preference is to explicitly provide inputs for invocations of `personId.uiSchema` rather than to rely on default values. Should this be changed throughout?

cvalarida (approved):

PR #5553, opened 2017-05-15
Hca rjsf children income

cvalarida:
Closes https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2736

This PR creates a new custom array field which allows us to pick out which fields we want to show for each item in the array. It's used for the children income section of the annual income page.

Complete with validation (finally)!
image


src/js/common/schemaform/SelectiveArrayField.jsx
```diff
@@ -0,0 +1,203 @@
...
+const Element = Scroll.Element;
+const scroller = Scroll.scroller;
+
+
+class SelectiveArrayField extends React.Component {
+  // This fills in an empty item in the array if it has minItems set
+  // so that schema validation runs against the fields in the first item
+  // in the array. This shouldn't be necessary, but there's a fix in rjsf
+  // that has not been released yet
+  componentDidMount = () => {
```
jbalboni: I don't think we need anything in this method, since we don't want to show any fields if the array is empty.

src/js/common/schemaform/SelectiveArrayField.jsx
```diff
@@ -0,0 +1,203 @@
...
+  }
+
+  onItemChange = (indexToChange, value) => {
+    const newItems = _.set(indexToChange, value, this.props.formData || []);
+    this.props.onChange(newItems);
+  }
+
+  getItemSchema = (index, selectedFields) => {
+    const schema = this.props.schema;
+    const itemSchema = schema.items.length > index
```
jbalboni: I think we can skip the fallback to `additionalItems`, since that's for showing an empty row, which we don't need to do here.

src/js/common/schemaform/SelectiveArrayField.jsx
```diff
@@ -0,0 +1,203 @@
...
+    const properties = _.pick(selectedFields, itemSchema.properties);
+    const required = _.intersection(selectedFields, itemSchema.required);
+
+    return _.assign(itemSchema, {
+      properties,
+      required
+    });
+  }
+
+  scrollToTop = () => {
```
jbalboni: Both of these scroll methods can go away.

src/js/common/schemaform/SelectiveArrayField.jsx
```diff
@@ -0,0 +1,203 @@
...
+      : null;
+    const hasTitleOrDescription = (!!title && !hideTitle) || !!description;
+    const selectedFields = uiSchema['ui:selectedFields'] || Object.keys(schema.properties);
+
+    // if we have form data, use that, otherwise use an array with a single default object
+    const items = (formData && formData.length)
+      ? formData
+      : [getDefaultFormState(schema, undefined, registry.definitions)];
+
+    // uiSchema's ui:order complains loudly if there are extraineous properties
```
jbalboni: Wanna correct my typo? :)

src/js/common/schemaform/SelectiveArrayField.jsx
```diff
@@ -0,0 +1,203 @@
...
+
+    const title = uiSchema['ui:title'] || schema.title;
+    const hideTitle = !!_.get(['ui:options', 'hideTitle'], uiSchema);
+    const description = uiSchema['ui:description'];
+    const textDescription = typeof description === 'string' ? description : null;
+    const DescriptionField = typeof description === 'function'
+      ? uiSchema['ui:description']
+      : null;
+    const hasTitleOrDescription = (!!title && !hideTitle) || !!description;
+    const selectedFields = uiSchema['ui:selectedFields'] || Object.keys(schema.properties);
```
jbalboni: We talked about this a little, but I'm sure we need this at all.

src/js/common/schemaform/SelectiveArrayField.jsx
```diff
@@ -0,0 +1,203 @@
...
+        <div>
+          {items.map((item, index) => {
+            const itemSchema = this.getItemSchema(index, selectedFields);
+            const itemIdPrefix = `${idSchema.$id}_${index}`;
+            const itemIdSchema = toIdSchema(itemSchema, itemIdPrefix, definitions);
+
+            // Use ui:setItemTitle function to set a title for each item
+            // Note: this overrides the ui:title
+            // Also, we have to create a new object or we'll have trouble with the reassignment
+            let itemUiSchema = _.assign({}, uiSchema.items);
```
jbalboni: You could do this in an `updateSchema` function and set `schema.title` instead. I'd like to keep the modifications to uiSchema to a minimum (or none, ideally), since it's the sort of blueprint for all the changes that happen to the regular schema. All of this becomes even harder to follow when the blueprint is changing, too.

src/js/hca-rjsf/config/form.js
```diff
@@ -674,31 +675,54 @@ const formConfig = {
             'view:spouseIncome': {
               'ui:title': 'Spouse income',
               'ui:options': {
-                hideIf: (formData) => formData.maritalStatus !== 'married' // Something else too?
+                hideIf: (formData) => formData.maritalStatus !== 'Married' // Something else too?
```
jbalboni: Should we normalize case here?

src/js/common/schemaform/SelectiveArrayField.jsx
cvalarida: Yeah, I guess if we're adding new items, we'll have to do a lot more than just this...Okay.

src/js/common/schemaform/SelectiveArrayField.jsx
cvalarida: Pretty sure that typo was all me 😛

src/js/common/schemaform/SelectiveArrayField.jsx
cvalarida: I think the alternative is `_.pick()`ing them in the `formConfig`, but we'd also need to make sure we do an `_.intersection()` manually with the fields we want and the fields in `ui:order` we may already have. Rather than doing that in the form config, I did it in here...

...er...I was supposed to, anyhow. Woops. Fixed now!
```js
if (uiSchema.items && uiSchema.items['ui:order']) {
      uiSchema.items['ui:order'] = _.intersection(
        // Make sure ALL the selectedFields are in there
        _.concat(uiSchema.items['ui:order'], selectedFields),
        selectedFields
      );
    }
```

Anywho, I'm interested to see what you're thinking, but the only way I think we can get around this is by doing the work manually in each `formConfig`, which seems silly.

src/js/common/schemaform/SelectiveArrayField.jsx
cvalarida: Yeah, good point. I want to say I tried that but couldn't get it working........I'll give it another go since that may have been broken for other reasons.

src/js/hca-rjsf/config/form.js
cvalarida: Thought about it, but it should always be title case, no? It wouldn't add much, to do so, but...Yeah, I guess I'll just play it safe.

src/js/common/schemaform/SelectiveArrayField.jsx
jbalboni: I think the disconnect is that I'm thinking of the child definitions for the two pages as largely separate, but you built the child definition as one thing. So you're coming from the angle of needing to split the schema and uiSchema apart, but I have been assuming it's built as two separate definitions already. 

So there are two approaches: build two separate definitions, and make the SelectiveArrayField component essentially the normal ArrayField component but stripped of any way to modify the number of rows. The second approach is the direction you're going, which is having one overall child definition, and have SelectiveArrayField show particular fields based on uiSchema configuration info. Me pushing for removing some of what you have in SelectiveArrayField to select fields puts us somewhere in the middle of the two approaches, which is probably a bad place. We should just pick one or the other.

I have been assuming the first approach because I don't see much (or any) overlap in the uiSchemas between the two pages, and it seems straightforward to use `pick` and `omit` to pull the fields out of the `child` schema definition (similar to how we pass lists of fields to other common definition files) for each page. So building two separate definitions seemed clearer to me.

I think this pattern is probably limited to HCA, and the changes to `uiSchema` are probably only going to impact this particular array component. So ultimately, I think either approach is fine. I think you should pick whichever one seems cleaner to you.

jbalboni:
@cvalarida The review page isn't working for the child info page because of an issue with the income fields being in `ui:order`. Not sure if you've looked at that yet.

src/js/common/schemaform/SelectiveArrayField.jsx
cvalarida: Ahhh, the two separate definitions make sense. I like that. As long as they're both called `children` in the `formSchema`, they'll be combined in `formData`, yeah? I guess that was the ambiguous part for me when I was building this, but in retrospect it doesn't make any more sense to build one definition and modify it than it does to build two definitions.

It does sort of take the point out of having a `SelectiveArray`, though...I'll have to come up with a more appropriate name, now. 😛

Thanks for clearing that up!

jbalboni: Right, since we moved to a single object for formData, both pages will modify the same children array. So we don't have to worry about data merge issues later.

cvalarida:
Ah right. I forgot about the review field. Good catch!

cvalarida:
None of the children information shows up on review--neither in the child information page, nor in the annual income page. 😞

cvalarida:
This is pretty much done, but still suffers from https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2852

jbalboni:
@cvalarida That's happening because the array field holds the array data in its local state and doesn't update it when it changes outside of that component (because before you always one array component per actual array in the data).

The quickest fix is to switch between `this.state.items` and `this.props.arrayData` depending on if the items count is locked.

src/js/common/schemaform/review/ArrayField.jsx
```diff
@@ -150,19 +150,26 @@ class ArrayField extends React.Component {
       pageKey: fieldName
     };
 
+    // TODO: Make this better; it's super hacky for now.
```
jbalboni: I think the non-hacky way to fix this is to get rid of our special casing of array data. But that needs an ok from Alex (there's an issue somewhere).

src/js/common/schemaform/review/ArrayField.jsx
```diff
@@ -201,11 +208,10 @@ class ArrayField extends React.Component {
...
                   <SchemaForm
                       reviewMode
-                      pageData={item}
-                      schema={this.getItemSchema(index)}
+                      data={item}
+                      schema={itemSchema}
                       uiSchema={arrayPageConfig.uiSchema}
-                      title={pageTitle}
-                      hideTitle
+                      title={itemSchema.title || ''}
```
jbalboni: I think we'd want to default to `pageTitle` if there's no title in the item schema?

src/js/hca-rjsf/config/form.js
```diff
@@ -670,31 +672,50 @@ const formConfig = {
             'view:spouseIncome': {
               'ui:title': 'Spouse income',
               'ui:options': {
-                hideIf: (formData) => formData.maritalStatus !== 'married' // Something else too?
+                hideIf: (formData) => !formData.maritalStatus || (formData.maritalStatus.toLowerCase() !== 'married') // Something else too?
```
jbalboni: I think we should do this case normalization everywhere for `maritalStatus` or just skip it altogether.

src/js/hca-rjsf/config/form.js
```diff
@@ -670,31 +672,50 @@ const formConfig = {
...
+                'ui:required': (formData) => formData.maritalStatus === 'Married'
               }
             },
-            'view:childrenIncome': {
+            children: {
+              // 'ui:title': 'Child income',
+              'ui:field': 'BasicArrayField',
+              items: _.merge(childIncomeUiSchema, {
+                'ui:options': {
+                  updateSchema: (formData, schema, uiSchema, index) => {
```
jbalboni: Why put this in here and not in `childIncomeUiSchema`?

test/hca-rjsf/test-data.json
```diff
@@ -0,0 +1,81 @@
+{
```
jbalboni: Should we turn this into data for schema tests, like the other form? Don't have to do that in this PR, though.

jbalboni (approved):
LGTM. Made a few comments, but nothing major.

test/hca-rjsf/test-data.json
cvalarida: Yeah, I was thinking about it, but in order for this to fill data nested in `view:` containers, they have to also be nested in `view:`. The schema test data expects it to be flattened like it will be when we submit.

Also, this isn't all the fields like we'll want to fill in anyhow; it's required only.

src/js/hca-rjsf/config/form.js
cvalarida: Because I skipped it while refactoring....No reason. It should be in there. I'm on it!

src/js/common/schemaform/review/ArrayField.jsx
cvalarida: I did it this way because as-is, it added the title only because it was required, but always hid it. This title will be the one next to the edit button, so I don't know that it'll make much sense to have the page title on the item we're editing.

src/js/hca-rjsf/config/form.js
cvalarida: Yeah, that's a good point. I forgot we were looking for this in another place as well. I'll fix that up while I'm here.

PR #5554, opened 2017-05-15
Common schemaform errors cv

cvalarida:
Closes https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1599

**Note:** I still need to actually _use_ `isValidSchema()`, but once I figure out where it should go, that'll be quick.


src/js/common/schemaform/helpers.js
```diff
@@ -331,6 +331,50 @@ export const pureWithDeepEquals = shouldUpdate((props, nextProps) => {
...
+      Object.keys(schema.properties).forEach((propName) => {
+        isValidSchema(schema.properties[propName], errors, [...path, propName]);
+      });
+    }
+  }
+
+  // Blech...need a better message for this one.
+  if (schema.type === 'array') {
+    if (typeof schema.items !== 'object') {
+      errors.push(`Missing items in ${path.join('.')} (array) schema.`);
```
jbalboni: I think if it said "Missing items schema in ..." it'd be clear enough.

jbalboni (commented):
Looks good so far. I can take another look when you add in the call to isValidSchema.

jbalboni (approved):
LGTM. I think the only thing left is to add the check to where `updateSchema` is called?

cvalarida:
Oh, right. I thought I had that covered already, but I moved it out of `recalculateSchemaAndData()`. Will do!

cvalarida:
@jbalboni Can I get 👀 on this again? Should be golden, but I just want to make sure my new checks are good.

jbalboni (approved):

PR #5556, opened 2017-05-16
Reduce duplication for completeServicePeriods

cvalarida:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2458

This was the only duplication I could find. I think it was mostly resolved with some of the refactoring I've already done, but I did want to clean at least this much up.


jbalboni (approved):
LGTM. Not sure what's up with the failed hca e2e test.

PR #5559, opened 2017-05-16
Remove nth child rule on body text in callouts

melwoodard:
Closes https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2787

Removes a rule that makes some body text larger in blue callouts on static content pages. This rule isn't necessary now with the ability to manually apply the `.va-introtext` class where needed.

Before:

image

After:

image

FYI @DanielleSoCompany @bethpotts this will fix this formatting issue in the health care PR: https://vetsgov-pr-5436.herokuapp.com/healthcare/eligibility/

### DIFF
--- a/src/sass/base/_va.scss
+++ b/src/sass/base/_va.scss
@@ -511,12 +511,6 @@ h6 {
     font: inherit !important;
     margin-bottom: 0;
     padding-bottom: .5em;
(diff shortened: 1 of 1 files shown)


jbalboni (approved):
LGTM

PR #5563, opened 2017-05-17
Wrap VA letters content in RequiredLoginView

raquelromano-gov:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2807

Was having some troubles with my vets-api/ID.me setup yesterday so couldn't test this manually. Will try again today, but sending out for review in the meantime.

Will add a rudimentary unit test as well.

### DIFF
--- a/src/js/va-letters/containers/VALettersApp.jsx
+++ b/src/js/va-letters/containers/VALettersApp.jsx
@@ -1,4 +1,3 @@
-import PropTypes from 'prop-types';
 import React from 'react';
 import { connect } from 'react-redux';
 
--- a/src/js/va-letters/containers/VALettersApp.jsx
+++ b/src/js/va-letters/containers/VALettersApp.jsx
@@ -19,7 +19,7 @@ function AppContent({ children, isDataAvailable }) {
   if (unregistered) {
     view = (
       <h4>
-        To view and download your VA letters, you need something something something.
+        To view and download your VA letters, you need to verify your identity (or whatever).
       </h4>
     );
   } else {
@@ -39,9 +39,11 @@ class VALettersApp extends React.Component {
     const { children, location } = this.props;
 
     return (
+      // TODO: change the service name below from "user-profile" to
+      // "evss-letters" or "va-letters" once its defined in vets-api
       <RequiredLoginView
           authRequired={3}
-          serviceRequired={"evss-letters"}
+          serviceRequired={"user-profile"}
           userProfile={this.props.profile}
           loginUrl={this.props.loginUrl}
           verifyUrl={this.props.verifyUrl}>
--- a/src/js/va-letters/containers/VALettersApp.jsx
+++ b/src/js/va-letters/containers/VALettersApp.jsx
@@ -1,3 +1,4 @@
+import PropTypes from 'prop-types';
 import React from 'react';
 import { connect } from 'react-redux';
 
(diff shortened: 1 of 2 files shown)


src/js/va-letters/containers/VALettersApp.jsx
```diff
@@ -1,19 +1,60 @@
...
+        To view and download your VA letters, you need something something something.
+      </h4>
+    );
+  } else {
+    view = children;
+  }
+
+  return (
+    // TODO: add HTML and CSS when design has settled
+    <div className="row">
```
annekainicUSDS: My PR will eventually change some of this, but I'm trying not to use new Foundation class names so we don't add to the problem of stripping out Foundation :). You should be able to use the equivalent USWDS class here, `.usa-grid`, instead. Let me know if that doesn't work though!

src/js/va-letters/containers/VALettersApp.jsx
```diff
@@ -1,19 +1,60 @@
...
+}
+
 class VALettersApp extends React.Component {
   render() {
-    const { children } = this.props;
     return (
-      <div>{children}</div>
+      <RequiredLoginView
+          authRequired={3}
+          serviceRequired={"evss-letters"}
```
annekainicUSDS: I assume we want to change this to `va-letters`?

src/js/va-letters/containers/VALettersApp.jsx
```diff
@@ -1,19 +1,60 @@
...
+          userProfile={this.props.profile}
+          loginUrl={this.props.loginUrl}
+          verifyUrl={this.props.verifyUrl}>
+        <AppContent>
+          {this.props.children}
```
annekainicUSDS: I don't totally understand how this works. Given the above definition of `<AppContent>`, I would have thought `children` would have to be passed as a prop, not as nested child content, if that makes sense. So it would have looked like this:
```
<App Content children={this.props.children} />
```
Googling for information on this isn't yielding anything fruitful so far. Do you happen to know how this works?

annekainicUSDS (commented):
Couple of small comments/questions!

src/js/va-letters/containers/VALettersApp.jsx
raquelromano-gov: Thanks--done.

src/js/va-letters/containers/VALettersApp.jsx
raquelromano-gov: It can be anything--was just following suit with `evss-claims` but right now it's just a placeholder for whatever vets-api decides to name this service, to be defined in https://github.com/department-of-veterans-affairs/vets-api/blob/master/lib/backend_services.rb

I just switched it to pretend to be `user-profile` since we're not going to be writing any submission code for a while and right now just need to test that the auth stuff works as intended.

src/js/va-letters/containers/VALettersApp.jsx
raquelromano-gov: Yeah, I wasn't aware of how `this.props.children` can be used either but found this article helpful: https://facebook.github.io/react/docs/composition-vs-inheritance.html

This is the funky way that all of our LOA3-protected apps show or hide content depending on the user (e.g. see [`MessagingApp`](https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/js/messaging/containers/MessagingApp.jsx#L59) or [`DisabilityBenefitsApp`](https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/js/disability-benefits/containers/DisabilityBenefitsApp.jsx#L46). Then the logic in https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/js/common/components/RequiredLoginView.jsx#L66 adds a prop to `AppContent` (actually to any child React component of `RequiredLoginView`) indicating whether or not to render the content.

raquelromano-gov (commented):
This seems to do the right thing on my local instance. The review instance deployment process on Jenkins still needs some work to fully support the node upgrade, so I can't send over a demo. Hopefully we can start doing that soon though.

raquelromano-gov:
FYI, just rebased on top of your progress bar changes.

src/js/va-letters/containers/VALettersApp.jsx
annekainicUSDS: Ah gotcha, sounds good.

annekainicUSDS (approved):
The code looks good to me. The only question I had is when I pull this down locally, and I'm not logged in, I see a generic message telling me to log in, but nothing is specific to VA Letters (it comes from `<LoginPrompt>`). It's a little confusing, and I think it would be better to have more specific language letting users know why they have to sign in or sign up in order to use this app. What do you think? Also signing in locally isn't working for me, something about a MVI outage, so I'm not able to fully test this. This worked for you locally?

raquelromano-gov:
Yeah, I got it working locally. In your local vets API setup, do you have a `config/mvi_schema/mock_mvi_responses.yml` file? At some point I started needing this (copied it from `config/mvi_schema/mock_mvi_responses.yml.example` to get login working locally.

Re `<LoginPrompt>` having better information, yes we should definitely update that bulleted list of services that require logging in since currently it only lists messaging, rx refills, and disability claims but should also list profile since https://www.vets.gov/profile/ requires logging in. We can file a ticket to update that page once we're ready to launch. (https://www.vets.gov/healthcare/health-records/ will also need this)

PR #5564, opened 2017-05-17
Add empty 1990 rjsf scaffold

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2810

Sets up the folder for the 1990 rjsf work.

### DIFF
--- a/src/js/edu-benefits/routes.jsx
+++ b/src/js/edu-benefits/routes.jsx
@@ -87,6 +87,23 @@ export default function createRoutes(store) {
       },
     },
     {
+      path: '1990-rjsf',
+      indexRoute: { onEnter: (nextState, replace) => replace('/1990-rjsf/introduction') },
+      component: asyncLoader(() => {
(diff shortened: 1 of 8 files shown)


src/js/edu-benefits/routes.jsx
```diff
@@ -87,6 +87,23 @@ export default function createRoutes(store) {
       },
     },
     {
+      path: '1990-rjsf',
+      indexRoute: { onEnter: (nextState, replace) => replace('/1990-rjsf/introduction') },
+      component: asyncLoader(() => {
```
cvalarida: This doesn't look hidden from production...Am I missing something?

jbalboni: Eh, I guess I should do that. It doesn't matter as much here because we already have approval for the 1990 and there's no way for a user to get to this version of the form.

cvalarida (approved):
Awesome!

PR #5565, opened 2017-05-17
Health records tweaks

bshyong:
- Removes Health Records from ignorelist
- add additional validation and error message for start date

screen shot 2017-05-17 at 12 36 20 am

### DIFF
--- a/src/js/health-records/containers/Main.jsx
+++ b/src/js/health-records/containers/Main.jsx
@@ -206,7 +206,7 @@ export class Main extends React.Component {
         { label: 'Last year', value: '1yr' },
         {
           label: (
-            <div className={customDateOptionClass}>
+            <div aria-live="assertive" className={customDateOptionClass}>
               {
                 invalidFormatError && <p className="date-range-error">
                 {validationErrorMessages.invalidFormatError}</p>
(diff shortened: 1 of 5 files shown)


src/js/health-records/containers/Main.jsx
```diff
@@ -179,6 +182,9 @@ export class Main extends React.Component {
     const missingDateError =
     [this.state.startDateError, this.state.endDateError]
       .indexOf('missing') > -1;
+    const startDateError =
+    [this.state.startDateError, this.state.endDateError]
```
U-DON: These could also probably be simplified by creating the error array once. Also, any reason not to use `includes`?

U-DON (approved):
LGTM. Just had some nits.

src/js/health-records/containers/Main.jsx
bshyong: Will do. thanks!

bshyong: Used indexOf because I am sure that it is supported across browsers. Includes would fail in IE, but since I believe we have polyfills now that work I'll switch to it for readability

PR #5567, opened 2017-05-17
Add title, progress bar, and nav header

annekainicUSDS:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2806

This PR does a few things:
1. Adds a title, progress bar, and nav header
2. Extracts out some shared logic around the nav header that was also being used for HCA
3. Adds a basic test of the main `VALettersApp` component (should have added it in my scaffolding PR)

It doesn't look perfect, but I think how it should look will change based on which Marvel App version we end up going with.
screen shot 2017-05-17 at 10 25 39 am

### DIFF
--- a/src/js/common/components/NavHeader.jsx
+++ b/src/js/common/components/NavHeader.jsx
@@ -1,30 +1,22 @@
 import PropTypes from 'prop-types';
 import React from 'react';
 
+import { getCurrentFormStep, getCurrentPageName } from '../utils/helpers';
+
 export default class NavHeader extends React.Component {
   render() {
     const { chapters, path, className } = this.props;
     const total = chapters.length;
 
-    let step;
-    let name;
-    chapters
-      .forEach((chapter, index) => {
-        if (chapter.pages.some(page => page.path === path)) {
-          step = index + 1;
-          name = chapter.name;
-        }
-      });
-
-    return step
+    return getCurrentFormStep(chapters, path)
       ? <h4
           role="progressbar"
-          aria-valuenow={step}
+          aria-valuenow={getCurrentFormStep(chapters, path)}
           aria-valuemin="1"
-          aria-valuetext={`Step ${step} of ${total}: ${name}`}
+          aria-valuetext={`Step ${getCurrentFormStep(chapters, path)} of ${total}: ${getCurrentPageName(chapters, path)}`}
           aria-valuemax={total}
           className={`nav-header ${className}`}>
-        <span className="form-process-step current">{step}</span> of {total} {name}
+        <span className="form-process-step current">{getCurrentFormStep(chapters, path)}</span> of {total} {getCurrentPageName(chapters, path)}
       </h4>
       : null;
   }
--- a/src/js/common/components/NavHeader.jsx
+++ b/src/js/common/components/NavHeader.jsx
@@ -7,16 +7,18 @@ export default class NavHeader extends React.Component {
   render() {
     const { chapters, path, className } = this.props;
     const total = chapters.length;
+    const step = getCurrentFormStep(chapters, path);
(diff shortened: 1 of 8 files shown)


src/js/common/utils/helpers.js
```diff
@@ -66,6 +66,17 @@ export function getInactivePages(pages, data) {
   return pages.filter(page => !isActivePage(page, data));
 }
 
+export function getCurrentFormStep(chapters, location) {
```
annekainicUSDS: This code was being used in HCA to handle some NavHeader logic. I knew I would need it for VA Letters, so I extracted it into the common helpers, and then revised HCA to use this helper as well.

src/js/va-letters/routes.jsx
```diff
@@ -1,18 +1,29 @@
...
+      chapter={chapterNames.downloadLetters}
+      name="Download Letters"
       key="/download-letters"
       path="/download-letters"/>
 ];
 
 export default routes;
+
+// Chapters are groups of form pages that correspond to the steps in the navigation components
+export const chapters = groupPagesIntoChapters(routes.map(r => r.props));
```
annekainicUSDS: The fact that this is needed in routes wasn't clear to me initially, and I only realized it after simply including the components for `SegmentedProgressBar` and `NavHeader` weren't actually rendering. Would love to find a better way to handle this than needing to know this must be included in the routes.

src/js/common/utils/helpers.js
raquelromano-gov: It looks like this was originally copied from https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/js/common/components/NavHeader.jsx#L12, so it probably makes sense to import this from there as well.

raquelromano-gov (approved):

src/js/common/utils/helpers.js
annekainicUSDS: Ah yes! will do

src/js/common/components/NavHeader.jsx
```diff
@@ -1,30 +1,22 @@
...
-          name = chapter.name;
-        }
-      });
-
-    return step
+    return getCurrentFormStep(chapters, path)
       ? <h4
           role="progressbar"
-          aria-valuenow={step}
+          aria-valuenow={getCurrentFormStep(chapters, path)}
```
jbalboni: These get calls seem repetitive and it might be better to set them to variables at the top of the render function.

src/js/va-letters/containers/VALettersApp.jsx
```diff
@@ -1,11 +1,30 @@
...
+
 class VALettersApp extends React.Component {
   render() {
-    const { children } = this.props;
+    const { children, location } = this.props;
+
     return (
-      <div>{children}</div>
+      <div className="usa-grid">
+        <div className="usa-width-two-thirds medium-8 columns">
```
jbalboni: `medium-8` is a Foundation thing (I think), should we make an effort to use just the uswds grid in new code if possible?

jbalboni (approved):
LGTM. I made some nitpicky comments you can do what you want with. The refactoring of common code looks good.

src/js/common/components/NavHeader.jsx
annekainicUSDS: Good point, I will change that!

src/js/va-letters/containers/VALettersApp.jsx
annekainicUSDS: Yes, I'll see if I can remove that without it causing an issue. I copied this code from somewhere else, and I remember thinking it was a bit strange that we're using both USWDS and Foundation  classes here.

PR #5569, opened 2017-05-17
Add ability to specify expandUnder condition on schemaforms

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2046

This adds an `expandUnderCondition` option to fields using `expandUnder`, so you can specify specific values or a function that need to match/be true for fields to expand underneath. I also updated this to hide fields if they use `expandUnder` and the field specified is hidden.

### DIFF
--- a/docs/schemaform/form-config.md
+++ b/docs/schemaform/form-config.md
@@ -238,6 +238,11 @@ We've also been adding some additional uiSchema functionality not found in the r
     // the expandUnder field as the first question.
     expandUnder: '',
 
+    // If you need to match on a specific value, you can use this option to specify a value
+    // that the expandUnder field's data should equal. This can also be a function, which
+    // receives the expandUnder field's data as an argument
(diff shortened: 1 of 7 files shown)


docs/schemaform/form-config.md
```diff
@@ -238,6 +238,11 @@ We've also been adding some additional uiSchema functionality not found in the r
     // the expandUnder field as the first question.
     expandUnder: '',
 
+    // If you need to match on a specific value, you can use this option to specify a value
+    // that the expandUnder field's data should equal. This can also be a function, which
+    // receives the expandUnder field's data as an argument
+    expandUnderCondition: 'someValue',
+
```
cvalarida: Can we also add here:
```js
expandUnderCondition: (formData) => formData.field === 'value'
```
or something?

cvalarida (approved):

cvalarida (approved):

PR #5571, opened 2017-05-17
MHV account creation and terms to master

aub:


src/js/common/components/RequiredTermsAcceptanceView.jsx
```diff
@@ -0,0 +1,56 @@
...
+
+import { checkAcceptance } from '../../user-profile/actions';
+
+import AcceptTermsPrompt from './AcceptTermsPrompt';
+import LoadingIndicator from '../../common/components/LoadingIndicator';
+
+class RequiredTermsAcceptanceView extends React.Component {
+  componentWillMount() {
+    // I get errors and the app crashes if this line is called.
+    // this.props.checkAcceptance();
```
bshyong: @aub what is the error that you see? This line kicks off the check which should ping the API for the acceptance check. It definitely should not crash so if you have more context I can see what's going on. 

My app doesn't crash but I get a 404 for the check which I presume means that the user has not accepted the terms.

```
code: "404"
detail: "The record identified by mhvac could not be found"
```

aub: @bshyong I think the crash may be fixed. It still bails when getting the 404, and I think that's the appropriate response from the backend, so we'll need to handle that on the client. You're right that the 404 does indicate that the user hasn't accepted the terms.

bshyong:
Great. I pushed some changes to pull the content from the API, so you'll need to seed your local database for it to work. The title needs to be 'mhvac' (that tripped me up a little bit when the API was unable to find the terms and conditions that I saved in the database)

The larger content fields take HTML and render it dangerously in the page to preserve formatting.
________________________________

bshyong:
We are passing specs! Also, currently MHV terms and conditions is only turned on in front of RX, but can be turned on for Secure Messaging as well

aub:
@bshyong I added the check to SM and BB, so I think we're good there. Looks like we need to do some refactoring of JS in order to make code climate happy, but looks great otherwise. Would you mind taking a look at that?

bshyong:
Couple of things I will do here:
- Fix specs for Secure Messaging. The latest commit broke tests because the Terms components blocks access to the app
- Refactor some common code 

We will also need to plan for a demo/deployment strategy because this PR introduces a major change that will block users from the app.

bshyong (approved):
LGTM

Issue #5572, opened 2017-05-17
Array fields are broken on review page


jbalboni:
No data can be entered in array fields on review page.

PR #5573, opened 2017-05-17
Fix review page array fields on schemaforms

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets-website/issues/5572

We were passing in a bad prop name, so no data was ever showing up on the review page array field rows.


cvalarida (approved):
:man_facepalming:

PR #5576, opened 2017-05-17
Rx Track Packages

U-DON:
### Changes
* Order history table moved to its own tab with different columns.
    * Resolves department-of-veterans-affairs/vets.gov-team#2705.
    * Resolves department-of-veterans-affairs/vets.gov-team#2706.
    * Clicking "Track Package" links should lead to this tab of the prescription page. Resolves department-of-veterans-affairs/vets.gov-team#2708.
* Prescription details is its own tab. Resolves department-of-veterans-affairs/vets.gov-team#2707.

Note: Various name changes for components.
- Prescription.jsx -> PrescriptionCard.jsx
- Detail.jsx -> Prescription.jsx, Details.jsx, TrackPackage.jsx


U-DON:
@NatalieEMoore: I want to reiterate that "fill date" here is not actually fill date, but is using the `shippedDate` provided by the API. That's the only date that's provided for each item in the package tracking data, while `fillDate` is only given for the single prescription that is being viewed (CARBAMAZEPINE in the screenshot). We don't have `fillDate` for each package.

Given that, should we display the column as "Ship date" instead for accuracy?

bshyong (approved):
LGTM. Failing test I think is due to the change from `rx-prescription` to `rx-prescription-card` as a CSS classname, should be an easy fix to pass e2e specs

U-DON:
Updating/fixing E2E tests.

patrickvinograd (dismissed):
Just putting a hold on this so it doesn't inadvertently get merged to master.

PR #5578, opened 2017-05-18
Health records tweaks

bshyong:
Address PR feedback from https://github.com/department-of-veterans-affairs/vets-website/pull/5565

Also makes a change to help prevent health record popups from being blocked


U-DON (approved):
LGTM.

PR #5579, opened 2017-05-18
WIP: VA letters confirm address

annekainicUSDS:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2808
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2824

I'm wondering if buttons is the right UI element for asking the user if their address is correct, for 2 reasons: 
1. Since the answer to this question is linked to the navigation of the app, if you select Yes, go to the download letters page, and then hit back, it hasn't saved your response that your address is correct. Seems a little odd, since we're storing this value. 
2. I'm also not sure about the accessibility of this, if screen readers will pick up on the fact that the the buttons are tied to a question. I tried to think about this in the markup I picked, putting the question in a `<label>` element, but if the screenreader jumps to the buttons there will be no context for which one to select.

screen shot 2017-05-18 at 1 21 27 pm

screen shot 2017-05-18 at 1 21 40 pm


jbalboni:
@annekainicUSDS The common store part looks good.

src/js/va-letters/containers/ConfirmAddress.jsx
```diff
@@ -1,19 +1,79 @@
...
+    this.confirmAddress = this.confirmAddress.bind(this);
+  }
+
+  confirmAddress() {
+    this.props.onConfirmationOfAddress(true);
+    this.props.router.push('/download-letters');
+  }
+
   render() {
+    const buttons = this.props.letters.addressConfirmed !== false ?
```
raquelromano-gov: Would it be possible to use `src/js/common/components/form-elements/ErrorableRadioButtons.jsx` with the `yesNo` option labels here, and pass in a handler to its `onValueChange` prop that does the same as below? I haven't looked carefully at it, but am guessing it takes care of the accessibility issues you were concerned about.

src/js/va-letters/containers/ConfirmAddress.jsx
```diff
@@ -1,19 +1,79 @@
...
+        (<div className="usa-grid-full form-progress-buttons">
+          <div className="usa-width-one-fourth">
+            <ProgressButton
+                onButtonClick={this.confirmAddress}
+                buttonText="Yes"
+                buttonClass="usa-button-primary"/>
+          </div>
+          <div className="usa-width-one-fourth">
+            <ProgressButton
+                onButtonClick={() => this.props.onConfirmationOfAddress(false)}
```
raquelromano-gov: You could also bind this in the constructor so you don't have to use `()  => ...`

src/js/va-letters/containers/ConfirmAddress.jsx
```diff
@@ -1,19 +1,79 @@
...
+  return {
+    profile: userState.profile,
+    letters: lettersState.letters
+  };
+}
+
+function mapDispatchToProps(dispatch) {
   return {
-    profile: userState.profile
+    onConfirmationOfAddress: (value) => {
```
raquelromano-gov: How about `onAddressConfirmation()` instead? (gets rid of the 'Of')

src/js/va-letters/reducers/index.js
```diff
@@ -0,0 +1,22 @@
...
+  addressConfirmed: null
+};
+
+export default function userReducer(state = initialState, action) {
+  let newState = undefined;
+
+  switch (action.type) {
+    case UPDATE_ADDRESS_CONFIRMATION: {
+      newState = Object.assign({}, state);
+      _.set(newState, 'addressConfirmed', action.value);
```
raquelromano-gov: Since`lodash/fp` copies the state and returns a new one, I think you can just return this and get rid of `newState`

src/js/va-letters/containers/ConfirmAddress.jsx
annekainicUSDS: Yes I think so, it just means that these buttons will look like radio buttons instead :) so I think we'd need designers to weigh in on that.

src/js/va-letters/containers/ConfirmAddress.jsx
annekainicUSDS: Yep! Will change

src/js/va-letters/containers/ConfirmAddress.jsx
annekainicUSDS: I don't think I fully understand. The reason I had to do `() => ...` was because without that `this.props.onConfirmationOfAddress(false)` would be executed immediately instead of passing the function as a prop to `ProgressButton` to be executed `onClick`. I don't think I can bind a function that's part of `this.props`, since that is read only.

src/js/va-letters/reducers/index.js
annekainicUSDS: Yep, will change

src/js/va-letters/containers/ConfirmAddress.jsx
raquelromano-gov: Oh, never mind then...radio buttons don't sound right or this.

raquelromano-gov: Re: accessibility, you might try giving each `ProgressButton` an `aria-labelledby` attribute and give it the id of the element holding "Is this address correct?": https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-labelledby_attribute

Or just put in a TODO to that effect for now.

raquelromano-gov:
Re: what should happen if the user clicks yes or no and then hits the browser back button: I think this is a design question. Since we're storing the value, we can change what this page looks like depending on what the user has input during this session, but I'm pretty sure that isn't something being tested.

annekainicUSDS:
This PR probably isn't going to be relevant anymore because the design of the app is going to change pretty significantly. Going to close for now and we can reopen or replicate other work when needed.

PR #5580, opened 2017-05-18
Deprecate old selectDropdown e2e helper

raquelromano-gov:
Was doing some accessibility testing stuff and noticed this helper can be deprecated now.

### DIFF
--- a/test/util/nightwatch-commands/selectDropdown.js
+++ b/test/util/nightwatch-commands/selectDropdown.js
@@ -1,6 +1,8 @@
 /**
- * For chaining selecting dropdowns
- * Note: This takes the field name, not the whole selector.
+ * Change select value and trigger change event programatically. This
+ * is necessary because long select boxes tend to render offscreen,
+ * causing Selenium to fail in unexpected ways.
+ * The first parameter is the field name, not the whole selector.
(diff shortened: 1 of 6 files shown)


test/util/nightwatch-commands/selectDropdown.js
```diff
@@ -1,6 +1,8 @@
 /**
- * For chaining selecting dropdowns
- * Note: This takes the field name, not the whole selector.
+ * Change select value and trigger change event programatically. This
+ * is necessary because long select boxes tend to render offscreen,
+ * causing Selenium to fail in unexpected ways.
```
cvalarida: Ooh, good point! I was wondering what the deal was.

cvalarida (approved):
Thanks for doing this!

test/util/nightwatch-commands/selectDropdown.js
raquelromano-gov: I just swiped this comment from the old function, but yes, good to know.

cvalarida: Shows how much I pay attention! ¯\\_(ツ)_/¯

PR #5582, opened 2017-05-18
Remove special casing for content updates

robertfairhead:
This removes the special handling for content team updates so they are handled like all other branches.

Full test suite is run so that they are merge-able to master if they pass.

Related to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2695


jkassemi (approved):
👍

PR #5583, opened 2017-05-18
Refactor common store to not include other app's reducers and actions

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2785

I was going to make these changes so they could be done by each team when they have time, but the changes are pretty rote and I needed to update the login entry file for everyone, so I went ahead and made them.

The changes basically are:

- Turn `common/store` into a function that can be passed app specific reducers
- Turn `login/login-entry` into a function that accepts a store
- Update each app to use the two new functions and pass in their reducers
- Update login to use the passed in store and pass a dispatch function to `getUserData`

Outstanding questions:
- Did I break anything in login? Need to test that a little more
- Do rx/messaging/health-records need access to each other's reducers/state? The way this is right now, they'll only have their specific reducer. That's easy to change if we need to.


src/js/common/store/index.js
```diff
@@ -1,8 +1,23 @@
...
+    profile
+  })
+};
+
+export default function createCommonStore(appReducer = {}) {
+  const reducer = _.assign(appReducer, commonReducer);
+
+  return createStore(combineReducers(reducer), compose(
+    applyMiddleware(thunk),
+    __BUILDTYPE__ === 'development' && window.devToolsExtension
```
omgitsbillryan: The short-hand if-statement syntax looks a bit gnarly.  A named function might read better.

This is a _very_ nitpicky comment, feel free to disregard at your discretion.

omgitsbillryan (approved):
This looks like a win to me.  In relation to my [webpack bundling optimization work](https://github.com/department-of-veterans-affairs/vets-website/pull/5584), I think this will reduce the size of the `commons.js` bundle - since in theory `login` _should_ now be the only thing within it.

U-DON (approved):
Noticed some places that used `combineReducers` when there was only one object nested (pointed out one example) and wasn't sure if that was necessary. But LGTM, and tested that Rx/Messaging/Health Records/GIBCT loaded properly for me.

src/js/disability-benefits/reducers/index.js
```diff
@@ -7,12 +7,16 @@ import uploads from './uploads';
...
-  claims,
-  claimDetail,
-  claimAsk,
-  claimSync,
-  uploads,
-  routing,
-  notifications
-});
+export default {
+  disability: combineReducers({
```
jbalboni: I think so, but I can try it without to be sure.

jbalboni: Looks like it is necessary. The claim status app breaks if I remove it there.

annekainicUSDS (approved):
LGTM! Tested locally and login works great.

raquelromano-gov:
LGTM--thanks @jbalboni for the refactor

One thing I don't quite understand is when `combineReducers()` is/isn't needed.  I get that if there is only 1 reducer for an app, then the `combineReducer()` in `common` will take care of combining them. But it looks like there are cases in which it was there and you removed it, and cases in which it wasn't and you added it, and it doesn't seem to have anything to do with how many reducers are in that app's reducer. Would love to understand this better at some point.

jbalboni:
@raquelromano-gov My understanding is you need `combineReducers` wherever you're nesting part of the state. I removed it from the top level of all the app reducers, so that I can call it in the `common/store/index` after merging the app reducer object with the user info.

PR #5584, opened 2017-05-18
WiP : Optimize Webpack Bundling

omgitsbillryan:
**Issue**
https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2842

**Change List**
- Add `moment` to Vendor.js bundle
- Remove timestamps from bundle filenames
- Add a `commons` bundle

These are the bundles produced:
image


omgitsbillryan:
@jbalboni im a little stuck - can you tell me why this branch is not including the `login.scss` in the generated `no-react.css`?

image

jbalboni:
@omgitsbillryan You should check with @jkassemi and @aub about the timestamps. They added them because of a caching issue, so I don't know if it's safe to remove them.

jbalboni:
@omgitsbillryan I haven't built this branch to check, but I suspect that there's another css file being output now that commons is it's own bundle. The login code imports the login scss file and I believe Webpack merges that with other css in whatever bundle it's in, which before was each app's css.

I don't really see why the login scss shouldn't be imported in style.scss, since every app uses it. I think that would fix your issue (assuming my guess is correct).

wyattwalter:
This is quite old - do we still need this change?

jkassemi:
@omgitsbillryan what additional changes (aside from conflict resolution) would be needed to get this integrated?

jkassemi:
@omgitsbillryan, @wyattwalter ping re: level of effort required to get this change in place.

jbalboni:
I think we should close this PR and create/update the issue to focus on the timestamp in the bundle file name change/issue. That's the most important thing here and the other changes probably need to be re-evaluated based on how things have changed over the past year.

omgitsbillryan:
This effort is still valid but this PR is wildly out of date.  Each bundle is timestamped because something was broken inside bundler, such that, a change to a given bundle was not yielding a different fingerprint and thus the browser would not re-download.  Only until that issue is fixed can we try optimizing.

PR #5585, opened 2017-05-18
Schemaform review array bug fix

cvalarida:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2852

Per @jbalboni's [suggestion](https://github.com/department-of-veterans-affairs/vets-website/pull/5553#issuecomment-302427388), I switch between `this.props.arrayData` and `this.state.items` depending on whether or not the items count is locked.

Note: This does do a weird thing where when I add another item to the array, the `updateSchema` is only called once and the title below (if one is set, as it is in this case) doesn't get updated to what it should be. Turns out like this:
image

Only the first letter of the first name makes it, but after I edit, the title is right as rain. Not sure what the deal is there, exactly.

### DIFF
--- a/src/js/common/schemaform/SchemaForm.jsx
+++ b/src/js/common/schemaform/SchemaForm.jsx
@@ -50,6 +50,8 @@ class SchemaForm extends React.Component {
   componentWillReceiveProps(newProps) {
     if (newProps.name !== this.props.name) {
       this.setState(this.getEmptyState(newProps));
+    } else if (newProps.title !== this.props.title) {
+      this.setState({ formContext: _.set('pageTitle', newProps.title, this.state.formContext) });
     }
   }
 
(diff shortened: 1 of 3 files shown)


jbalboni:
@cvalarida I pushed a fix for the title issue. The way it was originally written the title wasn't expected to change, so some internal state wasn't being updated.

jbalboni:
One thing I've noticed is that the scrolling/focusing when you add or remove rows is messed up now, I think because there are two instances of this component pointing to the same array.

jbalboni (commented):
This looks good, but I'd like to see the the scrolling issues when editing rows fixed. If that turns into a big thing, we can make a separate item for it.

cvalarida:
Yeah, I noticed that too...I'll see what I can figure out about that.

cvalarida:
I went with the workaround you mentioned, @jbalboni. It's not meant to be a solid solution, but it'll get us out the door without a problem. It will have problems later if we don't use arrays like we are right now, though...

PR #5587, opened 2017-05-19
Fix state label collision in common address definition

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2869

There are overlapping state codes between US and Canada and we had switched to a single object to hold all labels, so Nevada turned into Nova Scotia.

screen shot 2017-05-18 at 11 44 09 pm

### DIFF
--- a/src/js/common/schemaform/definitions/address.js
+++ b/src/js/common/schemaform/definitions/address.js
@@ -4,22 +4,20 @@ import { createSelector } from 'reselect';
 import { countries, states } from '../../utils/options-for-select';
 import { validateAddress } from '../validation';
 
-/*
- * These are schema definitions for some common form fields
- */
 const countryValues = countries.map(object => object.value);
 const countryNames = countries.map(object => object.label);
 const militaryStates = states.USA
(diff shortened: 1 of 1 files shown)


src/js/common/schemaform/definitions/address.js
```diff
@@ -66,18 +64,23 @@ export function uiSchema(label = 'Address', useStreet3 = false) {
       const isRequired = addressSchema.required.length > 0;
 
       let stateList;
+      let labelList;
       if (country === 'USA') {
         stateList = usaStates;
+        labelList = usaLabels;
```
cvalarida: Idle curiosity: Why `if` `else if` `else if` instead of a switch?

cehsu (approved):
LGTM

cvalarida (approved):
Looks good! Glad to see Nevada dry, hot, and landlocked again. 😄

src/js/common/schemaform/definitions/address.js
jbalboni: I dislike switch statements in JS. They're easy to mess up, especially if you're not returning from each branch.

iantopper:
LGTM

PR #5588, opened 2017-05-19
Working review and submit page for hca rjsf

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2237

Review page is all working now, plus confirmation and intro pages. This also includes more tests and some fixes to page ordering and validation.

Line changes look bigger because of the new schema tests and the rearranging of the form config.

screen shot 2017-05-19 at 11 34 03 am
screen shot 2017-05-19 at 11 35 23 am

Also updated error message:
screen shot 2017-05-18 at 1 09 39 pm

### DIFF
--- a/src/js/common/schemaform/definitions/address.js
+++ b/src/js/common/schemaform/definitions/address.js
@@ -27,10 +27,11 @@ function isMilitaryCity(city = '') {
   return lowerCity === 'apo' || lowerCity === 'fpo' || lowerCity === 'dpo';
 }
 
+const requiredFields = ['street', 'city', 'country', 'postalCode'];
 export function schema(currentSchema, isRequired = false) {
   return {
     type: 'object',
-    required: isRequired ? ['street', 'city', 'country', 'postalCode'] : [],
+    required: isRequired ? requiredFields : [],
     properties: _.assign(currentSchema.definitions.address.properties, {
       country: {
         'default': 'USA',
@@ -50,7 +51,7 @@ export function schema(currentSchema, isRequired = false) {
   };
 }
 
-export function uiSchema(label = 'Address', useStreet3 = false) {
+export function uiSchema(label = 'Address', useStreet3 = false, isRequired = null) {
   let fieldOrder = ['country', 'street', 'street2', 'street3', 'city', 'state', 'postalCode'];
   if (!useStreet3) {
     fieldOrder = fieldOrder.filter(field => field !== 'street3');
@@ -61,9 +62,12 @@ export function uiSchema(label = 'Address', useStreet3 = false) {
     ({ formData, path }) => _.get(path.concat('city'), formData),
     _.get('addressSchema'),
     (currentCountry, city, addressSchema) => {
-      const schemaUpdate = { properties: addressSchema.properties };
+      const schemaUpdate = {
+        properties: addressSchema.properties,
+        required: addressSchema.required
+      };
       const country = currentCountry || addressSchema.properties.country.default;
-      const isRequired = addressSchema.required.length > 0;
+      const required = addressSchema.required.length > 0;
 
       let stateList;
       if (country === 'USA') {
@@ -80,7 +84,7 @@ export function uiSchema(label = 'Address', useStreet3 = false) {
           schemaUpdate.properties = _.set('state.enum', stateList, schemaUpdate.properties);
 
           // all the countries with state lists require the state field, so add that if necessary
-          if (isRequired && !addressSchema.required.some(field => field === 'state')) {
+          if (required && !addressSchema.required.some(field => field === 'state')) {
             schemaUpdate.required = addressSchema.required.concat('state');
           }
         }
@@ -88,7 +92,7 @@ export function uiSchema(label = 'Address', useStreet3 = false) {
       // so we need to update it
       } else if (addressSchema.properties.state.enum) {
         schemaUpdate.properties = _.unset('state.enum', schemaUpdate.properties);
-        if (isRequired) {
+        if (required) {
           schemaUpdate.required = addressSchema.required.filter(field => field !== 'state');
         }
       }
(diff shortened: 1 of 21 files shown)


src/js/common/schemaform/definitions/address.js
```diff
@@ -120,12 +124,22 @@ export function uiSchema(label = 'Address', useStreet3 = false) {
       validateAddress
     ],
     'ui:options': {
-      updateSchema: (formData, addressSchema, addressUiSchema, index, path) =>
-        addressChangeSelector({
+      updateSchema: (formData, addressSchema, addressUiSchema, index, path) => {
+        let currentSchema = addressSchema;
+        if (isRequired) {
```
jbalboni: I'm open to better ways of dealing with conditionally required address fields.

src/js/common/schemaform/definitions/address.js
```diff
@@ -48,7 +49,7 @@ export function schema(currentSchema, isRequired = false) {
   };
 }
 
-export function uiSchema(label = 'Address', useStreet3 = false) {
+export function uiSchema(label = 'Address', useStreet3 = false, isRequired = null) {
```
cvalarida: Looks like `isRequired` takes a function? Can we put that in a docstring up here?

src/js/common/schemaform/validation.js
```diff
@@ -151,9 +151,9 @@ export function uiSchemaValidate(errors, uiSchema, schema, formData, path = '')
       validations.forEach(validation => {
         const pathErrors = path ? _.get(path, errors) : errors;
         if (typeof validation === 'function') {
-          validation(pathErrors, currentData, formData, schema, uiSchema['ui:errorMessages']);
+          validation(pathErrors, currentData, formData, schema, uiSchema['ui:errorMessages'], currentIndex);
```
cvalarida: Was this broken before, or is this for a better error message, or...?

src/js/hca-rjsf/config/form.js
```diff
@@ -128,7 +132,9 @@ const formConfig = {
         veteranInformation: {
           path: 'veteran-information/personal-information',
           title: 'Veteran information',
-          initialData: {},
+          initialData: {
+            maritalStatus: 'Married'
```
cvalarida: Should this (and the other `initialData`s) still be here? Are you just setting defaults?

src/js/hca-rjsf/config/form.js
```diff
@@ -672,19 +606,20 @@ const formConfig = {
             'view:spouseIncome': {
               'ui:title': 'Spouse income',
               'ui:options': {
-                hideIf: (formData) => !formData.maritalStatus || (formData.maritalStatus.toLowerCase() !== 'married') // Something else too?
+                hideIf: (formData) => !formData.maritalStatus ||
+                  (formData.maritalStatus.toLowerCase() !== 'married' && formData.maritalStatus.toLowerCase() !== 'separated')
```
cvalarida: Excellent, thanks!

cvalarida (approved):
Just had a couple questions.

src/js/common/schemaform/validation.js
jbalboni: It's so we can conditionally validate data in an array. Previously we hadn't had a situation where some data in an array item needed to be validated in combination with other data in that item (this happens for child dependent date).

src/js/hca-rjsf/config/form.js
jbalboni: Need to remove that.

src/js/common/schemaform/definitions/address.js
cehsu: Perhaps use a ternary operator to reassign the requiredFields (would have to be declared using `let`) variable similar to on line 32?
`requiredFields = (required && currentSchema.required.length === 0) ? requiredFields : [];`
then just 
`currentSchema = _.set('required', requiredFields, currentSchema);`

cehsu: Besides pulling that logic out into a helper, it looks as good as can be.

jbalboni: I don't want to call `_.set()` when the output of `isRequired()` hasn't changed, as that means we'll always create a new copy of addressSchema, which will cause the addressSelector to run more often than it needs to and force the address on the page to re-render unnecessarily.

cehsu (approved):
LGTM, lots of exciting changes in here.

src/js/common/schemaform/definitions/address.js
cehsu: That makes sense, I've updated my comment.

src/js/common/schemaform/validation.js
cvalarida: Ahhh, very cool. 👍

src/js/hca-rjsf/config/form.js
```diff
@@ -193,7 +195,9 @@ const formConfig = {
         demographicInformation: {
           path: 'veteran-information/demographic-information',
           title: 'Veteran information',
-          initialData: {},
+          initialData: {
+            isSpanishHispanicLatino: false
```
cvalarida: Hiya.

cvalarida (approved):
Missed a spot 😉

Other than that, looks great!

src/js/common/schemaform/definitions/address.js
cehsu: Could this be a chance to reuse the `required` defined on line 76?

src/js/hca-rjsf/config/form.js
jbalboni: This one is actually intentional! For some reason, the backend schema requires this field. So, I'm defaulting it to false (which causes no visual changes, because it's a checkbox) to get around that based on Mark's suggestion.

cvalarida: Oh. How bizarre. 😕
Well alrighty! Glad to hear it!

PR #5589, opened 2017-05-19
Add detail page to exceptions for piano key inclusion

melwoodard:
Closes https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2763

Adding the detail-page template as an exception from including the piano keys.

After all content goes live with the new templates, the `general-related-links` include won't be in use anywhere and should be revisited as part of https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2157

### DIFF
--- a/content/layouts/page-breadcrumbs.html
+++ b/content/layouts/page-breadcrumbs.html
@@ -43,7 +43,7 @@
     {{ contents }}
   {% endcase %}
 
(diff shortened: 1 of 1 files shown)


bshyong (approved):
LGTM

bshyong:
Feel free to merge after tests pass

PR #5596, opened 2017-05-22
merge 508 and e2e test for facility locator

bshyong:
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2203


raquelromano-gov (commented):
This e2e test should probably also load the facility detail page and run an aXe check on it, e.g. https://www.vets.gov/facilities/facility/vba_372b

U-DON (commented):
Move looks good, but agree with @raquelromano-gov that the details page is missing from the test.

Also not sure why it does the `axeCheck` after entering text in the input. The check after the submit seems sufficient.

bshyong:
Agreed with feedback. Added mocks for the missing routes and the test for the detail page.

bshyong:
@U-DON Feel free to take another look and merge if OK

test/e2e/facility-helpers.js
```diff
@@ -0,0 +1,267 @@
...
+      id: 'vba_343z',
+      type: 'va_facilities',
+      attributes: {
+        unique_id: '343z',
+        name: 'VetSuccess on Campus at Los Angeles City College',
+        facility_type: 'va_benefits_facility',
+        classification: 'VETSUCCESS ON CAMPUS',
+        website: null,
+        lat: 34.08751109,
+        'long': -118.2917626,
```
U-DON: I don't think the key needs single quotes.

test/e2e/facility-helpers.js
```diff
@@ -0,0 +1,267 @@
...
+      id: 'vha_691GE',
+      type: 'va_facilities',
+      attributes: {
+        unique_id: '691GE',
+        name: 'Los Angeles VA Clinic',
+        facility_type: 'va_health_facility',
+        classification: 'Multi-Specialty CBOC',
+        website: null,
+        lat: 34.05171284,
+        'long': -118.23874096,
```
U-DON: Single quotes can probably be removed here too.

test/e2e/facility-helpers.js
```diff
@@ -0,0 +1,267 @@
...
+      id: 'vba_343o',
+      type: 'va_facilities',
+      attributes: {
+        unique_id: '343o',
+        name: 'Los Angeles Ambulatory Care Center',
+        facility_type: 'va_benefits_facility',
+        classification: 'OUTBASED',
+        website: null,
+        lat: 34.05171315,
+        'long': -118.2387403,
```
U-DON: Single quotes here also.

test/e2e/facility-helpers.js
```diff
@@ -0,0 +1,267 @@
...
+    id: 'vba_343z',
+    type: 'va_facilities',
+    attributes: {
+      unique_id: '343z',
+      name: 'VetSuccess on Campus at Los Angeles City College',
+      facility_type: 'va_benefits_facility',
+      classification: 'VETSUCCESS ON CAMPUS',
+      website: null,
+      lat: 34.08751109,
+      'long': -118.2917626,
```
U-DON: Single quotes here also.

test/e2e/facility-helpers.js
```diff
@@ -0,0 +1,267 @@
+const mock = require('./mock-helpers');
+
+/* eslint-disable camelcase */
```
U-DON: Would it help to eventually make requests with `X-Key-Inflection: camel`?

test/e2e/facility-helpers.js
```diff
@@ -0,0 +1,267 @@
...
+    pagination: {
+      current_page: 1,
+      per_page: 20,
+      total_pages: 1,
+      total_entries: 3
+    }
+  }
+};
+
+const facilityData = {
```
U-DON: This looks equivalent to `resultsData.data[0]`. Would it be more succinct to set it to a deep copy of that?

test/facility-locator/00-required.e2e.spec.js
```diff
@@ -18,11 +20,16 @@ module.exports = E2eHelpers.createE2eTest(
     .setValue('input[name="streetCityStateZip"]', 'Seattle, WA');
 
     client
-      .axeCheck('.main')
       .click('input[type="submit"]');
 
     client
       .axeCheck('.main');
```
U-DON: Probably could just remove the `client` in this block and just append the `axeCheck` to the chain above, right?

test/facility-locator/00-required.e2e.spec.js
```diff
@@ -18,11 +20,16 @@ module.exports = E2eHelpers.createE2eTest(
...
 
+    client
+      .waitForElementVisible('.facility-result', Timeouts.normal)
```
U-DON: Does it make sense to move this `waitForElementVisible` to be before the previous `axeCheck`?

test/e2e/facility-helpers.js
U-DON: I'm surprised the linter didn't trip over this. Thought it usually complained quoted single-word keys for me.

test/facility-locator/00-required.e2e.spec.js
U-DON: In other words, the sensible chain might be `click` -> `waitForElementVisible` -> `axeCheck`.

test/e2e/facility-helpers.js
bshyong: Yes, good catch. We should do that for consistency. I can roll that into the PR for the appointment stats

https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2859

test/e2e/facility-helpers.js
bshyong: @U-DON I added this because long is a reserved word according to the linter :man_shrugging:

U-DON: Oh, got it. That's strange; they were apparently only reserved in older versions of ECMAScript (1-3) too. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#Future_reserved_keywords

test/e2e/facility-helpers.js
```diff
@@ -0,0 +1,221 @@
...
+    path: '/v0/facilities/va',
+    verb: 'get',
+    value: resultsData,
+  });
+
+  mock(token, {
+    path: '/v0/facilities/va/vba_343z',
+    verb: 'get',
+    value: {
+      data: resultsData[0]
```
U-DON: Probably mean `resultsData.data[0]` here?

U-DON:
I think that the tests may be failing on the incorrect object path. Should be good to go once that's fixed, though.

PR #5598, opened 2017-05-22
Change postal code to zip code for USA addresses on HCA

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2904

The HCA backend expects `zipcode` instead of `postalCode` for USA addresses, so this adds some code to convert them before sending the form off to the backend.

### DIFF
--- a/src/js/common/schemaform/definitions/address.js
+++ b/src/js/common/schemaform/definitions/address.js
@@ -39,7 +39,8 @@ export function schema(currentSchema, isRequired = false) {
       },
       state: {
         title: 'State',
-        type: 'string'
+        type: 'string',
+        maxLength: 51
       },
       postalCode: {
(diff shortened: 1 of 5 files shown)


src/js/common/schemaform/definitions/address.js
```diff
@@ -39,7 +39,8 @@ export function schema(currentSchema, isRequired = false) {
       },
       state: {
         title: 'State',
-        type: 'string'
+        type: 'string',
+        maxLength: 51
```
jbalboni: I noticed this when looking at the address schema, so adding it just to be careful.

cvalarida (approved):
Kind of a pain, isn't it? Ah well, looks good!

PR #5600, opened 2017-05-22
Fixed broken GIBCT test

U-DON:


src/js/gi/reducers/index.js
```diff
@@ -9,7 +11,7 @@ import InstitutionDetailsFilterReducer from './filter';
 import ProfileReducer from './profile';
 import CalculatorReducer from './calculator';
 
-const rootReducer = {
+const rootReducer = combineReducers({
```
jbalboni: Does the app work with this? I think the common store call will fail.

U-DON: Oh, I see now that the common store is doing the `combineReducers` already. So it looks like I'll have to do this within the tests themselves.

U-DON: In which case, probably using `createCommonStore` again in the tests would avoid the verbosity of using `createStore` and `combineReducers` in each of them.

jbalboni: Yeah. I think that's what I did in the other tests

jbalboni (approved):

bshyong (approved):
LGTM

PR #5602, opened 2017-05-23
Add empty pensions app

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2879

Adds the empty pensions app, based on hca.

### DIFF
--- a/script/build.js
+++ b/script/build.js
@@ -125,6 +125,7 @@ const ignore = require('metalsmith-ignore');
 const ignoreList = [];
 if (options.buildtype === 'production') {
   ignoreList.push('healthcare/rjsf/*');
+  ignoreList.push('pensions/application.md');
   ignoreList.push('va-letters/*');
 }
 smith.use(ignore(ignoreList));
--- a/src/js/pensions/pensions-entry.jsx
+++ b/src/js/pensions/pensions-entry.jsx
@@ -1,0 +1,36 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import { createHistory } from 'history';
+import { Router, useRouterHistory } from 'react-router';
+import { Provider } from 'react-redux';
+
+import initReact from '../common/init-react';
+import route from './routes';
+import createCommonStore from '../common/store';
+import createLoginWidget from '../login/login-entry';
+import reducer from './reducer';
+
+require('../common');
+require('../../sass/pensions.scss');
+
+const store = createCommonStore(reducer);
+createLoginWidget(store);
+
+// Change the basename path once we replace hca with this form
+// (should be 'healthcare/appy/application')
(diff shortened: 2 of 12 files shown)


script/build.js
```diff
@@ -125,6 +125,7 @@ const ignore = require('metalsmith-ignore');
 const ignoreList = [];
 if (options.buildtype === 'production') {
   ignoreList.push('healthcare/rjsf/*');
+  ignoreList.push('pensions/application.md');
```
cvalarida: What does this ignore list do, exactly?

src/js/pensions/pensions-entry.jsx
```diff
@@ -0,0 +1,36 @@
...
+import { Router, useRouterHistory } from 'react-router';
+import { Provider } from 'react-redux';
+
+import initReact from '../common/init-react';
+import route from './routes';
+import createCommonStore from '../common/store';
+import createLoginWidget from '../login/login-entry';
+import reducer from './reducer';
+
+require('../common');
```
cvalarida: Ohh, interesting. `require()`ing to inject all the js from `../common/index.js`, aye? I was a bit befuddled why we were mixing `import` and `require()`, but I think I get it.

I imagine we'd only ever do this at the entry point, yes?

src/sass/pensions.scss
```diff
@@ -0,0 +1,9 @@
+// Variables
+
+@import "style";
+@import "modules/m-process-list";
```
cvalarida: Isn't this the subway map? Will we need this module?

cvalarida (approved):
Had some questions, but it looks good as a scaffold.

script/build.js
jbalboni: It keeps us from generating an html page at `pensions/application` in production.

src/js/pensions/pensions-entry.jsx
jbalboni: We should be able to use import here, I just copied it from other entry pages.

src/sass/pensions.scss
jbalboni: I am assuming so for the intro page.

cvalarida: Ahh, the intro page. Okay, cool.

src/js/pensions/pensions-entry.jsx
cvalarida: Ah, [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import) says we can `import '../common` "for side-effects only". Okay.

:man_shrugging:

script/build.js
cvalarida: Ahhh, that's why I don't see the edu forms in there. Awesome! Thanks for the clarification. 😄

PR #5603, opened 2017-05-23
Fix timeline styling on claim status

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2930

These styles were missed in the process list updates, I think because they're dynamically generated and hard to search for.

screen shot 2017-05-23 at 12 45 38 pm


cvalarida (approved):
Looks like another case for visual regression testing. 😕

PR #5607, opened 2017-05-23
Use flag to pick up nav breadcrumbs

raquelromano-gov:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3120

@phiden non-urgent question about navigation for VA letters (if it's not decided yet, no worries--we will shelve this for now):

The URL for now will be www.vets.gov/va-letters (you can see it on staging), so if that's fine, then should the nav breadcrumbs be
image
or should they reflect that this is discoverable from the profile page:
image

### DIFF
--- a/content/pages/va-letters/index.md
+++ b/content/pages/va-letters/index.md
@@ -2,19 +2,13 @@
 title: Download VA Letters
 entryname: va-letters
 layout: page-react.html
+gatePage: true
 in_maintenance: false
-maintenance_line1: We're sorry. The healthcare application is currently down while we fix a few things. We will be back up as soon as we can.
-maintenance_line2: In the meantime, you can still call 1-877-222-VETS(8387) and press 2 to complete this application over the phone.
+maintenance_line1: We're sorry. The page you are looking for is currently down while we fix a few things. We will be back up as soon as we can.
+maintenance_line2: In the meantime, please try the search box above or one of the options listed below to find more information.
 ---
 <div id="main">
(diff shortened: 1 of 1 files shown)


annekainicUSDS (approved):
LGTM! Good find.

annekainicUSDS:
Just tested this locally. The only thing that might be an issue is that before we had the breadcrumbs path as `Home > Profile > Download VA Letters`. Using the breadcrumbs include will now make the path `Home > Download VA Letters`, because I think it's passed on the url, and not the path a user takes to get there. This might not be a huge issue, but just something to note. Maybe also a good question for Laura (who doesn't seem to be in this repo?).

raquelromano-gov:
Yes, good question for Laura and in general for vets.gov design team who should know if breadcrumbs are supposed to reflect location in the site hierarchy.

phiden:
@raquelromano-gov I don't know that this decision has been made -- I suspect it's more of a business than design decision at this point. @rroueche do you have any insight?

rroueche:
Yeah good questions. A consistent hierarchy and breadcrumb treatment has been established for all static content, but I don't think it has been talked about for products/tools yet.

For content at least, breadcrumbs reflect location in site hierarchy. I don't know how we want to approach it for all tools, but for this one at least I don't think the Profile crumb should be in there. Just  ```Home > Download VA Letters``` .

raquelromano-gov:
Thanks @rroueche !

PR #5610, opened 2017-05-24
Scaffolding for Post-9/11 GI Bill status

raquelromano-gov:
Closes https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2865


config/webpack.config.js
```diff
@@ -23,12 +24,12 @@ const entryFiles = {
   'hca-rjsf': './src/js/hca-rjsf/hca-rjsf-entry.jsx',
   'health-records': './src/js/health-records/health-records-entry.jsx',
   messaging: './src/js/messaging/messaging-entry.jsx',
+  pensions: './src/js/pensions/pensions-entry.jsx',
+  'post-911-gibs': './src/js/post-911-gibs/post-911-gibs-entry.jsx',
```
annekainicUSDS: The only question I have about the name `post-911-gibs` is that we're hiding the word `status`, which is I think the most critical part of the name of this application. Just looking at it, I might not know what `gibs` stands for. But if it was called `post-911-gib-status`, I might guess in context that `gib` is GI Bill, and `status` tells me what this application is used for. Might be overthinking this, but to me it seems a bit clearer that way.

And thanks for alphabetizing these!

annekainicUSDS (approved):
LGTM! Only one question about the name and perhaps maybe a basic test that the main component renders? I think we have an example for that for VA letters.

config/webpack.config.js
raquelromano-gov: Yeah, good point--I like `post-911-gib-status` better. Will update tomorrow and merge.

PR #5613, opened 2017-05-24
HCA rjsf timestamp fix

jbalboni:
I can't test this locally, so I keep guessing at the invalid date issue. This should actually fix it, because I was able to successfully submit in staging and debug the component.


cvalarida (approved):
Full disclosure: I don't know how to test this, so I'm just taking your word that it works. I just don't want you to be blocked for lack of a review.

PR #5614, opened 2017-05-24
Make an automated last update for content that is opt-in

aub:
This is an offshoot of #5083, where we were trying to make all of the content pages have an automated last updated date based on their git history. It turns out that we do not always want this, so with this PR it's now opt-in, as follows:

To show a specific last updated data, add something like this to the front matter:

`lastupdate: 2016-12-25`

To show the last updated date generated from the git commit history, add this to the front matter:

`show_git_lastupdate: true`

@rroueche let me know if this is what you were looking for.


ayaleloehr:
pinging @rroueche on this.

rroueche:
thanks @melwoodard ! I think this branch is pretty far behind master at this point, would it be better to replicate the changes done here in the health-care-rr branch where I have the cleaned up static health care content? or to merge master in here, and then merge this one to #5876 ?

melwoodard:
@rroueche I already updated this branch with master, so it should be pretty clean to merge into the health care branch if you want to do it that way.

melwoodard:
These are the two commits to review in particular:

https://github.com/department-of-veterans-affairs/vets-website/pull/5614/commits/f8a636a5ce3d2326f922c77e78775f28062cb0af

https://github.com/department-of-veterans-affairs/vets-website/pull/5614/commits/93a06c882b077ce5ad858fb334333d4e39178083

bshyong (approved):
LGTM!

PR #5624, opened 2017-05-25
Add auth header to schemaform submissions

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2963

Old HCA sends the auth header so it can be used on the backend, but the rjsf forms don't send this by default. It shouldn't break anything to send it for everything.

### DIFF
--- a/src/js/common/schemaform/actions.js
+++ b/src/js/common/schemaform/actions.js
@@ -54,7 +54,8 @@ export function submitForm(formConfig, form) {
     window.dataLayer.push({
       event: `${formConfig.trackingPrefix}-submission`,
     });
-    return fetch(`${environment.API_URL}${formConfig.submitUrl}`, {
+
+    const fetchOptions = {
       method: 'POST',
       mode: 'cors',
       headers: {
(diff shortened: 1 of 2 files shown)


markolson (approved):

cehsu (approved):

src/js/common/schemaform/actions.js
```diff
@@ -54,15 +54,23 @@ export function submitForm(formConfig, form) {
...
       mode: 'cors',
       headers: {
         'Content-Type': 'application/json',
         'X-Key-Inflection': 'camel'
       },
       body
-    })
+    };
+
+    const userToken = sessionStorage.userToken;
```
cvalarida: Oh good; I just looked this up for the save in progress work but I wanted verification this was right. Awesome!

test/common/schemaform/actions.unit.spec.js
```diff
@@ -124,6 +124,36 @@ describe('Schemaform actions:', () => {
...
+      };
+      const form = {
+        pages: {
+          testing: {},
+        },
+        data: {
+          test: 1
+        }
+      };
+      global.sessionStorage = { userToken: 'testing' };
```
cvalarida: I'm assuming it has something to do with running as a unit test, but why `global` here?

cvalarida (approved):

test/common/schemaform/actions.unit.spec.js
jbalboni: It's Node's fallback global scope (vs window in browsers).

PR #5625, opened 2017-05-25
Fix search bug on GIBCT

bshyong:
resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2770

Search results will update when input is blurred


U-DON (approved):
LGTM.

PR #5626, opened 2017-05-25
Search from landing page when blurred or suggestion is selected.

U-DON:
Resolves department-of-veterans-affairs/vets.gov-team#2771.

Note that this also triggers search on blur if the term is not empty. Unknown if this is the behavior we want, but it will match the search page's behavior (though in that case, empty search term is allowed).


bshyong (approved):
LGTM we can revise behavior later if needed. Maybe we can disable the blur behavior for the landing page only

PR #5627, opened 2017-05-25
Add financial disclosure pages for dependents

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2911

Sorry for how giant this one is, it's a new pattern for us. Each child in the children array has three financial pages. So this adds an `showPagePerItem` pattern, than dynamically generates pages pages depending the items in an array.

This is sort of hard to test without actually having a page to add and remove items, so it might need some later tweaks.

screen shot 2017-05-25 at 2 06 31 pm
screen shot 2017-05-25 at 2 06 44 pm
screen shot 2017-05-25 at 2 06 56 pm
screen shot 2017-05-25 at 2 07 02 pm

screen shot 2017-05-25 at 2 07 28 pm
screen shot 2017-05-25 at 2 07 32 pm


src/js/pensions/config/form.js
```diff
@@ -196,16 +195,116 @@ const formConfig = {
...
-        //   }
-        // }
+        dependentsNetWorth: {
+          path: 'financial-disclosure/net-worth/dependents/:index',
+          title: item => `${item.childFullName.first} ${item.childFullName.last} net worth`,
+          pageType: 'array',
+          arrayPath: 'children',
+          itemFilter: (item) => !item.childNotInHousehold,
+          initialData: {
+            children: [
```
jbalboni: I'll comment this out before merging.

jbalboni:
Summary of what's happening:

When you go to a page that's set as an array page:
- Instead of passing through the schema/uiSchema to `SchemaForm`, the current item schema for the array at `arrayPath` is pulled out of the page state and passed instead.
- Similarly, the `items` uiSchema and the data for just that particular item are passed

When you make a change on an array item page:
- The data returned from `SchemaForm` is just for the one item, so we update the array in the overall data objet and call `setData` with that.

When you go to the next/previous page
- When the list of pages is pulled, after we filter out inactive pages, any array pages are expanded to create items for each array item. We update the `path` for each of those pages to replace `:index` with the current item index.
- The current page index is found by checking the url, instead of the page name (since there will be duplicates

When you're on the review page
- Similar changes to when you're on a regular page, but the editing state for each page is held in an array instead of a single boolean. If you end up changing the number of items, all pages go back to review mode, since we can't keep track of which item goes with which edit state.

src/js/common/schemaform/SchemaForm.jsx
```diff
@@ -145,7 +145,7 @@ class SchemaForm extends React.Component {
 
 SchemaForm.propTypes = {
   name: PropTypes.string.isRequired,
-  title: PropTypes.string.isRequired,
+  title: PropTypes.any.isRequired,
```
cvalarida: Isn't it more
```js
title: PropTypes.oneOf([
  PropTypes.string,
  PropTypes.func
]).isRequired
```
?

test/common/schemaform/ObjectField.unit.spec.jsx
```diff
@@ -150,6 +150,8 @@ describe('Schemaform: ObjectField', () => {
     expect(tree.everySubTree('TitleField')).is.not.empty;
   });
   it('should render component title', () => {
+    const onChange = sinon.spy();
+    const onBlur = sinon.spy();
```
cvalarida: Doesn't look like these are used?

cvalarida (commented):
Missing tests: `title` as a function, array page with `itemFilter`...seems like I forgot one other...blast.

I won't guarantee I looked at everything super thoroughly, but it looks good! Quite a beast, eh?

src/js/common/schemaform/FormPage.jsx
```diff
@@ -54,12 +55,16 @@ class FormPage extends React.Component {
   }
 
   onChange(formData) {
-    this.props.setData(formData);
+    let newData = formData;
+    if (this.props.route.pageConfig.showPagePerItem) {
+      newData = _.set([this.props.route.pageConfig.arrayPath, this.props.params.index], formData, this.props.form.data);
```
cehsu: Perhaps this logic could be pulled out into a helper and documented, it appears you're setting state by creating an array with 'children' and the param.index, creating a key based on the changed form data, and assigning the value to be the relevant config for the page, as described in `config/form.js`?

cehsu: It seems like you've explained most of what's going on here, in your summary comment from yesterday, and that a lot of what you've written could be repurposed to document this.

src/js/common/schemaform/SchemaForm.jsx
jbalboni: You mean I should do it the right way, that I was too lazy to look up? Yes. :-)

src/js/pensions/config/form.js
```diff
@@ -216,16 +215,116 @@ const formConfig = {
...
-        //     childrenInHousehold: [{}]
-        //   }
-        // }
+        dependentsNetWorth: {
+          path: 'financial-disclosure/net-worth/dependents/:index',
+          title: item => `${item.childFullName.first} ${item.childFullName.last} net worth`,
+          showPagePerItem: true,
+          arrayPath: 'children',
+          itemFilter: (item) => !item.childNotInHousehold,
+          initialData: {
```
jbalboni: Yeah. I think I'll clean that up in a later PR, since it's easier to have some of this in the form until we get some of the earlier pages built.

src/js/common/schemaform/FormPage.jsx
jbalboni: @cehsu Yeah, that's a good point, I should work that comment into comments into the actual code.

@cvalarida I'll add a comment, but the data that comes back here is just for a single row in the array, so we need to update the full form data object before saving it in the redux store.

src/js/common/schemaform/FormNav.jsx
```diff
@@ -5,6 +5,10 @@ import SegmentedProgressBar from '../components/SegmentedProgressBar';
 
 import { createFormPageList, createPageList } from './helpers';
 
+function routesMatch(pagePath, currentPath) {
+  return pagePath.replace(/\/:[^\/]+$/, '').endsWith(currentPath.replace(/\/\d+$/, ''));
```
jbalboni: Adding a comment, but yes, that's pretty much it.

jbalboni:
@cvalarida @cehsu Updates have been made, ready for another look.

cehsu (approved):
Thanks for the quick updates!

PR #5629, opened 2017-05-25
Include chapter 1 - applicant information in pensions config

cehsu:
Connects to department-of-veterans-affairs/vets.gov-team#2211

- Use definitions copied from education and relocated to common

TODO:

- [x] Test for applicantInformation section of pensions/config/form.js


src/js/common/utils/labels.jsx
```diff
@@ -0,0 +1,72 @@
+import React from 'react';
```
jbalboni: I don't think we should copy this whole file. Let's make a labels file with just the ones we need in the `common/schemaform` folder.

src/js/pensions/config/form.js
```diff
@@ -26,6 +27,12 @@ const formConfig = {
     }
   },
   chapters: {
+    applicantInformation: {
```
jbalboni: We should write some tests for this, too. I think there are existing tests for this page, we should follow that and adapt them to fields being shown here.

jbalboni (commented):
Looks good so far. Needs some tests and we can point the existing edu forms to the new common file.

cehsu:
@jbalboni changes made. Any additional thoughts tests? Right now the new tests are the same, as the fields are the same.

src/js/common/schemaform/definitions/applicantInformation.js
```diff
@@ -1,10 +1,10 @@
 import _ from 'lodash/fp';
 
-import currentOrPastDateUI from '../../common/schemaform/definitions/currentOrPastDate';
-import fullNameUI from '../../common/schemaform/definitions/fullName';
+import currentOrPastDateUI from './currentOrPastDate';
```
jbalboni: Let's make a `pages` folder, instead of putting this in `definitions`. We've been using that separation up until now and I think it's useful context.

jbalboni:
@cehsu Right, I meant that we should have a test that pulls in the pensions config file and tests those fields, like we have for each of the edu forms. I think we can refactor those tests eventually like we've done in other areas so there isn't so much duplication, but for now I want to make sure we have tests that actually use the pensions config, so that we prevent any bugs that result in us not pulling in the applicant information page correctly.

jbalboni (approved):

cehsu:
@jbalboni Have set up unit test (modeled after 1995 config unit test for the applicationInformation page) and have updated the form to specify the veteran status and required fields. Am currently getting:
```
Error: uiSchema order list contains extraneous property 'veteranDateOfBirth'
```
When including `veteranDateOfBirth` in required and fields to overwrite defaults provided in applicantInformation.

jbalboni:
@cehsu Looks like the backend schema (the `fullSchemaPensions` object in the configt) schema was missing that field. I updated the schema in here, so that error should go away, but you'll have to adjust the tests slightly now that the date of birth field is showing up.

J-Quag:
Looks good. :)

PR #5630, opened 2017-05-25
simplify logic and resolve bugs in FL

bshyong:
- Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2942

Searches with the same parameters will return the same results


U-DON (approved):
LGTM.

PR #5634, opened 2017-05-26
Use local version of leaflet.css

bshyong:
- Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1885
- Upgrades React Tabs to address PropType deprecation warnings

@patrickvinograd we are free to remove the leaflet entry from the CSP once this fix is merged to production
https://github.com/department-of-veterans-affairs/devops/pull/1615

### DIFF
--- a/content/includes/header.html
+++ b/content/includes/header.html
@@ -49,9 +49,6 @@
 <!-- CSS -->
 {% if entryname %}
 <link rel="stylesheet" href="/generated/{{ entryname }}.css">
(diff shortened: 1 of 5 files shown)


U-DON (approved):
LGTM. Though I wonder why we had to use the version from npmcdn to begin with.

bshyong:
Not sure; I think there were some issues with the way they were handling in the beta version

PR #5635, opened 2017-05-26
Fix bad variable override

jbalboni:
A recent change moved our variables above the uswds element/component imports, which meant that a variable we were overriding is no longer overridden. This adds a spot to override element/component variables and moves the one affected to that file.


cehsu (approved):
LGTM

annekainicUSDS (approved):
LGTM

PR #5639, opened 2017-05-26
Additional error handling for FL

bshyong:
- adds FL error handling
- adds improvements and messaging for bad search strings
- improves handling of searches when navigating with back button

Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2559

### DIFF
--- a/src/js/facility-locator/containers/VAMap.jsx
+++ b/src/js/facility-locator/containers/VAMap.jsx
@@ -345,7 +345,6 @@ class VAMap extends Component {
         <div className="row">
           <div className="columns usa-width-one-third medium-4 small-12" style={{ maxHeight: '75vh', overflowY: 'auto' }} id="searchResultsContainer">
             <div className="facility-search-results">
-              <p>Search Results near <strong>"{currentQuery.context}"</strong></p>
               <div>
                 <ResultsList facilities={facilities} pagination={pagination} currentQuery={currentQuery} updateUrlParams={this.updateUrlParams}/>
               </div>
--- a/src/js/facility-locator/containers/VAMap.jsx
+++ b/src/js/facility-locator/containers/VAMap.jsx
@@ -27,6 +27,10 @@ class VAMap extends Component {
     this.zoomOut = debounce(() => this.refs.map.leafletElement.zoomOut(0.5), 2500, {
       leading: true,
     });
+
+    this.listener = browserHistory.listen((location) => {
+      this.syncStateWithLocation(location);
+    });
   }
 
   componentDidMount() {
@@ -38,6 +42,9 @@ class VAMap extends Component {
     }
 
     if (location.query.address) {
+      this.props.updateSearchQuery({
+        searchString: location.query.address,
+      });
       this.props.searchWithAddress({
         searchString: location.query.address,
         context: location.query.context,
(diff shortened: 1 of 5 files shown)


src/js/facility-locator/containers/VAMap.jsx
```diff
@@ -89,7 +96,14 @@ class VAMap extends Component {
...
     const newQuery = this.props.currentQuery;
 
-    if (isEmpty(this.props.facilities) && !newQuery.inProgress && currentQuery.inProgress && newQuery.bounds && parseInt(newQuery.zoomLevel, 10) > 2) {
+    if (
+      isEmpty(this.props.facilities) &&
+      !newQuery.inProgress &&
+      currentQuery.inProgress &&
+      newQuery.bounds &&
+      parseInt(newQuery.zoomLevel, 10) > 2 &&
+      !newQuery.error
```
U-DON: Usually I like to assign long conditions to descriptively named variables like `shouldUpdateSearchQuery`.

src/js/facility-locator/containers/VAMap.jsx
```diff
@@ -109,6 +123,22 @@ class VAMap extends Component {
     }
   }
 
+  componentWillUnmount() {
+    this.listener();
```
U-DON: Why does this invoke the listener? Is that how it gets unbound?

src/js/facility-locator/reducers/searchQuery.js
```diff
@@ -27,35 +27,33 @@ export default function (state = INITIAL_STATE, action) {
...
         ...action.payload,
+        error: false,
         inProgress: true,
       };
-    case SEARCH_SUCCEEDED:
-      return {
-        ...state,
-        ...action.payload,
-        inProgress: false,
-      };
```
U-DON: Is this not needed anymore, at least to set `inProgress` to `false`?

U-DON (commented):
I think it LGTM, but lacking some context here. Aside from the additional alert error and clearing out the results on a search failure, I'm not sure about the other changes.

It looks like it's updating the state according to the URL, which should make navigating back a better experience. Could the callback for the listener on `browserHistory` be done within `componentDidUpdate` if you wrapped the component in `withRouter`?

src/js/facility-locator/containers/VAMap.jsx
bshyong: Yes I believe so; they use a const called `unlisten` in the docs https://github.com/ReactTraining/history/

src/js/facility-locator/containers/VAMap.jsx
bshyong: This is a good idea; will do

bshyong:
I explored doing things inside `componentDidUpdate` and `componentWillReceiveProps`, but came to the conclusion that it is cleaner to separate out the URL behavior. Because we have a controlled input, the props are often changing based on the user typing things into the search box. 

Having the callback only trigger on URL changes avoids having that logic mixed in with the other logic inside the component lifecycle.

src/js/facility-locator/reducers/searchQuery.js
bshyong: I realized this is unneeded because the `FETCH_VA_FACILITIES` and `FETCH_VA_FACILITIES` are the success cases. `SEARCH_SUCCEEDED` is unused. It is a little confusing because there is an intermediate step where the Mapbox API is hit, and needs to return properly before we can fetch the VA facilities

U-DON (approved):

PR #5644, opened 2017-05-30
Add general military history page to pensions

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2221

Adds general military history page:

screen shot 2017-05-30 at 2 40 40 pm
screen shot 2017-05-30 at 2 40 45 pm
screen shot 2017-05-30 at 2 41 00 pm

### DIFF
--- a/src/js/pensions/components/FullNameField.jsx
+++ b/src/js/pensions/components/FullNameField.jsx
@@ -1,9 +1,10 @@
 import React from 'react';
 
 export default function FullNameField({ formData }) {
+  const { first, middle, last, suffix } = formData;
   return (
     <div>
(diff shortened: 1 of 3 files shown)


cehsu (approved):
LGTM. One question is whether it would also be helpful to test whether the service date range ended after 9/11 and whether the from and to dates are in chronological order (or is that already being validated elsewhere?).

jbalboni:
@cehsu We have a test for the date range definition that should verify the dates are in the correct order.

I updated the test for the combat zone question to test the logic more specifically (that it doesn't show up with a date pre-9/11 and does with a date after).

PR #5648, opened 2017-05-30
Adds E2e tests for Kudos LOA3-protected apps

bshyong:
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2843

Added tests for
- health records
 - prescriptions
 - secure messaging
- disability benefits


U-DON (approved):
LGTM. Some of login stuff is repeated a lot, which may lend itself to being refactored into a common helper function. It would be easier to change things in one place if any wording changes on that screen in the future.

On a side note, is the number before the E2E tests indicative of anything other than the order in which they were created?

bshyong:
I think the naming is mostly for help with visual organization. I believe they are also run in alphabetical order.

Do you think it is worth it to refactor this entire login section into a helper since we are testing teh same behavior? The helper would have the test for unauthed and non-LOA3 users, but would need to be triggered within each e2e test with a one-line call.

U-DON:
Yeah, looks like it could be done in one line in each test.

bshyong:
Refactored—common test is in the login-helpers

U-DON (approved):
Presumably, when `disability-benefits` gets a full E2E test, it can replace the `00` test and include the unauthed flow.

PR #5650, opened 2017-05-30
Fix accessibility issues

bshyong:
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2927

also resolves sass lint warning

accessibility test from Axe extension is now passing on home page with no errors

### DIFF
--- a/content/includes/veteran-navigation.html
+++ b/content/includes/veteran-navigation.html
@@ -138,5 +138,3 @@
     </nav>
   </div>
 </div>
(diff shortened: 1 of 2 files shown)


U-DON (approved):
LGTM. I'm assuming that `usa-search.js` is not being used at all.

bshyong:
it's included twice; once in the header and once in veteran nav so I removed the extra one

PR #5651, opened 2017-05-31
Reference correct attribute for 'Last submit date' in list view and allow sorting by it in card view.

U-DON:
Resolves department-of-veterans-affairs/vets.gov-team#2861.

### DIFF
--- a/src/js/rx/components/SortMenu.jsx
+++ b/src/js/rx/components/SortMenu.jsx
@@ -1,16 +1,13 @@
 import PropTypes from 'prop-types';
-/* Component that lets a veteran sort by prescription name,
- * facility, or last submit date.
-*/
-
 import React from 'react';
 import _ from 'lodash';
 import classNames from 'classnames';
 
 const isDateAttribute = (option) => {
   return _.includes([
-    'refillSubmitDate',
+    'lastFillDate',
     'lastSubmitDate',
+    'refillSubmitDate',
     'refillDate'
   ], option);
 };
--- a/src/js/rx/config.js
+++ b/src/js/rx/config.js
@@ -36,6 +36,7 @@ module.exports = {
       }
     ]
   },
+
   errorCodes: {
     acceptTerms: [
       'RX135',
@@ -57,6 +58,7 @@ module.exports = {
       'RX99',
     ],
   },
+
   rxStatuses: {
     active: 'Active',
     deleted: 'Unknown',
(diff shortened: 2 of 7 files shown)


src/js/rx/config.js
```diff
@@ -71,20 +73,5 @@ module.exports = {
...
-    },
-    {
-      value: 'facilityName',
-      label: 'Facility name'
-    },
-    {
-      value: 'lastSubmitDate',
-      label: 'Last submit date'
-    }
-  ]
```
U-DON: Specific to the cards in active view. Moved it to `PrescriptionList`, the only place where it would be used.

src/js/rx/components/SortMenu.jsx
```diff
@@ -1,16 +1,13 @@
...
 import _ from 'lodash';
 import classNames from 'classnames';
 
 const isDateAttribute = (option) => {
   return _.includes([
-    'refillSubmitDate',
+    'lastFillDate',
     'lastSubmitDate',
+    'refillSubmitDate',
     'refillDate'
```
U-DON: Two are actually used (`lastFillDate`, `lastSubmitDate`), and the other two just in case the raw attributes of the prescriptions objects are exposed.

`lastFillDate` and `lastSubmitDate` represent the formatted attributes that are displayed to the user. They're named in accord with the field names we also display to the user.

src/js/rx/containers/Active.jsx
```diff
@@ -162,11 +162,6 @@ class Active extends React.Component {
         <div>
           <p className="rx-tab-explainer">Your active VA prescriptions</p>
           {this.renderViewSwitch()}
-          {this.state.view === 'list' || <SortMenu
-              onChange={this.handleSort}
-              onClick={this.handleSort}
-              options={sortOptions}
-              selected={currentSort}/>}
```
U-DON: Moved the sort menu to the `PrescriptionList` component since it's specifically used there.

src/sass/rx/partials/_rx-prescriptions.scss
```diff
@@ -41,7 +41,6 @@
 .rx-prescription-title {
   color: $color-primary;
   font-size: 1.4em;
-  line-height: 1;
```
U-DON: Minor side fix for prescription names in the cards getting cut off because their line-heights were too small.

bshyong (approved):
LGTM

PR #5653, opened 2017-05-31
Relocate message from FormPage to applicantInformation uiSchema

cehsu:
Connects to department-of-veterans-affairs/vets.gov-team#3024

Only show "You aren’t required to fill in all fields" message on veteran info pages

### DIFF
--- a/src/js/common/schemaform/pages/applicantInformation.js
+++ b/src/js/common/schemaform/pages/applicantInformation.js
@@ -52,6 +52,7 @@ export default function applicantInformation(schema, options) {
     initialData: {},
     uiSchema: _.assign({
       'ui:order': fields,
+      'ui:description': 'You aren’t required to fill in all fields, but VA can evaluate your claim faster if you provide more information.',
       [`${prefix}FullName`]: fullNameUI,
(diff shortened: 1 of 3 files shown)


src/js/common/schemaform/pages/applicantInformation.js
```diff
@@ -52,6 +52,7 @@ export default function applicantInformation(schema, options) {
     initialData: {},
     uiSchema: _.assign({
       'ui:order': fields,
+      'ui:description': 'You aren’t required to fill in all fields, but VA can evaluate your claim faster if you provide more information.',
```
jbalboni: If you change the file extension to .jsx, you can keep the bolding of "all":

```jsx
'ui:description': <p>You aren’t required to fill in <strong>all</strong> fields, but VA can evaluate your claim faster if you provide more information.</p>,
```

jbalboni (approved):
LGTM, one nitpick

src/js/common/schemaform/pages/applicantInformation.js
cehsu: Thanks @jbalboni, changes made.

PR #5656, opened 2017-05-31
Add command line to run e2e tests against wcag2a accessibility rules

raquelromano-gov:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2945

Allows developers to run
```
npm run test:wcag2a -- path/to/test.e2e.spec.js
```
to see where tests need updating to satisfy WCAG 2.0 standards

Also allows the rule set to be specified as part of the `axeCheck` config argument (which will override the global set by the above command, e.g.
```
.axeCheck('.main', { rules: ['section508', 'wcag2a'] })
```

### DIFF
--- a/config/nightwatch.js
+++ b/config/nightwatch.js
@@ -57,23 +57,11 @@ module.exports = {
       },
     },
     accessibility: {
-      filter: './test/accessibility/*.spec.js',
+      filter: './test/accessibility/*.spec.js'
+    },
+    wcag2a: {
       globals: {
-        asyncHookTimeout: 20000,
-      },
-      desiredCapabilities: {
-        browserName: 'chrome',
-        javascriptEnabled: true,
-        acceptSslCerts: true,
-        webStorageEnabled: true,
-        chromeOptions: {
-          binary: electron,
-          args: ['--window-size=1024,768']
-        }
-      },
-      test_workers: {
-        enabled: false,
-        workers: parseInt(process.env.CONCURRENCY || 1, 10)
+        rules: ['section508', 'wcag2a']
       }
     }
   }
--- a/test/util/nightwatch-commands/axeCheck.js
+++ b/test/util/nightwatch-commands/axeCheck.js
@@ -18,23 +18,17 @@ export function command(context, config, _callback) {
     document.head.appendChild(script);
   }, [axeSource]);
 
-  const runAxe = (innerContext, done) => {
+  // Run axe checks and report
+  this.executeAsync((innerContext, rules, done) => {
     axe.run(document.querySelector(innerContext) || document, { // eslint-disable-line no-undef
       runOnly: {
         type: 'tag',
-        values: (config || {}).rules || ['section508']
+        values: rules
       }
     }, (err, results) => {
       done({ err, results });
     });
-  };
-
-  // Run axe checks and report
-  this.executeAsync((innerContext, axeRunFn, done) => {
-    (axeRunFn(innerContext, done));
-  },
-  [context, runAxe],
-  response => {
+  }, [context, this.globals.rules || ['section508']], response => {
     if (response.state !== 'success') {
       this.verify.fail(`${response.state}: ${JSON.stringify(response)}`);
       return;
--- a/test/util/nightwatch-commands/axeCheck.js
+++ b/test/util/nightwatch-commands/axeCheck.js
@@ -18,18 +18,23 @@ export function command(context, config, _callback) {
     document.head.appendChild(script);
   }, [axeSource]);
 
-  // Run axe checks and report
-  const rules = (config || {}).rules || ['section508'];
-  this.executeAsync((innerContext, done) => {
+  const runAxe = (innerContext, done) => {
     axe.run(document.querySelector(innerContext) || document, { // eslint-disable-line no-undef
       runOnly: {
         type: 'tag',
-        values: rules
+        values: (config || {}).rules || ['section508']
       }
     }, (err, results) => {
       done({ err, results });
     });
-  }, [context], response => {
+  };
+
+  // Run axe checks and report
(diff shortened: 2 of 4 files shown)


config/nightwatch.js
```diff
@@ -57,23 +57,11 @@ module.exports = {
...
     },
     accessibility: {
-      filter: './test/accessibility/*.spec.js',
+      filter: './test/accessibility/*.spec.js'
+    },
+    wcag2a: {
       globals: {
-        asyncHookTimeout: 20000,
-      },
-      desiredCapabilities: {
```
raquelromano-gov: These are identical to default so should be inherited

config/nightwatch.js
```diff
@@ -57,23 +57,11 @@ module.exports = {
       },
     },
     accessibility: {
-      filter: './test/accessibility/*.spec.js',
+      filter: './test/accessibility/*.spec.js'
+    },
+    wcag2a: {
       globals: {
-        asyncHookTimeout: 20000,
```
raquelromano-gov: Doesn't appear to be used by any tests

cvalarida (approved):
Looks good. 👍
Thanks for the cleanup!

U-DON (approved):
Although I don't know much about how Nightwatch works, the changes look sensible.

Are we currently separating this new `wcag2a` env from the default env so it doesn't interfere with the builds yet and so we can resolve any failures introduced by the stricter checks? Also, should this rule be included with the `accessibility` tests (which currently only care about 508) eventually?

raquelromano-gov:
@U-DON Yes, the idea is exactly as you describe so we can easily find problems and fix (tracking that in https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2198). Once all tests pass with this command-line, we can update the default ruleset for our CI to be stricter.

As for the `accessibility` tests, the only ones that should remain there after https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2203 is resolved are the `sitemap-*.spec.js` tests, and yes those should also be run against these rules and fixed.

Using this command, we can easily file tickets for all of the tests that need updating and give devs a way to reproduce locally.

test/util/nightwatch-commands/axeCheck.js
```diff
@@ -19,16 +19,16 @@ export function command(context, config, _callback) {
   }, [axeSource]);
 
   // Run axe checks and report
-  this.executeAsync((innerContext, done) => {
+  this.executeAsync((innerContext, rules, done) => {
```
jbalboni: Does this let us pass in `rules` to the `axeCheck` function in the e2e tests? Haven't really written these commands much.

U-DON: I was under the impression that this is executing the function in the first arg with the rest of the args passed into the function. On that note, I'm seeing now that the array containing `context` now includes `this.globals.rules || ['section508']`. By my initial understanding of how `executeAsync` is working, should the latter be separated out into its own array? In other words, aren't those two separate arguments?

U-DON: It could also be that the one array represents all the arguments between [the function being invoked asynchronously] and [the `done` callback]. The [docs](https://github.com/nightwatchjs/nightwatch-docs/blob/master/api/method/executeAsync.md) aren't that clear about the additional arg.

jbalboni: Oh, gotcha. Yeah, was not reading this correctly.

It would be nice to be able to specify the ruleset somehow when calling `axeCheck`, so we could catch regressions in updated apps before everything is switched over.

raquelromano-gov: No, though I did try  that in an earlier version: added a `rules` field to the `config` paramater in the `axeCheck` command. I think that was a bit messier because you'd have to go into your tests and update the `axeCheck` calls, whereas with this command-line, you can just see what failures are there without even editing the test.

We could actually put both capabilities, if there are e2e tests with many failures and it would be helpful to only check one page at a time for example. From what I've seen, most e2e tests only have a few failures and they're pretty easy to track down from the details provided in the failure message.

raquelromano-gov: Sorry, the above response was only to @jbalboni 's first comment...now seeing the rest. Will read through again.

raquelromano-gov: @U-DON To answer your question about what `executeAsync()`, the second argument is an array of arguments to be passed to the function in the first argument: http://nightwatchjs.org/api/#executeAsync
So in this case `rules` can be either
1. `this.gobals.rules` which should have the value `['section508', 'wcag2a']` if defined in the nightwatch config
or
2. default to `['section508']`

Here are docs: http://nightwatchjs.org/api/#executeAsync

raquelromano-gov: @jbalboni Back to your question---I think you're saying you'd prefer to catch new regressions by changing individual `axeCheck` commands to be stricter, so that they will be merged into master and stay fixed. Then when we're ready to make _all_ the test strict, we just change the default rule set, and at our leisure update any `axeCheck` commands that unnecessarily specify the rules. 

Is that right? I think that's a fair point. Will go back and add that to the config parameter.

jbalboni: Yep. Though I think it kinda depends on how many actual errors there are. If there are only a few, that approach might be overkill.

raquelromano-gov: OK, updated so now you can use
```
.axeCheck('.main', { rules: ['section508', 'wcag2a'] })
```
or any other rules from https://github.com/dequelabs/axe-core/blob/develop/doc/API.md#options-parameter to achieve extra awesome accessibility

PR #5657, opened 2017-05-31
Fix info message on applicant info

jbalboni:
I wasn't clear on what I meant earlier. We can put jsx in the description fields if we need to by changing the file extension, importing react, and not using quotes around the field.


cehsu (approved):
Thanks for the quick fix and explanation!

PR #5661, opened 2017-05-31
Add reducer tests for Facility Locator

bshyong:
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2686

### DIFF
--- a/src/js/facility-locator/reducers/facilities.js
+++ b/src/js/facility-locator/reducers/facilities.js
@@ -1,5 +1,3 @@
-import { FETCH_VA_FACILITY, FETCH_VA_FACILITIES, SEARCH_FAILED } from '../actions';
-
 const INITIAL_STATE = {
   facilities: [],
   selectedFacility: null,
@@ -8,18 +6,18 @@ const INITIAL_STATE = {
 
 export default function (state = INITIAL_STATE, action) {
   switch (action.type) {
-    case FETCH_VA_FACILITY:
+    case 'FETCH_VA_FACILITY':
       return {
         ...state,
         selectedFacility: action.payload,
       };
-    case FETCH_VA_FACILITIES:
+    case 'FETCH_VA_FACILITIES':
       return {
         ...state,
         facilities: action.payload.data,
         pagination: action.payload.meta.pagination,
       };
-    case SEARCH_FAILED:
+    case 'SEARCH_FAILED':
       return INITIAL_STATE;
     default:
       return state;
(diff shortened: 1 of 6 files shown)


src/js/facility-locator/reducers/facilities.js
```diff
@@ -1,5 +1,3 @@
-import { FETCH_VA_FACILITY, FETCH_VA_FACILITIES, SEARCH_FAILED } from '../actions';
```
U-DON: Are these constants not going to be used from the `reducers/index` module anymore? If not, might wanna remove the lines where they're defined.

U-DON: In theory, they'd be useful to share between the usage here and the tests, but I don't anticipate much scaling for FL, so I'm not opposed to using hard-coded strings in two places.

test/facility-locator/reducers/searchQuery.unit.spec.js
```diff
@@ -0,0 +1,88 @@
...
+
+    expect(state.error).to.eql(false);
+    expect(state.inProgress).to.eql(true);
+  });
+
+  it('should handle fetching list of facilities', () => {
+    const state = searchQueryReducer({
+      ...INITIAL_STATE,
+      inProgress: true,
+      error: true,
```
U-DON: Note there's no preset `error` in `INITIAL_STATE`.

test/facility-locator/reducers/searchQuery.unit.spec.js
```diff
@@ -0,0 +1,88 @@
...
+      type: 'SEARCH_STARTED',
+    });
+
+    expect(state.error).to.eql(false);
+    expect(state.inProgress).to.eql(true);
+  });
+
+  it('should handle fetching list of facilities', () => {
+    const state = searchQueryReducer({
+      ...INITIAL_STATE,
```
U-DON: `INITIAL_STATE` is destructured here but not in the subsequent tests. Either add it in later tests or remove this one for consistency?

U-DON (approved):
LGTM with some nits.

src/js/facility-locator/reducers/facilities.js
bshyong: There are some strange errors from Babel happening, when I include these so I removed their usage. I spent a bunch of time debugging this to no avail; we can remove these to your point about scaling. I'll remove the usage in the `reducers/index`

PR #5663, opened 2017-06-01
Suppress wcag2a rules for aXe checks due to a aXe API bug

raquelromano-gov:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2198

An open bug in the aXe API (https://github.com/dequelabs/axe-core/issues/160 and, possibly related, https://github.com/dequelabs/axe-core/issues/214) keeps these 2 checks from passing WCAG 2.0 rules, so explicitly configure the check to only run `section508` rules so that when we enable `wacag2a` on all e2e tests, this will continue to pass. Right now it should be a no-op.

### DIFF
--- a/test/gibct/00-required.e2e.spec.js
+++ b/test/gibct/00-required.e2e.spec.js
@@ -15,7 +15,8 @@ module.exports = E2eHelpers.createE2eTest(
     client
       .waitForElementVisible('body', Timeouts.normal)
       .waitForElementVisible('.gi-app', Timeouts.slow)
-      .axeCheck('.main');
+      // do not run 'wcag2a' rules because of open aXe bug https://github.com/dequelabs/axe-core/issues/214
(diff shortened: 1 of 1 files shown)


dylanb:
@raquelromano-gov can you reference axe-core issue 160 instead of 214 so you can track the correct issue?

raquelromano-gov:
@dylanb sure, done--it wasn't clear to me which issue we were hitting (or both)

raquelromano-gov:
@jbalboni found another good use for the aXeCheck parameter

jbalboni (approved):
Change looks good. Is this going to affect just these two pages? Or others, too?

raquelromano-gov:
For now just these, since the search box on the GIBCT is the source of the  false positive. I have yet to run these checks on messaging and rx where there could be others.

PR #5664, opened 2017-06-01
Fixed broken SM headers in mobile.

U-DON:
Resolves department-of-veterans-affairs/vets.gov-team#2738.

@bshyong: I looked into the history of changes for the files, because the headers were originally `h2` and are styled by the `.messaging-content-header h2` rule. It looks like you had made the change due to findings at a 508 audit. Didn't seem like there was more info given about why we had to change from `h2` to `h3`.

After quickly searching through the page hierarchy, `h2` seemed like it would make sense, but I changed pages with the mobile header to use `h3` anyway in case changing back to `h2` would break 508.

> screen shot 2017-05-31 at 9 14 55 pm

> screen shot 2017-05-31 at 9 15 17 pm

> screen shot 2017-05-31 at 9 34 28 pm

I also made it so that the header in desktop is the same size that it was originally intended to be, even though it's now `h3`. On another note, it seems strange to me that `h2` should be smaller than `h3` according to the default style rules, but anyway...
> screen shot 2017-05-31 at 9 33 44 pm

### DIFF
--- a/src/js/messaging/containers/Settings.jsx
+++ b/src/js/messaging/containers/Settings.jsx
@@ -18,7 +18,7 @@ export class Settings extends React.Component {
               onClick={this.props.toggleFolderNav}>
             Menu
           </button>
-          <h2>Settings</h2>
+          <h3>Settings</h3>
         </div>
         <div id="messaging-settings">
           <ul className="va-tabs" role="tablist">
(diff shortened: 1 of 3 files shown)


bshyong (approved):
LGTM. Thanks for catching and fixing this

PR #5668, opened 2017-06-01
Minor form/login performance improvements

jbalboni:
I was testing out HCA to see if there were any performance issues and found a couple things.

The biggest one is that the start up code in the schema form reducer was taking 125ms on my desktop, and 650ms when I throttled it to a low-end device. Removing the `_.merge` call improved that time on a low end device by around half a second.

I also found that the login/profile components were re-rendering after any Redux action, even if it didn't affect them, because `pure` was set to false. That's not necessary unless you're doing something very strange in your components and we aren't as far as I can tell.

### DIFF
--- a/src/js/common/schemaform/reducers/index.js
+++ b/src/js/common/schemaform/reducers/index.js
@@ -55,16 +55,17 @@ export default function createSchemaFormReducer(formConfig) {
       schema = updateItemsSchema(schema);
       const data = getDefaultFormState(schema, page.initialData, schema.definitions);
 
-      return _.merge(state, {
-        pages: {
-          [page.pageKey]: {
-            uiSchema: page.uiSchema,
-            schema,
-            editMode: false
-          }
-        },
-        data
-      });
+      /* eslint-disable no-param-reassign */
+      state.pages[page.pageKey] = {
+        uiSchema: page.uiSchema,
+        schema,
+        editMode: false
+      };
+
+      state.data = _.merge(state.data, data);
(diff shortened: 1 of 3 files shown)


jbalboni:
@annekainicUSDS or @raquelromano-gov, can you give the login changes a quick look?

src/js/common/schemaform/reducers/index.js
```diff
@@ -66,20 +66,19 @@ export default function createSchemaFormReducer(formConfig) {
...
-            schema,
-            editMode: isArrayPage ? [] : false,
-            showPagePerItem: page.showPagePerItem,
-            arrayPath: page.arrayPath
-          }
-        },
-        data
-      });
+      /* eslint-disable no-param-reassign */
+      state.pages[page.pageKey] = {
```
cvalarida: Ah, yeah. This doesn't need to be immutable after all. Good call.

cvalarida (approved):

raquelromano-gov (commented):
BTW, there are a bunch of places throughtout vets-website where we set `pure: false`, along with `TODO`s to remove that once we move to Immutable.js...should we create a cleanup issue to go through all those and make sure we only keep it if we need to?

jbalboni:
@raquelromano-gov Yeah, I think that'd be good. We really shouldn't ever need to use it.

raquelromano-gov:
Filed https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3139

PR #5669, opened 2017-06-01
Pensions reserve and national guard

cehsu:
Connects to department-of-veterans-affairs/vets.gov-team#2222


src/js/pensions/config/form.js
```diff
@@ -133,9 +138,48 @@ const formConfig = {
...
+              'ui:title': 'What is the information for your unit?',
+              'ui:options': {
+                expandUnder: 'nationalGuardActivation',
+              },
+              name: {
+                'ui:title': 'Name of Reserve/NG unit'
+              },
+              address: address.uiSchema(),
+              phone: phoneUI(),
+              date: dateUI
```
jbalboni: `dateUI` is a function (you'll see a prop type error in the dev console for this page because of it).

src/js/pensions/config/form.js
```diff
@@ -133,9 +138,48 @@ const formConfig = {
...
+            }
+          },
+          schema: {
+            type: 'object',
+            required: ['nationalGuardActivation'],
+            properties: {
+              nationalGuardActivation,
+              nationalGuard: {
+                type: 'object',
+                required: ['name', 'address', 'phone', 'date'],
```
jbalboni: Like I mentioned in Slack, I think you'll want to use `ui:required` functions in the schema to mark these fields as required if `nationalGuardActivation` is true. For the address part, the address uiSchema should have a parameter for that.

src/js/pensions/config/form.js
```diff
@@ -133,9 +138,48 @@ const formConfig = {
               combatSince911
             }
           }
+        },
+        reserveAndNationalGuard: {
+          path: 'military/reserveAndNationalGuard',
```
jbalboni: We normally use kebab case instead of camel case. I think `military/reserve-national-guard` would work.

src/js/pensions/config/form.js
```diff
@@ -136,9 +141,46 @@ const formConfig = {
...
+              phone: phoneUI('Unit phone number'),
+              date: dateUI('Service Activation Date')
+            }
+          },
+          schema: {
+            type: 'object',
+            required: ['nationalGuardActivation'],
+            properties: {
+              nationalGuardActivation,
+              nationalGuard: {
```
jbalboni: I think this would work, too:

`nationalGuard: _.set('properties.address', address.schema(fullSchemaPensions), nationalGuard)`

It looks like you only need to change the address, so doing it this way keeps us more in sync with the backend schema.

jbalboni (approved):
LGTM, just one nitpicky comment

src/js/pensions/config/form.js
cehsu: Much tidier @jbalboni, thanks! Changes made.

PR #5674, opened 2017-06-02
Actions for stubbed-out fetch from vets-api va-letters endpoint

raquelromano-gov:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2864

A few notes:
  * obviously still not styled in any particular way--just feeding some fake data through.
  * still not adding tests, as I think it's not worth the setup given they will probably  need to be rewritten, but open to counter-arguments

image

### DIFF
--- a/src/js/va-letters/containers/DownloadLetters.jsx
+++ b/src/js/va-letters/containers/DownloadLetters.jsx
@@ -13,7 +13,7 @@ class DownloadLetters extends React.Component {
       );
     });
     return (
-      <span>
+      <div>
         <div className="letters-form-panel">
           <p>The address on file for you with VA Compensation and Pension is:</p>
           <span>{destination.fullName}</span><br/>
@@ -26,7 +26,7 @@ class DownloadLetters extends React.Component {
             {letterItems}
           </ul>
         </div>
-      </span>
+      </div>
     );
   }
 }
--- a/src/js/va-letters/containers/DownloadLetters.jsx
+++ b/src/js/va-letters/containers/DownloadLetters.jsx
@@ -1,18 +1,44 @@
 import React from 'react';
 import { connect } from 'react-redux';
 
+// TODO: style this properly once design is settled
 class DownloadLetters extends React.Component {
   render() {
+    const destination = this.props.destination || {};
+    const letterItems = (this.props.letters || []).map((letter) => {
+      return (
+        <li key={letter.letterType}>
+          <a href="#">{letter.letterName}</a>
+        </li>
+      );
+    });
     return (
-      <div>Placeholder for download letters</div>
+      <span>
+        <div className="letters-form-panel">
+          <p>The address on file for you with VA Compensation and Pension is:</p>
+          <span>{destination.fullName}</span><br/>
+          <span>{destination.addressLine1}, {destination.addressLine2} {destination.addressLine3}</span><br/>
+          <span>{destination.city}, {destination.state} {destination.zipCode}</span>
+        </div>
+        <div>
+          <p>The following letters are available for download:</p>
+          <ul>
+            {letterItems}
+          </ul>
+        </div>
+      </span>
     );
   }
 }
 
 function mapStateToProps(state) {
   const userState = state.user;
+  const letterState = state.letters;
   return {
-    profile: userState.profile
+    profile: userState.profile,
+    letters: letterState.letters,
+    destination: letterState.destination,
+    available: letterState.available
   };
 }
 
--- a/src/js/va-letters/reducers/index.js
+++ b/src/js/va-letters/reducers/index.js
@@ -1,1 +1,27 @@
-export default {};
+import set from 'lodash/fp/set';
+
+const initialState = {
+  letters: [],
+  destination: {},
+  available: false
+};
+
+function letters(state = initialState, action) {
+  switch (action.type) {
+    case 'GET_LETTERS_SUCCESS':
+      return {
+        ...state,
+        letters: action.data.letters,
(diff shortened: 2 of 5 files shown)


src/js/va-letters/containers/DownloadLetters.jsx
```diff
@@ -1,18 +1,44 @@
...
+    const letterItems = (this.props.letters || []).map((letter) => {
+      return (
+        <li key={letter.letterType}>
+          <a href="#">{letter.letterName}</a>
+        </li>
+      );
+    });
     return (
-      <div>Placeholder for download letters</div>
+      <span>
```
annekainicUSDS: Small nit but probably want to wrap this in a `<div>` instead of a `<span>`.

src/js/va-letters/reducers/index.js
```diff
@@ -1 +1,27 @@
...
+  letters: [],
+  destination: {},
+  available: false
+};
+
+function letters(state = initialState, action) {
+  switch (action.type) {
+    case 'GET_LETTERS_SUCCESS':
+      return {
+        ...state,
```
annekainicUSDS: I'm not familiar with this syntax or how this works. Why just `return` these values instead of setting them within `state` as you do below with the `available` property? Are the values for `letters` and `destination` being set in `state` as well here?

src/js/va-letters/utils/helpers.js
```diff
@@ -1,4 +1,40 @@
+import merge from 'lodash/fp/merge';
+
+import environment from '../../common/helpers/environment';
+
 export const chapterNames = {
   confirmAddress: 'Confirm Address',
   downloadLetters: 'Download Letters'
 };
+
+function isJson(response) {
```
annekainicUSDS: Is this all copied from somewhere else? I noticed similar code in the rx app. If so, perhaps it's something we can extract out and share between apps?

annekainicUSDS (commented):
Looks great! Just a few questions/comments.

src/js/va-letters/containers/DownloadLetters.jsx
raquelromano-gov: Done, though presumably this will disappear.

src/js/va-letters/containers/DownloadLetters.jsx
```diff
@@ -1,18 +1,44 @@
...
 
 function mapStateToProps(state) {
   const userState = state.user;
+  const letterState = state.letters;
   return {
-    profile: userState.profile
+    profile: userState.profile,
+    letters: letterState.letters,
+    destination: letterState.destination,
+    available: letterState.available
```
raquelromano-gov: My understanding of what "available" means is here: https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2671
but according to that logic, this should be part of the letter-type response (i.e., it's a property of each individual letter, not a property of the letter list.

In that case, I guess a value of `false` would tell us we need to put up a notice that no letters are available due to a system being down or whatever

Tangentially, we need to sort out how the EVSS API provides information about availability and eligibility when it comes to individual letters, i.e.,
- if a user is eligible for a letter, but a backend service is down, so we should show the letter and gray it out with some explanation of what's going on
- if a user is not eligible for a letter, we won't show it in the UI

I don't like that they are confounding the concepts of availability and eligibility...will make sure we discuss this with them either via slack or at Thursday's meeting.

src/js/va-letters/reducers/index.js
raquelromano-gov: It's object spread syntax: http://redux.js.org/docs/recipes/UsingObjectSpreadOperator.html
It's just a little more compact than 2 `set` operations for the 2 fields that are being updated here (as compared to a single one below for `GET_LETTERS_FAILURE`

src/js/va-letters/utils/helpers.js
raquelromano-gov: Yes, it's copied in a [bunch of places](https://github.com/department-of-veterans-affairs/vets-website/search?p=2&q=apiRequest&type=&utf8=%E2%9C%93): health-records, messaging, rx, and I agree it might make sense to share instead of duplicating the code (though the function signature does vary slightly across those apps).  Not sure this PR is the place to do that though...we can ask around on #vg-frontend or at the next FE architecture syn.

src/js/va-letters/containers/DownloadLetters.jsx
annekainicUSDS: Yes, great point. For now, I think this is good, though.

src/js/va-letters/reducers/index.js
annekainicUSDS: Ah thanks! Perfect.

src/js/va-letters/utils/helpers.js
annekainicUSDS: Make sense, I can add that to the agenda for the next FE sync.

annekainicUSDS (approved):
LGTM!

src/js/va-letters/utils/helpers.js
raquelromano-gov: 👍 especially since I just added 4 code climate issues 😢

src/js/va-letters/containers/DownloadLetters.jsx
raquelromano-gov: Oh, I just realized the `available` here is unrelated to the EVSS values...I think I just added this so that if the data fetch failed, we have some indication of it and can update the UI appropriately.

PR #5675, opened 2017-06-02
Improve review page performance for hca

jbalboni:
The review page is a little slow on old phones, so I made some updates to address that. I also tweaked the layout of fields on the review page for mobile, since that was pretty rough (and fixing it slipped through the cracks before).

screen shot 2017-06-02 at 11 15 57 am

### DIFF
--- a/script/build.js
+++ b/script/build.js
@@ -59,6 +59,7 @@ const optionDefinitions = [
   { name: 'watch', type: Boolean, defaultValue: false },
   { name: 'entry', type: String, defaultValue: null },
   { name: 'host', type: String, defaultValue: 'localhost' },
+  { name: 'public', type: String, defaultValue: null },
 
   // Catch-all for bad arguments.
   { name: 'unexpected', type: String, multile: true, defaultOption: true },
@@ -219,6 +220,7 @@ if (options.watch) {
     port: options.port,
     publicPath: '/generated/',
     host: options.host,
+    'public': options.public,
     stats: {
       colors: true,
       assets: false,
--- a/src/js/common/schemaform/SchemaForm.jsx
+++ b/src/js/common/schemaform/SchemaForm.jsx
@@ -2,6 +2,7 @@ import PropTypes from 'prop-types';
 import React from 'react';
 import _ from 'lodash/fp';
 import Form from 'react-jsonschema-form';
+import { deepEquals } from 'react-jsonschema-form/lib/utils';
 
 import { uiSchemaValidate, transformErrors } from './validation';
 import FieldTemplate from './FieldTemplate';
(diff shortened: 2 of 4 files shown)


script/build.js
```diff
@@ -59,6 +59,7 @@ const optionDefinitions = [
   { name: 'watch', type: Boolean, defaultValue: false },
   { name: 'entry', type: String, defaultValue: null },
   { name: 'host', type: String, defaultValue: 'localhost' },
+  { name: 'public', type: String, defaultValue: null },
```
jbalboni: The newer version of Webpack dev server forces you to use this option now if you want to expose your local server to other devices on your network.

src/js/common/schemaform/SchemaForm.jsx
```diff
@@ -57,6 +58,25 @@ class SchemaForm extends React.Component {
     }
   }
 
+  /*
+   * If we're in review mode, we can short circuit updating
```
jbalboni: The review page is slow mainly if you open several chapters and then edit a page, because it's re-rendering all of the visible pages on every character you type. This cuts down on that.

src/js/common/schemaform/formState.js
```diff
@@ -6,6 +6,12 @@ function isHiddenField(schema) {
   return !!schema['ui:collapsed'] || !!schema['ui:hidden'];
 }
 
+function get(path, data) {
```
jbalboni: It turns out that lodash-fp's `get` and `findLast` are a lot slower on low end devices than I expected. Removing them here shaves ~75-100ms off the updates that happen on every data change.

It's likely there are still more improvements related to that, but this change gets a good enough improvement for now without the risk of changing lots of instances.

cehsu (approved):
LGTM. Had some minor questions.

src/js/common/schemaform/formState.js
cehsu: Is path[path.length - 1] not an option for replacing findLast here?

src/sass/modules/_m-schemaform.scss
```diff
@@ -45,22 +45,29 @@ dl.review {
...
       margin-top: 0;
       margin-bottom: 0;
     }
+    @media (min-width: $small-screen) {
+      flex-direction: row;
+      &> dd {
+        padding-left: 5px;
+        min-width: 100px;
+      }
+      &> dt {
```
cehsu: Are these `&`'s necessary?

src/js/common/schemaform/formState.js
jbalboni: It's not necessarily the last item in the path, just the last number item. It would be better theoretically to search from back to front (which is what `findLast` was doing), but since paths are pretty short (almost always single digit length), it didn't seem worth breaking out a for loop for.

src/sass/modules/_m-schemaform.scss
jbalboni: Probably not. I'll remove them.

src/js/common/schemaform/formState.js
cehsu: Thanks for the insight @jbalboni, unless `path` is available as a string somehow in the invocation scope for this function (in which case regex could provide a slighter faster option), this is the clearest, fastest approach.

src/sass/modules/_m-schemaform.scss
cehsu: It looks you are following an existing convention. Perhaps simply creating an issue for this is good enough here.

cvalarida (approved):
I _think_ the next low-hanging fruit would be to remove the arrow functions from `render()` ([source](https://flexport.engineering/optimizing-react-rendering-part-1-9634469dca02)). Looks good!
I don't suppose you can give before and after render time statistics, can you?

jbalboni:
@cvalarida We might save some time because of the overhead of creating functions in each render, but I've been using a `shouldComponentUpdate` check from the rjsf library that skips any functions when comparing old and new props. So it's not causing us to do unnecessary renders.

jbalboni:
@cvalarida I didn't measure before and after times exactly for the `shouldComponentUpdate` change on the review page, but the wasted time in the React perf tools went down by about 80-100ms in total, I think?

PR #5677, opened 2017-06-02
Save in progress forms

cvalarida:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2910

It's not quite finished, but I should have made a PR before this, so...it's high time.

## What this entails
- New login modal (@raquelromano-gov Please review!)
- Saving form data
- Loading form data

## Workflow
First, if we're not logged in, we'll have an info box on the introduction page telling us that we can save forms while they're in progress, but we'll have to log in first. While filling out the forms (still not logged in), we'll have a link at the bottom of each page directing us to sign in before we can save the application.

Once logged in, the introduction page (`FormIntroButtons` for lack of a better name) will allow us to continue the form we've already saved (if we've got one) or start over from scratch. If we continue, it'll load the data into the redux store at `form.data` and redirect us to the last page we were on.

## Behind the scenes
What's happening behind the scenes is a little hacky until we get an API endpoint which tells us all the forms we have saved.

The `FormIntroButtons` component is responsible for the actual querying of the saved data for now. When the component is loaded, it'll try to query for the data. If we signed in on the page that has the `FormIntroButtons`, it'll re-query because the last one will inevitably have failed (`no-auth`). This will update the buttons to allow us to "load" the data.

I struggled with semantics a little bit because there are only so many ways to express that we're "loading" the data. First, we load the saved data into the redux store at `form.loadedData`. This holds both `formData` and `metadata`. When we click the button to "load" the saved form, we transfer the data from `form.loadedData.formData` to `form.data`.

Ideally, we'd get the information about whether the form is filled in upon login and just check it in `FormIntroButtons`, then make the query when the button is clicked, but...Can't do that yet.

## Still to come
- Redirect upon save
- Error messages upon failed save (for various reasons)
- All the tests 😞


src/js/common/schemaform/SaveFormLink.jsx
```diff
@@ -0,0 +1,73 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
+
+import LoginModal from '../components/LoginModal';
+
+// TODO: Come up with a better name than SaveFormLink
```
cvalarida: Wide open to suggestions to this and all names.

src/js/common/schemaform/SaveFormLink.jsx
```diff
@@ -0,0 +1,73 @@
...
+          title="Sign in to save your application"
+          onClose={this.closeLoginModal}
+          visible={this.state.modalOpened}/>
+    </div>);
+
+    if (this.props.loggedIn) {
+      content = <a onClick={saveForm}>Save and come back later</a>;
+    }
+
+    if (savedStatus === 'no-auth') {
```
cvalarida: Shouldn't get here, but just in case.

src/js/common/schemaform/SaveFormLink.jsx
```diff
@@ -0,0 +1,73 @@
...
+      content = <p>no-auth message</p>;
+    } else if (savedStatus === 'failure') {
+      content = <p>failure message</p>;
+    } else if (savedStatus === 'pending') {
+      content = <p>spinner or something</p>;
+    } else if (savedStatus === 'success') {
+      content = <p>success message</p>;
```
cvalarida: Also shouldn't get here, but I'm just prototyping this for now.

src/js/common/schemaform/SaveFormLink.jsx
```diff
@@ -0,0 +1,73 @@
...
+    if (this.props.loggedIn) {
+      content = <a onClick={saveForm}>Save and come back later</a>;
+    }
+
+    if (savedStatus === 'no-auth') {
+      content = <p>no-auth message</p>;
+    } else if (savedStatus === 'failure') {
+      content = <p>failure message</p>;
+    } else if (savedStatus === 'pending') {
+      content = <p>spinner or something</p>;
```
cvalarida: @goldenmeanie What should we put here when the save request is pending and when it's failed.

src/js/common/schemaform/actions.js
```diff
@@ -44,6 +48,35 @@ export function setSubmitted(response) {
...
+}
+
+export function setLoadedData(data) {
+  return {
+    type: SET_LOADED_DATA,
+    data
+  };
+}
+
+export function loadData() {
```
cvalarida: How many ways _can_ one skin a cat??

src/js/common/schemaform/actions.js
```diff
@@ -81,3 +114,182 @@ export function submitForm(formConfig, form) {
...
+ *
+ * @param  {Object}  savedData    The formData from the old version of the form.
+ * @param  {Ingeter} savedVersion The version of the form the corresponding
+ *                                 data was saved with.
+ * @param  {Array}   migrations   An array of functions which transform the
+ *                                 data saved to work with the current version.
+ * @return {Object}               The modified formData which should work with
+ *                                 the current version of the form.
+ */
+function getUpdatedFormData(savedData, savedVersion, migrations) {
```
cvalarida: Still need to test this.

src/js/common/schemaform/actions.js
```diff
@@ -81,3 +114,182 @@ export function submitForm(formConfig, form) {
...
+
+/**
+ * Saves the form data to the back end
+ * @param  {String}  formId    The form's formId
+ * @param  {Ingeter} version   The form's version
+ * @param  {String}  returnUrl The last URL the user was at before saving
+ * @param  {Object}  formData  The data the user has entered so far
+ */
+export function saveFormData(formId, version, returnUrl, formData) {
+  // Double stringify because of api reasons. Olive Branch issues, methinks.
```
cvalarida: There was [some chatter](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3046#issuecomment-305308118) about this the other day, but until I'm "officially" told not to, this is working.

src/js/common/schemaform/reducers/index.js
```diff
@@ -76,7 +80,14 @@ export default function createSchemaFormReducer(formConfig) {
...
         hasAttemptedSubmit: false
-      }
+      },
+      savedStatus: 'not-attempted',
+      loadedStatus: 'not-attempted',
+      version: formConfig.version,
+      formId: formConfig.formId,
+      disableSave: formConfig.disableSave,
+      // loadedData: undefined
+      migrations: formConfig.migrations
```
cvalarida: I'm not convinced this is the best place to store the migrations, but I wasn't sure how else to get it from the form config to the loading function.

src/js/hca-rjsf/components/IntroductionPage.jsx
```diff
@@ -27,32 +59,27 @@ class IntroductionPage extends React.Component {
...
 }
 
-export default withRouter(IntroductionPage);
+function mapStateToProps(state) {
+  return {
+    loggedIn: state.user.login.currentlyLoggedIn
+  };
+}
+
+export default connect(mapStateToProps)(IntroductionPage);
```
cvalarida: I guess this means I should move it from `components/` to `containers/`, eh?

src/sass/base/_b-utils.scss
```diff
@@ -0,0 +1,3 @@
+.full-width {
+  width: 100%;
```
cvalarida: Is there something else that does this? I couldn't find anything...

src/sass/modules/_m-modal.scss
```diff
@@ -152,7 +152,7 @@
   white-space: nowrap;
 }
 
-.va-modal-close {
+button.va-modal-close {
```
cvalarida: The close button's size was being overridden by another style because the `<LoginModal>` was a child of a `.form-progress` element (`FormPage`?).

cvalarida (commented):
Still a WIP, but it's pretty close.

src/js/common/schemaform/SaveFormLink.jsx
```diff
@@ -0,0 +1,73 @@
...
+    return content;
+  }
+}
+
+SaveFormLink.propTypes = {
+  saveForm: PropTypes.func.isRequired,
+  savedStatus: PropTypes.string.isRequired
+};
+
+// I'd rather not have this linked to redux, but then again, I'd rather not pass
```
jbalboni: It seems like this is generally going to be the direct child of a connected component, so there shouldn't be much prop threading. And I'm not sure we should worry about re-renders if this is going to route to another page

src/js/common/schemaform/actions.js
```diff
@@ -6,6 +6,10 @@ export const SET_DATA = 'SET_DATA';
 export const SET_PRIVACY_AGREEMENT = 'SET_PRIVACY_AGREEMENT';
 export const SET_SUBMISSION = 'SET_SUBMISSION';
 export const SET_SUBMITTED = 'SET_SUBMITTED';
+export const SET_SAVED = 'SET_SAVED';
```
jbalboni: I think we should break these actions and action creators into a separate file.

src/js/common/schemaform/actions.js
```diff
@@ -81,3 +114,182 @@ export function submitForm(formConfig, form) {
...
+  //   },
+  //   ...
+  // ]
+  // The functions transform the data from version of their index to the next one up.
+  // This works because every time the version is bumped on the form, it's because
+  //  the saved data needs to be manipulated, so there will be no skipped versions.
+  if (!migrations || typeof migrations[savedVersion] !== 'function') {
+    return savedData;
+  }
+  return getUpdatedFormData(migrations[savedVersion](savedData), savedVersion + 1, migrations);
```
jbalboni: I think in this case making this function recursive obscures what's happening, which is that you're iterating through the migrations and applying each one. I think `reduce` might be clearer.

src/js/common/schemaform/SaveFormLink.jsx
jbalboni: Should probably put these statuses in an object that can be shared around, so these aren't magic strings.

src/js/hca-rjsf/components/IntroductionPage.jsx
jbalboni: Yeah, it would also be nice to figure out how to share more of this code between the different forms. This is sort of tedious to do for each form.

Maybe some kind of higher order component that passes the info box thing to an intro page component, so it can put it wherever it wants?

cvalarida: Yeah, I spent some time pondering that, but eventually just put it in here anyhow. It's a bit of a pain this way, but using a higher-order component didn't seem like it'd be much less painful. I figure either way, we'll have to modify all the intro pages, and while `connect()`ing them like this _does_ make them less reusable as components...they're very specific components anyhow. :man_shrugging:

Either way, I'm not attached to doing it like this. I was mostly going for a proof of concept.

src/js/common/schemaform/SaveFormLink.jsx
cvalarida: Yeah, I that was one of those things I meant to do but never got around too. Will do!

src/js/common/schemaform/SaveFormLink.jsx
cvalarida: So would it be better to pass the props after all?

src/js/common/schemaform/actions.js
cvalarida: Oh. Well yes, yes it would. Not sure why I immediately went to recursion, now that I think about it. 😕

src/js/common/schemaform/SaveFormLink.jsx
jbalboni: I think so.

src/js/common/schemaform/actions.js
jbalboni: This is my pass at the naming:

`saveInProgressForm`
`fetchInProgressForm`
`setFetchInProgressStatus`
`setSaveInProgressStatus`
`loadInProgressFormData`

src/js/common/schemaform/actions.js
```diff
@@ -81,3 +114,180 @@ export function submitForm(formConfig, form) {
...
+        //  transformed. Sentry error?
+        return Promise.reject('invalid-data');
+      }
+      // Set the data in the redux store
+      // NOTE: Until we get the api for the list of filled forms, we're using this
+      //  function to see if the form has been filled in, so this is setting the
+      //  data in a separate place to be pulled in when we _actually_ want to load
+      //  the formData.
+      // dispatch(setData(formData));
+      dispatch(setLoadedData({ formData, metadata: resBody.metadata }));
```
jbalboni: Do these two actions need to be separate? Could we set the status in the first action?

src/js/hca-rjsf/components/IntroductionPage.jsx
jbalboni: Yeah, this is fine for now.

src/js/common/schemaform/actions.js
cvalarida: 🤔 I'm not sure I follow you. What do you mean?

jbalboni: In the reducer for this action, can you set the data like you're doing now and also set the status to 'success?' Instead of firing two actions?

src/js/common/schemaform/actions.js
raquelromano-gov: Just read the chatter...sounds like a TODO would suffice saying to remove the extra stringify once the backend does whatever it needs to do.

src/js/common/schemaform/actions.js
cvalarida: Those are much better names. Thank you!

src/js/common/components/LoginModal.jsx
```diff
@@ -0,0 +1,95 @@
...
+    const wasLoggedIn = this.props.user.login.currentlyLoggedIn;
+    const isLoggedIn = nextProps.user.login.currentlyLoggedIn;
+    if (!wasLoggedIn && isLoggedIn) {
+      this.props.onClose();
+    }
+  }
+
+  // Copied from src/js/login/containers/Main.jsx
+  componentWillUnmount() {
+    if (this.loginUrlRequest && this.loginUrlRequest.abort) {
```
raquelromano-gov: Curious..when is `abort` not defined?

raquelromano-gov (approved):
LGTM on the login stuff! LMK if you want to offload some of the login testing stuff to me since that was a project I obviously left unfinished.

src/js/common/components/LoginModal.jsx
cvalarida: I don't remember exactly, but I did run into it, so I added that in there.

cvalarida:
@raquelromano-gov If you could tackle the login testing, that'd be stupendous!

src/js/common/schemaform/actions.js
cvalarida: Ohhh, gotcha. Yeah. Will do.

cvalarida:
Had to refactor the `FormIntroButtons` and `SaveFormLink` to not be `connect()`ed for testability purposes. I don't like moving all that logic out into the `IntroductionPage`, but I'm planning on moving it up still more into a HOC that we can wrap our intro pages in.  It's ready for review again. (Though apparently I have to figure out what's up with the e2e tests, so...more on that when I figure it out.)

src/js/common/schemaform/FormPage.jsx
```diff
@@ -94,13 +100,32 @@ class FormPage extends React.Component {
...
+    const returnUrl = this.props.location.pathname;
+    this.props.saveInProgressForm(formId, version, returnUrl, data);
+
+    // TODO: Build this page and make it accessible (and customizable) to all forms
+    // this.router.push('/form-saved');
+  }
+
   render() {
-    const { route, params, form } = this.props;
+    const { route, params } = this.props;
```
jbalboni: What's the reason for moving `form` out of here? Because it's only used once?

src/js/common/schemaform/save-load-actions.js
```diff
@@ -0,0 +1,239 @@
+import environment from '../helpers/environment.js';
+
```
jbalboni: Seems like eslint should be bugging you about the extra blank line.

src/js/edu-benefits/1990-rjsf/config/form.js
```diff
@@ -13,6 +13,9 @@ const formConfig = {
   urlPrefix: '/1990-rjsf/',
   submitUrl: '/v0/education_benefits_claims/1990',
   trackingPrefix: 'edu-',
+  formId: '1990',
+  version: 0,
+  disableSave: true,
```
jbalboni: Opting into disabling save in progress while it isn't done makes me nervous. I get that it would make it cleaner later. Maybe we should also put a `__BUILDTYPE__ === 'development'` check where it's used in the common code, too?

jbalboni (approved):
LGTM. Just had some minor comments (and I want to be very careful about this accidentally getting enabled).

src/js/common/schemaform/FormPage.jsx
cvalarida: Mmm...I think it was only used once before I merged master back in, but I don't remember.

src/js/common/schemaform/save-load-actions.js
cvalarida: Huh...Now that you mention it, yes. It doesn't get angry with me until I enter 6 blank lines, though. Weird.

src/js/edu-benefits/1990-rjsf/config/form.js
cvalarida: Sounds good. 👍

PR #5678, opened 2017-06-02
Actually default children array in hca rjsf

jbalboni:
I fixed this before, but multiple failures caused me to miss it:

1. The `children` array is only being checked if financial disclosure is true and/or is only stripped on the front end in that case
2. Chai's `to.be.defined` check was returning false positives.

The fix is is a pain, because we have multiple steps where this particular array can be removed from the output. I gave up trying to make some kind of interface to do this when calling `transformForSubmit` and just had hca overwrite the whole thing.

### DIFF
--- a/src/js/common/schemaform/helpers.js
+++ b/src/js/common/schemaform/helpers.js
@@ -204,7 +204,7 @@ export function filterInactivePages(pages, form) {
   }, form.data);
 }
 
-export function stringifyForm(key, value) {
+export function stringifyFormReplacer(key, value) {
   // an object with country is an address
   if (value && typeof value.country !== 'undefined' &&
     (!value.street || !value.city || (!value.postalCode && !value.zipcode))) {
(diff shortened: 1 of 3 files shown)


src/js/common/schemaform/helpers.js
```diff
@@ -204,6 +204,24 @@ export function filterInactivePages(pages, form) {
   }, form.data);
 }
 
+export function stringifyForm(key, value) {
```
cvalarida: Mmm...I think this name is misleading since it doesn't actually stringify the form, but is used in the stringification process. Maybe something more like `cullFields`...or something.

But then, I'm no good at naming; I already feel sorry for my unborn children.

test/hca-rjsf/schema/schema.unit.spec.js
```diff
@@ -26,17 +26,17 @@ describe('hca schema tests', () => {
         }
         if (data.veteranAddress.country === 'USA') {
           expect(data.veteranAddress.zipcode).to.not.be.empty;
-          expect(data.veteranAddress.postalCode).not.to.be.defined;
+          expect(typeof data.veteranAddress.postalCode).to.equal('undefined');
```
cvalarida: Well that's...odd... 😕

cvalarida (approved):
Well this looked like a fun bug to track down!

src/js/common/schemaform/helpers.js
jbalboni: Yeah, I agree. I'm going to change it to `stringifyFormReplacer`, since the mdn docs call that parameter a "replacer" function.

PR #5680, opened 2017-06-02
508 fix: change element ID and name on 404 page

raquelromano-gov:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3087

### DIFF
--- a/content/pages/404.md
+++ b/content/pages/404.md
@@ -28,7 +28,7 @@ permalink: false
                 <input name="utf8" type="hidden" value="&#x2713;" />
               </div>
               <div class="va-flex va-flex--top va-flex--jctr">
-                <input id="affiliate" name="affiliate" type="hidden" value="vets.gov_search" />
+                <input id="affiliate-1" name="affiliate-1" type="hidden" value="vets.gov_search" />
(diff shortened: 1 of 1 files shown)


omgitsbillryan (approved):
So long as nothing relies on this particular `ID` , LGTM.

cvalarida (approved):
Famous last words: Well that was easy.

raquelromano-gov:
Yeah, no idea if anything relies on those IDs being what they are, but will keep an eye out for anything after the next production deploy.

cvalarida:
I did a search in the entire project for it and found nothing, so :crossed_fingers: it looks good. Unless it's dynamically generated (which would be weird for this, I think).

raquelromano-gov:
Yeah, I did that before changing it, but still wondered if some monitoring tool could rely on it for something. You made me paranoid so I just searched sentry logs for `affiliate`: http://sentry.vetsgov-internal/vets-gov/website-production/?query=is%3Aunresolved+affiliate
I can't make any sense of the errors though. They've been around for a while and look like something inside the VA network. I can't even tell if it's a coincidence or the same `affiliate`

PR #5685, opened 2017-06-05
Implement Vet Center designs

bshyong:
tagging https://github.com/department-of-veterans-affairs/vets-api/pull/1025 for awareness


src/js/facility-locator/components/ServicesAtFacility.jsx
```diff
@@ -117,10 +68,12 @@ class ServicesAtFacility extends Component {
   }
 
   renderVetCenterServices() {
-    // to be hardcoded
     return (
       <div className="mb2">
         <ul>
+          {vetCenterServices.map(s => {
+            return <li key={s}>{s}</li>;
```
U-DON: Is a key separated by spaces valid?

src/js/facility-locator/components/FacilityMap.jsx
```diff
@@ -17,6 +17,7 @@ class HowToGetHere extends Component {
       va_health_facility: 'health',
       va_cemetery: 'cemetery',
       va_benefits_facility: 'benefits',
+      vet_center: 'vet-centers',
```
U-DON: Would it be easier to maintain this by transforming the values of `facilityTypes` from `config` using `_.kebabCase`?

U-DON (approved):
LGTM.

src/js/facility-locator/components/FacilityMap.jsx
bshyong: Yes, good catch. I'd like to do some clean up on this once both the Access to Care and Vet Center features are deployed..

The switch to camelCase I implemented in the Access to Care branch so once that gets in I can also implement your suggestion

PR #5686, opened 2017-06-05
Fix sidebar visibility issues for disability claims

bshyong:
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3076

### DIFF
--- a/src/sass/modules/_m-nav-sidebar.scss
+++ b/src/sass/modules/_m-nav-sidebar.scss
@@ -14,14 +14,25 @@
     padding: 1.6rem;
     position: absolute;
     top: 0;
-    transform: scale(0, 1);
-    transform-origin: right center;
-    transition: transform 250ms ease-out;
+    transform: translateX(100%) scale(0);
+    transition: transform ease-out 250ms;
     display: block;
 
     // Necessary to override .usa-grid .usa-width-one-fourth
     .va-l-detail-page & {
        width: 100%;
+    }
+
+    &::before {
+      background: $color-link-default-hover;
+      content: "\00A0";
+      display: block;
+      height: 100%;
+      left: 0;
+      position: fixed;
+      top: 0;
+      width: 100%;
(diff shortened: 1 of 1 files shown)


melwoodard (commented):
@bshyong this looks good. Is there an easy way to bring back the slide-out transition that used to happen when the menu is opened?

U-DON (approved):
LGTM. Might be a way to maintain the animated sliding effect, but not sure.

bshyong:
I'll add this in—we can't hide the element completely if we want to keep transitions; we have to make it invisible with a height/width of 0 instead.

melwoodard (changes requested):
The scale effect still isn't really right. If it isn't easy to do the slide-in without visible scaling, I say let's just keep it simple and take out the transition effect.

bshyong:
@melwoodard updated—take a look at this and see what you think

http://recordit.co/d1dZ8pGOS2

src/sass/modules/_m-nav-sidebar.scss
```diff
@@ -49,7 +39,19 @@
     }
 
     &--opened {
-      transform: translateX(0);
+      transform: scale(1);
+
+      &+ div:before {
```
U-DON: Super nitpick: usually I leave a space between the `&` and the operator.

melwoodard (commented):
@bshyong better, but still seems funky with the scale effect stretching the menu out when it appears. I'm looking at this locally too and now the gray transparent overlay is on top of the menu, preventing me from interacting with it. I don't want to keep you spinning on this forever, so let's go back to the static transition for now and I can put in an issue to come back to the fine-tuning later.

bshyong:
Updated—no animation now and we can improve this later and add animations back in

melwoodard (approved):
Thank you! Looks great, and I'll add the animation issue for the future.

PR #5689, opened 2017-06-05
Add max length for middle name and maiden name on hca

jbalboni:
The ES backend has a max length of 30 for some name fields, so we need to set that.

We may want to set this on the backend, too, but this is ok for now.

### DIFF
--- a/src/js/hca/config/form.js
+++ b/src/js/hca/config/form.js
@@ -152,7 +152,7 @@ const formConfig = {
             type: 'object',
             properties: {
               veteranFullName,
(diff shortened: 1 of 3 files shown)


src/js/hca/config/form.js
```diff
@@ -121,7 +122,7 @@ const formConfig = {
   defaultDefinitions: {
     date,
     provider,
-    fullName,
+    fullName: _.set('properties.middle.maxLength', 35, fullName),
```
markolson: Sorry, it's 30 for name parts

markolson (approved):

cvalarida (approved):

PR #5690, opened 2017-06-05
Add burials scaffolding

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2880

I decided to make this it's own app with subfolders, assuming the preneed form will be a separate app. There did not appear to be much burials specific overlap between the two.


src/js/burials/components/IntroductionPage.jsx
```diff
@@ -0,0 +1,58 @@
...
+          <div className="small-6 usa-width-five-twelfths medium-5 end columns">
+            <ProgressButton
+                onButtonClick={this.goForward}
+                buttonText="Continue"
+                buttonClass="usa-button-primary"
+                afterText="»"/>
+          </div>
+        </div>
+        <div className="omb-info--container">
+          <OMBInfo resBurden={30} ombNumber="2900-0091" expDate="05/31/2018"/>
```
cvalarida: I'm sure this will be changed when we know what to change it to, but can we fill this in with question marks until then to make it obvious?

src/js/burials/helpers.jsx
```diff
@@ -0,0 +1,161 @@
+import React from 'react';
```
cvalarida: We'll probably need to gut this, yeah?

cvalarida (approved):

src/js/burials/helpers.jsx
jbalboni: Huh, I thought I did.

src/js/burials/components/IntroductionPage.jsx
jbalboni: I can actually get that off the paper form, so I'll fill in the real values

cehsu (approved):

PR #5691, opened 2017-06-05
Create extended version of AskVAQuestions component and include in ED…

cehsu:
…U and HCA forms

Connects to department-of-veterans-affairs/vets.gov-team#2637
Connects to department-of-veterans-affairs/vets.gov-team#2638

EDU
__________________________________________________________________________________________________________
screen shot 2017-06-12 at 9 09 55 pm

HCA
__________________________________________________________________________________________________________
screen shot 2017-06-12 at 10 23 02 am


cehsu:
@J-Quag, @goldenmeanie, @jbalboni Is this also to be included in the pensions forms?

cehsu (commented):
These changes affect not only EDU, and HCA forms but so also pensions, is this the appropriate scope here? Also, some styles have been duplicated and I wonder if modularizing/reusing by renaming and relocating them in base/_va.scss should be done here or if another ticket should be created for that.

src/js/common/schemaform/FormApp.jsx
```diff
@@ -68,8 +71,7 @@ export default class FormApp extends React.Component {
           {formConfig.title && !isIntroductionPage && <FormTitle title={formConfig.title} subTitle={formConfig.subTitle}/>}
           {content}
         </div>
-        <div className="usa-width-one-third medium-4 columns show-for-medium-up">
-        </div>
+        <div className="columns usa-width-one-third medium-4"><AskVAQuestions/></div>
```
cehsu: By including this here, we are including in all forms that use this component, including pension forms, is that within the scope of these changes?

src/sass/modules/_m-schemaform.scss
```diff
@@ -231,3 +231,41 @@ label {
     padding-left: 25px !important;
   }
 }
+
+//duplicated from disability-benefits.scss
```
cehsu: Should these be rename and reused by relocating in base/_va.scss? Or should another ticket be created for that?

jbalboni:
@cehsu I think these are supposed to be at the bottom of the form all the time. That's what we decided in some planning meeting (and there's a comment into those two issues to that effect).

I'm not sure which forms this should apply to. That's a question for @J-Quag.

J-Quag:
hey thanks, Yeah, the screenshots showed the right side but during our grooming session everyone seemed to think the bottom might be a better place to start. 
I'm personally fine with either.

A few things, does it scroll with the page or is it static? Just looking at the height of it, thinking maybe it should appear under the page header and not aligned with it?

Right now this is for HCA and the Education forms. 
  I can get the numbers for the Pension and Burials forms and we can include those as well.

jbalboni:
So it sounds like there are going to be different numbers for different forms, so @cehsu I think you probably want to make this something we can pass in through the form config object. Similar to how we set `introduction` to an introduction page component, have an `askSupport` property that you can set to a form specific version of `AskVAQuestions`. The edu forms can probably all use the same one, at least.

src/js/common/schemaform/FormApp.jsx
cvalarida: I'm with @jbalboni. We already have access to `formConfig`, so it'd be super easy to do something like:
```jsx
const contactInfo = typeof formConfig.contactInfo === 'function'
  ? formConfig.contactInfo
  : null;

...
  {formConfig.title && !isIntroductionPage && <FormTitle title={formConfig.title} subTitle={formConfig.subTitle}/>}
  {content}
</div>
{contactInfo}
<span className="js-test-location hidden" data-location={trimmedPathname} hidden></span>
...
```
But with hopefully a better name than `contactInfo` :man_shrugging:

cvalarida (commented):
Yeah, we decided on this being below the buttons because of ad blindness; if we have stuff floating to the right, people tend to ignore it because they're usually ads. This may not hold as true with our particular audience, but it's still good practice, I imagine.

src/js/common/schemaform/FormApp.jsx
```diff
@@ -65,14 +63,16 @@ export default class FormApp extends React.Component {
...
+          <Element name="topScrollElement"/>
+          <div className="usa-width-two-thirds medium-8 columns">
+            {formConfig.title && !isIntroductionPage && <FormTitle title={formConfig.title} subTitle={formConfig.subTitle}/>}
+            {content}
+          </div>
+          <span className="js-test-location hidden" data-location={trimmedPathname} hidden></span>
         </div>
-        <div className="columns usa-width-one-third medium-4"><AskVAQuestions/></div>
-        <span className="js-test-location hidden" data-location={trimmedPathname} hidden></span>
+        {formConfig.getHelp && formConfig.getHelp(phoneNumber)}
```
cehsu: @cvalarida @jbalboni This is non-idiomatic, and I'd appreciate your thoughts, I realize this is contrary to the initial proposal of only specifying a component in the config (as opposed to the phone number as well), but thought it might be worth exploring.

cehsu:
Thanks @J-Quag, @jbalboni, and @cvalarida for feedback on where the help information should appear, changes made, now visible beneath buttons.

src/js/common/schemaform/AskVAQuestions.jsx
```diff
@@ -0,0 +1,32 @@
...
+  return (
+    <div className="row">
+      <div className="usa-width-two-thirds medium-8 columns">
+        <div className="ask-va-questions help-sidebar-box">
+          <h2 className="help-h2-border help-h2">Need Help?</h2>
+          <p className="talk">Enrollment or Eligibility questions:</p>
+          <p className="phone-number">
+            <a href="`tel:+${props.phoneNumber}`">{props.phoneNumber}</a><br/>
+            Monday - Friday, 8:00 a.m. - 8:00 p.m. ET<br/>
+            TTY: <a href="tel:+1-800-829-4833">800-829-4833</a>
```
jbalboni: If we passed a different number through props, this one would also likely change.

src/js/common/schemaform/FormApp.jsx
```diff
@@ -62,15 +63,16 @@ export default class FormApp extends React.Component {
...
+        <div className="row">
+          <Element name="topScrollElement"/>
+          <div className="usa-width-two-thirds medium-8 columns">
+            {formConfig.title && !isIntroductionPage && <FormTitle title={formConfig.title} subTitle={formConfig.subTitle}/>}
+            {content}
+          </div>
+          <span className="js-test-location hidden" data-location={trimmedPathname} hidden></span>
         </div>
-        <span className="js-test-location hidden" data-location={trimmedPathname} hidden></span>
+        {formConfig.getHelp && formConfig.getHelp(phoneNumber)}
```
cehsu: @cvalarida @jbalboni This is non-idiomatic, and I'd appreciate your thoughts, I realize this is contrary to the initial proposal of only specifying a component in the config (as opposed to the phone number as well), but thought it might be worth exploring.

jbalboni: It looks like the info for hca and edu differs by more than the phone number (there's some initial text that's different).

I wonder if we should make a default AskVAQuestions component that takes a form specific component as a child. It looks like the bottom paragraph will be the same across the forms. It's the just the top one that can vary.

So AskVAQuestions could be:

```
function AskVAQuestions(props) {
  return (
    <div className="row">
      <div className="usa-width-two-thirds medium-8 columns">
        <div className="ask-va-questions help-sidebar-box">
          <h2 className="help-h2-border help-h2">Need Help?</h2>
          {props.children}
          <p className="talk">To report a problem with this form,<br/>
          call the Vets.gov Technical Help Desk:</p>
          <p className="phone-number">
            <a href="tel:+1-855-574-7286">855-574-7286</a><br/>
            TTY: <a href="tel:+1-855-574-7286">855-574-7286</a><br/>
            Monday - Friday, 8:00 a.m. - 8:00 p.m. ET
          </p>
        </div>
      </div>
    </div>
  );
}
```
And then an hca specific component would have this content:

```
<p className="talk">Enrollment or Eligibility questions:</p>
<p className="phone-number">
  <a href="tel:+877-222-8387">877-222-8387</a><br/>
  Monday - Friday, 8:00 a.m. - 8:00 p.m. ET<br/>
  TTY: <a href="tel:+1-800-829-4833">800-829-4833</a>
</p>
```

And in FormPage you could have:

```
// towards the beginning of the render function
const GetFormHelp = formConfig.getHelp;
...
// In the jsx being returned
<AskVAQuestions><GetFormHelp/></AskVAQuestions>
```

src/js/common/schemaform/FormApp.jsx
```diff
@@ -62,15 +63,16 @@ export default class FormApp extends React.Component {
...
+      <div>
+        <div className="row">
+          <Element name="topScrollElement"/>
+          <div className="usa-width-two-thirds medium-8 columns">
+            {formConfig.title && !isIntroductionPage && <FormTitle title={formConfig.title} subTitle={formConfig.subTitle}/>}
+            {content}
+          </div>
+          <span className="js-test-location hidden" data-location={trimmedPathname} hidden></span>
         </div>
-        <span className="js-test-location hidden" data-location={trimmedPathname} hidden></span>
```
cehsu: @jbalboni, this is actually on line 73 now, unless you meant that is should be moved back down.

src/js/common/schemaform/AskVAQuestions.jsx
```diff
@@ -0,0 +1,32 @@
+import PropTypes from 'prop-types';
```
cvalarida: Maybe a bit premature, but this doesn't look like it's schemaform-specific (though it is only being used in schemaform for now, I realize). Might it be worth it to put this in `src/js/common/components`?

src/sass/modules/_m-schemaform.scss
cehsu: Thanks @jbalboni, created [here](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3193).

cehsu:
Thanks @jbalboni, changes made.

src/js/edu-benefits/components/GetFormHelp.jsx
```diff
@@ -0,0 +1,17 @@
...
+
+function GetFormHelp() {
+  return (
+    <div>
+      <p className="talk">For help filling out this form,<br/>
+      ask the Education Call Center:</p>
+      <p className="phone-number">
+        <a href="tel:+888-442-4551">888-442-4551</a><br/>
+        Monday - Friday, 8:00 a.m. - 7:00 p.m. ET<br/>
+        TTY: <a href="tel:+1-800-829-4833">???-???-????</a>
```
cehsu: I've put placeholders here as I'm following up with @J-Quag about what information should be provided (TTY number, email).

cehsu:
Thanks @J-Quag for the education service question submission link, changes made.

J-Quag (changes requested):
HCA TTY Number: TTY: 800-829-4833
   This is for both E&E and Vets.gov line. HRC apparently only has one TTY number. So the HRC TTY number in the EDU form should be that as well.

cehsu:
Thanks @J-Quag, changes made.

src/js/common/schemaform/FormApp.jsx
```diff
@@ -62,14 +64,17 @@ export default class FormApp extends React.Component {
...
-        <div className="usa-width-one-third medium-4 columns show-for-medium-up">
+      <div>
+        <div className="row">
+          <Element name="topScrollElement"/>
+          <div className="usa-width-two-thirds medium-8 columns">
+            {formConfig.title && !isIntroductionPage && <FormTitle title={formConfig.title} subTitle={formConfig.subTitle}/>}
+            {content}
+          </div>
         </div>
+        <AskVAQuestions>
```
jbalboni: Could we hide this on the confirmation page? Doesn't seem useful there (there are already phone numbers and they've completed the form).

jbalboni (approved):
Looks good, had one question about hiding this on the confirmation page, once that's resolved it's good to merge.

cehsu:
Thanks @jbalboni, changes made.

src/js/edu-benefits/1990/containers/EduBenefitsApp.jsx
```diff
@@ -82,6 +84,7 @@ class EduBenefitsApp extends React.Component {
 
     const trimmedPathname = currentLocation.pathname.replace(/\/$/, '');
     const isIntroductionPage = trimmedPathname.endsWith('introduction');
+    const isConfirmationPage = trimmedPathname.endsWith('confirmation');
```
jbalboni: On the 1990, the confirmation page has a url that ends with `submit-message`.

jbalboni:
@cehsu Found one thing to fix.

@J-Quag You need to approve the PR so we can merge (since you requested change).

src/js/edu-benefits/1990/containers/EduBenefitsApp.jsx
cehsu: Thanks again @jbalboni, change made.

J-Quag:
Gotcha @cehsu hate to do this. But the on edu, the education call center business line does not have a tty, so that is why the submit a question is there. The tech hrc desk uses that tty number...  

Sorry its annoying. Hrc staffs the e&e and vg number and they only have 1 tty number. But the education call ctr doesnt even have a tty line.

cehsu:
Thanks @J-Quag, that makes a lot of sense! Appreciate the additional explanation, change made, EDU TTY removed and screenshot updated.

J-Quag (approved):

PR #5692, opened 2017-06-05
Styling and logic tweaks to Terms Acceptance

bshyong:
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3070


U-DON (approved):
LGTM. Just now noticing the usage of `dangerouslySetInnerHTML` in a couple places that concern me, but that's a separate issue.

bshyong:
We should discuss if you have concerns—there is some balance between maintaining the formatting of the terms content that is located in the API and security IRT dangerously setting the HTML. Account creation is not finalized yet so we can easily make changes

PR #5693, opened 2017-06-05
Confirm navigating away when preferences have changed.

U-DON:
Resolves department-of-veterans-affairs/vets.gov-team#2780.

Because of how different browsers and versions handle custom messages when leaving pages, there can be different messages when navigating away from unsaved changes.

Below, the first scenario is handled by React Router and the latter is by plain JS, but some browsers that allow the custom message may show the former.


bshyong (approved):
LGTM

PR #5694, opened 2017-06-06
GIB status action to pull data

annekainicUSDS:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3095

Action to pull mock data for now. Waiting to add all the helper functions for the `apiRequest` in the event we extract that out into `common/`.

screen shot 2017-06-05 at 9 19 39 pm

### DIFF
--- a/script/build.js
+++ b/script/build.js
@@ -125,7 +125,7 @@ smith.metadata({ buildtype: options.buildtype });
 const ignore = require('metalsmith-ignore');
 const ignoreList = [];
 if (options.buildtype === 'production') {
-  ignoreList.push('education/gi-bill/post-9-11/status');
+  ignoreList.push('education/gi-bill/post-9-11/status/*');
   ignoreList.push('pensions/application.md');
   ignoreList.push('va-letters/*');
   ignoreList.push('education/apply-wizard/');
--- a/src/js/post-911-gib-status/actions/post-911-gib-status.js
+++ b/src/js/post-911-gib-status/actions/post-911-gib-status.js
@@ -1,0 +1,77 @@
+// TODO: remove this hard-coded response once we can fetch from a vets-api endpoint
+const chapter33EnrollmentResponse = {
+  chapter33EducationInfo: {
+    dateOfBirth: '1977-10-01T04:00:00.000+0000',
+    firstName: 'Jean',
+    lastName: 'Picard',
+    regionalProcessingOffice: 'Central Office Washington, DC',
+    vaFileNumber: '301010301',
+    enrollmentList: [
+      {
+        beginDate: '2012-11-01T04:00:00.000+0000',
+        endDate: '2012-12-01T05:00:00.000+0000',
+        facilityCode: '11902614',
+        facilityName: 'PURDUE UNIVERSITY',
+        fullTimeHours: 12,
+        intervalChoice: 'NO_RESTRICTIONS',
+        participantId: '11170323',
+        trainingType: 'UnderGrad',
+        vacationDayCount: 0
+      },
+      {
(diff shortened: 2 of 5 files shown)


src/js/post-911-gib-status/actions/post-911-gib-status.js
```diff
@@ -0,0 +1,77 @@
...
+    dateOfBirth: '1977-10-01T04:00:00.000+0000',
+    firstName: 'Jean',
+    lastName: 'Picard',
+    regionalProcessingOffice: 'Central Office Washington, DC',
+    vaFileNumber: '301010301',
+    enrollmentList: [
+      {
+        beginDate: '2012-11-01T04:00:00.000+0000',
+        endDate: '2012-12-01T05:00:00.000+0000',
+        facilityCode: '11902614',
```
raquelromano-gov: Not relevant to this PR, but comparing this response to the UI, we'll have to roll up the enrollment objects by facility code on the front end since we're keeping the vets-api piece as thin as possible. That seems fine with me given there won't be hundreds of enrollments (I assume?), but want to make sure we're all on board with this happening on the client.

src/js/post-911-gib-status/actions/post-911-gib-status.js
```diff
@@ -0,0 +1,77 @@
+// TODO: remove this hard-coded response once we can fetch from a vets-api endpoint
+const chapter33EnrollmentResponse = {
```
raquelromano-gov: Should we call it `post911GibStatusResponse` for consistency with the rest of the code? Even if the contents of the response will be an `chapter33EducationInfo` object

src/js/post-911-gib-status/reducers/index.js
```diff
@@ -0,0 +1,25 @@
+import set from 'lodash/fp/set';
+
+const initialState = {
+  enrollmentData: null,
+  available: false
+};
+
+function post911GIBillStatus(state = initialState, action) {
```
raquelromano-gov: We may want to pick either GIBillStatus or GibStatus throughout to keep us sane, but not critical

raquelromano-gov (approved):

src/js/post-911-gib-status/actions/post-911-gib-status.js
annekainicUSDS: Yep, that makes sense, will make that change!

src/js/post-911-gib-status/actions/post-911-gib-status.js
annekainicUSDS: Ah good thing to note. Yes I think that's fine too. We can probably write a helper function that transforms the data as we need it.

src/js/post-911-gib-status/reducers/index.js
annekainicUSDS: yes good point. let's go with GibStatus, to be consistent with the file names :) I'll make the change

annekainicUSDS:
Pushed up the naming changes mentioned in the review, will merge once tests pass.

script/build.js
```diff
@@ -125,7 +125,7 @@ smith.metadata({ buildtype: options.buildtype });
 const ignore = require('metalsmith-ignore');
 const ignoreList = [];
 if (options.buildtype === 'production') {
-  ignoreList.push('education/gi-bill/post-9-11/status');
+  ignoreList.push('education/gi-bill/post-9-11/status/*');
```
annekainicUSDS: @raquelromano-gov I noticed that this app isn't currently being blocked in production. I think this change will fix that.

PR #5696, opened 2017-06-06
Consistent alignment of columns for new templates

U-DON:
Resolves department-of-veterans-affairs/vets.gov-team#2988.


U-DON:
Looks like the visual issues were mostly related to the attempts to switch over to the USWDS grid system and not using the Foundation grid as a fallback. In particular, the breadcrumbs line uses the Foundation grid, but the content under it was using USWDS, and they have different paddings.

So I added back in Foundation-related classes to align the elements consistently, while leaving in some remnants of the USWDS bits that didn't seem to affect the alignment as a marker for the future when we can follow through with the refactor.

@melwoodard: Can you go through and see if I didn't miss anything? I checked these pages after making changes to the relevant templates:
* https://vetsgov-pr-5696.herokuapp.com/education/gi-bill/ (detail-page)
* https://vetsgov-pr-5696.herokuapp.com/education/ (level2-index)
* https://vetsgov-pr-5696.herokuapp.com/employment/job-seekers/ (topic-landing)

There were also changes to the `navigation-link-list` include and component (which, by the way, why are multiple things similarly and confusingly named?), so hopefully they didn't break other things...

melwoodard (commented):
Thanks @U-DON, this looks great! The only other place the `navigation-link-list` affects is the homepage, but this branch doesn't have those updates from master yet. Can you pull those down here so we can confirm the homepage layout still works?

melwoodard (changes requested):
I just noticed a z-index bug with the mobile slide-out menu. I don't see it on other branches:

melwoodard (approved):
Looks good!

U-DON:
Great, going to merge this to the `content-template-update` branch. Note that the build is still breaking from the broken link, but you probably have a handle on that.

PR #5697, opened 2017-06-06
Load hca confirmation page even if there's an error

jbalboni:
Sentry error: http://sentry.vetsgov-internal/vets-gov/website-production/issues/7204/

It looks like people are refreshing the confirmation page on hca, which is now an error due to how we changed the page. This avoids the error and hides the date if you refresh.

### DIFF
--- a/src/js/hca/containers/ConfirmationPage.jsx
+++ b/src/js/hca/containers/ConfirmationPage.jsx
@@ -34,9 +34,11 @@ class ConfirmationPage extends React.Component {
         </div>
         <div>
           <p>We are processing your application. The Department of Veterans Affairs will contact you when we finish our review.</p>
-          <div className="success-alert-box">
-            <p className="success-copy">Form submitted: {time && moment(time).format('MMMM D, YYYY, h:mm a')}</p>
-          </div>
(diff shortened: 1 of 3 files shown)


cehsu (approved):
LGTM.

src/js/hca/containers/ConfirmationPage.jsx
```diff
@@ -34,7 +35,7 @@ class ConfirmationPage extends React.Component {
         <div>
           <p>We are processing your application. The Department of Veterans Affairs will contact you when we finish our review.</p>
           <div className="success-alert-box">
-            <p className="success-copy">Form submitted: {moment(time).format('MMMM D, YYYY, h:mm a')}</p>
+            <p className="success-copy">Form submitted: {time && moment(time).format('MMMM D, YYYY, h:mm a')}</p>
```
cehsu: Should the entire alert box be conditionally rendered?

jbalboni: Yeah, you're right, it's probably better to hide the box entirely

cvalarida (approved):

PR #5703, opened 2017-06-07
Removed unnecessary class name and element and extraneous comment appearing in rendered JSX.

U-DON:
Resolves department-of-veterans-affairs/vets.gov-team#2976.


bshyong (approved):
LGTM

PR #5705, opened 2017-06-07
Gib status user info component

annekainicUSDS:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3101

This PR does a few things:
1. Adds the general message at the top of the page about being able to use this in place of the COE

2. Creates a component for the user information section
There is still more work to do here because I'm waiting on 2 things:
- Updated UI from MarvelApp: might result in slight tweaks
- Updated mock data from vets-api: depending on what the data looks like, we'll need to transform it to be formatted as we want when it is displayed to the user. I'm not tackling that yet as I don't yet know the final version of the data.

3. Fixes an issue with the `gatePage` flag
Background on this issue: I noticed that the bottom margin on the breadcrumbs on this page was being collapsed. I traced this to a class called `va-nav-breadcrumbs--gate` that was being set on the breadcrumbs if the `gatePage` flag was set to true (https://github.com/department-of-veterans-affairs/vets-website/blob/master/content/includes/breadcrumbs.html#L1). This `gatePage` flag was also used to determine whether or not to render the breadcrumbs entirely in the `page-react` layout (https://github.com/department-of-veterans-affairs/vets-website/blob/master/content/layouts/page-react.html#L3).

I wanted to include the breadcrumbs, but I didn't want to add that extra class to collapse the margin. Because the same flag was used to control both, I needed to create a separate flag, `includeBreadcrumbs`, just to control whether or not breadcrumbs should be included. I kept the `gatePage` flag to add that class; I removed it from this app specifically so it wouldn't be set.

screen shot 2017-06-08 at 11 19 57 am

### DIFF
--- a/src/js/post-911-gib-status/components/UserInfoSection.jsx
+++ b/src/js/post-911-gib-status/components/UserInfoSection.jsx
@@ -1,0 +1,63 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import moment from 'moment';
+
+class UserInfoSection extends React.Component {
+  render() {
+    const { userData } = this.props;
+
+    // Used to get today's date to show information current of
+    const today = new Date();
+    const todayFormatted = moment(today).format('MM/DD/YYYY');
+
+    return (
+      <div className="gibstatus">
+        <h3 className="section-header">{userData.firstName} {userData.lastName}</h3>
+        <div className="usa-alert usa-alert-info">
+          <div className="usa-alert-body">
+            <h4 className="usa-alert-heading">This information is current as of {todayFormatted}</h4>
+          </div>
+        </div>
+        <div className="usa-grid-full section-line">
+          <div className="usa-width-one-third">
+            <span><strong>Date of Birth: </strong></span>
+          </div>
+          <div className="usa-width-one-third">
+            {userData.dateOfBirth}
+          </div>
+        </div>
+        <div className="usa-grid-full section-line">
+          <div className="usa-width-one-third">
+            <span><strong>VA File Number: </strong></span>
+          </div>
+          <div className="usa-width-one-third">
+            {userData.vaFileNumber}
+          </div>
+        </div>
+        <div className="usa-grid-full section-line">
+          <div className="usa-width-one-third">
+            <span><strong>Regional Processing Office: </strong></span>
+          </div>
+          <div className="usa-width-one-third">
+            {userData.regionalProcessingOffice}
+          </div>
+        </div>
+        <div>
+          <h4>When You Can Receive Benefits</h4>
+          <div className="section-line">You are eligible to receive benefits between <strong>{userData.eligibilityDate}</strong> and <strong>{userData.delimitingDate}</strong></div>
+        </div>
+        <div>
+          <h4>Your Benefit Level</h4>
+          <div className="section-line">You are eligible to receive benefits at a rate of <strong>{userData.percentageBenefit}</strong></div>
+        </div>
+      </div>
+    );
+  }
+}
+
+UserInfoSection.propTypes = {
+  userData: PropTypes.object
+};
+
+export default UserInfoSection;
--- a/src/js/post-911-gib-status/components/UserInfoSection.jsx
+++ b/src/js/post-911-gib-status/components/UserInfoSection.jsx
@@ -1,24 +1,23 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-import { getDateFormatted, getPercentFormatted } from '../utils/helpers';
+import { formatDate, formatPercent } from '../utils/helpers';
 
 class UserInfoSection extends React.Component {
   render() {
     const { userData } = this.props;
 
     // Used to get today's date to show information current of
-    const today = new Date();
-    const todayFormatted = getDateFormatted(today);
+    const todayFormatted = formatDate(new Date());
 
     // Check if percent is a number
-    let percentageBenefit = 'unavailable';
-    if (getPercentFormatted(userData.percentageBenefit)) {
-      percentageBenefit = getPercentFormatted(userData.percentageBenefit);
-    }
+    const percentageBenefit = formatPercent(userData.percentageBenefit) || 'unavailable';
+
+    // TODO: Figure out what will be sent to us if the user is ineligible so we can
+    // conditionally show "Currently Disallowed" message
 
     return (
-      <div className="gibstatus">
+      <div>
         <h3 className="section-header">{userData.firstName} {userData.lastName}</h3>
         <div className="usa-alert usa-alert-info">
           <div className="usa-alert-body">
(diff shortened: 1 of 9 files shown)


src/js/post-911-gib-status/utils/helpers.js
```diff
@@ -0,0 +1,15 @@
+import moment from 'moment';
+
+export function getDateFormatted(date) {
```
raquelromano-gov: These can probably be called simply `formatDate()` and `formatPercent()` since we aren't actually getting the date or percent value from anywhere.

src/js/post-911-gib-status/components/UserInfoSection.jsx
```diff
@@ -0,0 +1,69 @@
...
+
+import { getDateFormatted, getPercentFormatted } from '../utils/helpers';
+
+class UserInfoSection extends React.Component {
+  render() {
+    const { userData } = this.props;
+
+    // Used to get today's date to show information current of
+    const today = new Date();
+    const todayFormatted = getDateFormatted(today);
```
raquelromano-gov: These could be collapsed into a single since we're not using `today` anywhere else:
```
const today = formatDate(new Date());
```

src/js/post-911-gib-status/components/UserInfoSection.jsx
```diff
@@ -0,0 +1,69 @@
...
+
+    // Check if percent is a number
+    let percentageBenefit = 'unavailable';
+    if (getPercentFormatted(userData.percentageBenefit)) {
+      percentageBenefit = getPercentFormatted(userData.percentageBenefit);
```
raquelromano-gov: Another possible 1-liner:
```
const percentageBenefit = formatPercent(userData.percentageBenefit) || 'unavailable';
```

raquelromano-gov (approved):
This makes me happy because I don't have to figure out half of the styling for the enrollment history component. :)

LGTM--we can tweak things as new design and data info comes in.

src/js/post-911-gib-status/components/UserInfoSection.jsx
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    return (
+      <div className="gibstatus">
+        <h3 className="section-header">{userData.firstName} {userData.lastName}</h3>
+        <div className="usa-alert usa-alert-info">
+          <div className="usa-alert-body">
+            <h4 className="usa-alert-heading">This information is current as of {todayFormatted}</h4>
```
raquelromano-gov: I had a question on the mocks for this: this notice isn't shown on the view where there is actual enrollment history data available (only when either the user is not eligible, or there is no data available). @phiden is that by design or should we always show this notice?

src/js/post-911-gib-status/components/UserInfoSection.jsx
```diff
@@ -0,0 +1,69 @@
...
+
+    // Check if percent is a number
+    let percentageBenefit = 'unavailable';
+    if (getPercentFormatted(userData.percentageBenefit)) {
+      percentageBenefit = getPercentFormatted(userData.percentageBenefit);
+    }
+
+    return (
+      <div className="gibstatus">
+        <h3 className="section-header">{userData.firstName} {userData.lastName}</h3>
```
raquelromano-gov: Also just noticed that the UI for pages (1) and (2) when there is no enrollment history available includes the header defined here showing the user's name, but for (3) when enrollment history is available, the user's name isn't in a header element; just listed with the data:
image

Going off this document:
https://github.com/department-of-veterans-affairs/vets.gov-team/blob/master/Products/EVSS%20Integration/Post-911%20Enrollment%20Status/design/post-911_complete_UI.pdf

src/js/post-911-gib-status/components/UserInfoSection.jsx
```diff
@@ -0,0 +1,69 @@
...
+          <div className="usa-width-one-third">
+            <span><strong>Regional Processing Office: </strong></span>
+          </div>
+          <div className="usa-width-one-third">
+            {userData.regionalProcessingOffice}
+          </div>
+        </div>
+        <div>
+          <h4>When You Can Receive Benefits</h4>
+          <div className="section-line">You are eligible to receive benefits between <strong>{getDateFormatted(userData.eligibilityDate)}</strong> and <strong>{getDateFormatted(userData.delimitingDate)}</strong></div>
```
raquelromano-gov: Maybe add a TODO here to find out how the enrollment data will indicate that the person is ineligible for Post-9/11 GIB benefits (so we can display the "Currently Disallowed" warning).

src/js/post-911-gib-status/components/UserInfoSection.jsx
```diff
@@ -0,0 +1,69 @@
...
+    const todayFormatted = getDateFormatted(today);
+
+    // Check if percent is a number
+    let percentageBenefit = 'unavailable';
+    if (getPercentFormatted(userData.percentageBenefit)) {
+      percentageBenefit = getPercentFormatted(userData.percentageBenefit);
+    }
+
+    return (
+      <div className="gibstatus">
```
raquelromano-gov: Is there a file missing that defines this? Not seeing it in the sass file

src/js/post-911-gib-status/utils/helpers.js
annekainicUSDS: Yep, good point, will change

src/js/post-911-gib-status/components/UserInfoSection.jsx
annekainicUSDS: Yes much better, thanks!

src/js/post-911-gib-status/components/UserInfoSection.jsx
annekainicUSDS: I thought I saw in Slack @phiden mention something that she was going to move this message to the top of this section for all screens, but hadn't updated the mocks yet, which is why I built it this way. But I could have misunderstood something!

src/js/post-911-gib-status/components/UserInfoSection.jsx
annekainicUSDS: I feel like they should be as consistent as possible, unless there's a good reason they're not?

src/js/post-911-gib-status/components/UserInfoSection.jsx
annekainicUSDS: Nope! I was using this earlier and then realized I didn't need it :) will remove!

src/js/post-911-gib-status/components/UserInfoSection.jsx
annekainicUSDS: Yes, good point, will add

annekainicUSDS:
Going to merge while we await final design confirmation so we can move forward on the other tickets.

PR #5706, opened 2017-06-07
Update fetch polyfill

jbalboni:

We're seeing some failures in IE11 when posting forms and I want to make sure we're not hitting bugs because we use an older version of the fetch polyfill. This change affects primarily IE11, but also some older iOS and Android devices.

The only breaking change listed for the fetch polyfill in 2.0 is using `getAll()` with the Headers object, which I can't find any instances of us doing.

I also removed some instances of `mode: 'cors'` because that's the default and is unnecessary.

### DIFF
--- a/package.json
+++ b/package.json
@@ -178,7 +178,7 @@
     "webpack": "^2.4.1",
     "webpack-manifest-plugin": "^1.1.0",
     "webpack-md5-hash": "0.0.5",
-    "whatwg-fetch": "http://registry.npmjs.org/whatwg-fetch/-/whatwg-fetch-1.0.0.tgz",
+    "whatwg-fetch": "^2.0.3",
     "winston": "^2.2.0"
   },
   "optionalDependencies": {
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -106,7 +106,7 @@
     "libxmljs": "^0.18.0",
     "lodash": "^4.5.1",
     "lodash-deep": "^2.0.0",
-    "mapbox": "^1.0.0-beta5",
```
jbalboni: We're already using beta7, this just cleans up the yarn.lock file a little (it had multiple versions of whatwg-fetch before I changed this).

cvalarida (approved):
It'd be nice to have some cross-browser testing right about now, eh? Hmm... 🤔 Well, it looks good to me.

U-DON (approved):

bshyong (approved):
LGTM

ayaleloehr:
@cvalarida if you want to use SauceLabs, here are the credentials: https://github.com/department-of-veterans-affairs/vets.gov-team/blob/master/Work%20Practices/Cross-Platform%20Testing/cross-platform-testing-plan.md

Not automated (yet?), but at least it's an easier way of using IE than I had before!

PR #5707, opened 2017-06-07
Pv terms soft launch

patrickvinograd:
Fixes implementation of terms acceptance. 

Adds an endpoint that logged-in users can visit to activate themselves for the beta features. The beta features are all conditionalized on the backend, frontend code does not need to vary. 

The only area that's not super-accurate for beta vs. not is the error messages for e.g. an ineligible non-patient.

### DIFF
--- a/src/js/health-beta/containers/HealthBetaEnrollment.jsx
+++ b/src/js/health-beta/containers/HealthBetaEnrollment.jsx
@@ -6,7 +6,7 @@ import { registerBeta } from '../actions';
 
 class HealthBetaEnrollment extends React.Component {
 
-  componentWillMount() {
+  componentDidMount() {
     this.props.registerBeta();
   }
 
--- a/src/js/health-beta/containers/HealthBetaEnrollment.jsx
+++ b/src/js/health-beta/containers/HealthBetaEnrollment.jsx
@@ -1,0 +1,70 @@
+import React from 'react';
+import { connect } from 'react-redux';
+
+import RequiredLoginView from '../../common/components/RequiredLoginView';
+import { registerBeta } from '../actions';
+
+class HealthBetaEnrollment extends React.Component {
+
+  componentWillMount() {
+    this.props.registerBeta();
+  }
+
+  render() {
+    let message;
+
+    if (this.props.loading === true) {
+      message = 'Activating beta features...';
+    }
+    if (this.props.stats === 'failed') {
+      message = 'Activation failed, please contact beta product manager.';
+    } else {
+      message = 'Beta features activated for user '.concat(this.props.username).concat('. Thank you.');
+    }
+    let view;
+
+    view = (
+      <div className="row">
+        <div className="usa-width-two-thirds medium-8 small-12 columns">
+          <h1>Health Beta Registration</h1>
+          <div>
+            <p>{message}</p>
+          </div>
+        </div>
+      </div>
+    );
+
+    return (
+      <div>
+        <RequiredLoginView
+            authRequired={1}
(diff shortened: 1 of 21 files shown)


patrickvinograd:
The entire health-beta endpoint is temporary and would go away once we launch for real.

src/js/user-profile/actions/index.js
```diff
@@ -72,5 +73,7 @@ export function acceptTerms(termsName) {
       }),
       () => dispatch({ type: ACCEPTING_LATEST_MHV_TERMS_FAILURE })
     );
+
+    getUserData(dispatch);
```
patrickvinograd: @raquelromano-gov I tagged you specifically in the context of what we discussed on Slack yesterday about being able to refresh the `/user` API endpoint. Invoking this here in this action seems to do the trick, and the corresponding redux state gets updated. Any issues with invoking this?

patrickvinograd:
The health terms and conditions modal on the user profile page is still broken in this PR, looking at how to plumb props through to it.

src/js/health-beta/health-beta-entry.jsx
```diff
@@ -0,0 +1,43 @@
...
+   *
+   * Plan is to make this trigger a sort when the query
+   * parameter is `sortby`.
+   */
+  const handleChangedURL = (event) => {
+    // Here so eslint doesn't tell us about an unused variable.
+    return event;
+  };
+  browserHistory.listen(handleChangedURL);
+  // End URL listening
```
U-DON: These irrelevant comments and history event listener keep getting copy-pasta'd as new apps are get created. They can be removed.

src/js/health-beta/containers/HealthBetaEnrollment.jsx
```diff
@@ -0,0 +1,70 @@
+import React from 'react';
+import { connect } from 'react-redux';
+
+import RequiredLoginView from '../../common/components/RequiredLoginView';
+import { registerBeta } from '../actions';
+
+class HealthBetaEnrollment extends React.Component {
+
+  componentWillMount() {
```
U-DON: Same comment as before about using `componentDidMount` instead for invoking API requests.

bshyong:
@patrickvinograd do you need any help on this?

src/js/user-profile/actions/index.js
U-DON: Do you intend to call this synchronously? If you want to `getUserData` in response to the POST request (which I'm assuming is what you wanted to do), you'll want to add it to the handlers.

Like instead of:

```
() => dispatch({ type: ACCEPTING_LATEST_MHV_TERMS_SUCCESS }),
```

Perhaps you might want:

```
() => {
  dispatch({ ... });
  getUserData(dispatch);
},
```

Otherwise, you're doing the fetch from `apiRequest` and then immediately calling `getuserData`. The completion of the request could occur after the `getUserData` since it's asynchronous.

patrickvinograd:
@bshyong I feel comfortable with putting what I have up in dev if you don't see any red flags.

src/js/user-profile/actions/index.js
bshyong: ^ 👍

raquelromano-gov: LGTM!

bshyong:
@patrickvinograd I can push a fix for the profile page model to the `acct` branch after this is merged

bshyong:
Fix is here https://github.com/department-of-veterans-affairs/vets-website/pull/5711

PR #5709, opened 2017-06-07
File upload prototype for rjsf forms

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3132

This is an initial pass at a file upload field. It's not clear if we're going to do file uploads in the form itself or have a special upload section on the review page. This work is for the former approach and we can adapt it to the latter approach if necessary.

Tests are a bit sparse because this is very much not the final design.
screen shot 2017-06-07 at 12 04 21 pm
screen shot 2017-06-07 at 11 46 06 am
screen shot 2017-06-07 at 11 44 07 am
screen shot 2017-06-07 at 11 44 00 am

### DIFF
--- a/src/js/common/schemaform/FileField.jsx
+++ b/src/js/common/schemaform/FileField.jsx
@@ -97,17 +97,18 @@ export default class FileField extends React.Component {
                       type="button">
                     Remove
                   </button>
-                </li>}
-                {!file.uploading && !!errors.length &&
-                  <li>{errors[0]}&nbsp;
-                    <button
-                        onClick={() => this.removeFile(index)}
-                        className="usa-button usa-button-outline schemaform-file-remove-button"
-                        type="button">
-                      Remove
-                    </button>
-                  </li>
-                }
+                </li>
+              }
+              {!file.uploading && !!errors.length &&
+                <li>{errors[0]}&nbsp;
+                  <button
+                      onClick={() => this.removeFile(index)}
+                      className="usa-button usa-button-outline schemaform-file-remove-button"
+                      type="button">
+                    Remove
+                  </button>
+                </li>
+              }
             </ul>
           );
         })}
--- a/src/js/common/schemaform/FileField.jsx
+++ b/src/js/common/schemaform/FileField.jsx
@@ -15,42 +15,35 @@ export default class FileField extends React.Component {
       .then(fileInfo => {
         const uploadingFiles = this.props.formData;
         this.props.onChange(_.set(uploadingFiles.length - 1, {
-          serverErrorMessage: null,
+          errorMessage: null,
           uploading: false,
           fileName: fileInfo.fileName,
-          fileSize: fileInfo.fileSize,
+          maxSize: fileInfo.maxSize,
           confirmationNumber: fileInfo.confirmationNumber
         }, uploadingFiles));
       })
       .catch(message => {
-        const uploadingFiles = this.props.formData;
+        const uploadingFiles = this.props.formData || [];
+        this.props.onBlur(this.props.idSchema.$id);
         this.props.onChange(_.set(uploadingFiles.length - 1, {
           uploading: false,
-          serverErrorMessage: message
+          errorMessage: message
         }, uploadingFiles));
       });
-    // setTimeout(() => {
-    //   const uploadingFiles = this.props.formData;
-    //   this.props.onChange(_.set(uploadingFiles.length - 1, {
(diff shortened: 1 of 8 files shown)


src/sass/modules/_m-schemaform.scss
```diff
@@ -231,3 +231,12 @@ label {
     padding-left: 25px !important;
   }
 }
+
+.usa-button.schemaform-file-remove-button {
+  width: auto;
```
cvalarida: Nitpicky, but I'd like to see a bit of left padding too (@goldenmeanie to confirm, though).

src/js/common/schemaform/FileField.jsx
```diff
@@ -0,0 +1,161 @@
...
+              {file.uploading && 'Uploading file...'}
+              {file.confirmationNumber &&
+                <li>{file.fileName} ({displayFileSize(file.maxSize)})&nbsp;
+                  <button
+                      onClick={() => this.removeFile(index)}
+                      className="usa-button usa-button-outline schemaform-file-remove-button"
+                      type="button">
+                    Remove
+                  </button>
+                </li>}
```
cvalarida: Nit: Can we move that `}` down a line? I skipped right over it when I was reading this.

src/js/common/schemaform/FileField.jsx
```diff
@@ -0,0 +1,161 @@
...
+      }
+
+      return Promise.reject(resp.statusText);
+    }, error => {
+      return Promise.reject(error.message);
+    });
+  }
+
+  removeFile = (index) => {
+    const newFileList = this.props.formData.filter((file, idx) => index !== idx);
```
cvalarida: Not that it matters much given the tiny numbers we'll be dealing with here, but I was curious, so I did a jsperf and found this:
image
So for what it's worth, if we're dealing with large arrays and we need to return a clone of the array with only one index removed, slice and splice is the fastest. Good to know! (Eventually....maybe...?)

src/js/common/schemaform/FileField.jsx
```diff
@@ -0,0 +1,161 @@
...
+        this.props.onChange(_.set(uploadingFiles.length - 1, {
+          errorMessage: null,
+          uploading: false,
+          fileName: fileInfo.fileName,
+          maxSize: fileInfo.maxSize,
+          confirmationNumber: fileInfo.confirmationNumber
+        }, uploadingFiles));
+      })
+      .catch(message => {
+        const uploadingFiles = this.props.formData || [];
```
cvalarida: Why define `uploadingFiles` here (and in the `.then()` above) instead of using `files`? I mean, we're using lodash/fp, so the `_.set()` on it should return a new array so immutability isn't the issue, yeah? (And if it was, I suppose we'd need to clone `this.props.formData` anyhow, eh?)

My guess is it's because of the possibility of having different props by the time the promise resolves or rejects. Is that it?

src/js/common/schemaform/definitions/file.js
```diff
@@ -0,0 +1,57 @@
...
+      required: 'You must upload a file',
+      minItems: 'You must upload a file'
+    },
+    'ui:validations': [
+      validateFileField
+    ]
+  };
+}
+
+// An example schema so we don't forget it for now
```
cvalarida: 👍
I imagine this will get moved into the schemaform docs when it's a little more stable, yeah?

cvalarida (approved):
Other than the sparse testing you mentioned, it looks good. I just had some non-blocking questions.

src/sass/modules/_m-schemaform.scss
jbalboni: Yeah, the UI is not good, I was just going for "doesn't look too broken." Alex and Stacy will make it look nice once they're done researching/testing.

src/js/common/schemaform/FileField.jsx
jbalboni: Yeah, there's an `onChange` call before the promise, so this will have re-rendered by the time the promise is resolved.

src/js/common/schemaform/FileField.jsx
jbalboni: Hmm, interesting.

src/js/common/schemaform/definitions/file.js
jbalboni: Yeah, we can put it there and I think it would also be a common definition in vets-json-schema.

PR #5710, opened 2017-06-07
In Progress Forms -- tests

cvalarida:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3164

~There are still like three tests to go because I'm not sure how to make them work, but they're mostly there.~

Done! The only thing I don't have tested (to my knowledge) is the various saving / loading UI error handling changes.

### DIFF
--- a/src/js/common/schemaform/FormIntroButtons.jsx
+++ b/src/js/common/schemaform/FormIntroButtons.jsx
@@ -11,6 +11,8 @@ class FormIntroButtons extends React.Component {
     // Load the form data into the redux store if possible
     // NOTE: This does not fill the data into the form yet; just grabs it and
     //  saves it for later (when we want to fill in the blanks).
+    // TODO: Probably should wrap this in
+    // if (this.props.loggedIn) { ... }
     this.props.fetchInProgressForm(this.props.form.formId, this.props.form.migrations);
   }
 
@@ -20,7 +22,7 @@ class FormIntroButtons extends React.Component {
     let wasLoggedIn;
     let isLoggedIn;
     try {
-      wasLoggedIn = this.props.user.login.currentlyLoggedIn;
+      wasLoggedIn = this.props.loggedIn;
     } catch (e) {
       if (e instanceof TypeError) {
         // Probably couldn't read a property somewhere along the way
@@ -28,7 +30,7 @@ class FormIntroButtons extends React.Component {
       }
     }
     try {
-      isLoggedIn = newProps.user.login.currentlyLoggedIn;
+      isLoggedIn = newProps.loggedIn;
     } catch (e) {
       if (e instanceof TypeError) {
         isLoggedIn = false;
--- a/src/js/common/schemaform/reducers/index.js
+++ b/src/js/common/schemaform/reducers/index.js
@@ -144,13 +144,14 @@ export default function createSchemaFormReducer(formConfig) {
         return _.set('savedStatus', action.status, state);
       }
       case SET_FETCH_FORM_STATUS: {
-        const newState = _.set('loadedStatus', action.status, state);
+        return _.set('loadedStatus', action.status, state);
+      }
+      case SET_IN_PROGRESS_FORM: {
+        const newState = _.set('loadedData', action.data, state);
+        // I think this should be in SET_IN_PROGRESS_FORM...
         newState.loadedStatus = LOAD_STATUSES.success;
 
         return newState;
-      }
-      case SET_IN_PROGRESS_FORM: {
-        return _.set('loadedData', action.data, state);
       }
       case LOAD_DATA_INTO_FORM: {
         // Mirrors SET_DATA, but uses state.loadedData
--- a/src/js/common/schemaform/reducers/index.js
+++ b/src/js/common/schemaform/reducers/index.js
@@ -148,7 +148,6 @@ export default function createSchemaFormReducer(formConfig) {
       }
       case SET_IN_PROGRESS_FORM: {
         const newState = _.set('loadedData', action.data, state);
-        // I think this should be in SET_IN_PROGRESS_FORM...
         newState.loadedStatus = LOAD_STATUSES.success;
 
         return newState;
--- a/src/js/common/schemaform/save-load-actions.js
+++ b/src/js/common/schemaform/save-load-actions.js
@@ -62,7 +62,7 @@ export function loadInProgressDataIntoForm() {
  * @return {Object}               The modified formData which should work with
  *                                 the current version of the form.
  */
-function migrateFormData(savedData, savedVersion, migrations) {
+export function migrateFormData(savedData, savedVersion, migrations) {
   // migrations is an array that looks like this:
   // [
   //   (savedData) => {
@@ -113,14 +113,15 @@ export function saveInProgressForm(formId, version, returnUrl, formData) {
     // If we don't have a userToken, fail safely
     if (!userToken) {
       dispatch(setSaveFormStatus(SAVE_STATUSES.noAuth)); // Shouldn't get here, but...
-      return;
+      return Promise.reject('no auth'); // Returning a rejected promise for testing purposes only
     }
 
     // Update UI while we're waiting for the API
     dispatch(setSaveFormStatus(SAVE_STATUSES.pending));
 
     // Query the api
-    fetch(`${environment.API_URL}/v0/in_progress_forms/${formId}`, {
+    // (returning for testing purposes only)
+    return fetch(`${environment.API_URL}/v0/in_progress_forms/${formId}`, {
       method: 'PUT',
       headers: {
         'Content-Type': 'application/json',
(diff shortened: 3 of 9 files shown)


src/js/common/schemaform/FormIntroButtons.jsx
```diff
@@ -20,15 +22,15 @@ class FormIntroButtons extends React.Component {
     let wasLoggedIn;
     let isLoggedIn;
     try {
-      wasLoggedIn = this.props.user.login.currentlyLoggedIn;
+      wasLoggedIn = this.props.loggedIn;
```
cvalarida: This got overlooked in a refactor and I must not have tested it. That's why we write these things!

src/js/common/schemaform/FormIntroButtons.jsx
```diff
@@ -11,6 +11,8 @@ class FormIntroButtons extends React.Component {
     // Load the form data into the redux store if possible
     // NOTE: This does not fill the data into the form yet; just grabs it and
     //  saves it for later (when we want to fill in the blanks).
+    // TODO: Probably should wrap this in
+    // if (this.props.loggedIn) { ... }
```
cvalarida: Let me know what you think about this. If they're not logged in, they'll get a 401 from the api, so I think we should go ahead with the change, but...I'd like an outside opinion.

src/js/common/schemaform/reducers/index.js
```diff
@@ -144,14 +144,14 @@ export default function createSchemaFormReducer(formConfig) {
         return _.set('savedStatus', action.status, state);
       }
       case SET_FETCH_FORM_STATUS: {
-        const newState = _.set('loadedStatus', action.status, state);
+        return _.set('loadedStatus', action.status, state);
+      }
+      case SET_IN_PROGRESS_FORM: {
+        const newState = _.set('loadedData', action.data, state);
         newState.loadedStatus = LOAD_STATUSES.success;
```
cvalarida: Realized I put this in the wrong `case`. Now it'll only set the load status to `success` after we successfully load, which, y'know, makes sense.

src/js/common/schemaform/save-load-actions.js
```diff
@@ -100,6 +100,7 @@ function migrateFormData(savedData, savedVersion, migrations) {
  */
 export function saveInProgressForm(formId, version, returnUrl, formData) {
   // Double stringify because of api reasons. Olive Branch issues, methinks.
+  // TODO: Stop double stringifying
```
cvalarida: Re: https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3046#issuecomment-305872275
Just thought I'd record this while I was thinking about it. :man_shrugging:

test/common/components/LoginModal.unit.spec.jsx
```diff
@@ -0,0 +1,58 @@
+import React from 'react';
+import { findDOMNode } from 'react-dom';
+import { expect } from 'chai';
+import sinon from 'sinon';
+import ReactTestUtils from 'react-dom/test-utils';
+
+import LoginModal from '../../../src/js/common/components/LoginModal';
+
+describe('<LoginModal>', () => {
```
cvalarida: Still need to test signing in, but since `handleLogin` is called from login helpers, I'm not sure how to do that yet.

test/common/schemaform/FormIntroButtons.unit.spec.jsx
```diff
@@ -0,0 +1,215 @@
...
+    });
+    expect(fetchSpy.calledTwice).to.be.true;
+  });
+  it('should go to the first page when "Continue" is clicked', () => {
+    const routerSpy = {
+      push: sinon.spy()
+    };
+    const fetchSpy = sinon.spy();
+    const loadDataSpy = sinon.spy();
+    const tree = ReactTestUtils.renderIntoDocument(
```
cvalarida: Using `ReactTestUtils.renderIntoDocument()` so I can do a `findDOM.querySelector('.usa-button-primary').click()` later.

test/common/schemaform/SaveFormLink.unit.spec.jsx
```diff
@@ -0,0 +1,84 @@
...
+    );
+    const findDOM = findDOMNode(tree);
+
+    // Open the login modal
+    findDOM.querySelector('a').click();
+
+    const modal = findDOM.querySelector('.va-modal');
+
+    // Find the login modal
+    expect(modal).to.not.be.null;
```
cvalarida: This fails because the `.click()` above isn't triggering `SaveFormLink`'s `openLoginModal()` and I don't know why.

test/common/schemaform/SaveFormLink.unit.spec.jsx
```diff
@@ -0,0 +1,84 @@
...
+            saveForm={saveFormSpy}
+            onUpdateLoginUrl={updateLoginSpy}/>
+      </div>
+    );
+    const findDOM = findDOMNode(tree);
+
+    // "Save" the form
+    findDOM.querySelector('a').click();
+
+    expect(saveFormSpy.called);
```
cvalarida: Confused about the previous test because this works just fine.

cvalarida (commented):
If it's ever tiresome that I leave so many review comments on my own PRs, do let me know. 😛

src/js/common/schemaform/FormIntroButtons.jsx
jbalboni: I think that makes sense.

jbalboni (commented):
This looks good so far. Once you get that one test working, I can take another quick look and approve it.

src/js/common/schemaform/save-load-actions.js
```diff
@@ -193,7 +202,8 @@ export function fetchInProgressForm(formId, migrations) {
       return Promise.reject(status);
     }).then((resBody) => {  // eslint-disable-line consistent-return
```
jbalboni: I think you can remove the eslint comment now.

test/common/schemaform/save-load-actions.unit.spec.js
```diff
@@ -0,0 +1,342 @@
...
+      Promise.all([thunk(dispatch)]).then(() => {
+        expect(dispatch.calledWith(setSaveFormStatus(SAVE_STATUSES.pending))).to.be.true;
+        done();
+      });
+    });
+    it('calls the api to save the form', (done) => {
+      const thunk = saveInProgressForm('hca', 0, 'some/path', {});
+      const dispatch = sinon.spy();
+
+      Promise.all([thunk(dispatch)]).then(() => {
```
jbalboni: Why use Promise.all here, since it's only one promise?

src/js/common/schemaform/save-load-actions.js
```diff
@@ -161,14 +170,14 @@ export function fetchInProgressForm(formId, migrations) {
     // If we don't have a userToken, fail safely
     if (!userToken) {
       dispatch(setFetchFormStatus(LOAD_STATUSES.noAuth)); // Shouldn't get here, but just in case
-      return;
+      return Promise.reject('no auth'); // Returns rejected Promise for testing only
```
jbalboni: If we hit this condition in production, we'll probably get an unhandled promise rejection error here.

jbalboni (approved):
LGTM. Some minor comments/questions.

src/js/common/schemaform/save-load-actions.js
cvalarida: Oh that's true. I was thinking about that the other day but never wrote that down to circle back around to it. I'm thinking wrapping it in an `if (__BUILDTYPE___ === 'development)`. What say ye?

test/common/schemaform/save-load-actions.unit.spec.js
cvalarida: Y'know, I think that was just me being dumb. `thunk(dispatch).then()` / `.catch()` should be fine. Originally, I had to have a way to `.then()` or `.catch()` when I wasn't returning a Promise, but that didn't work out so well after all anyhow, so :man_shrugging:

src/js/common/schemaform/save-load-actions.js
jbalboni: I don't think that's necessary at the moment, since you should hit this if you aren't using save in progress already.

src/js/common/schemaform/save-load-actions.js
```diff
@@ -112,14 +113,19 @@ export function saveInProgressForm(formId, version, returnUrl, formData) {
     // If we don't have a userToken, fail safely
     if (!userToken) {
       dispatch(setSaveFormStatus(SAVE_STATUSES.noAuth)); // Shouldn't get here, but...
-      return;
+      if (__BUILDTYPE__ === 'development') {
+        return Promise.reject('no auth'); // Returning a rejected promise for testing purposes only
+      }
+
+      return; // eslint-disable-line consistent-return
```
cvalarida: Adding these pointless `return`s because eslint gets angry when there's an inconsistent return, and I wanted to record _where_ the inconsistency was.

PR #5711, opened 2017-06-07
Fix terms modal on user profile page

bshyong:
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3072

also handles case where user has already accepted the terms and conditions.

screen shot 2017-06-07 at 4 33 10 pm


U-DON (approved):
LGTM.

PR #5713, opened 2017-06-08
Enrollment history component

raquelromano-gov:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3104

image


src/js/post-911-gib-status/components/EnrollmentHistory.jsx
```diff
@@ -0,0 +1,74 @@
...
+        <div className="feature">
+          Enrollment history does not include flight training, on-the-job-training, apprenticeship training, or correspondence training.
+        </div>
+      );
+    }
+
+    return (
+      <div>
+        <h3 className="section-header">Enrollment History</h3>
+        <div className="section-line">This information is current as of {today}</div>
```
annekainicUSDS: Do we need this line here? Now that we have the blue box appearing in the `UserInfoSection`.

src/js/post-911-gib-status/components/EnrollmentHistory.jsx
```diff
@@ -0,0 +1,74 @@
...
+        <div className="usa-grid-full section-line">
```
annekainicUSDS: Something I did in this PR (https://github.com/department-of-veterans-affairs/vets-website/pull/5717) is extract out this markup into a function that takes the label and the value to be rendered, and renders this markup for each line, since it's pretty repetitive and there are many lines throughout the page that are formatted exactly like this one. Perhaps I could extract that out into a helper to be used across components? Not necessarily for this PR, can do this refactor in a future PR.

src/js/post-911-gib-status/containers/Post911GIBStatusApp.jsx
```diff
@@ -56,7 +57,8 @@ class Post911GIBStatusApp extends React.Component {
                   print a copy of this screen for benefit and eligibility verification.
                 </p>
               </div>
-              <UserInfoSection userData={enrollmentState}/>
+              <UserInfoSection userData={enrollmentData}/>
+              <EnrollmentHistory enrollmentData={enrollmentData}/>
```
annekainicUSDS: Just noticed the props we pass to these two sections are called different things (`userData` and `enrollmentData`). Should we make it consistent since the value is the same? I don't care which one it is.

annekainicUSDS (approved):
LGTM! Just a couple of minor things.

src/js/post-911-gib-status/components/EnrollmentHistory.jsx
raquelromano-gov: Yeah, I saw that and yes, sounds good!

src/js/post-911-gib-status/containers/Post911GIBStatusApp.jsx
raquelromano-gov: Yeah, was thinking we should unify these names to make maintenance easier. I think we can do that in one of the next PRs.

src/js/post-911-gib-status/components/EnrollmentHistory.jsx
raquelromano-gov: Good question. I think after Shay's meeting we should all take a look at the designs  and the current demo on staging and talk through these points.

raquelromano-gov:
I'm going to go ahead and merge even though there is still an outstanding design question so that you can merge in your next PR and we can have a demo ready for design discussions.

PR #5715, opened 2017-06-08
Add veteran info chapter to burials form

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3130

screen shot 2017-06-08 at 11 46 05 am
screen shot 2017-06-08 at 11 46 10 am

screen shot 2017-06-08 at 11 46 19 am
screen shot 2017-06-08 at 11 46 23 am

### DIFF
--- a/src/js/burials/config/form.js
+++ b/src/js/burials/config/form.js
@@ -1,10 +1,22 @@
-// import _ from 'lodash/fp';
+import _ from 'lodash/fp';
 
-// import fullSchemaHca from 'vets-json-schema/dist/10-10EZ-schema.json';
+import fullSchemaBurials from 'vets-json-schema/dist/21P-530-schema.json';
 
 // import { transform } from '../helpers';
 import IntroductionPage from '../components/IntroductionPage';
 import ConfirmationPage from '../containers/ConfirmationPage';
+import { relationshipLabels } from '../labels.jsx';
+
+import fullNameUI from '../../common/schemaform/definitions/fullName';
+
+const {
+  relationship,
+  claimantFullName
+} = fullSchemaBurials.properties;
+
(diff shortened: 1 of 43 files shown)


src/js/burials/config/form.js
```diff
@@ -67,6 +79,66 @@ const formConfig = {
           }
         }
       }
+    },
+    veteranInformation: {
+      title: 'Veteran Information',
+      pages: {
+        veteranInformation: {
```
jbalboni: I could probably use the common applicantInformation page for this, but I'd have to add a couple more configurable bits and it seemed like it was stretching the purpose of that shared page too much.

src/js/burials/config/form.js
```diff
@@ -67,6 +79,66 @@ const formConfig = {
...
+          title: 'Deceased Veteran information',
+          path: 'veteran-information',
+          uiSchema: {
+            'ui:title': 'Deceased Veteran information',
+            veteranFullName: fullNameUI,
+            'view:veteranId': personId.uiSchema(
+              'veteran',
+              'view:veteranId.view:noSSN',
+              'I don’t have the Veteran’s Social Security number'
```
cehsu: Should "Number" in "Social Security Number" be capitalized throughout? You did not introduce this convention here, just wondering if we should follow the labeling instructions @stacyhlinn provided more closely.

cehsu (approved):
LGTM.

src/js/burials/config/form.js
jbalboni: I believe the guidance we've gotten before from @peggygannon and/or the previous copy person was that "number" should be lower case.

PR #5716, opened 2017-06-08
Error handling for client errors on forms

jbalboni:

Creates a different message if we get an error that is not an error from the server.

screen shot 2017-06-08 at 12 41 17 pm

### DIFF
--- a/src/js/common/schemaform/actions.js
+++ b/src/js/common/schemaform/actions.js
@@ -1,4 +1,5 @@
 import Raven from 'raven-js';
+import _ from 'lodash/fp';
 import { transformForSubmit } from './helpers';
 import environment from '../helpers/environment.js';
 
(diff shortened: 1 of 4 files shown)


markolson (approved):

src/js/common/schemaform/review/SubmitButtons.jsx
```diff
@@ -29,6 +29,21 @@ export default function SubmitButtons({ submission, onSubmit, onBack, errorMessa
...
           beforeText="&#10003;"/>
     );
+  } else if (submission.status === 'clientError') {
+    submitButton = (
+      <ProgressButton
+          onButtonClick={onSubmit}
+          buttonText="Submit Application"
+          buttonClass="usa-button-primary"/>
+    );
+    submitMessage = (
```
cvalarida: Not sure that abstracting this out makes much sense, but could we reference these duplicate blocks in a comment in case we have to update the text?

src/js/common/schemaform/actions.js
```diff
@@ -81,15 +82,22 @@ export function submitForm(formConfig, form) {
         return res.json();
       }
 
-      return Promise.reject(new Error(res.statusText));
+      return Promise.reject(new Error(`vets_server_error: ${res.statusText}`));
```
cvalarida: I'm pretty confident, but still want to ask: This `then()` will only be reached if we actually get a _response_ from the server, right? If there's a network issue, it'll bypass here and go straight to the following `catch()`, right? I assume you tested it by throttling your network connection to nil? (If not, I can always give it a go for my own peace of mind.)

cvalarida (approved):

src/js/common/schemaform/actions.js
jbalboni: Correct. You have get an http response code to get in here. Otherwise, fetch will throw an error. I've been testing this by shutting down my vets-api instance (which causes a connection refused error).

src/js/common/schemaform/review/SubmitButtons.jsx
jbalboni: Makes sense, but I don't want to restart the build for that, so I'll make an item for it.

PR #5717, opened 2017-06-08
GIB enrollment period component

annekainicUSDS:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3102

Creates a component for individual enrollment periods. This will eventually be rendered within the `EnrollmentHistory` component, but I put it in the top level app for now.

screen shot 2017-06-08 at 2 52 46 pm


src/js/post-911-gib-status/containers/Post911GIBStatusApp.jsx
```diff
@@ -38,6 +39,17 @@ class Post911GIBStatusApp extends React.Component {
     // TODO: change the service name below from "user-profile" to
     // something like "post-911-gib-status" once its defined in vets-api
     const { enrollmentState } = this.props;
+
+    const enrollmentHistory = enrollmentState && enrollmentState.enrollmentList.length > 0 ? (
+      enrollmentState.enrollmentList.map((enrollment, index) => {
+        return (
+          <EnrollmentPeriod
+              key={index}
```
annekainicUSDS: Apparently you're not supposed to use `index` for the `key`: https://facebook.github.io/react/docs/lists-and-keys.html#keys. But there's not a unique id for each enrollment that I could use instead, so either I just use index or make up a number. Do you have thoughts on what's best here @raquelromano-gov?

raquelromano-gov: How about pre-pending a string literal, e.g. `enrollment-{index}`

raquelromano-gov (approved):
LGTM

src/js/post-911-gib-status/containers/Post911GIBStatusApp.jsx
annekainicUSDS: Yep, will do!

src/sass/post-911-gib-status.scss
```diff
@@ -17,5 +24,9 @@ h4 {
...
 
 .feature {
-    min-height: 0;
-}
+  min-height: 0;
+
+  &.help-desk {
+  	margin-top: 2em;
+  }
+}
```
raquelromano-gov: Missing a newline, but will add it in a follow-up

PR #5719, opened 2017-06-08
Fix exit icon css

bshyong:
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2989

Merging into the content template branch so that these can go in together

### DIFF
--- a/src/sass/base/_va.scss
+++ b/src/sass/base/_va.scss
@@ -1,3 +1,14 @@
+@mixin exit-icon {
+  // Using longhand properties instead of the shorthand to limit
+  // risk and impact of side effects
+  background-image: url(/img/icons/exit-icon.png);
+  background-position: 100% 50%;
+  background-repeat: no-repeat;
+  background-size: 1em auto;
+  padding-right: 1.2em;
+  position: relative;
+}
+
 html, body {
   padding: 0;
   margin: 0;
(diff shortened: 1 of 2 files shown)


U-DON (commented):
Code change LGTM.

Although, I wonder if the intention is for the icon be next to the link itself, which would require a change in the markup rather than CSS.

> screen shot 2017-06-08 at 3 26 20 pm

bshyong:
@melwoodard can you clarify on the above comment? I can change the markup for these cases specifically

melwoodard:
Ah, yes ideally it should be next to the link text instead of at the end of the block. If it has to be a markup change does that mean it needs to be applied manually to every instance of an external link?

bshyong:
Fixed—this is done through css styles so no markup changes are needed

screen shot 2017-06-08 at 5 36 39 pm

U-DON:
Yeah, it can work with just styles too. But probably have to add `h1`, `h2`, etc. to the rule to apply to everything that might be contained within an external link. And probably should apply only to the first header in case there are multiple in that section. Ultimately it's copying the same style of an external link to a header within a nav link list (in which case, might it be helpful to make it a mixin?).

bshyong:
Yup, good suggestions. I'm not sure how much further to optimize in case things change, but will incorporate this feedback

U-DON (approved):

PR #5720, opened 2017-06-08
Always display calendar picker below input regardless of scroll height.

U-DON:
Resolves department-of-veterans-affairs/vets.gov-team#2809.

Apparently `react-datepicker` uses Tether, so I updated the `tetherConstraints` prop accordingly. http://tether.io/#constraints

Considered upgrading `react-datepicker` on the way, as it's up to 0.47.0 now, but it has a small font size that would require overriding in CSS to fix, so forwent it. The relevant props to the component haven't changed anyway.


bshyong (approved):
LGTM

PR #5721, opened 2017-06-08
Move to common styles file

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3066

This is a proof of concept for moving to a common css file for the site. Instead of compiling all of style.scss for each app, we compile it separately and include two css file per page.

This probably needs some discussion, but since I was experimenting and this looks pretty good, I'm going to push it up here.

On my machine, build time goes from ~284s to ~146s with this change.


cehsu (approved):
Exciting!

aub:
Woah, that's amazing! Great work!

src/sass/variables.scss
```diff
@@ -0,0 +1,12 @@
+@charset "UTF-8";
```
cvalarida: Is there some way we can differentiate this from `base/b-variables`? If I'm understanding this correctly, all the `@import`s here are for variables of all kinds, but I think it may get confusing to have a `variables.scss` and a `base/_b-variables.scss`.

cvalarida (approved):
This is awesome.

src/sass/variables.scss
jbalboni: Yeah, I couldn't think of good name for this file. Open to suggestions.

content/includes/header.html
```diff
@@ -47,6 +47,7 @@
 <meta name="msapplication-TileColor" content="#144073">
 
 <!-- CSS -->
+<link rel="stylesheet" href="/generated/style.css">
```
omgitsbillryan: Do any of our bundles _not_ require `style.css`?  Then a followup: if they all require `style.css`, does every bundle require every part of `style.css`? Otherwise `style.css` seems like a dumping ground for all styles and a first-time visitor  may end up downloading a large `style.css` file and only need to use some small percentage of it.

omgitsbillryan (approved):
Reducing build times by 50% is a huge win.  LGTM with one caveat.

content/includes/header.html
jbalboni: The answer to the first question is no, as far as I can tell.

The answer to the second question is probably no, but this is not any different from how it works currently. None of the apps try to break style.scss into just the styles they need. I think there's some room to whittle down the size of style.scss and this change doesn't stop us from doing that.

bshyong (approved):
LGTM

src/sass/variables.scss
cvalarida: How do y'all feel about `globals`? It's not a great name either, but at least there's not a name confusion... :man_shrugging:

content/includes/header.html
cvalarida: Yeah, I think that's a great next step. I think it'd be tough to figure out _how_ to split them all up without breaking things, but after we get visual regression testing, it'd be easier to test _if_ things are broken, anyhow.

Anybody know if there's a tool to find unused styles in a stylesheet? I'll do some poking around, but that seems like a kind of hairy problem to solve...

U-DON (approved):
👍

ayaleloehr:
After discussion in engineering meeting today, this is to be reviewed by either @raquelromano-gov or @annekainicUSDS tomorrow, then should be good to merge if no further questions arise. Very excited about this so thanks @jbalboni for putting it up and everyone else for the reviews!

raquelromano-gov:
Should this connect to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3066?

raquelromano-gov (approved):

src/sass/variables.scss
jbalboni: I changed it to "shared-variables," I think that's a reasonable name and I'll add some comments.

PR #5724, opened 2017-06-09
Switch between health care urls

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2981

The content team is going to be switching the health care urls from `healthcare` to `health-care`, which means the base url for hca needs to change. The change in here is what we did while we switched urls for the claim status app: checking the current url to see what's being used and match it.


cehsu (approved):
LGTM.

rroueche (approved):
Looks good to me - caveated w/the fact that I'm only approving the concept described here, I don't have the technical background to validate the code :)

J-Quag:
we good to close this?

PR #5730, opened 2017-06-09
Pensions pow severance

cehsu:
Connects to department-of-veterans-affairs/vets.gov-team#2223

Matches UI to back end schema

screen shot 2017-06-09 at 1 25 35 pm


src/js/pensions/config/form.js
```diff
@@ -199,10 +199,28 @@ const formConfig = {
...
               'ui:options': {
                 expandUnder: 'view:receivedSeverancePay'
+              },
+              amount: {
+                'ui:title': 'Amount'
+              },
+              type: {
+                'ui:title': 'Pay Type',
+                'ui:widget': 'radio',
+                'ui:options': {
```
jbalboni: You probably don't need the labels here if they're the same as the enum values.

jbalboni (approved):
LGTM, one minor comment.

src/js/pensions/config/form.js
```diff
@@ -199,10 +199,28 @@ const formConfig = {
               'ui:widget': 'yesNo'
             },
             severancePay: {
```
cehsu: Thanks @J-Quag, @jbalboni for noting the UI for this needed to be synced up with the schema.

src/js/pensions/config/form.js
cehsu: Thanks @jbalboni, I thought so too initially, but removing these causes a `enumOptions.map is not a function` error.

J-Quag:
Looks good :)

PR #5733, opened 2017-06-09
Update enrollment list to match backend plus some refactoring

raquelromano-gov:


src/js/post-911-gib-status/components/EnrollmentHistory.jsx
```diff
@@ -1,79 +1,51 @@
...
-        </div>
-        <div className="usa-grid-full section-line">
-          <div className="usa-width-one-third">
-            <span><strong>Remaining: </strong></span>
-          </div>
-          <div className="usa-width-one-third">
-            {enrollmentData.remainingEntitlement}
-          </div>
-        </div>
+        <InfoPair label={'Total months received'} value={enrollmentData.originalEntitlement}/>
```
annekainicUSDS: I don't think you need the curly brackets around this, you should be able to just do `label="Total months received"`.

annekainicUSDS (approved):
LGTM! One minor comment

src/js/post-911-gib-status/components/EnrollmentHistory.jsx
raquelromano-gov: Thanks! Fixed.

PR #5735, opened 2017-06-10
Add marriages pages to pensions form

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2906

This is another of our `pagePerItem` pages, where each item in an array is a page, but with the new wrinkle that you choose the number of items in a text box. So I've created a custom widget for that. I've also put the current page index in `formContext` so it can be used in titles/descriptions.

screen shot 2017-06-09 at 11 56 28 pm
screen shot 2017-06-09 at 11 56 38 pm
screen shot 2017-06-09 at 11 56 45 pm
screen shot 2017-06-09 at 11 56 51 pm

If the person is still marriage and we're at the end, don't show the separation fields:
screen shot 2017-06-09 at 11 57 06 pm

### DIFF
--- a/src/js/common/schemaform/widgets/ArrayCountWidget.jsx
+++ b/src/js/common/schemaform/widgets/ArrayCountWidget.jsx
@@ -25,9 +25,7 @@ export default class ArrayCountWidget extends React.Component {
   // user has chosen '' or 0
   componentWillReceiveProps(props) {
     const arrayData = props.value || [];
-    if (arrayData.length && this.state.value !== arrayData.length) {
-      this.setState({ value: arrayData.length });
-    } else if (!this.state.value && !!arrayData.length) {
+    if (!!arrayData.length && (!this.state.value || this.state.value !== arrayData.length)) {
       this.setState({ value: arrayData.length });
     }
   }
--- a/src/js/common/schemaform/widgets/ArrayCountWidget.jsx
+++ b/src/js/common/schemaform/widgets/ArrayCountWidget.jsx
@@ -1,0 +1,67 @@
+import React from 'react';
+
+// We're expanding or contracting the array based on the count
+// and returning undefined if the array should be empty
+function getValue(count, value = []) {
+  if (!count || parseInt(count, 10) < 1) {
+    return undefined;
+  }
+
+  const intCount = parseInt(count, 10);
+  if (intCount < value.length) {
+    return value.slice(0, intCount);
+  }
+
+  return value.concat(Array(intCount - value.length).fill({}));
+}
+
+export default class ArrayCountWidget extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = { value: !!props.value ? props.value.length : undefined };
+  }
+
+  // Need to keep the count state in sync, but don't want to overwrite if the
+  // user has chosen '' or 0
+  componentWillReceiveProps(props) {
+    const arrayData = props.value || [];
+    if (arrayData.length && this.state.value !== arrayData.length) {
+      this.setState({ value: arrayData.length });
+    } else if (!this.state.value && !!arrayData.length) {
+      this.setState({ value: arrayData.length });
+    }
+  }
+
+  updateArrayLength = (event) => {
+    const count = event.target.value;
+    this.setState({ value: count }, () => {
+      this.props.onChange(getValue(count, this.props.value));
+    });
+  }
+
+  render() {
+    const props = this.props;
+
+    if (props.formContext.reviewMode) {
+      return (
+        <div className="review-row">
+          <dt>{props.uiSchema['ui:title']}</dt><dd>{this.state.value}</dd>
+        </div>
+      );
+    }
+
+    return (
+      <input type="number"
+          step="1"
+          min="1"
+          id={props.id}
+          name={props.id}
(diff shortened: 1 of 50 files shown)


src/js/pensions/components/MarriageTitle.jsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import { getMarriageTitle } from '../helpers';
+
+export default function MarriageTitle({ id, formContext }) {
+  return (
+    <legend
+        className="schemaform-block-title"
+        id={id}
+        tabIndex="-1">
```
cehsu: Why assign to the default value?

test/pensions/components/MarriageTitle.unit.spec.jsx
```diff
@@ -0,0 +1,26 @@
...
+  it('should render first marriage title', () => {
+    const tree = SkinDeep.shallowRender(
+      <MarriageTitle
+          id="id"
+          formContext={{ pagePerItemIndex: 0 }}/>
+    );
+
+    expect(tree.text()).to.contain('First marriage');
+  });
+  it('should render tenth marriage title', () => {
```
cehsu: Tenth? or Eleventh?

test/pensions/config/marriageHistory.unit.spec.jsx
```diff
@@ -0,0 +1,124 @@
...
+      expect(result).to.be.false;
+    });
+  });
+
+  describe('page title', () => {
+    const pageTitle = marriageHistory.title;
+    it('uses word for index', () => {
+      expect(pageTitle({}, { pagePerItemIndex: 0 })).to.equal('First marriage');
+    });
+    it('uses number when over ten', () => {
```
cehsu: Perhaps this could be rephrased, "when contains over ten items/length is greater than or equal to ten" or "when at index ten or greater"

src/js/common/schemaform/widgets/ArrayCountWidget.jsx
```diff
@@ -0,0 +1,67 @@
...
+  constructor(props) {
+    super(props);
+    this.state = { value: !!props.value ? props.value.length : undefined };
+  }
+
+  // Need to keep the count state in sync, but don't want to overwrite if the
+  // user has chosen '' or 0
+  componentWillReceiveProps(props) {
+    const arrayData = props.value || [];
+    if (arrayData.length && this.state.value !== arrayData.length) {
```
cehsu: Should we also check for `this.state.value` here?

src/js/common/schemaform/widgets/ArrayCountWidget.jsx
```diff
@@ -0,0 +1,67 @@
...
+      this.setState({ value: arrayData.length });
+    } else if (!this.state.value && !!arrayData.length) {
+      this.setState({ value: arrayData.length });
+    }
+  }
+
+  updateArrayLength = (event) => {
+    const count = event.target.value;
+    this.setState({ value: count }, () => {
```
cehsu: Why do this here rather than `componentDidUpdate`?

cehsu (approved):
LGTM. Had a few questions.

src/js/pensions/components/MarriageTitle.jsx
jbalboni: Good question. I think this is copied and pasted a couple times. I should take it out.

test/pensions/components/MarriageTitle.unit.spec.jsx
jbalboni: Doh

src/js/common/schemaform/widgets/ArrayCountWidget.jsx
jbalboni: I'm not sure if you mean the `setState` call or the callback.

For the `setState` call: I won't necessarily know what the value in the input element is if I do that in a lifecycle method. The reason for using local state here is that I can track what a user has entered into the text box separately from the length of the array field being changed. If we don't do that, the text field acts weird and won't let you delete a 0.

For the callback: I think this makes the chain of events clearer. If I put the `onChange` call in a lifecycle method, I have to be careful about only calling it when I know it's necessary, which I think is harder code to read (sort of like the `componentWillReceiveProps` method in here, which isn't immediately clear, but I don't have a choice).

cehsu: Thanks for answering my question (the latter) even though it wasn't as clear as it could be. Your explanation makes sense. Just good to keep in mind that using the optional callback is [not the recommended approach](https://facebook.github.io/react/docs/react-component.html) and occasionally causes [unexpected behavior](https://github.com/facebook/react/issues/6320).

jbalboni: That's a good point. I gave it another shot with `componentDidUpdate`, but it leads to calling `onChange` twice on every change and it's hard to wrap your head around the logic to try to avoid that.

jbalboni:
Alex and I were talking about how exactly the spouse/marriage fields should work and that was one thing that came up (calling one of the marriages "Current marriage"). So I think we can do that, but need to get the exact layout of the household fields nailed down first.

PR #5736, opened 2017-06-10
Hide states that don't have facilities

jbalboni:
Sentry error: http://sentry.vetsgov-internal/vets-gov/website-production/issues/7246/
(There are several other related ones)

If you pick a state on the va facility page that doesn't have any facilities, we throw an error. They're stopped by validation and they can pick another state, but it's cluttering Sentry and it's likely confusing why we even show those states. This filters out those states without facilities (this is what the old HCA app was doing).


cehsu (approved):
LGTM.

PR #5737, opened 2017-06-10
Make wcag2a changes for rjsf forms and claim status

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3051
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2828

Fixes the wcag2a issues in forms and claim status.

Verified by running
`yarn test:wcag2a -- test/accessibility/disability-benefits-508.spec.js`
`yarn test:wcag2a -- test/edu-benefits/1990e/01-all-fields.e2e.spec.js`

### DIFF
--- a/src/js/common/components/Modal.jsx
+++ b/src/js/common/components/Modal.jsx
@@ -78,7 +78,7 @@ class Modal extends React.Component {
     }
 
     return (
-      <div className={modalCss} id={id} role="alertdialog" aria-labelledby={id && title ? `${id}-title` : undefined}>
+      <div className={modalCss} id={id} role="alertdialog" aria-labelledby={`${id}-title`}>
         <div className="va-modal-inner">
           {modalTitle}
           {closeButton}
--- a/src/js/common/components/Modal.jsx
+++ b/src/js/common/components/Modal.jsx
@@ -78,7 +78,7 @@ class Modal extends React.Component {
     }
 
     return (
-      <div className={modalCss} id={id} role="alertdialog" aria-labelledby={`${id}-title`}>
+      <div className={modalCss} id={id} role="alertdialog" aria-labelledby={id && title ? `${id}-title` : undefined}>
         <div className="va-modal-inner">
           {modalTitle}
           {closeButton}
--- a/src/js/common/schemaform/review/ReviewCollapsibleChapter.jsx
+++ b/src/js/common/schemaform/review/ReviewCollapsibleChapter.jsx
@@ -81,7 +81,7 @@ export default class ReviewCollapsibleChapter extends React.Component {
       const expandedPages = expandArrayPages(activePages, form.data);
 
       pageContent = (
-        <div id={`collapsible-${this.id}`} className="usa-accordion-content">
+        <div className="usa-accordion-content">
           {expandedPages.map(page => {
             const pageState = form.pages[page.pageKey];
             let pageSchema;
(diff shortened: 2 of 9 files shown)


src/js/common/components/Modal.jsx
```diff
@@ -78,7 +78,7 @@ class Modal extends React.Component {
     }
 
     return (
-      <div className={modalCss} id={id} role="alertdialog" aria-labelledby={`${id}-title`}>
+      <div className={modalCss} id={id} role="alertdialog" aria-labelledby={id && title ? `${id}-title` : undefined}>
```
raquelromano-gov: Should the `id` prop be required so that this is always defined? I found at least one place where it's not passed in (https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/js/rx/components/DiscardChangesModal.jsx#L36)  but it doesn't seem to be tested in e2e tests.

raquelromano-gov (commented):
Thanks, Jeff--I think these fixes are all we need to turn these checks on for the entire site 🎆

src/js/common/components/Modal.jsx
jbalboni: I think it needs an id if you include a title, but otherwise I don't think we would need to force all modals to have an id. I think the id is a new-ish addition to the component.

raquelromano-gov: OK, but how can we make sure the `aria-labelledby` attribute has a valid value if neither `id` nor `title` is defined? 

Also, did you want the value to be `${id}-${title}` or `${id}-title`?

jbalboni: Hmm, yeah, I didn't realize it had to have a label. I'll change that.

I think the latter for the title template, that's what it was before.

raquelromano-gov: Just guessing that screenreaders arent going to like an aria attribute with no value, but I could be wrong.

jbalboni: Well, it should be undefined, and React shouldn't include the attribute at all in the html. But it looks like if you use the `alertdialog` role, you should have a label.

For what it's worth, we did run through all the modals on claim status with the 508 office, and made fixes, but this particular issue didn't come up.

raquelromano-gov: Oh, so you're saying if React sees a div with `aria-labelledby=undefined` it will throw away the entire attribute? That seems more palatable.

I'm really curious to understand what kinds of things the 508 office does/doesn't catch, and similarly what our stricter tests can/can't test....it's all a bit mysterious since I haven't yet seen a comprehensive report from the 508 office about a particular product. Hopefully will soon.

jbalboni: Yeah, as long as it's not the string `'undefined'`.

Yeah, same. It's pretty hard to figure out what the right thing is without constant access to someone who knows how to use the main screen readers well.

jbalboni:
@raquelromano-gov Fixes made.

src/js/common/components/Modal.jsx
```diff
@@ -78,7 +78,7 @@ class Modal extends React.Component {
     }
 
     return (
-      <div className={modalCss} id={id} role="alertdialog" aria-labelledby={id && title ? `${id}-title` : undefined}>
+      <div className={modalCss} id={id} role="alertdialog" aria-labelledby={`${id}-title`}>
```
raquelromano-gov: Thanks for updating the other `<Modal>` components!

raquelromano-gov (approved):

PR #5738, opened 2017-06-12
Extract fetch of data to another component

annekainicUSDS:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3237.

Fixes the issue where login wasn't blocked the fetch of data from the vets-api endpoint. Also cleans up a few styles.

### DIFF
--- a/src/js/post-911-gib-status/components/EnrollmentHistory.jsx
+++ b/src/js/post-911-gib-status/components/EnrollmentHistory.jsx
@@ -8,8 +8,7 @@ import { formatDateShort, formatDateLong } from '../utils/helpers';
 
 class EnrollmentHistory extends React.Component {
   render() {
-    const { enrollmentData } = this.props || {};
-    const today = formatDateShort(new Date());
+    const enrollmentData = this.props.enrollmentData || {};
 
     // TODO: find out when this warning should be shown.
     const expirationWarning = (
@@ -41,8 +40,6 @@ class EnrollmentHistory extends React.Component {
     return (
       <div>
         <h3 className="section-header">Enrollment History</h3>
-        {/* Find out if this line should be present if the blue box with same info is rendered */}
-        <div className="section-line">This information is current as of {today}</div>
         <InfoPair label="Total months received" value={enrollmentData.originalEntitlement}/>
         <InfoPair label="Used" value={enrollmentData.usedEntitlement}/>
         <InfoPair label="Remaining" value={enrollmentData.remainingEntitlement}/>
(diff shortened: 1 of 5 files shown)


src/js/post-911-gib-status/components/EnrollmentHistory.jsx
```diff
@@ -8,8 +8,7 @@ import { formatDateShort, formatDateLong } from '../utils/helpers';
 
 class EnrollmentHistory extends React.Component {
   render() {
-    const { enrollmentData } = this.props || {};
-    const today = formatDateShort(new Date());
+    const enrollmentData = this.props.enrollmentData || {};
```
annekainicUSDS: I don't think this was working the way it was before. `enrollmentData` was being set to `null` when `this.props.enrollmentData` was null, when it should have been set to `{}` in that case. This was creating problems where calls to properties within `enrollmentData` would fail (e.g., `enrollmentData.percentageBenefit`) and nothing would render. This change fixes that.

src/js/post-911-gib-status/components/EnrollmentHistory.jsx
```diff
@@ -41,8 +40,6 @@ class EnrollmentHistory extends React.Component {
     return (
       <div>
         <h3 className="section-header">Enrollment History</h3>
-        {/* Find out if this line should be present if the blue box with same info is rendered */}
```
annekainicUSDS: @phiden said that we can include the box on every box, so I don't think this is needed anymore.

src/js/post-911-gib-status/containers/Post911GIBStatusApp.jsx
```diff
@@ -58,14 +51,7 @@ class Post911GIBStatusApp extends React.Component {
...
               </div>
-              <UserInfoSection enrollmentData={enrollmentData}/>
-              <EnrollmentHistory enrollmentData={enrollmentData}/>
-              <div className="feature help-desk">
-                <h2>Need help?</h2>
-                <div>Call the Vets.gov Help Desk</div>
-                <div>1-855-574-7286</div>
-                <div>Monday - Friday, 8:00am - 8:00pm (ET)</div>
-              </div>
+              <Main/>
```
annekainicUSDS: I tried to make this a child route, so that I could call `{this.props.children}` here instead of explicitly rendering this particular component, but for some reason ran into problems when I tried to do it that way. This works for now.

src/js/post-911-gib-status/containers/Post911GIBStatusApp.jsx
```diff
@@ -31,15 +29,10 @@ function AppContent({ children, isDataAvailable }) {
 }
 
 class Post911GIBStatusApp extends React.Component {
-  componentDidMount() {
-    this.props.getEnrollmentData();
-  }
 
   render() {
     // TODO: change the service name below from "user-profile" to
```
annekainicUSDS: We should probably change this now that we have a service available for this app. Tried doing it quickly in this PR, but I think it's going to be a more extensive change, so saving that for a separate PR.

raquelromano-gov (approved):

PR #5740, opened 2017-06-12
Add redirect page after saving form

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3163

Adds the redirect after saving a form. I've left the info icon in until we get a definite answer on the style for that box.

screen shot 2017-06-13 at 8 58 00 am


src/js/common/schemaform/save-load-actions.js
```diff
@@ -135,29 +139,33 @@ export function saveInProgressForm(formId, version, returnUrl, formData) {
...
-      // Probably a network error has occurred
-      // TODO: Log this in GA or something
-      console.error('Error saving form:', error.message); // eslint-disable-line no-console
-      dispatch(setSaveFormStatus(SAVE_STATUSES.failure));
-      if (__BUILDTYPE__ === 'development') {
-        return Promise.reject(); // For unit testing
-      }
 
-      return; // eslint-disable-line consistent-return
+        Raven.captureException(new Error(`Error saving in progress form: ${resOrError.statusText}`));
```
jbalboni: Maybe we should come up with a string to put in here that we can add to any SIP errors and make them searchable in Sentry.

src/js/common/schemaform/save-load-actions.js
```diff
@@ -1,4 +1,7 @@
+import moment from 'moment';
+import Raven from 'raven-js';
 import environment from '../helpers/environment.js';
+import 'isomorphic-fetch';
```
cvalarida: Why?

test/common/schemaform/save-load-actions.unit.spec.js
```diff
@@ -27,7 +27,7 @@ const setup = () => {
     userToken: '123abc'
   };
   global.fetch = sinon.stub();
-  global.fetch.returns(Promise.resolve({ okay: true }));
```
cvalarida: Heh, woops! Good catch!

src/js/common/schemaform/save-load-actions.js
```diff
@@ -135,29 +139,33 @@ export function saveInProgressForm(formId, version, returnUrl, formData) {
...
         }
-      }
-      return Promise.resolve(); // For unit testing
-    }).catch((error) => {
-      // Probably a network error has occurred
-      // TODO: Log this in GA or something
-      console.error('Error saving form:', error.message); // eslint-disable-line no-console
-      dispatch(setSaveFormStatus(SAVE_STATUSES.failure));
-      if (__BUILDTYPE__ === 'development') {
-        return Promise.reject(); // For unit testing
```
cvalarida: If this isn't in here, some of the unit tests won't actually run properly. You won't get any failures, but the `expect()` statements in the `.catch()`es won't be run (unless I'm missing something with the changes you made in the unit tests).

src/js/common/schemaform/save-load-actions.js
cvalarida: Yeah, that'd be great. Would just `sip-error save-failure specific reason here` work? (I still haven't really looked at Sentry to see how the search works.)

src/js/hca/config/form.js
```diff
@@ -116,7 +116,7 @@ const formConfig = {
   trackingPrefix: 'hca-',
   formId: 'hca',
   version: 0,
-  disableSave: true,
+  // disableSave: true,
```
cvalarida: I think you missed uncommenting this.

test/common/schemaform/save-load-actions.unit.spec.js
```diff
@@ -185,10 +190,10 @@ describe('Schemaform save / load actions:', () => {
       global.fetch.returns(Promise.reject(new Error('No network connection')));
 
       thunk(dispatch).then(() => {
-        done(new Error("Should not call the dispatch's .then() on a network failure"));
```
cvalarida: Why remove this?

cvalarida (commented):
Sorry for not getting this reviewed earlier; I started and got distracted. Turns out I had lots of comments / questions!

src/js/common/schemaform/FormPage.jsx
```diff
@@ -53,6 +53,14 @@ class FormPage extends React.Component {
     focusForm();
   }
 
+  // A successful form save will mean we go from pending to success,
+  // and we need to redirect
+  componentWillReceiveProps(newProps) {
+    if (this.props.form.savedStatus === SAVE_STATUSES.pending && newProps.form.savedStatus === SAVE_STATUSES.success) {
+      this.props.router.push(`${newProps.urlPrefix || ''}form-saved`);
```
jbalboni: Yes, hca doesn't have one.

src/js/common/schemaform/FormSaved.jsx
```diff
@@ -0,0 +1,91 @@
...
+    duration: 500,
+    delay: 0,
+    smooth: true,
+  });
+};
+
+class FormSaved extends React.Component {
+  componentDidMount() {
+    scrollToTop();
+    focusForm();
```
jbalboni: Correct

src/js/common/schemaform/FormSaved.jsx
```diff
@@ -0,0 +1,91 @@
...
+
+  render() {
+    const lastSavedDate = this.props.lastSavedDate;
+
+    return (
+      <div className="schemaform-intro">
+        <div className="usa-alert usa-alert-info">
+          <div className="usa-alert-body">
+            <strong>Your application has been saved!</strong><br/>
+            {!!lastSavedDate && <p>Last saved on {moment(lastSavedDate).format('M/D/YYYY [at] h:mma')}.</p>}
```
jbalboni: Based on the comment in the other issue, I'll remove the seconds.

src/js/common/schemaform/save-load-actions.js
jbalboni: Because I'm checking if something is an instance of `Response`, and this adds that (and `Request`) as globals that I can use in the tests.

src/js/common/schemaform/save-load-actions.js
jbalboni: I've updated all the `saveInProgressForm` tests to put expect statements inside `then` instead of `catch` and verified that the `then` callback with `expect` statements is run for each test. 

The way the code works (outside of unit tests) is that we are handling errors inside a `catch`, and continuing on normally, so I think that's what we should be testing against. We can test that the appropriate statuses were dispatched in the `catch` callback without having to use `catch` in the unit tests.

If we keep these rejections behind a flag, we could end up missing an error that triggers a rejected promise we weren't expecting, because the promise is artificially being rejected just for tests.

src/js/common/schemaform/save-load-actions.js
jbalboni: Yeah, that would work. I did `vets_server_error` for some on the rjsf forms, so maybe `vets_sip_error`?

test/common/schemaform/save-load-actions.unit.spec.js
jbalboni: It doesn't really make sense after I took out the unit test specific promise rejections. We really care that the failure status is set (tested in the `then` here) and that we don't get an unhandled/unexpected rejection (tested by rejecting in the `catch` here).

src/js/common/schemaform/reducers/index.js
```diff
@@ -104,6 +104,7 @@ export default function createSchemaFormReducer(formConfig) {
       loadedStatus: LOAD_STATUSES.notAttempted,
       version: formConfig.version,
       formId: formConfig.formId,
+      lastSaveDate: null,
```
jbalboni: @cvalarida Does this make sense? Or should I update this in the metadata for loadedData? We won't always have data in loadedData after a save, though.

src/js/common/schemaform/FormPage.jsx
jbalboni: Actually, it does! Just a `/`. I had been assuming it's not required to have one, I don't want to change that now. Having to put `/` for hca seems hackish.

cvalarida: Weird. Alright. :man_shrugging:

src/js/common/schemaform/save-load-actions.js
cvalarida: Ahh, okay. Awesome!

src/js/common/schemaform/reducers/index.js
cvalarida: Yeah, that's alright. The only time it's used is in that save page which, as you said, is distinct from the loadedData. The loaded data will have `metadata.savedAt` for when it was saved (to be displayed on the intro page) and they'll be _basically_ the same thing, but....not.

src/js/common/schemaform/save-load-actions.js
```diff
@@ -99,12 +102,14 @@ export function migrateFormData(savedData, savedVersion, migrations) {
...
 export function saveInProgressForm(formId, version, returnUrl, formData) {
+  const savedAt = Date.now();
   // Double stringify because of api reasons. Olive Branch issues, methinks.
   // TODO: Stop double stringifying
   const body = JSON.stringify({
     metadata: JSON.stringify({
       version,
-      returnUrl
+      returnUrl,
+      savedAt
```
jbalboni: @cvalarida I'm stealing what you did in your PR, since it makes more sense for me to use this date than get something from the http response.

cvalarida (approved):
Excellent!

cehsu (approved):

cehsu (approved):

PR #5741, opened 2017-06-13
Fix schema commit

jbalboni:
This is causing issues on some machines, because the commit is in a deleted branch.


cvalarida (approved):
Thanks!

PR #5742, opened 2017-06-13
Made app loading dependent on fetch for folders and recipients.

U-DON:
Related to department-of-veterans-affairs/vets.gov-team#3057.

The idea is to show the "Click here to try again" link to provide an option for users to retry a failed request for recipients. Previously, we were disabling the compose button with the possibility of it never getting enabled if the request failed.


bshyong (approved):
LGTM

PR #5743, opened 2017-06-13
Clean up the hours on FL in cases where they look funny

aub:
Closes #department-of-veterans-affairs/vets.gov-team/issues/3177

This finds cases where the hours are listed as "-" and replaces that with "Closed" and cases where they are listed like "800am-430pm" and puts colons in the hours for those. So it'll look like this:

screen shot 2017-06-13 at 9 21 18 am

### DIFF
--- a/src/js/facility-locator/components/FacilityHours.jsx
+++ b/src/js/facility-locator/components/FacilityHours.jsx
@@ -31,15 +31,34 @@ export default class FacilityHours extends Component {
       return null;
     }
 
-    const hourRows = Object.keys(hours).map(h => {
-      if (h !== 'notes' && hours[h] && hours[h] !== '') {
+    function colonizeTime(time) {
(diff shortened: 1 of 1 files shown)


src/js/facility-locator/components/FacilityHours.jsx
```diff
@@ -31,15 +31,34 @@ export default class FacilityHours extends Component {
       return null;
     }
 
-    const hourRows = Object.keys(hours).map(h => {
-      if (h !== 'notes' && hours[h] && hours[h] !== '') {
+    function colonizeTime(time) {
```
U-DON: "Colonize". 😛

U-DON (approved):
LGTM. I think we plan to add more tests to FL at some point, but having them would be quite nice for any regex-y things like this.

bshyong (approved):
LGTM

PR #5744, opened 2017-06-13
Enrollment change history

raquelromano-gov:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3194

 - [ ] adds "Change History" (amendments)
 - [ ] updates keys that changed in the backend schema
 - [ ] gets rid of all-caps institution names
 
Collapsed:
image

Expanded (the margin for the `<hr>` used here and for the enrollment periods still needs a tweak):
image

### DIFF
--- a/src/js/post-911-gib-status/components/EnrollmentPeriod.jsx
+++ b/src/js/post-911-gib-status/components/EnrollmentPeriod.jsx
@@ -1,5 +1,3 @@
-import drop from 'lodash/fp/drop';
-
 import React from 'react';
 import PropTypes from 'prop-types';
 import Scroll from 'react-scroll';
@@ -16,7 +14,7 @@ class EnrollmentPeriod extends React.Component {
     super();
     this.scrollToEnrollment = this.scrollToEnrollment.bind(this);
     this.toggleHistory = this.toggleHistory.bind(this);
-    this.state = { historyOpen: false };
+    this.state = { historyExpanded: false };
   }
 
   scrollToEnrollment() {
@@ -25,9 +23,9 @@ class EnrollmentPeriod extends React.Component {
   }
 
   toggleHistory() {
-    const willOpen = !this.state.historyOpen;
-    this.setState({ historyOpen: !this.state.historyOpen });
-    if (willOpen) {
+    const shouldExpand = !this.state.historyExpanded;
+    this.setState({ historyExpanded: !this.state.historyExpanded });
+    if (shouldExpand) {
       this.scrollToEnrollment();
     }
   }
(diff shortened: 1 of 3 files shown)


src/js/post-911-gib-status/components/EnrollmentPeriod.jsx
```diff
@@ -1,35 +1,99 @@
...
-    const { enrollment } = this.props;
+    const { enrollment, id } = this.props;
+    const amendments = enrollment.amendments || [];
+    const firstAmendment = amendments.length > 0 && amendments[0];
+
+    // TODO: should fullTimeHours be used anywhere?
+    // const fullTimeHours = <InfoPair label="Full-time Hours" value={enrollment.fullTimeHours}/>;
+    const onCampusHours = <InfoPair label="On-campus Hours" value={enrollment.onCampusHours}/>;
+    const onlineHours = <InfoPair label="Online Hours" value={enrollment.onlineHours}/>;
+    const typeOfChange = firstAmendment &&
```
raquelromano-gov: I am beginning to think that these 2 things don't belong here because
  * the first amendment has its own `onCampusHours` and `onlineHours` values which are being ignored here. Unless the first amendment overrides the original enrollment's hours? Design & product question so I'll add it to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3250

src/js/post-911-gib-status/components/EnrollmentPeriod.jsx
```diff
@@ -1,35 +1,99 @@
...
-      <InfoPair label="Type of Change" value={enrollment.amendmentList[0].type}/>;
-    const changeEffectiveDate = enrollment.amendmentList && enrollment.amendmentList.length > 0 &&
-      <InfoPair label="Change Effective Date" value={enrollment.amendmentList[0].changeEffectiveDate}/>;
+    const moreAmendments = amendments.length > 0 && drop(1, amendments);
+    const changeHistory = moreAmendments.length > 0 && (
+      <div className="usa-accordion-bordered">
+        <ul className="usa-unstyled-list">
+          <li>
+            <div id={`collapsible-${id}`} className="usa-accordion-content">
+            {/* <div id={`collapsible-${id}`} className="accordion-header clearfix"> */}
```
raquelromano-gov: Styling here isn't right yet...just getting the data piped properly.

src/js/post-911-gib-status/components/EnrollmentPeriod.jsx
```diff
@@ -1,35 +1,95 @@
...
-    const typeOfChange = enrollment.amendmentList && enrollment.amendmentList.length > 0 &&
-      <InfoPair label="Type of Change" value={enrollment.amendmentList[0].type}/>;
-    const changeEffectiveDate = enrollment.amendmentList && enrollment.amendmentList.length > 0 &&
-      <InfoPair label="Change Effective Date" value={enrollment.amendmentList[0].changeEffectiveDate}/>;
+    const amendmentList = this.state.historyExpanded ? (
+      <div id={`collapsible-${id}`} className="usa-accordion-content">
+        {amendments.map((amendment, index) => {
+          return (
+            <div key={`amendment-${this.props.key}-${index}`}>
+              {index > 0 ? <hr/> : null}
```
annekainicUSDS: I styled `hr` to have no margin on the bottom for this page, because it was creating styling issues elsewhere. But with the amendment list some margin bottom might be needed because it looks a little funny in the screenshot. What I might recommend is reversing the change on `hr` specifically to not have `margin-bottom: 0;`, and instead collapse the top margin of the next sibling element (specifically, we want to do this where we have a `hr` followed by a `h4`. So maybe something like this:

```
hr {
  border-style: solid;
  border-color: $color-gray-light;
  border-width: 1px 0 0 0;
  
  & + h4 {
    margin-top: 0;
  }
}
```

I think something like this will work, but not able to test right now because the data isn't populating when I run this locally. Let me know if this doesn't make sense!

annekainicUSDS (commented):
Looks good! Having trouble testing it because the mock data isn't populating for me right now. Will try to figure out why so I can see what this looks like. A few comments.

src/js/post-911-gib-status/components/EnrollmentPeriod.jsx
```diff
@@ -1,35 +1,95 @@
...
+      this.scrollToEnrollment();
+    }
+  }
+
   render() {
-    const { enrollment } = this.props;
+    const { enrollment, id } = this.props;
+    const amendments = enrollment.amendments || [];
+
+    // TODO: should fullTimeHours be used anywhere?
```
phiden: @annekainicUSDS Good question, that I unfortunately don't have a concrete answer to. If it's different from on/off-campus hours, then I would guess we do _not_ have to use it -- eBenefits does not. 

@rroueche can we add this question to our next check-in? I don't think it's worth an e-mail at this stage, but would like to confirm that we don't need to use that piece of data.

src/js/post-911-gib-status/components/EnrollmentPeriod.jsx
```diff
@@ -1,35 +1,95 @@
...
-    const distanceHours = <InfoPair label="Distance Hours" value={enrollment.distanceHours}/>;
-    const typeOfChange = enrollment.amendmentList && enrollment.amendmentList.length > 0 &&
-      <InfoPair label="Type of Change" value={enrollment.amendmentList[0].type}/>;
-    const changeEffectiveDate = enrollment.amendmentList && enrollment.amendmentList.length > 0 &&
-      <InfoPair label="Change Effective Date" value={enrollment.amendmentList[0].changeEffectiveDate}/>;
+    const amendmentList = this.state.historyExpanded ? (
+      <div id={`collapsible-${id}`} className="usa-accordion-content">
+        {amendments.map((amendment, index) => {
+          return (
+            <div key={`amendment-${this.props.key}-${index}`}>
```
raquelromano-gov: Oops, this is no longer needed. I had originally thought each key needed to be unique across the whole document, but it only needs to be unique among its siblings, so this will just be `amendment-${index}`

src/js/post-911-gib-status/components/EnrollmentPeriod.jsx
raquelromano-gov: Thanks! Will take a stab at it.

annekainicUSDS:
Tested this locally and it looks good! The only thing I noticed from testing is that the scrolling doesn't appear to be working correctly. Getting a console error about `key` not being a prop.

src/js/post-911-gib-status/components/EnrollmentPeriod.jsx
annekainicUSDS: I actually think you can just remove the `margin-bottom: 0;` on `hr` without adding additional styles. I think it looks fine without adjusting the default margin on `hr`. Once we do a thorough design review of the whole app they can let us know if they want that adjusted :).

src/js/post-911-gib-status/components/EnrollmentPeriod.jsx
rroueche: yes

raquelromano-gov:
Made some style changes that look better:
image

src/js/post-911-gib-status/components/EnrollmentPeriod.jsx
raquelromano-gov: talked to @phiden and this the training type is not shown anyhwere in e-benefits so yes, it is intentionally excluced here.

annekainicUSDS (approved):
LGTM!

PR #5746, opened 2017-06-13
Save in Progress -- Save and show savedAt timestamp

cvalarida:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3197

Looks like https://marvelapp.com/2hj59b1/screen/28358733

Not logged in:
image

Logged in with saved form:
image

Logged in with no saved form:
image


cehsu:
Do we want to include seconds in the timestamp?
screen shot 2017-06-13 at 9 57 42 am

cvalarida:
Heh, I actually asked @stacyhlinn the same thing in @jbalboni's PR for the save redirect. I'll ask it again here!

What sayest thou, @stacyhlinn? To seconds or not to seconds? That is the question.

stacyhlinn:
@cvalarida @cehsu Good question! 
Showing the seconds is probably not that valuable for the users, so let's just go with hour:minute am/pm. Thanks guys!

ayaleloehr:
Pedantic questions: @stacyhlinn what time zone will that time be in and should we put that on there? (those are both actual questions I don't know the answers to, not leading questions!)

stacyhlinn:
@ayaleloehr Thanks for bringing this up. And yes it should show the time in whichever timezone the users are in.  :)

jbalboni:
@cvalarida You should still include a screenshot, even with the link to the marvel app. And also connect it to your GH issue.

src/sass/base/_b-utils.scss
```diff
@@ -1,3 +1,7 @@
 .full-width {
   width: 100%;
 }
+
+.no-background-image {
```
jbalboni: Won't there still be left side padding here? I think making an alert specific no icon class would be good (i.e. a `usa-alert-no-icon` class that fixed the padding and hide the bg image).

jbalboni (approved):
Looks good to me. Can we add some tests for the intro page changes?

cvalarida:
@jbalboni No, the padding comes from an inner `div.usa-alert-body`.

I'm having trouble getting this to work again (hoping just because of a change in the api), so I'm going to wait to get a screenshot out until [this PR](https://github.com/department-of-veterans-affairs/vets-website/pull/5762) is merged. I'll then merge master into here and :crossed_fingers: hope it works again.

src/sass/base/_b-utils.scss
cvalarida: No, the padding comes from an inner `div.usa-alert-body`.

cehsu (approved):

PR #5749, opened 2017-06-14
Fix FL logic for displaying wait times

bshyong:


U-DON (approved):
Don't really have much context on this, but looks correct that not having specialty wait times should prevent rendering those sections instead of the whole thing.

U-DON:
There's a conflict after #5751 got merged. Note that `pull` can take multiple args and remove multiple elements in one call.

Also, the render functions within the overall render seem like they could be pulled out as methods at some point. Anything that would cause the component to re-render would recreate those functions on each cycle.

patrickvinograd:
Closing in favor of #5758 due to some changes that have gone in in the meantime.

PR #5756, opened 2017-06-14
Add expired session handling for save in progress

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3163

This handles the case where a user's session expires while working on the form (which I am testing by logging in, going to the form, then changing the userToken in session storage, so I get a 401 when saving).

screen shot 2017-06-14 at 12 21 58 pm

Also updated the normal link to match prototype:
screen shot 2017-06-14 at 12 21 44 pm


src/js/login/reducers/login.js
```diff
@@ -5,11 +5,10 @@ import {
   UPDATE_LOGIN_URL,
   UPDATE_VERIFY_URL,
   UPDATE_LOGOUT_URL,
-  UPDATE_SEARCH_HELP_USER_MENU
+  UPDATE_SEARCH_HELP_USER_MENU,
+  LOG_OUT
```
jbalboni: @raquelromano-gov It looks like `LOG_OUT` was not used or hooked up correctly?

src/js/common/schemaform/SaveFormLink.jsx
```diff
@@ -28,35 +28,34 @@ class SaveFormLink extends React.Component {
...
 
-    if (savedStatus === SAVE_STATUSES.noAuth) {
-      content = <p>no-auth message</p>;
-    } else if (savedStatus === SAVE_STATUSES.failure) {
-      content = <p>failure message</p>;
+    if (savedStatus === SAVE_STATUSES.failure) {
+      content = <span>failure message</span>;
     } else if (savedStatus === SAVE_STATUSES.pending) {
-      content = <p>spinner or something</p>;
+      content = <span>spinner or something</span>;
```
jbalboni: The jumpiness of switching between content with and without a surrounding `<p>` tag was annoying, though I realize we'll likely replace these messages.

jbalboni:
@stacyhlinn Does that message look ok to you? It will show up if you click the normal save link and you need to log in again.

stacyhlinn:
@jbalboni  Yes message looks good. I would just hyperlink "sign in" instead of the whole sentence "Please sign in again" to be consistent with other messaging.

src/js/common/schemaform/SaveFormLink.jsx
```diff
@@ -28,35 +28,34 @@ class SaveFormLink extends React.Component {
...
 
-    if (this.props.user.login.currentlyLoggedIn) {
-      content = <a onClick={saveForm}>Save and come back later</a>;
+    if (!this.props.user.login.currentlyLoggedIn) {
+      // if we have a noAuth status, that means they tried to save and got a 401, which
+      // likely means their session is expired (since they were logged in before)
+      content = (<div>
+        {savedStatus === SAVE_STATUSES.noAuth
+            ? <span>Sorry, your session has expired. <a onClick={this.openLoginModal}>Please sign in again</a></span>
+            : <a onClick={this.openLoginModal}>Sign in before saving your application</a>}
```
cvalarida: @stacyhlinn What do you think about having just "Sign in" linked here too? It would make it consistent with the intro page and session expiration (once that's updated per your comment).

Actually, in your design, this just said "Save and come back later" both when the user is and is not logged in. Is that what we want to go with?

cvalarida (approved):
Woot!

src/js/common/schemaform/SaveFormLink.jsx
```diff
@@ -28,35 +28,34 @@ class SaveFormLink extends React.Component {
...
+    let content = <a onClick={saveForm}>Save and finish later</a>;
 
-    if (this.props.user.login.currentlyLoggedIn) {
-      content = <a onClick={saveForm}>Save and come back later</a>;
+    if (!this.props.user.login.currentlyLoggedIn) {
+      // if we have a noAuth status, that means they tried to save and got a 401, which
+      // likely means their session is expired (since they were logged in before)
+      content = (<div>
+        {savedStatus === SAVE_STATUSES.noAuth
+            ? <span>Sorry, your session has expired. Please <a onClick={this.openLoginModal}>sign in</a> again</span>
```
cvalarida: Missing `.` at the end. Just noticed.

src/js/common/schemaform/SaveFormLink.jsx
jbalboni: @cvalarida I thought I pushed that change, but I did not! Sign in link is updated now.

Using the same "save and finish later" link for both logged in and out users is tricky, so I didn't try to do that in this PR.

src/js/common/schemaform/SaveFormLink.jsx
jbalboni: I wasn't sure if we wanted one there or not, so I left it off.

src/js/common/schemaform/SaveFormLink.jsx
cvalarida: Yeah, it doesn't need to be the same one, but the design just had them saying the same thing.

cehsu (approved):

src/js/login/reducers/login.js
raquelromano-gov: I have the vague recollection that this needed to be shared across apps at some point but it looks like that's no longer the case.

PR #5758, opened 2017-06-14
Gracefully handle no data e.g. due to running against old API

patrickvinograd:
Replaces #5749. Allows facility detail page to render even if running against old API. Tested locally by downgrading my local vets-api.


bshyong (approved):
LGTM

PR #5759, opened 2017-06-14
GIB print page

annekainicUSDS:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3169

Print page! The way I put this together is a little hacky, I'm afraid. I started by trying to create a new layout for print pages, one that wouldn't include the header, footer, etc. The problem I encountered with this approach is I wasn't sure how to render different layouts within the same React app. So instead I added styles that creates a fixed position container that overlays the regular Vets.gov layout, so all you see is the white page with the letter content. Not sure if this is the best way to go about it. Would love feedback from @raquelromano-gov and @jbalboni in particular on this, and any other front-end devs who take a look!

screen shot 2017-06-15 at 9 49 12 am


src/js/post-911-gib-status/components/UserInfoSection.jsx
```diff
@@ -19,7 +19,6 @@ class UserInfoSection extends React.Component {
 
     return (
       <div>
-        <h3 className="section-header">Chapter 33 Benefit Information</h3>
```
annekainicUSDS: Removed this because the latest mocks don't have this header anymore.

src/js/post-911-gib-status/routes.jsx
```diff
@@ -1,8 +1,13 @@
...
+import React from 'react';
+import { Route } from 'react-router';
 
-const routes = {
-  path: '/',
-  component: Post911GIBStatusApp
-};
+import PrintPage from './containers/PrintPage';
+
+const routes = [
```
annekainicUSDS: The only way I knew how to add a separate route for the print page was to change how we're handling our routes. In doing so, I tried to add a route for the `Main` component, where the path would just be `''`, but still having trouble getting that to work. So I'm still just rendering the `Main` component explicitly in the top level component. I think this would be a good topic for the next front end eng sync, since I'm pretty confused on how these routes work.

phiden:
@annekainicUSDS you should be using the marvelapp version, please. sorry for any confusion!

annekainicUSDS:
Gotcha, will make the changes!

src/sass/post-911-gib-status.scss
```diff
@@ -32,3 +32,28 @@ hr {
   	margin-top: 2em;
   }
 }
+
+.print-status {
```
jbalboni: I think this is a reasonable approach given that we can't have different static page layouts within the same React app. Another approach I've taken in the past is to create a class that can be added/removed on the `html` element that sets the header and footer inside it to `display: none`. I don't think that's better than what you have, but it's another option if this has unexpected issues.

jbalboni (approved):

src/sass/post-911-gib-status.scss
annekainicUSDS: Awesome, thanks! If we find any issues in testing I will try that other approach.

src/js/post-911-gib-status/routes.jsx
raquelromano-gov: Just browsed the above IndexRoute docs as well as [RouteConfiguration](https://github.com/ReactTraining/react-router/blob/v3/docs/guides/RouteConfiguration.md) and `IndexRoute` seems like the way to go.

annekainicUSDS:
Currently running into a few problems:
1. `<InfoPair>`s aren't being rendered in line in the print version. I think I need to specify separate styles for print.
2. This looks super bad on mobile. Fix that.
3. Using `<IndexRoute>` was resulting in a console error around children having separate keys. Fix that too.
4. When going to print, it's printing 3 identical sheets (?!?!?!). Fix that, but first figure out why this is happening.

annekainicUSDS:
@raquelromano-gov this is ready for review now!

src/js/post-911-gib-status/components/UserInfoSection.jsx
```diff
@@ -14,17 +14,20 @@ class UserInfoSection extends React.Component {
     const percentageBenefit = formatPercent(enrollmentData.percentageBenefit) || 'unavailable';
     const fullName = `${enrollmentData.firstName} ${enrollmentData.lastName}`;
 
-    // TODO: Figure out what will be sent to us if the user is ineligible so we can
-    // conditionally show "Currently Disallowed" message
-
-    return (
-      <div>
-        <h3 className="section-header">Chapter 33 Benefit Information</h3>
+    let currentOfAlert;
```
raquelromano-gov: Can we name these `currentAsOfAlert` and `showCurrentAsOfAlert`? Still a little awkward but I think more self-describing.

src/js/post-911-gib-status/containers/StatusPage.jsx
```diff
@@ -0,0 +1,60 @@
...
+import EnrollmentHistory from '../components/EnrollmentHistory';
+import UserInfoSection from '../components/UserInfoSection';
+
+class StatusPage extends React.Component {
+  componentDidMount() {
+    this.props.getEnrollmentData();
+  }
+
+  render() {
+    // TODO: change the service name below from "user-profile" to
```
raquelromano-gov: This TODO got pasted in here by mistake. Related, I filed https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3329 for us to update that service name

src/js/post-911-gib-status/routes.jsx
```diff
@@ -1,8 +1,22 @@
...
-const routes = {
-  path: '/',
-  component: Post911GIBStatusApp
-};
+import Main from './containers/Main';
+import PrintPage from './containers/PrintPage';
+import StatusPage from './containers/StatusPage';
+
+const routes = [
+  <Route
```
raquelromano-gov: This looks nice and clean.

raquelromano-gov (approved):

src/js/post-911-gib-status/containers/StatusPage.jsx
```diff
@@ -0,0 +1,60 @@
...
+            in your Certificate of Eligibility (COE) letter. In lieu of a COE letter, you can
+            print a copy of this screen for benefit and eligibility verification.
+          </p>
+        </div>
+        <div className="section-line">
+          <Link to="/print" target="_blank" className="usa-button-primary">
+            Print This Page as COE
+          </Link>
+        </div>
+        <UserInfoSection enrollmentData={enrollmentData} showCurrentOfAlert/>
```
raquelromano-gov: Sorry, one more question: where does this value come from?

src/js/post-911-gib-status/containers/StatusPage.jsx
annekainicUSDS: Yes, will fix this

src/js/post-911-gib-status/containers/StatusPage.jsx
annekainicUSDS: It doesn't come from anywhere; `showCurrentOfAlert` is just a prop passed to `UserInfoSection`. If the prop is passed, the value is `true`; if it's not passed, the value is `false`. So by including this prop, the value of `showCurrentOfAlert` once within `UserInfoSection` is `true`. I originally had `showCurrentOfAlert={true}`, but then React warned me that was unnecessary :)

src/js/post-911-gib-status/components/UserInfoSection.jsx
annekainicUSDS: Yes! that is definitely more clear, will make that change

src/js/post-911-gib-status/containers/StatusPage.jsx
raquelromano-gov: OK, got it. So we always want to show this on StatusPage. Thanks!

PR #5762, opened 2017-06-14
Save in Progress -- Use savedForms List

cvalarida:
We can now get a list of all the forms saved by a user when they log in. This makes things much cleaner (though the the prop handling still could use help, admittedly).

If I could also get @raquelromano-gov or @omgitsbillryan to take a look at login-helpers.js to make sure I'm not inadvertently doing a terrible thing, that'd be swell. 👌


src/js/common/helpers/login-helpers.js
```diff
@@ -27,6 +27,7 @@ export function getUserData(dispatch) {
       if (userData.first_name) {
         sessionStorage.setItem('userFirstName', userData.first_name);
       }
+      dispatch(updateProfileField('savedForms', json.data.attributes.in_progress_forms));
```
jbalboni: This list of dispatches makes me sad. It could be one dispatch call.

(Not that you need to fix that here).

src/js/common/schemaform/FormIntroButtons.jsx
```diff
@@ -4,66 +4,34 @@ import PropTypes from 'prop-types';
...
-    this.props.loadInProgressDataIntoForm();
-    // Navigate to the last page they were on
-    this.props.router.push(this.props.form.loadedData.metadata.returnUrl);
-    // TODO: Handle this scenario:
-    //  1) I fill out some information and save my progress.
-    //  2) The form is updated and a field I've not filled out yet gets moved
-    //     to a page I have already completed.
-    //  3) I load my saved progress.
-    //  4) I should be put in the page with the missing information.
+    this.props.fetchInProgressForm(this.props.formId, this.props.migrations).then((returnUrl) => {
```
jbalboni: I think it would be better to handle this in a lifecycle method, rather than using the promise. We're setting the correct statuses already and we probably want to make `fetchInProgressForm` work like `saveInProgressForm`, where we don't re-reject in the `catch` callback (in my outstanding PR). Plus, you should have the updated props with the metadata in that case.

src/js/common/helpers/login-helpers.js
cvalarida: Should we make a ticket out of it?

raquelromano-gov: Make Jeff happy.

raquelromano-gov: TBH, I intentionally ignored everything after `handleVerify` when I touched this file. This part looked scary and made me think our profile code could use some refactoring.

cvalarida: https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3311

src/js/common/schemaform/FormIntroButtons.jsx
```diff
@@ -4,40 +4,17 @@ import PropTypes from 'prop-types';
...
-    try {
-      isLoggedIn = newProps.loggedIn;
-    } catch (e) {
-      if (e instanceof TypeError) {
-        isLoggedIn = false;
-      }
-    }
-    if (!wasLoggedIn && isLoggedIn) {
-      this.props.fetchInProgressForm(this.props.form.formId, this.props.form.migrations);
+  componentWillReceiveProps = (newProps) => {
```
cehsu: Is this the appropriate lifecycle method? Or should `shouldComponentUpdate()` be used here?

cvalarida: No, I don't think that'd be the right lifecycle method. That's for determining whether or not the component should exit the update lifecycle to avoid pointless re-renders ([citation](https://developmentarc.gitbooks.io/react-indepth/content/life_cycle/update/using_should_component_update.html)). I think it'd be improper to have side-effects in `shouldComponentUpdate`.

src/js/hca/config/form.js
```diff
@@ -25,6 +25,7 @@ import {
 } from '../helpers';
 
 import IntroductionPage from '../components/IntroductionPage';
+// import IntroductionPage from '../components/SIPIntroductionPage';
```
cehsu: Is the plan to hide this behind a feature flag?

cvalarida: Yeah, when we get all these PRs merged into master, I'll put it behind a feature flag so we can test it in staging without putting it out on production.

cvalarida: I didn't realize I added this in this commit...But yeah, it's supposed to be `IntroductionPage` and not `SIPIntroductionPage` for now.

cehsu (approved):
LGTM. Thanks for already answering questions.

src/js/common/schemaform/FormIntroButtons.jsx
cehsu: Thanks @cvalarida, you're totally right. Good point. I'm wondering why this is being performed in the child component, rather than in when the returnURL is being set, if the check and routing could be performed there in `mapStateToProps`.

jbalboni: `mapStateToProps` is a selector, and shouldn't have side effects either. `returnUrl` is set in the action creator, so you could do the routing in there. That might actually be clearer, though you do have to pass the router to the action creator, which feels strange to me (though I have done that before in apps).

cvalarida: I think the answer to this is moving the routing to the wrapper `SIPIntroductionPage` when we make it. Not the current one, but a HOC we can wrap our `IntroductionPage`s with that handles all the save in progress logic. Making a ticket for it now...

cvalarida: https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3331

PR #5766, opened 2017-06-14
Show any appointments data available; dont use primary care as indicator

patrickvinograd:
Fixes https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3313

Determine whether to display overall component and new/established subsections based on whether any data is present rather than keying off of primary care. 

With this fix, we get just the existing patient wait times for a facility with no data for any new patient wait times (http://localhost:3001/facilities/facility/vha_635GF): 
screen shot 2017-06-14 at 4 19 41 pm


bshyong (approved):
LGTM

PR #5769, opened 2017-06-15
Skeletons for GIBS unit tests and actual InfoPair test

raquelromano-gov:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3206

### DIFF
--- a/src/js/post-911-gib-status/components/EnrollmentPeriod.jsx
+++ b/src/js/post-911-gib-status/components/EnrollmentPeriod.jsx
@@ -33,13 +33,10 @@ class EnrollmentPeriod extends React.Component {
   render() {
     const { enrollment, id } = this.props;
     const amendments = enrollment.amendments || [];
-
-    // TODO: should fullTimeHours be used anywhere?
-    // const fullTimeHours = <InfoPair label="Full-time Hours" value={enrollment.fullTimeHours}/>;
     const onCampusHours = <InfoPair label="On-campus Hours" value={enrollment.onCampusHours}/>;
     const onlineHours = <InfoPair label="Online Hours" value={enrollment.onlineHours}/>;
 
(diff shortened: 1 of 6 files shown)


src/js/post-911-gib-status/components/EnrollmentPeriod.jsx
```diff
@@ -79,7 +76,7 @@ class EnrollmentPeriod extends React.Component {
     return (
       <div>
         <hr/>
-        <h4>{formatDateShort(enrollment.beginDate)} to {formatDateShort(enrollment.endDate)} at <span className="facility">{enrollment.facilityName.toLowerCase()}</span> ({enrollment.facilityCode})</h4>
+        <h4>{formatDateShort(enrollment.beginDate)} to {formatDateShort(enrollment.endDate)} at <span className="facility">{(enrollment.facilityName || '').toLowerCase()}</span> ({enrollment.facilityCode})</h4>
```
annekainicUSDS: Something I've been wondering about is if we should be checking if data exists before rendering anything. I need to double check the swagger yaml, to see which fields are optional from EVSS, because it would be weird if we didn't have `facilityName` or `facilityCode` for some reason, but we're still rendering this whole line. So it would be `06/15/17 to 6/15/18 at` for this line. Should we be checking if all these fields exist first? Is there a clean way of doing that?

annekainicUSDS (approved):
LGTM! Just one general question I've been wondering about.

src/js/post-911-gib-status/components/EnrollmentPeriod.jsx
raquelromano-gov: Yeah, I was thinking the same thing. I'll open a ticket about that and try to enumerate the cases when an empty value would require a graceful fallback. Will ping it to you and then we can ask for design feedback if necessary.

PR #5775, opened 2017-06-15
530 claimant contact confirmation email

cehsu:
Connects to department-of-veterans-affairs/vets.gov-team#3152
screen shot 2017-06-15 at 5 09 58 pm


yarn.lock
```diff
@@ -8300,9 +8300,9 @@ verror@1.3.6:
   dependencies:
     extsprintf "1.0.2"
 
-"vets-json-schema@https://github.com/department-of-veterans-affairs/vets-json-schema.git#175be45a2005e1d2ade7c43419d3dd0ca82d6c07":
-  version "3.0.3"
-  resolved "https://github.com/department-of-veterans-affairs/vets-json-schema.git#175be45a2005e1d2ade7c43419d3dd0ca82d6c07"
+"vets-json-schema@https://github.com/department-of-veterans-affairs/vets-json-schema.git#b4c9f42f7f1182952d59bb23db5867ed1eb2c848":
```
cehsu: This is not the merged commit hash.

cehsu (commented):
Must update commit hash for vets-json-schema dependency.

src/js/burials/config/form.js
```diff
@@ -153,18 +155,29 @@ const formConfig = {
           title: 'Claimant Contact Information',
           path: 'claimant-contact-information',
           uiSchema: {
+            'ui:validation': validateMatch('claimantEmail', 'view:claimantConfirmationEmail'),
             claimantAddress: address.uiSchema('Address'),
             claimantEmail: {
-              'ui:title': 'Email address'
+              'ui:title': 'Email address',
+              'ui:errorMessages': {
+                pattern: 'Please put your email in this format x@x.xxx'
```
jbalboni: Does there need to be a colon after format?

src/js/burials/config/form.js
```diff
@@ -153,18 +155,29 @@ const formConfig = {
           title: 'Claimant Contact Information',
           path: 'claimant-contact-information',
           uiSchema: {
+            'ui:validation': validateMatch('claimantEmail', 'view:claimantConfirmationEmail'),
```
jbalboni: I think we could add a test to make sure there's a validation error when these don't match. To keep us from forgetting to change the property names here later if we make a change.

jbalboni (approved):
Looks good, had one minor question and I think it needs one more test.

cehsu:
Great idea @jbalboni, test added.

PR #5777, opened 2017-06-15
Use the same text regardless of login state

cvalarida:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3328

~Doesn't save the form upon login yet (didn't see that it should until just now). I'll circle back around for that...~ Done!

### DIFF
--- a/src/js/common/components/LoginModal.jsx
+++ b/src/js/common/components/LoginModal.jsx
@@ -57,7 +57,12 @@ class LoginModal extends React.Component {
     const loginUrl = this.props.user.login.loginUrl;
     const onUpdateLoginUrl = this.props.onUpdateLoginUrl;
     this.loginUrlRequest = handleLogin(loginUrl, onUpdateLoginUrl);
-    this.loginUrlRequest.then(() => this.props.onClose());
+    this.loginUrlRequest.then(() => {
+      this.props.onClose();
+      if (this.props.onLogin) {
+        this.props.onLogin();
+      }
+    });
   }
 
   render() {
@@ -76,7 +81,8 @@ LoginModal.propTypes = {
   onClose: PropTypes.func.isRequired,
   title: PropTypes.string,
   user: PropTypes.object.isRequired, // Taken from the redux store
-  onUpdateLoginUrl: PropTypes.func.isRequired // Dispatches updateLogInUrl()
+  onUpdateLoginUrl: PropTypes.func.isRequired, // Dispatches updateLogInUrl()
+  onLogin: PropTypes.func
 };
 
 
--- a/src/js/common/components/LoginModal.jsx
+++ b/src/js/common/components/LoginModal.jsx
@@ -11,15 +11,33 @@ class LoginModal extends React.Component {
   constructor(props) {
     super(props);
     this.handleLogin = this.handleLogin.bind(this);
+
+    // I don't like this, but we need to make sure componentWillReceiveProps
+    //  doesn't call onLogin() when a page with a LoginModal is refreshed and
+    //  a user is logged in.
+    this.state = {
+      loginButtonClicked: false
+    };
   }
 
   componentWillReceiveProps(nextProps) {
     // If the loggedIn status went from false to true, close the modal
     const wasLoggedIn = this.props.user.login.currentlyLoggedIn;
     const isLoggedIn = nextProps.user.login.currentlyLoggedIn;
-    if (!wasLoggedIn && isLoggedIn) {
+    if (!wasLoggedIn && isLoggedIn && this.state.loginButtonClicked) {
+      if (this.props.onLogin) {
+        this.props.onLogin();
+      }
+      this.setState({ loginButtonClicked: false });
       this.props.onClose();
     }
+  }
+
+  shouldComponentUpdate(nextProps) {
+    const wasLoggedIn = this.props.user.login.currentlyLoggedIn;
+    const isLoggedIn = nextProps.user.login.currentlyLoggedIn;
+
+    return (wasLoggedIn !== isLoggedIn) || (this.props.visible !== nextProps.visible);
   }
 
   // Copied from src/js/login/containers/Main.jsx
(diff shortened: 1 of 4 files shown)


test/common/components/LoginModal.unit.spec.jsx
```diff
@@ -16,13 +16,15 @@ const response = {
...
   }
 };
+let fetchPromise;
 
 const setup = () => {
   oldFetch = global.fetch;
   oldWindow = global.window;
+  fetchPromise = Promise.resolve(response); // Reset it every time.
 
   global.fetch = sinon.spy(() => {
```
cehsu: Where is this being used?

src/js/common/components/LoginModal.jsx
```diff
@@ -57,7 +57,12 @@ class LoginModal extends React.Component {
     const loginUrl = this.props.user.login.loginUrl;
     const onUpdateLoginUrl = this.props.onUpdateLoginUrl;
     this.loginUrlRequest = handleLogin(loginUrl, onUpdateLoginUrl);
-    this.loginUrlRequest.then(() => this.props.onClose());
+    this.loginUrlRequest.then(() => {
```
cehsu: if we return this promise, we should be able to test like so:
```
    var ev = {preventDefault: function(){console.log('howdy')}};
    tree.handleLogin(ev).then(() => {
      expect(loginSpy).to.have.been.called.once
    });
```

cvalarida: Yeah, that's more what I'm looking for, but I'd rather not return the promise from `handleLogin()` if I don't have to. If the test I wrote fails either consistently or even just spottily (spottily...spottily...spell checker says it's a word!), I'll rewrite it like that.

Also, I think I may start using "howdy" for debugging now. 😄

cehsu (approved):
LGTM. Thank you for clearly explaining thought process in comments and identifying potential areas for considering alternate approaches down the road. A+ for maintainability.

src/js/common/components/LoginModal.jsx
```diff
@@ -11,17 +11,35 @@ class LoginModal extends React.Component {
...
   componentWillReceiveProps(nextProps) {
     // If the loggedIn status went from false to true, close the modal
     const wasLoggedIn = this.props.user.login.currentlyLoggedIn;
     const isLoggedIn = nextProps.user.login.currentlyLoggedIn;
-    if (!wasLoggedIn && isLoggedIn) {
+    if (!wasLoggedIn && isLoggedIn && this.state.loginButtonClicked) {
+      if (this.props.onLogin) {
+        this.props.onLogin();
+      }
+      this.setState({ loginButtonClicked: false });
```
cehsu: Would it make sense to call `onClose` in the setState callback?

src/js/common/components/LoginModal.jsx
```diff
@@ -54,10 +72,10 @@ class LoginModal extends React.Component {
 
   handleLogin(e) {
     e.preventDefault(); // Don't try to submit the page
+    this.setState({ loginButtonClicked: true });
     const loginUrl = this.props.user.login.loginUrl;
```
cehsu: could be shortened to:
const { user.login.loginUrl: loginUrl, onUpdateLoginUrl } = this.props;

src/js/common/components/LoginModal.jsx
```diff
@@ -11,17 +11,35 @@ class LoginModal extends React.Component {
...
+      this.setState({ loginButtonClicked: false });
       this.props.onClose();
     }
   }
 
+  shouldComponentUpdate(nextProps) {
+    const wasLoggedIn = this.props.user.login.currentlyLoggedIn;
+    const isLoggedIn = nextProps.user.login.currentlyLoggedIn;
+
+    return (wasLoggedIn !== isLoggedIn) || (this.props.visible !== nextProps.visible);
```
jbalboni: Isn't this going to skip updating when `loginButtonClicked` changes?

src/js/common/components/LoginModal.jsx
jbalboni: It doesn't look like loginButtonClicked is ever used in rendering, so I think it should be an instance property instead (`this.loginButtonClicked = false`). Then we don't need to deal a `setState` callback.

src/js/common/components/LoginModal.jsx
```diff
@@ -11,17 +11,35 @@ class LoginModal extends React.Component {
   constructor(props) {
     super(props);
     this.handleLogin = this.handleLogin.bind(this);
+
+    // I don't like this, but we need to make sure componentWillReceiveProps
+    //  doesn't call onLogin() when a page with a LoginModal is refreshed and
+    //  a user is logged in.
+    this.state = {
```
jbalboni: I mentioned this below, but since it doesn't look like this is used in rendering, I think you can make it an instance property: `this.loginButtonClicked = false`.

src/js/common/components/LoginModal.jsx
```diff
@@ -11,17 +11,35 @@ class LoginModal extends React.Component {
...
+    if (!wasLoggedIn && isLoggedIn && this.state.loginButtonClicked) {
+      if (this.props.onLogin) {
+        this.props.onLogin();
+      }
+      this.setState({ loginButtonClicked: false });
       this.props.onClose();
     }
   }
 
+  shouldComponentUpdate(nextProps) {
```
jbalboni: This is fine, but it's worth noting that we have recompose, so you could use [onlyUpdateForKeys](https://github.com/acdlite/recompose/blob/master/docs/API.md#onlyupdateforkeys), too.

jbalboni (approved):
Looks good, just had one concern about using state for the clicked flag.

src/js/common/components/LoginModal.jsx
cvalarida: Ohh, interesting. I'll have to do some looking into that; I've 1) never heard of it and 2) didn't know we used it.

src/js/common/components/LoginModal.jsx
cvalarida: Good point. I forget about the ability to rename in deconstruction statements like this. :man_shrugging:

src/js/common/components/LoginModal.jsx
cvalarida: Yeah, I guess we're not using it `render()` after all; best to not trigger a re-render just because I'm using `setState()`. Good call. 👍

PR #5779, opened 2017-06-16
Avoid sending null services param for benefits facilities

patrickvinograd:
Fixes https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3345

Restores the added condition that removes the services query param. 

Tested locally with testing all the filter permutations I could think of.


bshyong (approved):
LGTM

PR #5780, opened 2017-06-16
GIB currently disallowed alert

annekainicUSDS:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3297

This is the solution for case 1 in the list:
1. "Currently disallowed" (due to ineligibility) **(@annekainicUSDS taking this)**   
\- https://marvelapp.com/8d6igd9/screen/29244254
\- when `percentageBenefit === 0` and `originalEntitlement === 0`

Data I used for this:
```
:body:
  chapter33_education_info:
    first_name: Srikanth
    last_name: Vanapalli
    name_suffix: II
    date_of_birth: '1995-11-12T06:00:00.000+0000'    
    regional_processing_office: Central Office Washington, DC
    va_file_number: '12345678'
    eligibility_date: '2016-12-01T05:00:00.000+0000'
    delimiting_date: '2017-12-07T05:00:00.000+0000'
    original_entitlement: '0'
    used_entitlement: '75'
    remaining_entitlement: '25'
    percentage_benefit: '0'
  
:status: 200
```

screen shot 2017-06-16 at 9 06 37 am

### DIFF
--- a/src/js/post-911-gib-status/components/UserInfoSection.jsx
+++ b/src/js/post-911-gib-status/components/UserInfoSection.jsx
@@ -27,20 +27,7 @@ class UserInfoSection extends React.Component {
     }
 
     let entitlementInfo;
-    if (!currentlyAllowed) {
-      entitlementInfo = (
-        <div>
-          <h4>When You Can Receive Benefits</h4>
-          <div className="usa-alert usa-alert-warning usa-content">
-            <div className="usa-alert-body">
-              <h2>Currently Not Qualified</h2>
-              We have received your application and have determined that you are not currently eligible
-              for Post-9/11 GI Bill benefits. Additional service time could change this determination.
-            </div>
-          </div>
-        </div>
-      );
-    } else {
+    if (currentlyAllowed) {
       entitlementInfo = (
         <div>
           <div>
@@ -58,6 +45,19 @@ class UserInfoSection extends React.Component {
           <InfoPair label="Total months received" value={enrollmentData.originalEntitlement}/>
           <InfoPair label="Used" value={enrollmentData.usedEntitlement}/>
           <InfoPair label="Remaining" value={enrollmentData.remainingEntitlement}/>
+        </div>
+      );
+    } else {
+      entitlementInfo = (
+        <div>
+          <h4>When You Can Receive Benefits</h4>
+          <div className="usa-alert usa-alert-warning usa-content">
+            <div className="usa-alert-body">
+              <h2>Currently Not Qualified</h2>
+              We have received your application and have determined that you are not currently eligible
+              for Post-9/11 GI Bill benefits. Additional service time could change this determination.
+            </div>
+          </div>
         </div>
       );
     }
--- a/src/js/post-911-gib-status/components/UserInfoSection.jsx
+++ b/src/js/post-911-gib-status/components/UserInfoSection.jsx
@@ -33,7 +33,7 @@ class UserInfoSection extends React.Component {
           <h4>When You Can Receive Benefits</h4>
           <div className="usa-alert usa-alert-warning usa-content">
             <div className="usa-alert-body">
-              <h2>Currently Disallowed</h2>
+              <h2>Currently Not Qualified</h2>
               We have received your application and have determined that you are not currently eligible
               for Post-9/11 GI Bill benefits. Additional service time could change this determination.
             </div>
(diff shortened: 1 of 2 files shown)


src/js/post-911-gib-status/components/UserInfoSection.jsx
```diff
@@ -25,6 +25,42 @@ class UserInfoSection extends React.Component {
       );
     }
 
+    let entitlementInfo;
+    if (enrollmentData.percentageBenefit === 0 && enrollmentData.originalEntitlement === 0) {
```
raquelromano-gov: It looks like `currentlyAllowed` is defined as both of these values being non-zero, and currently disallowed is defined as both of them being zero, so (while I'm guessing this never happens), the case when one is zero and one is non-zero falls between the cracks. I say we just treat that case as allowed (if one is nonzero, you're probably getting benefits), so above we would set
```
const currentlyAllowed = enrollmentData.percentageBenefit !== 0 || enrollmentData.originalEntitlement !== 0;
```
and here just use `!currentlyAllowed`

src/js/post-911-gib-status/components/UserInfoSection.jsx
```diff
@@ -25,6 +25,42 @@ class UserInfoSection extends React.Component {
...
+              for Post-9/11 GI Bill benefits. Additional service time could change this determination.
+            </div>
+          </div>
+        </div>
+      );
+    } else {
+      entitlementInfo = (
+        <div>
+          <div>
+            <h4>When You Can Receive Benefits</h4>
```
raquelromano-gov: I think it would be a little more readable to split these into 2 variables, maybe `entitlementInfo` for the stuff under  "When You Can Receive Benefits" and `benefitLevel` for the stuff under "Your Benefit Level".

Then the logic can be something like
```
if (currentlyAllowed) {
  entitlementInfo = blah;
  benefitLevel = blah;
} else  {
  entitlementInfo = /* currently disallowed warning */;
}
```
and below on line 84
```
{entitlementInfo}
{benefitLevel}
```
Open to debate though

src/js/post-911-gib-status/components/UserInfoSection.jsx
annekainicUSDS: Yes, good point! I will fix this

src/js/post-911-gib-status/components/UserInfoSection.jsx
annekainicUSDS: I started doing it that way, and then had doubts because I thought it would be confusing to be returning 2 separate variables when only one of them will always be null, but I agree the naming of this makes it weird. I'll change it, thanks!

src/js/post-911-gib-status/components/UserInfoSection.jsx
annekainicUSDS: This leads me to a more general question of where is the right place to be storing this logic. At first I thought all the logic should live in `Main` or in helpers, and then values like `currentlyAllowed` are just passed to the components as props. That might be the cleanest way to do this, but if we decide to go that route we can refactor later.

src/js/post-911-gib-status/components/UserInfoSection.jsx
annekainicUSDS: Ah I misunderstood your comment when I first read it. It might be tricky to do this because there are actually 3 types of data being rendered under entitlement:
1. When You Can Receive Benefits
2. Your Benefit Level
3. Months received, used, and remaining

I could create 3 separate variables for each of these, but the reason why I wanted to group them is to show that either all 3 of these pieces are rendered, or none. If none, instead the disallowed message appears. I think this way it's a bit more clear when these things should be rendered, even if we're grouping 3 disparate pieces of information together. What do you think?

raquelromano-gov: Yeah, that makes sense to keep them all together. The only thing I might change is to swap what's tested:  first test  `if (currentlyAllowed)` rather than `!currentlyAllowed` on line 30 above. Not super important though, so feel free to merge.

raquelromano-gov (approved):

src/js/post-911-gib-status/components/UserInfoSection.jsx
annekainicUSDS: yep, makes sense! I can do that

raquelromano-gov (approved):
LGTM

PR #5786, opened 2017-06-16
Appeals error handling

bshyong:
- Moves around the error handling for claims
- Begin adding style changes/structure for appeals detail page


aub:
Looks like this already got merged into appeals-fe, but it LGTM in any case ;)

PR #5788, opened 2017-06-16
530 military history

cehsu:
Connects to department-of-veterans-affairs/vets.gov-team#3131

Although the issue indicates that all fields should be optional, I've made previous names required if the radio button is selected. Thanks for thoughts on whether this makes sense (enforces requirements upon partial completion).

screen shot 2017-06-19 at 10 07 15 am
screen shot 2017-06-19 at 10 06 56 am


stacyhlinn:
1- The copy shouldn't ask "Did you serve under another name?" Since the person filling out this form is not the veteran him/herself. Let's use "Did the veteran serve under another name?"

2- Looking at the original 21P-530 pdf again, and I noticed the question is " IF VETERAN SERVED UNDER NAME OTHER THAN THAT SHOWN IN ITEM 1, GIVE FULL NAME AND **SERVICE RENDERED UNDER THAT NAME**." Hmmm Let me rethink through this and check in with Josh.

cehsu:
Thanks @stacyhlinn, I have fixed the copy you mentioned in 1. As for point 2, very interesting, this may involve changing the back end schema to something like how marriages are handled.

stacyhlinn:
@cehsu Just talked to the team about this. Here's how we're going to handle chapter 3. Basically swapping the military history & names, and provide a page break in between. No schema changes needed.

- Branch [text field]
- Date of entry [date field]
- Place of entry [text field]
- Date of separation [date field]
- Place of separation [text field]
- Service number [text field]
- Rank [text field]
  [Button] Add a new service period. If clicked would create another set of the same fields from Branch through Rank as above.
—page break—
Did the veteran serve under another name? [Y/N]
{if yes} Name [text field]
[Button] Add another

cehsu:
Thanks for the guidance @stacyhlinn, changes made.

jbalboni (approved):
LGTM, just had a couple questions

src/js/burials/config/form.js
```diff
@@ -162,6 +168,52 @@ const formConfig = {
...
+          }
+        },
+        previousNames: {
+          title: 'Previous Names',
+          path: 'military-history/previous-names',
+          uiSchema: {
+            previousNames: {
+              'ui:options': {
+                expandUnder: 'view:serveUnderOtherNames',
+                viewField: FullNameField
```
cehsu: The user is allowed to add names without providing service period information.

test/burials/config/previousNames.unit.spec.jsx
```diff
@@ -0,0 +1,165 @@
...
+      <DefinitionTester
+          schema={schema}
+          definitions={formConfig.defaultDefinitions}
+          onSubmit={onSubmit}
+          uiSchema={uiSchema}/>
+    );
+
+    const formDOM = findDOMNode(form);
+
+    ReactTestUtils.Simulate.change(formDOM.querySelector('input[type="radio"]'), {
```
cehsu: Thanks for the reminder @jbalboni, this logic has been replaced with that which uses `getFormDOM`.

cehsu:
Thanks @jbalboni, change made, question answered.

jbalboni (approved):

PR #5789, opened 2017-06-16
SiP Loading Screen and Error Page

cvalarida:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3359
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3413
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3299 (just the loading errors, though)

When we start using this for loading from the profile page as well, we'll have to do some fancy footwork to make sure the Back button gets us to the right place (profile page vs introduction page).

Loading screen (waiting for data).
image

Note there's some weird alignment thing going on that I haven't figured out yet:
image

401 -- Session expired
image

404 -- Application not found
We'll probably have to change the copy on that since they won't be able to start over when we remove that start over button (shown below)...so...we shouldn't ask them to.
image

500 -- Server error
image

Also note: The buttons aren't the right size yet. I don't want to spend a long time fiddling with it (or wading through the existing css to find what I'm looking for), so I'll make another ticket for that when this is done.


src/js/common/schemaform/FormIntroButtons.jsx
```diff
@@ -25,7 +25,9 @@ class FormIntroButtons extends React.Component {
   handleLoadForm = () => {
     // If successful, this will set form.loadedData.metadata.returnUrl and will
     //  trickle down to this.props to be caught in componentWillReceiveProps
-    this.props.fetchInProgressForm(this.props.formId, this.props.migrations);
+    this.props.fetchInProgressForm(this.props.formId, this.props.migrations)
+      // Should probably do GA or sentry or something, but for now...
+      .catch((e) => console.error('Error loading form:', e)); // eslint-disable-line no-console
```
jbalboni: I think this can go away. My Sentry errors PR should catch and log errors here.

src/js/common/schemaform/LoadingPage.jsx
```diff
@@ -0,0 +1,124 @@
...
+
+import { LOAD_STATUSES } from './save-load-actions';
+import LoadingIndicator from '../components/LoadingIndicator';
+import SignInLink from '../components/SignInLink';
+import ProgressButton from '../components/form-elements/ProgressButton';
+
+
+// TODO: Determine whether I can connect() and withRouter() this component and still
+//  be alright for testing; the contamination in FormApp is getting silly.
+class LoadingPage extends React.Component {
```
jbalboni: Like we talked about, this should probably be `LoadingScreen` or `SaveInProgressLoading` or something.

src/js/common/schemaform/FormApp.jsx
```diff
@@ -43,14 +56,40 @@ export default class FormApp extends React.Component {
...
     const isIntroductionPage = trimmedPathname.endsWith('introduction');
     const isConfirmationPage = trimmedPathname.endsWith('confirmation');
     const GetFormHelp = formConfig.getHelp;
     let content;
-    if (!isInProgress(trimmedPathname)) {
+
+    // TODO: As soon as we've redirected, remember to set loadedStatus back to 'not-attempted'
+    if (!formConfig.disableSave && this.props.loadedStatus !== LOAD_STATUSES.notAttempted) {
+      // Show the loading screen instead of the children.
+      // Agh! The prop threading!
```
jbalboni: I think part of the reason this is painful is because `LoadingPage` is trying to be two things: a separate page for the user to figure out what to do on an error, and a spinner indicating that there's a pending state. The way it is now is sort of confusing, because it feels like you're on a new page when there's an error, but the url hasn't changed.

I think you can split this in two. Make `LoadingPage` something like `ErrorPage` and remove the spinner. You can add it to the routes for the app. Then make a `InProgressFormLoading` component, that `FormApp` can put up whenever it sees the right loading status (or just use `LoadingIndicator`). `FormApp` would probably also handle routing to the `ErrorPage` or the return url, based on the load status, which I think is ok.

Prop threading would go away, because `ErrorPage` would be a connected component and `FormApp` would really just handling some global routing, which seems ok since it's the global form component.

src/js/common/schemaform/save-load-actions.js
```diff
@@ -253,7 +253,7 @@ export function fetchInProgressForm(formId, migrations) {
       dispatch(setFetchFormStatus(loadedStatus));
 
       // Return a rejected promise to tell the caller there was a problem
-      return Promise.reject();
+      return Promise.reject(status);
```
jbalboni: I took this out completely in another PR, fyi.

src/js/hca/config/form.js
```diff
@@ -120,6 +120,10 @@ const formConfig = {
   version: 0,
   // Disable save in progress for production
   disableSave: __BUILDTYPE__ === 'production',
+  savedFormErrorMessages: {
```
jbalboni: I kinda wish we had grouped all these config options into an object, like `saveInProgress`.

jbalboni:
@cvalarida Something else I noticed, when you click "Sign in" after getting a no-auth status, you can sign in again but nothing happens after that.

src/js/common/schemaform/FormIntroButtons.jsx
cvalarida: 👍 I'll leave it in here for now because it'll be taken out with [your PR](https://github.com/department-of-veterans-affairs/vets-website/pull/5821).

cvalarida:
@jbalboni Yeah, I noticed that too, but I assumed it was something to do with _how_ was testing it. Because, in order for the resume button to appear, I had to be logged in, so I was logged in and just faking a 401 from `fetch`, so when `FormApp`'s `componentWillReceiveProps` checks to see `if (!wasLoggedIn && isLoggedIn)`, that'll be false and it won't fire `onLogin`.

src/js/hca/config/form.js
cvalarida: Yeah, we should probably make a ticket for that.

src/js/common/schemaform/LoadingPage.jsx
cvalarida: Yes! I'll rename it.

src/js/common/schemaform/FormApp.jsx
cvalarida: Yeah, that's why I wanted to make `LoadingPage` an actual page. Making an error page probably makes more sense, though.

I remember talking about that, but I was thinking that was covered in https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3410. Turns out not so much.

jbalboni:
@cvalarida It looks like we may need to do the thing I added to `saveInProgressForm`: dispatch a log out action when we get a 401, so that the state updates.

cvalarida:
I just tried loading the form after the meeting we were just in and got a 401. Excitedly, I logged in and...!

...nothing. It didn't do a thing after I logged in. 😞

Yeah, I think dispatching a logout action upon receiving a 401 I think is the easy win; I just checked and saw that `currentlyLoggedIn` is `true`, so....yerp. Let's give that a go!

src/js/common/schemaform/FormApp.jsx
```diff
@@ -50,7 +67,11 @@ export default class FormApp extends React.Component {
     const isConfirmationPage = trimmedPathname.endsWith('confirmation');
     const GetFormHelp = formConfig.getHelp;
     let content;
-    if (!isInProgress(trimmedPathname)) {
+
+    // TODO: As soon as we've redirected, remember to set loadedStatus back to 'not-attempted'
```
jbalboni: Should we remove this TODO?

src/js/common/schemaform/FormIntroButtons.jsx
jbalboni: You might want to remove this specifically, so it doesn't overwrite my already merged change.

src/js/common/schemaform/SaveInProgressErrorPage.jsx
```diff
@@ -0,0 +1,126 @@
...
+          buttonClass={buttonClass}
+          buttonText="Back"
+          beforeText="«"/>
+    );
+  }
+
+  // Reload the form and try again.
+  reloadForm = () => {
+    // formConfig is put in this.props.routes[length - 1]
+    const formConfig = this.props.routes[this.props.routes.length - 1].formConfig;
```
jbalboni: I think you should be able to get this from `this.props.route` in here.

src/js/common/schemaform/SaveInProgressErrorPage.jsx
```diff
@@ -0,0 +1,126 @@
...
+
+import { updateLogInUrl } from '../../login/actions';
+
+
+// For now, this only handles loading errors, but it could feasibly be reworked
+//  to handle save errors as well if we need it to.
+class SaveInProgressErrorPage extends React.Component {
+  getBackButton = (primary = false) => {
+    const buttonClass = primary ? 'usa-button-primary' : 'usa-button-outline';
+    const goBack = () => {
```
jbalboni: I think you could make this a class method (would be slightly more efficient, and I think clearer).

src/js/common/schemaform/SaveInProgressErrorPage.jsx
```diff
@@ -0,0 +1,126 @@
...
+  // Reload the form and try again.
+  reloadForm = () => {
+    // formConfig is put in this.props.routes[length - 1]
+    const formConfig = this.props.routes[this.props.routes.length - 1].formConfig;
+    this.props.fetchInProgressForm(formConfig.formId, formConfig.migrations);
+  }
+
+  render() {
+    const { loadedStatus } = this.props;
+    const { noAuth, notFound } = this.props.errorMessages || {};
```
jbalboni: It looks like these aren't coming through correctly, I see a truncated sign in message when running this.

src/js/common/schemaform/SaveInProgressErrorPage.jsx
```diff
@@ -0,0 +1,126 @@
...
+    return (
+      <div>
+        {content}
+      </div>
+    );
+  }
+}
+
+SaveInProgressErrorPage.propTypes = {
+  formConfig: PropTypes.object.isRequired,
```
jbalboni: This is throwing a prop type error.

jbalboni (approved):
A few things to fix, but looks good overall.

src/js/common/schemaform/SaveInProgressErrorPage.jsx
cvalarida: Ah! Would you look at that? Cool!

Danke!

src/js/common/schemaform/SaveInProgressErrorPage.jsx
cvalarida: Right! Now that it's its own page, need to change this to get from `this.props.formConfig.errorMessages` and change the tests. Good catch.

src/js/common/schemaform/SaveInProgressErrorPage.jsx
cvalarida: During testing?

cvalarida: Ah, yes...I thought passing it in the way I was in `helpers.js#createRoutes` would pass it as a prop. And...it doesn't...so.......yerp. 😞

PR #5790, opened 2017-06-17
Unit tests for EnrollmentPeriod, EnrollmentHistory

raquelromano-gov:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3347
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3346

### DIFF
--- a/test/post-911-gib-status/components/EnrollmentHistory.unit.spec.js
+++ b/test/post-911-gib-status/components/EnrollmentHistory.unit.spec.js
@@ -4,18 +4,76 @@ import { expect } from 'chai';
 
 import EnrollmentHistory from '../../../src/js/post-911-gib-status/components/EnrollmentHistory.jsx';
 
-const props = {
+const defaultProps = {
   enrollmentData: {
+    firstName: 'Joe',
+    lastName: 'West',
+    dateOfBirth: '1995-11-12T06:00:00.000+0000',
+    vaFileNumber: '12345678',
+    regionalProcessingOffice: 'Central Office Washington, DC',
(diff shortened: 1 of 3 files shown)


annekainicUSDS (approved):
LGTM! Only one question.

test/post-911-gib-status/components/EnrollmentHistory.unit.spec.js
```diff
@@ -4,18 +4,76 @@ import { expect } from 'chai';
...
     const vdom = tree.getRenderOutput();
-    expect(vdom).to.not.be.undefined;
+    expect(vdom).to.exist;
   });
 
-  // test logic controlling which components are shown depending on data
-});
+  it('should show enrollments', () => {
+    const tree = SkinDeep.shallowRender(<EnrollmentHistory {...defaultProps}/>);
+    expect(tree.subTree('EnrollmentPeriod')).to.be.ok;
```
raquelromano-gov: It just tests that the EnrollmentPeriod component is there, but yes, I noticed that a lot of the ways we check for components in our unit tests are explicitly not recommended by chai.

We don't want to check for facility name because that's rendered by the EnrollmentPeriod component, not EnrollmentHistory, so I added a check for the "Enrollment History" header. I'm not sure how else to check that an EnrollmentPeriod is rendered--I changed it from `to.be.ok` to `to.exist` but that is *also* not recommended. Maybe it's not recommended if that's the only thing we're checking? Will ask around on #vg-frontend

PR #5791, opened 2017-06-19
Fix sitemap-1 accessibility issues

raquelromano-gov:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3289

### DIFF
--- a/content/pages/burials-and-memorials/burial-planning.md
+++ b/content/pages/burials-and-memorials/burial-planning.md
@@ -27,6 +27,7 @@ Yes. You can plan ahead to make the process of applying for a burial in a VA nat
 - Send it by U.S. mail to: 
   
 <dl class="va-address-block">
+    <dt></dt>
     <dd>National Cemetery Scheduling Office</dd>
     <dd>P.O. Box 510543</dd>
     <dd>St. Louis, MO 63151</dd>
--- a/content/pages/playbook/design/visual.html
+++ b/content/pages/playbook/design/visual.html
@@ -29,7 +29,7 @@ layout: page-playbook.html
 
           <dl class="card">
             <dt>Headings: Roboto Slab, Source Sans Pro</dt>
-            <p>Content headings (headings 1 and 2 are reserved for the vets.gov header and breadcrumb navigation)</p>
+            <dt>Note that headings 1 and 2 are reserved for the vets.gov header and breadcrumb navigation</dt>
             <dd><h3>Heading 3</h3></dd>
             <dd><h4>Heading 4</h4></dd>
             <dd><h5>Heading 5</h5></dd>
--- a/content/pages/playbook/design/visual.html
+++ b/content/pages/playbook/design/visual.html
@@ -29,7 +29,8 @@ layout: page-playbook.html
 
           <dl class="card">
             <dt>Headings: Roboto Slab, Source Sans Pro</dt>
-            <dt>Note that headings 1 and 2 are reserved for the vets.gov header and breadcrumb navigation</dt>
+            <dd><h1>Heading 1</h1></dd>
+            <dd><h2>Heading 2</h2></dd>
             <dd><h3>Heading 3</h3></dd>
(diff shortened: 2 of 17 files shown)


content/pages/playbook/design/visual.html
```diff
@@ -29,7 +29,7 @@
 
           <dl class="card">
             <dt>Headings: Roboto Slab, Source Sans Pro</dt>
-            <p>Content headings (headings 1 and 2 are reserved for the vets.gov header and breadcrumb navigation)</p>
+            <dt>Note that headings 1 and 2 are reserved for the vets.gov header and breadcrumb navigation</dt>
```
raquelromano-gov: @melwoodard I made an editorial change here, but LMK if I should leave as is or if you have a better suggestion.

test/e2e/sitemap-helpers.js
```diff
@@ -14,8 +14,8 @@ function sitemapURLs(callback) {
       const doc = libxmljs.parseXml(body);
 
       const urls = doc.find('//xmlns:loc', SITEMAP_LOC_NS)
-        .map(n => n.text().replace(BUILD_BASE_URL, E2eHelpers.baseUrl))
-        .filter(url => !(url.endsWith('facility-locator/') || url.endsWith('auth/login/callback/')));
```
raquelromano-gov: Facility locator seem to pass all checks, so taking it out of here.

test/accessibility/sitemap-1.spec.js
```diff
@@ -13,11 +13,23 @@ module.exports = {
       const mark = Math.ceil(urls.length / 4);
       const segment = urls.splice(0, mark);
 
+      // Whitelist of URLs on which to only run'section508' rules and not 'wcag2a'
+      // rules. Please provide an explanation for each exception.
+      const wcagSkip = [
```
raquelromano-gov: Actually, I think I'll move this whitelist into `sitemap-helpers.js` since it's not guaranteed which URL will be distributed to which of the `sitemap-*spec.js` files

@jbalboni  I'll ping you when this is ready for review

content/pages/playbook/design/visual.html
melwoodard: @raquelromano-gov this actually isn't true anymore (we use H1 for the page title, and breadcrumbs are now a `nav` element), so maybe we should remove that line altogether for and add Heading 1 and Heading 2 back in for the time being. @suzchap thoughts?

(FYI there's an issue in the Nebula backlog somewhere to revisit this hierarchy)

image

suzchap: @melwoodard yes, I agree!

raquelromano-gov:
@jbalboni I finally got back to this--would you be willing to review again?

content/pages/burials-and-memorials/burial-planning.md
```diff
@@ -27,6 +27,7 @@ Yes. You can plan ahead to make the process of applying for a burial in a VA nat
 - Send it by U.S. mail to: 
   
 <dl class="va-address-block">
+    <dt></dt>
```
jbalboni: It doesn't seem like a definition list makes sense for an address, does it? Should we switch this whole thing to be in a `p` tag?

jbalboni (approved):
The changes seem fine to avoid the errors, but I'm not sure the address markup makes sense. But maybe that's a bigger task that should be done separately.

content/pages/burials-and-memorials/burial-planning.md
raquelromano-gov: Yes, I thought that was weird too. There are ~15 `va-address-block` elements in vets-website, so I'd rather file at ticket against the content team to update those and not do it here if that's OK.

raquelromano-gov: https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3859

PR #5793, opened 2017-06-19
Add document upload pages to burials and pensions

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3326
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3376

This updates the file field to follow this spec,
image
with some deviations.

The deviations are:
- Document upload page does not use the Review and submit chapter header, it uses Additional Information, which we changed after that screenshot was created.
- Required field notation uses our standard (*Required), not just an asterisk
- No gray borders between file fields. I think we can do this, but it's a deviation from our current field layout and doesn't seem critical.
- On the review page, you don't see the upload button horizontal with the label, you see it as it looks on the regular upload page. This is something I couldn't find a good way of doing, so I want to punt it to a separate item and not spend a bunch of time on right now.

I think that's it. I will make an issue for items 2 and 3 above.

screen shot 2017-06-19 at 1 12 44 pm
screen shot 2017-06-19 at 1 12 32 pm
screen shot 2017-06-19 at 1 12 23 pm

screen shot 2017-06-19 at 1 10 40 pm

screen shot 2017-06-19 at 1 24 57 pm

screen shot 2017-06-19 at 1 10 14 pm
screen shot 2017-06-19 at 1 10 11 pm
screen shot 2017-06-19 at 1 10 02 pm

Note: above burials pages now have documents in an additional information chapter:
screen shot 2017-06-20 at 2 28 52 pm
screen shot 2017-06-20 at 2 27 45 pm


package.json
```diff
@@ -192,6 +192,6 @@
   "dependencies": {
     "camelcase-keys-recursive": "^0.8.2",
     "reselect": "^2.5.4",
-    "vets-json-schema": "https://github.com/department-of-veterans-affairs/vets-json-schema.git#1f656661958459d90e468c9ebd13c390d5bf483c"
```
jbalboni: This is not in master.

stacyhlinn:
@jbalboni the reason that document upload lives under the "Review & Submit" umbrella is because not all users will encounter document upload, so we don't know how many chapters there are to begin with if we separate it out. Putting it under Review & Submit seems like a reasonable compromise.

jbalboni:
@stacyhlinn Ok, I'll add that to the ticket I'm going to create. It doesn't fit in with how the rjsf forms work right now.

jbalboni:
FYI @U-DON, this would also probably be used for preneed.

src/js/pensions/helpers.jsx
```diff
@@ -39,3 +39,15 @@ export function getMarriageTitleWithCurrent(form, index) {
 
 export const spouseContribution = <span>How much do you <strong>contribute monthly</strong> to your spouse’s support?</span>;
 
+export function fileHelp({ formContext }) {
+  if (formContext.reviewMode) {
+    return <p/>;
```
cvalarida: Just `<p/>` for the spacing that gives, or would `null` be a problem?

src/sass/modules/_m-schemaform.scss
```diff
@@ -281,4 +281,40 @@ label {
...
+  > .schemaform-file-item-edit {
+    flex-direction: column;
+    align-items: left;
+  }
+}
+
+.schemaform-file-list-buttons-editing {
+  display: flex;
+  > .schemaform-file-remove-button {
+    width: inherit;
```
cvalarida: Is this overridden somewhere else? (That is, do you _have_ to specify an inherited with?)

I assume so, but figured I'd flag it just in case.

test/burials/config/documents.unit.spec.jsx
```diff
@@ -0,0 +1,133 @@
...
+  it('should render', () => {
+    const form = ReactTestUtils.renderIntoDocument(
+      <DefinitionTester
+          schema={schema}
+          definitions={formConfig.defaultDefinitions}
+          uiSchema={uiSchema}/>
+    );
+    const formDOM = getFormDOM(form);
+
+    expect(formDOM.querySelectorAll('input,select').length).to.equal(0);
```
cvalarida: This seems odd, testing that the component renders by testing that there are no `input` or `select` elements. Is there something more specific we can check for? Like the upload button? Text?

test/burials/config/documents.unit.spec.jsx
```diff
@@ -0,0 +1,133 @@
...
+            }
+          }}
+          definitions={formConfig.defaultDefinitions}
+          uiSchema={uiSchema}/>
+    );
+    const formDOM = getFormDOM(form);
+
+    expect(formDOM.querySelectorAll('input,select').length).to.equal(1);
+  });
+
```
cvalarida: Would it provide value to have a test that makes sure they both render together? I mean, I see that the next test does render them both and checks that they both have errors, so maybe that's sufficient..?

test/common/schemaform/FileField.unit.spec.jsx
```diff
@@ -0,0 +1,243 @@
...
+      <FileField
+          schema={schema}
+          uiSchema={uiSchema}
+          idSchema={idSchema}
+          formContext={formContext}
+          onChange={f => f}
+          requiredSchema={requiredSchema}/>
+    );
+
+    expect(tree.everySubTree('label')[0].text()).to.contain('Upload');
```
cvalarida: Should we be checking for anything else? Upload button, etc?

test/common/schemaform/FileField.unit.spec.jsx
```diff
@@ -0,0 +1,243 @@
...
+          schema={schema}
+          uiSchema={uiSchema}
+          idSchema={idSchema}
+          formData={formData}
+          formContext={formContext}
+          onChange={f => f}
+          requiredSchema={requiredSchema}/>
+    );
+
+    expect(tree.everySubTree('label')).to.be.empty;
```
cvalarida: Is the label on the button? Just seems odd for the test to be `should not render upload button if over max items` and for us to not check for a button. :man_shrugging:

Apparently I'm all about finding buttons to push... 🤔

test/common/schemaform/FileField.unit.spec.jsx
```diff
@@ -0,0 +1,243 @@
...
+            fileField: uiSchema
+          }}/>
+    );
+    const formDOM = getFormDOM(form);
+
+    formDOM.click('button.usa-button-outline');
+
+    expect(formDOM.querySelectorAll('.usa-button').length).to.equal(2);
+    expect(formDOM.querySelectorAll('.usa-button')[0].textContent).to.equal('Replace');
+    expect(formDOM.querySelectorAll('.usa-button')[1].textContent).to.equal('Cancel');
```
cvalarida: Nit: Could do a `const buttons = formDOM.querySelectorAll('.usa-button')` and use that in these three `expect`ations.

test/common/schemaform/actions.unit.spec.js
```diff
@@ -193,4 +194,195 @@ describe('Schemaform actions:', () => {
       });
     });
   });
+  describe('uploadFile', () => {
+    let fetchMock;
+    let oldFetch;
```
cvalarida: We pretty much never want to _actually_ use `fetch` in a unit test, right? Would it make sense to globalize this change and provide hooks to change what `fetch` returns?

cvalarida (approved):
Phew! Made it!

Yeah, I may be missing some things, but everything I read looks good. I figure if it passes the tests, it's golden, right?

src/js/common/schemaform/actions.js
```diff
@@ -101,3 +101,53 @@ export function submitForm(formConfig, form) {
     });
   };
 }
+
+export function uploadFile(file, filePath, uiOptions = {}) {
+  return (dispatch, getState) => {
+    if (file.size > uiOptions.maxSize) {
```
jbalboni: There is a default in the common definition for file.

src/js/pensions/helpers.jsx
jbalboni: Hmm, `null` should be ok.

src/sass/modules/_m-schemaform.scss
jbalboni: Yeah, I'm trying to override it being set to `auto` somewhere else, I believe.

test/burials/config/documents.unit.spec.jsx
jbalboni: I mainly wanted to make sure nothing showed up if you didn't make the right choices for form data. Normally you won't hit this state because of `depends`.

test/burials/config/documents.unit.spec.jsx
jbalboni: Yeah, I think that's enough.

test/common/schemaform/FileField.unit.spec.jsx
jbalboni: The upload button is the label here, because we want to hide the usual file input element and just have a button that triggers the file explorer. It's weird.

test/common/schemaform/FileField.unit.spec.jsx
jbalboni: The label is the button in this widget.

test/common/schemaform/actions.unit.spec.js
jbalboni: Yeah, that would probably be good, if we could figure out a reasonable way to make it reusable. There might be libraries for it, too.

test/burials/config/documents.unit.spec.jsx
cvalarida: Gotcha. So it's more a `'should render nothing if neither x nor y'`. Makes sense. Still weird. :man_shrugging:

test/common/schemaform/FileField.unit.spec.jsx
cvalarida: Oh, that is weird. Does make that bit of code make more sense to me now, though.

test/common/schemaform/actions.unit.spec.js
cvalarida: 👍 https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3460

PR #5796, opened 2017-06-19
Update build flag for save in progress forms

jbalboni:
Fixes the save in progress flag so that the functionality shows up in staging.


cvalarida (approved):
Awesome, thanks!

PR #5801, opened 2017-06-20
Remove confirm address page

raquelromano-gov:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3358

(Felt bad ripping this all out!)


annekainicUSDS (approved):
LGTM! I think we might want to change the component hierarchy to be similar to what GIBS does (`VALettersApp` > `Main` > etc.), where the fetch is in `Main`, but we can make those changes in a separate PR.

raquelromano-gov:
Yeah, good idea to introduce a `Main` component. I'll create a ticket.

PR #5803, opened 2017-06-20
Correct misspelling in burials form property names

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3343

No UI changes, just corrects my inability to spell lately.

### DIFF
--- a/package.json
+++ b/package.json
@@ -192,6 +192,6 @@
   "dependencies": {
     "camelcase-keys-recursive": "^0.8.2",
     "reselect": "^2.5.4",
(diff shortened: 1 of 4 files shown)


cvalarida (approved):
Assuming you did a search through everything relevant, it looks good!

cehsu (approved):

cehsu:
Spelling is hard. Would generating documentation + automating spellchecking of said documentation be helpful here?

cvalarida:
@cehsu I'm guessing that'll be more work than it's probably worth. Unless that documentation is used for something other than spelling, I wouldn't worry about it.

Having a spellchecker of user-facing text might be good, though, but again, I'm still not worried about it since it's all peer reviewed. It just falls to the reviewers to make sure the text is spelled right.

PR #5804, opened 2017-06-20
Split letters content into steps

raquelromano-gov:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3416

Working on unit tests for these new components, but glad for any feedback in the meantime.

image

### DIFF
--- a/src/js/va-letters/components/AddressSection.jsx
+++ b/src/js/va-letters/components/AddressSection.jsx
@@ -7,11 +7,12 @@ class AddressSection extends React.Component {
     const addressLines = [
       destination.addressLine1,
       destination.addressLine2 ? `, ${destination.addressLine2}` : '',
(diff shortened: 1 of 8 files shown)


annekainicUSDS (commented):
This looks good so far! There's something funky going on with the commas in the address for the first address line, but that's minor. Also I wonder if we want to make `StepHeader` accessible from the `/common` directory, since the schemaform forms use something pretty similar. That can be for a future PR, though.

raquelromano-gov:
Thanks, fixed the comma issue.

Yeah, this could be helpful in common. I should poke around to see where else people have stand-alone steps. I think i did actually, but now don't recall.

annekainicUSDS (approved):
This looks great! I think this is ready to merge :)

PR #5806, opened 2017-06-20
Block indexing of scorecard

robertfairhead:
Closes https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3419

Indexing should also be blocked through metatag on the scorecard pages. Adding here as well for defense in depth against different crawlers' behavior.


ayaleloehr (approved):
LGTM

PR #5809, opened 2017-06-20
Update SiP error messaging when saving

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3367

Updates the messaging for errors when saving. I also tweaked the Sentry errors a little to give us more info.

screen shot 2017-06-20 at 1 14 33 pm
screen shot 2017-06-20 at 12 19 42 pm


cvalarida (approved):
Looks good.

Is the idea here to get something out before we do the saving page, or are we going to wait on implementing that until...later? Until we get a firm charge from the design team?

jbalboni:
@cvalarida I'm not sure what you mean. There's a page that you get routed to after saving that exists now and the errors here are shown in line with the form, according to the design. I think there's some work to show a spinner while waiting, but I want to wait until your PR is done for that and it's separate from this ticket anyway, I think.

cvalarida:
Gotcha. I thought that the page we'll show while waiting to save will also show the errors if it was unsuccessful (just like it does for loading a saved form).

I guess the design has been updated (should have followed the link to the related issue). And if it's changed later, that's a separate issue, so... :man_shrugging:

PR #5811, opened 2017-06-20
Update save in progress Sentry errors

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3365

This adds Sentry errors and messaging in the save / load actions. They mostly should have a `vets_sip_error` prefix in the message (or a separate message with that I didn't want to lose the stack trace). I also updated the promise handling to match the save function.


test/common/schemaform/save-load-actions.unit.spec.js
```diff
@@ -200,40 +200,37 @@ describe('Schemaform save / load actions:', () => {
...
 
-    it('dispatches a no-auth if the user has no session token', (done) => {
+    it('dispatches a no-auth if the user has no session token', () => {
       const thunk = fetchInProgressForm('hca', 0, 'some/path', {});
       const dispatch = sinon.spy();
       delete sessionStorage.userToken;
 
-      thunk(dispatch).catch(() => {
+      return thunk(dispatch).then(() => {
+        expect(dispatch.calledOnce).to.be.true;
```
cvalarida: Returning a promise. Good to know. 👍

cvalarida (approved):
Thanks for cleaning up the tests, too!

PR #5818, opened 2017-06-21
Terms of service tweaks

aub:
This fixes two issues from department-of-veterans-affairs/vets.gov-team#3423, namely:

* It makes the path that you get to when hitting cancel on the terms of service dialog be configurable, and then uses that to stop sending users to the healthcare page when cancelling from the profile page.

* On the profile page, it stops us from automatically loading in the terms and conditions data every time the page is opened. Now it will only load them in when the user clicks the link to bring up the dialog.


patrickvinograd (approved):
LGTM 🚀

PR #5819, opened 2017-06-21
Appeals status page

bshyong:
Additional work on status page for displaying a specific appeal and pulling up appropriate copy.

### DIFF
--- a/src/js/disability-benefits/containers/AppealStatusPage.jsx
+++ b/src/js/disability-benefits/containers/AppealStatusPage.jsx
@@ -23,12 +23,16 @@ class AppealStatusPage extends React.Component {
           </div>
         </div>
         <div className="row">
-          <div className="next-action">
-            next action explanation
-          </div>
-        </div>
-        <div className="row">
           <div className="small-12 usa-width-two-thirds medium-8 columns">
+            <div className="row">
+              <div className="next-action">
+                <h4></h4>
+                next action explanation
+              </div>
+            </div>
+            <div className="row">
+
+            </div>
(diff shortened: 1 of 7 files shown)


aub (approved):
Looks great.

src/js/disability-benefits/containers/AppealStatusPage.jsx
```diff
@@ -23,13 +39,26 @@ class AppealStatusPage extends React.Component {
...
+              <div className="next-action">
+                <h4>{nextAction.title}</h4>
+                {nextAction.description}
+              </div>
+            </div>
+            <div className="row">
+              <div className="last-status">
+                <h4>{status.title}</h4>
+                <strong>{moment(lastEvent.date).format('MMM DD, YYYY')}</strong>
+                <p dangerouslySetInnerHTML={{ __html: status.description }}/>
```
U-DON: I think we should avoid using `dangerouslySetInnerHTML` where possible. I see you have some HTML in `appealStatusDescriptions`; you could just unstringify that and import React in that module and it would be fine.

With that, you could also remove the `eslint-disable` above, which seems to have never gotten enabled again.

U-DON (commented):
LGTM, but still take issue with the usage of `dangerouslySetInnerHTML`.

src/js/disability-benefits/containers/AppealStatusPage.jsx
bshyong: Got it - recommendation is to make appealStatusDescriptions return react nodes to avoid setting raw html?

U-DON: Yup!

PR #5820, opened 2017-06-21
Update fetch to go to vets-api

annekainicUSDS:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3190

This PR does 3 things:
1. Fetches data from vets-api endpoint (and has to change some other things due to the different response structure)
2. Adds a test for the reducer
3. Fixes the `ignoreList` addition for letters

This will have to be updated again once @raquelromano-gov's PR (https://github.com/department-of-veterans-affairs/vets-website/pull/5804) is merged, but wanted to get it up for now.

### DIFF
--- a/test/va-letters/reducers/index.unit.spec.js
+++ b/test/va-letters/reducers/index.unit.spec.js
@@ -1,0 +1,63 @@
+import { expect } from 'chai';
+
+import lettersReducer from '../../../src/js/va-letters/reducers';
+
+const initialState = {
+  letters: [],
+  destination: {},
+  available: false
+};
+
+describe('letters reducer', () => {
+  it('should handle failure to fetch letters', () => {
+    const state = lettersReducer.letters(
+      initialState,
+      { type: 'GET_LETTERS_FAILURE' }
+    );
+
+    expect(state.letters).to.be.empty;
+    expect(state.available).to.be.false;
+  });
+
+  it('should handle a successful request for letters', () => {
+    const state = lettersReducer.letters(
+      initialState,
+      {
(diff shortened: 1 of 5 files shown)


annekainicUSDS:
Working on adding a test for this as well

test/va-letters/reducers/index.unit.spec.js
```diff
@@ -0,0 +1,63 @@
...
+
+const initialState = {
+  letters: [],
+  destination: {},
+  available: false
+};
+
+describe('letters reducer', () => {
+  it('should handle failure to fetch letters', () => {
+    const state = lettersReducer.letters(
```
annekainicUSDS: The only thing I'm struggling to understand is the import of the reducers why the import is returning something like this: 
```
lettersReducer: {
  letters: function()
}
```
I would think the import would just import the single reducer function we need here and not an object with the reducer function within it.

jbalboni: Yeah, this is due to how the common store is set up. When `va-letters/reducer` is pulled in by the common store function, it's wrapped in a `combineReducers` call, which turns that object into a function. Since you're importing the reducer directly, you're skipping that step. We used to have the `combineReducers` call in the reducer code for each app, but we can't do that and easily combine it with the user profile/login reducers.

raquelromano-gov (approved):

test/va-letters/reducers/index.unit.spec.js
annekainicUSDS: Ah gotcha, thanks for explaining that!

PR #5821, opened 2017-06-21
Use prefill data for authenticated users without a saved form

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3392

This changes the flow to always request a form if you're logged in. If you don't have a saved form, you'll get the prefilled version and it will be loaded as usual. If there's no prefilled data or some kind of non-auth related error occurs, it'll load a blank form.

Note that this makes it pretty obvious that we need to either hide the start over button or delete an existing form, as without one of those changes, resume and start over do essentially the same thing.


src/js/common/schemaform/FormApp.jsx
```diff
@@ -31,6 +31,9 @@ class FormApp extends React.Component {
       newProps.router.push(newProps.returnUrl);
       // Set loadedStatus in redux to not-attempted to not show the loading page
       newProps.setFetchFormStatus(LOAD_STATUSES.notAttempted);
+    } else if (newProps.loadedStatus === LOAD_STATUSES.prefillComplete) {
+      newProps.router.push(newProps.routes[newProps.routes.length - 1].pageList[1].path);
```
cvalarida: So....I get that this is going to the "first" page in the form (after the intro), but how does that work? What does `newProps.routes` contain and why are we referencing the last element in there?

This question is probably long-overdue. 😳

src/js/hca/config/form.js
```diff
@@ -116,7 +116,7 @@ const formConfig = {
   urlPrefix: '/',
   submitUrl: '/v0/health_care_applications',
   trackingPrefix: 'hca-',
-  formId: 'hca',
+  formId: '1010ez',
```
cvalarida: Awesome. I was going to make a ticket for this. 😄

test/util/schemaform-utils.jsx
```diff
@@ -104,5 +104,9 @@ export function getFormDOM(form) {
     });
   };
 
+  formDOM.click = function click(selector) {
```
cvalarida: 🎉

src/js/common/schemaform/FormIntroButtons.jsx
```diff
@@ -22,12 +22,18 @@ class FormIntroButtons extends React.Component {
...
+      this.goToBeginning();
+    }
+  }
+
+  handleLoadForm = (prefill = false) => {
     // If successful, this will set form.loadedData.metadata.returnUrl and will
     //  trickle down to this.props to be caught in componentWillReceiveProps
-    this.props.fetchInProgressForm(this.props.formId, this.props.migrations)
-      // Should probably do GA or sentry or something, but for now...
-      .catch((e) => console.error('Error loading form:', e)); // eslint-disable-line no-console
```
cvalarida: I think I added this when I was testing the loading screen errors. Gave me an `unhandled rejection` error (or some such) without it, so I don't know if we want to get rid of this...unless I'm missing something.

cvalarida (approved):
As of https://github.com/department-of-veterans-affairs/vets-api/pull/1078 we can get a list of prefills available when the user logs in, so we may want to use that here. If you want to make another ticket for that and tackle it there, that's fine too.

src/js/common/schemaform/FormApp.jsx
jbalboni: I learned this today, actually, but when you're in an component that has children that are routes, you get an array of routes, one for each nested child used by the current route. So it looks like ["/", "introduction"] when you're on the intro page (though with objects, not strings).

src/js/common/schemaform/FormIntroButtons.jsx
jbalboni: That should be fixed in the error handling PRs, since I updated it to always catch promise rejections.

src/js/common/schemaform/FormApp.jsx
cvalarida: 💯

jbalboni:
This branches off of @cvalarida's loading screen PR, so I'm going to leave this open until that lands.

cvalarida (approved):
🎉

PR #5823, opened 2017-06-21
Add test for the reducer

annekainicUSDS:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3341

Adding a test for the two actions that are possible in this return: successful fetch of data and failed fetch of data.

### DIFF
--- a/test/post-911-gib-status/reducers/index.unit.spec.js
+++ b/test/post-911-gib-status/reducers/index.unit.spec.js
@@ -32,7 +32,7 @@ describe('post911GIBStatus reducer', () => {
       }
     );
 
-    expect(state.enrollmentData.firstName).to.eql('Jane');
+    expect(state.enrollmentData.firstName).to.equal('Jane');
     expect(state.available).to.be.true;
   });
 });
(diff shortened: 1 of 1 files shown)


test/post-911-gib-status/reducers/index.unit.spec.js
```diff
@@ -0,0 +1,38 @@
...
+        data: {
+          firstName: 'Jane',
+          lastName: 'Austen',
+          dateOfBirth: '9/1/1980',
+          vaFileNumber: '111223333'
+        }
+      }
+    );
+
+    expect(state.enrollmentData.firstName).to.eql('Jane');
```
raquelromano-gov: Can you just use `to.equal` here?  I wasn't familiar with `eql` but it looks like ti does a deep comparison.

raquelromano-gov (approved):
Just one question below

test/post-911-gib-status/reducers/index.unit.spec.js
annekainicUSDS: Ah interesting, didn't realize that. What's the problem with a deep comparison?

PR #5829, opened 2017-06-21
Fix missing bracket in wizard page

jbalboni:
Safari is throwing an error, not sure why other browsers allowed it.


cehsu (approved):
LGTM.

cvalarida (approved):

PR #5833, opened 2017-06-22
Appeals status page

bshyong:
Some refactoring and more changes to reach parity with mockup. Note that content is not finalized yet in screenshot below

screen shot 2017-06-21 at 6 03 31 pm


src/js/disability-benefits/utils/appeal-helpers.jsx
```diff
@@ -0,0 +1,64 @@
+import React from 'react';
+
+export const appealStatusDescriptions = (eventType) => {
```
U-DON: Why change this to a function?

U-DON (approved):
LGTM.

src/js/disability-benefits/utils/appeal-helpers.jsx
bshyong: Do react nodes behave properly if this is a javascript object?  I made this a functional component which seemed to accomplish what I needed

U-DON: I thought they would. You're still using them inside an object in the function. I don't know if wrapping them within a function is a requirement for React nodes. But I'm not opposed to doing it either way.

bshyong: Object worked—in addition to importing React, needed to change the file extension to `jsx`

PR #5835, opened 2017-06-22
Update burials endpoint and use newer confirmation page layout

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3157

I verified that the review page looks ok and updated the endpoint so we can submit (once the backend schema is updated to accept the privacy agreement).

We have the fields for the confirmation page, but it's enough work that I'm going to leave it as a separate item.

screen shot 2017-06-21 at 11 34 17 pm


src/js/burials/containers/ConfirmationPage.jsx
```diff
@@ -15,29 +15,53 @@ const scrollToTop = () => {
...
 
+  toggleExpanded = (e) => {
+    e.preventDefault();
+    this.setState({ isExpanded: !this.state.isExpanded });
+  }
+
   render() {
-    const time = this.props.form.submission.response.timestamp;
+    const form = this.props.form;
+    // const response = this.props.form.submission.response
```
cvalarida: Testing leftovers or explanation of something?

src/js/burials/containers/ConfirmationPage.jsx
```diff
@@ -15,29 +15,53 @@ const scrollToTop = () => {
...
+      <div className="edu-benefits-submit-success">
+        <h3 className="burial-page-title">Claim received</h3>
+        <p>Normally processed within <strong>30 days</strong></p>
+        <p>
+          We may contact you for more information or documents.<br/>
+          <i>Please print this page for your records.</i>
+        </p>
+        <div className="inset">
+          <h4>Burial Benefit Claim <span className="additional">(Form 21P-530)</span></h4>
+          <span>for {name.first} {name.middle} {name.last} {name.suffix}</span>
```
cvalarida: Does the `<span>` take out the extra space between first and last if there is no middle name? Looks like it in your screenshot, but I wanted to confirm.

cvalarida (approved):
I'm on board. 👍
After being confused that the connected issue seemed more about the document upload in the review page and this had nothing to do with that, I found that bit was split into a different ticket, so... 🚀

cehsu (approved):

src/js/burials/containers/ConfirmationPage.jsx
jbalboni: It'll be used in the future, when we add more info.

src/js/burials/containers/ConfirmationPage.jsx
jbalboni: Yeah, multiple spaces are collapsed, I believe.

src/js/burials/helpers.jsx
```diff
@@ -20,3 +22,13 @@ export function fileHelp({ formContext }) {
     </p>
   );
 }
+
+export function transform(formConfig, form) {
+  // delete form.data.privacyAgreementAccepted;
```
cehsu: Another potential zombie comment as already noted elsewhere by @cvalarida

PR #5836, opened 2017-06-22
Avoid presenting a no-op modal for terms and conditions

patrickvinograd:
Near-term solution for https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3452

After this is beyond beta we can switch to a modal that shows the accepted terms, but for now that would be misleading for non-beta users.


bshyong (approved):
LGTM

PR #5839, opened 2017-06-22
Point to pensions endpoint and add skeletal confirmation page

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3333

Same as the burials endpoint issue.

screen shot 2017-06-22 at 4 14 34 pm

### DIFF
--- a/src/js/pensions/helpers.jsx
+++ b/src/js/pensions/helpers.jsx
@@ -2,8 +2,6 @@ import React from 'react';
 import { transformForSubmit } from '../common/schemaform/helpers';
 
 export function transform(formConfig, form) {
-  // delete form.data.privacyAgreementAccepted;
-  // delete form.data.hasVisitedVAMC;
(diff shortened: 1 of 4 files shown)


src/js/pensions/helpers.jsx
```diff
@@ -4,7 +4,9 @@ import { transformForSubmit } from '../common/schemaform/helpers';
 export function transform(formConfig, form) {
   const formData = transformForSubmit(formConfig, form);
   return JSON.stringify({
-    form: formData
+    pensionClaim: {
```
cvalarida: I'll assume this is correct, but I guess I don't actually know.

cvalarida (approved):

src/js/pensions/helpers.jsx
jbalboni: It is, it doesn't work without it.

PR #5842, opened 2017-06-22
Disability Benefits beta entry point

bshyong:
This is a big PR but essentially a new entry point was created, and the current disability benefits folder was copied over so that we can modify the content pages without having to do crazy feature flagging.

The `/track-claims` app in the beta entry point will point to the existing disability benefits app because the only feature flag we need to do there is on the call to fetch appeals data.


U-DON (approved):
LGTM.  Just a content merge conflict to resolve.

PR #5843, opened 2017-06-22
Remove other field from netWorth

cehsu:
Connects to department-of-veterans-affairs/vets.gov-team#3490

screen shot 2017-06-22 at 4 30 56 pm


jbalboni (approved):

cvalarida (commented):
test/pensions/config/.financialDisclosure.unit.spec.jsx.swp added but shouldn't be. Remove that and it looks golden, though.

PR #5845, opened 2017-06-23
Pensions employment history question

cvalarida:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3489

image

image


cvalarida:
I was looking for the `employmentHistory` unit test and didn't find one, but I did find that `powAndSeverance` wasn't a `.unit.spec`, so I renamed it while I was here.

jbalboni:
@cvalarida That description in jsx in the helpers file, so you could add a class to the element and add some margin.

test/pensions/config/employmentHistory.unit.spec.jsx
```diff
@@ -0,0 +1,113 @@
...
+    const form = ReactTestUtils.renderIntoDocument(
+      <DefinitionTester
+          schema={schema}
+          definitions={formConfig.defaultDefinitions}
+          onSubmit={onSubmit}
+          uiSchema={uiSchema}/>
+    );
+
+    const formDOM = getFormDOM(form);
+    formDOM.setYesNo('#root_view\\:workedBeforeDisabledYes', true);
```
jbalboni: This doesn't seem notably better than `formDOM.fillData('#root_view\\:workedBeforeDisabledYes', 'Y');`

test/pensions/config/employmentHistory.unit.spec.jsx
```diff
@@ -0,0 +1,113 @@
...
+          uiSchema={uiSchema}/>
+    );
+    const formDOM = getFormDOM(form);
+
+    formDOM.setYesNo('#root_view\\:workedBeforeDisabledYes', true);
+
+    expect(formDOM.querySelectorAll('input,select').length).to.equal(18);
+  });
+
+  it('should not submit empty form', () => {
```
jbalboni: Can you add a test to make sure you can submit if you choose No to the first question?

yarn.lock
```diff
@@ -7787,7 +7787,7 @@ supports-color@1.2.0:
   version "1.2.0"
   resolved "https://registry.yarnpkg.com/supports-color/-/supports-color-1.2.0.tgz#ff1ed1e61169d06b3cf2d588e188b18d8847e17e"
 
-supports-color@3.1.2, supports-color@^3.1.0, supports-color@^3.1.1:
```
jbalboni: Doesn't seem like we need these lock file changes.

test/burials/config/employmentHistory.unit.spec.jsx
```diff
@@ -1,153 +0,0 @@
-import React from 'react';
```
jbalboni: lol, oops

jbalboni (approved):
Looks good. Made a comment about adding another test and fixing spacing.

yarn.lock
cvalarida: Yeah, I don't know what the deal is with these. I had to `yarn install` to get the correct version of vets-json-schema, but that also updated the lockfile...Usually I just ignore it, but...I guess I didn't this time. I'll change that back.

test/pensions/config/employmentHistory.unit.spec.jsx
cvalarida: Not really, no, but I made the ability for it to accept either `'Y'` or `true`.

cvalarida:
Updated the spacing.

PR #5846, opened 2017-06-23
Letters add main component

annekainicUSDS:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3431
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3462

This PR does a few things:
1. Adds a Main component to only fetch the data
2. Change url from `/va-letters` to `/letters`
3. Change any reference to `va-letters` or `VALetters` to `letters` and `Letters`, respectively, to be in line with the new url change
4. Add additional content to the bottom of the page below the collapsible panels (based on the mockups)
5. Tweak collapsible panel styles to be wider and shorter (to be closer in line with mocks)

Merging with open PR for now since I branched off that PR. Will switch to merge into master once that PR is merged.

screen shot 2017-06-23 at 12 26 36 pm


raquelromano-gov (approved):
It looks like the routes and style files got deleted/recreated instead of renamed, so we might lose some change history, but I'm OK with that if you are (not that much has happened to those files)

PR #5848, opened 2017-06-23
Add other expenses pages

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3484

Adds the expenses pages and also tweaks their handling on the review page. Arrays inside of arrays are a pain, so this one just displays a read-only version of the array and forces you to click Edit to add/edit rows. Ongoing conversation about this with Alex.

screen shot 2017-06-23 at 2 19 06 pm
screen shot 2017-06-23 at 2 19 55 pm


src/js/common/schemaform/SchemaForm.jsx
```diff
@@ -26,7 +27,7 @@ const fields = {
 
 const reviewFields = {
   ObjectField: ReviewObjectField,
-  ArrayField,
+  ArrayField: ReadOnlyArrayField,
```
cvalarida: Does this mean that `ArrayField`s never get used in review now?

src/js/common/schemaform/review/ReadOnlyArrayField.jsx
```diff
@@ -0,0 +1,92 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+
+import {
+  toIdSchema,
+  deepEquals
+} from 'react-jsonschema-form/lib/utils';
+
+class BasicArrayField extends React.Component {
```
cvalarida: Probably call this `ReadOnlyArrayField`, yeah?

src/js/common/schemaform/review/ReadOnlyArrayField.jsx
```diff
@@ -0,0 +1,92 @@
...
+}
+
+BasicArrayField.propTypes = {
+  schema: PropTypes.object.isRequired,
+  uiSchema: PropTypes.object,
+  errorSchema: PropTypes.object,
+  requiredSchema: PropTypes.object,
+  idSchema: PropTypes.object,
+  onChange: PropTypes.func.isRequired,
+  onBlur: PropTypes.func,
```
cvalarida: `onChange` and `onBlur` aren't used.

yarn.lock
```diff
@@ -7787,7 +7787,7 @@ supports-color@1.2.0:
   version "1.2.0"
   resolved "https://registry.yarnpkg.com/supports-color/-/supports-color-1.2.0.tgz#ff1ed1e61169d06b3cf2d588e188b18d8847e17e"
 
-supports-color@3.1.2, supports-color@^3.1.0, supports-color@^3.1.1:
```
cvalarida: Always with these bloody....Eh, whatever. :man_shrugging:

cvalarida (approved):

cehsu (approved):

src/js/common/schemaform/review/ReadOnlyArrayField.jsx
cehsu: same for `requiredSchema`?

src/js/common/schemaform/review/ReadOnlyArrayField.jsx
```diff
@@ -0,0 +1,92 @@
...
+    return (
+      <div className="schemaform-field-container">
+        <div>
+          {items.map((item, index) => {
+            const itemSchema = this.getItemSchema(index);
+            const itemIdPrefix = `${idSchema.$id}_${index}`;
+            const itemIdSchema = toIdSchema(itemSchema, itemIdPrefix, definitions);
+
+            return (
+              <div key={index} className="va-growable-background">
```
cehsu: Just want to make sure using index as key here won't be an issue as items won't be re-orderable, etc.

src/js/pensions/components/DisclosureTitle.jsx
```diff
@@ -8,7 +8,7 @@ export default function createDisclosureTitle(path, title) {
     // A little hackish
     if (path === 'spouse') {
       const marriages = formData.marriages || [];
-      fullName = _.get(['marriages', marriages.length - 1, 'spouseFullName'], formData);
+      fullName = _.get(['marriages', marriages.length - 1, 'spouseFullName'], formData) || {};
```
cehsu: Is specifying a default value for `_.get` an option here instead of after the `||`?

src/js/pensions/components/ExpenseField.jsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import moment from 'moment';
+
+export default function ExpenseField({ formData }) {
+  const noData = !formData.amount && !formData.date;
+
+  return noData
+      ? <div><strong>Expense</strong></div>
+      : <div><strong>${formData.amount}</strong><br/>{formData.date && moment(formData.date).format('MMM D, YYYY')}</div>;
```
cehsu: Elegant handling of lack of required fields.

yarn.lock
cehsu: `git add -p` might come in handy here.

src/js/common/schemaform/SchemaForm.jsx
jbalboni: No, what we've been doing is pulling array fields out of the schema and displaying them separately. That's still a thing and we use the review version of ArrayField there. 

This change means that, if you don't do that (with the `keepArrayInPageOnReview` option), you get this component in view mode.

src/js/common/schemaform/review/ReadOnlyArrayField.jsx
jbalboni: Yep!

src/js/common/schemaform/review/ReadOnlyArrayField.jsx
jbalboni: Right, the items won't be reordered when this component is mounted.

src/js/pensions/components/DisclosureTitle.jsx
jbalboni: I don't think the `lodash/fp` version has the default value parameter (all fp versions of functions have a fixed number of parameters).

J-Quag:
Can we do "Paid to:" as the bolded in the saved expense with the dollar amount underneath?
  When they are put together you can only see the dollar amount and date, but not what they were for. 
image

Would like to see something like this. Though I dont consider this a blocker. I can make a separate ticket for a change after we push it out. 

image

PR #5854, opened 2017-06-24
Download letter

raquelromano-gov:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3465

This is clearly flawed, but merging so we can continue developing. Launch-blocking issues are documented here: https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3597

### DIFF
--- a/src/js/letters/components/DownloadLetterLink.jsx
+++ b/src/js/letters/components/DownloadLetterLink.jsx
@@ -51,12 +51,14 @@ class DownloadLetterLink extends React.Component {
               downloadWindow.location.href = downloadUrl;
             }
             document.body.removeChild(link);
-          } else if (ie10) {
-            window.navigator.msSaveOrOpenBlob(blob, this.props.letterName);
-          } else {
-            // What should we do about Opera Mini?
+            urlObj.revokeObjectURL(downloadUrl);
+            return Promise.resolve();
           }
-          urlObj.revokeObjectURL(downloadUrl);
+          if (ie10) {
+            window.navigator.msSaveOrOpenBlob(blob, this.props.letterName);
+            return Promise.resolve();
+          }
+          return Promise.reject(new Error('DownloadLetterLink: could not download PDF'));
         });
       });
   }
--- a/src/js/letters/components/DownloadLetterLink.jsx
+++ b/src/js/letters/components/DownloadLetterLink.jsx
@@ -5,16 +5,6 @@ import { Link } from 'react-router';
 
 import { apiRequest } from '../utils/helpers';
 
-const Modernizr = require('modernizr');
-if (!Modernizr.bloburls) {
-  require('classlist-polyfill'); // DOM element classList support.
-}
-if (!Modernizr.adownload) {
-  require('dataset');  // dataSet accessor support.
-}
-
-// const FileSaver = require('file-saver');
-
 class DownloadLetterLink extends React.Component {
   constructor(props) {
     super(props);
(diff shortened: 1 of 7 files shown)


src/js/letters/components/DownloadLetterLink.jsx
```diff
@@ -0,0 +1,80 @@
...
+            const link = document.createElement('a');
+            if (typeof link.download !== 'undefined') {
+              document.body.appendChild(link);
+              link.style = 'display: none';
+              link.href = downloadUrl;
+              link.target = '_blank';
+              link.download = this.props.letterName;
+              link.click();
+            } else {
+              downloadWindow.location.href = downloadUrl;
```
annekainicUSDS: Because here you're just using the window you already opened...

annekainicUSDS (approved):
This works for me locally! Had a few questions related to the duplicate window opening up. I can see the comments that show the additional work/clean-up to be done, but I agree that can be tackled in a future PR so as not to hold up this one.

src/js/letters/components/DownloadLetterLink.jsx
```diff
@@ -0,0 +1,80 @@
...
+        response.blob().then(blob => {
+          // Blob URLs are not supported on Opera Mini or IE10 and below
+          const urlObj = window.URL || window.webkitURL;
+          const downloadUrl = urlObj.createObjectURL(blob);
+          const blobSupported = !!(/^blob:/.exec(downloadUrl));
+          if (blobSupported) {
+            // This is hack to give the file a nice name instead of an
+            // ugly hash, but is not supported on IE11 or iOS Safari.
+            const link = document.createElement('a');
+            if (typeof link.download !== 'undefined') {
```
raquelromano-gov: No, this is to check if the browser supports the `download` attribute for the `<a>` tag. I'll add a comment to that effect if we decide to keep this in (to give the file a readable name).

src/js/letters/components/DownloadLetterLink.jsx
```diff
@@ -0,0 +1,80 @@
...
+              document.body.appendChild(link);
+              link.style = 'display: none';
+              link.href = downloadUrl;
+              link.target = '_blank';
+              link.download = this.props.letterName;
+              link.click();
```
raquelromano-gov: Yeah, but I was thinking that this is the blank window that opens to download the file, which is the one we're OK with. But now that you mention it, maybe I'm wrong, because it makes more sense that this one would be blocked since it's in the callback instead of directly in the click handler. Hmm...! Will investigate how we might do this outside of the callback. Thanks!

src/js/letters/components/DownloadLetterLink.jsx
raquelromano-gov: Yep, I think you're right.

src/js/letters/components/DownloadLetterLink.jsx
```diff
@@ -0,0 +1,144 @@
...
+
+import { apiRequest } from '../utils/helpers';
+
+class DownloadLetterLink extends React.Component {
+  constructor(props) {
+    super(props);
+    this.downloadLetter = this.downloadLetter.bind(this);
+  }
+
+  // This opens a duplicate window for some inexplicable reason,
```
raquelromano-gov: @annekainicUSDS This experiment was mystifying. Even without opening a window or creating the blob URL, simply issuing a POST  with an empty response opens a new window. I don't see anything in the blob() docs that explains why this would happen, eg.  https://developer.mozilla.org/en-US/docs/Web/API/Body/blob or https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch

PR #5856, opened 2017-06-24
Save in progress prefill UX update and analytics events

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3461
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3327

screen shot 2017-06-23 at 11 31 24 pm


src/js/common/schemaform/FormApp.jsx
```diff
@@ -30,9 +30,9 @@ class FormApp extends React.Component {
       newProps.router.push(newProps.returnUrl);
       // Set loadedStatus in redux to not-attempted to not show the loading page
       newProps.setFetchFormStatus(LOAD_STATUSES.notAttempted);
-    } else if (newProps.loadedStatus === LOAD_STATUSES.prefillComplete) {
+    } else if (newProps.loadedStatus !== this.props.loadedStatus
```
jbalboni: This seems safer.

src/js/common/schemaform/FormIntroButtons.jsx
```diff
@@ -24,16 +24,16 @@ class FormIntroButtons extends React.Component {
 
   handleLoadPrefill = () => {
     if (this.props.prefillAvailable) {
-      this.handleLoadForm(true);
+      this.props.fetchInProgressForm(this.props.formId, this.props.migrations, true);
```
jbalboni: handleLoadForm is always passed an event as an argument by `ProgressButton`, so this works better for passing the prefill parameter.

src/js/common/schemaform/reducers/index.js
```diff
@@ -157,10 +159,22 @@ export default function createSchemaFormReducer(formConfig) {
         return _.set('loadedStatus', action.status, state);
       }
       case SET_IN_PROGRESS_FORM: {
-        const newState = _.set('loadedData', action.data, state);
+        let newState;
+
+        // if we're prefilling, we want to use whatever initial data the form has
+        if (action.prefilled) {
```
jbalboni: Currently, prefilling blows away all data, when it should really be merging with our default data. In progress forms can just replace, because we're saving all the default data along with them.

src/js/hca/config/form.js
```diff
@@ -159,6 +159,9 @@ const formConfig = {
             }),
             mothersMaidenName: {
               'ui:title': 'Mother’s maiden name'
+            },
+            'ui:options': {
+              showPrefillMessage: true
```
jbalboni: This seems easier than trying to guess what the first page is.

J-Quag:
Which events? When we prefill?

jbalboni:
@J-Quag I wrote that before you answered my question about the prefix. All sip and prefill events will have a form specific prefix.

src/js/common/schemaform/save-load-actions.js
```diff
@@ -111,12 +109,16 @@ export function saveInProgressForm(formId, version, returnUrl, formData) {
     }),
     formData: JSON.stringify(formData)
   });
-  return dispatch => {
+  return (dispatch, getState) => {
```
cvalarida: Adding to my TODO list: Learn more about redux thunk...

I didn't know you could also have `getState`. Handy!

src/js/common/schemaform/SchemaForm.jsx
```diff
@@ -56,6 +56,8 @@ class SchemaForm extends React.Component {
       this.setState({ formContext: _.set('pageTitle', newProps.title, this.state.formContext) });
     } else if (!!newProps.reviewMode !== !!this.state.formContext.reviewMode) {
       this.setState(this.getEmptyState(newProps));
+    } else if (!!newProps.prefilled !== !!this.state.formContext.prefilled) {
```
cvalarida: Heh, this might be a new record for the number of useful `!`s I've seen in an `if` statement. 😛

Curiosity, though: Doing `!!` will force it to a boolean, which means that if we get something really bizarre like `newProps.prefilled = 'false'` and `formContext.prefilled = true`, this `if`'s condition will be met when it shouldn't. That _shouldn't_ be a problem, but then, you're forcing a boolean because of defensive programming, right? Just in case we get a weird value?

cvalarida (approved):

src/js/common/schemaform/SchemaForm.jsx
jbalboni: Yeah, honestly, it's not really necessary in these if statements. It's more relevant if you're setting a value.

src/js/common/schemaform/FormApp.jsx
jbalboni: Yep!

cehsu (approved):

PR #5858, opened 2017-06-24
Show modal and delete form when starting over

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3371
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3301

There are a couple weird paths that make these changes make more sense. One is what happens when you get an error when starting over. I don't think we should put user who wants to start over on the error page, since they were explicitly trying to start the form without their saved data, regardless of if we have a problem actually deleting it. So I've done that, with the exception of 401 errors, in which case they're asked to sign in again. A similar thing happens with prefill; if they run into any errors, just put them on the form.

screen shot 2017-06-24 at 4 28 06 pm


jbalboni:
Tests are there now.

cehsu (approved):

test/common/schemaform/FormApp.unit.spec.jsx
```diff
@@ -83,4 +84,104 @@ describe('Schemaform <FormApp>', () => {
...
+      </FormApp>
+    );
+
+    tree.getMountedInstance().componentWillReceiveProps({
+      prefillStatus: PREFILL_STATUSES.complete,
+      router,
+      routes
+    });
+
+    expect(router.push.called).to.be.true;
```
cvalarida: Should we check to see what it was called with? Just in case it's called with `/error`, for example.

test/common/schemaform/reducers/index.unit.spec.js
```diff
@@ -214,15 +223,58 @@ describe('schemaform createSchemaFormReducer', () => {
...
       expect(state.data.existingProp).to.be.true;
       expect(state.data.field).to.equal('foo');
     });
+    it('should set fetch form pending', () => {
+      const state = reducer({
+      }, {
+        type: SET_FETCH_FORM_PENDING,
+      });
+
+      expect(state.loadedStatus).to.equal(LOAD_STATUSES.pending);
```
cvalarida: Should we also check to see if `!prefill`?

cvalarida (approved):

src/js/common/schemaform/FormIntroButtons.jsx
```diff
@@ -36,28 +42,54 @@ class FormIntroButtons extends React.Component {
...
+  startOver = () => {
+    this.toggleModal();
+    this.props.removeInProgressForm(this.props.formId, this.props.migrations);
+  }
+
   render() {
-    let resumeButton = null;
-    let firstPageButtonText = 'Continue';
-    let firstPageButtonClass = 'usa-button-primary';
     if (this.props.formSaved) {
```
jbalboni: Yeah, I agree. Not sure about a good name, though.

src/js/common/schemaform/reducers/index.js
```diff
@@ -147,35 +153,61 @@ export default function createSchemaFormReducer(formConfig) {
       }
       case SET_SAVE_FORM_STATUS: {
         const newState = _.set('savedStatus', action.status, state);
+        newState.startingOver = false;
+        newState.prefillStatus = PREFILL_STATUSES.notAttempted;
 
         // This is the only time we have a saved datetime
         if (action.status === SAVE_STATUSES.success) {
-          return _.set('lastSavedDate', action.lastSavedDate, newState);
+          newState.lastSavedDate = action.lastSavedDate;
```
jbalboni: Right, it doesn't change functionality. I doubt we'll see much of a difference, but it bugs me to do a bunch of these lodash fp functions in a row, since it creates new objects each time.

src/js/common/schemaform/reducers/index.js
```diff
@@ -147,35 +153,61 @@ export default function createSchemaFormReducer(formConfig) {
...
+      case SET_START_OVER: {
+        return _.assign(state, {
+          isStartingOver: true,
+          data: state.initialData,
+          loadedStatus: LOAD_STATUSES.pending
+        });
+      }
+      case SET_PREFILL_COMPLETE: {
+        return _.assign(state, {
+          prefillStatus: PREFILL_STATUSES.complete,
```
jbalboni: Yeah, it might be better to change that complete status to failed or something. It initially covered both success and failure.

src/js/common/schemaform/save-load-actions.js
```diff
@@ -282,3 +309,52 @@ export function fetchInProgressForm(formId, migrations, prefill = false) {
...
   };
 }
+
+export function removeInProgressForm(formId, migrations) {
+  return (dispatch, getState) => {
+    const userToken = sessionStorage.userToken;
+    const trackingPrefix = getState().form.trackingPrefix;
+    // If we don't have a userToken, fail safely
+    if (!userToken) {
+      dispatch(setFetchFormStatus(LOAD_STATUSES.noAuth));
```
jbalboni: Yeah, you shouldn't get here, it's carried over from the other actions. Might be able to take it out here, since it seems pretty unlikely (and would get caught with a 401 response later).

src/js/common/schemaform/reducers/index.js
```diff
@@ -147,35 +153,61 @@ export default function createSchemaFormReducer(formConfig) {
...
         return newState;
       }
       case SET_FETCH_FORM_STATUS: {
         return _.set('loadedStatus', action.status, state);
       }
+      case SET_FETCH_FORM_PENDING: {
+        const newState = _.set('loadedStatus', LOAD_STATUSES.pending, state);
+
+        if (action.prefill) {
+          newState.prefillStatus = PREFILL_STATUSES.pending;
```
jbalboni: I think that makes the form status action confusing, since it's directly specifying that we're changing the form status.

This is something I don't like about how we use Redux actions here (which is much my fault as anyone's): they're often focused on changing a specific piece of state instead of relating to an actual event that has happened. What really happened here is we started the process of fetching a form; the changing form status is a consequence of that. I probably should name this something like `START_FETCH_FORM`, but that felt like straying too far from the existing Redux actions.

src/js/common/schemaform/reducers/index.js
```diff
@@ -147,35 +153,61 @@ export default function createSchemaFormReducer(formConfig) {
...
         // if we're prefilling, we want to use whatever initial data the form has
-        if (action.prefilled) {
+        if (state.prefillStatus === PREFILL_STATUSES.pending) {
           const formData = _.merge(state.data, action.data.formData);
           const loadedData = _.set('formData', formData, action.data);
           newState = _.set('loadedData', loadedData, state);
+          newState.prefillStatus = PREFILL_STATUSES.success;
         } else {
           newState = _.set('loadedData', action.data, state);
+          newState.prefillStatus = PREFILL_STATUSES.notAttempted;
```
jbalboni: I think: start new form with prefillled info -> save form -> resume form

src/js/common/schemaform/FormIntroButtons.jsx
jbalboni: I'm going to save this for the last open SiP PR, so I don't cause a bunch of merge conflicts.

J-Quag:
@jbalboni 
Burial form looks good on this, but the Pension Form: 
image

jbalboni:
@J-Quag Good catch. I pushed a fix to that to the open PR I have for SiP.

PR #5869, opened 2017-06-26
Only allow burial dates on or after date of death

jbalboni:
Found this issue in testing.

screen shot 2017-06-26 at 12 12 12 pm


src/js/burials/validation.js
```diff
@@ -0,0 +1,8 @@
+import moment from 'moment';
+
+export function validateBurialDate(errors, page) {
+  const { burialDate, deathDate } = page;
+  if (burialDate && deathDate && moment(burialDate).isBefore(moment(deathDate))) {
+    errors.burialDate.addError('Date of burial must be on or before date of death');
```
cvalarida: Mmm...Is this right?

cvalarida (commented):
Validation text seems backwards, but it looks good other than that.

src/js/burials/validation.js
```diff
@@ -0,0 +1,8 @@
+import moment from 'moment';
+
+export function validateBurialDate(errors, page) {
```
cehsu: Will this be expanded to also check end of service?

cehsu (approved):

src/js/burials/validation.js
jbalboni: I think we would write separate validation functions for each one, though it might make sense to write a generic function for it (we have a date range validator, but it assumes particular date field names).

That the date is in the past should be checked by the validation included in the `currentOrPastDateUI` definition.

PR #5878, opened 2017-06-27
Update SiP error copy on connection errors, usability tweaks

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3552

This updates the error copy when saving a form the above text. I also made some focus/scroll handling tweaks to make the form work better if you're on a screen reader or just using a keyboard.

screen shot 2017-06-27 at 11 37 39 am
screen shot 2017-06-27 at 11 34 06 am

### DIFF
--- a/src/js/common/schemaform/FormApp.jsx
+++ b/src/js/common/schemaform/FormApp.jsx
@@ -17,7 +17,7 @@ const scrollToTop = () => {
   scroller.scrollTo('topScrollElement', window.VetsGov.scroll || {
     duration: 500,
     delay: 0,
-    smooth: true,
+    smooth: true
   });
 };
 
--- a/src/js/common/schemaform/FormApp.jsx
+++ b/src/js/common/schemaform/FormApp.jsx
@@ -6,12 +6,20 @@ import { connect } from 'react-redux';
 import FormNav from './FormNav';
 import FormTitle from './FormTitle';
 import AskVAQuestions from './AskVAQuestions';
-import { LOAD_STATUSES, PREFILL_STATUSES, setFetchFormStatus } from './save-load-actions';
+import { LOAD_STATUSES, PREFILL_STATUSES, SAVE_STATUSES, setFetchFormStatus } from './save-load-actions';
 import LoadingIndicator from '../components/LoadingIndicator';
 
 import { isInProgress } from '../utils/helpers';
 
 const Element = Scroll.Element;
+const scroller = Scroll.scroller;
+const scrollToTop = () => {
+  scroller.scrollTo('topScrollElement', window.VetsGov.scroll || {
+    duration: 500,
+    delay: 0,
+    smooth: true,
+  });
+};
 
 /*
  * Primary component for a schema generated form app.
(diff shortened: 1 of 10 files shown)


src/js/common/schemaform/FormPage.jsx
```diff
@@ -48,15 +48,9 @@ class FormPage extends React.Component {
...
 
   componentDidMount() {
-    scrollToTop();
-    focusForm();
-  }
-
-  // A successful form save will mean we go from pending to success,
-  // and we need to redirect
-  componentWillReceiveProps(newProps) {
-    if (this.props.form.savedStatus === SAVE_STATUSES.pending && newProps.form.savedStatus === SAVE_STATUSES.success) {
```
jbalboni: Figured I would consolidate this sort of logic in FormApp

src/sass/base/_b-utils.scss
```diff
@@ -5,3 +5,43 @@
 .no-background-image {
   background-image: none;
 }
+
+.va-button-link {
```
jbalboni: This is ugly, but we need a way to style buttons as links, since we're using the link element design for actions in places.

cvalarida: Why can't we just use anchors instead of buttons? Accessibility reasons?

cvalarida (approved):

src/js/common/schemaform/FormApp.jsx
```diff
@@ -6,12 +6,20 @@ import { connect } from 'react-redux';
...
 
 import { isInProgress } from '../utils/helpers';
 
 const Element = Scroll.Element;
+const scroller = Scroll.scroller;
+const scrollToTop = () => {
+  scroller.scrollTo('topScrollElement', window.VetsGov.scroll || {
+    duration: 500,
+    delay: 0,
+    smooth: true,
```
cehsu: extra comma here?

src/js/common/schemaform/SaveFormLink.jsx
```diff
@@ -1,8 +1,20 @@
...
+import { SAVE_STATUSES, saveErrors } from './save-load-actions';
+import { focusElement } from '../utils/helpers';
+
+const Element = Scroll.Element;
+const scroller = Scroll.scroller;
+const scrollToTop = () => {
+  scroller.scrollTo('saveFormLinkTop', window.VetsGov.scroll || {
+    duration: 500,
+    delay: 0,
+    smooth: true,
```
cehsu: again here

src/js/common/schemaform/SaveFormLink.jsx
```diff
@@ -63,9 +72,18 @@ class SaveFormLink extends React.Component {
             user={this.props.user}
             onUpdateLoginUrl={this.props.onUpdateLoginUrl}
             onLogin={this.saveFormAfterLogin}/>
-        {savedStatus === SAVE_STATUSES.failure &&
-          <div className="usa-alert usa-alert-error no-background-image schemaform-save-error">We’re sorry, but something went wrong. Please try saving your application again.</div>}
-        {content}
+        {saveErrors.has(savedStatus) &&
+          <div role="alert" className="usa-alert usa-alert-error no-background-image schemaform-save-error">
```
cehsu: Is it now common practice for alerts to not include the background image (was this an exclamation mark before)? It seems appropriate here?

cehsu (approved):

src/sass/base/_b-utils.scss
jbalboni: Yeah, anchors without `href` attributes can't be tabbed to by default, plus `a` tags are meant for links, not actions.

src/js/common/schemaform/SaveFormLink.jsx
```diff
@@ -1,8 +1,20 @@
...
 import PropTypes from 'prop-types';
 
 import LoginModal from '../components/LoginModal';
-import { SAVE_STATUSES } from './save-load-actions';
+import { SAVE_STATUSES, saveErrors } from './save-load-actions';
+import { focusElement } from '../utils/helpers';
+
+const Element = Scroll.Element;
+const scroller = Scroll.scroller;
+const scrollToTop = () => {
```
jbalboni: Yeah, it's tempting, but we normally need to put an `Element` on the page, too, and you're not saving a whole lot if you still have to important react-scroll wherever you're using it.

src/js/common/schemaform/SaveFormLink.jsx
jbalboni: It seems like we're inconsistent about that. Something worth bringing up to the designers, since the icon is the default styling from uswds, but in certain spots where the text in the alert doesn't have paragraph or line breaks this more compact format seems to fit better.

src/js/common/schemaform/SaveFormLink.jsx
cvalarida: That's what I was thinking.

PR #5882, opened 2017-06-27
Add e2e tests for HCA save in progress

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3526

End to end tests for SiP. Pretty limited, since we can only do so much with the mocks.


cvalarida (approved):
Not sure I completely understand the `initSaveInProgress` helper, but if it works the way I think it does, then it looks good. We may circle back on this after June to see if we can test it more thoroughly, though.

cehsu (approved):

PR #5884, opened 2017-06-27
Bold purpose on top and move regular weight amount to line two

cehsu:
Connects to department-of-veterans-affairs/vets.gov-team#3536

screen shot 2017-06-27 at 1 02 10 pm

### DIFF
--- a/src/js/pensions/components/ExpenseField.jsx
+++ b/src/js/pensions/components/ExpenseField.jsx
@@ -1,10 +1,9 @@
 import React from 'react';
-import moment from 'moment';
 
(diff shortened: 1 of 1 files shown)


src/js/pensions/components/ExpenseField.jsx
```diff
@@ -1,10 +1,9 @@
 import React from 'react';
-import moment from 'moment';
 
 export default function ExpenseField({ formData }) {
   const noData = !formData.amount && !formData.date;
```
jbalboni: We should change date to purpose here.

jbalboni (approved):

PR #5885, opened 2017-06-27
Pass the correct parameter to getActivePages

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3555

Yep. This is correct in `FormPage`, but not in `ReviewPage`.


cehsu (approved):

cvalarida (approved):
D'oh!
Good catch.

PR #5893, opened 2017-06-28
Create GetFormHelp for pensions and burials

cehsu:
Connects to department-of-veterans-affairs/vets.gov-team#3428
screen shot 2017-06-28 at 2 26 28 pm


src/js/pensions/components/GetFormHelp.jsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+
+function GetFormHelp() {
+  return (
+    <div>
+      <p className="talk">For benefit related questions please call:</p>
+      <p className="phone-number">
+        <a href="tel:+800-827-100">800-827-100</a><br/>
```
cehsu: It looks like we've been including the `+` here which is not required, but not the country code that would go along with it. Should we include the country code? or remove the `+`?

src/js/burials/components/GetFormHelp.jsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+
+function GetFormHelp() {
+  return (
+    <div>
+      <p className="talk">For benefit related questions please call:</p>
+      <p className="phone-number">
+        <a href="tel:+1-800-827-100">800-827-100</a><br/>
```
jbalboni: We probably want the hours, too? I think they're on the confirmation page.

jbalboni: Also, I think you're missing a number in here somewhere.

src/js/pensions/components/GetFormHelp.jsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+
+function GetFormHelp() {
```
jbalboni: I think we could use the same component for pension and burials; it's the same call center for both.

jbalboni (approved):
A couple minor things, and you should get Josh's ok, but looks good to me.

src/js/burials/components/GetFormHelp.jsx
cehsu: Thanks @jbalboni, I thought so too but then saw this number referenced on several websites. @J-Quag can you please confirm that this is the correct number?

cehsu: Thanks @jbalboni, I thought so too but then saw this number referenced on several websites. @J-Quag can you please confirm that this is the correct number?

cehsu: Thanks @jbalboni, I found these on the burials confirmation page and will include this information here, should they also be on the pensions confirmation page as well?

src/js/pensions/components/GetFormHelp.jsx
J-Quag: yes we can

src/js/burials/components/GetFormHelp.jsx
J-Quag: sorry its 1000, probably did a copy/paste jumble

J-Quag: 8am to 9pm est. It'll match burials as Jeff said.

J-Quag (commented):
Just needs the extra 0 on the phone number.

src/js/pensions/components/GetFormHelp.jsx
cehsu: Thanks @jbalboni, have moved the shared component to common/schemaform folder and renamed as `GetPensionOrBurialFormHelp`, please let me know if another name or location for this would be better.

src/js/common/schemaform/GetPensionOrBurialFormHelp.jsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+
+function GetFormHelp() {
+  return (
+    <div>
+      <p className="talk">For benefit related questions please call:</p>
+      <p className="phone-number">
+        <a href="tel:+1-800-827-1000">800-827-1000</a><br/>
```
cehsu: Would it make more sense to have these match?

cvalarida (approved):
👍 🚀

src/js/burials/config/form.js
```diff
@@ -18,6 +18,7 @@ import currentOrPastDateUI from '../../common/schemaform/definitions/currentOrPa
 import toursOfDutyUI from '../definitions/toursOfDuty';
 import fileUploadUI from '../../common/schemaform/definitions/file';
 import { validateBurialDate } from '../validation';
+import GetFormHelp from '../common/schemaform/GetPensionOrBurialFormHelp';
```
cvalarida: Needs another `../` to get the right pathname.

src/js/common/schemaform/GetPensionOrBurialFormHelp.jsx
jbalboni: I think this is ok.

src/js/pensions/components/GetFormHelp.jsx
jbalboni: That's a fine spot.

src/js/common/schemaform/GetPensionOrBurialFormHelp.jsx
```diff
@@ -0,0 +1,16 @@
...
+        Monday - Friday, 8:00 a.m. - 9:00 p.m. ET<br/>
+        <p className="talk">For Telecommunications Relay Service (TRS): dial <a href="tel:711">711</a></p>
```
cvalarida: Does this make sense as a tel link? Is it an extension? Is it what we punch in after we call to get to the telecommunications relay service (whatever that is)?

cehsu: Thanks @cvalarida, really appreciate you highlighting this question, although this number should get folks through to TRS, perhaps additional instructions would bring help even closer within reach. Also interested in @J-Quag reflections on the best way to present this.

cehsu: I'll merge this for now and include any additional guidance regarding this in a subsequent PR.

PR #5895, opened 2017-06-28
Avoid parallel requests when loading messaging app

patrickvinograd:
Temporary fix for issue described in https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3591. Tested locally. 

Would like confirmation that the ordering of steps in componentDidMount is okay - I don't quite understand what that redirect stuff is doing. And also confirmation that I don't need a similar call in componentDidUpdate.


patrickvinograd:
This doesn't work because the Threads container expects recipients to be populated already...working on it in #5903 which encompasses this change as well.

patrickvinograd:
Closing in favor of #5903.

PR #5897, opened 2017-06-28
Configure common definition date widget to enforce full date

cehsu:
Connects to department-of-veterans-affairs/vets.gov-team#3522


jbalboni:
@cehsu If you get the latest backend schema, it should automatically be preventing partial dates. I don't think there are any code changes necessary (though worth testing).

cehsu:
Correct @jbalboni, if any date fields are not provided, the user is presented with the default required error message: `Please provide a response`. I'll close this.

PR #5898, opened 2017-06-28
Fix flashing on intro page

jbalboni:
We get some flashing of content on form intro pages with save in progress, because we couldn't tell when we were waiting for profile info from the server vs a user had an expired session and the request received a 401 error.

This addresses that by creating a loading flag in the profile state and setting it when the login status changes (or on a specific action).

While the profile is loading, we display this:

screen shot 2017-06-28 at 3 59 48 pm

### DIFF
--- a/src/js/common/helpers/login-helpers.js
+++ b/src/js/common/helpers/login-helpers.js
@@ -1,6 +1,6 @@
 import environment from './environment.js';
 import { updateLoggedInStatus } from '../../login/actions';
-import { updateProfileField } from '../../user-profile/actions';
+import { updateProfileField, profileLoadingFinished } from '../../user-profile/actions';
 
 export function handleVerify(verifyUrl) {
   window.dataLayer.push({ event: 'verify-link-clicked' });
@@ -39,6 +39,8 @@ export function getUserData(dispatch) {
       dispatch(updateProfileField('status', json.data.attributes.va_profile.status));
       dispatch(updateProfileField('services', json.data.attributes.services));
       dispatch(updateLoggedInStatus(true));
+    } else {
+      dispatch(profileLoadingFinished());
     }
   });
 }
--- a/src/js/user-profile/reducers/profile.js
+++ b/src/js/user-profile/reducers/profile.js
@@ -1,6 +1,7 @@
 import _ from 'lodash/fp';
 
-import { UPDATE_PROFILE_FIELD } from '../actions';
+import { UPDATE_PROFILE_FIELD, PROFILE_LOADING_FINISHED } from '../actions';
+import { UPDATE_LOGGEDIN_STATUS } from '../../login/actions';
 
 // TODO(crew): Romove before this goes to production.
 const initialState = {
@@ -15,13 +16,20 @@ const initialState = {
   gender: null,
   accountType: null,
   savedForms: [],
-  prefillsAvailable: []
+  prefillsAvailable: [],
+  loading: true
 };
 
 function profileInformation(state = initialState, action) {
   switch (action.type) {
     case UPDATE_PROFILE_FIELD: {
       return _.set(action.propertyPath, action.value, state);
+    }
+    case PROFILE_LOADING_FINISHED: {
+      return _.set('loading', false, state);
+    }
+    case UPDATE_LOGGEDIN_STATUS: {
+      return _.set('loading', false, state);
     }
 
     default:
--- a/test/user-profile/reducers/profile.unit.spec.js
+++ b/test/user-profile/reducers/profile.unit.spec.js
@@ -1,0 +1,22 @@
+import { expect } from 'chai';
+
+import reducer from '../../../src/js/user-profile/reducers/profile.js';
+import { UPDATE_LOGGEDIN_STATUS } from '../../../src/js/login/actions';
+import { PROFILE_LOADING_FINISHED } from '../../../src/js/user-profile/actions';
+
+describe('Profile reducer', () => {
+  it('should set loading to false when profile is done loading', () => {
+    const state = reducer({}, {
+      type: PROFILE_LOADING_FINISHED
+    });
+
+    expect(state.loading).to.be.false;
+  });
+  it('should set loading to false when logged in status changes', () => {
+    const state = reducer({}, {
+      type: UPDATE_LOGGEDIN_STATUS
+    });
+
+    expect(state.loading).to.be.false;
+  });
+});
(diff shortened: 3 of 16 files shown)


src/js/common/helpers/login-helpers.js
```diff
@@ -39,6 +39,8 @@ export function getUserData(dispatch) {
       dispatch(updateProfileField('status', json.data.attributes.va_profile.status));
       dispatch(updateProfileField('services', json.data.attributes.services));
       dispatch(updateLoggedInStatus(true));
+    } else {
+      dispatch(profileLoadingFinished());
```
jbalboni: I could just dispatch a `updateLoggedInStatus(false)` action here instead of creating a separate one, but that felt odd? The logged in status is already false in the current state. I'm open to changing that depending on what other people think.

src/js/user-profile/reducers/profile.js
```diff
@@ -1,6 +1,7 @@
 import _ from 'lodash/fp';
 
-import { UPDATE_PROFILE_FIELD } from '../actions';
+import { UPDATE_PROFILE_FIELD, PROFILE_LOADING_FINISHED } from '../actions';
+import { UPDATE_LOGGEDIN_STATUS } from '../../login/actions';
 
 // TODO(crew): Romove before this goes to production.
```
cvalarida: Not a part of this PR, but should this still be here?

src/js/common/helpers/login-helpers.js
cvalarida: Works for me. Less DRY, but more semantic.

annekainicUSDS: No I think it makes sense to have a separate action for this!

annekainicUSDS: The only thing I'm a little confused about is when this `else` condition gets hit. Looks like when `!json.data`? When is that? Not sure I totally understand how fetch works in this case.

test/user-profile/reducers/profile.unit.spec.js
```diff
@@ -0,0 +1,22 @@
+import { expect } from 'chai';
+
+import reducer from '../../../src/js/user-profile/reducers/profile.js';
+import { UPDATE_LOGGEDIN_STATUS } from '../../../src/js/login/actions';
+import { PROFILE_LOADING_FINISHED } from '../../../src/js/user-profile/actions';
+
+describe('Profile reducer', () => {
```
cvalarida: Huh, I guess there _wasn't_ a reducer unit test for user-profile. Should we add the other action (`UPDATE_PROFILE_FIELD`) in here too even though it isn't in the scope of this PR?

cvalarida (approved):
Cool beans! Looks good.

src/js/user-profile/reducers/profile.js
```diff
@@ -15,14 +16,21 @@ const initialState = {
   gender: null,
   accountType: null,
   savedForms: [],
-  prefillsAvailable: []
+  prefillsAvailable: [],
+  loading: true
```
annekainicUSDS: So anytime the app is first loaded this will be defaulted to `true`, and only changed to `false` once we finish the fetch of the data? Makes sense to me.

annekainicUSDS (approved):
Looks good to me!

src/js/common/helpers/login-helpers.js
raquelromano-gov: If we're going to have a new action, then can it also be dispatched above in the case where the response has user data? So that would dispatch both `updateLoggedInStatus(true)` and `profileLoadingFinished()` and below  in src/js/user-profile/reducers/profile.js you wouldn't have to set `loading` to false for an `UPDATE_LOGGEDIN_STATUS` action.

jbalboni: @raquelromano-gov So, after thinking about it, I think that's probably clearer for the code that's here now.

That said: I mentioned this to Chris the other day: we often dispatch actions in Redux that are very granular and focused on updating one piece of state, but I think it's better to have actions that describe an actual event in the system. Ideally, I think the profile here code would dispatch actions like FETCH_PROFILE, UPDATE_PROFILE_INFO, and PROFILE_FETCH_FAILED (or something like that). The logged in and pending status would be set in actions like that (along with other info) and not have their own specific actions. That way, if you look at a list of actions that have occurred, it's a (somewhat) meaningful history of what your application is doing.

jbalboni: @annekainicUSDS It looks like the else is hit whenever there's a non-200 server response. We get json back from the server, but it has no data attribute.

That's sort of brittle; we should probably be checking `reponse.ok` and acting accordingly.

jbalboni: @raquelromano-gov Actually, the other reason I did this is because there are other places where the login status is set to true or false and I didn't want to put an extra dispatch in each place along with it.

test/user-profile/reducers/profile.unit.spec.js
jbalboni: I think we know we want to refactor that code anyway, so I think it's ok to wait.

src/js/user-profile/reducers/profile.js
jbalboni: It's also set to false if we don't have a user token. Any time the logged in status is updated, essentially.

src/js/common/helpers/login-helpers.js
cvalarida: Yeah, I think I like the "do this thing" approach over "make these changes." When we dispatch actions to "do this thing," if we need to change what "this thing" does, it's all consolidated in the reducer instead of wherever the actions were `dispatch`ed. Also, like @jbalboni said, it makes the history more meaningful.

I think we'll also get minute performance gains by not waiting around for the next redux cycle to catch the remaining dispatched actions (however that works under the hood).

raquelromano-gov: Yeah, I saw that discussion in your other PR (I think Chris posted it in a channel) and it makes sense.  So we're saying a given action may result in different reducers updating different types of state, but the action doesn't/shouldn't need to know about that.

Sounds like a fun and messy refactoring project that will hold our testing approach to the fire!

src/js/user-profile/reducers/profile.js
raquelromano-gov: LOL. I wondered about this long ago but part of me really wanted to know why one would delete the initial state  definition before going to production, and I don't know which crew to ask. Maybe the comment was originally for a different chunk of code that's long since been removed?

ayaleloehr: "crew" is Courtney E-W's initials.

cvalarida: 💡 Thanks for that elucidation, @ayaleloehr!

PR #5903, opened 2017-06-28
Consistent errorView guard in all MHV-based apps

patrickvinograd:


patrickvinograd:
I know there is some duplication of the ErrorView component that could be factored out later.

patrickvinograd:
Consistent UX for the case where we get an `MHVAC1` error code back from the backend:

screen shot 2017-06-28 at 1 55 20 pm
screen shot 2017-06-28 at 3 53 10 pm
screen shot 2017-06-28 at 3 53 23 pm

bshyong (approved):
Looks great; yes error view and other things we can refactor later

PR #5904, opened 2017-06-29
Acct master 20170628

patrickvinograd:
Getting acct branch onto master before doing the reverse.


bshyong:
Great; do you need help with the conflict resolution or do you have it?

patrickvinograd:
Got it.

PR #5912, opened 2017-06-29
Remove email confirmation and make email and phone not required

cehsu:
Connects to department-of-veterans-affairs/vets.gov-team#3589

screen shot 2017-06-29 at 10 27 06 am


src/js/burials/config/form.js
```diff
@@ -426,21 +426,14 @@ const formConfig = {
             claimantEmail: {
               'ui:title': 'Email address'
             },
-            'view:claimantEmailConfirmation': {
```
jbalboni: Need to remove the `validateMatch` validation as well.

jbalboni (approved):
Need to remove the validateMatch validation, but otherwise looks good.

src/js/burials/config/form.js
cehsu: Thanks @jbalboni, change made.

PR #5913, opened 2017-06-29
VA benefit options for benefit summary letter

annekainicUSDS:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3362.

This PR creates a new container component for the benefit summary letter. It creates a new action to update which benefits the user selects and unselects, storing that value in the store to eventually be sent with the additional post to get the benefit summary letter pdf. There's some hacky things happening in the code, added as many comments in the code as I could to explain what's going on. Will add comments in the PR as well.

Unit tests coming in a separate PR.

screen shot 2017-06-29 at 4 38 59 pm

### DIFF
--- a/script/build.js
+++ b/script/build.js
@@ -111,8 +111,13 @@ smith.destination(`../build/${options.buildtype}`);
 // This lets us access the {{buildtype}} variable within liquid templates.
 smith.metadata({ buildtype: options.buildtype });
 
-// To block an app from production add the following to the below list:
-//  ignoreList.push('<path-to-content-file>');
+// To use:
+// const ignore = require('metalsmith-ignore');
+// const ignoreList = [];
+// if (options.buildtype === 'production') {
+//   ignoreList.push('track-claims/*');
+// }
+// smith.use(ignore(ignoreList));
 const ignore = require('metalsmith-ignore');
 const ignoreList = [];
 if (options.buildtype === 'production') {
--- a/script/build.js
+++ b/script/build.js
@@ -111,13 +111,8 @@ smith.destination(`../build/${options.buildtype}`);
 // This lets us access the {{buildtype}} variable within liquid templates.
 smith.metadata({ buildtype: options.buildtype });
 
-// To use:
-// const ignore = require('metalsmith-ignore');
-// const ignoreList = [];
-// if (options.buildtype === 'production') {
-//   ignoreList.push('track-claims/*');
-// }
-// smith.use(ignore(ignoreList));
+// To block an app from production add the following to the below list:
+//  ignoreList.push('<path-to-content-file>');
 const ignore = require('metalsmith-ignore');
 const ignoreList = [];
 if (options.buildtype === 'production') {
--- a/src/js/letters/actions/letters.js
+++ b/src/js/letters/actions/letters.js
@@ -31,3 +31,11 @@ export function getBenefitSummaryOptions() {
     );
   };
 }
+
+export function updateBenefitSummaryOptionsStatus(propertyPath, value) {
+  return {
+    type: 'UPDATE_BENEFIT_SUMMARY_OPTIONS_STATUS',
+    propertyPath,
+    value
+  };
+}
--- a/src/js/letters/actions/letters.js
+++ b/src/js/letters/actions/letters.js
@@ -32,9 +32,9 @@ export function getBenefitSummaryOptions() {
   };
 }
 
-export function updateBenefitSummaryOptionsStatus(propertyPath, value) {
+export function updateBenefitSummaryOption(propertyPath, value) {
   return {
-    type: 'UPDATE_BENEFIT_SUMMARY_OPTIONS_STATUS',
+    type: 'UPDATE_BENEFIT_SUMMARY_OPTION',
     propertyPath,
     value
   };
--- a/src/js/letters/components/LetterList.jsx
+++ b/src/js/letters/components/LetterList.jsx
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 
 import CollapsiblePanel from '../../common/components/CollapsiblePanel';
 import DownloadLetterLink from './DownloadLetterLink';
-import VeteranBenefitSummaryLetter from './VeteranBenefitSummaryLetter';
+import VeteranBenefitSummaryLetter from '../containers/VeteranBenefitSummaryLetter';
 
 import { letterContent } from '../utils/helpers';
 
@@ -15,7 +15,7 @@ class LetterList extends React.Component {
       let content;
 
       if (letter.letterType === 'benefit_summary') {
-        content = (<VeteranBenefitSummaryLetter benefitSummaryOptions={this.props.benefitSummaryOptions}/>);
+        content = (<VeteranBenefitSummaryLetter/>);
       } else {
         content = letterContent[letter.letterType] || '';
       }
--- a/src/js/letters/components/LetterList.jsx
+++ b/src/js/letters/components/LetterList.jsx
@@ -10,8 +10,6 @@ import { letterContent } from '../utils/helpers';
 class LetterList extends React.Component {
   render() {
     const letterItems = (this.props.letters || []).map((letter, index) => {
-      // TODO: if letter.letterType === 'benefit_summary', render the bsl
-      // custom component and pass in this.props.benefitSummaryOptions
       let content;
 
       if (letter.letterType === 'benefit_summary') {
--- a/src/js/letters/components/LetterList.jsx
+++ b/src/js/letters/components/LetterList.jsx
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types';
 
 import CollapsiblePanel from '../../common/components/CollapsiblePanel';
 import DownloadLetterLink from './DownloadLetterLink';
+import VeteranBenefitSummaryLetter from './VeteranBenefitSummaryLetter';
 
 import { letterContent } from '../utils/helpers';
 
@@ -11,13 +12,19 @@ class LetterList extends React.Component {
     const letterItems = (this.props.letters || []).map((letter, index) => {
       // TODO: if letter.letterType === 'benefit_summary', render the bsl
       // custom component and pass in this.props.benefitSummaryOptions
-      let content = letterContent[letter.letterType] || '';
+      let content;
+
+      if (letter.letterType === 'benefit_summary') {
+        content = (<VeteranBenefitSummaryLetter benefitSummaryOptions={this.props.benefitSummaryOptions}/>);
+      } else {
+        content = letterContent[letter.letterType] || '';
+      }
 
       return (
         <CollapsiblePanel
             panelName={letter.name}
             key={`collapsiblePanel-${index}`}>
-          <p>{content}</p>
+          <div>{content}</div>
           <DownloadLetterLink
               letterType={letter.letterType}
               letterName={letter.name}
--- a/src/js/letters/containers/VeteranBenefitSummaryLetter.jsx
+++ b/src/js/letters/containers/VeteranBenefitSummaryLetter.jsx
@@ -4,7 +4,7 @@ import { connect } from 'react-redux';
 import _ from 'lodash';
 
 import { veteranBenefitSummaryOptionText } from '../utils/helpers';
-import { updateBenefitSummaryOptionsStatus } from '../actions/letters';
+import { updateBenefitSummaryOption } from '../actions/letters';
 
 class VeteranBenefitSummaryLetter extends React.Component {
   constructor() {
@@ -13,13 +13,14 @@ class VeteranBenefitSummaryLetter extends React.Component {
   }
 
   handleChange(domEvent) {
-    this.props.updateBenefitSummaryOptionsStatus(domEvent.target.id, domEvent.target.checked);
+    this.props.updateBenefitSummaryOption(domEvent.target.id, domEvent.target.checked);
   }
 
   render() {
     const benefitInfo = this.props.benefitSummaryOptions.benefitInfo;
     const optionsToInclude = this.props.optionsToInclude;
     let militaryServiceInfo;
+    let vaBenefitInformation;
     let vaBenefitInfoRows = [];
 
     _.forIn(benefitInfo, (value, key) => {
@@ -30,12 +31,14 @@ class VeteranBenefitSummaryLetter extends React.Component {
       // it means this benefit is not for veterans, and therefore it shouldn't
       // be displayed.
       // 2. If the value of the current benefit is false, we don't want to
-      // display it.
+      // display it. Values can be either true, false, or some other value,
+      // which is why we're checking that it's not false.
       if (optionText && value !== false) {
         vaBenefitInfoRows.push(
           <tr key={`option${key}`}>
             <th scope="row">
               <input
+                  aria-labelledby={`${key}Label`}
                   autoComplete="false"
                   checked={optionsToInclude[key]}
                   id={key}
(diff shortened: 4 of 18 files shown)


src/js/letters/reducers/index.js
```diff
@@ -23,16 +27,34 @@ function letters(state = initialState, action) {
       // the various error scenarios
       return set('lettersAvailable', false, state);
     case 'GET_BENEFIT_SUMMARY_OPTIONS_SUCCESS':
+      // Create object for which options to include in post for summary letter
+      // Default all options to true so they appear checked in the UI
+      options = mapValues(action.data.data.attributes.benefitInformation, (value) => {
```
annekainicUSDS: What this section is doing is mapping through the list of options we get in the response from EVSS and putting all of those in a new object called `optionsToInclude`, which will store a boolean value for each option that will reflect whether the option is selected in the UI (`true`) or not (`false`).

Where it gets tricky is how the front-end handles concealing options that are sent to us by EVSS that we don't want to show in the UI, either because they're not eligible for those options (since the value is sent to us as `false`) or because the option is actually for the dependent summary letter (this shouldn't ever happen in real life, but you never know). What I do below is check if the value for each option is `!false`; if so I put it in `optionsToInclude`, defaulted to `true`. If the value is `false`, I put it in `optionsToInclude` as `false`. Hopefully that makes sense!

src/js/letters/containers/VeteranBenefitSummaryLetter.jsx
```diff
@@ -0,0 +1,108 @@
...
+        <h2>VA Benefit Information</h2>
+        <table className="usa-table-borderless">
+          <thead>
+            <tr>
+              <th scope="col">Include</th>
+              <th scope="col">Something</th>
+            </tr>
+          </thead>
+          <tbody>
+            {vaBenefitInfoRows}
```
annekainicUSDS: Just realized that I should probably check is `optionsAvailable` is `true` before trying to render anything. If not, I should probably display some kind of error message. Will add that.

src/js/letters/actions/letters.js
```diff
@@ -31,3 +31,11 @@ export function getBenefitSummaryOptions() {
     );
   };
 }
+
+export function updateBenefitSummaryOptionsStatus(propertyPath, value) {
+  return {
+    type: 'UPDATE_BENEFIT_SUMMARY_OPTIONS_STATUS',
```
raquelromano-gov: I wonder if this could just be `UPDATE_BENEFIT_SUMMARY_OPTION` and `updateBenefitSummaryOption()`

src/js/letters/containers/VeteranBenefitSummaryLetter.jsx
```diff
@@ -0,0 +1,108 @@
...
+    _.forIn(benefitInfo, (value, key) => {
+      const optionText = veteranBenefitSummaryOptionText(benefitInfo, key);
+
+      // There are 2 conditions this is checking for
+      // 1. If the current benefit is not in the option text list from helpers,
+      // it means this benefit is not for veterans, and therefore it shouldn't
+      // be displayed.
+      // 2. If the value of the current benefit is false, we don't want to
+      // display it.
+      if (optionText && value !== false) {
```
raquelromano-gov: Alternatively, `(optionText && value)`, unless you explicitly want to test not false because it's semantically closer to the business logic.

src/js/letters/containers/VeteranBenefitSummaryLetter.jsx
```diff
@@ -0,0 +1,108 @@
...
+            <th scope="row">
+              <input
+                  autoComplete="false"
+                  checked={optionsToInclude[key]}
+                  id={key}
+                  name={key}
+                  type="checkbox"
+                  onChange={this.handleChange}/>
+            </th>
+            <td>{veteranBenefitSummaryOptionText(benefitInfo, key)}</td>
```
raquelromano-gov: I think here you'll need to connect this text to the input using `aria-labelledby` and maybe wrapping this in a `label` tag as well. Here's an example for a `textarea`: https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/js/common/components/form-elements/ErrorableTextarea.jsx#L92

src/js/letters/reducers/index.js
```diff
@@ -23,16 +27,34 @@ function letters(state = initialState, action) {
...
+        let isSelected = false;
+        // If the value of the benefit option is not false, it means the user
```
raquelromano-gov: I like this whole comment so it's clear what's happening. I'd shorten the code to
```
return value && isSelected
```

src/js/letters/reducers/index.js
```diff
@@ -23,16 +27,34 @@ function letters(state = initialState, action) {
       // the various error scenarios
       return set('lettersAvailable', false, state);
     case 'GET_BENEFIT_SUMMARY_OPTIONS_SUCCESS':
+      // Create object for which options to include in post for summary letter
```
raquelromano-gov: Since `action.data.data.attributes.benefitInformation` is so horribly ugly and used 2x below, we may want to define `const benefitInfo = ...` here.

src/js/letters/utils/helpers.js
```diff
@@ -1,8 +1,9 @@
 /* eslint-disable camelcase */
-
+import React from 'react';
 import merge from 'lodash/fp/merge';
 
 import environment from '../../common/helpers/environment';
+import { formatDateShort } from '../../post-911-gib-status/utils/helpers';
```
raquelromano-gov: Should we move this to a common place? Have been meaning to check if there are already helpers like this anywhere in `src/js/common`

src/js/letters/utils/helpers.js
```diff
@@ -46,3 +47,22 @@ export const letterContent = {
...
+  const textForAllOptions = {
+    awardEffectiveDate: <div>The effective date of the last change to your current award was <strong>{formatDateShort(allOptions.awardEffectiveDate)}</strong></div>,
+    hasAdaptedHousing: <div>You <strong>have</strong> been found entitled to a Specially Adapted Housing (SAH) and/or Special Home Adaptation (SHA) grant</div>,
+    hasChapter35Eligibility: <div>You <strong>are</strong> considered to be totally and permanently disabled solely due to your service-connected disabilities</div>,
+    hasDeathResultOfDisability: <div></div>,
+    hasIndividualUnemployabilityGranted: <div>You <strong>are</strong> being paid at the 100 percent rate because you are unemployable due to your service-connected disabilities</div>,
+    hasNonServiceConnectedPension: <div>Your non-service connected pension information</div>,
+    hasServiceConnectedDisabilities: <div>You have one or more service-connected disabilities</div>,
+    hasSpecialMonthlyCompensation: <div>You <strong>are</strong> service-connected for loss of or loss of use of a limb, or you are totally blind in or missing at least one eye</div>,
```
raquelromano-gov: @phiden is there a missing word in the mock? "for loss of or loss of use of a limb"

src/js/letters/utils/helpers.js
```diff
@@ -46,3 +47,22 @@ export const letterContent = {
   benefit_summary: 'This letter shows what benefits you\'re receiving from the VA, military service, and disability status. Below, you can choose if you want military service and disability status to be included.',
   benefit_verification: 'This letter shows what benefits you\'re receiving from the VA. It is different from the benefit summary because it includes [x] and does not give you the option to choose what is included in the letter.'
 };
+
+export function veteranBenefitSummaryOptionText(allOptions, currentOption) {
```
raquelromano-gov: Instead of having an `allOptions`, can this just take the value of the current option? Or do you expect in the future that the text for a given option will depend on the values of other options (which isn't the case right now). If the latter, then maybe add a comment to this function explaining why we pass in all the options.

src/sass/letters.scss
```diff
@@ -3,30 +3,43 @@
...
-  padding-top: 1.5rem;
-  padding-bottom: 1.5rem;
-}
+  .accordion-header > button {
+    min-height: 6rem;
+    padding-top: 1.5rem;
+    padding-bottom: 1.5rem;
+  }
+
+  input[type="checkbox"] {
```
raquelromano-gov: I think this will need be only for an `input` inside a `th` or `td`

src/js/letters/containers/VeteranBenefitSummaryLetter.jsx
annekainicUSDS: Actually I want to use not false because sometimes the value could be something else other than true or false. For example, `awardEffectiveDate`'s value is a date, and `monthlyAwardAmount` is a number.

src/js/letters/containers/VeteranBenefitSummaryLetter.jsx
annekainicUSDS: Yep yep, will do thanks!

src/js/letters/reducers/index.js
annekainicUSDS: Haha I started with that but then eslint yelled at me for `Unexpected lexical declaration in case block`. Apparently you can't declare variables inside a case block.

src/js/letters/utils/helpers.js
annekainicUSDS: Yes definitely, I thought about it and then was too lazy to do it in the moment. Will update!

src/js/letters/utils/helpers.js
annekainicUSDS: Yep, good question, I was wondering about that too. The reason why I thought I had to pass `allOptions` in is because some of the text embeds data from the options. For example, for `awardEffectiveDate`, the actual date is embedded in the text. I'm not sure how else to handle it, as these aren't even really strings of text but rather React markup, so I don't think I could push it on the end or something like that.

src/sass/letters.scss
annekainicUSDS: ah yes, good point, will fix!

src/js/letters/containers/VeteranBenefitSummaryLetter.jsx
```diff
@@ -0,0 +1,126 @@
...
+          </tbody>
+        </table>
+      );
+    } else {
+      vaBenefitInformation = (
+        <div className="feature">
+          <h4>VA Benefit Information is currently unavailable</h4>
+          <div>Please try again later. You can call the Vets.gov Help Desk for additional assistance.</div>
+          <div>1-855-574-7286</div>
+          <div>Monday - Friday, 8:00am - 8:00pm (ET)</div>
```
annekainicUSDS: @phiden could you provide feedback on this language? This message appears if the VA benefit information options aren't available for some reason.

phiden:
```
<h4>Your VA Benefit Summary letter is currently unavailable</h4>
<div>We weren't able to retrieve your VA Benefit Summary letter. Please call 1-855-574-7286 between Monday‒Friday, 8:00 a.m.‒8:00 p.m. (ET).</div>
```

I would go with this ^ but tagging @rroueche for confirmation/edits.

src/js/letters/utils/helpers.js
phiden: @raquelromano-gov no, it's correct. Just worded oddly -- that's the exact language from the letter template: "You are service-connected for loss of or loss of use of a limb, or you are totally blind in or missing at least one eye:"

If it's ok with stakeholders we could edit that to read "You are service-connected for the loss of a limb, or the loss of the use of a limb, or you are totally blind in or missing at least one eye." cc @rroueche

raquelromano-gov: Ah, ok...yeah makes sense now! It definitely looked like a typo at first. Let's leave it as is for the moment, but if you want to add it to the list of stakeholder questions, and they are OK with the above clearer language, we can change it.

src/js/letters/components/LetterList.jsx
```diff
@@ -11,13 +12,19 @@ class LetterList extends React.Component {
     const letterItems = (this.props.letters || []).map((letter, index) => {
       // TODO: if letter.letterType === 'benefit_summary', render the bsl
```
raquelromano-gov: This comment can go away

src/js/letters/containers/VeteranBenefitSummaryLetter.jsx
```diff
@@ -0,0 +1,126 @@
...
+          <strong>Please note:</strong>This letter can only show up to the 3 most
+          recent periods of service.
+        </p>
+        <h2>Military Service Information</h2>
+        {militaryServiceInfo}
+        <h2>VA Benefit Information</h2>
+        {vaBenefitInformation}
+        <p>
+          If you see incorrect information for service periods or disability status,
+          please send a question using VA's <a href="/">Inquiry Routing & Information
```
raquelromano-gov: We should create a ticket to update all of the missing links in both letters and GIBS

src/js/letters/reducers/index.js
raquelromano-gov: oh, did not know!

src/js/letters/reducers/index.js
raquelromano-gov: Ah, ok thanks for explaining why you have to check that it's actually `false`.

So can these 5 lines be shortened to this single one?
```
return (value !== false) && isSelected;
```

src/js/letters/utils/helpers.js
raquelromano-gov: But is there ever a case where the text for a given option relies on the value of a **different** option? If not, then this function signature could be
```
veteranBenefitSummaryOptionText(key, value)
```
and e.g. below for the text that depends on the value
```
awardEffectiveDate: <div>The effective date of the last change to your current award was <strong>{formatDateShort(value)}</strong></div>,
```
and
```
monthlyAwardAmount: <div>Your current monthly award amount is <strong>${value}</strong></div>,
```
and at the end
```
return textForAllOptions[key];
```

And the `forIn` statement in `VeteranBenefitSummaryLetter.jsx` can call
```
const optionText = veteranBenefitSummaryOptionText(key, value);
```
instead of passing in `benefitInfo`

src/js/letters/components/LetterList.jsx
annekainicUSDS: Yep! will remove that

src/js/letters/containers/VeteranBenefitSummaryLetter.jsx
annekainicUSDS: definitely will do

src/js/letters/utils/helpers.js
annekainicUSDS: oh! yeah I think that would work! great idea, will change that

src/js/letters/reducers/index.js
annekainicUSDS: hm I think actually all I need to return is: 
```
return value !== false
```
The only reason I created the `isSelected` variable is so I could return a boolean, where `value === false` returns `false` and `value === true || value === _something else_` returns `true`. I think simple returning `value !== false` gets me that. Thanks for pointing this out!

script/build.js
```diff
@@ -111,8 +111,13 @@ smith.destination(`../build/${options.buildtype}`);
 // This lets us access the {{buildtype}} variable within liquid templates.
 smith.metadata({ buildtype: options.buildtype });
 
-// To block an app from production add the following to the below list:
-//  ignoreList.push('<path-to-content-file>');
+// To use:
```
raquelromano-gov: I think James merged a separate PR to replace these comments, so you probably didn't pick that up yet.

raquelromano-gov (approved):

script/build.js
annekainicUSDS: oops, resolved the rebase the wrong way. will fix!

PR #5916, opened 2017-06-29
Remove parents from dependents page

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3492
screen shot 2017-06-29 at 4 14 40 pm

### DIFF
--- a/src/js/pensions/config/form.js
+++ b/src/js/pensions/config/form.js
@@ -17,7 +17,6 @@ import {
   directDepositWarning,
   isMarried
 } from '../helpers';
-import { relationshipLabels } from '../labels';
 import IntroductionPage from '../components/IntroductionPage';
 import DisabilityField from '../components/DisabilityField';
 import SpouseMarriageTitle from '../components/SpouseMarriageTitle';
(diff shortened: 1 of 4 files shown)


src/js/pensions/config/form.js
```diff
@@ -99,10 +98,6 @@ function isCurrentMarriage(form, index) {
   return isMarried(form) && numMarriages - 1 === index;
 }
 
-function isChild(item) {
-  return item.dependentRelationship === 'child';
```
cehsu: Will `dependentRelationship` ever not be child now?

cehsu (approved):

src/js/pensions/config/form.js
```diff
@@ -853,9 +830,8 @@ const formConfig = {
                 minItems: 1,
                 items: {
                   type: 'object',
-                  required: ['dependentRelationship', 'fullName'],
+                  required: ['fullName', 'childDateOfBirth'],
                   properties: {
-                    dependentRelationship: dependents.items.properties.dependentRelationship,
```
cehsu: Are there any changes that should be made to the back end schema to reflect this?

jbalboni: Yep, I made a change to remove this field in the backend schema.

src/js/pensions/config/form.js
```diff
@@ -853,9 +830,8 @@ const formConfig = {
                 minItems: 1,
                 items: {
                   type: 'object',
-                  required: ['dependentRelationship', 'fullName'],
+                  required: ['fullName', 'childDateOfBirth'],
```
cvalarida: Does adding the `childDateOfBirth` here make it required even when there are no depended children to report?

jbalboni: No, they won't see this array/page at all and won't hit validation.

PR #5919, opened 2017-06-29
Remove from ignoreList

annekainicUSDS:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3540

This should not be merged until we have confirmed password protection is configured on these apps

### DIFF
--- a/script/build.js
+++ b/script/build.js
@@ -125,11 +125,9 @@ smith.metadata({ buildtype: options.buildtype });
 const ignore = require('metalsmith-ignore');
 const ignoreList = [];
 if (options.buildtype === 'production') {
-  ignoreList.push('education/gi-bill/post-9-11/ch-33-benefit.md');
   ignoreList.push('pension/application/527EZ.md');
   ignoreList.push('burials-and-memorials/application/530.md');
   ignoreList.push('burials-and-memorials/burial-planning/application.md');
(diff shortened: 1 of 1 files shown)


rroueche:
@jkassemi and/or @patrickvinograd gettin' close! could you confirm that auth is ready for:
`http://www.vets.gov/letters` and `http://www.vets.gov/education/gi-bill/post-9-11/ch-33-benefit` and all their subroutes? (for example: `/education/gi-bill/post-9-11/ch-33-benefit/print`)

script/build.js
```diff
@@ -125,11 +125,9 @@ smith.metadata({ buildtype: options.buildtype });
 const ignore = require('metalsmith-ignore');
```
raquelromano-gov: While you're editing this, there are some comments that could be deleted:
  * the `TODO(awong):` above can go away since that issue is closed
  * the "To use" can mostly go away since that code is already below. We can just add
```
To block an app from production, add the following to the below list:
  ignoreList.push('<path-to-content-file>');
```

raquelromano-gov (approved):

script/build.js
jkassemi: Wanted to get this merged without incurring another CI run. The changes you suggested are now part of #5923.

PR #5924, opened 2017-06-30
Refactor health-beta to avoid RequiredLoginView conflict

patrickvinograd:
Fixes the health-beta registration app. 

Problem was that I was doing everything in HealthBetaApp, but RequiredLoginView was rendering its view which cause a refresh of HealthBetaApp which cause RequiredLoginView to go again...putting the betaRegistration stuff inside a child container (Main) fixes that.


patrickvinograd:
Beta registration worked if logged in, but if logged out it would get stuck in a re-render loop. @bshyong you may want to adopt this for the appeals beta endpoint if you have not already done something similar.

bshyong:
@patrickvinograd thanks will adopt! will copy paste this your changes and we can refactor at some point if this Beta pattern is something that we will use often..

bshyong (approved):
🥇  Great

PR #5929, opened 2017-06-30
Logic fix for appeals access

bshyong:
- logic to bypass a check for `va_profile` if a user is trying to access appeals

Currently, we check for LOA3 and also existence of a `va_profile`. Appeals test users have LOA3 access but do not have `va_profile` so they cannot access the app even with access to `appeals-status` in their profile. 

This logic can be deleted in the future


patrickvinograd (approved):
LGTM for intended purpose, but let's get this code out of here soon :P

bshyong:
@annekainicUSDS here is the merged PR for reference. I tested it in staging and it looks good for user 300, can access `appeals status` app but cannot access `health-records`. trying to access `health-records` shows the normal MVI unavailable error message, meaning `health-records` is still checking for existence of `va_profile` as expected

PR #5932, opened 2017-06-30
Burial benefits selection page changes

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3566

screen shot 2017-06-30 at 3 35 34 pm

Also, no extra incurred expenses question:
screen shot 2017-06-30 at 3 32 14 pm

### DIFF
--- a/src/js/common/schemaform/validation.js
+++ b/src/js/common/schemaform/validation.js
@@ -340,7 +340,7 @@ export function validateFileField(errors, fileList) {
 export function validateBooleanGroup(errors, userGroup, form, schema, errorMessages = {}) {
   const { atLeastOne = 'Please choose at least one option' } = errorMessages;
   const group = userGroup || {};
-  if (!Object.keys(group).filter(item => !!group[item]).length) {
+  if (!Object.keys(group).filter(item => group[item] === true).length) {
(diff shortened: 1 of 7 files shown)


src/js/common/schemaform/validation.js
```diff
@@ -340,7 +340,7 @@ export function validateFileField(errors, fileList) {
 export function validateBooleanGroup(errors, userGroup, form, schema, errorMessages = {}) {
   const { atLeastOne = 'Please choose at least one option' } = errorMessages;
   const group = userGroup || {};
-  if (!Object.keys(group).filter(item => !!group[item]).length) {
+  if (!Object.keys(group).filter(item => group[item] === true).length) {
```
cvalarida: Idle curiosity: Why wasn't `!!` good enough here?

cvalarida (approved):

cehsu (approved):

src/js/common/schemaform/validation.js
jbalboni: There are other fields in this object that are not booleans. This is sort of lazy, I should probably make the validator take a list of keys.

PR #5937, opened 2017-06-30
Military service options

annekainicUSDS:
Connects https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3361

Just pushing all the work @raquelromano-gov did in a PR :)


src/js/letters/containers/VeteranBenefitSummaryLetter.jsx
```diff
@@ -17,9 +21,23 @@ class VeteranBenefitSummaryLetter extends React.Component {
...
 
   render() {
+    const serviceInfo = this.props.benefitSummaryOptions.serviceInfo || [];
+    const militaryServiceRows = serviceInfo.map((service, index) => {
+      return (
+        <tr key={`service${index}`}>
+          <th scope="row" className="service-info">{(service.branch || '').toLowerCase()}</th>
+          <td className="service-info">
+            {characterOfServiceContent[(service.characterOfService).toLowerCase()]}
+            {/* _.get([(service.characterOfService).toLowerCase(), ''], characterOfServiceContent) */}
```
raquelromano-gov: I forget why my `_.get()` wasn't working, but thinking more generally, this is another case of weird/missing/unexpected data, i.e., an unexpected character of service type for which we have no content. However, backend schema validation should ensure this never happens. Just glanced a the benefit options schema and these values are strings but should probably be enums: https://github.com/department-of-veterans-affairs/vets-api/blob/master/spec/support/schemas/letter_beneficiary.json#L54

src/js/letters/utils/helpers.js
```diff
@@ -37,6 +37,11 @@ export function apiRequest(url, optionalSettings = {}, success, error) {
     .then(success, error);
 }
 
+export const characterOfServiceContent = {
```
annekainicUSDS: Are we sure these are the only two options?

annekainicUSDS (commented):
LGTM! With one question.

src/js/letters/utils/helpers.js
raquelromano-gov: No, this was just a placeholder. I just looked up the types in the EVSS swagger docs and updated https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3689.

Possible values are ['HONORABLE', 'OTHER_THAN_HONORABLE', 'UNDER_HONORABLE_CONDITIONS', 'GENERAL', 'UNCHARACTERIZED', 'UNCHARACTERIZED_ENTRY_LEVEL', 'DISHONORABLE']

I can update these and guess at the content (for the most part just follow the pattern e.g., `under_honorable_conditions` -> "Under honorable conditions" and verify with product/UX later.

raquelromano-gov (commented):
I can take this back over and add the other discharge types since you're busy with loading spinner and other changes.

src/js/letters/containers/VeteranBenefitSummaryLetter.jsx
```diff
@@ -3,10 +3,14 @@ import { connect } from 'react-redux';
...
-import { veteranBenefitSummaryOptionText } from '../utils/helpers';
 import { updateBenefitSummaryOption } from '../actions/letters';
+import {
+  characterOfServiceContent,
+  veteranBenefitSummaryOptionText
+} from '../utils/helpers';
+import { formatDateShort } from '../../common/utils/helpers';
 
-class VeteranBenefitSummaryLetter extends React.Component {
+export class VeteranBenefitSummaryLetter extends React.Component {
```
raquelromano-gov: Need to export this class in order to import the unconnected component for testing.

test/letters/containers/VeteranBenefitSummaryLetter.unit.spec.jsx
```diff
@@ -0,0 +1,44 @@
+import React from 'react';
+import SkinDeep from 'skin-deep';
+import { expect } from 'chai';
+
+import { VeteranBenefitSummaryLetter } from '../../../src/js/letters/containers/VeteranBenefitSummaryLetter';
+
+const defaultProps = {
+  benefitSummaryOptions: {
+    benefitInfo: {
+      awardEffectiveDate: '1965-01-01T05:00:00.000+00:00',
```
raquelromano-gov: These date value trigger a moment() warning in unit tests, though they still pass. Leaving them as is (copied from a vets-api response) so we can remember to figure out if the values returned from EVSS need to be formatted or parsed differently on the front-end.

```
Deprecation warning: value provided is not in a recognized RFC2822 or ISO format.
moment construction falls back to js Date(), which is not reliable across all browsers and versions.
Non RFC2822/ISO date formats are discouraged and will be removed in an upcoming major release.
Please refer to http://momentjs.com/guides/#/warnings/js-date/ for more info.
```

test/letters/containers/VeteranBenefitSummaryLetter.unit.spec.jsx
```diff
@@ -0,0 +1,44 @@
...
+    const vdom = tree.getRenderOutput();
+    expect(vdom).to.exist;
+  });
+
+  it('should show benefit info options', () => {
+    const tree = SkinDeep.shallowRender(<VeteranBenefitSummaryLetter {...defaultProps}/>);
+    expect(tree.subTree('hasAdaptedHousing')).to.exist;
+  });
+
+  it('should show service info options', () => {
```
raquelromano-gov: In writing this, I began to wonder if we should be handling the case of an empty military service list (although that shouldn't be possible/make sense). I think https://github.com/department-of-veterans-affairs/vets-api/blob/master/spec/support/schemas/letter_beneficiary.json#L54 should be updated to require a non-empty array. I'll update the above issue.

raquelromano-gov:
@annekainicUSDS this is ready for another look!

raquelromano-gov:
Oh, can't add you as a reviewer since you opened a PR, but will wait for you to take a look anyway before merging.

raquelromano-gov (approved):

test/letters/containers/VeteranBenefitSummaryLetter.unit.spec.jsx
annekainicUSDS: ah yes I noticed this too, not sure what the fix for this is yet but we can address is a separate ticket

src/js/letters/containers/VeteranBenefitSummaryLetter.jsx
```diff
@@ -92,7 +109,33 @@ class VeteranBenefitSummaryLetter extends React.Component {
           recent periods of service.
         </p>
         <h2>Military Service Information</h2>
-        {militaryServiceInfo}
+        <div className="form-checkbox">
+          <input
+              autoComplete="false"
+              id="serviceInfoCheckboxId"
```
annekainicUSDS: Minor comment: because this input is tied to `this.handleChange`, the id that will be set in the `optionsToInclude` object in the store is the id of the input, which in this case is `serviceInfoCheckboxId`. I wonder if we need to name this something closer to what EVSS will accept on their side in the bsl post for the pdf.

annekainicUSDS (commented):
LGTM! Noticed one minor thing, but we can address that separately, would like to merge as is for now.

src/js/letters/containers/VeteranBenefitSummaryLetter.jsx
raquelromano-gov: I can change this now.

annekainicUSDS: Awesome, you'll want to pull down the branch before you make that change, because I just rebased this on master :/ sorry about that, wanted to update it before merging

raquelromano-gov: This is a little brittle--I  believe the request body for posting the BSL options expects different keys from the ones returned in benefitInfo and serviceInfo, so I"ll add a map in helpers.js to define those and look them up in the change handler.

annekainicUSDS: Yeah I'm not totally positive on what the request body should be, but I think for military service it's looking for this: `"militaryService": false,` (found in this documentation at the bottom: https://github.com/department-of-veterans-affairs/vets.gov-team/blob/master/Products/EVSS%20Integration/VA%20Letters/letters-responses-from-vets-api.md). But I see that the other options do have slightly different keys. That'll be great to map those. Thanks!

PR #5938, opened 2017-07-03
Update text and add min file check on burial/pension forms

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3563
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3516
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3583

screen shot 2017-07-03 at 9 46 38 am
screen shot 2017-07-03 at 9 46 27 am
screen shot 2017-07-03 at 9 45 47 am
screen shot 2017-07-03 at 9 55 09 am

### DIFF
--- a/src/js/pensions/helpers.jsx
+++ b/src/js/pensions/helpers.jsx
@@ -61,7 +61,7 @@ export function fileHelp({ formContext }) {
   return (
     <p>
       Files we accept: pdf, jpg, png<br/>
-      Maximum file size: 2MB
+      Maximum file size: 20MB
     </p>
   );
 }
--- a/src/js/burials/config/form.js
+++ b/src/js/burials/config/form.js
@@ -184,7 +184,7 @@ const formConfig = {
           title: 'Burial information',
           path: 'veteran-information/burial',
           uiSchema: {
-            burialDate: currentOrPastDateUI('Date of burial'),
+            burialDate: currentOrPastDateUI('Date of burial (includes cremation or interment)'),
             deathDate: currentOrPastDateUI('Date of death'),
             locationOfDeath: {
               location: {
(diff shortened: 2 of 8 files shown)


src/js/pensions/helpers.jsx
```diff
@@ -61,7 +61,7 @@ export function fileHelp({ formContext }) {
   return (
     <p>
       Files we accept: pdf, jpg, png<br/>
-      Maximum file size: 2MB
+      Maximum file size: 20MB
```
cvalarida: I wonder...Would it be worth making this match the `maxSize` in the `formConfig`? That is, pass the `maxSize` in and have it do some calculations and print out the size in MB.

Might be a bit of a pain, but that'd prevent us from displaying the wrong number. Just a precautionary thing.

"It _says_ I can upload a 20MB jpg, but it won't let me upload a 17.5MB one! Ugh!"

cvalarida (approved):
Just idly wondering about file size display syncing to the actual maximum file size.

src/js/pensions/helpers.jsx
jbalboni: I didn't do that mainly because there's not an obvious way to tie the two together. This is a regular description field and doesn't know anything about the file field. On the burial form, this description covers two file fields, so we can't really have the common file definition include this description, either.

We could make a higher order component that you pass in file types and a max size, but that's only slightly (if at all) better.

cvalarida: Yeah, I was just pondering through it and I couldn't come up with a good answer. We'll just have to be careful to make sure they match is all.

PR #5940, opened 2017-07-03
Make iterative updates except for no data case

annekainicUSDS:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3481

Checks off all the boxes in that ticket of items to change **except** for the no data case, which is still blocked while we wait for information from vets-api on what the data looks like in that case.

screen shot 2017-07-03 at 11 32 52 am

screen shot 2017-07-03 at 11 33 04 am

### DIFF
--- a/src/js/post-911-gib-status/components/InfoPair.jsx
+++ b/src/js/post-911-gib-status/components/InfoPair.jsx
@@ -9,24 +9,41 @@ class InfoPair extends React.Component {
         `usa-grid-full ${this.props.spacingClass}`
         : 'usa-grid-full';
 
-    return (
-      this.props.value &&
-        <div className={gridRowClasses}>
-          <div className="usa-width-one-third">
-            <span><strong>{this.props.label}: </strong></span>
-          </div>
-          <div className="usa-width-one-third">
-            {this.props.value}
-          </div>
+    const row = (
+      <div className={gridRowClasses}>
+        <div className="usa-width-one-third">
+          <span><strong>{this.props.label}: </strong></span>
         </div>
-    || null);
+        <div className="usa-width-two-thirds">
+          {this.props.value}
+        </div>
+      </div>
+    );
+
+    let rowToDisplay;
+    // Some rows should display the value even if the value is 0,
+    // while other rows shouldn't display anything if the value is 0.
+    // If this prop is passed, the value should be displayed regardless
+    // of if it is 0 or not.
+    // If the prop is not passed, check first if the value exists before
+    // determining whether or not to display it.
+    if (this.props.displayIfZero) {
+      rowToDisplay = row;
+    } else {
+      rowToDisplay = this.props.value && row;
+    }
+
+    return (
+      rowToDisplay || null
+    );
   }
 }
 
 InfoPair.propTypes = {
   label: PropTypes.string.isRequired,
   value: PropTypes.any,
-  spacingClass: PropTypes.string
+  spacingClass: PropTypes.string,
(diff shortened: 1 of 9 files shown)


src/js/post-911-gib-status/components/InfoPair.jsx
```diff
@@ -9,24 +9,41 @@ class InfoPair extends React.Component {
...
-    || null);
+      </div>
+    );
+
+    let rowToDisplay;
+    // Some rows should display the value even if the value is 0,
+    // while other rows shouldn't display anything if the value is 0.
+    // If this prop is passed, the value should be displayed regardless
+    // of if it is 0 or not.
+    // If the prop is not passed, check first if the value exists before
```
raquelromano-gov: If the displayIfZero prop is not passed

(the comment originally confused me since I thought "the prop" referred to this.props.value

raquelromano-gov: Actually, something like this might be a little clearer:

The displayIfZero prop may be passed in because some rows should display 0 values, while other rows should not display any 0 values. If displayIfZero is false or is not passed in, only display the row if the value is defined and nonzero.

src/js/post-911-gib-status/components/InfoPair.jsx
```diff
@@ -9,24 +9,41 @@ class InfoPair extends React.Component {
...
+    // determining whether or not to display it.
+    if (this.props.displayIfZero) {
+      rowToDisplay = row;
```
raquelromano-gov: Can these just be return statements since we don't use `rowToDisplay` anywhere else?

src/js/post-911-gib-status/containers/PrintPage.jsx
```diff
@@ -16,10 +15,9 @@ class PrintPage extends React.Component {
       <div className="print-status">
         <div className="print-screen">
           <img src="/img/design/logo/va-logo.png" alt="VA logo" width="300"/>
-          <h1 className="section-header">Post-9/11 GI Bill Certificate of Eligibility</h1>
+          <h1 className="section-header">Post-9/11 GI Bill Benefit Information</h1>
+          <p>Date generated: {todayFormatted}</p>
```
raquelromano-gov: I notice that not all the mocks have the "Date generated" message  (e.g., https://marvelapp.com/8d6igd9/screen/29016777 is missing it), but I'm guessing it should show up for at least all of the cases where the person is enrolled. @phiden can you confirm or correct?

src/js/post-911-gib-status/components/InfoPair.jsx
annekainicUSDS: Yes, thanks! Will fix the comment

src/js/post-911-gib-status/containers/PrintPage.jsx
phiden: @raquelromano-gov we removed the date from all screens except the print screen, for the time being.

raquelromano-gov: ah, OK...sorry, was not looking carefully. 
NM: looks good!

raquelromano-gov (approved):
Going ahead and approving so you can merge as soon as you want.

src/js/post-911-gib-status/components/InfoPair.jsx
annekainicUSDS: I just wasn't sure about having multiple return statements :/ what do you think?

PR #5941, opened 2017-07-03
Initial events for account creation

robertfairhead:
Closes https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3469

Tracks acceptances of the terms
Tracks how many times it is shown. It uses the end of the URL to differentiate which app caused it to be shown (per requested analytics). Seems to be the best proxy for that.
By taking # shown - # accepted we can get # of those who did not accept.

Per @leannammiller i'm tagging @bshyong and @patrickvinograd for review

I'll add in the config to report these events through to GA as well now. Once this is merged, I'll verify things on staging.


patrickvinograd (approved):
LGTM. I think in general common use should result in the expected app names as the last path component, but it's always possible people will navigate directly to `messaging/compose` or something like that and be presented terms in which case we may see some oddball event names. I think we can revisit and see if that's an issue in practice.

PR #5944, opened 2017-07-03
VBA requested changes 7.3.17

DanielleSoCompany:
Beth: I revised the intro to accommodate the additions of foreign countries and faxing option. Also, I thought we should also probably following the VA before the state name in the links to regional benefit offices, but please revert if you disagree. Finally, I think we need to see how the table adjustments look in heroku preview so let me know if they don't look right and we can adjust further as needed. Thanks!

### DIFF
--- a/content/pages/pension/pension-management-center.md
+++ b/content/pages/pension/pension-management-center.md
@@ -8,11 +8,11 @@ order: 6
 
 <div class="va-introtext">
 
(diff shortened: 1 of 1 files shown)


bethpotts (approved):
Content changes look good!

rroueche (approved):

PR #5949, opened 2017-07-03
Update spouse marriage question to include current marriage

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3642

Also removed Yes/No question because it no longer makes sense.

screen shot 2017-07-03 at 4 24 14 pm

### DIFF
--- a/src/js/common/schemaform/widgets/ArrayCountWidget.jsx
+++ b/src/js/common/schemaform/widgets/ArrayCountWidget.jsx
@@ -1,20 +1,24 @@
 import React from 'react';
 
-export default class SpouseMarriageWidget extends React.Component {
+export default class ArrayCountWidget extends React.Component {
   constructor(props) {
     super(props);
-    this.state = { value: !!props.value
+    this.state = { userCount: !!props.value
       ? props.value.length - (this.props.options.countOffset || 0)
       : undefined };
   }
 
   componentDidUpdate(prevProps, prevState) {
-    if (prevState.value !== this.state.value) {
-      let count = parseInt(this.state.value, 10);
+    if (prevState.userCount !== this.state.userCount) {
+      let count = parseInt(this.state.userCount, 10);
       if (isNaN(count)) {
         count = 0;
       }
 
+      // Too high of a count can crash the browser. We've been using
+      // this for marriage counts and 29 is the record, so 29 seems like
+      // a good upper limit. Filling out more than 29 pages is probably
+      // not reasonable in a form anyway
       if (count > 29) {
         count = 29;
       }
@@ -31,6 +35,10 @@ export default class SpouseMarriageWidget extends React.Component {
 
     const intCount = count + (this.props.options.countOffset || 0);
 
+    if (intCount < 0) {
+      return undefined;
+    }
+
     if (intCount < value.length) {
       return value.slice(0, intCount);
     }
@@ -39,7 +47,7 @@ export default class SpouseMarriageWidget extends React.Component {
   }
 
   updateArrayLength = (event) => {
-    this.setState({ value: event.target.value });
+    this.setState({ userCount: event.target.value });
   }
 
   render() {
@@ -48,7 +56,7 @@ export default class SpouseMarriageWidget extends React.Component {
     if (props.formContext.reviewMode) {
       return (
         <div className="review-row">
-          <dt>{props.uiSchema['ui:title']}</dt><dd>{this.state.value}</dd>
+          <dt>{props.uiSchema['ui:title']}</dt><dd>{this.state.userCount}</dd>
         </div>
       );
     }
@@ -62,7 +70,7 @@ export default class SpouseMarriageWidget extends React.Component {
           disabled={props.disabled}
           autoComplete={props.options.autocomplete || false}
           className={props.options.widgetClassNames}
-          value={typeof this.state.value === 'undefined' ? '' : this.state.value}
+          value={typeof this.state.userCount === 'undefined' ? '' : this.state.userCount}
           onBlur={() => props.onBlur(props.id)}
           onChange={this.updateArrayLength}/>
     );
--- a/src/js/common/schemaform/widgets/ArrayCountWidget.jsx
+++ b/src/js/common/schemaform/widgets/ArrayCountWidget.jsx
@@ -1,45 +1,45 @@
 import React from 'react';
 
-// We're expanding or contracting the array based on the count
-// and returning undefined if the array should be empty
-function getValue(count, value = []) {
-  if (!count || parseInt(count, 10) < 1) {
-    return undefined;
-  }
-
-  const intCount = parseInt(count, 10);
-  if (intCount < value.length) {
-    return value.slice(0, intCount);
-  }
-
-  return value.concat(Array(intCount - value.length).fill({}));
-}
-
-export default class ArrayCountWidget extends React.Component {
+export default class SpouseMarriageWidget extends React.Component {
   constructor(props) {
     super(props);
-    this.state = { value: !!props.value ? props.value.length : undefined };
+    this.state = { value: !!props.value
+      ? props.value.length - (this.props.options.countOffset || 0)
+      : undefined };
   }
 
-  // Need to keep the count state in sync, but don't want to overwrite if the
-  // user has chosen '' or 0
-  componentWillReceiveProps(props) {
-    const arrayData = props.value || [];
-    if (!!arrayData.length && (!this.state.value || this.state.value !== arrayData.length)) {
-      this.setState({ value: arrayData.length });
+  componentDidUpdate(prevProps, prevState) {
+    if (prevState.value !== this.state.value) {
+      let count = parseInt(this.state.value, 10);
+      if (isNaN(count)) {
+        count = 0;
+      }
+
+      if (count > 29) {
+        count = 29;
+      }
+
+      this.props.onChange(this.getValue(count, this.props.value));
     }
+  }
+  // We're expanding or contracting the array based on the count
+  // and returning undefined if the array should be empty
+  getValue = (count, value = []) => {
+    if (count === 0) {
+      return undefined;
+    }
+
+    const intCount = count + (this.props.options.countOffset || 0);
+
+    if (intCount < value.length) {
+      return value.slice(0, intCount);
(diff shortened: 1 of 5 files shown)


src/js/pensions/config/form.js
```diff
@@ -459,6 +459,7 @@ const formConfig = {
           },
           schema: {
             type: 'object',
+            required: ['view:workedBeforeDisabled'],
```
jbalboni: Unrelated to this issue, but I noticed we missed making this required.

cehsu (approved):

src/js/common/schemaform/widgets/ArrayCountWidget.jsx
```diff
@@ -1,45 +1,45 @@
...
-// We're expanding or contracting the array based on the count
-// and returning undefined if the array should be empty
-function getValue(count, value = []) {
-  if (!count || parseInt(count, 10) < 1) {
-    return undefined;
+export default class SpouseMarriageWidget extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = { value: !!props.value
+      ? props.value.length - (this.props.options.countOffset || 0)
```
cvalarida: After reading this through for _waaaay_ too long, I think I finally figured out why we're _subtracting_ the offset. Correct me if I'm wrong:

We start with (usually) a length of 0, so when we have an offset of -1 (as in we know the spouse has 1 marriage more than we'll ask information for), then the value we'll put into the `input` field will be `0 - (-1) = 1`. Later, when we update the field to trigger `getValue()`, that will call `this.props.onChange` with an array sized `this.state.value + this.props.offsetCount` so if we enter 3, the array of marriages we'll ask information for will have a length of `3 + (-1) = 2`.

Also, can we rename `value` to `arrayLength` or something more descriptive?

cehsu: When would the zero be used? Is countOffset not always set at -1 for the spouse?

jbalboni: Other uses of this widget might not have an offset.

test/common/schemaform/widgets/ArrayCountWidget.unit.spec.jsx
```diff
@@ -55,6 +72,27 @@ describe('Schemaform <ArrayCountWidget>', () => {
...
+          disabled={false}
+          formContext={{}}
+          onChange={onChange}
+          options={{
+            countOffset: -1
+          }}/>
+    );
+    tree.subTree('input').props.onChange({
+      target: {
+        value: '3'
```
cehsu: Are we making sure the minimum value is 1?

src/js/common/schemaform/widgets/ArrayCountWidget.jsx
```diff
@@ -1,45 +1,45 @@
...
+  componentDidUpdate(prevProps, prevState) {
+    if (prevState.value !== this.state.value) {
+      let count = parseInt(this.state.value, 10);
+      if (isNaN(count)) {
+        count = 0;
+      }
 
-  return value.concat(Array(intCount - value.length).fill({}));
-}
+      if (count > 29) {
```
cvalarida: This seems pretty arbitrary. Maybe just make a note that the browser will probably die if we need to add more than 29. You had one in there before and I think it'd still be helpful when we come back later (or somebody new entirely looks at it).

test/common/schemaform/widgets/ArrayCountWidget.unit.spec.jsx
jbalboni: I should at least add a test for that.

src/js/common/schemaform/widgets/ArrayCountWidget.jsx
```diff
@@ -1,45 +1,45 @@
...
-  constructor(props) {
-    super(props);
-    this.state = { value: !!props.value ? props.value.length : undefined };
+      this.props.onChange(this.getValue(count, this.props.value));
+    }
   }
+  // We're expanding or contracting the array based on the count
+  // and returning undefined if the array should be empty
+  getValue = (count, value = []) => {
+    if (count === 0) {
```
cvalarida: Is there value in passing a negative count? This will later be used to slice off the end of the array. I'm okay with that, but if we do that without expecting the result we'll get, we may run into trouble. (Also, I just noticed it used to be in here, but you took it out, so I'm wondering why.)

src/js/common/schemaform/widgets/ArrayCountWidget.jsx
```diff
@@ -1,45 +1,45 @@
...
-  // user has chosen '' or 0
-  componentWillReceiveProps(props) {
-    const arrayData = props.value || [];
-    if (!!arrayData.length && (!this.state.value || this.state.value !== arrayData.length)) {
-      this.setState({ value: arrayData.length });
+    if (intCount < value.length) {
+      return value.slice(0, intCount);
     }
+
+    return Array(intCount - value.length).fill({}).concat(value);
```
cvalarida: Adding on the front of the array seems kinda weird, but I guess that's alright.

src/js/common/schemaform/widgets/ArrayCountWidget.jsx
```diff
@@ -1,45 +1,45 @@
...
-    if (count > 29) {
-      count = 29;
-    }
-    this.setState({ value: count }, () => {
-      this.props.onChange(getValue(count, this.props.value));
-    });
+    this.setState({ value: event.target.value });
   }
 
   render() {
```
cvalarida: Should the `min` on the `input` be `countOffset`?

src/js/common/schemaform/widgets/ArrayCountWidget.jsx
```diff
@@ -1,45 +1,45 @@
...
+    }
   }
+  // We're expanding or contracting the array based on the count
+  // and returning undefined if the array should be empty
+  getValue = (count, value = []) => {
+    if (count === 0) {
+      return undefined;
+    }
+
+    const intCount = count + (this.props.options.countOffset || 0);
```
cvalarida: Mmm...what if `abs(count) > abs(countOffset) && countOffset < 0`? So `count = 1` and `countOffset = -2`.

cvalarida (approved):
Well that took a while...I think I get it now, but maybe some comments will aid in reading later. Like, why subtract the offset here and add it there? What is this black magic!?

src/js/common/schemaform/widgets/ArrayCountWidget.jsx
```diff
@@ -1,45 +1,45 @@
 import React from 'react';
 
-// We're expanding or contracting the array based on the count
-// and returning undefined if the array should be empty
-function getValue(count, value = []) {
-  if (!count || parseInt(count, 10) < 1) {
-    return undefined;
+export default class SpouseMarriageWidget extends React.Component {
```
jbalboni: No, I was copying and pasting from different attempts and missed updating this.

src/js/common/schemaform/widgets/ArrayCountWidget.jsx
jbalboni: I think that's right. It's an equation that has to balance:

`userCount + offset = arrayLength`

If we need to get the userCount, then we have to subtract the offset from both sides:

`userCount = arrayLength - offset`

src/js/common/schemaform/widgets/ArrayCountWidget.jsx
jbalboni: Yeah, I have to guess here, and I picked this way because the last one is titled "current marriage" and seems like that's the one they've mostly likely entered correctly.

src/js/common/schemaform/widgets/ArrayCountWidget.jsx
jbalboni: Should be caught now.

src/js/common/schemaform/widgets/ArrayCountWidget.jsx
jbalboni: Without knowing other uses cases for this widget, I'm not sure if that would be correct. I'm going to leave it for now.

src/js/common/schemaform/widgets/ArrayCountWidget.jsx
jbalboni: I think I have this accounted for now.

test/common/schemaform/widgets/ArrayCountWidget.unit.spec.jsx
```diff
@@ -133,6 +133,50 @@ describe('Schemaform <ArrayCountWidget>', () => {
...
+          options={{
+            countOffset: -2
+          }}/>
+    );
+    tree.subTree('input').props.onChange({
+      target: {
+        value: '1'
+      }
+    });
+    expect(onChange.firstCall.args[0]).to.equal(undefined);
```
cvalarida: Start with 2 values, offset -2
Change down to 1 value, offset -2
Expect `undefined` to be returned because the number of values < number of offset.

Got it. The description wasn't super descriptive, but I think it makes sense after reading it. Not sure what a better description would be, though.

test/common/schemaform/widgets/ArrayCountWidget.unit.spec.jsx
```diff
@@ -133,6 +133,50 @@ describe('Schemaform <ArrayCountWidget>', () => {
...
+          options={{
+            countOffset: -1
+          }}/>
+    );
+    tree.subTree('input').props.onChange({
+      target: {
+        value: '1'
+      }
+    });
+    expect(onChange.firstCall.args[0]).to.eql([]);
```
cvalarida: Why should it be `[]`? I was expecting `[{}]`.

jbalboni: Setting the value to one, with an offset of 1, means the array length should be 0.

cvalarida: Duh. Okay, that makes perfect sense. It was `undefined` above because we can't have a negative length, but we can have a length of 0.

Issue #5889, opened 2017-06-28
GI Bill CT: SEO needs improvement


saintsoup52:
Earlier this week the RAND Corporation came out with a report about the Post-9/11 GI Bill.

https://www.rand.org/pubs/research_reports/RR1766.html  

The most interest part of the report to me was the word cloud showing what servicemembers and veterans are google searching when they are researching the GI Bill. “How much is the gi bill” was the most searched GI Bill phrase last year, which has been our working theory with the Comparison Tool all along.

![image](https://user-images.githubusercontent.com/4397508/27638264-a654bea8-5be0-11e7-8fd1-c18cc42e0836.png)

Sadly, if you google "how much is the gi bill" right now…the first three links are to non-VA sites and google highlights information from a non-VA site about the Montgomery GI Bill ironically with a picture of the Post-9/11 GI Bill. 

 
![image](https://user-images.githubusercontent.com/4397508/27638297-bf732e2e-5be0-11e7-8e9a-5c0b8fe2f400.png)

And the GI Bill comparison tool is way down the list and the subtext is the legal disclaimer which does nothing to explain why a user should click on this link versus the 10 above it.

 
![image](https://user-images.githubusercontent.com/4397508/27638301-c538bf7c-5be0-11e7-936c-b4f5e5c9df16.png)

What can we do to improve the GI Bill Comparison Tool's SEO... especially for this search phrase?

saintsoup52:
Sorry forgot to tag folks @leannammiller @melwoodard

ayaleloehr:
Great point @saintsoup52 . Also tagging @robertfairhead and @suzchap as I know this topic is on your radar as well. This is a high traffic page, so could be a good candidate to focus on first for SEO.

robertfairhead:
We can take some basic steps to improve SEO on the pages around metadata and descriptions. I'm still catching up from being out for my move, but I can do that next week. I'm assigning myself.

@leannammiller let me know if there's someone who i should use to review the PR when it's done.

robertfairhead:
I took at a look at this today. I opened a PR that'll link to this that adds description tags and fills in one for the GIBCT landing page. That description is our "suggestion" to Google, Bing, etc on what to use as the short description. Ultimately they will use whatever they want, but that tag is the best way to suggest clear language. I pulled the subhead text from the landing page to use but if there's a preferred alternative, feel free to drop that in the PR.

[...]

PR #5955, opened 2017-07-04
Adding support for og:description

robertfairhead:
Closes https://github.com/department-of-veterans-affairs/vets-website/issues/5889
Closes https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3601
Part of https://github.com/department-of-veterans-affairs/vets.gov-team/issues/612

Adds support for og:description meta tags in our headers
Puts an initial description on the GIBCT landing page

### DIFF
--- a/content/includes/header.html
+++ b/content/includes/header.html
@@ -26,6 +26,7 @@
 {% endif %}
 {% if description %}
 <meta content="{{ description }}" property="og:description">
+<meta content="{{ description }}" name="description">
 {% endif %}
 <meta content="https://www.vets.gov/img/design/logo/og-image.jpg" property="og:image">
 {% if tags %}
--- a/content/includes/header.html
+++ b/content/includes/header.html
@@ -24,6 +24,9 @@
 {% if path %}
 <meta content="https://www.vets.gov/{{ path }}" property="og:url">
 {% endif %}
+{% if description %}
+<meta content="{{ description }}" property="og:description">
+{% endif %}
 <meta content="https://www.vets.gov/img/design/logo/og-image.jpg" property="og:image">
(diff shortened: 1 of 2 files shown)


saintsoup52:
Sorry for not knowing this... But any particular reason why the entryname is just "gi" and not "gi bill"?

Also is there an ideal character limit for the page description? I imagine VA EDU might want to put a little more descriptive explanation of why people should use the comparison tool.

Lastly... since GI Bill is a registered trademark do we need to include the (r) there too?

robertfairhead:
No particular reason on `gi` versus anything else that I'm aware of. The entryname is only really used internally by the front-end build system so it doesn't impact the Veterans, SEO, etc. in any meaningful way. The only restriction on it is that it couldn't have a space.

For the page description, Google (and the other engines) do not publish any specific numbers. That's an internal thing for them. I'm not an SEO specialist, but it's roughly two lines at roughly 75-80 characters per line, so something in the range of 150-160 characters would be reasonable. 

With respect to the registered trademark symbol, the phrase isn't used in the new description and the rest of the page's copy have already been reviewed and approved by the appropriate VA stakeholder, so I wouldn't be in a position to change that here.

ayaleloehr:
pinging @melwoodard , but feel free to redirect if you think there is someone better to review.

melwoodard:
@ayaleloehr makes functional sense to me, though I'd appreciate a look from someone more familiar with meta tags.

cvalarida (approved):
I poked around for a bit, looking at how to make metatags useful, and this looks great.

It did look funny without the self-closing bit in there, but Google tells me that's right.

🚀

PR #5956, opened 2017-07-04
Loading spinner for letters and gibs

annekainicUSDS:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3567

This adds a loading spinner to both letters and gibs until data is returned from the fetch and stored in state. This solves the problem of the component starting to render before data is returned, resulting in a lot of blank or "undefined" field values in the UI for the second(s) it takes for the fetch and store update to complete.

The reason why the components were rendering without all of the data is because the fetch for the data happens in the `componentDidMount`, which is called after the component is mounted and rendering begins. The action to fetch the data would be dispatched from `componentDidMount`, then upon a successful response it would update the store in the reducer. This store update would cause a re-rendering of the component, at which point the data would be rendered in the component.

What this PR does is to check if the app is available (passed down in props), meaning the fetch has returned, before determining what to render. There are three options of what `this.props.available` could be:
1. If `this.props.available === true`, it means the fetch successfully returned the data, so the children components can be rendered. 
2. If `this.props.available === undefined`, it means the fetch has not returned the data yet because this value has not yet been set, and the loading indicator can be rendered.
3. If `this.props.available === false`, it means there was a failure to fetch the data, in which case a generic error message is rendered (need content for this still)

This is a very simplistic way of handling this. In looking at how other apps are doing this on Vets.gov, some actually dispatch separate actions to determine the loading status of the app, and store that status in the store as a boolean value, which is then passed down through props to the components (the normal store update cycle). I felt that this was a bit overkill for simply determining whether or not to render a loading indicator, and didn't want to unnecessarily store another value in state just for the loading status. We have a value in the store already (`available`) that should be enough to tell us what the status of the fetch is, so I wanted to just leverage that. Curious to get thoughts on this approach.

### DIFF
--- a/src/js/letters/containers/Main.jsx
+++ b/src/js/letters/containers/Main.jsx
@@ -14,10 +14,12 @@ class Main extends React.Component {
   render() {
     let appContent;
 
-    if (this.props.destination) {
+    if (this.props.lettersAvailable) {
       appContent = this.props.children;
-    } else {
+    } else if (this.props.lettersAvailable === undefined) {
       appContent = <LoadingIndicator message="Loading your application..."/>;
+    } else {
+      appContent = <div>Warning message: need content for when fetch of data from vets-api fails</div>;
     }
 
     return (
--- a/src/js/letters/containers/Main.jsx
+++ b/src/js/letters/containers/Main.jsx
@@ -1,6 +1,8 @@
 import React from 'react';
 import { connect } from 'react-redux';
 
+import LoadingIndicator from '../../common/components/LoadingIndicator';
+
 import { getBenefitSummaryOptions, getLetterList } from '../actions/letters';
 
 class Main extends React.Component {
@@ -10,9 +12,17 @@ class Main extends React.Component {
   }
 
   render() {
+    let appContent;
+
+    if (this.props.destination) {
+      appContent = this.props.children;
+    } else {
+      appContent = <LoadingIndicator message="Loading your application..."/>;
+    }
+
     return (
       <div className="letters">
-        {this.props.children}
+        {appContent}
       </div>
     );
   }
--- a/src/js/letters/containers/Main.jsx
+++ b/src/js/letters/containers/Main.jsx
@@ -14,12 +14,25 @@ class Main extends React.Component {
   render() {
     let appContent;
 
-    if (this.props.lettersAvailable) {
+    if (this.props.lettersAvailability === 'available') {
       appContent = this.props.children;
-    } else if (this.props.lettersAvailable === undefined) {
-      appContent = <LoadingIndicator message="Loading your application..."/>;
-    } else {
-      appContent = <div>Warning message: need content for when fetch of data from vets-api fails</div>;
+    } else if (this.props.lettersAvailability === 'awaitingResponse') {
+      appContent = <LoadingIndicator message="Loading your letters..."/>;
+    } else if (this.props.lettersAvailability === 'unavailable') {
+      appContent = (
+        <div>
+          <div className="usa-alert usa-alert-error" role="alert">
+            <div className="usa-alert-body">
+              <h4 className="usa-alert-heading">Letters Unavailable</h4>
+              <p className="usa-alert-text">
+                We weren't able to retrieve your VA letters. Please call
+                1-855-574-7286 between Monday-Friday 8:00 a.m. - 8:00 p.m. (ET).
+              </p>
+            </div>
+          </div>
+          <br/>
+        </div>
+      );
     }
 
     return (
@@ -35,7 +48,7 @@ function mapStateToProps(state) {
   return {
     letters: letterState.letters,
     destination: letterState.destination,
-    lettersAvailable: letterState.lettersAvailable,
+    lettersAvailability: letterState.lettersAvailability,
     benefitSummaryOptions: {
       benefitInfo: letterState.benefitInfo,
       serviceInfo: letterState.serviceInfo
--- a/src/js/letters/reducers/index.js
+++ b/src/js/letters/reducers/index.js
@@ -5,7 +5,7 @@ import _ from 'lodash/fp';
 const initialState = {
   letters: [],
   destination: {},
-  lettersAvailable: false,
+  lettersAvailable: undefined,
   benefitInfo: {},
   serviceInfo: [],
   optionsAvailable: false,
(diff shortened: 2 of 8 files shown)


src/js/letters/reducers/index.js
```diff
@@ -4,7 +4,7 @@ import mapValues from 'lodash/mapValues';
 const initialState = {
   letters: [],
   destination: {},
-  lettersAvailable: false,
+  lettersAvailable: undefined,
```
annekainicUSDS: Because I'm now checking if the `available` prop for each app is `true` (if fetch is successful) or `false` (if fetch is unsuccessful), I need to default this the something else for when the fetch has not completed yet.

src/js/post-911-gib-status/containers/StatusPage.jsx
```diff
@@ -10,10 +10,6 @@ import EnrollmentHistory from '../components/EnrollmentHistory';
 import UserInfoSection from '../components/UserInfoSection';
 
 class StatusPage extends React.Component {
-  componentDidMount() {
```
annekainicUSDS: Noticed a duplicate fetch of data here and it didn't seem necessary so removed it.

src/js/letters/containers/Main.jsx
```diff
@@ -10,9 +12,19 @@ class Main extends React.Component {
   }
 
   render() {
+    let appContent;
+
+    if (this.props.lettersAvailable) {
+      appContent = this.props.children;
+    } else if (this.props.lettersAvailable === undefined) {
```
jbalboni: The downside to doing this is that since `lettersAvailable` is a boolean, it's surprising for it to have three different states instead of two. It's also easy to forget and add a check that catches both `false` and `undefined`, since they're both false-y. 

It might not be an issue if these two apps are pretty small and simple, but if you end up needing to add different failure states, you might end up wanting a loading or status property to make the logic clearer.

src/js/post-911-gib-status/containers/Main.jsx
```diff
@@ -9,17 +11,28 @@ class Main extends React.Component {
   }
 
   render() {
+    let appContent;
+
+    if (this.props.available) {
+      appContent = this.props.children;
+    } else if (this.props.available === undefined) {
+      appContent = <LoadingIndicator message="Loading your application..."/>;
```
jbalboni: I think we've been trying to make these messages more specific to what's being loaded.

jbalboni (approved):
Had one note, but this looks good to me. I think this is the simplest approach, but might have some clarity issues in the future.

src/js/post-911-gib-status/containers/Main.jsx
annekainicUSDS: Good point, will update to be more specific!

src/js/letters/containers/Main.jsx
annekainicUSDS: Ah yes good point, in that case perhaps the possible values should be `available`, `unavailable`, `awaitingResponse` or something like that? Although that will mean the checks for `this.props.available` everywhere will have to change and be a little less obvious. And yes agree about perhaps needing a different status if the app grows in complexity.

jbalboni: Yeah, something like that. The checks should still be pretty clear, since you should be making comparisons like `this.props.letterStatus === LETTER_STATUSES.available` (or however you want to create constants for the different statuses).

robbiethegeek-usds:
This looks clean to me, happy to LGTM, if you want to wait on @raquelromano-gov for a final 👍 that makes sense she is more familiar with this code.

src/js/letters/containers/Main.jsx
```diff
@@ -10,9 +12,32 @@ class Main extends React.Component {
...
   render() {
+    let appContent;
+
+    if (this.props.lettersAvailability === 'available') {
+      appContent = this.props.children;
+    } else if (this.props.lettersAvailability === 'awaitingResponse') {
+      appContent = <LoadingIndicator message="Loading your letters..."/>;
+    } else if (this.props.lettersAvailability === 'unavailable') {
+      appContent = (
+        <div>
```
raquelromano-gov: I am guessing we will be using a warning like this in other circumstances, both for letters and for GIBS. We have these warnings  sprinkled throughout vets.gov code--I tried to enumerate some of them in this ticket: https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3243#issuecomment-309521640

I have a secret desire to stop the proliferation of hard-coded messages like this and just have a single shared component that is used everywhere (possibly taking a prop that has the display name of the product, to customize the message). But for now it would be a start to just define this stuff in each app's helpers file.

src/js/letters/reducers/index.js
```diff
@@ -5,7 +5,7 @@ import _ from 'lodash/fp';
 const initialState = {
   letters: [],
   destination: {},
-  lettersAvailable: false,
+  lettersAvailability: 'awaitingResponse',
```
raquelromano-gov: I think we should define constants for these 3 state values. Not sure if in this file or elsewhere, but then import them so all files are referring to the same constants. All-caps would also be good for readability.

src/js/letters/reducers/index.js
```diff
@@ -38,12 +38,12 @@ function letters(state = initialState, action) {
         ...state,
         letters: letterList,
         destination: action.data.meta.address,
-        lettersAvailable: true
+        lettersAvailability: 'available'
       };
     case 'GET_LETTERS_FAILURE':
       // We are currently ignoring this; consider removing once we're sure we've handled
```
raquelromano-gov: Delete stale comment

src/js/post-911-gib-status/containers/Main.jsx
```diff
@@ -9,17 +11,41 @@ class Main extends React.Component {
...
+    let appContent;
+
+    if (this.props.availability === 'available') {
+      appContent = this.props.children;
+    } else if (this.props.availability === 'awaitingResponse') {
+      appContent = <LoadingIndicator message="Loading your Post-9/11 GI Bill benefit information..."/>;
+    } else if (this.props.availability === 'unavailable') {
+      appContent = (
+        <div>
+          <div className="usa-alert usa-alert-error" role="alert">
```
raquelromano-gov: Same comment as above.

raquelromano-gov (approved):
Approving so we can get this merged for the demo, so comments can be addressed in a follow-up PR.

PR #5957, opened 2017-07-04
Post benefit summary letter options

raquelromano-gov:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3599
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3634
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3597

Got this working (ish) locally, connecting to CI, although they are not returning valid BSL or COE pdfs (all others look fine). Requesting the BSL or COE returns this response instead of a PDF blob.

Also fixes problem of extra window opening when download is clicked.

For COE:
```
{
  "messages" : [ {
    "key" : "lettergenerator.serviceError",
    "severity" : "ERROR",
    "text" : "Error calling external service"
  } ]
}
```
For BSL:
```
<!DOCTYPE html>
<html>
        <head>
                <title>Unexpected Error</title>
        </head>
        <body>
                Unexpected Error
        </body>
</html>
```

Also get the above `lettergenerator.serviceError` when fetching the BSL from dev and staging without any options in the request from the command-line:
```
$ curl  --request POST  --header "Authorization: Token token=<token-copied-from-dev-tools>" https://dev-api.vets.gov/v0/letters/benefit_summary
``` 

Also built a review instance, but something is not configured correctly because the vets-website review instance is getting a CORS error for hitting the vets-api review instance:
```
Fetch API cannot load http://48dcc38cb8fb4069a8962c60858cade7-api.review.vetsgov-internal/v0/sessions/new?level=1.
No 'Access-Control-Allow-Origin' header is present on the requested resource.
Origin 'http://48dcc38cb8fb4069a8962c60858cade7.review.vetsgov-internal' is therefore not allowed access.
The response had HTTP status code 502.
If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.
```

Also tested against the mock service just to see if the request is handled with a generic PDF but `POST http://localhost:3000/v0/letters/benefit_summary` returns a 500.

This is the request payload:
```
{"hasNonServiceConnectedPension":false,"hasServiceConnectedDisabilities":true,"hasSurvivorsIndemnityCompensationAward":false,"hasSurvivorsPensionAward":false,"monthlyAwardAmount":true,"serviceConnectedPercentage":true,"awardEffectiveDate":false,"hasAdaptedHousing":false,"hasChapter35Eligibility":false,"hasDeathResultOfDisability":false,"hasIndividualUnemployabilityGranted":false,"hasSpecialMonthlyCompensation":false}
```
And this is the beginning of the vets-api stack trace:
```
{"errors":[{"title":"Internal server error","detail":"Internal server error","code":"500","status":"500","meta":{"exception":"no implicit conversion of nil into String","backtrace":["/usr/local/Cellar/ruby/2.3.1_2/lib/ruby/2.3.0/pathname.rb:409:in `initialize'","/usr/local/Cellar/ruby/2.3.1_2/lib/ruby/2.3.0/pathname.rb:409:in `new'","/usr/local/Cellar/ruby/2.3.1_2/lib/ruby/2.3.0/pathname.rb:409:in `join'","/Users/romano/dsva/vets-api/lib/evss/letters/mock_service.rb:48:in `download_by_type'","/Users/romano/dsva/vets-api/app/controllers/v0/letters_controller.rb:21:in `download'"
```

### DIFF
--- a/src/js/letters/components/DownloadLetterLink.jsx
+++ b/src/js/letters/components/DownloadLetterLink.jsx
@@ -12,52 +12,17 @@ export class DownloadLetterLink extends React.Component {
     this.downloadLetter = this.downloadLetter.bind(this);
   }
 
-  // This opens a duplicate window for some inexplicable reason,
-  // possibly related to some blob response nuance.
-  /*
-  downloadLetter() {
-    const requestUrl = `/v0/letters/${this.props.letterType}`;
-    apiRequest(
-      requestUrl,
-      { method: 'POST' },
-      response => {
-        response.blob().then(blob => {
-        });
-      });
-  */
-
-  // This opens a new duplicate window, focuses on it, and then opens
-  // the PDF in the original window (so it appears in the browser tab to
-  // the left of the current tab, which is weird.
-  /*
-  downloadLetter() {
-    const requestUrl = `/v0/letters/${this.props.letterType}`;
-    apiRequest(
-      requestUrl,
-      { method: 'POST' },
-      response => {
-        response.blob().then(blob => {
-          window.URL = window.URL || window.webkitURL;
-          const downloadUrl = window.URL.createObjectURL(blob);
-          window.location.href = downloadUrl;
-        });
-      });
-  }
-  */
-
-  // This opens a new blank window and renders the pdf in it (to the
-  // right of the current browser as expected, but also tries to open
-  // a window from the original window as above (which may be blocked
-  // by the popup blocker, or if not blocked will pop up an extra window
-  // for no apparent reason).
-  downloadLetter() {
+  // Either download the pdf or open it in a new window, depending on the browser.
+  // Needs to bbe manually tested on a variety of vets.gov-supported
+  // platforms, particularly iOS/Safari
+  downloadLetter(e) {
+    e.preventDefault();
     window.dataLayer.push({
       event: 'letter-download',
       'letter-type': this.props.letterType
     });
-    const requestUrl = `/v0/letters/${this.props.letterType}`;
-    const downloadWindow = window.open();
 
+    const requestUrl = `/v0/letters/${this.props.letterType}`;
     let settings;
     if (this.props.letterType === 'benefit_summary') {
       settings = {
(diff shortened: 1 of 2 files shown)


src/js/letters/components/DownloadLetterLink.jsx
```diff
@@ -140,9 +90,18 @@ class DownloadLetterLink extends React.Component {
   }
 }
 
+function mapStateToProps(state, ownProps) {
+  return {
+    letterType: ownProps.letterType,
```
raquelromano-gov: I made this a connected component that still has its own props since we would rather get the option values from the store but pass in the letter type and name.

src/js/letters/components/DownloadLetterLink.jsx
```diff
@@ -1,132 +1,82 @@
...
-  // right of the current browser as expected, but also tries to open
-  // a window from the original window as above (which may be blocked
-  // by the popup blocker, or if not blocked will pop up an extra window
-  // for no apparent reason).
-  downloadLetter() {
+  // Either download the pdf or open it in a new window, depending on the browser.
+  // Needs to bbe manually tested on a variety of vets.gov-supported
+  // platforms, particularly iOS/Safari
+  downloadLetter(e) {
+    e.preventDefault();
```
raquelromano-gov: This suppresses that extra browser window opening.

test/letters/components/DownloadLetterLink.unit.spec.js
```diff
@@ -3,43 +3,33 @@ import SkinDeep from 'skin-deep';
...
 const defaultProps = {
   letterName: 'Commissary Letter',
   letterType: 'commissary'
 };
 
-
 let oldWindow;
 let oldFetch;
 
+// TODO: fix this warning and improve test coverage for various scenarios:
```
raquelromano-gov: This still needs a ton of work but just trying to get something functional for usability tests tomorrow.

src/js/letters/components/DownloadLetterLink.jsx
```diff
@@ -1,132 +1,84 @@
...
-              // The download attribute is not supported on IE11 or
-              // iOS Safari, so live with the ugly hash name.
               downloadWindow.location.href = downloadUrl;
             }
-            // document.body.removeChild(link);
-            // urlObj.revokeObjectURL(downloadUrl);
-            return Promise.resolve();
           }
-          // Make sure this gets to sentry
-          return Promise.reject(new Error('Cannot download pdf blob'));
```
annekainicUSDS: Not sure I fully understand why we got rid of the Promises here?

annekainicUSDS (approved):
Looks good to me! Hard for me to test locally because I'm still not connected to CI, but the commissary letter still downloaded just fine for me.

Do you think the reason why we're getting the bsl errors is because the military service periods probably aren't being sent correctly yet?

I think it's probably best to merge this as is so we can test in staging and make sure things are working (apart from the bsl) for the user testing session this afternoon. I might also be able to help debug this a bit better once it's in staging.

src/js/letters/components/DownloadLetterLink.jsx
raquelromano-gov: Because I need to figure out exactly what these should be doing instead of guessing :)

annekainicUSDS: Gotcha :) Something for me to figure out too!

PR #5961, opened 2017-07-05
Hard code military service data for purposes of user testing

annekainicUSDS:
Military service data doesn't exist for users in PINT right now. Hard code this data so it works at least for testing with users today.


robbiethegeek-usds (approved):
Looks good for the moment.

raquelromano-gov (approved):

PR #5963, opened 2017-07-05
Hardcode 2 letters for usability testing

raquelromano-gov:


src/js/letters/components/LetterList.jsx
```diff
@@ -9,7 +10,24 @@ import { letterContent } from '../utils/helpers';
 
 class LetterList extends React.Component {
   render() {
-    const letterItems = (this.props.letters || []).map((letter, index) => {
+    // Hard-coding a few extra letters for usability testing purposes. Revert after testing.
+    // const letterItems = (this.props.letters || []).map((letter, index) => {
+    let letters = this.props.letters;
```
raquelromano-gov: Oh wait--this has the problem that the letters response may not have arrived yet, in which case the letter could show up twice. Let me move this into the api request code.

annekainicUSDS (approved):
LGTM! worked for me locally when commenting out those letters in the mock response from vets-api

PR #5966, opened 2017-07-05
Get HCA e2e test working in both dev and production

cvalarida:
HCA currently has save in progress behind a build type flag, so it loads the SIPIntroduction page if we're not in production, and the "regular" intro page if we are in production.

The selector for that first `.click()` had `.schemaform-buttons`which isn't used in the SIPIntroduction page.

FYI: @aub


aub (approved):
If the tests pass then I guess it's LGTM by definition. :)

jbalboni (approved):

PR #5970, opened 2017-07-06
Fix benefit summary request body keys and conditional option text

raquelromano-gov:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3723
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3685

Tested this against CI, and the POST successfully downloads a customized BSL.
Now writing unit tests.

This also includes some refactoring, so here is a summary of the changes:
  - move pdf fetch into action creator so that success or failure actions are dispatched; currently we don't do much with these, but they should be handled with an appropriate error message (we may want to include this in https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3780)
  - removed hard-coded military options for usability test
  - maps the BSL options returned by the initial request to the request body options for customizing the PDF
  - renames `optionsToInclude` to `requestOptions`
  - guts our `apiRequest` helper function so we can use `common/helpers/api/commonApiClient` (which should probably happen in a bunch of places throughout the codebase to reduce code duplication)

### DIFF
--- a/src/js/letters/actions/letters.js
+++ b/src/js/letters/actions/letters.js
@@ -5,12 +5,10 @@ export function getLetterList() {
     apiRequest(
       '/v0/letters',
       null,
-      (response) => {
-        return dispatch({
-          type: 'GET_LETTERS_SUCCESS',
-          data: response,
-        });
-      },
+      response => dispatch({
+        type: 'GET_LETTERS_SUCCESS',
+        data: response,
+      }),
       () => dispatch({ type: 'GET_LETTERS_FAILURE' })
     );
   };
@@ -21,16 +19,75 @@ export function getBenefitSummaryOptions() {
     apiRequest(
       '/v0/letters/beneficiary',
       null,
-      (response) => {
-        return dispatch({
-          type: 'GET_BENEFIT_SUMMARY_OPTIONS_SUCCESS',
-          data: response,
-        });
-      },
+      response => dispatch({
+        type: 'GET_BENEFIT_SUMMARY_OPTIONS_SUCCESS',
+        data: response,
+      }),
       () => dispatch({ type: 'GET_BENEFIT_SUMMARY_OPTIONS_FAILURE' })
     );
   };
 }
+
+export function getLetterPdf(letterType, letterName, letterOptions) {
+  let settings;
+  if (letterType === 'benefit_summary') {
+    settings = {
+      method: 'POST',
+      headers: { 'Content-Type': 'application/json' },
+      body: JSON.stringify(letterOptions)
+    };
+  } else {
+    settings = {
+      method: 'POST'
+    };
+  }
+
+  // We handle IE10 separately but assume all other vets.gov-supported
+  // browsers have blob URL support.
+  // TODO: possibly want to explicitly check for blob URL support with something like
+  // const blobSupported = !!(/^blob:/.exec(downloadUrl));
+  const ie10 = !!window.navigator.msSaveOrOpenBlob;
+  const save = document.createElement('a');
+  let downloadWindow;
+  const downloadSupported = typeof save.download !== 'undefined';
+  if (!downloadSupported) {
+    // Instead of giving the file a readable name and downloading
+    // it directly, open it in a new window with an ugly hash URL
+    downloadWindow = window.open();
+  }
+  let downloadUrl;
+  return (dispatch) => {
+    apiRequest(
+      `/v0/letters/${letterType}`,
+      settings,
+      response => {
+        response.blob().then(blob => {
+          if (ie10) {
+            window.navigator.msSaveOrOpenBlob(blob, letterName);
+          } else {
+            window.URL = window.URL || window.webkitURL;
+            downloadUrl = window.URL.createObjectURL(blob);
+            if (downloadSupported) {
+              // Give the file a readable name if the download attribute is supported.
+              save.download = letterName;
+              save.href = downloadUrl;
+              save.target = '_blank';
+              document.body.appendChild(save);
+              save.click();
+              document.body.removeChild(save);
+            } else {
+              downloadWindow.location.href = downloadUrl;
+            }
+          }
+        });
+        window.URL.revokeObjectURL(downloadUrl); // make sure this doesn't cause problems
+        dispatch({ type: 'GET_LETTER_PDF_SUCCESS' });
+      },
+      () => dispatch({ type: 'GET_LETTER_PDF_FAILURE', data: letterType }) // add something to reducer
+    );
+  };
+}
+
 
 export function updateBenefitSummaryOption(propertyPath, value) {
   return {
--- a/src/js/letters/actions/letters.js
+++ b/src/js/letters/actions/letters.js
@@ -89,9 +89,9 @@ export function getLetterPdf(letterType, letterName, letterOptions) {
 }
 
 
-export function updateBenefitSummaryOption(propertyPath, value) {
+export function updateBenefitSummaryRequestOption(propertyPath, value) {
   return {
-    type: 'UPDATE_BENEFIT_SUMMARY_OPTION',
+    type: 'UPDATE_BENFIT_SUMMARY_REQUEST_OPTION',
     propertyPath,
     value
   };
--- a/src/js/letters/actions/letters.js
+++ b/src/js/letters/actions/letters.js
@@ -83,7 +83,7 @@ export function getLetterPdf(letterType, letterName, letterOptions) {
         window.URL.revokeObjectURL(downloadUrl); // make sure this doesn't cause problems
         dispatch({ type: 'GET_LETTER_PDF_SUCCESS' });
       },
-      () => dispatch({ type: 'GET_LETTER_PDF_FAILURE', data: letterType }) // add something to reducer
+      () => dispatch({ type: 'GET_LETTER_PDF_FAILURE', data: letterType })
     );
   };
 }
--- a/src/js/letters/reducers/index.js
+++ b/src/js/letters/reducers/index.js
@@ -41,7 +41,7 @@ function letters(state = initialState, action) {
       };
     }
     case 'GET_LETTERS_FAILURE':
-      return set('lettersAvailability', 'unavailable', state);
+      return _.set('lettersAvailability', 'unavailable', state);
     case 'GET_BENEFIT_SUMMARY_OPTIONS_SUCCESS': {
       // Gather all possible displayed options that the user may toggle on/off.
       const benefitInfo = action.data.data.attributes.benefitInformation;
--- a/src/js/letters/reducers/index.js
+++ b/src/js/letters/reducers/index.js
@@ -47,7 +47,7 @@ function letters(state = initialState, action) {
       const benefitInfo = action.data.data.attributes.benefitInformation;
       const possibleOptions = [];
       Object.keys(benefitInfo).forEach(key => {
-        if (optionsToAlwaysDisplay.includes(key) || (benefitInfo[key] !== false) &&
+        if ((optionsToAlwaysDisplay.includes(key) || (benefitInfo[key] !== false)) &&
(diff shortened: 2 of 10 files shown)


src/js/letters/reducers/index.js
```diff
@@ -40,34 +39,45 @@ function letters(state = initialState, action) {
...
-      // Default all options to true so they appear checked in the UI
-      options = mapValues(action.data.data.attributes.benefitInformation, (value) => {
-        // If the value of the benefit option is not false, it means the user
-        // is eligible for that value. Default benefit options they are eligible for
-        // to true. Keep benefit options that have a value of false as false, so
-        // they are sent as false in the request for the pdf.
-        return value !== false;
+      return _.set('lettersAvailable', false, state);
+    case 'GET_BENEFIT_SUMMARY_OPTIONS_SUCCESS': {
+      // Gather all possible displayed options that the user may have toggled on/off.
```
annekainicUSDS: At this point they haven't toggled anything though, correct? This happens before the user toggles anything, and is used as a default list of the options that apply to this user that can then be rendered in the UI.

annekainicUSDS (commented):
This is a great start. This type of conditional logic is really hard to handle in a clear, straightforward way. I'm finding the `getBenefitOptionText` helper a little hard to reason through (the inconsistencies with checking just if `currentValue` exists vs. also checking if `currentValue !== 'unavailable'`. It's hard to handle this cleanly because the data we're getting isn't clean or consistent. But what if we had this data structured more similarly to what Rachael has in the ticket? Such as: 
```
const benefitOptionText = {
  hasServiceConnectedDisabilities: {
    true: {
      veteran: 'You have one or more service-connected disabilities.',
      dependent: undefined
    },
    false: {
      veteran: 'You do not have one or more service-connected disabilities.',
      dependent: undefined
    },
  },
  ...
}

getBenefitOptionText(currentOption, currentValue, userType) {
  return benefitOptionText[currentOption][currentValue][userType];
}
```
Then if what `getBenefitOptionText` returns is undefined we know not to display anything. Not sure if this will work or not, and we'll have to handle the `currentValue` differently if it's an actual value as opposed to `true` or `false` (can probably pass this in `getBenefitOptionText(currentOption, currentValue && currentValue !== false, userType)` or something like that). What do you think?

src/js/letters/utils/helpers.js
```diff
@@ -58,21 +58,75 @@ export const letterContent = {
...
+    hasChapter35Eligibility: currentValue ?
+      <div>You are considered to be totally and permanently disabled solely due to your service-connected disabilities</div> :
+      <div>Information on your totally and permanently disabled status which resulted from service-connected disabilities</div>, // This doeson't look right; clarify in https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3685
+    monthlyAwardAmount: (currentValue && currentValue !== 'unavailable') ?
+      <div>Your current monthly award amount is <strong>${currentValue}</strong></div> : undefined
   };
 
-  return textForAllOptions[currentOption];
+  if (isVeteran) {
+    merge({
```
raquelromano-gov: OK, will take a look.

raquelromano-gov:
I like the idea of restructuring the function to get the display text! Will make things much more readable.  I'll give that a shot later today.

src/js/letters/reducers/index.js
raquelromano-gov: Yeah, this should say "all possible displayed options that the user may toggle on/off"

src/js/letters/actions/letters.js
```diff
@@ -21,17 +19,76 @@ export function getBenefitSummaryOptions() {
...
+            } else {
+              downloadWindow.location.href = downloadUrl;
+            }
+          }
         });
+        window.URL.revokeObjectURL(downloadUrl); // make sure this doesn't cause problems
+        dispatch({ type: 'GET_LETTER_PDF_SUCCESS' });
       },
-      () => dispatch({ type: 'GET_BENEFIT_SUMMARY_OPTIONS_FAILURE' })
+      () => dispatch({ type: 'GET_LETTER_PDF_FAILURE', data: letterType })
```
raquelromano-gov: We should decide what to do about failed PDF downloads and add to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3780

src/js/letters/utils/helpers.js
```diff
@@ -58,21 +37,165 @@ export const letterContent = {
...
+      dependent: undefined
+    }
+  },
+  hasChapter35Eligibility: {
+    'true': {
+      veteran: <div>You are considered to be totally and permanently disabled solely due to your service-connected disabilities</div>,
+      dependent: <div>You <strong>are</strong> eligible for Dependents' Educational Assistance (Chapter 35).</div>
+    },
+    'false': {
+      // This doesn't look right: getting clarification in
```
raquelromano-gov: https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3685 has some updates (bolding and language), but still has one detail to be resolved.

src/js/letters/actions/letters.js
annekainicUSDS: Yes, great point. Probably a question for @rroueche and @phiden. I assume display some kind of error message to the user?

src/js/letters/utils/helpers.js
```diff
@@ -58,21 +37,169 @@ export const letterContent = {
...
-    hasServiceConnectedDisabilities: <div>You have one or more service-connected disabilities</div>,
-    hasSpecialMonthlyCompensation: <div>You <strong>are</strong> service-connected for loss of or loss of use of a limb, or you are totally blind in or missing at least one eye</div>,
-    hasSurvivorsIndemnityCompensationAward: <div></div>,
-    hasSurvivorsPensionAward: <div></div>,
-    monthlyAwardAmount: <div>Your current monthly award amount is <strong>${currentValue}</strong></div>,
-    serviceConnectedPercentage: <div>Your combined service-connected evaluation is <strong>{currentValue}%</strong></div>,
-  };
+// Options returned by the benefit summary letter request that should be offered in
+// the checkbox list regardless of their values (e.g., true, false, 'unavailable', or other)
+// All other options are conditionall displayed, depending on the value
```
annekainicUSDS: Minor typo: `conditionall` should be `conditionally`

src/js/letters/utils/helpers.js
```diff
@@ -58,21 +37,169 @@ export const letterContent = {
...
+      }
+      return undefined;
+    case 'serviceConnectedPercentage':
+      if (value && value !== 'unavailable') {
+        if (isVeteran) {
+          return (<div>Your combined service-connected evaluation is <strong>{value}%</strong></div>);
+        }
+        return (<div>The Veteran's combined service-connected evaluation is <strong>{value}%</strong></div>);
+      }
+      break;
```
annekainicUSDS: Why the break here?

src/js/letters/utils/helpers.js
```diff
@@ -58,21 +37,169 @@ export const letterContent = {
...
+  hasDeathResultOfDisability: 'deathResultOfDisability',
+  hasIndividualUnemployabilityGranted: 'unemployable',
+  hasNonServiceConnectedPension: 'nonServiceConnectedPension',
+  hasServiceConnectedDisabilities: 'serviceConnectedDisabilities',
+  hasSpecialMonthlyCompensation: 'specialMonthlyCompensation',
+  // A given user should only see one of these survivor award options and never both,
+  // so both map to the same request body option
+  hasSurvivorsIndemnityCompensationAward: 'survivorsAward',
+  hasSurvivorsPensionAward: 'survivorsAward',
+  awardEffectiveDate: 'monthlyAward',
```
annekainicUSDS: I don't think this one is right. From what Ian told me (and we might want to triple check this), `awardEffectiveDate` shouldn't appear as a clickable option in the list because it's not something they can optionally include in their letter or not. Therefore it doesn't map to anything that we send in the request for the pdf for bsl.

src/js/letters/utils/helpers.js
```diff
@@ -58,21 +37,169 @@ export const letterContent = {
...
+    valueString = 'true';
+  } else {
+    valueString = value;
+  }
+
+  if (!['awardEffectiveDate', 'monthlyAwardAmount', 'serviceConnectedPercentage'].includes(option)) {
+    return benefitOptionText[option][valueString][personType];
+  }
+  switch (option) {
+    case 'awardEffectiveDate':
```
annekainicUSDS: Not sure we have to worry about `awardEffectiveDate` (see comment below), let's confirm with @rroueche

annekainicUSDS (approved):
Looks great! Tested locally well too (only thing was because `awardEffectiveDate` and `monthlyAward` are being mapped to the same thing, they both toggle when one is clicked, but I think we can remove `awardEffectiveDate` anyway, something to confirm with @rroueche). Awesome PR!

src/js/letters/utils/helpers.js
raquelromano-gov: Ah, OK I did not know that--thanks. I will remove it and double-check with Rachael tomorrow.

I also plan to add more unit test to the components before merging, but will try to merge this tomorrow.

raquelromano-gov: Actually, the request body option `monthlyAward` for generating the BSL seems to simultaneously control whether both of these appear in the PDF (the award amount and the award effective date). I've asked in #evss channel if this is by design, and if this is what is done in eBenefits.

raquelromano-gov:
Another edge case that I think should be considered in a separate PR (maybe once we figure out the `awardEffectiveDate` issue: if all options are unselected, the letter looks like this:
image

Maybe if EVSS changes the letter generator service to always show the award effective date (as opposed to coupling it with the award amount), then that date can always be shown in the "VA Benefit Information" section

test/letters/containers/VeteranBenefitSummaryLetter.unit.spec.jsx
```diff
@@ -39,11 +42,39 @@ describe('<VeteranBenefitSummaryLetter>', () => {
 
   it('should show benefit info options', () => {
     const tree = SkinDeep.shallowRender(<VeteranBenefitSummaryLetter store={store} {...defaultProps}/>);
-    expect(tree.subTree('hasAdaptedHousing')).to.exist;
+    const rows = tree.dive(['div', '#benefitInfoTable', 'tbody']).everySubTree('tr');
```
raquelromano-gov: This is all more brittle than I would like, even though I added an ID to each table to make it simpler. I think I've spent too much time on this test already, so wondering what you think about making any testing improvements in a separate PR.

src/js/letters/actions/letters.js
phiden: do we have a global error message for failed downloads from blue button or similar we could use?

annekainicUSDS: I have no idea. Perhaps you could ask someone on the Kudos team?

phiden: asking!

annekainicUSDS (approved):
Looks good! I'm having trouble testing this locally because I seem to be having some trouble with id.me. The only consideration is last time I tested this, because `awardEffectiveDate` and `monthlyAward` are mapped to the same field, they both toggle when you check or uncheck one of them. Not a big deal temporarily if we're eventually changing how `awardEffectiveDate` is going to work, but something we'll definitely have to revisit at some point. Perhaps add a TODO so we remember to get back to that?

raquelromano-gov:
Yeah idme is misbehaving for me too--I dropped a note to Tanel in the #identity channel.

For now, I think the coupling of those 2 checkboxes is the best way to go given what Ian has said about what the PDF generator service. I tried to get him to explain in more detail how it works on eBen right now, but it was hard since we had multiple threads.

I think the best thing to do is to get this on staging, show Rachael and explain the behavior of that service, compare to eBen, make a product decision, and go back to EVSS with any requests.

PR #5975, opened 2017-07-06
Link to appeals correctly

aub:
This ugly line of code creates links to either an appeal or a claim, but previously was always using the same form of URL for both, so it didn't work correctly for appeals.


U-DON (approved):
LGTM.

U-DON:
Certainly an ugly line. Could be refactored later to be cleaner.

PR #5977, opened 2017-07-06
Reformat va file number to hide everything but last 4 digits

annekainicUSDS:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3744.

This replaces everything but the last 4 digits in the VA file number with x's. I couldn't make it look exactly like the mock-up, which includes 2 hyphens (e.g. `xxx-xx-6789`) because I won't know how many total digits the number is going to be. While VA file number is usually just going to be the user's ssn, sometimes it's not, and original file numbers can be a range of possible lengths. Therefore, what this will render is x's for each digit up until the last 4, a hyphen, and then the last 4 digits (e.g., `xxxx-5678`).

### DIFF
--- a/src/js/post-911-gib-status/components/UserInfoSection.jsx
+++ b/src/js/post-911-gib-status/components/UserInfoSection.jsx
@@ -4,7 +4,7 @@ import PropTypes from 'prop-types';
 import InfoPair from './InfoPair';
 
 import { formatDateShort } from '../../common/utils/helpers';
-import { formatPercent, formatVAFilNumber } from '../utils/helpers';
+import { formatPercent, formatVAFileNumber } from '../utils/helpers';
 
 class UserInfoSection extends React.Component {
   render() {
@@ -78,7 +78,7 @@ class UserInfoSection extends React.Component {
           {/* TODO: find out whether this should be only partially displayed  xxxx1234 */}
           <InfoPair
               label="VA file number"
-              value={formatVAFilNumber(enrollmentData.vaFileNumber)}
+              value={formatVAFileNumber(enrollmentData.vaFileNumber)}
               spacingClass="section-line"/>
           <InfoPair
               label="Regional processing office"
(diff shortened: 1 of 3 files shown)


src/js/post-911-gib-status/utils/helpers.js
```diff
@@ -12,6 +12,11 @@ export function formatPercent(percent) {
   return validPercent;
 }
 
+export function formatVAFilNumber(number) {
+  const lengthOfXString = number.length - 4;
```
omgitsbillryan: `va_file_number` may be `null`.  Is that accounted for at some earlier place?  Otherwise, this will throw a:
```
Uncaught TypeError: Cannot read property 'length' of null
```

omgitsbillryan (approved):
LGTM with a small comment.

src/js/post-911-gib-status/components/UserInfoSection.jsx
```diff
@@ -4,7 +4,7 @@ import PropTypes from 'prop-types';
 import InfoPair from './InfoPair';
 
 import { formatDateShort } from '../../common/utils/helpers';
-import { formatPercent } from '../utils/helpers';
+import { formatPercent, formatVAFilNumber } from '../utils/helpers';
```
raquelromano-gov: missing e in File

raquelromano-gov:
I didn't realize the VA File Number could vary in length. Was going to suggest that if the string length were exactly 9, we could assume it's an SSN and format like xxx-xx-1234 and otherwise just do xxxx-1234 for all other lengths (as it's implemented now)

But I think this is fine for now!

src/js/post-911-gib-status/components/UserInfoSection.jsx
annekainicUSDS: Thanks! Fixed

src/js/post-911-gib-status/utils/helpers.js
annekainicUSDS: Thanks! Fixed

PR #5978, opened 2017-07-06
Wizard fix

cehsu:
- Deduplicates update-non-dependent id
- Enables multiple alternates

### DIFF
--- a/src/js/edu-benefits/education-wizard.js
+++ b/src/js/edu-benefits/education-wizard.js
@@ -41,12 +41,14 @@ function reInitWidget() {
 
   content.addEventListener('change', (e) => {
     const radio = e.target;
-    const otherChoice = radio.dataset.alternate;
-    const otherNextQuestion = container.querySelector(`#${otherChoice}`).dataset.nextQuestion;
-    if (otherNextQuestion) {
-      const otherNextQuestionElement = container.querySelector(`[data-question=${otherNextQuestion}]`);
-      closeStateAndCheckChild(otherNextQuestionElement, container);
-    }
+    const otherChoices = radio.dataset.alternate.split(' ');
+    otherChoices.forEach((otherChoice) => {
+      const otherNextQuestion = container.querySelector(`#${otherChoice}`).dataset.nextQuestion;
+      if (otherNextQuestion) {
+        const otherNextQuestionElement = container.querySelector(`[data-question=${otherNextQuestion}]`);
+        closeStateAndCheckChild(otherNextQuestionElement, container);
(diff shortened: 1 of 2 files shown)


cehsu:
The original test failure that prompted this is no longer a problem, but these changes still make sense.

src/js/edu-benefits/education-wizard.js
```diff
@@ -41,12 +41,14 @@ function reInitWidget() {
 
   content.addEventListener('change', (e) => {
     const radio = e.target;
-    const otherChoice = radio.dataset.alternate;
-    const otherNextQuestion = container.querySelector(`#${otherChoice}`).dataset.nextQuestion;
-    if (otherNextQuestion) {
-      const otherNextQuestionElement = container.querySelector(`[data-question=${otherNextQuestion}]`);
-      closeStateAndCheckChild(otherNextQuestionElement, container);
-    }
+    const otherChoices = radio.dataset.alternate.split(' ');
```
jbalboni: Why do this instead of making the name something like `update-non-dependent-transfer`?

cehsu: My understanding is that benefits are transferred from sponsors to dependents, but if that is mistaken, `update-non-dependent-transfer` would be clearer. Perhaps @goldenmeanie can advise, unless you're familiar with benefits be transferred to non-dependents.

goldenmeanie: I believe this is governing this part of the wizard

transfered-or-not

First option is not transferred (not a dependent)
Second option is transferred (a dependent/spouse)
Third option is not transferred (a dependent)

Third option is not considered transferred even though it is using a sponsor’s eligibility. The sponsor was not available to officially transfer the benefit. Does this help?

cehsu: This is very helpful, thanks @goldenmeanie, @jbalboni is the current name ok given the provided explanation?

jbalboni: My main concern there was just if we could come up with a name that didn’t require you to add the code to split dataset.alternate and loop through each item. If that doesn't make sense or isn't feasible, this looks fine to me.

jbalboni (approved):
Had one comment to see if we can remove some of the code that's added, but otherwise looks fine.

src/js/edu-benefits/education-wizard.js
jbalboni: I was misunderstanding how `data-alternate` was being used; this change makes sense.

PR #5980, opened 2017-07-07
Update MHV account creation catch-all error message

patrickvinograd:
Fixes https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3578

Just updating the catch-all error message per guidance from Danielle.

### DIFF
--- a/src/js/health-records/components/ErrorView.jsx
+++ b/src/js/health-records/components/ErrorView.jsx
@@ -22,7 +22,7 @@ class ErrorView extends React.Component {
       title = "We couldn't access your health tools";
       detail = (
         <p>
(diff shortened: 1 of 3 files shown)


U-DON (approved):
LGTM. Although, are the curly single quotes intentional? If they are, it's fine; just wanted to point it out in case there wasn't a reason to use them. Could miss out on matches when searching for these phrases with the usual single quote.

U-DON:
There also seems to be a mix of curly and regular single quotes in the content.

patrickvinograd:
Whoops, probably copy-paste problem. Good catch. Will fix.

PR #5981, opened 2017-07-07
Set page focus on appeals detail page

bshyong:
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3752

Sets page focus on Appeals detail page to help with tab ordering issues


U-DON (approved):
LGTM. I wonder why disability-benefits is the only app to use `setPageFocus`, though. Seems like it would be an issue anywhere else when content is rendered by React when navigating.

bshyong:
@U-DON yes I think those page utils could be used site-wide at some point. Seems like an improvement that the original creators made but the other teams have not identified this as an issue or it didn't come up during testing

PR #5983, opened 2017-07-07
Add progress bar and file type check to file upload field

jbalboni:
This addresses a few things I noticed in testing:

- The occasional slowness of staging makes it clear that putting up "Uploading files..." is not a good solution to telling the user to wait while files upload. So, this adds a progress bar. However, to get progress events, you can't use fetch, you have to use xhr. So that's most of the changes here.
- The `accepts` property on file inputs is really just a guideline, not a rule, so we need to actual prevent files that aren't the right type on the front end.
- IE11 was centering some of the buttons, because I used `left` instead of `flex-start`. The former is incorrect, but Chrome looked correct with it.

screen shot 2017-07-07 at 4 50 25 pm

### DIFF
--- a/src/js/common/schemaform/FileField.jsx
+++ b/src/js/common/schemaform/FileField.jsx
@@ -54,12 +54,11 @@ export default class FileField extends React.Component {
         filePath,
         this.props.uiSchema['ui:options'],
         this.updateProgress
-      )
-        .then(() => {
-          // rather not use the promise here, but seems better than trying to pass
-          // a blur function
-          this.props.onBlur(this.props.idSchema.$id);
-        });
+      ).then(() => {
+        // rather not use the promise here, but seems better than trying to pass
+        // a blur function
+        this.props.onBlur(this.props.idSchema.$id);
+      });
     }
   }
 
--- a/src/js/common/schemaform/FileField.jsx
+++ b/src/js/common/schemaform/FileField.jsx
@@ -11,7 +11,6 @@ export default class FileField extends React.Component {
     const files = this.props.formData || [];
     this.state = {
       editing: files.map(() => false),
-      isUploading: false,
       progress: 0
     };
   }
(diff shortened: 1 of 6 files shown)


src/sass/modules/_m-schemaform.scss
```diff
@@ -284,7 +284,7 @@ label {
   }
   > .schemaform-file-item-edit {
     flex-direction: column;
-    align-items: left;
```
cehsu: Should we automate this check somehow (e.g. [doiuse](https://github.com/anandthakker/doiuse))?

cehsu (approved):

src/js/common/schemaform/FileField.jsx
```diff
@@ -41,15 +49,24 @@ export default class FileField extends React.Component {
...
       const filePath = this.props.idSchema.$id.split('_').slice(1).concat(idx);
-      this.props.formContext.uploadFile(event.target.files[0], filePath, this.props.uiSchema['ui:options'])
-        .catch(() => {
+      this.props.formContext.uploadFile(
+        event.target.files[0],
+        filePath,
+        this.props.uiSchema['ui:options'],
+        this.updateProgress
+      )
+        .then(() => {
```
cehsu: Does this indentation make sense?

src/sass/modules/_m-schemaform.scss
jbalboni: I don't think it would help in this case since I just messed up a property value, but that seems like a really useful tool.

test/common/schemaform/FileField.unit.spec.jsx
```diff
@@ -235,10 +263,9 @@ describe('Schemaform <FileField>', () => {
...
 
-    expect(uploadFile.firstCall.args).to.eql([
-      {},
-      ['fileField', 0],
-      uiSchema['ui:options']
-    ]);
+    expect(uploadFile.firstCall.args[0]).to.eql({});
+    expect(uploadFile.firstCall.args[1]).to.eql(['fileField', 0]);
+    expect(uploadFile.firstCall.args[2]).to.eql(uiSchema['ui:options']);
+    expect(uploadFile.firstCall.args[3]).to.be.a('function');
```
cvalarida: I take it this is why you can't use `expect(...).to.eql()`, yeah?

cvalarida (approved):
🚀

test/common/schemaform/FileField.unit.spec.jsx
jbalboni: Yep!

PR #5984, opened 2017-07-08
Fix FL copy issue

bshyong:
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3344

Minor change to fix date presentation issue


U-DON (approved):
LGTM.

PR #5986, opened 2017-07-08
Do not move FL pins after clicking to not allow geolocation

bshyong:
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2947

Do not move FL pins after clicking to not allow geolocation


U-DON (approved):
LGTM.

PR #5987, opened 2017-07-08
Set up pension e2e test scaffold

cehsu:
Connects to department-of-veterans-affairs/vets.gov-team#3736


cvalarida (approved):
Awesome! The only other thing I'd ask for is a _full_ maximal-test.json, but with pensions, that's...a lot of work. Ideally, this would be gotten by actually filling out the form and taking either the submission or the data from the redux store (using the wonderful redux dev tools).

PR #5988, opened 2017-07-10
GIBS e2e test

annekainicUSDS:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3381

### DIFF
--- a/test/e2e/post-911-gib-status-helpers.js
+++ b/test/e2e/post-911-gib-status-helpers.js
@@ -71,8 +71,8 @@ const enrollmentData = {
 };
 
 // Create API routes
-function initApplicationMock() {
-  mock(null, {
+function initApplicationMock(token) {
+  mock(token, {
     path: '/v0/post911_gi_bill_status',
     verb: 'get',
     value: enrollmentData
--- a/test/e2e/post-911-gib-status-helpers.js
+++ b/test/e2e/post-911-gib-status-helpers.js
@@ -1,73 +1,77 @@
 const mock = require('./mock-helpers');
 
 const enrollmentData = {
-  dataOfBirth: '1995-11-12T06:00:00.000+0000',
-  delimitingDate: '2017-12-07T05:00:00.000+0000',
-  eligibilityDate: '2016-12-01T05:00:00.000+0000',
-  enrollments: [
-    {
-      amendments: [],
-      beginDate: '2012-11-01T04:00:00.000+00:00',
-      endDate: '2012-12-01T05:00:00.000+00:00',
-      facilityCode: '11902614',
-      facilityName: 'PURDUE UNIVERSITY',
-      fullTimeCreditHourUnderGrad: null,
-      fullTimeHours: 12,
-      hourType: null,
-      onCampusHours: 12,
-      onlineHours: 0,
-      participantId: '11170323',
-      status: 'Approved',
-      termId: null,
-      trainingType: 'UnderGrad',
-      vacationDayCount: 0,
-      yellowRibbonAmount: 0
-    },
-    {
-      amendments: [
+  data: {
+    attributes: {
+      firstName: 'First',
+      lastName: 'Last',
+      nameSuffix: 'II',
+      originalEntitlement: 100,
+      percentageBenefit: 25,
+      regionalProcessingOffice: 'Central Office Washington, DC',
+      remainingEntitlement: 25,
+      usedEntitlement: 75,
+      vaFileNumber: '12345678',
+      dateOfBirth: '1995-11-27T05:00:00.000+0000',
+      delimitingDate: '2017-12-07T05:00:00.000+0000',
+      eligibilityDate: '2016-12-01T05:00:00.000+0000',
+      enrollments: [
         {
-          changeEffectiveDate: '2017-06-06T17:01:03.926Z',
+          amendments: [],
+          beginDate: '2012-11-01T04:00:00.000+00:00',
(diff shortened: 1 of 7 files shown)


test/post-911-gib-status/00-basic-flow.e2e.spec.js
```diff
@@ -0,0 +1,33 @@
...
+module.exports = E2eHelpers.createE2eTest(
+  (client) => {
+    const token = LoginHelpers.getUserToken();
+
+    GibsHelpers.initApplicationMock(token);
+
+    AccountCreationHelpers.initMHVTermsMocks(token);
+
+    // Test flow for unauthed and LOA1 users
+    LoginHelpers.testUnauthedUserFlow(client, '/education/gi-bill/post-9-11/ch-33-benefit');
```
raquelromano-gov: I would separate out the unauthed and authed flow into 2 separate tests.

The unauthed test ca probably be nearly identical to this: https://github.com/department-of-veterans-affairs/vets-website/blob/master/test/disability-benefits/e2e/00.unauthed-user.e2e.spec.js

Then this test can do 
```
LoginHelpers.logIn(token, client, '/education/gi-bill/post-9-11/ch-33-benefit', 3)
```
(or whatever the URL currently is)

test/post-911-gib-status/01-authed.e2e.spec.js
```diff
@@ -0,0 +1,40 @@
...
+
+    // Print page
+    client
+      .click('#print-button')
+      .windowHandles((result) => {
+        const newWindow = result.value[1];
+        client.switchWindow(newWindow);
+      });
+
+    // Ensure main status page renders.
```
raquelromano-gov: Change comment to "Ensure print page renders"

We probably want a comment explaining that Electron loses the session data so in the test environment we need to log back in to see the print page.

test/e2e/post-911-gib-status-helpers.js
```diff
@@ -0,0 +1,89 @@
+const mock = require('./mock-helpers');
+
+const enrollmentData = {
```
raquelromano-gov: Saw the intermediate commit where you were looking for specific data values. Maybe something in between is just to look for a single data-dependent item in each section?

test/post-911-gib-status/01-authed.e2e.spec.js
```diff
@@ -0,0 +1,48 @@
...
+        client.switchWindow(newWindow);
+      });
+
+    // Ensure print page renders.
+    // We need to log in again here because Electron loses the session
+    // when opening a new window.
+    LoginHelpers.logIn(token, client, '/education/gi-bill/post-9-11/ch-33-benefit/print', 3)
+      .waitForElementVisible('body', Timeouts.normal)
+      .axeCheck('.main')
+      .assert.title('Check Benefit: Vets.gov')
```
raquelromano-gov: Looks like this is where it fails on Jenkins:
```
 ✖ Testing if the page title equals "Check Benefit: Vets.gov".  - expected "Check Benefit: Vets.gov" but got: "Error"
    at /application/test/post-911-gib-status/01-authed.e2e.spec.js:39:15

```
2 ideas, though both could be red herring:

1) From reading the same pages you probably read about `windowHandles` it looks like sometimes people call `client.switchWindow()` and other times `this.switchWindow()`. Is there a difference?

2) I wonder if you put this entire block inside the callback passed to `windowHandles` how it would behave. Again, not sure whether to use `client` or `this`

Otherwise,  I'm stumped too....maybe time to send a plea to #vetsgov-fe pointing to the jenkins log and this PR

annekainicUSDS: Tried both of these suggestions and neither has passed Jenkins :(. What has most recently been pushed is moving the entire block into `.window_handles` (I also changed it from `windowHandles` to `window_handles` because that's what the nightwatch docs specify: http://nightwatchjs.org/api/switchWindow.html).

Everything is still passing locally, just not in Jenkins. Might ping @robbiethegeek to see if he can take a look at this and see if anything else might be going on.

raquelromano-gov:
😞  same error. I'm stumped.

annekainicUSDS:
Commented out print page for now, will merge this without that check and revisit that aspect next week.

test/post-911-gib-status/01-authed.e2e.spec.js
```diff
@@ -0,0 +1,49 @@
...
+    //     LoginHelpers.logIn(token, client, '/education/gi-bill/post-9-11/ch-33-benefit/print', 3)
+    //       .waitForElementVisible('body', Timeouts.normal)
+    //       .axeCheck('.main')
+    //       .assert.title('Check Benefit: Vets.gov')
+    //       .waitForElementVisible('.print-status', Timeouts.slow);  // First render of React may be slow.
+
+    //     client.expect.element('.section-header').text.to.contain('Post-9/11 GI Bill Benefit Information');
+    //   });
+
+    client.axeCheck('.main');
```
raquelromano-gov: This looks like an extra check that's not needed since it's called on line 15.

test/e2e/post-911-gib-status-helpers.js
raquelromano-gov: This isn't critical, but since we're just testing for one thing to be rendered (since unit tests are testing the details), we can probably simplify this object to just a couple of fields.

raquelromano-gov (approved):

PR #5989, opened 2017-07-10
Pull in some claims/appeals changes

aub:
This brings in some of the changes for department-of-veterans-affairs/vets.gov-team#3729 that can be merged in before the launch of appeals. For some of those changes, we don't wish to make them live until after the launch, but for others we can go ahead.


bshyong (approved):
LGTM

PR #5994, opened 2017-07-10
Chunk manifest compatibility for CSP

jkassemi:
CSP script-src currently requires unsafe-inline, which is considered overly permissive. This change removes the inline chunk manifest declaration and instead writes the manifest to a js file later included with a script tag.

### DIFF
--- a/content/includes/header.html
+++ b/content/includes/header.html
@@ -67,11 +67,7 @@
 <!-- We participate in the US government's analytics program. See the data at analytics.usa.gov. -->
 <script async type="text/javascript" src="https://dap.digitalgov.gov/Universal-Federated-Analytics-Min.js?agency=VA" id="_fed_an_ua_tag"></script>
 
-<script>
-//<![CDATA[
-window.webpackManifest = 'CHUNK_MANIFEST_PLACEHOLDER';
-//]]>
-</script>
+<script src="/generated/chunk-manifest.js"/></script>
 
 <script src="/generated/vendor.js"></script>
 {% if entryname %}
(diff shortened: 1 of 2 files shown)


content/includes/header.html
```diff
@@ -67,11 +67,7 @@
 <!-- We participate in the US government's analytics program. See the data at analytics.usa.gov. -->
 <script async type="text/javascript" src="https://dap.digitalgov.gov/Universal-Federated-Analytics-Min.js?agency=VA" id="_fed_an_ua_tag"></script>
 
-<script>
-//<![CDATA[
-window.webpackManifest = 'CHUNK_MANIFEST_PLACEHOLDER';
-//]]>
-</script>
+<script src="/generated/chunk-manifest.js"/></script>
```
jkassemi: slight issue here - since the file doesn't exist in development we get a 404

ayaleloehr:
Ping @jkassemi and @aub

jkassemi:
FWIW the 404 issue really doesn't cause any problems, I was just noting it. Safe for review/merge.

content/includes/header.html
ayaleloehr: would it make sense to do this async or no?

aub (approved):
This looks fine to me.

wyattwalter:
Do we still want to do this?

jkassemi:
cc @edpaget due to recent work on the CSP.

Assigned to @leannammiller for prioritization.

edpaget:
This shouldn't be necessary any more.

PR #6006, opened 2017-07-11
Restore alert on no messaging recipients

patrickvinograd:
Addresses the SM part of https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3657

Commit history looks like first there was an attempt to show this alert if no recipients, then there was a change to prevent loading the entire app if loading folders or recipients failed, but they kind of conflicted. User should still be able to load app if they have an empty recipients list (versus loading recipients list failed). 

This restores that functionality and also goes back to disabling the compose button. If the user somehow navigates directly to the `/compose` screen, they still get the alert there as well. 

screen shot 2017-07-11 at 1 02 09 pm
screen shot 2017-07-11 at 1 02 23 pm

### DIFF
--- a/src/js/messaging/containers/Main.jsx
+++ b/src/js/messaging/containers/Main.jsx
@@ -113,7 +113,7 @@ export class Main extends React.Component {
             <ButtonClose
                 className="messaging-folder-nav-close"
                 onClick={this.props.toggleFolderNav}/>
-            <ComposeButton/>
+            <ComposeButton disabled={_.isEmpty(this.props.recipients)}/>
             <FolderNav
                 currentFolderId={this.props.currentFolderId}
                 folders={this.props.folders}
--- a/src/js/messaging/containers/Main.jsx
+++ b/src/js/messaging/containers/Main.jsx
@@ -113,7 +113,7 @@ export class Main extends React.Component {
             <ButtonClose
                 className="messaging-folder-nav-close"
                 onClick={this.props.toggleFolderNav}/>
-            <ComposeButton disabled={_.isEmpty(this.props.recipients)}/>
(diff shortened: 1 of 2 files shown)


bshyong (approved):
LGTM 🚢

ayaleloehr:
I haven't looked into this far enough to really know, but I got a strange feeling on this one so wanted to comment. 

[This PR](https://github.com/department-of-veterans-affairs/vets-website/pull/5742) about a month ago fixed an issue where the Compose button was grey when it shouldn't have been (because an API call either failed or hadn't returned yet, I believe). I just want to make sure we aren't going back to the [issue](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3057) that caused that PR to go in in the first place. 

Please ignore if irrelevant!

U-DON (approved):
LGTM. This does revert the change I made to no longer disable the compose button, but I think this disables it properly.

In other words, if the user is able to get to the point where they can see the button (and the rest of the app in general), it means they've successfully loaded folders and recipients. If it's disabled at this point, it's because the response from MHV must have had an empty recipients list (which is a valid result), not because the request failed out. In the latter case, we would show the retry link.

U-DON:
`isEmpty` returns `true` if the argument is `null`, which is the default value of `recipients` before it gets populated, so erroring out on the request will result in it remaining `null`.

Perhaps to make it explicit, the condition for disabling the button could be clarified as `recipients && _.isEmpty(recipients)`. Similar to the condition for displaying the banner.

patrickvinograd:
In the interest of minimizing churn and possible regressions I am going to go back to leaving the Compose button always activated for now. The presence of the banner should be enough to instruct users.

PR #6010, opened 2017-07-11
Add city/state parenthetical to Place of birth on burial form

cehsu:
Connects to department-of-veterans-affairs/vets.gov-team#3831

screen shot 2017-07-11 at 2 20 31 pm

### DIFF
--- a/src/js/burials/config/form.js
+++ b/src/js/burials/config/form.js
@@ -160,7 +160,7 @@ const formConfig = {
             },
             veteranDateOfBirth: currentOrPastDateUI('Date of birth'),
             placeOfBirth: {
-              'ui:title': 'Place of birth (city, state or country)'
+              'ui:title': 'Place of birth (city and state or foreign country)'
             }
           },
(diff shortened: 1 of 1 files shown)


src/js/burials/config/form.js
```diff
@@ -160,7 +160,7 @@ const formConfig = {
             },
             veteranDateOfBirth: currentOrPastDateUI('Date of birth'),
             placeOfBirth: {
-              'ui:title': 'Place of birth'
+              'ui:title': 'Place of birth (city, state or country)'
```
jbalboni: It looks like the text in other fields is `(city and state or foreign country)`, so let's use that.

jbalboni (approved):
Looks good, had a tweak to the text.

src/js/burials/config/form.js
cehsu: Thanks @jbalboni, change made.

cvalarida (approved):

PR #6008, opened 2017-07-11
Update intro page on burials

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3838

screencapture-localhost-3001-burials-and-memorials-application-530-introduction-1499804374333

PR #6014, opened 2017-07-11
Make review notification language consistent

cehsu:
Language has been changed to match updated language in burial form (#6008).

screen shot 2017-07-11 at 4 30 29 pm


jbalboni (approved):
This is fine, but there's another issue with all the pension intro page content updates that we'll need to do tomorrow.

PR #6016, opened 2017-07-12
Preneed multiple applicants

U-DON:
Updates preneed form to accommodate multiple applicants.

The schema is updated on department-of-veterans-affairs/vets-json-schema#125 as well. Basically organizes it into multiple applications, each with its own claimant, veteran, and applicant. The form config was modified accordingly (similar nesting of applications -> items -> claimant/veteran/application).

Note that there are a bunch of fields that should be required but are not yet marked as such. This is to make it easier to quickly go through the form on staging for anyone to check whether everything's in the right place. Will add in more requirements and validations later.

Still need to add a `transformForSubmit` to fill out the data further to comply with the API.

@jbalboni: Encountered a couple issues:

1. I found that if you have `expandUnder` sections with required fields, the page won't proceed unless those required fields are filled out, even though the sections aren't expanded. Is this a known problem? How do we require fields in expanded sections only if they've been expanded?

2. Upon submitting the form, the `depends` property of a page behaved strangely. Printing out the `formData` that it was using, I would get something like:

```
Object {view:preparer: "s s", applications: Array(2), privacyAgreementAccepted: false}
```

while going through the form. Then when I try to submit, the `formData` apparently becomes this:

```
Object {data: Object, pages: Object, submission: Object, savedStatus: "not-attempted", loadedStatus: "not-attempted"…}
```

So then the fields that the original `depends` referred to are `undefined` according to this latter structure, so submission results in an error. Is this a known problem? I've been able to get around the error by doing `depends: formData => formData.field && formData.field.subfield`.


jbalboni:
@U-DON 

1. That's a known thing, we normally use `ui:required` on the fields that are conditionally hidden to set if they're required or not.
2. I haven't seen that before. That looks like we're sometimes passing the whole form state (instead of just the data) to the depends function. When exactly were you seeing that issue? You said after submitting, but I'm assuming the submit wasn't successful, since you were editing the form info after that.

U-DON:
@jbalboni: Right, the submit was unsuccessful, though, that was expected for my local testing. It happened when I clicked the "Submit Application" button.

jbalboni:
PR for issue 2 is here: https://github.com/department-of-veterans-affairs/vets-website/pull/6022

bshyong (approved):
LGTM

PR #6017, opened 2017-07-12
Updates to pension intro page

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3611

There are a few differences from the last comment in the above issue:

- I'm hiding the FDC text that has the TBD links until we get those
- I only listed Work history in the "under 65" section, since that's the only time someone would see that page
- I kept the mention of the VA file number in the first bullet with the SSN, since you can use either of those.

screencapture-localhost-3001-pension-application-527ez-introduction-1499871950070

### DIFF
--- a/src/js/pensions/components/IntroductionPage.jsx
+++ b/src/js/pensions/components/IntroductionPage.jsx
@@ -19,7 +19,7 @@ class IntroductionPage extends React.Component {
     return (
       <div className="schemaform-intro">
         <FormTitle title="Apply for pension benefits"/>
-        <p>This application is equivalent to Form 21-527EZ (Application for Pension).</p>
+        <p>Equal to Form 21-527EZ</p>
         {__BUILDTYPE__ !== 'production' && <SaveInProgressIntro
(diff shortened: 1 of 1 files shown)


peggygannon:
@jbalboni looks good. One minor edit: under Apply, can you add a period after:
 Complete this pension benefits form.

Thank you!

cehsu (approved):
Looks good to me. Wondering whether it makes more sense to note optional/conditionally required documents, rather than required documents as the section title states "...you'll need your:"

jbalboni:
@cehsu I think the wording means that the military info is required, not the documents, but maybe we can make it clearer? @peggygannon, do you have an opinion on that?

peggygannon:
@jbalboni @cehsu We discussed noting "optional" items, rather than "required" with Alex and Stacy, but the decision was to keep "required" for now, as it follows the style and treatment of the other form intro pages. We will be revisiting this style and how all intro pages are formatted in the next few months.

PR #6022, opened 2017-07-12
Pass correct data to depends when submitting rjsf form

jbalboni:
@U-DON noticed that we're passing the entire form state to `depends` functions when submitting, unlike when we're creating lists of active pages.

This has been in production for a little while, but since it results in depends always being false it means we're skipping validation, rather than enforcing it when we shouldn't, so we were likely not preventing anyone from submitting.


U-DON (approved):
LGTM. :+1: Thanks for looking into this. Also, I think you meant to refer to #6016.

PR #6023, opened 2017-07-12
Set to lower case before comparing file types in rjsf

jbalboni:
Sometimes files are upper case and that was causing problems when checking their extensions.


cehsu (approved):
Looks good to me.

PR #6027, opened 2017-07-13
Make copy/logic updates to bsl options

annekainicUSDS:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3830

Copy/logic updates based on latest. I think we still need to figure out how we're handling `awardEffectiveDate`.

### DIFF
--- a/src/js/letters/utils/helpers.js
+++ b/src/js/letters/utils/helpers.js
@@ -172,10 +172,8 @@ export function getBenefitOptionText(option, value, isVeteran) {
     }
 
     case 'serviceConnectedPercentage': {
-      if (value && value !== 'unavailable') {
-        if (isVeteran) {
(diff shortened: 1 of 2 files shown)


src/js/letters/utils/helpers.js
```diff
@@ -176,7 +176,6 @@ export function getBenefitOptionText(option, value, isVeteran) {
         if (isVeteran) {
```
raquelromano-gov: This can now go up into the above if clause:
```
if (value && value !== 'unavailable' && isVeteran) {
```

raquelromano-gov (approved):
Just 1 suggestion above

annekainicUSDS:
Thanks for the suggestion @raquelromano-gov! Made that change, will merge once the build passes.

PR #6028, opened 2017-07-13
Upgrade node 6.10.3 -> 6.11.1

jkassemi:
Heroku notified us of a node vulnerability (https://nodejs.org/en/blog/vulnerability/july-2017-security-releases/). It doesn't appear that this causes any direct issue for our users, but
can affect review applications (and potentially review instances).

Accompanies https://github.com/department-of-veterans-affairs/devops/pull/1789


robbiethegeek-usds:
Will warn the front end team they will need to do this update...

U-DON (approved):
:+1:

robbiethegeek-usds:
@jkassemi let's merge this in tomorrow and get it to production so we can make sure there are no issues early this week will let the front end know they need to update their locals

PR #6029, opened 2017-07-13
Add message to upload page on pension form

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3798

This only adds the message on the upload page. Other changes in that issue need some feedback from stakeholders.

screen shot 2017-07-13 at 12 49 52 pm


cehsu (approved):
Looks good to me. Minor content question, would it be helpful to clarify that the intent of the mail in option is convenience (and that the mailing address will be provided on the confirmation page specifically)? For example:
~~If you have many documents to submit (or any documents that exceed the 20MB size limit) and would prefer to mail them in rather than upload them here, a mailing address will be provided along with your confirmation details.~~
**UPDATE**: If you have many documents to submit (or any documents that exceed the 20MB size limit) and would prefer to mail them in rather than upload them here, a mailing address will be provided on your confirmation page.  (<-- might be nice to indicate that this will be provided here rather than say via email or otherwise misinterpreted, perhaps too much information though)

jbalboni:
@cehsu I think "after you finish" is clearer than "confirmation details" since a user doesn't necessarily know that they will end up on a page with those details.

@goldenmeanie or @peggygannon, do you have any input on Claire's question?

goldenmeanie:
I also like "after you finish" or "after you finish the application"

peggygannon:
@goldenmeanie @jbalboni Yes, I agree. ... "after you finish the application" is the most clear..

cvalarida (approved):

PR #6033, opened 2017-07-13
Letters fix checkbox styling when focused

annekainicUSDS:
I noticed on letters when you clicked on a checkbox it would add a blue box shadow around the outside of the checkbox. 

screen shot 2017-07-13 at 5 04 43 pm

This was really bothering me, so just submitting a small css change to stop that box shadow from being added on `:focus` state of the checkbox.

Also fixed the lint warning we started getting a week ago with a css change. I just took the color value and put it in a variable so there are no more lint warnings! @bshyong I think this affects appeals status so wanted to make sure you take a look at this before merging.


raquelromano-gov (approved):

U-DON (approved):
LGTM.

PR #6034, opened 2017-07-13
Display wartimeWarning when veteran didn't serve during wartime

cehsu:
Connects to department-of-veterans-affairs/vets.gov-team#3793

screen shot 2017-07-13 at 4 43 58 pm

### DIFF
--- a/src/js/pensions/config/form.js
+++ b/src/js/pensions/config/form.js
@@ -16,7 +16,8 @@ import {
   directDepositWarning,
   isMarried,
   applicantDescription,
-  otherExpensesWarning
+  otherExpensesWarning,
+  wartimeWarning
 } from '../helpers';
 import IntroductionPage from '../components/IntroductionPage';
 import DisabilityField from '../components/DisabilityField';
@@ -111,6 +112,25 @@ function isCurrentMarriage(form, index) {
 
 function usingDirectDeposit(formData) {
   return formData['view:noDirectDeposit'] !== true;
+}
+
+function servedDuringWartime(period) {
+  const warDates = [
+    ['1916-05-09', '1917-04-05'], // Mexican Border Period (May 9, 1916 - April 5, 1917)
+    ['1917-04-06', '1918-11-11'], // World War I (April 6, 1917 - November 11, 1918)
+    ['1941-12-07', '1946-12-31'], // World War II (December 7, 1941 - December 31, 1946)
+    ['1950-06-27', '1955-01-31'], // Korean Conflict (June 27, 1950 - January 31, 1955)
+    ['1964-08-05', '1975-05-07'], // Vietnam Era (August 5, 1964 - May 7, 1975)
+    ['1990-08-02']// Gulf War (August 2, 1990)
+  ];
+  return warDates.some((warTime) => {
+    const warStart = warTime[0];
+    const warEnd = warTime[1];
+    const periodStart = period.from;
+    const periodEnd = period.to;
+    const overlap = (warStart <= periodStart && periodStart <= warEnd) || (warStart <= periodEnd && periodEnd <= warEnd);
+    return (warEnd ? overlap : warStart < periodEnd);
+  });
 }
 
 const marriageProperties = marriages.items.properties;
@@ -245,7 +265,28 @@ const formConfig = {
                   'Date entered service must be before date left service'
                 )
               }
-            }
+            },
+            'view:wartimeWarning': (() => {
+              const rangeIncludesWartime = createSelector(
+                form => form.servicePeriods,
+                (periods) => {
+                  return (periods || []).some(period => {
+                    const isFullDate = /^\d{4}-\d{2}-\d{2}$/;
+                    return !period.activeServiceDateRange ||
+                      !isFullDate.test(period.activeServiceDateRange.to) ||
+                      !isFullDate.test(period.activeServiceDateRange.from) ||
+                      servedDuringWartime(period.activeServiceDateRange);
+                  });
+                }
+              );
+
+              return {
+                'ui:description': wartimeWarning,
+                'ui:options': {
+                  hideIf: rangeIncludesWartime
+                }
+              };
+            })()
           },
           schema: {
             type: 'object',
@@ -265,6 +306,10 @@ const formConfig = {
                     })
                   }
                 }
+              },
+              'view:wartimeWarning': {
+                type: 'object',
+                properties: {}
               }
             }
           }
--- a/src/js/pensions/config/form.js
+++ b/src/js/pensions/config/form.js
@@ -3,6 +3,7 @@ import moment from 'moment';
 import { createSelector } from 'reselect';
 
 import fullSchemaPensions from 'vets-json-schema/dist/21P-527EZ-schema.json';
+import { isFullDate } from '../../common/utils/validations';
 
 import * as address from '../../common/schemaform/definitions/address';
 import bankAccountUI from '../../common/schemaform/definitions/bankAccount';
@@ -114,7 +115,6 @@ function isCurrentMarriage(form, index) {
 function usingDirectDeposit(formData) {
   return formData['view:noDirectDeposit'] !== true;
 }
-
 
 const marriageProperties = marriages.items.properties;
 
@@ -254,10 +254,9 @@ const formConfig = {
                 form => form.servicePeriods,
                 (periods) => {
                   const completePeriods = (periods || []).filter(period => {
-                    const isFullDate = /^\d{4}-\d{2}-\d{2}$/;
                     return period.activeServiceDateRange &&
-                      isFullDate.test(period.activeServiceDateRange.to) &&
-                      isFullDate.test(period.activeServiceDateRange.from);
+                      isFullDate(period.activeServiceDateRange.to) &&
+                      isFullDate(period.activeServiceDateRange.from);
                   });
                   if (!completePeriods.length) {
                     return true;
@@ -326,10 +325,8 @@ const formConfig = {
                 form => form.servicePeriods,
                 (periods) => {
                   return (periods || []).every(period => {
-                    const isFullDate = /^\d{4}-\d{2}-\d{2}$/;
-
                     return !period.activeServiceDateRange ||
-                      !isFullDate.test(period.activeServiceDateRange.to) ||
+                      !isFullDate(period.activeServiceDateRange.to) ||
                       !moment('2001-09-11').isBefore(period.activeServiceDateRange.to);
                   });
                 }
--- a/src/js/pensions/config/form.js
+++ b/src/js/pensions/config/form.js
@@ -17,7 +17,8 @@ import {
   isMarried,
   applicantDescription,
   otherExpensesWarning,
-  wartimeWarning
+  wartimeWarning,
+  servedDuringWartime
 } from '../helpers';
 import IntroductionPage from '../components/IntroductionPage';
 import DisabilityField from '../components/DisabilityField';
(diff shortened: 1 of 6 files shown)


src/js/pensions/config/form.js
```diff
@@ -113,6 +114,24 @@ function usingDirectDeposit(formData) {
   return formData['view:noDirectDeposit'] !== true;
 }
 
+function servedDuringWartime(period) {
+  const warDates = [
+    [-1692982800000, -1664384400000], // Mexican Border Period (May 9, 1916 - April 5, 1917)
```
cehsu: Calculations for reference:
screen shot 2017-07-13 at 4 45 34 pm

src/js/pensions/config/form.js
```diff
@@ -113,6 +114,25 @@ function usingDirectDeposit(formData) {
...
+    ['1950-06-27', '1955-01-31'], // Korean Conflict (June 27, 1950 - January 31, 1955)
+    ['1964-08-05', '1975-05-07'], // Vietnam Era (August 5, 1964 - May 7, 1975)
+    ['1990-08-02']// Gulf War (August 2, 1990)
+  ];
+  return warDates.some((warTime) => {
+    const warStart = warTime[0];
+    const warEnd = warTime[1];
+    const periodStart = period.from;
+    const periodEnd = period.to;
+    const overlap = (warStart <= periodStart && periodStart <= warEnd) || (warStart <= periodEnd && periodEnd <= warEnd);
```
jbalboni: I think it would be better to use moment for this.

src/js/pensions/config/form.js
```diff
@@ -113,6 +114,25 @@ function usingDirectDeposit(formData) {
...
+  const warDates = [
+    ['1916-05-09', '1917-04-05'], // Mexican Border Period (May 9, 1916 - April 5, 1917)
+    ['1917-04-06', '1918-11-11'], // World War I (April 6, 1917 - November 11, 1918)
+    ['1941-12-07', '1946-12-31'], // World War II (December 7, 1941 - December 31, 1946)
+    ['1950-06-27', '1955-01-31'], // Korean Conflict (June 27, 1950 - January 31, 1955)
+    ['1964-08-05', '1975-05-07'], // Vietnam Era (August 5, 1964 - May 7, 1975)
+    ['1990-08-02']// Gulf War (August 2, 1990)
+  ];
+  return warDates.some((warTime) => {
+    const warStart = warTime[0];
```
jbalboni: You can destructure both of these:

`const [warStart, warEnd] = warTime;`
`const {from, to} = period;`

src/js/pensions/config/form.js
```diff
@@ -245,7 +265,28 @@ const formConfig = {
                   'Date entered service must be before date left service'
                 )
               }
-            }
+            },
+            'view:wartimeWarning': (() => {
+              const rangeIncludesWartime = createSelector(
+                form => form.servicePeriods,
+                (periods) => {
+                  return (periods || []).some(period => {
```
jbalboni: I feel like this might be clearer if you separated out the full date checks from the call to check the service period. The way I understand the logic is that we want to filter out any periods with incomplete dates (since those are likely in progress). Then, if there's at least one period with full dates, we want to show the warning if none of those periods are in wartime.

```js
const completePeriods = (periods || []).filter(period => {
  const isFullDate = /^\d{4}-\d{2}-\d{2}$/;
  return period.activeServiceDateRange &&
    isFullDate.test(period.activeServiceDateRange.to) &&
    isFullDate.test(period.activeServiceDateRange.from);
});

if (!periods.length) {
  return true;
}

return completePeriods.some(period => servedDuringWartime(period.activeServiceDateRange);
```

src/js/pensions/config/form.js
```diff
@@ -113,6 +114,25 @@ function usingDirectDeposit(formData) {
   return formData['view:noDirectDeposit'] !== true;
 }
 
+function servedDuringWartime(period) {
```
jbalboni: I think we should pull this into the helper file and write some unit tests against it, rather than only doing the DefinitionTester test.

src/js/pensions/config/form.js
```diff
@@ -245,7 +265,28 @@ const formConfig = {
                   'Date entered service must be before date left service'
                 )
               }
-            }
+            },
+            'view:wartimeWarning': (() => {
+              const rangeIncludesWartime = createSelector(
```
jbalboni: This selector is probably better named something like `hideWartimeWarning`, since it includes a check to hide it if we have incomplete dates or no periods.

src/js/pensions/config/form.js
```diff
@@ -113,6 +114,25 @@ function usingDirectDeposit(formData) {
   return formData['view:noDirectDeposit'] !== true;
 }
 
+function servedDuringWartime(period) {
+  const warDates = [
```
jbalboni: I wonder if we should move this out of the function, so it's not created and destroyed on every call.

src/js/pensions/config/form.js
cehsu: Thanks @jbalboni, changes made.

src/js/pensions/config/form.js
cehsu: Thanks @jbalboni, changes made.

src/js/pensions/config/form.js
cehsu: Thanks @jbalboni, changes made.

src/js/pensions/config/form.js
cehsu: Thanks @jbalboni, changes made.

src/js/pensions/config/form.js
cehsu: Thanks @jbalboni, changes made.

src/js/pensions/config/form.js
cehsu: Thanks @jbalboni, changes made.

test/pensions/servedDuringWartime.unit.spec.js
```diff
@@ -0,0 +1,32 @@
+import { expect } from 'chai';
+
+import { servedDuringWartime } from '../../src/js/pensions/helpers';
+
+describe('Wartime service validation', () => {
```
cehsu: Thanks for your recommendation to create these tests @jbalboni, please let me know if additional changes would make sense (naming, additional/different test cases, etc.).

cehsu (commented):
Changes in response to @jbalboni's helpful PR comments have been made here, please let me know if additional changes would make sense as well.

src/js/pensions/helpers.jsx
```diff
@@ -97,3 +98,27 @@ export const otherExpensesWarning = (
...
+  ['1964-08-05', '1975-05-07'], // Vietnam Era (August 5, 1964 - May 7, 1975)
+  ['1990-08-02']// Gulf War (August 2, 1990)
+];
+
+export function servedDuringWartime(period) {
+  return warDates.some((warTime) => {
+    const [warStart, warEnd] = warTime;
+    const { from: periodStart, to: periodEnd } = period;
+    const overlap = moment(periodStart).isBetween(warStart, warEnd) || moment(periodEnd).isBetween(warStart, warEnd);
+    return (warEnd ? overlap : warStart < periodEnd);
```
cvalarida: What if I served from 1950-01-01 to 1956-01-01? It's the whole Korean "conflict," but this function would return false because the start and end date aren't between the start and end of the conflict (and there is a `warEnd` date).

I think you need another `|| (moment(periodStart).isBefore(warStart) && moment(periodEnd).isAfter(warEnd))`.

test/pensions/config/servicePeriods.unit.spec.jsx
```diff
@@ -38,6 +38,52 @@ describe('Pensions service periods', () => {
...
+          schema={schema}
+          definitions={formConfig.defaultDefinitions}
+          onSubmit={onSubmit}
+          uiSchema={uiSchema}/>
+    );
+
+    const formDOM = getFormDOM(form);
+
+    formDOM.fillData('#root_servicePeriods_0_serviceBranch', 'Army');
+    formDOM.fillData('#root_servicePeriods_0_activeServiceDateRange_toMonth', '1');
```
cvalarida: You can use `formDOM.fillDate('root_servicePeriods_0_activeServiceDateRange_to', '1-1-1983');` for this. (Though perhaps we should change the format of that to 'YYYY-MM-DD'...would require some refactoring, though.)

test/pensions/servedDuringWartime.unit.spec.js
```diff
@@ -0,0 +1,32 @@
...
+    });
+    it('should not allow for service date outside of wartime period', () => {
+      const allow = servedDuringWartime({
+        from: '1980-11-01',
+        to: '1981-12-02'
+      });
+
+      expect(allow).to.be.false;
+    });
+  });
```
cvalarida: Add a test to check if the service period completely envelopes a wartime period.

cvalarida (commented):
I just think it won't work if a service period starts before the wartime period and ends after.

src/js/pensions/helpers.jsx
```diff
@@ -97,3 +98,27 @@ export const otherExpensesWarning = (
   </div>
 );
 
+export const wartimeWarning = (
+  <div className="usa-alert usa-alert-warning no-background-image">
+    <span><strong>Note:</strong> You have indicated that you did not serve during an <a href="http://www.benefits.va.gov/pension/wartimeperiod.asp"> eligible wartime period</a>. Find out if you still qualify. <a href="/pension/eligibility">Check your eligibility.</a></span>
```
cehsu: Good point @cvalarida!

src/js/pensions/helpers.jsx
cehsu: Thanks again @cvalarida.

test/pensions/config/servicePeriods.unit.spec.jsx
cehsu: Thanks, nice tip!

src/js/pensions/config/form.js
```diff
@@ -245,7 +248,33 @@ const formConfig = {
...
                 )
               }
-            }
+            },
+            'view:wartimeWarning': (() => {
+              const hideWartimeWarning = createSelector(
+                form => form.servicePeriods,
+                (periods) => {
+                  const completePeriods = (periods || []).filter(period => {
+                    const isFullDate = /^\d{4}-\d{2}-\d{2}$/;
```
jbalboni: I believe so, yes.

src/js/pensions/helpers.jsx
```diff
@@ -97,3 +98,31 @@ export const otherExpensesWarning = (
...
+  ['1964-08-05', '1975-05-07'], // Vietnam Era (August 5, 1964 - May 7, 1975)
+  ['1990-08-02']// Gulf War (August 2, 1990)
+];
+
+export function servedDuringWartime(period) {
+  return warDates.some((warTime) => {
+    const [warStart, warEnd] = warTime;
+    const { from: periodStart, to: periodEnd } = period;
+    const overlap =
+      moment(periodStart).isBetween(warStart, warEnd) ||
```
jbalboni: I think these all need to be inclusive, which is not the default in moment. Would look like `.isBetween(warStart, warEnd, null, '[]')`. https://momentjs.com/docs/#/query/is-between/

cvalarida: Oh! That was the other thing I was going to mention. Good call, @jbalboni.

cehsu: Thanks @jbalboni!

test/pensions/servedDuringWartime.unit.spec.js
cehsu: Thanks @cvalarida, change made.

jbalboni (approved):

PR #6037, opened 2017-07-14
Removed overly fancy terms events naming scheme

robertfairhead:
Follow-up on https://github.com/department-of-veterans-affairs/vets-website/pull/5941

Looking at it in practice @patrickvinograd was right and folks are coming in at sub-URLs so the fancy naming scheme isn't actually all that useful.

screen shot 2017-07-14 at 3 19 07 pm

Worth a shot, I guess.

Reverting to a simpler naming scheme before too much data piles up. We can use the pages view of events (as shown above) to determine what page caused the terms to be shown so there's real data loss from dropping it from the event name.

### DIFF
--- a/src/js/common/components/RequiredTermsAcceptanceView.jsx
+++ b/src/js/common/components/RequiredTermsAcceptanceView.jsx
@@ -16,8 +16,7 @@ export class RequiredTermsAcceptanceView extends React.Component {
       this.props.fetchLatestTerms(this.props.termsName);
       window.scrollTo(0, 0);
     }
-    const app = window.location.pathname.split('/').pop();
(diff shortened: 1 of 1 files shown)


ayaleloehr:
@robertfairhead is this just waiting on a review from someone on Kudos (likely @U-DON @saneshark or @patrickvinograd ) or is there more to be done?

robertfairhead:
Yeah, just needs the review :+1:

This is a low-speed one as the change isn't particularly impactful. The event prefix will remain the same so it's easy to filter before/after. This is mostly a long-term cleanliness thing.

Was working with Patrick on some Facility Locator items that natalie wanted so this was back-burnered.

U-DON (approved):
LGTM.

PR #6038, opened 2017-07-14
Gibs month day fields

annekainicUSDS:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3733

screen shot 2017-07-14 at 4 29 01 pm

### DIFF
--- a/src/js/post-911-gib-status/utils/helpers.js
+++ b/src/js/post-911-gib-status/utils/helpers.js
@@ -22,7 +22,7 @@ export function formatVAFileNumber(n) {
 export function formatMonthDayFields(field) {
   let displayValue;
   if (field) {
-    displayValue = `${field.months || 0} months, ${field.days || 0} days`;
+    displayValue = `${field.months} months, ${field.days} days`;
   } else {
     displayValue = 'unavailable';
   }
--- a/src/js/post-911-gib-status/utils/helpers.js
+++ b/src/js/post-911-gib-status/utils/helpers.js
@@ -20,7 +20,13 @@ export function formatVAFileNumber(n) {
 }
 
 export function formatMonthDayFields(field) {
-  return `${field.months || 0} months, ${field.days || 0} days`;
+  let displayValue;
+  if (field) {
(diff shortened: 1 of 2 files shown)


src/js/post-911-gib-status/utils/helpers.js
```diff
@@ -19,6 +19,16 @@ export function formatVAFileNumber(n) {
   return number.replace(number.substring(0, lengthOfXString), `${'x'.repeat(lengthOfXString)}-`);
 }
 
+export function formatMonthDayFields(field) {
+  let displayValue;
+  if (field) {
+    displayValue = `${field.months || 0} months, ${field.days || 0} days`;
```
raquelromano-gov: If the entitlement object is given, both months and days are required (in both the EVSS and vets-api swagger docs), so we can display them without testing and defaulting to 0.

src/js/post-911-gib-status/utils/helpers.js
```diff
@@ -19,6 +19,16 @@ export function formatVAFileNumber(n) {
...
+  } else {
+    displayValue = 'unavailable';
```
raquelromano-gov: Let's check with @phiden on what this should say if the entitlement isn't supplied. None of the entitlement objects (total, used, remaining) are required, so it very well be that one is not defined, though it probably shouldn't happen in the wild (this is the problem I was hoping to address with this old ticket: https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3336)

I think you're right that for any of these values that we can't rely in b/c they are not required fields in the API docs, we should supply a human-readable placeholder. Maybe we should also log a Sentry error so that we can monitor whether/how often this happens if at all.

raquelromano-gov (approved):
LGTM except for the outstanding question about what to display when a field that's not required in the API isn't defined, which we can/should address more broadly.

annekainicUSDS:
Need to merge this as I think it's breaking the build on staging. Will revisit what message should be shown in a separate PR.

PR #6039, opened 2017-07-14
Adding an event for when an option is checked or unchecked in benefits letters

robertfairhead:
Re: https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3446

I think this would do the trick but maybe isn't the cleanest with so many string interpolations. Feel free to tweak it. Once this is merged I'll do the config on the tag manager side so the reporting will flow through.

### DIFF
--- a/src/js/letters/containers/VeteranBenefitSummaryLetter.jsx
+++ b/src/js/letters/containers/VeteranBenefitSummaryLetter.jsx
@@ -17,7 +17,11 @@ export class VeteranBenefitSummaryLetter extends React.Component {
   }
 
   handleChange(domEvent) {
-    window.dataLayer.push({ event: `letter-benefit-${domEvent.target.id}-${domEvent.target.checked ? 'checked' : 'unchecked'}` });
+    window.dataLayer.push({
+      event: 'letter-benefit-option-clicked',
+      'letter-benefit-option': domEvent.target.id,
+      'letter-benefit-option-status': domEvent.target.checked ? 'checked' : 'unchecked'
+    });
     this.props.updateBenefitSummaryRequestOption(
       benefitOptionsMap[domEvent.target.id],
       domEvent.target.checked);
--- a/src/js/letters/containers/VeteranBenefitSummaryLetter.jsx
+++ b/src/js/letters/containers/VeteranBenefitSummaryLetter.jsx
@@ -17,6 +17,7 @@ export class VeteranBenefitSummaryLetter extends React.Component {
   }
 
   handleChange(domEvent) {
+    window.dataLayer.push({ event: `letter-benefit-${domEvent.target.id}-${domEvent.target.checked ? 'checked' : 'unchecked'}` });
     this.props.updateBenefitSummaryRequestOption(
       benefitOptionsMap[domEvent.target.id],
       domEvent.target.checked);


robertfairhead:
Upon further reflection, i decided to split out the details into variables. That'll make the reporting less brittle in GA since I can just parameterize there and not worry about having to manually adding each new option.

src/js/letters/containers/VeteranBenefitSummaryLetter.jsx
```diff
@@ -17,6 +17,11 @@ export class VeteranBenefitSummaryLetter extends React.Component {
   }
 
   handleChange(domEvent) {
+    window.dataLayer.push({
+      event: 'letter-benefit-option-clicked',
+      'letter-benefit-option': domEvent.target.id,
```
raquelromano-gov: Yeah, I agree this looks cleaner than packing all the identifiers into the event name.

I was just browsing the GTM docs and don't understand their comment in this section: https://developers.google.com/tag-manager/devguide#multipush which says that pushing a variable with the same name and a new value will overwrite the old value (e.g., for `letter-benefit-option`, which will take on various values as the user clicks on different checkboxes). Their comment doesn't make sense since `dataLayer` is an array, so pushing should just add a new one. Maybe just poorly written documentation, and you will indeed have access to all of the pushed variables on the reporting side?

robertfairhead: The pushes just queue objects locally that eventually get sent to the Google servers for processing where the actual "state" resides.

So if we had a tag that reported out something like `Category: Interaction`, `Action: {{ letter-benefit-option }} - {{ letter-benefit-option-clicked }}`, `Label: {{ Event }}` into GA, the value of those variables `letter-benefit-option` and `letter-benefit-option-clicked` is stored on the GA side.

If events like the below were pushed locally:

```
{
  event: 'letter-benefit-option-clicked',
  'letter-benefit-option': 'foo',
  'letter-benefit-option-status': 'checked'
}

{
  event: 'letter-benefit-option-clicked',
  'letter-benefit-option': 'foo',
}

{
  event: 'letter-benefit-option-clicked',
  'letter-benefit-option': 'foo',
  'letter-benefit-option-status': 'unchecked'
}
```

The event actions reported in GA should be:

```
foo - checked
foo - checked
foo - unchecked
```

If those event pushes arrived out of order at the GA servers with the second one arriving last (because the internet), you'd get:

```
foo - checked
foo - unchecked
foo - unchecked
```

Which is what that comment is meant to warn you about. Once you overwrite the value of variable on the GA server, it'll persist just that new value. So, if your actual reporting of an event depends on a variable, you should always push them together to be sure the variable has the correct value at processing time.

Hope that clarifies!

robertfairhead: While I'm thinking about it, i should also note that everything in GA resides in a single, global namespace. So, if you had a variable `checked` in two different places they'd each be overwriting each other constantly. Or more perniciously, if you set a `product` variable *only* on app startup it'd have like `gibct` and then use it in your event reporting and then th user opens `letters` in a different tab, overwriting `product`, all the future GIBCT events would get reported through tags labeled as `letters`. 

Which is why we try to always apply a meaningful prefix to our GA variables.

raquelromano-gov: Thank very  much for the explanation! Definitely would not have inferred all of this from that developer guide post.

raquelromano-gov (approved):

PR #6040, opened 2017-07-17
Handle various HTTP status codes

raquelromano-gov:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3780
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3839 (Scenario G)


src/js/post-911-gib-status/utils/helpers.js
```diff
@@ -1,62 +0,0 @@
-import merge from 'lodash/fp/merge';
```
raquelromano-gov: Sorry, this diff seems to be ignoring my `git mv`

src/js/post-911-gib-status/actions/post-911-gib-status.js
```diff
@@ -1,18 +1,57 @@
...
+} from '../utils/constants';
 
 export function getEnrollmentData() {
   return (dispatch) => {
-    apiRequest('/v0/post911_gi_bill_status',
+    apiRequest(
+      '/post911_gi_bill_status',
       null,
       (response) => {
+        // TODO: clarify with backend devs whether vets-api will continue to
```
raquelromano-gov: Asked this in #unicorns

src/js/post-911-gib-status/actions/post-911-gib-status.js
```diff
@@ -1,18 +1,57 @@
...
+        }
+        if (response.status === 403) {
+          // Backend authentication problem
+          return dispatch({ type: BACKEND_AUTHENTICATION_ERROR });
+        }
+        if (response.status === 404) {
+          // EVSS partner service has no record of this user
+          return dispatch({ type: NO_CHAPTER33_RECORD_AVAILABLE });
+        }
+        // TODO: find out if we have any best practices for logging Sentry errors
```
raquelromano-gov: Will ask this in #vetsgov-fe

src/js/post-911-gib-status/containers/Main.jsx
```diff
@@ -12,26 +17,27 @@ class Main extends React.Component {
...
+        appContent = this.props.children;
+        break;
+      case 'awaitingResponse':
+        appContent = <LoadingIndicator message="Loading your Post-9/11 GI Bill benefit information..."/>;
+        break;
+      case 'backendServiceError':
+        appContent = systemDownMessage();
+        break;
+      case 'backendAuthenticationError':
+        appContent = chapter33InfoUnavailableWarning();
```
raquelromano-gov: Asked Sophia about this in Slack

src/js/post-911-gib-status/utils/constants.js
```diff
@@ -0,0 +1,14 @@
+// Action types
+export const BACKEND_AUTHENTICATION_ERROR = 'BACKEND_AUTHENTICATION_ERROR';
```
raquelromano-gov: I kept misspelling these, so defining constants for all of them.

src/js/post-911-gib-status/utils/helpers.jsx
```diff
@@ -0,0 +1,83 @@
...
+        <div className="react-container">
+          <h3>Sorry, our system is temporarily down while we fix a few things. Please try again later.</h3>
+          <a href="/"><button>Go Back to Vets.gov</button></a>
+        </div>
+      </div>
+    </div>
+  );
+}
+
+export function chapter33InfoUnavailableWarning() {
```
raquelromano-gov: Asked Sophia in Slack if we are still using this pink warning since it's no longer in the mocks.

raquelromano-gov:
@annekainicUSDS This is ready for comments, though I'll work on getting answers to some of the above outstanding questions tomorrow, as well as updating unit tests.

src/js/post-911-gib-status/utils/helpers.jsx
raquelromano-gov: Sophia says to replace this with a generic no-records warning like this one: https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/js/disability-benefits/components/MviRecordsUnavailable.jsx

src/js/post-911-gib-status/actions/post-911-gib-status.js
```diff
@@ -1,18 +1,63 @@
...
+        }
+        throw (response);
       })
-    );
+      .catch((error) => {
+        // TODO: find out why Raven does not treat this object as an error
+        // Raven.captureException(error);
+        let errorMessage;
+        if (error.errors && error.errors.length > 0) {
+          errorMessage = `${error.errors[0].title}: ${error.errors[0].detail}`;
```
annekainicUSDS: Why do we only care about the first error message in the list?

src/js/post-911-gib-status/containers/Main.jsx
```diff
@@ -12,26 +17,27 @@ class Main extends React.Component {
...
+        appContent = systemDownMessage();
+        break;
+      case 'backendAuthenticationError':
+        appContent = unableToFindRecordWarning();
+        break;
+      case 'noChapter33Record':
+        appContent = noChapter33BenefitsWarning();
+        break;
+      case 'unavailable':
+      default:
```
annekainicUSDS: This is great! So much more organized and clear.

src/js/post-911-gib-status/actions/post-911-gib-status.js
```diff
@@ -1,18 +1,63 @@
...
+        }
+        if (response.status === 403) {
+          // Backend authentication problem
+          return dispatch({ type: BACKEND_AUTHENTICATION_ERROR });
+        }
+        if (response.status === 404) {
+          // EVSS partner service has no record of this user
+          return dispatch({ type: NO_CHAPTER33_RECORD_AVAILABLE });
+        }
+        throw (response);
```
annekainicUSDS: Reading up on the docs for `throw`, a little confused what throwing an exception is doing here. Is this to prevent the `.catch` from executing?

src/js/post-911-gib-status/reducers/index.js
```diff
@@ -1,19 +1,33 @@
...
         availability: 'available'
       };
-    case 'GET_ENROLLMENT_DATA_FAILURE':
+    case BACKEND_SERVICE_ERROR:
+      return set('availability', 'backendServiceError', state);
+    case BACKEND_AUTHENTICATION_ERROR:
+      return set('availability', 'backendAuthenticationError', state);
+    case NO_CHAPTER33_RECORD_AVAILABLE:
+      return set('availability', 'noChapter33Record', state);
+    case GET_ENROLLMENT_DATA_FAILURE:
```
annekainicUSDS: What's the relationship between the generic `GET_ENROLLMENT_DATA_FAILURE` and the more specific types of errors? I see `GET_ENROLLMENT_DATA_FAILURE` gets triggered in the `.catch`, but not quite clear on how that works and when that gets executed.

annekainicUSDS (commented):
Looks great so far! Mostly just a bunch of questions for myself to understand what the code is doing better.

src/js/post-911-gib-status/actions/post-911-gib-status.js
raquelromano-gov: It's actually to trigger the `catch` below. So if none of the response status codes are things we recognize, then something weird and unexpected is happening, so the below `catch` will log a sentry error. 

This one I did test by hacking the vets-api controller to raise an unexpected error.

src/js/post-911-gib-status/actions/post-911-gib-status.js
raquelromano-gov: You're right--we should probably show them all. I was hoping to avoid all of this unpacking by just calling `Raven.captureException(error)` but it appears that Raven.js wasn't able to understand this error, so it didn't do what it should when capturing an exception (dump a stack trace etc.), so I decided to hand-craft the error message.

@jbalboni any chance you can look at the above and LMK if there's a better way to log a sentry error here?

src/js/post-911-gib-status/reducers/index.js
raquelromano-gov: I think I answered this in the above question, but this is the catch-all for any failure not recognized/expected. I have no idea when/why/if this would happen.

src/js/post-911-gib-status/actions/post-911-gib-status.js
jbalboni: I'm not sure where this error object is coming from. It doesn't look like a normal `Error` object. Is it the `Response` object that's being thrown above?

Generally, we should throw errors, not a `Response` object. You can reject a promise with the `Response` object if you need to `return Promise.reject(response)`. I'm not sure if that's the issue, but it wouldn't hurt.

src/js/post-911-gib-status/actions/post-911-gib-status.js
annekainicUSDS: Ah gotcha, that makes sense!

src/js/post-911-gib-status/reducers/index.js
annekainicUSDS: Yep, you did, thanks!

annekainicUSDS (approved):
LGTM!

src/js/post-911-gib-status/utils/helpers.jsx
```diff
@@ -0,0 +1,79 @@
...
+        </div>
+      </div>
+    </div>
+  );
+}
+
+export function unableToFindRecordWarning() {
+  return (
+    <div>
+      <div className="usa-alert usa-alert-warning claims-no-icon claims-unavailable">
```
raquelromano-gov: FYI, these are left-over classes from copy-pasting from another error message, so I'm removing them and tweaking the layout.

annekainicUSDS:
After testing this locally, a couple of thoughts on how the error messages should be rendered.

In general, we should probably make each error message look like this page: https://marvelapp.com/8d6igd9/screen/30345545. In order to do that, I think there are a couple of changes we could make:

1. We should include the title of the page with each error message. Right now, it's not being included. This is because the title is included in `StatusPage`, which is rendered within `Main` (where the error messages are being calculated). One solution is to move the title (`<FormTitle title="Post-9/11 GI Bill Benefit Information"/>`) from `StatusPage` to `Main`.

2. Let's make the html for each message follow the USWDS alert styles. So something like this:
```
<div className="feature">
    <h2>Sorry, our system is temporarily down while we fix a few things. Please try again later.</h2>
    <div><a href="/"><button>Go Back to Vets.gov</button></a></div>
</div>
```

Also, I was only successfully able to replicate responses for a 200, 503, and 500. For some reason, when I tried to replicate responses for a 403, 404, and 504, I just got a generic 500 error response from vets-api on the front-end, so I wasn't able to see the messages that those error codes render. Seems like this issue is coming from vets-api, but haven't dug into this too much yet.

(btw the way I'm mocking these responses is by commenting out everything in https://github.com/department-of-veterans-affairs/vets-api/pull/1207/files#diff-a75f655aa85a597d3de913c368a0c1f8 exception for the response I want.

src/js/post-911-gib-status/actions/post-911-gib-status.js
raquelromano-gov: Thanks! I changed this to reject the promise and return a new Error object, so `Raven.captureException()` should do the right thing with it

raquelromano-gov:
Just for the record, we decided to leave the vets.gov-wide types of messages styled like they are elsewhere, i.e., usually `<h4>` elements with no form title, just breadcrumbs, but we'll run all this by product and design. Screenshots, based on https://github.com/department-of-veterans-affairs/vets.gov-team/blob/master/Products/EVSS%20Integration/Letters%20and%20GIBS%20error%20messages%20mapping.md:

For HTTP status codes 504 and 503:
image

For HTTP status code 403:
image

For HTTP status code 404:
image

src/js/post-911-gib-status/containers/Post911GIBStatusApp.jsx
```diff
@@ -37,7 +38,7 @@ class Post911GIBStatusApp extends React.Component {
           verifyUrl={this.props.verifyUrl}>
         <AppContent>
           <div className="row">
-            <div className="usa-width-two-thirds medium-8 columns">
+            <div className="small-12 columns">
```
raquelromano-gov: I had to swap where some of the classes are defined so that the system-down and no-records errors span the full width.

src/js/post-911-gib-status/actions/post-911-gib-status.js
```diff
@@ -1,18 +1,58 @@
-import { apiRequest } from '../utils/helpers';
+import Raven from 'raven-js';
```
raquelromano-gov: The test for these actions is a WIP.

raquelromano-gov:
OK, after a frightening merge, and then realizing I needed to update my mock GIBS service yaml to properly handle some of the conditional stuff, this is as ready as it's going to get for tomorrow's tests. I will try to test on staging before that to make sure things are working as intended.

Still outstanding are
- actions unit tests
- one other Main.jsx test that's commented out (seemed straightforward, but when I moved classes around, I couldn't get any selectors to work anymore)

And of course running these screenshots by Sophia and Rachael for input/polish.

For the record, I tested some of this stuff by editing the vets-api controller, but I also couldn't reproduce the various HTTP status codes, so I grabbed the screenshots by just hacking the fetch error function to dispatch different actions.

PR #6044, opened 2017-07-17
Make testing in Sauce totally work

aub:
This does a few things:

* Sets up nightwatch configuration for Sauce so that it can be executed correctly and with the proper environments for different browsers we would like to test.
* Adds a nightwatch-globals file that defines a callback to be run after each test for letting Sauce know if it passed or fail (annoyingly, this is not done automatically)
* Adds some packages and test commands in package.json for running these things
* Updates timeouts to be longer when Sauce is in effect
* Removes a check for the status in axeCheck because not all browsers return their results in the expected format

Note that annoyingly, all of the tests do not yet pass on all browsers when running on sauce. In particular, I hope to fix a problem where on Firefox the browser won't select the correct item from a select box for the education 1995 form, which causes it to fail. I have not yet run this against every browser we have configured and cataloged the errors, but this should be ok to set up the framework for it.


config/nightwatch-globals-sauce.js
```diff
@@ -0,0 +1,30 @@
+module.exports = {
```
raquelromano-gov: There is also this directory that has nightwatch stuff in it: `test/util/nightwatch-commands` although a bunch of those are helper functions. Wondering if anyone has strong opinions about whether these should live in a single place.

test/user-profile/e2e/user-profile.e2e.spec.js
```diff
@@ -7,9 +7,11 @@ module.exports = E2eHelpers.createE2eTest(
     let token = LoginHelpers.getUserToken();
 
     LoginHelpers.logIn(token, client, '/profile', 1)
-      .waitForElementVisible('.loading-indicator-container', Timeouts.normal)
-      .waitForElementNotPresent('.loading-indicator-container', Timeouts.slow)
-//      .waitForElementVisible('.usa-button-big', Timeouts.normal)
+      // The loading indicator tests seem to be problematic because the thing goes away fairly quickly
```
raquelromano-gov: Is there a way to pass a parameter in that indicates this is running on Sauce, and only skip stuff if that's the case?

test/util/nightwatch-commands/axeCheck.js
```diff
@@ -29,11 +29,6 @@ export function command(context, config, _callback) {
       done({ err, results });
     });
   }, [context, (config || {}).rules || this.globals.rules || ['section508', 'wcag2a']], response => {
-    if (response.state !== 'success') {
```
raquelromano-gov: Same here. Or if it's only a specific browser for which the aXe response is weird, is there a way to pass in that information and only skip this check in that case.

OR, just change this test to `if (response.state && response.state !== 'success')` and add a comment explaining which browsers this is needed for. (Curious if that's documented anywhere, or if it's just what you encountered when running it.)

test/user-profile/e2e/user-profile.e2e.spec.js
aub: In this particular case, it seems like the test is not really testing much, right? It's also not really a sauce thing... the test is flawed in that it requires the timing to work in a specific way, so since it's (IMO) both useless and flawed it seemed worth removing.

config/nightwatch-globals-sauce.js
aub: This is somewhere between configuration and executable code... prior to this change the configuration all went in the config directory, so I've stuck with that pattern. The nightwatch-commands directory is a bit different, since they're extensions to the testing library, so even if we could split out the executable part of this I'm not sure it would belong there.

test/util/nightwatch-commands/axeCheck.js
aub: It's just what I encountered while running it. I haven't catalogued which browsers respond in which ways, but I could put that together pretty easily. The thing that struck me here was that this test is really just a short circuit for things that are tested with more depth below. If the response is totally bogus, it's going to fail one of those other tests, and if it's well-formed but has violations then it's going to be flagged as well.

jbalboni (approved):
This looks great.

cehsu (approved):
Exciting!

U-DON (commented):
:+1:

PR #6049, opened 2017-07-17
GIBS QA fixes

annekainicUSDS:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3702

This fixes all the issues outlined in the above ticket except for 2:
1. For letters, address is in all caps: I believe this is something EVSS is going to fix on their side. Will follow up with them about that.
2. For GIBS, print page scrolling issues. While the scrolling is a bit odd for the print page, I wasn't able to reproduce the issues @phiden identified in the ticket. Might be useful to do a screenshare so I can see what Sophia is seeing.

### DIFF
--- a/content/pages/health-care/about-va-health-care.md
+++ b/content/pages/health-care/about-va-health-care.md
@@ -193,7 +193,7 @@ Call our toll-free hotline at 877-222-VETS (<a href="tel:+1-877-222-8387">877-22
 Yes. If you're signed up for the VA health care program, you can manage your VA health and benefits online:
 
 [Refill your prescriptions](/health-care/prescriptions/). <br> 
-[Send a message to your VA health care provider](/health-care/messaging/). 
+[Send a secure message to your health care team](/health-care/messaging/). 
 
 You can also check the status of a disability or pension claim online. [Check your claim status](/track-claims/).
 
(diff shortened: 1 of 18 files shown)


raquelromano-gov (approved):
If you want to fix the all-caps stuff now so we don't have to think about it and remove later once EVSS fixes on their side, you can add `text-transform: capitalize;` (we do this already for the facility name)

annekainicUSDS:
@raquelromano-gov hm, `text-transform: capitalize;` didn't work for me. I did some Googling and apparently `capitalize` only affects the first letters of each word. If the rest of the letters in the word are already capitalized, it won't make them lowercase. I don't think this is actually working elsewhere, at least what I can find. It looks like the data for service info (only other place in letters I see we're using `text-transform: capitalize;` is already coming in lowercase, so capitalize works in that case.

raquelromano-gov:
Yeah, you first have to transform to lower case, then CSS will properly capitalize. We do this here: https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/js/post-911-gib-status/components/EnrollmentPeriod.jsx#L86

annekainicUSDS:
@raquelromano-gov ah gotcha, okay will do! Thanks!

annekainicUSDS:
Messed up a rebase :/ Will reopen once I can merge this into master.

PR #6050, opened 2017-07-17
Burial allowance type question changes

cehsu:
Connects to department-of-veterans-affairs/vets.gov-team#3899

 * Remove parenthetical after non-service connected
 
 * Make VA medical center option conditional if state or other is chosen as location of death

 * Display warning if burial not within 2 years and non-service selected

screen shot 2017-07-17 at 4 51 36 pm

### DIFF
--- a/src/js/burials/config/form.js
+++ b/src/js/burials/config/form.js
@@ -366,11 +366,11 @@ const formConfig = {
                   const filterAllowanceType = createSelector(
                     _.get('locationOfDeath.location'),
                     (locationData) => {
-                      const filterOut = [];
+                      let allowanceTypes = burialAllowanceTypes;
                       if (locationData !== 'vaMedicalCenter' && locationData !== 'nursingHome') {
-                        filterOut.push('vaMC');
+                        allowanceTypes = allowanceTypes.filter(type => type !== 'vaMC');
                       }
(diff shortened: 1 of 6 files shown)


src/js/burials/config/form.js
```diff
@@ -354,7 +360,31 @@ const formConfig = {
...
-                labels: allowanceLabels
+                labels: allowanceLabels,
+                updateSchema: (() => {
+                  const burialAllowanceTypes = burialAllowanceRequested.enum;
+                  const filterAllowanceType = createSelector(
+                    _.get('locationOfDeath.location'),
+                    (locationData) => {
+                      const filterOut = [];
+                      if (locationData !== 'vaMedicalCenter' && locationData !== 'nursingHome') {
+                        filterOut.push('vaMC');
```
jbalboni: I think you could simplify this. What about:

```js
let allowanceTypes = burialAllowanceTypes;

if (locationData !== 'vaMedicalCenter' && locationData !== 'nursingHome') {
  allowanceTypes = allowanceTypes.filter(type => type === 'vaMC');
}

return { 'enum': allowanceTypes };
```

jbalboni (approved):
One minor comment, but looks good.

PR #6055, opened 2017-07-18
Tell users where new pages have been added due to data changes

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3869

We don't do a good job of telling users when new pages pop in, this item attempts to address that. There are two parts: one is that arrays that need an item now show a message (screenshot about disabilities). The second is when new pages are shown, they're highlighted (other screenshots). The logic is listed in that ticket.

There are some follow ups to this, I think. I'd like to move the setViewedPages logic to Redux, but I should do that along with the chapter open/closed state and this PR is already long enough. Second, we can build off this pattern to show validation errors, but we have to put page/chapter validation state in Redux.

Arrays when minItems validation error is tripped:

screen shot 2017-07-18 at 3 32 41 pm
screen shot 2017-07-18 at 3 32 19 pm

When you add a dependent:

screen shot 2017-07-18 at 3 33 04 pm
screen shot 2017-07-18 at 3 33 11 pm
screen shot 2017-07-18 at 3 33 33 pm


src/js/common/schemaform/helpers.js
```diff
@@ -488,3 +488,24 @@ export function expandArrayPages(pageList, data) {
 
   return result.currentList;
 }
+
+/**
+ * getPageKeys returns a list of keys for the currently active pages
+ *
+ * @param pages {Array} List of page configs
+ * @param formData {Object} Current form data
+ * @returns {Array} A list of strings with the page key from the page config
```
cehsu: It looks like this might be missing a parameter name before the data type.

cehsu: There are some options for documenting anonymous objects ([name](http://usejsdoc.org/tags-name.html), [type](http://usejsdoc.org/tags-type.html), [closure compiler example](https://github.com/google/closure-compiler/wiki/Annotating-JavaScript-for-the-Closure-Compiler#record-type)), would any of these make sense here?

jbalboni: We could do `Array<string>`, I think.

src/sass/base/_b-utils.scss
```diff
@@ -12,11 +12,11 @@
   border: 0;
   border-radius: 0;
   outline: 0;
-  padding: 0;
+  padding: 0 !important;
```
cehsu: Would it be helpful to use comments to explain why !important is being used here?

src/js/pensions/components/FinancialDisclosureDescription.jsx
```diff
@@ -0,0 +1,47 @@
...
+  if (!unViewedPages.length) {
+    return null;
+  }
+
+  const jumpToFirst = () => scrollToPage(unViewedPages[0]);
+  let alertBody;
+  if (unViewedPages.some(key => key.includes('spouse'))) {
+    alertBody = (
+      <div className="usa-alert-body">
+        Does your spouse have any net worth, income, or expenses?
```
cehsu: Could it ever be helpful for the alert to display both messages?

src/js/pensions/components/FinancialDisclosureDescription.jsx
```diff
@@ -0,0 +1,47 @@
+import React from 'react';
+import Scroll from 'react-scroll';
+
+const scroller = Scroll.scroller;
+
+function scrollToPage(key) {
+  scroller.scrollTo(`${key}ScrollElement`, {
+    duration: 500,
+    delay: 2,
+    smooth: true,
```
cehsu: Look like there's an extra comma here.

src/js/common/schemaform/review/ReviewPage.jsx
```diff
@@ -45,6 +53,29 @@ class ReviewPage extends React.Component {
     }
   }
 
+  setPagesViewed = (keys) => {
+    const viewedPages = keys.reduce((pages, key) => {
+      if (!pages.has(key)) {
+        // if we hit a page that we need to add, check to see if
+        // we haven't cloned the set yet; we should only do that once
+        if (pages === this.state.viewedPages) {
```
cehsu: Should we use `_.isEqual` here? Will this always be false?

src/js/common/schemaform/review/ReviewPage.jsx
```diff
@@ -45,6 +53,29 @@ class ReviewPage extends React.Component {
...
+          return newPages;
+        }
+
+        pages.add(key);
+      }
+
+      return pages;
+    }, this.state.viewedPages);
+
+    if (viewedPages !== this.state.viewedPages) {
```
cehsu: Same question as above.

src/js/common/schemaform/review/ReviewPage.jsx
```diff
@@ -45,6 +53,29 @@ class ReviewPage extends React.Component {
     }
   }
 
+  setPagesViewed = (keys) => {
+    const viewedPages = keys.reduce((pages, key) => {
+      if (!pages.has(key)) {
```
cehsu: Could we toggle a flag variable here rather than conducting the comparison on line 61?

src/js/common/schemaform/review/ReviewCollapsibleChapter.jsx
```diff
@@ -55,47 +62,60 @@ export default class ReviewCollapsibleChapter extends React.Component {
...
-    if (this.state.open) {
-      const { form, pages } = this.props;
-      const activePages = getActivePages(pages, form.data);
-      const expandedPages = expandArrayPages(activePages, form.data);
+    const { form, pages, viewedPages, chapter } = this.props;
+    const activePages = getActivePages(pages, form.data);
+    const expandedPages = expandArrayPages(activePages, form.data);
 
+    this.pageKeys = getPageKeys(pages, form.data);
+    const hasUnViewedPages = this.pageKeys.some(key =>
```
cehsu: Should this arrow function either be brought onto one line or given curly brackets?

cehsu (approved):
Looks good to me. Had a few minor questions.

src/js/common/schemaform/review/ReviewPage.jsx
jbalboni: We could, but I don't see much benefit to it.

src/js/common/schemaform/review/ReviewPage.jsx
jbalboni: No, I specifically want to do a reference equality check here. If we hit a page that wasn't already in the viewed pages above we clone the Set. That means that we can use a `===` check to see if there are any changes.

src/js/pensions/components/FinancialDisclosureDescription.jsx
jbalboni: I don't _think_ so, but I'm not positive. I think it's likely easier to direct them to the earlier new pages (which are the spouse ones) and they'll notice the later ones below the spouse pages when they complete them.

PR #6058, opened 2017-07-19
Adding events to record success/failure of Post-9/11 GI Bill Status

robertfairhead:
I added the analytics events in with the actions. There are several places the call could be, and I don't feel strongly they have to be in the actions if there's a cleaner place for them. Actions just feels best to me b/c it seems like the place to mutate state, but that's more personal preference than any sort of reasoned argument.

I did the GTM config to report these out so once it's merged, we'll be good to go.

Related to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3446


raquelromano-gov:
Oops--I just updated this file yesterday, so there are a bunch of other action types (specifically, different failure actions). Not sure if you want to separately report each type or just group all the failures together.

robertfairhead:
@raquelromano-gov Ok i fixed up the conflicts. I added in failure events in each path where needed (but doublecheck that).

I didn't differentiate the failures b/c the ask was only to determine if they could see the option to print or not and there did not seem a simple method to pull out the kind of failure. I'd be concerned about the analytics drifting from the code (in the future) so the simpler solution seemed best for now.

raquelromano-gov (approved):
Yep, I agree this level of detail looks good, esp. since the details here about various ways to fail could still change.

Thanks!!

PR #6065, opened 2017-07-19
Change terms acceptance input to checkbox

U-DON:
Resolves department-of-veterans-affairs/vets.gov-team#3823.

I noticed there was the possibility of a 'no' button. But after discussing with @altangel, it seems we would never provide a means to revoke acceptance of terms, so I removed it. If you think it's better that we keep it, I could put it back in.


aub (approved):
LGTM

PR #6068, opened 2017-07-19
Update burial tests with recent changes, make them live

jbalboni:
I forgot to remove the build type flag on the e2e tests when we went live behind a password, so we got out of sync with some recent changes to the form.


cehsu (approved):
Looks good to me.

robbiethegeek-usds (approved):
Thanks @jbalboni :) LGTM

PR #6071, opened 2017-07-19
Update the reset-environment to start using yarn@0.27.5

robbiethegeek-usds:


jbalboni:
@robbiethegeek-usds If we want people to update Yarn, there are some other changes we need to make, to the jenkins and docker scripts (at least).

robbiethegeek-usds:
Created a ticket : https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3943 for adding to the playbook @jbalboni

jbalboni:
@robbiethegeek-usds I'm a little surprised at how different that lock file is from the current one. Even with the version bump, that seems like a lot.

robbiethegeek-usds:
I am going to reset the lock file and then submit it after this goes through

jbalboni (approved):

script/reset-environment.sh
```diff
@@ -1,13 +1,26 @@
...
     if [ $? -eq 0 ]; then
         echo "Yarn successfulling installed globally."
     else 
         echo "Yarn failed to install...please install manually."
         exit 1
     fi
+else
+    yarn_version=$(yarn --version)
+    if [ "$yarn_version" != "0.27.5" ]; then
+        echo "Install yarn version 0.27.5 [y/N]: "
```
cvalarida: Just for fun, can we also display the current version of yarn (to make it obvious why we're asking)?

```bash
echo "Current yarn version is $yarn_version. Install yarn version 0.27.5 [y/N]: "
```
Or something.

cvalarida (approved):

jkassemi (approved):
🚀

PR #6075, opened 2017-07-20
Fix LoginModal test with timeout

jbalboni:
This test was causing random failures whenever it ran fast enough to execute the code inside the timeout (which was rarely).

The way it was previously written, `loginSpy` would never be called, because there's nothing that will force the component to re-render and run `componentWillReceiveProps`. But since it was wrapped in `setTimeout`, Mocha would not normally wait for the timeout to finish (you have to include `done` in the parameters of the test function for that to happen). But sometimes it appears to finish while the schemaform action tests are running (not sure why that test specifically).

I split the test into two, to cover the logic we're trying to test. If you wanted to an actual integration test between this modal and the login code, you would need to set up Redux or do a lot of mocking, which seems like stretching the bounds of unit tests (in my opinion, anyway).


cehsu (approved):
Thanks for splitting this up.

PR #6078, opened 2017-07-20
Move rjsf form submission to xhr for better error visibility

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3144

Using xhr instead of fetch gives us some more error visibility by being able to see timeout and abort errors. This may not help at all, but we'll at least be consistent across browsers. It should address the issue with fetch and Edge 14, too.

### DIFF
--- a/src/js/common/schemaform/actions.js
+++ b/src/js/common/schemaform/actions.js
@@ -47,7 +47,6 @@ export function setSubmitted(response) {
   };
 }
 
-
 export function submitForm(formConfig, form) {
   const body = formConfig.transformForSubmit
     ? formConfig.transformForSubmit(formConfig, form)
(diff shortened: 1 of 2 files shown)


src/js/common/schemaform/actions.js
```diff
@@ -59,46 +58,86 @@ export function submitForm(formConfig, form) {
...
+          const results = JSON.parse(responseBody);
+          resolve(results);
+        } else {
+          const error = new Error(`vets_server_error: ${req.statusText}`);
+          Raven.captureException(error, {
+            extra: {
+              clientError: false,
+              statusText: req.statusText
+            }
+          });
```
U-DON: This gets duplicated quite a bit. Might it be helpful to have this as a function within the scope of this promise?

U-DON (approved):
LGTM.

cehsu (approved):
Looks good to me.

src/js/common/schemaform/actions.js
jbalboni: I ended up moving this out more and calling it from the promise's `catch` function. It's possible the way I had it before would miss errors in the promise callback function.

PR #6082, opened 2017-07-21
GIBS Update link

annekainicUSDS:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3952


rroueche:
@annekainicUSDS is there a reason for using the absolute URL instead of relative?

annekainicUSDS:
@rroueche I've heard pros and cons for both, apparently absolute URLs are better for SEO. But seems like a relative url is mostly used from within the same domain, so I'll switch to that.

rroueche:
@annekainicUSDS i might be misunderstanding, but wouldn't the relative URL be `/gi-bill-comparison-tool` instead of `/vets.gov/gi-bill-comparison-tool` ?

rroueche (approved):
LGTM

PR #6086, opened 2017-07-21
Pensions e2e documents ceh

cehsu:
Connects to department-of-veterans-affairs/vets.gov-team#3417
Connects to department-of-veterans-affairs/vets.gov-team#3743


jbalboni:
@cehsu I want to try testing this locally, but I get an error at the second dependent's expected income page. Any idea?

cvalarida (commented):
What's `test/pensions/test.png` for?

cehsu:
Thanks @jbalboni, that's an interesting bug. The logic in `00-all-fields.e2e.spec.js` and test data in `maximal-test.json` for both dependents are the same.
Thanks @cvalarida, the `test/pensions/test.png` is to test out mocking document upload, thanks for any thoughts on whether there is another better way to go about this, or if you also have thoughts on the issue @jbalboni pointed out, as I do not currently encounter that error locally.

jbalboni:
It looks like on my machine we need some wait calls before the additional sources inputs. I'm not entirely sure why. Changes pushed here.

jbalboni (approved):

cvalarida (approved):

PR #6088, opened 2017-07-24
Add trademark symbol to print page

raquelromano-gov:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3983

image


rroueche:
@raquelromano-gov please see mockup for how to implement:https://marvelapp.com/8d6igd9/screen/29023992

The symbols should appear only in the page title, and should be an R instead of TM (per research that Laura did and I validated.)


Rachael

rroueche (approved):

PR #6089, opened 2017-07-24
Update tests to work with pagePerItem pages

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3690
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3473

Updates our `DefinitionTester` component to do the same logic as `FormPage` when you pass in an `arrayPath` prop.

### DIFF
--- a/test/pensions/config/childrenAddress.unit.spec.jsx
+++ b/test/pensions/config/childrenAddress.unit.spec.jsx
@@ -7,8 +7,7 @@ import { DefinitionTester, getFormDOM } from '../../util/schemaform-utils.jsx';
 import formConfig from '../../../src/js/pensions/config/form.js';
 
 describe('Child address page', () => {
-  const schema = formConfig.chapters.householdInformation.pages.childrenAddress.schema.properties.dependents.items;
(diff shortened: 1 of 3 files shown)


cehsu (approved):
LGTM. So exciting!

test/pensions/config/childrenAddress.unit.spec.jsx
```diff
@@ -34,11 +35,30 @@ describe('Child address page', () => {
...
+          arrayPath={arrayPath}
+          pagePerItemIndex={0}
+          definitions={formConfig.defaultDefinitions}
+          schema={schema}
+          data={nameData}
+          uiSchema={uiSchema}/>
+    );
+
+    const formDOM = getFormDOM(form);
+    formDOM.selectRadio('root_childInHouseholdNo', 'N');
```
cvalarida: Out of curiosity, is this the preferred way now? You could also do
```js
formDOM.selectYesNo('root_childInHousehold', 'N');
```
I wrote that one before `selectRadio`, though, so we could depreciate `selectYesNo` if we wanted.

Have you an opinion?

PR #6093, opened 2017-07-24
Updating the yarn.lock file after yarn update to 0.27.5.

robbiethegeek-usds:
@jbalboni as promised :)


jbalboni:
@robbiethegeek What did you do to generate the new file? There are a bunch of minor version bumps, like going from Webpack 2.4 to 2.7. They're updates we should do, but it should be intentional, at least.

robbiethegeek-usds:
just removed the yarn.lock and did an install.  If that is an issue it seems like the package.json needs to be locked down more.

jbalboni:
@robbiethegeek Ah, I see. That's what I would expect to happen in that case. What happens if you run `yarn install` with the old yarn.lock file? Does it change? I think that might be all we need to do to make sure it's up to date with the new yarn version.

I think the point of the lock file is partly so that we don't have to spend a bunch of time locking down package.json (which only works to a point). This is sort of like running `yarn upgrade`, rather than just moving to a new version of yarn.

Anyway, I guess my point is: if we want to upgrade our dependencies like this is doing, that's fine, but we should probably have everyone do some testing and sign off. If we just want to get everyone on the newer version of yarn, I think we might just be able to run `yarn install` with the old lock file in place and commit any changes.

robbiethegeek-usds:
@jbalboni it doesn't actually change the lock file if I run yarn install with 0.27.5 against the current yarn.lock

robbiethegeek-usds:
@jbalboni any thoughts?

jbalboni:
@robbiethegeek-usds In that case, I don't think we need to do this to be on the most recent version of yarn? The changes here are upgrading a bunch of dependencies, which if you want to take that on, that's fine with me, but it doesn't have to be done now.

robbiethegeek-usds:
Closing for now

PR #6094, opened 2017-07-24
Add GA event to capture facility type in FL search

robertfairhead:
Per Slack convo with @NatalieEMoore this adds a GA event to capture which facility type was selected on a FL search.

Tagging @patrickvinograd for review per Natalie

I added the event push in the onClick handler b/c from looking through it seemed like multiple API calls may get triggered from a single user action so I didn't want to push this lower in the stack near the `fetch` call like I usually would. Definitely open to thoughts if a better place to capture this since I only spent an hour or two looking over FL so I may have missed an edge case or something.

### DIFF
--- a/src/js/facility-locator/components/AppointmentInfo.jsx
+++ b/src/js/facility-locator/components/AppointmentInfo.jsx
@@ -69,6 +69,7 @@ export default class AppointmentInfo extends Component {
       }
 
       const onClick = () => {
+        window.dataLayer.push({ event: 'fl-show-waittimes' });
         this.setState({
           [showHideKey]: !this.state[showHideKey],
         });
--- a/src/js/facility-locator/components/SearchControls.jsx
+++ b/src/js/facility-locator/components/SearchControls.jsx
@@ -59,6 +59,13 @@ class SearchControls extends Component {
     const { onSearch } = this.props;
     e.preventDefault();
 
+    const { facilityType } = this.props.currentQuery;
+    // Report event here to only send analytics event when a user clicks on the button
+    window.dataLayer.push({
+      event: 'fl-search',
+      'fl-search-fac-type': facilityType
+    });
+
     onSearch();
   }
 


patrickvinograd (commented):
Putting the event in this handler will capture all clicks of the `Search` button. It omits the queries that occur upon loading the facility locator by submitting a URL (vs. clicking Search once already in the app). 

Historically none of those initial load views would have a facilityType populated, but as of this pending PR: 
https://github.com/department-of-veterans-affairs/vets-website/pull/6098
it's possible to navigate directly to a URL like "www.vets.gov/facilities?facilityType=vet_center" and the facilityType param will be respected. So you'd be undercounting these initial-load page views, either when geocoding is enabled or when somebody loads a bookmarked URL with an address populated.

Eventing on actions/index.js in `searchWithBounds` would cover literally every API query, but that might be too much because it would include e.g. zooming/panning around the map. But if that's what you're going for, that's the place to do it.

robertfairhead:
Dude, way to leave me in suspense. "But as of this PR:" AS OF THIS PR??? CLIFFHANGER!!!

patrickvinograd (commented):
See comment.

robertfairhead:
Yeah, I originally had it in the actions but that was my concern, re: not respecting user actions.

In that PR, if I'm reading it correctly, facilityType will always be sent via the URL as queryParam if it's selected, correct? If so, then I don't *need* the custom event in the JS. I can just parse the URL for the query param when it's sent to GA. It'll have the same issue with panning/zooming I assume b/c a page view gets fired on any history change and if I understand it correctly, you update the URL params on pan/zoom, etc.

I'll check with Natalie on her preference and update/close this PR as necessary.

robertfairhead:
Discussed with Natalie via Slack, will keep the event in the onClick handler on the assumption that entering via a URL param won't be a significant percentage of traffic.

Once that PR is done and facilityType is set in the URL, I can parse that out after the fact to check on that assumption and ensure the relative percentages of facilityTypes being reported look accurate.

That sound good @patrickvinograd?

robertfairhead:
@patrickvinograd I added the "Show More" event that @NatalieEMoore requested in Slack to this PR.

It doesn't differentiate showing from hiding, but the ask seemed mostly about overall usage versus that difference, so I thought I'd keep it simple.

patrickvinograd (approved):
LGTM.

PR #6095, opened 2017-07-24
Raise the search bar above the nav menu

cvalarida:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3967

image

image

I also tested to make sure it was under a modal, and that was fine.

@goldenmeanie mentioned possibly expanding the banner vertically when the search bar is shown, but this is significantly more complicated. If that's still the route we want to go, we can figure out how to make it happen, but for the sake of shipping a fix, I wanted to get this out sooner rather than later.


cehsu (approved):
Looks good to me.

jbalboni (approved):

PR #6098, opened 2017-07-24
Read/write filter parameters to URL query parameters.

patrickvinograd:
Fixes https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3989

With this you can use a URL like: `http://localhost:3001/facilities/?facilityType=vet_center` and it will load with the facility-type pre-selected. Other behavior (whether it geolocates or defaults to Washington DC) remains the same.

Also fixed a latent bug I observed where if you toggled between the two facility types that have service filters (benefits and vet centers), it would not reset the service type, leading to the possibility of submitting a search for vet centers with a benefits-appropriate service filter (or vice versa) which would result in an API error.

### DIFF
--- a/src/js/facility-locator/components/SearchControls.jsx
+++ b/src/js/facility-locator/components/SearchControls.jsx
@@ -85,17 +85,19 @@ class SearchControls extends Component {
     }
   }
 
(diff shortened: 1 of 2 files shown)


U-DON (approved):
LGTM.

PR #6100, opened 2017-07-25
Moved up facility type line in info.

U-DON:
Resolves department-of-veterans-affairs/vets.gov-team#1075.

@altangel: Note that I also moved it up in the facility's detail page. Figured the ordering of lines should be analogous to its search result block. Let me know if I should revert it for the detail page.


altangel:
@U-DON That's good! Thanks for doing that.

patrickvinograd (approved):
LGTM.

PR #6101, opened 2017-07-25
Show expiration date for SiP forms

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3590

The time formatting is what I thought was reasonable, but open to suggestions. Moment's calculations are fuzzy and that felt wrong for an expiration time that someone might look at more carefully.

screen shot 2017-07-25 at 10 32 26 am

### DIFF
--- a/src/js/common/schemaform/SaveInProgressIntro.jsx
+++ b/src/js/common/schemaform/SaveInProgressIntro.jsx
@@ -8,6 +8,7 @@ import { fetchInProgressForm, removeInProgressForm } from './save-load-actions';
 import SignInLink from '../components/SignInLink';
 import LoadingIndicator from '../components/LoadingIndicator';
 import FormStartControls from './FormStartControls';
+import { dateDiffDesc } from '../utils/helpers';
 
 export default class SaveInProgressIntro extends React.Component {
   getAlert(savedForm) {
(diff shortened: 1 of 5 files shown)


src/js/common/schemaform/SaveInProgressIntro.jsx
```diff
@@ -18,13 +19,14 @@ export default class SaveInProgressIntro extends React.Component {
         const savedAt = this.props.lastSavedDate
           ? moment(this.props.lastSavedDate)
           : moment.unix(savedForm.last_updated);
+        const expirationDate = moment(savedAt).add(30, 'days');
```
jbalboni: Should figure out if this is in the backend somewhere, maybe use that instead.

markolson: Copying from Slack, this _should_ be available in `savedForm.expires_at` if `savedForm` references the `metadata` of the saved claim.

jbalboni: This is the data that's in the profile, so I think it's just partial info from the `metadata`. It might make sense to add `expires_at` to that data.

src/js/common/schemaform/SaveInProgressIntro.jsx
```diff
@@ -18,13 +19,14 @@ export default class SaveInProgressIntro extends React.Component {
...
 
         alert = (
           <div>
             <div className="usa-alert usa-alert-info no-background-image schemaform-sip-alert">
               <div style={{ paddingBottom: '8px' }}>Application status: <strong>In progress</strong></div>
               <br/>
-              <div>Last saved on {savedAt.format('MM/DD/YYYY [at] h:mma')}</div>
+              <div>Last saved on {savedAt.format('MM/DD/YYYY [at] h:mma')}. <span className="schemaform-sip-expires">Expires in {dateDiffDesc(expirationDate)}</span>.</div>
```
jbalboni: Yes, that is it (mentioned that it's "fuzzy" in the PR desc).

cvalarida: Ah, that's what you meant by that. Suppose I should read those more carefully 😛

Looks good to me, then!

cvalarida (approved):

cehsu (approved):
Looks good to me.

src/js/common/schemaform/SaveInProgressIntro.jsx
jbalboni: I added an issue to change the backend so we can update it here later: https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4050

PR #6103, opened 2017-07-25
Remove expanding group error style

cehsu:
Connects to department-of-veterans-affairs/vets.gov-team#3686

screen shot 2017-07-25 at 11 23 57 am

### DIFF
--- a/src/sass/modules/_m-form-elements.scss
+++ b/src/sass/modules/_m-form-elements.scss
@@ -141,7 +141,7 @@ button.form-button-disabled {
   margin-top: 0 !important;
 }
 
-.form-expanding-group {
+.form-expanding-group > span {
   .usa-input-error {
     padding-left: calc(1.5rem - 6px);
     right: calc(1.9rem - 6px);
--- a/src/sass/modules/_m-form-elements.scss
+++ b/src/sass/modules/_m-form-elements.scss
@@ -146,6 +146,12 @@ button.form-button-disabled {
     padding-left: calc(1.5rem - 6px);
     right: calc(1.9rem - 6px);
   }
+  .schemaform-expandUnder-indent {
+    .usa-input-error {
+      padding-left: 1.5rem;
+      right: 1.9rem;
+    }
+  }
 }
 
 .form-expanding-group {
--- a/src/js/pensions/config/form.js
+++ b/src/js/pensions/config/form.js
@@ -1038,7 +1038,7 @@ const formConfig = {
                 },
                 previouslyMarried: {
                   'ui:title': 'Has your child ever been married?',
-                  'ui:widget': 'yesNo',
+                  'ui:widget': 'yesNo'
                 },
                 married: {
                   'ui:title': 'Are they currently married?',


src/sass/modules/_m-form-elements.scss
```diff
@@ -142,13 +142,6 @@ button.form-button-disabled {
 }
 
 .form-expanding-group {
-  .usa-input-error {
```
cvalarida: Oh, that's...really bizarre. Why was this here in the first place?

cvalarida (approved):

src/sass/modules/_m-form-elements.scss
jbalboni: I must have added it a while back so that the bars lined up underneath a group. Not sure why it would be different now.

jbalboni:
@cehsu @cvalarida This is why the styles were there:

screen shot 2017-07-27 at 9 37 05 am

cehsu:
@jbalboni @cvalarida Is the desired behavior to remove the blue line?

jbalboni:
I think the desired behavior is to have the red line to the right of the blue line, but maybe @goldenmeanie can weigh in.

goldenmeanie:
Yes, original intent was to the right of the blue bar.

Can see a couple ways forward here:
1. To right of bar, fields stay lined up
2. To right of bar, fields bumped over to maintain spacing with alert highlight
3. Total eclipse of the blue bar, effectively turning part of it red

cehsu:
Is this open to a vote? If so, I vote option 3. If not, that's cool too.

cvalarida:
I think the way we had it before is fine, but if there is no blue line, the red line needs to line up with the other red lines. I'd need to look at it more closely, but I think we can manage this with some css-fu. (So option 1.)

Basically, the trouble is that the fields inside an `expandUnder` get the blue line, but fields inside a `hideIf` do not, but the shifted red line applies to both.

cehsu:
@cvalarida so would hideIf need to add a class, and then if an element has that class, we would unset the styles added in the currently deleted code (which would be restored)?

jbalboni:
I don't think `form-expanding-group` is used for `hideIf` fields.

cehsu:
Thanks for the clarification @jbalboni, so the desired behavior is to keep things the same? Should we close this PR?

jbalboni:
@cehsu I think what @cvalarida created the original issue for is still a problem. I think you should look into that and figure out the cause, and/or a way to fix it while keeping the red par position inside expanded groups.

cehsu:
It looks like we should be using expandUnderClassNames for this instead?

cehsu:
Is this the desired behavior?
### Not expanded
screen shot 2017-07-27 at 3 14 34 pm

### Expanded
screen shot 2017-07-27 at 3 13 33 pm

jbalboni:
@cehsu Yeah, that's how it was intended to look before. If this is what the normal behavior should be, I don't want to make us have to set `expandUnderClasses` everywhere to make it happen, though.

cehsu:
#### Not expanded
screen shot 2017-07-28 at 10 54 19 am

#### Expanded
screen shot 2017-07-28 at 10 54 38 am

#### Expanded with `schemaform-expandUnder-indent` class
screen shot 2017-07-28 at 10 53 41 am

src/sass/modules/_m-form-elements.scss
```diff
@@ -141,11 +141,17 @@ button.form-button-disabled {
...
 
-.form-expanding-group {
+.form-expanding-group > span {
   .usa-input-error {
     padding-left: calc(1.5rem - 6px);
     right: calc(1.9rem - 6px);
   }
+  .schemaform-expandUnder-indent {
+    .usa-input-error {
+      padding-left: 15px;
```
jbalboni: Can we use the original units for this? `rem`, I think.

jbalboni (approved):
LGTM, had one nitpick

cehsu:
Thanks @jbalboni, is the updated code what you had in mind?

jbalboni:
@cehsu I'm not sure you need the calc. Shouldn't those values match the default from USWDS?

cehsu:
Thanks @jbalboni, you're right. Change made.

PR #6087, opened 2017-07-22
Show status descriptions for previous appeals history events in accordion style.

U-DON:
Resolves department-of-veterans-affairs/vets.gov-team#3870.

screen shot 2017-07-21 at 8 37 44 pm

PR #6109, opened 2017-07-26
Appeals spacing tweaks

U-DON:
Addendum to #6087 to center the icons. Also semantic correction to define the list as an `ol`.

The latter also correctly styles the spacing of paragraphs in the description contents due to a rule for `ul li p` that removes margins and padding and is the cause of department-of-veterans-affairs/vets.gov-team#3964. That rule should be removed separately, but it's still more correct to declare this list as `ol` than `ul`.

screen shot 2017-07-25 at 8 00 53 pm


aub (approved):
LGTM

PR #6110, opened 2017-07-26
Leave margins and padding for 'ul li p' by default.

U-DON:
Resolves department-of-veterans-affairs/vets.gov-team#3964.

Removing a rule that causes problem for some content. `p` elements would otherwise have some default margins, and I think it should be up to individual use cases to remove them within `ul` elements. Additionally, it seems like it would be overkill if I wanted to add back margins using a specific class. Also strange that this wasn't applied to paragraphs within `ol` elements.

### DIFF
--- a/src/sass/base/_va.scss
+++ b/src/sass/base/_va.scss
@@ -1170,14 +1170,6 @@ li {
   }
 }
 
-// List overrides
-
-ul li p {
-  margin: 0;
-  padding: 0;
-}
-
-
 // vets.gov branded apps
 .vets-app {
   border-left: 3px solid $color-gold;


jbalboni:
Yeah, I'm in favor of removing this, but not sure what it would affect. I can take a spin through the forms later today.

cvalarida (approved):
Oh man. That was from [PR #1](https://github.com/department-of-veterans-affairs/vets-website/blame/master/src/sass/base/_va.scss#L1175). I'd run through a bunch of stuff with Chrome's coverage tool running, but I'm also in favor of nixing this.

U-DON:
Going to merge this. Generally, I've found that the markdown that would get affected by this already seems to differentiate itself from the list items that do not have margins between them by including newlines between the bullets. So without this seemingly arbitrary rule, it would actually display as expected.

Example: https://github.com/department-of-veterans-affairs/vets-website/blob/master/content/pages/burials-and-memorials/survivor-and-dependent-benefits/compensation.md

PR #6120, opened 2017-07-27
Use expiration date from backend for SiP

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4050

Uses the backend expiration date instead of one from the front end. Also filters out expired forms, while we figure out what the best way of handling that is.


cvalarida (approved):
:100:

PR #6121, opened 2017-07-27
Remove awardEffectiveDate as option and add to monthlyAward option

annekainicUSDS:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3872

While both `awardEffectiveDate` and `monthlyAward` are sent to us in the response for the benefit options list, only one is needed to be displayed (`monthlyAward`). However, the value for `awardEffectiveDate` should now be shown in the display text for the `monthlyAward` checkbox. To do this I added an argument to `getBenefitOptionText` so I could pass the `awardEffectiveDate` value, because `getBenefitOptionText` is run while mapping over each option, and could only take the value of the current option it's mapped on, which means it couldn't accept the values for multiple options. This is kind of a hacky way of doing this, so I wonder if there's a more extensible solution. Would love feedback on this.

screen shot 2017-07-27 at 3 46 19 pm


src/js/letters/utils/helpers.js
```diff
@@ -142,7 +142,7 @@ const benefitOptionText = {
   }
 };
 
-export function getBenefitOptionText(option, value, isVeteran) {
+export function getBenefitOptionText(option, value, isVeteran, awardEffectiveDate) {
```
annekainicUSDS: Not sure if I should make the argument so specific, or just something more generic like `additionalInfo` in case we want to use this for other things in the future? Or if there's a better way to handle this entirely.

test/letters/utils/helpers.unit.spec.js
```diff
@@ -69,6 +68,10 @@ describe('Letters helpers: ', () => {
         expect(getBenefitOptionText(option, 'unavailable', true)).to.be.undefined;
       }, ['monthlyAwardAmount', 'serviceConnectedPercentage']);
     });
+    it('should include the awardEffectiveDate in the text for monthlyAward', () => {
+      const tree = getBenefitOptionText('monthlyAwardAmount', 20, true);
+      expect(tree.props.children[1].props.children[0]).to.contain('The effective date');
```
annekainicUSDS: This is really hacky. I tried lots of different solutions to get the text content of a child element of what I rendered, but I couldn't figure out a good way to do this. I think I ran into issues because what I'm rendering in `getBenefitOptionText` is not a React component, but just some React elements  (or DOM elements?). Can't figure out a good way of finding text content for an element that is lower down in a hierarchy of React/DOM elements.

raquelromano-gov:
Hey, finally back and checking github notifications for the first time in a week. I will try to review this tomorrow evening (sorry for the delay...have not slept much and reviewing it now would be fruitless)

src/js/letters/utils/helpers.js
jbalboni: I don't have another obvious solution. You could pass `benefitInfo`, `key`, and `isVeteran` to this function instead, and pull `value` and `awardEffectiveDate` out of `benefitInfo` in the function body. But I'm not sure that's a lot better.

test/letters/utils/helpers.unit.spec.js
jbalboni: I haven't looked at this at all, but it seems like you should be able to do `SkinDeep.shallowRender(tree)`, but if that's one of the things you tried and didn't work, this seems ok.

src/js/letters/utils/helpers.js
jbalboni: Sort of random nitpick: this file should probably be `.jsx` instead of `.js`.

jbalboni (approved):

src/js/letters/utils/helpers.js
annekainicUSDS: thanks! will fix

test/letters/utils/helpers.unit.spec.js
annekainicUSDS: I got that to work by doing that and then testing `tree.text()`. Thanks for the suggestion! So much better this way.

PR #6122, opened 2017-07-28
Style fixes for acceptance line of MHV terms and conditions.

U-DON:
Resolves department-of-veterans-affairs/vets.gov-team#3962.

Moved the yes content to one line.

Ensured a consistent cursor when hovering around the label (default when disabled and pointer when enabled). @altangel: Let me know if that's the desired behavior. Also whether the cursor should be a pointer (hand) when hovering over the text or the line in general (as in the white space next to the text).

screen shot 2017-07-27 at 9 44 30 pm

### DIFF
--- a/src/sass/login.scss
+++ b/src/sass/login.scss
@@ -73,6 +73,7 @@
       }
 
       label {
+        display: inline-block;
         max-width: none;
       }
     }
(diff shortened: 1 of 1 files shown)


altangel:
@U-DON I think you should be able to hover over the box through text and it would show the pointer and be clickable. No need whitespace to the right. Thanks!

U-DON:
Pinging @aub or @patrickvinograd to review.

aub (approved):
LGTM

PR #6123, opened 2017-07-28
Fix case when no primary care wait time but yes other wait time data

patrickvinograd:
Fixes https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4112

Only affected a small number of facilities - those that were reporting wait times for some specialty categories but not for primary care. Avoids assuming that the primary care element will be present.

### DIFF
--- a/src/js/facility-locator/components/AppointmentInfo.jsx
+++ b/src/js/facility-locator/components/AppointmentInfo.jsx
@@ -1,5 +1,5 @@
 import React, { Component } from 'react';
-import { some, pull, startCase } from 'lodash';
+import { get, some, pull, startCase } from 'lodash';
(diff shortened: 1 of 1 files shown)


patrickvinograd:
Verified locally - 757QC loads as expected; ran through my list of other corner case facilities and they all look fine.

U-DON (approved):
LGTM. If you wanted, you could also use `get` from `lodash`. Note that whether the path resolves to `null` or `undefined`, it would be regarded as a falsy value and evaluate as `false` in the `&&` expression, thus not rendering anything.

```
return get(accessAttrs, ['primaryCare', category]);
```

https://lodash.com/docs/4.17.4#get

PR #6124, opened 2017-07-28
Log review page validation errors to Sentry

jbalboni:
We're seeing enough validation errors on the review page for the edu rjsf forms that we want to log what they are in Sentry. This creates a Sentry error when this happens and adds the errors to the extra info. The `instance` property is filtered out so we don't send PII.


cehsu (approved):
Looks good to me.

PR #6127, opened 2017-07-28
Address the Moment.js deprecation warning

cvalarida:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4045

I added a `DATE_FORMAT` in `src/js/common/utils/helpers.js` to address the depreciation warning moment was giving us. If we run into more of these warnings, we just have to track down where we're calling `moment(dateString)` without specifying the format and give it the `DATE_FORMAT` (unless it's not a `YYYY-MM-DD` string, of course).

### DIFF
--- a/src/js/pensions/config/form.js
+++ b/src/js/pensions/config/form.js
@@ -2,6 +2,8 @@ import _ from 'lodash/fp';
 import moment from 'moment';
 import { createSelector } from 'reselect';
 
+import { DATE_FORMAT } from '../../common/utils/helpers';
+
 import fullSchemaPensions from 'vets-json-schema/dist/21P-527EZ-schema.json';
 import { isFullDate } from '../../common/utils/validations';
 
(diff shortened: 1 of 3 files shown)


src/js/pensions/config/form.js
```diff
@@ -1025,14 +1035,14 @@ const formConfig = {
...
+                    hideIf: (formData, index) => !isBetween18And23(_.get(['dependents', index, 'childDateOfBirth'], formData))
                   }
                 },
                 disabled: {
                   'ui:title': 'Is your child seriously disabled?',
-                  'ui:required': (formData, index) => !isEligibleForDisabilitySupport(_.get(['dependents', index, 'childDateOfBirth'], formData)),
+                  'ui:required': (formData, index) => isEligibleForDisabilitySupport(_.get(['dependents', index, 'childDateOfBirth'], formData)),
                   'ui:options': {
-                    hideIf: (formData, index) => isEligibleForDisabilitySupport(_.get(['dependents', index, 'childDateOfBirth'], formData))
+                    hideIf: (formData, index) => !isEligibleForDisabilitySupport(_.get(['dependents', index, 'childDateOfBirth'], formData))
```
cvalarida: The logic in `isEligibleForDisabilitySupport()` is reversed now, hence the `!` swap.

test/pensions/config/childrenInformation.unit.spec.jsx
```diff
@@ -50,7 +53,8 @@ describe('Child information page', () => {
     );
     const formDOM = getFormDOM(form);
     formDOM.submitForm(form);
-    expect(formDOM.querySelectorAll('.usa-input-error').length).to.equal(4);
+
+    expect(formDOM.querySelectorAll('.usa-input-error').length).to.equal(3);
```
cvalarida: Previously, the date passed in (`undefined`) was causing `isEligibleForDisabilitySupport()` to return true, so the `disabled` field was showing up when it shouldn't have. That's why we went from 4 to 3 required fields.

test/pensions/config/childrenInformation.unit.spec.jsx
```diff
@@ -68,12 +72,13 @@ describe('Child information page', () => {
...
     formDOM.setCheckbox('#root_view\\:noSSN', true);
+
     formDOM.submitForm(form);
     const errors = formDOM.querySelectorAll('.usa-input-error-label');
 
-    errors.forEach(e => console.log(e.getAttribute('for'))); // eslint-disable-line no-console
+    // errors.forEach(e => console.log(e.getAttribute('for')));
 
-    expect(errors.length).to.equal(3);
+    expect(errors.length).to.equal(2);
```
cvalarida: Same as above.

ayaleloehr:
pinging on status here

jbalboni:
I'm going to close this for now and leave the branch up. We decided in a meeting to address this by walking through several forms and addressing specific instances of the warning. The person who does that can resurrect this PR/branch if they want.

cvalarida:
When we do that, I still think it makes sense to use a single format variable in all the places instead of magic strings.

Issue #4061, opened 2016-11-21
Investigate shutting off date changes on scroll


cweber:
This is in Chrome and seems to be a browser default. So cool! ....not cool. Can we turn this off?

![date-scrolling](https://cloud.githubusercontent.com/assets/417657/20497420/0bd06890-aff7-11e6-8250-1147020a38be.gif)

J-Quag:
Moving this out of ice box for after the pension form is released. Happens on more than just dates. the pension form has a large number of number fields.

PR #6129, opened 2017-07-28
Prevent scrolling from changing number inputs

cvalarida:
Connects to https://github.com/department-of-veterans-affairs/vets-website/issues/4061

I couldn't find a better solution, so this works. It also stops scrolling when hovering over the number input in focus, but...I couldn't find a better solution.

### DIFF
--- a/src/js/common/index.js
+++ b/src/js/common/index.js
@@ -9,3 +9,13 @@ require('./utils/sticky-action-box.js');
 
 // New navigation menu
 require('../legacy/mega-menu.js');
+
+// Disable scrolling on input[type="number"] because it was changing the value unexpectedly
+// NOTE: This also just stops the user from scrolling when hovering over a number
+//  input that has focus. This isn't ideal, but until a solution is found to that,
+//  this is what we've got.
+document.addEventListener('mousewheel', (event) => {
+  if (event.target.type === 'number' && document.activeElement === event.target) {
+    event.preventDefault();
+  }
+});
--- a/src/js/common/polyfills.js
+++ b/src/js/common/polyfills.js
@@ -24,3 +24,13 @@ if (!Modernizr.fetch) {
 
 // This polyfill has its own test logic so no need to conditionally require.
 require('polyfill-function-prototype-bind');
+
+// Disable scrolling on input[type="number"] because it was changing the value unexpectedly
+// NOTE: This also just stops the user from scrolling when hovering over a number
+//  input that has focus. This isn't ideal, but until a solution is found to that,
+//  this is what we've got.
+document.addEventListener('mousewheel', (event) => {
+  if (event.target.type === 'number' && document.activeElement === event.target) {
+    event.preventDefault();
+  }
+});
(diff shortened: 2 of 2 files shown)


src/js/common/polyfills.js
```diff
@@ -24,3 +24,13 @@ if (!Modernizr.fetch) {
 
 // This polyfill has its own test logic so no need to conditionally require.
 require('polyfill-function-prototype-bind');
+
+// Disable scrolling on input[type="number"] because it was changing the value unexpectedly
+// NOTE: This also just stops the user from scrolling when hovering over a number
+//  input that has focus. This isn't ideal, but until a solution is found to that,
+//  this is what we've got.
+document.addEventListener('mousewheel', (event) => {
```
cvalarida: This can be pulled into a different file if it makes more sense to do that. Also, I can make this more like:

```js
export function preventScrolling(element) {
  element.addEventListener('mousewheel', (event) => {
    if (event.target.type === 'number' && document.activeElement === event.target) {
      event.preventDefault();
    }
  });
}
```
So we can call it only when we want to instead of making this global on `document`.

I'm not super fond of this, but it works. Mostly.

cehsu: This looks great, there is some debate about whether blur is also helpful in this situation [here](https://stackoverflow.com/questions/9712295/disable-scrolling-on-input-type-number
). Perhaps worth considering if it makes UX sense.

cehsu (approved):
Looks good to me. Had two minor questions.

src/js/common/polyfills.js
jbalboni: I think always adding it on `document` is fine until we have a reason not to. But maybe we should just put this in common/index.js? It's not really a polyfill. I don't have a strong opinion on that.

jbalboni (approved):
One minor comment.

src/js/common/polyfills.js
cvalarida: @cehsu Yeah, I wasn't sure about that. I was shying away from the accessibility implications there might be if we use `.blur()`, but I'll let @goldenmeanie weigh in on what's less bad.

cvalarida: @cehsu Oh, and yes, we do need that `document.activeElement === event.target`, otherwise it'll stop scrolling on _all_ number fields whether or not they have the focus. The problem with the scrolling affecting the value in the number field only happens when the field has focus and we're hovering over it, so this takes care of that. Just narrows the weird UX is all. 😛

src/js/common/index.js
```diff
@@ -9,3 +9,13 @@ require('./utils/sticky-action-box.js');
 
 // New navigation menu
 require('../legacy/mega-menu.js');
+
+// Disable scrolling on input[type="number"] because it was changing the value unexpectedly
+// NOTE: This also just stops the user from scrolling when hovering over a number
+//  input that has focus. This isn't ideal, but until a solution is found to that,
+//  this is what we've got.
+document.addEventListener('mousewheel', (event) => {
```
cehsu: It looks like [`wheel` should be used instead of `mousewheel`](https://developer.mozilla.org/en-US/docs/Web/Events/mousewheel).

src/js/common/polyfills.js
goldenmeanie: The main accessibility issue is about trapping people in an input. I'll ask around about this, but it seems, on first review, to be okay.

PR #6130, opened 2017-07-29
Three letter months for appeals past activity. Tweaked spacing.

U-DON:
Resolves department-of-veterans-affairs/vets.gov-team#4106.

### DIFF
--- a/src/sass/disability-benefits.scss
+++ b/src/sass/disability-benefits.scss
@@ -364,20 +364,21 @@ a.claim-list-item {
     }
   }
 
+  .event-date {
+    flex: 0 0 10rem;
+    margin-right: 1rem;
+  }
+
   .event-header {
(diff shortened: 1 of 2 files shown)


src/sass/disability-benefits.scss
```diff
@@ -364,20 +364,21 @@ a.claim-list-item {
...
 
-    .date {
-      flex: 0 1 12rem;
-    }
-
     i.fa {
       &:first-child {
         color: $color-gray;
-        flex: 0 1 3.25rem;
+        flex: 0 0 3.25rem;;
```
patrickvinograd: double semi-colon - is that intentional?

patrickvinograd (approved):
LGTM - but check that double-semicolon in the .scss file, and fix before merging if needed.

PR #6133, opened 2017-07-31
Restructure expected vets-api response, and remove hard-coded code

annekainicUSDS:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3722

This PR does 2 things:
1. The structure of the response we get from vets-api has changed: instead of nesting the address info under `meta`, it's now under the regular `data: attributes` part of the response with everything else we get. Change the call for this address data to match the new response structure.
2. Remove some old hard-coded code that we put in place for a usability test.


robbiethegeek-usds (approved):
LGTM :shipit:

PR #6134, opened 2017-07-31
Clarify dependent disability notification

cehsu:
screen shot 2017-08-01 at 11 01 22 am
screen shot 2017-08-01 at 11 01 17 am
screen shot 2017-08-01 at 10 16 02 am

### DIFF
--- a/src/js/pensions/helpers.jsx
+++ b/src/js/pensions/helpers.jsx
@@ -210,7 +210,7 @@ export const expeditedProcessDescription = (
 export const dependentWarning = (
   <div className="usa-alert usa-alert-warning">
     <div className="usa-alert-body">
-      Your child won't qualify for the dependent credit unless they are in school or disabled.
+      Your child won't qualify as a dependent unless they're in school or disabled.
     </div>
   </div>
 );
--- a/src/js/pensions/helpers.jsx
+++ b/src/js/pensions/helpers.jsx
@@ -166,14 +166,18 @@ export const aidAttendanceEvidence = (
 );
 
 export const disabilityDocs = (
-  <div className="usa-alert usa-alert-warning no-background-image">
-    You'll need to provide all private medical records for your child's disability.
+  <div className="usa-alert usa-alert-warning">
+    <div className="usa-alert-body">
+      You'll need to provide all private medical records for your child's disability.
+    </div>
   </div>
 );
 
 export const schoolAttendanceWarning = (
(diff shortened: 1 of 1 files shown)


jbalboni:
@cehsu I don't think this is a change they wanted. I think they wanted the mention of the dependent credit on the other warning removed. Maybe @peggygannon or @goldenmeanie can say for sure?

peggygannon:
@cehsu @jbalboni  Please note "they are" should be "they're":
-- Your child won't qualify as a dependent unless they're in school or disabled.   

This box should appear on the household/dependents/children/information/ page if the user enters "no" for the questions is your child is seriously disabled? and is your child in school"

jbalboni:
Thanks @peggygannon. 

@cehsu I think you need to change `dependentWarning` instead of this one.

peggygannon:
@jbalboni @cehsu Does this warning appear in two different places/pages? If so, we will want to update the language in both.  I was only able to see it pop up on the /children/information/ page

jbalboni:
@peggygannon Not that I know of. There are two disability related warnings on the child info page:

>You'll need to provide all private medical records for your child's disability.

and 

>Your child won't qualify for the dependent credit unless they are in school or disabled.

The first is for when you choose Yes to the disabled question, the second is when you choose No for the disabled question and the school question.

My understanding is we want to update the second one to be

>Your child won't qualify as a dependent unless they are in school or disabled.

peggygannon:
@jbalboni Ok, thanks for clarifying.

cehsu:
Thanks @jbalboni @peggygannon for clarifying the changes involved. Please see screenshots for the changes above and here. I wonder if we want the style of these panels to be consistent? It looks like one has a background image and the other does not. Is this deliberate or simply due to the potentially different times at which these changes were made?
screen shot 2017-08-01 at 10 11 20 am
screen shot 2017-08-01 at 10 11 14 am

jbalboni:
@cehsu I think the background difference is due to one of them originally having a title. @stacyhlinn should they both have background images, or neither?

stacyhlinn:
@jbalboni @cehsu Yes they should be consistent. Let's stick with the background image. I'd make sure the rest of the yellow warning messages on the form all get the treatment.

jbalboni (approved):

cehsu:
Thanks @stacyhlinn changes made.

PR #6137, opened 2017-07-31
Letters error states

annekainicUSDS:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3977
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3969

This PR adds the basic logic around handling various HTTP status codes and showing the corresponding UI messages. The logic works and the specified messages are shown for each status code. However, I believe we need some design review of these messages and to make sure we're showing the right thing for each status code. I think this PR can still be reviewed and merged in the meantime so we get the basic logic in place; but we will need to make more permanent updates later once design has had a chance to take a look at it.

For now, here's what is rendered for each status code:

#### 503 (backend service unavailable) or 504 (backend service timeout)
screen shot 2017-08-02 at 2 00 58 pm

#### 403 (authentication failure/ forbidden)
403 error lettergeneration lettereligibilityerror screenshot

#### 422 (address error)
screen shot 2017-08-14 at 11 00 47 am

The other case where this warning box is shown is when the address is blank for some reason, but an error is not sent to us about the address. This scenario is unlikely, but handling it as a precaution:
screen shot 2017-08-14 at 10 59 06 am

#### 502 (letter eligibility error, some letters may be unavailable; either some letters will appear, and the warning will appear above the letters that are sent, or no letters will be sent and the user will just see the warning)
screen shot 2017-08-14 at 10 58 03 am

#### Any other status code (500, etc.)
screen shot 2017-08-01 at 12 51 07 pm


src/js/common/utils/error-messages.jsx
```diff
@@ -0,0 +1,28 @@
+/* eslint-disable camelcase */
+import React from 'react';
+
+export function systemDownMessage() {
```
jbalboni: You could export these as regular variables:

```jsx
export const systemDownMesssage = (
  <div>...</div>
);
```

src/js/letters/containers/Main.jsx
```diff
@@ -14,29 +16,45 @@ class Main extends React.Component {
...
-        <div>
-          <div className="usa-alert usa-alert-error" role="alert">
-            <div className="usa-alert-body">
-              <h4 className="usa-alert-heading">Letters Unavailable</h4>
-              <p className="usa-alert-text">
-                We weren't able to retrieve your VA letters. Please call
-                1-855-574-7286 between Monday-Friday 8:00 a.m. - 8:00 p.m. (ET).
-              </p>
+    switch (this.props.lettersAvailability) {
+      case 'available':
```
jbalboni: Does it make sense for these strings to be shared constants? They're used in a couple places, it looks like.

src/js/letters/actions/letters.js
```diff
@@ -1,16 +1,52 @@
...
+            // Backend authentication problem
+            return dispatch({ type: BACKEND_AUTHENTICATION_ERROR });
+          }
+          if (error.status === '422') {
+            // Something about the address is invalid, unable to process the request
+            return dispatch({ type: INVALID_ADDRESS_PROPERTY });
+          }
+        }
+        return Promise.reject(
+          new Error('letters getLetterList() received unexpected status code: `${response.status}`')
```
jbalboni: Also, should this be `error.status` instead of `response.status`? That's what's being checked above.

src/js/letters/containers/Main.jsx
```diff
@@ -14,29 +16,45 @@ class Main extends React.Component {
...
+        break;
+      case 'unavailable':
+        appContent = (
+          <div id="lettersUnavailable">
+            <div className="usa-alert usa-alert-error" role="alert">
+              <div className="usa-alert-body">
+                <h4 className="usa-alert-heading">Letters Unavailable</h4>
+                <p className="usa-alert-text">
+                  We weren't able to retrieve your VA letters. Please call
+                  1-855-574-7286 between Monday-Friday 8:00 a.m. - 8:00 p.m. (ET).
```
jbalboni: I think this needs to be a `tel` link, too (and maybe not have the `1-`).

src/js/letters/utils/helpers.jsx
```diff
@@ -15,6 +15,23 @@ export function apiRequest(resource, optionalSettings = {}, success, error) {
...
+export function invalidAddressProperty() {
+  return (
+    <div id="invalidAddress" className="usa-width-three-fourths">
+      <div className="usa-alert usa-alert-error">
+        <div className="usa-alert-body">
+          <h2 className="usa-alert-heading">Address unavailable</h2>
+          <p className="usa-alert-text">
+            We're encountering an error with your address information. This is not
+            required for your letters, but if you'd like to see the address we have
+            on file, or to update it, please visit <a href="/" target="_blank">this link</a>.
```
jbalboni: Assuming this is going to be filled in later?

jbalboni (approved):
Looks good overall. I had some minor questions/comments.

src/js/common/utils/error-messages.jsx
annekainicUSDS: ah yes good point! will fix

src/js/common/utils/error-messages.jsx
```diff
@@ -0,0 +1,28 @@
...
+  );
+}
+
+export function unableToFindRecordWarning() {
+  return (
+    <div id="recordNotFound">
+      <div className="small-12 columns">
+        <div className="react-container">
+          <h3>We weren't able to find your records.</h3>
+          <h4>Please call 855-574-7286 between Monday - Friday, 8:00 a.m. - 8:00 p.m. ET.</h4>
```
annekainicUSDS: good point! will add

cehsu: Should this include a `<tel>`?

src/js/letters/containers/LettersApp.jsx
```diff
@@ -12,7 +12,8 @@ function AppContent({ children, isDataAvailable }) {
   if (unregistered) {
     view = (
       <h4>
-        To view and download your VA letters, you need to verify your identity (or whatever).
+        We weren't able to find information about your VA letters.
+        If you think you should be able to access this information, please call the Vets.gov Help Desk at 855-574-7286 (TTY: 800-829-4833). We're here Monday–Friday, 8:00 a.m.–8:00 p.m. (ET).
```
cehsu: Same as above.

src/js/letters/containers/Main.jsx
cehsu: Same as above.

src/js/letters/reducers/index.js
```diff
@@ -3,6 +3,18 @@ import {
...
 } from '../utils/helpers.jsx';
+import {
+  BACKEND_AUTHENTICATION_ERROR,
+  BACKEND_SERVICE_ERROR,
+  INVALID_ADDRESS_PROPERTY,
+  GET_LETTERS_FAILURE,
+  GET_LETTERS_SUCCESS,
+  GET_BENEFIT_SUMMARY_OPTIONS_FAILURE,
+  GET_BENEFIT_SUMMARY_OPTIONS_SUCCESS,
+  // GET_LETTER_PDF_FAILURE,
```
cehsu: Are these comments part of a TODO? If so should it be noted?

src/js/letters/utils/helpers.jsx
cehsu: Will this be provided?

cehsu (approved):
Looks good to me.

src/js/letters/actions/letters.js
annekainicUSDS: Good question. I think you're right. Confirming with @raquelromano-gov because I copied this from some code she wrote for GIBS, but in testing this I don't think there is such a thing as `response.status`; it seems like this is a holdover from when we thought the status would be returned directly in the response as `response.status`, but then realized we're actually getting an errors array in the response, and the status is within that.

src/js/letters/reducers/index.js
annekainicUSDS: Good catch, thanks! Will remove.

src/js/letters/utils/helpers.jsx
annekainicUSDS: Yep, we're actually going to be changing this content anyway so it's fine for now.

PR #6138, opened 2017-07-31
Don't stop scrolling on number inputs

cvalarida:
Connect to https://github.com/department-of-veterans-affairs/vets-website/issues/4061

Turns out this _was_ really easy to do after all. Basically all the credit goes to @cehsu on this one! Thanks, Claire!

My previous fix made stopped scrolling when the cursor collided with the active number input. This keeps scrolling naturally.

I did a performance profile to see what would happen and...came up with nothing. Could be I don't know what I'm doing (true), or it could be that the performance impact is super minimal (probably true). I'd like to do some cross-platform testing, though, to make sure this will work as expected.

### DIFF
--- a/src/js/common/index.js
+++ b/src/js/common/index.js
@@ -10,7 +10,8 @@ require('./utils/sticky-action-box.js');
 // New navigation menu
 require('../legacy/mega-menu.js');
 
-// Disable scrolling on input[type="number"] because it was changing the value unexpectedly
+// Prevent some browsers from changing the value when scrolling while hovering
+//  over an input[type="number"] with focus.
 document.addEventListener('wheel', (event) => {
   if (event.target.type === 'number' && document.activeElement === event.target) {
     event.preventDefault();
--- a/src/js/common/index.js
+++ b/src/js/common/index.js
@@ -11,11 +11,9 @@ require('./utils/sticky-action-box.js');
 require('../legacy/mega-menu.js');
 
 // Disable scrolling on input[type="number"] because it was changing the value unexpectedly
-// NOTE: This also just stops the user from scrolling when hovering over a number
-//  input that has focus. This isn't ideal, but until a solution is found to that,
-//  this is what we've got.
-document.addEventListener('mousewheel', (event) => {
+document.addEventListener('wheel', (event) => {
   if (event.target.type === 'number' && document.activeElement === event.target) {
     event.preventDefault();
+    document.body.scrollTop += event.deltaY;
   }
 });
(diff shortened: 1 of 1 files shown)


src/js/common/index.js
```diff
@@ -10,12 +10,11 @@ require('./utils/sticky-action-box.js');
...
-// NOTE: This also just stops the user from scrolling when hovering over a number
-//  input that has focus. This isn't ideal, but until a solution is found to that,
-//  this is what we've got.
-document.addEventListener('mousewheel', (event) => {
+// Prevent some browsers from changing the value when scrolling while hovering
+//  over an input[type="number"] with focus.
+document.addEventListener('wheel', (event) => {
   if (event.target.type === 'number' && document.activeElement === event.target) {
     event.preventDefault();
+    document.body.scrollTop += event.deltaY;
```
cehsu: Wow! That is really elegant @cvalarida, we can share the credit 60/40 :)

cehsu (approved):
Looks good to me.

cehsu (approved):
LGTM. Thanks for testing this so thoroughly.

src/js/common/index.js
```diff
@@ -10,12 +10,12 @@ require('./utils/sticky-action-box.js');
 // New navigation menu
 require('../legacy/mega-menu.js');
 
-// Disable scrolling on input[type="number"] because it was changing the value unexpectedly
-// NOTE: This also just stops the user from scrolling when hovering over a number
-//  input that has focus. This isn't ideal, but until a solution is found to that,
-//  this is what we've got.
-document.addEventListener('mousewheel', (event) => {
+// Prevent some browsers from changing the value when scrolling while hovering
```
cehsu: Should this be updated to reflect these changes?

cvalarida: Mmm...Could. It's still true. Only some browsers did change it in the first place. Now it technically prevents all browsers, but...same difference, eh?

cehsu: Thanks @cvalarida, is the focus bit still true? I thought that's what the second boolean clause was ensuring before.

cvalarida: Yeah, `document.activeElement === event.target` makes sure we only `event.preventDefault()` when we're hovering over the element that has focus.

cehsu: Is the second boolean clause still required now that scrolling is no longer interrupted?

cehsu: Ah, got it, the problematic behavior only occurred while the input was focused, hence the specification.

PR #6139, opened 2017-08-01
Create currency definition and widget

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4116

It turns out that number inputs in some browsers in React will return a parsed value in some cases but not others. If you enter "10", you'll get the number 10, but if you enter "10.0" you'll get a string. That's annoying, so I created a currency input that always parses the numbers (and also adds some other currency configuration).

I've only updated HCA, but we need to do this on all forms.

screen shot 2017-08-01 at 10 50 01 am
screen shot 2017-08-01 at 10 49 54 am

### DIFF
--- a/src/js/common/schemaform/widgets/CurrencyWidget.jsx
+++ b/src/js/common/schemaform/widgets/CurrencyWidget.jsx
@@ -1,0 +1,43 @@
+import React from 'react';
+
+export default class CurrencyWidget extends React.Component {
+  onBlur = () => {
+    this.props.onBlur(this.props.id);
+  }
+
+  handleChange = (event) => {
+    const val = event.target.value;
+    if (val === '') {
+      this.props.onChange();
+    } else {
+      // Sometimes we get a string back (e.g. when you enter 10.00)
+      // so we need to parse it to be sure
+      const parsed = parseFloat(val);
+      if (!isNaN(parsed)) {
+        this.props.onChange(parsed);
+      } else {
+        this.props.onChange(val);
+      }
+    }
+  }
+
+  render() {
+    let val = this.props.value;
+    if (typeof val === 'number') {
+      val = val.toFixed(2);
+    }
+
+    return (
+      <input type="number"
+          step=".01"
+          id={this.props.id}
+          name={this.props.id}
+          disabled={this.props.disabled}
+          autoComplete={this.props.options.autocomplete || false}
+          className={this.props.options.widgetClassNames}
+          value={typeof val === 'undefined' ? '' : val}
+          onBlur={this.onBlur}
+          onChange={this.handleChange}/>
+    );
+  }
+}
--- a/src/js/common/schemaform/widgets/CurrencyWidget.jsx
+++ b/src/js/common/schemaform/widgets/CurrencyWidget.jsx
@@ -29,7 +29,6 @@ export default class CurrencyWidget extends React.Component {
 
     return (
       <input type="number"
-          step=".01"
           id={this.props.id}
           name={this.props.id}
           disabled={this.props.disabled}
--- a/src/js/hca/definitions/child.js
+++ b/src/js/hca/definitions/child.js
@@ -83,7 +83,7 @@ export const uiSchema = {
     'ui:title': 'If child is between 18 and 23 years of age, did child attend school during the last calendar year?',
     'ui:widget': 'yesNo'
   },
-  childEducationExpenses: currencyUI('Expenses paid by your dependent child for college, vocational rehabilitation, or training (e.g., tuition, books, materials)?'),
+  childEducationExpenses: currencyUI('Expenses paid by your dependent child for college, vocational rehabilitation, or training (e.g., tuition, books, materials)'),
   childCohabitedLastYear: {
     'ui:title': 'Did your child live with you last year?',
     'ui:widget': 'yesNo'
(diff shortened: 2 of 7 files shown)


src/js/common/schemaform/widgets/CurrencyWidget.jsx
```diff
@@ -0,0 +1,43 @@
...
+
+  render() {
+    let val = this.props.value;
+    if (typeof val === 'number') {
+      val = val.toFixed(2);
+    }
+
+    return (
+      <input type="number"
+          step=".01"
```
cvalarida: Hard to say whether anybody would actually use this, but if so, incrementing by pennies seems silly. Still, what other assumptions do we make, right?

:man_shrugging:

src/js/common/schemaform/widgets/CurrencyWidget.jsx
```diff
@@ -0,0 +1,43 @@
...
+    if (val === '') {
+      this.props.onChange();
+    } else {
+      // Sometimes we get a string back (e.g. when you enter 10.00)
+      // so we need to parse it to be sure
+      const parsed = parseFloat(val);
+      if (!isNaN(parsed)) {
+        this.props.onChange(parsed);
+      } else {
+        this.props.onChange(val);
```
cvalarida: Why call `onChange()` with the non-parsed value if it isn't a number(ish)? I'm assuming it has to do with passing an empty value, but I'm not sure how that would play out.

test/common/schemaform/review/CurrencyWidget.unit.spec.jsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import { expect } from 'chai';
+import SkinDeep from 'skin-deep';
+
+import CurrencyWidget from '../../../../src/js/common/schemaform/review/CurrencyWidget';
+
+describe('Schemaform review <CurrencyWidget>', () => {
+  it('should format ssn', () => {
```
cvalarida: I'm guessing that's not the description you want, right?

test/common/schemaform/widgets/CurrencyWidget.unit.spec.jsx
```diff
@@ -0,0 +1,47 @@
...
+import CurrencyWidget from '../../../../src/js/common/schemaform/widgets/CurrencyWidget';
+
+describe('Schemaform <CurrencyWidget>', () => {
+  it('should render', () => {
+    const tree = SkinDeep.shallowRender(
+      <CurrencyWidget
+          options={{}}
+          value={178}/>
+    );
+    expect(tree.subTree('input').props.value).to.equal('178.00');
```
cvalarida: Minor efficiency boost: Pull out `const input = tree.subTree('input');` and use that in each of these `expect`s.

cvalarida (approved):
Looks pretty good generally. I had some questions and suggestions, but I don't think they're blocking, so I'm approving this.

src/js/common/schemaform/widgets/CurrencyWidget.jsx
jbalboni: I don't want to erase user input, if it all possible. I'd rather have a user enter "aaa" and see a validation error than have the input act like you're not typing in it.

This branch isn't actually hit much, since most (maybe all?) browsers return an empty string if the value isn't a number. Chrome won't even let you type non-numbers.

src/js/common/schemaform/widgets/CurrencyWidget.jsx
jbalboni: I actually had this in there because I had read it was necessary for number inputs to treat decimals correctly, but that's not the case in Chrome. So, I'll do some more digging on that. I would rather use a step of 1.

test/common/schemaform/review/CurrencyWidget.unit.spec.jsx
jbalboni: Noooope.

src/js/common/schemaform/widgets/CurrencyWidget.jsx
cvalarida: Oh, I guess it would be a browser-specific functionality, but I was thinking since they _are_ number inputs, we can't even enter non-numbers. Good point, though.

src/js/common/schemaform/widgets/CurrencyWidget.jsx
cehsu: What if we set the increment value by the last number place changed (so if they change a 100 value, the increment step is 10)?

jbalboni: This doesn't seem necessary, so I'll remove it.

cvalarida: 😬 That'd be nifty, but I don't know that it makes sense from a UX perspective for one, but also that it'd be worth the effort.

src/js/hca/definitions/child.js
```diff
@@ -82,9 +83,7 @@ export const uiSchema = {
     'ui:title': 'If child is between 18 and 23 years of age, did child attend school during the last calendar year?',
     'ui:widget': 'yesNo'
   },
-  childEducationExpenses: {
-    'ui:title': 'Expenses paid by your dependent child for college, vocational rehabilitation, or training (e.g., tuition, books, materials)?'
-  },
+  childEducationExpenses: currencyUI('Expenses paid by your dependent child for college, vocational rehabilitation, or training (e.g., tuition, books, materials)?'),
```
cehsu: Do we need the question mark at the end?

cehsu (approved):
Looks good to me.

src/js/hca/definitions/child.js
jbalboni: That does look weird.

Issue #4217, opened 2016-12-05
Button should not break to two lines


goldenmeanie:
The `Get Started` button in the EDU Benefits form should not break to two lines. It’s behaving as if it were part of a side-by-side two-button layout. It should be as wide as its content requires to be on one line.

currently:
![](https://cloud.githubusercontent.com/assets/4106356/20844337/22df311c-b874-11e6-895b-b4bc3531b4df.png)

ayaleloehr:
This looks resolved here: https://www.vets.gov/education/apply-for-education-benefits/application/1990/introduction

@goldenmeanie can you close if so?

cvalarida:
I think I remember seeing it broken in IE on someone's screenshare...somewhere...but haven't gone about testing it thoroughly yet.

J-Quag:
This happens in IE11 when you make the window super small. Is there even anything we can do about that in IE?

PR #6147, opened 2017-08-01
1990 Start Button Fix

cvalarida:
Connects to https://github.com/department-of-veterans-affairs/vets-website/issues/4217

I made some style changes to the intro page to make it look better.

Before:
image

After:
image

Full-width button on a small screen:
image

### DIFF
--- a/src/js/common/components/NavButtons.jsx
+++ b/src/js/common/components/NavButtons.jsx
@@ -164,7 +164,7 @@ export default class NavButtons extends React.Component {
       </div>);
     } else if (path.endsWith('submit-message')) {
       buttons = (
-        <div className="row form-progress-buttons">
+        <div className="row">
           <div className="small-12 medium-5 columns">
             <a href="/">
               <button className="usa-button-primary">Back to Main Page</button>
--- a/src/js/common/components/NavButtons.jsx
+++ b/src/js/common/components/NavButtons.jsx
@@ -165,7 +165,7 @@ export default class NavButtons extends React.Component {
     } else if (path.endsWith('submit-message')) {
       buttons = (
         <div className="row form-progress-buttons">
-          <div className="small-6 medium-5 columns">
+          <div className="small-12 medium-5 columns">
             <a href="/">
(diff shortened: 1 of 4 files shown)


src/js/common/components/NavButtons.jsx
```diff
@@ -165,7 +165,7 @@ export default class NavButtons extends React.Component {
     } else if (path.endsWith('submit-message')) {
       buttons = (
         <div className="row form-progress-buttons">
-          <div className="small-6 medium-5 columns">
+          <div className="small-12 medium-5 columns">
```
cvalarida: I also changed this one because I didn't want it to be tiny on a small screen when it's the only button. Same problem as the "Get Started" button.

src/js/common/components/NavButtons.jsx
```diff
@@ -174,8 +174,8 @@ export default class NavButtons extends React.Component {
       );
     } else if (path.endsWith('introduction')) {
       buttons = (
-        <div className="row form-progress-buttons">
```
cvalarida: `form-progress-buttons` gave it `max-width: 47rem`, but the `small` breakpoint > 47rem, so there was this weird gap where the button was really long, but didn't take up the whole width.

src/sass/modules/_m-schemaform.scss
```diff
@@ -161,12 +161,6 @@ label {
   margin-bottom: 7rem;
 }
 
-// Align the intro content of the 22-1990 to match the title
-// Necessary for the 22-1990 and HCA since they don't use schemaform
-.intro-content {
```
cvalarida: The 1990 intro page was the only one to use this...and it didn't need it any more.

jbalboni:
@cvalarida Should we fix the margin on the need help content, too?

jbalboni (approved):
Looks like you need to update an e2e test, and I made one comment, but otherwise it looks good.

cehsu (approved):
Looks good to me.

PR #6156, opened 2017-08-02
Clean up Sentry error grouping for form submissions

jbalboni:
Related to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3168

The submission errors are not grouped well in Sentry because of the variety of stacktraces you can get, so this adds [fingerprinting](https://docs.sentry.io/learn/rollups/#customize-grouping-with-fingerprints) to group them more specifically (by form and error message).

I also found that we get a lot of `undefined` error messages in Sentry because not all unhandled promise rejections have a reason. So, this adds a default message and also sends the event as extra info so we can figure out if there are other properties we can use to organize the errors.

### DIFF
--- a/src/js/common/schemaform/actions.js
+++ b/src/js/common/schemaform/actions.js
@@ -57,7 +57,8 @@ export function submitForm(formConfig, form) {
     Raven.captureException(error, {
       extra: {
         clientError,
-        statusText: error.statusText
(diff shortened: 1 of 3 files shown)


cehsu (approved):
Looks good to me.

U-DON (approved):
LGTM.

PR #6157, opened 2017-08-02
[WIP] remove current page breadcrumb

phiden:
Per https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2990

### DIFF
--- a/content/includes/breadcrumbs.html
+++ b/content/includes/breadcrumbs.html
@@ -7,9 +7,9 @@
     <li><a href="/">Home</a></li>
       {% assign current_page = breadcrumb_path | last %}
       {% for crumb in breadcrumb_path %}
-        {% if current_page.name == crumb.name %}
+        {% if current_page.name != crumb.name %}
           <!-- removed to remove 'active' breadcrumb -->
-        {% else %}
+        
           {% comment %}
             Each breadcrumb item has a child object. One of the properties of
             that object is `file` which loads the front matter metadata from
--- a/content/includes/breadcrumbs.html
+++ b/content/includes/breadcrumbs.html
@@ -8,13 +8,7 @@
       {% assign current_page = breadcrumb_path | last %}
       {% for crumb in breadcrumb_path %}
         {% if current_page.name == crumb.name %}
-          <li class="active">
-            {% if title %}
-              {{ title }}
-            {% else %}
-              {{ crumb.name | replace: '-', ' ' | capitalize }}
-            {% endif %}
-          </li>
+          <!-- removed to remove 'active' breadcrumb -->
         {% else %}
           {% comment %}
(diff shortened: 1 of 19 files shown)


content/includes/breadcrumbs.html
```diff
@@ -8,13 +8,7 @@
       {% assign current_page = breadcrumb_path | last %}
       {% for crumb in breadcrumb_path %}
         {% if current_page.name == crumb.name %}
```
robertfairhead: Since we only care about the `else` here, maybe switch this to `{% if current_page.name != crumb.name %}` to simplify

src/js/gi/components/heading/Breadcrumbs.jsx
```diff
@@ -30,15 +24,12 @@ class Breadcrumbs extends React.Component {
       crumbs.push(<span key="profile"><strong>{this.props.profileName || 'Profile'}</strong></span>);
```
robertfairhead: Don't you need to remove this too? When showing the profile, the profile is currently in prod the "active" crumb:

screen shot 2017-08-04 at 4 33 57 pm

content/pages/burials-and-memorials/application/530.md
```diff
@@ -10,7 +10,6 @@ layout: page-react.html
       <li><a href="/burials-and-memorials"> Burials and Memorials </a></li>
       <li><a href="/burials-and-memorials/survivor-and-dependent-benefits"> Survivors and Dependents </a></li>
       <li><a href="/burials-and-memorials/survivor-and-dependent-benefits/burial-costs/">Handling Burial Costs (Allowances)</a></li>
-      <li class="active">Apply for Burial Benefits</li>
```
robertfairhead: Just a question, is there any reason why we can't use the `breadcrumbs.html` partial instead of doing this manually?

If not, then now may be a good time to make the switch since we're touching all those files anyway.

robertfairhead (changes requested):
the `breadcrumbs.html` and general app start page comments are basically take it or leave it.

But i'm fairly sure the GIBCT one needs to get removed. if not, let me know and i'll approve.

src/js/gi/components/heading/Breadcrumbs.jsx
phiden: Good catch!

content/pages/burials-and-memorials/application/530.md
phiden: Turns out that the breadcrumbs partial renders different nav than the manually entered nav, so I'm going to leave it as-is for now.

phiden:
@robertfairhead changes made mostly as requested -- see comments above.

robertfairhead (approved):

PR #6158, opened 2017-08-02
Text updates for FDC page on pension form

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4153

screen shot 2017-08-02 at 4 22 26 pm

### DIFF
--- a/src/js/pensions/config/form.js
+++ b/src/js/pensions/config/form.js
@@ -1648,12 +1648,12 @@ const formConfig = {
           }
         },
         expedited: {
-          title: 'Expedited process',
+          title: 'Fully developed claim program',
           path: 'additional-information/fdc',
           uiSchema: {
             'ui:description': expeditedProcessDescription,
             noRapidProcessing: {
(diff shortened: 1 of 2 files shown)


src/js/pensions/config/form.js
```diff
@@ -1648,12 +1648,12 @@ const formConfig = {
...
         },
         expedited: {
-          title: 'Expedited process',
+          title: 'Fully developed claim program',
           path: 'additional-information/fdc',
           uiSchema: {
             'ui:description': expeditedProcessDescription,
             noRapidProcessing: {
-              'ui:title': 'Do you want to use the expedited process?',
+              'ui:title': 'Do you want to apply using the Fully Developed Claim program?',
```
cehsu: One minor thought is whether "via" or "through" might be clearer than "using"?

cehsu (approved):
Looks good to me. Had a minor question but should not block this change.

PR #6159, opened 2017-08-02
Turn on SiP again

jbalboni:
This turns on SiP for pensions, burials, and HCA.


cehsu (approved):
This looks good to me. Are there any tests that need to be updated?

jbalboni:
@cehsu Good catch, I missed removing the flag on the e2e test.

PR #6160, opened 2017-08-02
Updates for pension A&A page

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4185

screen shot 2017-08-02 at 5 35 28 pm
screen shot 2017-08-02 at 5 34 30 pm
screen shot 2017-08-02 at 10 18 39 pm

### DIFF
--- a/src/js/pensions/helpers.jsx
+++ b/src/js/pensions/helpers.jsx
@@ -81,14 +81,17 @@ export function fileHelp({ formData }) {
 
   return (
     <div>
-      <p>Please upload all documentation to support your claim. This includes:</p>
+      <p>Please upload all documentation to support your claim. {(hasSchoolChild || hasDisabledChild) && 'This includes:'}</p>
       <ul>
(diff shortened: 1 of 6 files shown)


src/js/pensions/helpers.jsx
```diff
@@ -148,7 +147,7 @@ export const uploadMessage = (
 
 export const aidAttendanceEvidence = (
   <div className="usa-alert usa-alert-info no-background-image">
-    <strong>If you’re claiming non-service-connected pension benefits with Aid and Attendance</strong>, the evidence must show that you:
+    <strong>If you’re claiming non-service-connected pension with Aid and Attendance benefits</strong>, your supporting documents must show that you:
```
cehsu: Should the opening <strong> tag be moved to just before non-service to align with the style on line 159?

cehsu (approved):
Looks good to me. I had a weakly held assumption about a strong tag. I also understand there might be another change to split the `aidAttendanceEvidence` panel into two.

src/js/pensions/helpers.jsx
jbalboni: They were different before, so I'm not sure.

PR #6162, opened 2017-08-03
Don't error out when a form page isn't active

jbalboni:
Sentry errors: http://sentry.vetsgov-internal/vets-gov/website-production/?query=is%3Aunresolved+chapterKey

If you open a page that has a `depends` property that isn't true by default, FormNav is erroring out. We should display the page, even if the nav info might be slightly off.

You can see this by going to https://www.vets.gov/health-care/apply/application/household-information/annual-income


cehsu (approved):
Looks good to me.

PR #6165, opened 2017-08-03
Add message about upleveling LOA1 accounts after saving a form

cehsu:
Connect to department-of-veterans-affairs/vets.gov-team#4196

- [x] Revise existing content
- [x] Add new loa uplevel message
- [x] Insert link into message
- [x] Provide screenshots
- [x] Write/update tests
- [x] Add expiration time (from team-wide testing session)

### Verified
screen shot 2017-08-07 at 4 04 25 pm

### Not verified
screen shot 2017-08-07 at 12 55 18 pm


src/js/common/schemaform/FormSaved.jsx
```diff
@@ -41,10 +41,15 @@ class FormSaved extends React.Component {
...
+            <strong>Your application has been saved.</strong><br/>
             {!!lastSavedDate && <p>Last saved on {moment(lastSavedDate).format('M/D/YYYY [at] h:mma')}.</p>}
             {success}
-            If you're on a public computer, please sign out before you leave to ensure your data is secure.
+            If you're logged in through a public computer, please sign out of your account before you log off to keep your information secure.
+          </div>
+        </div>
+        <div className="usa-alert usa-alert-warning">
+          <div className="usa-alert-body">
+            <strong>Please verify your identity.</strong> We want to provide you with the highest level of security to keep your information safe. Please go to your account (link here) and verify your identity.
```
cehsu: @stacyhlinn @peggygannon is it alright that I added a "with" in "We want to provide you with the highest level..."?

cehsu: Also, it looks like there are two options for the link here, either:
-We can have a link that starts the verification process directly, we don't have to send them to the profile page. Either way works, just wanted to point that out in case that was preferred. (As proposed by @jbalboni [here](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4196#issuecomment-320063431))
- or as already phrased above
Which option seems preferable? And what would be the actual link to provide?

stacyhlinn:
LGTM!

peggygannon:
@cehsu @stacyhlinn Could we make the following tweaks? 
--add a space btw. the time and p.m., and add periods in a.m. p.m.
so 4:02 p.m. 12:53 p.m.
--lowercase with in the blue button: Continue with Application

cehsu:
Thanks @stacyhlinn @peggygannon for your comments. Kindly refer to the revised screenshot below. Also, could you please confirm that the meridiam should be consistently displayed in this way? If so, I will remove trailing periods from all instances that I find (so there are no instances of "12:00 p.m..", for example).
screen shot 2017-08-07 at 9 55 00 pm

src/js/common/schemaform/FormSaved.jsx
```diff
@@ -7,9 +7,19 @@ import Scroll from 'react-scroll';
 
 import { focusElement } from '../utils/helpers';
 import { fetchInProgressForm, removeInProgressForm } from './save-load-actions';
+import { handleVerify } from '../../common/helpers/login-helpers.js';
 
 import FormStartControls from './FormStartControls';
 
+moment.updateLocale('en', {
```
cehsu: @cvalarida @jbalboni thanks for any insights on whether there is a better place to locate this.

src/js/common/schemaform/FormStartControls.jsx
```diff
@@ -53,8 +53,8 @@ class FormStartControls extends React.Component {
         <div>
           <ProgressButton
               onButtonClick={this.handleLoadForm}
-              buttonText="Resume previous application"
-              buttonClass="usa-button-primary"/>
+              buttonText="Continue with Application"
+              buttonClass="usa-button-primary title-case"/>
```
cehsu: Is `title-case` an appropriate name for this? It is actually only setting `text-transform` to `none`, but for situations intended for title cased text.

test/user-profile/components/AuthApplicationSection.unit.spec.jsx
```diff
@@ -1,36 +1,26 @@
...
-  // TODO: figure out how to get spy to be called
-  /*
-  it('should call handler when verify link is clicked', () => {
-    const verifySpy = sinon.spy(handleVerify);
-    const section = ReactTestUtils.renderIntoDocument(
-      <AuthApplicationSection {...props }/>
-    );
-    ReactTestUtils.Simulate.click(
-      findDOMNode(section).querySelector("a[href='#']"));
-    expect(verifySpy.called.with(props.verifyUrl)).to.be.true;
```
cehsu: `handleVerify` is already being tested in `test/common/helpers/login-helpers.unit.spec.jsx`, so this is only testing whether conditionally rendered components are behaving as expected.

src/js/user-profile/components/AuthApplicationSection.jsx
```diff
@@ -26,7 +35,7 @@ class AuthApplicationSection extends React.Component {
           <p><span className="label">Your account will allow you to:</span></p>
           <p><a href="/health-care/apply">Apply for health care</a></p>
           <p><a href="/education/apply-for-education-benefits">Apply for education benefits</a></p>
-          <p><span className="label">You need to <a href="#" onClick={() => handleVerify(this.props.verifyUrl)}>verify your account</a> in order to:</span></p>
+          <p><span className="label">You need to <a href="#" onClick={this.handleVerify}>verify your account</a> in order to:</span></p>
```
cehsu: This seems more idiomatic but please let me know if needs to be reverted.

peggygannon:
@ceshu. LGTM. thank you!

src/js/common/schemaform/FormSaved.jsx
```diff
@@ -31,22 +46,32 @@ class FormSaved extends React.Component {
...
 
+  handleVerify() {
+    handleVerify(this.props.user.login.verifyUrl);
+  }
+
   render() {
     const { profile } = this.props.user;
     const lastSavedDate = this.props.lastSavedDate;
     const prefillAvailable = !!(profile && profile.prefillsAvailable.includes(this.props.formId));
+    const notVerified = profile.accountType !== 3;
```
jbalboni: I think a comment or a constant for this would be helpful. It's not clear what 3 means without context.

cehsu:
Thanks @jbalboni looping @peggygannon in as she actually requested the meridiem and title case change, which is not currently in place, rather uppercase is the standard behavior (which also capitalizes articles).

src/js/common/schemaform/FormSaved.jsx
cehsu: Thanks @jbalboni, should all meridiems be made uppercase? I think there are at least a few lowercase instances.

jbalboni: I'm not exactly sure, it does look like we have both in places. I don't see any uses of periods, though, so I'm hoping we can avoid making this `updateLocale` call and just use the moment formatting (which can do upper or lower case).

peggygannon:
@cehsu @jbalboni @stacyhlinn Here's some background on my suggested edits:

--Vets.gov style is to lowercase prepositions and articles in button text. I think we should be consistent across the site with how we treat content in button text. 

--for the a.m. p.m., this is also Vets.gov style to have these lowercase and with periods. It is more user-friendly and typographically easier to read. 

 I will create a task for myself to look at all button content/style for consistency, so we're not having to make a one-off decision each time this comes up.

src/js/common/schemaform/FormStartControls.jsx
jbalboni: Oh, I see, I think I misunderstood this initially. We're still title casing, we just don't want to do it the way text-transform does. I think that's ok, then, but I think we should choose something other than `title-case`. `no-text-transform` or something.

src/js/common/schemaform/FormSaved.jsx
jbalboni: @cehsu We should put this somewhere else, I think. It's going to affect all date formatting in form bundles, so it shouldn't be in here. Maybe FormApp? Not really sure.

cehsu:
Thanks @jbalboni @peggygannon, changes made. Kindly refer to screen shots for additional visual changes below:

#### Add periods to meridiem
screen shot 2017-08-08 at 10 32 27 am


#### Make all meridiems lowercase
screen shot 2017-08-08 at 10 48 29 am

src/js/hca/helpers.jsx
```diff
@@ -161,5 +161,5 @@ export const disclosureWarning = (
 );
 
 export const resumeMessage = (
-  <p>To resume the application when you come back, either bookmark this page or come back to the <a href="/health-care/apply/application">health care application page</a>.</p>
+  <p>To continue with your saved application when you come back, go to <a href="/health-care/apply/application" target="_blank">Health Care Application Process</a> or add this page to your web browser's bookmark list.</p>
```
cehsu: Thanks for the pointer @jbalboni, change made.

src/js/common/schemaform/FormSaved.jsx
cehsu: Thanks @jbalboni, change made.

src/js/common/schemaform/FormStartControls.jsx
cehsu: Thanks @jbalboni, change made.

src/js/common/schemaform/FormSaved.jsx
cehsu: Thanks @jbalboni, constant and comment added.

src/js/user-profile/components/AuthApplicationSection.jsx
```diff
@@ -4,6 +4,15 @@ import React from 'react';
 import { handleVerify } from '../../common/helpers/login-helpers.js';
 
 class AuthApplicationSection extends React.Component {
+  constructor(props) {
+    super(props);
+    this.handleVerify = this.handleVerify.bind(this);
+  }
+
+  handleVerify() {
```
cehsu: Thanks @jbalboni, changes made (now named `verifyUser`).

test/user-profile/components/AuthApplicationSection.unit.spec.jsx
cehsu: Thanks for your insights @annekainicUSDS @raquelromano-gov on whether the decision to remove these comments (and also add new tests) makes sense.

src/js/common/schemaform/FormSaved.jsx
```diff
@@ -31,22 +32,33 @@ class FormSaved extends React.Component {
...
     const { success } = this.props.route.formConfig.savedFormMessages || {};
 
     return (
       <div>
         <div className="usa-alert usa-alert-info">
           <div className="usa-alert-body">
-            <strong>Your application has been saved!</strong><br/>
-            {!!lastSavedDate && <p>Last saved on {moment(lastSavedDate).format('M/D/YYYY [at] h:mma')}.</p>}
+            <strong>Your application has been saved.</strong><br/>
+            {!!lastSavedDate && <p>Last saved on {moment(lastSavedDate).format('M/D/YYYY [at] h:mm a')}</p>}
```
jbalboni: Was the period intentionally removed?

src/js/common/schemaform/FormSaved.jsx
```diff
@@ -31,22 +32,33 @@ class FormSaved extends React.Component {
...
             {success}
-            If you're on a public computer, please sign out before you leave to ensure your data is secure.
+            If you're logged in through a public computer, please sign out of your account before you log off to keep your information secure.
           </div>
         </div>
+        {notVerified && <div className="usa-alert usa-alert-warning">
+          <div className="usa-alert-body">
+            We want to keep your information safe with the highest level of security. Please <a href="#" onClick={this.verifyUser}>verify your identity</a>.
```
jbalboni: I've been trying to make these buttons, since they're actions and not navigation. There's a class called `va-button-link` that styles buttons like links, so it would be: `<button className="va-button-link" onClick={this.verifyUser}>verify your identity</button>`.

src/sass/modules/_m-button.scss
```diff
@@ -174,4 +174,8 @@ button.short {
   margin: 0;
   text-indent: -9999rem;
   width: 1rem;
-}
\ No newline at end of file
+}
+
+.no-text-transform {
```
jbalboni: We could put this in `src/sass/base/_b-utils.scss`. There's a similar `no-background-image` class in there.

jbalboni (approved):
A couple other minor comments, but looks good otherwise. I think the only outstanding is if it's ok for us to delete the commented out test.

test/user-profile/components/AuthApplicationSection.unit.spec.jsx
annekainicUSDS: Hm, looking at this a bit, I assume what we were trying to do here is test that `handleVerify` actually gets called when this link is clicked. While this would be a nice to have, seems like it was difficult to figure out how to get the spy actually working. If it's not too much trouble it seems useful to keep the commented out code in for now. We should, however, come up with a plan for how to revisit these kinds of things, there are lots of commented out tests right now :/ curious to hear what @raquelromano-gov thinks too though!

cehsu: Thanks @annekainicUSDS @jbalboni @raquelromano-gov for indicating that we should also check that this event is also getting triggered properly here, I've replicated the handleVerify spy logic from `test/common/helpers/login-helpers.unit.spec.jsx`. Please let me know if another approach might make more sense.

cehsu:
Thanks for your guidance so far @peggygannon @stacyhlinn, It looks like the expiration date/time for the saved form should also be included (per [this comment](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4256#issuecomment-321041284)). Are there some good examples for how to present this?

stacyhlinn:
@cehsu on the entry page this is how we present it: 
screen shot 2017-08-08 at 3 11 44 pm
Can we do something similar here, adding in red text how many days til it expires after the save date/time?

src/js/common/schemaform/FormSaved.jsx
cehsu: Yes, in order to avoid a situation with "12:00 p.m..". Thanks for letting me know if this should be reverted.

src/sass/modules/_m-button.scss
cehsu: Great tip @jbalboni, relocated to `src/sass/base/_b-utils.scss`.

cehsu:
Thanks @stacyhlinn, kind refer to the updated screenshot below:
screen shot 2017-08-08 at 4 18 27 pm

peggygannon:
@cehsu @stacyhlinn Here is a suggestion for the expiration message. Please let me know if this will work. 

Last saved on 8/8/2017 at 4:29 p.m. Your saved application will expire in 59 days.

peggygannon:
@cehsu @stacyhlinn Can we also update that other message? the one that appears on the entry page?

cehsu:
Thanks @peggygannon, changes made. Kindly see the updated screenshots below:

screen shot 2017-08-09 at 10 02 20 am

screen shot 2017-08-09 at 10 02 39 am

peggygannon:
@cehsu Thanks.. Can we also update this  second line on the entry page? 

Please complete the 10-10EZ form to apply for health care benefits.

cehsu:
Thanks @peggygannon, changes made. Kindly refer to the updated screenshot below:
screen shot 2017-08-09 at 10 44 17 am

peggygannon:
@cehsu LGTM!

PR #6168, opened 2017-08-04
Use text type for currency input in rjsf forms

jbalboni:
We got some reports from the call center about users not being able to enter income info in HCA. It appears to stem from parsing issues when using `type="number"` on `input` fields. Using that type means that browsers will attempt to parse the value and return an empty string in the change event if it is not valid. This causes all sorts of issues, especially around comma usage (they are sometimes used as decimal separators).

So this sets the input back to text and does more user friendly parsing of numbers that are entered. There's no reliable way to make the numeric keyboard show up for text inputs, so we do lose that (see https://www.filamentgroup.com/lab/type-number.html).


cehsu (approved):
This looks good to me.

src/js/common/schemaform/widgets/CurrencyWidget.jsx
```diff
@@ -1,40 +1,54 @@
...
       this.props.onChange();
     } else {
-      // Sometimes we get a string back (e.g. when you enter 10.00)
-      // so we need to parse it to be sure
-      const parsed = parseFloat(val);
-      if (!isNaN(parsed)) {
-        this.props.onChange(parsed);
-      } else {
+      // Needs to look like a currency
+      if (!val.match(/^[\$0-9\.,]*$/)) {
```
cvalarida: I don't know if it's worth it, but we could be a little more strict and use
```js
/^[0-9,]*(\.\d{1,2})?$/
```
This prevents multiple `.`s, commas after the `.`, trailing `.`s with no digits after, and limits the decimal to hundredths.

cvalarida (approved):

src/js/common/schemaform/widgets/CurrencyWidget.jsx
jbalboni: Works for me. I would say that we want to prevent anything that looks like an error or will parse in a surprising way. 

I added `$` to it, since I wanted to allow people to put that in there (accidentally). It's something that's easy to strip out.

cehsu: Nice!

PR #6170, opened 2017-08-04
Letters letter download loading

annekainicUSDS:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3724

This PR adds logic around changing the content of the "Download Letter" button depending on the state of the download. The 4 possible states are:
1. `pending`: not yet attempted to be downloaded (what the state is initialized at)
screen shot 2017-08-04 at 4 02 36 pm

2. `downloading`: button has been clicked and downloading has started
screen shot 2017-08-04 at 4 01 47 pm

3. `success`: letter has been successfully downloaded
screen shot 2017-08-04 at 3 57 39 pm

4. `failed`: letter has failed to download
screen shot 2017-08-04 at 3 57 47 pm

This PR may need some design/content/product review. While I was passing this design pattern off of what we do for form submit buttons, there is a difference between how these buttons behave. While we don't want to enable multiple attempts to submit a form, do we want to enable multiple attempts to download a letter? The questions I think need answering are:
1. During which states should the button be disabled? If it fails to download should you be allowed to try again? If it successfully downloads should you be allowed to try again?
2. What should the content and look of the button be during each state?

### DIFF
--- a/src/js/letters/components/DownloadLetterLink.jsx
+++ b/src/js/letters/components/DownloadLetterLink.jsx
@@ -27,7 +27,7 @@ export class DownloadLetterLink extends React.Component {
     let buttonClasses;
     let buttonText;
     let buttonDisabled;
-    let unavailableMessage;
+    let message;
     switch (this.props.downloadStatus) {
       case 'downloading':
         buttonClasses = 'usa-button-disabled';
@@ -36,14 +36,24 @@ export class DownloadLetterLink extends React.Component {
         break;
       case 'success':
         buttonClasses = 'usa-button-primary va-button-primary';
-        buttonText = 'Download Again';
+        buttonText = 'Download Letter';
         buttonDisabled = false;
+        message = (
+          <div className="usa-alert usa-alert-success" role="alert">
+            <div className="usa-alert-body">
+              <h2 className="usa-alert-heading">Your letter has successfully downloaded.</h2>
+              <p className="usa-alert-text">
+                If you want to download your letter again, please press the button below.
+              </p>
+            </div>
+          </div>
+        );
         break;
       case 'failure':
         buttonClasses = 'usa-button-primary va-button-primary';
         buttonText = 'Retry Download';
         buttonDisabled = false;
-        unavailableMessage = (
+        message = (
           <div className="usa-alert usa-alert-error" role="alert">
             <div className="usa-alert-body">
               <h2 className="usa-alert-heading">Your letter didn't download.</h2>
@@ -71,7 +81,7 @@ export class DownloadLetterLink extends React.Component {
               transitionAppearTimeout={700}
               transitionEnterTimeout={700}
               transitionLeave={false}>
-            {unavailableMessage}
+            {message}
           </ReactCSSTransitionGroup>
         </div>
         <div className="download-button">
--- a/src/js/letters/components/DownloadLetterLink.jsx
+++ b/src/js/letters/components/DownloadLetterLink.jsx
@@ -1,7 +1,6 @@
 import React from 'react';
 import { connect } from 'react-redux';
 import PropTypes from 'prop-types';
-import { Link } from 'react-router';
 
 import { getLetterPdf } from '../actions/letters';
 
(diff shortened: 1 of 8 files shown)


goldenmeanie:
1, 2, and 4 make sense to me.

3 seems like it could use some other success message or say "download again"

rroueche:
for 3, I like Alex's suggestion of "Download again"

src/sass/letters.scss
```diff
@@ -3,6 +3,8 @@
 @import "modules/m-progress-bar";
 @import "modules/m-schemaform";
 
+$color-gray-lighter: #d6d7d9;
```
jbalboni: Is this being overridden? That variable should already exist.

src/sass/letters.scss
```diff
@@ -35,6 +37,15 @@
     padding-bottom: 1.5rem;
   }
 
+  .download-button {
+    margin: 1.5em 0 1em;
+
+    button:disabled {
```
jbalboni: Maybe `> button:disabled`? I think it's good practice to make sure nested styles like this are as narrowly specified as possible.

jbalboni (approved):
LGTM. Only a couple really minor questions.

src/js/letters/components/DownloadLetterLink.jsx
```diff
@@ -24,11 +23,39 @@ export class DownloadLetterLink extends React.Component {
...
+        buttonDisabled = true;
+        break;
+      case 'success':
+        buttonClasses = 'usa-button-primary va-button-primary';
+        buttonText = 'Downloaded';
+        buttonDisabled = false;
+        break;
+      case 'failure':
+        buttonClasses = 'usa-button-disabled';
+        buttonText = 'Failed to download';
```
jbalboni: I think button text is supposed to be title case?

src/sass/letters.scss
cehsu: Is this for performance? Do we anticipate nesting for a button attribute?

cehsu (approved):
Looks good to me.

goldenmeanie:
I agree with @rroueche about 4. Error messaging and a more decisive button. If there is precedent for the modal error, then cool. If that's new, then I'd suggest and inline error message above the "Try downloading again?" button

annekainicUSDS:
Sounds good, I can add an error message for that case and change the button language

annekainicUSDS:
@goldenmeanie when you have a chance to review, here's a new screenshot of the failed download view.

screen shot 2017-08-11 at 4 46 15 pm

src/sass/letters.scss
jbalboni: In this specific case, it may not matter much. But nesting selectors like this in general will lead to a codebase that's really hard to modify and understand (it's already causing problems in ours, like with the accordion styles). It's better to avoid it if at all possible, especially when using element names (vs classes).

rroueche:
@peggygannon new letters buttons and error messages for your review :) could you give your opinion on the text, and suggest updates for 1-4 if they're not PL enough or not in line w/our text styling?

1. `pending`: not yet attempted to be downloaded (what the state is initialized at)
image

2. `downloading`: button has been clicked and downloading has started


image

You can just put your suggested text in a comment in this PR.

goldenmeanie:
I don't think the failure state (4) needs the blue (conditional group) bar on the left

peggygannon:
@rroueche 
Screen 1:  button text looks good. 

Screen 2: button text looks good

Screen 3; button text TK?
Is this button meant to have users download another letter? or download the same letter again?  or just tell users that the download was successful?

Screen 4:  **Your letter didn't download.**
Your letter isn't available at this time. If you need help with accessing your letter, please call 855-574-7286, Monday-Friday, 8 a.m. - 8 p.m. (ET).

Button text; looks good

annekainicUSDS:
@peggygannon, Screen 3 is meant to show the case where the letter was successfully downloaded. The current button language of "Download again" is a little confusing for this case, even though they are able to download it again. Maybe we can show a message above the button, like we do for Screen 4 where there's an error, but show a success message instead explaining that it downloaded and that you can download it again if you want? And then just keep the button language as "Download Letter"?

This is what it looks like now:
screen shot 2017-08-15 at 10 35 24 am

rroueche:
@peggygannon thank you!!

@annekainicUSDS , if you have worked in Peggy's PL updates for 4, then we're ready to merge. 3 is good as is!

rroueche:
@annekainicUSDS actually before merging, could you show us what the success message for 3. would look like? @peggygannon could you provide what you believe the message text should be?

peggygannon:
@annekainicUSDS @rroueche 

Ok, here is the success message and updated button text. Let me know what you think. 

Do we need a message if users want to download a different letter? What is the user's experience to download a different letter. I think we might want to provide some text on how they can download a different letter. 

Screen 3: **Your letter has successfully downloaded.**
If you want to download your letter again, please press the button below.
Button text: Download Letter

Since we're using the "Download Letter" button for screen 3, then I suggest we update screen 4 button language to "Download Again"

annekainicUSDS:
@peggygannon @goldenmeanie @rroueche this is what the success message looks like now:
screen shot 2017-08-15 at 1 35 35 pm

peggygannon:
@annekainicUSDS @rroueche @goldenmeanie thank you! LGTM. 

What are your thoughts about updating the button text for when the download fails from "Retry Download" to "Download Again"?

annekainicUSDS:
Personally I prefer "Retry Download", because it's clear that something didn't work the first time, whereas "Download Again" suggests that it has already been downloaded, when it hasn't.

rroueche:
@peggygannon @goldenmeanie @annekainicUSDS I prefer "Retry Download" bc "retry" better implies it didn't work the first time than "again".

goldenmeanie:
Agree with "Retry" reenforcing the message.

I wonder if that success message could be simplified to just the headline if the button went back to "Download Again"? Again with button reenforcing alert message.

annekainicUSDS:
I'm going to merge this for now so we can make sure this is in staging before tomorrow. If we want to make additional changes to the content I can open a new PR for that. Thanks all for the feedback!

PR #6174, opened 2017-08-07
Use currency definition on pension form

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4206

Updates pension form to use currency definition. Also changes currency widget to use the right regex method.

Only notable visual change (0 is now 0.00):

screen shot 2017-08-07 at 1 07 56 pm

### DIFF
--- a/package.json
+++ b/package.json
@@ -198,6 +198,6 @@
   "dependencies": {
     "camelcase-keys-recursive": "^0.8.2",
     "reselect": "^2.5.4",
-    "vets-json-schema": "https://github.com/department-of-veterans-affairs/vets-json-schema.git#75dc05646af4c553c54d70e9aab1a9639fb28ce7"
+    "vets-json-schema": "https://github.com/department-of-veterans-affairs/vets-json-schema.git#0e5db3eaa46d9a666ce96fe35ee7d6b71ff77b8b"
   }
 }
--- a/src/js/common/schemaform/widgets/CurrencyWidget.jsx
+++ b/src/js/common/schemaform/widgets/CurrencyWidget.jsx
@@ -22,7 +22,7 @@ export default class CurrencyWidget extends React.Component {
       this.props.onChange();
     } else {
       // Needs to look like a currency
-      if (!val.test(/^\${0,1}[0-9,]*(\.\d{1,2})?$/)) {
+      if (!/^\${0,1}[0-9,]*(\.\d{1,2})?$/.test(val)) {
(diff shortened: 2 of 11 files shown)


cehsu (approved):
Looks good to me.

src/js/common/schemaform/widgets/CurrencyWidget.jsx
```diff
@@ -22,7 +22,7 @@ export default class CurrencyWidget extends React.Component {
       this.props.onChange();
     } else {
       // Needs to look like a currency
-      if (!val.match(/^\${0,1}[0-9,]*(\.\d{1,2})?$/)) {
+      if (!/^\${0,1}[0-9,]*(\.\d{1,2})?$/.test(val)) {
```
cvalarida: Is this just so we return a boolean instead of an array / null (for cleanliness)?

package.json
```diff
@@ -198,6 +198,6 @@
   "dependencies": {
     "camelcase-keys-recursive": "^0.8.2",
     "reselect": "^2.5.4",
-    "vets-json-schema": "https://github.com/department-of-veterans-affairs/vets-json-schema.git#75dc05646af4c553c54d70e9aab1a9639fb28ce7"
+    "vets-json-schema": "https://github.com/department-of-veterans-affairs/vets-json-schema.git#0e5db3eaa46d9a666ce96fe35ee7d6b71ff77b8b"
```
cvalarida: Why was this updated? (Mostly out of curiosity.)

cvalarida (approved):

package.json
jbalboni: The schema was using integer for a bunch of these fields.

src/js/common/schemaform/widgets/CurrencyWidget.jsx
jbalboni: Yeah. I think it's also slightly faster, but not enough to notice, I would think.

jbalboni:
This can't be merged until the backend is using the updated schema.

src/js/common/schemaform/widgets/CurrencyWidget.jsx
cvalarida: I would imagine so; it can probably fail fast where `val.match` can't (maybe?) and doesn't have to find all the parentheses matches.

I'm not committed enough to make a benchmark right now, though. :man_shrugging:

PR #6175, opened 2017-08-07
Fix mistake about fingerprint property location

jbalboni:
The `fingerprint` property is in the wrong place, so it's not picked up by Sentry.


cehsu (approved):
LGTM.

cvalarida (commented):
Oi, that's annoying. Good catch, though!

PR #6176, opened 2017-08-07
Stub Out 1990 RJSF Chapters

cvalarida:
This isn't connected to a single issue, but it'll theoretically make merge conflicts less painful. Also, this way we know where the chapter we're working on goes.

I didn't look for shared pages, so that'll still need to be done when we actually _make_ the pages, but this should be the right chapter order. Hopefully.

### DIFF
--- a/src/js/edu-benefits/1990-rjsf/config/form.js
+++ b/src/js/edu-benefits/1990-rjsf/config/form.js
@@ -198,6 +198,20 @@ const formConfig = {
             }
           }
         },
+        dependents: {
+          title: 'Dependent information',
+          path: 'personal-information/dependents',
+          depends: {
+            // hasServiceBefore1978
+          },
+          uiSchema: {
+          },
+          schema: {
+            type: 'object',
+            properties: {
+            }
+          }
+        },
         directDeposit: {
           title: 'Direct deposit',
           path: 'personal-information/direct-deposit',
--- a/src/js/edu-benefits/1990-rjsf/config/form.js
+++ b/src/js/edu-benefits/1990-rjsf/config/form.js
@@ -24,6 +24,193 @@ const formConfig = {
   title: 'Apply for education benefits',
   subTitle: 'Form 22-1990',
   chapters: {
+    veteranInformation: {
+      title: 'Veteran Information',
+      pages: {
+        benefitEligibility: {
+          title: 'Veteran information',
+          path: 'veteran-information',
+          uiSchema: {
+          },
+          schema: {
+            type: 'object',
+            properties: {
+            }
+          }
+        }
+      }
+    },
+    benefitsEligibility: {
+      title: 'Benefits Eligibility',
+      pages: {
+        benefitEligibility: {
+          title: 'Benefits eligibility',
+          path: 'benefits-eligibility/benefits-selection',
+          uiSchema: {
+          },
+          schema: {
+            type: 'object',
+            properties: {
+            }
+          }
+        },
+        benefitRelinquishment: {
+          title: 'Benefits relinquishment',
+          path: 'benefits-eligibility/benefits-relinquishment',
+          depends: {
+            chapter33: true
+          },
+          uiSchema: {
+          },
+          schema: {
+            type: 'object',
+            properties: {
+            }
+          }
+        }
+      }
+    },
+    militaryHistory: {
+      title: 'Military History',
+      pages: {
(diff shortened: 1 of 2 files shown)


src/js/edu-benefits/1990-rjsf/config/form.js
```diff
@@ -24,6 +24,193 @@ const formConfig = {
...
+          },
+          schema: {
+            type: 'object',
+            properties: {
+            }
+          }
+        }
+      }
+    },
+    educationhistory: {
```
jbalboni: Capital H?

src/js/edu-benefits/1990-rjsf/config/form.js
```diff
@@ -24,6 +24,193 @@ const formConfig = {
   title: 'Apply for education benefits',
   subTitle: 'Form 22-1990',
   chapters: {
+    veteranInformation: {
+      title: 'Veteran Information',
+      pages: {
+        benefitEligibility: {
```
jbalboni: Think this name should be something else

src/js/edu-benefits/1990-rjsf/config/form.js
```diff
@@ -24,6 +24,193 @@ const formConfig = {
...
+            properties: {
+            }
+          }
+        }
+      }
+    },
+    benefitsEligibility: {
+      title: 'Benefits Eligibility',
+      pages: {
+        benefitEligibility: {
```
jbalboni: benefit**s**Eligibility?

jbalboni (approved):
Few minor things, looks good.

src/js/edu-benefits/1990-rjsf/config/form.js
```diff
@@ -24,6 +24,193 @@ const formConfig = {
...
+      }
+    },
+    educationhistory: {
+      title: 'Education History',
+      pages: {
+        educationHistory: {
+          title: 'Education history',
+          // There's only one page in this chapter (right?), so this url seems a
+          //  bit heavy-handed.
+          path: 'education-history/education-information',
```
cehsu: Should we make this consistent with the current 1990 routes or take this opportunity to make the 1990 routes consistent with the other education forms (e.g. `path: 'education/history'`)?

cehsu (approved):
LGTM. Had a question about whether paths should be consistent with existing 1990 paths or edu-benefits in general (assuming these are already consistent, I just checkout out a few). But if the second option is chosen, this can be changed later, as each chapter is built as well, so no need to block this.

src/js/edu-benefits/1990-rjsf/config/form.js
```diff
@@ -24,6 +24,193 @@ const formConfig = {
   title: 'Apply for education benefits',
```
cehsu: Like in pensions, could it be helpful to specify "Veteran education benefits"? Not sure when/if the specification is helpful.

src/js/edu-benefits/1990-rjsf/config/form.js
cvalarida: Naaaaaaaah!

...ahem, copy pasta. Fixing now.

src/js/edu-benefits/1990-rjsf/config/form.js
cvalarida: who needs camel case? i'm going for plane-case!

...ahem, fixing...

src/js/edu-benefits/1990-rjsf/config/form.js
cvalarida: An excellent question! I think @goldenmeanie is the one to answer that, right?

src/js/edu-benefits/1990-rjsf/config/form.js
cvalarida: I don't _think_ this is necessary, but I'll let somebody else make that decision. @J-Quag maybe? (If not, pull in whomever it should be.)

At any rate, this can be changed in another PR since it's not specifically related to the changes made here.

src/js/edu-benefits/1990-rjsf/config/form.js
jbalboni: I think we know we want to make them consistent with the rest of the ed forms, just a question of when. There's a ticket for it, so it doesn't have to be here.

I believe the only effect it will have is on analytics, which isn't too hard to update.

PR #6178, opened 2017-08-07
Update copy and make button 'visitable' for feedback on clicks on analytics optout

robertfairhead:
related to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3624


robertfairhead:
Tried to clarify the caveat about needing to visit in for each device/browser to opt out with new copy.

Made the button "visitable" try changing to it an `<a>` to no where. That will trigger the primary button's `:visited` state to give feedback that the click took effect.

Normal:
screen shot 2017-08-07 at 2 10 57 pm

Visited:
screen shot 2017-08-07 at 2 11 05 pm

The other option would be to change the button's class to `.usa-button-disabled` in the `onClick` handler but i didn't want to shove to much functionality into an inline piece of JS, so this seemed cleaner.

robertfairhead:
I totally changed my mind and went the disabled + new text route. Note, it doesn't actually stop the event from being sent a second time. It doesn't matter if it's sent multiple times or not, the custom Dimension won't change it'll just get overwritten with the same value.

Before click:

screen shot 2017-08-07 at 3 38 05 pm

After click:

screen shot 2017-08-07 at 3 38 21 pm

robertfairhead:
@phiden could you give this a quick once over? I don't love how long the onclick is, but given this is a purely internal thing it didn't seem worth it to break it out.

phiden (approved):
fixed a couple of typos but otherwise LGTM

PR #6179, opened 2017-08-07
1990 rjsf veteran info

cvalarida:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2811

Old on the left, new on the right:
image

image

### DIFF
--- a/src/js/edu-benefits/1990-rjsf/config/form.js
+++ b/src/js/edu-benefits/1990-rjsf/config/form.js
@@ -1,4 +1,5 @@
-// import _ from 'lodash/fp';
+import _ from 'lodash/fp';
+import moment from 'moment';
 
 import fullSchema1990 from 'vets-json-schema/dist/22-1990-schema.json';
 
@@ -29,7 +30,7 @@ const formConfig = {
     applicantInformation: {
       title: 'Applicant Information',
       pages: {
-        applicantInformation: applicantInformation(fullSchema1990, {
+        applicantInformation: _.merge(applicantInformation(fullSchema1990, {
           isVeteran: true,
           fields: [
             'veteranFullName',
@@ -37,6 +38,19 @@ const formConfig = {
             'veteranDateOfBirth',
             'gender'
           ],
+        }), {
+          uiSchema: {
+            veteranDateOfBirth: {
+              'ui:validations': [
+                (errors, dob) => {
+                  // TODO: Use a constant instead of 'YYYY-MM-DD'
+                  if (moment(dob, 'YYYY-MM-DD').isAfter(moment().endOf('day').subtract(17, 'years'))) {
+                    errors.addError('You must be at least 17 to apply');
+                  }
+                }
+              ]
+            }
+          }
         })
       }
     },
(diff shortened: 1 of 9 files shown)


src/js/edu-benefits/1990-rjsf/config/form.js
```diff
@@ -24,20 +26,18 @@ const formConfig = {
...
-            properties: {
-            }
-          }
-        }
+        applicantInformation: applicantInformation(fullSchema1990, {
+          isVeteran: true,
+          fields: [
+            'veteranFullName',
+            'veteranSocialSecurityNumber',
+            'veteranDateOfBirth',
```
jbalboni: I think there's some extra validation around date of birth, where you have to be over 17. Look in `src/js/edu-benefits/1990/utils/validations.js`.

src/js/edu-benefits/1990-rjsf/config/form.js
```diff
@@ -24,20 +26,18 @@ const formConfig = {
   title: 'Apply for education benefits',
   subTitle: 'Form 22-1990',
   chapters: {
-    veteranInformation: {
-      title: 'Veteran Information',
+    applicantInformation: {
```
jbalboni: I think we need some unit tests for this page, too.

cvalarida: Ohhhh, right. Because our current unit tests don't count for this. Duh.

cehsu (approved):
Looks good to me.

src/js/edu-benefits/1990-rjsf/config/form.js
```diff
@@ -24,20 +27,31 @@ const formConfig = {
...
           uiSchema: {
-          },
-          schema: {
-            type: 'object',
-            properties: {
+            veteranDateOfBirth: {
+              'ui:validations': [
+                (errors, dob) => {
+                  // TODO: Use a constant instead of 'YYYY-MM-DD'
+                  if (moment(dob, 'YYYY-MM-DD').isAfter(moment().endOf('day').subtract(17, 'years'))) {
```
jbalboni: This makes me a little nervous, since it implicitly relies on moment returning false when you call `isAfter` (or anything else, I think) on an invalid date (which is the case when `dob` is undefined).

Curious what you think. I worry that if we get used to doing this, we'll eventually hit a case where it does something unexpected.

jbalboni (approved):

src/js/edu-benefits/1990-rjsf/config/form.js
cvalarida: Yeah, I mean it will return `false` for invalid dates (I tested), but when the date doesn't fit the format supplied, that's when some funny things happen.

If the `dob` is `12-12-1991`, funny things happen: http://jsbin.com/balenepimo/2/edit?js,console
This will come out to `Wed Dec 19 2012 00:00:00 GMT-0800` which...is wrong.

I'm open to suggestions, but I'm not sure how else to do that. I mean, I could do the parsing myself, but that seems a little heavy-handed, doesn't it?

jbalboni: I'm not too worried about the format being incorrect, just if it's undefined or not. All the dates come from our date widget or backend prefill code, so I'm pretty confident that we're always going to have an ISO date, if it's not empty.

So I was at most thinking of a check that the date is undefined, or that it matches `/\d{4}-\d{2}-\d{2}/`, just so we're explicitly handling the empty case.

But like I said, I'm not sure that's worth it.

PR #6180, opened 2017-08-07
Content/wip/add var to healthcare

DanielleSoCompany:
Suggested placements for VAR links.

### DIFF
--- a/content/pages/health-care/about-va-health-care/where-you-get-care.md
+++ b/content/pages/health-care/about-va-health-care/where-you-get-care.md
@@ -71,7 +71,7 @@ Your VA provider will work with you to make sure you get the care you need. You 
 - [Where should I go for care in an emergency?](#care-emergency)
 - [What should I do if I move?](#care-if-moving)
 - [What if I need care while traveling? ](#care-while-traveling)
-- [How do I make an appointment at a VA health facility?](#care-appointment)
+- [How do I request an appointment at a VA health facility?](#care-appointment)
 - [Can you help me get to and from appointments?](#care-transportation)
 - [Can you help me cover travel expenses if I need to travel for care?](#care-travel-expenses)
 
@@ -140,11 +140,11 @@ For more information, contact your care team or the Traveling Veteran Coordinato
 
 <br>
 
-<span id="care-appointment>
+<span id="care-appointment">
 
 #### How do I make an appointment at a VA health facility?
 
-You can call the VA health facility where you go for care. [Find your VA health facility's phone number](/facility).
+You can call the VA health facility where you receive care. [Find your VA health facility's phone number](/facility).
 
 You can also use our VA Appointments scheduling app to request an appointment and to schedule select appointments. [Start using the VA Appointments app](https://veteran.mobilehealth.va.gov/veteran-appointment-requests/). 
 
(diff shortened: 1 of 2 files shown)


DanielleSoCompany:
@amybehrens @bethpotts Please see my suggested placement for the VAR link on the "Your Care Team" and "Where You'll Get Care" pages. I had 2 thoughts on where we could place the link on these pages: Either at the bottom of the first main chunk of content or as its own separate question. I've added both so you can review both options and let me know which you prefer. Thanks!

bethpotts:
@amybehrens @DanielleSoCompany Thanks for the two options to review. I like it broken out as a separate question. It stands out more with its own subhead.

amybehrens:
@DanielleSoCompany and @bethpotts I it as its own question too (which surprised me!). 

I'm wary of calling it an app because it's a web app (not something people need to download from the app store). I also think it's important to mention they need a DS Logon to use it so we don't set up an even more frustrating experience. 

Can we go with something like:
You can also use your DS Logon to request and schedule non-urgent visits online. (link the whole sentence)

DanielleSoCompany:
@amybehrens yes, i like that! @bethpotts I'm updating now and will ping you when this is ready for proofing.

DanielleSoCompany:
okay, @bethpotts this is ready for proofing! @amybehrens just an fyi that i added this additional text to define DS Logon: "Your DS Logon is the same logon used for eBenefits." In recent identity/logon research, we found that users didn't recognize what "DS Logon" was even though they had it for eBenefits. Thanks!

bethpotts (approved):
@DanielleSoCompany @amybehrens I've proofed this and it's ready to run past leadership.

DanielleSoCompany:
Hi @robertfairhead. The VA Appointments Request (VAR) team has requested that we set up tracking for user clicks on the links to VAR from Vets.gov. Per the PR, there will be links to VAR from /health-care/about-va-health-care/your-care-team and from /health-care/about-va-health-care/where-you-get-care. Could you set this up? Thank you! And just let @amybehrens or I know if you have any questions about this.

robertfairhead:
We track all clicks on links to non `vets.gov` domains so we would pick up any clicks on those links. Outbound click tracking like that isn't perfect so we may miss some because the internet isn't always perfect, but we can definitely give them a sense of it from the reporting.

DanielleSoCompany:
okay, great - thank you @robertfairhead!

DanielleSoCompany:
Update: This relates to Issue #4247 and is with Jeff Grandon for review.

DanielleSoCompany:
@amybehrens Jeff has approved this PR so we're ready to publish!

ghost (approved):

DanielleSoCompany:
@bethpotts can you take one more copy editing look at this before I ask Sophia to merge? Thank you!

bethpotts:
@DanielleSoCompany I avoided repetition of "use" in one sentence by substituting "access"--take a look at what I did to make sure you agree. Content looks great! Approved.

bethpotts:
@phiden Would you please review this PR and merge if all looks good? Thanks! We are not waiting on anyone else to review.

phiden (approved):

PR #6182, opened 2017-08-08
Save defaulted form data on page change in rjsf forms

jbalboni:
On the Pension form, if you have a dependent and you don't change their financial info, no financial information will be included in the array for each item, even though all fields are defaulted to 0. This happens because when create the object for each dependent, we're using a schema that doesn't have the financial fields.

This change makes it so that we always save the data rjsf sends back in the onSubmit callback (which has the defaults set) when we switch pages. I've limited this to just for page per item pages, but it could probably be used for all pages.

### DIFF
--- a/src/js/common/schemaform/FormPage.jsx
+++ b/src/js/common/schemaform/FormPage.jsx
@@ -72,7 +72,16 @@ class FormPage extends React.Component {
     this.props.setData(newData);
   }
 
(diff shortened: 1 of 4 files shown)


cvalarida (approved):

cehsu (approved):
Looks good to me.

PR #6183, opened 2017-08-08
Adding google search console verification

robertfairhead:
This adds the required html file to verify to google that we control vets.gov to get access to their search console (and related GA integration). We had this setup previously, but it appears to have broken about three weeks ago and was disabled two weeks ago:

screen shot 2017-08-08 at 11 11 00 am

Uploading this page is the first step. Once live on production, I will complete our reverification per Google instructions:

screen shot 2017-08-08 at 11 12 10 am

Tagging @ayaleloehr who originally set this up and @robbiethegeek-usds for his Google search work.


ayaleloehr (approved):
Clearly I missed the "to stay verified, don't remove this HTML page," as I just put it up initially and took it down. This seems more sustainable. LGTM. I haven't actually tested this, but it also won't break anything if it's not right, so doesn't feel very risky either.

PR #6184, opened 2017-08-08
Gibs utc time

annekainicUSDS:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4281

We heard of an issue where some users were seeing their DOB as being one day off. We suspect this is due to a timezone issue where the dates sent to us from EVSS are specifying UTC, and when we display these dates on the front-end it changes the day in their DOB depending on the timezone of their birth. This PR simply drops the time and timezone before reformatting the dates.

### DIFF
--- a/src/js/common/utils/helpers.js
+++ b/src/js/common/utils/helpers.js
@@ -101,16 +101,21 @@ export function dateToMoment(dateField) {
   });
 }
 
-export function formatDateLong(date) {
-  return moment(date).format('MMMM DD, YYYY');
-}
-
-export function formatDateWithoutTimeLong(date) {
+export function dropTimeAndTimezone(date) {
   let newDate = moment(date).format();
   const indexOfTime = newDate.indexOf('T');
   if (indexOfTime >= 0) {
     newDate = newDate.substring(0, indexOfTime);
   }
+  return newDate;
+}
+
+export function formatDateLong(date) {
+  return moment(date).format('MMMM DD, YYYY');
+}
+
+export function formatDateWithoutTimeLong(date) {
+  const newDate = dropTimeAndTimezone(date);
   return moment(newDate).format('MMMM DD, YYYY');
 }
 
@@ -119,11 +124,7 @@ export function formatDateShort(date) {
 }
 
 export function formatDateWithoutTimeShort(date) {
-  let newDate = moment(date).format();
-  const indexOfTime = newDate.indexOf('T');
-  if (indexOfTime >= 0) {
-    newDate = newDate.substring(0, indexOfTime);
-  }
+  const newDate = dropTimeAndTimezone(date);
   return moment(newDate).format('MM/DD/YYYY');
 }
 
--- a/src/js/common/utils/helpers.js
+++ b/src/js/common/utils/helpers.js
@@ -105,8 +105,26 @@ export function formatDateLong(date) {
   return moment(date).format('MMMM DD, YYYY');
 }
 
+export function formatDateWithoutTimeLong(date) {
+  let newDate = moment(date).format();
+  const indexOfTime = newDate.indexOf('T');
+  if (indexOfTime >= 0) {
+    newDate = newDate.substring(0, indexOfTime);
+  }
+  return moment(newDate).format('MMMM DD, YYYY');
+}
+
 export function formatDateShort(date) {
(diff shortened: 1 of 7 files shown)


src/js/common/utils/helpers.js
```diff
@@ -101,14 +101,33 @@ export function dateToMoment(dateField) {
   });
 }
 
+export function dropTimeAndTimezone(date) {
+  let newDate = moment(date).format();
+  const indexOfTime = newDate.indexOf('T');
```
omgitsbillryan: This will always resolve to `-1`.  I think you meant to write:
```javascript
const indexOfTime = date.indexOf('T'); // not newDate
```

A side note, I'd recommend renaming `newDate` to `formattedDate`.

src/js/common/utils/helpers.js
```diff
@@ -101,14 +101,33 @@ export function dateToMoment(dateField) {
...
+    newDate = newDate.substring(0, indexOfTime);
+  }
+  return newDate;
+}
+
 export function formatDateLong(date) {
   return moment(date).format('MMMM DD, YYYY');
 }
 
+export function formatDateWithoutTimeLong(date) {
```
omgitsbillryan: `formatDateWithoutTimeLong` & `formatDateWithoutTimeShort` seem excessive to me.  Just leave it up to the caller to make two calls like so,

```javascript 
var myDateTime = "1995-11-12T06:00:00.000+0000";
var date = dropTimeAndTimezone(myDateTime);
var formattedDate = formatDateLong(date);
```

omgitsbillryan (changes requested):
There's an error in your `dropTimeAndTimezone ` function that needs to be corrected.

src/js/common/utils/helpers.js
annekainicUSDS: Are you sure? I don't see that happening for me... All these 2 lines of code are doing first formatting the date passed in to be like `YYYY-MM-DDThh:mm:ssTZD` to guarantee all dates are in the same format. Then I'm just looking for the index of the character `T` in that string, which should always return `10`.

omgitsbillryan: Whoops, you're right, my mistake.  Ironically, this code is still problematic.  I just ran a test case on this function:
```javascript
> dropTimeAndTimezone('1995-11-12T03:00:00.000+0000');
"1995-11-11"
// expected = 1995-11-12
```

omgitsbillryan: This is because you're passing that date into `moment()` straight away on line 105.

annekainicUSDS: Gah. You're right. It's the `moment(date).format()` at the beginning of the function that is changing the date. I will fix that.

src/js/common/utils/helpers.js
omgitsbillryan: Fair enough.  I would still question the need for two functions `formatDateWithoutTimeLong` & `formatDateLong` implying that in some places we _do_ and in other places we _do not_ want timezone incorporated into the formatting of the date.

annekainicUSDS: We're not using `formatDateLong` anywhere in GIBS, but as this function exists in a common helper file, I'm leaving it in case other apps use or plan to use it in the future. They may want a generic function that doesn't drop time and timezone from the date.

src/js/common/utils/helpers.js
```diff
@@ -101,14 +101,33 @@ export function dateToMoment(dateField) {
   });
 }
 
+export function dropTimeAndTimezone(date) {
+  let formattedDate = date.toString();
```
annekainicUSDS: Instead of immediately formatting the `date` input, I'm converting it to a string. The reason being is because calling `.indexOf()` on anything other than a string will throw an error, preventing the app from even rendering. While the changes of this happening are small, I wanted to at least convert the date to a string first so that the entire app won't fail to load if this happens.

omgitsbillryan (approved):
LGTM.  Needs tests though;  e-to-e automated tests or unit.  Probably doesn't need to block this PR given the urgency.

annekainicUSDS:
Yep, totally agree on the tests. Created a ticket to specifically track this: https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4281. Will try to get a test PR in tomorrow.

PR #6186, opened 2017-08-08
Adding Secondary Jenkins File to just test build

CyberKoz:
Adding a secondary Jenkins file that can be used just to test the build on the k8s Jenkins cluster

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -77,4 +77,4 @@ RUN yarn install --production=false
 # Copy application source to image
 
 COPY . .
-RUN chmod -R 777 /application
+RUN chmod -R 777 /
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,10 +2,6 @@
 
 FROM buildpack-deps:jessie
 
-# Match the jenkins uid/gid on the host (504)
-RUN groupadd --gid 504 jenkins \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -82,4 +82,6 @@ RUN yarn install --production=false
 
 COPY . .
 RUN chown -R jenkins:jenkins /application
+RUN chmod -R 777 /application
```
jkassemi: Can you give a little info on this change? I know the jenkins user workaround for volume permissions was wonky, but maybe there's a better way?

jkassemi (approved):

Dockerfile
CyberKoz: So on the cluster I am building on the Jenkins user is a different UID, I initially had changed the docker file to use the UID on the different Jenkins cluster, but I didn't want to change the existing build path.

As to why this occurs, we don't have a properly implemented "out of source" build with build artifacts getting written to both /application and $HOME. We could retool everything to only write artifacts to a specific directory and then open up that directory, but I don't know enough about the NPM build stack to comment on how hard that would be. I get around $HOME by setting it to be something world writable in Jenkinsfile.pr.

PR #6189, opened 2017-08-08
Fix formatting of today date

annekainicUSDS:
Missed a formatting fix for a `new Date()` in the previous PR (https://github.com/department-of-veterans-affairs/vets-website/pull/6184).

### DIFF
--- a/src/js/post-911-gib-status/components/UserInfoSection.jsx
+++ b/src/js/post-911-gib-status/components/UserInfoSection.jsx
@@ -1,10 +1,9 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import moment from 'moment';
 
 import InfoPair from './InfoPair';
 
-import { formatDateWithoutTimeShort, formatDateWithoutTimeLong } from '../../common/utils/helpers';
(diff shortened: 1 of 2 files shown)


src/js/post-911-gib-status/components/UserInfoSection.jsx
```diff
@@ -18,7 +19,7 @@ class UserInfoSection extends React.Component {
 
     // Get today's date to show information current as of
     const today = new Date();
-    const todayFormatted = formatDateWithoutTimeShort(today);
+    const todayFormatted = formatDateWithoutTimeShort(moment(today).format());
```
omgitsbillryan: Wouldn't this be a good candidate for `formatDateShort`?  Either way, LGTM

omgitsbillryan (approved):
LGTM

src/js/post-911-gib-status/components/UserInfoSection.jsx
annekainicUSDS: True, that would probably be sufficient here :) I'll clean that up

PR #6190, opened 2017-08-09
Reset buy-up to 600 if the input exceeds that amount

### LINKED ISSUE #5405: GI Bill CT: Buy-up should revert back to $600
This can be backlogged.

If a user inputs more than $600 in the buy-up input field the field should return to $600, the max value. Since the calculations limit the buy-up to $600 this is not critical for launch.

image

@leannammiller @saneshark @mphprogrammer

U-DON:
Resolves #5405.

Also refactored (1) an action so that event handling occurs directly within the component and (2) the use of components that should be rendered by class methods due to all props coming directly from the parent component anyway.

### DIFF
--- a/src/js/gi/actions/index.js
+++ b/src/js/gi/actions/index.js
@@ -188,9 +188,7 @@ export function fetchProfile(facilityCode, version) {
   };
 }
 
-export function calculatorInputChange(e) {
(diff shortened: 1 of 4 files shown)


patrickvinograd (approved):
LGTM - appreciate the refactor work.

PR #6192, opened 2017-08-09
Gibs parse zone fix

annekainicUSDS:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4281

Better fix for the time/timezone problem with DOBs. `moment.parseZone()` goes a step beyond `moment.utc()` by fixing the date instead of converting the date to local or utc time (https://momentjs.com/docs/#/parsing/parse-zone/).

Also added some unit tests with various edge cases to ensure that the correct date is getting rendered every time.


omgitsbillryan (approved):
Nice! 🎅

PR #6195, opened 2017-08-09
Use 'school' for every non-OJT institution.

### LINKED ISSUE #5726: GI BILL CT: Change Public/Private/Foreign "Institution" to "School"
@leannammiller  @melwoodard @jmcneal65

We need ensure the language describing schools on the GI Bill Comparison Tool is consistent between types of schools. 

In the header on the school profile page we should be saying Public /  Private / For Profit /Foreign "School" not institution.  Currently we only use the term school when a "For Profit" school is selected.

The consistency is important because it is unfair and confusing to call one a school and the other an institution. Additionally we are drafting the updates to the TAP curriculum and instructors use the GI Bill CT to teach the difference between types of schools and the feedback was that the word "institution" is confusing.

**Private School**
https://www.vets.gov/gi-bill-comparison-tool/profile/31000139

image

**Public School**
https://www.vets.gov/gi-bill-comparison-tool/profile/11801605

**Foreign School**
https://www.vets.gov/gi-bill-comparison-tool/profile/31002871

U-DON:
Resolves #5726.

### DIFF
--- a/src/js/gi/components/profile/HeadingSummary.jsx
+++ b/src/js/gi/components/profile/HeadingSummary.jsx
@@ -65,8 +65,7 @@ class HeadingSummary extends React.Component {
                 On-the-job training
               </IconWithInfo>
               <IconWithInfo icon="institution" present={it.type && it.type !== 'ojt'}>
-                {_.capitalize(it.type)}&nbsp;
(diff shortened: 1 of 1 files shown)


patrickvinograd (approved):
LGTM

PR #6198, opened 2017-08-10
Fixed breadcrumbs on GIBCT search and profile pages

U-DON:
Resolves department-of-veterans-affairs/vets.gov-team#2101.

Added "Search Results" to breadcrumbs on a profile page when search has been done already (the user arrived at the profile from a search).

If the user refreshes while on the profile page, "Search Results" will disappear from the breadcrumbs. But going back in the browser will still return to the search results page with the search filters that were last active.


U-DON:
@melwoodard: Just checked the breadcrumbs on some content pages and noticed that the last item in the list is linked too. For consistency, that should probably be the case here, right?

Should "Search Results" in the breadcrumbs on the search results page link to an empty search or the same search with the same filters (we can carry over the search filters in the URL but not the eligibility selections)?

It got me thinking about how to treat them consistently when it comes to React apps. When clicking on breadcrumb links in a React app, should it do a full page refresh or a partial page load (we're mostly doing the latter in React apps right now, I believe)? Also, a link that partially loads effectively does nothing if you're already on the page, so I'm not sure if we should even do anything with the last breadcrumb other than style it as a link.

melwoodard:
@U-DON actually it's a little different than that -- we made a change recently to take the _current_ page out of the breadcrumbs since it's not actionable and takes up valuable space on mobile. So in this case on a profile page, Harvard University wouldn't appear and either Search Results/GI Bill Comparison Tool (based on the two use cases above) would be the last link in the line.

U-DON:
@melwoodard: Oh, okay. That's how it was before for GIBCT, but I most recently saw that the last link on the content was the current page, so thought I had to change it here. I'll change it back.

U-DON:
Updated:

#### When user gets to the profile from a search
> screen shot 2017-08-11 at 2 40 18 pm

#### When user has opened profile directly or refreshed the page after getting there from search
> screen shot 2017-08-11 at 2 39 59 pm

melwoodard (approved):
Design looks good, thanks!

bshyong (approved):
LGTM

PR #6206, opened 2017-08-11
Edu 1990 rjsf military history

cehsu:
Connects to department-of-veterans-affairs/vets.gov-team#2815

TODO:
- [x] Configure military service and service periods UI
- [x] Write tests
- [x] Fix className bug (Thanks @cvalarida for solving this!)

screen shot 2017-08-14 at 4 14 31 pm

#### Service Periods
screen shot 2017-08-11 at 5 23 17 pm

#### Expanded
screen shot 2017-08-11 at 5 23 29 pm

#### Multiple periods
screen shot 2017-08-11 at 5 24 12 pm

### DIFF
--- a/package.json
+++ b/package.json
@@ -198,6 +198,6 @@
   "dependencies": {
     "camelcase-keys-recursive": "^0.8.2",
     "reselect": "^2.5.4",
-    "vets-json-schema": "https://github.com/department-of-veterans-affairs/vets-json-schema.git#07e1e0f333491559a23c80074a44eee333d36b16"
+    "vets-json-schema": "https://github.com/department-of-veterans-affairs/vets-json-schema.git#91379d76189f605d53b9ec810da1171aa4975ea4"
   }
 }
--- a/package.json
+++ b/package.json
@@ -198,6 +198,6 @@
   "dependencies": {
     "camelcase-keys-recursive": "^0.8.2",
     "reselect": "^2.5.4",
-    "vets-json-schema": "https://github.com/department-of-veterans-affairs/vets-json-schema.git#0e5db3eaa46d9a666ce96fe35ee7d6b71ff77b8b"
+    "vets-json-schema": "https://github.com/department-of-veterans-affairs/vets-json-schema.git#07e1e0f333491559a23c80074a44eee333d36b16"
   }
 }
--- a/src/js/common/schemaform/definitions/year.js
+++ b/src/js/common/schemaform/definitions/year.js
@@ -1,0 +1,17 @@
+import { validateYear } from '../validation';
+
+export default function uiSchema(title = 'Year') {
+  return {
+    'ui:title': title,
+    'ui:widget': 'updown',
+    'ui:options': {
+      widgetClassNames: 'usa-input-medium'
+    },
+    'ui:validations': [
+      validateYear
+    ],
+    'ui:errorMessages': {
+      pattern: 'Please provide a valid year'
+    }
+  };
+}
--- a/src/js/common/schemaform/definitions/year.js
+++ b/src/js/common/schemaform/definitions/year.js
@@ -1,9 +1,12 @@
 import { validateYear } from '../validation';
+import TextWidget from '../widgets/TextWidget.jsx';
+import * as ReviewWidget from '../review/widgets.jsx';
 
 export default function uiSchema(title = 'Year') {
   return {
     'ui:title': title,
-    'ui:widget': 'updown',
+    'ui:widget': TextWidget,
+    'ui:reviewWidget': ReviewWidget.TextWidget,
     'ui:options': {
       widgetClassNames: 'usa-input-medium'
     },
--- a/src/js/common/schemaform/definitions/year.js
+++ b/src/js/common/schemaform/definitions/year.js
@@ -1,16 +1,17 @@
-import { validateYear } from '../validation';
-import TextWidget from '../widgets/TextWidget.jsx';
+import { validateCurrentOrPastYear } from '../validation';
 import * as ReviewWidget from '../review/widgets.jsx';
 
 const uiSchema = {
   'ui:title': 'Year',
-  'ui:widget': TextWidget,
   'ui:reviewWidget': ReviewWidget.TextWidget,
   'ui:validations': [
-    validateYear
+    validateCurrentOrPastYear
   ],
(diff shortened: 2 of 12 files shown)


test/edu-benefits/1990-rjsf/config/servicePeriods.unit.spec.jsx
```diff
@@ -0,0 +1,76 @@
...
+import sinon from 'sinon';
+import ReactTestUtils from 'react-dom/test-utils';
+
+import { DefinitionTester, submitForm } from '../../../util/schemaform-utils.jsx';
+import formConfig from '../../../../src/js/edu-benefits/1990-rjsf/config/form';
+
+describe('Edu 1995 servicePeriods', () => {
+  const { schema, uiSchema } = formConfig.chapters.militaryHistory.pages.servicePeriods;
+  const definitions = formConfig.defaultDefinitions;
+  it('should render', () => {
```
jbalboni: This test looks identical to the next one, just without an exact count of fields. Maybe we should remove it?

src/js/common/schemaform/definitions/year.js
```diff
@@ -0,0 +1,17 @@
...
+
+const uiSchema = {
+  'ui:title': 'Year',
+  'ui:widget': TextWidget,
+  'ui:reviewWidget': ReviewWidget.TextWidget,
+  'ui:validations': [
+    validateYear
+  ],
+  'ui:errorMessages': {
+    pattern: 'Please provide a valid year'
```
jbalboni: I don't think this will ever be used, since this is an integer field and we don't set `pattern` schema properties for those. You could change `pattern` to `type`, since we that could theoretically be triggered.

package.json
```diff
@@ -198,6 +198,6 @@
   "dependencies": {
     "camelcase-keys-recursive": "^0.8.2",
     "reselect": "^2.5.4",
-    "vets-json-schema": "https://github.com/department-of-veterans-affairs/vets-json-schema.git#0e5db3eaa46d9a666ce96fe35ee7d6b71ff77b8b"
+    "vets-json-schema": "https://github.com/department-of-veterans-affairs/vets-json-schema.git#07e1e0f333491559a23c80074a44eee333d36b16"
```
jbalboni: I made a PR, this will need to be updated again: https://github.com/department-of-veterans-affairs/vets-json-schema/pull/142

jbalboni (commented):
Left some comments, looks good in general.

src/js/edu-benefits/1990-rjsf/config/form.js
```diff
@@ -150,21 +162,51 @@ const formConfig = {
...
           }
         },
         militaryService: {
           title: 'Military service',
           path: 'military-history/military-service',
           uiSchema: {
+            serviceAcademyGraduationYear: _.assign(yearUI, {
+              'ui:title': 'If you received a commission from a military service academy, what year did you graduate?',
+              'ui:options': {
+                widgetClassNames: 'usa-input-medium'
```
cehsu: Although it is not always the desired class (e.g. ROTC scholarship year/amount field on the next page), it is the default, so this change is still a great idea.

src/js/common/schemaform/definitions/year.js
cehsu: Thanks @jbalboni, great point, change made.

src/js/common/schemaform/definitions/year.js
```diff
@@ -0,0 +1,17 @@
+import { validateYear } from '../validation';
+import TextWidget from '../widgets/TextWidget.jsx';
+import * as ReviewWidget from '../review/widgets.jsx';
+
+const uiSchema = {
+  'ui:title': 'Year',
+  'ui:widget': TextWidget,
```
cehsu: Wow! That's convenient. Change made

src/js/common/schemaform/validation.js
```diff
@@ -255,6 +255,12 @@ export function validateFutureDateIfExpectedGrad(errors, dateString, formData) {
   }
 }
 
+export function validateYear(errors, year) {
```
cehsu: The logic here is a little different as it takes a year rather than a full date, also it checks whether the year is between 1900 and the current year. I wonder if making changes to `isValidCurrentOrPastYear` to accommodate these differences is what you had in mind? Or is simply checking that a year is in the past sufficient and the after 1900 check unnecessary?

package.json
jbalboni: That PR is merged, so this needs to be updated with the latest version from vets-json-schema.

src/js/common/schemaform/validation.js
jbalboni: `isValidCurrentOrPastYear` looks like it takes a year and does the 1900 check:

```js
function isValidCurrentOrPastYear(value) {
  return Number(value) >= 1900 && Number(value) < moment().year() + 1;
}
```

cehsu: Ah you're right, thanks for the reminder and clarification, change made.

test/edu-benefits/1990-rjsf/config/servicePeriods.unit.spec.jsx
cehsu: Good idea, removed.

cehsu: Thanks @jbalboni, great point, change made.

test/edu-benefits/1990-rjsf/config/servicePeriods.unit.spec.jsx
```diff
@@ -0,0 +1,76 @@
...
+  });
+  it('should render service fields', () => {
+    const form = ReactTestUtils.renderIntoDocument(
+      <DefinitionTester
+          schema={schema}
+          data={{}}
+          uiSchema={uiSchema}
+          definitions={definitions}/>
+    );
+    expect(ReactTestUtils.scryRenderedDOMComponentsWithTag(form, 'input').length).to.equal(5);
```
cehsu: Thanks @jbalboni, great point, change made.

package.json
cehsu: Thanks for the heads up, change made.

jbalboni (approved):

PR #6208, opened 2017-08-11
Fix search submission corner cases from landing page

patrickvinograd:
Fixes https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3697

Selection of autocomplete suggestion via mouse was not passing params to `handleFilterChange` so it was using the `autocomplete.searchTerm` prop instead (i.e. whatever was in the input box).

Changing the above broke the scenario of typing a string and hitting enter (i.e. not involving autocomplete at all). That case uses the form-level `handleSubmit`, so that too changed to pass the searchTerm to `handleFilterChange`.


U-DON (approved):
LGTM. Good catch.

PR #6209, opened 2017-08-14
Adjust Sentry error tracking for SiP

jbalboni:
Sentry error: http://sentry.vetsgov-internal/vets-gov/website-production/issues/13015/

We're tracking the above case as errors, but it's just a normal session expiration, so it doesn't need to be in Sentry. We're doing the same thing when loading a form, so I've removed that error as well.

Both cases are still tracking in GA, so we can see how often people are having to login in again.


cehsu (approved):
Looks good to me.

cvalarida (approved):

PR #6211, opened 2017-08-14
Adjust radio label alignment

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3726

Fixes the alignment. Seems ok in both RJSF and non-RJSF locations now.

screen shot 2017-08-14 at 3 11 49 pm
screen shot 2017-08-14 at 3 09 26 pm
screen shot 2017-08-14 at 3 09 19 pm

### DIFF
--- a/src/sass/modules/_m-form-elements.scss
+++ b/src/sass/modules/_m-form-elements.scss
@@ -55,7 +55,7 @@ button.form-button-disabled {
     display: block;
     float: left;
     margin-left: -1.7em;
-    margin-top: 0.14em;
+    margin-top: 0.25em;
     margin-right: 0;
     pointer-events: none;
   }


cehsu (approved):

src/sass/modules/_m-form-elements.scss
```diff
@@ -55,7 +55,7 @@ button.form-button-disabled {
     display: block;
     float: left;
     margin-left: -1.7em;
-    margin-top: 0.14em;
+    margin-top: 0.25em;
```
cvalarida: Seems odd that the labels were too low, but increasing the `margin-top` fixes it. What's up with that? I feel like I'm probably missing something.

cvalarida (approved):
Odd, but it looks good.

src/sass/modules/_m-form-elements.scss
jbalboni: This moves the radio button image down, instead of moving the labels up.

cvalarida: Oh weird. I guess it's the `label::before` that's the button image. Makes sense. I was reading that as just the label itself which, now that I think about it, is silly.

Thanks for the clarification!

PR #6212, opened 2017-08-14
Pass duplicate ID test for firefox

bshyong:
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4168

@aub this fix should pass all sauce tests successfully on firefox. I searched for the `dnsrsearch.com` issue in your test logs, and it seems like it may be related to an ISP or local web browser issue causing the test to fail


aub (approved):
LGTM

PR #6213, opened 2017-08-14
Fix for claims/appeals error msg

bshyong:
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4346
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4358

Adds explicit check to ensure that error messages do not show if app is in a loading state.

### DIFF
--- a/src/js/disability-benefits/containers/YourClaimsPage.jsx
+++ b/src/js/disability-benefits/containers/YourClaimsPage.jsx
@@ -88,7 +88,11 @@ class YourClaimsPage extends React.Component {
   }
 
   renderErrorMessages() {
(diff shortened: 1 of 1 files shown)


aub (approved):
LGTM

patrickvinograd:
LGTM although this looks like a good candidate for some unit tests based on state - each of claims and appeals can be loading, loaded successfully, or loaded with an error, so that's a lot of variants with slightly different rendering requirements.

bshyong:
Agree with adding coverage for these cases into unit tests, will do

PR #6214, opened 2017-08-15
Creating a prebuild script and adding command.

robbiethegeek-usds:
When starting to look at the build.js seems like it could be broken into some more semantic pieces, and figured I would start to tease apart. This is just a pull all the stuff that happens before the build starts into a prebuild script and create a new npm run script for prebuild that will be run automatically on npm run build.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "description": "Frontend build for vets.gov website static assets.",
   "scripts": {
     "build": "node script/build.js",
+    "prebuild": "node script/prebuild.js",
     "heroku-serve": "http-server",
     "heroku-postbuild": "node script/build.js --entry no-react,style",
     "lint": "npm run lint:js && npm run lint:sass",
--- a/script/build.js
+++ b/script/build.js
@@ -1,5 +1,5 @@
 // Builds the site using Metalsmith as the top-level build runner.
-
+const path = require('path');
 const Metalsmith = require('metalsmith');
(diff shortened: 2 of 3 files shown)


robbiethegeek-usds:
When starting to look at the build.js seems like it could be broken into some more semantic pieces, and figured I would start to tease apart.  This is just a pull all the stuff that happens before the build starts into a prebuild script and create a new npm run script for prebuild that will be run automatically on npm run build.

cvalarida (commented):
Looks good, but how is the prebuild script run automatically when the build script is?

robbiethegeek-usds:
@cvalarida it is a magical feature of npm scripts :) if there is a pre or a post it will be run before and/or after

cvalarida:
Good to know! Thanks, @robbiethegeek-usds!

cvalarida (approved):

jbalboni (approved):

jkassemi (approved):

robertfairhead:
@robbiethegeek-usds did you check on if this will impact the heroku builds? it uses a lightly different build path than the normal site builds

robbiethegeek-usds:
@jkassemi can you confirm that the heroku build for this PR was successful and make sure nothing there needs to change?

jkassemi:
@robbiethegeek-usds @robertfairhead not seeing any issues with the Heroku build! 🚂

PR #6217, opened 2017-08-15
Increase sauce timeouts

aub:
Related to department-of-veterans-affairs/vets.gov-team#4167

We have specs randomly failing on Sauce due to timeouts because things take forever to render.


bshyong (approved):
LGTM!

PR #6218, opened 2017-08-15
Navigate back when canceling health T&C

bshyong:
Adds a fix to navigate back in the browser history when terms and conditions are canceled instead of going to a hardcoded link. Preserves the use of the hard coded `cancelPath` prop if the history  does not exist.

Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4347


patrickvinograd:
This doesn't quite work as expected for the T&C modal from the user profile page. If you cancel out of the modal I'd expect it to return you to the profile page, instead it goes back to whatever you were on before the profile page. Since opening the modal from there doesn't push anything onto the history.

bshyong:
Ah, right. That won't work because the prompt shows up in a modal. I will add an override prop which forces the cancel button to be the normal hardcoded link

src/js/common/components/AcceptTermsPrompt.jsx
```diff
@@ -91,7 +101,7 @@ class AcceptTermsPrompt extends React.Component {
           </div>
           <div>
             {submitButton}
-            <a href={this.props.cancelPath} className="usa-button usa-button-outline">
+            <a href="#" onClick={this.onCancel} className="usa-button usa-button-outline">
```
U-DON: Should this just be a `button` instead of `a` now?

U-DON (approved):
LGTM. Minor nitpick.

src/js/common/components/AcceptTermsPrompt.jsx
bshyong: makes sense—will change this

PR #6226, opened 2017-08-16
Update testing modules

jbalboni:

This is the first half, just the testing related modules. In addition, I updated Mocha (just a minor version bump) and removed dirty-chai, which we don't use.

Almost all of the fixes I made were due to the chai upgrade, which exposed some tests that were running incorrectly and some breaking changes.


test/util/unit-helpers.js
```diff
@@ -10,20 +8,6 @@ chai.use(chaiAsPromised);
 
 const expect = chai.expect;
 
-function setupJSDom() {
```
jbalboni: This wasn't used, it's duplicated in mocha-setup.js

cehsu (approved):

test/common/schemaform/reducers/index.unit.spec.js
```diff
@@ -50,10 +50,10 @@ describe('schemaform createSchemaFormReducer', () => {
     const reducer = createSchemaFormReducer(formConfig);
     const state = reducer(undefined, {});
 
-    expect(state.submission).to.be.defined;
+    expect(state.submission).not.to.be.undefined;
     expect(state.data.privacyAgreementAccepted).to.be.false;
     expect(state.data.field).to.eql(formConfig.chapters.test.pages.page1.initialData.field);
-    expect(state.page2).to.be.defined;
+    expect(state.page2).to.be.undefined;
```
robbiethegeek-usds: Any idea bout these flip flops?

jbalboni: I don't think `defined` was a valid option before and Chai didn't fail because of it, which it does now. So, I think this exposed a test that was never properly working.

cvalarida: Does this test make sense? Should it be `undefined` because the page is essentially empty?

cvalarida (commented):
Looks mostly okay, but there are some weird flops as @robbiethegeek mentioned. Without being super familiar with the tests concerned, I can't say whether they're good or not, so :man_shrugging: 

Any idea how to check for unused node modules? Specifically, I'm looking at the one removed `import polyfillDataset from 'element-dataset'` and I'm wondering if it's used anywhere else...or if there are unused dependencies anywhere. Probably it's tangential to this PR, though.

test/common/schemaform/reducers/index.unit.spec.js
jbalboni: Yeah, this might not actually be a needed check.

jbalboni:
@cvalarida There's a `depcheck` module you could use to find unused modules, but I haven't tried it. That's probably a worthwhile thing to do once we get things upgraded.

bshyong (approved):
Great upgrades here; LGTM

PR #6227, opened 2017-08-16
Change default state of facility locator map

bshyong:
resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4349

Sets default map view to the following:
screen shot 2017-08-16 at 3 36 09 pm


aub (approved):
LGTM

PR #6228, opened 2017-08-16
Handles Incomplete RJSF Forms Better

cvalarida:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4286

When we enter a page in the middle of a form, it redirects us back to the introduction page (technically, the first page, but that's the intro, so :man_shrugging:). It doesn't redirect us back if we're in development, but we can recreate the functionality for testing purposes by appending the URL with `?redirect`.

image

### DIFF
--- a/src/js/common/schemaform/FormApp.jsx
+++ b/src/js/common/schemaform/FormApp.jsx
@@ -42,12 +42,13 @@ class FormApp extends React.Component {
     }
 
     // If we start in the middle of a form, redirect to the beginning
-    const currentPath = this.props.location.pathname;
+    const currentPath = this.props.currentLocation.pathname;
     const firstPagePath = this.props.routes[this.props.routes.length - 1].pageList[0].path;
     // If we're in production, we'll redirect if we start in the middle of a form
     // In development, we won't redirect unless we append the URL with `?redirect` ()
-    const devRedirect = __BUILDTYPE__ !== 'development' || this.props.location.search.includes('redirect');
+    const devRedirect = __BUILDTYPE__ !== 'development' || this.props.currentLocation.search.includes('redirect');
     if (currentPath !== firstPagePath && devRedirect) {
+      // If the first page is not the intro and uses `depends`, this will probably break
       this.props.router.push(firstPagePath);
     }
   }
--- a/src/js/common/schemaform/FormApp.jsx
+++ b/src/js/common/schemaform/FormApp.jsx
@@ -40,6 +40,16 @@ class FormApp extends React.Component {
     if (window.History) {
       window.History.scrollRestoration = 'manual';
     }
+
+    // If we start in the middle of a form, redirect to the beginning
+    const currentPath = this.props.location.pathname;
+    const firstPagePath = this.props.routes[this.props.routes.length - 1].pageList[0].path;
+    // If we're in production, we'll redirect if we start in the middle of a form
+    // In development, we won't redirect unless we append the URL with `?redirect` ()
+    const devRedirect = __BUILDTYPE__ !== 'development' || this.props.location.search.includes('redirect');
+    if (currentPath !== firstPagePath && devRedirect) {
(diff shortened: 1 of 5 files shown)


cvalarida:
We might want to move the alert down below the buttons since that's where the connection problem alert is. Also, it might just be better accessibility-wise since it'll be after the focused element (presumably). But to be honest, I'm just making that up; I don't actually know.

jbalboni:
@cvalarida Yeah, I think that message should go after the button.

src/js/common/schemaform/review/ReviewPage.jsx
```diff
@@ -37,7 +37,8 @@ class ReviewPage extends React.Component {
       // to be slower than shallow cloning objects
       viewedPages: new Set(
         getPageKeys(props.route.pageList, props.form.data)
-      )
+      ),
+      showValidationWarning: false
```
jbalboni: Another approach to this is to make another submission state (like `validationError` or something), call `this.props.setSubmission`, and add the error message in SubmitButtons (along with the other failure states).

src/js/common/schemaform/FormApp.jsx
```diff
@@ -40,6 +40,16 @@ class FormApp extends React.Component {
     if (window.History) {
       window.History.scrollRestoration = 'manual';
     }
+
+    // If we start in the middle of a form, redirect to the beginning
+    const currentPath = this.props.location.pathname;
+    const firstPagePath = this.props.routes[this.props.routes.length - 1].pageList[0].path;
```
jbalboni: This is making an assumption that the first page is always shown, which is true now, but we might want to make a note of it here. It's possible for someone to create a form that skips the intro page and makes the first page use `depends`.

I think that's rare enough that we don't need to account for it, but I think a comment would be helpful.

jbalboni (commented):
I think you'll need to make this set the submission property in the Redux state to be able to put the message below the button, but other than that it looks good.

src/js/common/schemaform/FormApp.jsx
```diff
@@ -40,6 +40,16 @@ class FormApp extends React.Component {
     if (window.History) {
       window.History.scrollRestoration = 'manual';
     }
+
+    // If we start in the middle of a form, redirect to the beginning
+    const currentPath = this.props.location.pathname;
```
jbalboni: I think you may want to use `currentLocation` instead of `location`. That's what the other code in this component uses. Or we could probably switch to using `location` and remove the `currentLocation` prop.

src/js/common/schemaform/FormApp.jsx
cvalarida: Yeah, I think the only reason we'd ever make the first "real" page in a form use `depends` is if we wanted it to depend on some pre-filled data, and that seems like an awful idea (famous last words). But! It's safer to put the comment in, so I'm on it!

src/js/common/schemaform/FormApp.jsx
cvalarida: As far as I can tell `this.props.location` == `this.props.currentLocation` == `this.props.router.location`. I'll use `currentLocation`, but we might consider revisiting this to see _why_ we've got three instances of the same object (basically).

cvalarida:
I just moved the error alert down below the buttons instead of messing around with redux and another file. If we need visibility outside of `ReviewPage`, we can put something in the redux store, but I'd prefer to keep it as simple as the component's state for now.

The original comment has been updated with the new screenshot.

cvalarida:
Mmm...on the other hand, it's not super clean to have error messages in two places...Mmm...

cvalarida:
I liked keeping all the error messages in one place, so I went with your way, @jbalboni. 👍

cehsu (approved):

jbalboni (approved):

PR #6229, opened 2017-08-17
Ensure prebuild script runs on Heroku

robertfairhead:
Heroku builds currently invoke the build.js script directly, which I believe bypasses the NPM pre- and post- script hooks.

Items currently in `prebuild.js` are not needed for Heroku builds so they're working fine. But if that changes, will want to be sure it gets picked up.

- [X] confirm current setup is skipping `prebuild.js` on Heroku
- [X] Confirm the fix runs prebuild

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
     "build": "node script/build.js",
     "prebuild": "node script/prebuild.js",
     "heroku-serve": "http-server",
-    "heroku-postbuild": "node script/build.js --entry no-react,style",
+    "heroku-postbuild": "npm run build --  --entry no-react,style",
     "lint": "npm run lint:js && npm run lint:sass",
     "lint:changed": "npm run lint:js:changed && npm run lint:sass",
     "lint:js": "eslint --quiet --ext .js --ext .jsx .",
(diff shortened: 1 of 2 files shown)


robertfairhead:
Prior to the change, no invocation of prebuild:

screen shot 2017-08-17 at 12 11 50 pm

After change: confirmed it ran our prebuild with `console.log`

screen shot 2017-08-17 at 12 14 05 pm

Final version shows NPM running both:

screen shot 2017-08-17 at 12 15 49 pm

robbiethegeek (approved):
Sorry about that

robbiethegeek-usds (approved):

jkassemi:
Do we actually want these to run on Heroku? I'd think making this stage something we expect to run on developer systems only would be ideal.

robertfairhead:
Yeah, agreed that the current checks in prebuild aren't needed on Heroku and could be skipped.

The Jenkinsfile uses `npm run build` so it'll invoke these as well. I figured it was worth having them run on Heroku mainly as future insurance against other things getting factored out into `prebuild.js` (or `postbuild.js`). Local and Jenkins would work fine and Heroku would be broken. Figured it was worth avoiding that special casing.

If we decide `prebuild.js` should be local dev box only then i'm 💯  with that but would suggest we add comments to make that clear in `prebuild.js` and alter the Jenkinsfiles to avoid `npm run build` so it will also skip them.

PR #6230, opened 2017-08-17
Update ESLint packages and adjust to new rules

jbalboni:

This updates the ESLint packages to the latest, except for the a11y rules, which are a version back because that's what the latest airbnb config requires.

I've tried to make only small changes, and turn off rules with big impacts. Still, it's a lot of changes. My hope is to turn on some or all of the rules that are newly turned off in our config, per discussion with the team.

I've also added comments to ignore some rules in certain files, if the rules seemed like ones we obviously want on, but were too difficult to fix in some instances. This mostly affects a11y rules.

If you're curious, this is the frequency of those turned off errors:

screen shot 2017-08-17 at 2 27 02 pm


src/js/pensions/config/form.js
```diff
@@ -110,7 +110,8 @@ const {
 const nonRequiredFullName = createNonRequiredFullName(fullName);
 
 function isUnder65(formData) {
-  return moment().startOf('day').subtract(65, 'years').isBefore(formData.veteranDateOfBirth);
+  return moment().startOf('day').subtract(65, 'years')
+    .isBefore(formData.veteranDateOfBirth);
```
cvalarida: Is this just because the line is long?

src/js/rx/client.js
```diff
@@ -1,25 +0,0 @@
-// TODO(awong): Convert from webpack common.js require() format to ES6 modules.
```
cvalarida: Why is this removed?

src/js/rx/reducers/errors.js
```diff
@@ -2,7 +2,7 @@ const initialState = {
   errors: [],
 };
 
-export default function errors(state = initialState, action) {
```
cvalarida: No default param? Is that right?

test/common/model/fields.spec.js
```diff
@@ -1,90 +0,0 @@
-const chai = require('chai');
```
cvalarida: `src/js/common/model/fields.js` still exists, so should this test really be deleted?

test/common/model/schema/definitions.spec.js
```diff
@@ -1,62 +0,0 @@
-const jsonValidator = require('../../../src/common/schema/validator');
```
cvalarida: Is this intentional?

test/edu-benefits/1990-rjsf/config/applicantInformation.unit.spec.jsx
```diff
@@ -1,10 +1,9 @@
 import React from 'react';
 import { expect } from 'chai';
 import ReactTestUtils from 'react-dom/test-utils';
-import { getFormDOM } from '../../../util/schemaform-utils';
 import moment from 'moment';
 
-import { DefinitionTester } from '../../../util/schemaform-utils.jsx';
+import { DefinitionTester, getFormDOM } from '../../../util/schemaform-utils.jsx';
```
cvalarida: Probably don't need the extension here.

test/edu-benefits/1990-rjsf/config/benefitsEligibility.unit.spec.jsx
```diff
@@ -1,9 +1,8 @@
 import React from 'react';
 import { expect } from 'chai';
 import ReactTestUtils from 'react-dom/test-utils';
-import { getFormDOM } from '../../../util/schemaform-utils';
 
-import { DefinitionTester } from '../../../util/schemaform-utils.jsx';
+import { DefinitionTester, getFormDOM } from '../../../util/schemaform-utils.jsx';
```
cvalarida: Or here.

test/edu-benefits/1990-rjsf/config/benefitsRelinquishment.unit.spec.jsx
```diff
@@ -1,11 +1,10 @@
 import React from 'react';
 import { expect } from 'chai';
 import ReactTestUtils from 'react-dom/test-utils';
-import { getFormDOM } from '../../../util/schemaform-utils';
 import sinon from 'sinon';
 import _ from 'lodash/fp';
 
-import { DefinitionTester } from '../../../util/schemaform-utils.jsx';
+import { DefinitionTester, getFormDOM } from '../../../util/schemaform-utils.jsx';
```
cvalarida: ...or here...

test/pensions/config/employmentHistory.unit.spec.jsx
```diff
@@ -1,11 +1,10 @@
 import React from 'react';
 import moment from 'moment';
-import { getFormDOM } from '../../util/schemaform-utils';
 import { expect } from 'chai';
 import sinon from 'sinon';
 import ReactTestUtils from 'react-dom/test-utils';
 
-import { DefinitionTester } from '../../util/schemaform-utils.jsx';
+import { DefinitionTester, getFormDOM } from '../../util/schemaform-utils.jsx';
```
cvalarida: ...ahem... :man_shrugging:

cvalarida (commented):
Not sure how some of these weren't already fixed, actually.

That notwithstanding, they mostly look alright, but there were some tests that were deleted entirely (which didn't look right) and some other odd goings on.

Um...also, feel free to ignore the extension comments if you like. I just thought I'd point it out while I saw it.

src/js/pensions/config/form.js
jbalboni: There's a rule about breaking chained calls into multiple lines. It seems weird, but it also only seemed to cause problems in two places, so I left it on.

src/js/rx/client.js
jbalboni: It's never used, as far as I can tell.

src/js/rx/reducers/errors.js
jbalboni: The default parameter is unneeded, there's a rule that catches that now.

test/common/model/fields.spec.js
jbalboni: This test is never run. That source file is also never used, so I should delete it, too.

test/common/model/schema/definitions.spec.js
jbalboni: This test isn't ever run and doesn't make sense anymore, as far as I can tell.

test/edu-benefits/1990-rjsf/config/applicantInformation.unit.spec.jsx
jbalboni: Nope, but I think we'll end up turning on the rules to prevent using extensions and I can auto-fix them then.

test/common/model/fields.spec.js
jbalboni: Never mind, it is still used. I've updated and moved this test.

cvalarida (approved):

PR #6233, opened 2017-08-18
Fix bug with HCA array items

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4410

I can't reproduce the above error directly, but I think it's due to the fact that the review version of `ArrayField` doesn't account for array data changing outside of itself, which is a possibility. You can see that cause problems by changing the number of children on the HCA review page, then trying to edit their financial info. It jumps back to the old version that's in `ArrayField`'s state.

In addition, I made the `hideIf` function that's erroring to use `_.get`, so it won't throw an error now anyway.

### DIFF
--- a/src/js/common/schemaform/review/ArrayField.jsx
+++ b/src/js/common/schemaform/review/ArrayField.jsx
@@ -39,6 +39,20 @@ class ArrayField extends React.Component {
     this.isLocked = this.isLocked.bind(this);
   }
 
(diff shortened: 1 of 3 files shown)


cvalarida (approved):
"Woops! Forgot to add Johnny!"

Yeah, that looks good. Clever.

cehsu (approved):
This looks good to me.

PR #6234, opened 2017-08-18
Archive builds on S3 pt1

jkassemi:
* Uploads builds to `vetsgov-website-builds-s3-upload`
* Includes source ref in BUILD.txt
* All buildtypes built (so we can reliably deploy any build to any environment)

The last bullet has performance implications that may require scaling Jenkins nodes further (we'll need to keep a closer eye on them).


patrickvinograd (approved):
LGTM.

PR #6237, opened 2017-08-18
Auto-Load SiP / Pre-Filled Form

cvalarida:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3579

I tested this to make sure it works and I think it's golden (my unit tests say it is, too), but it could definitely use some more eyes than just mine, so please test this manually as well.

There are no visual changes.


test/util/unit-helpers.js
```diff
@@ -77,5 +79,12 @@ function fillDate(formDOM, partialId, dateString) {
   });
 }
 
+export function mockFetch(returnVal) {
+  global.fetch = sinon.stub().returns(returnVal);
+}
+
+export function resetFetch() {
```
cvalarida: I wrote these before I realized I didn't actually need them. Still, they might be helpful later. :man_shrugging:

test/common/schemaform/FormApp.unit.spec.jsx
```diff
@@ -128,7 +127,6 @@ describe('Schemaform <FormApp>', () => {
           currentLocation={currentLocation}
           loadedStatus={LOAD_STATUSES.pending}
           prefillStatus={PREFILL_STATUSES.pending}
-          isLoggedIn
```
cvalarida: These weren't used before. I actually added this prop (back) and it's only used in the auto-load bits.

src/js/common/schemaform/FormApp.jsx
```diff
@@ -102,6 +124,27 @@ class FormApp extends React.Component {
...
+        props.fetchInProgressForm(currentForm, props.formConfig.migrations, isPrefill);
+      } else {
+        // No forms to load; go to the beginning
+        // If the first page is not the intro and uses `depends`, this will probably break
+        props.router.push(firstPagePath);
+      }
+    } else {
+      // Can't load a form; go to the beginning
+      // If the first page is not the intro and uses `depends`, this will probably break
+      props.router.push(firstPagePath);
```
cvalarida: I feel like I can refactor this, but it's just one repeated line with comments for clarity, so I didn't want to complicate things too much.

src/js/common/schemaform/FormApp.jsx
```diff
@@ -41,19 +47,35 @@ class FormApp extends React.Component {
...
+      this.redirectOrLoad(this.props);
     }
   }
 
   componentWillReceiveProps(newProps) {
+    // When a user is logged in, the profile finishes loading after the component
+    //  has mounted, so we check here.
+    // If we're done loading the profile, check to see if we should load or redirect
+    if (this.props.profileIsLoading && !newProps.profileIsLoading && this.shouldRedirectOrLoad) {
+      this.redirectOrLoad(newProps);
```
cvalarida: Is it clear why this is being called in two places in the lifecycle?

src/js/common/schemaform/FormApp.jsx
```diff
@@ -102,6 +124,27 @@ class FormApp extends React.Component {
...
   }
 
+  redirectOrLoad(props) {
+    const firstPagePath = props.routes[props.routes.length - 1].pageList[0].path;
+
+    // If we're logged in and have a saved / pre-filled form, load that
+    if (props.isLoggedIn) {
+      const currentForm = props.formConfig.formId;
+      const isPrefill = props.prefillsAvailable.includes(currentForm);
+      if (props.savedForms.includes(currentForm) || isPrefill) {
```
cvalarida: @jbalboni Yeah, I dunno. I can see it go either way. I'm good with whatever, but to be thorough, I'd like to wrangle in UX for this.

@stacyhlinn The way this works right now is if a logged-in user has a pre-filled form but not a saved form, this will load their pre-filled form and put them on page 1 (after the intro). The alternative is landing them on the intro page and having them explicitly load their pre-filled form.

Which way should we go?

jbalboni: Ah, I was thinking of this incorrectly. If we're always redirecting to the last page for a saved/prefilled form, it matters less. I still think the intro is better, but it's up to @stacyhlinn.

cvalarida: Yeah, I mean, we just go wherever we'll be going when `fetchInProgressForm` finishes, so the last page they were on for saved forms, and the very first "real" page for a pre-fill page.

:man_shrugging:

src/js/common/schemaform/FormApp.jsx
```diff
@@ -102,6 +124,27 @@ class FormApp extends React.Component {
...
+        props.fetchInProgressForm(currentForm, props.formConfig.migrations, isPrefill);
+      } else {
+        // No forms to load; go to the beginning
+        // If the first page is not the intro and uses `depends`, this will probably break
+        props.router.push(firstPagePath);
```
jbalboni: What do you think about using replace instead of push here (and the other places)? If they click back after being redirected it's going to be surprising, since they won't have seen the page they're going back to.

src/js/common/schemaform/FormApp.jsx
jbalboni: I think so.

src/js/common/schemaform/FormApp.jsx
jbalboni: Also, this is always failing the first condition, because `savedForms` is an array of objects.

src/js/common/schemaform/FormApp.jsx
```diff
@@ -41,19 +47,35 @@ class FormApp extends React.Component {
...
+    if (!this.props.profileIsLoading && this.shouldRedirectOrLoad) {
+      this.redirectOrLoad(this.props);
     }
   }
 
   componentWillReceiveProps(newProps) {
+    // When a user is logged in, the profile finishes loading after the component
+    //  has mounted, so we check here.
+    // If we're done loading the profile, check to see if we should load or redirect
+    if (this.props.profileIsLoading && !newProps.profileIsLoading && this.shouldRedirectOrLoad) {
```
jbalboni: Should we reset `shouldRedirectOrLoad` once this condition is hit? It seems like this might get tripped inadvertently if the session expires and the user logs in again.

cvalarida: Probably safer, yeah.

src/js/common/schemaform/FormApp.jsx
cvalarida: Ahhh, good catch!

src/js/common/schemaform/FormApp.jsx
```diff
@@ -41,19 +47,35 @@ class FormApp extends React.Component {
...
-      this.props.router.push(firstPagePath);
+      // We started on a page that isn't the first, so after we know whether
+      //  we're logged in or not, we'll load or redirect as needed.
+      this.shouldRedirectOrLoad = true;
+    }
+  }
+
+  componentDidMount() {
+    // When a user isn't logged in, the profile finishes loading before the component mounts
+    if (!this.props.profileIsLoading && this.shouldRedirectOrLoad) {
```
jbalboni: Should this specifically check that they're not logged in?

src/js/common/schemaform/FormApp.jsx
```diff
@@ -41,19 +47,35 @@ class FormApp extends React.Component {
...
     // If we're in production, we'll redirect if we start in the middle of a form
-    // In development, we won't redirect unless we append the URL with `?redirect` ()
+    // In development, we won't redirect unless we append the URL with `?redirect`
     const devRedirect = __BUILDTYPE__ !== 'development' || this.props.currentLocation.search.includes('redirect');
     if (currentPath !== firstPagePath && devRedirect) {
-      // If the first page is not the intro and uses `depends`, this will probably break
-      this.props.router.push(firstPagePath);
+      // We started on a page that isn't the first, so after we know whether
+      //  we're logged in or not, we'll load or redirect as needed.
+      this.shouldRedirectOrLoad = true;
```
jbalboni: Do we need to check if SiP is enabled in some of the conditions in here?

cvalarida: Ahh! Yes. Yes we should.

src/js/common/schemaform/FormApp.jsx
cvalarida: Mmmm...I don't think so...I mean, the need to call `redirectOrLoad()` in two places is for whether the profile is done loading before the component mounts or not, not _really_ whether the user is logged in or not. It just happens that the profile will be done loading before the component mounts if the user isn't logged in.

src/js/common/schemaform/FormApp.jsx
cvalarida: Yeah...unless we need to pre-fill even when SiP is disabled, then we'd want to check in `redirectOrLoad()`.

Should we pre-fill when SiP is disabled? I'm not sure where we stand on that, actually.

src/js/common/schemaform/FormApp.jsx
```diff
@@ -102,6 +124,32 @@ class FormApp extends React.Component {
...
+
+    const firstPagePath = props.routes[props.routes.length - 1].pageList[0].path;
+
+    // If we're logged in and have a saved / pre-filled form, load that
+    if (props.isLoggedIn) {
+      const currentForm = props.formConfig.formId;
+      const isSaved = props.savedForms.some((savedForm) => savedForm.form === currentForm);
+      const isPrefill = props.prefillsAvailable.includes(currentForm);
+      const saveEnabled = !this.props.formConfig.disableSave;
+      if (saveEnabled && (isSaved || isPrefill)) {
```
cvalarida: I added the SiP check in here so we still redirect back to the beginning if the form doesn't have SiP enabled.

src/js/common/schemaform/FormApp.jsx
stacyhlinn: @jbalboni @cvalarida Taking the users to the intro page when they haven't started an application makes sense.

cehsu (approved):
Looks good to me.

jbalboni (approved):

PR #6238, opened 2017-08-18
HCA - Add future discharge date

cvalarida:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1869

image

image

@peggygannon Is that error message good? Do we need to change any other content?

image

### DIFF
--- a/src/js/hca/config/form.js
+++ b/src/js/hca/config/form.js
@@ -1,4 +1,5 @@
 import _ from 'lodash/fp';
+import moment from 'moment';
 
 import fullSchemaHca from 'vets-json-schema/dist/10-10EZ-schema.json';
 
@@ -361,11 +362,17 @@ const formConfig = {
             // TODO: this should really be a dateRange, but that requires a backend schema change. For now
             // leaving them as dates, but should change these to get the proper dateRange validation
             lastEntryDate: currentOrPastDateUI('Start of service period'),
-            lastDischargeDate: currentOrPastDateUI('Date of discharge'),
+            lastDischargeDate: {
+              'ui:title': 'Date of discharge',
+              'ui:widget': 'date'
+            },
             dischargeType: {
               'ui:title': 'Character of discharge',
+              // TODO: Use a constant instead of a magic string
+              'ui:required': (formData) => !moment(_.get('lastDischargeDate', formData), 'YYYY-MM-DD').isAfter(moment().startOf('day')),
               'ui:options': {
-                labels: dischargeTypeLabels
+                labels: dischargeTypeLabels,
+                hideIf: (formData) => moment(_.get('lastDischargeDate', formData), 'YYYY-MM-DD').isAfter(moment().startOf('day'))
               }
             },
             'ui:validations': [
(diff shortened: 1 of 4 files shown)


peggygannon:
@cvalarida Sorry, i thought this issue was blocked? Otherwise I would have provided message. 

Can we please edit the message to this: 
Discharge date must be after the service period start date and before December 16, 2017 (120 days from today).

cvalarida:
@peggygannon Yeah, it is blocked. I learned that after I already put up this PR. :man_shrugging:

I can make that change so it's ready to go when we're not blocked. I labeled the original ticket as blocked, but I should label this one too. Good point.

cvalarida:
Updated screenshot:

image

test/hca/config/serviceInformation.unit.spec.js
```diff
@@ -51,50 +51,37 @@ describe('Hca serviceInformation', () => {
...
+    formDOM.submitForm();
+    expect(formDOM.querySelectorAll('.usa-input-error').length).to.equal(0);
+    expect(onSubmit.called).to.be.true;
+  });
+
+  it('should hide and not require discharge status if discharge date is in the future', () => {
+    const onSubmit = sinon.spy();
+    const form = ReactTestUtils.renderIntoDocument(
+      <DefinitionTester
+          schema={schema}
```
cehsu: I think this indentation might trip the linter.

cehsu (approved):
This looks good to me. Had one minor syntax comment, perhaps merging with master (if that hasn't already been done) will make this easier to verify.

cvalarida:
Yeah, I'll have to merge master into this before merging this into master. Good call. 👍

cvalarida:
@rtluu / @J-Quag What's the status with this?

rtluu:
@cvalarida I believe this work is ready to be merged.

src/js/hca/config/form.js
```diff
@@ -361,11 +362,17 @@ const formConfig = {
...
             lastEntryDate: currentOrPastDateUI('Start of service period'),
-            lastDischargeDate: currentOrPastDateUI('Date of discharge'),
+            lastDischargeDate: {
+              'ui:title': 'Date of discharge',
+              'ui:widget': 'date'
+            },
             dischargeType: {
               'ui:title': 'Character of discharge',
+              // TODO: Use a constant instead of a magic string
+              'ui:required': (formData) => !moment(_.get('lastDischargeDate', formData), 'YYYY-MM-DD').isAfter(moment().startOf('day')),
```
cehsu: Is it necessary to include anything in .isAfter() for a future date? We use an alternative syntax in src/js/common/schemaform/SaveInProgressIntro.jsx on line 67.

cvalarida: I'm not sure. [The docs](https://momentjs.com/docs/#/query/is-after/) say that omitting the argument defaults to the current time, but what if the date is earlier in the same day? I think this is the right way to do it, but it's been so long ago I forget the exact reasons why.

Basically the argument in the `.isAfter()` just ensures that we're working with the beginning of the day instead of an arbitrary time in the middle of the day. Like I said, I don't remember why that was important, though.

PR #6241, opened 2017-08-21
Upgrade node-sass version to meet USWDS requirements

cehsu:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1463

No errors encountered when upgrading.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -5866,7 +5866,25 @@ node-pre-gyp@^0.6.29, node-pre-gyp@~0.6.32:
     tar "^2.2.1"
     tar-pack "^3.4.0"
 
-node-sass@^3.10.1, node-sass@^3.4:
+node-sass@3.4.2:
+  version "3.4.2"
+  resolved "https://registry.yarnpkg.com/node-sass/-/node-sass-3.4.2.tgz#ef61069927f1578ae51408ed60298449c4cdd294"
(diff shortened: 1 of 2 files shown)


yarn.lock
```diff
@@ -5866,7 +5866,25 @@ node-pre-gyp@^0.6.29, node-pre-gyp@~0.6.32:
     tar "^2.2.1"
     tar-pack "^3.4.0"
 
-node-sass@^3.10.1, node-sass@^3.4:
+node-sass@3.4.2:
```
jbalboni: The newest version of node-sass is 4.5.3. That version should get installed if you run `yarn upgrade node-sass`.

cehsu: It looks like USWDS requires "node-sass": "^3.4.2", but we are currently requiring "^3.10.1", and installing "3.13.1". This PR requires the version of node-sass that USWDS requires. Is it better to simply upgrade to the latest version?

jbalboni: Ah, I see. No, let's not get out of sync with uswds. This version is fine.

jbalboni (approved):

cvalarida (approved):
If we're going from 3.10.1 -> 3.4.2, it's not really an upgrade, is it? 😞  Still, best to not get out of sync just in case, eh?

PR #6242, opened 2017-08-21
Modularize wizard

cehsu:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4331


content/includes/header.html
```diff
@@ -34,6 +34,9 @@
 <meta content="{{ tag }}" property="article:tag">
 {% endfor %}
 {% endif %}
+{% if title == 'Education Benefits Application Process' or title == 'Education Benefits Eligibility' %}
+<link rel="import" href="/education/wizard/index.html">
```
jbalboni: Did you try to do this with the `include` directive from Metalsmith (or our markdown processor, not sure exactly where it's from)? We use it for a similar reusable component in the content pages, the sticky action box:

`{% include "content/includes/sticky-action-box.html" %}` (in 6-info-page.html).

cehsu: This would be really nice, but it doesn't look like it's possible to render liquid within [collections](https://github.com/jekyll/jekyll/issues/6211).

content/includes/header.html
```diff
@@ -34,6 +34,9 @@
 <meta content="{{ tag }}" property="article:tag">
 {% endfor %}
 {% endif %}
+{% if title == 'Education Benefits Application Process' or title == 'Education Benefits Eligibility' %}
```
cehsu: Thanks @jbalboni, great idea.

content/includes/header.html
jbalboni: I'm not entirely sure how that relates to `include` (that issue is also for Jekyll). What you're doing looks really similar to the way the sticky action box worked. It looks like you want to be able to include some common html on two different pages, which I think you can do just by putting `{% include ... }` on each of those pages, in the right spot.

cehsu: Thanks @jbalboni, using liquid tags doesn't work as expected. There are two types of includes syntax used in our content (with/without quotes):
using with quotes, yields:
screen shot 2017-08-22 at 9 37 40 am

using  without, yields:
screen shot 2017-08-22 at 9 37 53 am

The linked documentation was the only I could find to provide insight on why using liquid was not rendering in the collection pages.

content/includes/header.html
```diff
@@ -34,6 +34,9 @@
 <meta content="{{ tag }}" property="article:tag">
 {% endfor %}
 {% endif %}
+{% if includes_wizard %}
```
cvalarida: I hate to be so verbose, but just in case we need to make other wizards, should we be more specific and call this something like `includes_edu_wizard`?

src/js/edu-benefits/education-wizard.js
```diff
@@ -89,4 +89,14 @@ function reInitWidget() {
   resetForm();
 }
 
-document.addEventListener('DOMContentLoaded', reInitWidget);
+function loadImports() {
+  const link = document.querySelector('link[rel=import]');
```
cvalarida: Can we add an id to this `<link>` element? Again, hedging for the future when we may have multiple import links.

jbalboni:
@cehsu This is the alternative I came up with to using html import: https://github.com/department-of-veterans-affairs/vets-website/compare/modular-wizard-ceh...ms-include-jsb

I want to avoid webcomponents/html imports because it adds more page weight and adds another async call for something that we know ahead of time and can include in the page itself.

cvalarida (approved):
Ahhhh, this is much nicer, thank you! I was wondering how we might nicely include the wizard on only the pages we want to. I think this is a great way!

I just have a few suggestions to help future-proof it. Other than that, it looks great!

cvalarida:
I was thinking more along the lines of @jbalboni's branch--using a `script` tag instead of `link`. I like shying away from adding more dependencies.

cehsu:
Thanks @jbalboni, now I see what you meant, that's beautiful.

cvalarida (approved):

PR #6232, opened 2017-08-17
Letters copy edits 08/17/2017

goldenmeanie:
PL copy updates added. Will be easier for @peggygannon to review on staging.

References [#6527](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3947#issuecomment-323076527)

PR #6244, opened 2017-08-21
Letters copy edits 8/21/17

annekainicUSDS:
Same changes as in this PR: https://github.com/department-of-veterans-affairs/vets-website/pull/6232. Opened a new PR because the other one's git commit history got messed up at some point and I wasn't able to cleanly rebase on master.

References https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3947

### DIFF
--- a/src/js/letters/components/AddressSection.jsx
+++ b/src/js/letters/components/AddressSection.jsx
@@ -23,6 +23,9 @@ class AddressSection extends React.Component {
     } else {
       addressContent = (
         <div className="step-content">
+          <p>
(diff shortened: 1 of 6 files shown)


rroueche (approved):
lgtm

heads up @annekainicUSDS for the "address unavailable" portion, per gh issue #4135 that won't ever exist, since the whole page will be down if the address data is down. I'm fine to merge in like this though, and then that update will be taken care of when you get to that other gh issue later this week

PR #6245, opened 2017-08-21
Fix chrome tests on sauce

aub:
Addresses department-of-veterans-affairs/vets.gov-team#4167

1) Uploads don't work, so skip that
2) Use a fake login url that looks like a URL


bshyong (approved):
LGTM!

cehsu (approved):
Looks good to me.

PR #6247, opened 2017-08-21
Use month/year dates on education history pages

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4399#issuecomment-323433288

This affects the 1990 (rjsf), 1990e, and 5490. Education history and grad dates are now month/year dates. The backend date format isn't any different from what would happen if you filled them in now without a day, so there are no changes needed there.

screen shot 2017-08-21 at 3 02 02 pm
screen shot 2017-08-21 at 2 31 26 pm


cehsu (approved):
This looks good to me.

src/js/common/utils/validations.js
```diff
@@ -188,6 +188,11 @@ function isValidFutureDate(day, month, year) {
   return momentDate.isAfter(moment().endOf('day'), 'day');
 }
 
+function isValidCurrentOrFutureMonthYear(month, year) {
+  const momentDate = moment({ month: parseInt(month, 10) - 1, year });
+  return momentDate.isSameOrAfter(moment().endOf('month'), 'month');
```
cvalarida: Do we need the `.endOf('month')`?

cvalarida (approved):

src/js/common/utils/validations.js
jbalboni: Hmm, looks like I need that or the 'month' param, but maybe not both.

PR #6250, opened 2017-08-21
Fixes the sentry error

cvalarida:
There was a problem with refreshing on the confirmation page. We were getting a lot of `Cannot read property '0' of undefined` because the confirmation page wasn't passed `pageList`. This references the first page's `childRoutes` to get the path. It's a bit convoluted, but should be okay.

### DIFF
--- a/src/js/common/schemaform/FormApp.jsx
+++ b/src/js/common/schemaform/FormApp.jsx
@@ -61,7 +61,7 @@ class FormApp extends React.Component {
       newProps.setFetchFormStatus(LOAD_STATUSES.notAttempted);
     } else if (newProps.prefillStatus !== this.props.prefillStatus
       && newProps.prefillStatus === PREFILL_STATUSES.unfilled) {
-      newProps.router.push(newProps.routes[newProps.routes.length - 1].pageList[1].path);
+      newProps.router.push(newProps.routes[0].childRoutes[0].pageList[0].path);
     } else if (status !== LOAD_STATUSES.notAttempted
       && status !== LOAD_STATUSES.pending
       && status !== this.props.loadedStatus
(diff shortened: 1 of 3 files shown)


jbalboni (approved):
LGTM, but there's a build failure.

src/js/common/schemaform/FormApp.jsx
```diff
@@ -43,7 +43,7 @@ class FormApp extends React.Component {
 
     // If we start in the middle of a form, redirect to the beginning
     const currentPath = this.props.currentLocation.pathname;
-    const firstPagePath = this.props.routes[this.props.routes.length - 1].pageList[0].path;
```
cvalarida: This seems to be a more reliable way to get the path to the first page.

src/js/common/schemaform/FormApp.jsx
```diff
@@ -42,12 +42,11 @@ class FormApp extends React.Component {
     }
 
     // If we start in the middle of a form, redirect to the beginning
-    const currentPath = this.props.currentLocation.pathname;
-    const firstPagePath = this.props.routes[this.props.routes.length - 1].pageList[0].path;
     // If we're in production, we'll redirect if we start in the middle of a form
     // In development, we won't redirect unless we append the URL with `?redirect` ()
     const devRedirect = __BUILDTYPE__ !== 'development' || this.props.currentLocation.search.includes('redirect');
-    if (currentPath !== firstPagePath && devRedirect) {
+    if (isInProgress(this.props.currentLocation.pathname) && devRedirect) {
```
jbalboni: Might need to remove trailing `/`s?

jbalboni (approved):
LGTM, just had one question.

src/js/common/schemaform/FormApp.jsx
cvalarida: I just tested with and without the trailing `/` in the URL and it worked as expected in both cases. 🤔  It doesn't seem like it _should_ work with the trailing `/`. So that's odd...

PR #6252, opened 2017-08-21
Refactor updateProfileField

cvalarida:
I was looking through the login code today and figured I could take a quick shot at refactoring this so we didn't have to dispatch a thousand actions when logging in.

### DIFF
--- a/src/js/auth/containers/AuthApp.jsx
+++ b/src/js/auth/containers/AuthApp.jsx
@@ -4,7 +4,6 @@ import { connect } from 'react-redux';
 import environment from '../../common/helpers/environment.js';
 
 import { updateLoggedInStatus } from '../../login/actions';
-import { updateProfileFields } from '../../user-profile/actions';
 
 class AuthApp extends React.Component {
   constructor(props) {
@@ -111,10 +110,6 @@ const mapDispatchToProps = (dispatch) => {
   return {
     onUpdateLoggedInStatus: (update) => {
       dispatch(updateLoggedInStatus(update));
-    },
-    // Is this ever used?
-    onUpdateProfile: (field, update) => {
-      dispatch(updateProfileFields({ [field]: update }));
     }
   };
 };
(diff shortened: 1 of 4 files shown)


src/js/auth/containers/AuthApp.jsx
```diff
@@ -112,8 +112,9 @@ const mapDispatchToProps = (dispatch) => {
     onUpdateLoggedInStatus: (update) => {
       dispatch(updateLoggedInStatus(update));
     },
+    // Is this ever used?
     onUpdateProfile: (field, update) => {
```
cvalarida: I guess I didn't really need to keep the comment, but...is this ever used? It's the only instance of `onUpdateProfile` in the file and it's not being exported or anything, so...

Still, it seemed a bit out of scope to nix it in this PR.

src/js/user-profile/reducers/profile.js
```diff
@@ -35,8 +35,8 @@ const initialState = {
 
 function profileInformation(state = initialState, action) {
   switch (action.type) {
-    case UPDATE_PROFILE_FIELD: {
-      return _.set(action.propertyPath, action.value, state);
+    case UPDATE_PROFILE_FIELDS: {
+      return _.merge(state, action.newState);
```
jbalboni: I'm not sure merge is correct, I think you want to do `assign`. You could get old and new array items mixed together with merge, I think.

src/js/auth/containers/AuthApp.jsx
jbalboni: Yeah, it doesn't look used to me.

jbalboni (approved):
LGTM, but I think we should have @annekainicUSDS or @raquelromano-gov look at it, too.

src/js/user-profile/reducers/profile.js
cvalarida: Ohh, yeah. Good point. 👍

raquelromano-gov (approved):
LGTM--thanks!

PR #6256, opened 2017-08-22
Fix some more Sauce stuff

aub:
1) Allow the use of an identified tunnel
2) Use wait for element visible rather than expect visible
3) In selects for dates, use the value not the name of the option
4) Always use selectDropdown for selects
5) Fix selectDropdown for IE

Addresses department-of-veterans-affairs/vets.gov-team#4169
Addresses department-of-veterans-affairs/vets.gov-team#4168
Addresses department-of-veterans-affairs/vets.gov-team#4167

### DIFF
--- a/config/nightwatch-sauce.js
+++ b/config/nightwatch-sauce.js
@@ -31,7 +31,8 @@ module.exports = {
         recordScreenshots: true,
         recordLogs: true,
         captureHtml: true,
-        seleniumVersion: '3.4.0'
+        seleniumVersion: '3.4.0',
+        tunnelIdentifier: process.env.SAUCE_CONNECT_TUNNEL_IDENTIFIER || ''
       },
       selenium: {
         port: process.env.SELENIUM_PORT || 4444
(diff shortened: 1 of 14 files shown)


bshyong (approved):
LGTM!

raquelromano-gov:
@aub I'm late to this, but question about removing `selectDropdown` -- should this be completely removed from all tests? We have helper functions that still use it: https://github.com/department-of-veterans-affairs/vets-website/search?utf8=%E2%9C%93&q=selectDropdown&type=
If we don't want Sauce to run into this problem, should we remove it completely, including deleting
https://github.com/department-of-veterans-affairs/vets-website/blob/master/test/util/nightwatch-commands/selectDropdown.js

Also wondering what about that (admittedly weird) function doesn't play nicely with Sauce.

aub:
@raquelromano-gov on the contrary, this PR switches us to *always* use selectDropdown. It appears that setValue doesn't always work for selects for some reason that I don't entirely understand.

raquelromano-gov:
Ha...OK, clearly I browsed that too quickly. Thanks for the correction!

PR #6257, opened 2017-08-22
Letters error message update

annekainicUSDS:
Connects  to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4430
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4433

Decided to tackle 2 tickets in this PR because they both seemed small. Moving the IRIS content below the download letter button for the benefit summary letter (4433) turned out to be more complicated than I initially realized. The code in BSL had the download letter button always at the very end of the content, which means if I wanted to render it any other way I would have to refactor this.

There were 2 ways to handle this:
1. Importing `DownloadLetterLink` in `VeteranBenefitSummaryLetter`, so that I could place it in the middle of the content. This created an issue with one of the unit tests failing (something with the fact that `DownloadLetterLink` is a connected component and needs access to the store. When trying to render the whole component tree for the `VeteranBenefitSummaryLetter` unit test, since `DownloadLetterLink` is one of its children, it would fail saying `DownloadLetterLink` didn't have access to the store, even when I explicitly passed the store to it. This was the failure message I got: `Invariant Violation: Could not find "store" in either the context or props of "Connect(DownloadLetterLink)". Either wrap the root component in a <Provider>, or explicitly pass "store" as a prop to "Connect(DownloadLetterLink)"`. I couldn't figure out how to fix it, after trying several things).

So instead I did something hacky where I have the preliminary content for `VeteranBenefitSummaryLetter` in that component, then in `LetterList` I do the following:
```
<VeteranBenefitSummaryLetter/>
<DownloadLetterLink/>
<p>IRIS message content</p>
```

Not great because what this means is that all the content that's needed for the `VeteranBenefitSummaryLetter` is no longer contained within it, but it was the best way I could figure out without breaking tests.

Would love feedback on this, either on the solution or how to get that unit test to pass.

This is what is looked like before:
screen shot 2017-08-21 at 5 18 47 pm

This is what it looked like after:
screen shot 2017-08-21 at 4 54 18 pm

### DIFF
--- a/src/js/letters/components/LetterList.jsx
+++ b/src/js/letters/components/LetterList.jsx
@@ -12,34 +12,18 @@ class LetterList extends React.Component {
     const downloadStatus = this.props.letterDownloadStatus;
     const letterItems = (this.props.letters || []).map((letter, index) => {
       let content;
-
+      let bslHelpInstructions;
       if (letter.letterType === 'benefit_summary') {
-        content = (
-          <div>
-            <VeteranBenefitSummaryLetter/>
-            <DownloadLetterLink
-                letterType={letter.letterType}
-                letterName={letter.name}
-                downloadStatus={downloadStatus[letter.letterType]}
-                key={`download-link-${index}`}/>
-            <p>
-              If your service period or disability status information is incorrect, please send us
-              a message through VA’s <a target="_blank" href="https://iris.custhelp.com/app/ask">
-              Inquiry Routing & Information System (IRIS)</a>. VA will respond within 5 business days.
-            </p>
-          </div>
+        content = (<VeteranBenefitSummaryLetter/>);
+        bslHelpInstructions = (
+          <p>
+            If your service period or disability status information is incorrect, please send us
+            a message through VA’s <a target="_blank" href="https://iris.custhelp.com/app/ask">
+            Inquiry Routing & Information System (IRIS)</a>. VA will respond within 5 business days.
+          </p>
         );
       } else {
-        content = (
-          <div>
-            <div>{letterContent[letter.letterType] || ''}</div>
-            <DownloadLetterLink
-                letterType={letter.letterType}
-                letterName={letter.name}
-                downloadStatus={downloadStatus[letter.letterType]}
-                key={`download-link-${index}`}/>
-          </div>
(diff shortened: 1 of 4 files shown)


src/js/letters/components/LetterList.jsx
```diff
@@ -14,21 +14,39 @@ class LetterList extends React.Component {
...
+          </div>
+        );
       }
 
       return (
         <CollapsiblePanel
             panelName={letter.name}
             key={`collapsiblePanel-${index}`}>
           <div>{content}</div>
-          <DownloadLetterLink
```
raquelromano-gov: To avoid duplicating the DownloadLetterLink code, what if we left all of the above as-is, and then just before the `return` statement for this `map` callback, define something like
```
let bslHelpInstructions;
if (letter.letterType === 'benefit_summary) {
   bslHelpnstructions = (<p>If your service period or disability status information is incorrect, please send us a message through VA’s <a target="_blank" href="https://iris.custhelp.com/app/ask"> Inquiry Routing & Information System (IRIS)</a>. VA will respond within 5 business days.</p>);
}
```
and then after the download link add
```
{bslHelpInstructions}
```

raquelromano-gov:
In case you run into the above test failure again, `Invariant Violation: Could not find "store" in either the context or props of "Connect(DownloadLetterLink)". Either wrap the root component in a <Provider>, or explicitly pass "store" as a prop to "Connect(DownloadLetterLink)"`, I had run into this a while back and found that there were times I needed to test the component in isolation (not as a connected component) by explicitly passing it props. This issue comment explains how to do that: https://github.com/reactjs/redux/issues/2091#issuecomment-259754771 (we currently do that for some of our containers, like `VeteranBenefitSummaryLetter`)

src/js/letters/components/LetterList.jsx
annekainicUSDS: That's so much better! Will do that, thanks!

annekainicUSDS:
@raquelromano-gov Refactored per your suggestion. Should be ready for another review now!

raquelromano-gov (approved):

PR #6261, opened 2017-08-23
WIP new FAQ/signup page

bshyong:
Connected to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3504

I'm putting up this WIP PR for prelim feedback/questions/concerns. Key to note is that I made some changes in the `Main` login container to leverage the handleSignin/handleSignup methods for the buttons on the page. I also minimized the use of React vs. making the entire signup page a React component to facilitate SEO on the content.

screen shot 2017-08-22 at 5 10 02 pm

### DIFF
--- a/content/layouts/page-breadcrumbs.html
+++ b/content/layouts/page-breadcrumbs.html
@@ -4,7 +4,7 @@
 
 <div id="content" class="interior {{ id }}">
 
-  {% case template %}  
+  {% case template %}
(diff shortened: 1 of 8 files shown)


bshyong:
Update: the plan is to no longer have this page have the signin/login buttons. Instead, the FAQ page will be only static content. I will revise the PR accordingly.

PR #6262, opened 2017-08-23
Update indentation so that we can enable ESLint indent rules again

jbalboni:

I had to turn off the indentation rules because of our weird config, this changes everything back to two space indentation, which is the airbnb config default.


cehsu (approved):
Looks good to me.

U-DON (approved):
LGTM.

cvalarida (approved):

PR #6263, opened 2017-08-23
After reviewing current best practices reverting the br switch.

robbiethegeek-usds:


robbiethegeek-usds:
@bethpotts I stand corrected :)  https://developer.mozilla.org/en-US/docs/Web/HTML/Element/br 

``` 
<br /> - is only valid on XHTML 1.0 Strict doctype. 
```

phiden (approved):

PR #6264, opened 2017-08-23
Set up saved form list

cehsu:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3357
https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3272#issuecomment-309592280

TODO: 
- [x] Wire delete button up to removeInProgressForm
- [x] Refine responsive styles
- [x] Write tests (unit test for list, e2e tests for links, unit test for helper)

screen shot 2017-08-29 at 9 48 43 am

screen shot 2017-08-29 at 9 43 17 am

screen shot 2017-08-29 at 12 02 41 am

screen shot 2017-08-28 at 11 55 53 pm


src/js/user-profile/components/AuthApplicationSection.jsx
```diff
@@ -12,6 +14,49 @@ class AuthApplicationSection extends React.Component {
     let content;
     const verifiedAccountType = 3;// verified ID.me accounts are type 3
     const isVerifiedUser = this.props.userProfile.accountType === verifiedAccountType;
+    const savedForms = this.props.userProfile.savedForms;
+    const hasSavedForms = !!savedForms;
+    
+    //TODO:  (ceh) verify edu and burial form codes
+    const formTitles = {
```
cehsu: This feels like it should be broken down into two different helper objects and relocated to a helper file, but I didn't want to set up a file for just two helpers that aren't shared. Thanks for other thoughts on this.

cvalarida:
Before I forget, I'll make a comment here: If we're going to hard-code all the links in, we're going to need tests for all of them to make sure they link to where they should (to catch when we change the urls later).

src/js/user-profile/components/AuthApplicationSection.jsx
jbalboni: Yeah, I think we need to make this separate. I think we also want to figure out what the best way to tell people to keep this list up to date.

Off the top of my head: throw a console error when you load a form with a `formId` that isn't in this list and throw a Sentry error if you're on the profile page and there's a form in the list that doesn't have an entry.

@cvalarida Other ideas?

cvalarida: I'd like to use the data from the `formConfig`s themselves, but this might be tricky since they don't contain the `path`.

Either way, that still doesn't solve the problem of making sure this file contains all SiP-enabled forms. I think @jbalboni's idea of throwing an error up to Sentry is great as a just-in-case measure, but I'd like to prevent that from happening.

Now this might be overkill, but bear with me. `vets-json-schema/dist/schemas` has a list of all the schemas, so if we use the keys there as a master list, we can then use them in a unit test that checks to see if the form is SiP-enabled, and if so, that this list (`formTitles`) contains the key (of the form ID).

```js
import formConfig as formConfig1990 from '../path/to/config';
import formConfig as formConfig1995 from '../path/to/config';
import sipEnabledForms from '../path/to/helper/file';
import schemas from 'vets-json-schema/dist/schemas';

const allFormIds = Object.keys(schemas);
const configsArray = [formConfig1990, formConfig1995];

const configs = {};

// Transmogrify the array of configs into an object
configsArray.forEach((formConfig) => {
  configs[formConfig.formId] = formConfig;
});

allFormIds.forEach((formId) => {
  describe(`${formId}`, () => {
    // To make sure we're at least checking to see if it needs to be in the
    //  list of SiP-enabled forms
    it('should be in the configsArray', () => {
      expect(configs[formId]).to.not.be.undefined;
    });

    if (configs[formId] && !configs[formId].disableSave) {
      it('should be in the list of SiP-enabled forms', () => {
        expect(sipEnabledForms[formId]).to.not.be.undefined;
      });
    }
  });
});
```
When :point_up: fails, we know we need to either add a new form to the SiP-enabled list, disable saving on that form, or add it to this unit test to perform the necessary checks on it.

What say ye?

cehsu: Thanks @cvalarida, very clever idea!

cehsu: Thanks also @jbalboni, combined this makes for a great belt and suspenders approach.

src/js/common/helpers/login-helpers.js
```diff
@@ -12,11 +13,22 @@ export function handleVerify(verifyUrl) {
 }
 
 export function getUserData(dispatch) {
+  
   fetch(`${environment.API_URL}/v0/user`, {
     method: 'GET',
     headers: new Headers({
       Authorization: `Token token=${sessionStorage.userToken}`
     })
+  }).catch(res => {
```
jbalboni: I'm curious about the reasoning for adding this. It looks like it changes how this promise works a little and might be better at the end of the chain.

test/user-profile/e2e/user-profile.e2e.spec.js
```diff
@@ -2,24 +2,58 @@ const E2eHelpers = require('../../e2e/e2e-helpers');
...
+    // .waitForElementVisible('.usa-button-big', Timeouts.normal)
+    .axeCheck('document');
+
+  token = LoginHelpers.getUserToken();
+
+  LoginHelpers.logIn(token, client, '/profile', 3)
+    .waitForElementVisible('.react-container', Timeouts.slow)
+    .axeCheck('document');
+
+  client
```
jbalboni: What are all these for?

src/js/common/schemaform/save-load-actions.js
```diff
@@ -350,18 +364,30 @@ export function removeInProgressForm(formId, migrations) {
...
-      // If there’s some error when deleting, there’s not much we can
-      // do aside from not stop the user from continuing on
-      if (!res || res.status !== 401) {
-        window.dataLayer.push({
-          event: `${trackingPrefix}sip-form-start-over`
-        });
-        // after deleting, go fetch prefill info if they’ve got it
-        return dispatch(fetchInProgressForm(formId, migrations, true));
+
+      if (startOver) {
```
jbalboni: I wonder if a better approach would be to make another action in the profile application and not add the branching logic in here. I think this is somewhat confusing to read and there's not a whole lot of common code, outside of the initial `fetch` call. If we needed to, we could make a function for wrapping the initial fetch call and use it both here and in the action in the profile app.

src/js/user-profile/containers/UserProfileApp.jsx
```diff
@@ -1,40 +1,47 @@
...
-          userProfile={this.props.profile}
-          loginUrl={this.props.loginUrl}
-          verifyUrl={this.props.verifyUrl}>
-          {view}
-        </RequiredLoginView>
-      </div>
-    );
-  }
+
+ handleClick = (formId) => {
```
jbalboni: I think it would be better to call this something more specific than `handleClick` here and in the props in subsequent components. It makes it easier to understand when looking at some of the other components.

src/js/user-profile/helpers.jsx
```diff
@@ -0,0 +1,56 @@
...
+  '1010ez': 'health care (10-10EZ)',
+  '22-1990': 'education benefits (22-1990)',
+  '22-1990E': 'education benefits (22-1990E)',
+  '22-1990N': 'education benefits (22-1990N)',
+  '22-1995': 'education benefits (22-1995)',
+  '22-5490': 'education benefits (22-5490)',
+  '22-5495': 'education benefits (22-5495)'
+};
+
+const formLinks = {
```
jbalboni: Can we take off "introduction" from these links? They should redirect to the right place.

src/js/user-profile/components/AuthApplicationSection.jsx
```diff
@@ -42,6 +53,7 @@ class AuthApplicationSection extends React.Component {
 
     return (
       <div className="profile-section medium-12 columns">
+        {hasVerifiedSavedForms && <FormList className="sip-form-list" handleClick={this.handleClick} savedForms={verifiedSavedForms}/>}
```
jbalboni: Is the `className` prop used?

src/js/user-profile/components/FormList.jsx
```diff
@@ -0,0 +1,24 @@
...
+import FormItem from './FormItem';
+
+class FormList extends React.Component {
+  render() {
+    const handleClick = this.props.handleClick;
+    const forms = this.props.savedForms;
+    return (
+      <div>
+        <h4 className="section-header">Saved applications</h4>
+        {forms.map((form, index) =>  <FormItem key={index} savedFormData={form} handleClick={handleClick}/>)}
```
jbalboni: They `key` could be the form id, there's only one saved application per form.

src/js/user-profile/helpers.jsx
```diff
@@ -0,0 +1,56 @@
...
+const sipEnabledForms = new Set(['1010ez', '21P-527EZ', '21P-530']);
+
+export function handleNonSIPEnabledForm(formNumber) {
+  throw new Error(`Could not find form ${formNumber} in list of sipEnabledForms`);
+}
+
+export function handleIncompleteInformation(formNumber) {
+  Raven.captureMessage('vets_sip_list_item_missing_info');
+  window.dataLayer.push({
+    event: `${formNumber}sip-list-item-missing-info`
```
jbalboni: I think this would need to be `trackingPrefix` instead of `formNumber`

test/user-profile/e2e/user-profile.e2e.spec.js
cehsu: This is just the reformatted code from before.

src/js/common/schemaform/save-load-actions.js
cehsu: That's a better idea, thanks @jbalboni.

src/js/common/helpers/login-helpers.js
cehsu: This was added in the interest of being consistent, but if it's not achieving that goal, I will remove.

src/js/user-profile/containers/UserProfileApp.jsx
cehsu: Great point, I will change here and should I also change this in src/js/messaging/components/buttons/ButtonDelete.jsx?

src/js/user-profile/components/FormList.jsx
cehsu: Great idea, will change.

src/js/user-profile/components/AuthApplicationSection.jsx
cehsu: No, good catch, the styles now feature classes at the lowest possible levels per your comments elsewhere regarding nesting, it will be removed.

src/js/user-profile/helpers.jsx
cehsu: Yes, these will be removed.

src/js/user-profile/helpers.jsx
cehsu: Thanks, `formNumber` will be replaced with `trackingPrefix`.

src/js/user-profile/components/FormItem.jsx
```diff
@@ -0,0 +1,44 @@
...
+              <span className="saved-form-item-metadata">Last saved on {moment(lastSavedDate).format('MMMM D, YYYY')}.  </span>
+              <span className="saved-form-item-metadata saved-form-item-expires">Expires in {dateDiffDesc(expirationDate)}.</span>
+            </div>}
+          </div>
+          <div className="small-12 large-4 columns">
+            <a className="usa-button-primary resume-saved-application float-right" href={formLinks[formId]}>Resume application</a>
+          </div>
+        </div>
+        <div className="remove-saved-application-container">
+          <a className="remove-saved-application-link" onClick={() => {this.props.removeForm(formId);}}>
```
cvalarida: Should probably be a button styled like a link since there's no navigation, yeah?

src/js/user-profile/helpers.jsx
```diff
@@ -0,0 +1,56 @@
...
+  '22-1990N': '/education/apply-for-education-benefits/application/1990n/',
+  '22-1995': '/education/apply-for-education-benefits/application/1995/',
+  '22-5490': '/education/apply-for-education-benefits/application/5490/',
+  '22-5495': '/education/apply-for-education-benefits/application/5495/'
+};
+
+const sipEnabledForms = new Set(['1010ez', '21P-527EZ', '21P-530']);
+
+export function handleNonSIPEnabledForm(formNumber) {
+  throw new Error(`Could not find form ${formNumber} in list of sipEnabledForms`);
```
cvalarida: Are we capturing this error in Sentry and GA elsewhere?

Also, why throw an error?

src/js/user-profile/helpers.jsx
```diff
@@ -0,0 +1,56 @@
...
+}
+
+export function handleIncompleteInformation(trackingPrefix) {
+  Raven.captureMessage('vets_sip_list_item_missing_info');
+  window.dataLayer.push({
+    event: `${trackingPrefix}sip-list-item-missing-info`
+  });
+  return false;
```
cvalarida: Why return false?

src/js/user-profile/helpers.jsx
```diff
@@ -0,0 +1,56 @@
...
+
+export function isSIPEnabledForm(savedForm) {
+  const formNumber = savedForm.form;
+  if (!sipEnabledForms.has(formNumber)) {
+    handleIncompleteInformation(formNumber);
+  }
+  if (!formTitles[formNumber] || !formLinks[formNumber]) {
+    handleNonSIPEnabledForm(formNumber);
+  }
+  return true;
```
cvalarida: This will always return true...unless it throws an error, which is perhaps unexpected when I check to see if the form is SiP-enabled.

src/js/user-profile/helpers.jsx
```diff
@@ -0,0 +1,56 @@
...
+}
+
+export {
```
cvalarida: Can we export these inline like the functions?

test/user-profile/components/FormList.unit.spec.jsx
```diff
@@ -0,0 +1,51 @@
...
+        metadata: {
+          last_updated: '1503688891',  // eslint-disable-line camelcase
+          expires_at: '1503688891'  // eslint-disable-line camelcase
+        }
+      }
+    ]
+  };
+  let windowOpen;
+  let oldWindow;
+  const setup = () => {
```
cvalarida: Can we pull this out into a helper file so we don't have to duplicate this boilerplate elsewhere?

We'd just want to make it customizable on what the stub returns when you call it, but that should be easy.

cvalarida (commented):
I'm still pondering the tests; I'll get back to you on those.

src/js/user-profile/components/FormItem.jsx
cehsu: Excellent point. Thanks @cvalarida. Will change.

src/js/common/helpers/login-helpers.js
```diff
@@ -1,3 +1,4 @@
+import Raven from 'raven-js';
```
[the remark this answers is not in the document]
cehsu: Good catch, removed.

src/js/user-profile/helpers.jsx
cehsu: Nice catch, exported inline.

src/js/user-profile/components/FormList.jsx
```diff
@@ -0,0 +1,24 @@
...
+import FormItem from './FormItem';
+
+class FormList extends React.Component {
+  render() {
+    const removeForm = this.props.removeForm;
+    const forms = this.props.savedForms;
+    return (
+      <div className="profile-section medium-12 columns">
+        <h4 className="section-header">Saved applications</h4>
+        {forms.map((form) =>  <FormItem key={form.form} savedFormData={form} removeForm={removeForm}/>)}
```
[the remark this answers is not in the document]
cehsu: Thanks, good catch.

src/js/user-profile/helpers.jsx
cehsu: In order for `isSIPEnabledForm` to return false.

test/user-profile/helpers.unit.spec.js
```diff
@@ -0,0 +1,90 @@
...
+import fullSchema5490 from '../../src/js/edu-benefits/5490/config/form';
+import fullSchema5495 from '../../src/js/edu-benefits/5495/config/form';
+import fullSchema527EZ from '../../src/js/pensions/config/form';
+import fullSchema530 from '../../src/js/burials/config/form';
+import fullSchema10007 from '../../src/js/pre-need/config/form';
+
+import schemas from 'vets-json-schema/dist/schemas';
+
+// Maps schema id to config id
+const schemaToConfigIds = {
```
cehsu: While the keys on this object are used, the properties are essentially for visibility, as formIds are the basis of the removal logic, yet:
- schema ids and config ids do not align
- not all configs have a formId (e.g. pre-need)

test/user-profile/helpers.unit.spec.js
```diff
@@ -0,0 +1,90 @@
...
+      const configs = [
+        fullSchema1010ez,
+        fullSchema1990,
+        fullSchema1990e,
+        fullSchema1990n,
+        fullSchema1995,
+        fullSchema5490,
+        fullSchema5495,
+        fullSchema527EZ,
+        fullSchema530
```
cehsu: pre-need is not included here as it does not include a formId.

src/js/user-profile/helpers.jsx
```diff
@@ -0,0 +1,73 @@
...
+export const formTitles = {
+  '21P-527EZ': 'Veterans pension benefits (21P-527EZ)',
+  '21P-530': 'burial benefits (21P-530)',
+  '1010ez': 'health care (10-10EZ)',
+  '22-1990': 'education benefits (22-1990)',
+  '22-1990E': 'education benefits (22-1990E)',
+  '22-1990N': 'education benefits (22-1990N)',
+  '22-1995': 'education benefits (22-1995)',
+  '22-5490': 'education benefits (22-5490)',
+  '22-5495': 'education benefits (22-5495)'
```
cehsu: @peggygannon can you please confirm that the titles such as can be seen in the screen shot, and which are listed here, should be lower case as with the form introduction pages?

src/js/common/helpers/api.js
```diff
@@ -21,8 +21,8 @@ export function apiRequest(resource, optionalSettings = {}, success, error) {
     }
   };
 
-  const settings = merge(defaultSettings, optionalSettings);
-
+  // if settings are complete, no need to merge with default
+  const settings = (complete ? optionalSettings :  merge(defaultSettings, optionalSettings));
```
cehsu: The default setting for X-Key-Inflection is not required for form removal, so it is not included.

jbalboni: Does it break if it's included? I don't think it's bad practice to always send it.

src/js/user-profile/components/FormItem.jsx
```diff
@@ -0,0 +1,58 @@
...
+        <button className="va-alert-close notification-close va-expired-item-close" onClick={() => {this.props.toggleModal(formId);}} aria-label="Close notification">
+          <i className="fa fa-close" aria-label="Close icon"></i>
+        </button>
+        <div className="usa-alert-body">
+          <h5>Your saved application has expired</h5>
+          <p>The application you started for {formTitles[formId]} on {moment(lastSavedDate).format('MMMM D, YYYY')} has expired. If you would still like to apply, you will need to <a href={formLinks[formId]}>start a new application</a>.</p>
+        </div>
+      </div>
+    );
+    const view = (isExpired ? expiredView : activeView);
```
jbalboni: I think `return isExpired ? expiredView : activeView;` is clearer.

src/js/user-profile/components/FormList.jsx
```diff
@@ -0,0 +1,24 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+
+import FormItem from './FormItem';
+
+class FormList extends React.Component {
+  render() {
+    const toggleModal = this.props.toggleModal;
```
jbalboni: You can use destructuring here: `const { toggleModal, forms: savedForms } = this.props`

src/js/user-profile/containers/UserProfileApp.jsx
```diff
@@ -1,40 +1,84 @@
...
+    this.state = { modalOpen: false };
   }
+
+ removeForm = () => {
+   this.toggleModal();
+   this.props.removeSavedForm(this.state.formId);
+ }
+
+ toggleModal = (formId) => {
+   this.setState({ formId, modalOpen: !this.state.modalOpen });
```
jbalboni: You should be able to leave out `formId` here.

src/js/user-profile/containers/UserProfileApp.jsx
```diff
@@ -1,40 +1,84 @@
...
+   const savedForms = this.props.profile.savedForms;
+   const verifiedSavedForms = savedForms.filter(isSIPEnabledForm);
+   const hasVerifiedSavedForms = !!verifiedSavedForms.length;
+
+   const view = (
+     <div className="row user-profile-row">
+       <div className="usa-width-two-thirds medium-8 small-12 columns">
+         <h1>Your Account</h1>
+         <div>
+           {hasVerifiedSavedForms && <FormList
```
jbalboni: I think I would move as much as possible of the logic you added in this component this into your `FormsList` component (or make another component that wraps that one). This is the main component for the user profile, so it would be nice to pare it down to calling a single component with a few props for the change we're making. It keeps this component from getting to cluttered with logic from different features in the profile.

src/js/user-profile/helpers.jsx
```diff
@@ -0,0 +1,73 @@
...
+  window.dataLayer.push({
+    event: `${trackingPrefix}sip-list-item-missing-info`
+  });
+  return false;
+}
+
+export function isSIPEnabledForm(savedForm) {
+  const formNumber = savedForm.form;
+  if (!sipEnabledForms.has(formNumber)) {
+    return handleIncompleteInformation(formNumber);
```
jbalboni: I don't we need to track a GA event for this case, so maybe this code should just be:

```js
Raven.captureMessage('vets_sip_list_item_missing_info');
return false;
```

src/js/user-profile/helpers.jsx
```diff
@@ -0,0 +1,73 @@
...
+  }
+  if (!formTitles[formNumber] || !formLinks[formNumber]) {
+    return handleNonSIPEnabledForm(trackingPrefixes[formNumber]);
```
jbalboni: I think you can just throw in here, without the return.

src/sass/user-profile.scss
```diff
@@ -55,4 +55,52 @@
       margin-right: 5px;
     }
   }
+  
+  .saved-form-information {
+    margin: 0;
+    
+    div {
```
jbalboni: Can we give this div a class and use that in the styles?

src/sass/user-profile.scss
```diff
@@ -55,4 +55,52 @@
...
+      margin-bottom: 8px;
+    }
+    
+    span.saved-form-item-metadata {
```
jbalboni: Probably don't need the `span` part of this.

src/js/user-profile/components/FormItem.jsx
```diff
@@ -0,0 +1,58 @@
...
+import { formTitles, formLinks } from '../helpers';
+
+class FormItem extends React.Component {
+  render() {
+    const savedFormData = this.props.savedFormData;
+    const formId = savedFormData.form;
+    const { last_updated: lastSavedTime, expires_at: expirationTime } = savedFormData.metadata;
+    const lastSavedDate = moment.unix(lastSavedTime);
+    const expirationDate = moment.unix(expirationTime);
+    const isExpired = moment(expirationDate).isBefore(moment().startOf('minute'));
```
jbalboni: I don't think we need the `startOf('minute')` part. This is the (reversed) check in the intro component:

`moment.unix(f.metadata.expires_at).isAfter()`

jbalboni:
Locally I see the Resume button shifted left some:

screen shot 2017-08-29 at 3 56 04 pm

src/js/user-profile/components/FormItem.jsx
```diff
@@ -0,0 +1,58 @@
...
+        <div className="row small-collapse saved-form-information">
+          <div className="small-12 large-8 columns">
+            <h5 className="saved-form-title">Application for {formTitles[formId]}</h5>
+            {!!lastSavedDate && !!expirationDate && <div>
+              <span className="saved-form-item-metadata">Last saved on {moment(lastSavedDate).format('MMMM D, YYYY')}</span>
+              <span className="saved-form-item-metadata saved-form-item-expires">Expires in {dateDiffDesc(expirationDate)}</span>
+            </div>}
+          </div>
+          <div className="small-12 large-4 columns">
+            <a className="usa-button-primary resume-saved-application" href={formLinks[formId]}>Resume application</a>
```
jbalboni: This avoids the broken link issue, since we just put them on the intro page, but it would be a better experience to get them on the page they left off on without the extra click.

Maybe we should make a separate issue for that?

src/js/common/helpers/api.js
cehsu: It breaks and is commented it out in [`removeInProgressForm`](https://github.com/department-of-veterans-affairs/vets-website/blob/0d8b592f954dda22ca9bfa8157ff7018bdd6c8b4/src/js/common/schemaform/save-load-actions.js#L339).

src/js/user-profile/components/FormItem.jsx
cehsu: Good to know!

src/js/user-profile/components/FormItem.jsx
cehsu: Good idea, [created](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4537).

src/js/user-profile/containers/UserProfileApp.jsx
cehsu: Really? How should `removeForm` be fed the `formId`?

src/sass/user-profile.scss
cehsu: Specificity is being used here to override other styles.

src/sass/user-profile.scss
cehsu: Good idea.

src/js/user-profile/helpers.jsx
cehsu: Good point.

src/js/user-profile/helpers.jsx
cehsu: Sounds good.

src/js/user-profile/containers/UserProfileApp.jsx
jbalboni: I mean that I don't think this `setState` call will erase `formId` if you leave it out here. I could be wrong, though.

src/js/user-profile/components/FormItem.jsx
cehsu: Indeed, change made.

src/js/user-profile/components/FormList.jsx
cehsu: Thanks, change made.

src/js/user-profile/helpers.jsx
peggygannon: @cehsu Hi, Claire. Can I have more context about where these titles will be seen--will they appear in a save application message? Also, do you need me to provide content edits to the screen shots above?

cehsu:
Updated after @jbalboni catch:
screen shot 2017-08-29 at 1 48 51 pm

src/js/user-profile/helpers.jsx
cehsu: @peggygannon, these titles will be displayed in the user's profile, where they will be able to delete or resume their applications. Additional content edits would be greatly appreciated.

src/js/user-profile/containers/UserProfileApp.jsx
cehsu: I see, thanks. Change made.

src/js/common/helpers/api.js
cvalarida: Yeah, I remember that also being a problem in `fetchInProgressForm` because of back end Olive Branch reasons...I don't like this extra parameter to filter it out, though. Feels gross. 😕

cehsu: Should another version of this function be setup with different defaults? Then the other form related actions could also use it?

src/js/user-profile/helpers.jsx
peggygannon: @cehsu OK, thanks. Please make the form names lowercase to match the intro pages. 
Here are my suggested edits: 

Application for Veterans pension benefits (21P-527EZ)
Last saved on 8/29/2017 at [provide time?]  a.m. or p.m.
Your saved application will expire in 59 days. 

Button: Continue Your Application 

**Your saved application has expired**
Your application for Veterans pension benefits (21P-527EZ) has expired. If you want to still apply, please [start a new application].

cehsu: Thanks for the guidance, @peggygannon. Kindly refer to these updated screenshots.
screen shot 2017-08-29 at 4 05 09 pm

screen shot 2017-08-29 at 4 06 23 pm


screen shot 2017-08-29 at 4 05 58 pm

peggygannon: @cehsu Thanks, can you also update this line: 
--Your saved application will expire in 59 days.
--Add period after this sentence:...Your saved application has expired.

cehsu: Thanks again for the helpful pointers @peggygannon, kindly refer to the updated screenshots:

screen shot 2017-08-29 at 4 52 43 pm


screen shot 2017-08-29 at 4 53 04 pm


screen shot 2017-08-29 at 4 53 42 pm

peggygannon: @cehsu LGTM!

src/js/user-profile/helpers.jsx
```diff
@@ -0,0 +1,59 @@
...
+  '22-1990': '/education/apply-for-education-benefits/application/1990/',
+  '22-1990E': '/education/apply-for-education-benefits/application/1990e/',
+  '22-1990N': '/education/apply-for-education-benefits/application/1990n/',
+  '22-1995': '/education/apply-for-education-benefits/application/1995/',
+  '22-5490': '/education/apply-for-education-benefits/application/5490/',
+  '22-5495': '/education/apply-for-education-benefits/application/5495/'
+};
+
+export const trackingPrefixes = {
+  '21P-527EZ': 'pensions-527EZ-',
```
cehsu: These have been delinked from the formConfigs as the link was causing the following error:
```
There was an error while starting the test runner:

ReferenceError: location is not defined
    at Object.<anonymous> (/Users/clairehsu/Projects/vets-website/src/js/common/helpers/environment.js:5:31)
    at Module._compile (module.js:573:30)
    at loader (/Users/clairehsu/Projects/vets-website/node_modules/babel-register/lib/node.js:144:5)
    at Object.require.extensions.(anonymous function) [as .js] (/Users/clairehsu/Projects/vets-website/node_modules/babel-register/lib/node.js:154:7)
    at Module.load (module.js:507:32)
    at tryModuleLoad (module.js:470:12)
    at Function.Module._load (module.js:462:3)
    at Module.require (module.js:517:17)
    at require (internal/module.js:11:18)
    at Object.<anonymous> (/Users/clairehsu/Projects/vets-website/src/js/common/helpers/login-helpers.js:1:1)

npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! vets-website@1.0.0 nightwatch: `nightwatch -c config/nightwatch.js "test/user-profile/e2e/user-profile.e2e.spec.js"`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the vets-website@1.0.0 nightwatch script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/clairehsu/.npm/_logs/2017-08-30T05_23_56_794Z-debug.log
kill: usage: kill [-s sigspec | -n signum | -sigspec] pid | jobspec ... or kill -l [sigspec]
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! vets-website@1.0.0 test:e2e: `./script/run-nightwatch.sh "test/user-profile/e2e/user-profile.e2e.spec.js"`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the vets-website@1.0.0 test:e2e script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/clairehsu/.npm/_logs/2017-08-30T05_23_56_847Z-debug.log
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! vets-website@1.0.0 test: `npm run test:unit && npm run test:e2e "test/user-profile/e2e/user-profile.e2e.spec.js"`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the vets-website@1.0.0 test script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/clairehsu/.npm/_logs/2017-08-30T05_23_56_877Z-debug.log
```

src/js/user-profile/helpers.jsx
```diff
@@ -0,0 +1,59 @@
...
+  '22-5495': 'edu-5495-'
+};
+
+export const sipEnabledForms = new Set(['1010ez', '21P-527EZ', '21P-530']);
+
+export function handleNonSIPEnabledForm(formNumber) {
+  throw new Error(`Could not find form ${formNumber} in list of sipEnabledForms`);
+}
+
+export function handleIncompleteInformation() {
```
jbalboni: Do these two functions (this and `handleNonSIPEnabledForm`) need to be separate? I don't think they add much and it looked like they're only export for testing purposes?

test/user-profile/helpers.unit.spec.js
jbalboni: I think we can add an id for it.

src/js/user-profile/components/FormItem.jsx
```diff
@@ -0,0 +1,57 @@
...
+          </div>
+        </div>
+        <div className="remove-saved-application-container">
+          <button className="va-button-link" onClick={() => {this.props.toggleModal(formId);}}>
+            <i className="fa fa-trash"></i><span className="remove-saved-application-label">Delete</span>
+          </button>
+        </div>
+      </div>
+    );
+    const expiredView = (
```
jbalboni: Doesn't really affect your changes, but this view is really only ever going to be visible for an hour, where the form is expired but before the scheduled job removes it.

src/js/user-profile/containers/UserProfileApp.jsx
```diff
@@ -1,18 +1,39 @@
...
+    const savedForms = this.props.profile.savedForms;
+    const verifiedSavedForms = savedForms.filter(isSIPEnabledForm);
+    const hasVerifiedSavedForms = !!verifiedSavedForms.length;
+
     const view = (
       <div className="row user-profile-row">
         <div className="usa-width-two-thirds medium-8 small-12 columns">
           <h1>Your Account</h1>
           <div>
+            {hasVerifiedSavedForms && <FormList
```
jbalboni: Can you add a build type check here so it's not visible in production yet? Per https://dsva.slack.com/archives/C446YB1K8/p1504101843000103

jbalboni (approved):
Some more minor comments, but looks good after that.

test/user-profile/helpers.unit.spec.js
```diff
@@ -0,0 +1,85 @@
...
+      ];
+      const allFormIds = Object.keys(schemas);
+      const allMappedIds = Object.keys(schemaToConfigIds);
+      const sipEnabledConfigs = configs.filter(config => !config.disableSave);
+      const sipEnabledFormIds = sipEnabledConfigs.map(sipEnabledConfig => sipEnabledConfig.formId);
+      expect(_.isEqual(allFormIds, allMappedIds)).to.be.true;
```
jbalboni: You should be able to use something like `expect(allFormIds).to.equal(allMappedIds)`

test/user-profile/helpers.unit.spec.js
```diff
@@ -0,0 +1,85 @@
...
+import fullSchema530 from '../../src/js/burials/config/form';
+
+import schemas from 'vets-json-schema/dist/schemas';
+
+// Maps schema id to config id
+const schemaToConfigIds = {
+  '10-10EZ': '1010ez',
+  '21P-527EZ': '21P-527EZ',
+  '21P-530': '21P-530',
+  '22-1990': '1990',
```
jbalboni: I think the current ids for the education forms are made up, and should match the id in the schema file name.

src/js/user-profile/helpers.jsx
cvalarida: 🤔 How peculiar...

src/js/user-profile/components/FormItem.jsx
cehsu: Thanks @jbalboni, does it make sense to inform them of this should this wish to retrieve information from their saved form? @stacyhlinn @peggygannon could you please share your thoughts on this?

src/js/common/helpers/api.js
```diff
@@ -22,7 +22,37 @@ export function apiRequest(resource, optionalSettings = {}, success, error) {
...
+        if (response.status === 401) { return window.location.reload(); }
+        return data.then(Promise.reject.bind(Promise));
+      }
+
+      return data;
+    })
+    .then(success, error);
+}
+
+export function savedFormRequest(resource, optionalSettings = {}, success, error) {
```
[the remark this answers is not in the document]
cehsu: Great point, moved.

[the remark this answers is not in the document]
cehsu: Great point, reorganized.

src/js/user-profile/components/FormList.jsx
```diff
@@ -0,0 +1,55 @@
...
+
+import ProgressButton from '../../common/components/form-elements/ProgressButton';
+import Modal from '../../common/components/Modal';
+
+import FormItem from './FormItem';
+
+class FormList extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = { modalOpen: false };
```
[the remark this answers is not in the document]
cehsu: Sure thing, included.

src/js/user-profile/containers/UserProfileApp.jsx
cehsu: Good idea, added. @rtluu should we include removing this to the testing ticket created for this?

src/js/user-profile/helpers.jsx
cehsu: Good idea, changed. Kindly refer to updated screen shot below:
screen shot 2017-08-30 at 12 14 08 pm

src/js/user-profile/helpers.jsx
cehsu: Sounds good. These have been combined.

test/user-profile/helpers.unit.spec.js
cehsu: @jbalboni, just to confirm, is HCA the exception in that it should not match? In which case this key will still be required but will be updated with updated education formIds?

src/js/user-profile/components/FormItem.jsx
jbalboni: I don't think we can do much to change any of this right now, the backend would have to work differently.

test/user-profile/helpers.unit.spec.js
jbalboni: HCA is the exception in that the form id doesn't match the id in the schema file name (i.e. it isn't uppercase). All the other ids should match.

test/user-profile/helpers.unit.spec.js
cehsu: Sounds good, added.

test/user-profile/helpers.unit.spec.js
cehsu: Thanks again, @jbalboni. Sounds good, changed.

src/js/user-profile/helpers.jsx
cvalarida: Ahh, that looks great! 👍

src/js/user-profile/components/FormItem.jsx
stacyhlinn: @jbalboni We should somehow notify the user that their application has expired, either by keeping the yellow card, or including in the notification message that we're exploring separately.

stacyhlinn:
@jbalboni @cehsu @cvalarida 
We should keep the expired message indefinitely until
1. Users start a new application via the link in the yellow card
2. Users click the little "x" on the top right of the yellow card
3. While not signed in, users start a new application, signs in while in the form, and save.

I will write a separate ticket to update the application entry pages to include the expired message as well.

jbalboni:
@stacyhlinn We'll have to figure out how to do that on the backend, it's not possible with the way things work right now.

PR #6266, opened 2017-08-23
Update yarn.lock and upgrade to latest node-sass version

annekainicUSDS:
Ran `yarn install` to upgrade the `yarn.lock` file, and ran `yarn upgrade node-sass` to update latest version of node-sass.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -5411,7 +5411,7 @@ lodash.merge@^4.4.0:
   version "4.6.0"
   resolved "https://registry.yarnpkg.com/lodash.merge/-/lodash.merge-4.6.0.tgz#69884ba144ac33fe699737a6086deffadd0f89c5"
 
-lodash.mergewith@^4.0.0:
+lodash.mergewith@^4.0.0, lodash.mergewith@^4.6.0:
   version "4.6.0"
   resolved "https://registry.yarnpkg.com/lodash.mergewith/-/lodash.mergewith-4.6.0.tgz#150cf0a16791f5903b8891eab154609274bdea55"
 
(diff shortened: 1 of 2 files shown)


yarn.lock
```diff
@@ -6281,27 +6281,30 @@ node-pre-gyp@^0.6.29, node-pre-gyp@~0.6.32:
     tar "^2.2.1"
     tar-pack "^3.4.0"
 
-node-sass@3.4.2:
-  version "3.4.2"
-  resolved "https://registry.yarnpkg.com/node-sass/-/node-sass-3.4.2.tgz#ef61069927f1578ae51408ed60298449c4cdd294"
+node-sass@^3.4:
+  version "3.13.1"
```
annekainicUSDS: This is odd... The package.json file clearly was updated with the latest version of node-sass, but here it went back to using an older version. Not sure what's going on here @jbalboni

jbalboni: Maybe try `yarn remove node-sass`, then updating it again? It looks like a duplicate for not reason.

annekainicUSDS: didn't work for me :/ still stuck on node-sass version 3.13.1. I did the following commands:
```
yarn remove node-sass
yarn upgrade node-sass
yarn install
```

annekainicUSDS: Looks like this is related to bourbon-neat. While the dependency for bourbon-neat specifies `node-sass "^3.4"`, node-sass v3.13.1 gets introduced during install here: `> node-sass@3.13.1 install /Users/vacokainia/Desktop/coding/vets-website/node_modules/bourbon-neat/node_modules/node-sass`. When looking at the version of node-sass that's included in the node-modules under bourbon-neat, it is indeed 3.13.1. Not sure how to reset that, as we're not installing bourbon-neat directly (I think it's installed as a dependency of something else), so running `yarn remove bourbon-neat` to reset it doesn't work.

jbalboni: Well, that's annoying. However, it does appear that 4.5.3 is the version going in `node_modules/node-sass`, and since this does make the deprecation errors go away and doesn't appear to cause other problems, I think it's ok.

jbalboni (approved):

yarn.lock
annekainicUSDS: Okay great, sounds good!

PR #6267, opened 2017-08-23
Gibs accessibility improvements

annekainicUSDS:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3973

This PR fixes the couple of small accessibility issues that were identified in the above ticket. There is a larger accessibility issue of fixing the print page that was discovered during the course of this work, but going to reserve that for a separate PR so we can at least get these couple of smaller changes in.

### DIFF
--- a/src/js/letters/containers/VeteranBenefitSummaryLetter.jsx
+++ b/src/js/letters/containers/VeteranBenefitSummaryLetter.jsx
@@ -77,7 +77,7 @@ export class VeteranBenefitSummaryLetter extends React.Component {
                 type="checkbox"
                 onChange={this.handleChange}/>
             </th>
-            <td><label id={`${key}Label`}>{optionText}</label></td>
+            <td><label id={`${key}Label`} htmlFor={key}>{optionText}</label></td>
           </tr>
         );
       }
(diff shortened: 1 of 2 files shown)


raquelromano-gov (approved):
LGTM

Aside: I have been meaning to ask vetsgov product leads if we can add an explicit line to the launch checklist for a front-end dev or designer to go through the entire product with a screenreader (OS X / VoiceOver fine) to manually test for any accessibility issues.

Looking at https://github.com/department-of-veterans-affairs/vets.gov-team/blob/master/Work%20Practices/Product%20Management/Product_Launch_Guide.md, there is a line, "Tested with screen reader software for accessibility and 508 office" which is vague. We do not currently block launch if the 508 office hasn't tested it--they are going through a backlog of launched products and flagging issues for us that we file tickets for, but because we have automated testing, we have permission to launch before the 508 office does manual QA. However, automatic testing will never catch anything, so manual QA should still be a launch requirement.

Obviously this topic extends far beyond this PR! Just writing it here because it's on my mind--I will cut/paste it elsewhere to bring up with the product team. :)

PR #6269, opened 2017-08-23
Add SiP alert for logged in with no saved form

cvalarida:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4329

image
image


src/js/common/schemaform/SaveInProgressIntro.jsx
```diff
@@ -32,13 +32,24 @@ export default class SaveInProgressIntro extends React.Component {
             <br/>
           </div>
         );
+      } else {
```
jbalboni: I wonder if we should flatten these two if statements to make it clearer:

```js
if (this.props.user.login.currentlyLoggedIn && savedForm) {
...
} else if (this.props.user.login.currentlyLoggedIn) {
...
} else {
...
}
```

jbalboni (approved):
One nitpick, LGTM

cehsu (approved):
This looks good to me. Are those prop errors due to the eslint upgrade?

PR #6272, opened 2017-08-23
Fixed folder names in dropdown menu

U-DON:
Fixed a bug that resulted from #6212. It was changing the actual displayed folder name in the dropdown when the intent was just to change the HTML id on the label and input pair.


bshyong (approved):
LGTM! Thanks for catching this.

PR #6273, opened 2017-08-24
[WIP] 1047 modals

phiden:
So many modals, so many styles. Trying to consolidate them to match what's in our Sketch styleguide.


phiden:
blocked -- mhv isn't playing nice with review instances; can't verify the work.

content/includes/footer.html
```diff
@@ -65,12 +65,19 @@
...
-        <h4 class="va-crisis-panel-body-title">Get help from Veterans Crisis Line</h4>
+<div id="modal-crisisline" class="va-overlay va-modal va-modal-large" role="alertdialog">
+    <div class="va-crisis-panel va-modal-inner">
+
+      <h3>Get help from Veterans Crisis Line</h3>
+
+      <button class="va-modal-close va-overlay-close" type="button">
+        <i class="fa fa-close"></i>
+        <span class="usa-sr-only">Close this modal</span>
+      </button>
```
ncksllvn: I'm noticing that if you click on the "x", and not the surrounding button area, that the modal doesn't close because the element that the event is originating doesn't contain the `va-overlay-close` class. The click handler in src/js/legacy/menu.js either needs to be refactored or you can add that `va-overlay-close` to all of the elements in the button.

robertfairhead: Ahh. i slacked sophia about that problem yesterday but didn't debug as thoroughly as you 😬

phiden: yeah @ncksllvn making my life easier -- thanks! will adjust.

robertfairhead (changes requested):
in addition to the close button not closing on the crisis line modal, I noticed one other problem.

The login modal is opening with the close button in the "active state":

screen shot 2017-09-27 at 1 43 59 pm

Close up:

screen shot 2017-09-27 at 1 43 53 pm

I got a 500 on the review instance when i tried to login to use the claims and Rx modals. Is there a path test that that works other than running locally?

phiden:
@robertfairhead are you using the Heroku instance or this one? http://c6ed27e8d19632dfc5ed82f455edb00e.review.vetsgov-internal/ <--the one I've been using to test

src/js/common/components/OMBInfo.jsx
```diff
@@ -4,6 +4,7 @@ import Modal from './Modal';
...
   </div>
```
ncksllvn: This looks good! Unrelated to modals, but is there no link to `burials-and-memorials/burial-planning/application/introduction` on the site? I couldn't find one.

phiden: I dont' think there is at this point -- that's a question for the content team, ultimately

robertfairhead:
Yeah that review instance on internal. When I tried to login via IDme it goes through the flow, 2FA then i get:

screen shot 2017-09-27 at 1 46 55 pm

phiden:
@robertfairhead my guess is that's related to the new identity flow that Kudos has been testing, and not so much to the modals. Maybe @saneshark has some insight?

robertfairhead:
yeah not blaming the PR for it, just saying i wasn't able to check the Rx and Claims modals visually.

ncksllvn:
On my localhost, the Privacy Act Statement on the Burials Application also opens in the focused state as well

ncksllvn:
Actually that issue with the close button in focus I see happening on the production site as well - https://www.vets.gov/burials-and-memorials/application/530/introduction. Is this maybe intentional, something for accessibility?

phiden:
@ncksllvn that's a really good question; I don't actually know. But since that's how it is in production, let's leave it as-is until we hear otherwise.

src/js/rx/components/GlossaryList.jsx
```diff
@@ -9,21 +9,32 @@ class GlossaryList extends React.Component {
...
+
+    if (termsList.length === 2) {
+      return (
+        <div>
+          <h3 className="modalTitle">Glossary: {termsList[0]}</h3>
+          {termsList[1]}
+        </div>
+      );
+    }
     return (
```
ncksllvn: Just to confirm, there's no scenario that `termsList` could wind up as an empty array, right?

src/js/rx/components/GlossaryModal.jsx
```diff
@@ -21,14 +21,12 @@ class GlossaryModal extends React.Component {
...
-          <div className="va-modal-button-group cf">
-            <button type="button" onClick={this.handleCloseModal}>Close</button>
-            <Link
-              to="/glossary"
-              onClick={this.props.onCloseModal}>
-              See all status definitions
-            </Link>
-          </div>
+          <button className="va-modal-close" type="button" onClick={this.handleCloseModal}><i className="fa fa-close"></i><span className="usa-sr-only">Close this modal</span></button>
+          <Link
```
ncksllvn: This is picky and you don't need to worry about changing it unless you want to, but when lines get real long like this it can be a little easier to read by breaking it into separate lines

ncksllvn:
Okay, final thing! It may be worth getting confirmation on the close buttons autofocusing like we're seeing. I say that because I would think that the first focusable actual content within the modal would be what's auto-focused and now that's the close button so we may have to change something. 

If that's not intended, I see the Modal components accepts a prop called [`focusSelector`](https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/js/common/components/Modal.jsx#L105) so we'll either have to start passing that prop or update that base class. 

Otherwise, great job!

src/js/rx/components/GlossaryList.jsx
phiden: Nope, none. It's being fed by a predefined list.

phiden:
@robertfairhead fixed the crisis line button; would you mind reviewing again please?

robertfairhead (approved):

PR #6279, opened 2017-08-25
Mark JS assets as async and defer where appropriate

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2839

We have JS assets that can be marked as async and defer so that the static content pages load faster. It has a minor impact on the React app pages, since those pages are rendered with JS.

For the home page, throttled to Good 3G.

Before:
screen shot 2017-08-25 at 10 07 13 am
screen shot 2017-08-25 at 9 37 48 am

We're not putting any content on the screen for around 5 seconds, just because we're waiting for JS that only renders the menus. 

After:
screen shot 2017-08-25 at 10 07 20 am
screen shot 2017-08-25 at 9 32 37 am

Deferring our JS assets means that the page doesn't stop rendering to wait for the scripts to load, so we're able to render most of the content. The effect of this is that users see a page with the content rendered, but the header is missing the menus, starting 3 seconds in. I think this is better, since people can start reading the page they're on sooner.

The next step after this would be to attempt to pare down style.css. It's pretty big and is now the main bottleneck for rendering on the static pages.

Note: I chose `async` for some scripts that are external and don't require a particular loading order. Our JS bundles can't use `async` because they need to run in a specific order, which `async` doesn't guarantee.

### DIFF
--- a/content/includes/header.html
+++ b/content/includes/header.html
@@ -68,21 +68,21 @@
 <link rel='stylesheet' href='/assets/css/ie.css'>
 <![endif]-->
 
-<!-- We participate in the US government’s analytics program. See the data at analytics.usa.gov. -->
-<script async type="text/javascript" src="https://dap.digitalgov.gov/Universal-Federated-Analytics-Min.js?agency=VA" id="_fed_an_ua_tag"></script>
-
 <script>
 //<![CDATA[
 window.webpackManifest = 'CHUNK_MANIFEST_PLACEHOLDER';
 //]]>
 </script>
 
-<script src="/generated/vendor.js"></script>
+<script defer src="/generated/vendor.js"></script>
 {% if entryname %}
-<script src="/generated/{{ entryname }}.entry.js"></script>
+<script defer src="/generated/{{ entryname }}.entry.js"></script>
 {% else %}
-<script src="/generated/no-react.entry.js"></script>
+<script defer src="/generated/no-react.entry.js"></script>
 {% endif %}
+
+<!-- We participate in the US government’s analytics program. See the data at analytics.usa.gov. -->
+<script async type="text/javascript" src="https://dap.digitalgov.gov/Universal-Federated-Analytics-Min.js?agency=VA" id="_fed_an_ua_tag"></script>
 
 </head>
 <body class="{{ body_class }}">
@@ -151,4 +151,4 @@ window.webpackManifest = 'CHUNK_MANIFEST_PLACEHOLDER';
 
 </header>
 
-<script src="/js/usa-search.js"></script>
+<script async src="/js/usa-search.js"></script>
--- a/src/sass/base/_va.scss
+++ b/src/sass/base/_va.scss
@@ -196,9 +196,11 @@ hr {
   clear: both;
   margin: 0;
   padding: 0;
+  min-height: 147.19px;
 
   @include media($medium-screen) {
     position: relative;
+    min-height: 168.5px;
   }
 
 .login-container {
--- a/content/includes/footer.html
+++ b/content/includes/footer.html
@@ -92,7 +92,7 @@ visit <a href="https://www.veteranscrisisline.net/">VeteransCrisisLine.net</a> f
   {% include "content/includes/modal.html" %}
 {% endif %}
 
-<script src="/js/user-voice.js"></script>
+<script async src="/js/user-voice.js"></script>
 
 <!-- htmllint attr-bans="[]" -->
 {% if buildtype === 'development' %}
--- a/src/js/legacy/mega-menu.js
+++ b/src/js/legacy/mega-menu.js
@@ -146,6 +146,7 @@ class MegaMenu {
 export default MegaMenu;
 
 function reInitMenu() {
+  console.log('menu');
   const menuElement = document.querySelector('#vetnav');
   const openMenuElement = document.querySelector('.vetnav-controller-open');
   const closeMenuElement = document.querySelector('.vetnav-controller-close');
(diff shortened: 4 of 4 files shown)


content/includes/header.html
```diff
@@ -151,4 +151,4 @@
 
 </header>
 
-<script src="/js/usa-search.js"></script>
+<script async src="/js/usa-search.js"></script>
```
jbalboni: This script still causes problems because it synchronously loads another script that has a permanent redirect and ends up delaying things by 613ms, according to Chrome's audit.

src/sass/base/_va.scss
```diff
@@ -196,9 +196,11 @@ hr {
   clear: both;
   margin: 0;
   padding: 0;
+  min-height: 147.19px;
```
jbalboni: Reduces jankiness when the menus pop in, needs some testing in other browsers still.

raquelromano-gov: Yeah, curious to see how/whether things jump around unpredictably depending on the device, but sounds like you're looking into it.  If you need people to QA on various devices once the change is on staging, maybe put out a request on #vetsdotgov

cehsu: Should we create a sign up sheet for this? A markdown table in the description where folks can add their name and indicate if things look ok on a specific browser?

jbalboni: So far, testing on BrowserStack, it seems ok. I think zoom levels might affect it, too, so I'm going to play around with that some.

cehsu (approved):
Looks good to me. How can we help test? (Which browsers/configurations?)

src/sass/base/_va.scss
jbalboni: @cehsu I think we're good. I was able to test our supported browsers without any trouble. It's the same on every page, so there are not a lot of things to test.

content/includes/header.html
robbiethegeek-usds: Did you check to see if the login / signup links we be functional on a slower connection?  It is the one thing I am worried about because we fetch the id.me url on load.

content/includes/header.html
```diff
@@ -68,22 +68,22 @@
...
-<script async type="text/javascript" src="https://dap.digitalgov.gov/Universal-Federated-Analytics-Min.js?agency=VA" id="_fed_an_ua_tag"></script>
-
 <script>
 //<![CDATA[
 window.webpackManifest = 'CHUNK_MANIFEST_PLACEHOLDER';
 //]]>
 </script>
 
-<script src="/generated/vendor.js"></script>
+<script defer src="/generated/vendor.js"></script>
```
cehsu: Do we need to worry about IE <= 9  support for these?

content/includes/header.html
raquelromano-gov: 2 separate discussions here:
1) usa-search.js: @jbalboni, can you open a ticket and assign to me to look into contacting the folks who maintain this? Maybe we can get them to asynchronously load that second script

2) login/signup links (and anything else we're deferring): maybe one more manual QA task to try this out on a slow connection, and if fine, mention that in the PR description (which I will now go re-read in case that's already been done)

raquelromano-gov:
Wondering what your rule of thumb is for using `defer` vs `async` (maybe something to add to the PR description).

content/includes/header.html
jbalboni: No, we support IE11 and up, and display a warning if you use anything older.

cvalarida: Our minimum supported version of IE is 11, so no. Worst case scenario, somebody with IE <= 9 opens it up and they load synchronously, right? (I don't actually know that it won't break things, but I think it's a safe assumption.)

jbalboni:
@raquelromano-gov Updated the PR description.

content/includes/header.html
jbalboni: @raquelromano-gov 

1. Sure. Thinking about it more, I'm not positive that changing it to async is the right course, but we can ask what they think.
2. I've tried it with Chrome's throttling and it works fine for me. This change is unlikely to affect that, since all of the login code is in React and runs after DOMContentLoaded whether the scripts use `defer` or not.

bshyong (approved):
LGTM!

content/includes/header.html
robbiethegeek: Thanks for checking @jbalboni

cvalarida:
@raquelromano-gov Looks like [this page](https://peter.sh/experiments/asynchronous-and-deferred-javascript-execution-explained/) gives a pretty good explanation of the difference between `defer` and `async`.

cvalarida (approved):
Didn't do any manual testing, to be honest, but it _looks_ good.

src/sass/base/_va.scss
```diff
@@ -196,9 +196,11 @@ hr {
...
 
   @include media($medium-screen) {
     position: relative;
+    min-height: 168.5px;
```
U-DON: Will these min-heights still apply if we ever add content to the header in the future? Just wondering how these values were chosen.

U-DON (approved):
LGTM.

src/sass/base/_va.scss
jbalboni: They'll have to be changed if we change the design of the header. I picked them because they're the final height in the browser, after everything has loaded. It's been pretty stable for a while, and doesn't affect content added in menus, so I don't think it'll cause much of a maintenance burden.

raquelromano-gov (approved):

PR #6285, opened 2017-08-25
Update facility locator test data and fix accessibility issues

bshyong:
- updates test data for FL tests
- fixes accessibility issues for facility locator

Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3278

### DIFF
--- a/src/js/facility-locator/components/AppointmentInfo.jsx
+++ b/src/js/facility-locator/components/AppointmentInfo.jsx
@@ -43,11 +43,11 @@ export default class AppointmentInfo extends Component {
       return null;
     }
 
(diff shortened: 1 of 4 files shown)


U-DON (approved):
LGTM.

I'm not sure having a header like "Specialty care" or interactive button like "Show more/less" on the same level as the times is exactly semantic. Also unsure of the accessibility issue that was solved.

bshyong:
@U-DON accessibility rules are strict about only allowing <li> elements as children inside <ul> elements. I think there are multiple solutions here but I went with the simplest one

PR #6287, opened 2017-08-28
Fetch GA client ID and append it to id.me log in & out urls

omgitsbillryan:
Connects: https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3440

Append the Google Analytics client ID as a query parameter onto all login and logout ID.me URLs

### DIFF
--- a/src/js/common/helpers/login-helpers.js
+++ b/src/js/common/helpers/login-helpers.js
@@ -81,13 +81,19 @@ export function handleLogin(loginUrl, onUpdateLoginUrl) {
   return Promise.reject('Could not log in; loginUrl not provided.');
 }
 
+function isGaLoaded() {
+  return !!(window.ga && ga.create);
+}
+
 // google analytics client Id
 /* global gaClientId ga:true */
 export function gaClientId() {
   let clientId;
-  for (const data of ga.getAll()) {
-    if (data.get('cookieDomain') === 'vets.gov') {
-      clientId = data.get('clientId');
+  if (isGaLoaded()) {
+    for (const data of ga.getAll()) {
+      if (data.get('cookieDomain') === 'vets.gov') {
+        clientId = data.get('clientId');
+      }
     }
   }
   return clientId;
--- a/src/js/common/helpers/login-helpers.js
+++ b/src/js/common/helpers/login-helpers.js
@@ -80,3 +80,15 @@ export function handleLogin(loginUrl, onUpdateLoginUrl) {
   }
   return Promise.reject('Could not log in; loginUrl not provided.');
 }
+
+// google analytics client Id
+/* global gaClientId ga:true */
+export function gaClientId() {
+  let clientId;
+  for (const data of ga.getAll()) {
+    if (data.get('cookieDomain') === 'vets.gov') {
+      clientId = data.get('clientId');
+    }
+  }
+  return clientId;
+}
--- a/src/js/login/reducers/login.js
+++ b/src/js/login/reducers/login.js
@@ -1,4 +1,6 @@
 import _ from 'lodash/fp';
+import { gaClientId } from '../../common/helpers/login-helpers';
+import appendQuery from 'append-query';
 
 import {
   UPDATE_LOGGEDIN_STATUS,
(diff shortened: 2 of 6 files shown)


src/js/common/helpers/login-helpers.js
```diff
@@ -80,3 +80,15 @@ export function handleLogin(loginUrl, onUpdateLoginUrl) {
   }
   return Promise.reject('Could not log in; loginUrl not provided.');
 }
+
+// google analytics client Id
+/* global gaClientId ga:true */
+export function gaClientId() {
+  let clientId;
+  for (const data of ga.getAll()) {
```
jbalboni: We load the GA script with `async`, so what happens if GA finishes loading after our main JS bundles? That's very unlikely, but possible (I think).

omgitsbillryan: That's an excellent point.  I found, [this stack overflow post](https://stackoverflow.com/questions/1954910/javascript-detect-if-google-analytics-is-loaded-yet) on the subject.  Let me know if you have a preference as to which solution - otherwise standby for another commit to address this point.

robertfairhead: You may also want to consider if there is not cookie (e.g., the user has turned off cookies). Also, some plugins like Ghostery prevent GA from ever loading.

There will be clients that **never** have a `ga` object or `clientId`

src/js/common/helpers/login-helpers.js
```diff
@@ -80,3 +80,21 @@ export function handleLogin(loginUrl, onUpdateLoginUrl) {
...
+
+function isGaLoaded() {
+  return !!(window.ga && ga.create);
+}
+
+// google analytics client Id
+/* global gaClientId ga:true */
+export function gaClientId() {
+  let clientId;
+  if (isGaLoaded()) {
```
omgitsbillryan: @jbalboni @robertfairhead I _think_ this is sufficient.  We expect that GA is already loaded 99.9% of the time, but this safeguard should prevent any kind of `Uncaught ReferenceError: ga is not defined` JS error.s

src/js/login/reducers/login.js
```diff
@@ -35,13 +37,13 @@ function loginStuff(state = initialState, action) {
       return _.set('currentlyLoggedIn', action.value, state);
 
     case UPDATE_LOGIN_URL:
-      return _.set('loginUrl', action.value, state);
+      return _.set('loginUrl', appendQuery(action.value, { clientId: gaClientId() }), state);
```
robertfairhead: I think your check is fine. I'd go with:
function getLoginURL(default_url) {
 let clientId = gaClientId()
 return (clientId === undefined) ? default_url : appendQuery(default_url, { clientId: clientId })
} 
...
return _.set('loginUrl', getLoginURL(action.value) , state)

But that's basically just out of my own ignorance of how JS and the appendQuery library would handle the `undefined` return in the case GA isn't there. The proverbial abundance of caution approach.

omgitsbillryan: I can confirm that when you try to add a query param of value `undefined`, `append-query` ignores that param.  E.g:
```
appendQuery('http://foo', { bar: undefined })
```
will result as 
```
http://foo
```

src/js/login/reducers/login.js
```diff
@@ -35,13 +37,13 @@ function loginStuff(state = initialState, action) {
...
 
     case UPDATE_VERIFY_URL:
-      return _.set('verifyUrl', action.value, state);
+      return _.set('verifyUrl', appendQuery(action.value, { clientId: gaClientId() }), state);
```
jbalboni: Reducers are supposed to be pure functions in Redux, it would be better if you could call `gaClientId()` in the action creator functions for these and pass the id in on the action object.

src/js/common/helpers/login-helpers.js
jbalboni: Looks good to me.

ayaleloehr:
Checking on status here, as getting this in will really help our Google Analytics

jbalboni (approved):

omgitsbillryan:
@ayaleloehr I'm going to hold off the merge to `master` until after the long weekend since this is a site-wide change.

PR #6286, opened 2017-08-26
Address update request

raquelromano-gov:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4460

Sets up the address update request with a dummy "Update" button and hard-coded address object, both of which should be replaced when the address form UI is built.

Outstanding questions:
1. Does it matter whether we issue a PATCH or PUT? (@kreek)
2. The example JSON request body is taken from the vets-api swagger, but it fails validation (returns a 422). @kreek can you double-check the validation code and/or let me know if you see an issue with this example address? We're adding the form fields and FE validation soon and I want to make sure I have the right source of truth for what constitutes a valid address.

PR #6293, opened 2017-08-29
Letters address fields

annekainicUSDS:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4459

In order to add the address fields, I tried at first to use the common component `Address`. However, I wasn't able to because `Address` expects the address date to be formatted differently from what we have, both in terms of the names of the fields (e.g., `street` vs. `addressLine1`, as well as the actual structure of the data (e.g., `street: { dirty: false, value: '2000 Main St'}` vs. `addressLine1: '2000 Main St'`). So in order to easily and quickly build these address fields, I had to recreate the common `Address` component in our directory, as well as the `ErrorableSelect` and `ErrorableTextInput` components `Address` relies upon. It's not great to have nearly identical components, and we should figure out a way to make a common `Address` component that every app can use, but it was too much work to refactor this at this time. I will create a ticket to go back and refactor this at a later date.

Added a couple of unit tests to test the expand/collapse functionality. Didn't write tests for the new components I added because we want to move back to using the common ones soon.

Still left to do: Post the updated address to vets-api. I think this will be easier to make work in a separate PR once https://github.com/department-of-veterans-affairs/vets-website/pull/6286 is merged.

#### Collapsed
screen shot 2017-08-29 at 12 23 42 pm

#### Expanded
screen shot 2017-08-29 at 12 23 58 pm


src/js/letters/components/Address.jsx
```diff
@@ -0,0 +1,151 @@
...
+    let stateList = [];
+    const selectedCountry = this.props.value.country === 'US' ? 'USA' : this.props.value.country;
+    if (states[selectedCountry]) {
+      stateList = states[selectedCountry];
+      if (this.props.value.city && this.isMilitaryCity(this.props.value.city)) {
+        stateList = stateList.filter(state => state.value === 'AE' || state.value === 'AP' || state.value === 'AA');
+      }
+    }
+
+    const stateProvince = _.hasIn(states, this.props.value.country === 'US' ? 'USA' : this.props.value.country)
```
omgitsbillryan: This part 
```
this.props.value.country === 'US' ? 'USA' : this.props.value.country
```
looks like it can be replaced with `selectedCounrty`.

I see this little bit of logic duplicated in a few places, maybe it belongs in some kind of helper function so we can re-use?

src/js/letters/components/Address.jsx
```diff
@@ -0,0 +1,151 @@
...
+    // if country is changing we should clear the state
+    if (path === 'country') {
+      address = set('state', '', address);
+    }
+
+    this.props.onUserInput(address);
+  }
+
+  validateAddressField(field) {
+    if (this.props.required || !isBlankAddress(this.props.value)) {
```
omgitsbillryan: Excuse my ignorance - what is `this.props.required` as it relates to an `Address` component?

src/js/letters/components/Address.jsx
```diff
@@ -0,0 +1,151 @@
...
+      return isNotBlank(field);
+    }
+
+    return true;
+  }
+
+  validatePostalCode(postalCodeField) {
```
omgitsbillryan: Since this returns a `boolean`, I'd recommend a rename to `isValidPostalCode`.  Same goes for `validateAddressField`

omgitsbillryan (commented):
You said it yourself - the code duplication is quite serious.  For me, "fix it later" usually turns into "never".  I'd recommend getting it right the first time as opposed to trying to come back and fix later.

That said, I defer to your judgement and I only had mostly cosmetic comments.

src/js/letters/components/Address.jsx
annekainicUSDS: This code has changed since you made the comment, but basically the reason why I had to do this logic of `this.props.value.country === 'US' ? 'USA' : this.props.value.country` in several places is because the address we get for the user under `destination` has the country listed as `US`, but in the list of countries we get from EVSS from the `v0/address/countries` endpoint, it is listed as `USA`. Because they don't match, in order for the country to be selected at US once the fields load I have to set the value as `USA`. This is a question I plan on asking EVSS about, if there's a reason why we're storing 2 different values for the U.S., so this may change.

src/js/letters/components/Address.jsx
annekainicUSDS: Good point. I can make that change.

src/js/letters/components/Address.jsx
annekainicUSDS: It's meant to represent the question of whether the address as a whole is required. If it's not required, we may not care that you submit a valid and complete address. If it is required, we do care that the address is valid and complete. A valid address needs to have at least a street, city, state, zip, and country. So we can use the general sense of whether or not the address is required to require each of those individual fields.

src/js/letters/containers/AddressSection.jsx
```diff
@@ -3,8 +3,15 @@ import { connect } from 'react-redux';
 import isEmpty from 'lodash/isEmpty';
 
 import { invalidAddressProperty } from '../utils/helpers.jsx';
+import { updateAddress } from '../actions/letters';
+import Address from '../components/Address';
 
 export class AddressSection extends React.Component {
+  constructor() {
+    super();
+    this.state = { editAddress: false };
```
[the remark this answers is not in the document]
annekainicUSDS: Yep, I like `isEditingAddress`, since it's also a boolean. Thanks!

annekainicUSDS:
@omgitsbillryan thanks so much for your review! In regards to your general concern about replicating the components, I agree, it's not great. One thing to keep in mind is that the common component `Address` is actually soon to no longer be used anywhere on Vets.gov. The reason why is that all of our forms that require an address have moved or are moving to using react-json-schema-form. The component for an address under react-json-schema-form is fundamentally different than a basic React component because it builds the component using json schema. So as long as we're building the letters app as a basic React app and not using react-json-schema-form, we can't use the address component that all other apps use anyway.

The common, basic React component for an address is a leftover component from when we used to build forms with basic React apps. The last form to use that component is the 1990, which is soon to be moved to rjsf anyway (I think this work is almost done), so no apps will be using the common `Address` component soon. Therefore, I think the risk of having essentially duplicated the common `Address` component is basically null because that component isn't even being used.

I think it's a good goal to have common components for frequently asked questions. But the fact that we build our apps using two different techniques (React and rjsf) has complicated this goal. Do we now have 2 sets of common components, one as a basic React component and one in json-schema? Do we coordinate those components to be essentially the same, but just built differently? If so, we also have to have agreement across the design team and across the product teams to make sure that the same component can be used everywhere. Otherwise, we'll have to make it _so_ flexible, by passing tons of props to handle the various edge cases that each app requires, that it may no longer be beneficial to have a single common component. So I think this is a larger issue than just what pertains to letters at this moment.

annekainicUSDS:
@omgitsbillryan updated the PR with a few changes based on your review, it's ready for another look!

src/js/letters/components/Address.jsx
omgitsbillryan: Sounds good.  I don't doubt the need for the logic, just that it's not DRY.

If you can get EVSS to clean up their API, then the need to this logic disappears altogether! 👍

omgitsbillryan:
> ... all of our forms that require an address have moved or are moving to using react-json-schema-form

> ... so no apps will be using the common Address component soon

How come we're not using rjsf for the letters address form?

annekainicUSDS:
@omgitsbillryan because when we first started building this app it wasn't a form: it was just a single page that included custom components for the letter download functionality, without form fields. we've since decided to make it into 2 pages with form fields (design finalized about a week ago), so at this point in the process we're working with what we have. Also rjsf is a little heavy for what we need for this app. I could see an argument for going back to it at this point, even though this is a very simple, lightweight app, but when we started building this there really was no use case for it.

src/js/letters/components/Address.jsx
omgitsbillryan: So a pretty easy change - line 81 could read:
```jsx
const stateProvince = _.hasIn(states, selectedCountry)
```

& line 104
```jsx
value={selectedCountry}
```

annekainicUSDS: I've changed this in a separate branch I have for #4520 that is branched off of this one. I really want to just merge this as is, and this will be fixed in a separate PR.

omgitsbillryan (approved):

PR #6295, opened 2017-08-29
Fix HCA demographic default data

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4517

We have to default `isSpanishHispanicLatino` because of the backend schema, but the default was set incorrectly and always overrode the user's choice. This fixes that.

### DIFF
--- a/test/hca/schema/minimal-test.json
+++ b/test/hca/schema/minimal-test.json
@@ -34,7 +34,9 @@
       "last": "Doe"
     },
     "privacyAgreementAccepted": true,
-    "isSpanishHispanicLatino": false,
+    "view:demographicCategories": {
+      "isSpanishHispanicLatino": false
+    }
     "spouseFullName": {}
   }
 }
--- a/test/hca/schema/minimal-test.json
+++ b/test/hca/schema/minimal-test.json
@@ -36,7 +36,7 @@
     "privacyAgreementAccepted": true,
     "view:demographicCategories": {
       "isSpanishHispanicLatino": false
-    }
+    },
     "spouseFullName": {}
   }
 }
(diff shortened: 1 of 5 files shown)


jbalboni:
Actually, I need to make a migration for this.

jbalboni:
Migration added!

cehsu (approved):
This looks good to me.

test/hca/schema/minimal-test.json
```diff
@@ -34,7 +34,9 @@
       "last": "Doe"
     },
     "privacyAgreementAccepted": true,
-    "isSpanishHispanicLatino": false,
+    "view:demographicCategories": {
+      "isSpanishHispanicLatino": false
```
cvalarida: Is this required in the minimal test? If we leave it out, it's like the user didn't check any of the boxes.

cvalarida (commented):
When loading the form, does are the migrations run? Last I remember, they're not hooked up...

cvalarida:
I traced it back to `src/js/hca/components/SIPIntroductionPage.jsx`; When making the `<SaveInProgressIntro>`, we still need to pass `migrations={this.props.route.formConfig.migrations}` for this to run when loaded.

jbalboni:
@cvalarida It's already running when loaded. We put the migrations in the Redux state and the selector in SaveInProgressIntro pulls it in.

cvalarida (commented):
Ahh, the redux state. I do remember that now. Meh. I still don't like it.

But this is golden as far as that's concerned. 👍

cvalarida (approved):
:point_up: was supposed to be approved...

PR #6306, opened 2017-08-30
Adjustment open / close behavior of global menus

ncksllvn:
These are patches so that the legacy and React dropdowns will close automatically when the user clicks somewhere other than the dropdown.

### DIFF
--- a/src/js/common/components/DropDown.jsx
+++ b/src/js/common/components/DropDown.jsx
@@ -6,6 +6,26 @@ class DropDown extends React.Component {
   constructor(props) {
     super(props);
     this.toggleDropDown = this.toggleDropDown.bind(this);
+    this.handleDocumentClick = this.handleDocumentClick.bind(this);
+  }
+
+  componentWillMount() {
+    this.props.container.addEventListener('click', this.handleDocumentClick, false);
+    this.props.container.addEventListener('touchend', this.handleDocumentClick, false);
+  }
+
+  componentWillUnmount() {
+    this.props.container.removeEventListener('click', this.handleDocumentClick, false);
+    this.props.container.removeEventListener('touchend', this.handleDocumentClick, false);
+  }
+
+  handleDocumentClick(event) {
+    // If this dropdown is open, and it's not an element within this dropdown being clicked,
+    // then the user clicked elsewhere and we should invoke the click handler to toggle this
+    // dropdown to closed.
+    if (this.props.isOpen && !this.dropdownDiv.contains(event.target)) {
+      this.props.clickHandler();
+    }
   }
 
   toggleDropDown() {
@@ -20,7 +40,7 @@ class DropDown extends React.Component {
     );
 
     return (
-      <div className="va-dropdown">
+      <div className="va-dropdown" ref={div => { this.dropdownDiv = div; }}>
         <button className={buttonClasses}
           aria-controls={this.props.id}
           aria-expanded={this.props.isOpen}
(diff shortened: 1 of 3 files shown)


src/js/common/components/DropDown.jsx
```diff
@@ -6,6 +6,26 @@ class DropDown extends React.Component {
...
+    this.props.container.removeEventListener('click', this.handleDocumentClick, false);
+    this.props.container.removeEventListener('touchend', this.handleDocumentClick, false);
+  }
+
+  handleDocumentClick(event) {
+    // If this dropdown is open, and it's not an element within this dropdown being clicked,
+    // then the user clicked elsewhere and we should invoke the click handler to toggle this
+    // dropdown to closed.
+    if (this.props.isOpen && !this.dropdownDiv.contains(event.target)) {
+      this.props.clickHandler();
```
U-DON: Perhaps `this.toggleDropdown()` for clarity. They may be the same right now, but `clickHandler` isn't used directly anywhere else, and it would be cleaner to bind both this event and the button click to the same function.

U-DON (approved):
LGTM. I had been meaning to convert the "legacy" menus into React components, but hadn't gotten around to it. Something to tackle separately.

src/js/legacy/mega-menu.js
```diff
@@ -38,8 +39,16 @@ class MegaMenu {
...
     this.closeControl.addEventListener('click', this.hideMenu);
 
+    document.addEventListener('click', this.handleDocumentClick);
+    document.addEventListener('touchend', this.handleDocumentClick);
     window.addEventListener('resize', this.resetMenu);
+  }
 
+  handleDocumentClick(event){
+    const target = event.target;
+    if (!target.classList.contains('vetnav-level1')){
```
U-DON: Upon testing this out, if you click on any of the space not covered by links in the nav menus, you still close the menu. I don't think this is expected behavior.

U-DON (changes requested):
Actually, found some unexpected behavior while clicking around...

src/js/legacy/mega-menu.js
U-DON: Also nitpick: space before the `{`.

ncksllvn: Good call! Previously there wasn't an event listener there at all so the menu stayed open unless you clicked the toggle button again. But now with an event listener on the document that closes the menu, there needs to be another event listener on the actual dropdown menu to keep that event from bubbling up. Working on it now, thanks!

src/js/common/components/DropDown.jsx
ncksllvn: That makes sense for sure. Will update that!

ncksllvn:
PTAL @U-DON

src/js/legacy/mega-menu.js
```diff
@@ -108,10 +108,11 @@ class MegaMenu {
   }
 
   toggleSubMenu(event) {
-    event.stopPropagation();
     const submenus = Array.from(this.menu.querySelectorAll('.vetnav-panel--submenu'));
     const triggers = Array.from(this.menu.querySelectorAll('.vetnav-level2'));
 
+    event.stopPropagation();
```
U-DON: Why does moving `stopPropagation` to after the variable declarations fix things?

U-DON (approved):
Looks like it works great now! Had one question about how moving `stopPropagation` fixes things.

test/common/components/DropDown.unit.spec.jsx
```diff
@@ -1,24 +1,49 @@
...
+    container,
     contents: (<h1>Hi</h1>),
     cssClass: 'testClass',
     isOpen: true,
     icon: (<svg><rect x="50" y="50" width="50" height="50"/></svg>),
     id: 'testId'
   };
 
-  const tree = SkinDeep.shallowRender(<DropDown {...props}/>);
+  ReactDOM.render(<DropDown {...props}/>, container);
```
cvalarida: Out of curiosity, why was `SkinDeep` not working for you?

I'm guessing either because `shallowRender()` doesn't go through the whole React lifecycle or it doesn't render children without `.dive()`ing (or both), but I'd like to confirm since I haven't had to use `ReactDOM.render()` in a unit test yet.

cvalarida (approved):

test/common/components/DropDown.unit.spec.jsx
ncksllvn: Exactly - I really wanted to have a test case written for the `componentWillUnmount` lifecycle method to confirm that the event handler on the document is being removed correctly. It just looks like a potential memory leak to me otherwise.

cvalarida: Awesome. Good work!

src/js/legacy/mega-menu.js
ncksllvn: Sorry, overlooked this before - that was just a stylistic change to have variable declarations at the top.

test/common/components/DropDown.unit.spec.jsx
ncksllvn: Thank you!

PR #6308, opened 2017-08-31
Fix messaging date display issue on iOS

bshyong:
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4282

Prevents date from being cut off on iOS when it is a full date with a message attachment

screen shot 2017-08-30 at 6 36 15 pm


U-DON (approved):
LGTM.

PR #6309, opened 2017-08-31
Mark GA script as async

jbalboni:
I missed this during my initial pass, but this GA script should be async, even though the actual Google script it includes is already async.


jbalboni:
Holding this until I talk to James about CSP.

omgitsbillryan:
Will this actually do anything?  Since this `<script>` tag is already at the very bottom of the page, I don't think adding `async` to it will have any effect at all.

jbalboni:
@omgitsbillryan Since the other scripts are now marked as `defer`, without marking this one as `async`, it becomes the top priority. You can see it block other scripts in production:

screen shot 2017-08-31 at 5 21 11 pm

cehsu (approved):
LGTM.

PR #6311, opened 2017-08-31
Letter populate address dropdowns

annekainicUSDS:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4520 and connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4669

Warning message screenshot:
screen shot 2017-09-18 at 5 48 27 pm


annekainicUSDS:
What's done so far for this PR:
- added actions to fetch countries and states
- hard-coded the countries and states for now because the fetch only works if you're on CI
- updated the address component to use the hard coded states and countries

Still left to do:
- make sure the fetch actually works (Alastair has a branch on vets-API that mocks the response using betamocks, this may be easier than getting CI set up)
- add the data that is in the response for the death of countries and states to the store
- figure out if EVSS can send us complete names for states instead of/in addition to the abbreviated names
- add the military states to the state list
- calculate the "type" of address based on the country (for "domestic" and "international") or state (for "military")
- make sure everything works :)

src/js/letters/components/Address.jsx
```diff
@@ -60,6 +60,37 @@ class Address extends React.Component {
     return lowerCity === 'apo' || lowerCity === 'fpo' || lowerCity === 'dpo';
   }
 
+  adjustStateNames(stateList, militaryStates, city) {
```
annekainicUSDS: There are 3 different formats the list of states could be in:
1. The list fetched from EVSS is just an array of abbreviations
e.g., `['AL', 'AK', 'AZ'...]`

2. The list of constants we have hard-coded on our side, which is what we decided to actually use, is an object of key-value pairs
e.g., `{AL: 'Alabama', AK: 'Alaska', AZ: 'Arizona'....}`

3. There's the list that `ErrorableSelect` expects, which is an array of objects with 2 keys: `label` and `value`
e.g., `[{label: 'Alabama', value: 'AL'}, {label: 'Alaska', value: 'AK'}....]`

This code transforms list 2 (hard-coded object list) into list 3 (array of objects) so that it can be processed by the `ErrorableSelect` component.

This logic is a little weird. There are basically a few steps I have to take to reformat the state list.
1. Create an empty array
2. Map through the hard-coded state list (list 2) and create a new object for each state, and then push that object into the newly created array
3. Add the military states into that array list
4. Alphabetize the list
5. Filter the list down to the military states if the user has entered a military address city (this is a feature from our common `Address` component across Vets.gov, figured we wanted to maintain that for this app)

I'm having a hard time figuring out how to test this functionality. I figured the important things to test here are to make sure the list is populated and that the military states have been added to the list (could also test the alphabetization and the filtering). So far I tried rendering this component, getting the input for the state dropdown, and then trying to test the values of the options inside that select, but apparently that doesn't work. Any other thoughts?

cvalarida: I'm curious how you tried doing this. Did you use `findDOMNode` (or `getFormDOM` from `test/util/schemaform-utils.jsx`)? I think you could do something like

```js
const component = ReactTestUtils.renderIntoDocument( ... );
const tree = findDOMNode(component);

const options = Array.from(tree.querySelectorAll('option'));
expect(options).to.include('APO'); // etc.
// TODO: Test that all the options are in order, etc
```

annekainicUSDS: I tried this and unfortunately it's not working for me. The `options` that get rendered are React elements, and I'm not sure how to access their value. Ordinarily doing `.value` on a React element in a test works, but for some reason it doesn't work for the options, only the select itself.

src/js/letters/components/Address.jsx
```diff
@@ -60,6 +60,37 @@ class Address extends React.Component {
...
+        return -1;
+      }
+      if (a.label > b.label) {
+        return 1;
+      }
+      return 0;
+    });
+    // Filter state list if military city is entered
+    if (this.props.value[city] && this.isMilitaryCity(this.props.value[city])) {
+      states = states.filter(state => state.value === 'AE' || state.value === 'AP' || state.value === 'AA');
```
cvalarida: Should we save the filtering and just use the `militaryStateNames` from `helpers.jsx`? Also, if we do that, we should probably gate setting `states` based on whether a military city is selected:

```js
let states = [];
if (this.props.value[city] && this.isMilitaryCity(this.props.value[city])) {
  states = militaryStateNames;
} else {
  _.mapKeys(stateList, (value, key) => {
    states.push({ label: value, value: key });
  });
  // Add military states to full state list
  // Would this still be necessary?
  militaryStates.forEach((militaryState) => {
    states.push(militaryState);
  });
}
```
That way we don't sort all the states and then discard them, y'know? Avoids wasted cpu cycles.

src/js/letters/containers/AddressSection.jsx
```diff
@@ -82,6 +85,14 @@ export class AddressSection extends React.Component {
           {invalidAddressProperty}
         </div>
       );
+    // If countries and states are not available when they try to update their address,
+    // they will see this warning message instead of the address fields.
```
cvalarida: This only happens if
1. There's a problem between the api and EVSS and the api no longer has these values in the cache
2. We haven't gotten the response back from the api yet
    - So...how does that play with slow connections? Will this warning message pop in while we're waiting?

src/js/letters/utils/constants.js
```diff
@@ -20,10 +20,15 @@ export const GET_LETTER_PDF_SUCCESS = 'GET_LETTER_PDF_SUCCESS';
...
 export const UPDATE_ADDRESS = 'UPDATE_ADDRESS';
+
 export const SAVE_ADDRESS_PENDING = 'SAVE_ADDRESS_PENDING';
 export const SAVE_ADDRESS_SUCCESS = 'SAVE_ADDRESS_SUCCESS';
 export const SAVE_ADDRESS_FAILURE = 'SAVE_ADDRESS_FAILURE';
 
+export const GET_ADDRESS_COUNTRIES_SUCCESS = 'GET_ADDRESS_COUNTRIES_SUCCESS';
+export const GET_ADDRESS_COUNTRIES_FAILURE = 'GET_ADDRESS_COUNTRIES_FAILURE';
+export const GET_ADDRESS_STATES_SUCCESS = 'GET_ADDRESS_STATES_SUCCESS';
+export const GET_ADDRESS_STATES_FAILURE = 'GET_ADDRESS_STATES_FAILURE';
```
cvalarida: I'm not sure what all the ui considerations are here, but would it make sense to have pending states as well?

cvalarida (approved):
I think the only other thing I'd like to see is throwing an error up to Sentry if we get some unexpected values from EVSS. When they change things, we want to know, right?

src/js/letters/utils/constants.js
annekainicUSDS: Good question. This fetch happens when the app initially loads, but the states and countries aren't loaded until the user clicks the "Update" button, so there's no lag from the user's perspective. I think it's fine without a pending state.

src/js/letters/containers/AddressSection.jsx
annekainicUSDS: The fetch is attempted in `Main`, so I assume we'll get a response back from that fetch before the user is able to click the `Update` button and `isEditingAddress` is changed to `true`, especially because the address content won't render until the fetch for the address is also returned (which happens at the same time as the fetch for the countries and states). I wonder if we should tie the loading indicator for the whole app on the responses from *all* the fetches in `Main`. What do you think @cvalarida and @markgreenburg?

src/js/letters/components/Address.jsx
annekainicUSDS: Ah gotcha, makes sense!

src/js/letters/containers/AddressSection.jsx
annekainicUSDS: Created a new ticket to handle this separately: https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4990

src/js/letters/reducers/index.js
```diff
@@ -123,6 +133,41 @@ function letters(state = initialState, action) {
       return _.set('address', action.address, newState);
     }
     // Add SAVE_ADDRESS_FAILURE
+    case GET_ADDRESS_COUNTRIES_SUCCESS: {
+      let countriesAvailable = true;
+      const countryList = action.countries.data.attributes.countries;
+
+      // Log error if the countries response is not what we expect
+      if (!_.isArray(countryList) || countryList.length === 0) {
```
annekainicUSDS: @cvalarida I took your suggestion of logging a Sentry error if the response for countries and states isn't what we expect. Do you mind taking another look at this to make sure it makes sense? I'm basically checking if either the response isn't an array or if it's an empty array, and then both logging a Sentry error and making the countries/states unavailable.

cvalarida: Looks good! I'd just like to reduce lodash dependency, so my only nitpick here is that I'd rather see `Array.isArray(countryList)`.

If you're tired of changing things, though, that's totally fine. 😛

cvalarida (commented):
Do we have a ticket made to write that test you couldn't get to work?

If not, we should make one, if so, this is golden!

PR #6312, opened 2017-08-31
Convert fields that are groups of inputs to use fieldset

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4493
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4495
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1905 (old issue)

This changes the fields that are groups of inputs to use fieldset/legend. This is mainly radio/yesNo fields and date fields. The rest of the changes are adjustments to make sure they still look correct.

### DIFF
--- a/src/js/common/schemaform/FieldTemplate.jsx
+++ b/src/js/common/schemaform/FieldTemplate.jsx
@@ -68,7 +68,7 @@ export default function FieldTemplate(props) {
     return children;
   }
 
-  const labelElement = isFieldGroup
+  const labelElement = (isFieldGroup || showFieldLabel)
     ? <legend className={labelClassNames}>{label}{requiredSpan}</legend>
     : <label className={labelClassNames} htmlFor={id}>{label}{requiredSpan}</label>;
 
@@ -84,7 +84,7 @@ export default function FieldTemplate(props) {
     </div>
   );
 
-  if (isFieldGroup) {
+  if (isFieldGroup || showFieldLabel) {
     return (
       <fieldset className={containerClassNames}>
         {content}
--- a/src/js/common/schemaform/FieldTemplate.jsx
+++ b/src/js/common/schemaform/FieldTemplate.jsx
@@ -27,6 +27,7 @@ export default function FieldTemplate(props) {
   const isDateField = uiSchema['ui:widget'] === 'date';
   const showFieldLabel = uiSchema['ui:options'] && uiSchema['ui:options'].showFieldLabel;
   const hideLabelText = uiSchema['ui:options'] && uiSchema['ui:options'].hideLabelText;
+  const useLabel = uiSchema['ui:options'] && uiSchema['ui:options'].useLabel;
 
   const description = uiSchema['ui:description'];
   const textDescription = typeof description === 'string' ? description : null;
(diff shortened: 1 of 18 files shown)


src/js/common/schemaform/FieldTemplate.jsx
```diff
@@ -56,21 +58,46 @@ export default function FieldTemplate(props) {
...
+  if (useFieldsetLegend) {
+    return (
+      <fieldset className={containerClassNames}>
+        {content}
+      </fieldset>
     );
+  }
+
+  return (
+    <div className={containerClassNames}>
```
cvalarida: Idle curiosity: What would happen if we always used a `fieldset`? Is this separate return so we don't use a `label` in a `fieldset` (because of reasons I don't know about)?

src/js/pensions/config/form.js
```diff
@@ -614,6 +614,7 @@ const formConfig = {
                 showFieldLabel: true,
                 keepInPageOnReview: true,
                 expandUnder: 'maritalStatus',
+                useLabel: true,
```
cvalarida: Because `showFieldLabel` is `true`? (Also, why is that part of the determination to use a `<legend>` or not?)

cvalarida (approved):

src/js/common/schemaform/FieldTemplate.jsx
jbalboni: It feels semantically wrong to always use `fieldset`, since it's meant to group related elements. If it's just one field, it's not really a group. I don't think it would break anything, though.

src/js/pensions/config/form.js
jbalboni: Yeah, I suppose that's a little convoluted. Most of the time when `showFieldLabel` is true, it's because we want to show a label for an object property, which is going to be a label for a group of elements, and therefore should be a `legend`. `useLabel` is an option to force the template to use a `label` instead of a `legend`.

So `showFieldLabel` is using the word "label" generically, and `useLabel` is using it specifically to refer to `label` elements. Which is confusing, yeah.

`useLabel` could be `useLabelInsteadOfLegend`? I'm open to suggestions.

cehsu: Is it possible to consolidate this logic into one option, such as labelType, and specify label legend or none?

src/sass/modules/_m-schemaform.scss
```diff
@@ -195,6 +195,22 @@ label {
...
+  font-size: inherit;
+  color: inherit;
+  line-height: inherit;
+  max-width: 46rem;
+  padding-bottom: 0;
+}
+
+fieldset.schemaform-field-template {
+  margin-top: 3rem;
+  &.schemaform-first-field {
```
cehsu: Is this ampersand necessary?

cehsu (approved):
Looks good to me. Had a few minor questions.

src/js/pensions/config/form.js
jbalboni: A `labelType` option would work, but I'm not sure I want to force us to choose each time we use this option. Normally you would want to let the FieldTemplate component choose the right one.

I'm going to change `showFieldLabel` to allow a 'label' string for when you want a label element, I think that's actually clearer.

src/sass/modules/_m-schemaform.scss
jbalboni: I believe so, since `schemaform-first-field` is added to the fieldset when it's the first field.

PR #6316, opened 2017-08-31
Fix for android touch events on T&Cs

bshyong:
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3582

uses touch event to address Android buginess


src/js/common/components/Modal.jsx
```diff
@@ -71,7 +71,7 @@ class Modal extends React.Component {
       closeButton = (<button
         className="va-modal-close"
         type="button"
-        onClick={this.props.onClose}>
+        onClick={this.props.onClose} onTouchStart={this.props.onClose}>
```
U-DON: Nitpick: Can this be on its own line like the other props?

U-DON: Furthermore, can the click and touch events somehow double trigger the handler? If it just closes in this case, it probably won't be an issue, but something like toggle could be problematic.

U-DON (approved):
LGTM.

PR #6320, opened 2017-09-01
Remove empty objects from arrays

cvalarida:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2773

### DIFF
--- a/src/js/common/schemaform/helpers.js
+++ b/src/js/common/schemaform/helpers.js
@@ -235,6 +235,13 @@ export function stringifyFormReplacer(key, value) {
     }
   }
 
+  // Clean up empty objects in arrays
+  if (Array.isArray(value)) {
+    const newValues = value.filter(v => !!stringifyFormReplacer(key, v));
+    // If every item in the array is cleared, remove the whole array
+    return newValues.length > 0 ? newValues : undefined;
+  }
+
   return value;
 }
 
--- a/test/common/schemaform/helpers.unit.spec.js
+++ b/test/common/schemaform/helpers.unit.spec.js
@@ -432,6 +432,38 @@ describe('Schemaform helpers:', () => {
       expect(output.someField).to.be.undefined;
       expect(output.someField2).to.be.undefined;
     });
+    it('should remove empty objects within an array', () => {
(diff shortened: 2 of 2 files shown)


src/js/common/schemaform/helpers.js
```diff
@@ -235,6 +235,13 @@ export function stringifyFormReplacer(key, value) {
     }
   }
 
+  // Clean up empty objects in arrays
+  if (Array.isArray(value)) {
+    const newValues = value.filter(v => !!stringifyFormReplacer(key, v));
+    // If every item in the array is cleared, remove the whole array
+    return newValues.length > 0 ? newValues : undefined;
```
cvalarida: It made sense to me that if all of the objects in an array were somehow empty, we should probably not send an empty array. If we want to change that, we can, but I figured I'd address it while I was here since we could feasibly add an item to an array (with information), hit "Add another", leave that one blank, edit the first one to remove all the information, update, and hit "Continue". This would (in theory; haven't tested) leave us with an array of two blank objects which would be removed when we submit the form.

jbalboni: I think this is the correct thing to do, but it could theoretically break backend code. I would make sure to test HCA and validate that you don't get schema errors (I think HCA is always going to fail for you locally, but you should be able to tell if the error is from the ES backend or a schema error).

jbalboni (approved):
Looks ok to me.

cehsu (approved):
Looks good to me.

src/js/common/schemaform/helpers.js
cvalarida: Just tested HCA and I wasn't able to break it. Probably because everything is required. 😛 
We'll have to keep an eye out for schema errors, but I think this will only affect the 1990e and 5490 since I _think_ they're the only two forms that have arrays where no field in the array item is required. The 1990-rjsf might fall into that category as well, so we'll have to watch out for it.

Currently, if no information is given for any items, the array isn't passed at all. This duplicates that functionality so I _think_ (:crossed_fingers:) we should be good.

PR #6322, opened 2017-09-01
Incorporate resume routing

cehsu:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4537

### DIFF
--- a/src/js/common/schemaform/FormApp.jsx
+++ b/src/js/common/schemaform/FormApp.jsx
@@ -51,8 +51,10 @@ class FormApp extends React.Component {
     //  saved form / prefill
     // If we're in production, we'll redirect if we start in the middle of a form
     // In development, we won't redirect unless we append the URL with `?redirect`
+    const resumeForm = this.props.currentLocation.search.includes('resume');
     const devRedirect = __BUILDTYPE__ !== 'development' || this.props.currentLocation.search.includes('redirect');
-    if (isInProgress(this.props.currentLocation.pathname) && devRedirect) {
+    const goToStartPage = resumeForm || devRedirect;
+    if (isInProgress(this.props.currentLocation.pathname) && goToStartPage) {
       // We started on a page that isn't the first, so after we know whether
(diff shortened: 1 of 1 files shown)


src/js/common/schemaform/FormApp.jsx
```diff
@@ -51,8 +51,10 @@ class FormApp extends React.Component {
     //  saved form / prefill
     // If we're in production, we'll redirect if we start in the middle of a form
     // In development, we won't redirect unless we append the URL with `?redirect`
+    const resumeForm = this.props.currentLocation.search.includes('resume');
     const devRedirect = __BUILDTYPE__ !== 'development' || this.props.currentLocation.search.includes('redirect');
-    if (isInProgress(this.props.currentLocation.pathname) && devRedirect) {
+    const goToStartPage = resumeForm || devRedirect;
+    if (isInProgress(this.props.currentLocation.pathname) && goToStartPage) {
```
cvalarida: I'm guessing the idea is to link to `/path/to/form/some-page?resume` from the profile page so we can auto-load the form, yeah? We'll still need to link to a page _inside_ the form (not just the intro) for that to work, but if that's no problem, then great!

cvalarida (approved):

cvalarida (changes requested):
Actually, now that I _look_ at the ticket this is linked to ( :man_facepalming: ), this isn't doing what that ticket says. https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4537 says to create a `/resume` route, but this looks in the `GET` params to see if `resume` is in there. Is that intended?

I guess we could just link to `/path/to/form/?resume`, but I'm not convinced that would work because the logic as it stands checks to see if the page we navigated to is in the middle of the form.

cvalarida:
**Update**: The implementation of resuming the application from the profile page can be found in the [`profile-resume-application` PR](https://github.com/department-of-veterans-affairs/vets-website/pull/6323/files).

Since that has the same changes as this branch does, can we close this one and just use `profile-resume-application`?

PR #6325, opened 2017-09-05
Adding a defer to the usa-search.js.

robbiethegeek-usds:
Seems like this file has been updated since https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4513 was opened, but think this could prevent this from blocking by adding a defer.

### DIFF
--- a/assets/js/usa-search.js
+++ b/assets/js/usa-search.js
@@ -1,7 +1,7 @@
 var usasearch_config = { siteHandle:"vets.gov_search" };
 var script = document.createElement("script");
 script.src = "https://search.usa.gov/javascripts/remote.loader.js";
-script.defer = true;
+script.async = true;
 document.getElementsByTagName("head")[0].appendChild(script);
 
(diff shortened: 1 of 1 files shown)


assets/js/usa-search.js
```diff
@@ -1,6 +1,7 @@
 var usasearch_config = { siteHandle:"vets.gov_search" };
 var script = document.createElement("script");
 script.src = "https://search.usa.gov/javascripts/remote.loader.js";
+script.defer = true;
```
jbalboni: I think `async` is more often used for this sort of external script, but should be an improvement either way.

jbalboni (approved):

assets/js/usa-search.js
robbiethegeek-usds: would you like it as async?  I am happy to :)

robbiethegeek-usds: converted

PR #6327, opened 2017-09-05
Enable SiP for 5495

cvalarida:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4479


src/js/edu-benefits/5495/config/form.js
```diff
@@ -13,7 +13,7 @@ import fullNameUI from '../../../common/schemaform/definitions/fullName';
 
 import * as personId from '../../../common/schemaform/definitions/personId';
 
-import IntroductionPage from '../components/IntroductionPage';
+import IntroductionPage from '../components/SIPIntroductionPage';
```
jbalboni: I think you can just update the current intro page. It's easy enough to just revert this commit if we need to.

src/js/edu-benefits/5495/config/form.js
```diff
@@ -43,7 +43,7 @@ const formConfig = {
   trackingPrefix: 'edu-5495-',
   formId: '22-5495',
   version: 0,
-  disableSave: true,
+  // disableSave: true,
```
jbalboni: I had also added in `savedFormMessages`, based on what we did for pensions and hca.

jbalboni (approved):

cehsu (approved):
LGTM. No additional comments beyond what @jbalboni mentioned.

src/js/edu-benefits/5495/components/IntroductionPage.jsx
```diff
@@ -60,26 +60,34 @@ class IntroductionPage extends React.Component {
...
-              afterText="»"/>
-          </div>
-        </div>
-        <div className="omb-info--container">
+        <SaveInProgressIntro
+          pageList={this.props.route.pageList}
+          messages={this.props.route.formConfig.savedFormMessages}
+          {...this.props.saveInProgressActions}
+          {...this.props.saveInProgress}>
+          Please complete the 22-5495 form to apply for education benefits.
```
cehsu: It looks like we have inconsistent language here and in the 1990e, do we need to confirm the content for this across these forms?

jbalboni: Yeah, we probably should. I think this one can at least be changed to match the format in HCA/1990e/pension/burials. I think that language is from Peggy. We can create an issue for her to go back through them and confirm.

jbalboni: Yeah, we probably should. I don't think it needs to hold this up, but we need to have her look at them.

PR #6330, opened 2017-09-05
4302 homepage image

phiden:
CSS change to keep homepage image from getting cut off. See: https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4302


robertfairhead (approved):
LGTM. Verified on heroku page it's scaled correctly now.

PR #6334, opened 2017-09-06
Hide invisible elements from tab order and keep focus on selection

cehsu:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4454

This enables keyboard navigation of the wizard. In researching [radio accessibility](http://webaim.org/techniques/keyboard/), I learned that arrow keys should enable toggling, which is not how the forms behave. Is this desirable? Or is it beyond the scope of our requirements?

### DIFF
--- a/src/js/edu-benefits/education-wizard.js
+++ b/src/js/edu-benefits/education-wizard.js
@@ -2,9 +2,11 @@ function toggleClass(element, className) {
   element.classList.toggle(className);
 }
 function openState(element) {
+  element.setAttribute('hidden', false);
   element.setAttribute('data-state', 'open');
 }
 function closeState(element) {
+  element.setAttribute('hidden', true);
   element.setAttribute('data-state', 'closed');
   const selectionElement = element.querySelector('input:checked');
   if (selectionElement) {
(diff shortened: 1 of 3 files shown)


jbalboni:
It's not clear to me what this change fixes. Form elements already have a tab index and I don't see a difference when using the form.

Trying the wizard again with the keyboard, I think the issue is that we (Chris and I) are used to radio buttons working like they do in the form, where you can tab through each option. This appears to be wrong, and like your link says, you should use the arrow keys to move between them and tab should jump to the next set of form controls. 

The wizard is working correctly in this regard, but it looks like we focus on the next question after you choose an option, which is fine when you're using the mouse, but is super confusing when using the keyboard. I think the original intent of that focusing was to make sure you were still focused at a reasonable spot when a question disappears, but I'm not sure exactly.

I think what we need to do is 1) change the focus code so that we're not stealing it from users when they choose an option and 2) create an issue to fix the forms so that tabbing doesn't step through each radio button.

cehsu:
Thanks @jbalboni, I'll reimplement this so that the focus doesn't jump unexpectedly (especially to elements that aren't visible).

cehsu:
Also, the [issue](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4609) for the radio buttons on the forms has been created.

cehsu:
Thanks @jbalboni, changes made.

jbalboni (approved):
LGTM! Works much better now.

PR #6338, opened 2017-09-06
WIP: refactor address editing

raquelromano-gov:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4520

This was an attempt to refactor the current VA Letters address update feature, which required a new address component because CorpDB uses its own custom address format that our existing component (still used only by the pre-RJSF 1990 form)  doesn't support.

This refactor doesn't yet do any validation, as we are still ironing out the requirements for the various fields with EVSS.  However, after doing this refactoring, I began to think it would make much more sense to just make this a 1-page RJSF form and take advantage of the simplicity of defining field dependencies and validation functions in the schema. (The original product requirements for VA letters did not have a form at all--it was just an app to view and download PDFs, which is why it was originally not built with RJSF.)

Getting a second opinion from @cvalarida this week.


cvalarida:
@raquelromano-gov I'm guessing we want to just close this now, right?

raquelromano-gov:
yup! thanks

PR #6347, opened 2017-09-06
Make Privacy Act Statement link tab accessible

cehsu:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4500#issuecomment-327580489


cvalarida (approved):
In that original ticket, you said you needed to investigate more about why it wasn't working before. I _think_ I know why...

Disclaimer: I haven't worked with screen readers much....like...yesterday was the first time.

From my understanding, this works because we would have expected it under the links menu (which is a screenreader thing; it grabs all the links in a page so we can quickly navigate them), but it didn't have a `href`, so I'm guessing that's why it didn't show up in the menu. I'm also guessing we weren't able to open the link once it was tabbed to because the modal is opened `onClick`, but the screenreader was trying to follow the link instead of clicking it..? I'm really not sure.

At any rate, making it a button styled like a link is the right thing to do. I just tested with VoiceOver and it works great.

jbalboni (approved):

PR #6352, opened 2017-09-07
changed 'survivors pension' to the possessive 'survivor's pension'

phiden:
See [#4616](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4616)


robbiethegeek-usds (approved):

peggygannon:
@phiden Hi, Sophia, this change should not be made. Survivors is correct as is. We don't need the '. Can we revert this PR back?

PR #6354, opened 2017-09-07
Fix ArrayField Sentry Error

cvalarida:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4688

This was only happening in HCA because that's the only place we had an array field with a locked item count (I'm pretty sure). I'm not sure how this was possible, but I'm thinking maybe they reported that they didn't have any children? Seems odd that there were so few errors if that were the case.

At any rate, this just defaults to an empty array if the item count is locked and no `arrayData` is passed, so it'll generate the validation error, but the application won't die ignominiously.

### DIFF
--- a/src/js/common/schemaform/review/ArrayField.jsx
+++ b/src/js/common/schemaform/review/ArrayField.jsx
@@ -175,7 +175,9 @@ class ArrayField extends React.Component {
 
     // TODO: Make this better; it’s super hacky for now.
     const itemCountLocked = this.isLocked();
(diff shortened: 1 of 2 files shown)


jbalboni (approved):
Makes sense to me. I think this only happens when you don't have any children, you're including financials, and you open the right chapter on the review page, so it may be pretty unlikely.

cehsu (approved):
LGTM.

PR #6356, opened 2017-09-07
Fix breadcrumbs on claim status

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4326

This fixes the breadcrumbs for the claim status app. It looks like there were some generic breadcrumbs added for Appeals, but the claims pages have some nested page and need to generate specific links for breadcrumbs. So I've moved the global breadcrumbs into an appeals specific layout component for those pages.

screen shot 2017-09-07 at 1 51 37 pm
screen shot 2017-09-07 at 1 51 09 pm
screen shot 2017-09-07 at 1 50 50 pm
screen shot 2017-09-08 at 9 16 22 am


src/js/disability-benefits/components/AppealLayout.jsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import Breadcrumbs from './Breadcrumbs';
+
+export default function AppealLayout({ children }) {
+  return (
+    <div>
+      <div>
+        <div className="row">
```
cvalarida: Why is this nested in an otherwise empty `div`?

src/js/disability-benefits/containers/DisabilityBenefitsApp.jsx
```diff
@@ -30,11 +29,6 @@ class DisabilityBenefitsApp extends React.Component {
         loginUrl={this.props.loginUrl}
         verifyUrl={this.props.verifyUrl}>
         <AppContent>
-          <div>
-            <div className="row">
-              <Breadcrumbs location={this.props.location}/>
```
cvalarida: Why was this removed?

cvalarida (approved):
The screenshots look good, but I had a few question.

src/js/disability-benefits/containers/DisabilityBenefitsApp.jsx
cehsu: It looks like it's included in the children now instead.

src/js/disability-benefits/components/AppealLayout.jsx
cehsu: Should the classes of the nested `div`s be consistent with other usages of Breadcrumbs?
```
<div>
  <div className="row">
    <div className="medium-12 columns">
      <Breadcrumbs/>
    </div>
  </div>
  <div className="row">
```

src/js/disability-benefits/routes.jsx
```diff
@@ -12,6 +12,7 @@ import AskVAPage from './containers/AskVAPage.jsx';
 import DocumentRequestPage from './containers/DocumentRequestPage.jsx';
 import AdditionalEvidencePage from './containers/AdditionalEvidencePage.jsx';
 import ClaimEstimationPage from './containers/ClaimEstimationPage.jsx';
+import AppealLayout from './components/AppealLayout';
```
cehsu: It looks like AppealLayout is containing a container, should it also be a container rather than a component (located in the component directory)?

src/js/disability-benefits/containers/YourClaimsPage.jsx
```diff
@@ -173,14 +174,19 @@ class YourClaimsPage extends React.Component {
 
     return (
       <div className="your-claims">
+        <Breadcrumbs/>
```
cehsu: It looks like the class nesting is slightly different here. Does the `your-claims` class serve a styling purpose? Does the specificity override other styling?

cehsu (approved):
LGTM.

src/js/disability-benefits/components/AppealLayout.jsx
jbalboni: I am not sure why I added that extra div. Probably can remove it.

@cehsu I was matching the way the Appeals app was using it. The layout is a little finicky, so I think I'd like to leave it the way it is since it looks right.

src/js/disability-benefits/containers/DisabilityBenefitsApp.jsx
jbalboni: The breadcrumbs are now either in specific page components (claims) or in the AppealsLayout (appeals).

src/js/disability-benefits/routes.jsx
jbalboni: I think since it's a regular component it should go in components.

src/js/disability-benefits/containers/YourClaimsPage.jsx
jbalboni: Yeah, `your-claims` overwrites some styles in its children (a lot of the css for this app is a tangled mess). I'm going to create a tech debt ticket for that.

src/js/disability-benefits/components/AppealLayout.jsx
jbalboni: I tried add the columns div, but it adds more padding and affects the alignment, so I'd have to correct for it again here, which doesn't seem like an improvement.

src/js/disability-benefits/components/Breadcrumbs.jsx
```diff
@@ -2,14 +2,13 @@ import React from 'react';
...
-    ];
-
-    return (<div className="db-breadcrumbs">
-      {crumbs.reduce((content, e) => { return [...content, ' › ', e]; }, []).slice(1)}
-    </div>);
+    return (<nav className="va-nav-breadcrumbs">
+      <ul className="row va-nav-breadcrumbs-list columns claims-breadcrumbs" role="menubar" arialLabel="Primary">
+        <li><a href="/" key="home">Home</a></li>
+        <li><a href="/disability-benefits" key="disability-benefits">Disability Benefits</a></li>
+        {this.props.children}
```
bshyong: 👍

bshyong (approved):
LGTM

PR #6357, opened 2017-09-07
Add JSX whitespace to correct for React's EOL whitespace stripping

markgreenburg:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4664

Doesn't get any smaller than this :)


U-DON (changes requested):
I don't think this is the nicest looking way to resolve this issue. Typically we handle this by just moving at least the the opening `a` tag (if not the whole thing) to the preceding line of text.

markgreenburg:
@U-DON That makes sense. I was thinking that this way we wouldn't need to worry about where line breaks happen (which could break things down the road) but agree it doesn't look great and good to know how you all handle this normally. I'll push again in a bit.

markgreenburg:
Moved fix to new PR [6374](https://github.com/department-of-veterans-affairs/vets-website/pull/6374)

PR #6359, opened 2017-09-07
Make radio buttons more accessible

cvalarida:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4609

So there were a couple problems here. First, when we hit tab, it would cycle through all the radio options instead of the navigating to / from the whole `<fieldset>`. Second, the arrow keys wouldn't cycle through the options in the `<fieldset>`.

All of this is because the radio buttons had different names. With this PR, radio buttons inside the same `<fieldset>` all share the same name.

For more information, see [this WebAIM article](http://webaim.org/techniques/keyboard/) (search for "Keyboard Testing") and for an example, see [this WebAIM example](http://webaim.org/techniques/forms/controls) (search for "Radio Buttons").

### DIFF
--- a/src/js/common/schemaform/widgets/RadioWidget.jsx
+++ b/src/js/common/schemaform/widgets/RadioWidget.jsx
@@ -28,7 +28,7 @@ export default function RadioWidget({
               autoComplete="false"
               checked={checked}
               id={`${id}_${i}`}
-              name={`${id}_${i}`}
+              name={`${id}`}
               value={option.value}
               disabled={disabled}
               onChange={_ => onChange(option.value)}/>
--- a/test/e2e/edu-helpers.js
+++ b/test/e2e/edu-helpers.js
@@ -84,11 +84,11 @@ function completeApplicantInformation(client, data, prefix = 'relative') {
 
   if (typeof data.relationship !== 'undefined') {
     client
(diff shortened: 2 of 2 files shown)


src/js/common/schemaform/widgets/RadioWidget.jsx
```diff
@@ -28,7 +28,7 @@ export default function RadioWidget({
               autoComplete="false"
               checked={checked}
               id={`${id}_${i}`}
-              name={`${id}_${i}`}
```
jbalboni: Oh, wow. Using controlled components seems to have kept us from ever showing multiple radio buttons selected at once, even though the html allowed it.

jbalboni (approved):

src/js/common/schemaform/widgets/RadioWidget.jsx
cvalarida: Heh, huzzah for being accidentally correct!  🎉

cehsu (approved):
LGTM.

src/js/common/schemaform/widgets/RadioWidget.jsx
cehsu: Hm, fascinating find.

PR #6360, opened 2017-09-07
Rationalize isDataAvailable - set to true/false instead of undefined/false

patrickvinograd:
Fixes https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2752

### DIFF
--- a/src/js/common/components/RequiredLoginView.jsx
+++ b/src/js/common/components/RequiredLoginView.jsx
@@ -74,7 +74,7 @@ class RequiredLoginView extends React.Component {
         // DOM elements.)
         const props = { isDataAvailable: this.isServiceAvailable() };
         return React.Children.map(this.props.children, (child) => {
-          React.cloneElement(child, typeof child.type === 'function' : props, null);
+          React.cloneElement(child, (typeof child.type === 'function' ? props : null));
         });
       } else if (this.props.userProfile.accountType === 1) {
         return  <Main renderType="verifyPage"/>;
--- a/src/js/common/components/RequiredLoginView.jsx
+++ b/src/js/common/components/RequiredLoginView.jsx
@@ -66,22 +66,15 @@ class RequiredLoginView extends React.Component {
           return <SystemDownView messageLine1="We couldn’t find your records with that information." messageLine2="Please call the Vets.gov Help Desk at 1-855-574-7286. We’re open Monday‒Friday, 8:00 a.m.‒8:00 p.m. (ET)."/>;
         }
 
-        // If va_profile has any other value, continue on to check if this user can
-        // use this specific service.
-        if (this.isServiceAvailable()) {
-          // If you have the required service, show the application view.
-          return this.props.children;
-        }
-
+        // Otherwise if va_profile is present and populated, show the application view.
         // If the required service is not available, the component will still be rendered,
-        // but we pass an `isDataAvailable` prop to child components indicating there is
-        // no data. (Only add this prop to React components (functions), and not ordinary
+        // but we pass an `isDataAvailable` prop to child components indicating whether 
+        // or not there is data.
+        // (Only add this prop to React components (functions), and not ordinary
         // DOM elements.)
+        const props = { isDataAvailable: this.isServiceAvailable() };
         return React.Children.map(this.props.children, (child) => {
-          const props = typeof child.type === 'function'
-            ? { isDataAvailable: false }
-            : null;
-          return React.cloneElement(child, props);
+          React.cloneElement(child, typeof child.type === 'function' : props, null);
         });
       } else if (this.props.userProfile.accountType === 1) {
         return  <Main renderType="verifyPage"/>;
--- a/src/js/common/components/RequiredTermsAcceptanceView.jsx
+++ b/src/js/common/components/RequiredTermsAcceptanceView.jsx
@@ -27,7 +27,6 @@ export class RequiredTermsAcceptanceView extends React.Component {
 
   render() {
     const { terms, topContent } = this.props;
-    const enabled = this.props.isDataAvailable === true || typeof this.props.isDataAvailable === 'undefined';
 
     let view;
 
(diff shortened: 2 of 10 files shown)


src/js/common/components/RequiredLoginView.jsx
```diff
@@ -65,23 +65,21 @@ class RequiredLoginView extends React.Component {
...
-            ? { isDataAvailable: false }
-            : null;
-          return React.cloneElement(child, props);
-        });
+        return React.Children.map(this.props.children,
+          (child) => {
+            let props = null;
+            if (typeof child.type === 'function') {
+              props = { isDataAvailable: this.isServiceAvailable() };
+            }
```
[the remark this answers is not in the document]
U-DON: Also, rather than recreate `props = { isDataAvailable: this.isServiceAvailable() };` for each child, it might be better to create it once outside of the `map`. Although, for the most part, it's usually only going to occur once for a particular child.

```
const props = { isDataAvailable: this.isServiceAvailable() };
return React.Children.map(
  this.props.children,
  (child) => React.cloneElement(child, (typeof child.type === 'function' ? props : null))
);

File: test/common/components/RequiredLoginView.unit.spec.jsx [hunk 2]
```diff
@@ -104,7 +104,7 @@ describe('<RequiredLoginView>', () => {
 
     // Child components should not be passed an isDataAvailable prop
     tree.props.children.forEach((child) => {
-      expect(child.props.isDataAvailable).to.be.undefined;
+      expect(child.props.isDataAvailable).to.equal(true);
```
U-DON: `to.be.true` is the more idiomatic way to write that.

U-DON (approved):
LGTM. Some minor comments.

U-DON:
Also, maybe not for this set of changes, but the `AppContent` component seems like something that can be pulled out to `common` and shared across the apps.

src/js/common/components/RequiredLoginView.jsx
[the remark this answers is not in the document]
raquelromano-gov: Yeah, I like the above to create the props outside of the `map` and just add it when necessary, but probably a  negligible efficiency gain, so your call.

src/js/common/components/RequiredLoginView.jsx
```diff
@@ -65,23 +65,21 @@ class RequiredLoginView extends React.Component {
...
         }
 
-        // If va_profile has any other value, continue on to check if this user can
-        // use this specific service.
-        if (this.isServiceAvailable()) {
-          // If you have the required service, show the application view.
-          return this.props.children;
-        }
-
+        // If va_profile has any other value, show the application view.
```
raquelromano-gov: This comment seems out of place now.

Am I reading the logic correctly that we go ahead and pass in `props = { isDataAvailable: this.isServiceAvailable() };` regardless of whether `this.isServiceAvailable()` evaluates to true or false to get the same effect as what's currently here?

I should be able to verify that, but it's late Friday afternoon so feeling lazy.

raquelromano-gov: Just looked at the specs and confirmed that yes, that's what's happening.

raquelromano-gov: Would you be willing to update the comments that refer to `va_profile` to make more sense? Otherwise, LGTM

raquelromano-gov (approved):
Thanks for doing this!

patrickvinograd:
Incorporated review feedback and rebased on current master - wouldn't mind another quick review.

src/js/common/components/RequiredLoginView.jsx
```diff
@@ -66,22 +66,15 @@ class RequiredLoginView extends React.Component {
...
-
+        // Otherwise if va_profile is present and populated, show the application view.
         // If the required service is not available, the component will still be rendered,
-        // but we pass an `isDataAvailable` prop to child components indicating there is
-        // no data. (Only add this prop to React components (functions), and not ordinary
+        // but we pass an `isDataAvailable` prop to child components indicating whether 
+        // or not there is data.
+        // (Only add this prop to React components (functions), and not ordinary
         // DOM elements.)
+        const props = { isDataAvailable: this.isServiceAvailable() };
```
raquelromano-gov: I have the desire to rename a bunch of variables and prop names in this code, but maybe that's better done in a later PR that moves `AppContent` into shared code as @bshyong suggested:
`serviceRequired` and `requiredApp` should both just be `serviceName`
`isDataAVailable` should just be `isServiceAvailable`

Would you mind filing a tech debt ticket for that for the FE tech debt warriors to follow up on? (you can add labels `sitewide` and `front end`)

src/js/common/components/RequiredTermsAcceptanceView.jsx
```diff
@@ -38,7 +37,7 @@ export class RequiredTermsAcceptanceView extends React.Component {
         (child) => {
           let props = null;
```
raquelromano-gov: Collapse all this into 2 lines like you did above? 
```
const props = { isDataAvailable: this.props.isDataAvailable };
React.cloneElement(child, (typeof child.type === 'function' ? props : null));
```

test/common/components/RequiredLoginView.unit.spec.jsx
```diff
@@ -108,7 +108,7 @@ describe('<RequiredLoginView>', () => {
 
     // Child components should not be passed an isDataAvailable prop
```
raquelromano-gov: This comment also needs updating

test/claims-status/components/ClaimsStatusApp.unit.spec.jsx
```diff
@@ -20,10 +20,11 @@ describe('<ClaimsStatusApp>', () => {
     expect(tree.subTree('RequiredLoginView').props.authRequired).to.equal(3);
   });
   it('should render children', () => {
+    const props = { isDataAvailable: true };
     const tree = SkinDeep.shallowRender(
       <AppContent
         available
-        authorized>
+        authorized {...props}>
```
U-DON: Since `props` only has one property in it, could just have:

```
<AppContent available authorized isDataAvailable>
```

In which case, there is a case for renaming that prop name...

bshyong (approved):

jkassemi:
Closing out in favor of the approach @U-DON recommends in https://github.com/department-of-veterans-affairs/vets.gov-team/issues/9220.

PR #6363, opened 2017-09-07
Wait to show the sign in link

cvalarida:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4628

If we're not logged, we may just be waiting for the profile to load, so this makes the sign in link not show up until we're sure we're not logged in.


jbalboni (approved):
Works for me locally

cehsu (approved):
LGTM.

PR #6366, opened 2017-09-07
Prefill notification

cehsu:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4330

### DIFF
--- a/src/js/common/schemaform/ObjectField.jsx
+++ b/src/js/common/schemaform/ObjectField.jsx
@@ -147,6 +147,7 @@ class ObjectField extends React.Component {
       ? this.props.formData
       : getDefaultFormState(schema, {}, definitions);
     const uiOptions = uiSchema['ui:options'] || {};
+    const hasApplicantPrefillInfo = !!(formData.veteranFullName || formData.claimantFullName);
 
     // description and title setup
     const showFieldLabel = uiOptions.showFieldLabel;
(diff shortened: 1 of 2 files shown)


src/js/common/schemaform/ObjectField.jsx
```diff
@@ -147,6 +147,7 @@ class ObjectField extends React.Component {
       ? this.props.formData
       : getDefaultFormState(schema, {}, definitions);
     const uiOptions = uiSchema['ui:options'] || {};
+    const hasApplicantPrefillInfo = !!(formData.veteranFullName || formData.claimantFullName);
```
cvalarida: I don't know that there's any reason they wouldn't have a name if they have any information to pre-fill, but this seems kind of brittle. I'd say it's not a blocker for this PR, though, since it'd be a really weird state when that happens.

Who knows where we get this information, what the information is, and which bits we'll always get? @J-Quag?

cvalarida (approved):
It looks good, but I'd like to know more about the information we _should_ be getting if we're getting a pre-filled form.

src/js/common/schemaform/ObjectField.jsx
jbalboni: I don't want to assume specific field names in the generic schemaform code. What is actually in the object when there's visible form data? Can we check that it's not empty?

cvalarida: Ohhh, yeah. This is just generic schemaform code...On that note, I'll second @jbalboni's concern. I'd like to compare `formData` to whatever we'll get when we have no pre-fill data returned from the api.

cehsu: Good point @jbalboni @cvalarida, changes made.

jbalboni (approved):

PR #6368, opened 2017-09-08
Consistent checkbox size

markgreenburg:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4360

Affects the Benefit Summary Letter checkable options

Goals were to 1.) increase size of all checkboxes to make them a bigger hit target 2.) improve consistency between these checkboxes and the standard USWDS checkboxes that are used elsewhere (including in the same component).

Due to the fact that the USWDS checkbox styling and behavior are heavily dependent on a sibling label, there were several potential ways to accomplish these goals while keeping the options rendered in a table (vs. a list). In the end, this approach seemed least disruptive.

Basically:
1. Introduce an empty `<label />` sibling after the `<input />` element to enable USWDS default styling / behavior on the input
2. Introduce slight style tweaks to the benefit options table to switch to a pointer cursor on the clickable area of the pre-existing label (and generally clean up the alignment of everything)

**Before:**
screen shot 2017-09-06 at 11 07 51 pm

**After:**
screen shot 2017-09-07 at 8 22 48 pm

### DIFF
--- a/src/js/letters/containers/VeteranBenefitSummaryLetter.jsx
+++ b/src/js/letters/containers/VeteranBenefitSummaryLetter.jsx
@@ -76,7 +76,7 @@ export class VeteranBenefitSummaryLetter extends React.Component {
                 name={key}
                 type="checkbox"
                 onChange={this.handleChange}/>
-              <label/>
+              <label />
             </th>
             <td><label id={`${key}Label`} htmlFor={key}>{optionText}</label></td>
           </tr>
--- a/src/js/letters/containers/VeteranBenefitSummaryLetter.jsx
+++ b/src/js/letters/containers/VeteranBenefitSummaryLetter.jsx
@@ -76,7 +76,7 @@ export class VeteranBenefitSummaryLetter extends React.Component {
                 name={key}
                 type="checkbox"
                 onChange={this.handleChange}/>
-              <label />
+              <label/>
             </th>
             <td><label id={`${key}Label`} htmlFor={key}>{optionText}</label></td>
           </tr>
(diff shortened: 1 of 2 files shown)


src/js/letters/containers/VeteranBenefitSummaryLetter.jsx
```diff
@@ -76,6 +76,7 @@ export class VeteranBenefitSummaryLetter extends React.Component {
                 name={key}
                 type="checkbox"
                 onChange={this.handleChange}/>
+              <label/>
```
raquelromano-gov: I haven't checked, but I'm wondering if this could cause accessibility issues.  It could be that the `aria-labelledby` attribute above overrides this blank label being associated with the input, but we should probably verify manually.

Suzanne wrote up some Voice Over instructions in case you haven't used it before: https://github.com/department-of-veterans-affairs/vets.gov-team/blob/master/Work%20Practices/Accessibility%20and%20508/Mac_VoiceOver_guide.md
I guess I'd just check that it doesn't announce the extraneous blank label and confuse the user.

markgreenburg: Yeah, was wondering about that but seemed to pass accessibility checks. I'll verify manually and share an update here shortly - thanks for the instructions.

markgreenburg:
@raquelromano-gov I just tested with screenreader on OS X and the behavior of the BSL checkboxes is consistent with using a screenreader on the example checkboxes here: https://standards.usa.gov/components/form-controls/#checkboxes

Looks as though the the labels are working correctly (i.e., the aria-* attributes are over-riding the blank labels).

src/js/letters/containers/VeteranBenefitSummaryLetter.jsx
raquelromano-gov: Thanks for checking!

raquelromano-gov (approved):

PR #6369, opened 2017-09-08
Prevent validation from running twice

cehsu:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4284

@jbalboni explained the issue thus:

The double validation thing was happening because of this sequence:
1. User makes a change
2. rjsf calls validation in its onChange method
3. rjsf calls our onChange, we run our updates in Redux
4. New data is sent to rjsf, it calls validation again in its componentWillReceiveProps method

This change stops step 4. That means that we’re running validation against data with the immediate change from the user, but not with any changes from custom updateSchema or required functions in the config. Theoretically, that could cause problems, but I have not been able to find an example of it doing so. I think rjsf is still running the basic JSON Schema validation twice, just not our custom validation.

This change is fine, since our custom validation is unlikely to change based on custom updates in the config. It would be more efficient if rjsf didn’t run the json schema validation twice, but that would require a change to rjsf itself.

### DIFF
--- a/src/js/common/schemaform/FormPage.jsx
+++ b/src/js/common/schemaform/FormPage.jsx
@@ -62,16 +62,18 @@ class FormPage extends React.Component {
   }
 
   onChange(formData) {
-    let newData = formData;
-    if (this.props.route.pageConfig.showPagePerItem) {
+    if (event && event.type !== 'message') {
+      let newData = formData;
+      if (this.props.route.pageConfig.showPagePerItem) {
       // If this is a per item page, the formData object will have data for a particular
       // row in an array, so we need to update the full form data object and then call setData
-      newData = _.set([this.props.route.pageConfig.arrayPath, this.props.params.index], formData, this.props.form.data);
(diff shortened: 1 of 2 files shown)


src/js/common/schemaform/FormPage.jsx
```diff
@@ -140,7 +140,7 @@ class FormPage extends React.Component {
           pagePerItemIndex={params ? params.index : undefined}
           uploadFile={this.props.uploadFile}
           prefilled={this.props.form.prefillStatus === PREFILL_STATUSES.success}
-          onChange={this.onChange}
+          onChange={_.once(this.onChange)} // Wrapped in _.once to prevent message events from retriggering validation
```
jbalboni: I still see validations happening twice. I think the issue is more likely in how RJSF handles state (i.e. it has local state and we also have Redux state).

cehsu: Thanks @jbalboni, `validate` is now wrapped in `once`, and only fires once per change.

jbalboni (approved):
LGTM

PR #6370, opened 2017-09-08
Fix claims message sequence

cehsu:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4445

I was not able to test this locally.


jbalboni (approved):
I tested it out in the review instance and it does remove the confusing message. I think it's still slightly confusing, since you might be sitting and waiting for claims with no indication that they're still being pulled. I think addressing that needs some design work, though, so I think this particular change is good.

cehsu:
Thanks @jbalboni, I've created a ticket for the additional work for this[ here](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4732).

PR #6372, opened 2017-09-08
Use the same name for yes/no radio buttons

cvalarida:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4609

I did this for "regular" radio buttons, but skipped the yes / no ones, so this takes care of them.


cvalarida:
Humbug. Fiddly tests. Looks like I'm still missing some, but the problem isn't immediately apparent to me, so I'm leaving this alone for now.

@jbalboni @cehsu If either of you want to pick this up and fix the tests, that'd be awesome, otherwise I'll look at it when I've got more time, but I need to move on to unicorns stuff for now.

jbalboni:
@cvalarida I can look at it.

cehsu (approved):
LGTM.

PR #6374, opened 2017-09-08
Add whitespace before anchor tag by breaking line on anchor tag inste…

markgreenburg:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4664

Introduces whitespace between text and beginning of inline anchor tag by moving the line break to be within anchor element instead of within text (which strips the EOL whitespace).


markgreenburg:
@U-DON Was too tired when I pushed update to the original PR last night and ended up duplicating some commits that already went to master, so decided to delete that branch and re-create this change in a new branch since it's such a small thing anyway (rather than try to clean up the commit history on the old branch). Sorry for the confusion :/

U-DON (approved):
That's okay. Looks good!

PR #6379, opened 2017-09-11
Check if veteran or dependent

markgreenburg:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3890

This PR simply adds isVeteran and veteranStatus returns from vets-api to store, connects it to the BSL component, and removes the hard-coded veteran status in this component's rendering.

This allows us to conditionally render BSL options / text based on whether the logged-in user is a veteran or a dependent.

Also includes some related tests.

### DIFF
--- a/test/e2e/login-helpers.js
+++ b/test/e2e/login-helpers.js
@@ -39,6 +39,10 @@ function initUserMock(token, level) {
             gender: 'F',
             birth_date: '1985-01-01'
           },
+          veteran_status: {
+            status: "OK",
+            is_veteran: true,
+          },
           in_progress_forms: [],
           prefills_available: [],
           services: ['facilities', 'hca', 'edu-benefits', 'evss-claims', 'user-profile', 'rx', 'messaging'],
--- a/test/e2e/login-helpers.js
+++ b/test/e2e/login-helpers.js
@@ -40,7 +40,7 @@ function initUserMock(token, level) {
             birth_date: '1985-01-01'
           },
           veteran_status: {
-            status: "OK",
+            status: 'OK',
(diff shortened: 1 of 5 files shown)


test/e2e/login-helpers.js
```diff
@@ -39,6 +39,10 @@ function initUserMock(token, level) {
             gender: 'F',
             birth_date: '1985-01-01'
           },
+          veteran_status: {
+            status: 'OK',
```
cvalarida: Is there a list somewhere of all the possible statuses?

cvalarida (approved):

test/e2e/login-helpers.js
raquelromano-gov: These look like the common vets-api error statuses defined [here](https://github.com/department-of-veterans-affairs/vets-api/blob/master/lib/common/client/concerns/service_status.rb) and shared by [EVSS responses](https://github.com/department-of-veterans-affairs/vets-api/blob/master/lib/evss/response.rb#L26) and [MVI responses](https://github.com/department-of-veterans-affairs/vets-api/blob/master/lib/mvi/responses/find_profile_response.rb#L60):  OK, NOT_FOUND, SERVER_ERROR, NOT_AUTHORIZED

These should probably be defined in a constants.js file, perhaps in src/js/common/utils

Sitewide ticket?

cvalarida: Sounds like a site-wide ticket to me. 👍

raquelromano-gov: https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4849

PR #6386, opened 2017-09-12
Save confirmation redesign

cehsu:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4734

screen shot 2017-09-12 at 5 07 23 pm

screen shot 2017-09-12 at 5 11 02 pm


src/js/common/schemaform/FormSaved.jsx
```diff
@@ -50,7 +50,10 @@ class FormSaved extends React.Component {
         <div className="usa-alert usa-alert-info">
           <div className="usa-alert-body">
             <strong>Your application has been saved.</strong><br/>
-            {!!lastSavedDate && !!expirationDate && <p>Last saved on {moment(lastSavedDate).format('M/D/YYYY [at] h:mm a')} <span className="schemaform-sip-expires">Your saved application will expire in {dateDiffDesc(expirationDate)}</span>.</p>}
+            {!!lastSavedDate && !!expirationDate && <p className="saved-form-metadata-container">
+              <span className="saved-form-metadata">Last saved on {moment(lastSavedDate).format('M/D/YYYY [at] h:mm a')}</span>
+              <p className="expires-container">Your saved application <span className="expires">will expire in {dateDiffDesc(expirationDate)}.</span></p>
+              </p>}
             {success}
             If you’re logged in through a public computer, please sign out of your account before you log off to keep your information secure.
```
cehsu: @stacyhlinn do users no longer need this information?

jbalboni:
@cehsu It looks like there are still changes left in order to match this screenshot: https://user-images.githubusercontent.com/27507402/30297633-3c4006aa-96fd-11e7-8b7a-4cd7a48c7f35.png

cehsu:
Thanks @jbalboni, some changes made. I want to double check with @peggygannon @stacyhlinn about removing the information explaining to users that they should sign out of public computers, as this seems helpful?

src/js/common/schemaform/FormSaved.jsx
```diff
@@ -39,18 +39,22 @@ class FormSaved extends React.Component {
...
     const { profile } = this.props.user;
     const lastSavedDate = this.props.lastSavedDate;
-    const prefillAvailable = !!(profile && profile.prefillsAvailable.includes(this.props.formId));
+    const formId = this.props.formId;
+    const prefillAvailable = !!(profile && profile.prefillsAvailable.includes(formId));
     const verifiedAccountType = 3;// verified ID.me accounts are type 3
     const notVerified = profile.accountType !== verifiedAccountType;
     const { success } = this.props.route.formConfig.savedFormMessages || {};
-    const expirationDate = moment.unix(this.props.expirationDate);
+    const expirationDate = moment.unix(this.props.expirationDate).format('D/M/YYYY');
```
jbalboni: I think the month and day are swapped here.

src/js/common/schemaform/SaveInProgressIntro.jsx
```diff
@@ -19,14 +19,16 @@ export default class SaveInProgressIntro extends React.Component {
...
-        const expirationDate = moment.unix(savedForm.metadata.expires_at);
+        const expirationDate = moment.unix(savedForm.metadata.expires_at).format('M/D/YYYY');
 
         alert = (
           <div>
             <div className="usa-alert usa-alert-info no-background-image schemaform-sip-alert">
-              <div style={{ paddingBottom: '8px' }}>Application status: <strong>In progress</strong></div>
-              <br/>
-              <div>Last saved on {savedAt.format('MM/DD/YYYY [at] h:mm a')} <span className="schemaform-sip-expires">Your saved application will expire in {dateDiffDesc(expirationDate)}</span>.</div>
+              <div style={{ paddingBottom: '16px' }}>Application status: <strong>In progress</strong></div>
```
jbalboni: Can we move this inline style into a css class?

src/js/common/schemaform/SaveInProgressIntro.jsx
```diff
@@ -19,14 +19,16 @@ export default class SaveInProgressIntro extends React.Component {
...
+              <div className="saved-form-metadata-container">
+                <span className="saved-form-metadata">Last saved on {savedAt.format('MM/DD/YYYY [at] h:mm a')}</span>
```
jbalboni: This is a different date format that we're using in other places above, should they match?

src/js/hca/config/form.js
```diff
@@ -124,9 +123,7 @@ const formConfig = {
   migrations,
   savedFormMessages: {
     notFound: 'Please start over to apply for health care.',
-    noAuth: 'Please sign in again to resume your application for health care.',
-    success: resumeMessage,
-    startOver: 'This will remove anything you have put into the Health Care Application.'
```
jbalboni: Do we want to remove this custom start over text? That shows up in the start over modal, I think?

src/js/user-profile/helpers.jsx
```diff
@@ -1,18 +1,28 @@
...
+  '22-5490': 'education benefits',
+  '22-5495': 'education benefits',
   '40-10007': 'pre-need determination of eligibility in a VA National Cemetery'
 };
 
+export const formTitles = Object.keys(formBenefits).reduce((titles, key) => {
+  titles[key] = key === '40-10007' // eslint-disable-line no-nested-ternary no-param-reassign  
+    ? formBenefits[key]
+    : key === '1010ez'
+      ? `${formBenefits[key]} (10-10EZ)`
```
jbalboni: Instead of this extra conditional, you could always upper case the key: ``${formBenefits[key]} (${key.toUpperCase()})``

src/sass/user-profile.scss
```diff
@@ -56,52 +56,64 @@
...
   }
 
-  .saved-form-title {
+  .saved {
     color: $color-primary;
   }
 
+  .form-title {
+    @include media ($medium-large-screen) {
+    padding-right: 24px;
```
jbalboni: Indentation looks a little odd here.

jbalboni (approved):
LGTM with some minor comments, pending confirmation from Stacy and Peggy about the public computer info.

stacyhlinn:
Good catch @cehsu. Let's keep the message re: logging out at a public computer.

src/js/user-profile/helpers.jsx
cehsu: Thanks for the recommendation @jbalboni. It looks like a dash would need to be inserted into the formID, is that an option?

src/js/hca/config/form.js
cehsu: Thanks for pointing that out @jbalboni, @stacyhlinn @peggygannon could you please share your thoughts on whether the HCA start over model should have a custom modal message, or whether it should use the generic message that the rest of the forms use?
Generic
screen shot 2017-09-14 at 11 13 51 am

Current
screen shot 2017-09-14 at 11 17 21 am

src/sass/user-profile.scss
cehsu: Good catch.

src/js/common/schemaform/FormSaved.jsx
cehsu: Good catch, changed.

src/js/common/schemaform/SaveInProgressIntro.jsx
cehsu: Good idea, changed.

src/js/user-profile/helpers.jsx
```diff
@@ -1,18 +1,28 @@
...
+  '22-5490': 'education benefits',
+  '22-5495': 'education benefits',
   '40-10007': 'pre-need determination of eligibility in a VA National Cemetery'
 };
 
+export const formTitles = Object.keys(formBenefits).reduce((titles, key) => {
+  titles[key] = key === '40-10007' // eslint-disable-line  
+    ? formBenefits[key]
+    : key === '1010ez'
+      ? `${formBenefits[key]} (10-10EZ)`
```
jbalboni: Ah, I see your point about the dash. In that case, I would just remove the nested ternary (those are hard to read) in favor of something like:

```
let formNumber;
if (key === '40-10007') {
  formNumber = '';
} else if (key === '1010ez') {
  formNumber = ' (10-10EZ)';
} else {
  formNumber = ` (${key})`;
}

return `${formBenefits[key]}${formNumber}`;

src/js/hca/config/form.js
stacyhlinn: The more specific one (current version) seems less confusing to me. Would rely on @peggygannon to make the right judgement though.

jbalboni: It would be nice to not have custom text here, from a technical perspective, since we have to remember to pull it in in several places now.

stacyhlinn: Makes sense! A generic one would work too. Just want to make sure Peggy has seen and reviewed the message.

src/js/common/schemaform/SaveInProgressIntro.jsx
cehsu: Great idea, this style has been moved into `schemaform-sip-alert-title`.

src/js/user-profile/helpers.jsx
cehsu: Thanks @jbalboni, that is much more readable.

src/js/hca/config/form.js
peggygannon: @jbalboni @stacyhlinn @cehsu I think we should use the generic message, as we do for other forms. 
--Can we please make the below edit to the bold sentence? ("delete" is more clear/direct than "erase"--which reminds me of school chalkboard!) 
--Can this edited message be applied across all the forms? 

Starting over will delete your in-progress form. 
[[rest of message is OK as is]]

cehsu: Thanks for the insights @peggygannon, we can make that change, and it will be on a shared component that all of the forms use.

PR #6388, opened 2017-09-12
New signin page

bshyong:
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4595

- New sign in page
- New headerless react layout

TODO in #4599:
- refactor header signin/sign up buttons
- remove deprecated and duplicated code, consolidate signin & login folders

image


melwoodard:
@bshyong is this ready for a design review?

bshyong:
Yes let's do it.

melwoodard (changes requested):
A few formatting tweaks:

sign_in__vets_gov

And a couple layout issues on mobile:

sign_in__vets_gov

sign_in__vets_gov

bshyong:
Changes made!

src/sass/login.scss
```diff
@@ -1,5 +1,190 @@
...
+    border-color: $color-gray-lighter;
+  }
+
+  @include media-maxwidth($small-screen) {
+    .hide-for-small {
+      display: none;
+    }
+
+    .explanation-content {
+      padding: 1em 2em;
```
melwoodard: I think you can remove this - I see you put it in to make the text break nicely, but it pads the text too much on  mobile. This content is subject to change any way so I think we don't need to fuss over it too much.

src/sass/login.scss
```diff
@@ -1,5 +1,190 @@
...
+
+      img {
+        height: inherit;
+        vertical-align: baseline;
+        padding: 0 0.25em;
+      }
+    }
+
+    .signin-actions {
+      padding: 2em;
```
melwoodard: @bshyong the intention was for the larger side padding to only apply to large screens. Is it possible to add an include here so that large screen will be `2em 2em 1em`, and small will be `2em 1em 1em`?

melwoodard (changes requested):
Added a couple more small tweaks.

src/sass/login.scss
```diff
@@ -1,5 +1,190 @@
...
+        p:first-of-type {
+          margin-top: 0.5em;
+          color: $color-gray-dark;
+        }
+      }
+    }
+  }
+
+  .explanation-content {
+    padding: 0 2em;
```
melwoodard: Padding on the buttons looks great. I think this line is still adding padding to the right-column content.

image

bshyong:
@U-DON how does this look? would like to merge this into kudos-identity branch soon

src/js/signin/components/SearchHelpSignIn.jsx
```diff
@@ -0,0 +1,75 @@
...
+      content = (<SignInProfileMenu
+        clickHandler={() => {
+          this.props.onClickSearchHelpSignIn('account', !login.utilitiesMenuIsOpen.account);
+        }}
+        greeting={greeting}
+        isOpen={login.utilitiesMenuIsOpen.account}
+        onUserLogout={this.props.onUserLogout}/>);
+    } else {
+      content = (<div>
+        <a href="#" onClick={this.props.onUserLogin}>Sign In</a><span className="signin-spacer">|</span><a href="#" onClick={this.props.onUserSignup}>Sign up</a>
```
U-DON: "Sign In" is both capitalized whereas "Sign up" only has the first word capitalized. Should be consistent.

src/js/signin/components/SearchHelpSignIn.jsx
```diff
@@ -0,0 +1,75 @@
...
+import HelpMenu from '../../common/components/HelpMenu';
+import SearchMenu from '../../common/components/SearchMenu';
+import SignInProfileMenu from './SignInProfileMenu';
+
+import { updateLoggedInStatus, toggleSearchHelpUserMenu } from '../actions';
+
+class SearchHelpSignIn extends React.Component {
+  render() {
+    let content;
+    const login = this.props.login;
```
U-DON: Minor improvement could be made here; destructuring the `this.props.login` to pull out `currentlyLoggedIn` and `utilitiesMenuIsOpen`. Also destructuring `profile` out of `this.props`.

src/js/signin/components/SearchHelpSignIn.jsx
```diff
@@ -0,0 +1,75 @@
...
+  };
+};
+
+const mapDispatchToProps = (dispatch) => {
+  return {
+    onUpdateLoggedInStatus: (update) => {
+      dispatch(updateLoggedInStatus(update));
+    },
+    onClickSearchHelpSignIn: (menu, isOpen) => {
+      dispatch(toggleSearchHelpUserMenu(menu, isOpen));
```
U-DON: I don't think mapping the actions to differently named functions is useful for understanding what's happening, so I'd like to see `mapDispatchToProps` just directly map the actions to the same names.

```
  const mapDispatchToProps = {
    updateLoggedInStatus,
    toggleSearchHelpUserMenu
  };
```

src/js/signin/containers/Signin.jsx
```diff
@@ -0,0 +1,143 @@
+import React from 'react';
+
+class Main extends React.Component {
```
U-DON: `class Signin`?

src/js/signin/containers/Signin.jsx
```diff
@@ -0,0 +1,143 @@
...
+              </div>
+            </div>
+          </div>
+        </div>
+      </main>
+    );
+  }
+}
+
+export default Main;
```
U-DON: `export default SignIn`?

src/js/signin/containers/Signin.jsx
```diff
@@ -0,0 +1,143 @@
...
+              <hr/>
+              <div className="fed-warning">
+                <p>
+                  By entering authentication information, you are attempting to access a United States Federal Government information system. This system is for the use of authorized users only.
+                </p>
+                <p>
+                  System usage may be monitored, recorded, and subject to audit. By accessing this system, you are consenting to have your activiteis monitored, recorded, and made available to auditors or law enforcement officials.
+                </p>
+                <p>
+                  Unauthorized use of this information system or use in excess of your approved authority is prohibited, and may be subject ot disciplinary action including criminal or civil penalties.
```
U-DON: Typo: "subject __to__ disciplinary action".

src/js/signin/containers/Signin.jsx
```diff
@@ -0,0 +1,143 @@
...
+                  We're here Monday – Friday, 8:00am – 8:00pm (ET).
+                </p>
+              </div>
+              <hr/>
+              <div className="fed-warning">
+                <p>
+                  By entering authentication information, you are attempting to access a United States Federal Government information system. This system is for the use of authorized users only.
+                </p>
+                <p>
+                  System usage may be monitored, recorded, and subject to audit. By accessing this system, you are consenting to have your activiteis monitored, recorded, and made available to auditors or law enforcement officials.
```
U-DON: Typo: "activiteis" -> "activities".

src/js/signin/containers/Signin.jsx
```diff
@@ -0,0 +1,143 @@
...
+              </div>
+            </div>
+          </div>
+          <div className="row">
+            <div className="columns small-12">
+              <div className="help-info">
+                <h4>Having trouble signing in?</h4>
+                <p><a href="/faq">Get answers to Frequently Asked Questions</a></p>
+                <p>
+                  Call the Vets.gov Help Desk at <a href="tel:+18555747286">1-855-574-7286</a> (TTY: <a href="tel:+18008294833">1-800-829-4833</a>).<br/>
```
U-DON: Hm... Super small thing, but we have inconsistencies about having spaces between the days and times and the hyphens. Should make sure they're all the same at some point. @melwoodard

src/js/signin/reducers/login.js
```diff
@@ -0,0 +1,59 @@
...
+
+function closeAllMenus(menuState) {
+  const menus = menuState.utilitiesMenuIsOpen;
+
+  Object.keys(menus).forEach(menu => {
+    menus[menu] = false;
+  });
+}
+
+function loginStuff(state = initialState, action) {
```
U-DON: `loginStuff`...

src/js/signin/containers/Signin.jsx
U-DON: Actually, maybe make this class `SignIn` (with the capitalized "In") for proper camel-casing.

src/js/signin/containers/Main.jsx
```diff
@@ -0,0 +1,167 @@
...
+        }
+      }
+    } else {
+      this.props.onUpdateLoggedInStatus(false);
+    }
+  }
+
+  render() {
+
+    if (this.props.verify) {
```
U-DON: Where is `this.props.verify` coming from?

U-DON: Instead of returning the following in a separate conditional block, could just return one block:

```
const currentlyLoggedIn =
  this.props.verify
  ? undefined
  : this.props.login.currentlyLoggedIn;

return (
  <SignIn
    currentlyLoggedIn={currentlyLoggedIn}
    handleSignup={...}
    handleLogin={...}/>
);
```

src/js/signin/containers/Main.jsx
```diff
@@ -0,0 +1,167 @@
...
+      dispatch(updateVerifyUrl(update));
+    },
+    onUpdateLogoutUrl: (update) => {
+      dispatch(updateLogoutUrl(update));
+    },
+    onUpdateLoggedInStatus: (update) => {
+      dispatch(updateLoggedInStatus(update));
+    },
+    dispatch
+  };
```
U-DON: Not a fan of needlessly renaming all the actions here either. Would rather just map them directly to the same names. Also, mapping `dispatch` is redundant since it should already be available in props, I believe...

U-DON (approved):
I noticed you copied over some things from the `login` app.

In `SearchHelpSignIn`, I wondered why you removed the null content for the case when `this.props.profile.loading`. Do we want to show the "Sign In/Sign Up" links while the props are loading?

I also suggested a couple of minor refactors in similar files since they could use some cleanup while being moved over.

Overall, LGTM. Assuming most of these are just copied over from `login` with the exception of the SignIn component, which hasn't been hooked up with the finalized login/signup actions yet.

src/js/signin/containers/Signin.jsx
melwoodard: I think this is the format that the content team prefers. We can circle back at a later point to address any that are inconsistent with this

src/sass/login.scss
```diff
@@ -1,5 +1,186 @@
...
+    border-color: $color-gray-lighter;
+  }
+
+  @include media-maxwidth($small-screen) {
+    .hide-for-small {
+      display: none;
+    }
+
+    .help-info {
+      padding: 2em !important;
```
melwoodard: @bshyong one more...this padding still needs to go. Then I _think_ we should be good!

image

src/js/signin/containers/Main.jsx
bshyong: A little explanation here—things might have gotten a little mixed up with all the branching. I am making the `Verify` and `Signin` pages into components, while the Main container holding the shared logic around fetching the tokens, urls, etc.  So the `verify` prop on `Main` tells it whether to render the Verify page.

Would it be helpful to pull in the `Verify` page branch into this PR and review it together? There are a bunch of other changes and logic improvements on the Verify branch

src/js/signin/components/SearchHelpSignIn.jsx
bshyong: This should be refactored but it's currently copied 1:1 from the existing `login` folder; will note this for refactoring in the future

bshyong:
Combined with verify page PR  https://github.com/department-of-veterans-affairs/vets-website/pull/6389

PR #6390, opened 2017-09-12
Remove old form related code

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4553

This removes all the 1990 and hca related form code. There are probably some generic form components that are no longer used, but I chose to leave them in case they're useful in other apps.

I'm going to rename 1990 to 1990-rjsf in a second PR linked to the above issue.


cvalarida (approved):
_[Insert cyberman joke here]_

PR #6391, opened 2017-09-12
Remove dark-blue color property for intro text

ncksllvn:
Removes the intro text's color property so that it defaults to the color of the body text. Resolves [Issue 4498](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4498).


melwoodard (approved):
Looks good!

PR #6395, opened 2017-09-12
Error message button fix

cvalarida:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4679

This affects the generic error-messages.jsx, so I'm requesting the review of all of you!

Also, I can't log in to reproduce the error (this is an ongoing issue), so could one of you guys try it out? I think if you go to `localhost:3001/letters` while logged in, it should get you to that error page.

### DIFF
--- a/src/js/common/utils/error-messages.jsx
+++ b/src/js/common/utils/error-messages.jsx
@@ -6,7 +6,7 @@ export const systemDownMessage = (
     <div className="small-12 columns">
       <div className="react-container">
         <h3>Sorry, our system is temporarily down while we fix a few things. Please try again later.</h3>
-        <a href="/"><button>Go Back to Vets.gov</button></a>
+        <a href="/" className="usa-button-primary">Go Back to Vets.gov</a>
       </div>
     </div>
   </div>


U-DON (approved):
LGTM.

jbalboni:
@cvalarida The change seems fine, but I can't get that error message to display locally.

cehsu (approved):
LGTM.

cehsu:
@cvalarida this is what I see currently when navigating to `/letters`:
screen shot 2017-09-14 at 9 35 36 am

cvalarida:
@cehsu Yeah...that's an issue with the letters app itself. You logged in with a user that isn't mocked in the right services (mvi & evss, if I remember correctly).

The trouble I was having before is when I was logged in but the `/v0/user` endpoint was sending back `SERVER_ERROR` for...I forget exactly, but whatever ends up in `store.user.profile.status`. The `RequiredLoginView` checks to see if the profile status is a server error, and if it is, it displays a generic error message with that button I changed.

It's all kind of convoluted. 😛

PR #6396, opened 2017-09-13
Fix unknown prop in breadcrumbs

bshyong:
resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4806


U-DON (approved):
LGTM.

PR #6299, opened 2017-08-29
WIP: Upgrade USWDS to 1.3.0

cehsu:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1463

Errors encountered in upgrading USWDS:

-SASS can't easily be used when installing via npm + git (documented [here](https://github.com/18F/web-design-standards/issues/2032))

There are two options indicated in the above issue on this error. One is to include node modules in the sass path, the other is to install these as dependencies rather than dev dependencies.

```
[e_uswds-upgrade-version-ceh-QH4ZB256PMXBOICES3NH7NINIMNBPJHMI7GPNRRGHBLIYKQG6ZLQ] Running shell script
+ cd /application
+ npm --no-color run lint
npm info it worked if it ends with ok
npm info using npm@3.10.10
npm info using node@v6.11.1
npm info lifecycle vets-website@1.0.0~prelint: vets-website@1.0.0
npm info lifecycle vets-website@1.0.0~lint: vets-website@1.0.0

> vets-website@1.0.0 lint /application
> npm run lint:js && npm run lint:sass

npm info it worked if it ends with ok
npm info using npm@3.10.10
npm info using node@v6.11.1
npm info lifecycle vets-website@1.0.0~prelint:js: vets-website@1.0.0
npm info lifecycle vets-website@1.0.0~lint:js: vets-website@1.0.0

> vets-website@1.0.0 lint:js /application
> eslint --quiet --ext .js --ext .jsx .
/application/config/webpack.config.js
  7:25  error  Unable to resolve path to module 'bourbon'       import/no-unresolved
  8:22  error  Unable to resolve path to module 'bourbon-neat'  import/no-unresolved

✖ 2 problems (2 errors, 0 warnings)


npm info lifecycle vets-website@1.0.0~lint:js: Failed to exec lint:js script
npm ERR! Linux 4.4.8-20.46.amzn1.x86_64
npm ERR! argv "/usr/local/bin/node" "/usr/local/bin/npm" "run" "lint:js"
npm ERR! node v6.11.1
npm ERR! npm  v3.10.10
npm ERR! code ELIFECYCLE
npm ERR! vets-website@1.0.0 lint:js: `eslint --quiet --ext .js --ext .jsx .`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the vets-website@1.0.0 lint:js script 'eslint --quiet --ext .js --ext .jsx .'.
npm ERR! Make sure you have the latest version of node.js and npm installed.
[...]

PR #6400, opened 2017-09-13
Fix for button border inside an accordion

ncksllvn:
Currently, there is a style applied to all buttons inside an accordion that changes the `box-shadow` property when the button is in the `focus` state:

```
.usa-accordion button:focus, .usa-accordion-bordered button:focus {
    box-shadow: 0 0 3px #3e94cf, 0 0 7px #3e94cf;
}
```

This isn't desirable for the outline-style buttons. This problem is really in the USWDS repo and has since been [fixed](https://github.com/18F/web-design-standards/commit/d77ee35e26a21d805367c3dddb9e2bb3168b9c37), but our version is frozen in our [package.json](https://github.com/department-of-veterans-affairs/vets-website/blob/master/package.json#L182), probably for good reason, so we need an override in our own style.

Connects to department-of-veterans-affairs/vets.gov-team/issues/4662.

### DIFF
--- a/src/sass/modules/_m-button.scss
+++ b/src/sass/modules/_m-button.scss
@@ -25,8 +25,9 @@ button,
   }
 }
 
-.usa-button-outline {
-  .usa-accordion &:focus {
+// @todo Remove this once the USWDS upgrade is complete.
+.usa-accordion {
+  .usa-button-outline:focus {
     box-shadow: inset 0 0 0 2px $color-primary-darkest, 0 0 3px $color-focus, 0 0 7px $color-focus;
   }
 }
--- a/src/sass/modules/_m-button.scss
+++ b/src/sass/modules/_m-button.scss
@@ -25,6 +25,12 @@ button,
   }
 }
 
+.usa-button-outline {
+  .usa-accordion &:focus {
+    box-shadow: inset 0 0 0 2px $color-primary-darkest, 0 0 3px $color-focus, 0 0 7px $color-focus;
+  }
+}
+
 .usa-button-outline-exit {
   &::after {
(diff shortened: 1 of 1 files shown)


src/sass/modules/_m-button.scss
```diff
@@ -25,6 +25,12 @@ button,
   }
 }
 
+.usa-button-outline {
+  .usa-accordion &:focus {
+    box-shadow: inset 0 0 0 2px $color-primary-darkest, 0 0 3px $color-focus, 0 0 7px $color-focus;
+  }
+}
```
ncksllvn: With nesting something higher on the visual hierarchy with something lower, it for some reason made sense to me in my head to group it like that since that's the button module, but I don't feel strongly either way so I don't mind reversing it.

cehsu: Thanks for pointing this out @ncksllvn, since you mentioned that this has been resolved in newer versions of USWDS (which we are upgrading [to here](https://github.com/department-of-veterans-affairs/vets-website/pull/6299)), this way of structuring the nesting -- to accommodate additional overrides -- may not be necessary. Also, it may be helpful to include a comment to remove this once the upgrade has been performed.

cehsu (approved):
LGTM. Had a few minor comments.

src/sass/modules/_m-button.scss
ncksllvn: Sorry, saw this just now - but thanks @cehsu! I'll add a comment to that about the upgrade in just a sec!

ncksllvn:
PTAL @cehsu

ncksllvn:
Made changes as requested. Thanks @cehsu!

PR #6403, opened 2017-09-13
Update letters content

markgreenburg:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4463, https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4464

Updates the descriptions of several letters, as well as some general 'more resources' type information in the letters app.

### DIFF
--- a/src/js/letters/containers/LetterList.jsx
+++ b/src/js/letters/containers/LetterList.jsx
@@ -74,7 +74,7 @@ export class LetterList extends React.Component {
         <br/>
         <h4>Can’t find what you’re looking for?</h4>
         <p>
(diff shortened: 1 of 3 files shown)


cvalarida (approved):
Looks good! I'm not sure where the Commissary Card is retitled to Commissary Letter, but I looked at it locally (on another branch) and it looks like it's called Commissary Letter, so no gripes here--just mild confusion.

markgreenburg:
Yep, that's a good point. I checked on master and that letter's already coming back as Commissary Letter so no changes made to that part. Thanks for checking!

PR #6404, opened 2017-09-14
Additional logic on new sign-in and identity-proofing flows

bshyong:
Additional changes and clean up

- supports new API endpoints for pulling multiple auth URLs
- supports new API endpoint for identity proofing

- Ensure existing sign-in flows work; resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4804
- Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4599; need to verify that all scenarios in https://github.com/department-of-veterans-affairs/vets.gov-team/blob/master/Products/Identity/User%20Login/discovery/User%20Scenarios.md work with the proper test users (which users can we use to test these scenarios?)
- Ensure DSLogon flow works; resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4596


bshyong:
@NatalieEMoore or @leannammiller do we know which test users are suitable for testing the scenarios here?: https://github.com/department-of-veterans-affairs/vets.gov-team/blob/master/Products/Identity/User%20Login/discovery/User%20Scenarios.md

U-DON (approved):
LGTM. Why not keep a copy of the actions and reducers in `signin`, though? Seems like it would be easier to refactor later with a copy rather than having to modify the existing ones in `login`.

src/js/login/containers/Main.jsx
```diff
@@ -14,7 +14,6 @@ class Main extends React.Component {
     this.setMyToken = this.setMyToken.bind(this);
     this.getLogoutUrl = this.getLogoutUrl.bind(this);
     this.getLoginUrl = this.getLoginUrl.bind(this);
-    this.getVerifyUrl = this.getVerifyUrl.bind(this);
```
U-DON: Did you mean to remove this from the original Main component in `login`?

bshyong:
Given that the actions are shared, I figure it's easier to have them all in one place. I cant think of a case right now where we would modify the actions in one place but not the other. 

That said, maybe we want to have a copy if we want to delay refactoring the tests...

Let me know if you have other thoughts here. The copied code was distracting so I decided to consolidate - Im not sure what you mean when you say it might be easier to refactor with a copy

src/js/login/containers/Main.jsx
bshyong: Good catch will add this in. It's also included in the login component but we need it for this component too in case a user goes straight to the /signin page (the header nav would never have been loaded in this case)

bshyong: I misspoke; I looked through and did mean to remove this. we only need this one in the `Main` container in the `signin` folder because only that one uses the VerifyScreen where this is applicable. There is no need to fetch the verify URL for the `login` folder because it is never used

U-DON: Shouldn't we probably keep this around anyway, at least for soft launch?

PR #6412, opened 2017-09-14
Add classes in/out for animating the drawer

ncksllvn:
On the static site on mobile (screenshot), the sidebar currently just appears and disappears without any effect. This PR adds a slide-in and slide-out effect from the right side.

Connects to department-of-veterans-affairs/vets.gov-team/issues/3191.

screen shot 2017-09-14 at 4 11 54 pm

### DIFF
--- a/src/js/legacy/sidebar-navigation.js
+++ b/src/js/legacy/sidebar-navigation.js
@@ -30,16 +30,34 @@ class SideBarMenu {
   }
 
   openMenu(trigger) {
-    this.getMenu(trigger).classList.add('va-sidebarnav--opened');
+    const menu = this.getMenu(trigger);
+
+    menu.classList.add('va-sidebarnav--opened');
     document.body.classList.add('va-pos-fixed');
+
+    // Delay the animation classes slightly so that switching the display property
+    // to block doesn't cause the animations to skip.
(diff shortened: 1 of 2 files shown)


U-DON (approved):
LGTM. There is a CSS-only solution if you're interested.

Instead of using `display`, use `visibility`. Also, if you don't want to keep track of the `$sidebar-width`, you can also use `transform: translateX(100%)` on `.va-sidebarnav > div` and `transform: translateX(0)` on `.va-sidebarnav--opened > div`.

Also have to keep `visibility: visible` on `.va-sidebarnav > div` for the close animation to work.

ncksllvn:
@U-DON thanks! I'll check out that CSS-only solution. If I recall correctly, doesn't `visibility` still cause the element to take up space, therefore stopping the use form clicking other elements on the page? I tried that briefly. Good idea with the `transform` property!

ncksllvn:
@U-DON Wow, I stand corrected with the `visibility` property! These were awesome suggestions. I had no idea the visibility property worked like that. Thanks man!

cehsu (approved):

ncksllvn:
Thanks @cehsu!

PR #6413, opened 2017-09-14
Move "other ways to apply" options into list

ncksllvn:
Instead of just using headings, the options under the "Other ways to apply" have been moved into a standard bulleted list for clarity purposes.

Connects to department-of-veterans-affairs/vets.gov-team/issues/4767.

### DIFF
--- a/content/pages/burials-and-memorials/survivor-and-dependent-benefits/burial-costs.md
+++ b/content/pages/burials-and-memorials/survivor-and-dependent-benefits/burial-costs.md
@@ -64,6 +64,9 @@ You may need a copy of:
 
 #### Other ways to apply
 
+<ul>
+<li>
+
 ##### By mail
 
 Apply by mail using the Application for Burial Benefits (VA Form 21P-530). [Download VA Form 21P-530](http://www.vba.va.gov/pubs/forms/VBA-21P-530-ARE.pdf).
@@ -71,6 +74,9 @@ Apply by mail using the Application for Burial Benefits (VA Form 21P-530). [Down
 Mail the application and other paperwork to your local regional benefits office. [Find your local regional benefits office](/facilities).
 
 If you have questions, call <a href="tel:+18008271000">800-827-1000</a> (our TDD number for the hearing-impaired is 711) or call your regional benefits office.
+
+</li>
+</ul>
 
 #### What are the burial allowance amounts for a service-connected death?
 
--- a/content/pages/disability-benefits/apply.md
+++ b/content/pages/disability-benefits/apply.md
@@ -62,13 +62,22 @@ For all disability claims, please provide:
 
 ### Other ways to apply
 
-#### Work with a trained professional
+<ul>
+<li>
+
+<strong>Work with a trained professional</strong>
 
 You can work with a trained professional who can help you file a claim.  [Find an accredited representative](/disability-benefits/apply/help/index.html).
 
-#### Apply in person
+</li>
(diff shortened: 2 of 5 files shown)


phiden:
Would love to see this on a review instance -- code looks good, though

phiden (approved):

melwoodard:
@phiden it should be up here, no? https://vetsgov-pr-6413.herokuapp.com/health-care/apply/

Also should this PR cover all the instances of "other ways to apply" content on the site? This structure appears for example on `/education/apply`, `/disability-benefits/apply`, etc.

phiden:
@melwoodard wompwomp. Yes, thanks -- i was definitely looking for a page in the app. So --  LGTM!

ncksllvn:
@melwoodard I don't mind expanding the PR to cover those as well; I was just keeping it confined to the original [issue](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4767). Should I go ahead and update those pages as well?

melwoodard:
@ncksllvn that would be great. Somehow we lost the tracking to this original issue, but it should definitely be consistent sitewide: https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3078

ncksllvn:
@melwoodard @phiden Done! So in total, this includes moving the content of the "Other ways to apply" sections into a list for the following pages:

- `/health-care/apply/`
- `/education/apply`
- `/disability-benefits/apply`
- `/burials-and-memorials/survivor-and-dependent-benefits/burial-costs/`

Please let me know if I missed any!

melwoodard (approved):
Looks good!

leannammiller:
@ncksllvn 

image

ncksllvn:
@leannammiller It made sense to me to still have it as a list, since it's sort of set up to be displayed like the other pages in this PR (even if it's only a one-item list.) Would you rather that I remove that? Or would it make sense to rename that section to "Apply by mail" and change it to a list later if other options become available?

ncksllvn:
@leannammiller oh wait, did you mean just remove the bullets altogether? They're shown in the Marvel mockup so I thought they were supposed to be there.

leannammiller:
@bethpotts ^

bethpotts:
I have some concerns about this approach of making the elements under "Other ways to apply" be a bulleted list. The first issue is when there's only one other way to apply--then we have a lonely bullet. Our site style is to only use bullets when there's more than one.

The second issue is how this will look when taking the page into account holistically. What if there are already a couple bulleted lists on the page?

Like, this page now is pretty bullet happy: https://www.vets.gov/education/apply/

And there are sub-bullets now under "Other ways to apply"

ncksllvn:
@bethpotts I do think the bullets add something in terms of clarity but I see what you mean in that if they're used too much they clutter up the page.

I'll go ahead and revert this for now. I was under the impression the changes were pretty set-in-stone and just waiting to be implemented so I'm glad we got this discussion going.

bethpotts:
@ncksllvn Thanks for reverting for now! Amy and I will chat about it and I'll get back to you.

ncksllvn:
@bethpotts No problem! Sounds great! Just let me know and I'm sure the changes won't take long to make.

phiden:
So the original intent was to fix spacing issues on this particular page, as I understood it, not so much that it was meant to be a universal change. This goes to the bigger issue overall that we have spacing issues across the board! We should revisit.

PR #6415, opened 2017-09-15
Implement autosave

cehsu:
Not yet saved
screen shot 2017-09-15 at 9 05 55 am

Not logged in
screen shot 2017-09-15 at 9 06 16 am

Saving
screen shot 2017-09-15 at 9 08 27 am

Saved
screen shot 2017-09-15 at 9 12 09 am


src/js/common/schemaform/FormPage.jsx
```diff
@@ -112,13 +127,28 @@ class FormPage extends React.Component {
...
     const { route, params, form } = this.props;
+    const { profile } = this.props.user;
     let {
       schema,
       uiSchema
     } = form.pages[route.pageConfig.pageKey];
 
     let data = form.data;
-
+    let savedAt;
```
jbalboni: Might be cleaner to make a SaveStatus component and put this logic in it.

src/js/common/schemaform/FormPage.jsx
```diff
@@ -102,6 +104,18 @@ class FormPage extends React.Component {
     return { pages: expandedPageList, pageIndex };
   }
 
+  autoSave() {
+    const { route, params, form } = this.props;
+    let data = form.data;
+    const { formId, version } = form;
+    const returnUrl = this.props.locationPathname;
+
+    if (route.pageConfig.showPagePerItem) {
```
jbalboni: I don't think this is necessary. I think you can just save `form.data`.

src/js/common/schemaform/SaveStatus.jsx
```diff
@@ -0,0 +1,46 @@
...
+    const { form, profile } = this.props;
+
+    let savedAt;
+    let savedAtMessage;
+    const savedForm = profile && profile.savedForms.length > 0 && profile.savedForms
+      .filter(f => moment.unix(f.metadata.expires_at).isAfter())
+      .find(f => f.form === form.formId);
+    if (savedForm || form.lastSavedDate) {
+
+      savedAt = form.lastSavedDate
```
jbalboni: I think we set `lastSavedDate` whenever we've auto-saved? You also have an `autoSuccess` status. I think you could simplify this logic and just use those values.

src/js/common/schemaform/FormPage.jsx
```diff
@@ -158,7 +172,9 @@ class FormPage extends React.Component {
                 afterText="»"/>
             </div>
           </div>
-          {!form.disableSave && <div className="row">
+          {!form.disableSave && <SaveStatus
```
jbalboni: It feels a little weird to me to pass SaveFormLink as a child to SaveStatus. I think I would just pull the row div in here and put them next to each other.

jbalboni (commented):
Looks good so far, had some comments/questions.

src/js/common/schemaform/FormPage.jsx
cehsu: Thanks @jbalboni, removed.

src/js/common/schemaform/FormPage.jsx
cehsu: Good point, changes made.

src/js/common/schemaform/SaveStatus.jsx
cehsu: That's much cleaner, change made.

src/js/common/schemaform/FormPage.jsx
cehsu: Good point, change made.

src/js/common/schemaform/SchemaForm.jsx
```diff
@@ -88,6 +88,7 @@ class SchemaForm extends React.Component {
   }
 
   onBlur(id) {
+    this.props.autoSave();
```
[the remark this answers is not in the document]
cehsu: Good point, 1 second debounce added.

src/js/common/schemaform/FormPage.jsx
```diff
@@ -102,6 +104,15 @@ class FormPage extends React.Component {
     return { pages: expandedPageList, pageIndex };
   }
 
+  autoSave() {
+    const { form } = this.props;
+    const data = form.data;
+    const { formId, version } = form;
+    const returnUrl = this.props.locationPathname;
```
jbalboni: This should be `this.props.location.pathname`.

src/js/common/schemaform/SchemaForm.jsx
```diff
@@ -88,6 +88,7 @@ class SchemaForm extends React.Component {
   }
 
   onBlur(id) {
+    _.debounce(1000, this.props.autoSave);
```
jbalboni: This isn't doing anything right now, it returns a function you can call that will be debounced.

I think it's slightly better if instead of passing an `autoSave` prop we pass an `onBlur` one, and do the debouncing/saving all in FormPage, which keeps it all in one place.

So in here you would have something like `this.props.onBlur(id);`.

And in `FormPage` you can put `this.autoSaveDebounced = _.debounce(1000, this.autoSave)` in the constructor, and pass `onBlur={this.autoSaveDebounced}` in the props passed to `SchemaForm`.

src/js/common/schemaform/FormPage.jsx
```diff
@@ -102,6 +104,15 @@ class FormPage extends React.Component {
...
+
+    this.props.saveInProgressForm(formId, version, returnUrl, data, true);
```
jbalboni: I think we want to avoid auto-saving if they're not logged in, so we don't show them that red box the whole time they're filling out the form if they're not going to log in.

src/js/common/schemaform/FormPage.jsx
cehsu: Nice catch, changed.

src/js/common/schemaform/FormPage.jsx
cehsu: Good idea, `SaveStatus` is now rendered if `this.props.isLoggedIn` is true.

src/js/common/schemaform/FormPage.jsx
```diff
@@ -111,14 +125,12 @@ class FormPage extends React.Component {
   }
 
   render() {
-    const { route, params, form } = this.props;
+    const { route, params, form, isLoggedIn } = this.props;
```
jbalboni: Is `isLoggedIn` ever true? I can't find where it's passed into this component. You might need to get it from the `user` prop.

cehsu: You're totally right. Thanks, this has been changed.

jbalboni:
@cehsu It looks like everything is working except for what happens when your session expires. I simulate this by going into the console and doing something like `sessionStorage.userToken = 'asdfasdf'`, then triggering a blur. When you do that, you get the green box and the red box with the sign in link stacked below the buttons. And when you click the sign in link and sign in, you are sent to the saved form page, which isn't ideal, since the user didn't ask to save and finish later.

cehsu:
Thanks @jbalboni, those are both good points. I've made a change so that conflicting messages won't appear. However, I wonder if they click "sign in" on: "Please sign in again to save your application." they would expect for their form to be saved?

jbalboni:
@cehsu Yeah, I suppose that's true. I think it's more that, if your session expires, we don't give you a way to sign in again without going to the form saved page. I think we could leave to a separate item, though, since like you pointed out, the message itself is pretty accurate.

jbalboni (approved):

cehsu:
Another good point, an issue for providing users with a way to sign in again that doesn't navigate to form saved page has been created [here](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4932).

PR #6417, opened 2017-09-15
Kudos identity - new sign in flows

bshyong:
- New sign in page
- New verify identity page
- Changes to navbar signin/signup link
- Changes in Required Login View to support new signin and verify flows
- Support for multiple auth urls from API
- Multifactor opt-in on user profiles
- Changes to E2E login helpers

EDIT: Review instance here http://f083c2d09c619d6af6e4d8b606d9ec96.review.vetsgov-internal/


src/js/login/components/SearchHelpSignIn.jsx
```diff
@@ -30,7 +32,7 @@ class SearchHelpSignIn extends React.Component {
       content = null;
     } else {
       content = (<div>
-        <a href="#" onClick={this.props.onUserLogin}>Sign In</a><span className="signin-spacer">|</span><a href="#" onClick={this.props.onUserSignup}>Sign up</a>
+        <a href={signInUrl}>Sign In</a><span className="signin-spacer">|</span><a href={signInUrl}>Sign up</a>
```
U-DON: Probably lowercase the "In" to make it "Sign in | Sign up" for consistency.

src/js/common/components/RequiredLoginView.jsx
```diff
@@ -32,15 +30,15 @@ class RequiredLoginView extends React.Component {
       return <LoadingIndicator setFocus message="Loading your information..."/>;
     }
 
-    const loginComponent = <LoginPrompt loginUrl={this.props.loginUrl}/>;
-    const verifyComponent = <VerifyPrompt verifyUrl={this.props.verifyUrl}/>;
+    const currentPath = window.location.pathname;
```
U-DON: Could leverage `appendQuery` here. Since the point is to get the path for the next query, could be something like:

```
const nextQuery = { next: window.location.pathname };
const signInUrl = appendQuery('signin', nextQuery);
const verifyUrl = appendQuery('verify', nextQuery);
```

src/js/signin/components/Signin.jsx
```diff
@@ -0,0 +1,151 @@
...
+
+  checkLoggedInStatus(loggedIn) {
+    if (this.props.currentlyLoggedIn || loggedIn) {
+      const nextParams = new URLSearchParams(window.location.search);
+      const nextPath = nextParams.get('next');
+
+      if (nextPath) {
+        return window.location.replace(nextPath);
+      }
+      return window.location.replace('/');
```
U-DON:
```
return window.location.replace(nextPath || '/');
```

src/js/signin/verify-entry.jsx
```diff
@@ -0,0 +1,29 @@
...
+      <Provider store={store}>
+        <Main verify/>
+      </Provider>
+    ), document.getElementById(rootElementId));
+  }
+
+  initReact(init);
+}
+
+createLoginWidget(commonStore, 'react-root');
```
U-DON: Could just do the same as `signin-entry` and do `document.getElementById('react-root')` instead of using `createLoginWidget`.

U-DON (approved):
LGTM. Left some minor comments to keep track of things to refactor after everything is merged.

src/js/signin/containers/Main.jsx
```diff
@@ -0,0 +1,173 @@
...
+import { getUserData, addEvent, handleLogin, getLoginUrls } from '../../common/helpers/login-helpers';
+
+import { updateLoggedInStatus, updateLogInUrl, updateVerifyUrl, updateLogoutUrl, updateLogInUrls } from '../../login/actions';
+import Signin from '../components/Signin';
+import Verify from '../components/Verify';
+
+class Main extends React.Component {
+  constructor(props) {
+    super(props);
+    this.setMyToken = this.setMyToken.bind(this);
```
jbalboni: On Rainbows we've mostly switched to writing class methods like this, to avoid these bind calls:

```js
setMyToken = (event) => {
...
}
```

src/js/signin/containers/Main.jsx
```diff
@@ -0,0 +1,173 @@
...
+  const userState = state.user;
+  return {
+    login: userState.login,
+    profile: userState.profile
+  };
+};
+
+
+const mapDispatchToProps = (dispatch) => {
+  return {
```
jbalboni: Any reason to not use the object syntax for mapDispatchToProps?

src/js/signin/verify-entry.jsx
```diff
@@ -0,0 +1,29 @@
...
+require('../common');  // Bring in the common javascript.
+require('../../sass/login.scss');
+
+const commonStore = createCommonStore(reducer);
+
+export default function createLoginWidget(store, rootElementId = 'login-root') {
+  function init() {
+    ReactDOM.render((
+      <Provider store={store}>
+        <Main verify/>
```
jbalboni: Do we need to have separate signin and verify bundles? It seems like they could be the same bundle, but set this verify param (and skip createLoginWidget) based on the url.

It might be slightly bigger, but then it would be cached for the second request on the verify step.

src/sass/user-profile.scss
```diff
@@ -12,6 +12,29 @@
...
+      padding-right: 0;
+      text-align: right;
+
+      button {
+        margin-right: 0;
+      }
+    }
+  }
+
+  .usa-alert {
```
jbalboni: Can we make a class for these styles? If someone else adds an alert in the profile page, they might have to override these.

src/sass/login.scss
```diff
@@ -1,5 +1,202 @@
...
+    text-align: right;
+  }
+
+  &:after {
+    margin-left: 2%;
+  }
+}
+
+.verify, .login {
+  button {
```
jbalboni: I've been pushing people on Rainbows to avoid writing styles like this, with generic elements and nesting that's not limited by direct descendent selectors. I think it leads to styles that are hard to understand as they grow and are modified (the claim status styles are good examples of this). And they make styles harder to overwrite, since the nested styles have two or more selectors, which increases specificity. That matters less here, since these pages are small and pretty focused, but it does make a difference for pages with a broader set of features (like the user profile) or in common styles.

The last sass/less codebase I worked on ended up with some selectors that were 7+ classes deep trying to overwrite nested styles that had built up for several years.

jbalboni (approved):
LGTM, had some questions/comments.

src/js/common/components/RequiredLoginView.jsx
```diff
@@ -32,15 +30,15 @@ class RequiredLoginView extends React.Component {
...
 
     if (this.props.authRequired === 1) {
       if (this.props.userProfile.accountType >= 1) {
         return this.props.children;
       }
 
-      return loginComponent;
+      return window.location.replace(`/signin?next=${currentPath}`);
```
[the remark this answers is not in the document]
U-DON: This used to return the gated login component, but now redirects to a route that loads the SignIn component. So in this case it doesn't really need to return an actual value, I don't think.

src/js/signin/containers/Main.jsx
U-DON: I'd like to see that too. Sounds like a potential future sweep that can be done across all products. Might be good to bring up at the next FE dev meeting.

src/js/signin/containers/Main.jsx
U-DON: A lot of this logic is copied over from the `login` app, which heavily used `mapDispatchToProps` in this style. We're going to be refactoring the code after the initial launch, and the object syntax for `mapDispatchToProps` is on the plate.

src/sass/login.scss
U-DON: What's the style you're advocating for? I do think we should be using a lot more of the `>` selector.

jbalboni: I have used BEM in the past, but I don't think we need to go that far. I think we should shoot for using single class selectors as much as possible, > selectors when necessary, and almost never style generic elements (aside from site wide widget styles). Some nesting is unavoidable, but I think we should try to minimize it.

src/js/signin/containers/Main.jsx
bshyong: Agree. These issues are related to code style but more of a thing we could establish across the entire team instead of having variance in different apps

raquelromano-gov:
@bshyong @saneshark  Is there a ticket open to test all of the new components and functionality added in this PR? Or is this the piece that's getting rolled back after a weekend of testing?

bshyong:
@raquelromano-gov yes we plan to test extensively this weekend. The functionality has been tested throughout development by us with ID.me but having it in staging will let us test more thoroughly. It will be rolled back before going to production on Monday

PR #6419, opened 2017-09-15
Identity cleanup

bshyong:
- Fixes a bug with identity requests
- Cleans up old signin code
- Connected to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4856, maintains use of IDme signin URL for in progress form saving

@patrickvinograd please make sure this gets merged to `kudos-identity` before that branch is deployed to staging

### DIFF
--- a/src/js/common/helpers/login-helpers.js
+++ b/src/js/common/helpers/login-helpers.js
@@ -1,5 +1,5 @@
 import environment from './environment.js';
-import { updateLoggedInStatus } from '../../login/actions';
+import { updateLoggedInStatus, updateLogInUrls } from '../../login/actions';
 import { updateProfileFields, profileLoadingFinished } from '../../user-profile/actions';
 
 export function handleVerify(verifyUrl) {
@@ -99,13 +99,13 @@ export function getLoginUrls(onUpdateLoginUrls) {
   return loginUrlsRequest;
 }
 
-export function handleLogin(loginUrl) {
+export function handleLogin(loginUrl, onUpdateLoginUrls) {
   window.dataLayer.push({ event: 'login-link-clicked' });
   if (loginUrl) {
     window.dataLayer.push({ event: 'login-link-opened' });
     const receiver = window.open(`${loginUrl}&op=signin`, '_blank', 'resizable=yes,scrollbars=1,top=50,left=500,width=500,height=750');
     receiver.focus();
-    return null;
+    return getLoginUrls(onUpdateLoginUrls || updateLogInUrls);
   }
   return Promise.reject('Could not log in; loginUrl not provided.');
 }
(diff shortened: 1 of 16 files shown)


src/js/signin/containers/Main.jsx
```diff
@@ -45,11 +49,14 @@ class Main extends React.Component {
     this.loginUrlRequest = getLoginUrls(this.props.onUpdateLoginUrls);
   }
 
-  getVerifyUrl() {
-    if (!this.props.login.currentlyLoggedIn) return;
+  getVerifyUrl(forceRequest) {
+    if (!forceRequest && !this.props.login.currentlyLoggedIn) return;
```
U-DON: Why was this necessary?

U-DON (approved):
LGTM. Might want @jbalboni to verify the SiP stuff.

src/js/common/helpers/login-helpers.js
```diff
@@ -111,13 +99,13 @@ export function getLoginUrls(onUpdateLoginUrls) {
...
 
-export function handleLogin(loginUrl, onUpdateLoginUrl) {
+export function handleLogin(loginUrl) {
   window.dataLayer.push({ event: 'login-link-clicked' });
   if (loginUrl) {
     window.dataLayer.push({ event: 'login-link-opened' });
     const receiver = window.open(`${loginUrl}&op=signin`, '_blank', 'resizable=yes,scrollbars=1,top=50,left=500,width=500,height=750');
     receiver.focus();
-    return getLoginUrl(onUpdateLoginUrl);
+    return null;
```
U-DON: Should this not be `return getLoginUrls(onUpdateLoginUrls)`? This is a change that's causing some unit test failures in `LoginModal`.

bshyong: What would be the purpose of calling getLoginUrls again? I can look at the tests but getting rid of this second call didn't break any of the signin flows. 

We already have a URL and open the modal with that URL in this method so I'm not sure what the purpose of the second call would be

src/js/login/reducers/login.js
```diff
@@ -39,9 +37,7 @@ function loginStuff(state = initialState, action) {
     case UPDATE_LOGGEDIN_STATUS:
       return _.set('currentlyLoggedIn', action.value, state);
 
-    case UPDATE_LOGIN_URL:
-      return _.set('loginUrl', appendQuery(action.value, { clientId: action.gaClientId }), state);
-
+    // being explicit here to avoid confusion
```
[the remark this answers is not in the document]
bshyong: The duplicated appendQuery code in the reducer for each type of URl

src/js/signin/containers/Main.jsx
bshyong: This is so that the method can be properly called in willReceiveProps if the loggedIn prop changes due to someone logging in, and if somebody lands on the verify page and does a page refresh

PR #6420, opened 2017-09-16
Disallow address update

markgreenburg:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4113

This adds conditional logic to edit address button so that it only shows for vets that can update their address per the vets-api.

Still needs tests.

Also, not sure I'm happy with how complex this component has gotten. Would like to refactor at some point and break out into smaller components that just do One Thing.

### DIFF
--- a/test/letters/containers/AddressSection.unit.spec.jsx
+++ b/test/letters/containers/AddressSection.unit.spec.jsx
@@ -18,7 +18,8 @@ const defaultProps = {
     state: 'VA',
     zipCode: '12345'
   },
-  saveAddress: saveSpy
+  canUpdateAddress: true,
+  saveAddress: saveSpy,
 };
 
 describe('<AddressSection>', () => {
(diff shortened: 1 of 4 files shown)


test/letters/containers/AddressSection.unit.spec.jsx
```diff
@@ -45,6 +46,18 @@ describe('<AddressSection>', () => {
     expect(tree.subTree('.step-content').text()).to.contain('2476 main street, ste #12 west');
   });
 
+  it('should not render an edit button if user not allowed to edit address', () => {
+    const cannotEditProps = { ...defaultProps, canUpdateAddress: false };
+    const component = ReactTestUtils.renderIntoDocument(<AddressSection {...cannotEditProps}/>);
+    expect(() => ReactTestUtils.findRenderedDOMComponentWithTag(component, 'button')).to.throw();
```
cvalarida: Does this make sense as an arrow function? Can't we just
```js
expect(ReactTestUtils.findRenderedDOMComponentWithTag(component, 'button')).to.throw();
```
?

Alternatively, what we've done before is:
```js
import { findDOMNode } from 'react-dom';
import { ReactTestUtils } from 'react-dom/test-utils';
...
const component = ReactTestUtils.renderIntoDocument(<AddressSection {...cannotEditProps}/>);
const tree = findDOMNode(component);
expect(tree.querySelectorAll('button.usa-button-outline')).to.be.empty; // or .to.have.lengthOf(0)
```

What you've got is probably a bit faster since we don't have to use `findDOMNode`, but the above I think is clearer that we want to make sure that there are no such buttons. It's also maybe more robust in case `findRenderedDOMComponentWithTag` throws an error for some reason other than not being able to find the element. I have no idea if that's even possible, but paranoid programming. :man_shrugging:

cvalarida (approved):
Does this still need tests like the original comment suggests? I think the ones you've got are good, though I did have a thought on one.

test/letters/containers/AddressSection.unit.spec.jsx
markgreenburg: As far as I know, the function that's throwing actually has to be wrapped within the `expect()` so that the error's handled / propagated properly. If you run without wrapping, the test itself actually throws an error. If you wrap it like I have but return something other than an error out of the inner function (e.g. `expect(() => true).to.throw()`, the test fails properly, notifying you that it expected to fail but didn't 🎊 

The alternative approach makes sense as well...happy to swap if you think it'd work better!

PR #6421, opened 2017-09-18
Fix infinite verify URL loop and broken verify link in user profile.

U-DON:
### Infinite `identity_proof` fetches

Going to the verify page when logged in would infinitely fetch the identity proof URL. When the user was `currentlyLoggedIn` in `getVerifyUrl` in `Verify.jsx`, it would never return before the fetch. And each time it finished fetching, it would update the store and trigger `componentWillReceiveProps`.

I changed it to stop once it set the `verifyUrl`. I also changed it to use the `componentDidUpdate` lifecycle method instead so that it would deal with the current props if it had to call `getVerifyUrl`. With `componentWillReceiveProps`, the comparison for `forceRequest` was between the current props and what they would change to, but then `getVerifyUrl` would look at the current props, not what they were changing to, which is why you had to pass in the force flag. It felt a little confusing for me to think about.

### Broken verify URL in user profile

The "You need to verify your account" link wasn't working because `this.props.login.verifyUrl` wasn't populated, so I changed it to make the container fetch the URL if it wasn't available. Since these two instances did the same thing, I pulled the promise chain into its own helper function like the other identity URL helpers.

### DIFF
--- a/src/js/common/helpers/login-helpers.js
+++ b/src/js/common/helpers/login-helpers.js
@@ -20,6 +20,21 @@ export function handleMultifactor(multifactorUrl) {
   }
 }
 
+export function getVerifyUrl(onUpdateVerifyUrl) {
+  const verifyUrlRequest = fetch(`${environment.API_URL}/v0/sessions/identity_proof`, {
+    method: 'GET',
+    headers: new Headers({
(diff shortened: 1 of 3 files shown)


bshyong (approved):
LGTM! Also matches the same pattern as the other URL requests. Thanks for the fix

PR #6422, opened 2017-09-18
Switch mobile navigation menu to use a calculated height

ncksllvn:
Currently, the height of the navigation menu's container (on mobile) is set at `100%` and the actual contents at `62rem`. This causes some weird behavior, because the container is then offscreen by the distance from the navigation to the top of the window. This causes the content to get cut off in some instances, specifically if you have the "Explore benefits" menu expanded.

To fix this, these changes include using the CSS `calc()` function to set the container's height to `100%` of the screen minus the distance from the navigation bar to the top of the window. That second value I determined to be `111.16px` by inspecting each element - there wasn't really a clear way to do this otherwise without involving JavaScript. It then applies that value to the `top` value to offset it from the top of the menu and allows the navigation menu's contents to increase the height according to the length of the content.

Connects to department-of-veterans-affairs/vets.gov-team/issues/4619.

### DIFF
--- a/src/sass/modules/_m-vet-nav.scss
+++ b/src/sass/modules/_m-vet-nav.scss
@@ -1,3 +1,6 @@
+// The distance from the top of the sceen that the mobile navigation menu should appear
+$height-offset: 111.16px;
+
 body.va-pos-fixed {
   width: 100%;
   height: 100%;
@@ -5,7 +8,8 @@ body.va-pos-fixed {
 
 #vetnav {
   background-color: $color-primary-darkest;
-  height: 100%;
+  height: calc(100% - #{$height-offset});
+  top: $height-offset;
   left: 0;
   overflow-y: scroll;
   position: absolute;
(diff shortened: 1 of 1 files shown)


src/sass/modules/_m-vet-nav.scss
```diff
@@ -1,11 +1,15 @@
...
 body.va-pos-fixed {
   width: 100%;
   height: 100%;
 }
 
 #vetnav {
   background-color: $color-primary-darkest;
-  height: 100%;
+  height: calc(100% - #{$height-offset});
+  top: $height-offset;
```
U-DON: Is this necessary? From a quick test on the Heroku instance, removing `top` didn't seem to affect the menu's layout.

U-DON (approved):
LGTM. 👍

src/sass/modules/_m-vet-nav.scss
ncksllvn: Now that you mention it I guess it's not necessary since the `$height-offset` is so close to being exactly equal to the default location the navigation menu appears. I kinda think it makes sense to be there for clarity though. What do you think?

U-DON: Doesn't hurt to have it, I suppose. It can be clearer to specify both a `top`/`bottom` and `left`/`right`.

PR #6425, opened 2017-09-18
Enable SiP on 5490 form

jbalboni:
screen shot 2017-09-18 at 4 08 57 pm

### DIFF
--- a/test/user-profile/helpers.unit.spec.js
+++ b/test/user-profile/helpers.unit.spec.js
@@ -77,7 +77,7 @@ describe('profile helpers:', () => {
       expect(isSIPEnabledForm('missingInfoForm')).to.be.false;
     });
     it('should throw an error if a form is not included the list of sipEnabledForms', () => {
-      expect(() => isSIPEnabledForm({ form: '22-5490' })).to.throw('Could not find form');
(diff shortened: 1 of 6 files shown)


test/user-profile/helpers.unit.spec.js
```diff
@@ -77,7 +77,7 @@ describe('profile helpers:', () => {
       expect(isSIPEnabledForm('missingInfoForm')).to.be.false;
     });
     it('should throw an error if a form is not included the list of sipEnabledForms', () => {
-      expect(() => isSIPEnabledForm({ form: '22-5490' })).to.throw('Could not find form');
+      expect(() => isSIPEnabledForm({ form: '40-10007' })).to.throw('Could not find form');
```
cvalarida: Suppose we should probably just make a random form id. Is that what this does?

cvalarida (approved):

test/user-profile/helpers.unit.spec.js
jbalboni: The test above is for random ids, this one is for forms that are in our list of forms, but don't have SiP on yet. I think this test will go away once pre-needs adds SiP.

PR #6427, opened 2017-09-19
Add dependent messaging

markgreenburg:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4592

Replaces the old 404 [post-9/11 GIBS message](https://github.com/department-of-veterans-affairs/vets.gov-team/blob/master/Products/EVSS%20Integration/Letters%20and%20GIBS%20error%20messages%20mapping.md) with a new one:

screen shot 2017-09-18 at 8 34 18 pm

screen shot 2017-09-18 at 8 40 15 pm

### DIFF
--- a/src/js/post-911-gib-status/utils/helpers.jsx
+++ b/src/js/post-911-gib-status/utils/helpers.jsx
@@ -105,7 +105,7 @@ export function notQualifiedWarning() {
 
 export function noChapter33BenefitsWarning() {
   return (
-    <div className="record-not-found">
+    <div id="noChapter33Benefits">
       <header>
         <h1>We couldn’t find your Post-9/11 GI Bill information.</h1>
       </header>
--- a/src/js/post-911-gib-status/utils/helpers.jsx
+++ b/src/js/post-911-gib-status/utils/helpers.jsx
@@ -120,7 +120,8 @@ export function noChapter33BenefitsWarning() {
       <h2>Need GI Bill help?</h2>
       <hr className="divider"/>
       <p>If you have questions or need help looking up your GI Bill information, please call <a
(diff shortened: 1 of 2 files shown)


raquelromano-gov:
removed @akainic and added @annekainicUSDS  (I also made this mistake when I first joined--Anne, maybe you should remove the other handle from the org since people probably tag you there and you may not see things)

markgreenburg:
Ahh thanks for catching that, Raquel.

src/js/post-911-gib-status/utils/helpers.jsx
```diff
@@ -104,25 +104,24 @@ export function notQualifiedWarning() {
...
+      <div className="usa-alert usa-alert-warning">
+        <div className="usa-alert-body">
+          <p className="usa-alert-heading">
+            <a target="_blank" href="https://www.ebenefits.va.gov/ebenefits/about/feature?feature=post-911-gi-bill-enrollment-status">
+            If you’re a dependent, please go to eBenefits to look up your GI Bill information.</a>
+          </p>
+        </div>
+      </div>
+      <h2>Need GI Bill help?</h2>
+      <hr className="divider"/>
```
annekainicUSDS: I wonder if you could use the classes that this page for health care uses: http://localhost:3001/health-care/apply/application/introduction. The classes are in the schemaform scss, so may need to be abstracted out of that and into a common file. Not necessary for this PR, just noticing common styles across pages and wondering if we should think about consolidating them.

annekainicUSDS (approved):
LGTM! Just one question about the css

src/js/post-911-gib-status/utils/helpers.jsx
markgreenburg: Thanks Anne...I need to dig around a bit and see what we've already got built.

PR #6428, opened 2017-09-19
Increase z index so it overrides nav

annekainicUSDS:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4907

The z-index of one of the nav elements was recently changed to 1011, which was causing it to show up on the print page. This PR increase the z-index of the print page to 2000 to override that.

Before:
screen shot 2017-09-19 at 12 34 58 pm

After:
screen shot 2017-09-19 at 12 34 30 pm

### DIFF
--- a/src/sass/post-911-gib-status.scss
+++ b/src/sass/post-911-gib-status.scss
@@ -63,7 +63,7 @@
     width: 100%;
     height: 100%;
     overflow-y: scroll;
-    z-index: 10;
+    z-index: 2000;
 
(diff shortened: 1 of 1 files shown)


cvalarida (approved):
Awesome. This works for now, but I'm wondering if there's some fancy footwork we can do to make the nav menu `display:none` on this page...Probably we'd have to have some javascript on this page that adds a class to the nav menu.

PR #6437, opened 2017-09-20
Address update fail

markgreenburg:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4109

@raquelromano-gov @annekainicUSDS @cvalarida Would appreciate some early eyes / thoughts on this...fairly significant refactor of the AddressSection container driven by one too many nested conditionals 😛 .
This would be to enable error handling for saving addresses the way it's designed to look in the mock...surprisingly, doing so without changing the structure of the `AddressSection` container is kinda tricky.


markgreenburg:
One other thought I have is that down the road it may make sense to extract the `usa-alert-error`s (yellow alert boxes) into a common component that just takes a header & text as props (I think I've made...3 of these now?).

src/js/letters/containers/AddressSection.jsx
```diff
@@ -8,26 +8,63 @@ import {
...
+    this.state = {
+      hasLoadedAddress: false,
+      isEditingAddress: false,
+      editableAddress: {},
+    };
   }
 
-  handleUpdate = () => {
+  componentWillReceiveProps(nextProps) {
+    if (!this.state.hasLoadedAddress && Object.keys(nextProps.savedAddress).length > 0) {
```
cvalarida: We should have put a comment in here when we knew why this was necessary. Can you do that?

If I remember, it was something like...`savedAddress` is coming from an api call that may or may not get returned before this component mounts, so we had to update the editable address when we get it (but only the first time).

src/js/letters/containers/AddressSection.jsx
```diff
@@ -8,26 +8,63 @@ import {
...
+import AddressContent from '../components/AddressContent';
 
 export class AddressSection extends React.Component {
   constructor() {
     super();
-    this.state = { isEditingAddress: false };
+    this.state = {
+      hasLoadedAddress: false,
+      isEditingAddress: false,
+      editableAddress: {},
```
cvalarida: Given that the `savedAddress` _may_ come in before this component mounts, can we do something like this:

```js
constructor(props) {
  super(props);
  this.state = {
    hasLoadedAddress: false,
    isEditingAddress: false,
    editableAddress: props.savedAddress || {}
  }

  // If the request has already come in, make sure we know that we've loaded it already
  if (Object.keys(this.state.editableAddress) !== 0) {
    this.state.hasLoadedAddress = true;
  }
}
```

I don't know that it ever _will_ come in before this component mounts, but I just like to play it safe when it comes to asynchronous whatnots. This is maybe overkill. :man_shrugging:

src/js/letters/containers/AddressSection.jsx
```diff
@@ -8,26 +8,63 @@ import {
...
+  }
+
+  handleChange = (path, update) => {
+    this.setState(({ editableAddress }) => {
+      // reset state code when user changes address country but don't add or
+      // modify state property otherwise
+      return (path === 'country'
+        ? {
+          editableAddress: {
+            ...editableAddress,
```
cvalarida: If I remember right, the object spread operator is pretty slow and this is going to be run (along with the validations for potentially all the fields) on every keystroke, so I'd like to optimize it as best we can. I think we can use `Object.assign()` to do the same thing.

src/js/letters/containers/AddressSection.jsx
```diff
@@ -8,26 +8,63 @@ import {
...
+        }
+      );
+    });
   }
 
   render() {
-    const address = this.props.address || {};
+    // We want to use the Redux address as source of truth. Address in
+    // container state is only used to control the form input for the
+    // <Address/> component
```
cvalarida: For clarification, you mean like if the save fails, we want to use / display the original address, so we can't override what's in the redux store.

Can we update this comment to include that? It makes perfect sense to me now, but I don't know that it would in six months is all.

src/js/letters/containers/AddressSection.jsx
```diff
@@ -95,33 +129,26 @@ export class AddressSection extends React.Component {
...
+        </AddressContent>
       );
     }
 
-    return (
-      <div>
-        {addressContent}
-      </div>
-    );
+    return (addressContent);
```
cvalarida: Probably don't need these `()`s.

src/js/letters/components/InvalidAddress.jsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+
+const InvalidAddress = () => (
+  <div className="step-content">
+    <div id="invalidAddress">
+      <div className="usa-alert usa-alert-error">
+        <div className="usa-alert-body">
+          <h2 className="usa-alert-heading">Address unavailable</h2>
+          <p className="usa-alert-text">
+            We’re encountering an error with your address information. This is not required for your letters, but if you’d like to see the address we have on file, or to update it, please visit <a href="/" target="_blank">this link</a>.
```
cvalarida: This isn't a new change, I know, but I don't like having a dangling link that goes nowhere helpful.

Do we have a ticket to track this? I feel like I wrote one up, but I can't seem to find it...

test/letters/components/AddressBlock.unit.spec.jsx
```diff
@@ -0,0 +1,14 @@
...
+import { expect } from 'chai';
+import AddressBlock from '../../../src/js/letters/components/AddressBlock';
+
+const defaultProps = { name: 'Gary Todd' };
+
+describe('<AddressBlock/>', () => {
+  it('should render', () => {
+    const tree = SkinDeep.shallowRender(<AddressBlock { ...defaultProps }/>);
+    const vdom = tree.getRenderOutput();
+    expect(vdom).to.exist;
```
cvalarida: Mmm...Can we check to make sure it's just got some text we expect it to have? That way we know for sure it actually rendered rather than there being DOM.

Maybe it's trivial, but I think we've always checked against some element we knew it should render to avoid it failing to render but passing the test.

test/letters/components/AddressContent.unit.spec.jsx
```diff
@@ -0,0 +1,38 @@
...
+  saveError: false,
+  addressObject: addressLines,
+  name: 'Gary Todd',
+};
+
+describe('<AddressContent/>', () => {
+  it('should render', () => {
+    const tree = SkinDeep.shallowRender(<AddressContent { ...defaultProps }/>);
+    const vdom = tree.getRenderOutput();
+    expect(vdom).to.exist;
```
cvalarida: Same as the comment above with all these `.to.exist` checks.

test/letters/components/InvalidAddress.unit.spec.jsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import SkinDeep from 'skin-deep';
+import { expect } from 'chai';
+import InvalidAddress from '../../../src/js/letters/components/InvalidAddress';
+
+describe('InvalidAddress', () => {
+  it('should render', () => {
+    const tree = SkinDeep.shallowRender(<InvalidAddress/>);
+    const vdom = tree.getRenderOutput();
+    expect(vdom).to.exist;
```
cvalarida: But wait, there's more!

test/letters/containers/AddressSection.unit.spec.jsx
```diff
@@ -28,37 +27,62 @@ const defaultProps = {
...
-    const tree = SkinDeep.shallowRender(<AddressSection {...defaultProps}/>);
+    const tree = SkinDeep.shallowRender(<AddressSection { ...defaultProps }/>);
     const vdom = tree.getRenderOutput();
     expect(vdom).to.exist;
   });
 
+  it('should display an address if one is provided in props', () => {
+    const tree = SkinDeep.shallowRender(<AddressSection { ...defaultProps }/>);
+    const addressContent = tree.dive(['AddressContent']);
+    expect(addressContent).to.exist;
```
cvalarida: Sorry to be such a pain... 😬

test/letters/containers/AddressSection.unit.spec.jsx
```diff
@@ -28,37 +27,62 @@ const defaultProps = {
 
 describe('<AddressSection>', () => {
   it('should render', () => {
-    const tree = SkinDeep.shallowRender(<AddressSection {...defaultProps}/>);
+    const tree = SkinDeep.shallowRender(<AddressSection { ...defaultProps }/>);
```
cvalarida: Oh weird...And the linter doesn't care about this? Previously, we always had to make sure we had no spaces between our `{}`s inside jxs tags. Huh. :man_shrugging:

test/letters/containers/AddressSection.unit.spec.jsx
```diff
@@ -28,37 +27,62 @@ const defaultProps = {
...
+    const invalidAddress = tree.dive(['InvalidAddress']);
+    expect(invalidAddress).to.exist;
+  });
+
   it('should format 1 address line', () => {
     const tree = SkinDeep.shallowRender(<AddressSection {...defaultProps}/>);
-    expect(tree.subTree('.step-content').text()).to.contain('2476 main street');
+    const addressBlock = tree.dive(['AddressContent', 'AddressBlock']);
+    const addressBlockText = addressBlock.subTree('.address-block', 'div.letters-address').text();
+    // NOTE: have to pass in uncapitalized addresses to assert against because shallowRender apparently strips caps...
```
cvalarida: :man_facepalming:

cvalarida (commented):
It's shaping up! Looks good. I had a few comments...Okay, so I sometimes get carried away with my comments. Sorry about that. 😬  I won't let it be said I wasn't thorough!

src/js/letters/containers/AddressSection.jsx
markgreenburg: Yep, I like this! We'll provide better UX at least some of the time

src/js/letters/containers/AddressSection.jsx
markgreenburg: Yep...not available at time of mounting in most cases, so the user would get a blank form instead of one pre-populated with their existing info. Will add comment

annekainicUSDS (commented):
Looking good so far! So if I understand the main changes in this PR, they are:

1. Remove `updateAddress` action, as we will now only update the Redux store with the `saveAddress` action, which is triggered when the user hits the "Update" button in the address form. Save incremental changes to the address in local state instead.
2. Extract out parts of the address content/functionality into separate components.

New component hierarchy is:
```
AddressSection
  AddressContent
    UpdateFailureAlert OR AddressBlock
      Address
```
Missing any other major changes here?

src/js/letters/containers/AddressSection.jsx
markgreenburg: Actually, React seems to not like relying on an undefined prop in the constructor. I could put in some more logic to guard against this but not sure it's worth it.

src/js/letters/containers/AddressSection.jsx
markgreenburg: 👍

src/js/letters/containers/AddressSection.jsx
markgreenburg: 👍

src/js/letters/components/InvalidAddress.jsx
markgreenburg: Hmmm. @rroueche ? Do you know if we're tracking empty links somewhere?

markgreenburg: We should probably make sure we have these all covered across the app...

markgreenburg: Opened new issue to track this: https://github.com/department-of-veterans-affairs/vets.gov-team/issues/5004

test/letters/components/AddressBlock.unit.spec.jsx
markgreenburg: Yeah, I saw this convention being used elsewhere but it shouldn't be used as the sole test for a component. Fixing.

test/letters/components/AddressContent.unit.spec.jsx
markgreenburg: 👍

test/letters/components/InvalidAddress.unit.spec.jsx
markgreenburg: Don't wait, act now!

test/letters/containers/AddressSection.unit.spec.jsx
markgreenburg: My thinking with these is that we want to make sure each child component is getting rendered, and should be testing _what_ gets rendered in the actual unit tests for that subcomponent. Seem like you don't agree 😛 (which is completely fine), but just wanted to get clarification on your thinking.

src/js/letters/containers/AddressSection.jsx
markgreenburg: Doh, that's what happens when you don't pass props to the constructor...fixed

src/js/letters/components/Address.jsx
```diff
@@ -116,31 +104,32 @@ class Address extends React.Component {
...
-      selectedCountry = this.props.value.country;
+      selectedCountry = this.props.address.country;
     }
 
+
     const adjustedStateNames = this.adjustStateNames(STATE_CODE_TO_NAME, militaryStateNames, cityField);
 
     const stateProvince = selectedCountry === 'USA'
-      ? (<ErrorableSelect errorMessage={this.isValidAddressField(this.props.value[stateField]) ? undefined : 'Please enter a valid state/province'}
+      ? (<ErrorableSelect errorMessage={this.isValidAddressField(this.props.address[stateField]) ? undefined : 'Please enter a valid state/province'}
```
annekainicUSDS: The `stateField` and `cityField` stuff has changed in @cvalarida latest PR he merged, so this will change once you rebase/pull from master next.

annekainicUSDS (commented):
Looks good! Wondering if you saw my earlier comments as well?

src/js/letters/actions/letters.js
```diff
@@ -75,18 +75,20 @@ export function getMailingAddress() {
...
           address.state = address.militaryStateCode;
           delete address.militaryPostOfficeTypeCode;
           delete address.militaryStateCode;
         }
+        responseCopy.data.attributes.address = address;
 
         dispatch({
           type: GET_ADDRESS_SUCCESS,
-          data: address,
+          data: responseCopy
```
cvalarida: Thanks again for fixing this!

src/js/letters/containers/AddressSection.jsx
cvalarida: Maybe? Depends on how it's done. We'd have to hold off on constructing `<AddressSection/>` until the fetches are finished for it to fix this.

For now, though, I think this is pretty efficient and effective...if also a bit smelly. :man_shrugging:

src/js/letters/containers/AddressSection.jsx
```diff
@@ -39,31 +51,28 @@ export class AddressSection extends React.Component {
...
-          },
-        }
-        : {
-          editableAddress: {
-            ...editableAddress,
-            [path]: update,
-          },
+          [path]: update,
+          state: '',
+          militaryStateCode: '',
```
cvalarida: I don't want to block this merge, so I'm tracking this [here](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/5015): We don't need this `militaryStateCode`.

src/js/letters/containers/AddressSection.jsx
```diff
@@ -79,12 +88,12 @@ export class AddressSection extends React.Component {
...
-      const state = getStateName(address.stateCode);
+      const state = getStateName(address.state);
       cityStatePostal = `${city}, ${state} ${zipCode}`;
     } else if (isMilitaryAddress(address)) {
-      const militaryPostOfficeTypeCode = address.militaryPostOfficeTypeCode || '';
-      const militaryStateCode = address.militaryStateCode || '';
-      cityStatePostal = `${militaryPostOfficeTypeCode}, ${militaryStateCode} ${zipCode}`;
+      const city = address.city || '';
+      const militaryStateCode = address.state || '';
+      cityStatePostal = `${city}, ${militaryStateCode} ${zipCode}`;
```
cvalarida: Ideally, I'd ask you to nix this right here, but again, I want to get this merged as soon as possible, so we'll clean this up in the aforementioned ticket.

test/letters/components/AddressBlock.unit.spec.jsx
```diff
@@ -8,7 +8,8 @@ const defaultProps = { name: 'Gary Todd' };
 describe('<AddressBlock/>', () => {
   it('should render', () => {
     const tree = SkinDeep.shallowRender(<AddressBlock { ...defaultProps }/>);
-    const vdom = tree.getRenderOutput();
-    expect(vdom).to.exist;
+    const helpText = tree.subTree('p').text();
+
+    expect(helpText).to.contain('A correct address is not required, but');
```
cvalarida: Awesome!

cvalarida (approved):
Looks great! I'm excited to get this sucker merged into master!

src/js/letters/components/UpdateFailureAlert.jsx
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+
+const UpdateFailureAlert = (props) => {
```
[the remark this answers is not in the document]
markgreenburg: Also implements SAVE_ADDRESS_FAILURE action and introduces new address save error message

src/js/letters/components/InvalidAddress.jsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+
+const InvalidAddress = () => (
```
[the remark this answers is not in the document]
markgreenburg: Yep, makes sense. Fixed.

src/js/letters/components/UpdateFailureAlert.jsx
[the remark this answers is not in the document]
markgreenburg: Yeah that's an oversight, changed to destructure props

PR #6439, opened 2017-09-20
Improvements for mobile navigation

ncksllvn:
This includes fixes for when the bottom-most level of the navigation menu is open on mobile devices. Previously the menu opened overtop the other parent menu options, which is sort of confusing in terms of the user experience because some of the menu options are cut off while others are not. See the issue below for more information and screenshots.

Connects to department-of-veterans-affairs/vets.gov-team/issues/4632.


ncksllvn:
The one potential issue I see at the moment is that the "This site is a work in progress..." message doesn't appear when the bottom-most level is open, because I didn't see any easy way to do that without repeating the same HTML. I'm not sure if this is problematic or not.

src/sass/modules/_m-vet-nav.scss
```diff
@@ -52,6 +53,10 @@ body.va-pos-fixed {
     position: relative;
   }
 
+  .vetnav--submenu-expanded & {
+    height: 0;
+  }
+
```
ncksllvn: It maybe be better to put this in the parent class instead to avoid the sort-of-confusing higher-level selector inside a lower-level element, but I liked the way this was nested and this way I didn't have to repeat the `#vetnav-menu` selector anywhere else.

melwoodard (changes requested):
@ncksllvn when I dig into the second level menu (e.g. Health Care) I'm unable to scroll within that menu. I'm on iOS 10/Safari.

ncksllvn:
@melwoodard Thanks for the feedback. I'm having a little trouble reproducing it - is the "Apply Now" button visible? That should be last element visible as part of that menu.

melwoodard:
Yes, on any of the longer menus that overflow the viewport, I can't scroll down to see all the links like the Apply now button. I'm on an iPhone 5s, so small screen problems FTW :)

ncksllvn:
@melwoodard good catch! Do you mind taking a look now?

src/sass/modules/_m-vet-nav.scss
```diff
@@ -1,5 +1,5 @@
 // The distance from the top of the sceen that the mobile navigation menu should appear
-$height-offset: 111.16px;
+$height-offset: 121.16px;
```
ncksllvn: This value was increased because I removed the margin properties from the menu. The margin was causing weird scroll issues on mobile where some menus would be cut off by the value of the margin.

src/js/legacy/mega-menu.js
```diff
@@ -67,11 +67,13 @@ class MegaMenu {
...
-    const menu = target.getAttribute('aria-controls');
+    const dropdown = this.getMenu(target.getAttribute('aria-controls'));
 
     event.stopPropagation();
     target.setAttribute('aria-expanded', false);
-    this.getMenu(target.getAttribute('aria-controls')).setAttribute('hidden','hidden');
+    dropdown.setAttribute('hidden','hidden');
+    
+    this.menu.classList.remove('vetnav--submenu-expanded');
   }
```
ncksllvn: The submenu's new style hides the parent menu items, so there needs to be a top-level CSS class added in order to apply that style. I didn't see a more elegant way than to just add that logic here.

src/sass/modules/_m-vet-nav.scss
```diff
@@ -35,18 +35,25 @@ body.va-pos-fixed {
...
     white-space: normal;
   }
+
+  &.vetnav--submenu-expanded {
+    #vetnav-menu {
+      overflow: hidden;
+      height: 0;
+    }
+  }
 }
```
ncksllvn: The submenu works by hiding the upper-level menu via setting the height to zero. This hides all of the parent menu items, while the submenu is positioned absolute so it still shows up.

melwoodard (approved):

melwoodard:
Works now, thanks!

src/sass/modules/_m-vet-nav.scss
```diff
@@ -36,29 +35,30 @@ body.va-pos-fixed {
...
 
 #vetnav-menu {
   height: auto;
   list-style: none;
-  margin: 1rem 0;
-  overflow-y: auto;
+  margin: 0;
```
U-DON: Is removing the top margin on mobile necessary? I personally think it looks better to balance out the spacing above and below the close button.

U-DON (approved):
LGTM. Yeah, the WIP message doesn't show up on the submenu, but I think it's fine.

src/js/legacy/mega-menu.js
```diff
@@ -67,11 +67,13 @@ class MegaMenu {
 
   closeMenu(event) {
     const target = event.target;
-    const menu = target.getAttribute('aria-controls');
+    const dropdown = this.getMenu(target.getAttribute('aria-controls'));
 
     event.stopPropagation();
     target.setAttribute('aria-expanded', false);
-    this.getMenu(target.getAttribute('aria-controls')).setAttribute('hidden','hidden');
+    dropdown.setAttribute('hidden','hidden');
```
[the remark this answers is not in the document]
ncksllvn: Good catch!

src/sass/modules/_m-vet-nav.scss
ncksllvn: The `$height-offset` has been increased to compensate for the margins being dropped so it should look the same. The margins were causing some of the content to be slightly cut off when scrolling on mobile devices because of the way height is calculated. We can increase that `$height-offset` though if you want to balance it out more.

U-DON: It's fine. Testing on the menu before the change indicates that the spacing under the button shouldn't be there anyway.

ncksllvn:
Thanks @melwoodard, thanks @U-DON!

PR #6441, opened 2017-09-20
Update file number validation

jbalboni:
Pulls in the updated schema to get the validation pattern change and updates the message based on Peggy's feedback.

screen shot 2017-09-20 at 2 08 15 pm


cehsu (approved):
LGTM.

src/js/burials/config/form.js
```diff
@@ -162,7 +162,7 @@ const formConfig = {
                 widgetClassNames: 'usa-input-medium'
               },
               'ui:errorMessages': {
-                pattern: 'File number must be 8 digits'
+                pattern: 'Your VA file number must be between 7 to 9 digits.'
```
cehsu: Are periods common at the end of validation messages? Just asking as it looks like the previous message didn't have one.

jbalboni: @peggygannon asked for it, but no, I don't think most of our messages have them.

cehsu: @peggygannon should we be adding periods to the end of validation messages of a certain length (or even all of them)?

peggygannon: @cehsu I had originally indicated to add a period, but if previous validation messages don't have them, then I'm OK with keeping that style.

jbalboni: I will go look through the ones we have to be sure.

peggygannon: @cehsu @jbalboni one more thought... it's be great to audit our validation messages to see if they have/don't have periods.

jbalboni: I took a quick look, I didn't see an instance of one with a period.

peggygannon: @jbalboni OK, great. Thanks for checking!

PR #6443, opened 2017-09-20
Update HCA with new dependent options

jbalboni:
This updates HCA so that the children list is now a dependent list and adds some extra relationship type (in the latest version of the schema).

localhost-3001-health-care-apply-application-household-information-child-information 1


cehsu (approved):

src/js/hca/definitions/child.js
```diff
@@ -64,35 +60,35 @@ export const uiSchema = {
...
-  childEducationExpenses: currencyUI('Expenses paid by your dependent child for college, vocational rehabilitation, or training (e.g., tuition, books, materials)'),
+  childEducationExpenses: currencyUI('Expenses your dependent paid for college, vocational rehabilitation, or training (e.g., tuition, books, materials)'),
   childCohabitedLastYear: {
-    'ui:title': 'Did your child live with you last year?',
+    'ui:title': 'Did your dependent live with you last year?',
     'ui:widget': 'yesNo'
   },
   'view:childSupportDescription': {
-    'ui:description': 'Count child support contributions even if not paid in regular set amounts. Contributions can include tuition payments or payments of medical bills.'
+    'ui:description': 'Please count all support contributions even if they weren’t paid in regular and set amounts. Support can include tuition or medical bill payments.'
```
cehsu: Does this pertain to the following question? If so, should it be visually located closer to it?

jbalboni: Yeah, it looks like it's grouped with the question above it, which is odd. I'll make a ticket for it.

PR #6444, opened 2017-09-20
Modal based signin flow

bshyong:
Refactor of identity and signin code to support a modal-based UX

Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4856


bshyong:
currently passing all tests except for accessibility; will dive into it and fix those tomorrow

src/js/login/components/SearchHelpSignIn.jsx
```diff
@@ -64,15 +67,9 @@ const mapStateToProps = (state) => {
...
-    onUpdateLoggedInStatus: (update) => {
-      dispatch(updateLoggedInStatus(update));
-    },
-    onClickSearchHelpSignIn: (menu, isOpen) => {
-      dispatch(toggleSearchHelpUserMenu(menu, isOpen));
-    }
-  };
+const mapDispatchToProps = {
+  toggleLoginModal,
+  toggleSearchHelpUserMenu,
```
U-DON: :+1:

src/js/login/containers/Main.jsx
```diff
@@ -103,21 +185,37 @@ const mapStateToProps = (state) => {
...
+      getUserData(dispatch);
+    },
   };
 };
 
+Main.propTypes = {
+  onLoggedIn: PropTypes.func,
+  renderType: PropTypes.oneOf([
+    'navComponent',
+    'signinModal',
```
U-DON: This one doesn't seem to get used.

jbalboni:
@bshyong @U-DON I pushed a commit to fix up the SiP components based on Ben's changes (and Eugene's comments). There's duplicated logic between SaveFormLink and SignInLink, which I can fix, but I don't want to clutter up this PR. I can always do it as tech debt later.

jbalboni:
@bshyong I'm not getting redirected off the verify page when I sign in. It also seems like I shouldn't be getting the verify page when the test users.

Steps:
1. Click claim status link from nav while not signed out
2. Modal popups, sign in
3. Verify page opens
4. Click link to verify, id.me popup opens, then closes without an user input
5. Nothing happens

If I then go directly to the /track-claims page, it loads for me.

bshyong:
@jbalboni I pushed a fix for the verification issue. It looks like a normal ID.me test user comes back with account level = 1 from the backend, so it's correct that they should be sent to verification. After verification the API shows that they are level 3, in which case they should be redirected to the correct app. This flow is working now. 

The issue of whether the test users should be 1 or 2 or 3 is more of a data issue that we can look into

saneshark:
@bshyong are you we in a position to get this merged into your kudos-identity branch and pushed to master so we can test on staging? Was hoping to have it in there for testing at 1:00 pm EST.

PR #6445, opened 2017-09-21
Modal based signin flow

bshyong:
For testing purposes


saneshark (approved):
@jbalboni and @U-DON i'm going to go ahead and give this a green-light so we can merge to staging. It will get rolled back and maybe one of you could give it a good look over.

PR #6447, opened 2017-09-21
Use only `city` and `state` on the front end

cvalarida:
This should simplify handling the city and state fields.

The idea is that we'll translate their military address equivalent when talking to the api, but keep them as `city` and `state` in the front-end code.


cvalarida:
I ended up skipping the tests 😞 

We can circle back around to this later in [this ticket](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4961).

src/js/letters/utils/constants.js
```diff
@@ -133,3 +133,9 @@ export const STATE_CODE_TO_NAME = {
   PI: 'Virgin Islands'
 };
 
+export const addressTypes = Object.freeze({
```
annekainicUSDS: Is it weird to add a constant variable that's not capitalized? I don't know too much about the conventions here, but that seems to be what we're following for the other constants.

src/js/letters/actions/letters.js
```diff
@@ -215,10 +227,18 @@ export function saveAddressFailure(address) {
 }
 
 export function saveAddress(address) {
+  const transformedAddress = Object.assign({}, address);
+  if (transformedAddress.type === addressTypes.military) {
```
annekainicUSDS: I assume the code to actually set the `type` before saving will be somewhere else?

src/js/common/helpers/api.js
```diff
@@ -37,5 +37,6 @@ export function apiRequest(resource, optionalSettings = {}, success, error) {
 
       return data;
     })
-    .then(success, error);
+    .then(success)
+    .catch(error);
```
annekainicUSDS: Is there a reason for this change? It behaves the same I assume, but might just be more descriptive of what's actually happening?

annekainicUSDS (commented):
LGTM! Just a few questions

src/js/letters/utils/constants.js
cvalarida: Yeah, that's fair. I can capitalize it. 👍

src/js/letters/actions/letters.js
cvalarida: Yup! In another PR.

src/js/common/helpers/api.js
cvalarida: Yeah, this one was tricky. Basically, if the `success` function threw an error for whatever reason, it wouldn't be caught by `error`. I was running into that while trying to mock out the `fetch`.

annekainicUSDS: Ah okay, good catch!

annekainicUSDS (approved):
LGTM!

PR #6451, opened 2017-09-21
Address validation

cvalarida:
**Update:** This PR is ready for review, but it doesn't cover all the validations. I'll make a ticket to cover the rest of the validations and get cracking on it after this gets merged (hopefully!). Once this is merged, adding validations should be as easy as adding a function to an array, so shouldn't take too long.

**Note:** I made a little helper action to get me past the error gate. It's still in this PR, but completely commented out.


cvalarida:
**Update:** I've now got the validation "engine" up and running, but I still need to write the individual checks. Also, it'll be super easy to add new validations now, so we should be golden ~if~ when requirements change.

Moving forward, I just need to write the validations and resolve the hundred-and-one merge conflicts when @markgreenburg gets his branch merged into master.

src/js/letters/components/Address.jsx
```diff
@@ -101,74 +95,84 @@ class Address extends React.Component {
...
+  addressTwo: PropTypes.string,
+  addressThree: PropTypes.string,
+  city: PropTypes.string,
+  state: PropTypes.string,
+  country: PropTypes.string,
+  zipCode: PropTypes.string
+});
+
+Address.propTypes = {
+  // value = address
```
annekainicUSDS: Think you can remove this?

src/js/letters/containers/AddressSection.jsx
```diff
@@ -43,38 +63,122 @@ export class AddressSection extends React.Component {
     }
   }
 
-  handleSave = () => {
-    this.setState({ isEditingAddress: false });
+  /**
+   * Runs all the valiations against the address passed as a prop for a given field.
```
annekainicUSDS: Minor typo: `validations`

src/js/letters/containers/AddressSection.jsx
```diff
@@ -43,38 +63,122 @@ export class AddressSection extends React.Component {
...
+    // If there is no validations array for that field, assume it has no validations
+    if (!validations) {
+      return undefined;
+    }
+
+    let errorMessage = false;
+    for (let i = 0; i < validations.length; i++) {
+      // this.props.value = address
+      errorMessage = validations[i](fullAddress[fieldName], fullAddress);
+      if (typeof errorMessage === 'string') {
```
annekainicUSDS: So will this just return the first errorMessage that is applicable to this field? But there could be multiple error messages if there are multiple validations?

src/js/letters/containers/AddressSection.jsx
```diff
@@ -43,38 +63,122 @@ export class AddressSection extends React.Component {
...
-      );
+  /**
+   * Infers the address type from the address supplied and returns the address
+   *  with the "new" type.
+   */
+  inferAddressType = (address) => {
+    let type = ADDRESS_TYPES.domestic;
+    if (!['USA', 'US'].includes(address.country)) {
+      type = ADDRESS_TYPES.international;
+    } else if (address.militaryStateCode) {
```
annekainicUSDS: Hm, will this ever be true? Aren't we deleting `militaryStateCode` in the actions? Think we may need to instead check if the `['AE', 'AA', 'AP'].includes(address.state)` or something?

src/js/letters/containers/AddressSection.jsx
```diff
@@ -43,38 +63,122 @@ export class AddressSection extends React.Component {
...
+    if (fieldName === 'country') {
+      address.state = '';
+    }
+
+    // Make sure we've got the right address type (domestic, military, international)
+    address = this.inferAddressType(address);
+
+    // Update the error messages
+    // TODO: This might get super slow, so we can debounce this part if necessary...probably
+    const errorMessages = this.validateAll(address);
```
annekainicUSDS: Is there a reason we can't run `this.validateField(fieldName, address)` here instead of running the validations on all fields?

src/js/letters/utils/validations.js
```diff
@@ -0,0 +1,56 @@
...
+  // Could return anything that isn't a string, really
+  return true;
+};
+
+
+/**
+ * These validation functions must return an error message string if the validation fails.
+ */
+export const addressOneValidations = [
+  (input, fullAddress) => requiredValidator(input, fullAddress, 'Please enter an address')
```
annekainicUSDS: How does this syntax work? How does it know what `input` and `fullAddress` are? This looks like magic!

src/js/letters/utils/validations.js
```diff
@@ -0,0 +1,56 @@
...
+];
+
+export const postalCodeValidations = [
+  // Require zip for US address
+  (postalCode, fullAddress) => {
+    if (fullAddress.country === 'USA' && !postalCode) {
+      return 'Please enter a zip code';
```
annekainicUSDS: I think it should be "ZIP"

src/js/letters/utils/validations.js
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    return true;
+  },
+  // Check for valid US zip codes
+  (postalCode, fullAddress) => {
+    if (['US', 'USA'].includes(fullAddress.country) && !isValidUSZipCode(postalCode)) {
+      return 'Please enter a valid zip code';
```
annekainicUSDS: Same here with "ZIP"

annekainicUSDS (commented):
Looks great so far! Few questions and comments. Also super helpful comments in the code, that was great!

src/js/letters/components/Address.jsx
```diff
@@ -13,26 +14,19 @@ import { isNotBlank, isBlankAddress, isValidUSZipCode } from '../../common/utils
...
-    this.isValidAddressField = this.isValidAddressField.bind(this);
-    this.isValidPostalCode = this.isValidPostalCode.bind(this);
-  }
-
   componentWillMount() {
     this.id = _.uniqueId('address-input-');
   }
 
-  isValidAddressField(field) {
+  isValidAddressField = (field) => {
```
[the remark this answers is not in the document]
markgreenburg: ES6's arrow functions don't have their own 'this', so just use the 'this' of the calling function. So basically you don't have to bind the method in the constructor.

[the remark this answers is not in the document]
cvalarida: Yup, exactly what @markgreenburg said. I just think it's nicer than using the constructor to bind `this` to all the functions. It's really not super important in this case, but when you pass functions as props to children, things get confusing since `this` is perhaps not what you expect. I just like to be consistent with it when I think about it.

src/js/letters/components/Address.jsx
```diff
@@ -42,33 +36,33 @@ class Address extends React.Component {
...
 
-  isMilitaryCity(city) {
+  isMilitaryCity = (city) => {
     const lowerCity = city.toLowerCase().trim();
 
     return lowerCity === 'apo' || lowerCity === 'fpo' || lowerCity === 'dpo';
   }
 
-  adjustStateNames(stateList, militaryStates) {
+  adjustStateNames(state, militaryStates) {
```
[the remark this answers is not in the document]
cvalarida: There's actually no need to pass any parameters, so...I just nixed them and used the constants we're importing.

src/js/letters/components/Address.jsx
```diff
@@ -101,74 +95,84 @@ class Address extends React.Component {
...
           name="city"
           autocomplete="address-level2"
           charMax={30}
           value={this.props.address.city}
           required={this.props.required}
-          onValueChange={(update) => {this.props.onInput('city', update);}}/>
-        {stateProvince}
-        <ErrorableTextInput errorMessage={this.isValidPostalCode(this.props.address.zipCode) ? undefined : 'Please enter a valid Postal code'}
+          onValueChange={(update) => this.props.onInput('city', update)}/>
+        <ErrorableSelect errorMessage={errorMessages.state}
```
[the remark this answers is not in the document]
cvalarida: So I was looking at the [validation table](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4110#issuecomment-328640426) and saw that state is not even allowed for international addresses...which I thought was odd, but I was gonna make it and then get somebody who knows better to review the thing when I was done. In the comment for that field, it says "...BGS does not provide a means to store state or province for international addresses..."

I'll make a list of things that I thought were weird and want confirmation on before we call it final.

What I missed, though, was actually hiding the field if USA isn't the country...

src/js/letters/components/Address.jsx
```diff
@@ -101,74 +95,84 @@ class Address extends React.Component {
...
+          label="State"
+          name="state"
+          autocomplete="address-level1"
+          options={adjustedStateNames}
+          value={this.props.address.state}
+          required={this.props.required}
+          onValueChange={(update) => this.props.onInput('state', update)}/>)
+
+        {/* Hide the zip code for addresseses that aren't in the US */}
+        {selectedCountry === 'USA' && <ErrorableTextInput errorMessage={errorMessages.zipCode}
```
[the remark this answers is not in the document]
cvalarida: I don't actually know if it'll fail the back end, but the table says it's not allowed for international addresses, so I figured we might as well take it out.

src/js/letters/components/Address.jsx
cvalarida: Ah! Yes. Because now `this.props.value` actually _is_ `this.props.address`. Good catch!

src/js/letters/containers/AddressSection.jsx
cvalarida: Eagle-eye Anne to the rescue! Thanks. 😄

src/js/letters/containers/AddressSection.jsx
cvalarida: Yes, exactly. So this will return the first validation message, but if they have multiple validation errors, they'll have to be displayed / resolved one at a time. I thought it made more sense than something like "Please enter a zip code and please enter a valid zip code" or something, so we just show one message at a time.

src/js/letters/containers/AddressSection.jsx
cvalarida: This smells like a bad merge...Good catch; I'll look into it.

src/js/letters/containers/AddressSection.jsx
cvalarida: Yeah, I'd prefer to do that, but the reason I have it this way is because if one field changes, it may change the validations on other fields.

For example, we might have a validation that says the city needs to be a valid `militaryPostOfficeTypeCode` (FPO and friends, I think) if it's a military address. So if we set our country to USA and state to AE, but city to San Francisco, there'll be a validation error on the city, but as soon as we change the state to California, we want that validation error to go away, so if we don't validate all the fields, we'd need to touch the city again to reset the validation.

To keep it as performant as possible, I keep track of all the fields that have been modified so we only run validation on them instead of _all_ the fields. I _think_ that'll be okay, but...I haven't really tested it out extensively, if I'm honest. 😳

src/js/letters/utils/validations.js
cvalarida: In `AddressSection`, `validateField()` passes the validation functions these params:
```js
errorMessage = validations[i](fullAddress[fieldName], fullAddress);
```

All validations functions have to have this signature to work.

Would it be helpful for me to write up some documentation in the comments about this?

src/js/letters/containers/AddressSection.jsx
```diff
@@ -144,14 +144,22 @@ export class AddressSection extends React.Component {
     let type = ADDRESS_TYPES.domestic;
     if (!['USA', 'US'].includes(address.country)) {
       type = ADDRESS_TYPES.international;
-    } else if (address.militaryStateCode) {
-      // TODO: Make sure we clear this out if a state code is selected
+    } else if (['AE', 'AA', 'AP'].includes(address.state)) {
+      // Are these ^^ constants anywhere?
```
markgreenburg: I haven't seen them anywhere, I think you'd have to map over the `militaryStateNames` array in `helpers` (but I think you already know this...)

src/js/letters/containers/AddressSection.jsx
```diff
@@ -144,14 +144,22 @@ export class AddressSection extends React.Component {
...
       type = ADDRESS_TYPES.military;
     }
 
     return Object.assign({}, address, { type });
   }
 
+  handleCancel = () => {
```
markgreenburg: Thought this was already in master? Hm.

markgreenburg: Oh I guess this came in from the merge, forgot this wasn't rebased in.

cvalarida: Yeah, it was just part of a merge conflict, so here it is...again. Yay, reruns!

src/js/letters/containers/AddressSection.jsx
cvalarida: Ahh, I didn't realize it was there...not sure if I should use that or not right here, though.

src/js/letters/components/Address.jsx
```diff
@@ -79,96 +48,113 @@ class Address extends React.Component {
...
+    const errorMessages = this.props.errorMessages;
 
     // Our hard-coded list of countries has the value for the U.S. as
     // 'USA', but the value sent to us from EVSS is 'US'. This will cause
     // the field to not populate correctly.
     // This will be changed once we pull the real country list from the
     // address endpoint.
     let selectedCountry;
-    if (this.props.address.country === undefined && addressType === 'MILITARY') {
+    if (this.props.address.country === undefined && this.props.address.type === ADDRESS_TYPES.military) {
```
markgreenburg: Are there cases where the country could be undefined and the address type is something other than Military? If so, looks like we're defaulting to setting selectedCountry to the country in the address prop. I take it this maps to an empty country dropdown on the front end?

Do we need the `undefined` check?

cvalarida: Hmmm...I'm actually no sure; I didn't write that bit. Lemme look!

cvalarida: My guess is when we get an address from the api which has a type of `MILITARY`, but no country. We want to select 'USA' on the front end so we can do everything we need to...

@annekainicUSDS to verify.

src/js/letters/containers/AddressSection.jsx
```diff
@@ -43,45 +63,130 @@ export class AddressSection extends React.Component {
...
+  validateAll = (address = this.state.editableAddress) => {
+    const errorMessages = {};
+    Object.keys(fieldValidations).forEach((fieldName) => {
+      errorMessages[fieldName] = this.validateField(fieldName, address);
+    });
+
+    return errorMessages;
+  }
 
+  // TODO: Make sure this doesn't allow us to save the address if there are validation errors
```
markgreenburg: Could just disable the button if there are error messages instead of touching the `handleSave` logic, right?

cvalarida: Yeah, that's an option. We're not doing that anywhere else as far as I know, so I'd have to rope design in on that decision. I'll write another ticket for it.

markgreenburg (approved):
LGTM

src/js/letters/components/Address.jsx
annekainicUSDS: @markgreenburg @cvalarida according to the table in this ticket (https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4110), I believe it is possible for `country` to be `undefined`, and for the `addressType` to be `domestic`. In the table it states that `countryName` is not required for domestic addresses. `country` is only required for international addresses. So in order to guarantee the address is military, I think we need to do both of these checks.

However, since we're now pulling the countries from the API, I'm not sure any of this code is needed anymore? I can address that in a separate PR once this is merged.

src/js/letters/utils/validations.js
annekainicUSDS: Ah okay, yes that helps!

src/js/letters/containers/AddressSection.jsx
annekainicUSDS: Gotcha, that makes sense! We can test more today during QA testing.

src/js/letters/components/Address.jsx
```diff
@@ -79,96 +48,113 @@ class Address extends React.Component {
...
+        {isUSA && <ErrorableSelect errorMessage={errorMessages.state}
+          label="State"
+          name="state"
+          autocomplete="address-level1"
+          options={adjustedStateNames}
+          value={this.props.address.state}
+          required={this.props.required}
+          onValueChange={(update) => this.props.onInput('state', update)}/>}
+
+        {/* Hide the zip code for addresseses that aren't in the US */}
```
annekainicUSDS: I'm still not confident that we're handling this correctly. Even though `state` and `zipCode` aren't required for international addresses, I wonder if sending them will fail validation on the back-end. If not, and if they can be accepted, I think it better to still provide text fields for those questions so that someone can enter that information if they want to. Otherwise, it may not be a valid international address without it. I will add a ticket to check on this separately so it doesn't have to hold up this PR, but would like to return to this question.

src/js/letters/components/ErrorableSelect.jsx
```diff
@@ -109,7 +109,10 @@ ErrorableSelect.propTypes = {
     ])).isRequired,
   required: PropTypes.bool,
   includeBlankOption: PropTypes.bool,
-  value: PropTypes.string.isRequired,
+  value: PropTypes.shape({
+    value: PropTypes.string,
+    dirty: PropTypes.bool
```
annekainicUSDS: Are we doing anything with `dirty` now? Do we know the select will be able to handle this new data structure? Above we're just passing `this.props.value`, if we're changing this to an object we may need to instead pass it `this.props.value.value`.

annekainicUSDS (commented):
Testing this locally I still found a couple of issues:
1. If I delete the `city` (which is defaulted to a military city), the `state` and `zipCode` fields are hidden. If I unselect `USA` and reselect it again, the fields appear again, but this obviously isn't a great UX.
2. Updating the `city` to "Cambridge", the `state` to "Massachusetts", and the `zipCode` to "02138" (which is a valid address) fails validation when I try to update.

I'm trying to debug locally on what is causing both of these issues, will update here once I figure it out! But wanted to provide review feedback asap.

annekainicUSDS:
Figured out the cause of the first issue I identified above (where the `state` and `zipCode` fields are being hidden when they shouldn't be):

The reason why they are being hidden when a military address undergoes a change (like deleting the value in the `city` field) is because the `type` is being set to "INTERNATIONAL", which then causes the `state` and `zipCode` fields to be hidden until "USA" is reselected for the `country`. The reason why the `type` is being set to "INTERNATIONAL" is because on this line https://github.com/department-of-veterans-affairs/vets-website/pull/6451/files#diff-23b2fcfb82ec34456090de782cb49c98R145 we're checking if the `country` is not "US" or "USA". When military addresses are sent to us, they are sent without a value for `country`, so they would satisfy this check, therefore setting the type to "INTERNATIONAL".

Proposed fix: In the actions, when an address is sent to us as a "MILITARY" address, we delete the military state and city fields and reset the values for the regular `city` and `state` fields (https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/js/letters/actions/letters.js#L82). At the same time, make sure the `country` is set to "USA".

annekainicUSDS (approved):
LGTM to merge for now and will fix remaining issues in a separate PRs!

PR #6454, opened 2017-09-22
Convert edu forms to separate bundles instead of lazy loading

cehsu:
- [x] Created form-specific application pages in `/content`
- [x] Created form-specific webpack entries
- [x] Wire up routing and replace lazy loading
- [x] Relocate entries and routes to form-specific directories
- [x] Write a migration to remove the form number from return urls 
- [x] Update migration logic to include metadata and formId in data to be migrated
- [x] Update/write test for new migration
- [x] [DevOps changes](https://github.com/department-of-veterans-affairs/devops/pull/1942)


src/js/edu-benefits/1995-edu-benefits-entry.jsx
```diff
@@ -0,0 +1,35 @@
...
+
+import initReact from '../common/init-react';
+import createRoutes from './1995-routes';
+import form1995Reducer from './1995/reducer';
+import createLoginWidget from '../login/login-entry';
+import createCommonStore from '../common/store';
+
+require('../../sass/edu-benefits.scss');
+
+const store = createCommonStore(form1995Reducer);
```
cehsu: Form-specific reducer is now used here.

src/js/edu-benefits/1995-routes.jsx
```diff
@@ -0,0 +1,18 @@
+import Form1995App from './1995/Form1995App';
+import routes from './1995/routes';
```
cehsu: Form-specific routes are now used here.

src/js/edu-benefits/routes.jsx
```diff
@@ -1,126 +0,0 @@
...
-          callback(null, require('./1990/routes').default);
-        }, 'edu-1990');
-      }
-    }
-  ];
-
-  childRoutes.push({
-    path: '*',
-    onEnter: (nextState, replace) => replace('/')
-  });
```
cehsu: The Pensions app doesn't include this @jbalboni, is it correct to remove here?

config/webpack.config.js
```diff
@@ -17,7 +17,12 @@ const timestamp = new Date().getTime();
 const entryFiles = {
   burials: './src/js/burials/burials-entry.jsx',
   'claims-status': './src/js/claims-status/claims-status-entry.jsx',
-  'edu-benefits': './src/js/edu-benefits/edu-benefits-entry.jsx',
+  '1990-edu-benefits': './src/js/edu-benefits/1990/1990-edu-benefits-entry.jsx',
+  '1990e-edu-benefits': './src/js/edu-benefits/1990e/1990e-edu-benefits-entry.jsx',
+  '1990n-edu-benefits': './src/js/edu-benefits/1990n/1990n-edu-benefits-entry.jsx',
+  '1995-edu-benefits': './src/js/edu-benefits/1995/1995-edu-benefits-entry.jsx',
+  '5490-edu-benefits': './src/js/edu-benefits/5490/5490-edu-benefits-entry.jsx',
+  '5495-edu-benefits': './src/js/edu-benefits/5495/5495-edu-benefits-entry.jsx',
```
cehsu: Is it redundant or helpful to include the form number at the beginning of the entry file name?

jbalboni: I think you can take off the file name

src/js/edu-benefits/routes.jsx
jbalboni: We should probably add it to pensions, actually.

src/js/edu-benefits/1990e/1990e-edu-benefits-entry.jsx
```diff
@@ -0,0 +1,35 @@
...
+import createCommonStore from '../../common/store';
+
+require('../../../sass/edu-benefits.scss');
+
+const store = createCommonStore(reducer);
+
+createLoginWidget(store);
+
+const browserHistory = useRouterHistory(createHistory)({
+  basename: '/education/apply-for-education-benefits/application'
```
jbalboni: You probably need to add the form number to the base url here, and change it to '/' in the form config file.

src/js/edu-benefits/routes.jsx
cehsu: What about adding it in the `createRoutes` helper?

config/webpack.config.js
cehsu: Thanks, removed.

script/build.js
```diff
@@ -422,7 +422,12 @@ if (options.watch) {
     historyApiFallback: {
       rewrites: [
         { from: '^/track-claims(.*)', to: '/track-claims/' },
-        { from: '^/education/apply-for-education-benefits/application(.*)', to: '/education/apply-for-education-benefits/application/' },
+        { from: '^/education/apply-for-education-benefits/application/1990(?![ne])(\/.*)?$', to: '/education/apply-for-education-benefits/application/1990/' },
```
cehsu: Thanks @cvalarida for the regex solution here.

script/build.js
```diff
@@ -422,7 +422,12 @@ if (options.watch) {
     historyApiFallback: {
       rewrites: [
         { from: '^/track-claims(.*)', to: '/track-claims/' },
-        { from: '^/education/apply-for-education-benefits/application(.*)', to: '/education/apply-for-education-benefits/application/' },
+        { from: '^/education/apply-for-education-benefits/application/1990(\/.*)?$', to: '/education/apply-for-education-benefits/application/1990/' },
```
cehsu: Thanks @cvalarida for making the regex solution here more elegant.

cehsu: The redirect currently adds several trailing `/`'s briefly before navigating to `/introduction`

script/build.js
```diff
@@ -422,7 +422,12 @@ if (options.watch) {
     historyApiFallback: {
       rewrites: [
         { from: '^/track-claims(.*)', to: '/track-claims/' },
-        { from: '^/education/apply-for-education-benefits/application(.*)', to: '/education/apply-for-education-benefits/application/' },
+        { from: '^/education/apply-for-education-benefits/application/1990(\/.*)?$', to: '/education/apply-for-education-benefits/application/1990/' }, // eslint-disable-line no-useless-escape
```
cehsu: Thanks to @cvalarida for finding an elegant solution for this.

cehsu: The 1990 redirect adds several `/`'s before navigating to the introduction page. I'm not sure why this is happening only here (for the 1990). Any thoughts on this @jbalboni @cvalarida?

script/build.js
```diff
@@ -422,7 +422,12 @@ if (options.watch) {
     historyApiFallback: {
       rewrites: [
         { from: '^/track-claims(.*)', to: '/track-claims/' },
-        { from: '^/education/apply-for-education-benefits/application(.*)', to: '/education/apply-for-education-benefits/application/' },
+        { from: '^/education/apply-for-education-benefits/application/1990(\/.*)?', to: '/education/apply-for-education-benefits/application/1990/' }, // eslint-disable-line no-useless-escape
+        { from: '^/education/apply-for-education-benefits/application/1990E(.*)', to: '/education/apply-for-education-benefits/application/1990E/' },
```
jbalboni: These appear to be case-sensitive, so I think it should be `^/education/apply-for-education-benefits/application/1990[eE](.*)` (similar change for the 1990N).

src/js/common/schemaform/helpers.js
```diff
@@ -129,6 +129,10 @@ export function createRoutes(formConfig) {
     {
       path: 'confirmation',
       component: formConfig.confirmation
+    },
+    {
+      path: '*',
+      onEnter: (nextState, replace) => replace('/')
```
jbalboni: I think we would want to use `urlPrefix` here: `replace(formConfig.urlPrefix || '/')`.

src/js/edu-benefits/1990/config/migrations.js
```diff
@@ -0,0 +1,15 @@
+import _ from 'lodash/fp';
+
+export default [
+  // 0 -> 1, we have split the edu bundle into form-specific bundles with a new urlPrefix
+  // and this replaces the saved returnUrl if necessary
+  (savedData) => {
+    const savedReturnUrl = _.get('returnUrl', savedData);
```
jbalboni: Have you tested this? I don't think `returnUrl` is in here, I think we need to change some of the common SiP code to pass through the returnUrl and use any updates to it.

src/js/edu-benefits/1990/config/migrations.js
```diff
@@ -0,0 +1,15 @@
+import _ from 'lodash/fp';
+
+export default [
+  // 0 -> 1, we have split the edu bundle into form-specific bundles with a new urlPrefix
+  // and this replaces the saved returnUrl if necessary
+  (savedData) => {
```
jbalboni: Can we put this function in a common location somewhere and pass in the form number, so it isn't duplicated for each form?

script/build.js
```diff
@@ -422,7 +422,12 @@ if (options.watch) {
     historyApiFallback: {
       rewrites: [
         { from: '^/track-claims(.*)', to: '/track-claims/' },
-        { from: '^/education/apply-for-education-benefits/application(.*)', to: '/education/apply-for-education-benefits/application/' },
+        { from: '^/education/apply-for-education-benefits/application/1990(\/.*)?', to: '/education/apply-for-education-benefits/application/1990/' }, // eslint-disable-line no-useless-escape
```
jbalboni: This doesn't work for me, I still get 1990N redirected to the 1990 form. What does work for me is moving this line below the other two 1990 forms and changing it to 

`{ from: '^/education/apply-for-education-benefits/application/1990(.*)', to: '/education/apply-for-education-benefits/application/1990/' }`

That lets the other two 1990 forms be matched first.

cehsu: Great catch and idea @jbalboni, changes made.

script/build.js
cehsu: Great catch and idea @jbalboni, changes made.

src/js/edu-benefits/1990/config/migrations.js
cehsu: Good point, change made.

src/js/edu-benefits/1990/config/migrations.js
cehsu: Good point, change made.

src/js/edu-benefits/config/migrations.js
```diff
@@ -0,0 +1,16 @@
+import _ from 'lodash/fp';
+
+export default [
+  // 0 -> 1, we have split the edu bundle into form-specific bundles with a new urlPrefix
+  // and this replaces the saved returnUrl if necessary
+  (savedData) => {
+    const savedReturnUrl = _.get('return_url', savedData.metadata);
+
+    if (savedReturnUrl.includes(savedData.formId)) {
```
jbalboni: The formId doesn't necessarily match the url prefix. For the 1990, the url prefix is 1990 but the form id is 22-1990.

I think you can make this a standalone function that takes a formId and returns the migration function. You can use that for each form's migration.

So like:

```js
export function urlMigration(formId) {
  return (savedData) => {
    //migration code that you have now, but using formId from above
  };
}
```

Then each form's migration list can look something like `[urlMigration('1990')]`

src/js/edu-benefits/config/migrations.js
```diff
@@ -0,0 +1,16 @@
...
+      const formId = new RegExp(`/${savedData.formId}`);
```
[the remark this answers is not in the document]
cehsu: I had been using the regex functionality to add the forward slash, but as this is can now simply be included when the function is used: e.g. `urlMigration('/1990')`, it is no longer necessary here.

src/js/edu-benefits/config/migrations.js
cehsu: Thanks @jbalboni, that's much simpler, changes made.

src/js/edu-benefits/config/migrations.js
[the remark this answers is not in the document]
cehsu: Good point, removed.

src/js/common/schemaform/save-load-actions.js
```diff
@@ -278,9 +279,15 @@ export function fetchInProgressForm(formId, migrations, prefill = false) {
       // If we’ve made it this far, we’ve got valid form
 
       let formData;
+      let metadata;
       try {
         // NOTE: This may change to be migrated in the back end before sent over
-        formData = migrateFormData(resBody.form_data, resBody.metadata.version, migrations);
+        const dataToMigrate = {
+          formId,
+          formData: Object.assign({}, resBody.form_data),
```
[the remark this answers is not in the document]
cehsu: Good point, change made.

test/edu-benefits/config/migrations.unit.spec.js
```diff
@@ -18,84 +18,84 @@ describe('EDU migrations', () => {
   describe('first migration', () => {
     it('should remove formId from return url', () => {
       const data1990 = {
-        formId: '1990',
+        formId: formConfig1990.formId,
```
cvalarida: Oh interesting. Good idea. Was this causing a problem at any point?

src/js/common/schemaform/save-load-actions.js
```diff
@@ -244,9 +246,15 @@ export function fetchInProgressForm(formId, migrations, prefill = false) {
       // If we’ve made it this far, we’ve got valid form
 
       let formData;
+      let metadata;
       try {
         // NOTE: This may change to be migrated in the back end before sent over
-        formData = migrateFormData(resBody.form_data, resBody.metadata.version, migrations);
+        const dataToMigrate = {
+          formId,
+          formData: _.clone(resBody.form_data),
```
jbalboni: I don't think we need to clone these. I can't see any errors it would help us avoid.

src/js/edu-benefits/config/migrations.js
```diff
@@ -0,0 +1,15 @@
+import _ from 'lodash/fp';
+
+export function urlMigration(formId) {
+  // 0 -> 1, we have split the edu bundle into form-specific bundles with a new urlPrefix
+  // and this replaces the saved returnUrl if necessary
+  return (savedData) => {
+    const savedReturnUrl = _.get('return_url', savedData.metadata);
```
jbalboni: It should be `returnUrl`.

src/js/edu-benefits/config/migrations.js
```diff
@@ -0,0 +1,15 @@
...
+
+    if (savedReturnUrl.includes(formId)) {
```
jbalboni: Can we make this check case insensitive? I think we're likely storing lower case url prefixes, but it seems possible that we have a couple upper case ones (like 1990e versus 1990E).

`savedReturnUrl.toLowerCase().includes(formId.toLowerCase())` or something similar.

jbalboni (approved):
A couple things to fix, but after that it should be all set.

Let's make sure to merge this after 3 today so we don't release it on a Friday. We can coordinate with DevOps to get the changes reviewed and into staging so we can test this.

test/edu-benefits/config/migrations.unit.spec.js
cehsu: Yes, I had hard coded the url prefix rather than the complete form ID, which would've led to a different migration implementation.

src/js/edu-benefits/config/migrations.js
cehsu: The comment here is misleading, `return_url` is correct, I have updated the comment to match, and have also renamed the parameter `urlPrefix` (rather than `formId`, which is also misleading).

src/js/edu-benefits/config/migrations.js
cehsu: Great point, `savedReturnUrl` has been made case insensitive.

src/js/common/schemaform/save-load-actions.js
cehsu: Thanks @jbalboni, change made.

PR #6456, opened 2017-09-22
Update canUpdate field

annekainicUSDS:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4895

This value has already been changed in vets-api (https://github.com/department-of-veterans-affairs/vets-api/pull/1383), so changing on the front-end as well.


markgreenburg (approved):
LGTM! 🦄

PR #6458, opened 2017-09-22
Better error and messaging handling for auto saving forms

jbalboni:
As part of this, I refactoring the save action to hopefully make the code clearer by separating out the error tracking and analytics around the api call from the Redux logic.

Updated messages for auto save error states:

screen shot 2017-09-22 at 3 45 02 pm
screen shot 2017-09-22 at 3 42 42 pm
screen shot 2017-09-22 at 3 40 51 pm

The link text includes the word save if you're not signed in:
screen shot 2017-09-22 at 2 10 57 pm


src/js/common/schemaform/reducers/index.js
```diff
@@ -159,11 +162,32 @@ export default function createSchemaFormReducer(formConfig) {
...
+        if (saveErrors.has(action.status)) {
+          newState.autoSavedStatus = SAVE_STATUSES.notAttempted;
+        }
+
+        return newState;
+      }
+      case SET_AUTO_SAVE_FORM_STATUS: {
+        const newState = _.set('autoSavedStatus', action.status, state);
+
+        // This is the only time we have a saved datetime
```
cehsu: This comment is no longer relevant.

src/js/common/schemaform/save-load-actions.js
```diff
@@ -126,97 +130,54 @@ export function migrateFormData(savedData, savedVersion, migrations) {
...
       });
   };
 }
 
+export function autoSaveForm(...args) {
+  return saveForm('auto', ...args);
+}
+
+export function saveAndRedirectToReturnUrl(...args) {
+  return saveForm('saveAndRedirect', ...args);
```
cehsu: Would it be helpful to use constants for the statuses?

src/js/common/schemaform/sip-api.js
```diff
@@ -29,3 +29,65 @@ export function removeFormApi(formId) {
...
+      event: `${trackingPrefix}sip-form-save-failed`
+    });
+    return Promise.reject(new Error('Missing token'));
+  }
+
+  return fetch(`${environment.API_URL}/v0/in_progress_forms/${formId}`, {
+    method: 'PUT',
+    headers: {
+      'Content-Type': 'application/json',
+      'X-Key-Inflection': 'camel',
```
cehsu: Not directly related to these changes but, do you have any thoughts on whether it's possible to uncomment this where commented out elsewhere (such as in `removeFormApi` above and `fetchInProgressForm` in `src/js/common/schemaform/sip-api.js`) or why this is needed? Should a bug be filed with the [library](https://github.com/vigetlabs/olive_branch), if one isn't already?

jbalboni: I need to check in with Ryan and see if we've updated the backend with the newer version of that library.

test/common/schemaform/SaveFormLink.unit.spec.jsx
```diff
@@ -34,14 +34,24 @@ describe('Schemaform <SaveFormLink>', () => {
   //  prop requirements; they're only passed to LoginModal which we test elsewhere
   const saveInProgressForm = sinon.spy();
   const updateLoginSpy = sinon.spy();
-  it('should render login message when not logged in', () => {
+  it('should render save message when not logged in', () => {
```
cehsu: Interesting, this makes me wonder whether it might be helpful for the user to know that clicking on "Save and finish this application later" will prompt them to sign in. Could it be helpful to rephrase this as "Sign in to save and finish this application later"? @peggygannon @stacyhlinn (Not a blocking question.)

cehsu (approved):

test/common/schemaform/SaveFormLink.unit.spec.jsx
jbalboni: I think the latter phrasing is a little ambiguous as to what happens when you click the link: are you just signing in or are you signing in and saving the application?

cehsu: Good point, if it seems helpful, perhaps "Sign in and save this application to finish it later" is another option. Whether it's helpful is still open to debate.

peggygannon: @cehsu @jbalboni @stacyhlinn I'm wondering if we could set this link as a box, and mirror the SIP messaging treatment that is on form introduction pages. So copy would be in a shaded box with the following message:
You can save this form, and come back later to finish filling it out. [Sign in to save your form in progress]. 

This would also provide consistent messaging to the user throughout the forms.

jbalboni:
I'm letting this PR sit open until the login work is merged (unless it gets delayed), so I can fix any merge issues.

PR #6459, opened 2017-09-22
Add analytics events for login tracking

bshyong:
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4885


robertfairhead (approved):
👍  LGTM All the events match our current setup so we'll pick them up whenever this goes to prod.

PR #6462, opened 2017-09-25
URLsearchparams polyfill

bshyong:
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4999


U-DON (approved):
LGTM. Probably could just use `react-router` as well.

PR #6463, opened 2017-09-25
Handle Edge 14 fetch errors and add more tracking

jbalboni:
We see lots of TypeMismatchErrors when making fetch calls on Edge 14 (http://sentry.vetsgov-internal/vets-gov/website-production/issues/7140/). This appears to be specific to that browser's fetch implementation. This adds some code to force the fetch polyfill on that browser.

The other changes in here are to add some error tracking for client-side errors in our different api functions. These errors are swallowed currently, but they're common enough on Rainbows apps to generate call center tickets, so I would like to see if they're common on other features.

### DIFF
--- a/src/js/common/polyfills.js
+++ b/src/js/common/polyfills.js
@@ -19,9 +19,16 @@ if (!Modernizr.classlist) {
 if (!Modernizr.dataset) {
   require('dataset');  // dataSet accessor support.
 }
-if (!Modernizr.fetch) {
-  require('whatwg-fetch');  // dataSet accessor support.
+
+// Edge 14's fetch implementation throws TypeMismatchErrors seemingly without
+// reason. This is fixed in fetch 15, but we should use the (xhr based) polyfill
+// for 14.
+// https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/8546263/
+if (navigator.userAgent.includes('Edge/14')) {
+  window.fetch = undefined;
 }
+
+require('whatwg-fetch');
 
 // This polyfill has its own test logic so no need to conditionally require.
 require('polyfill-function-prototype-bind');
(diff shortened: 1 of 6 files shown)


cehsu (approved):

test/claims-status/actions.unit.spec.js
```diff
@@ -205,7 +205,7 @@ describe('Actions', () => {
     it('should fetch claims', (done) => {
       const appeals = [];
       fetchMock.returns({
-        then: (fn) => fn({ ok: true, json: () => Promise.resolve(appeals) })
+        'catch': () => ({ then: (fn) => fn({ ok: true, json: () => Promise.resolve(appeals) }) }),
```
cvalarida: Does eslint not complain about unnecessary `'`s?

src/js/common/polyfills.js
```diff
@@ -19,9 +19,16 @@ if (!Modernizr.classlist) {
 if (!Modernizr.dataset) {
   require('dataset');  // dataSet accessor support.
 }
-if (!Modernizr.fetch) {
```
cvalarida: So...why take this check out?

cvalarida (commented):
Wow, that was tedious. Thanks for doing that!

Would it make sense to do something crazy like wrap the global `fetch` to auto-catch and throw an error up to Sentry? Seems a bit overkill, but...curiosity compels me to ask anyhow.

If not, it looks good. I'm only commenting instead of checking because of my question about the modernizr check.

test/claims-status/actions.unit.spec.js
jbalboni: It in fact complains if I don't use them, because `catch` is a reserved word.

src/js/common/polyfills.js
jbalboni: It's not needed, the fetch polyfill already has conditional logic for native fetch implementations: https://github.com/github/fetch/blob/master/fetch.js#L4

jbalboni:
@cvalarida I think using `apiRequest` more is the better way to make sure we're including this tracking everywhere, vs monkey patching `fetch`.

cvalarida:
That makes much more sense. 👍

U-DON (approved):

PR #6466, opened 2017-09-25
Fix for mobile menu resetting when non-link is clicked

ncksllvn:
I was reviewing the functionality for the navigation on both desktop and mobile when I found that on the mobile menu on desktop, if you navigate within the menu, then click the bottom section, the menu will reset. This is because the document's click event is firing while the menu is open. This doesn't occur on the mobile devices I tested because the document's click event does not fire on mobile but I still think it should be fixed just in case there are some mobile devices that do fire the document's click event. 

This is a fix so that anywhere within the whole menu will prevent the document's click event from firing.


U-DON (approved):
LGTM.

PR #6471, opened 2017-09-25
Letters change loading indicator condition

annekainicUSDS:
This PR maintains the loading indicator as long as letters or address is still being fetched. We heard from EVSS that the address for PCIU is fetched from a different source than the address for the letters, so we're going to still load the app if address is unavailable, but just display a message in the address box instead.

### DIFF
--- a/src/js/letters/containers/Main.jsx
+++ b/src/js/letters/containers/Main.jsx
@@ -96,7 +96,7 @@ function mapStateToProps(state) {
   return {
     letters: letterState.letters,
     lettersAvailability: letterState.lettersAvailability,
-    destination: letterState.destination,
+    address: letterState.address,
     addressAvailability: letterState.addressAvailability,
     benefitSummaryOptions: {
       benefitInfo: letterState.benefitInfo,
--- a/src/js/letters/containers/Main.jsx
+++ b/src/js/letters/containers/Main.jsx
@@ -31,11 +31,6 @@ export class Main extends React.Component {
     // If letters are available, but address is still awaiting response, consider the entire app to still be awaiting response
     if (lettersAvailability === AVAILABILITY_STATUSES.awaitingResponse || addressAvailability === AVAILABILITY_STATUSES.awaitingResponse) {
       return AVAILABILITY_STATUSES.awaitingResponse;
-    // If letters have an eligibility error, the regular app content is still rendered,
-    // but address may still be awaiting response. In this case, consider the entire app
-    // to be awaiting response.
-    } else if (lettersAvailability === AVAILABILITY_STATUSES.letterEligibilityError && addressAvailability === AVAILABILITY_STATUSES.awaitingResponse) {
-      return AVAILABILITY_STATUSES.awaitingResponse;
     }
 
     return lettersAvailability;
--- a/src/js/letters/containers/Main.jsx
+++ b/src/js/letters/containers/Main.jsx
@@ -15,6 +15,16 @@ import {
   getAddressStates
 } from '../actions/letters';
 
+const {
+  awaitingResponse,
+  available,
+  backendServiceError,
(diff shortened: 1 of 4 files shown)


src/js/letters/containers/Main.jsx
```diff
@@ -26,10 +26,20 @@ export class Main extends React.Component {
     this.props.getAddressStates();
   }
 
+
+  appAvailability(lettersAvailability, addressAvailability) {
```
markgreenburg: This could probably be defined as a static method instead, but whatever

src/js/letters/containers/Main.jsx
```diff
@@ -26,10 +26,20 @@ export class Main extends React.Component {
...
+    // If letters are available, but address is still awaiting response, consider the entire app to still be awaiting response
+    if (lettersAvailability === AVAILABILITY_STATUSES.awaitingResponse || addressAvailability === AVAILABILITY_STATUSES.awaitingResponse) {
```
markgreenburg: If you want to DRY this up a bit, you could always do a destructuring assignment first and then just use the `awaitingResponse` property in the logic:
`const { awaitingResponse } = AVAILABILITY_STATUSES`. Or you could just leave it alone :)

test/letters/containers/Main.unit.spec.jsx
```diff
@@ -50,51 +51,57 @@ describe('<Main>', () => {
...
+  it('shows a loading spinner when awaiting response', () => {
     const tree = SkinDeep.shallowRender(<Main {...defaultProps}/>);
-    const childText = tree.subTree('span').text();
-    expect(childText).to.equal(testText);
+    expect(tree.subTree('LoadingIndicator')).to.not.be.false;
   });
 
-  it('shows a loading spinner when awaiting response', () => {
-    const props = _.merge({}, defaultProps, { lettersAvailability: awaitingResponse });
+  it('shows a loading spinner when letters is available but address is awaiting response', () => {
```
markgreenburg: This is probably not even necessary, but should we throw in another test to ensure that a loading spinner is displayed when address is available but letters are still pending?

markgreenburg (approved):
👍 Looks great! A few comments that you can use your discretion on but nothing super important. Also tested with local website + docker api + CI and everything seems to work as expected.

test/letters/containers/Main.unit.spec.jsx
annekainicUSDS: Yes I was thinking that too, I will add it!

src/js/letters/containers/Main.jsx
annekainicUSDS: good point, will add that!

src/js/letters/containers/Main.jsx
annekainicUSDS: Not sure I totally understand the suggestion, could you explain a bit more?

markgreenburg: Yeah sorry, that wasn't super helpful. I was terse because I really don't think this makes much difference, but here's a little more context if you're interested:
There's no need for this helper to have any understanding of what's going on inside a specific instance of `<Main />` because we're passing everything it needs via arguments (i.e., there's no need to call `this.whatever` from within the method). Defining as a static will keep the method on the prototype and you'd just use it by calling that way:
```
export class Main extends React.Component {
    static appAvailability(lettersAvail, addressAvail) { . . . }
    render() {
        switch(Main.appAvailability(
            this.props.lettersAvailability,
            this.props.addressAvailability
        ) { . . . }
    . . .
    }
};
```
Another option could be to define the method as a standalone function in the module but outside the class definition so it's clear that it's just a utility helper.

I mean, you could just leave it alone too. It's not like it doesn't work well the way it is 👍

PR #6484, opened 2017-09-26
Change name of Proof of Service Letter

annekainicUSDS:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4944

### DIFF
--- a/src/js/letters/containers/LetterList.jsx
+++ b/src/js/letters/containers/LetterList.jsx
@@ -27,6 +27,7 @@ export class LetterList extends React.Component {
         );
       } else if (letter.letterType === LETTER_TYPES.proofOfService) {
         letterTitle = 'Proof of Service Card';
+        content = letterContent[letter.letterType] || '';
       } else {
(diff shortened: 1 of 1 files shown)


src/js/letters/containers/LetterList.jsx
```diff
@@ -25,6 +25,8 @@ export class LetterList extends React.Component {
             Inquiry Routing & Information System (IRIS)</a>. VA will respond within 5 business days.
           </p>
         );
+      } else if (letter.letterType === LETTER_TYPES.proofOfService) {
+        letterTitle = 'Proof of Service Card';
```
markgreenburg: Do we need to define content like we define content in the `else`?

annekainicUSDS: oops! good catch, will fix that

annekainicUSDS: fixed!

markgreenburg (approved):
:shipit:

PR #6488, opened 2017-09-26
Fix IE bug for signin signup link

bshyong:
resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/5043

### DIFF
--- a/src/js/login/containers/Main.jsx
+++ b/src/js/login/containers/Main.jsx
@@ -89,7 +89,7 @@ class Main extends React.Component {
   }
 
   bindNavbarLinks() {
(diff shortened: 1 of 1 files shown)


U-DON (approved):
Strange. Why does it have to be done this way for IE to work?

U-DON (commented):
Would this have worked?

```
const loginRequiredLinks = document.querySelectorAll('.login-required');
loginRequiredLinks.forEach(...);
```

bshyong:
I believe `document.querySelectorAll` returns a `NodeList` which does not support `forEach`

PR #6492, opened 2017-09-26
Track form data for client failures

jbalboni:
This adds the form data to the extra info sent to Sentry for client errors, to see if that will help us track down the cause of some TypeErrors. I'm filtering out the obvious PII in them, which is pretty consistent across forms.

### DIFF
--- a/src/js/common/schemaform/helpers.js
+++ b/src/js/common/schemaform/helpers.js
@@ -536,7 +536,7 @@ export function getActiveChapters(formConfig, formData) {
 
 export function sanitizeForm(formData) {
   try {
-    const suffixes = ['first', 'last', 'accountNumber', 'socialSecurityNumber', 'dateOfBirth'];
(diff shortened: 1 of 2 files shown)


jkassemi (approved):

src/js/common/schemaform/helpers.js
```diff
@@ -533,3 +533,18 @@ export function getActiveChapters(formConfig, formData) {
 
   return _.uniq(expandedPageList.map(p => p.chapterKey).filter(key => !!key && key !== 'review'));
 }
+
+export function sanitizeForm(formData) {
+  try {
+    const suffixes = ['first', 'last', 'accountNumber', 'socialSecurityNumber', 'dateOfBirth'];
```
cehsu: should `vaFileNumber` also be included here? or any others?

cehsu (approved):
LGTM. Had a question about whether other fields should be included in the list.

src/js/common/schemaform/helpers.js
jbalboni: vaFileNumber is a good one to include, since that might be an SSN.

PR #6494, opened 2017-09-26
Fixes for verify page.

U-DON:
The other half of the conditional would keep the user on the verify page when not logged in, so I removed it.

Also, the styles for modals would be missing, since the verify page isn't a `no-react` page, so it doesn't include it by default. There's a short flash of the modal while it's redirecting, which can look really messed up without it. I think it's worth splitting the Verify page out of Main later when refactoring, but for now, let's include the styles.


bshyong (approved):
LGTM

PR #6497, opened 2017-09-26
Bugfixes for identity

bshyong:
Ensures DS Logon on verify page triggers DSlogon flow

Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/5054
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/5058

### DIFF
--- a/src/js/common/components/RequiredLoginView.jsx
+++ b/src/js/common/components/RequiredLoginView.jsx
@@ -5,7 +5,7 @@ import { intersection } from 'lodash';
 
 import SystemDownView from './SystemDownView';
 import LoadingIndicator from '../../common/components/LoadingIndicator';
-import Verify from '../../login/components/Verify';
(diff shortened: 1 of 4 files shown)


bshyong:
@U-DON if you can get review and merge this one into `kudos-identity` before we merge to master

U-DON (approved):
LGTM. Seeing more and more of a case for Verify to be its own container, but still something to refactor later.

By the way, does Verify need to get `login` and `handleLogin` passed in from RequiredLoginView as well?

bshyong:
Good catch yes I will add that.

bshyong:
Interestingly enough Verify started out as a separate container, then I refactored into a component, now it probably makes more sense as its own container again. We can discuss when we're ready to refactor.

PR #6503, opened 2017-09-27
Clear validations mg

markgreenburg:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/5077#event-1268097423

Resets `errorMessages` and `shouldValidate` to empty so that user doesn't see old validations if they cancel address update and then go to update again.

### DIFF
--- a/src/js/letters/containers/AddressSection.jsx
+++ b/src/js/letters/containers/AddressSection.jsx
@@ -155,9 +155,9 @@ export class AddressSection extends React.Component {
   handleCancel = () => {
     this.setState({
       isEditingAddress: false,
-      editableAddress: this.props.savedAddress,
       errorMessages: {},
-      shouldValidate: {}
+      shouldValidate: {},
+      editableAddress: this.props.savedAddress
     });
   }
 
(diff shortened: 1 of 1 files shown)


src/js/letters/containers/AddressSection.jsx
```diff
@@ -155,6 +155,8 @@ export class AddressSection extends React.Component {
   handleCancel = () => {
     this.setState({
       isEditingAddress: false,
+      errorMessages: {},
```
markgreenburg: @cvalarida this will make you unhappy but I didn't end up defining an initial state object because we don't want to reset `hadLoadedAddress` and I figured that defining an initial state just for `shouldValidate` and `errorMessages` was adding complexity.

cvalarida: Ah, right. I didn't think about that. Good call!

cvalarida (approved):

src/js/letters/containers/AddressSection.jsx
markgreenburg: er, `hasLoadedAddress`

PR #6507, opened 2017-09-27
Switch to display savedAddress in address block

markgreenburg:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/5099, https://github.com/department-of-veterans-affairs/vets.gov-team/issues/5116

Quick fix that just swaps the address used for display from `this.state.editableAddress` to `this.props.savedAddress`


annekainicUSDS (approved):
LGTM!

PR #6508, opened 2017-09-27
Remove code that is no longer needed because EVSS is sending us USA f…

annekainicUSDS:
…or all US country values

Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/5076

Checked several test users and verified all domestic addresses are sent with a country value of "USA".

### DIFF
--- a/src/js/letters/components/Address.jsx
+++ b/src/js/letters/components/Address.jsx
@@ -4,7 +4,7 @@ import _ from 'lodash';
 
 import ErrorableSelect from './ErrorableSelect';
 import ErrorableTextInput from './ErrorableTextInput';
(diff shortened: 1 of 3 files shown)


cvalarida (approved):
If we're confident we'll always get `USA`, it looks good!

annekainicUSDS:
I think so. I tested multiple users, and the response from `v0/address/countries` lists the address as "USA". The only reason it was ever "US" wasn't actually EVSS, but vets-api that had written "US" in the mock response. EVSS has only ever sent us "USA". So I feel confident about it :). Worst case scenario, it won't break anything, it just won't automatically select something in the dropdown for country.

PR #6509, opened 2017-09-27
Fix touched functionality for array field items

jbalboni:
This updates our array field and additional sources field to set the whole array item as touched and updates the field template to read that and act accordingly. This was the original intent, I'm not sure why we moved away from it.


cehsu (approved):
Looks good to me.

PR #6510, opened 2017-09-27
Redirect after successfully verifying.

U-DON:
Resolves department-of-veterans-affairs/vets.gov-team#5119.

When testing out verification from basic DS Logon to ID.me LOA 3, upon successful verification, the main tab would remain the on the verify page instead of redirecting. I don't think the user is actually logged out, so I'm not sure if the `currentlyLoggedIn` prop changes. The account type is sure to change, though, so the change is to condition on that instead.


bshyong (approved):
LGTM

PR #6513, opened 2017-09-27
Finish validations

cvalarida:
There were a bunch of little things. Namely:

- Not requiring country name for military addresses
- Setting the max length for the street address fields to 35
- Inferring a military address type if the city is APO, DPO, or FPO
- Ensuring the validations run the first time a field is edited
- Clearing out state and zip code for when the address type changes to international
  - Didn't want to accidentally send these to the api and get rejected because of that
- Ensuring saving actually runs the validations (and is not allowed if there are validation errors)
- Minor refactors for clarity
- Some added documentation

I hope it all makes enough sense, but as always, please let me know if anything is confusing or just plum wrong!

**Note:** This doesn't have any tests; we should write them in our general test-all-the-things ticket we wrote up the other day to speed this merge along. (I hate doing that...)

---
Character length (won't let me type any more once the max is reached)
image

Required fields (Domestic)
image

Required fields (International)
image

Required fields (Military)
image

Invalid Zip code
image

Invalid Zip suffix
image

Valid Zip code
image

Valid Zip suffix
image

State selection is limited when a military city is chosen
image

Military city is required for military addresses
image

---
**Update:** ~After I changed the city to be a dropdown for military addresses, I removed the `(or APO/FPO/DPO)` but I didn't want to retake all the screenshots, so...just act like that's gone! You can see what it looks like in the last few screenshots.~

**Update again:** Changed it back to `City (or APO/FPO/DPO)`

### DIFF
--- a/src/js/letters/components/Address.jsx
+++ b/src/js/letters/components/Address.jsx
@@ -92,7 +92,7 @@ class Address extends React.Component {
           label="Street address"
           name="address"
           autocomplete="street-address"
-          charMax={30}
+          charMax={35}
           value={this.props.address.addressOne}
           required={this.props.required}
           onValueChange={(update) => this.props.onInput('addressOne', update)}/>
@@ -100,14 +100,14 @@ class Address extends React.Component {
           label="Street address (optional)"
           name="address"
           autocomplete="street-address"
-          charMax={30}
+          charMax={35}
           value={this.props.address.addressTwo}
           onValueChange={(update) => this.props.onInput('addressTwo', update)}/>
         <ErrorableTextInput
           label="Street address (optional)"
           name="address"
           autocomplete="street-address"
-          charMax={30}
+          charMax={35}
           value={this.props.address.addressThree}
           onValueChange={(update) => this.props.onInput('addressThree', update)}/>
         <ErrorableTextInput errorMessage={errorMessages.city}
--- a/src/js/letters/components/Address.jsx
+++ b/src/js/letters/components/Address.jsx
@@ -71,7 +71,7 @@ class Address extends React.Component {
           autocomplete="country"
           options={this.props.countries}
           value={this.props.address.countryName}
-          required={this.props.required && this.props.address.type !== ADDRESS_TYPES.military}
+          required={this.props.required}
           onValueChange={(update) => this.props.onInput('countryName', update)}/>
         <ErrorableTextInput errorMessage={errorMessages.addressOne}
           label="Street address"
@@ -106,7 +106,7 @@ class Address extends React.Component {
             required={this.props.required}
             onValueChange={(update) => this.props.onInput('city', update)}/>
           : <ErrorableSelect errorMessage={errorMessages.city}
-            label="Military Post Office Type Code"
+            label="City"
             name="city"
             options={militaryCityNames}
             value={this.props.address.city}
--- a/src/js/letters/components/Address.jsx
+++ b/src/js/letters/components/Address.jsx
@@ -5,7 +5,7 @@ import _ from 'lodash';
 import ErrorableSelect from './ErrorableSelect';
 import ErrorableTextInput from './ErrorableTextInput';
 import { STATE_CODE_TO_NAME, ADDRESS_TYPES } from '../utils/constants';
-import { militaryStateNames } from '../utils/helpers';
+import { militaryStateNames, militaryCityNames } from '../utils/helpers';
 
 /**
  * Input component for an address.
@@ -61,6 +61,7 @@ class Address extends React.Component {
     const errorMessages = this.props.errorMessages;
     const isUSA = this.props.address.countryName === 'USA';
     const adjustedStateNames = this.getAdjustedStateNames();
+    const isMilitaryAddress = this.props.address.type === ADDRESS_TYPES.military;
 
     return (
       <div>
@@ -94,14 +95,24 @@ class Address extends React.Component {
           charMax={35}
           value={this.props.address.addressThree}
           onValueChange={(update) => this.props.onInput('addressThree', update)}/>
-        <ErrorableTextInput errorMessage={errorMessages.city}
-          label={<span>City <em>(or APO/FPO/DPO)</em></span>}
-          name="city"
-          autocomplete="address-level2"
-          charMax={30}
-          value={this.props.address.city}
-          required={this.props.required}
-          onValueChange={(update) => this.props.onInput('city', update)}/>
+        {/* Text field for non-military, select box for military */}
+        {!isMilitaryAddress
+          ? <ErrorableTextInput errorMessage={errorMessages.city}
+            label="City"
+            name="city"
+            autocomplete="address-level2"
+            charMax={30}
+            value={this.props.address.city}
+            required={this.props.required}
+            onValueChange={(update) => this.props.onInput('city', update)}/>
+          : <ErrorableSelect errorMessage={errorMessages.city}
+            label="Military Post Office Type Code"
+            name="city"
+            options={militaryCityNames}
+            value={this.props.address.city}
+            required={this.props.required}
+            onValueChange={(update) => this.props.onInput('city', update)}/>
+        }
         {/* Hide the state for addresses that aren't in the US */}
         {isUSA && <ErrorableSelect errorMessage={errorMessages.stateCode}
           label="State"
--- a/src/js/letters/components/Address.jsx
+++ b/src/js/letters/components/Address.jsx
@@ -4,8 +4,8 @@ import _ from 'lodash';
 
 import ErrorableSelect from './ErrorableSelect';
 import ErrorableTextInput from './ErrorableTextInput';
-import { STATE_CODE_TO_NAME, ADDRESS_TYPES } from '../utils/constants';
-import { militaryStateNames, militaryCityNames } from '../utils/helpers';
+import { STATE_CODE_TO_NAME } from '../utils/constants';
+import { militaryStateNames } from '../utils/helpers';
 
 /**
  * Input component for an address.
@@ -61,7 +61,6 @@ class Address extends React.Component {
     const errorMessages = this.props.errorMessages;
     const isUSA = this.props.address.countryName === 'USA';
     const adjustedStateNames = this.getAdjustedStateNames();
-    const isMilitaryAddress = this.props.address.type === ADDRESS_TYPES.military;
 
     return (
       <div>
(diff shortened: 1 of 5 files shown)


markgreenburg:
@cvalarida I actually don't think it's possible to have an international military address. All military addresses should have a USA country selection attached, right?

src/js/letters/utils/validations.js
```diff
@@ -16,24 +17,31 @@ const requiredValidator = (input, fullAddress, message) => {
...
-  (input, fullAddress) => requiredValidator(input, fullAddress, 'Please enter an address')
+  (input, fullAddress) => requiredValidator(input, fullAddress, 'Please enter a street address')
 ];
 
 export const postalCodeValidations = [
-  // Require zip for US address
+  // Require zip for US addresses
   (postalCode, fullAddress) => {
-    if (fullAddress.country === 'USA' && !postalCode) {
+    if (fullAddress.countryName === 'USA' && !postalCode) {
```
annekainicUSDS: Ah good catch!

src/js/letters/utils/validations.js
```diff
@@ -42,15 +50,28 @@ export const postalCodeValidations = [
...
+    return true;
+  }
 ];
 
 export const countryValidations = [
-  (input, fullAddress) => requiredValidator(input, fullAddress, 'Please enter a country')
+  (countryName, fullAddress) => {
+    // Country is required for domestic and international, but not military
+    if (fullAddress.type !== ADDRESS_TYPES.military && !countryName) {
+      return 'Please enter select a country';
```
annekainicUSDS: Maybe just `Please select a country`?

annekainicUSDS (approved):
The logic looks good here! My only concerns are more from a UX perspective. I happen to be on a user testing session now, where the user is mentioning that it's overwhelming to get all of these "angry error messages" as soon as he edits a field. Should we think about only showing error messages after they click "Update" the first time? Should we think about scrolling to the first error message after clicking "Update", or showing a message at the top that explains there are error messages to fix? Some sort of UI signal that their click has registered and there's a reason it's not doing anything, because there might be an error that's out of view and it looks like nothing is happening when they click "Update". We can certainly address this in a separate PR, just things to think about as a follow-up!

src/js/letters/utils/validations.js
cvalarida: Heh, woops! I was previously "Please enter a country," but "select" made more sense. Apparently I didn't proofread at all!

Good catch. 😄

cvalarida:
@markgreenburg I thought the same thing about the military address at first, but reading through the [Table (and Comments) of Truth](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4110#issuecomment-328640426), it looks like we can have "international" military addresses. I came to that conclusion by the following evidences:

1. Point 7 in the above issue comment mentions "non-overseas 'military' addresses" are treated like domestic addresses in that they require state and zip
  - Sounds like overseas 'military' addresses shouldn't be treated that way?
2. Military addresses don't require (and in fact disallow) `countryName`, so it sounds like they can be overseas

I do also have the following concerns:

1. Point 2 says to get to a military address, the user must select USA
2. Point 3 says addresses are only considered military if they have a military state code
  - My concern is that it doesn't check for `militaryPostOfficeTypeCode`...but maybe that really doesn't matter?

Hmm...The more I look at it, the more I'm wondering why I changed it...I think because I saw that military addresses don't require (disallow) selecting the country, and if we don't select a country, we can't select a military state code to indicate that it's a military address, so the only check left is in `city` / `militaryPostOfficeTypeCode`.

@raquelromano-gov I'm thinking I need to change the logic back to only determine that the address type is military if the `stateCode` matches one of the military state codes. What do you think?

cvalarida:
@annekainicUSDS Does this branch still have the problem of validating fields before they've been modified? I squashed some bugs that had to do with when validations were run in this branch, but I could have missed something.

The one thing I know is still a "problem" is when you delete the contents of a required field, it'll yell at you because it's empty...even though you're still in the middle of editing it. I couldn't think of a good fix for that except to pass a blur handler from `AddressSection` to `Address` and attach it to all the fields. Then we could add a field `fieldsToValidate` only when it's is blurred, which is probably a bit better UX. We could make another ticket for that.

raquelromano-gov:
> Hmm...The more I look at it, the more I'm wondering why I changed it...I think because I saw that
> military addresses don't require (disallow) selecting the country, and if we don't select a country,
> we can't select a military state code to indicate that it's a military address, so the only check left
> is in `city / militaryPostOfficeTypeCode`.

> @raquelromano-gov I'm thinking I need to change the logic back to only determine that the
> address type is military if the `stateCode` matches one of the military state codes. What do you
> think?

Yes, that was my initial thought when reading through the issue description.

This is super confusing because
1) There are _conceptually_ 2 types of military addresses, but only 1 type as far as the API is concerened
2) The way we infer address types relies on this USA + `militaryStateCode` trick

I read through the [table](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4110#issuecomment-328640426)  and think it is still accurate but maybe this will help.

 * a `MILITARY` type address ONLY refers to overseas addresses, and we are relying on the fact that people living on overseas military bases consider their address to be a USA address
 * based on that expectation, we expect them to choose USA, then see the state dropdown, and pick `AA`, `AE`, or `AP`, so yes, I would switch on that to assign the type as `MILITARY`
 * and then of course, throw away the USA `countryName` before sending the update because the backend doesn't want it
 * a US military base is not recognized as military; it's just a US address; no military state code, no military postal code, just the `DOMESTIC` type fields

> Military addresses don't require (and in fact disallow) `countryName`, so it sounds like they can be overseas

In fact they MUST be overseas. Just assume there's no such thing as a US military address. Ignorance is bliss.

> Point 3 says addresses are only considered military if they have a military state code
> My concern is that it doesn't check for `militaryPostOfficeTypeCode`...but maybe that really doesn't matter?

You're right that both are required. I wrote this before the UI began repurposing the city field to store the `militaryPostOfficeTypeCode` (I thought a new input would pop up, but I think what you all are doing is correct as users are accustomed to using `APO` etc. as city). So  if you currently read the city field and switch on whether it's `APO`, `FPO`, `DPO` to set the address type, that could work, assuming it won't actually pass validation until USA is chosen and a proper military state code is chosen.

annekainicUSDS:
@cvalarida seems to be behaving as expected to me! I'll created a separate ticket to revisit when we want the validations to trigger. LGTM!

cvalarida:
@annekainicUSDS Awesome, thanks!

@raquelromano-gov That makes a lot of sense, thank you! I have one last question, though: If we have country = USA and state = AA (or friends), we have a military address and require `city` (`militaryPostOfficeTypeCode`). Do we require it to be APO (or friends)?

I'm guessing yes, in which case I'll make it a dropdown like you mentioned in that ticket.

raquelromano-gov:
Oh...now you've exceeded my expertise in military addresses, but I think what you're saying is that if your code is APO then you must be in state AA, so why not constrain that in the UI. I didn't know that was true though--defer to someone who knows more about military addresses than I do. @goldenmeanie maybe, who did a bit of research on them?

cvalarida:
No, I was just thinking if your state code is AA, your city can't be San Francisco, so we could make it a dropdown of the available options. I don't actually know what any of the military address bits even mean! 😛

raquelromano-gov:
Ah, OK--yes, I think you're right that a dropdown makes more sense instead of a confusing validation error that says, "you can't live in a regular US city if you have a military state code or vice versa"

Either way, we can't send the update since it will fail validation. Good one for me to add to Alastair's script!

src/js/letters/components/Address.jsx
```diff
@@ -76,32 +77,42 @@ class Address extends React.Component {
...
           onValueChange={(update) => this.props.onInput('addressThree', update)}/>
-        <ErrorableTextInput errorMessage={errorMessages.city}
-          label={<span>City <em>(or APO/FPO/DPO)</em></span>}
-          name="city"
-          autocomplete="address-level2"
-          charMax={30}
-          value={this.props.address.city}
-          required={this.props.required}
-          onValueChange={(update) => this.props.onInput('city', update)}/>
+        {/* Text field for non-military, select box for military */}
```
annekainicUSDS: Hm, will this work? I'm afraid it's going to put users on a weird loop they can't get out of. Let's say they have a defaulted military address, but they want to change it to a regular domestic address. The only options they'll see for the city are the military ones, because their address is defaulted to military, which in turn is based on if a military city or state is selected. And the only options they'll see for the state are the military ones too because those are filtered based on the presence of a military city. So how will they be able to change those fields?

src/js/letters/utils/helpers.jsx
```diff
@@ -292,66 +301,27 @@ export function getStateName(stateCode) {
...
+  if (address.countryName !== 'USA') {
+    type = ADDRESS_TYPES.international;
+  } else if (MILITARY_STATES.has(address.stateCode)) {
+    type = ADDRESS_TYPES.military;
+  }
+
+  return Object.assign({}, address, { type });
+}
 
+export function resetDisallowedAddressFields(address) {
```
annekainicUSDS: Would you mind adding a comment before this function explaining why it's necessary, that we don't pass state and zip for international addresses so we'd want to clear those out? I think that would be helpful documentation to add here, thanks!

annekainicUSDS (commented):
Hello! Just a question about the new logic with which input type we're going to use for `city`. I think it may be creating a problem, so want to review that before we merge.

src/js/letters/components/Address.jsx
cvalarida: Hmm, you make an excellent point... So it _is_ possible to change back to a domestic address, but it's not intuitive.

The problem as it stands right now:
1. To make an address military, the state selection needs to be AA / AE/ AP
2 When the address is military, the city becomes a dropdown where only military post office type codes are displayed
3. When a military "city" is chosen, the state list is restricted to AA / AE / AP
4. This creates a situation where I can't select a "regular" state to change the address type to domestic

The workaround is to select a blank city before changing the state, but that's not really acceptable, is it?

Even without changing it to a dropdown, we'd still have to clear out the city field to expand the list of states to include "regular" states, so this isn't new with this PR, but...still a problem.

I think we've got a couple options:
1. Not limit the states in the dropdown to AA / AE / AP if the city is military
    - The trouble with this is they're likely to enter the city first, which may be...misleading
2. Not have a dropdown for the city
    - They'd have to type out the military "city" name instead of selecting it from a list, but by the time they get to state, they could select whatever state they're in

For now, I'll change it back to a text field and write a ticket with ☝️ in it for @goldenmeanie to decide what's best.

raquelromano-gov: I agree we need input from @goldenmeanie as this is a tough one and I'm not convinced we've nailed a good user experience yet

PR #6514, opened 2017-09-28
Show city for international address types

markgreenburg:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/5116

Adds a simple check to ensure that `cityStatePostal` is defined for international address types - should be equal to address's city for these international addresses per comments here: https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4110#issuecomment-328640426

### DIFF
--- a/src/js/letters/containers/AddressSection.jsx
+++ b/src/js/letters/containers/AddressSection.jsx
@@ -199,18 +199,17 @@ export class AddressSection extends React.Component {
 
     // City, state, postal code: second line of address
     const zipCode = getZipCode(address);
+    const city = address.city || '';
     let cityStatePostal;
     if (isDomesticAddress(address)) {
-      const city = (address.city || '').toLowerCase();
       const state = getStateName(address.stateCode);
       cityStatePostal = `${city}, ${state} ${zipCode}`;
     } else if (isMilitaryAddress(address)) {
-      const city = address.city || '';
       const militaryStateCode = address.stateCode || '';
       cityStatePostal = `${city}, ${militaryStateCode} ${zipCode}`;
     } else {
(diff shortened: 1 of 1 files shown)


src/js/letters/containers/AddressSection.jsx
```diff
@@ -208,7 +208,11 @@ export class AddressSection extends React.Component {
       const city = address.city || '';
       const militaryStateCode = address.stateCode || '';
       cityStatePostal = `${city}, ${militaryStateCode} ${zipCode}`;
+    } else {
```
markgreenburg: Suppose I could have just initialized cityStatePostal in the variable's declaration, but thought it would be easier to understand this way.

src/js/letters/containers/AddressSection.jsx
```diff
@@ -208,7 +208,11 @@ export class AddressSection extends React.Component {
...
+      // Must be an international address, only show a city
+      cityStatePostal = `${address.city || ''}`;
```
annekainicUSDS: While outside of the scope of what this PR is trying to do, this logic looks like it could be cleaned up. Each condition basically checks for `address.city || ''`, so we could probably set `const city = address.city || ''` outside of the condition statements, as is done for `const zipCode`. And then this line could just be:
```cityStatePostal = `${city}`;
```
And then you can get rid of lines 204 and 208. I think it's a bit more concise that way.

annekainicUSDS (approved):
LGTM! One minor comment.

src/js/letters/containers/AddressSection.jsx
markgreenburg: Thanks, good call!

PR #6516, opened 2017-09-28
Set lettersAvailability to unavailable when address fetch fails

markgreenburg:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/5106

Updates reducer for `GET_ADDRESS_FAILURE` action types to also set the `lettersAvailability` store property to `invalidAddressProperty`.

This change results in the Letters app displaying the general system down error message when we get an error while trying to fetch the user's address, instead of showing an address block-level error message about the address not being available.

### DIFF
--- a/src/js/letters/reducers/index.js
+++ b/src/js/letters/reducers/index.js
@@ -85,7 +85,11 @@ function letters(state = initialState, action) {
       };
     }
     case GET_ADDRESS_FAILURE:
-      return _.set('addressAvailable', false, state);
+      return {
+        ...state,
+        addressAvailable: false,
+        lettersAvailability: AVAILABILITY_STATUSES.invalidAddressProperty
+      };
     case GET_BENEFIT_SUMMARY_OPTIONS_SUCCESS: {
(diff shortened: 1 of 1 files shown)


annekainicUSDS:
@markgreenburg hm is this something we want to do? I thought we were going to allow a user to still download their letters even if their address was unavailable. I'd be curious to know if this impacts the letters themselves or only their ability to update their address. Perhaps @rroueche can confirm?

markgreenburg:
@annekainicUSDS hmm yeah confirming with Rachael. I'm inferring intended behavior from the linked ticket, but probably a good idea to double check before moving forward.

cvalarida:
+1 to @annekainicUSDS's concern.

markgreenburg:
I was also under the impression that letters should never be available when the address isn't available, but that's likely an over-generalization given that Rachael mentioned being able to continue downloading letters even after this address error.

markgreenburg:
Determination is that this is not something we want to do right now. Closing for now.

PR #6520, opened 2017-09-28
Fix state path when canceling uploads on claim status

jbalboni:
The path being used to pull the uploader from the Redux state was wrong and was causing an error whenever someone tried to cancel uploads on claim status.


cehsu (approved):
LGTM.

PR #6522, opened 2017-09-28
Fix circular dependency

jbalboni:
This is causing errors in prod right now: http://sentry.vetsgov-internal/vets-gov/website-production/issues/17609/

login-helpers imports login actions, which imports login-helpers.

### DIFF
--- a/src/js/login/actions/index.js
+++ b/src/js/login/actions/index.js
@@ -1,4 +1,4 @@
-import { gaClientId } from '../../common/helpers/login-helpers';
+import { gaClientId } from '../../common/utils/helpers';
 
 export const LOG_OUT = 'LOG_OUT';
 export const TOGGLE_LOGIN_MODAL = 'TOGGLE_LOGIN_MODAL';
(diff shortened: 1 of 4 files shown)


cehsu (approved):

bshyong (approved):
LGTM

src/js/login/actions/index.js
```diff
@@ -62,9 +69,3 @@ export function toggleSearchHelpUserMenu(menu, isOpen) {
   };
 }
 
-export function toggleLoginModal(isOpen) {
```
cehsu: Curious why this got moved (not blocking)?

U-DON (approved):
LGTM.

omgitsbillryan:
LGTM.  I had originally put the GA client ID stuff in `login-helpers` since it was login'ish.  I have no problem moving it to `'helpers` instead.

On a side note, the sentry issue you linked to is about a `TypeError`
```
n.props.toggleLoginModal is not a function
```
you had to dig further to find the circular dependency?

omgitsbillryan (approved):

jbalboni:
@omgitsbillryan Yeah, it totally makes sense to put it in login-helpers, it just happens to create the circular dependency.

Yeah, so the problem was manifesting as everything imported from login/actions in my FormPage component being undefined. I spent 45 minutes tearing my hair out trying to find the reason and started commenting things out in login/actions and then in login-helpers until it magically started working again.

There must be some webpack weirdness around when it manifests, since it definitely doesn't seem to be happening everywhere.

PR #6524, opened 2017-09-29
Pv vic post

patrickvinograd:
Changing VIC from GET to POST. Open to suggestions about a more React-ful way of building/submitting the hidden form. 

Goes with corresponding API route changes in https://github.com/department-of-veterans-affairs/vets-api/pull/1423

### DIFF
--- a/src/js/veteran-id-card/containers/Main.jsx
+++ b/src/js/veteran-id-card/containers/Main.jsx
@@ -11,7 +11,13 @@ class Main extends React.Component {
     super(props);
     this.handleSubmit = this.handleSubmit.bind(this);
     this.renderButton = this.renderButton.bind(this);
-    this.submitVicForm = this.submitVicForm.bind(this);
+    this.renderVicForm = this.renderVicForm.bind(this);
+  }
+
+  componentDidUpdate() {
+    if (this.props.vicUrl) {
+      document.getElementById('vicForm').submit();
+    }
   }
 
   handleSubmit(e) {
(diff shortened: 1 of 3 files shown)


bshyong (approved):
Instead of constructing a form and rendering it into the document, can we submit the object with the data via an API request? the submitForm function could call an action that uses the `apiRequest` helper to submit the form (you'd also be able to dispatch actions on success or error if desired)

src/js/veteran-id-card/containers/Main.jsx
```diff
@@ -58,8 +74,8 @@ class Main extends React.Component {
 
     if (this.props.errors) {
       message = this.renderErrors();
-    } else if (this.props.redirect) {
-      window.location.href = this.props.redirect;
+    } else if (this.props.vicUrl) {
+      this.submitVicForm();
```
jbalboni: Generally, we try not to trigger actions in the render method and let it only render things. It would be better to put this in one of the lifecycle methods, probably `componentDidUpdate()`.

src/js/veteran-id-card/containers/Main.jsx
```diff
@@ -11,13 +11,29 @@ class Main extends React.Component {
...
+    this.submitVicForm = this.submitVicForm.bind(this);
   }
 
   handleSubmit(e) {
     e.preventDefault();
     this.props.initiateIdRequest();
   }
 
+  submitVicForm() {
+    const vicForm = document.createElement('form');
```
jbalboni: I think it would be more React-ish to render this in the render method below, and trigger the submit in this method.

So you could put this in the jsx return in render():

```jsx
{!!this.props.vicUrl && 
  <form id="vicForm" method="POST" action={this.props.vicUrl}>
    {Object.entries(this.props.traits).map(([key, value] => (
      <input type="hidden" name={key} key={key} value={value}/>
    )}
  </form>
}
```

And have the submit in a lifecycle method:

```js
componentDidUpdate() {
  if (this.props.vicUrl) {
    document.getElementById('vicForm').submit();
  }
}
```

jbalboni (approved):
I made some comments assuming that you're stuck with the form submit method, and can't do what Ben mentions above.

patrickvinograd:
We specifically want to submit the form and move the user over to the remote site, rather than submitting an async request. I will use Jeff's approach.

U-DON (approved):
LGTM.

src/js/veteran-id-card/containers/Main.jsx
```diff
@@ -11,15 +11,36 @@ class Main extends React.Component {
     super(props);
     this.handleSubmit = this.handleSubmit.bind(this);
     this.renderButton = this.renderButton.bind(this);
+    this.renderVicForm = this.renderVicForm.bind(this);
+  }
+
+  componentDidUpdate() {
+    if (this.props.vicUrl) {
```
U-DON: Does the presence of certain traits affect whether we want to submit the request? That can be caught here too.

patrickvinograd: Not so far...we do the "is veteran, has edipi" enforcement on the backend and will return an error code accordingly. Other business logic (like making sure honorable discharge) is all responsibility of VIC team so far.

PR #6528, opened 2017-09-29
Check request options against supported vets-api BSL customization op…

markgreenburg:
…tions

Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/5111, https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3872

We should ensure that we're only passing valid options to vets-api when we specify how a user wants their BSL customized. The least "why is this happening" approach seemed to me to create a new constant that reflects which of the benefit options are actually able to be customized in our request to vets-api, and then ensure that our request options object only contains these recognized keys.

In the future, if the `BENEFIT_OPTIONS` we receive from `vets-api` continue to diverge from `REQUEST-OPTIONS` accepted by `vets-api`, we'll be able to just modify the `REQUEST-OPTIONS` constant further.

### DIFF
--- a/src/js/letters/reducers/index.js
+++ b/src/js/letters/reducers/index.js
@@ -6,29 +6,30 @@ import {
   optionsToAlwaysDisplay
 } from '../utils/helpers.jsx';
 import {
+  AVAILABILITY_STATUSES,
   BACKEND_AUTHENTICATION_ERROR,
   BACKEND_SERVICE_ERROR,
-  INVALID_ADDRESS_PROPERTY,
-  GET_LETTERS_FAILURE,
-  GET_LETTERS_SUCCESS,
+  DOWNLOAD_STATUSES,
   GET_ADDRESS_FAILURE,
   GET_ADDRESS_SUCCESS,
-  GET_BENEFIT_SUMMARY_OPTIONS_FAILURE,
-  GET_BENEFIT_SUMMARY_OPTIONS_SUCCESS,
-  GET_LETTER_PDF_DOWNLOADING,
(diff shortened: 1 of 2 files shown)


cvalarida (approved):
Looks great! I was confused at why a bunch of the imports are all changed up, but I see you alphabetized them. I like it. 😄

src/js/letters/reducers/index.js
```diff
@@ -90,8 +91,14 @@ function letters(state = initialState, action) {
...
       Object.keys(benefitInfo).forEach(key => {
-        if ((optionsToAlwaysDisplay.includes(key) || (benefitInfo[key] !== false)) &&
-            !possibleOptions.includes[key]) {
+        if (
+          // the option should always be displayed or vets-api says it is available
+          (optionsToAlwaysDisplay.includes(key) || (benefitInfo[key] !== false))
+          // and the option is not yet in the possibleOptions array
+          && !possibleOptions.includes[key]
+          // and the option is a customization option that vets-api supports
+          &&  REQUEST_OPTIONS[key]
```
annekainicUSDS: This is the smallest of nits, but it looks like there might be an extra space after `&&`?

annekainicUSDS (approved):
LGTM! This would be a great thing to add a unit test for, but I assume we'll get to that in one of the many new unit test tickets :)

src/js/letters/reducers/index.js
markgreenburg: ahh you're right, glad you found it :)
Kind of surprised that the linter didn't yell at me.

PR #6529, opened 2017-09-29
Fix for sidebar disappearing too soon on top landing pages

ncksllvn:
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4600

### DIFF
--- a/content/components/navigation-sidebar.html
+++ b/content/components/navigation-sidebar.html
@@ -15,7 +15,7 @@ Used to display a side navigation bar of pages in a collection.
 
 {% assign relatedPages = collections | get: collection %}
 
-<nav {% if relatedPages != empty %}class="usa-width-one-fourth va-sidebarnav columns medium-3"{% endif %} id="va-detailpage-sidebar">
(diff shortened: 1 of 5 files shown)


ncksllvn:
To expand on this, we had a great set up using Foundation. Foundation's grid was set to collapse the main content area (right) at 640px. However, we had a custom breakpoint defined at 768px, where the left sidebar would disappear (visibility to 0) and a button for toggling a side drawer would appear above the main content. Because of this discrepancy between the breakpoints, the left sidebar would disappear but its place in the grid would still be there on viewports between 640px and 768px. 

The fix for this was to switch to USWDS's grid system (it was using both framework classes before) and a few adjustments.

### Before:
image

### Proposed:
image

ncksllvn:
Just to add, after looking through the SASS, we import the modules strategically so that we override USWDS's medium breakpoint to be 768px (it's 600px by default) before the grid classes are defined. That's why switching entirely to USWDS's grid system worked.

melwoodard (approved):
Looks good from a visual perspective!

ncksllvn:
Thanks @melwoodard! I'm gonna go ahead and merge it because I feel really good about this fix.

PR #6532, opened 2017-09-29
Scroll to first error on save

cvalarida:
Well that was easy. Standing on the shoulders of giants helps. 😄


annekainicUSDS (approved):
This PR makes me so happy :) LGTM!

markgreenburg (approved):
Thanks for fixing this!

PR #6536, opened 2017-10-02
Letters e2e test

annekainicUSDS:
Adds 2 e2e tests, one for unauthed and one for authed flow; each run the axe check as well. 

Not sure if I should be testing more of the download letters functionality, like actually expecting a letter to download. Thoughts?

### DIFF
--- a/test/e2e/letters-helpers.js
+++ b/test/e2e/letters-helpers.js
@@ -1,0 +1,157 @@
+const mock = require('./mock-helpers');
+
+const letters = {
+  data: {
+    attributes: {
+      fullName: 'William Shakespeare',
+      letters: [
+        { name: 'Commissary Letter', letterType: 'commissary' },
+        { name: 'Proof of Service Letter', letterType: 'proof_of_service' },
+        { name: 'Service Verification Letter', letterType: 'service_verification' },
+        { name: 'Benefit Verification Letter', letterType: 'benefit_verification' }
+      ]
+    }
+  }
+};
+
+const benefitSummaryOptions = {
+  data: {
+    attributes: {
+      benefitInformation: {
+        awardEffectiveDate: '1965-01-01T05:00:00.000+00:00',
+        hasAdaptedHousing: false,
+        hasChapter35Eligibility: true,
+        hasIndividualUnemployabilityGranted: false,
+        hasNonServiceConnectedPension: false,
+        hasServiceConnectedDisabilities: true,
+        hasSpecialMonthlyCompensation: false,
+        monthlyAwardAmount: 123,
+        serviceConnectedPercentage: 2
+      },
+      militaryService: [
+        {
+          branch: 'ARMY',
+          characterOfService: 'HONORABLE',
+          enteredDate: '1965-01-01T05:00:00.000+00:00',
+          releasedDate: '1972-10-01T04:00:00.000+00:00'
+        },
+        {
+          branch: 'ARMY',
+          characterOfService: 'UNCHARACTERIZED_ENTRY_LEVEL',
+          enteredDate: '1973-01-01T05:00:00.000+00:00',
+          releasedDate: '1977-10-01T04:00:00.000+00:00'
+        }
+      ]
+    }
+  }
+};
+
+const address = {
+  data: {
+    attributes: {
+      address: {
+        addressEffectiveDate: '2012-04-03T04:00:00.000+00:00',
+        addressOne: '57 COLUMBUS STRASSA',
+        addressThree: '',
+        addressTwo: 'BEN FRANKLIN VILLAGE',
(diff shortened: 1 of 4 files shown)


test/e2e/letters-helpers.js
```diff
@@ -0,0 +1,157 @@
+const mock = require('./mock-helpers');
+
+const letters = {
+  data: {
+    attributes: {
+      fullName: 'William Shakespeare',
```
cvalarida: 💯

test/e2e/letters-helpers.js
```diff
@@ -0,0 +1,157 @@
...
+  mock(token, {
+    path: '/v0/address/states',
+    verb: 'get',
+    value: states
+  });
+
+  mock(token, {
+    path: '/v0/address',
+    verb: 'put',
+    value: newAddress,
```
cvalarida: This mocks returning the `newAddress` when we do a put request, right? Do we know this is what the api returns?

test/letters/00-unauthed.e2e.spec.js
```diff
@@ -0,0 +1,18 @@
...
+const LoginHelpers = require('../e2e/login-helpers');
+
+module.exports = E2eHelpers.createE2eTest(
+  (client) => {
+    const token = LoginHelpers.getUserToken();
+
+    LettersHelpers.initApplicationMock(token);
+
+    // Test flow for unauthed and LOA1 users
+    LoginHelpers.testUnauthedUserFlow(client, '/letters');
```
cvalarida: This is to test that the `RequiredLoginView` doesn't let us in, right?

I wonder if it would make more sense to bundle all these into a single test that pokes all the apps that require the user to be logged in. I think it'd save time in setup and teardown of a browser instance. Multiply that by however many "privileged" apps we have (minus one) and it might be worth our time.

What thinkest thou?

test/letters/01-authed.e2e.spec.js
```diff
@@ -0,0 +1,65 @@
...
+
+    // Update address and cancel
+    client
+      .click('.usa-button-outline')
+      .expect.element('select').to.be.present;
+
+    client
+      .setValue('select[name="country"]', 'United Kingdom')
+      .clearValue('input[name="city"]')
+      .fill('input[name="city"]', 'Stratford-upon-Avon')
```
cvalarida: I was just reading MacBeth the other day, so this makes me probably happier than it should. 😄

test/letters/01-authed.e2e.spec.js
```diff
@@ -0,0 +1,65 @@
...
+      .assert.title('Download VA Letters: Vets.gov')
+      .waitForElementVisible('.letters', Timeouts.slow);  // First render of React may be slow.
+
+    client.axeCheck('.main');
+
+    // Checking if full name has rendered
+    client.expect.element('h5.letters-address').text.to.contain('William Shakespeare');
+
+    // Checking if address has rendered correctly
+    client.expect.element('.street').text.to.contain('57 Columbus Strassa, Ben Franklin Village');
```
cvalarida: Probably should use the address properties in `LettersHelpers` instead of a "magic" string. Not a big deal, though.

cvalarida (approved):
Looks good. I think we should probably take a second look at it when all the unit tests are done so we can see what gaps need to be filled in the e2e test, though.

We can just run through all the unit tests (yeah..."just" run through...), make notes of which tests aren't super comprehensive--that is, when they don't simulate user input--and address those. Basically, we just want to make sure all the things are wired up properly, y'know?

test/e2e/letters-helpers.js
annekainicUSDS: I'm not totally sure how this works, but I think there is a special `/mock` endpoint that we're using for these mock requests. I would have assumed `newAddress` in this case is what is being sent in the request body, but I'm not positive how `/mock` works.

test/letters/00-unauthed.e2e.spec.js
annekainicUSDS: I agree! I think that would be much more efficient. I'll create a ticket for that and address that with the whole FE team at the next sync.

test/letters/01-authed.e2e.spec.js
annekainicUSDS: 💯

test/letters/01-authed.e2e.spec.js
annekainicUSDS: Good point, I can change that!

test/e2e/letters-helpers.js
annekainicUSDS: You're right, this means `newAddress` will be sent in the response when a put request is made. In building the `newAddress` data in the e2e helpers for letters, I based it off of what the real vets-api endpoint returns.

PR #6539, opened 2017-10-02
Helpers unit tests

cvalarida:
Pretty sure I got them all, but would like a second opinion.

Testing `getStateName` seems pretty pointless since it's just an object property lookup + sentry error if the state doesn't exist, so I skipped it by just commenting out some stubs (with comments inline saying why the tests aren't fleshed out).

### DIFF
--- a/test/letters/utils/helpers.unit.spec.jsx
+++ b/test/letters/utils/helpers.unit.spec.jsx
@@ -6,8 +6,18 @@ import { ADDRESS_TYPES, MILITARY_STATES } from '../../../src/js/letters/utils/co
 
 import {
   getBenefitOptionText,
-  inferAddressType
+  inferAddressType,
+  resetDisallowedAddressFields
 } from '../../../src/js/letters/utils/helpers';
+
+const address = {
+  countryName: 'USA',
+  addressOne: '123 Main St N',
+  stateCode: 'MA',
+  zipCode: '12345',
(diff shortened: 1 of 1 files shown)


test/letters/utils/helpers.unit.spec.jsx
```diff
@@ -2,42 +2,61 @@ import { expect } from 'chai';
...
+  type: ADDRESS_TYPES.domestic,
+  city: 'Bygtowne'
+};
 
 describe('Letters helpers: ', () => {
   describe('getBenefitOptionText', () => {
+    // The following tests check for options that should be available / not available given certain conditions
     it('should be defined for both veterans and dependents', () => {
       _.forEach(option => {
         expect(getBenefitOptionText(option, true, true)).not.to.be.undefined;
```
markgreenburg: `to.exist` should ensure that the function doesn't return something that is `===` to `undefined` or `null` - should we use that as a non-negating check or do you think the current way makes more sense?

markgreenburg (approved):
Woohoo, this is huge! LGTM

test/letters/utils/helpers.unit.spec.jsx
cvalarida: 🤔  Yeah......I don't really know what the "right" way to test these is. Maybe truthy / falsey? I'm just gonna merge it for now and we can loop back around later if it makes sense.

PR #6542, opened 2017-10-03
DownloadLetterLink unit tests

cvalarida:


src/js/letters/components/DownloadLetterLink.jsx
```diff
@@ -7,15 +7,10 @@ import { getLetterPdf } from '../actions/letters';
...
-  constructor(props) {
-    super(props);
-    this.downloadLetter = this.downloadLetter.bind(this);
-  }
-
   // Either download the pdf or open it in a new window, depending on the
   // browser. Needs to be manually tested on a variety of
   // vets.gov-supported platforms, particularly iOS/Safari
-  downloadLetter(e) {
+  downloadLetter = (e) => {
```
markgreenburg: yesssss thank you!

markgreenburg (approved):
LGTM! Couple of quick ideas I had since you're here, not related to your changes though:  
* I believe we can remove `SkinDeep` as a dependency here by just swapping out for `ReactTestUtils` in the first couple of tests . 
* The first 'it renders' test may be duplicative, since it's just checking that the vdom exists. What do you think?

cvalarida:
Yeah, it might make more sense to check that it renders something we'd expect it to.

I can do that and swap skin deep for ReactTestUtils at the same time. 👍

PR #6544, opened 2017-10-03
ErrorableSelect unit tests

cvalarida:
These were just copied from the original ErrorableSelect's unit test file with minimal modifications (to remove the dirty field).

I could think of more tests to run than were in the original, but it seemed maybe a bit overkill to add them. :man_shrugging:

### DIFF
--- a/test/letters/components/ErrorableSelect.unit.spec.jsx
+++ b/test/letters/components/ErrorableSelect.unit.spec.jsx
@@ -22,10 +22,10 @@ describe('<ErrorableSelect>', () => {
     });
 
     const select = ReactTestUtils.findRenderedDOMComponentWithTag(errorableSelect, 'select');
-    select.value = '';
+    select.value = 1;
     ReactTestUtils.Simulate.change(select);
 
-    return expect(updatePromise).to.eventually.eql('');
+    return expect(updatePromise).to.eventually.eql('1');
   });
 
   it('no error styles when errorMessage undefined', () => {
--- a/test/letters/components/ErrorableSelect.unit.spec.jsx
+++ b/test/letters/components/ErrorableSelect.unit.spec.jsx
@@ -1,0 +1,103 @@
+import React from 'react';
+import ReactTestUtils from 'react-dom/test-utils';
+import SkinDeep from 'skin-deep';
+import chaiAsPromised from 'chai-as-promised';
+import chai, { expect } from 'chai';
+
+import ErrorableSelect from '../../../src/js/letters/components/ErrorableSelect';
+
+chai.use(chaiAsPromised);
+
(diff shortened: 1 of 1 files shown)


test/letters/components/ErrorableSelect.unit.spec.jsx
```diff
@@ -0,0 +1,103 @@
...
+    let errorableSelect;
+
+    const updatePromise = new Promise((resolve, _reject) => {
+      errorableSelect = ReactTestUtils.renderIntoDocument(
+        <ErrorableSelect label="test" options={options} value={testValue} onValueChange={(update) => { resolve(update); }}/>
+      );
+    });
+
+    const select = ReactTestUtils.findRenderedDOMComponentWithTag(errorableSelect, 'select');
+    select.value = '';
```
markgreenburg: It looks as though the initial value is set to an empty string on L12 / L20, then we're changing it to an empty string again here. So it seems as though we're testing that the change propagated but using the same pre-/post- change values. Might it be better to change one or the other?

And I think the same thing that applied to `ErrorableTextInput` applies here, right? We're only testing that `updatePromise` resolves with the new value (when `onValueChange` is called), but not sure if that's comprehensive (I seriously have no idea, this isn't a leading question!). Either way I don't think another test is strictly necessary in this PR, just wanted to bring this point up.

cvalarida: Good catch on the update. Changed!

As far as the comprehensiveness, I think it's fine because we actually are simulating what happens when the select box is changed. We just have to trust that `ReactTestUtils.Simulate.change` does what it's supposed to and emulates it properly ;)

test/letters/components/ErrorableSelect.unit.spec.jsx
```diff
@@ -0,0 +1,103 @@
...
+
+    const labels = tree.everySubTree('.usa-input-error-label');
+    expect(labels).to.have.lengthOf(1);
+    expect(labels[0].text()).to.equal('my label');
+
+    const errorMessages = tree.everySubTree('.usa-input-error-message');
+    expect(errorMessages).to.have.lengthOf(1);
+    expect(errorMessages[0].text()).to.equal('error message');
+
+    // No error means no aria-describedby to not confuse screen readers.
```
annekainicUSDS: I think this is the wrong comment here? This should be that there is an aria-describedby if error.

annekainicUSDS (approved):
LGTM! One minor nit

test/letters/components/ErrorableSelect.unit.spec.jsx
cvalarida: :man_shrugging: Copy pasta!

PR #6545, opened 2017-10-03
ErrorableTextInput unit tests

cvalarida:
Like the ErrorableSelect, I just copied the tests from the original component. I modified the first test that looked for a dirty value, and removed the test about blur dirtying the field, but the rest is the same.

### DIFF
--- a/test/letters/components/ErrorableTextInput.unit.spec.jsx
+++ b/test/letters/components/ErrorableTextInput.unit.spec.jsx
@@ -1,0 +1,101 @@
+import React from 'react';
+import ReactTestUtils from 'react-dom/test-utils';
+import SkinDeep from 'skin-deep';
+import chaiAsPromised from 'chai-as-promised';
+import chai, { expect } from 'chai';
+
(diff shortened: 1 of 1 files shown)


test/letters/components/ErrorableTextInput.unit.spec.jsx
```diff
@@ -0,0 +1,101 @@
...
+      errorableInput = ReactTestUtils.renderIntoDocument(
+        <ErrorableTextInput field={makeField(1)} label="test" onValueChange={(update) => { resolve(update); }}/>
+      );
+    });
+
+    const input = ReactTestUtils.findRenderedDOMComponentWithTag(errorableInput, 'input');
+    input.value = 'newValue';
+    ReactTestUtils.Simulate.change(input);
+
+    return expect(updatePromise).to.eventually.eql('newValue');
```
markgreenburg: Per discussion, we probably need another test for the `field` prop.

markgreenburg (approved):
Feel free to add that `field` prop test if you think it's worthwhile, but I don't think that should be a blocker. LGTM!

PR #6546, opened 2017-10-03
UpdateFailureAlert unit tests

cvalarida:
I almost got really into the weeds in getting a functional component to render using ReactTestUtils, but wrapping it in a `<div>` did the trick, fortunately.

Also, because it wasn't rendering for me without the `<div>` and I didn't know what was going on, I added a (more) helpful error message to `getFormDOM` for when `findDOMNode()` isn't passed something it can use...like when `ReactTestUtils.renderIntoDocument()` just returns null.

### DIFF
--- a/test/letters/components/UpdateFailureAlert.unit.spec.jsx
+++ b/test/letters/components/UpdateFailureAlert.unit.spec.jsx
@@ -1,0 +1,30 @@
+import React from 'react';
+import ReactTestUtils from 'react-dom/test-utils';
+import { expect } from 'chai';
+
+import { getFormDOM } from '../../util/schemaform-utils';
+
+import UpdateFailureAlert from '../../../src/js/letters/components/UpdateFailureAlert';
+
+const defaultProps = {
+  addressObject: {
+    addressOne: '123 Main St N',
+    city: 'Bigtowne',
+    stateCode: 'BS',
+    countryName: 'Elsweyre'
+  }
+};
+
+describe('<UpdateFailureAlert>', () => {
+  it('should render address and warning', () => {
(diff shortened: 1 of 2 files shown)


test/letters/components/UpdateFailureAlert.unit.spec.jsx
```diff
@@ -0,0 +1,30 @@
...
+    addressOne: '123 Main St N',
+    city: 'Bigtowne',
+    stateCode: 'BS',
+    countryName: 'Elsweyre'
+  }
+};
+
+describe('<UpdateFailureAlert>', () => {
+  it('should render address and warning', () => {
+    // Have to wrap the stateless (function) component in <div></div> so renderIntoDocument() is happy
```
markgreenburg: Good catch! I'm surprised that none of the other teams have run into this problem before. I guess you could also abstract this away by making a wrapping class-based component (e.g., `functionalRender()` that just renders its children) somewhere in the common test helpers but that's probably overkill if the alternative is to wrap things in a `<div/>`.

markgreenburg (approved):
LGTM - thanks for picking this up!

test/letters/components/UpdateFailureAlert.unit.spec.jsx
cvalarida: Heh, I tried doing something like that first, but when I found I could just wrap the silly thing in a `<div>` I nixed it for a few reasons:

1) Because I don't have to write the thing
2) When I write the tests, it's still easier to wrap it in a `<div>` than to import and use that component
3) It's less likely to break in weird, unexpected ways because of the testing wrapper

PR #6548, opened 2017-10-04
[WIP] fixing spacing issues!

phiden:


phiden:
Have fixed footer + header alignments. Nav is still off -- this is a bigger discussion to be had -- but would like to merge the changes made thus far!

ncksllvn (approved):
LGTM! The only things is that `columns` class is for Foundation and as we migrate to USWDS we'll have to figure something out for that (probably a backfill) but otherwise great work!

PR #6549, opened 2017-10-04
Fix pagination component alignment

jbalboni:
Fixes the alignment in Safari and other browsers.

screen shot 2017-10-06 at 9 34 06 am


U-DON (changes requested):
Note that this breaks on mobile/smaller screens with the potential for a number to overlap the "Next". Ignore that the page numbers are all 3; I just kept appending to `va-pagination-inner` in the DOM navigator until I got this result.

> screen shot 2017-10-05 at 1 44 07 pm

jbalboni:
@U-DON Updated now. I added a screenshot of what it looks like with 30 pages at a small screen size above.

U-DON (approved):
Looks good! Yeah, we really should've used flexbox to begin with.

cehsu (approved):
LGTM. It looks like there are some slight spacing differences on different devices (numbers and "next" grouped together in the center vs. "next" being right aligned), but whether that deserves any additional attention is an open question, and if so would make sense to create a different ticket for.

PR #6553, opened 2017-10-05
Change HCA dependent page field names from child to depenent

jbalboni:
Renames fields that reference "child" or "children" to remove that or change it to "dependent/s." Also includes a migration and a change to some text that we missed before:

screen shot 2017-10-05 at 11 51 18 am


peggygannon (commented):
@jbalboni Hi, there. I made two content edits. I also have a question about this sentence: Include your wages, bonuses, tips, severance pay and other accrued benefits and your dependent's income information if it could have been used to pay your household expenses. 

As it's currently written, it's a little wishy-washy about when to include dependent's income Can we remove "could have been" and replace it with "was"
" . . .and your dependent's income information if it was used to pay your household expenses."

cehsu (approved):
LGTM. Had one question about a comment. It looks like the replacement also needs to be made on line 39 in test/hca/schema/schema.unit.spec.js as well.

rlbaker (approved):
LGTM. When would it be appropriate to remove the fallback support of the old fields from the schema and the backend?

jbalboni:
@peggygannon That "could have been" change seems ok to me, but I'm not sure if we need some approval for it. Can you make a separate ticket for it?

jbalboni:
@rlbaker Once it's merged, I think we should wait a few days to make sure anyone with the form sitting open from before the prod release has time to complete it.

PR #6557, opened 2017-10-05
Update spinner to show when claims or appeals are loading

jbalboni:
These changes make it so that the loading spinner shows up when at least one of claims or appeals are still loading, but doesn't hide any of the claims/appeals that have been loaded.

screen shot 2017-10-05 at 5 19 58 pm

### DIFF
--- a/src/js/claims-status/containers/YourClaimsPage.jsx
+++ b/src/js/claims-status/containers/YourClaimsPage.jsx
@@ -54,7 +54,7 @@ class YourClaimsPage extends React.Component {
       this.props.getAppeals(this.getFilter(this.props));
     }
 
-    if (this.props.loading) {
+    if (this.props.claimsLoading && this.props.appealsLoading) {
       scrollToTop();
     } else {
       setUpPage();
@@ -90,9 +90,9 @@ class YourClaimsPage extends React.Component {
   }
 
   renderErrorMessages() {
-    const { loading, appealsAvailable, canAccessAppeals, canAccessClaims, claimsAvailable, claimsAuthorized } = this.props;
+    const { claimsLoading, appealsLoading, appealsAvailable, canAccessAppeals, canAccessClaims, claimsAvailable, claimsAuthorized } = this.props;
 
-    if (loading) {
+    if (claimsLoading && appealsLoading) {
       return null;
     }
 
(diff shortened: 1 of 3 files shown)


stacyhlinn (approved):

src/js/claims-status/containers/YourClaimsPage.jsx
```diff
@@ -118,27 +118,49 @@ class YourClaimsPage extends React.Component {
...
+                <LoadingIndicator message="Loading your claims and appeals..."/>
+                <br/>
+              </div>
+            }
             {list.map(claim => this.renderListItem(claim))}
             <Pagination page={page} pages={pages} onPageSelect={this.changePage}/>
           </div>
         </div>);
-      } else if (!this.props.canAccessClaims) {
+      } else if (!this.props.canAccessClaims && bothRequestsLoaded) {
```
cehsu: Is it possible to have appeals if `!this.props.canAccessClaims`?

src/js/claims-status/containers/YourClaimsPage.jsx
```diff
@@ -118,27 +118,49 @@ class YourClaimsPage extends React.Component {
...
     let innerContent;
+    const bothRequestsLoaded = !claimsLoading && !appealsLoading;
+    const bothRequestsLoading = claimsLoading && appealsLoading;
+    const atLeastOneRequestLoading = claimsLoading || appealsLoading;
+    const emptyList = !list || !list.length;
 
-    if (loading) {
-      content = <LoadingIndicator message="Loading a list of your claims and appeals..." setFocus/>;
+    if (bothRequestsLoading || (atLeastOneRequestLoading && emptyList)) {
+      content = <LoadingIndicator message="Loading your claims and appeals..." setFocus/>;
```
cehsu: Does `setFocus` need to be set to true?

cehsu (approved):
LGTM. Had two minor questions.

src/js/claims-status/containers/YourClaimsPage.jsx
bshyong: I believe it's definitely possible, but in this conditional the first statement on top should catch it (having appeals would mean that the list is not empty).

bshyong (approved):

src/js/claims-status/containers/YourClaimsPage.jsx
jbalboni: I don't think so, since it's possible some claims or appeals are visible and the user is interacting with them already. I don't want to steal focus in that case.

PR #6558, opened 2017-10-05
Resolve focus issue in signin overlay

bshyong:
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4960


src/js/login/containers/Main.jsx
```diff
@@ -171,7 +171,7 @@ class Main extends React.Component {
         return (
           <div>
             <SearchHelpSignIn onUserLogout={this.handleLogout}/>
-            <Modal cssClass="va-modal-large" visible={this.props.login.showModal} onClose={this.handleCloseModal} id="signin-signup-modal" title="Sign in to Vets.gov">
+            <Modal cssClass="va-modal-large" visible={this.props.login.showModal}           focusSelector="button" onClose={this.handleCloseModal} id="signin-signup-modal" title="Sign in to Vets.gov">
```
U-DON: With this many props, we should separate each one to a new line.

U-DON (approved):
LGTM. But I'm not sure if you're merging into the right branch here.

U-DON:
Actually, none of the MHV changes are in here, so should be good. Only minor thing is that the formatting on that line is a bit wonky.

PR #6559, opened 2017-10-06
AddressSection unit tests

cvalarida:
So...this is possibly the most complex component of the whole app and I tried to test all the important things, but if I missed anything that you can think of, please point it out! There are already a ton of tests in here, but it's important for our sanity later to test all the things now.

I had to do some fancy footwork with the validations, but I hope my little comment sufficiently explained _why_ it was necessary. If not, please let me know and I'll draft up another one to be clearer. Also, if you think it's a terrible idea to use a static property on `AddressSection` to hold the validations, please let me know that too! I wasn't too keen on the idea, but I couldn't think of any other way to get it working. I'd be happy to hear your reasoning. 😄


test/letters/components/Address.unit.spec.jsx
```diff
@@ -142,4 +142,46 @@ describe('<Address>', () => {
...
     });
   });
+
+  it('should show state name for domestic addresses', () => {
+    // default is domestic
+    const component = ReactTestUtils.renderIntoDocument(<Address {...defaultProps}/>);
+    const tree = getFormDOM(component);
+
+    // Get the text that's actually displaying for the selected option
+    const stateName = tree.getElement('[name="state"]').selectedOptions[0].textContent;
```
markgreenburg: Talk about traversing the DOM...

test/letters/components/Address.unit.spec.jsx
```diff
@@ -142,4 +142,46 @@ describe('<Address>', () => {
...
+    //  the state and zip fields.
+    props.address.countryName = 'Elsweyre';
+
+    const component = ReactTestUtils.renderIntoDocument(<Address {...props}/>);
+    const tree = getFormDOM(component);
+
+    // Make sure we have as many fields as we expect...
+    expect(tree.querySelectorAll('input, select').length).to.equal(5);
+
+    // ...and that they don't include what they shouldn't
```
markgreenburg: Hah

test/letters/components/Address.unit.spec.jsx
```diff
@@ -142,4 +142,46 @@ describe('<Address>', () => {
...
+    // When the element isn't found, it'll throw an error
+    expect(() => tree.getElement('[name="state"]')).to.throw();
+    expect(() => tree.getElement('[name="postalCode"]')).to.throw();
```
markgreenburg: Nice, seems as though these two pair with the length check above for super-comprehensiveness :)

test/letters/containers/AddressSection.unit.spec.jsx
```diff
@@ -3,17 +3,35 @@ import ReactTestUtils from 'react-dom/test-utils';
...
 import sinon from 'sinon';
+import { cloneDeep } from 'lodash';
 
 import { getFormDOM } from '../../util/schemaform-utils';
 import { AddressSection } from '../../../src/js/letters/containers/AddressSection';
 import { ADDRESS_TYPES } from '../../../src/js/letters/utils/constants';
 
+const originalValidations = AddressSection.fieldValidations;
+const spyOnValidators = () => {
+  const validatorSpies = Object.keys(AddressSection.fieldValidations).reduce((spies, key) => {
```
markgreenburg: Nice, I like this. Glad this approach finally worked :)

test/letters/containers/AddressSection.unit.spec.jsx
```diff
@@ -3,17 +3,35 @@ import ReactTestUtils from 'react-dom/test-utils';
...
+    // Each property is an array of functions; wrap each function in a spy
+    return Object.assign({}, spies, { [key]: AddressSection.fieldValidations[key].map((validator) => sinon.spy(validator)) });
+  }, {});
+
+  // Infiltrate!
+  AddressSection.fieldValidations = validatorSpies;
```
markgreenburg: Whatever works, right?

test/letters/containers/AddressSection.unit.spec.jsx
```diff
@@ -141,4 +159,156 @@ describe('<AddressSection>', () => {
...
+
+    const instance = tree.getMountedInstance();
+    expect(instance.state.editableAddress).to.equal(props.savedAddress);
+
+    const newProps = Object.assign({}, props, { savedAddress: { addressOne: '123 Main St' } });
+    instance.componentWillReceiveProps(newProps);
+    expect(instance.state.editableAddress).to.equal(newProps.savedAddress);
+  });
+
+  it('should not call saveAddress when Cancel is clicked', () => {
```
markgreenburg: Do you think we need this test? I guess you could say that we're proactively preventing someone from breaking things in the future if they don't understand the difference between saved & editable addresses and decide to always call saveAddress?

test/letters/containers/AddressSection.unit.spec.jsx
```diff
@@ -141,4 +159,156 @@ describe('<AddressSection>', () => {
...
+  });
+
+  // Not sure how to test this bit yet...
+  // it('should scroll to first error', () => {});
+
+  describe('validation', () => {
+    // Spy on all the validation functions!
+    beforeEach(spyOnValidators);
+
+    // Clean up the spies so nobody finds out
```
markgreenburg: shhh

test/letters/containers/AddressSection.unit.spec.jsx
```diff
@@ -141,4 +159,156 @@ describe('<AddressSection>', () => {
...
+
+    it('should return the first error message it finds', () => {
+      const tree = SkinDeep.shallowRender(<AddressSection {...defaultProps}/>);
+      const instance = tree.getMountedInstance();
+
+      instance.handleChange('city', '');
+      // The required validator (first in the list) should return an error message and no other validators should run
+      expect(AddressSection.fieldValidations.city[0].called).to.be.true;
+      expect(AddressSection.fieldValidations.city.every(validator => validator.called)).to.be.false;
+    });
```
markgreenburg: Since we're only expecting the first validator to be called, should this check just look to make sure that `AddressSection.fieldValidations.city[1]` wasn't called?

Alternatively, we could expect that every validator except the first was *not* called. That might be more comprehensive?

markgreenburg (approved):
Awesome job! A few comments and one suggestion but nothing worthy of blocking a merge so go for it. Appreciate you figuring out comprehensive unit tests for all the validation craziness!

test/letters/containers/AddressSection.unit.spec.jsx
cvalarida: Yeah, basically. Just making sure that the button does what we expect it to for future-proofing. :man_shrugging:

Maybe overkill, but I like being thorough.

PR #6560, opened 2017-10-06
Remove the hurricane banner

ncksllvn:
See the issue below for the full discussion:
> can we take down the banner. Both the va.gov and mhv.gov sites have removed the emergency banners and have "hurrican updates" as lower leverl links on their sites

Resolves department-of-veterans-affairs/vets.gov-team/issues/4920

### DIFF
--- a/content/pages/index.md
+++ b/content/pages/index.md
@@ -3,7 +3,7 @@ layout: home.html
 body_class: home
 title: Home
 plainlanguage: 11-1-16 Ready for Beth review
-enablewarning: true
+enablewarning: false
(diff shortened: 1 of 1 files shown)


content/pages/index.md
```diff
@@ -3,7 +3,7 @@ layout: home.html
 body_class: home
 title: Home
 plainlanguage: 11-1-16 Ready for Beth review
-enablewarning: true
+enablewarning: false
```
ncksllvn: Instead of deleting the banner's markup and style, I figured that the `enablewarning` boolean is there for a reason and that we can save the banner in case something else comes up and we need it again.

bshyong (approved):
LGTM

PR #6561, opened 2017-10-06
Refactor claim status styles for status and detail tabs

jbalboni:
This updates the styles for the status and detail tab on the claim detail page. It also updates some app-wide styles when necessary.

screen shot 2017-10-06 at 1 24 53 pm
screen shot 2017-10-06 at 1 25 03 pm

### DIFF
--- a/src/js/claims-status/components/AddingDetails.jsx
+++ b/src/js/claims-status/components/AddingDetails.jsx
@@ -3,8 +3,8 @@ import React from 'react';
 class AddingDetails extends React.Component {
   render() {
     return (
-      <div className="usa-alert usa-alert-info claims-no-icon claims-alert-status">
-        <h4>We’re adding your details</h4>
+      <div className="usa-alert usa-alert-info no-background-image claims-alert-status">
+        <h4 className="claims-alert-header">We’re adding your details</h4>
         We’ve received your claim and are still adding some of your information. Check back soon to see the complete details of your claim.
       </div>
     );
--- a/src/js/claims-status/components/ClaimPhase.jsx
+++ b/src/js/claims-status/components/ClaimPhase.jsx
@@ -44,28 +44,28 @@ export default class ClaimPhase extends React.Component {
     switch (event.type) {
       case 'phase_entered':
         return (
-          <div className="claims-evidence-item columns usa-width-three-fourths medium-9">
+          <div className="claims-evidence-item">
             Your claim moved to {getUserPhaseDescription(this.props.phase)}
           </div>
         );
 
       case 'filed':
-        return <div className="claims-evidence-item columns usa-width-three-fourths medium-9">Thank you. VA received your claim</div>;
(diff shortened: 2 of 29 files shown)


src/js/claims-status/components/ClaimPhase.jsx
```diff
@@ -44,28 +44,28 @@ export default class ClaimPhase extends React.Component {
     switch (event.type) {
       case 'phase_entered':
         return (
-          <div className="claims-evidence-item columns usa-width-three-fourths medium-9">
+          <div className="claims-evidence-item">
```
cehsu: Is `claims-evidence-item` being used?

cehsu (approved):
LGTM. These styles are much easier to navigate now.

src/js/claims-status/components/AddingDetails.jsx
```diff
@@ -3,8 +3,8 @@ import React from 'react';
 class AddingDetails extends React.Component {
   render() {
     return (
-      <div className="usa-alert usa-alert-info claims-no-icon claims-alert-status">
-        <h4>We’re adding your details</h4>
+      <div className="usa-alert usa-alert-info no-background-image claims-alert-status">
```
cehsu: The original issue recommends using single class selectors as much as possible. Should we consider using `@extend` selectively to achieve this?

src/sass/claims-status.scss
```diff
@@ -138,80 +133,13 @@
...
-    display: inline-block;
-    font-weight: bold;
-    margin-right: 10px;
-  }
-
-}
-
-.claim-decision-is-ready,
 .you-have-no-claims {
   h4, p {
```
cehsu: Is this style being used (`.you-have-no-claims`)? If not, it would be a good opportunity to remove some additional tag name selectors.

src/js/claims-status/components/AddingDetails.jsx
jbalboni: We generally shouldn't use `@extend`: https://github.com/department-of-veterans-affairs/vets-website/blob/master/docs/CSSandSCSSBestPractices.md#dont-use-extend

Also, to be clear, I don't mean single classes in markup, I mean that when you write a rule in a scss file, we should try to avoid complicated selectors.

src/js/claims-status/components/ClaimPhase.jsx
jbalboni: Looks like not, aside from tests. I'll remove it.

src/sass/claims-status.scss
jbalboni: I think it's used on the claim list page, if you don't have any claims.

src/js/claims-status/components/ClaimPhase.jsx
jbalboni: I changed my mind on this, the markup here makes writing the tests without this class a pain, so I'm going to leave it.

PR #6562, opened 2017-10-06
Pre-need intro page.

U-DON:
Resolves department-of-veterans-affairs/vets.gov-team#3878.

Some of the spacing between subway map items and the lists and paragraphs within them is wonky. I think it's a problem across all form intro pages, and should be styled more generally, so I won't be fixing those issues here for this form individually.


bshyong (approved):
LGTM

PR #6565, opened 2017-10-06
Remove unused files and general cleaning up

ncksllvn:
This PR does the following:

- Removes unreferenced files from the content/includes directory
- Removes unreferenced files from the content/layouts directory
- Remove the "sticky action box" and all related code
- Removes references to the "extraspace" variable (after removing the above files there was only one left - content/pages/burials-and-memorials/index.md)

I determined what files could be removed by looking at top-level content pages, removing files that aren't referenced, and restarting the server to ensure it builds correctly. As for the "sticky action box", I don't believe that exists anymore in the markup so could be safely removed but please double check me in case I was misunderstanding what was being asked in the corresponding issue below.

Resolves department-of-veterans-affairs/vets.gov-team/issues/2157


phiden (approved):
Just to make sure -- none of these affect the Playbook? If no, then lgtm!

content/layouts/page-breadcrumbs.html
```diff
@@ -17,12 +17,6 @@
     {% when '3-panel-filters' %}
```
melwoodard: Since the `3-panel-filters` include is being deleted, can this be removed from the breadcrumbs too?

content/layouts/page-breadcrumbs.html
```diff
@@ -17,12 +17,6 @@
...
        {% include "content/includes/3-panel-filters.html" %}
 
-    {% when '4-action-page' %}
-      {% include "content/includes/4-action-page.html" %}
-
-    {% when '4-action-page-pending' %}
-      {% include "content/includes/4-action-page-pending.html" %}
-
     {% when '7-get-page' or '7-get-page-alt' %}
```
melwoodard: Since the two `7-get-page` includes are being deleted, can they be removed from the breadcrumbs too?

melwoodard (commented):
Added a couple comments to lines that look like they can be removed.

Another include that doesn't appear to be in use anymore is `/content/includes/general-related-links.html`
in which case that file and lines 27-29 in `/layouts/page-breadcrumbs.html` could also go away.

content/layouts/page-breadcrumbs.html
ncksllvn: Good call!

content/layouts/page-breadcrumbs.html
ncksllvn: Good call!

ncksllvn:
@melwoodard `general-related-links.html` was still referenced by the [breadcrumbs layout](https://github.com/department-of-veterans-affairs/vets-website/blob/master/content/layouts/page-breadcrumbs.html#L34) but it never does anything. The only time that condition is true is on the Privacy page and then it doesn't even print anything. So yep, good call on that one too!

PR #6567, opened 2017-10-06
Ensure attributes for identity proofed accounts are fetched

bshyong:
Resolves department-of-veterans-affairs/vets.gov-team#5267

This PR ensures that all users who are identity proofed "loa_highest" are automatically leveled up on signin.

### DIFF
--- a/src/js/login/containers/Main.jsx
+++ b/src/js/login/containers/Main.jsx
@@ -128,7 +128,7 @@ class Main extends React.Component {
   }
 
   handleLogin(loginUrl = 'idme') {
-    this.loginUrlRequest = handleLogin(this.props.login.loginUrls[loginUrl], this.props.onUpdateLoginUrl);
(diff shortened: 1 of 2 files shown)


src/js/login/containers/Main.jsx
```diff
@@ -128,7 +128,7 @@ class Main extends React.Component {
   }
 
   handleLogin(loginUrl = 'idme') {
-    this.loginUrlRequest = handleLogin(this.props.login.loginUrls[loginUrl], this.props.onUpdateLoginUrl);
+    this.loginUrlRequest = handleLogin(this.props.login.loginUrls[loginUrl], this.props.updateLogInUrls);
```
U-DON: Hm, looking this over, I'm not sure this second argument really serves any purpose... `handleLogin`, if given the `loginUrl`, ends up doing `getLoginUrls`, but presumably we already have that in order to pass it into the function.

U-DON (approved):
LGTM.

PR #6570, opened 2017-10-08
Claim status styles list ceh

cehsu:
This updates the styles for the claims list page. It also updates some app-wide styles when necessary. There are no visual changes. This branch is based on the `claim-status-styles-detail-jsb` branch in order to prevent merge conflicts.


src/sass/claims-status.scss
```diff
@@ -265,22 +193,18 @@ a.claim-list-item {
...
 .claims-status {
-  padding: 1em 0 3em 0;
+  padding: 1em 0 3em;
 }
 
-.your-claims, .claims-status {
+.claims-container, .claims-status {
   padding-bottom: 3em;
 
   h3 {
```
cehsu: It doesn't look like this style is ever used, should it be removed?

src/sass/claims-status.scss
```diff
@@ -265,22 +193,18 @@ a.claim-list-item {
 }
 
 .claims-status {
-  padding: 1em 0 3em 0;
+  padding: 1em 0 3em;
 }
 
-.your-claims, .claims-status {
+.claims-container, .claims-status {
```
cehsu: When appeals status is cleaned, would it make sense to consolidate claims-status and claims-container (formerly your-claims) as claims-container?

src/sass/claims-status.scss
```diff
@@ -265,22 +193,18 @@ a.claim-list-item {
...
   padding-bottom: 3em;
 
   h3 {
     margin-top: 1em;
   }
 
-  h1:first-child {
+  h1:first-child, .claims-container-title {
```
cehsu: When appeals status is cleaned, would it make sense to consolidate h1:first-child and claims-container-title as claims-container-title?

jbalboni: That seems like it would make sense.

jbalboni (commented):
It looks like there are more styles to update for the list page:

`.you-have-no-claims`
`.claim-list-item-container`
`.your-claims, .claims-status`
`.claims-list-sort`
`.claims-list-alert` (maybe, this looks very similar to the other alert styles in the file)

src/sass/claims-status.scss
```diff
@@ -133,46 +133,41 @@
   }
 }
 
-.you-have-no-claims {
```
cehsu: This style doesn't appear to be used.

src/sass/claims-status.scss
```diff
@@ -218,127 +209,121 @@ a.claim-list-item {
...
+}
 
-  .last-status {
-    margin-top: 1em;
-    padding: 0 1em 1em 0;
+.next-action-text {
+  margin-top: 0;
+}
 
-    h4:first-of-type {
```
cehsu: This style isn't used.

cehsu:
Thanks @jbalboni, I've updated the additional styles you pointed out.

src/sass/claims-status.scss
```diff
@@ -218,127 +208,121 @@ a.claim-list-item {
...
-      padding-left: 3.25rem;
+.content {
+  position: relative;
+  padding-left: 3.25rem;
+}
 
-      h5 + p {
-        margin-top: 0;
-      }
+h5 + .content-text {
```
jbalboni: I don't think this is actually used. The h5 has date content between it and the content

src/js/claims-status/utils/appeal-helpers.jsx
```diff
@@ -4,35 +4,35 @@ import moment from 'moment';
...
   },
   central_office: { // eslint-disable-line camelcase
     title: 'You asked for a Board hearing in Washington, DC',
-    description: <p>You’ll get a letter in the mail at least 30 days before your hearing is scheduled. It will let you know the date and time of the hearing. The Board’s central office hearings take place at 425 Eye Street, NW, Washington, DC, 20001.</p>
+    description: <p className="next-action-text">You’ll get a letter in the mail at least 30 days before your hearing is scheduled. It will let you know the date and time of the hearing. The Board’s central office hearings take place at 425 Eye Street, NW, Washington, DC, 20001.</p>
   },
 };
 
-export function appealStatusDescriptions(lastEvent, previousHistory = []) {
+export function appealStatusDescriptions(lastEvent, previousHistory = [], style) {
```
jbalboni: I don't think having this style parameter is necessary. The classes passed into it seem to always try to make the first `p` element in a group have a top margin of 0. Can we create a class that sets the top margin to 0 directly, and add it to the first `p` element that needs it in each description? Will that cause any problems you can think of?

src/js/claims-status/components/ClaimsListItem.jsx
```diff
@@ -18,17 +18,17 @@ export default function ClaimsListItem({ claim }) {
         <p className="status"><span className="claim-item-label">Status:</span> {listPhase(claim.attributes.phase)}</p>
         <div className="communications">
           {inProgress && claim.attributes.developmentLetterSent
-            ? <p><i className="fa fa-envelope"></i>We sent you a development letter</p>
+            ? <p className="claim-list-item-text"><i className="fa fa-envelope claim-list-item-icon"></i>We sent you a development letter</p>
```
jbalboni: I wonder if these should even be `p` elements? This is a list of items, so might be worth just changing the markup to a list

src/sass/claims-status.scss
```diff
@@ -628,41 +612,45 @@ h1:focus {
...
   }
-  > .usa-alert-body {
-    padding-left: 3.7rem;
-  }
-  .va-alert-close .fa {
-    font-size: 1.1em !important;
-  }
+}
+
+.claims-list-alert-body {
```
jbalboni: I think you could get away with not changing this (i.e. not creating a class for it and leaving it under `.claims-list-alert` like before). For reusable widgets and patterns like alerts and the subway map, I think it's ok to have one level of nesting in the styles if we just need to override a small portion of the default styling.

jbalboni (commented):
Looks good so far, had a few comments.

src/sass/claims-status.scss
cehsu: Good point, plus this should be moved into the status directly, rather than assigned at invocation, as you mentioned elsewhere.

src/sass/claims-status.scss
cehsu: Thanks @jbalboni, good point, reverted

src/sass/claims-status.scss
cehsu: There was one instance where this was used (for the `bva_final_decision` status description), but it can be specified directly with the new `description-text` (margin-top: 0) class.

src/js/claims-status/components/AppealListItem.jsx
```diff
@@ -10,7 +10,7 @@ const renderNextAction = (lastEvent, previousHistory) => {
   if (lastEvent.type === 'ssoc' && previousHistory[0].type !== 'soc') {
     return null;
   }
-  const nextAction = appealStatusDescriptions(lastEvent, previousHistory).nextAction;
+  const nextAction = appealStatusDescriptions(lastEvent, previousHistory, 'claim-list-item-text').nextAction;
```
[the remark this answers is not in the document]
cehsu: Thanks for pointing this out, change made.

src/js/claims-status/components/AppealListItem.jsx
```diff
@@ -26,7 +26,7 @@ export default function AppealListItem({ appeal }) {
       <div className="last-changed-label"><strong>Last changed: </strong> {moment(lastEvent.date).format('MMM D, YYYY')}</div>
       <Link className="claim-list-item" to={`appeals/${appeal.id}/status`}>
         <h4 className="claim-list-item-header">Compensation Appeal – Received {moment(firstEvent.date).format('MMM D, YYYY')}</h4>
-        <p className="status"><span className="claim-item-label">Status:</span> {appealStatusDescriptions(lastEvent, previousHistory).status.title}</p>
+        <p className="status"><span className="claim-item-label">Status:</span> {appealStatusDescriptions(lastEvent, previousHistory, 'claim-list-item-text').status.title}</p>
```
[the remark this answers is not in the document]
cehsu: Thanks for pointing this out, change made.

src/js/claims-status/containers/AppealStatusPage.jsx
```diff
@@ -25,7 +25,7 @@ class AppealStatusPage extends React.Component {
   }
 
   renderStatusNextAction(lastEvent, previousHistory) {
-    const { nextAction } = appealStatusDescriptions(lastEvent, previousHistory);
+    const { nextAction } = appealStatusDescriptions(lastEvent, previousHistory, 'next-action-text');
```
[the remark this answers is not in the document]
cehsu: Good point, a new class `description-text` has been created and included directly.

src/js/claims-status/components/ClaimsListItem.jsx
cehsu: Good point, change made.

src/js/claims-status/utils/appeal-helpers.jsx
```diff
@@ -4,15 +4,15 @@ import moment from 'moment';
 export const hearingDescriptions = {
   video: {
     title: 'You have asked for a videoconference hearing.',
-    description: <p>You will get a letter in the mail at least 30 days before your hearing is scheduled letting you know the date, time, and location of the hearing.</p>
+    description: <p className="description-text">You will get a letter in the mail at least 30 days before your hearing is scheduled letting you know the date, time, and location of the hearing.</p>
```
jbalboni: Should this class be named something different? It's really for the first paragraph in a section, maybe `claims-first-paragraph`?

jbalboni (approved):
LGTM. I would suggest taking another pass through the app to make sure everything looks right before merging. Not doing that for style refactors always seems to come back to bite me.

src/js/claims-status/utils/appeal-helpers.jsx
cehsu: Thanks for the recommendation, that's much clearer and easier to reason about.

cehsu:
Good point, @jbalboni, a visual check yielded the following difference in last status presentation:

screen shot 2017-10-11 at 4 07 05 pm

As the original styles relied on the user agent stylesheet to add the following for `p`:
```
    -webkit-margin-before: 1em;
    -webkit-margin-after: 1em;
```
Therefore the programmatic assignment of styles has been reintroduced. Please let me know if another way to go about this might also work.

jbalboni:
@cehsu I think that's actually a bug in the current code. It looks like the intent was to have that paragraph not have a top margin below an h5, and that's the only place where that would occur, but there's a date in between it. I'm trying to get the mock ups to confirm that.

If that's not the case, though, I think the easier solution would be to add a class to the date to add a margin underneath it, rather than programmatically setting the class.

src/sass/claims-status.scss
```diff
@@ -218,128 +208,118 @@ a.claim-list-item {
...
+}
 
-    .content {
-      position: relative;
-      padding-left: 3.25rem;
+.last-status-text {
+  h5 + & {
+    margin-top: 0;
+  }
+}
```
cehsu: Thanks @jbalboni, unless the mocks indicate that there should be a margin between h5 and p (such as in the `bva_final_decision` case in `appealStatusDescriptions`), how about we play it safe and simply add both cases where `margin-top` should be zero (all previous activity `p`, and when there is an instance of `h5` + `p` as described above in last status) as rules to `claims-first-paragraph`, so:
```
.claims-first-paragraph {
  .h4 + &, h5 + & {
    margin-top: 0;
  }
}
```

cehsu:
Thanks @jbalboni, additional recommendations integrated, as well as another bug fix. Screen shots below:
screen shot 2017-10-12 at 11 01 27 am

screen shot 2017-10-12 at 10 57 40 am

PR #6574, opened 2017-10-09
Switch edu md pages to upper case, handle return urls safely

jbalboni:
Two things:

1. We need to use upper case for 1990E and 1990N md pages, since we're redirecting to the upper case version and staging/prod is case sensitive.
2. It looks like we can get both return_url and returnUrl in saved form data. returnUrl is the format the front end uses consistently, so this will switch to that format in the migration.


patrickvinograd (approved):
LGTM.

PR #6575, opened 2017-10-09
Standardize loading message design

ncksllvn:
- [x] Messaging - While a message is sending, switch to an inline loader instead of full-screen.
- [x] RX - Same as above but for sending prescriptions instead.
- [x] Auth, landing pages - Switch from miscellaneous loaders to the standards Vets.gov circle loader

Resolves department-of-veterans-affairs/vets.gov-team/issues/4449

### DIFF
--- a/content/pages/auth/login/callback/index.md
+++ b/content/pages/auth/login/callback/index.md
@@ -8,8 +8,9 @@ entryname: auth
   <div class="section">
     <div id="react-root">
       <div class="loading-message">
-        <h3>Please wait while we log you in.</h3>
-        <img src="/img/preloader-primary-darkest.gif" alt="Loading">
+        <div class="loading-indicator-container">
+          <div class="loading-indicator" role="progressbar" aria-valuetext="lease wait while we log you in." tabIndex="0"></div> Please wait while we log you in.
+        </div>
       </div>
     </div>
   </div>
--- a/content/pages/burials-and-memorials/application/530.md
+++ b/content/pages/burials-and-memorials/application/530.md
@@ -15,8 +15,9 @@ layout: page-react.html
   <div class="section">
     <div id="react-root">
       <div class="loading-message">
-        <h3>Please wait while we load the application for you.</h3>
-        <img src="/img/preloader-primary-darkest.gif" alt="Loading">
+        <div class="loading-indicator-container">
+          <div class="loading-indicator" role="progressbar" aria-valuetext="Please wait while we load the application for you." tabIndex="0"></div> Please wait while we load the application for you.
+        </div>
       </div>
     </div>
   </div>
--- a/src/js/messaging/components/compose/MessageSend.jsx
+++ b/src/js/messaging/components/compose/MessageSend.jsx
@@ -62,7 +62,7 @@ class MessageSend extends React.Component {
 
   render() {
     const isDisabled = this.props.disabled;
-    const spinnerIcon = this.props.sendingMessage && <span className="fa fa-spinner fa-spin"/>;
+    const spinnerIcon = this.props.sendingMessage && <i className="fa fa-spinner fa-spin"/>;
 
     return (
       <div className="msg-send-group">
--- a/src/js/messaging/components/compose/MessageSend.jsx
+++ b/src/js/messaging/components/compose/MessageSend.jsx
@@ -62,6 +62,7 @@ class MessageSend extends React.Component {
 
   render() {
     const isDisabled = this.props.disabled;
+    const spinnerIcon = this.props.loading.sendingMessage ? <span className="fa fa-spinner fa-spin"/> : null;
 
     return (
       <div className="msg-send-group">
@@ -69,7 +70,7 @@ class MessageSend extends React.Component {
           <button
             disabled={isDisabled}
             type="button"
-            onClick={this.props.onSend}>Send</button>
+            onClick={this.props.onSend}>{spinnerIcon} Send</button>
           <button
             disabled={isDisabled}
             className="usa-button-outline msg-btn-save"
--- a/src/js/messaging/components/compose/MessageSend.jsx
+++ b/src/js/messaging/components/compose/MessageSend.jsx
@@ -62,7 +62,7 @@ class MessageSend extends React.Component {
 
   render() {
     const isDisabled = this.props.disabled;
-    const spinnerIcon = this.props.loading.sendingMessage ? <span className="fa fa-spinner fa-spin"/> : null;
+    const spinnerIcon = this.props.sendingMessage && <span className="fa fa-spinner fa-spin"/>;
 
     return (
       <div className="msg-send-group">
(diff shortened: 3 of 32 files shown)


ncksllvn:
In the original issue, I think the loading screen that says "You are now returning to..."  is actually ID.me, even though the address bar says staging

ncksllvn:
Just to summarize the scale of this update. I think these are actually really nice improvements -

1. Secure Message - after typing a message and clicking send, previously the editor would disappear and be replaced by a loader. Instead, after click, we disable the submit button and show a spinner in the button text.

2. Prescription Refill - same as above, except for requesting a prescription refill.

3. Various app landing pages - while waiting for React to boot up, many of our pages showed a header with "Please wait while we load the application". These have been replaced by our standard app loader.

phiden (approved):

phiden:
LGTM @ncksllvn -- nicely done!

leannammiller:
@eli-lewis we should have @ncksllvn and @phiden talk about this improvement during next demo day! what do you think?

src/js/messaging/containers/Compose.jsx
```diff
@@ -153,6 +153,7 @@ export class Compose extends React.Component {
           onSaveMessage={this.saveDraftIfNoAttachments}
           onSendMessage={this.sendMessage}
           onSubjectChange={this.props.setMessageField.bind(null, 'message.subject')}
+          sendingMessage={this.props.loading.sendingMessage}
```
ncksllvn: I decided to pass only the `sendingMessage` property instead of the whole `loading` object because that way if the boolean isn't passed it'll be falsy by default and we don't have to check for the existence of the `loading` object.

content/pages/auth/login/callback/index.md
```diff
@@ -8,8 +8,9 @@ entryname: auth
   <div class="section">
     <div id="react-root">
       <div class="loading-message">
-        <h3>Please wait while we log you in.</h3>
-        <img src="/img/preloader-primary-darkest.gif" alt="Loading">
+        <div class="loading-indicator-container">
+          <div class="loading-indicator" role="progressbar" aria-valuetext="lease wait while we log you in." tabIndex="0"></div> Please wait while we log you in.
+        </div>
```
ncksllvn: I would have much preferred to have a template and just use a Jekyll include statement, but I don't think our build is set up to allow include-statements from markdown files because I encountered some weird behavior trying that.

src/js/rx/components/ConfirmRefillModal.jsx
```diff
@@ -25,13 +24,10 @@ class ConfirmRefillModal extends React.Component {
     // Initialize to prevent console errors
     let innerElement = '';
 
-    if (this.props.isLoading) {
-      innerElement = (
-        <LoadingIndicator
-          message="Submitting your refill request..."/>
-      );
-    } else if (this.props.prescription) {
+    if (this.props.prescription) {
```
jbalboni: There's not a situation where isLoading is true and there's no prescription prop?

content/pages/burials-and-memorials/application/530.md
```diff
@@ -15,8 +15,9 @@ layout: page-react.html
   <div class="section">
     <div id="react-root">
       <div class="loading-message">
-        <h3>Please wait while we load the application for you.</h3>
-        <img src="/img/preloader-primary-darkest.gif" alt="Loading">
+        <div class="loading-indicator-container">
+          <div class="loading-indicator" role="progressbar" aria-valuetext="Please wait while we load the application for you." tabIndex="0"></div> Please wait while we load the application for you.
```
jbalboni: Something to note with this, the last time I checked the new spinner svg was twice the size of the older gif spinner. That could mean there's a longer delay for seeing the spinner itself.

jbalboni (approved):
LGTM with a couple minor comments, but I would have @U-DON look at the Rx/SM changes, too.

src/js/rx/components/ConfirmRefillModal.jsx
ncksllvn: Good question, this was just what I determined from looking at the logic around it. The previous loader's message was "submitting your refill request" so I would think not but I'll see if @U-DON can confirm.

content/pages/burials-and-memorials/application/530.md
ncksllvn: That makes sense. I think that may occur on the first request but not subsequent because it'll be in cache. If it becomes a problem at all maybe we can start preloading it with JS or somewhere on the page.

src/js/messaging/components/compose/MessageSend.jsx
```diff
@@ -62,14 +62,15 @@ class MessageSend extends React.Component {
...
+    const spinnerIcon = this.props.sendingMessage && <span className="fa fa-spinner fa-spin"/>;
 
     return (
       <div className="msg-send-group">
         <div className="msg-send-buttons">
           <button
             disabled={isDisabled}
             type="button"
-            onClick={this.props.onSend}>Send</button>
+            onClick={this.props.onSend}>{spinnerIcon} Send</button>
```
U-DON: Is this always going to have a space before the "Send" if it's not currently sending?

src/js/messaging/components/forms/NewMessageForm.jsx
```diff
@@ -65,6 +65,7 @@ export class NewMessageForm extends React.Component {
           allowedMimeTypes={allowedMimeTypes}
           errorMessage={validations.isValidMessageBody(message.body) ? undefined : composeMessage.errors.message}
           files={this.props.message.attachments}
+          loading={this.props.loading}
```
U-DON: I think you forgot to remove this, since you decided not to pass in the `loading` object anymore.

src/js/rx/components/ConfirmRefillModal.jsx
U-DON: I don't think it should occur, since `isLoading` becomes true only upon making a refill request. It does make me a little uneasy about what edge cases we might be missing. Like perhaps we shouldn't be showing a modal at all if there's no prescription currently being viewed.

src/js/messaging/components/forms/NewMessageForm.jsx
ncksllvn: Good call!

src/js/messaging/components/compose/MessageSend.jsx
ncksllvn: Good question, I know it doesn't in Chrome but I need to confirm for other browsers

ncksllvn: Just did a little digging - [leading spaces are trimmed](https://developer.mozilla.org/en-US/docs/Web/CSS/white-space) from HTML elements unless there is special style applied like in the `pre` tag or you use `nbsp;`. So we should be good here.

eli-lewis:
@leannammiller pls. remind me and/or @rroueche for next Tuesday's end of PI demo

PR #6576, opened 2017-10-09
Show loading spinner while savePending address updates

markgreenburg:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/5128 . 

Wires `savePending` state from redux store to `AddressSection`, then uses that to conditionally display the standard loading spinner in place of an address within `AddressSection` while the save is still pending.

**TO-DO:**
[x] Show loading spinner in place of old address while address update is in progress
[x] Add tests to ensure spinner shows when its supposed to and doesn't when it's not

Screenshot:
screen shot 2017-10-09 at 5 48 15 pm


cvalarida (approved):
Looks good! Haven't tested locally, but I'll trust your screenshot. 😉

PR #6580, opened 2017-10-10
Update claims filing message

cehsu:
This updates the claims filing message.
Please note the following screenshot has been created by hardcoding the changed value into the message variable on the `AdditionalEvidencePage`, like so:
```
      const trackedItem = {displayName: 'Request X'};
      const message = {
        title: 'We have your evidence',
        body: <span>Thank you for sending us {trackedItem ? trackedItem.displayName : 'additional evidence'}. We’ll let you know when we’ve reviewed it.<br/>Note: It may take a few minutes for your uploaded file to show here. If you don’t see your file, please try refreshing the page.</span>
      }

```

screen shot 2017-10-10 at 1 39 47 pm


src/js/claims-status/actions/index.js
```diff
@@ -221,7 +222,7 @@ export function submitFiles(claimId, trackedItem, files) {
               });
               dispatch(setNotification({
                 title: 'We have your evidence',
-                body: `Thank you for filing ${trackedItem ? trackedItem.displayName : 'additional evidence'}. We’ll let you know when we’ve reviewed it.`
+                body: <span>Thank you for sending us {trackedItem ? trackedItem.displayName : 'additional evidence'}. We’ll let you know when we’ve reviewed it.<br/>Note: It may take a few minutes for your uploaded file to show here. If you don’t see your file, please try refreshing the page.</span>
```
jbalboni: If we're putting jsx in here, we should change the file extension.

jbalboni (approved):
LGTM with one comment. A screenshot would be helpful, though you might have to use a review instance for that. It's hard to do locally (file uploads don't work in the mock api I gave you).

src/js/claims-status/actions/index.js
cehsu: Thanks @jbalboni, changed.

cehsu:
Thanks @jbalboni, I'll include a screen shot as well via review instance as you recommended.

PR #6581, opened 2017-10-10
Facility locator accessibility

cehsu:
These changes address the facility locator accessibility issues as well as the exit icon accessibility issues, which are also initially addressed by https://github.com/department-of-veterans-affairs/vets-website/pull/6863. Regarding the icon-related work, these additional changes have been determined and worked on together with @markgreenburg.

Including @bshyong and @cvalarida as they have worked on this component in the past and are making similar changes elsewhere on the site, respectively.

Including @stacyhlinn for feedback as well as this includes some changes to look and feel/design, and in case there are other visual changes that should be included here and as this work has identified the need for additional assessment as to whether this design requires updating (issue on this [here](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/6327)).

Including @peggygannon for feedback as well for visibility regarding handling of external links.

These changes address the following issues:

- Visual focus is provided to dropdown menus:

screen shot 2017-11-30 at 9 32 13 am

- Address edit field no longer an issue as entrance into field does not require additional input, and as enter retains focus on address field.

- Users can now select all facilities:
screen shot 2017-12-04 at 7 40 54 pm

References used:
- [Accessible Custom Select Dropdowns](http://www.webaxe.org/accessible-custom-select-dropdowns/)
- [selects](https://github.com/dequelabs/pattern-library/wiki/selects)


src/js/facility-locator/components/search-results/FacilityDirectionsLink.jsx
```diff
@@ -10,6 +10,7 @@ class FacilityDirectionsLink extends Component {
     return (
       <span>
         <a href={`https://maps.google.com?saddr=Current+Location&daddr=${addressArray.join(', ')}`} target="_blank">
+          <span className="usa-sr-only">External link</span>
```
cehsu: @markgreenburg is this consistent with the planned link accessibility approach?

cehsu: Thanks for investigating content solutions for external links @peggygannon, it looks like we will use the existing external link background image that applies for all external links (as confirmed by @stacyhlinn) and now include a hidden span for screen readers only, according to best practices:
- [Links to New Windows, Pop-ups, Other Frames, or External Web Sites](https://webaim.org/techniques/hypertext/hypertext_links#new_window)
- [Background images](https://webaim.org/techniques/alttext/#background)

src/js/facility-locator/components/SearchControls.jsx
```diff
@@ -18,11 +15,69 @@ class SearchControls extends Component {
...
+    this.navigateServiceDropdown = this.navigateServiceDropdown.bind(this);
     this.toggleFacilityDropdown = this.toggleFacilityDropdown.bind(this);
     this.toggleServiceDropdown = this.toggleServiceDropdown.bind(this);
-    this.handleFacilityFilterSelect = this.handleFacilityFilterSelect.bind(this);
+    this.setFacilityDropdownFocus = this.setFacilityDropdownFocus.bind(this);
+    this.setServiceDropdownFocus = this.setServiceDropdownFocus.bind(this);
+  }
+
+  setFacilityDropdownFocus(newState) {
+    const dropdownState = false ? !this.state.facilityDropdownActive : this.state.facilityDropdownActive;
```
jbalboni: Is this supposed to be `newState` instead of `false`? Also, are the focused and active states ever different?

src/js/facility-locator/components/SearchControls.jsx
```diff
@@ -129,11 +234,11 @@ class SearchControls extends Component {
...
     return (
-      <ul className="dropdown">
+      <ul className="dropdown" role="listbox" id="service-list">
         {
-          services.map(k => {
-            return (<li key={k} value={k} onClick={this.handleServiceFilterSelect.bind(this, k)}>
-              <button type="button" className="facility-option">
+          services.map((k, i) => {
+            return (<li role="option" aria-selected={k === facilityType} className={`${this.state.focusedServiceIndex === i ? 'is-hovered' : ''}`} key={k} value={k}>
+              <button tabIndex="-1" ref={ elem => { this.services[i] = elem; }} id={k} type="button" className="facility-option" onClick={this.handleServiceFilterSelect.bind(this, k)}>
```
jbalboni: Should this be a button? That seems sort of confusing inside a listbox option.

jbalboni (commented):
I took a pass through and made some comments. One difficulty I'm having in reviewing this is that there's a lot of code that's generic to handling a custom select widget, mixed in with some facility/service specific code (and also duplicated). What do you think about making a separate, generic component that could be used for both dropdowns?

src/js/facility-locator/components/search-results/FacilityDirectionsLink.jsx
markgreenburg: @cehsu I took a CSS-only approach due to 1.) not wanting to add spans for all the existing links and 2.) feeling that we may fail to consistently apply the correct markup going forward given that we don't currently have any checks / tests in place. Here's essentially what I came up with:
```

Comment (markgreenburg) [reply src/js/facility-locator/components/search-results/FacilityDirectionsLink.jsx]: Also for reference, here's the PR (still waiting on some input but should give you an idea of how I think I could solve for our external links specifically) https://github.com/department-of-veterans-affairs/vets-website/pull/6863

Also, note that I don't love this approach as a long-term solution - I think markup is the best way to go.

**Comment by cehsu** [comment]
Great idea, @jbalboni, I will create a separate select component as well as implement your other suggestions.

File: src/js/facility-locator/components/search-results/FacilityDirectionsLink.jsx (hunk 1 above)
Comment (cehsu) [reply src/js/facility-locator/components/search-results/FacilityDirectionsLink.jsx]: That makes sense and thanks for sharing this additional approach @markgreenburg.
It seems like this may contradict some guidance, but is a good initial step until we update all external links and links that open in new tabs:
>Sometimes image sprites or other background images are used to present content. While these should generally be avoided, if they are used, any content conveyed through the background image must be made accessible within the page markup. If the PDF icon image in Example 7 above were instead presented via a CSS background image, you could use a text replacement technique to present the content within the link:
<a href="application.pdf">Download the Employment Application<span class="pdficon"> (PDF)</span></a>
You would then use CSS to position the "(PDF)" text (the span class="pdficon" element) off-screen. See CSS in Action: Invisible Content Just for Screen Reader Users for more details. Sighted users would see the PDF background image and screen reader users would hear the off-screen "(PDF)" text.
Source: https://webaim.org/techniques/alttext/
One of the considerations here is that background images may be disabled in high-contrast display modes.

File: content/includes/common-and-popular.html [hunk 4]
```diff
@@ -20,7 +20,7 @@ <h3 class="va-h-ruled">Popular on vets.gov</h3>
       <li>
         <a href="/gi-bill-comparison-tool/">View education benefits available by school</a>
       </li>
-      <li><a target="_blank" href="https://www.veteranscrisisline.net/" class="external">Contact the Veterans Crisis Line</a></li>
+      <li><a target="_blank" href="https://www.veteranscrisisline.net/" class="external"><span class="usa-sr-only"></span>Contact the Veterans Crisis Line</a></li>
```
cehsu: @markgreenburg, the following regex might facilitate the transition if we have determined the appropriate pattern:
Search: `<a (.*?href="http.*?)(.*?)>(.*?)</a>`
Replace (Content): `<a $1$2><span class="usa-sr-only"></span>$3</a>`
Replace (**/*.jsx): `<a $1$2><span className="usa-sr-only"></span>$3</a>`

src/sass/facility-locator.scss
```diff
@@ -9,7 +9,13 @@
   .highlight {
     margin-bottom: 0.5em !important;
   }
-
+  .is-focused {
+    box-shadow: 0 0 3px #3e94cf, 0 0 7px #3e94cf;
```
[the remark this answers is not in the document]
cehsu: Good call, changed.

src/js/facility-locator/components/SearchControls.jsx
```diff
@@ -194,34 +300,35 @@ class SearchControls extends Component {
...
             <input ref="searchField" name="streetCityStateZip" type="text" onChange={this.handleQueryChange} value={currentQuery.searchString} aria-label="Street, City, State or Zip" title="Street, City, State or Zip"/>
           </div>
           <div className="columns usa-width-one-fourth medium-3">
             <label htmlFor="facilityType">Select Facility Type</label>
-            <div tabIndex="1" className={`facility-dropdown-wrapper ${facilityDropdownActive ? 'active' : ''}`} onClick={this.toggleFacilityDropdown}>
+            <div ref={ elem => {this.facilityDropdown = elem; }} onFocus={() => this.setFacilityDropdownFocus(true)} id="facility-list" onBlur={() => this.setFacilityDropdownFocus(false)} onKeyDown={this.toggleFacilityDropdown} tabIndex="0" className={`facility-dropdown-wrapper ${facilityDropdownFocused ? 'is-focused' : ''} ${facilityDropdownActive ? 'active' : ''}`} aria-autocomplete="none" aria-owns="facility-list" aria-controls="expandable" aria-expanded="false" aria-labelledby="facility-label" aria-required="false" aria-activedescendant={facilityType} role="combobox" onClick={this.toggleFacilityDropdown}>
```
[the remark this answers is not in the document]
cehsu: Good point, changed.

src/js/facility-locator/components/SearchControls.jsx
cehsu: Good point, in fact, this logic is no longer necessary due to the focus update you mentioned below, and it has been removed.

content/includes/common-and-popular.html
markgreenburg: That's a good idea. From what I've seen, some of the content already has the correct spans. One caveat I can see is that the `md` content pages may not work with this.

cehsu: Thanks @markgreenburg, I will remove duplicates. Could you please clarify what you anticipate the problem being with the `md` content pages?

content/includes/footer.html
```diff
@@ -4,23 +4,23 @@
...
-      <li><a href="https://www.vets.gov/pension/">Pension Benefits</a></li>
-      <li><a href="https://www.vets.gov/burials-and-memorials">Burials and Memorials</a></li>
+      <li><a href="https://www.vets.gov/disability-benefits/"><span class="usa-sr-only"></span>Disability Benefits</a></li>
+      <li><a href="https://www.vets.gov/health-care/"><span class="usa-sr-only"></span>Health Care Benefits</a></li>
+      <li><a href="https://www.vets.gov/education/"><span class="usa-sr-only"></span>Education Benefits</a></li>
+      <li><a href="https://www.vets.gov/housing-assistance/"><span class="usa-sr-only"></span>Housing Assistance</a></li>
+      <li><a href="https://www.vets.gov/employment/"><span class="usa-sr-only"></span>Careers and Employment</a></li>
+      <li><a href="https://www.vets.gov/life-insurance/"><span class="usa-sr-only"></span>Life Insurance</a></li>
+      <li><a href="https://www.vets.gov/pension/"><span class="usa-sr-only"></span>Pension Benefits</a></li>
+      <li><a href="https://www.vets.gov/burials-and-memorials"><span class="usa-sr-only"></span>Burials and Memorials</a></li>
```
cvalarida: Could you clarify why we're adding empty `<span>`s?

`.usa-sr-only` does the following (from `node_modules/uswds/dist/css/uswds.css` line 823):
```css
.usa-sr-only {
  position: absolute;
  left: -999em;
}
```
It looks like this is used for adding content that we want to be read off (and and interacted with) by screen readers but not shown visually. Is the plan to add "External link" to all of them? If so, @markgreenburg's [PR](https://github.com/department-of-veterans-affairs/vets-website/pull/6863) may already take care of all this.

content/includes/common-and-popular.html
markgreenburg: I'm still not sure how our markdown-based content pages are generated in the build, but if they're written in markdown which is then compiled to html, I'm assuming we'd need to either change how links are transpiled from md to html, or add the spans to the html output directly.  I may have a fundamental misunderstanding of how these content pages work, though.

cehsu: Thanks for clarifying @markgreenburg, I'll investigate this.

content/includes/footer.html
cehsu: You are correct @cvalarida, the plan is to add "External Link" here pending input on @markgreenburg's PR, where I have asked if that is the appropriate text. As for the correct approach, @markgreenburg and I are exploring that in [this](https://github.com/department-of-veterans-affairs/vets-website/pull/6581#discussion_r153851797) thread.

content/includes/common-and-popular.html
cehsu: @markgreenburg, it looks like the best way to make this consistent is to convert the external markdown links into html first.

content/includes/footer.html
cvalarida: Okay, cool. Just wanted to make sure. Thanks! 😄

src/js/facility-locator/components/search-results/FacilityDirectionsLink.jsx
cvalarida: @markgreenburg Why do you prefer markup over the css solution?

content/includes/common-and-popular.html
markgreenburg: Oh right, I forgot you can just inline html within markdown. That seems pretty reasonable to me.

cehsu: Sounds good @markgreenburg, thanks for the reminder about the markdown links.

bshyong:
Most of these changes look pretty straightforward, thanks for helping to clean this up for accessibility! I would make sure we test this in IE after the changes are all in—I remember that weird visual/styling issues always seemed to crop up in IE when we were making simple changes

cehsu:
Thanks @bshyong, a preliminary caniuse/doiuse scan identified a number of IE issues. I will further investigate these and manually test on IE.

src/js/facility-locator/components/Listbox.jsx
```diff
@@ -0,0 +1,167 @@
...
+  constructor() {
+    super();
+
+    this.state = {
+      facilityDropdownActive: false,
+      serviceDropdownActive: false,
+      focusedFacilityIndex: 0,
+      focusedServiceIndex: 0
+    };
+    this.options = [];
```
cehsu: State and options are a copy paste error and will be removed.

src/js/facility-locator/components/Select.jsx
```diff
@@ -0,0 +1,160 @@
...
+    const titleType = optionType[0].toUpperCase() + optionType.slice(1);
+    const dropdownClasses = classNames({
+      'facility-dropdown-wrapper': true,
+      active: this.state.dropdownActive,
+      disabled
+    });
+    return (
+      <div onBlur={this.handleBlur} className="columns usa-width-one-fourth medium-3">
+        <label htmlFor={`${optionType}Type`} id={`${optionType}-label`}>{`Select ${titleType} Type`}</label>
+        <button
```
jbalboni: When this button is disabled, we should set tabIndex to -1 so that it's taken out of the normal tab flow.

src/js/facility-locator/components/Select.jsx
```diff
@@ -0,0 +1,160 @@
...
+          onKeyDown={e => this.navigateDropdown(e, optionType)}
+          className={dropdownClasses}
+          ref={elem => { this.dropdown = elem; }}
+          aria-expanded="false"
+          aria-labelledby={`${optionType}-label`}
+          onClick={e => {e.preventDefault(); this.toggleDropdown(e);}}>
+          <Listbox
```
jbalboni: Is nesting these options inside of a button ok, a11y-wise? I'm curious how you ended up with that structure.

src/js/facility-locator/components/Select.jsx
```diff
@@ -0,0 +1,160 @@
...
+      this.setState({
+        focusedIndex: index
+      });
+    }
+  }
+
+  render() {
+    const { optionType, currentQuery, isMobile, hasIcons } = this.props;
+    const { facilityType, serviceType } = currentQuery;
+    const disabled = optionType === 'service' && !['benefits', 'vet_center'].includes(facilityType);
```
jbalboni: Could we move this into the facility locator component? I think it may make things clearer if we get as much of the optionType switching out of this component and passed in via props.

src/js/facility-locator/components/Select.jsx
```diff
@@ -0,0 +1,160 @@
...
+    let queryType = optionType === 'service' ? serviceType : facilityType;
```
jbalboni: Same comment as above

src/js/facility-locator/components/Select.jsx
```diff
@@ -0,0 +1,160 @@
...
+    this.state = {
+      dropdownActive: false,
+      focusedIndex: 0
+    };
+    this.options = [];
+  }
+
+  handleFilterSelect = (newOption, type) => {
+    const { currentQuery } = this.props;
+    const { facilityType, serviceType } = currentQuery; // eslint-disable-line no-unused-vars
```
jbalboni: Do we need this eslint exception?

src/js/facility-locator/components/Select.jsx
```diff
@@ -0,0 +1,160 @@
...
+      const increment = getDirection(e.keyCode);
+      const newIndex = this.state.focusedIndex + increment;
+      return this.focusOption(this.options[newIndex], newIndex, type);
+    }
+    return false;
+  }
+
+  handleBlur = (e) => {
+    const { currentTarget } = e;
+    setTimeout(() => {
```
jbalboni: The reason for this method isn't super clear to me. Why do we need to check the activeElement?

src/js/facility-locator/components/Select.jsx
```diff
@@ -0,0 +1,160 @@
...
+      }
+    });
+  }
+
+  toggleDropdown = (e) => {
+    const type = this.props.optionType;
+    const isActive = this.state.dropdownActive;
+    const { currentQuery: { facilityType, serviceType } } = this.props; // eslint-disable-line no-unused-vars
+    const queryType = type === 'service' ? serviceType : facilityType;
+    if (type === 'service' && noServices(type, facilityType)) {
```
jbalboni: If we put a disabled check in the onClick handler, could this check go away?

src/js/facility-locator/utils/helpers.js
```diff
@@ -0,0 +1,79 @@
...
+  if (code === keyMap.DOWN) return 1;
+  return false;
+}
+
+export function getSelection(options, queryType) {
+  /* eslint-disable no-param-reassign */
+  let selection = options[0];
+  if (queryType) {
+    const selectionType = queryType;
+    options = options.filter(item => !!item);
```
jbalboni: Do you need to overwrite options? Is that something the callers of this function rely on?

src/js/facility-locator/components/Select.jsx
cehsu: Yes good idea.

src/js/facility-locator/components/Select.jsx
cehsu: Good idea.

src/js/facility-locator/components/Select.jsx
cehsu: Good idea.

src/js/facility-locator/components/Select.jsx
cehsu: Good point.

src/js/facility-locator/components/Select.jsx
cehsu: Yes, but buttons are not allowed in buttons. This structure is based on Cauldron's [model](https://pattern-library.dequelabs.com/components/selects) but updated so that it conforms with the eslint-plugin-jsx-a11y no-noninteractive-element-interactions rule.

src/js/facility-locator/utils/helpers.js
cehsu: Options needs to be reset, so that it does not retain null references once facility is switched from benefits (which has more service options) to vet center (which has less services), the appropriate place to do this however is in `renderSelectOptions` via `this.props.options = [];`

src/js/facility-locator/components/Select.jsx
cehsu: We do not.

src/js/facility-locator/utils/helpers.js
cehsu: Now I remember why it was set up that way, props is read only. I will pass down another function prop to reset options.

src/js/facility-locator/components/Select.jsx
cehsu: This ensures that the select (or one of its descendents) is currently receiving focus, otherwise (such as if the user clicks elsewhere), it will close the dropdown menu.

src/js/facility-locator/components/Select.jsx
cehsu: Actually, this is resolved by your other recommendation to disable the button.

src/js/facility-locator/components/Select.jsx
cehsu: Actually there is no switching happening here, this line checks whether the dropdown is disabled, originally for styling purposes, but now also to determine the disabled attribute on the button in light of your recommendation.

src/js/facility-locator/components/Select.jsx
cehsu: It sounds like additional documentation would be useful here or perhaps some comments?

src/js/facility-locator/components/Listbox.jsx
```diff
@@ -0,0 +1,161 @@
...
+      'facility-option': true
+    });
+    return (
+      <div className="flex-center">
+        <div
+          id="serviceDropdown"
+          className={serviceOptionClasses}>
+          <span className="flex-center">
+            <span className="legend spacer"></span>
+            {truncate((benefitsServices[serviceType] || serviceType || 'All'), { length: (isMobile ? 38 : 27) })}
```
jbalboni: Can we document these magic numbers somehow? With constants and a comment, I think.

src/js/facility-locator/components/Listbox.jsx
```diff
@@ -0,0 +1,161 @@
...
+    const { currentQuery: { facilityType, serviceType } } = this.props;
+    return (
+      <div>
+        {hasIcons &&
+          <div>
+            <div className="flex-center">
+              {this.renderSelectButtonWithIcon(facilityType)}
+            </div>
+            <ul
+              role={!isDisabled ? 'listbox' : undefined}
```
jbalboni: Does the role need to be toggled? That seems odd.

src/js/facility-locator/components/Select.jsx
jbalboni: Is that not always going to be the case if `onBlur` is triggered? It looks like this is called on the parent element of this component.

src/js/facility-locator/components/Select.jsx
jbalboni: It still seems slightly different than that, plus I don't see anything in that model that would run counter to that eslint rule. You can put click handlers on div elements. Was there a reason for choosing the button over a div and the combobox role?

src/js/facility-locator/utils/helpers.js
```diff
@@ -0,0 +1,76 @@
...
+  if (code === keyMap.UP) return -1;
+  if (code === keyMap.DOWN) return 1;
+  return false;
+}
+
+export function getSelection(options, queryType) {
+  let selection = options[0];
+  if (queryType) {
+    const selectionType = queryType;
+    selection = options.reduce((acc, elem) => {
```
jbalboni: This seems more like `find` than reduce:

```
selection = options.find(elem => elem.id === selectionType) || selection;
```

src/js/facility-locator/utils/helpers.js
```diff
@@ -0,0 +1,76 @@
...
+}
+
+export function isSelect(code) {
+  return code === keyMap.ENTER || code === keyMap.SPACE;
+}
+
+export function isTraverse(keyEvent) {
+  const { keyCode: code } = keyEvent;
+  const result = code === keyMap.UP || code === keyMap.DOWN;
+  if (result) keyEvent.preventDefault();
```
jbalboni: This feels odd to have in an `isTraverse` function. Reading that name, I wouldn't expect it to call anything on the event itself. Applies to a couple other of these functions, too.

src/js/facility-locator/components/Select.jsx
```diff
@@ -0,0 +1,161 @@
...
+    this.setState({
+      dropdownActive: !this.state.dropdownActive
+    });
+  }
+
+  focusOption = (option, index) => {
+    if (option) {
+      option.focus();
+    }
+    if (this.options[index]) {
```
jbalboni: Can you have a case where this method is called with a valid option to focus on, but it's not in the options array? It's not clear to me why there are two slightly different conditionals in here.

src/js/facility-locator/components/Listbox.jsx
```diff
@@ -0,0 +1,161 @@
...
+      [`${kebabCase(facilityType)}-icon`]: facilityType
+    });
+
+    return (
+      <li
+        role="option"
+        tabIndex={this.props.dropdownActive ? '0' : '-1'}
+        aria-selected={isHovered}
+        ref={ elem => { this.props.options[index] = elem; }}
+        id={index > -1 ? (facilityType || 'AllFacilities') : null}
```
jbalboni: It looks like index is always greater than -1.

src/js/facility-locator/components/Listbox.jsx
```diff
@@ -0,0 +1,161 @@
...
+      </div>
+    );
+  }
+
+  render() {
+    const { hasIcons, isMobile, isDisabled } = this.props;
+    const { currentQuery: { facilityType, serviceType } } = this.props;
+    return (
+      <div>
+        {hasIcons &&
```
jbalboni: I'm having a hard time understanding why the code paths are so different depending on if there are icons or not. It seems like the difference should be limited to how the option and selected value are rendered inside each option. Are there are other reasons for the divergence that I am missing?

src/js/facility-locator/components/Listbox.jsx
```diff
@@ -0,0 +1,161 @@
...
+
+          return (
+            <li
+              tabIndex={this.props.dropdownActive ? '0' : '-1'}
+              role="option"
+              aria-selected={isHovered}
+              className={serviceOptionClasses}
+              key={k}
+              value={k}
+              ref={ elem => { this.props.options[i] = elem; }}
```
jbalboni: I think React discourages modifying props passed into components. Is there another way we can handle this? I'm also not quite understanding what the options array is used for.

cehsu: Good point, I will pass a callback down to update this. The options array stores the refs for the options and for the dropdown button to help manage the focus.

src/js/facility-locator/components/Listbox.jsx
cehsu: Good catch.

src/js/facility-locator/components/Listbox.jsx
cehsu: Good idea.

src/js/facility-locator/components/Listbox.jsx
cehsu: You're right, this component should't be rendered if isDisabled is true.

src/js/facility-locator/components/Select.jsx
cehsu: You are right, this is paranoid programming.

src/js/facility-locator/utils/helpers.js
cehsu: Good point.

src/js/facility-locator/utils/helpers.js
cehsu: Good point, I will relocate the preventDefault.

src/js/facility-locator/components/Listbox.jsx
cehsu: I'm not sure why the original logic was set up this way, but I will refactor it to make them more similar.

src/js/facility-locator/components/Select.jsx
cehsu: Good catch, this is now a div, the combobox role would cause the following violation here: Required ARIA child role not present: textbox (as it also does for Cauldron's model above)

cehsu: Actually adding the requisite tabIndex to the div causes a violation: `tabIndex` should only be declared on interactive elements
This was actually the reason a button was used. I will add the button role to the new div.

src/js/facility-locator/components/Select.jsx
cehsu: It appears that onBlur is triggered each time a descendent is blurred (navigating through options), this way, it only toggles when the new target is not a descendent. Also, onBlur isn't triggering on Listbox, this may have something to do with how React handles reconciliation. I will create a tech debt ticket to investigate this further.

src/js/facility-locator/components/Listbox.jsx
```diff
@@ -0,0 +1,110 @@
+import React, { Component } from 'react';
+import classNames from 'classnames';
+import { truncate, kebabCase } from 'lodash';
```
jbalboni: Can you switch this to the `import _ from 'lodash';` syntax? I'm not sure the lodash babel plugin picks this up to optimize.

jbalboni (commented):
This looks pretty good to me, but it looks like I can't use the arrow keys to move through the service dropdown. Works on on the facility one.

src/js/facility-locator/components/Select.jsx
cehsu: Hm, I think this caused an error which was why I made it a button:
  - sitemap 1/4 (1m 50s)
   Failed [fail]: ('http://localhost:3333/facilities/: [critical] Certain ARIA roles must contain particular children\nSee https://dequeuniversity. undefined undefined)
   See https://dequeuniversity.com/rules/axe/2.3/aria-required-children?application=axeAPI
   
   <ul class="dropdown" role="listbox"></ul>
   #react-root > div > .row > .facility-locator > div > .desktop-container > div:nth-child(1) > .search-controls-container.clearfix > form > div:nth-child(3) > [role="button"] > div > ul[role="listbox"]"
       at /application/test/util/nightwatch-commands/axeCheck.js:56:19
       at Array.forEach (native)
       at Object.<anonymous> (/application/test/util/nightwatch-commands/axeCheck.js:48:16)

I should better document these decisions next time.

jbalboni:
@cehsu Nevermind what I said about arrow keys not working, I didn't have the latest code.

jbalboni (approved):

bshyong (approved):
LGTM 

good work on the new components!

src/js/facility-locator/components/Select.jsx
jbalboni: @cehsu There's actually a bug in aXe here. I made a comment about it on your pre-need e2e test PR

cehsu: Thanks for the context @jbalboni, indeed, this seems the inverse of issues 214, 618 on the axe-core repo (which relate to the erroneous requirement that a specific child role be added when it shouldn't need to be), whereas in this case, it appears to be disallowing the assignment of child roles (listbox) to the button role (as the listbox children all posses the requisite option role -- so this is how I am interpreting the error). I do not see any issues related to this on the axe-core repo, do you propose we open a a bug issue on their repo and disable the check here for now?

jbalboni: @cehsu I'm not sure what you mean, the text you pasted in above is the message we've been getting elsewhere because of the bug in 618.

cehsu: Thanks @jbalboni I will make use of the workaround you shared in that case.

PR #6582, opened 2017-10-10
Update date validation in pension form

jbalboni:
This adds two new pieces of date validation to the pension form:

1. Date of birth must be before date entered service
2. Date of marriage must be before date of separation (for regular marriages and spouse marriages)

This also adds a migration to put users back on a page where the validation error exists.

screen shot 2017-10-09 at 3 03 42 pm
screen shot 2017-10-09 at 3 00 55 pm


jbalboni:
@peggygannon Can you take a look at the validation messages?

test/pensions/migrations.unit.spec.js
```diff
@@ -0,0 +1,100 @@
+import { expect } from 'chai';
+
+import migrations from '../../src/js/pensions/migrations';
+
+describe.only('Pension migrations', () => {
```
cvalarida: Probably don't want this anymore. 😉

test/pensions/migrations.unit.spec.js
```diff
@@ -0,0 +1,100 @@
...
+        }],
+        veteranDateOfBirth: '2013-01-01'
+      },
+      metadata: {
+        returnUrl: 'asdf'
+      }
+    });
+
+    expect(metadata.returnUrl).to.equal('/military/history');
+    expect(formData).to.be.an('object');
```
cvalarida: Out of curiosity, why check for this?

cvalarida (commented):
Looks good, but I'm wondering how it works if we save before the military service page, for example. We won't have the data that the migration looks for, so...I didn't look too hard at it, but the unit tests seem to show it's fine, so no worries.

Only reason it's not approved right here and now is because we can't let that `.only` into master. 🤖

test/pensions/migrations.unit.spec.js
jbalboni: Bleh, forgot to push a commit.

test/pensions/migrations.unit.spec.js
jbalboni: I want to make sure we don't accidentally forget to return `formData`.

src/js/pensions/validation.js
```diff
@@ -0,0 +1,21 @@
...
+    errors.addError('Date marriage ended must be after date of marriage');
+  }
+}
+
+export function validateServiceBirthDates(errors, service, formData) {
+  const fromDate = convertToDateField(formData.veteranDateOfBirth);
+  const toDate = convertToDateField(_.get('activeServiceDateRange.from', service));
+
+  if (!isValidDateRange(fromDate, toDate)) {
+    errors.activeServiceDateRange.from.addError('Date entering active service should be after birth date');
```
cehsu: Is legal enlistment age, also being checked here? If not, would it make sense to?

cvalarida (approved):
Beautiful!

src/js/pensions/config/form.js
```diff
@@ -60,6 +60,9 @@ import otherExpensesUI from '../definitions/otherExpenses';
 import currencyUI from '../../common/schemaform/definitions/currency';
 import GetFormHelp from '../../common/schemaform/GetPensionOrBurialFormHelp';
 
+import { validateServiceBirthDates, validateAfterMarriageDate } from '../validation';
```
cehsu: Would it make sense to also check child DOB?

cehsu (approved):
LGTM. Had a few minor questions about scope of changes, which if relevant could be converted into new tickets.

jbalboni:
@cehsu Those could be useful. I think the issue with date validation is that we can almost always add more checks, but eventually it becomes confusing to both us and users when something goes wrong. If you want to create a ticket with those suggestions that's fine with me, I just want to make sure @stacyhlinn thinks they would be valuable checks before we add them.

jbalboni:
@cvalarida To answer your question about saving before military history, the migration has a check for each of those arrays, and also the date range validation passes by default if one of the dates is empty. So we shouldn't hit any problems if users don't get to a particular page or opt not to fill in some information.

PR #6585, opened 2017-10-10
Fix prefill message on forms

jbalboni:
The prefill message was no longer showing up. I removed the previous check and added another one in the reducer, which should catch the usual cases where we try to prefill data and it is empty.

screen shot 2017-10-10 at 3 35 53 pm


cehsu (approved):
LGTM.

PR #6586, opened 2017-10-10
Merge claim status 508 tests with regular e2e tests

jbalboni:
Adds the axe call to the individual claim status e2e tests.


annekainicUSDS (approved):
LGTM! There are a lot of e2e tests for claim status, is this something we eventually want to condense? Or is there a good reason to have all of these? Only asking because I believe we discussed at one point trying to have fewer e2e tests per app.

jbalboni:
@annekainicUSDS It probably makes sense to consolidate some of them, they're more detailed than necessary, I think.

PR #6587, opened 2017-10-10
Add error handling when login urls fail to fetch

bshyong:
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/5245

adds error handling and messaging when login URLs fail to be fetched


src/js/common/helpers/login-helpers.js
```diff
@@ -102,25 +102,30 @@ export function addEvent(element, eventName, callback) {
...
-export function handleLogin(loginUrl, onUpdateLoginUrls) {
+export function handleLogin(loginUrl) {
   window.dataLayer.push({ event: 'login-link-clicked' });
   if (loginUrl) {
     window.dataLayer.push({ event: 'login-link-opened' });
     const receiver = window.open(`${loginUrl}&op=signin`, 'signinPopup', 'resizable=yes,scrollbars=1,top=50,left=500,width=500,height=750');
     receiver.focus();
-    return getLoginUrls(onUpdateLoginUrls || updateLogInUrls);
+  } else {
+    Promise.reject('Could not log in; loginUrl not provided.');
```
U-DON: Shouldn't this be returned? Actually, from a quick glance, none of our uses of `handleLogin` seem to expect a return value except for LoginModal, and it's not even used there correctly because it's only for aborting a request upon unmounting the component. So I'm not sure this `Promise.reject` would do anything.

https://github.com/department-of-veterans-affairs/vets-website/blob/cc4f3019eceba107269f7d51ab58b7ee599f0a9e/src/js/common/components/LoginModal.jsx#L74

U-DON (approved):
LGTM.

PR #6588, opened 2017-10-10
Resolve identity proof logic

bshyong:
- simplifies identity proof logic:
- No LOA3 users will be required to identity proof
- Users who are LOA1 or LOA2, with LOA3 available will be forced to level up/identity proof


saneshark (approved):
LGTM, tested locally.

IDme LOA1 -> LOA3 will prompt for MFA as part of the initial LOA1, as long as the leveling up to LOA3 happens within a 5 minute window (which should be the case seeing as we do it right away), then there should be no need for this mfaStart stuff anymore.

U-DON (commented):
LGTM. That `mfa_start` always did look suspect...

PR #6589, opened 2017-10-11
Main unit tests mg

markgreenburg:
connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/5204

- Updated unit test for case where letters are available
- Added unit tests for two uncovered cases
- Added unit tests for redux action creators called in `componentDidMount`

### DIFF
--- a/test/letters/containers/Main.unit.spec.jsx
+++ b/test/letters/containers/Main.unit.spec.jsx
@@ -86,4 +86,49 @@ describe('<Main>', () => {
     const tree = SkinDeep.shallowRender(<Main {...props}/>);
     expect(tree.subTree('#systemDownMessage')).to.be.ok;
   });
+
+  it('fetches letter data after mounting', () => {
+    const getLetterList = sinon.spy();
+    const props = { ...defaultProps, getLetterList };
+    const tree = SkinDeep.shallowRender(<Main {...props}/>);
+    const instance = tree.getMountedInstance();
+    instance.componentDidMount();
+    expect(getLetterList.callCount).to.equal(1);
+  });
+
+  it('fetches mailing address after mounting', () => {
+    const getMailingAddress = sinon.spy();
+    const props = { ...defaultProps, getMailingAddress };
+    const tree = SkinDeep.shallowRender(<Main {...props}/>);
+    const instance = tree.getMountedInstance();
+    instance.componentDidMount();
+    expect(getMailingAddress.callCount).to.equal(1);
+  });
+
+  it('fetches benefit summary options after mounting', () => {
+    const getBenefitSummaryOptions = sinon.spy();
+    const props = { ...defaultProps, getBenefitSummaryOptions };
+    const tree = SkinDeep.shallowRender(<Main {...props}/>);
+    const instance = tree.getMountedInstance();
+    instance.componentDidMount();
+    expect(getBenefitSummaryOptions.callCount).to.equal(1);
+  });
+
+  it('fetches country list after mounting', () => {
+    const getAddressCountries = sinon.spy();
+    const props = { ...defaultProps, getAddressCountries };
+    const tree = SkinDeep.shallowRender(<Main {...props}/>);
+    const instance = tree.getMountedInstance();
+    instance.componentDidMount();
+    expect(getAddressCountries.callCount).to.equal(1);
+  });
+
+  it('fetches state list after mounting', () => {
+    const getAddressStates = sinon.spy();
+    const props = { ...defaultProps, getAddressStates };
+    const tree = SkinDeep.shallowRender(<Main {...props}/>);
+    const instance = tree.getMountedInstance();
+    instance.componentDidMount();
+    expect(getAddressStates.callCount).to.equal(1);
+  });
 });
--- a/test/letters/containers/Main.unit.spec.jsx
+++ b/test/letters/containers/Main.unit.spec.jsx
@@ -2,13 +2,34 @@ import React from 'react';
 import SkinDeep from 'skin-deep';
 import { expect } from 'chai';
 import sinon from 'sinon';
+import _ from 'lodash';
 
 import { Main } from '../../../src/js/letters/containers/Main';
+import { AVAILABILITY_STATUSES } from '../../../src/js/letters/utils/constants';
+
+/**
+ * Define a simple child element for our component to render in tests. This
+ * gets passed in on props.children per defaultProps below. We'll use the
+ * testText to assert against as needed
+ */
+const testText = 'test child element';
+const childElement = (<span>{testText}</span>);
+
+// Destructure AVAILABILITY_STATUSES object for easier access
+const {
+  awaitingResponse,
+  available,
+  backendServiceError,
+  backendAuthenticationError,
+  invalidAddressProperty,
+  unavailable,
+  letterEligibilityError
+} = AVAILABILITY_STATUSES;
 
 const defaultProps = {
   letters: { },
   destination: { },
-  lettersAvailability: 'available',
+  lettersAvailability: available,
   benefitSummaryOptions: {
     benefitInfo: '',
     serviceInfo: '',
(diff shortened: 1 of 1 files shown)


test/letters/containers/Main.unit.spec.jsx
```diff
@@ -1,55 +1,134 @@
...
     const tree = SkinDeep.shallowRender(<Main {...props}/>);
-    expect(tree.subTree('LoadingIndicator')).to.be.false;
+    const childText = tree.subTree('span').text();
+    expect(childText).to.equal(testText);
   });
 
-  it('should show loading spinner when waiting for response', () => {
-    const props = _.merge({}, defaultProps, { lettersAvailability: 'awaitingResponse' });
+  it('shows a loading spinner when awaiting response', () => {
+    const props = { ...defaultProps, lettersAvailability: 'awaitingResponse' };
```
cvalarida: Out of curiosity, why change these from `_.merge()`?

test/letters/containers/Main.unit.spec.jsx
```diff
@@ -1,55 +1,134 @@
...
+    const tree = SkinDeep.shallowRender(<Main {...props}/>);
+    expect(tree.subTree('#systemDownMessage')).to.be.ok;
+  });
+
+  it('fetches letter data after mounting', () => {
+    const getLetterList = sinon.spy();
+    const props = { ...defaultProps, getLetterList };
+    const tree = SkinDeep.shallowRender(<Main {...props}/>);
+    const instance = tree.getMountedInstance();
+    instance.componentDidMount();
```
cvalarida: Did you check that you had to call this explicitly? Seems weird that `getMountedInstance()` wouldn't go through these mounting lifecycle methods, but I wouldn't be surprised. Just wondering if you checked.

Probably wouldn't hurt to put in a comment explaining _why_ you had to call this manually for anybody who looks later and perhaps isn't familiar with how SkinDeep does things.

test/letters/containers/Main.unit.spec.jsx
```diff
@@ -1,55 +1,134 @@
...
+    expect(getAddressCountries.callCount).to.equal(1);
+  });
+
+  it('fetches state list after mounting', () => {
+    const getAddressStates = sinon.spy();
+    const props = { ...defaultProps, getAddressStates };
+    const tree = SkinDeep.shallowRender(<Main {...props}/>);
+    const instance = tree.getMountedInstance();
+    instance.componentDidMount();
+    expect(getAddressStates.callCount).to.equal(1);
```
cvalarida: I wonder if it'd make sense to bundle these tests into a single one that tests that `it('should fetch all the things after mounting')`. That way, we would only shallow render once instead of five times. If the test fails, we'd see which line failed, so we'd know what needs to be fixed, but...it may be a micro-optimization. I can see it going either way.

test/letters/containers/Main.unit.spec.jsx
```diff
@@ -1,55 +1,134 @@
...
+  it('shows a loading spinner when awaiting response', () => {
+    const props = { ...defaultProps, lettersAvailability: 'awaitingResponse' };
     const tree = SkinDeep.shallowRender(<Main {...props}/>);
     expect(tree.subTree('LoadingIndicator')).to.be.ok;
   });
 
-  it('should show system down message for backend service error', () => {
-    const props = _.merge({}, defaultProps, { lettersAvailability: 'backendServiceError' });
+  it('shows a system down message for backend service error', () => {
+    const props = { ...defaultProps, lettersAvailability: 'backendServiceError' };
```
cvalarida: Can we change these magic strings into `ADDRESS_TYPES.*` from `src/js/letters/utils/constants.js`?

test/letters/containers/Main.unit.spec.jsx
```diff
@@ -1,55 +1,134 @@
...
+    const childText = tree.subTree('span').text();
+    expect(childText).to.equal(testText);
   });
 
-  it('should show loading spinner when waiting for response', () => {
-    const props = _.merge({}, defaultProps, { lettersAvailability: 'awaitingResponse' });
+  it('shows a loading spinner when awaiting response', () => {
+    const props = { ...defaultProps, lettersAvailability: 'awaitingResponse' };
     const tree = SkinDeep.shallowRender(<Main {...props}/>);
     expect(tree.subTree('LoadingIndicator')).to.be.ok;
```
cvalarida: I know this wasn't you, but can we change this to `expect(...).to.not.be.false`? [SkinDeep's docs](https://github.com/glenjamin/skin-deep#treesubtreeselector--matcher) say that `.subTree()` will return a tree or false. [Chai's docs](http://chaijs.com/api/bdd/#method_ok) say that the `ok` assertion isn't recommended.

test/letters/containers/Main.unit.spec.jsx
```diff
@@ -1,55 +1,134 @@
...
-import _ from 'lodash';
+import sinon from 'sinon';
 
 import { Main } from '../../../src/js/letters/containers/Main';
 
+const getBenefitSummaryOptionsSpy = sinon.spy();
+const getLetterListSpy = sinon.spy();
+const getMailingAddressSpy = sinon.spy();
+const getAddressCountriesSpy = sinon.spy();
+const getAddressStateSpy = sinon.spy();
```
cvalarida: You don't actually use these spies anywhere, right? Took me a while to notice that. I'd say just define `defaultProps.getMailingAddress()` and friends as empty arrow functions `() => {}` and then override them with spies in the tests where you check to see if they're called.

cvalarida (commented):
Looks good! I had a few comments on how we can harden the tests, but the only reason this isn't marked as approved is because there are unused spies defined at the top that we should nix.

test/letters/containers/Main.unit.spec.jsx
markgreenburg: @cvalarida Yeah, I was looking at a couple of different approaches and forgot to clean up :/
I'll move things around and push up a new commit

test/letters/containers/Main.unit.spec.jsx
markgreenburg: Yep, good catch!

test/letters/containers/Main.unit.spec.jsx
markgreenburg: Yeah...I'm with you on this, although I'd rather do something more straightforward. Like `expect(tree.find('LoadingIndicator').length).to.equal(1)`, but...pretty sure yours is our best option.

test/letters/containers/Main.unit.spec.jsx
markgreenburg: Yeah, I couldn't find great documentation on this, but from what I can tell in testing, you need to call this lifecycle method explicitly. I think it might vary from method to method, though. Will add comment.

test/letters/containers/Main.unit.spec.jsx
markgreenburg: Eh, wasn't explicitly necessary, I was just trying to remove unnecessary dependencies. Seemed like getting rid of lodash here was a good candidate since the prop object doesn't need to be nested and shallow clone functionality is easily replicable using native features. That being said, the prop object actually *is* nested, and while this nuance doesn't affect the current tests, I'd hate for it to break things down the road, so after thinking about this some more I'm changing it back to `lodash`. Oh well!

cvalarida: Heh, I did check to see if it was going to break because of that shallow cloning nonsense, but it's golden. I don't think you have to worry about changing it back; if we have to have another test down the road that needs the default props deeply cloned, we can cross that bridge when we get there.

test/letters/containers/Main.unit.spec.jsx
markgreenburg: Yeah, I was on the fence about this and actually initially wrote them this way. At some point I thought it might be simpler to just leave them separated but with you weighing in I'm leaning toward keeping things DRY.

annekainicUSDS (approved):
Tests look great! Liked Chris's suggestions too.

PR #6593, opened 2017-10-11
Auto save on change, skip warning after save

jbalboni:
No real visual changes, other than the "Saving..." message is now the same height as the saved message, to reduce the flickering when you save.


cehsu (approved):
LGTM. These are great changes, should any tests be updated as well?

PR #6595, opened 2017-10-12
Adds events to track clicks on the prev/next links

robertfairhead:
Closes https://www.github.com/department-of-veterans-affairs/vets.gov-team/issues/5296

I added support to the GTM config so the events will be reported once this deploys to production.


ncksllvn (approved):
LGTM!

PR #6597, opened 2017-10-12
Update child info url to dependent

jbalboni:
Step 6 in the connected issue. This updates the url from child-information to dependent-information and adds a migration.

### DIFF
--- a/src/js/hca/config/migrations.js
+++ b/src/js/hca/config/migrations.js
@@ -50,5 +50,19 @@ export default [
     }
 
     return { formData: newData, metadata };
+  },
+  // 2 -> 3, we're updating the url for the dependent info page since it's for dependents
+  // and not just children anymore
+  ({ formData, metadata }) => {
+    const url = metadata.returnUrl || metadata.return_url;
+
+    if (url === '/household-information/child-information') {
+      return {
+        formData,
+        metadata: _.set('returnUrl', '/household-information/dependent-information', metadata)
+      };
+    }
+
+    return { formData, metadata };
   }
 ];
--- a/src/js/hca/config/migrations.js
+++ b/src/js/hca/config/migrations.js
@@ -55,14 +55,12 @@ export default [
   // and not just children anymore
   ({ formData, metadata }) => {
     const url = metadata.returnUrl || metadata.return_url;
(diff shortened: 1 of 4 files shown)


src/js/hca/config/migrations.js
```diff
@@ -50,5 +50,17 @@ export default [
     }
 
     return { formData: newData, metadata };
+  },
+  // 2 -> 3, we're updating the url for the dependent info page since it's for dependents
+  // and not just children anymore
+  ({ formData, metadata }) => {
+    const url = metadata.returnUrl || metadata.return_url;
```
cehsu: Would `metadata` ever contain the camelcase version of these two, should only the snake case version be used here? The camelcase version is available on formData I believe, but I'm not sure it is available on metadata.

test/hca/config/migrations.unit.spec.js
```diff
@@ -88,4 +88,35 @@ describe('HCA migrations', () => {
...
   });
+  describe('third migration', () => {
+    const migration = migrations[2];
+    it('should update url when it matches', () => {
+      const data = {
+        formData: {
+          'view:reportChildren': false
+        },
+        metadata: {
+          returnUrl: '/household-information/child-information'
```
cehsu: Same questions as above.

src/js/hca/config/migrations.js
```diff
@@ -50,5 +50,17 @@ export default [
...
+    let newMetadata = metadata;
+
+    if (url === '/household-information/child-information') {
+      newMetadata = _.set('returnUrl', '/household-information/dependent-information', metadata);
```
cehsu: Answer to above question will influence how this `returnUrl` or `return_url` key is named.

cehsu (approved):
LGTM. Had a question about whether camelcase should be used.

src/js/hca/config/migrations.js
jbalboni: The vast majority of the time it should be camel case. The front end only ever reads and writes the return url in camel case. I believe the backend will send it in snake case when prefilling data, which the front end ignores in general, but I think we have to account for in migrations.

PR #6598, opened 2017-10-12
Fix file input validation error on claim status

jbalboni:
The steps for reproducing this are:

1. Add a file here: http://localhost:3001/track-claims/your-claims/344/additional-evidence
2. Remove the file
3. Click Add File, then click cancel in the file dialog

This is happening because the file input still has the old file after you pick one and this has some weird effects when we add more files using the same input.

### DIFF
--- a/src/js/common/components/form-elements/ErrorableFileInput.jsx
+++ b/src/js/common/components/form-elements/ErrorableFileInput.jsx
@@ -14,7 +14,12 @@ class ErrorableFileInput extends React.Component {
   }
 
   handleChange(domEvent) {
-    this.props.onChange(domEvent.target.files);
+    const files = domEvent.target.files;
+    // occasionally it seems like this event is triggering with an
+    // empty array, which is not helpful
+    if (files.length) {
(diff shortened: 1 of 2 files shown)


cehsu (approved):
LGTM.

src/js/common/components/form-elements/ErrorableFileInput.jsx
```diff
@@ -15,6 +15,10 @@ class ErrorableFileInput extends React.Component {
 
   handleChange(domEvent) {
     this.props.onChange(domEvent.target.files);
+    // clear the original input, otherwise events will be triggered
+    // with empty file arrays and sometimes uploading a file twice will
+    // not work
+    domEvent.target.value = null; // eslint-disable-line no-param-reassign
```
cehsu: How did you decide on null over undefined out of curiosity?

jbalboni: Mostly just preference, but `undefined` can technically be redefined, so it's theoretically slightly safer.

PR #6600, opened 2017-10-12
Remove build type flag for in progress applications on profile page

jbalboni:
@rtluu said we can release this now, so this PR removes the build type flag.

### DIFF
--- a/src/js/common/schemaform/FormApp.jsx
+++ b/src/js/common/schemaform/FormApp.jsx
@@ -138,10 +138,10 @@ class FormApp extends React.Component {
     if (props.isLoggedIn) {
       const currentForm = props.formConfig.formId;
       const isSaved = props.savedForms.some((savedForm) => savedForm.form === currentForm);
-      const isPrefill = props.prefillsAvailable.includes(currentForm);
+      const hasPrefillData = props.prefillsAvailable.includes(currentForm);
       const saveEnabled = !this.props.formConfig.disableSave;
-      if (saveEnabled && (isSaved || isPrefill)) {
-        props.fetchInProgressForm(currentForm, props.formConfig.migrations, !isSaved && isPrefill);
+      if (saveEnabled && (isSaved || hasPrefillData)) {
+        props.fetchInProgressForm(currentForm, props.formConfig.migrations, !isSaved && hasPrefillData);
       } else {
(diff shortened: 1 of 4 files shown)


src/js/common/schemaform/FormApp.jsx
```diff
@@ -141,7 +141,7 @@ class FormApp extends React.Component {
       const isPrefill = props.prefillsAvailable.includes(currentForm);
       const saveEnabled = !this.props.formConfig.disableSave;
       if (saveEnabled && (isSaved || isPrefill)) {
-        props.fetchInProgressForm(currentForm, props.formConfig.migrations, isPrefill);
+        props.fetchInProgressForm(currentForm, props.formConfig.migrations, !isSaved && isPrefill);
```
jbalboni: I found this when testing. The prefill message was being shown for saved apps when you use the `/resume` route. We know if there's a saved application, so we can tell it to not prefill in that case.

cehsu: Good catch, should we move this logic to line 141 when declaring isPrefill?

src/js/user-profile/containers/UserProfileApp.jsx
```diff
@@ -32,10 +32,10 @@ class UserProfileApp extends React.Component {
         <div className="usa-width-two-thirds medium-8 small-12 columns">
           <h1>Your Account</h1>
           <div>
-            {(__BUILDTYPE__ !== 'production') && <FormList
+            <FormList
```
cehsu: Should the tests for UserProfileApp be updated to reflect this? Or are the tests for FormList sufficient and that would that be a case of testing React and not our code?

cehsu (approved):
LGTM. Had two minor questions.

src/js/common/schemaform/FormApp.jsx
jbalboni: I think the logic is clearer the way it is now, but I did change the prefill variable name to make it clearer what's being checked.

src/js/user-profile/containers/UserProfileApp.jsx
jbalboni: Added a saved form to the mock data and a basic test.

PR #6602, opened 2017-10-12
Reducers unit tests

cvalarida:
A lot of these were already written, but I added some stuff. It's been a brain fog kind of day, so I very easily could have missed things. Feel free to tell me if the tests are wrong / not comprehensive enough.

### DIFF
--- a/test/letters/reducers/index.unit.spec.js
+++ b/test/letters/reducers/index.unit.spec.js
@@ -23,7 +23,6 @@ import {
   INVALID_ADDRESS_PROPERTY,
   LETTER_ELIGIBILITY_ERROR,
   LETTER_TYPES,
-  REQUEST_OPTIONS,
   SAVE_ADDRESS_PENDING,
   SAVE_ADDRESS_SUCCESS,
   SAVE_ADDRESS_FAILURE,
@@ -70,7 +69,7 @@ describe('letters reducer', () => {
     );
 
     expect(state.letters).to.be.empty;
-    expect(state.lettersAvailability).to.equal('unavailable');
+    expect(state.lettersAvailability).to.equal(AVAILABILITY_STATUSES.unavailable);
   });
 
   it('should handle backend service error', () => {
@@ -80,7 +79,7 @@ describe('letters reducer', () => {
     );
 
     expect(state.letters).to.be.empty;
-    expect(state.lettersAvailability).to.equal('backendServiceError');
+    expect(state.lettersAvailability).to.equal(AVAILABILITY_STATUSES.backendServiceError);
   });
 
   it('should handle backend authentication error', () => {
@@ -90,7 +89,7 @@ describe('letters reducer', () => {
     );
 
     expect(state.letters).to.be.empty;
-    expect(state.lettersAvailability).to.equal('backendAuthenticationError');
+    expect(state.lettersAvailability).to.equal(AVAILABILITY_STATUSES.backendAuthenticationError);
   });
 
   it('should handle invalid address', () => {
@@ -100,7 +99,7 @@ describe('letters reducer', () => {
     );
 
     expect(state.letters).to.be.empty;
-    expect(state.lettersAvailability).to.equal('invalidAddressProperty');
+    expect(state.lettersAvailability).to.equal(AVAILABILITY_STATUSES.invalidAddressProperty);
   });
 
   it('should handle a successful request for letters', () => {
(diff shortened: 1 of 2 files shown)


test/letters/reducers/index.unit.spec.js
```diff
@@ -38,71 +65,74 @@ describe('letters reducer', () => {
   it('should handle failure to fetch letters', () => {
```
markgreenburg: Might want to include a test to make sure the reducer returns `initialState` when it isn't passed an action? This would test the default case in the switch.

test/letters/reducers/index.unit.spec.js
```diff
@@ -184,22 +217,43 @@ describe('letters reducer', () => {
     );
 
     expect(state.countries).to.contain('USA');
+    expect(state.countriesAvailable).to.be.true;
```
markgreenburg: I think this is bordering on :paranoid-parrot: but it seems that we're only testing a specific value of the new state in general. Wouldn't it be more comprehensive that the whole new state is `eql` to what we expect? That way, we're not inadvertently letting other state updates or mutations get through our tests. E.g., if one of the reducers is changed down the road to update the original thing *and* something else, it'll fail our tests until the test is updated appropriately.

test/letters/reducers/index.unit.spec.js
```diff
@@ -38,71 +65,74 @@ describe('letters reducer', () => {
   it('should handle failure to fetch letters', () => {
     const state = lettersReducer.letters(
```
markgreenburg: Given that this is used so many times, I wonder if it would make sense to abstract this out into something like a `reduce(action) { ...call the letters reducer with a pre-defined initialState and action passed in as arg, return new state...}` that can just manually be called in each test, so: `const newState = reduce(action); expect(newState).to.deep-eql({ oldState with changes });`.

I mean, this is already written so probably not worth it at this point. I'm just genuinely curious what you think about that approach for testing reducers in general.

test/letters/reducers/index.unit.spec.js
cvalarida: I agree with the :paranoid-parrot:, but I like it. Should be a pretty easy update.

test/letters/reducers/index.unit.spec.js
cvalarida: :man_shrugging: 

Why not?

test/letters/reducers/index.unit.spec.js
cvalarida: 🤔 
Yeah...yeah, that's not bad. Then if we needed to change the initial state to test a thing, we can just call it with an optional parameter if we were so inclined. I like it.

test/letters/reducers/index.unit.spec.js
```diff
@@ -35,86 +54,78 @@ const benefitSummaryOptionData = {
...
             attributes: {
               letters: [
                 {
-                  letterType: 'commissary',
+                  letterType: LETTER_TYPES.commissary,
                   name: 'Commissary Letter'
                 }
-              ]
+              ],
+              fullName: 'Johann Bach'
```
annekainicUSDS: <3

test/letters/reducers/index.unit.spec.js
```diff
@@ -157,68 +167,183 @@ describe('letters reducer', () => {
...
+
+  it('should handle failure to fetch states', () => {
+    const state = reduce({ type: GET_ADDRESS_STATES_FAILURE });
+
     expect(state.states).to.be.empty;
+    expect(state.statesAvailable).to.be.false;
+  });
+
+
+  it('should handle a a letter eligibility error', () => {
```
annekainicUSDS: Minor typo: extra `a`

test/letters/reducers/index.unit.spec.js
```diff
@@ -157,68 +167,183 @@ describe('letters reducer', () => {
...
+  it('should handle failing to save an address', () => {
+    const state = reduce(
+      {
+        type: SAVE_ADDRESS_FAILURE,
+        address: '123 Main St'
+      }
+    );
+
+    expect(state.savePending).to.be.false;
+    expect(state.saveAddressError).to.be.true;
```
annekainicUSDS: This is interesting, I hadn't noticed before that the code worked this way. I wonder why we storing 2 separate fields here instead of doing what we do for fetching letters and address, and store one field and have the values be something like: `pending`, `failure`, `success`, etc?

annekainicUSDS (approved):
LGTM! One question

test/letters/reducers/index.unit.spec.js
cvalarida: Yeah... 🤔  Now that you mention it, that is odd. Let's write up a ticket to address that.

PR #6603, opened 2017-10-13
BSL unit tests

markgreenburg:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/5207

Cleaning up some existing tests for the veteran benefit summary letter and adding some more to cover important functionality.

### DIFF
--- a/test/letters/containers/VeteranBenefitSummaryLetter.unit.spec.jsx
+++ b/test/letters/containers/VeteranBenefitSummaryLetter.unit.spec.jsx
@@ -25,7 +25,8 @@ const defaultProps = {
   },
   isVeteran: true,
   optionsAvailable: true,
-  requestOptions: {}
+  requestOptions: {},
+  updateBenefitSummaryRequestOption: () => {}
 };
 
 describe('<VeteranBenefitSummaryLetter>', () => {
@@ -46,15 +47,10 @@ describe('<VeteranBenefitSummaryLetter>', () => {
 
   it('should show service info options', () => {
     const tree = SkinDeep.shallowRender(<VeteranBenefitSummaryLetter {...defaultProps}/>);
-    expect(tree.dive(['div', 'div', '#militaryService'])).not.to.be.empty;
+    expect(tree.subTree('#militaryService')).not.to.be.empty;
   });
 
   it('should update request option when checked', () => {
-    const oldWindow = global.window;
-    global.window = {
-      dataLayer: [],
-    };
-
     const updateOption = sinon.spy();
     const props = _.set('updateBenefitSummaryRequestOption', updateOption, defaultProps);
     const component = ReactTestUtils.renderIntoDocument(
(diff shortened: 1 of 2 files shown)


test/letters/containers/VeteranBenefitSummaryLetter.unit.spec.jsx
```diff
@@ -51,52 +46,93 @@ describe('<VeteranBenefitSummaryLetter>', () => {
...
       target: {
-        checked: false
-      }
-    });
+        id: DOMid,
+        checked: checkedValue,
+      },
+    };
+
+    ReactTestUtils.Simulate.change(inputs[0], mockDOMEvent);
```
cvalarida: If you use `getFormDOM` from the `test/util/schemaform-utils`, we have a `setCheckbox` method to shorten this.

test/letters/containers/VeteranBenefitSummaryLetter.unit.spec.jsx
```diff
@@ -51,52 +46,93 @@ describe('<VeteranBenefitSummaryLetter>', () => {
...
+    const doubleService = defaultProps.benefitSummaryOptions.serviceInfo.concat(navyService);
+    const props = _.set('benefitSummaryOptions.serviceInfo', doubleService, defaultProps);
+    const tree = SkinDeep.shallowRender(<VeteranBenefitSummaryLetter {...props}/>);
+    const serviceRows = tree.dive(['#militaryServiceTable', 'tbody']).everySubTree('tr');
+
+    expect(serviceRows.length).to.equal(doubleService.length);
+  });
+
+  it('handles check and uncheck events', () => {
+    const oldWindow = { ...global.window };
```
cvalarida: This is a shallow clone, so no guarantees this won't mess things up later. :man_shrugging:

I'm pretty sure `global.window.dataLayer` strictly equals `oldWindow.dataLayer` (so they have the same reference) because `dataLayer` is an object. Consequently, this precaution doesn't really do anything. Honestly, it's probably fine for now, but something to keep in mind for if future tests fail unexpectedly. We can always change it to use `_.cloneDeep()` later as a drop-in replacement.

cvalarida (approved):
:shipit:
Couple o' comments, but nothing blocking.

test/letters/containers/VeteranBenefitSummaryLetter.unit.spec.jsx
markgreenburg: Yeah, rather than spend more effort thinking about this I'm just going to deep-clone it :)
This will probably never matter, but would probably be a nasty bug to track down if it did come up.

test/letters/containers/VeteranBenefitSummaryLetter.unit.spec.jsx
markgreenburg: Ahh, good to know! I'm just gonna keep this version only because I'm also testing the contents of `global.window.dataLayer`, which has an `id` property for each item in the array - using `formDOM` doesn't appear to support specifying an id property for the event.

PR #6604, opened 2017-10-13
Added transformForSubmit to pre-need.

U-DON:
There can be multiple claimants (persons checking for their eligibility) in a pre-need form, and each one can designate another claimant as their sponsor (a veteran) and/or as their preparer (applicant, or the person filling out the form on all claimants' behalves).

In the schema that's received by the API, each application has a claimant, veteran, and applicant. But in the actual form that's being displayed, there can be multiple claimants (called "applicants" in the form), and each claimant chooses a sponsor by name (`view:sponsor`) and there is one preparer (`view:preparer`) for the whole form. So this transform function does the work of copying over the appropriate data for the API to receive.

### DIFF
--- a/src/js/pre-need/config/form.js
+++ b/src/js/pre-need/config/form.js
@@ -24,6 +24,7 @@ import {
   formatName,
   isVeteran,
   requiresSponsorInfo,
+  transform,
   veteranUISchema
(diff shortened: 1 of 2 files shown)


bshyong (approved):
LGTM

PR #6608, opened 2017-10-13
Refactor address actions and bsl logic

annekainicUSDS:
This PR does a few things:
1. Alphabetize and reorder actions with comments
2. Remove actions that aren't being used
3. Remove all logic for error response from `getMailingAddress()`, just dispatch the action for a failed fetch of the address.
4. Amend logic in the BSL so that military service history or DownloadLetterLink doesn't render if `optionsAvailable === false`. What will render instead is:
screen shot 2017-10-13 at 5 10 51 pm


src/js/letters/actions/letters.js
```diff
@@ -2,28 +2,28 @@ import Raven from 'raven-js';
 
 import { apiRequest } from '../utils/helpers.jsx';
 import {
+  ADDRESS_TYPES,
   BACKEND_AUTHENTICATION_ERROR,
   BACKEND_SERVICE_ERROR,
-  GET_LETTERS_FAILURE,
-  GET_LETTERS_SUCCESS,
+  GET_ADDRESS_COUNTRIES_SUCCESS,
+  GET_ADDRESS_COUNTRIES_FAILURE,
```
cvalarida: As long as we're alphabetizing, `FAILURE` comes before `SUCCESS`.

...how philosophical...

src/js/letters/actions/letters.js
```diff
@@ -2,28 +2,28 @@ import Raven from 'raven-js';
...
   GET_ADDRESS_FAILURE,
   GET_ADDRESS_SUCCESS,
+  GET_ADDRESS_STATES_FAILURE,
+  GET_ADDRESS_STATES_SUCCESS,
   GET_BENEFIT_SUMMARY_OPTIONS_FAILURE,
   GET_BENEFIT_SUMMARY_OPTIONS_SUCCESS,
+  GET_LETTERS_FAILURE,
```
cvalarida: Again, as long as we're re-ordering these, we can move this to right above `GET_LETTERS_SUCCESS`.

src/js/letters/actions/letters.js
```diff
@@ -92,33 +92,8 @@ export function getMailingAddress() {
...
-        );
-      }
-    ).catch((error) => {
-      if (error.message.match('vets_address_error_server_get')) {
-        Raven.captureException(error);
-        return dispatch({ type: GET_ADDRESS_FAILURE });
-      }
-      throw error;
-    });
+      () => dispatch({ type: GET_ADDRESS_FAILURE })
```
cvalarida: This also means we're not going to log any of these errors to Sentry. Is that okay? They're already logged on the back end, right?

src/js/letters/utils/constants.js
```diff
@@ -1,34 +1,40 @@
...
+// getAddressCountries() actions
+export const GET_ADDRESS_COUNTRIES_FAILURE = 'GET_ADDRESS_COUNTRIES_FAILURE';
+export const GET_ADDRESS_COUNTRIES_SUCCESS = 'GET_ADDRESS_COUNTRIES_SUCCESS';
 
-export const UPDATE_BENFIT_SUMMARY_REQUEST_OPTION = 'UPDATE_BENFIT_SUMMARY_REQUEST_OPTION';
+// getAddressStates() actions
+export const GET_ADDRESS_STATES_FAILURE = 'GET_ADDRESS_STATES_FAILURE';
+export const GET_ADDRESS_STATES_SUCCESS = 'GET_ADDRESS_STATES_SUCCESS';
 
-export const UPDATE_ADDRESS = 'UPDATE_ADDRESS';
```
cvalarida: There were no actions created and this was the only one removed...am I reading that right?

cvalarida (approved):
Looks great! I had a couple questions for my peace of mind, but they're non-blocking for sure.

:shipit:

src/js/letters/actions/letters.js
annekainicUSDS: hehe you're right :) will fix!

src/js/letters/actions/letters.js
annekainicUSDS: yep!

src/js/letters/actions/letters.js
annekainicUSDS: Yep, I confirmed with Alastair that they're all being logged on the back-end, so I think we're good.

src/js/letters/utils/constants.js
annekainicUSDS: yep!

PR #6613, opened 2017-10-16
Avoid dispatching some login actions when there's an error

jbalboni:
These errors are happening whenever users open a page with an expired session. See connected issue for an example Sentry error.

Note: I grabbed this because I initially thought it was related to some GA changes that I made, but that turned out not to be the case.

### DIFF
--- a/src/js/common/helpers/login-helpers.js
+++ b/src/js/common/helpers/login-helpers.js
@@ -29,7 +29,9 @@ export function getVerifyUrl(onUpdateVerifyUrl) {
   }).then(response => {
     return response.json();
   }).then(json => {
-    onUpdateVerifyUrl(json.identity_proof_url);
+    if (json.identity_proof_url) {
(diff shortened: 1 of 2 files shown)


src/js/common/helpers/login-helpers.js
```diff
@@ -29,7 +29,9 @@ export function getVerifyUrl(onUpdateVerifyUrl) {
   }).then(response => {
     return response.json();
   }).then(json => {
-    onUpdateVerifyUrl(json.identity_proof_url);
+    if (json.identity_proof_url) {
```
jbalboni: It's more "correct" probably to check response.ok and throw an error if not, but that would leave uncaught promise rejections. We don't appear to need to handle rejections, so this seemed like a reasonable way to avoid the errors.

bshyong (approved):
LGTM

U-DON (approved):
LGTM.

PR #6614, opened 2017-10-16
Show error when migration fails for saved form

jbalboni:
If we fail migrating a saved form to an updated version, we're showing an empty error message. This updates it to show an error message.

Generic error:
screen shot 2017-10-17 at 4 54 39 pm

Migration error:
screen shot 2017-10-17 at 4 52 49 pm


jbalboni:
@peggygannon There are two error states here with the above message:

1. Some sort of connection or server error happened. This might go away when retrying, or it might be something we need to fix.
2. There was an error updating their saved application to the most recent form configuration. This is something we would need to fix, and they won't be able to continue until we do. They could probably start their application over, though.

Let me know if you want to update the message(s) shown for those cases.

peggygannon:
@jbalboni  See below for my suggested messages. I think we need two messages for these: 
For scenario 1:
We're sorry. We're having some server issues and are working to fix them. Please try applying again in a few moments.
Button:  Back and Continue Your Application

For scenario 2:
We're sorry. Something went wrong when we tried to access your application. We're working to fix this. You can try applying again in a few moments or start your application over. 
Buttons: Back  and  Continue Your Application

peggygannon (commented):
@jbalboni See below for my suggested messages. I think we need two messages for these:
For scenario 1:
We're sorry. We're having some server issues and are working to fix them. Please try applying again in a few moments.
Button: Back and Continue Your Application

For scenario 2:
We're sorry. Something went wrong when we tried to access your application. We're working to fix this. You can try applying again in a few moments or start your application over.
Buttons: Back and Continue Your Application

src/js/common/schemaform/SaveInProgressErrorPage.jsx
```diff
@@ -78,10 +78,21 @@ class SaveInProgressErrorPage extends React.Component {
...
+          </div>
+        );
+        break;
+      case LOAD_STATUSES.invalidData:
+        content = (
+          <div>
+            <div className="usa-alert usa-alert-error no-background-image">We’re sorry. Something went wrong when we tried to access your application. We’re working to fix this. You can try applying again in a few moments or start your application over.</div>
+            <div style={{ marginTop: '30px' }}>
+              {this.getBackButton()}
+              <button className="usa-button-primary" onClick={this.reloadForm}>Continue Application</button>
```
cehsu: Would we want to provide a button option to start over since starting over is mentioned in the error message?

test/common/schemaform/SaveInProgressErrorPage.unit.spec.jsx
```diff
@@ -70,7 +70,7 @@ describe('<SaveInProgressErrorPage>', () => {
     );
     const findDOM = findDOMNode(tree);
 
-    expect(findDOM.querySelector('.usa-alert').textContent).to.contain('We’re sorry, but something went wrong.');
+    expect(findDOM.querySelector('.usa-alert').textContent).to.contain('We’re sorry, but something went wrong. We can’t find your application');
```
cehsu: Should this text match the updated text above?

cehsu (approved):
LGTM. Had a few minor questions.

src/js/common/schemaform/SaveInProgressErrorPage.jsx
```diff
@@ -78,10 +78,21 @@ class SaveInProgressErrorPage extends React.Component {
       case LOAD_STATUSES.failure:
         content = (
           <div>
-            <div className="usa-alert usa-alert-error no-background-image">We’re sorry, but something went wrong. Please try applying again in a few moments.</div>
+            <div className="usa-alert usa-alert-error no-background-image">We’re sorry. We’re having some server issues and are working to fix them. Please try applying again in a few moments.</div>
             <div style={{ marginTop: '30px' }}>
               {this.getBackButton()}
-              <button className="usa-button-primary" onClick={this.reloadForm}>Resume previous application</button>
+              <button className="usa-button-primary" onClick={this.reloadForm}>Continue Application</button>
```
cehsu: It looks like this should be "Continue Your Application" (missing "Your")

src/js/common/schemaform/SaveInProgressErrorPage.jsx
cehsu: Same as above.

test/common/schemaform/SaveInProgressErrorPage.unit.spec.jsx
```diff
@@ -85,9 +85,9 @@ describe('<SaveInProgressErrorPage>', () => {
     );
     const findDOM = findDOMNode(tree);
 
-    expect(findDOM.querySelector('.usa-alert').textContent).to.contain('We’re sorry, but something went wrong.');
+    expect(findDOM.querySelector('.usa-alert').textContent).to.contain('We’re sorry. We’re having some server issues');
     expect(findDOM.querySelector('.usa-button-outline').textContent).to.contain('Back');
-    expect(findDOM.querySelector('.usa-button-primary').textContent).to.contain('Resume previous application');
+    expect(findDOM.querySelector('.usa-button-primary').textContent).to.contain('Continue Application');
```
cehsu: Same as above.

test/common/schemaform/SaveInProgressErrorPage.unit.spec.jsx
jbalboni: It's a different message, but I'll make it more in line with the other ones.

src/js/common/schemaform/SaveInProgressErrorPage.jsx
jbalboni: I'm not sure we want to encourage starting over. @stacyhlinn, should we show the start over button on this error page? Or maybe change the text to mention that they need to click Back first to start over?

stacyhlinn (approved):

PR #6615, opened 2017-10-16
Defer Validations

cvalarida:
When select elements are changed, they're marked as dirty immediately and validation is run against them. The text inputs are dirtied only after blurred, so validation isn't run on them until after the user is finished with editing the field for the first time.

### DIFF
--- a/src/js/letters/containers/AddressSection.jsx
+++ b/src/js/letters/containers/AddressSection.jsx
@@ -153,17 +153,31 @@ export class AddressSection extends React.Component {
     });
   }
 
-
-  handleChange = (fieldName, update) => {
+  dirtyInput = (fieldName) => {
     let fieldsToValidate = this.state.fieldsToValidate;
-    // When a field is changed, make sure we validate it
+    // When a field is blurred, add it to the fields we want to validate
     if (!fieldsToValidate[fieldName]) {
-      // If we set state here, the validation won't run when the field is modified the first time
-      // This is because the state won't be updated in time for validateAll to catch it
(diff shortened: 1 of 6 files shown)


src/js/letters/containers/AddressSection.jsx
```diff
@@ -223,6 +242,7 @@ export class AddressSection extends React.Component {
         <div>
           <Address
             onInput={this.handleChange}
+            onBlur={this.dirtyInput}
```
markgreenburg: Probably totally fine, but worth pointing out that both `this.dirtyInput` and `this.handleChange` call `setState`, with the results of the `setState` call in`onBlur` getting reused in the `setState` call in `handleChange`. Since these are async I suppose we could get into a weird state where the two are out of sync, but I'm not sure that's an actual problem - everything seems to work as expected in local testing and hopefully there's enough time between `onBlur` triggering and `onInput` triggering that this'll never come up.

markgreenburg (approved):
Nice, looks great!

src/js/letters/containers/AddressSection.jsx
cvalarida: Yeah, I was trying to figure out how to abstract that in a way that made sense, but I came to that same conclusion--they're not going to get messed up with normal use. The only way I can think of for the state to get weird is if `dirtyInput` is called before `handleChange` is finished, but that would take _really_ fast input and blur. Even then, the out-of-sync-ness (there's probably a proper word for that...) isn't a big deal, I don't think, since the next edit will ensure everything's back to normal.

I'm pretty sure, anyhow.

PR #6619, opened 2017-10-17
Add tests and delete unused vars

markgreenburg:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/5202

There's a considerable amount of stuff going on in this component so I tried to cover a few more pieces of logic with tests. Unfortunately, these probably still aren't as comprehensive as they could be because I found a couple of things to be hard to test. Namely, it's kind of tough to test the content of each collapsible panel, as well as whether or not a download link renders for each. I think this should be sufficient though.

Changes:
- Deleted a few unused vars from test as well as the actual component
- Added tests that ensure certain props are passed down to children
- Added test for rendering an error message when letters unavailable

### DIFF
--- a/test/letters/containers/LetterList.unit.spec.jsx
+++ b/test/letters/containers/LetterList.unit.spec.jsx
@@ -3,6 +3,7 @@ import SkinDeep from 'skin-deep';
 import { expect } from 'chai';
 
 import { LetterList } from '../../../src/js/letters/containers/LetterList.jsx';
+import { AVAILABILITY_STATUSES } from '../../../src/js/letters/utils/constants';
 
 const defaultProps = {
   letters: [
@@ -11,7 +12,7 @@ const defaultProps = {
       letterType: 'commissary'
     },
     {
-      name: 'Benefit Summary Letter',
+      name: 'Benefit Summary and Service Verification Letter',
       letterType: 'benefit_summary'
     },
     {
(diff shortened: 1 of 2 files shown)


test/letters/containers/LetterList.unit.spec.jsx
```diff
@@ -11,27 +12,45 @@ const defaultProps = {
...
   ],
-  letterDownloadStatus: {}
+  lettersAvailability: AVAILABILITY_STATUSES.available,
+  letterDownloadStatus: {},
+  optionsAvailable: true,
 };
 
 describe('<LetterList>', () => {
-  it('should render', () => {
+  it('renders', () => {
```
cvalarida: The pattern we've been using is `it('should...')`. Why remove the 'should'?

test/letters/containers/LetterList.unit.spec.jsx
```diff
@@ -11,27 +12,45 @@ const defaultProps = {
...
     const tree = SkinDeep.shallowRender(<LetterList {...defaultProps}/>);
-    const vdom = tree.getRenderOutput();
-    expect(vdom).to.exist;
+    expect(tree.type).to.equal('div');
```
cvalarida: Very nice.

cvalarida (approved):

test/letters/containers/LetterList.unit.spec.jsx
markgreenburg: Yeah, that's fair. I'll stick to the pattern going forward. Some of the test names got a little long-winded when written out this way (i.e. too much horizontal scrolling in my split panes) so I cut them all for consistency. Likely not necessary.

PR #6620, opened 2017-10-17
Render trackedItem's displayName or "Document Request"

cehsu:
If trackedItem's displayName property (often something like "Request 10") is missing, "Document Request" will be rendered instead. This is based on how the `document.title` is conditionally determined on [lines 37-41](https://github.com/department-of-veterans-affairs/vets-website/blob/bbd97694d8e6e926da2eff212fef217543c233b2/src/js/claims-status/containers/DocumentRequestPage.jsx#L39) of the same file (`src/js/claims-status/containers/DocumentRequestPage.jsx`):
```
    if (this.props.trackedItem) {
      document.title = `Request for ${this.props.trackedItem.displayName}`;
    } else {
      document.title = 'Document Request';
    }
```

localhost-3001-track-claims-your-claims-339-document-request-211443

### DIFF
--- a/src/js/claims-status/containers/DocumentRequestPage.jsx
+++ b/src/js/claims-status/containers/DocumentRequestPage.jsx
@@ -97,7 +97,7 @@ class DocumentRequestPage extends React.Component {
                     <Element name="uploadError"/>
                     <Notification title={message.title} body={message.body} type={message.type}/>
                   </div>}
-                <h1 className="claims-header">{trackedItem.displayName || 'Document Request'}</h1>
+                <h1 className="claims-header">{trackedItem ? trackedItem.displayName : 'Document Request'}</h1>
                 {trackedItem.type.endsWith('you_list') ? <DueDate date={trackedItem.suspenseDate}/> : null}
                 {trackedItem.type.endsWith('others_list')
                   ? <div className="optional-upload">
--- a/src/js/claims-status/containers/DocumentRequestPage.jsx
+++ b/src/js/claims-status/containers/DocumentRequestPage.jsx
@@ -98,13 +98,13 @@ class DocumentRequestPage extends React.Component {
                     <Notification title={message.title} body={message.body} type={message.type}/>
                   </div>}
                 <h1 className="claims-header">{trackedItem ? trackedItem.displayName : 'Document Request'}</h1>
-                {trackedItem.type.endsWith('you_list') ? <DueDate date={trackedItem.suspenseDate}/> : null}
-                {trackedItem.type.endsWith('others_list')
+                {trackedItem && trackedItem.type.endsWith('you_list') ? <DueDate date={trackedItem.suspenseDate}/> : null}
+                {trackedItem && trackedItem.type.endsWith('others_list')
                   ? <div className="optional-upload">
                     <p><strong>Optional</strong> - We’ve asked others to send this to us, but you may upload it if you have it.</p>
                   </div>
                   : null}
-                <p>{trackedItem.description}</p>
+                {trackedItem && <p>{trackedItem.description}</p>}
                 <AddFilesForm
                   field={this.props.uploadField}
                   progress={this.props.progress}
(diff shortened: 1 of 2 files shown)


src/js/claims-status/containers/DocumentRequestPage.jsx
```diff
@@ -97,7 +97,7 @@ class DocumentRequestPage extends React.Component {
                     <Element name="uploadError"/>
                     <Notification title={message.title} body={message.body} type={message.type}/>
                   </div>}
-                <h1 className="claims-header">{trackedItem.displayName}</h1>
+                <h1 className="claims-header">{trackedItem.displayName || "Document Request"}</h1>
```
jbalboni: I don't think this will fix the error, since it's trackedItem that is `undefined`, based on the error message.

cehsu: Thanks Jeff, this has been replaced by the following:
```
<h1 className="claims-header">{trackedItem ? trackedItem.displayName : 'Document Request'}</h1>
```
As well as making all other references to trackedItem similar.
In the event that trackedItem is undefined, how should AddFilesForm and its onSubmit prop handle this?

jbalboni: This page in general is not very useful if the tracked item doesn't exist. I wonder if we should just redirect back to the claim detail page when that happens?

cehsu: Thanks @jbalboni, good idea, changes made (and unmade).

src/js/claims-status/containers/DocumentRequestPage.jsx
```diff
@@ -37,7 +37,7 @@ class DocumentRequestPage extends React.Component {
     if (this.props.trackedItem) {
       document.title = `Request for ${this.props.trackedItem.displayName}`;
     } else {
-      document.title = 'Document Request';
+      this.props.router.push(`/your-claims/${this.props.params.id}/status`);
```
jbalboni: I don't think this is the right place for redirecting. This method is only called once, after rendering, so it probably won't avoid the error we're seeing. We also need to wait for claim information to load, before we know for sure that the tracked item is missing.

I think you want to put this in `componentWillReceiveProps` and redirect when the `loading` prop becomes true and there's no `trackedItem`.

cehsu: Thanks for pointing that out @jbalboni, change made.

src/js/claims-status/containers/DocumentRequestPage.jsx
```diff
@@ -34,18 +34,17 @@ const Element = Scroll.Element;
 class DocumentRequestPage extends React.Component {
   componentDidMount() {
     this.props.resetUploads();
-    if (this.props.trackedItem) {
```
cehsu: Given the new redirect logic, is this check no longer required?

jbalboni: I think so. If this component mounts while loading, we wouldn't have a tracked item yet.

src/js/claims-status/containers/DocumentRequestPage.jsx
```diff
@@ -34,18 +34,17 @@ const Element = Scroll.Element;
...
+    document.title = `Request for ${this.props.trackedItem.displayName}`;
     if (!this.props.loading) {
       setUpPage();
     } else {
       scrollToTop();
     }
   }
   componentWillReceiveProps(props) {
+    if (props.loading && !!props.trackedItem) {
+      this.props.router.push(`/your-claims/${this.props.params.id}/status`);
```
jbalboni: I think we want to use `replace` instead of `push`. Otherwise, using the back button will get someone stuck.

src/js/claims-status/containers/DocumentRequestPage.jsx
```diff
@@ -46,6 +46,9 @@ class DocumentRequestPage extends React.Component {
     }
   }
   componentWillReceiveProps(props) {
+    if (props.loading && !!props.trackedItem) {
```
jbalboni: I think this should be `if (!props.loading && !!props.trackedItem) {`

cehsu: Thanks @jbalboni, you're right, the boolean clauses were reversed, changes made.

jbalboni (approved):

PR #6626, opened 2017-10-18
Update SIP intro prefill/sip alert message

cehsu:
### User is not signed in
localhost-3001-health-care-apply-application- 1

### User is signed in (but does not yet have any saved messages)
localhost-3001-health-care-apply-application-introduction 5

### DIFF
--- a/src/js/common/schemaform/SaveInProgressIntro.jsx
+++ b/src/js/common/schemaform/SaveInProgressIntro.jsx
@@ -11,8 +11,9 @@ import FormStartControls from './FormStartControls';
 export default class SaveInProgressIntro extends React.Component {
   getAlert(savedForm) {
     let alert;
-
-    if (this.props.user.login.currentlyLoggedIn) {
+    const { profile, login } = this.props.user;
+    const prefillAvailable = !!(profile && profile.prefillsAvailable.includes(this.props.formId));
+    if (login.currentlyLoggedIn) {
       if (savedForm) {
         const savedAt = this.props.lastSavedDate
           ? moment(this.props.lastSavedDate)
@@ -32,7 +33,7 @@ export default class SaveInProgressIntro extends React.Component {
             <br/>
           </div>
         );
-      } else {
+      } else if (prefillAvailable){
         alert = (
           <div>
             <div className="usa-alert usa-alert-info schemaform-sip-alert">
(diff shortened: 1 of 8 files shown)


src/js/common/schemaform/SaveInProgressIntro.jsx
```diff
@@ -37,7 +37,7 @@ export default class SaveInProgressIntro extends React.Component {
           <div>
             <div className="usa-alert usa-alert-info schemaform-sip-alert">
               <div className="usa-alert-body">
-                You can save this form in progress, and come back later to finish filling it out.
+                <strong>Note:</strong> As you are signed in to your account, we can prefill part of your application based on your account details. You can also save your form in progress, and come back later to finish filling it out.
```
cehsu: @peggygannon I made a slight tweak for if a user is logged in but doesn't yet have a saved form. Please let me know if additional changes should be made here.

jbalboni: Since we don't always have prefill information, we should conditionally show this new message. There's a `prefillsAvailable` array in the profile that you can check to make sure the current form is in (the render method for this component does that for another reason already).

cehsu: Thanks @jbalboni, change made.

src/js/common/schemaform/SaveInProgressIntro.jsx
```diff
@@ -44,13 +56,25 @@ export default class SaveInProgressIntro extends React.Component {
           </div>
         );
       }
+    } else if (prefillAvailable) {
```
jbalboni: This is going to be false if they're not signed in, so probably can't conditionally show the message here.

cehsu: Thanks @jbalboni, change made.

src/js/common/schemaform/SaveInProgressIntro.jsx
```diff
@@ -49,8 +61,8 @@ export default class SaveInProgressIntro extends React.Component {
         <div>
           <div className="usa-alert usa-alert-info schemaform-sip-alert">
             <div className="usa-alert-body">
-              You can save this form, and come back later to finish filling it out.<br/>
-              <button className="va-button-link" onClick={() => this.props.toggleLoginModal(true)}>Sign in to save your form in progress.</button>
+              You can save this form in progress, and come back later to finish filling it out.<br/>
```
jbalboni: Thinking about this more, it sort of defeats the purpose of the issue to not show the message about prefilling when someone isn't signed in. But we can't check the prefills available list in the profile without being signed in.

So, I guess we need to create a list of forms with prefill available in the front end code? Or add something to the form config to conditionally show this message. Right now it would just show for pension, burials, and hca.

Sorry about going back and forth on this.

cehsu: Great point, thanks for revisiting this. Changes made for option 2, which seems more consistent with how we've handled similar logic on the SIP cards. Not sure what the implications are for the existing `prefillsAvailable` logic?

jbalboni: I think it can stay the way it is. I don't want to force us to make a FE change to turn prefill on for a form if we don't have to. This way, just changing the message requires a front end change.

jbalboni (approved):

peggygannon (changes requested):
@cehsu 
Screen 1: SIP message looks good. No changes
Screen 2: please tweak message to this: Since you're signed in to your account, we can prefill part of your application based on your account details. You can also save your form in progress, and come back later to finish filling it out.

cehsu:
Thanks @peggygannon, changes made. Kindly refer to the updated screenshot in the PR description.

peggygannon (approved):
@cehsu LGTM!

PR #6627, opened 2017-10-18
Actions unit tests mg

markgreenburg:
Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/5190
Also closes https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4961

This PR adds a bunch of unit tests for our async action creators. Also bundled in are a couple of bugfixes for the action creators themselves:
- Primarily, keeping promises alive by returning each `apiRequest` instead of just calling it
- Splitting out some predictable errors for better reporting

These tests are still missing anything for `getLetterPdf()` just because we're still testing some things with how that action creator works.

### DIFF
--- a/src/js/letters/actions/letters.js
+++ b/src/js/letters/actions/letters.js
@@ -75,9 +75,8 @@ export function getMailingAddress() {
       null,
       // on fetch success
       (response) => {
-        const { address } = response.data.attributes;
         const responseCopy = Object.assign({}, response);
-        const addressCopy = Object.assign({}, address);
+        const addressCopy = Object.assign({}, response.data.attributes.address);
         // Translate military-only fields into generic ones; we'll translate
         // them back later if necessary
         if (addressCopy.type === ADDRESS_TYPES.military) {
--- a/src/js/letters/actions/letters.js
+++ b/src/js/letters/actions/letters.js
@@ -36,29 +36,50 @@ export function getLetterList() {
         data: response,
       }),
       (response) => {
-        const error = response.errors.length > 0 ? response.errors[0] : undefined;
-        if (error) {
-          if (error.status === '503' || error.status === '504') {
+        if (typeof response.errors === 'undefined' || response.errors.length === 0) {
+          return Promise.reject(new Error('vets_letters_error_server_get: undefined error'));
+        }
+        const error = response.errors[0];
+        switch (error.status) {
+          case ('503'): // Handled same as 504
+          case ('504'):
             // Either EVSS or a partner service is down or EVSS times out
             return dispatch({ type: BACKEND_SERVICE_ERROR });
-          }
-          if (error.status === '403') {
+          case ('403'):
             // Backend authentication problem
             return dispatch({ type: BACKEND_AUTHENTICATION_ERROR });
-          }
-          if (error.status === '502') {
-            // Some of the partner services are down, so we cannot verify the eligibility
-            // of some letters
+          case ('502'):
+            // Some of the partner services are down, so we cannot verify the
+            // eligibility of some letters
             return dispatch({ type: LETTER_ELIGIBILITY_ERROR });
-          }
-          return Promise.reject(
-            new Error(`vets_letters_error_server_get: error status ${error.status}`)
-          );
+          default:
+            return Promise.reject(
+              new Error(`vets_letters_error_server_get: ${error.status || 'unknown'}`)
+            );
         }
-        return Promise.reject(
-          new Error('vets_letters_error_server_get: unknown error status')
-        );
       }
+      //   if (error) {
+      //     if (error.status === '503' || error.status === '504') {
+      //       // Either EVSS or a partner service is down or EVSS times out
+      //       return dispatch({ type: BACKEND_SERVICE_ERROR });
+      //     }
+      //     if (error.status === '403') {
+      //       // Backend authentication problem
+      //       return dispatch({ type: BACKEND_AUTHENTICATION_ERROR });
+      //     }
+      //     if (error.status === '502') {
+      //       // Some of the partner services are down, so we cannot verify the eligibility
+      //       // of some letters
+      //       return dispatch({ type: LETTER_ELIGIBILITY_ERROR });
+      //     }
+      //     return Promise.reject(
+      //       new Error(`vets_letters_error_server_get: error status ${error.status}`)
+      //     );
+      //   }
+      //   return Promise.reject(
+      //     new Error('vets_letters_error_server_get: unknown error status')
+      //   );
+      // }
     ).catch((error) => {
       if (error.message.match('vets_letters_error_server_get')) {
         Raven.captureException(error);
--- a/src/js/letters/actions/letters.js
+++ b/src/js/letters/actions/letters.js
@@ -166,7 +166,7 @@ export function getLetterPdf(letterType, letterName, letterOptions) {
           }
         });
         window.URL.revokeObjectURL(downloadUrl);
-        dispatch({ type: GET_LETTER_PDF_SUCCESS, data: letterType });
+        return dispatch({ type: GET_LETTER_PDF_SUCCESS, data: letterType });
       },
       () => dispatch({ type: GET_LETTER_PDF_FAILURE, data: letterType })
     );
@@ -228,7 +228,7 @@ export function saveAddress(address) {
 
 export function getAddressCountries() {
   return (dispatch) => {
-    apiRequest(
+    return apiRequest(
       '/v0/address/countries',
       null,
       response => dispatch({
@@ -242,7 +242,7 @@ export function getAddressCountries() {
 
 export function getAddressStates() {
   return (dispatch) => {
-    apiRequest(
+    return apiRequest(
       '/v0/address/states',
       null,
       response => dispatch({
--- a/src/js/letters/actions/letters.js
+++ b/src/js/letters/actions/letters.js
@@ -73,34 +73,29 @@ export function getMailingAddress() {
     apiRequest(
       '/v0/address',
       null,
+      // on fetch success
       (response) => {
-        try {
-          const { address } = response.data.attributes;
-          const responseCopy = Object.assign({}, response);
-          const addressCopy = Object.assign({}, address);
-          // Translate military-only fields into generic ones; we'll translate
-          // them back later if necessary
-          if (addressCopy.type === ADDRESS_TYPES.military) {
-            addressCopy.city = addressCopy.militaryPostOfficeTypeCode;
-            addressCopy.stateCode = addressCopy.militaryStateCode;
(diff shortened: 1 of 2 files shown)


src/js/letters/actions/letters.js
```diff
@@ -36,29 +36,50 @@ export function getLetterList() {
...
+            // eligibility of some letters
             return dispatch({ type: LETTER_ELIGIBILITY_ERROR });
-          }
-          return Promise.reject(
-            new Error(`vets_letters_error_server_get: error status ${error.status}`)
-          );
+          default:
+            return Promise.reject(
+              new Error(`vets_letters_error_server_get: ${error.status || 'unknown'}`)
+            );
```
markgreenburg: Decided to _switch_ this action up a bit 🥁 
Mainly though I noticed when I was making some test cases that there was an unhandled Promise rejection happening when the response had no `errors` property. That's the early return at the top there. The test for this is the 'generic' case below.

src/js/letters/actions/letters.js
```diff
@@ -36,29 +36,50 @@ export function getLetterList() {
...
       (response) => {
-        const error = response.errors.length > 0 ? response.errors[0] : undefined;
-        if (error) {
-          if (error.status === '503' || error.status === '504') {
+        if (typeof response.errors === 'undefined' || response.errors.length === 0) {
+          return Promise.reject(new Error('vets_letters_error_server_get: undefined error'));
+        }
+        const error = response.errors[0];
+        switch (error.status) {
+          case ('503'): // Handled same as 504
```
cvalarida: Don't think we need the parens.

src/js/letters/actions/letters.js
```diff
@@ -36,29 +36,50 @@ export function getLetterList() {
...
+      //       return dispatch({ type: LETTER_ELIGIBILITY_ERROR });
+      //     }
+      //     return Promise.reject(
+      //       new Error(`vets_letters_error_server_get: error status ${error.status}`)
+      //     );
+      //   }
+      //   return Promise.reject(
+      //     new Error('vets_letters_error_server_get: unknown error status')
+      //   );
+      // }
```
cvalarida: Looks good! Don't need this comment block. 👍

test/letters/actions/letters.unit.spec.js
```diff
@@ -21,6 +36,208 @@ const frontEndAddress = {
...
+    const dispatch = sinon.spy((action) => {
+      // keep track of which dispatch() call we're asserting against
+      callCount += 1;
+      const { type } = action;
+      if (callCount === 1) {
+        try {
+          expect(type).to.equal(SAVE_ADDRESS_PENDING);
+          done();
+        } catch (error) {
+          done(error);
```
cvalarida: Is this necessary? I think we can just do an `expect()` then `done()`; if `expect()` throws an error, isn't it caught by mocha? I mean, this should work just as well, but I think we don't need to try / catch it.

I could be wrong, though. :man_shrugging:

test/letters/actions/letters.unit.spec.js
```diff
@@ -21,6 +36,208 @@ const frontEndAddress = {
...
+    thunk(dispatch, getState);
+  });
+
+  it('dispatches SAVE_ADDRESS_SUCCESS on update success', (done) => {
+    const thunk = saveAddress(frontEndAddress);
+    const dispatch = sinon.spy((action) => {
+      const { type, address } = action;
+      if (type === SAVE_ADDRESS_SUCCESS) {
+        try {
+          expect(type).to.equal(SAVE_ADDRESS_SUCCESS);
```
cvalarida: At this point, we already know this is going to pass, so there's no need for this assertion. The trick here is to catch if `dispatch` isn't called with the success action.

🤔  I think it's fine because `done` is never called if the action is never `SAVE_ADDRESS_SUCCESS`, and the test will timeout. The error message won't be super useful, but it should be obvious after we look at the test, though.

Alright, after puzzling this out, I'm okay with this. 👍

test/letters/actions/letters.unit.spec.js
```diff
@@ -21,6 +36,208 @@ const frontEndAddress = {
...
+  });
+
+  it('dispatches SAVE_ADDRESS_FAILURE on update failure', (done) => {
+    global.fetch.returns(Promise.reject(new Error('something went wrong')));
+    const thunk = saveAddress(frontEndAddress);
+    let callCount = 0;
+    const dispatch = sinon.spy((action) => {
+      const { type } = action;
+      callCount += 1;
+      if (callCount === 2) {
```
cvalarida: This could get brittle...but it should get caught right away when we change the action, so that's probably okay.

cvalarida (commented):
Yeah, I think you're on the right track. I'm not super fond of the way we have to embed the `expect`s in the promise resolution callback, but it works. I don't have a better solution at the tip of my fingers.

So far so good!

src/js/letters/actions/letters.js
```diff
@@ -74,26 +73,34 @@ export function getMailingAddress() {
...
-        const address = Object.assign({}, response.data.attributes.address);
-        // Translate military-only fields into generic ones; we'll translate them back later if necessary
-        if (address.type === ADDRESS_TYPES.military) {
-          address.city = address.militaryPostOfficeTypeCode;
-          address.stateCode = address.militaryStateCode;
-          address.countryName = 'USA';
-          delete address.militaryPostOfficeTypeCode;
-          delete address.militaryStateCode;
+      (response) => {
+        try {
```
markgreenburg: There are a few things that could go wrong in here, especially if the response doesn't have the shape we think it does (came up a bunch when I was writing tests), so thought it'd be safer to wrap the logic here and reject the promise in the catch.

src/js/letters/actions/letters.js
markgreenburg: ^ the unhandled promise rejection should be fixed now that `apiRequest()` is being returned from the thunk instead of just called, but I still think it's nicer to break out this error.

src/js/letters/actions/letters.js
```diff
@@ -76,35 +75,37 @@ export function getAddressFailure() {
...
         const responseCopy = Object.assign({}, response);
-        const address = Object.assign({}, response.data.attributes.address);
-        // Translate military-only fields into generic ones; we'll translate them back later if necessary
-        if (address.type === ADDRESS_TYPES.military) {
-          address.city = address.militaryPostOfficeTypeCode;
-          address.stateCode = address.militaryStateCode;
-          address.countryName = 'USA';
-          delete address.militaryPostOfficeTypeCode;
-          delete address.militaryStateCode;
+        const addressCopy = Object.assign({}, response.data.attributes.address);
```
markgreenburg: renaming address to addressCopy makes this whole block look new, but that's all that's going on here.

src/js/letters/actions/letters.js
```diff
@@ -76,35 +75,37 @@ export function getAddressFailure() {
...
+          addressCopy.city = addressCopy.militaryPostOfficeTypeCode;
+          addressCopy.stateCode = addressCopy.militaryStateCode;
+          addressCopy.countryName = 'USA';
+          delete addressCopy.militaryPostOfficeTypeCode;
+          delete addressCopy.militaryStateCode;
         }
-        responseCopy.data.attributes.address = address;
-
-        dispatch({
+        responseCopy.data.attributes.address = addressCopy;
```
cvalarida: Eh, because  `responseCopy` is only a shallow copy (with `Object.assign()`), this technically isn't necessary, but it does read a lot nicer with it.

test/letters/actions/letters.unit.spec.js
```diff
@@ -1,86 +1,495 @@
...
+    const dispatch = sinon.spy();
+    thunk(dispatch, getState)
+      .then(() => {
+        const action = dispatch.firstCall.args[0];
+        const { address } = action.data.data.attributes; // Actual
+        const { militaryPostOfficeTypeCode, militaryStateCode } = militaryAddress; // Test
+        expect(address.city).to.equal(militaryPostOfficeTypeCode);
+        expect(address.stateCode).to.equal(militaryStateCode);
+        expect(address.militaryPostOfficeTypeCode).to.be.undefined;
+        expect(address.militaryStateCode).to.be.undefined;
```
cvalarida: Excellent! In your PR comment, you might mention that this closes https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4961

test/letters/actions/letters.unit.spec.js
```diff
@@ -1,86 +1,495 @@
...
+      id: null,
+      type: 'evss_letters_letters_response'
+    }
+  };
+
+  it('dispatches GET_LETTERS_SUCCESS when GET succeeds', (done) => {
+    global.fetch.returns(Promise.resolve({
+      headers: { get: () => 'application/json' },
+      ok: true,
+      json: () => Promise.resolve(lettersResponse)
```
cvalarida: 🤔  We can't write a helper function to mock `fetch`, but we can make a helper function to mock what `fetch` _returns_.

Something like:
```js
function craftApiResponse(json, ok = true) {
  const resolveOrReject = ok ? Promise.resolve : Promise.reject;
  return resolveOrReject({
    headers: { get: () => 'application/json' },
    ok,
    json: () => resolveOrReject(json)
  });
}

...

global.fetch.returns(craftApiResponse(lettersResponse));
```

test/letters/actions/letters.unit.spec.js
```diff
@@ -1,86 +1,495 @@
...
+    data: {
+      attributes: {
+        address: {
+          type: 'DOMESTIC',
+          addressEffectiveDate: '1973-01-01T05:00:00.000+00:00',
+          addressOne: '140 Rock Creek Church Rd NW',
+          addressTwo: '',
+          addressThree: '',
+          city: 'Washington',
+          stateCode: 'DC',
```
cvalarida: 🤔 :man_shrugging:

cvalarida (approved):
Phew! We can go ahead and get this merged in before tackling the `getLetterPdf` action creator, if you like. This is getting pretty lengthy and we can split that up.

test/letters/actions/letters.unit.spec.js
markgreenburg: Yeah, I stole this from the `vets-api` Swagger

test/letters/actions/letters.unit.spec.js
markgreenburg: Oh right, nice.

test/letters/actions/letters.unit.spec.js
markgreenburg: This makes total sense, but was running into a strange issue with my implementation so decided to push this off to another PR that will also probably address pdf download tests.

PR #6628, opened 2017-10-18
Fix address line autocomplete names

annekainicUSDS:
Changed the `autocomplete` attributes values for the address lines to be in line with what Google and Mozilla recommend for multiple line addresses:
1. https://developers.google.com/web/updates/2015/06/checkout-faster-with-autofill
2. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input

Also changed the `name` attributes, since those were also all the same, to correspond with the name of the field we get from the response.

### DIFF
--- a/test/letters/reducers/index.unit.spec.js
+++ b/test/letters/reducers/index.unit.spec.js
@@ -174,7 +174,7 @@ describe('letters reducer', () => {
     const state = reduce({ type: GET_ADDRESS_FAILURE });
 
     expect(state.address).to.be.empty;
-    expect(state.addressAvailable).to.be.false;
+    expect(state.addressAvailability).to.equal(AVAILABILITY_STATUSES.unavailable);
   });
 
   it('should handle successful request for the countries', () => {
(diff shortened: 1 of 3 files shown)


cvalarida (approved):

test/letters/reducers/index.unit.spec.js
```diff
@@ -174,7 +174,7 @@ describe('letters reducer', () => {
     const state = reduce({ type: GET_ADDRESS_FAILURE });
 
     expect(state.address).to.be.empty;
-    expect(state.addressAvailable).to.be.false;
+    expect(state.addressAvailability).to.equal(AVAILABILITY_STATUSES.unavailable);
```
annekainicUSDS: Would love @cvalarida or @markgreenburg to take a second look at this, was getting a test failure on this test after rebasing on the latest from master. I think the issue is that the value that actually stores the result of the address fetch is `addressAvailability`, not `addressAvailable`, and it's not stored as a boolean but as an option from `AVAILABILITY_STATUSES`. Not sure how this test passed before, but I think this should fix it!

markgreenburg: Well that's super weird . . . the old test should have failed the strict equality check against `false`, right? Either way, looks like this test just didn't get updated (probably because it passed anyway! What? 😕 ) in the [PR that made these changes](https://github.com/department-of-veterans-affairs/vets-website/commit/43adf4c9c5a1146849de2f7b5608ef8de0a6fade#diff-3bdff36195cc2f3072a099de8cfe6e2bL84) initially.

LGTM!

markgreenburg (approved):
LGTM

PR #6635, opened 2017-10-19
Change menu item to CareerScope

ncksllvn:
Per Beth:

> Change the language for one of the links in the right-side call-out, under the header "What's your plan?" where the video is. "Explore your career" needs to say: "Get started with CareerScope." This is a SME request.


bshyong (approved):
LGTM

PR #6640, opened 2017-10-19
Turn on key inflection for SiP

jbalboni:
Previously, we could not turn this on on the backend, but now we can, and it also fixes an issue where the prefilled postal code for HCA is `postal_code` instead of `postalCode`.


src/js/common/schemaform/save-load-actions.js
```diff
@@ -213,7 +213,7 @@ export function fetchInProgressForm(formId, migrations, prefill = false) {
       // TODO: These headers should work, but trigger an api error right now
```
cvalarida: This was in reference to the key inflection header; it can be removed now.

cvalarida (approved):
:shipit:

cehsu (approved):

PR #6646, opened 2017-10-20
Add custom events for address updating

robertfairhead:
Closes https://github.com/department-of-veterans-affairs/vets.gov-team/issues/5450

Take a look at the `getAddressFailure` function in `letters.js` in particular since I refactor that bit out.

### DIFF
--- a/src/js/letters/containers/AddressSection.jsx
+++ b/src/js/letters/containers/AddressSection.jsx
@@ -136,13 +136,11 @@ export class AddressSection extends React.Component {
       return;
     }
 
-
     this.setState({
       isEditingAddress: false,
       // Reset all the error messages in case they go to edit again; should be pointless
       errorMessages: {}
     });
-    window.dataLayer.push({ event: 'letter-update-address-submit' });
     this.props.saveAddress(this.state.editableAddress);
   }
 
--- a/src/js/letters/containers/AddressSection.jsx
+++ b/src/js/letters/containers/AddressSection.jsx
@@ -127,7 +127,7 @@ export class AddressSection extends React.Component {
       // If we had any errors, make sure to validate them from here on out so we don't loose them
       const fieldsToValidate = Object.assign({}, this.state.fieldsToValidate);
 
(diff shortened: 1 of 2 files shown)


src/js/letters/containers/AddressSection.jsx
```diff
@@ -142,10 +142,12 @@ export class AddressSection extends React.Component {
       // Reset all the error messages in case they go to edit again; should be pointless
       errorMessages: {}
     });
+    window.dataLayer.push({ event: 'letter-update-address-submit' });
```
markgreenburg: For consistency, you could probably throw this into the `saveAddressPending` action creator like you did with `success` and `failure`. Is there a reason why you opted to put this one in the component instead?

src/js/letters/containers/AddressSection.jsx
```diff
@@ -142,10 +142,12 @@ export class AddressSection extends React.Component {
...
     this.props.saveAddress(this.state.editableAddress);
   }
 
   handleCancel = () => {
+    window.dataLayer.push({ event: 'letter-update-address-cancel' });
```
markgreenburg: So I guess the intent with this and `'letter-update-address-started'` is to log each time a user hits edit or cancel, understanding that there may be several of these per user? If so, 👍

src/js/letters/containers/AddressSection.jsx
```diff
@@ -154,6 +156,11 @@ export class AddressSection extends React.Component {
     });
   }
 
+  startEditing = () => {
```
markgreenburg: 👍

markgreenburg (approved):
LGTM! Also checked out this branch and ran locally without any problems.

src/js/letters/containers/AddressSection.jsx
robertfairhead: Yeah fair enough, pushed a commit to move it.

I added that one first and didn't know i'd be touching actions at the time. But that makes sense.

PR #6650, opened 2017-10-20
Fix IE download...hopefully

cvalarida:
I tried really hard to get a local dev environment accessible to my Windows machine, but I could neither successfully serve from my Macbook to the local network (couldn't log in because of reasons ID.me) nor get vets-website to run on Windows for strange permission issues (`npm install` failed because it ran into an illegal operation--said it was due to permissions, but I was running the git cli as Administrator and there is no `sudo` in that environment).

This PR is based on me hacking around in IE's console to see what _should_ happen. If this still doesn't work, I'll have to take @kreek's suggestion and map a remote file to a local one with a tool he recommended so I can _actually_ debug the silly thing.

### DIFF
--- a/src/js/letters/actions/letters.js
+++ b/src/js/letters/actions/letters.js
@@ -130,29 +130,24 @@ export function getLetterPdf(letterType, letterName, letterOptions) {
     };
   }
 
-  // We handle IE10 separately but assume all other vets.gov-supported
-  // browsers have blob URL support.
-  // TODO: possibly want to explicitly check for blob URL support with something like
-  // const blobSupported = !!(/^blob:/.exec(downloadUrl));
-  const ie10 = !!window.navigator.msSaveOrOpenBlob;
-  const save = document.createElement('a');
-  let downloadWindow;
-  const downloadSupported = typeof save.download !== 'undefined';
-  if (!downloadSupported) {
-    // Instead of giving the file a readable name and downloading
-    // it directly, open it in a new window with an ugly hash URL
-    downloadWindow = window.open();
-  }
-  let downloadUrl;
   return (dispatch) => {
     dispatch({ type: GET_LETTER_PDF_DOWNLOADING, data: letterType });
     apiRequest(
       `/v0/letters/${letterType}`,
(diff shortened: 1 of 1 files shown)


src/js/letters/actions/letters.js
```diff
@@ -165,6 +160,9 @@ export function getLetterPdf(letterType, letterName, letterOptions) {
               save.click();
               document.body.removeChild(save);
             } else {
+              // Instead of giving the file a readable name and downloading
+              // it directly, open it in a new window with an ugly hash URL
+              const downloadWindow = window.open();
```
cvalarida: Previously, we'd set `ie10` to true _and_ `downloadSupported` to false, so we'd open a window and do nothing with it.

src/js/letters/actions/letters.js
```diff
@@ -130,29 +130,24 @@ export function getLetterPdf(letterType, letterName, letterOptions) {
...
+          // browsers have blob URL support.
+          // TODO: possibly want to explicitly check for blob URL support with something like
+          // const blobSupported = !!(/^blob:/.exec(downloadUrl));
+          const ie10 = !!window.navigator.msSaveOrOpenBlob;
+          const save = document.createElement('a');
+          const downloadSupported = typeof save.download !== 'undefined';
+
           if (ie10) {
-            window.navigator.msSaveOrOpenBlob(blob, letterName);
+            window.navigator.msSaveOrOpenBlob(blob, `${letterName}.pdf`);
```
cvalarida: :crossed_fingers: Hopefully this works; need to test it. The file previously downloaded just fine, but without the extension. Manually adding the extension made it work like a charm, so hopefully this does the trick.

kreek (approved):
LGTM 🚀

raquelromano-gov (approved):
Worth a try--I can no longer remember why I had pulled that logic up to the top, but what you have here makes sense, and thanks for the monumental effort to get your windows machine to talk to your dev environment...let's do some aggressive manual testing on staging on a bunch of platform/browser combos once this is merged.

PR #6653, opened 2017-10-23
Switch from "require" syntax to "import" where possible

ncksllvn:
This PR switches from `require()` to `import` where possible. Exceptions include conditional-requires, because I don't believe that's possible yet with `import`. For consistency I also moved imports that do not contain an export to the top of each file.

Resolves department-of-veterans-affairs/vets.gov-team/issues/2944

### DIFF
--- a/src/js/health-beta/health-beta-entry.jsx
+++ b/src/js/health-beta/health-beta-entry.jsx
@@ -11,9 +11,9 @@ import reducer from './reducers';
 import createCommonStore from '../common/store';
 import createLoginWidget from '../login/login-entry';
 
-require('../common');  // Bring in the common javascript.
-require('../../sass/rx/rx.scss');
-require('../../sass/user-profile.scss');
+import '../common';  // Bring in the common javascript.
+import '../../sass/rx/rx.scss';
+import '../../sass/user-profile.scss';
 
 const commonStore = createCommonStore(reducer);
 createLoginWidget(commonStore);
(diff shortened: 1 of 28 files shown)


ncksllvn:
Ugh, Code Climate is giving an error about repetitive code. Weren't we talking about removing that?

jbalboni:
It shouldn't block merging, once someone approves this PR

cvalarida (approved):
I'll admit my ignorance of how `import`ing SCSS works, but assuming you tested that manually (or using visual regression testing!) and did a search for `require(` in all the JS, I'd call this good!

jbalboni (approved):

src/js/health-beta/health-beta-entry.jsx
```diff
@@ -1,4 +1,8 @@
 import 'core-js';
+import '../common';
+import '../../sass/rx/rx.scss';
+import '../../sass/user-profile.scss';
```
U-DON: I don't think styles need to be imported for this app.

U-DON (approved):
LGTM.

src/js/health-beta/health-beta-entry.jsx
ncksllvn: So I can safely remove those 3 lines?

ncksllvn:
Thanks @cvalarida, @jbalboni, and @U-DON!

src/js/health-beta/health-beta-entry.jsx
U-DON: Yeah, I don't think those two lines are used.

ncksllvn: Ah I would have removed it but I'm gonna leave it to whoever maintains that just in case there are some styles being used out of there for a modal or something. I'm not familiar with that app at all.

PR #6657, opened 2017-10-23
Edit address if 'empty'

cvalarida:
This checks to make sure that if the address is somehow "empty," we start editing. I tested it by overriding the `address` in `getMailingAddress()` after I got a response.

### DIFF
--- a/src/js/letters/containers/AddressSection.jsx
+++ b/src/js/letters/containers/AddressSection.jsx
@@ -1,6 +1,5 @@
 import React from 'react';
 import { connect } from 'react-redux';
-import isEmpty from 'lodash/isEmpty';
 
 import { scrollToFirstError } from '../../common/utils/helpers';
 import LoadingIndicator from '../../common/components/LoadingIndicator';
@@ -29,6 +28,12 @@ import {
 } from '../utils/validations';
 import { AVAILABILITY_STATUSES } from '../utils/constants';
 
+// The address is empty if every field except type is falsey.
+// NOTE: It "shouldn't" ever happen, but it did in testing, so...paranoid programming!
+function isAddressEmpty(address) {
+  return Object.keys(address).reduce((emptySoFar, nextField) => emptySoFar && (nextField === 'type' || !address[nextField]), true);
+}
+
 export class AddressSection extends React.Component {
   constructor(props) {
     super(props);
@@ -45,6 +50,8 @@ export class AddressSection extends React.Component {
     // savedAddress values
     if (Object.keys(this.state.editableAddress).length > 0) {
       this.state.hasLoadedAddress = true;
+      // If we start with an empty address, go straight to editing
+      this.state.isEditingAddress = isAddressEmpty(this.state.editableAddress);
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/js/letters/containers/AddressSection.jsx
```diff
@@ -29,6 +28,12 @@ import {
 } from '../utils/validations';
 import { AVAILABILITY_STATUSES } from '../utils/constants';
 
+// The address is empty if every field except type is falsey.
+// NOTE: It "shouldn't" ever happen, but it did in testing, so...paranoid programming!
+function isAddressEmpty(address) {
+  return Object.keys(address).reduce((emptySoFar, nextField) => emptySoFar && (nextField === 'type' || !address[nextField]), true);
+}
```
markgreenburg: Think I'm reviewing this too late in the day because this took longer to understand than I care to admit 🐢

src/js/letters/containers/AddressSection.jsx
```diff
@@ -45,6 +50,8 @@ export class AddressSection extends React.Component {
     // savedAddress values
     if (Object.keys(this.state.editableAddress).length > 0) {
       this.state.hasLoadedAddress = true;
+      // If we start with an empty address, go straight to editing
+      this.state.isEditingAddress = isAddressEmpty(this.state.editableAddress);
```
markgreenburg: Don't think this is a huge deal, but I do get some UI thrashing with this approach - at first I see the 'Address Update Not Available' warning, then the editor. Takes a second or two to switch over. This is probably because the full states and countries list isn't available right away (Address Section L298), but we probably don't need to care about that since this is already an edge case.

markgreenburg (approved):
LGTM!

test/letters/containers/AddressSection.unit.spec.jsx
```diff
@@ -46,6 +46,9 @@ const defaultProps = {
   statesAvailable: true
 };
 
+// For running ./node_modules/.bin/mocha directly on this file
```
annekainicUSDS: what is doing this for?

annekainicUSDS (approved):
Looks great!

src/js/letters/containers/AddressSection.jsx
cvalarida: Yeaaaaaaah, it's not _super_ clear, but the alternative was breaking the callback onto multiple lines. If that'll make it clearer, I'd be happy to do that. I wasn't thrilled with the length of this line. 0.o

src/js/letters/containers/AddressSection.jsx
cvalarida: I don't think that's because of this, actually. We'll get that warning when `this.props.addressAvailability === AVAILABILITY_STATUSES.unavailable`, so it's just the lag between when this component mounts and when we get the response back from the api to set `addressAvailability` to `available`.

We should probably write a ticket for that, but it's not anything new here.

test/letters/containers/AddressSection.unit.spec.jsx
cvalarida: I prefer running tests using the command `./node_modules/.bin/mocha test/letters/blah.unit.spec.js` instead of adding (and potentially forgetting to delete) an `.only` to the `describe` and running `npm run test:unit`.

The trouble I had, though, is about half the tests failed because `global.window.dataLayer` was `undefined`. Adding this line in here doesn't actually change anything when running `npm run test:unit`, but it did let me run `./node_modules/.bin/mocha ...`.

PR #6659, opened 2017-10-23
Make download button show for all letters

markgreenburg:
PR updates conditions under which a Download Letter link should render and adds associated test.

Basically, a download link should be included for all letters that are in the letters array, regardless of whether `optionsAvailable` for BSL is `true` or `false`.

Also took the opportunity to extract a snippet of static helper text to `helpers` module to tighten things up a bit.

### DIFF
--- a/src/js/letters/containers/LetterList.jsx
+++ b/src/js/letters/containers/LetterList.jsx
@@ -15,9 +15,11 @@ export class LetterList extends React.Component {
     const letterItems = (this.props.letters || []).map((letter, index) => {
       let content;
       let letterTitle;
+      let helpText;
       if (letter.letterType === LETTER_TYPES.benefitSummary) {
         letterTitle = 'Benefit Summary and Service Verification Letter';
         content = (<VeteranBenefitSummaryLetter/>);
+        helpText = bslHelpInstructions;
       } else if (letter.letterType === LETTER_TYPES.proofOfService) {
         letterTitle = 'Proof of Service Card';
         content = letterContent[letter.letterType] || '';
@@ -26,19 +28,22 @@ export class LetterList extends React.Component {
         content = letterContent[letter.letterType] || '';
       }
 
+      let conditionalDownloadButton;
+      if (letter.letterType !== LETTER_TYPES.benefitSummary || this.props.optionsAvailable) {
+        conditionalDownloadButton = (<DownloadLetterLink
+          letterType={letter.letterType}
+          letterName={letter.name}
+          downloadStatus={downloadStatus[letter.letterType]}
+          key={`download-link-${index}`}/>);
+      }
+
       return (
         <CollapsiblePanel
           panelName={letterTitle}
           key={`collapsiblePanel-${index}`}>
           <div>{content}</div>
-          <DownloadLetterLink
-            letterType={letter.letterType}
-            letterName={letter.name}
-            downloadStatus={downloadStatus[letter.letterType]}
-            key={`download-link-${index}`}/>
-          { letter.letterType === LETTER_TYPES.benefitSummary &&
-            bslHelpInstructions
-          }
+          {conditionalDownloadButton}
+          {helpText}
         </CollapsiblePanel>
       );
     });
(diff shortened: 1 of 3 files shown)


src/js/letters/containers/LetterList.jsx
```diff
@@ -34,24 +26,19 @@ export class LetterList extends React.Component {
         content = letterContent[letter.letterType] || '';
       }
 
-      let downloadLetterLink;
-      if (this.props.optionsAvailable) {
```
annekainicUSDS: So this logic was added in this PR: https://github.com/department-of-veterans-affairs/vets-website/pull/6608. I think the mistake I made was to hide the download button for ALL letters if options aren't available. However, the logic should still exist that if `optionsAvailable === false`, the download button should not appear for the BSL. That's because I checked with CSRA and they said if the options aren't available the letter won't be able to generate, so we might as well hide the button. So I think we need to still account for that here. So the logic should be something like:
```
if (letter.letterType !== LETTER_TYPES.benefitSummary || this.props.optionsAvailable) {
  downloadLetterLink = (<DownloadLetterLink ... />);
}
```

src/js/letters/containers/LetterList.jsx
```diff
@@ -34,24 +26,19 @@ export class LetterList extends React.Component {
...
           key={`collapsiblePanel-${index}`}>
           <div>{content}</div>
-          <div>{downloadLetterLink}</div>
-          <div>{bslHelpInstructions}</div>
+          <DownloadLetterLink
+            letterType={letter.letterType}
+            letterName={letter.name}
+            downloadStatus={downloadStatus[letter.letterType]}
+            key={`download-link-${index}`}/>
+          { letter.letterType === LETTER_TYPES.benefitSummary &&
```
annekainicUSDS: I guess this is a style difference, but I liked how before all the checks for the types of letters and the condition logic that rendered because of that was in one place up above. Now we're checking in 2 places if `letter.letterType === LETTER_TYPES.benefitSummary`. And is that slightly more efficient because we're only doing that check once? Thoughts?

test/letters/containers/LetterList.unit.spec.jsx
```diff
@@ -25,7 +25,7 @@ const defaultProps = {
   optionsAvailable: true,
 };
 
-describe('<LetterList>', () => {
+describe.only('<LetterList>', () => {
```
annekainicUSDS: I think you want to delete this `only`

test/letters/containers/LetterList.unit.spec.jsx
```diff
@@ -46,6 +46,21 @@ describe('<LetterList>', () => {
     });
   });
 
+  it('renders DL buttons for all letters in list', () => {
```
annekainicUSDS: Eventually, once the logic is fixed up above, I think we'll want another test here that makes sure if the letter is BSL and options aren't available that the download button doesn't render.

annekainicUSDS (commented):
A few comments!

src/js/letters/containers/LetterList.jsx
markgreenburg: That's fine; when deciding how to fix this I reached out to Tim Barto and was told that it was not possible to get a `benefit_summary` letter returned from `v0/letters` and *not* get any available `benefit_information` options returned by `v0/letters/beneficiary`, which is why I removed this bit entirely. It sounds like you've run into this before though, so the check for the BSL at least should go back in.

test/letters/containers/LetterList.unit.spec.jsx
markgreenburg: For sure!

test/letters/containers/LetterList.unit.spec.jsx
markgreenburg: Agreed, will write one up

src/js/letters/containers/LetterList.jsx
annekainicUSDS: Ah gotcha, I think I had either gotten slightly different info or misinterpreted what he said. Either way, I think this is a good safeguard just in case we get some unexpected data from them!

annekainicUSDS (approved):
LGTM!

markgreenburg:
Thanks for the review @annekainicUSDS !

PR #6660, opened 2017-10-23
Prevent double triggering of onclick for Modals

bshyong:
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/5250


U-DON (approved):
LGTM.

PR #6661, opened 2017-10-24
Add visual regression testing to readme

ncksllvn:
This is a super minimal addition to the readme about VRT. I also opened a PR to add an engineering doc. It contains more information about the road ahead. 

department-of-veterans-affairs/vets.gov-team/pull/5525

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,8 @@ very secret.
 | run only javascript linter               | `npm run lint:js`                        |
 | run only sass linter                     | `npm run lint:sass`                      |
 | run automated accessibility tests        | `npm run build && npm run test:accessibility` |
-| test for broken links                    | Build the site. Broken Link Checking is done via a Metalsmith plugin during build. Note that it only runs on *build* not watch. |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -250,6 +251,11 @@ and you should see:
 ```
 * Selenium requires **Java 8** to run
 
+### Visual Regression Testing
+This is the first iteration of visual regression testing. It is useful to detect side effects or scope of visual changes.
+
+VRT works by gathering the links for the site using the sitemap, then using Nightwatch to navigate throughout the site, capturing an image of each page that will either be used as the baseline for future comparisons or compared to the baseline. The developer must first create the baseline image set for comparisons (sometimes called the golden set), then after making their changes, run an additional task to execute the comparison.
+
```
cvalarida: Maybe add a note about it not delving deep into apps, so just because our VRT didn't catch a change doesn't mean there isn't one nested in some specific UI state. User discretion advised.

cvalarida (approved):

README.md
ncksllvn: Good idea, will add that in a sec. Thanks man!

jbalboni (approved):

PR #6670, opened 2017-10-25
Download letters properly in Safari <= 10

cvalarida:
Fun fact: Safari won't let you `window.open()` unless it's a direct result of a user interaction. This means we need to open the new window _before_ the API callback for it to work.

I'm guessing that's why you had that pulled out before, @raquelromano-gov.

### DIFF
--- a/src/js/letters/actions/letters.js
+++ b/src/js/letters/actions/letters.js
@@ -133,21 +133,31 @@ export function getLetterPdf(letterType, letterName, letterOptions) {
 
   return (dispatch) => {
     dispatch({ type: GET_LETTER_PDF_DOWNLOADING, data: letterType });
+
+    // We handle IE10 separately but assume all other vets.gov-supported
+    // browsers have blob URL support.
+    // TODO: possibly want to explicitly check for blob URL support with something like
+    // const blobSupported = !!(/^blob:/.exec(downloadUrl));
+    const isIE = !!window.navigator.msSaveOrOpenBlob;
+    const save = document.createElement('a');
+    const downloadSupported = typeof save.download !== 'undefined';
+    let downloadWindow;
+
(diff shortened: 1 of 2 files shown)


src/js/letters/actions/letters.js
```diff
@@ -133,21 +133,31 @@ export function getLetterPdf(letterType, letterName, letterOptions) {
...
+    // const blobSupported = !!(/^blob:/.exec(downloadUrl));
+    const isIE = !!window.navigator.msSaveOrOpenBlob;
+    const save = document.createElement('a');
+    const downloadSupported = typeof save.download !== 'undefined';
+    let downloadWindow;
+
+    if (!downloadSupported) {
+      // Instead of giving the file a readable name and downloading
+      // it directly, open it in a new window with an ugly hash URL
+      // NOTE: We're opening the window here because Safari won't open
```
raquelromano-gov: Yeah, I think this may be the case for more browsers than just Safari <= 10 (can't open a window unless it's in a handler for a user action) but we just don't notice it because if `download` is supported, we don't exercise this code.

raquelromano-gov (approved):

src/js/letters/actions/letters.js
cvalarida: Yup, exactly. All Safari does the `window.open()` prevention, but only Safari <= 10 doesn't support `download`. I don't know about other browsers, though.

raquelromano-gov: I had to fix this once for our login modals and found the top answer to this stackoverflow question a little helpful re: older browsers definitely not allowing windows to be opened if not a direct consequence of user action, and newer browsers being a bit more lenient:
https://stackoverflow.com/questions/2587677/avoid-browser-popup-blockers

PR #6672, opened 2017-10-25
LettersApp unit tests

cvalarida:
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/5203

I left `RequiredLoginView` alone because we already test that it works elsewhere. instead, I just made sure we were rendering `AppContent` and then tested that _that_ did what we expect.

### DIFF
--- a/test/letters/containers/LettersApp.unit.spec.jsx
+++ b/test/letters/containers/LettersApp.unit.spec.jsx
@@ -2,18 +2,57 @@ import React from 'react';
 import SkinDeep from 'skin-deep';
 import { expect } from 'chai';
 
-import LettersApp from '../../../src/js/letters/containers/LettersApp';
+import { LettersApp, AppContent } from '../../../src/js/letters/containers/LettersApp';
 
 import reducer from '../../../src/js/letters/reducers';
 import createCommonStore from '../../../src/js/common/store';
 
 const store = createCommonStore(reducer);
 
+const mockRoutes = [{ path: '/fake' }];
+const defaultProps = {
+  store,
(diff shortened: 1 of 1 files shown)


test/letters/containers/LettersApp.unit.spec.jsx
```diff
@@ -2,18 +2,57 @@ import React from 'react';
...
 describe('<LettersApp>', () => {
-  it('should render', () => {
-    const mockRoutes = [{ path: '/fake' }];
-    const tree = SkinDeep.shallowRender(<LettersApp store={store} location={{ pathname: '/blah' }} route={{ childRoutes: mockRoutes }}/>);
-    const vdom = tree.getRenderOutput();
-    expect(vdom).to.not.be.undefined;
+  it('should render AppContent', () => {
+    const tree = SkinDeep.shallowRender(<LettersApp {...defaultProps}/>);
+    // div() throws an exception if it can't find the selector
+    expect(tree.dive(['AppContent'])).to.not.be.null;
```
markgreenburg: Since `dive()` throws if it doesn't find the selector, would something like `expect(() => tree.dive(['AppContent'])).to.not.throw()` be more closely aligned with the assertion? Probably not worth messing with.

markgreenburg (approved):
LGTM!

test/letters/containers/LettersApp.unit.spec.jsx
cvalarida: Yeah, it's debatable, honestly. If it does throw, the test will fail, so I did it this way because I think it's a little more obvious what we're actually testing.

PR #6674, opened 2017-10-25
Letters use new address mg

markgreenburg:
- Switch to using address returned from API in the UI after a successful address update (involved some transforming of the response back into the shape required by the front end)
- Log any mismatches between PUT and response addresses to Sentry
- Update related tests

Also successfully tested all update permutations manually against CI (M -> D; M -> I; M -> M; etc...)

### DIFF
--- a/src/js/letters/actions/letters.js
+++ b/src/js/letters/actions/letters.js
@@ -1,7 +1,7 @@
 import Raven from 'raven-js';
 import { isEqual } from 'lodash';
 
-import { apiRequest } from '../utils/helpers.jsx';
+import { apiRequest, stripEmpties } from '../utils/helpers.jsx';
 import {
   ADDRESS_TYPES,
   BACKEND_AUTHENTICATION_ERROR,
@@ -242,11 +242,18 @@ export function saveAddress(address) {
       '/v0/address',
       settings,
       (response) => {
+        // translate military addresses back to front end address
         const responseAddress = Object.assign({}, response.data.attributes.address);
-        if (!isEqual(address, responseAddress)) {
+        if (transformedAddress.type === ADDRESS_TYPES.military) {
+          responseAddress.city = responseAddress.militaryPostOfficeTypeCode;
+          responseAddress.stateCode = responseAddress.militaryStateCode;
+          responseAddress.countryName = 'USA';
+          delete responseAddress.militaryPostOfficeTypeCode;
+          delete responseAddress.militaryStateCode;
+        }
+        if (!isEqual(stripEmpties(address), stripEmpties(responseAddress))) {
           const mismatchError = new Error('letters-address-update addresses don\'t match');
           Raven.captureException(mismatchError, { submitted: address, returned: responseAddress });
-          return Promise.reject(mismatchError);
         }
         return dispatch(saveAddressSuccess(responseAddress));
       },
--- a/src/js/letters/actions/letters.js
+++ b/src/js/letters/actions/letters.js
@@ -237,7 +237,7 @@ export function saveAddress(address) {
         const responseAddress = militaryToGeneric(response.data.attributes.address);
         if (!isEqual(stripEmpties(address), stripEmpties(responseAddress))) {
           const mismatchError = new Error('letters-address-update addresses don\'t match');
-          Raven.captureException(mismatchError, { submitted: address, returned: responseAddress });
+          Raven.captureException(mismatchError);
         }
         return dispatch(saveAddressSuccess(responseAddress));
       },
--- a/src/js/letters/actions/letters.js
+++ b/src/js/letters/actions/letters.js
@@ -83,17 +83,8 @@ export function getMailingAddress() {
       // on fetch success
       (response) => {
         const responseCopy = Object.assign({}, response);
-        // Translate military-only fields into generic ones; we'll translate
-        // them back later if necessary
+        // translate military address properties to generic properties for use in front end
         responseCopy.data.attributes.address = militaryToGeneric(response.data.attributes.address);
-        // if (addressCopy.type === ADDRESS_TYPES.military) {
-        //   addressCopy.city = addressCopy.militaryPostOfficeTypeCode;
-        //   addressCopy.stateCode = addressCopy.militaryStateCode;
-        //   addressCopy.countryName = 'USA';
-        //   delete addressCopy.militaryPostOfficeTypeCode;
-        //   delete addressCopy.militaryStateCode;
-        // }
-        // responseCopy.data.attributes.address = addressCopy;
         return dispatch({
           type: GET_ADDRESS_SUCCESS,
           data: responseCopy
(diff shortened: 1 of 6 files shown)


src/js/letters/actions/letters.js
```diff
@@ -236,13 +228,19 @@ export function saveAddress(address) {
   };
   window.dataLayer.push({ event: 'letter-update-address-submit' });
   return (dispatch) => {
-    // TODO: Show a spinner or some kind of indication we're waiting on this to return
```
cvalarida: 👍

src/js/letters/actions/letters.js
```diff
@@ -236,13 +228,19 @@ export function saveAddress(address) {
...
     return apiRequest(
       '/v0/address',
       settings,
-      () => dispatch(saveAddressSuccess(address)),
+      (response) => {
+        // translate military address properties back to front end address
+        const responseAddress = militaryToGeneric(response.data.attributes.address);
+        if (!isEqual(stripEmpties(address), stripEmpties(responseAddress))) {
+          const mismatchError = new Error('letters-address-update addresses don\'t match');
+          Raven.captureException(mismatchError, { submitted: address, returned: responseAddress });
```
cvalarida: Let's make sure we can log this before we merge it in. Not sure if it counts as PII.

src/js/letters/utils/helpers.jsx
```diff
@@ -339,3 +339,37 @@ export function resetDisallowedAddressFields(address) {
...
   return newAddress;
 }
+
+/**
+ * Traverses a single-level object and removes its zero-length own-enumerable properties
+ * @param {Object} input an object with no nested properties
+ * @returns a cloned object with no empty properties
+ */
+export const stripEmpties = (input) => {
+  const newObject = { ...input };
```
cvalarida: I just [confirmed on jsbin](http://jsbin.com/qifequzure/edit?js,console) that passing variables to a function does a shallow clone, so this line doesn't actually do anything. If you're looking to not modify `input`, you'll want to `_.deepClone()`.

src/js/letters/utils/helpers.jsx
```diff
@@ -339,3 +339,37 @@ export function resetDisallowedAddressFields(address) {
...
+  const deleteProperty = (key) => (delete newObject[key]);
```
cvalarida: Out of idle curiosity, why split this out here instead of defining the function directly in the `.forEach()` call?

test/e2e/letters-helpers.js
```diff
@@ -74,19 +74,47 @@ const address = {
...
+//   addressEffectiveDate: '2012-04-03T04:00:00.000+00:00',
+//   addressOne: '57 COLUMBUS STRASSA',
+//   addressThree: '',
+//   addressTwo: 'BEN FRANKLIN VILLAGE',
+//   city: 'Chicago',
+//   state: 'IL',
+//   type: 'DOMESTIC',
+//   zipCode: '60602',
+//   zipSuffix: ''
+// };
```
cvalarida: Can probably get rid of this.

test/letters/actions/letters.unit.spec.js
```diff
@@ -60,8 +60,21 @@ const getState = () => ({});
 describe('saveAddress', () => {
   const frontEndAddress = {
     type: ADDRESS_TYPES.military,
-    city: 'apo',
-    state: 'secret'
```
cvalarida: Ahh, should have been `state: 'Confusion'`! 😜 
Too bad!

src/js/letters/utils/helpers.jsx
```diff
@@ -339,3 +339,37 @@ export function resetDisallowedAddressFields(address) {
...
+ * Takes an address object as returned from vets-api and translates its properties to
+ * generic properties that are consumed by the front end
+ * @param {Object} address an address object as formatted by vets-api
+ * @returns {Object} shallow clone of address with military properties swapped for generics
+ */
+export const militaryToGeneric = (address) => {
+  if (address.type !== ADDRESS_TYPES.military) {
+    return { ...address };
+  }
+  const genericAddress = { ...address };
```
cvalarida: Same with these; you might want to `_.cloneDeep()`.

Not that it's a huge deal. :man_shrugging:

test/letters/utils/helpers.unit.spec.jsx
```diff
@@ -135,5 +137,90 @@ describe('Letters helpers: ', () => {
...
+    it('lets non-military address types pass through unmodified', () => {
+      const actualAddress = militaryToGeneric(domesticAddress);
+      expect(actualAddress).to.eql(domesticAddress);
+    });
+
+    it('returns a clone for all cases', () => {
+      const militaryTest = militaryToGeneric(militaryAddress);
+      const nonMilitaryTest = militaryToGeneric(domesticAddress);
+      expect(militaryTest).to.not.equal(militaryAddress);
+      expect(nonMilitaryTest).to.not.equal(domesticAddress);
```
cvalarida: See, this is tricky because the "top-level" objects _are_ different refs, but the "sub" objects will have the same refs.

For the address objects, though, it's probably totally fine because we won't have any nested objects. Suppose this is more academic then anything, really. 🤔

cvalarida (commented):
Looks good. Despite my repeated comments about not actually cloning the address, I think it's golden. The only reason I'm not going to approve this right now is because I want to make sure we can actually log the user's address first.

src/js/letters/actions/letters.js
markgreenburg: Oh, that's a great call...I'll check into it

src/js/letters/utils/helpers.jsx
cvalarida: Er, I stand corrected. Passing happens by reference, so cloning is necessary.

This is still a shallow clone, but that's probably fine since the address passed as `input` isn't going to have any sub-objects.

test/letters/utils/helpers.unit.spec.jsx
markgreenburg: Yeah, so these are just shallow objects with string values. But, that's just an implicit invariant and not actually specced out anywhere in the codebase, which is probably not great.

src/js/letters/actions/letters.js
markgreenburg: Checked with @kreek. Seems as though you could reverse engineer PII from the address itself so changed the code to ensure we don't log any of the actual address information itself.

src/js/letters/utils/helpers.jsx
markgreenburg: Yeah, I still don't think it's super great that this fact isn't actually enforced by the code, but don't think we need to do anything about it right now.

markgreenburg:
@cvalarida Addressed some of your comments and cleared the address logging with @kreek. And by 'cleared', I mean 'I changed the code by not logging the actual addresses to Sentry' 😉 
Would appreciate another run-through when you're feeling up to it so we can get this sucker merged. Thanks!

cvalarida (approved):
:shipit:

PR #6681, opened 2017-10-27
Letters accessibility fixes

annekainicUSDS:
This PR does a couple of things:
1. Focuses on certain elements when the page loads; this helps with tabbing and with the screen reader starting to read from the right place
2. Adds some scrolling functionality to the address section (this does not help with accessibility, but just something I added to this PR quickly)
3. Added some aria attributes that will read sections when the content changes.

Would love if the person who reviews this can pull the branch down and test it themselves to make sure the scrolling and focus elements work as they would expect.

### DIFF
--- a/src/js/letters/components/Address.jsx
+++ b/src/js/letters/components/Address.jsx
@@ -20,7 +20,7 @@ class Address extends React.Component {
   }
 
   componentDidMount() {
-    focusElement('select');
+    focusElement('h5');
   }
 
   getAdjustedStateNames = () => {
--- a/src/js/letters/components/Address.jsx
+++ b/src/js/letters/components/Address.jsx
@@ -2,6 +2,8 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import _ from 'lodash';
 
+import { focusElement } from '../../common/utils/helpers';
+
 import ErrorableSelect from './ErrorableSelect';
 import ErrorableTextInput from './ErrorableTextInput';
 import { STATE_CODE_TO_NAME, MILITARY_CITIES } from '../utils/constants';
@@ -15,6 +17,10 @@ import { militaryStateNames } from '../utils/helpers';
 class Address extends React.Component {
   componentWillMount() {
     this.id = _.uniqueId('address-input-');
+  }
+
+  componentDidMount() {
+    focusElement('select');
   }
 
   getAdjustedStateNames = () => {
--- a/src/js/letters/containers/AddressSection.jsx
+++ b/src/js/letters/containers/AddressSection.jsx
@@ -307,7 +307,7 @@ export class AddressSection extends React.Component {
       );
     }
 
-    return <div>{addressContent}</div>;
+    return <div aria-live="polite">{addressContent}</div>;
   }
 }
 
--- a/src/js/letters/containers/AddressSection.jsx
+++ b/src/js/letters/containers/AddressSection.jsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import { connect } from 'react-redux';
+import Scroll from 'react-scroll';
 
 import { scrollToFirstError } from '../../common/utils/helpers';
 import LoadingIndicator from '../../common/components/LoadingIndicator';
@@ -25,6 +26,16 @@ import {
   countryValidations,
   cityValidations
 } from '../utils/validations';
+
+const Element = Scroll.Element;
+const scroller = Scroll.scroller;
+const scrollToTop = () => {
+  scroller.scrollTo('addressScrollElement', window.VetsGov.scroll || {
+    duration: 500,
+    delay: 0,
+    smooth: true
+  });
+};
 
 // The address is empty if every field except type is falsey.
 // NOTE: It "shouldn't" ever happen, but it did in testing, so...paranoid programming!
@@ -167,6 +178,7 @@ export class AddressSection extends React.Component {
   startEditing = () => {
     window.dataLayer.push({ event: 'letter-update-address-started' });
     this.setState({ isEditingAddress: true });
+    scrollToTop();
   }
 
   dirtyInput = (fieldName) => {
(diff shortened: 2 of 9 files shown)


src/js/letters/containers/AddressSection.jsx
```diff
@@ -55,6 +66,9 @@ export class AddressSection extends React.Component {
     }
   }
 
+  componentDidMount() {
+    focusElement('h1');
```
markgreenburg: Had a really great conversation with an accessibility guru, who mentioned that it may be more intuitive to not steal focus or focus on top of page for the first page of the app, because focusing on H1 may give the false impression that there is no content preceding this element. He mentioned that voice assist users generally understand the first tabbable element on the page is the 'skip to content' link, which he recommended we use to go to the h1 instead of auto-focusing it on page load.

src/js/letters/containers/LetterList.jsx
```diff
@@ -6,10 +6,15 @@ import CollapsiblePanel from '../../common/components/CollapsiblePanel';
 import DownloadLetterLink from '../components/DownloadLetterLink';
 import VeteranBenefitSummaryLetter from './VeteranBenefitSummaryLetter';
 
+import { focusElement } from '../../common/utils/helpers';
 import { letterContent, bslHelpInstructions } from '../utils/helpers';
 import { AVAILABILITY_STATUSES, LETTER_TYPES } from '../utils/constants';
 
 export class LetterList extends React.Component {
+  componentDidMount() {
```
markgreenburg: Sam accessibility guru's recommendation was totally in line with what you did here - basically, for the second page of the app, focus on the beginning of the change, which is the 'step 2 of 2' `.nav-header`. This worked really nicely when I tried it out! 👍

src/js/letters/components/Address.jsx
```diff
@@ -17,6 +19,10 @@ class Address extends React.Component {
     this.id = _.uniqueId('address-input-');
   }
 
+  componentDidMount() {
+    focusElement('select');
```
cvalarida: Wonder if it might make more sense to have an "Edit address" header before this that we can focus on. I think generally speaking, this would be fine because to get here, they'd have to have clicked the "Edit" button, but if their address is returned as blank from the API, we automatically open this edit component. Focusing on the country field without any indication of what they're doing would be jarring, methinks.

I'll defer to @goldenmeanie on that one, though.

src/js/letters/containers/LetterList.jsx
```diff
@@ -66,7 +71,7 @@ export class LetterList extends React.Component {
     }
 
     return (
-      <div className="step-content">
+      <div className="step-content" aria-live="polite">
```
markgreenburg: For some reason, this works great on the first edit, but I had trouble getting the address to read back on subsequent edits. One 'hack' I was told is that sometimes `aria-live` can be spotty and needs to be combined with an `aria-relevant` attribute. I tried all the relevant aria-relevant values and from [MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-relevant_attribute#Values) and `aria-relevant=additions` seemed to work the best, including on subsequent edits. I'd love to hear what you think if once you're able to compare running the page with & without that extra aria attribute.

src/js/letters/containers/AddressSection.jsx
```diff
@@ -312,7 +329,14 @@ export class AddressSection extends React.Component {
       );
     }
 
-    return <div>{addressContent}</div>;
+    return (
+      <div>
+        <Element name="addressScrollElement"/>
+        <div aria-live="polite">
```
cvalarida: Oh how quaint.
> Would you be a dear and read the address to me?

So polite!

markgreenburg (approved):
Looks great! I'd also love to get your thoughts on some of my open-ended comments since I'm no accessibility expert.

src/js/letters/containers/LetterList.jsx
cvalarida: I'm not super familiar with this--does the `aria-live` attribute propagate to this element's children? That is, the only things I can think of changing in here are the download buttons, but does this cover that, or do we need to make each button `aria-live="polite"`?

cvalarida (approved):

src/js/letters/containers/AddressSection.jsx
annekainicUSDS: Good to know! I'll remove this.

src/js/letters/containers/LetterList.jsx
annekainicUSDS: In reading the docs on this it's hard to understand completely how this works, but my understanding is that this designates this element and its children as a live region, which means updates to that region will be read. I added this on `LetterList` so that when `CollapsiblePanel` is expanded, the content that describes each letter is read.

annekainicUSDS: I tried recreating this issue and wasn't able to, the address is always read back for me when edited. Just want to clarify that you're referring to an issue within `AddressSection`, as this comment is on `LetterList`. Want to make sure I'm not missing something!

src/js/letters/components/Address.jsx
annekainicUSDS: Ah that's a great point, will see what @goldenmeanie thinks!

goldenmeanie: Header sounds like a great idea @cvalarida @annekainicUSDS

src/js/letters/containers/LetterList.jsx
markgreenburg: Yep, you're not missing anything, comment is just in the wrong place.

annekainicUSDS:
Updated the header of the expanded address fields as Chris and Alex suggested, it now looks like this:
screen shot 2017-10-30 at 1 32 59 pm

In order to make this change, I had to do some refactoring that may make some of the abstraction of the address components unnecessary. I think we can probably get rid of `AddressBlock`, but leaving that to a separate PR.

PR #6683, opened 2017-10-27
If address is unavailable, show system unavailable message

cvalarida:
The unavailability of the address doesn't actually affect the veteran's ability to download their letters, but it may be confusing for that bit to not work but for them to _still_ be able to download their letters.

Before merging this, I want to be 100% certain this is the right thing to do since it will affect how we serve our Veterans.

**Update:** I confirmed with @kreek that this _is_ what we want to do.

### DIFF
--- a/src/js/letters/containers/Main.jsx
+++ b/src/js/letters/containers/Main.jsx
@@ -30,8 +30,6 @@ export class Main extends React.Component {
     this.props.getLetterList();
     this.props.getMailingAddress();
     this.props.getBenefitSummaryOptions();
-    // FOR TESTING PURPOSES ONLY; DO NOT LET THIS INTO PRODUCTION
-    // this.props.getAddressSuccessAction();
     this.props.getAddressCountries();
     this.props.getAddressStates();
   }
@@ -41,6 +39,11 @@ export class Main extends React.Component {
     // If letters are available, but address is still awaiting response, consider the entire app to still be awaiting response
     if (lettersAvailability === awaitingResponse || addressAvailability === awaitingResponse) {
       return awaitingResponse;
+    }
+
+    // If address isn't available, take the whole system down
+    if (addressAvailability === unavailable) {
+      return backendServiceError;
     }
 
     return lettersAvailability;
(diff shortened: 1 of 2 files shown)


src/js/letters/containers/Main.jsx
```diff
@@ -30,8 +30,6 @@ export class Main extends React.Component {
     this.props.getLetterList();
     this.props.getMailingAddress();
     this.props.getBenefitSummaryOptions();
-    // FOR TESTING PURPOSES ONLY; DO NOT LET THIS INTO PRODUCTION
```
markgreenburg: touche

src/js/letters/containers/Main.jsx
```diff
@@ -43,6 +41,11 @@ export class Main extends React.Component {
       return awaitingResponse;
     }
 
+    // If address isn't available, take the whole system down
+    if (addressAvailability === unavailable) {
+      return backendServiceError;
```
markgreenburg: Stepping through `letters.js`, it seems as though we have an action.type that maps to INVALID_ADDRESS_PROPERTY in `constants`. This type doesn't get used anywhere in the app...except in the unit tests where we make sure it works. Might `invalidAddressProperty` be a better one to use in this case? If not, can we just delete it and all its references, which might be even nicer. Thoughts?

markgreenburg (approved):
Had a small idea I left in comments but LGTM!

src/js/letters/containers/Main.jsx
cvalarida: Yeah, so what I did to test this was use the redux dev tools to dispatch a `GET_ADDRESS_FAILURE`. This made the `addressAvailability` `unavailable`. If we then look at the render function, it will display the system down message if the value returned from this function is `backendServiceError`, so I just made it return that if the address is down. :man_shrugging: 

The `invalidAddressProperty` seems more like it's for when we do get an address, but it's no good. Originally, I think it was used for blank addresses, but now we're just starting in edit mode if we get a blank address.

I think removing it is good. 👍

cvalarida: Removed in https://github.com/department-of-veterans-affairs/vets-website/pull/6689

PR #6690, opened 2017-10-28
Update pre-need form help info in footer.

U-DON:
Resolves department-of-veterans-affairs/vets.gov-team#3766.

Tagging @cehsu since you seem the most involved in creating/using the `AskVAQuestions` component.

Not sure if the copy change is inconsistent with other content, but perhaps something that should be decided one way or another by content people.

### DIFF
--- a/src/js/common/schemaform/AskVAQuestions.jsx
+++ b/src/js/common/schemaform/AskVAQuestions.jsx
@@ -7,12 +7,14 @@ function AskVAQuestions(props) {
         <div className="ask-va-questions help-sidebar-box">
           <h2 className="help-h2-border help-h2">Need Help?</h2>
           {props.children}
-          <p className="talk">To report a problem with this form,<br/>
-          call the Vets.gov Technical Help Desk:</p>
+          <p className="talk">
+            To report a problem with this form,
+            please call the Vets.gov Technical Help Desk:
+          </p>
           <p className="phone-number">
             <a href="tel:+1-855-574-7286">1-855-574-7286</a><br/>
-            TTY: <a href="tel:+18008778339">1-800-877-8339</a><br/>
-            Monday - Friday, 8:00 a.m. - 8:00 p.m. ET
+            TTY: <a href="tel:+1-800-877-8339">1-800-877-8339</a><br/>
+            Monday - Friday, 8:00 a.m. - 8:00 p.m. (ET)
           </p>
         </div>
       </div>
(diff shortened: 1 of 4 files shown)


bshyong (approved):
LGTM

src/js/common/schemaform/AskVAQuestions.jsx
```diff
@@ -7,12 +7,14 @@ function AskVAQuestions(props) {
...
-          call the Vets.gov Technical Help Desk:</p>
+          <p className="talk">
+            To report a problem with this form,
+            please call the Vets.gov Technical Help Desk:
+          </p>
           <p className="phone-number">
             <a href="tel:+1-855-574-7286">1-855-574-7286</a><br/>
-            TTY: <a href="tel:+18008778339">1-800-877-8339</a><br/>
-            Monday - Friday, 8:00 a.m. - 8:00 p.m. ET
+            TTY: <a href="tel:+1-800-877-8339">1-800-877-8339</a><br/>
```
bshyong: These links still work normally with the added dashes right?

bshyong (approved):

src/js/common/schemaform/AskVAQuestions.jsx
```diff
@@ -7,12 +7,14 @@ function AskVAQuestions(props) {
         <div className="ask-va-questions help-sidebar-box">
           <h2 className="help-h2-border help-h2">Need Help?</h2>
           {props.children}
-          <p className="talk">To report a problem with this form,<br/>
-          call the Vets.gov Technical Help Desk:</p>
+          <p className="talk">
+            To report a problem with this form,
```
cehsu: Is the `<br/>` removal deliberate? According to the design in the connected issue, it looks like the previous layout is still desired, unless this is now being achieved another way?

cehsu (approved):

src/js/common/schemaform/AskVAQuestions.jsx
U-DON: Should be. The other number has always had the dashes too.

src/js/common/schemaform/AskVAQuestions.jsx
U-DON: I'm not sure if it's still desired. I got no indication in the copy given for this change that the longer lines were to be broken into two lines (at least not forced instead of naturally due to limited screen space) and thought it would be inconsistent to have this be split but not the others.

PR #6692, opened 2017-10-30
Set up demo branch for testing

cehsu:
UPDATE: This is now intended to be merged and the test features available on staging. Thanks for feedback on this approach.

Applicant Information and Military History page prefill messages set up and sample data provided.

localhost-3001-health-care-apply-application-introduction

localhost-3001-health-care-apply-application-introduction 1

localhost-3001-health-care-apply-application-veteran-information-birth-information

localhost-3001-health-care-apply-application-military-service-service-information


peggygannon (commented):
@cehsu The only thing to fix is to make sure the quote mark in "We've" is facing the right way. Right now it looks like it's facing toward the "ve", when it should be facing toward "We"

cehsu:
Thanks @peggygannon, this has been fixed.

stacyhlinn:
Thanks @cehsu! To help the research session run more smoothly, I'd love to propose some updates to avoid confusion with the participants.
1. On the header, could we show the account name so it looks like the user is logged in?
2. Similarly, could we display the blue box message with the signed in language?
screenshot-localhost-3001-2017-10-31-09-43-28-136

3. Activate the green buttons under eligibility and application process.
4. Could we prefill more fields? I'm especially interested in seeing the participants' reaction when they see very personal information prefilled.
- Personal info: Social Security number + Birthday
- Military service: Start of service period & date of discharge, character of discharge so we can talk to participants about whether they find this prefill work valuable .

cehsu:
Thanks @stacyhlinn, the changes you mentioned have been made, kindly refer to the updated screenshots in the description above.

stacyhlinn:
Thank you @cehsu!

cehsu:
Looping in @jbalboni as it is now of interest to test this out on staging, and as there is a question of how best to simulate the prefilling of military history data?

src/js/common/schemaform/ObjectField.jsx
```diff
@@ -210,7 +211,7 @@ class ObjectField extends React.Component {
             {DescriptionField && <DescriptionField formData={formData} formContext={formContext} options={uiSchema['ui:options']}/>}
             {!textDescription && !DescriptionField && description}
           </div>}
-          {uiOptions.showPrefillMessage && formContext.prefilled && <PrefillMessage/>}
+          {process.env.BRANCH_NAME !== 'production' && uiOptions.showPrefillMessage && <PrefillMessage message={prefillMessage}/>}
```
jbalboni: I don't think `process.env.BRANCH_NAME` name is right, I think it's `__BUILDTYPE__` you want.

jbalboni (approved):
Seems fine for doing user testing against. I made one comment about how it's being hidden in production. You should do a production build locally and make sure it's hidden as you expect in prod.

jbalboni:
@cehsu I think it would be best to find a user that has that information. I think `+36` is the one that has some military history? Otherwise, I think we could work with Patrick and Lihan to figure out what users to test with.

src/js/common/schemaform/ObjectField.jsx
cehsu: Thanks @jbalboni, I've also reorganized this a bit more now so that the conditions reflect the different behaviors/scenarios.
- If in production or not on the military service page, check that information has been prefilled (in addition to uiOptions.showPrefillMessage being true).
- If not in production, and on the military service page, show the prefill notification as long as uiOptions.showPrefillMessage is true (information needs not be prefilled).

cehsu: Does that work for you @stacyhlinn?

src/js/common/schemaform/ObjectField.jsx
```diff
@@ -117,6 +118,15 @@ class ObjectField extends React.Component {
     return _.values(groupedProperties);
   }
 
+  showPrefill(uiOptions, formContext) {
+    const isProduction = process.env.__BUILDTYPE__ === 'production';
```
jbalboni: You should be able to leave off the `process.env` part.

jbalboni (approved):
LGTM, but made one comment.

src/js/common/schemaform/ObjectField.jsx
```diff
@@ -210,7 +221,7 @@ class ObjectField extends React.Component {
             {DescriptionField && <DescriptionField formData={formData} formContext={formContext} options={uiSchema['ui:options']}/>}
             {!textDescription && !DescriptionField && description}
           </div>}
-          {uiOptions.showPrefillMessage && formContext.prefilled && <PrefillMessage/>}
+          {this.showPrefill(uiOptions, formContext) && <PrefillMessage message={prefillMessage}/>}
```
jbalboni: It's generally recommended that you return true or false (not truthy or falsy) values when using the `{condition && ..}` pattern in React.

src/js/common/schemaform/ObjectField.jsx
cehsu: Thanks @jbalboni, removed.

src/js/common/schemaform/ObjectField.jsx
cehsu: Thanks @jbalboni, `showPrefill` now returns booleans.

src/js/hca/config/form.js
```diff
@@ -381,6 +384,10 @@ const formConfig = {
                 labels: dischargeTypeLabels
               }
             },
+            'ui:options': {
+              showPrefillMessage: true,
+              prefillMessage: 'military'
```
cehsu: If testing suggests that the additional message is helpful, we may consider using an array to store the pages on which the message can appear, and update the current logic for setting prefillStatust to indicate which information specifically was prefilled.

PR #6700, opened 2017-10-31
Download action tests mg

markgreenburg:
Adds missing unit tests for pdf downloads now that the action creator has stabilized.

### DIFF
--- a/test/letters/actions/letters.unit.spec.js
+++ b/test/letters/actions/letters.unit.spec.js
@@ -1,6 +1,5 @@
 import { expect } from 'chai';
 import sinon from 'sinon';
-import { cloneDeep } from 'lodash';
 
 import {
   ADDRESS_TYPES,
--- a/test/letters/actions/letters.unit.spec.js
+++ b/test/letters/actions/letters.unit.spec.js
@@ -1,5 +1,6 @@
 import { expect } from 'chai';
 import sinon from 'sinon';
+import { cloneDeep } from 'lodash';
 
 import {
   ADDRESS_TYPES,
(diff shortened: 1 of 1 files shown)


test/letters/actions/letters.unit.spec.js
```diff
@@ -359,49 +370,61 @@ describe('getBenefitSummaryOptions', () => {
...
+
+  it('dispatches FAILURE action if download fails', (done) => {
+    global.fetch.returns(Promise.reject(new Error('Oops, this failed')));
+    const { letterType, letterName, letterOptions } = benefitSLetter;
+    const thunk = getLetterPdf(letterType, letterName, letterOptions);
+    const dispatch = sinon.spy();
+    thunk(dispatch, getState)
+      .then(() => {
+        const action = dispatch.secondCall.args[0];
+        expect(action.type).to.equal(GET_LETTER_PDF_FAILURE);
```
cvalarida: We assume it should fail because we never configure `fetch` to return anything?

test/letters/actions/letters.unit.spec.js
```diff
@@ -359,49 +370,61 @@ describe('getBenefitSummaryOptions', () => {
...
-
-  it('sets up fetch for benefit summary letter', () => {});
-
-  it('sets up fetch for non-benefit-summary letters', () => {});
-
-  it('dispatches PENDING action when download initiated', () => {});
-
-  it('handles ie10 stuff', () => {});
-
-  it('downloads stuff conditionally', () => {});
```
cvalarida: Could you clarify why you didn't implement some of these tests? I know some of them are pretty hard to write tests for, but was that the only reason?

cvalarida (approved):

test/letters/actions/letters.unit.spec.js
markgreenburg: Yeah, any sort of error like this should get caught in the `onFailure` argument of `apiRequest`, so the thinking is that instead of mocking out a response object with errors this is simpler (since we don't distinguish between error types / codes anyway).

test/letters/actions/letters.unit.spec.js
markgreenburg: Eh, I thought that most of this stuff ended up being duplicative. Looking at it again today, I think it may still be worth testing IE10 as well as DOWNLOAD_SUCCESS for non-BSL types.

markgreenburg: Added those two tests if you want to give it another look. Otherwise I'll merge this afternoon.

PR #6705, opened 2017-11-01
Fix linting error, add missing validation for discharge date

jbalboni:
This is currently causing master to fail to build and there's also some missing validation.


ncksllvn (approved):
LGTM!

PR #6709, opened 2017-11-01
Add updated yarn.lock after uswds upgrade

ncksllvn:
Somehow forgot the `yarn.lock` after upgrading our version of USWDS.


robertfairhead (approved):
:+1:

PR #6710, opened 2017-11-01
Revert discharge date work

jbalboni:
We don't have the ok to move this forward and it also requires a schema change.


cvalarida (approved):
Looks good. I thought we were good to hit the big green button last night, so after checking with Ryan, I did. Forgot the stakeholder considerations. My bad 😬 

Thanks for reverting this!

PR #6711, opened 2017-11-01
Show expenses under non-empty field

jbalboni:
Updates the expenses warning to show up under the last non-empty field on the deductible expenses page.

screen shot 2017-11-01 at 1 18 34 pm
screen shot 2017-11-01 at 1 18 24 pm
screen shot 2017-11-01 at 1 18 13 pm

### DIFF
--- a/src/js/hca/helpers.jsx
+++ b/src/js/hca/helpers.jsx
@@ -170,50 +170,69 @@ export const expensesGreaterThanIncomeWarning = (
   </div>
 );
 
-export function expensesLessThanIncome(formData) {
-  const {
-    veteranGrossIncome = 0,
-    veteranNetIncome = 0,
-    veteranOtherIncome = 0,
-    dependents = []
-  } = formData;
-
-  const {
-    spouseGrossIncome = 0,
-    spouseNetIncome = 0,
-    spouseOtherIncome = 0
-  } = formData['view:spouseIncome'] || {};
-
-  const vetSpouseIncome =
-    veteranGrossIncome +
-    veteranNetIncome +
-    veteranOtherIncome +
-    spouseGrossIncome +
-    spouseNetIncome +
-    spouseOtherIncome;
-
-  const income = dependents.reduce((sum, dependent) => {
+export function expensesLessThanIncome(fieldShownUnder) {
+  const fields = ['deductibleMedicalExpenses', 'deductibleFuneralExpenses', 'deductibleEducationExpenses'];
+  return (formData) => {
     const {
-      grossIncome = 0,
-      netIncome = 0,
-      otherIncome = 0,
-    } = dependent;
+      veteranGrossIncome = 0,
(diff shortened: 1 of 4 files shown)


stacyhlinn:
@jbalboni Is the yellow line on the left of the card a new pattern? Or I just haven't encountered or noticed them before? I'm guessing it is similar to the blue line - that it is associated with the field above, correct? I like how it looks.

jbalboni:
@stacyhlinn It's new with the uswds upgrade.

stacyhlinn:
@jbalboni Got it. Looks good!

stacyhlinn (approved):

src/js/hca/helpers.jsx
```diff
@@ -170,50 +170,69 @@ export const expensesGreaterThanIncomeWarning = (
...
+      deductibleFuneralExpenses;
+
+    const hideBasedOnValues = income > expenses;
+
+    // If we're not going to hide based on values entered,
+    // then we need to make sure the current field is the last non-empty field
+    if (!hideBasedOnValues) {
+      const nonEmptyFields = fields.filter(field => formData[field]);
+      // If all fields are filled in, put it under the last one
+      if (!nonEmptyFields.length && fieldShownUnder === fields[fields.length - 1]) {
```
cehsu: According to the comment on line 224, should this be `nonEmptyFields.length === 3` (i.e. the expected total number have been filled in)? Or would it make sense to change the filter condition on line 223 to be `!formData[field]` so that empty fields are captured and `!emptyFields.length` can be used here instead? Is it necessary to have this clause? Could just the other two suffice?

jbalboni: Yep, looks like I just need the one.

test/hca/helpers.unit.spec.jsx
```diff
@@ -38,7 +38,35 @@ describe('HCA helpers', () => {
...
+        deductibleEducationExpenses: 0,
+        deductibleFuneralExpenses: 0,
+        deductibleMedicalExpenses: 4
+      };
+
+      expect(expensesLessThanIncome('deductibleMedicalExpenses')(formData)).to.be.false;
+      expect(expensesLessThanIncome('deductibleEducationExpenses')(formData)).to.be.true;
+      expect(expensesLessThanIncome('deductibleFuneralExpenses')(formData)).to.be.true;
+    });
+    it('should shown only on last field if all expenses are filled', () => {
```
cehsu: Should we also remove this test as well?

cehsu (approved):

test/hca/helpers.unit.spec.jsx
jbalboni: I think the test is still useful.

cehsu: Perhaps it could be rephrased? Show only if all expenses are filled?

PR #6715, opened 2017-11-01
Letters address update fix mg

markgreenburg:
- Don't store `addressEffectiveDate` property of `address` in Redux store
- ^ means that `addressEffectiveDate` isn't submitted to `vets-api` when an updated address is PUT
- Strip out `addressEffectiveDate` property from PUT response so that we can compare the response against the PUT values accurately
- Update associated tests

### DIFF
--- a/test/letters/actions/letters.unit.spec.js
+++ b/test/letters/actions/letters.unit.spec.js
@@ -59,19 +59,30 @@ const getState = () => ({});
 
 describe('saveAddress', () => {
   const frontEndAddress = {
-    type: ADDRESS_TYPES.military,
+    addressOne: '123 Any Street',
+    addressThree: '',
+    addressTwo: 'Apt 102',
     city: 'APO',
+    countryName: 'USA',
     stateCode: 'AE',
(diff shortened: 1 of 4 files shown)


test/letters/actions/letters.unit.spec.js
```diff
@@ -59,19 +59,30 @@ const getState = () => ({});
...
+    zipCode: '12345',
+    zipSuffix: ''
   };
 
   const successResponse = {
     data: {
       attributes: {
         address: {
-          type: frontEndAddress.type,
+          addressEffectiveDate: null,
```
cvalarida: This gets returned as `null` when we save the address? I think I remember that's what you said, but I want to verify.

cvalarida (approved):

test/letters/actions/letters.unit.spec.js
markgreenburg: Yeah...in theory it shouldn't matter anymore with these changes but that's what `vets-api` returned from the PUT.

PR #6724, opened 2017-11-03
Stop opening a new window in IE

cvalarida:
The window to oblivion is back. This shuts that portal to never open it again!

The history is something like this: I fixed this by moving that `window.open()` chunk down in the callback, but to get it to work in Safari, we have to call `window.open()` _before_ the callback. Unfortunately, IE opens that window because it doesn't support `download`...and then it never uses it. To fix that, we just make sure we don't open that tab for IE. Problem solved.


markgreenburg (approved):
🎉

PR #6729, opened 2017-11-03
Letters address help modal

markgreenburg:
Adds a modal to the Letters address section explaining the address shown.
Adds a link inside the address section to expand the modal.

Would love confirmation that the placement of the new 'link' (it's actually a button) within the address container is OK. I ended up not making it inline with the Edit button because our mobile layout uses a full-width Edit button on mobile, and it didn't seem worth it to introduce breakpoint-specific placements for this new link. Some mocks:

Small Phone:
screen shot 2017-11-03 at 11 39 42 am

screen shot 2017-11-03 at 11 39 48 am

Full Width:
screen shot 2017-11-03 at 11 39 03 am

screen shot 2017-11-03 at 11 39 23 am


src/js/letters/containers/AddressSection.jsx
```diff
@@ -4,15 +4,17 @@ import Scroll from 'react-scroll';
...
+  addressModalContent,
+  addressUpdateUnavailable,
   getStateName,
   getZipCode,
+  inferAddressType,
   isDomesticAddress,
-  isMilitaryAddress,
   isInternationalAddress,
-  addressUpdateUnavailable,
-  inferAddressType,
```
cvalarida: Took me way too long to figure out this just meant they were being rearranged. I was starting to think maybe there was some black magic goings on all up in these parts. 😨

src/js/letters/containers/AddressSection.jsx
```diff
@@ -295,9 +301,16 @@ export class AddressSection extends React.Component {
           <div className="letters-address street">{streetAddress}</div>
           <div className="letters-address city-state">{cityStatePostal}</div>
           <div className="letters-address country">{country}</div>
+          <button className="address-help-btn" onClick={this.openAddressHelp}>What is This?</button>
```
cvalarida: Should it be a lowercase `t` in `this`?

cvalarida: Is there a class that we can reuse to style this like a link?

cvalarida (approved):

src/js/letters/containers/AddressSection.jsx
markgreenburg: Mm, kind of - it's `usa-button-unstyled` but its padding / margins, outline and coloring are off, so would have needed to modify it pretty heavily anyway.

markgreenburg: Gah, good catch. The issue description does use a lowercase 't'.

markgreenburg: Gah, good catch. The issue description does use a lowercase 't'.

src/js/letters/containers/AddressSection.jsx
markgreenburg: Yeah sorry, when the list starts to grow out of control I like to alphabetize things. I meant to put a comment on this line to clarify; and then...I didn't.

PR #6731, opened 2017-11-03
Create migration test

cehsu:
These changes resolve https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4677. Thanks for any feedback or ideas for improvement.

### DIFF
--- a/test/migrations.unit.spec.js
+++ b/test/migrations.unit.spec.js
@@ -1,0 +1,34 @@
+import { expect } from 'chai';
+
+import fullSchema1010ez from '../src/js/hca/config/form';
+import fullSchema1990 from '../src/js/edu-benefits/1990/config/form';
+import fullSchema1990e from '../src/js/edu-benefits/1990e/config/form';
+import fullSchema1990n from '../src/js/edu-benefits/1990n/config/form';
+import fullSchema1995 from '../src/js/edu-benefits/1995/config/form';
+import fullSchema5490 from '../src/js/edu-benefits/5490/config/form';
+import fullSchema5495 from '../src/js/edu-benefits/5495/config/form';
+import fullSchema527EZ from '../src/js/pensions/config/form';
+import fullSchema530 from '../src/js/burials/config/form';
+import fullSchema10007 from '../src/js/pre-need/config/form';
+
+describe('form migrations:', () => {
+  it('should have a length equal to the version number', () => {
+    const configs = [
+      fullSchema1010ez,
+      fullSchema1990,
+      fullSchema1990e,
+      fullSchema1990n,
+      fullSchema1995,
+      fullSchema5490,
+      fullSchema5495,
+      fullSchema527EZ,
+      fullSchema530,
+      fullSchema10007
+    ];
+    configs.forEach(form => {
+      if (form.migrations || form.version > 0) {
+        expect(form.migrations.length).to.equal(form.version);
+      }
+    });
+  });
+});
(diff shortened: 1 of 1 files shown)


test/migrations.unit.spec.js
```diff
@@ -0,0 +1,34 @@
...
+      fullSchema1995,
+      fullSchema5490,
+      fullSchema5495,
+      fullSchema527EZ,
+      fullSchema530,
+      fullSchema10007
+    ];
+    configs.forEach(form => {
+      if (form.migrations || form.version > 0) {
+        expect(form.migrations.length).to.equal(form.version);
```
cvalarida: Can we make sure they're SiP-enabled first? Something like:

```js
import { isSIPEnabledForm } from '../src/js/user-profile/helpers';
...
if (isSIPEnabledForm(form.formID) && (form.migrations || form.version > 0)) {
  ...
}
```

cvalarida (commented):
Yeah, this is what I had in mind initially. I'm wondering, though, how can we make sure we add new forms to this list when we build them? Just want to make sure nothing slips through the cracks later on, y'know?

test/migrations.unit.spec.js
cehsu: Thanks @cvalarida. I could imagine this test failure being helpful even if SIP has been disabled for a form (or even in the unlikely event that it has not yet been enabled). Or would it make the most sense to combine the work involved in resolving such a test failure along with SIP re-enablement work?

cvalarida: Yeah, after talking with you in slack, I'm with you. If the form has never been SiP-enabled, it won't have migrations or a version > 0, and if it's only temporarily disabled, we still want to make sure the migrations and form version match up. You're 101% right!

Disregard my previous comment. 👍

cvalarida (approved):

test/migrations.unit.spec.js
```diff
@@ -0,0 +1,34 @@
+import { expect } from 'chai';
+
+import fullSchema1010ez from '../src/js/hca/config/form';
```
jbalboni: I'm a little concerned that we have more than one place that devs are expected to update when they add a new form. It's hard to surface these places effectively.

jbalboni (approved):
Looks good, but I'd like to see if we can make it as clear and easy as possible for devs to get new forms into a test like this.

test/migrations.unit.spec.js
cehsu: Great point @jbalboni, I'll add a test to check for that, similar to what @cvalarida designed for the user-profile helpers test.

test/migrations.unit.spec.js
```diff
@@ -0,0 +1,72 @@
...
+      fullSchema10007
+    ];
+    const allFormIds = Object.keys(schemas);
+    let reformattedIds = mappedIds.slice(0);
+    reformattedIds.splice(0, 1, '1010ez');
+    reformattedIds = reformattedIds.slice(0, -1);
+    const includedFormIds = configs.map(form => form.formId);
+    expect(allFormIds).to.deep.equal(mappedIds);
+    expect(includedFormIds).to.deep.equal(reformattedIds);
+  });
```
cehsu: @cvalarida @jbalboni thanks for any feedback on this additional test.

test/migrations.unit.spec.js
```diff
@@ -0,0 +1,72 @@
...
+      fullSchema10007
+    ];
+    const allFormIds = Object.keys(schemas);
+    const reformattedIds = mappedIds.slice(0);
+    reformattedIds.splice(0, 1, '1010ez');
+    reformattedIds.pop();
+    const includedFormIds = configs.map(form => form.formId);
+    expect(allFormIds).to.deep.equal(mappedIds);
+    expect(includedFormIds).to.deep.equal(reformattedIds);
+  });
```
cehsu: @cvalarida @jbalboni thanks for any feedback on this additional test.

PR #6733, opened 2017-11-03
Validate against special characters

cvalarida:
This just runs a specific validation against `addressOne` to filter out certain special characters. Okay, technically, it just whitelists certain characters, but that's probably splitting hairs.

Things I'm uncertain of:
- Does this validation match what's in production?
  - The swagger docs are unclear, saying only that "Address must specify a valid address one"
- Should we run this against _only_ `addressOne`?
- Should it really disallow `-`?
  - The error message didn't include that in the list of accepted special characters, but it seems like that's pretty likely to show up in an actual (valid) address

@kreek can you answer any of those questions or point me in the right direction?

### DIFF
--- a/src/js/letters/utils/validations.js
+++ b/src/js/letters/utils/validations.js
@@ -15,6 +15,14 @@ const requiredValidator = (input, fullAddress, message = requiredMessage) => {
   return true;
 };
 
+const specialCharacterValidator = (input) => {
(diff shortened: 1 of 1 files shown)


markgreenburg (approved):
LGTM! Though I do feel as though we might need the same validations for the other two address line fields.

kreek (approved):
LGTM 🚀

PR #6734, opened 2017-11-06
Include USWDS Grid in Print View

ncksllvn:
Previously, USWDS's grid was only in effect on the "screen" media-type. This was a problem because we lost some of our formatting in the print view. This PR does some overrides to add the grid into the print view.

##### Questions
1. Is `_b-breakpoints.scss` still an appropriate name for the file containing several USWDS overrrides? It still makes some sense since everything in there deals with media queries and breakpoints, but not sure if it would be better to include the word "overrides" in there.
2. This is obviously an ugly solution - I had to copy Neat's `media` mixin in order to add "print" into the media query. It was the only way I could think to do it, since as far as I know you can't override a mixin then call its predecessor. I think it sorta of makes sense to just log this as an issue in Neat or USWDS, but I don't think we'd ever see this problem resolved that way.

Resolves department-of-veterans-affairs/vets.gov-team/issues/3321


phiden (approved):

ncksllvn:
@annekainicUSDS can you confirm that this fixes the problem as shown in the screenshot of the original issue? I don't see where I can find credentials for checking that part of the application.

jbalboni:
@ncksllvn That change looks like it will work, but it reinforces that we should really make a decision about what our grid solution is going forward. Based on the lack of feedback so far, I don't think we can reasonably hope for uswds's grid to become a complete solution for us.

src/sass/base/_b-breakpoints.scss
```diff
@@ -5,12 +5,65 @@
...
+// We do a custom override here, setting our site's medium breakpoint to 768 instead of how USWDS defines it at 600.
 $xsmall-screen:       320px; // QVGA display
 $medium-large-screen: 768px;
 $medium-screen:       $medium-large-screen;
 
 $medium: new-breakpoint(min-width $medium-large-screen 6);
+
+// This is an override on Neat's media mixin to create media queries for both screen and print
+// https://github.com/thoughtbot/neat/blob/v1.8.0/app/assets/stylesheets/grid/_media.scss
+@mixin media($query: $feature $value $columns, $total-columns: $grid-columns) {
```
jbalboni: It's probably worth checking any uses of `@media()` in our code to see if affects a page that is meant to be printed (or already has print styles). This also potentially affects any uswds component that uses the grid and has print styling (I'm not sure if any do).

ncksllvn: It's actually even more confusing than how it looks here, because in the USWDS's imports, they first use this function as shown here for the grid SASS, then they import their utility SASS, which overrides `@media()` as shown here with a simpler `@media()` function that is used throughout the rest of the SASS. It was such a headache.

ncksllvn:
@jbalboni I completely agree. I've lost a lot of faith in USWDS's. The features are very lacking and there isn't a roadmap for planned features. On one hand, we're invested in Foundation's grid so I'm fine with that. But ideally, we'd have a standalone, lightweight grid that we can easily manage as a dependency. The latter would be nice but changing all of that markup sounds very difficult.

src/sass/base/_b-breakpoints.scss
ncksllvn: Here's the new [media mixin](https://github.com/18F/web-design-standards/blob/develop/src/stylesheets/core/_utilities.scss#L46). So really, these changes should only affect the grid, because only the grid uses that version of the `@media` mixin. I didn't override the second `@media` mixin, which only applied to the screen.

This is way too confusing lol.

ncksllvn: I would guess that most pages are using Foundation's grid instead so I don't see anything too crazy happening to the print style, but yeah, you're definitely right.

cvalarida:
image
Looks good! 🎉

ncksllvn:
Thank you @cvalarida!

PR #6736, opened 2017-11-06
Address edit letters mg

markgreenburg:
Gates address update functionality for users with blank address to ensure only users who `canUpdate` are presented with the edit form by default.

Note - blank addresses shouldn't be possible to get back from EVSS, so this state shouldn't be possible in Prod. We're seeing it in Staging, probably because of bad test data, so this change is purely precautionary.


cvalarida (approved):
:shipit:

PR #6737, opened 2017-11-06
Moved vr&e index to employment root...

phiden:
in order to remove it from the left nav.

So basically: `employment/vocational-rehab-and-employment/index.md` became `employment/vocational-rehab-and-employment.md` -- this is inline with the way the rest of the site is structured, preserves the existing IA and removes the extraneous leftnav item.


mnorthuis:
where can I review?  
Doesn't look correct at this link, but not sure if I'm in the right spot...
https://vetsgov-pr-6337.herokuapp.com/employment/vocational-rehab-and-employment/eligibility/

phiden:
@mnorthuis -- I think that's the wrong review instance. Try this one: https://vetsgov-pr-6737.herokuapp.com/employment/vocational-rehab-and-employment/eligibility/

mnorthuis (approved):
Looks good!

PR #6739, opened 2017-11-06
Fix for session timeout dialog not showing up

ncksllvn:
The `entryTime` property, which is used as a timestamp after a user token is retrieved, was not being set on the session storage after retrieving a new user token, which caused the session timeout confirmation box to never appear. This initializes that property after retrieving a new user token.

Resolves department-of-veterans-affairs/vets.gov-team/issues/3337


jbalboni (approved):
LGTM, but I am not the login expert.

U-DON (approved):
LGTM.

PR #6741, opened 2017-11-07
Convert help to tooltip (AdditionalInfo)

cehsu:
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/5452#event-1328241754
screen shot 2017-11-07 at 2 51 36 pm
screen shot 2017-11-07 at 2 51 52 pm


jbalboni:
@cehsu I think Peggy meant the pattern in the `AdditionalInfo` component, not the collapsible panel.

cehsu:
@jbalboni Thanks for pointing me toward the appropriate component. I will use that instead.

stacyhlinn:
@cehsu I also want to make sure that the "learn more" additional info is going to appear above the radio buttons. Thanks!

cehsu:
Thanks @jbalboni @peggygannon @stacyhlinn for the comments, kindly refer to the updated screenshots above at your convenience.

cehsu:
Thanks @jbalboni @peggygannon @stacyhlinn for the feedback. Kindly refer to the updated screenshot at your convenience. @jbalboni I have used `ui:description` in a hacky way, to achieve the desired placement. Should we consider looking into changing the default `ui:help` behavior in the future?

jbalboni:
@cehsu I think that's a good idea, to use `ui:help`. It shouldn't be too hard to set up, you would add it to `FieldTemplate`.

jbalboni:
@cehsu I think the way you have it for `ui:description` works fine. That's what that field is meant for.

src/js/hca/config/form.js
```diff
@@ -729,12 +730,16 @@ const formConfig = {
             isMedicaidEligible: {
               'ui:title': 'Are you eligible for Medicaid?',
               'ui:widget': 'yesNo',
-              'ui:help': 'Medicaid is a United States health program for eligible individuals and families with low income and few resources.'
+              'ui:description': <AdditionalInfo triggerText="Learn more about Medicaid.">
```
jbalboni: I would be interested to see what this looks like when there's an error with the field.

src/js/hca/config/form.js
```diff
@@ -1,3 +1,4 @@
+import React from 'react';
```
jbalboni: Should change this file to jsx or move the description element into helpers.

jbalboni (approved):
LGTM, had a couple questions.

src/js/hca/config/form.js
cehsu: @jbalboni great point, change made.

src/js/hca/config/form.js
cehsu: Great point @jbalboni, screenshot provided.

peggygannon (commented):
@ceshu. I'm not sure we need the "Please provide response" line that is shown in the first screenshot. We already have (*Required) next to the question. @stacyhlinn What do you think? 
Otherwise it looks good!

stacyhlinn:
@peggygannon I think what we're seeing here is when a user tries to hit continue without answering this required question, and the page is showing validation error. 
This looks good to me!

stacyhlinn (approved):

peggygannon (approved):

peggygannon:
@stacyhlinn OK, thanks! looks good.

PR #6742, opened 2017-11-07
Adds questions and answers for discharge wizard

bshyong:
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/5614

image

This is a prototype for Veteran testing purposes.

Some not yet addressed TODOs:
- fix scrollTo behavior
- debug question show/hide edge cases
- add additional questions to "review answers" section
- check mobile views
- accessibility check


src/js/discharge-wizard/components/FormQuestions.jsx
```diff
@@ -0,0 +1,305 @@
...
+        </div>
+        <table className="usa-table-borderless">
+          <tbody>
+            {Object.keys(this.props.formValues).map(k => {
+              const value = this.props.formValues[k];
+              if (Array.isArray(value)) { return null; }
+
+              if (k === '4_dischargeYear') {
+                const dischargeMonth = months.find(e => { return e.value.toString() === this.props.formValues['5_dischargeMonth']; });
+                if (k === '4_dischargeYear' && dischargeMonth) { return null; }
```
U-DON: This shouldn't need `k === '4_dischargeYear`' since it's already nested under that same condition. Should only need to check `dischargeMonth`.

U-DON (approved):
This was a gigantic change set, so I only looked at whatever lived in the `discharge-wizard` folder. It's very complex to create the form from scratch; is there no way to use `vets-json-schema` form config to build this out?

Otherwise, LGTM.

src/js/discharge-wizard/components/FormQuestions.jsx
bshyong: @U-DON thanks—yes I considered using vets json schema but this form logic looked too crazy to work with.. I definitely want to look at this to see about refactoring for use with `vets-json-schema`

PR #6745, opened 2017-11-07
Add helper text for facility input

jbalboni:
Adds helper text on the facility page, since we can't make it optional.

screen shot 2017-11-07 at 3 32 05 pm

### DIFF
--- a/src/js/hca/config/form.js
+++ b/src/js/hca/config/form.js
@@ -18,7 +18,7 @@ import {
   transform,
   dischargeTypeLabels,
   lastServiceBranchLabels,
-  FacilityHelp,
+  facilityHelp,
   medicalCentersByState,
   medicalCenterLabels,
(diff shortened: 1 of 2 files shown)


stacyhlinn:
@jbalboni Does that link open up facility locator in a new tab?

jbalboni:
@stacyhlinn Yes. All those links open in a new tab. The facility locator line is also currently there already, what's below it is new.

jbalboni:
@stacyhlinn It might be good to try to surface this information earlier (if we aren't already), so people don't fill out a chunk of the form and get here and realize they can't continue.

cehsu (approved):
LGTM.

peggygannon (approved):
@jbalboni LGTM. Thanks!

stacyhlinn (approved):

PR #6747, opened 2017-11-07
Add important to usa-button color

cehsu:


src/sass/modules/_m-button.scss
```diff
@@ -8,6 +8,7 @@ button,
 [type="reset"],
 [type="image"] {
   background-color: $color-primary;
+  color: $color-white !important;
```
jbalboni: This applies to more than just the primary button, I think. I think we should apply it to just the primary one.

cehsu: Thanks @jbalboni, change made.

jbalboni (approved):

PR #6748, opened 2017-11-07
Catch loading errors and log data to Sentry

jbalboni:
We're getting some Sentry errors in the formState logic when loading a form, this PR changes the error handling to sanitize and send the saved form data to Sentry so we can do further debugging.

### DIFF
--- a/src/js/common/schemaform/FormApp.jsx
+++ b/src/js/common/schemaform/FormApp.jsx
@@ -107,7 +107,11 @@ class FormApp extends React.Component {
       && status !== this.props.loadedStatus
       && !window.location.pathname.endsWith('/error')
     ) {
-      newProps.router.push(`${newProps.formConfig.urlPrefix || ''}error`);
+      let action = 'push';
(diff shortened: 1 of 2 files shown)


src/js/common/schemaform/FormApp.jsx
```diff
@@ -107,7 +107,11 @@ class FormApp extends React.Component {
       && status !== this.props.loadedStatus
       && !window.location.pathname.endsWith('/error')
     ) {
-      newProps.router.push(`${newProps.formConfig.urlPrefix || ''}error`);
+      let action = 'push';
+      if (window.location.pathname.endsWith('resume')) {
+        action = 'replace';
+      }
+      newProps.router[action](`${newProps.formConfig.urlPrefix || ''}error`);
```
jbalboni: I found that if you're on the error page and you came from a content page or the account page, clicking Back sends you back to the resume page, and puts you in a loop. This replaces the resume entry in the history and avoids that.

cehsu (approved):
LGTM.

PR #6749, opened 2017-11-07
Adds guidance page questions

bshyong:
Includes optional step and step 1

discharge wizard_ vets_guidance


src/js/discharge-wizard/containers/GuidancePage.jsx
```diff
@@ -1,24 +1,131 @@
...
+    return null;
+  }
+
+  renderStepOne() {
+    const questionOneResponse = this.props.formValues['1_reason'];
+    const specReason = ['1', '2', '4'].indexOf(questionOneResponse) > -1;
+    const boardToSubmit = board(this.props.formValues);
+    /* eslint-disable quote-props */
+    const reasons = {
+      '1': {
```
U-DON: Even unquoted, both `reasons['1']` and `reasons[1]` will return the same. So you shouldn't need the quotes (and the eslint-disable) here.

src/js/discharge-wizard/containers/GuidancePage.jsx
```diff
@@ -1,24 +1,131 @@
...
+        type: 'condition',
+      },
+      '4': {
+        name: 'sexual assault',
+        type: 'experience',
+      },
+    };
+    /* eslint-enable quote-props */
+
+    const strongCaseTips = () => {
```
U-DON: It doesn't seem like this content needs to be in a function as opposed to just being conditionally assigned to a variable.

U-DON (commented):
LGTM.

PR #6750, opened 2017-11-08
Discharge wizard

bshyong:
Merging for staging test


src/js/discharge-wizard/components/FormQuestions.jsx
```diff
@@ -0,0 +1,304 @@
...
+        break;
+      case '3':
+        labelYear = 2011;
+        break;
+      case '4':
+        labelYear = 2017;
+        break;
+      default:
+        break;
+    }
```
ncksllvn: Would it possibly be better to move these into a map? Like:

`const labelYears = { '1': 2014, '2': 2014, '3'...`

bshyong: yes it would. 100% want to refactor this entire app and clean things up once the dust settles

src/js/discharge-wizard/reducers/discharge-wizard.js
```diff
@@ -0,0 +1,104 @@
...
+      break;
+    case '6_courtMartial':
+      next = '7_branchOfService';
+      break;
+    case '7_branchOfService':
+      next = '8_prevApplication';
+      break;
+    case '8_prevApplication':
+      if (answer === '1' && parseInt(state['1_reason'], 10) < 5) {
+        next = '9_prevApplicationYear';
```
ncksllvn: Should any of these integer values be stored as constants?

ncksllvn (approved):
Made a couple tiny suggestions but LGTM either way! Great work!

PR #6758, opened 2017-11-09
Handle schema validation errors for non-veteran filled out pre-needs forms

jbalboni:
This fixes the following validation errors I saw when submitting the pre-needs form as a non-veteran applicant:

- `view:` fields were sometimes included
- `serviceName` was not set to `currentName` for the veteran
- Partial addresses were being sent through to the backend


mphprogrammer (approved):
I'm not well versed in vets-website, but it LGTM. Just out of curiosity, in the case where the sponsor, applicant, and claimant are the same do we fill in missing name data, and if so is it also from veteran.currentName?

rlbaker (approved):

jbalboni:
@mphprogrammer If the veteran and claimant are the same, it looks like we use either the serviceName or the claimant name. The currentName field only shows up if the claimant isn't the veteran. 

It looks like we do that here: https://github.com/department-of-veterans-affairs/vets-website/pull/6758/files#diff-98955615c4713be246b10a312c358359R47

cehsu (approved):
LGTM.

PR #6761, opened 2017-11-09
Appeals housekeeping

cvalarida:
1. Makes the appeals v2 api requests in development only
2. Puts the appeals from :point_up: in the `claims` store (franchise?)


markgreenburg (approved):
🎉  :shipit:

PR #6762, opened 2017-11-09
Fix for session timeout dialog causing error

ncksllvn:
This PR is a patch for the session timeout dialog. It was originally created in a PR a long time ago, but wasn't popping up. I thought that the functionality was all working except for that one piece, but I was wrong. This fixes the other problems, which is mostly that the website was trying to restore the user's session as soon as the user clicks "ok" on the timeout dialog, which opens as soon as the window loads. However, we needed to wait for the application to also retrieve the login URL's before doing opening that dialog or else we'd get a null reference, which looks like:

```
Cannot read property 'idme' of null
```

Resolves department-of-veterans-affairs/vets.gov-team/issues/5902

#### Full Breakdown

I opened this PR: department-of-veterans-affairs/vets-website/pull/6739

To fix this issue: department-of-veterans-affairs/vets.gov-team/issues/3337

That came as a result of this PR: department-of-veterans-affairs/vets-website/pull/3876


U-DON (approved):
LGTM.

ncksllvn:
Thanks @U-DON!

PR #6764, opened 2017-11-10
Appeals v2 CurrentStatus component & helpers

markgreenburg:
Adds a component that displays the current status for an appeal, as well as a helper that retrieves the right content based on the status type.

Lots of fake content in here as it's still being finalized. Also need to get some feedback on implementation (the content list will grow by more than 3x so the current approach may get messy).

### DIFF
--- a/src/js/claims-status/utils/helpers.js
+++ b/src/js/claims-status/utils/helpers.js
@@ -245,7 +245,7 @@ export function getClaimType(claim) {
 }
 
 // TO DO: Replace made up properties and content with real versions once finalized.
-export const STATUS_CONTENT = {
+export const STATUS_TYPES = {
   nod: 'nod',
   awaitingHearingDate: 'awaiting_hearing_date',
   bvaDecision: 'bva_decision',
@@ -263,7 +263,7 @@ export const STATUS_CONTENT = {
  * @returns {Contents}
  */
 export function getStatusContents(type, details) {
-  const { nod, awaitingHearingDate, bvaDecision } = STATUS_CONTENT;
+  const { nod, awaitingHearingDate, bvaDecision } = STATUS_TYPES;
 
   const contents = {};
   if (type === nod) {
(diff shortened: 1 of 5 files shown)


src/js/claims-status/utils/helpers.js
```diff
@@ -243,3 +243,97 @@ export function getCompletedDate(claim) {
 export function getClaimType(claim) {
   return claim.attributes.claimType || 'Disability Compensation';
 }
+
+// TO DO: Replace made up properties and content with real versions once finalized.
+export const STATUS_CONTENT = {
```
cvalarida: Can we rename this to clarify that this holds the types? Maybe `CURRENT_STATUS_TYPES`? Also, we might want to switch the keys and values around so we can use the snake_case string we get from the api as the key.

🤔  Are the current status types different than past event types, or do they share a common status type enum?

src/js/claims-status/utils/helpers.js
```diff
@@ -243,3 +243,97 @@ export function getCompletedDate(claim) {
...
+ * @returns {Contents}
+ */
+export function getStatusContents(type, details) {
+  const { nod, awaitingHearingDate, bvaDecision } = STATUS_CONTENT;
+
+  const contents = {};
+  if (type === nod) {
+    const office = details.regionalOffice || 'Regional Office';
+    contents.title = `The ${office} is reviewing your appeal`;
+    contents.description = `The ${office} received your Notice of Disagreement and is revewing
```
cvalarida: 🤔  Verification: The newlines and beginning-of-line spaces will be captured in that string, but when it gets put into the `<p>`, it'll get stripped out and appear like we expect it...right?

cvalarida (approved):
Looks great! Just had one small naming question.

src/js/claims-status/utils/helpers.js
markgreenburg: I believe they're the same, so going with STATUS_TYPES. Based on what I've seen so far, each status has its place in the chronological appeal order, and can either be the current status, past status, or next status. Let's confirm with appeals team.

PR #6765, opened 2017-11-10
Add autosuggest widget for cemetery lookup

jbalboni:
This adds a type ahead widget based on the Autosuggest component used by GIBCT.

screen shot 2017-11-09 at 11 03 31 pm


src/js/common/schemaform/FieldTemplate.jsx
```diff
@@ -74,8 +74,8 @@ export default function FieldTemplate(props) {
   const useFieldsetLegend = (isFieldGroup || !!showFieldLabel) && !useLabelElement;
 
   const labelElement = useFieldsetLegend
-    ? <legend className={labelClassNames}>{label}{requiredSpan}</legend>
-    : <label className={labelClassNames} htmlFor={id}>{label}{requiredSpan}</label>;
+    ? <legend id={`${id}-label`} className={labelClassNames}>{label}{requiredSpan}</legend>
```
jbalboni: Need this for the labeled by attribute in the Autosuggest

src/js/common/schemaform/fields/AutosuggestField.jsx
```diff
@@ -0,0 +1,122 @@
...
+        inputProps={{
+          id,
+          name: id,
+          value: this.state.input,
+          onChange: this.onChange,
+          'aria-labelledby': `${id}-label`,
+          onBlur: this.handleBlur
+        }}/>
+    );
+  }
```
jbalboni: Should write some prop types, I think.

src/js/common/schemaform/helpers.js
```diff
@@ -245,6 +245,10 @@ export function stringifyFormReplacer(key, value) {
     if (fields.length === 0 || fields.every(field => value[field] === undefined)) {
       return undefined;
     }
+    // autosuggest widgets save value and label info, but we should just return the value
+    if (value.widget === 'autosuggest') {
```
jbalboni: We need to have both the id and the label in the form data, in case a user comes back to the autosuggest field. But we don't want to send both. So I'm adding some extra data to tag this as an autosuggest field and convert it to the id.

mphprogrammer:
@jbalboni Be sure to ask EOAS to whether State cemeteries are valid (they will pass validation, however). I cannot recall if EOAS only works with National cemeteries. If so, you may want to filter by type of cemetery (its returned in getCemeteries).

cehsu (approved):
LGTM.

src/js/common/schemaform/definitions/autosuggest.js
```diff
@@ -0,0 +1,27 @@
+import _ from 'lodash/fp';
+
+import AutosuggestField from '../fields/AutosuggestField';
+
+export const schema = {
+  type: 'object',
+  properties: {
+    id: { type: 'any' },
```
cehsu: What are the different types that this can be?

src/sass/modules/_m-schemaform.scss
```diff
@@ -498,3 +498,41 @@ fieldset.schemaform-field-template {
...
+ul.react-autosuggest__suggestions-list {
+  background: $color-white;
+  border: 1px solid $color-gray;
+  // Box shadow only sides and bottom of suggestion list.
+  box-shadow: 0px 3px 3px -3px $color-focus,
+              3px 0 3px -3px $color-focus,
+              -3px 0 3px -3px $color-focus,
+              0 7px 7px -7px $color-focus,
+              7px 0 7px -7px $color-focus,
+              -7px 0 7px -7px $color-focus;
```
cehsu: The blur radius looks a little funny here, could it make sense to set that to zero so the shadow more closely reflects the shape of the options list? Or are rounded box-shadows best practice in this situation?

src/js/common/schemaform/definitions/autosuggest.js
jbalboni: Whatever is necessary for the form/api being used. I didn't want to assume anything. For this form, it's a string.

src/sass/modules/_m-schemaform.scss
jbalboni: I copied this from the existing autosuggest styles. I think we can make another issue to ask if we should tweak them and share the css between the two places where it's used.

jbalboni: https://github.com/department-of-veterans-affairs/vets.gov-team/issues/5977

PR #6771, opened 2017-11-13
Handle null field error

jbalboni:
Treats nulls as undefined values in the form state updates. This is necessary because of how Rails 4 handles data set as parameters.


cehsu (approved):
LGTM.

PR #6774, opened 2017-11-13
Appeals Timeline

cvalarida:
Because of how we have to build the timeline visually, I went ahead and made a Timeline component that's in charge of expanding and collapsing the history as well as showing the current event.

### What this does
- [x] Gets the past events and puts them into `<li>`s
- [x] Mostly styles the past and current events like we want
- [x] Shows event title and description based on the event type
    - This is just to get some basic functionality out the door; the content has definitely not reached officialdom

### What still needs to be done outside this PR
- [ ] Add the little arrow head to the bottom of the timeline
- [ ] Upon expanding the past events, scroll the expander to the middle of the screen
    - Or...similar experience
- [ ] Fix the little chunk of left border on the first `<li>` above the bullet when the screen is small

---

image

image


src/js/claims-status/components/appeals-v2/Timeline.jsx
```diff
@@ -0,0 +1,77 @@
...
+          <h3>{title || 'Title here'}</h3>
+          <span className="appeal-event-date">on {moment(event.date, 'YYYY-MM-DD').format('MMMM d, YYYY')}</span>
+          <p>{description}</p>
+        </li>
+      );
+    });
+  }
+
+  toggleExpanded = () => {
+    this.setState({ expanded: !this.state.expanded });
```
markgreenburg: state changes based on prev. state can get tricky since `setState` doesn't guarantee when state changes. For these types of state changes, React team recommends using the functional version - `setState((prevState, props) => ({ expanded: !prevState.expanded }));`
Ref: [State Updates May Be Asynchronous](https://reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous)

src/js/claims-status/components/appeals-v2/Timeline.jsx
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  render() {
+    const eventList = this.state.expanded ? this.getPastEvents() : [];
+    const dateRange = this.props.events[0] ? `${moment(this.props.events[0].date, 'YYYY-MM-DD').format('MMMM d, YYYY')} - ${moment(this.props.events[this.props.events.length - 1].date, 'YYYY-MM-DD').format('MMMM d, YYYY')}` : '';
```
markgreenburg: Since we use this format throughout in at least 3 places, do you think it might be worth it to just extract this formatting into its own little function? E.g., 
`format = (rawDate) => {
    // do some formatting with moment...
    return formattedDate
}`

Might make this easier to read. 🤷‍♂️

src/js/claims-status/components/appeals-v2/Timeline.jsx
```diff
@@ -0,0 +1,77 @@
...
+
+    // Add the expander
+    const expanderClassName = this.state.expanded ? 'section-expanded' : 'section-unexpanded';
+    eventList.push(
+      <li key={eventList.length} className={`process-step ${expanderClassName}`}>
+        {/* Giving this a margin top to help center the text to the li bullet */}
+        <button onClick={this.toggleExpanded} className="va-button-link">
+          <h4 style={{ color: 'inherit' }}>{this.state.expanded ? 'Hide past events' : 'See past events'}</h4>
```
markgreenburg: out of curiosity, why did you decide to style this inline?

src/js/claims-status/components/appeals-v2/Timeline.jsx
```diff
@@ -0,0 +1,77 @@
...
+        <div className="appeal-event-date">{dateRange}</div>
+      </li>
+    );
+
+    // Add the current status
+    const { title, description } = this.props.currentStatus;
+    eventList.push(<CurrentStatus key={eventList.length} title={title} description={description}/>);
+
+    return (
+      // May not want this as an ol...
```
markgreenburg: 🤷‍♂️

src/js/claims-status/components/appeals-v2/Timeline.jsx
cvalarida: So this `h4` needs to inherit the color of the `.va-button-link`, otherwise its color is just black (and thus doesn't look like a link). I _could_ have given it a class to make that happen, or I could have done something like

```css
.va-button-link {
  h4 { color: inherit; }
}
```
but it just seemed more straightforward to give it an inline style.

src/js/claims-status/components/appeals-v2/Timeline.jsx
cvalarida: Yeah, I thought about it. I'll go ahead and make that happen. 👍

src/js/claims-status/components/appeals-v2/Timeline.jsx
cvalarida: Oooh, didn't know this was a thing. Good call!

src/js/claims-status/containers/AppealsV2StatusPage.jsx
```diff
@@ -18,16 +18,18 @@ class AppealsV2StatusPage extends React.Component {
   componentDidMount() {
     // Make sure we grab the appeals if we don't have them already
     // Useful if the user goes directly to the appeal status without going to the list first
-    if (!this.props.appeal) {
+    if (this.props.appeal === AppealsV2StatusPage.defaultProps.appeal) {
```
markgreenburg: Will this change once we have a real API?

src/js/claims-status/containers/AppealsV2StatusPage.jsx
```diff
@@ -36,29 +38,55 @@ class AppealsV2StatusPage extends React.Component {
...
+  }).isRequired,
+  appeal: PropTypes.shape({
+    id: PropTypes.string.isRequired,
+    type: PropTypes.string.isRequired,
+    attributes: PropTypes.object.isRequired, // Can flesh this out later
+  })
 };
 
-function mapStateToProps(state, ownProps) {
+function isolateAppeal(state, id) {
```
markgreenburg: Might we want to extract this into helpers or somewhere like that? That would let us reuse this for other components.

src/js/claims-status/utils/appeals-v2-helpers.jsx
```diff
@@ -0,0 +1,195 @@
+// TO DO: Replace made up properties and content with real versions once finalized.
+export const STATUS_TYPES = {
```
markgreenburg: Seems like EVENT_TYPES is the new version of this. Let's standardize on the other one and delete this one?

markgreenburg: Also...maybe we throw it into a new `constants` file? 🤷‍♂️

src/sass/claims-status.scss
```diff
@@ -602,6 +602,76 @@ h1:focus {
...
+    margin-left: -2.65em;
+
+    &:before {
+      margin-left: -2.65em;
+    }
+  }
+
+  .section-complete:before {
+    font-family: FontAwesome;
+    content: "\f00c";
```
markgreenburg: :tada

src/sass/claims-status.scss
```diff
@@ -602,6 +602,76 @@ h1:focus {
...
+      margin-left: -3.85em;
+      margin-top: 11px; // Pull the bullet down to align it better
+    }
+  }
+
+  .appeal-event-date {
+    color: $color-gray-light;
+  }
+}
+
```
markgreenburg: This must've taken you forever...toyed around with it on desktop as well as mobile sizes and everything looks great!

test/claims-status/components/appeals-v2/Timeline.unit.spec.jsx
```diff
@@ -0,0 +1,74 @@
...
+describe('<Timeline/>', () => {
+  it('should render', () => {
+    const wrapper = shallow(<Timeline {...defaultProps}/>);
+    expect(wrapper.type()).to.equal('ol');
+  });
+
+  it('should expand and collapse past history events', () => {
+    const component = shallow(<Timeline {...defaultProps}/>);
+    const expander = component.find('.section-unexpanded button');
+    expect(expander.exists()).to.be.true;
```
markgreenburg: Not sure if this works - I believe .find() will always return a`ShallowWrapper` even if it doesn't find a matching node, so `exists()` might always return `true`. From the enzyme docs, it seems as though `expect(wrapper.find('.some-class').length).to.equal(1)` is the idiomatic way of doing it...which I'm on the fence about since it kind of obscures what's being tested.

markgreenburg (approved):
🎉 🌮 :shipit:

test/claims-status/components/appeals-v2/Timeline.unit.spec.jsx
cvalarida: I just tested this by changing the `.find()` selector to something that doesn't exist (`.section-unexpanded buttons` with an s) and the test failed, so I think we're golden. :crossed_fingers:

src/js/claims-status/utils/appeals-v2-helpers.jsx
cvalarida: Per our "phone" conversation, we'll check with @cmgiven to make sure the current status types map directly to the past event types first. If so, we can address it in another little PR.

cmgiven: Responded in Slack, but note that there will be two different sets of types for events and statuses.

gnakm:
@cvalarida gah! These are looking great!

Here are some answers to your questions, followed by more questions:

- We should stick to the default vets.gov link color, so the the image above is 👍 
- Bullet size looks 👍 

Questions for you:
- I'm assuming that the text styles are all in the Vets.gov default colors. Can you confirm?
- Does the vets.gov team have standardized spacing for the text styles? Some of the text styles look like they can use some bottom padding. 
- I think that the date under the event is `$color-gray-light` Could we make it `$color-gray-medium` for accessibility purposes?

cvalarida:
@gnakm Answers!

- I didn't mess with any of the text styles or colors, so I'm assuming they're all default
- Great question about the spacing...I really don't know... :man_shrugging: 
- Yeah, it is `$color-gray-light`; I'll add that to [the CSS tweaks issue](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/5998)

PR #6775, opened 2017-11-14
Tabbed navigation for Appeals

markgreenburg:
To Do:
[x] - Set up routing & scaffolding for appeal, appeal status, appeal details
[x] - Set up tabs: set active link by default, navigate between tabs w/functional history.push()
[x] - Conditionally render content based on active tab

Split off into other PRs:
- Better formatting to more closely match layout in mocks
- CSS: 1. convert reused claims-styles to appeals-styles for better control, 2. add styles for things that don't look right to match the mocks


src/js/claims-status/containers/AppealsV2DetailPage.jsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+
+const AppealsV2DetailPage = () => {
+  return (
+    <div>
+      Hey I'm the details page!
```
cvalarida: 😂

src/js/claims-status/containers/AppealInfo.jsx
```diff
@@ -0,0 +1,35 @@
...
+    </div>
+  );
+};
+
+AppealInfo.PropTypes = {
+  params: PropTypes.shape({ id: PropTypes.string.isRequired }).isRequired,
+  route: PropTypes.shape({ path: PropTypes.string.isRequired }).isRequired,
+};
+
+export default AppealInfo;
```
cvalarida: Could you clarify why you export this twice--once as a named export and once as the default?

src/js/claims-status/routes.jsx
```diff
@@ -41,10 +42,19 @@ const routes = [
...
+      key=":id"
+      from=":id"
+      to=":id/status"/>
     <Route
-      component={AppealsV2StatusPage}
+      component={AppealInfo}
       key=":id/status"
       path=":id/status"/>
+    <Route
+      component={AppealInfo}
```
cvalarida: Can we wrap these `Route`s in another `Route` that uses `AppealInfo` as its component and have that just renders its children?

```jsx
<Route
  component={AppealInfo}
  key=":id"
  path=":id">
  <Redirect
    key="/"
    from="/" {/* May need to be an IndexRedirect or something; I don't remember */}
    to="/status"/>
  <Route
    component={AppealsV2StatusPage}
    key="/status"
    path="/status"/>
  <Route
    component={AppealsV2DetailsPage}
    key="/details"
    path="/details"/>
</Route>
```

This way, we won't have to do any route-based rendering from inside `AppealInfo`, y'know?

src/js/claims-status/containers/AppealInfo.jsx
```diff
@@ -0,0 +1,35 @@
...
+import AppealsV2DetailPage from './AppealsV2DetailPage';
+import AppealsV2TabNav from '../components/appeals-v2/AppealsV2TabNav';
+
+const PATHS = { status: 'status', detail: 'detail' };
+
+const setActiveTab = (currentPath, paths = { status: '', detail: '' }, appealId) => {
+  if (currentPath.includes(paths.status)) {
+    return <AppealsV2StatusPage appealId={appealId}/>;
+  } else if (currentPath.includes(paths.detail)) {
+    return <AppealsV2DetailPage appealId={appealId}/>;
```
cvalarida: 🤔  I'm not super keen on this bit; see my comment in `routes.jsx`. I think we can get around this with a bit more fancy footwork in `react-router`.

cvalarida (commented):
It looks pretty good, but can we use `react-router` to handle rendering the status and details page? This works, but looking at it later might get confusing.

src/js/claims-status/containers/AppealInfo.jsx
markgreenburg: Yeah, I got rid of these. Generally I export twice to isolate e.g. connected components in unit tests. Obviously that's not needed here. Not to mention, it broke the tests.

src/js/claims-status/containers/AppealsV2DetailPage.jsx
markgreenburg: Finalize content: ✓

src/js/claims-status/routes.jsx
markgreenburg: 🤔 Let me look into that. I'm all for reducing complexity in the components.

src/js/claims-status/containers/AppealInfo.jsx
markgreenburg: I decided to keep things consistent-ish with what was done for Claims before but...I think I like the router-based implementation better. Will play around and see what I can do.

cvalarida (approved):
Beautiful!

PR #6779, opened 2017-11-14
Implement helper function for rendering all site-wide components

ncksllvn:
Eventually we'll be adding a React-based footer and header, and since we already have a login component rendering on nearly all pages, it seems like a good idea to create a single helper function for rendering all site-wide/fixed components to avoid adding new code to all of the app entry points whenever we add a new one. This PR implements that function and replaces calls to create the single login component to use that instead.

### DIFF
--- a/src/js/common/store/index.js
+++ b/src/js/common/store/index.js
@@ -12,10 +12,6 @@ export const commonReducer = {
   })
 };
 
-export function renderCommonComponents(commonStore) {
-  createLoginWidget(commonStore);
-}
-
 export default function createCommonStore(appReducer = {}) {
   const reducer = _.assign(appReducer, commonReducer);
   const useDevTools = __BUILDTYPE__ === 'development' && window.devToolsExtension;
--- a/src/js/common/store/index.js
+++ b/src/js/common/store/index.js
@@ -12,7 +12,7 @@ export const commonReducer = {
   })
 };
 
-export function renderCommonComponents(commonStore){
+export function renderCommonComponents(commonStore) {
   createLoginWidget(commonStore);
 }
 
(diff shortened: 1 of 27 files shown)


src/js/common/store/index.js
```diff
@@ -11,6 +12,10 @@ export const commonReducer = {
   })
 };
 
+export function renderCommonComponents(commonStore) {
```
jbalboni: It feels weird to me for this helper to be in this file, which is focused on Redux store things.

jbalboni (approved):
LGTM, had one minor issue.

src/js/common/store/index.js
ncksllvn: Yeah I know what you mean. It felt weird to me to put it anywhere really. Any suggestions? I didn't want to mix it with all of the components.

cehsu (approved):
LGTM.

src/js/common/store/index.js
jbalboni: Good question. Somewhere in /common. It could be an `init-components` file in the main common directory?

jbalboni: Alternatively, we could change the name of this file and treat it as the common set up code?

ncksllvn: The second idea sounds good to me. Makes sense because if the common store is going to be used then we'll most likely want the common components along with it

ncksllvn:
@jbalboni mind taking another look? I moved that `createCommonComponents` function into `common/init-common.js`, which export a helper function that creates the common store and the common components.

jbalboni (approved):
LGTM

U-DON (approved):
LGTM.

bshyong (approved):
LGTM

ncksllvn:
Thanks everybody

PR #6781, opened 2017-11-14
Add loading indicator

cvalarida:
### The problem
When I refresh the status page, I have to re-query the api to get the list of all appeals and then pull out the one this page is looking at. That takes time, so while we're waiting, we see only a half-baked page.

### The solution
Rather than showing this page with no information yet, we show a loading indicator instead.

It feels a bit awkward to be "Loading your information..." and then "Please wait while we load your appeal...", but it's probably better than showing everything empty.

@gnakm How's it look to you?

@peggygannon What should the loading spinner say? The "Please wait while we load your appeal..." is just filler I put in to get this gif up.

loading indicators


gnakm:
@cvalarida This looks great! I have the same questions that I dropped in this PR: https://github.com/department-of-veterans-affairs/vets-website/pull/6774 e.g. Standardized padding, colors, etc. I think I'm going to do a check in with Mel today to confirm these things with her. I say visually, lgtm 👍 Maybe we can revisit the spacing stuff in another ticket?

src/js/claims-status/actions/index.jsx
```diff
@@ -157,7 +157,7 @@ export function getAppealsV2() {
       ]
     })
       .then((response) => dispatch(fetchAppealsSuccess(response)))
-      .catch(() => dispatch({ type: SET_APPEALS_UNAVAILABLE }));
+      .catch(() => dispatch({ type: SET_APPEALS_UNAVAILABLE })), 4000);
```
markgreenburg: Probably realistic 😞

src/js/claims-status/containers/AppealsV2StatusPage.jsx
```diff
@@ -24,6 +25,9 @@ class AppealsV2StatusPage extends React.Component {
   }
 
   render() {
+    if (this.props.appealsLoading) {
```
markgreenburg: I guess we'll need the same thing in the `Detail` page, in case someone navigates there directly. An alternative might be to (once my tabs PR gets merged in) do this in AppealInfo and conditionally render either a `LoadingIndicator` or `this.props.children`. That would also mean moving some of the appeal fetching logic up there as well. But we'll cross that bridge when we get there :)

src/js/claims-status/containers/AppealsV2StatusPage.jsx
```diff
@@ -85,7 +89,7 @@ function isolateAppeal(state, id) {
 function mapStateToProps(state, ownProps) {
   return {
     appeal: isolateAppeal(state, ownProps.params.id),
-    loading: state.loading
+    appealsLoading: state.disability.status.appeals.appealsLoading
```
markgreenburg: So I take it loading status was already being stored in redux?

markgreenburg (approved):
LGTM!

src/js/claims-status/containers/AppealsV2StatusPage.jsx
cvalarida: Yeah, and it probably makes the most sense to pull that fetching logic into `AppealInfo` anyhow. 👍

src/js/claims-status/containers/AppealsV2StatusPage.jsx
cvalarida: Yessir!

peggygannon:
@cvalarida could we just have one message "Please wait while we load your appeal"  for both screens, rather than having  "Loading your information" first? 

Either way it's fine.. the language is fine. Thanks for checking!

cvalarida:
@peggygannon Great question. Let me see how it's all structured to know if that's feasible.

cvalarida:
Looks like it's done with two loading indicators right now in v1, so I think I'll leave that as an enhancement for later. I'll write a ticket up to investigate it further.

gnakm (approved):
lgtm!

PR #6785, opened 2017-11-15
Appeals v2 timeline style tweaks

cvalarida:
The original design can be found [here](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/5637).

image

**Note:** There's still some work left to do concerning the text spilling out past the divider lines, but that's an easy fix after we get the tabs PR merged in. Should just be able to take away the `right-margin` from the `.separator` and call it good. :crossed_fingers:

### DIFF
--- a/src/sass/claims-status.scss
+++ b/src/sass/claims-status.scss
@@ -685,14 +685,14 @@ h1:focus {
     width: 0px;
     border-left: 0.6em solid transparent;
     border-right: 0.6em solid transparent;
-    border-top: 1.2em solid #d6d7d9;
+    border-top: 1.2em solid $color-gray-lighter;
     margin-top: -32px;
     margin-left: 0.8em;
   }
 
   .separator {
     margin: 0.8em 2em 0 -0.7em;
-    border-top: 1px solid #d6d7d9;
+    border-top: 1px solid $color-gray-lighter;
   }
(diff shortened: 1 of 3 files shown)


gnakm (approved):
Looking g8!

src/sass/claims-status.scss
```diff
@@ -668,7 +678,21 @@ h1:focus {
...
   .appeal-event-date {
-    color: $color-gray-light;
+    color: $color-gray-medium;
+  }
+
+  & + .down-arrow {
+    width: 0px;
+    border-left: 0.6em solid transparent;
+    border-right: 0.6em solid transparent;
+    border-top: 1.2em solid #d6d7d9;
```
annekainicUSDS: Very minor, but is there a variable already for this hex color? Might be `$color-gray-lighter`?

annekainicUSDS (approved):
LGTM! One minor comment

src/sass/claims-status.scss
cvalarida: 😂 
@robbiethegeek Just asked about this too and I made a PR to fix what's already been merged into master. Looped back around to fix this up and saw this comment.

Good call! `$color-gray-lighter` is exactly what I want. 😄

PR #6787, opened 2017-11-15
Add attachment type to file upload, remove file edit states

jbalboni:
This adds the attachment type to the pre need upload page. It also updates the file upload field on pensions/burials to follow @stacyhlinn's new design, in order to make everything consistent between forms (except for the attachment type).

Some things to note:
- I made changes to the common FileField component to allow attachmentId specifically. This appears to be a common requirement for VA systems, to be able to specify a document type, so I figured that would be a reasonable thing to build in.
- I adjusted how validation errors are shown on this field. The old way of wrapping all files in an error message doesn't mesh well with having a document type. So now files with errors are highlighted individually

screen shot 2017-11-15 at 1 51 22 pm
screen shot 2017-11-15 at 1 51 04 pm
screen shot 2017-11-15 at 1 50 44 pm
screen shot 2017-11-15 at 1 50 38 pm
screen shot 2017-11-15 at 1 50 09 pm


test/common/components/SignInLink.unit.spec.jsx
```diff
@@ -100,7 +100,7 @@ describe('<SignInLink>', () => {
   // TODO: While this _does_ test that the function is called, it doesn't test
   //  that the function is called _at the right time_. We should circle back to
   //  this and fix that.
-  it('should call onLogin after a successful login', () => {
+  xit('should call onLogin after a successful login', () => {
```
jbalboni: This test is async but isn't written properly. I'm excluding it, since it causes confusion when some other tests fail.

stacyhlinn (approved):
LGTM.

src/js/pre-need/config/form.js
```diff
@@ -566,7 +570,30 @@ const formConfig = {
...
-                  attachments
+                  preneedAttachments: _.merge(preneedAttachments, {
+                    items: {
+                      properties: {
+                        attachmentId: {
+                          'enum': [
+                            '1',
+                            '2',
+                            '3',
+                            // '4',
```
cehsu: Is this planned in the future?

cehsu (approved):
LGTM. Had a small question.

src/js/pre-need/config/form.js
jbalboni: No, this is a type that they said we could exclude, since it's called "VA PRENEED FORM," which is sort of redundant when that's the form you're filling out.

rlbaker (approved):

PR #6792, opened 2017-11-15
[WIP] More Appeals V2 Formatting!

cvalarida:
I'm not sure I like the breakpoints where they're at, or how big the buttons are, but those can be changed (within reason for the breakpoints; I'd rather not roll out something too custom).

### Small
image

### Medium
image

### Large
image


annekainicUSDS (approved):
LGTM!

gnakm:
@cvalarida I think this looks great and I agree that we shouldn't customize this too much.

cvalarida:
Hmm... 🤔 I wonder if that's just an optical illusion because the tabs are so big; the line goes across the full 8 columns in both medium and large. Would making those tabs a bit smaller be better? They're currently 10 of the 12 columns wide (inside the 8-column context just to make it confusing).

Shrinking the tabs to 8 columns looks like this:
image

_641px wide--the smallest "medium" viewport we can have_

gnakm:
Ah sorry -- perhaps these smaller things are being a blocker. I think the way you originally had it above looks great.

cvalarida:
With the new typography from master:
image

PR #6796, opened 2017-11-16
Tweak VIC eligibility

patrickvinograd:
Eligibility expands beyond title 38 veterans so isVeteran is no longer accurate; backend implements correct logic when populating service list.


patrickvinograd:
Backend changes in https://github.com/department-of-veterans-affairs/vets-api/pull/1536

U-DON (approved):
LGTM.

PR #6801, opened 2017-11-17
Appeals v2 breadcrumbs

cvalarida:
Our component hierarchy wasn't really great for the breadcrumbs as they were, so I pondered it over with @cehsu and decided to nix the `AppealLayout`, moving the breadcrumb logic into `AppealInfo`. I then pulled the appeals redux state grabbing logic out of `AppealsV2StatusPage` and put it in `AppealInfo`, which I had wanted to do already for when we start building the details page.

**Note:** This doesn't test the `isolateAppeal()` functionality, but I'm not sure if 1) we want to or 2) the best way to do that if we do want to.

image

image

### DIFF
--- a/src/js/claims-status/actions/index.jsx
+++ b/src/js/claims-status/actions/index.jsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import environment from '../../common/helpers/environment';
-import { makeAuthRequest } from '../utils/helpers';
+import { makeAuthRequest, mockData } from '../utils/helpers';
 
 export const SET_CLAIMS = 'SET_CLAIMS';
 export const SET_APPEALS = 'SET_APPEALS';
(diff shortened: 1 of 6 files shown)


src/js/claims-status/containers/AppealInfo.jsx
```diff
@@ -1,20 +1,69 @@
...
+
+  return appeal;
+}
+
+class AppealInfo extends React.Component {
+  render() {
+    const { params, children } = this.props;
+    const appealId = params.id;
+    // TODO: Make sure this is where we should get the date from
+    const appealDate = this.props.appeal ? moment(this.props.appeal.attributes.events[0].date, 'YYYY-MM-DD').format('MMMM YYYY') : '';
```
cvalarida: @cmgiven Should the date that shows in the breadcrumbs come from the first event in the array?

cmgiven: The first event with type `claim`. We should also be able to handle the absence of such an event, in which case it might just read "Appeal of Claim Decision".

cc @gnakm

gnakm: I was thinking something like Appeal of Claim Decision [Month Year]. This is something I want to go over with Amy Behrens when she comes town. 

image

gnakm: Maybe stick with Appeal of Claim Decision.

src/js/claims-status/actions/index.jsx
```diff
@@ -91,71 +91,7 @@ export function getAppealsV2() {
...
-              {
-                description: 'Service treatment records',
-                date: '2017-09-30'
-              }
-            ]
-          }
-        }
-      ]
-    })
+    return Promise.resolve(mockData)
```
markgreenburg: Good call

src/js/claims-status/containers/AppealInfo.jsx
```diff
@@ -1,20 +1,69 @@
...
+    appeal.attributes.events.unshift({
+      type: appeal.attributes.type === 'post_cavc_remand' ? 'cavc_decision' : 'bva_remand',
+      date: appeal.attributes.prior_decision_date,
+    });
+  }
+
+  return appeal;
+}
+
+export class AppealInfo extends React.Component {
```
markgreenburg: Does this need to be a class-based component? I *think* functional components are `connect`able as well

src/js/claims-status/containers/AppealInfo.jsx
```diff
@@ -1,20 +1,69 @@
...
+          <Breadcrumbs>
+            <li><Link to="your-claims">Your Claims and Appeals</Link></li>
+            <li><strong>Appeal of {appealDate} Claim Decision</strong></li>
+          </Breadcrumbs>
+        </div>
+        <div className="row">
+          <div className="medium-8 columns">
+            <AppealsV2TabNav appealId={appealId}/>
+            <div className="va-tab-content va-appeals-content">
+              {React.Children.map(children, child => React.cloneElement(child, { appeal: this.props.appeal }))}
```
markgreenburg: Nice! I take it this is necessary so that you can pass the `appeal` prop to `{this.props.children}`. Nifty.

markgreenburg (approved):
🎉 LGTM!

src/js/claims-status/containers/AppealInfo.jsx
cvalarida: Yup! That's exactly why. 😄

src/js/claims-status/containers/AppealInfo.jsx
cvalarida: Um, I'm not sure. I had it in my head that they weren't, so I just made it a class... :man_shrugging:

I can try keeping it functional and :crossed_fingers: just for fun.

src/js/claims-status/containers/AppealInfo.jsx
cvalarida: @cmgiven Excellent. I can make that filter and leave off the date if there are no events of type `claim`. 👍

src/js/claims-status/containers/AppealInfo.jsx
cvalarida: Good news! You were right. Totally works. 💥

PR #6808, opened 2017-11-17
Fix sitemap issues found in SEO audit

robertfairhead:
Closes: https://github.com/department-of-veterans-affairs/vets.gov-team/issues/5950


robertfairhead:
@ncksllvn You can confirm at https://vetsgov-pr-6808.herokuapp.com/sitemap.xml that the file now uses HTTPS and the mentioned pages are excluded

ncksllvn (approved):
Very nice! I would not have known you can add that `private` property to the markdown like that.

PR #6811, opened 2017-11-20
Add GA event to track 404s

robertfairhead:
Part of: https://github.com/department-of-veterans-affairs/vets.gov-team/issues/5671

This PR adds a custom event fired on any page that renders in the `404.md` content. If the user goes to nonexistent page like `https://www.vets.gov/nothing-here.html` in GA we currently see that URL show up but do not know if it rendered actual content or the `404.md` content for being a missing page.

This adds an event so we can see which URLs are turning up the `404.md` content over time.

Also, there was a slight refactor of the `footer.html` code. It's a general best practice to not overwrite the `window.dataLayer` variable if it exists. That way regardless of the `404.md` or the `footer.html` executes first they'll both use the `push` function and not overwrite anything that may exist. I added a comment with a ref on that, but that can be axed if it's overkill.

### DIFF
--- a/content/includes/footer.html
+++ b/content/includes/footer.html
@@ -106,9 +106,9 @@ visit <a href="https://www.veteranscrisisline.net/">VeteransCrisisLine.net</a> f
 {% else %}
   <!-- Google Tag Manager -->
   <script>
-    dataLayer = [{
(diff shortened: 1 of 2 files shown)


ncksllvn (approved):
LGTM!

PR #6812, opened 2017-11-20
Remove VIC from ignoreList

patrickvinograd:
Application going into production for internal testing; still behind beta enrollment link, and also VIC integration only available on-VA-network.


ncksllvn (approved):
LGTM

PR #6814, opened 2017-11-20
Appeals v2 header spacing fix

cvalarida:
This was kind of a pain to track down, but once found, it was super easy to fix. 🎉 

## Before
image

## After
image


markgreenburg (approved):
LGTM

PR #6815, opened 2017-11-20
Add user feedback form above footer

ncksllvn:
This adds a widget into the footer for users to submit feedback. See the issues below for more info.

### Issues
Resolves department-of-veterans-affairs/vets.gov-team/issues/3445
Resolves department-of-veterans-affairs/vets.gov-team/issues/4126


src/js/feedback/components/FeedbackForm.js
```diff
@@ -0,0 +1,93 @@
...
+import PropTypes from 'prop-types';
+import AlertBox from '../../common/components/AlertBox';
+
+const MIN_INPUT = 1;
+
+class FeedbackForm extends React.Component {
+
+  constructor(props) {
+    super(props);
+    this.state = {
```
jbalboni: I think we've generally leaned toward using Redux for state (https://github.com/department-of-veterans-affairs/vets.gov-team/blob/master/Work%20Practices/Engineering/Style%20Guides/Javascript/vets.gov%20Front%20End.md#using-setstate-in-react-components), unless it's pretty emphemeral UI state. Why'd you decide to make this local state vs putting it with the other Redux state?

src/js/feedback/components/DefaultView.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+function GiveUsFeedback({ feedbackButtonClicked }) {
```
jbalboni: Should the file name and the function name match?

src/js/feedback/components/FeedbackForm.js
```diff
@@ -0,0 +1,93 @@
...
+          </div>
+        </div>
+        <input id="should-send-response" type="checkbox" value={this.state.shouldSendResponse} onChange={this.shouldSendResponseChanged}/>
+        <label htmlFor="should-send-response">I would like to receive a response about my feedback.</label>
+        <div className="usa-grid-full">
+          <div className="usa-width-one-third">
+            {this.state.shouldSendResponse ? (
+              <div>
+                <label htmlFor="email">Your email address</label>
+                <input name="email" value={this.state.email} onChange={this.emailChanged}/>
```
jbalboni: Should these inputs match our (and uswds's) normal form validation patterns? We have an `ErrorableTextarea` and `ErrorableTextInput` components that could be used for that. We also have an email validation function in `src/js/common/utils/validations.js`.

src/js/feedback/components/DefaultView.js
```diff
@@ -0,0 +1,24 @@
...
+  return (
+    <div className="usa-grid-full">
+      <div className="usa-width-one-half">
+        <h4 className="feedback-widget-title">Tell us what you think</h4>
+        <p className="feedback-widget-intro">
+          We are always looking for ways to make Vets.gov better.
+        </p>
+      </div>
+      <div className="usa-width-one-half">
+        <button onClick={feedbackButtonClicked} className="usa-button-secondary feedback-button">Give us feedback</button>
```
jbalboni: I don't think it hurts to name this `revealForm` instead of `feedbackButtonClicked`. That's what this is going to do, and it's not generic enough that we could do a variety of things based on this button click. So `revealForm` tells you more about what's going to happen when this button is clicked.

src/sass/modules/_m-footer.scss
```diff
@@ -19,7 +19,94 @@
...
+        margin-top: 0;
+      }
+    }
+  }
+
+  .feedback-widget-intro {
+    margin: 0;
+  }
+
+  label[for=feedback] {
```
jbalboni: It seems odd that you need this much custom css for these fields. The input components we have might help with that.

src/sass/modules/_m-footer.scss
```diff
@@ -19,7 +19,94 @@
...
+.feedback-widget {
+  padding: 1.5em 0;
+  background-color: $color-gray-lightest;
+  color: $color-gray-dark;
+
+  .row {
+    padding: 0 0.9375rem;;
+  }
+
+  .feedback-widget-title {
```
jbalboni: Do these need to be nested? You've prefixed the class names, so it seems unnecessary.

src/sass/modules/_m-footer.scss
```diff
@@ -19,7 +19,94 @@
   }
 }
 
-  .footer-inner {
+.feedback-widget {
+  padding: 1.5em 0;
+  background-color: $color-gray-lightest;
+  color: $color-gray-dark;
+
+  .row {
```
jbalboni: This is something that might trip someone up if we make design changes in there again later and they need to use `row` in a different context.

src/sass/modules/_m-footer.scss
```diff
@@ -19,7 +19,94 @@
...
+  }
+
+  .feedback-widget-need-help {
+    order: 0;
+    flex: 0 1 100%;
+    .feedback-widget-need-help-inner {
+      background-color: $color-aqua-lightest;
+      padding: 1em;
+      margin: 1em 0;
+      h3 {
```
jbalboni: I think I would use `>` or add a class to this `h3`. It's another one of those things that can become a pain later if someone makes changes in here later and wants to tweak the header styles.

src/js/feedback/containers/Main.jsx
```diff
@@ -0,0 +1,81 @@
...
+
+  render() {
+    let content = null;
+
+    if (this.props.feedbackReceived) {
+      content = (
+        <FeedbackSubmitted
+          shouldSendResponse={this.props.shouldSendResponse}/>
+      );
+    } else if (!this.state.formIsVisible) {
```
jbalboni: I'm curious how a screen reader would deal with the trigger for showing the form being completely replaced. In other places where we have collapsing/expanding content we use aria-expanded and aria-controls to convey some of that.

jbalboni (commented):
Looks pretty good. Left some comments.

src/js/feedback/components/DefaultView.js
ncksllvn: Yeah that would probably be better.

src/js/feedback/containers/Main.jsx
```diff
@@ -0,0 +1,81 @@
...
+import { sendFeedback, clearError } from '../actions';
+import DefaultView from '../components/DefaultView';
+import FeedbackForm from '../components/FeedbackForm';
+import FeedbackSubmitted from '../components/FeedbackSubmitted';
+
+class Main extends React.Component {
+
+  constructor(props) {
+    super(props);
+    this.state = { formIsVisible: false, requestPending: false };
```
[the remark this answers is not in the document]
ncksllvn: Yeah good call, the second one isn't needed. Might be better to have the first one in Redux too.

src/js/feedback/components/FeedbackForm.js
ncksllvn: Yeah I was keeping the form values in local state until the form is submitted, then it makes it into Redux.

src/js/feedback/components/FeedbackForm.js
ncksllvn: Oh, I did not even see those components in our common directory. Hopefully our design system will be useful for finding these things in the future.

src/js/feedback/containers/Main.jsx
```diff
@@ -0,0 +1,81 @@
...
+  }
+}
+
+function mapStateToProps(state) {
+  return {
+    ...state.feedback
+  };
+}
+
+function mapDispatchToProps(dispatch) {
```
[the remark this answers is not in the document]
ncksllvn: Wow, did not know you could do that!

src/sass/modules/_m-footer.scss
ncksllvn: Yeah good call. Also, unless I overlooked something, we have no simple wrapper class for just adding something to the page that lines up with the rest of the elements, which is why this has that ugly padding property. It looks like we got mixed up between USWDS, Foundation, and our own classes.

src/sass/modules/_m-footer.scss
ncksllvn: No I just like the style of having it all contained like that. I prefixed the classes just to be explicit. Should I try to flatten this?

src/sass/modules/_m-footer.scss
ncksllvn: Yeah that makes sense, I'll try to get the style cleaned up in general

src/sass/modules/_m-footer.scss
ncksllvn: Yeah, I didn't enjoy writing that CSS but I feel like there were a lot of edge cases in that tiny widget with different viewports.

src/js/feedback/components/FeedbackForm.js
jbalboni: I don't think there's ever a clear answer to that question, but I think we've tried to keep things in Redux unless we have a reason not to.

src/sass/modules/_m-footer.scss
jbalboni: I think avoiding nesting makes sass easier to maintain, so that would be my recommendation.

test/feedback/actions/feedback.unit.spec.js
```diff
@@ -0,0 +1,86 @@
...
+  });
+
+  it('dispatches SEND_FEEDBACK and FEEDBACK_RECEIVED when there is a successful request', (done) => {
+
+    const actionCreator = sendFeedback({ description: 'My feedback description', email: 'test@test.com' });
+    const result = actionCreator(dispatch);
+
+    result.then(() => {
+
+      chai.assert.isTrue(fetch.calledOnce, 'Fetch was called');
```
jbalboni: We use the `expect` syntax for chai almost exclusively (in unit tests). Might be good to switch to that for consistency.

ncksllvn: Okay, I can switch to that, but ahhh, I so much prefer the traditional assert syntax.

src/sass/modules/_m-footer.scss
ncksllvn: I guess I usually look at in inversely, in that I try to keep object state as contained as possible and only let it propagate throughout Redux state if there's a clear incentive to do so.

src/js/feedback/components/FeedbackForm.js
ncksllvn: I guess I usually look at in inversely, in that I try to keep object state as contained as possible and only let it propagate throughout Redux state if there's a clear incentive to do so.

jbalboni: I don't think that's the wrong approach necessarily, but you lose one of the benefits of Redux, which is that all (or the vast majority) of your state is in one place and is changed in similar ways.

In any case: we don't have a hard and fast rule about this and I don't think this way is wrong, just that our other apps more often take a different approach.

ncksllvn:
@jbalboni Thanks for your suggestions! Just published a bunch of changes.  Can you take another look and let me know what you think?

src/js/feedback/components/FeedbackForm.jsx
```diff
@@ -0,0 +1,86 @@
...
+          <div className="feedback-widget-need-help-inner">
+            <h3>Need help?</h3>
+            Calls the Vets.gov Help Desk<br/>
+            <a href="tel:18555747286">1-855-574-7286</a><br/>
+            TTY:&nbsp;<a href="tel:+18008778339">1-800-877-8339</a><br/>
+            Monday – Friday, 8:00 a.m. – 8:00 p.m. (<abbr title="eastern time">ET</abbr>)
+          </div>
+        </div>
+      </div>
+      <input
```
jbalboni: I think there's a component for this one, too: `ErrorableCheckbox`.

src/js/feedback/components/FeedbackForm.jsx
```diff
@@ -0,0 +1,86 @@
...
+        value={props.formValues.shouldSendResponse}
+        onChange={({ target: { checked: shouldSendResponse } }) => props.setFormValues({ shouldSendResponse })}/>
+      <label htmlFor="should-send-response">I would like to receive a response about my feedback.</label>
+      <div className="usa-grid-full">
+        <div className="usa-width-one-third">
+          {props.formValues.shouldSendResponse ? (
+            <div className="feedback-email-container">
+              <ErrorableTextInput
+                label="Your email address"
+                name="email"
```
jbalboni: Probably want a `type="email"` on here, too.

src/js/feedback/components/FeedbackForm.jsx
```diff
@@ -0,0 +1,86 @@
...
+
+function FeedbackForm(props) {
+  return (
+    <form
+      className="feedback-form"
+      onSubmit={(event) => {
+        event.preventDefault();
+        props.sendFeedback();
+      }}>
+      {props.errorMessage ? (
```
jbalboni: This is ultimately up to the designers, but our longer forms tend to put these global errors below the submit buttons (less chance of missing them because of scroll position).

src/js/feedback/actions/index.js
```diff
@@ -0,0 +1,62 @@
...
+  return {
+    type: SET_FORM_VALUES,
+    formValues,
+    formErrors: runValidation(formValues)
+  };
+}
+
+export function sendFeedback() {
+  return (dispatch, getState) => {
+    const { formValues } = getState().feedback;
```
jbalboni: `getState` is sometimes necessary, but it's also very easy to break without knowing it if you change the structure of the Redux state. You could also pass formValues as a parameter.

src/js/feedback/components/FeedbackSubmitted.jsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+
+function FeedbackSubmitted({ shouldSendResponse }) {
+  return (
+    <div id="feedback-submitted">
+      <h4 className="feedback-widget-title">Thank you for your feedback!</h4>
+      {shouldSendResponse ? (
```
jbalboni: Ignore me if you know about this and just prefer the ternary syntax, but this is slightly more concise: `{shouldSendResponse && <p className="feedback-widget-intro">We'll get back to you soon.</p>}`

src/js/feedback/components/FeedbackForm.jsx
```diff
@@ -0,0 +1,86 @@
...
+      <label htmlFor="should-send-response">I would like to receive a response about my feedback.</label>
+      <div className="usa-grid-full">
+        <div className="usa-width-one-third">
+          {props.formValues.shouldSendResponse ? (
+            <div className="feedback-email-container">
+              <ErrorableTextInput
+                label="Your email address"
+                name="email"
+                field={{ value: props.formValues.email, dirty: false }}
+                errorMessage={props.formErrors.email}
```
jbalboni: Does this show an error as you're typing in an email? One pattern we have on our forms is that we don't show validation errors until you blur a field (to let the user finish typing without being yelled at). After that, validation changes are shown on change.

This PR is making me realize we need an easier way of writing small forms with the best practices we have on our bigger forms.

jbalboni (approved):
I made some more comments, but nothing that should stop you from merging. LGTM.

src/js/feedback/components/FeedbackForm.jsx
ncksllvn: I think the blur is probably a better idea but there is some stuff in place that works okay. I'm gonna clean it up though.

src/js/feedback/components/FeedbackSubmitted.jsx
ncksllvn: I swear there was a version of React where that printed `false` in the rendered HTML lol

ncksllvn:
Thanks @jbalboni! I think things are finally in-sync with the way the rest of our project does things. Test cases can be tightened up but it's a fine stopping point I think.

test/feedback/components/FeedbackSubmitted.unit.spec.jsx
```diff
@@ -0,0 +1,21 @@
...
+import { expect } from 'chai';
+
+import FeedbackSubmitted from '../../../src/js/feedback/components/FeedbackSubmitted';
+
+describe('<FeedbackSubmitted/>', () => {
+
+  it('should render with follow-up message', () => {
+    const wrapper = enzyme.shallow(<FeedbackSubmitted shouldSendResponse/>);
+    const text = wrapper.text();
+    expect(text.includes('We\'ll get back to you soon.')).to.be.true;
```
cvalarida: Oh, I think the `'` in `We'll` should be a curly apostrophe (`’`).

test/feedback/containers/Main.unit.spec.jsx
```diff
@@ -0,0 +1,56 @@
...
+  errorMessage: null
+};
+
+describe('<Main/>', () => {
+
+  it('should render with DefaultView', () => {
+    const wrapper = enzyme.shallow(<Main {...defaultProps}/>);
+    const text = wrapper.text();
+    expect(text.includes('Give feedback on this page')).to.be.true;
+    expect(wrapper.find(DefaultView)).to.have.lengthOf(1);
```
cvalarida: Alternatively, you can do `wrapper.find('DefaultView')` as a string like that so you don't have to import it.

test/feedback/feedback.e2e.spec.js
```diff
@@ -0,0 +1,36 @@
...
+function begin(browser) {
+  browser.url(E2eHelpers.baseUrl)
+    .waitForElementVisible(selectors.root, Timeouts.slow)
+    .axeCheck('document');
+
+  browser.click(selectors.revealFormButton).pause(pause);
+  browser.setValue(selectors.formDescription, 'This is my feedback').pause(pause);
+  browser.click(selectors.formShouldSendResponse).pause(pause);
+  browser.setValue(selectors.formEmail, 'test@adhocteam.us').pause(pause);
+  browser.click(selectors.formSubmit).pause(pause);
```
cvalarida: What are these pauses for?

cvalarida (approved):
Looks great! Had some questions and suggestions, but nothing major.

src/js/feedback/components/FeedbackForm.jsx
```diff
@@ -0,0 +1,121 @@
...
+    if (dirty) this.setState({ suppressEmailErrors: false });
+    this.props.setFormValues({ email });
+  }
+
+  sendFeedback = (event) => {
+    event.preventDefault();
+    this.props.sendFeedback({
+      description: this.props.description,
+      email: this.props.email,
+      shouldSendResponse: this.props.shouldSendResponse
```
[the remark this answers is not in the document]
ncksllvn: Yeah that `shouldSendResponse` is strictly used to show and hide the email field on the UI. I put it in Redux for consistency, but I could see having it in state as well.

test/feedback/components/FeedbackSubmitted.unit.spec.jsx
ncksllvn: Good call, will fix

test/feedback/feedback.e2e.spec.js
ncksllvn: It's to make sure each command is finished before starting the next command, and so that I don't need to use callbacks. Are these not necessary? I may not have tested otherwise.

src/js/feedback/actions/index.js
```diff
@@ -0,0 +1,52 @@
...
+    formErrors
+  };
+}
+
+export function sendFeedback(formValues) {
+  return (dispatch) => {
+    const { description, email: ownerEmail } = formValues;
+    const targetPage = window.location.pathname;
+    const settings = {
+      headers: { 'Content-Type': 'application/json', Authorization: '' },
```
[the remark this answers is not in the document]
ncksllvn: @jbalboni, do you all think this is something we should patch in the mock API or API helper function?

src/js/feedback/reducers/index.js
```diff
@@ -0,0 +1,78 @@
...
+      return {
+        ...state,
+        formIsVisible: true
+      };
+    case SET_FORM_VALUES:
+    {
+      const formValues = { ...state.formValues, ...action.formValues };
+      const formErrors = { ...state.formErrors, ...action.formErrors };
+
+      let formIsSubmittable = true;
```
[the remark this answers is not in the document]
cvalarida: Honestly, it's fine like it is. It could be clearer by explicitly splitting
```js
const mapStateToProps = (state) => state.feedback;
```
into
```js
const mapStateToProps = (state) => ({
  formIsSubmittable: state.feedback.formIsSubmittable
  // Etc.
});
```
It seems kinda dumb to do that when you want the whole `state.feedback`, though, so I wouldn't worry about it. I was just being dense. :man_shrugging:

test/feedback/feedback.e2e.spec.js
cvalarida: I think the only time you'd need to pause is when you're waiting for an animation to finish. Even then, I think there's a way @jbalboni had to bypass the animation for tests, but I'd have to dig it back up since I don't remember how it worked exactly.

When we're waiting for things, we usually prefer [`.waitForElementVisible()`](http://nightwatchjs.org/api#waitForElementVisible), but that doesn't seem like it'd be useful here.

If you try it without the pauses and it breaks, I'd be really interested to find out why! Seems like it should be fine without pausing.

cehsu:
Would it make sense to tackle https://github.com/department-of-veterans-affairs/vets.gov-team/issues/6633 as part of this work?

ncksllvn:
@cehsu yeah good idea, that link will be removed but we should make sure the new widget appears in the right order

src/js/feedback/components/FeedbackForm.jsx
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  render() {
+    return (
+      <form className="feedback-form" onSubmit={this.sendFeedback}>
+        <h4 className="feedback-widget-title">Tell us what you think</h4>
+        <div className="va-flex">
+          <div className="feedback-widget-description-container">
+            <ErrorableTextarea
+              label="What can we do to make Vets.gov better?"
```
cehsu: It looks like the class for this label's asterisk (`hca-required-span`) is inconsistent with that of the email label's asterisk (`form-required-span`). This inconsistency is not introduced by these changes, but it might make sense to address it along with them.

cehsu: Also, should the tab order be to go through the form and then to go to the contact information?

ncksllvn: I would rather not dig into the errorable inputs. I'm not the author and this PR is heating up to be pushed through soon. Maybe @jbalboni has some input?

The tab order needs to be fixed though. I don't think the contact info is a big deal because it actually switches order on mobile, but the form itself should probably show/hide instead of an actual render.

PR #6820, opened 2017-11-21
Add official USA Banner

ncksllvn:
The PR adds the official [USWDS banner](https://standards.usa.gov/components/headers/) (top of the screen) and the toggle behavior. I did not use React for this because the only interactivity it has is the toggle and it felt like overkill to use React to render that on every page, but if anyone feels strongly it can be moved into the React pipeline with little effort. If we're planning on integrating it into the design system, maybe it should be anyway.

### Default
image

### Toggled
image

Resolves department-of-veterans-affairs/vets.gov-team/issues/5904


ncksllvn:
This also removes the WIP message from the mobile navigation menu. I don't think we need that message anymore but definitely correct me if I'm wrong.

phiden (approved):

bethpotts:
@ncksllvn @amybehrens I'd like to make two PL tweaks:
From this current language:
The https:// ensures that you are connecting to the official website and that any information you provide is encrypted and transmitted securely.

To this:
The https:// ensures that you're connecting to the official website and that any information you provide is encrypted and sent securely.

bethpotts:
@ncksllvn The wording changes you made look great. Thanks!

src/sass/base/_va.scss
```diff
@@ -215,11 +215,11 @@ table {
   clear: both;
   margin: 0;
   padding: 0;
-  min-height: 147.19px;
+  min-height: 167.19px;
```
jbalboni: Could we add a comment on this change and the next one saying that they're used to set the header height so that it doesn't flash when the menus are rendered? I should have done that when I initially made that change.

jbalboni (approved):

src/sass/base/_va.scss
ncksllvn: Good idea, will do!

PR #6822, opened 2017-11-21
Add message when profile is slow

jbalboni:
This adds a Sentry message when loading the user profile is slow. There's no UI change, I want to get some more info before making changes there.


cehsu (approved):
LGTM.

PR #6823, opened 2017-11-21
Don't show name twice while upload file

jbalboni:
Was doing this:

screen shot 2017-11-21 at 1 56 58 pm

Now it's not (screen disappears very quickly, so I haven't taken a shot of it)

Also fixed the title.


cehsu (approved):
LGTM.

PR #6825, opened 2017-11-21
Add TTF Font

ncksllvn:
Google Web Fonts does not support Internet Explorer. This was news to me, but I believe adding the TTF font format should resolve this.

Resolves department-of-veterans-affairs/vets.gov-team/issues/6144


cehsu (approved):
LGTM. This is a bit different but I assume as you intended: 
0003screenshot

ncksllvn:
@cehsu ah that is actually Georgia I think. I've tested this deployment from Sauce Labs and on my home desktop computer and it looks like IE is working with the TTF format of Bitter now. Is that IE 11 on Windows 10? I haven't tested Edge but I can't imagine that Edge doesn't support WOFF or TTF.

cehsu:
Sounds good @ncksllvn, the screenshot is of Edge (looks the same on 13, 14, 15).

ncksllvn:
Thank you @cehsu!

I'm gonna mark this as done because IE and Edge both look nice in my Sauce Labs and desktop testing. We can add another format if we hear otherwise but I feel good about it.

Images below show Bitter rendering correctly.

## IE9 
image

## Edge 13
image

PR #6828, opened 2017-11-21
Appeals v2 -- Timeline scrolling

cvalarida:
So it's not so much scrolling as it is popping in and out, but keeping the button at the same position relative to the viewport as it was before.

I wasn't able to record a gif for reasons that remain a mystery to me, but it works, honest! Give it a shot!

Cross-browser testing is probably warranted, too.


src/js/claims-status/components/appeals-v2/Timeline.jsx
```diff
@@ -21,6 +21,18 @@ class Timeline extends React.Component {
...
+    if (prevState.expanded !== this.state.expanded) {
+      const x = (window.pageXOffset || document.documentElement.scrollLeft);
+      // I think this is too slow...should be happening after rendering, but it's like it's getting the previous button location
+      const topOfButton = document.querySelector('li button.va-button-link').getBoundingClientRect().top;
+      const y = (window.pageYOffset || document.documentElement.scrollTop);
+      // Put the expander button in the same position it was in before relative to the window
+      window.scrollTo(x, y + topOfButton - this.prevTopOfButton);
+    }
+  }
+
```
markgreenburg: Cool!

markgreenburg (approved):
LGTM; nifty trick! I tested on Chrome, FF, & Safari. I did notice a little bit of thrashing / flashing on Safari but couldn't reproduce in either of the other browsers. One thing to note is that I think this approach may have accessibility implications - automatically moving users further down the page may not make sense to a screen reader user - we probably need to think about what to do with focus when the timeline is toggled. Probably worth looking into in another ticket.

cvalarida:
@markgreenburg Yeah, that's a good point accessibility-wise. I don't _think_ it'll affect them since I'm not changing the focus at all. We're only moving the viewport down. Definitely something to watch out for when we do the 508 work on this, though.

PR #6832, opened 2017-11-22
Post 9/11 print page fix

cvalarida:
Because of the session issue, we'll just keep it in the same window for now.

@rroueche I can make a "Go back" button and have it not show up on the print page if that'll help. I'm guessing the reason behind opening in a new tab is so the veteran doesn't get confused about needing to go back.

### DIFF
--- a/src/js/post-911-gib-status/containers/StatusPage.jsx
+++ b/src/js/post-911-gib-status/containers/StatusPage.jsx
@@ -1,5 +1,4 @@
 import React from 'react';
-import { withRouter } from 'react-router';
 import { connect } from 'react-redux';
 
 import FormTitle from '../../common/schemaform/FormTitle';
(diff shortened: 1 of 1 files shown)


rroueche:
@cvalarida go ahead and merge w/o a back link - I think we need research before we'd add it. 3 reasons:

1) VA has given an image of this "letter" to schools so they know what to accept from students, so I'm wary to change anything about it unless/until we're ready for a V2 that we can also circulate to the schools.

2) Back button is a super highly used browser feature, and in this case I don't think there would be confusion of what it would bring the user back to, so I think it's ok to rely on it for this.

3) We have V2 functionality for print/download that we should launch in December, so it'll only be in this state for about a month if that.

markgreenburg (approved):
LGTM!

PR #6836, opened 2017-11-24
Close modals when the escape key is pressed

ncksllvn:
Resolves department-of-veterans-affairs/vets.gov-team/issues/5161

### DIFF
--- a/src/js/common/components/Modal.jsx
+++ b/src/js/common/components/Modal.jsx
@@ -27,10 +27,6 @@ class Modal extends React.Component {
     this.state = { lastFocus: null, focusListener: null };
   }
 
-  componentWillMount() {
-    document.addEventListener('keyup', this.handleDocumentKeyUp, false);
-  }
-
   componentDidMount() {
     if (this.props.visible) {
       document.body.classList.add('modal-open');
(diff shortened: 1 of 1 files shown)


src/js/common/components/Modal.jsx
```diff
@@ -32,8 +35,10 @@ class Modal extends React.Component {
 
   componentWillReceiveProps(newProps) {
     if (newProps.visible && !this.props.visible) {
+      document.addEventListener('keyup', this.handleDocumentKeyUp, false);
       this.setState({ lastFocus: document.activeElement, focusListener: focusListener(newProps.focusSelector) });
     } else if (!newProps.visible && this.props.visible) {
+      document.removeEventListener('click', this.handleDocumentClick, false);
```
U-DON: What is `handleDocumentClick`? Doesn't seem to be defined in this class.

ncksllvn: Wow, good call. Not sure how that wound up there.

src/js/common/components/Modal.jsx
```diff
@@ -32,8 +35,10 @@ class Modal extends React.Component {
 
   componentWillReceiveProps(newProps) {
     if (newProps.visible && !this.props.visible) {
+      document.addEventListener('keyup', this.handleDocumentKeyUp, false);
       this.setState({ lastFocus: document.activeElement, focusListener: focusListener(newProps.focusSelector) });
     } else if (!newProps.visible && this.props.visible) {
+      document.removeEventListener('click', this.handleDocumentKeyUp, false);
```
U-DON: Isn't `keyup` the right event to unbind the listener from rather than `click`?

src/js/common/components/Modal.jsx
```diff
@@ -50,10 +55,17 @@ class Modal extends React.Component {
   }
 
   componentWillUnmount() {
+    document.removeEventListener('click', this.handleDocumentClick, false);
```
U-DON: `handleDocumentClick` is still here too.

U-DON (changes requested):

src/js/common/components/Modal.jsx
ncksllvn: Yeah it definitely is. Wow, this is sloppiest PR I've ever submitted. Sorry man, thanks for the thorough CR.

U-DON (approved):
LGTM.

PR #6837, opened 2017-11-24
Added link in caution flag specific to Ashford University.

U-DON:
Resolves department-of-veterans-affairs/gi-bill-comparison-tool#477.

> screen shot 2017-11-24 at 4 30 26 pm

The message before the link will have to be updated separately in GIDS. The Iowa link removed from the modal pertains to Ashford as well, and the relevant section is directly below the content that the "more information" link leads to.

### DIFF
--- a/src/js/gi/components/profile/CautionaryInformation.jsx
+++ b/src/js/gi/components/profile/CautionaryInformation.jsx
@@ -39,9 +39,18 @@ export class CautionaryInformation extends React.Component {
     const it = this.props.institution;
     if (!it.complaints) { return null; }
 
+    // If Ashford, show specific link.
+    const schoolSpecificLink =
+      (it.facilityCode === '21007103' || it.website === 'http://www.ashford.edu') && (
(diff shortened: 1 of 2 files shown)


U-DON:
CC: @melwoodard

melwoodard:
@U-DON can you make sure there's a period after the word "benefits" and before the more info link?

bshyong (approved):
LGTM besides Mel's comment on adding a period to the end of the `cautionFlagReason` content

U-DON:
@melwoodard: The period has to come from the copy that's set in GIDS. Note that [in Ashford's profile on prod](https://www.vets.gov/gi-bill-comparison-tool/profile/21007103), there is a period, but we aren't setting that ourselves. That's just the data entered in GIDS.

melwoodard:
Ah, that's right. Sounds good then!

SPriceVA:
When will this go live?

SPriceVA:
Also we would like to keep this language 'VA has notified Ashford that benefit payments may be suspended as early as January 8, 2018, unless corrective actions are taken'

PR #6842, opened 2017-11-27
Add tests for pre-need sponsor other name page

jbalboni:


cehsu (approved):
LGTM. Is it useful to include a minimal submission test case (with no selected)?

PR #6845, opened 2017-11-27
Appealsv2 alerts bugfix mg

markgreenburg:
Addresses a few small issues from the original Alerts component PR, namely:
- Shape of the Alert object didn't make sense and wasn't likely to match the eventual API. Needed to update some helpers, as well as mock data and unit tests.
- Some linting
- Deleted unnecessary ref to own key in `Alert` component. Not only do React components not receive their own `key` prop, but they also don't need it in this case since the key was given to the subcomponent in the `Alerts` component's `render` function. Oops.

### DIFF
--- a/src/js/claims-status/components/appeals-v2/Alerts.jsx
+++ b/src/js/claims-status/components/appeals-v2/Alerts.jsx
@@ -22,7 +22,6 @@ const Alerts = (props) => {
 Alerts.propTypes = {
   alerts: PropTypes.arrayOf(PropTypes.shape({
     type: PropTypes.string.isRequired,
-    date: PropTypes.string,
     details: PropTypes.object
   }))
 };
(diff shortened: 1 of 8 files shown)


src/js/claims-status/utils/helpers.js
```diff
@@ -283,18 +283,21 @@ export const mockData = {
         ],
         alerts: [
           {
-            type: 'warning',
+            type: 'waiting_on_action',
             date: '09-21-2017',
             details: {
-              type: 'waiting_on_action'
+              representative: 'Mr. Spock'
```
cvalarida: 🖖

src/js/claims-status/components/appeals-v2/Alerts.jsx
```diff
@@ -23,7 +23,7 @@ Alerts.propTypes = {
   alerts: PropTypes.arrayOf(PropTypes.shape({
     type: PropTypes.string.isRequired,
     date: PropTypes.string,
```
cvalarida: `date` wasn't in the Swagger docs; does it still make sense here instead of inside `details`?

cvalarida (approved):

src/js/claims-status/components/appeals-v2/Alerts.jsx
markgreenburg: Oh hmm, I could've sworn it was in there but it definitely isn't. I think I must've pulled the date property from `events`

PR #6846, opened 2017-11-28
Additional question for discharge wizard

bshyong:
- adds additional question and logic regarding previous period of service
- also includes copy changes and logic adjustments based on latest discussions 

screen shot 2017-11-28 at 1 17 11 am

resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/6059


U-DON (approved):
LGTM. The numbering/ordering seems to be getting more and more difficult to maintain and seems like it will be in much need of a refactor once everything has settled.

bshyong:
Definitely. I am thinking about the best way to refactor this when we are settled. We ended up reordering the questions and are adding some more conditional content..

PR #6847, opened 2017-11-28
Fix order of imported files

jbalboni:
This fixes a Sentry error that I think is related to the connected issue.

### DIFF
--- a/src/js/common/polyfills.js
+++ b/src/js/common/polyfills.js
@@ -28,7 +28,9 @@ if (navigator.userAgent.includes('Edge/14')) {
   window.fetch = undefined;
 }
 
-import 'whatwg-fetch';
+// This needs to stay as require because import causes it to be executed before the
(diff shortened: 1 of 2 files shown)


src/js/common/polyfills.js
```diff
@@ -28,7 +28,9 @@ if (navigator.userAgent.includes('Edge/14')) {
   window.fetch = undefined;
 }
 
-import 'whatwg-fetch';
+// This needs to stay as require because import causes it to be executed before the
+// above code
+require('whatwg-fetch');
```
ncksllvn: Interesting, did not know that!

ncksllvn (approved):
LGTM

src/js/common/polyfills.js
jbalboni: Yeah, me neither.

PR #6848, opened 2017-11-28
Add sponsor/applicant service tests, show branch name

jbalboni:
Also updated the service period view to show the label instead of the id:

screen shot 2017-11-28 at 11 44 36 am


cehsu (approved):
LGTM.

src/js/pre-need/components/ServicePeriodView.jsx
```diff
@@ -0,0 +1,19 @@
...
+  let to = '';
+  if (formData.dateRange) {
+    from = formatReviewDate(formData.dateRange.from);
+    to = formatReviewDate(formData.dateRange.to);
+  }
+
+  return (
+    <div>
+      <strong>{serviceLabels[formData.serviceBranch]}</strong><br/>
+      {from} &mdash; {to}
```
cehsu: Is there is a reason for &mdash; rather than —, as we use both, and with varying spacing conventions.

src/js/pre-need/utils/helpers.js
```diff
@@ -291,6 +292,7 @@ export const veteranUI = {
   }
 };
 
+
```
cehsu: Perhaps this is an extra line?

src/js/pre-need/utils/labels.js
```diff
@@ -0,0 +1,57 @@
...
+  AC: 'US ARMY AIR CORPS',
+  AF: 'US AIR FORCE',
+  AR: 'US ARMY',
+  CG: 'US COAST GUARD',
+  CV: 'CIVILIAN WAKE ISLAND NAS',
+  FP: 'CIVILIAN FERRY PILOT',
+  MM: 'US MERCHANT MARINE',
+  PH: 'US PUBLIC HEALTH SERVICE',
+  NN: 'NAVY NURSE CORPS',
+  WA: 'WOMEN\'S ARMY AUX CORPS',
```
cehsu: Do these need to be curly instead?

test/util/schemaform-utils.jsx
```diff
@@ -208,7 +208,7 @@ export function getFormDOM(form) {
 
   // TODO: Remove fillDate from unit-helpers and prefer this one
   formDOM.fillDate = function populateDate(partialId, dateString) {
-    fillDate(this, partialId, dateString);
+    oldFillDate(this, partialId, dateString);
```
cehsu: Obsolete comment on line 210?

cehsu (approved):
LGTM. Had a few minor questions/comments.

src/js/pre-need/components/ServicePeriodView.jsx
jbalboni: Good question, I don't remember. I copied this from another component, so it's at least following what we normally do for the service periods view field.

PR #6851, opened 2017-11-28
Add tests for sponsor mailing address

jbalboni:


cehsu (approved):
LGTM.

test/pre-need/config/sponsorMailingAddress.unit.spec.jsx
```diff
@@ -0,0 +1,59 @@
...
+    const onSubmit = sinon.spy();
+    const form = mount(
+      <DefinitionTester
+        schema={schema}
+        definitions={formConfig.defaultDefinitions}
+        onSubmit={onSubmit}
+        uiSchema={uiSchema}/>
+    );
+
+    fillData(form, 'input#root_application_veteran_address_street', 'Test');
```
cehsu: Should we consider creating reusable constants for these commonly used values (such as "Test"), or is it better to have the values exposed/visible in the tests themselves for maintainability?

cehsu (approved):

test/pre-need/config/sponsorMailingAddress.unit.spec.jsx
jbalboni: I don't think the constants buys us much, but we could potentially use the data in the json files for the schema tests.

PR #6855, opened 2017-11-28
Update error message for ineligible VIC users

patrickvinograd:
Fixes https://github.com/department-of-veterans-affairs/vets.gov-team/issues/6283

Duplication is probably something to be cleaned up, but guards against the id_card API returning an error status if somehow they get past the fact that they don't have the veteran_id_card in their list of available services.


ncksllvn (approved):
LGTM

PR #6856, opened 2017-11-29
Add GA events for VIC submission, success, failure

robertfairhead:
Adds three events to track how many submissions total and how many succeed and fail through GA.

I don't try to track the exact error codes. That can be added if needed, but the assumption is those will be picked up by one of the backend monitoring systems already for service improvements.


U-DON (approved):
LGTM.

PR #6863, opened 2017-11-29
Add screen reader content to exit-icon mixin, over-ride in footer

markgreenburg:
Adds "opens in new tab" screen reader content to all instances where the `exit-icon` mixin is being actively applied. To check for side-effects, I looked at all refs in our css to `background: none;` as well as `background-image: none;`, and checked all of the instances where related classes are used. As far as I can tell, I've hidden the screenreader content from all cases where the icon was hidden as well.

As for functionality, I'm only able to test on VoiceOver. Even though this content is applied with an `:after` pseudo-element selector on the links themselves, VoiceOver announces the SR content first:
> `link, opens in a new tab, download Form 21P-527EZ`. . .

which is probably a good thing anyway.

Would really appreciate some IE testing

### DIFF
--- a/src/sass/base/_b-mixins.scss
+++ b/src/sass/base/_b-mixins.scss
@@ -63,5 +63,19 @@
   background-size: 1em auto;
   padding-right: 1.2em;
   position: relative;
+
+  &:after {
+    // adds non-visible alt content for screen readers
+    position: absolute;
+    height: 1px;
+    width: 1px;
+    overflow: hidden;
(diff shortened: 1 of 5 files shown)


cvalarida (approved):
I'm gonna be straight with you--I didn't test in IE. We can make sure it doesn't look super messed up using browserstack once this hits staging.

Tested in Chrome on El Capitan using VoiceOver and it was great! 👍

PR #6865, opened 2017-11-30
Add veteran info test for pre-need

jbalboni:
I also added a missing title that was causing a prop type warning, and a missing validation message:

screen shot 2017-11-30 at 9 49 21 am


cehsu (approved):
LGTM.

PR #6866, opened 2017-11-30
Add support for footer navigation analytics

robertfairhead:
Closes: department-of-veterans-affairs/vets.gov-team#6119

Adds a custom event on a link clicks in the footer navigation items so we can distinguish their use

Also changes footer nav links to be relative so that staging, etc. don't always point to production

### DIFF
--- a/content/includes/footer.html
+++ b/content/includes/footer.html
@@ -4,30 +4,30 @@
     <div class="row va-flex">
     <ul class="va-footer-linkgroup" id="footer-services">
       <li><h4 class="va-footer-linkgroup-title">Services and Benefits</h4></li>
-      <li><a href="/disability-benefits/" onClick=footerNavClick>Disability Benefits</a></li>
-      <li><a href="/health-care/" onClick=footerNavClick>Health Care Benefits</a></li>
-      <li><a href="/education/" onClick=footerNavClick>Education Benefits</a></li>
-      <li><a href="/housing-assistance/" onClick=footerNavClick>Housing Assistance</a></li>
-      <li><a href="/employment/" onClick=footerNavClick>Careers and Employment</a></li>
-      <li><a href="/life-insurance/" onClick=footerNavClick>Life Insurance</a></li>
-      <li><a href="/pension/" onClick=footerNavClick>Pension Benefits</a></li>
-      <li><a href="/burials-and-memorials/" onClick=footerNavClick>Burials and Memorials</a></li>
+      <li><a href="/disability-benefits/" onClick="footerNavClick">Disability Benefits</a></li>
+      <li><a href="/health-care/" onClick="footerNavClick">Health Care Benefits</a></li>
+      <li><a href="/education/" onClick="footerNavClick">Education Benefits</a></li>
+      <li><a href="/housing-assistance/" onClick="footerNavClick">Housing Assistance</a></li>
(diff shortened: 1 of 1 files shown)


robertfairhead:
Confirmed the events are firing properly

screen shot 2017-11-30 at 1 31 41 pm

I don't _love_ doing it this way with the random JS at the bottom and adding the `onClick` manually to each link. I played around with using some YAML to define the links and a liquid for loop to roll it out but it required some more serious changes to the build system to add in some metalsmith plugins. didn't feel worth it just for here.

but i may revisit that when header nav comes around since that may be more complex.

phiden (approved):

content/includes/footer.html
```diff
@@ -126,6 +126,14 @@ <h3>Get help from Veterans Crisis Line</h3>
 “To care for him who shall have borne the battle and for his widow, and his orphan.”
 - Abraham Lincoln
 -->
+<script type="text/javascript">
+  function footerNavClick(elem) {
+    window.dataLayer.push({
+      event: `nav-${elem.parentElement.parentElement.id}`,
```
ncksllvn: I'm not sure if we should use ES6 templates here because I don't think our preprocessor will parse JS in content files. Just an idea, but would it possibly be better to have this as a second parameter instead of traversing the DOM? Just thinking that in case there are some changes to the footer later we don't have to revisit that function.

ncksllvn (approved):
LGTM

content/includes/footer.html
robertfairhead: yeah, it kinda sucks either way because it makes the onclick pretty long and the HTML harder to read. i was trying to use the templates to insert it while keeping the HTML readable but sort of gave up.

```
    <ul class="va-footer-linkgroup" id="footer-services">
      <li><h4 class="va-footer-linkgroup-title">Services and Benefits</h4></li>
      <li><a href="/disability-benefits/" onClick="footerNavClick(this, 'footer-services');">Disability Benefits</a></li>
      <li><a href="/health-care/" onClick="footerNavClick(this, 'footer-services');">Health Care Benefits</a></li>
      <li><a href="/education/" onClick="footerNavClick(this, 'footer-services');">Education Benefits</a></li>
      <li><a href="/housing-assistance/" onClick="footerNavClick(this, 'footer-services');">Housing Assistance</a></li>
      <li><a href="/employment/" onClick="footerNavClick(this, 'footer-services');">Careers and Employment</a></li>
      <li><a href="/life-insurance/" onClick="footerNavClick(this, 'footer-services');">Life Insurance</a></li>
      <li><a href="/pension/" onClick="footerNavClick(this, 'footer-services');">Pension Benefits</a></li>
      <li><a href="/burials-and-memorials/" onClick="footerNavClick(this, 'footer-services');">Burials and Memorials</a></li>
```

felt kinda verbose and repetitive to me. but i agree. i don't like tying it to the DOM. I'm like 2 out of 10 on how strong i feel about it either way. If you're higher than that one way or another let me know, and i'll change it up.

but good call on removing the ES6 template string. i'll break that out.

ncksllvn: Yeah I see what you mean for sure, not really a pretty way to do it here. No strong feelings here, probably a 1. It's unlikely that markup will ever not be a list like that anyway.

PR #6867, opened 2017-11-30
Refactor FormPage to remove SiP logic

jbalboni:
This removes the SiP and routing logic from `FormPage`. The routing logic is now in a `routing.js` file, which should eventually contain the other routing related helpers in the `helpers.js` file.

The SiP logic is now in `RoutedSiPPage`, which wraps `FormPage`.

I also moved some of the SiP components to a `save-in-progress` folder. I think we can continue doing this as we go along, but I didn't want to clutter up this PR more.

I also think we should change `FormPage` to `RoutedFormPage`, to make it more clear that it depends on React Router, but similar to above, I don't want to make the diff harder to read.


src/js/common/schemaform/FormPage.jsx
```diff
@@ -77,11 +54,10 @@ class FormPage extends React.Component {
       newData = _.set([this.props.route.pageConfig.arrayPath, this.props.params.index], formData, this.props.form.data);
     }
     this.props.setData(newData);
-    this.debouncedAutoSave();
   }
 
   onSubmit({ formData }) {
```
cvalarida: Just making sure this is intentionally not `onSubmit = ({ formData }) => {...}` after you got rid of `this.onSubmit = this.onSubmit.bind(this)`.

cvalarida (commented):
Preliminary review--I didn't make it through the whole thing before I got distracted; I'll loop back around and look at the whole thing later.

src/js/common/schemaform/FormPage.jsx
jbalboni: Nope, not intentional. Good catch.

src/js/common/schemaform/helpers.js
```diff
@@ -71,11 +72,12 @@ export function createPageList(formConfig, formPages) {
 export function createRoutes(formConfig) {
   const formPages = createFormPageList(formConfig);
   const pageList = createPageList(formConfig, formPages);
+  const Page = formConfig.disableSave ? FormPage : RoutedSiPPage;
```
annekainicUSDS: This is also very minor, but maybe it's safer to actually have `enableSave` in the formConfig vs. defaulting to enabling it by having to specify `disableSave`.

annekainicUSDS (approved):
LGTM! A few comments

src/js/common/schemaform/helpers.js
```diff
@@ -1,6 +1,7 @@
 import _ from 'lodash/fp';
-import FormPage from './FormPage';
 import ReviewPage from './review/ReviewPage';
+import FormPage from './FormPage';
+import RoutedSiPPage from './save-in-progress/RoutedSiPPage';
```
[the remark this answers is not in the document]
jbalboni: Yep, that sounds better to me.

src/js/common/schemaform/helpers.js
jbalboni: We did that because the expectation is that all of the forms will have save in progress, but yeah, maybe the reverse would make sense if non vets teams use this. We can consider changing that later, I think.

cehsu (approved):
LGTM. Had a minor question.

PR #6871, opened 2017-11-30
Add custom events for side nav analytics

robertfairhead:
Closes: department-of-veterans-affairs/vets.gov-team#6120


robertfairhead:
These are more straightforward since they're already in templates

ncksllvn (approved):
LGTM

PR #6873, opened 2017-11-30
Discharge Wizard updated copy and URL

bshyong:
- Adds content changes
- Sets correct URL for app
- gets rid of static entry page 

- resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/5645


U-DON (approved):
LGTM.

PR #6874, opened 2017-11-30
Force all users to be considered as veterans, update test

markgreenburg:
Makes all users veterans, left some stuff in comments to make this easy to revert once we want to split out dependents and veterans again.


cvalarida (approved):
Pretty straightforward. Good call on leaving the comment. 👍

PR #6875, opened 2017-12-01
Nav accessibility

cvalarida:
It's a bit of a beast. Sorry.

This PR does a few things to make the main navigation more accessible:
1. Makes the menu buttons read like pop up buttons (which is apparently the preferred way for screen readers to communicate this)
2. Removes excess `aria-*` attributes until they're necessary
3. Adds keyboard navigation to the menu
    - Arrow keys and escape are now supported

There are a few things that are wonky with this, but I think it's probably alright for now. Notably weird are:
- The only way to get back to the main nav from a submenu is to navigate back up to it
- Hitting the left arrow key from the second nav menu (the one without a third-level menu) moves the focus to the main nav menu button
- When the last submenu is reached with `->` and I hit `->` again, I expect it to go to the next main nav menu, but it doesn't
  - Now that I think about it, I have visibility on the top-level `menuElement`, so I can make this happen, but Trevor, Ad Hoc's Accessibility Guru™, said he expected it to do nothing in this case

I'm probably missing some other edge cases, but like I said, I think it's good enough to merge now and address the issues later.

There are some outstanding things yet to be done such as:
1. Document the assumptions on the DOM structure being passed into `addMenuListeners`
2. Possibly create another helper we could feed the DOM to which adds all the appropriate `aria-*` attributes
3. Write unit tests?
4. Remove dependency on `src/js/legacy/mega-menu.js`
    - This should be pretty simple now


src/js/common/utils/accessible-menus.js
```diff
@@ -0,0 +1,349 @@
+const LEFT_ARROW = 37;
+const UP_ARROW = 38;
+const RIGHT_ARROW = 39;
+const DOWN_ARROW = 40;
+const ESCAPE = 27;
+
+function isWideScreen() {
+  return matchMedia('(min-width: 768px)').matches;
```
markgreenburg: That's nifty

src/js/common/utils/accessible-menus.js
```diff
@@ -0,0 +1,349 @@
...
+ * @param {HTMLElement}  element
+ * @param {function}     testCB   When this returns true on an ancestor, that ancestor is returned
+ * @return {HTMLElement|null}     If there is no ansestor for which the test returns true, null is returned
+ */
+function findNearestAncestor(element, testCB) {
+  const parent = element.parentElement;
+  if (!parent) {
+    return null;
+  }
+  const returnVal = testCB(parent) ? parent : findNearestAncestor(parent, testCB);
```
markgreenburg: OK, so implicitly `testCB` should return a bool

src/js/common/utils/accessible-menus.js
```diff
@@ -0,0 +1,349 @@
...
+function isMenuButton(element) {
+  // There is no element.getElementById() :matrix:
+  const menuElement = element.parentElement.querySelector(`#${element.getAttribute('aria-controls')}`);
+  if (!menuElement) {
+    return false;
+  }
+
+  const menuRole = (menuElement.getAttribute('role') || '').toLowerCase();
+
+  return (element.tagName.toLowerCase() === 'button' || element.getAttribute('role').toLowerCase() === 'button') &&
```
markgreenburg: This is fine, but IMO it might help readability if the conditions were split up and assigned to constants or one-line functions that described what's being tested.

src/js/common/utils/accessible-menus.js
```diff
@@ -0,0 +1,349 @@
...
+    ['menu', 'true'].includes(element.getAttribute('aria-haspopup')) &&
+    // The menuElement either is or contains a menu element
+    !!(menuRole === 'menu' || menuElement.querySelector('[role="menu"]'));
```
markgreenburg: This is fine, but IMO it might help readability if the conditions were split up and assigned to constants or one-line functions that described what's being tested so that this big check reads more naturally

src/js/common/utils/accessible-menus.js
```diff
@@ -0,0 +1,349 @@
...
+ *  is reached, it wraps to the last or first element respectively. If an element isn't visible, it's
+ *  skipped.
+ *
+ * @param {HTMLLIElement} element  The <li> containing the menu item
+ * @param {String} direction     The direction to move the focus. Possible options: 'previous', 'next'
+ */
+function moveFocus(element, direction) {
+  const focusElement = direction === 'previous' ?
+    (element.previousElementSibling || element.parentNode.lastElementChild) :
+    (element.nextElementSibling || element.parentNode.firstElementChild);
```
markgreenburg: Nice, so previous / next or wrap if it doesn't exist.

src/js/common/utils/accessible-menus.js
```diff
@@ -0,0 +1,349 @@
...
+
+  // If focusElement isn't visible, recurse
+  if (!isVisible(focusElement)) {
+    moveFocus(focusElement, direction);
+  } else {
+    // The focusElement should be a <li>, so we want to focus on the contents
+    focusElement.firstElementChild.focus();
+  }
```
markgreenburg: Nice!

src/js/common/utils/accessible-menus.js
```diff
@@ -0,0 +1,349 @@
...
+/**
+ * Gets the menu and associated menu button from an element if possible. Returns null if not.
+ *
+ * @param {HTMLElement} menuLi  The element (probably <li>) that may contain a menu
+ * @return {Object|null}        If the element contains a proper menu structure, returns an object
+ *                               { menu: <HTMLElement>, menuButton: <HTMLElement> }
+ *                               Otherwise, returns null
+ */
+function getMenuStructure(menuLi) {
+  const menuButton = menuLi.querySelector('button, [role="button"]');
```
markgreenburg: So we have buttons inside links. 🤔

src/js/common/utils/accessible-menus.js
```diff
@@ -0,0 +1,349 @@
...
+  if (openSubMenu) {
+    openMenu(menu.firstElementChild, false, false);
+  }
+}
+
+
+/**
+ * Creates a closure which closes all menus in the provided element.
+ *
+ * @param {HTMLElement}
```
markgreenburg: I assume this `{HTMLElement}` is called `menuElement`

src/js/common/utils/accessible-menus.js
```diff
@@ -0,0 +1,349 @@
...
+ */
+function closeAll(menuElement) {
+  // Get all the parent <li>s of open menus
+  const allMenus = Array.from(menuElement.querySelectorAll('[aria-expanded=true]'))
+    .map(e => findNearestLi(e));
+
+  // Add the top-level 
+  if (menuElement.getAttribute('aria-expanded') === true) {
+    allMenus.push(findNearestLi(menuElement));
+  }
```
markgreenburg: Interesting. Why isn't this found with the `allMenus` query selector?

src/js/common/utils/accessible-menus.js
```diff
@@ -0,0 +1,349 @@
...
+ */
+export default function addMenuListeners(menuElement) {
+  const menuRole = menuElement.getAttribute('role');
+  if (!['menubar', 'menu'].includes(menuRole)) {
+    // If we don't have a menubar or menu, don't continue
+    return;
+  }
+
+  // For all the sub menus, add listeners for:
+  //  Up, down, left, right, enter (if necessary), space (if necessary)
```
markgreenburg: 😬

src/js/common/utils/accessible-menus.js
```diff
@@ -0,0 +1,349 @@
...
+      case UP_ARROW: {
+        const isMB = isMenuButton(event.target);
+        if (inMenubar && isMB) {
+          // Open the menu, focus on the last item
+          event.preventDefault();
+
+          // Only open the first menu; we'll manually open the next level
+          openMenu(targetLi);
+
+          if (isWideScreen()) {
```
markgreenburg: yikes

markgreenburg (approved):
LGTM. This is a ton of work, nice going! Wish I had more constructive criticism but there wasn't a lot that jumped out at me. I pulled your branch down and tested with VO + Chrome - seems to work like a charm!

src/js/common/utils/accessible-menus.js
cvalarida: 😛

src/js/common/utils/accessible-menus.js
cvalarida: Ahh, right. Make like a Cyberman and DELETE!

src/js/common/utils/accessible-menus.js
```diff
@@ -0,0 +1,349 @@
...
+              moveFocus(targetLi, 'previous');
+            }
+          } else {
+            moveFocus(targetLi, 'previous');
+          }
+        }
+        break;
+      }
+
+      case DOWN_ARROW: {
```
cehsu: I am not familiar with the expected behavior of mega menus, if this opens a menu, should space and enter do the same ([example 1](http://adobe-accessibility.github.io/Accessible-Mega-Menu/), [example 2](https://dequeuniversity.com/library/aria/menus/megamenu#))?
It also looks like [some examples](https://staff.washington.edu/tft/tests/menus/simplyaccessible/index.html)  open as soon as a menu option receives focus, so it's not clear that there is a well-established standard (but you probably found more helpful resources), though the first two examples seem pretty solid.
It is interesting to note the use of anchor tags in the Deque and W3 examples, as this sort of duplicate event handling can be simplified by using click (as according to the W3 example: "Despite the name, click events are activated regardless of the input method as soon as the link gets activated.")

cehsu (approved):
LGTM. Had some general questions.

src/js/common/utils/accessible-menus.js
```diff
@@ -0,0 +1,349 @@
...
+export default function addMenuListeners(menuElement) {
+  const menuRole = menuElement.getAttribute('role');
+  if (!['menubar', 'menu'].includes(menuRole)) {
+    // If we don't have a menubar or menu, don't continue
+    return;
+  }
+
+  // For all the sub menus, add listeners for:
+  //  Up, down, left, right, escape
+  menuElement.addEventListener('keydown', (e) => {
```
cehsu: Could it be advantageous to attach separate listeners to menubar, menu, and menuitem or would that unnecessarily duplicate code?

src/js/common/utils/accessible-menus.js
```diff
@@ -0,0 +1,349 @@
...
+  }
+
+  // For all the sub menus, add listeners for:
+  //  Up, down, left, right, enter (if necessary), space (if necessary)
+  menuElement.addEventListener('keydown', (event) => {
+    const targetLi = event.target.parentElement;
+    // Target's grandparent because the parent is a <li>
+    const inMenubar = targetLi.parentElement.getAttribute('role').toLowerCase() === 'menubar';
+    const inMenu = targetLi.parentElement.getAttribute('role').toLowerCase() === 'menu';
+
```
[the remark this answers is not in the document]
cvalarida: Since this code will be run on _all_ `keydown` events (in the scope of `menuElement`), I didn't want to call `isMenuButton()` every time because it's relatively expensive (uses `querySelector()` once or twice) and not always needed.

As far as using a class goes, I was hoping to use this function to apply to any menu, so I wanted to reduce markup assumptions as best I could. It may be that adding a class _is_ better / more clear after all, but I was hoping to make this fairly generic.

src/js/common/utils/accessible-menus.js
cvalarida: That's how I was originally going to do it, actually. As I was starting to implement it, however, I went this way because it seemed like a lot of code duplication. It might still make more sense to split them up, but...I'm not sure, honestly.

src/js/common/utils/accessible-menus.js
cvalarida: Because `menuElement.querySelector()` won't select `menuElement`, so I have to look for it specifically. Alternatively, I could look at `menuElement`'s parent, but then it'd also look at any of `menuElement`'s siblings and could have unexpected side-effects, so I decided to just do it this way. :man_shrugging:

src/js/common/utils/accessible-menus.js
cvalarida: Actually, now that I think about it, it might be best to not add it to the individual items to allow for items added to the menu dynamically after its creation. May not ever happen, but future proofing is rarely bad (if it's still performant and maintainable, of course).

cehsu: Right, event delegation would be used in that case, but the idea would be to separate the logic if desirable.

PR #6876, opened 2017-12-01
Limit VIC render to 10% of requests

jkassemi:
VIC services are currently struggling to accomodate the increased load. This limits the number of Veterans who will be making a request by limiting the interface to 10% of requests.

While folks may refresh and try again, this should ease the burden somewhat. We can tie into beta registrations if we need to enable this on a per-user basis, but I'd prefer being able to simply revert this change when the service is better able to handle the traffic.


patrickvinograd (approved):
Looks...well I won't say "good"...to me. 

Approved pending general consensus on message content.

jkassemi:
We have approval on language, just waiting for tests to pass and will confirm we need to ship before merging and setting up a production release.

jkassemi:
One last update to the content. Will merge after tests run and get this out to prod

PR #6881, opened 2017-12-01
change sourcemap to direct s3 link

wyattwalter:
Download sourceMap directly from S3 rather than using the production URL.

Addresses: https://github.com/department-of-veterans-affairs/vets.gov-team/issues/6285


ncksllvn (approved):
LGTM

PR #6882, opened 2017-12-01
Comment out session refresh code

ncksllvn:
This disables the confirm dialog from popping up on page load after the user's session has reached 45 minutes. `window.open` can only be used after some sort of user interaction through an event listener, otherwise the browser's popup blocker will block it from opening. So, in most cases this was probably just a random dialog that popped up and didn't do anything.

Soon, we should definitely add an actual modal that refreshes the session to alleviate this issues.

### DIFF
--- a/src/js/login/containers/Main.jsx
+++ b/src/js/login/containers/Main.jsx
@@ -145,12 +145,14 @@ class Main extends React.Component {
   checkTokenStatus() {
     if (sessionStorage.userToken) {
       if (moment() > moment(sessionStorage.entryTime).add(SESSION_REFRESH_INTERVAL_MINUTES, 'm')) {
-        // TODO(crew): make more customized prompt.
-        if (confirm('For security, you’ll be automatically signed out in 2 minutes. To stay signed in, click OK.')) {
-          this.handleLogin();
-        } else {
-          this.handleLogout();
-        }
+
+        // @todo once we have time to replace the confirm dialog with an actual modal we should uncomment this code.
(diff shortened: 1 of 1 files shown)


src/js/login/containers/Main.jsx
```diff
@@ -145,12 +145,14 @@ class Main extends React.Component {
   checkTokenStatus() {
     if (sessionStorage.userToken) {
       if (moment() > moment(sessionStorage.entryTime).add(SESSION_REFRESH_INTERVAL_MINUTES, 'm')) {
-        // TODO(crew): make more customized prompt.
-        if (confirm('For security, you’ll be automatically signed out in 2 minutes. To stay signed in, click OK.')) {
-          this.handleLogin();
-        } else {
-          this.handleLogout();
-        }
+
```
jbalboni: What happens if the entryTime is older than 45 minutes now? Do the login links show up?

ncksllvn:
@jbalboni Good call, had to comment out that whole block. Now if there's an access token it will always hit the API and do everything based on that response. So if it's older than 45 minutes, or 45 hours, it doesn't matter, because it's just going to send the login request anyway.

ncksllvn:
Also, you can test this locally by editing the `entryTime` property of your `sessionStorage`.

image

cvalarida (approved):
Do we have a ticket anywhere outlining the work that needs to be done on the modal? This looks good like a stopgap in the meantime.

jbalboni (approved):

ncksllvn:
@cvalarida added one here - department-of-veterans-affairs/vets.gov-team/issues/6582

PR #6884, opened 2017-12-01
Add a rate limit event

robertfairhead:
This should fire off the event whenever we hit the rate limited view.

### DIFF
--- a/src/js/veteran-id-card/components/RequiredVeteranView.jsx
+++ b/src/js/veteran-id-card/components/RequiredVeteranView.jsx
@@ -12,6 +12,12 @@ class RequiredVeteranView extends React.Component {
     };
   }
 
+  componentDidMount() {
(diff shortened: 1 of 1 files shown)


robertfairhead:
I added the GTM config to report this out once this hits prod. I'll keep an eye on Jenkins to make sure it builds clean.

jkassemi:
@bshyong helped guide me through potential issues with a double call to `render` earlier. I believe this should go into `componentDidMount` to fully avoid the potential for a rerender on state update: 

```javascript
   componentDidMount() {
     if (this.state.serviceRateLimited) {
       window.dataLayer.push({ event: 'vic-ratelimited' });
     };
   }
```

bshyong (commented):
^ What James said. This needs to go in the mount method so that it only fires one time

jkassemi (approved):

PR #6886, opened 2017-12-01
Update copy for VIC rate limit message

jkassemi:
> The Veteran Identification Card application is getting a lot of traffic right now. If your application isn't going through, please try again in 24 hours. We apologize, and we want to make sure you know we're working to fix the problem.

From https://github.com/department-of-veterans-affairs/vets.gov-team/issues/6390#issuecomment-348609239


jkassemi:
cc @wyattwalter and @robertfairhead for review so we can potentially get this out along with your changes this evening

raquelromano-gov (approved):

PR #6889, opened 2017-12-02
Fix config for feature flag

jkassemi:
JSON.stringify is not necessary and causes problems when setting boolean values.

I scanned the history for examples before building the initial implementation. Caught 8650328b86398deac8817383081be47985a3001a, which was addressed in 113170e0bfcc5e7a9f52009f4086391689d300da.


robertfairhead (approved):
:shipit:

PR #6902, opened 2017-12-04
Production Release 20171204pre1

jkassemi:


ncksllvn (approved):
LGTM. This touches a lot of files but they're all very minor. 99% is just adding a trailing slash to links or changing http to https.

PR #6903, opened 2017-12-04
VIC announcement subscription copy updates

jkassemi:


jkassemi:
screen shot 2017-12-04 at 12 18 51 pm

screen shot 2017-12-04 at 12 19 01 pm

bethpotts:
@jkassemi Two more tiny tweaks, then we should be good to go. It helps so much to see the text in the screenshots, as it will appear on the site. Thanks! So could you edit two sentences?:
1) We're sorry, but your application didn't go through. 
2) ...please enter your email address below.

bethpotts:
After convo with James, here's where we landed:

We’re sorry. Our system is having trouble handling the many requests for Veteran ID cards, and can’t accept your application right now. We’re working to fix the problem as fast as we can.

bethpotts:
@jkassemi I just noticed that the link to our privacy policy is embedded. Usually we link the whole line of text. If it's not too late, could you link the whole shebang?: [See our privacy policy.]

jkassemi:
@bethpotts thanks! Just made the update. We're waiting on an update to how we pull the rate limit configuration to get this out.

bethpotts (approved):
Looks great!

PR #6904, opened 2017-12-04
Block SM if there is an error related to account creation.

U-DON:
Rx and BB were blocking access when getting the error `MHVAC1` and showing the ErrorView, but SM was not. Just fixing that inconsistency.


bshyong (approved):
LGTM

PR #6908, opened 2017-12-05
Gibs action tests mg

markgreenburg:
- Adds unit tests for the `getEnrollmentData()` async action creator in the `post-911-gib-status` app.
- Removes old logic that shouldn't be required any longer. Checked with @kreek that `meta` object will always return `OK`, so while it's still returned in the response, we shouldn't need to check it for errors any longer.

Adding Raquel & Alastair to reviewers list as FYI about deleting the check against the `meta` object in the response.


test/post-911-gib-status/actions/post-911-gib-status.unit.spec.js
```diff
@@ -0,0 +1,171 @@
...
+  global.fetch.returns(Promise.resolve({
+    headers: { get: () => 'application/json' },
+    ok: true,
+    json: () => Promise.resolve({})
+  }));
+  global.window.dataLayer = [];
+  global.window.URL = {
+    createObjectURL: () => { },
+    revokeObjectURL: () => { }
+  };
```
cvalarida: I'm really lamenting not being able to do this globally, but I'm not going down _that_ rabbit hole again so soon!

test/post-911-gib-status/actions/post-911-gib-status.unit.spec.js
```diff
@@ -0,0 +1,171 @@
...
+  global.fetch = oldFetch;
+  global.window = oldWindow;
+};
+
+describe('getEnrollmentData', () => {
+  const successResponse = {
+    meta: {
+      status: 'OK'
+    },
+    data: {
```
cvalarida: Where did you get this data (structure) from?

test/post-911-gib-status/actions/post-911-gib-status.unit.spec.js
```diff
@@ -0,0 +1,171 @@
...
+            vacationDayCount: 0,
+            onCampusHours: 12,
+            onlineHours: 0,
+            yellowRibbonAmount: 0,
+            status: 'Approved',
+            amendments: [
+              {
+                onCampusHours: 'Unknown Type: number,null',
+                onlineHours: 'Unknown Type: number,null',
+                yellowRibbonAmount: 'Unknown Type: number,null',
```
cvalarida: These seem...odd...

test/post-911-gib-status/actions/post-911-gib-status.unit.spec.js
```diff
@@ -0,0 +1,171 @@
...
+      .then(() => {
+        const action = dispatch.firstCall.args[0];
+        expect(action.type).to.equal(GET_ENROLLMENT_DATA_FAILURE);
+      }).then(done, done);
+  });
+
+  it('dispatches matching error action on known error code', (done) => {
+    global.fetch.returns(Promise.reject({
+      errors: [
+        { status: '503' }
```
cvalarida: I think we could pretty easily map this so we test all our error codes return what we expect...but I guess I don't know what value that would provide. I mean, it'd ensure they don't get messed up or deleted accidentally, but...

:man_shrugging: Not necessary, just a thought.

cvalarida (approved):
I assume that mocked `successResponse` reflects the same structure that we'd get from the api, but figured I'd still verify that it's golden.
img

test/post-911-gib-status/actions/post-911-gib-status.unit.spec.js
markgreenburg: Good question - copy / pasted from the Swagger for this endpoint

test/post-911-gib-status/actions/post-911-gib-status.unit.spec.js
markgreenburg: Ahh, you're right. I changed this to just take random numbers instead of strings, since that's what the type hints suggest.

test/post-911-gib-status/actions/post-911-gib-status.unit.spec.js
markgreenburg: Yeah, I deliberated over this for a while. My thought here was that I didn't want to duplicate the exact application logic, just test that the condition tree works as expected. 🤷‍♂️

PR #6909, opened 2017-12-05
Hca insurance expands ceh

cehsu:
This creates the ACA tooltip, as well as updates the `AdditionalInfo` component so that fields are not selected when it is closed (as was happening in this case). In particular updating the design as per https://github.com/department-of-veterans-affairs/vets.gov-team/issues/6045 has been combined with this work, as it will automatically resolve this issue.

screen shot 2017-12-04 at 10 08 19 pm
screen shot 2017-12-04 at 10 08 06 pm


peggygannon (approved):
@cehsu LGTM! thanks...

src/js/common/components/AdditionalInfo.jsx
```diff
@@ -10,7 +10,8 @@ export default class AdditionalInfo extends React.Component {
     this.state = { open: false };
   }
 
-  toggle = () => {
+  toggle = (e) => {
+    if (e) e.preventDefault();
```
jbalboni: Instead of doing this, we could also add `type="button"` to the button element. It's treated as submit button without that attribute.

jbalboni (approved):
Minor comment, but LGTM.

src/js/common/components/AdditionalInfo.jsx
cehsu: Thanks @jbalboni, it looks like the button element already has type="button".

jbalboni: @cehsu The close button doesn't.

cehsu: Thanks again for clarifying @jbalboni, that makes sense.

cehsu: Thanks for your recommendation @jbalboni, the proposed change does not prevent the automatic/unintentional selection.
screen shot 2017-12-07 at 12 02 52 pm
screen shot 2017-12-07 at 12 03 12 pm

jbalboni: Ah, I see. In that case, the way it was is fine, assuming it's focusing on the link again when you close it.

cehsu: It's not focusing on the link again, now that you mention it, nor does it with the button role assigned. How about I pull https://github.com/department-of-veterans-affairs/vets.gov-team/issues/6045 into this work so that there's no more need for the close button, rather than writing logic to set the focus?

jbalboni: Sounds good to me

PR #6911, opened 2017-12-05
Server error copy change for VIC

jkassemi:
When a server error prevents determining Veteran status, use an error
message that points less to a records problem and more to a temporary
systems problem.


bethpotts:
I'd recommend this PL version:
We're sorry. We can't process your request for a Veteran ID card right now because of a temporary system error.

jkassemi:
Thanks @bethpotts - so you think we shouldn't suggest retrying after a few minutes?

bethpotts:
@jkassemi Sorry about that! I DO think you should leave that in. I just had no changes to that line of text. So, yes, please keep: Please try again in a few minutes.

jkassemi:
Closing, further updates can be made in a new PR.

PR #6913, opened 2017-12-05
Add more save in progress e2e tests

jbalboni:
This adds some more e2e tests for save in progress and moves them to their own folder. They're using HCA, which seems ok, since the code is the same across forms.

A couple other changes:
1. I updated our mock api to support mocking up non-200 status codes
2. I created a `mockData` command for Nightwatch, so that we can change the mocked data at different points when running tests. I think this is clearer and it also avoids some of the execution order issues with trying to call it directly in tests.

Also, I couldn't find a good way to test connection failures, so that scenario isn't tested (but the only difference there is the message).


cvalarida (approved):
Looks great!

test/save-in-progress/01-sip-auto.e2e.spec.js
```diff
@@ -0,0 +1,102 @@
...
+      .click('.usa-button-primary')
+      .waitForElementVisible('.schemaform-sip-save-link', Timeouts.normal)
+      .mockData({
+        path: '/v0/in_progress_forms/1010ez',
+        verb: 'put',
+        value: {},
+        status: 500
+      }, token)
+      .fill('input[name="root_view:placeOfBirth_cityOfBirth"]', 'Amherst, MA')
+      .pause(1500);
```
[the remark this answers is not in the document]
jbalboni: I figured it need to be slightly higher than 1000, because that's the delay we have for auto-saving. Could probably lower it a little, though.

test/save-in-progress/01-sip-auto.e2e.spec.js
```diff
@@ -0,0 +1,102 @@
...
+      .mockData({
+        path: '/v0/in_progress_forms/1010ez',
+        verb: 'put',
+        value: {
+          data: {
+            attributes: {
+              metadata: {
+                version: 0,
+                returnUrl: '/veteran-information/birth-information',
+                savedAt: 1498588443698,
```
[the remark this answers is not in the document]
jbalboni: Yeah, I believe expires_at is all that's checked.

test/save-in-progress/01-sip-review.e2e.spec.js
```diff
@@ -0,0 +1,53 @@
...
+    E2eHelpers.overrideSmoothScrolling(client);
+
+    client
+      .click('.schemaform-chapter-accordion-header:first-child > .usa-button-unstyled')
+      .click('.edit-btn')
+      .fill('input[name="root_veteranFullName_first"]', 'Jane')
+      .pause(1200);
+
+    // client
+    //   .expect.element('.saved-success-container').to.be.visible;
```
annekainicUSDS: Did you mean to leave this commented out?

test/save-in-progress/01-sip-review.e2e.spec.js
```diff
@@ -0,0 +1,53 @@
...
+
+    // save and finish a form later
+    client
+      .click('.schemaform-sip-save-link');
+
+    E2eHelpers.expectNavigateAwayFrom(client, '/review-and-submit');
+    client.assert.urlContains('form-saved');
+
+    client
```
annekainicUSDS: Minor note but maybe a comment above here like you do elsewhere explaining this is testing for a 500 error on the review and submit page?

test/save-in-progress/01-sip.e2e.spec.js
```diff
@@ -1,11 +1,12 @@
+const moment = require('moment');
```
annekainicUSDS: What's the difference between this test file and `01-sip-auto`?

annekainicUSDS (approved):
LGTM! Just a few minor comments

test/save-in-progress/01-sip.e2e.spec.js
jbalboni: This is for save and finish later, the other is for auto saving. I'll change the file names to make that clearer.

test/save-in-progress/01-sip-auto.e2e.spec.js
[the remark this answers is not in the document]
cvalarida: Heh, if not, we'll find out in a month or three, eh? 😜

test/e2e/sitemap-helpers.js
```diff
@@ -27,7 +27,9 @@ function sitemapURLs(callback) {
         // which violates WCAG 2.0 standards. This element id is referenced by
         // https://search.usa.gov/assets/sayt_loader_libs.js, so if we change the ID
         // of one of the elements, search won't work.
-        '/404.html'
+        '/404.html',
+        // This is here because an aXe bug flags the autosuggest component on this page
+        '/gi-bill-comparison-tool/'
```
jbalboni: FYI @raquelromano-gov. I'm not sure why this page is fully rendering in the sitemap tests now, but I believe this is how the bug was handled in the gibct e2e test.

PR #6906, opened 2017-12-04
Add a settings file for quickly changing global variables

ncksllvn:
Resolves department-of-veterans-affairs/vets.gov-team/issues/6521

PR #6916, opened 2017-12-06
WIP: Rate limit both authenticated and unauthenticated users

raquelromano-gov:
This moves all rate-limiting logic and analytics events into VeteranIDCard.jsx and out of RequiredVeteranView.jsx, and adds a second variable so we can independently adjust the rate limit for authenticated and unauthenticated users.

This is WIP, so have yet to get a demo up to test behavior, but in the meantime @bshyong @ncksllvn  can you review? We'll probably want to deploy this manually tomorrow morning so we can begin sending a fraction of unauthenticated users to the VIC page

@robertfairhead can you review the changes analytics events? This renames events to more clearly reflect the conditions, but am also open to preserving the existing naming conventions instead, and settling for event names that are a bit harder to interpret

### DIFF
--- a/config/create-settings.js
+++ b/config/create-settings.js
@@ -12,7 +12,7 @@ function createSettings(options) {
     },
     vic: {
       rateLimitAuthed: 1,
-      rateLimitUnauthed: 1
+      rateLimitUnauthed: 0.05
     }
   };
 }
--- a/config/create-settings.js
+++ b/config/create-settings.js
@@ -11,7 +11,8 @@ function createSettings(options) {
       BASE_URL: process.env.BASE_URL
     },
     vic: {
-      rateLimit: options.buildtype === 'development' ? 1 : 0.1
+      rateLimitAuthed: 1,
+      rateLimitUnauthed: 1
     }
   };
 }
--- a/src/js/veteran-id-card/containers/VeteranIDCard.jsx
+++ b/src/js/veteran-id-card/containers/VeteranIDCard.jsx
@@ -27,6 +27,15 @@ class VeteranIDCard extends React.Component {
         window.dataLayer.push({ event: 'vic-unauthenticated' });
       }
     }
+
+    if (this.renderEmailCapture === true ){
+      // Report if they will see an error message around eMIS status
+      if (this.props.profile.veteranStatus === 'NOT_FOUND') {
+        window.dataLayer.push({ events: 'vic-emis-lookup-failed' });
+      } else if (this.props.profile.veteranStatus === 'SERVER_ERROR' ) {
+        window.dataLayer.push({ events: 'vic-emis-error' });
+      }
+    }
   }
 
   render() {
--- a/src/js/veteran-id-card/containers/VeteranIDCard.jsx
+++ b/src/js/veteran-id-card/containers/VeteranIDCard.jsx
@@ -28,11 +28,11 @@ class VeteranIDCard extends React.Component {
       }
     }
 
-    if (this.renderEmailCapture === true ){
+    if (this.renderEmailCapture === true) {
       // Report if they will see an error message around eMIS status
       if (this.props.profile.veteranStatus === 'NOT_FOUND') {
         window.dataLayer.push({ events: 'vic-emis-lookup-failed' });
-      } else if (this.props.profile.veteranStatus === 'SERVER_ERROR' ) {
+      } else if (this.props.profile.veteranStatus === 'SERVER_ERROR') {
         window.dataLayer.push({ events: 'vic-emis-error' });
       }
     }
--- a/src/js/veteran-id-card/containers/VeteranIDCard.jsx
+++ b/src/js/veteran-id-card/containers/VeteranIDCard.jsx
@@ -5,17 +5,32 @@ import RequiredLoginView from '../../common/components/RequiredLoginView';
 import RequiredVeteranView from '../components/RequiredVeteranView';
 import EmailCapture from './EmailCapture';
 
+const rateLimitAuthed = window.settings.vic.rateLimitAuthed;
+const rateLimitUnauthed = window.settings.vic.rateLimitUnauthed;
+
 class VeteranIDCard extends React.Component {
 
   componentDidMount() {
-    if (!this.props.profile.accountType) {
-      window.dataLayer.push({ event: 'vic-unauthenticated' });
+    // Redirect users to email form based on rate limits for unauthed or authed users
+    if (this.props.profile.accountType) {
+      if (this.props.serviceRateLimitedAuthed) {
+        window.dataLayer.push({ event: 'vic-authenticated-ratelimited' });
+        this.renderEmailCapture = true;
+      } else {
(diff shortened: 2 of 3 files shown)


src/js/veteran-id-card/containers/VeteranIDCard.jsx
```diff
@@ -42,6 +57,8 @@ const mapStateToProps = (state) => {
     profile: userState.profile,
     loginUrl: userState.login.loginUrl,
     verifyUrl: userState.login.verifyUrl,
+    serviceRateLimitedAuthed: Math.random() > rateLimitAuthed,
```
raquelromano-gov: @ncksllvn curious why in https://github.com/department-of-veterans-affairs/vets-website/pull/6906/files for RequiredLoginView.jsx you moved this out of the constructor and into defaultProps. Thinking this could go in `componentDidMount()` instead, right?

src/js/veteran-id-card/containers/VeteranIDCard.jsx
```diff
@@ -5,17 +5,32 @@ import RequiredLoginView from '../../common/components/RequiredLoginView';
...
 class VeteranIDCard extends React.Component {
 
   componentDidMount() {
-    if (!this.props.profile.accountType) {
-      window.dataLayer.push({ event: 'vic-unauthenticated' });
+    // Redirect users to email form based on rate limits for unauthed or authed users
+    if (this.props.profile.accountType) {
+      if (this.props.serviceRateLimitedAuthed) {
+        window.dataLayer.push({ event: 'vic-authenticated-ratelimited' });
+        this.renderEmailCapture = true;
```
raquelromano-gov: Storing this as a plain old class member instead of state variable because it's not something that changes, but open to arguments for why it should be part of the component state.

robertfairhead:
The changes to the analytics events are fine. Changing the meaning of the `vic-unauthenticated` event means todays data will be a little wonky--part of the day it'll mean they got blocked; the rest they got through. But, if we're gonna make that change we may as well do it now.

I opened a short PR https://github.com/department-of-veterans-affairs/vets-website/pull/6917 against this branch to add some events to capture when they see the error messages around eMIS lookup from `RequiredVeteranView.jsx`. Figured those would be nice to have explicitly so we can differentiate the cases, but it's take it or leave it.

I have the updated GTM config queued up already. So, whenever this deploys the data will be picked up.

raquelromano-gov:
@robertfairhead thanks for those additional events in https://github.com/department-of-veterans-affairs/vets-website/pull/6917 
I'll probably just add those to this PR

src/js/veteran-id-card/containers/VeteranIDCard.jsx
ncksllvn: Yeah it may be better have it determined in `componentDidMount()` since that's where the rest of the logic is, or maybe into `mapStateToProps`. At the time I didn't really see a reason to keep it in state, because that value is only calculated once (since after page load that component will only be rendered at most one time) and then should remain static, so defaultProps gives us that while also exposing that value as something we could add unit tests for or manipulate from a parent component if we wanted to.

ncksllvn (approved):
LGTM

config/create-settings.js
```diff
@@ -11,7 +11,8 @@ function createSettings(options) {
       BASE_URL: process.env.BASE_URL
     },
     vic: {
-      rateLimit: options.buildtype === 'development' ? 1 : 0.1
+      rateLimitAuthed: 1,
```
raquelromano-gov: @jkassemi Just realized: if we default this to 1, will we have to tweak it on S3 after every deploy? Until we get to the point where we're not rate-limiting, that is

wyattwalter: Yeah, with this implementation that is correct.

raquelromano-gov: OK, then we should probably change this default to 0.05 for now. We'll still have to tweak on deploys since it's looking like we'll want to ramp up and down throughout the week.

robbiethegeek-usds:
Any idea what the throttling limit is from vets.gov to VIC? Getting asked in a meeting.

ncksllvn:
@robbiethegeek do you mean what is it currently? It's currently (11:30 AM EST) set to allow 100% of users but it can be adjusted to anywhere between 0 and 100%. After this PR, that number will be configurable for both authenticated and unauthenticated users.

src/js/veteran-id-card/containers/VeteranIDCard.jsx
```diff
@@ -7,33 +7,39 @@ import EmailCapture from './EmailCapture';
...
-      }
-    }
+  componentWillReceiveProps(nextProps) {
+
+    // Once the login logic is all done...
+    // This will occur even for unauthenticated users and should only occur once.
+    if (this.props.profile.loading && !nextProps.profile.loading) {
+      const userProfile = nextProps.profile;
+      const serviceRateLimitedAuthed = this.props.serviceRateLimitedRandomizer > this.props.rateLimitAuthed;
+      const serviceRateLimitedUnauthed = this.props.serviceRateLimitedRandomizer > this.props.rateLimitUnauthed;
```
bshyong: Why is the randomizer mapped to props instead of accessing it from the constant defined above?

ncksllvn: That prop is still mapped to that constant (see `mapStateToProps`). I moved it so that the value is exposed as a prop and so I can see it in the React dev tools. Also it sort of sets it up to move that value into the Redux layer. Our next step is to not apply the `rateLimitAuthed` rate limit to users who logged in after being allowed through by the `rateLimitUnauthed`. I think that logic would probably be best in Redux logic, right?

raquelromano-gov: @ncksllvn was wondering the same--is there a reason you moved it into props?

ncksllvn: Yeah I answered that a moment ago but it may not have come through on your comments yet

> That prop is still mapped to that constant (see mapStateToProps). I moved it so that the value is exposed as a prop and so I can see it in the React dev tools. Also it sort of sets it up to move that value into the Redux layer. Our next step is to not apply the rateLimitAuthed rate limit to users who logged in after being allowed through by the rateLimitUnauthed. I think that logic would probably be best in Redux logic, right?

ncksllvn: If anyone disagrees, we can pull it out of props. I just don't like having values that we can't see or change from unit testing or parent components.

bshyong: I don't have a strong opinion on props or not; I have thoughts on moving this into Redux (e.g. reducers need to be pure) but we can revisit that later.

ncksllvn: Yeah I was thinking using these constants as `initialState` and then just never overriding them unless we find ourselves under special circumstances, not sure what those would be right now though

PR #6919, opened 2017-12-06
Turn on maintenance message for HCA

jbalboni:
screen shot 2017-12-06 at 2 47 33 pm


rtluu:
Looks good to me!

rtluu (approved):
Great that we have a message about the saved forms

kerryilenahan:
@jbalboni - This message is for when the ES is unavailable, right?

If so, was thinking our messaging for these scenarios would be the same across the site.  So taking what we are using for VIC.  I would think that HCA would display something like....

We’re sorry. The Enrollment System is current unavailable and can’t accept your healthcare application right now. We’re working to fix a few things as fast as we can.  

If you have a saved form, you will be able to access it as soon as the Enrollment System is working again.

jbalboni:
@kerryilenahan That looks good to me. I think I'm going to close this PR for now, though, since ES came back up and we no longer need to do anything here. I think the platform team has some plans on making putting up these messages easier and more consistent.

PR #6921, opened 2017-12-06
VIC - Only use one method to rate limit users

ncksllvn:
This adds a value to `sessionStorage` called `vicInitialAuthStatus` that is used to keep the user from being allowed through as unauthorized user, but then rate limited again as an authorized user after logging in.

Resolves department-of-veterans-affairs/vets.gov-team/issues/6569

### DIFF
--- a/src/js/veteran-id-card/containers/VeteranIDCard.jsx
+++ b/src/js/veteran-id-card/containers/VeteranIDCard.jsx
@@ -27,6 +27,15 @@ class VeteranIDCard extends React.Component {
         window.dataLayer.push({ event: 'vic-unauthenticated' });
       }
     }
+
+    if (this.renderEmailCapture === true ){
+      // Report if they will see an error message around eMIS status
+      if (this.props.profile.veteranStatus === 'NOT_FOUND') {
+        window.dataLayer.push({ events: 'vic-emis-lookup-failed' });
+      } else if (this.props.profile.veteranStatus === 'SERVER_ERROR' ) {
+        window.dataLayer.push({ events: 'vic-emis-error' });
+      }
+    }
   }
 
   render() {
--- a/src/js/veteran-id-card/containers/VeteranIDCard.jsx
+++ b/src/js/veteran-id-card/containers/VeteranIDCard.jsx
@@ -34,11 +34,11 @@ class VeteranIDCard extends React.Component {
           window.dataLayer.push({ event: 'vic-authenticated' });
         }
       } else {
-        window.sessionStorage.setItem('vicInitialAuthStatus', 'unauthenticated');
         if (serviceRateLimitedUnauthed) {
           window.dataLayer.push({ event: 'vic-unauthenticated-ratelimited' });
           this.renderEmailCapture = true;
         } else {
+          window.sessionStorage.setItem('vicInitialAuthStatus', 'unauthenticated');
           window.dataLayer.push({ event: 'vic-unauthenticated' });
         }
       }
--- a/src/js/veteran-id-card/containers/VeteranIDCard.jsx
+++ b/src/js/veteran-id-card/containers/VeteranIDCard.jsx
@@ -5,10 +5,10 @@ import RequiredLoginView from '../../common/components/RequiredLoginView';
 import RequiredVeteranView from '../components/RequiredVeteranView';
 import EmailCapture from './EmailCapture';
 
-const rateLimiteMethod = window.sessionStorage.getItem('vicRateLimitMethod');
-const rateLimitAuthed = rateLimiteMethod === 'unauthenticated' ? 1 : window.settings.vic.rateLimitAuthed;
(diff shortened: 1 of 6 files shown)


src/js/veteran-id-card/containers/VeteranIDCard.jsx
```diff
@@ -5,9 +5,10 @@ import RequiredLoginView from '../../common/components/RequiredLoginView';
 import RequiredVeteranView from '../components/RequiredVeteranView';
 import EmailCapture from './EmailCapture';
 
-const rateLimitAuthed = window.settings.vic.rateLimitAuthed;
-const rateLimitUnauthed = window.settings.vic.rateLimitUnauthed;
-const serviceRateLimitedRandomizer = Math.random();
+const vicSettings = {
+  serviceRateLimitedUnauthed: Math.random() > window.settings.vic.rateLimitUnauthed,
+  serviceRateLimitedAuthed: window.sessionStorage.getItem('vicInitialAuthStatus') !== 'unauthenticated' && Math.random() > window.settings.vic.rateLimitAuthed
+};
```
ncksllvn: This is not pretty but I felt like it was getting too cluttered and therefore counterproductive to define a bunch of constants that will ultimately only be used for determining the `serviceRateLimitedUnauthed` and the `serviceRateLimitedAuthed` booleans.

src/js/veteran-id-card/containers/VeteranIDCard.jsx
```diff
@@ -72,9 +73,7 @@ const mapStateToProps = (state) => {
     profile: userState.profile,
     loginUrl: userState.login.loginUrl,
     verifyUrl: userState.login.verifyUrl,
-    rateLimitAuthed,
-    rateLimitUnauthed,
-    serviceRateLimitedRandomizer
+    vicSettings
```
ncksllvn: I pass these to props only to expose them to React dev tools since we aren't planning on moving these into Redux.  If we wanted to bind these to the class instead that would make sense to me too.

raquelromano-gov (approved):

src/js/veteran-id-card/containers/VeteranIDCard.jsx
```diff
@@ -34,11 +34,11 @@ class VeteranIDCard extends React.Component {
           window.dataLayer.push({ event: 'vic-authenticated' });
         }
       } else {
-        window.sessionStorage.setItem('vicInitialAuthStatus', 'unauthenticated');
         if (serviceRateLimitedUnauthed) {
           window.dataLayer.push({ event: 'vic-unauthenticated-ratelimited' });
           this.renderEmailCapture = true;
         } else {
+          window.sessionStorage.setItem('vicInitialAuthStatus', 'unauthenticated');
```
raquelromano-gov: Did not catch this, but thanks. Should we add a comment above that says something like, set this variable only for users who have made it past the unauthenticated rate-limiting gate and to the sign-in/log-in page, not for users who are sent to the email capture form

src/js/veteran-id-card/containers/VeteranIDCard.jsx
```diff
@@ -5,9 +5,10 @@ import RequiredLoginView from '../../common/components/RequiredLoginView';
 import RequiredVeteranView from '../components/RequiredVeteranView';
 import EmailCapture from './EmailCapture';
 
-const rateLimitAuthed = window.settings.vic.rateLimitAuthed;
-const rateLimitUnauthed = window.settings.vic.rateLimitUnauthed;
-const serviceRateLimitedRandomizer = Math.random();
+const vicSettings = {
+  serviceRateLimitedUnauthed: Math.random() > window.settings.vic.rateLimitUnauthed,
+  serviceRateLimitedAuthed: window.sessionStorage.getItem('vicInitialAuthStatus') !== 'unauthenticated' && Math.random() > window.settings.vic.rateLimitAuthed
```
raquelromano-gov: I'd add a comment here that says, users who have made it past the unauthenticated rate-limiting gate, and are now authenticated and accessing VIC should not be subject to the authenticated rate-limiting gate

Or something much more succinct! Just think the double negative above, `!== 'unauthenticated'` makes this hard to grok and I want to make sure the intent is clear.

src/js/veteran-id-card/containers/VeteranIDCard.jsx
bshyong: +1 to adding explanatory comment

ncksllvn: Awesome, will do! I may change the name of that key in sessionStorage to something more clear, not sure at the moment what that would be though.

src/js/veteran-id-card/containers/VeteranIDCard.jsx
ncksllvn: Yeah good idea, I'll do that and wrap that logic in a function so I can list where the variables come from and add some meaning to them.

PR #6925, opened 2017-12-07
Set up pre-need e2e tests

cehsu:
These changes set up the e2e tests for preneed. There are two tests currently disabled, file upload and confirmation page.
Also, is name not required for service name?
screen shot 2017-12-06 at 5 28 51 pm
This seems different from how we handle other forms:
screen shot 2017-12-06 at 7 52 42 pm

And how is the document type information being captured/used?
screen shot 2017-12-06 at 6 28 23 pm


jbalboni:
@cehsu Yeah, it looks like the sponsor other name fields should be required when you choose Yes there.

The document type is `attachmentId` in that object in your screenshot.

test/e2e/preneed-helpers.js
```diff
@@ -0,0 +1,137 @@
...
+  completeVeteranInformation,
+  completeServicePeriods,
+  completeServiceName,
+  completeBenefitSelection,
+  completeClaimantContactInformation,
+  completeVeteranContactInformation,
+  completeApplicantContactInformation,
+
+  initApplicationSubmitMock,
+  initDocumentUploadMock
```
jbalboni: You'll probably want another one of these for the cemetery api call.

jbalboni:
@cehsu Looks like there are some aXe failures for the cemetery widget. If you want to tackle those, go for it, otherwise I can fix them in this PR.

jbalboni:
Actually, I think the axe failures are due a bug, which we worked around here: https://github.com/department-of-veterans-affairs/vets-website/pull/5663/files

cehsu:
Thanks for pointing out the axe failures, and for the recommendation regarding the cemetery api call. There is now appears to be a bug in the preneed application that is blocking this, I will investigate it to see whether an issue needs to be created.
After filling out the name field, the information is erased onBlur:
screen shot 2017-12-07 at 4 26 24 pm
screen shot 2017-12-07 at 4 26 36 pm

src/js/common/schemaform/FormPage.jsx
```diff
@@ -7,7 +7,7 @@ import _ from 'lodash/fp';
 
 import SchemaForm from './SchemaForm';
 import ProgressButton from '../components/form-elements/ProgressButton';
-import { setData, uploadFile } from '../actions';
+import { setData, uploadFile } from './actions';
```
cehsu: This was causing the bug.

jbalboni: Ah, good catch

test/e2e/preneed-helpers.js
```diff
@@ -0,0 +1,3004 @@
...
+  });
+}
+
+function completeServiceName(client, data) {
+  client.selectRadio('root_application_veteran_view:hasServiceName', 'Y')
+    .fillName('root_application_veteran_serviceName', data.serviceName);
+}
+
+function completeBenefitSelection(client, data) {
+  // TODO: Include autosuggest interaction
```
cehsu: Right now this test is not making use of the autosuggest functionality, it is simply filling the field. It seems like a more robust implementation that makes use of the getCemeteries mock request, and the autosuggest functionality that involves only partially filling the field and then selecting one of the suggested options should be built as well? Should a ticket be created for that or is that beyond the scope of the e2e tests?

test/pre-need/00-all-fields.e2e.spec.js
```diff
@@ -0,0 +1,121 @@
...
+const PageHelpers = require('../e2e/preneed-helpers');
+const testData = require('./schema/maximal-test.json');
+
+const runTest = E2eHelpers.createE2eTest(
+  (client) => {
+    PageHelpers.initApplicationSubmitMock();
+    PageHelpers.initDocumentUploadMock();
+    PageHelpers.initGetCemeteriesMock();
+
+    if (process.env.BUILDTYPE !== 'production') {
```
cehsu: Is this the appropriate flag to use? __BUILDTYPE__ is undefined here.

jbalboni: Yeah, as long as it's passing on Jenkins you should be good.

test/pre-need/00-all-fields.e2e.spec.js
```diff
@@ -0,0 +1,121 @@
...
+      // Review and Submit Page.
+      client
+        .waitForElementVisible('label[name="privacyAgreement-label"]', Timeouts.slow)
+        .pause(1000)
+        .click('input[type="checkbox"]')
+        .axeCheck('.main')
+        .click('.form-progress-buttons .usa-button-primary');
+      // E2eHelpers.expectNavigateAwayFrom(client, '/review-and-submit');
+      // client.expect.element('.js-test-location').attribute('data-location')
+      // .to.not.contain('/review-and-submit').before(Timeouts.slow);
```
jbalboni: We should be able to get this commented out code to work

jbalboni (approved):
LGTM, just need to get the submit check uncommented and working

PR #6927, opened 2017-12-07
Add analytics events for header navigation

robertfairhead:
Closes: department-of-veterans-affairs/vets.gov-team#1420


robertfairhead:
This mostly just adds `onClick` handlers to the various elements. I manually added them in.

Details on the reporting side are in the linked issue but that's mainly handled through GTM and not in our code.

I tried some more "clever" approaches but none of them really struck me as that durable in the face of changes. This way, while repetitious, at least seemed obvious and fairly simple to debug.

But, i'm open to thoughts if you see a better way to instrument these.

content/includes/veteran-navigation.html
```diff
@@ -227,22 +237,33 @@
         <li>
           <button
             class="vetnav-level1 vetnav-trigger"
-            aria-controls="vetnav-benefits"
-            aria-haspopup="true">Manage your health &amp; benefits</button>
+            role="button"
+            aria-controls="vetnav-benefits" aria-haspop="true"
```
phiden: should be `areia-haspopup`?

robertfairhead: yep. fixed it.

phiden (approved):

content/includes/veteran-navigation.html
```diff
@@ -227,22 +236,34 @@
...
   </div>
 </div>
+
+<script>
+function reportHeaderNav(navpath) {
+  window.dataLayer.push({
+    event: 'nav-header-link',
+    'nav-path': navpath
+  });
+}
```
ncksllvn: It may be worth considering at some point using JS to attach event listeners:

``` html
<li><a class="login-required" href="/health-care/prescriptions" data-navpath="Manage Health and Benefits">Refill prescriptions</a></li>
```

``` javascript
document.querySelectorAll('#vetnav li > a').forEach(function(element){
    element.addEventListener('click', function(e){
        const navpath = element.dataset.navpath;
        window.dataLayer.push({
         event: 'nav-header-link',
         'nav-path': navpath
       });
    });
});
```

ncksllvn (approved):
LGTM

content/includes/veteran-navigation.html
robertfairhead: yeah, i had that originally but the but i felt like the `onClick` versus the `data-navpath` didn't save much in the readability of the HTML and was a little more clear to show something happened on click right in the element. 

What i _really_ wanted to do was move all the nav elements into some YAML and then use liquid templates to make all the HTML so we could separate the data from the presentation more cleanly. But, it just sort of grew into a furball. But i think the header nav could probably use a good refactor. I saw this first PR as an incremental step towards that.

ncksllvn: Yeah I know you mean, I agree. This looks good the way it is. The navigation has gone through all sorts of changes lately with accessibility and viewports and it would be nice to see it all simplified at some point.

PR #6929, opened 2017-12-07
Fix dependent income on HCA

jbalboni:
Fixes the bad dependent display issue (and related issue moving to next page)

screen shot 2017-12-07 at 3 11 41 pm


cehsu (approved):
LGTM.

PR #6930, opened 2017-12-07
Fetch BSL options after letter list

cvalarida:
Calls the fetch to get the BSL options after the fetch to get the available letters list so the response from the letters list fetch can be cached for the BSL options request in an EVSS partner services...to my understanding.

### DIFF
--- a/src/js/letters/actions/letters.js
+++ b/src/js/letters/actions/letters.js
@@ -27,48 +27,71 @@ import {
   SAVE_ADDRESS_SUCCESS
 } from '../utils/constants';
 
-export function getLetterList() {
+export function getLetterList(dispatch) {
+  return apiRequest(
+    '/v0/letters',
+    null,
+    response => dispatch({
+      type: GET_LETTERS_SUCCESS,
+      data: response,
+    }),
+    (response) => {
+      window.dataLayer.push({ event: 'letter-list-failure' });
+      if (typeof response.errors === 'undefined' || response.errors.length === 0) {
+        return Promise.reject(new Error('vets_letters_error_server_get: undefined error'));
+      }
+      const error = response.errors[0];
+      switch (error.status) {
+        case '503': // Handled same as 504
+        case '504':
+          // Either EVSS or a partner service is down or EVSS times out
+          return dispatch({ type: BACKEND_SERVICE_ERROR });
+        case '403':
+          // Backend authentication problem
+          return dispatch({ type: BACKEND_AUTHENTICATION_ERROR });
+        case '502':
+          // Some of the partner services are down, so we cannot verify the
+          // eligibility of some letters
+          return dispatch({ type: LETTER_ELIGIBILITY_ERROR });
+        default:
+          return Promise.reject(
+            new Error(`vets_letters_error_server_get: ${error.status || 'unknown'}`)
+          );
+      }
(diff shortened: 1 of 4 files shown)


src/js/letters/actions/letters.js
```diff
@@ -27,49 +27,72 @@ import {
...
+}
+
+
+// Call getLetterList then getBenefitSummaryOptions
+export function getLetterListAndBSLOptions() {
+  return (dispatch) => {
+    return getLetterList(dispatch)
+      // Maybe shouldn't try to get BSO options if we get an error or we don't get a BSL...
+      .then(() => getBenefitSummaryOptions(dispatch))
+      .catch(rejectedPromise => rejectedPromise); // getLetterList should return a rejected promise if it fails
```
markgreenburg: I think getLetterList actually returns a `dispatch({ type: GET_LETTERS_FAILURE })` _or_ throws an error (the rejected promise should get caught by the `catch()` on L#62). It might make sense to move the `catch()` for the `getLetterList()`'s `apiRequest()`into `getLetterListAndBSLOptions()`, so that we can consolidate top-level error handling and logging to Sentry (which is missing from `getBenefitSummaryOptions()` entirely).

markgreenburg (commented):
LGTM! Would like to discuss error handling comment before approving.

src/js/letters/actions/letters.js
cvalarida: Oh blast. You're right. It _either_ returns the result of `dispatch()` _or_ throws the error. Missed that `return` before.

Yeah, I like the idea of moving that error handling logic up a level. Lemme see what I can come up with.

test/letters/actions/letters.unit.spec.js
```diff
@@ -117,6 +117,8 @@ describe('saveAddress', () => {
     const dispatch = sinon.spy();
     thunk(dispatch, getState)
       .then(() => {
+        throw new Error('Should not resolve the promise when GET fails');
```
markgreenburg: If I understand correctly, this is necessary to mimic the nested structure of the new consolidated action creator

src/js/letters/actions/letters.js
```diff
@@ -27,49 +27,77 @@ import {
...
+
+
+// Call getLetterList then getBenefitSummaryOptions
+export function getLetterListAndBSLOptions() {
+  return (dispatch) => {
+    return getLetterList(dispatch)
+      // Maybe shouldn't try to get BSO options if we get an error or we don't get a BSL...
+      .then(() => getBenefitSummaryOptions(dispatch))
+      .catch(error => {
+        Raven.captureException(error);
```
markgreenburg: Maybe a silly question, but is logging to Sentry all we want to do here? I guess we're dispatching failure actions for each fetch that fails so it's probably the case that we're already covered, but wanted to ask to double check.

cvalarida: Yeah, I think so. Like you said, we're dispatching failure actions as we go and I don't think there's anything else to do.

Actually, changing it like this will catch more errors, but I don't think that's a problem.

markgreenburg (commented):
LGTM 🎉

markgreenburg (approved):
Meant to approve last time around.

PR #6936, opened 2017-12-08
Fix application status spinner

jbalboni:
There are two issues here:

1. The apply now button flashes before showing the spinner, even when logged out. This seems to happen because the page renders with the button (because it's in the markup), then JS takes over and replaces the button with a spinner. Now the button is hidden by default, and JS has to remove a class to show anything.
2. When there's an error fetching the login urls (like a connection issue), the loading state for the profile is never changed and so we have the spinner show indefinitely. I'm not sure this is all of the problems from the call center, but it's probably some of them. There are regular client-side failures of the login api calls.

### DIFF
--- a/src/js/login/containers/Main.jsx
+++ b/src/js/login/containers/Main.jsx
@@ -40,9 +40,9 @@ class Main extends React.Component {
     });
     this.bindNavbarLinks();
 
-    // If there is a window.opener, then this window was spawned by another instance of the website as the auth app.
-    // In this case, we don't need to actually login, because that data will be passed to the parent window and done there instead.
-    if (!window.opener) {
+    // In some cases this component is mounted on a url that is part of the login process and doesn't need to make another 
+    // request, because that data will be passed to the parent window and done there instead.
+    if (!window.location.pathname.includes('auth/login/callback')) {
       window.onload = () => {
         this.loginUrlRequest.then(this.checkTokenStatus);
       };
--- a/src/js/login/containers/Main.jsx
+++ b/src/js/login/containers/Main.jsx
@@ -40,9 +40,9 @@ class Main extends React.Component {
     });
     this.bindNavbarLinks();
 
-    // In some cases this component is mounted on a url that is part of the login process and doesn't need to make another 
-    // request, because that data will be passed to the parent window and done there instead.
-    if (!window.location.pathname.includes('auth/login/callback')) {
+    // If there is a window.opener, then this window was spawned by another instance of the website as the auth app.
+    // In this case, we don't need to actually login, because that data will be passed to the parent window and done there instead.
+    if (!window.opener) {
       window.onload = () => {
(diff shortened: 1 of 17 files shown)


src/js/login/containers/Main.jsx
```diff
@@ -40,9 +40,9 @@ class Main extends React.Component {
     });
     this.bindNavbarLinks();
 
-    // If there is a window.opener, then this window was spawned by another instance of the website as the auth app.
-    // In this case, we don't need to actually login, because that data will be passed to the parent window and done there instead.
-    if (!window.opener) {
+    // In some cases this component is mounted on a url that is part of the login process and doesn't need to make another 
+    // request, because that data will be passed to the parent window and done there instead.
+    if (!window.location.pathname.includes('auth/login/callback')) {
```
raquelromano-gov: Hard to test this thoroughly, but it does make sense. Is the worst case that this is too restrictive, so we end up making a few unnecessary extra fetches for login URLs (which I believe we were doing before for a while), but no user-perceived effect?

jbalboni: I believe so. I tested the login flow with the check completely removed and I was still able to login fine.

saneshark:
@U-DON @bshyong could you guys look this over as well?

@jbalboni I thought we addressed #2 in your description some time ago with respect to the sign-in modal / spinner. If the backend raises an exception because (lets say redis is down for example), my understanding was that the spinner would go away. Only network latency, vets-api not returning anything at all would result in spinner being around indefinitely.

Also, can we make sure that SAML popup flow never does a redirect showing the modal. If an error is shown there, the sign-in modal should not be shown. Instead, the error message from the saml response should be shown. I think I ran into this issue on staging the other day, but it's difficult to reproduce. In my particular case, I believe that DS Logon / ID.me interaction was causing an error which resulted in an error to vets-api about saml response being invalid. Previously this would be a generic 
"Oops something went wrong." I did see this error but only for brief second before it was overlayed with the sign-in modal. I don't believe that sign-in modal should ever be shown in the sign-in flow popup.

saneshark:
Additionally, and perhaps out of scope for this PR. We make requests for identity_proof url for nearly every single request and I still do not understand why the destroy action needs to be invoked immediately after sign-in. I would think it could be fetched when user clicks on logout, we would prompt them with a "are you sure" during which time the url for actually destroying should have returned, and then we fire off that url if they say "yes." I know these are unrelated, but on my wish list of improvements. :)

jbalboni:
@saneshark I see 2 mostly (maybe solely) in the context of client side failures, where there's no response from the backend. This is also for the sign in/up links and spinners that are waiting on the profile on different pages, not necessarily in the login flow itself, which I think handles errors alright.

cehsu (approved):
LGTM.

jbalboni:
@saneshark The popup flow as far as I can tell uses the Auth app and whenever there's an error it just displays an error, no modal or anything.

PR #6938, opened 2017-12-08
Fix infinite spinner on new tab issue

jbalboni:
When vets.gov is opened in a new tab, window.opener is set and login requests are not made. This results in infinite spinners in many places. I've changed the logic to look for the auth callback page instead, because I believe that's the only case where we wouldn't want to make the login request.

### DIFF
--- a/src/js/login/containers/Main.jsx
+++ b/src/js/login/containers/Main.jsx
@@ -40,9 +40,9 @@ class Main extends React.Component {
     });
     this.bindNavbarLinks();
 
-    // If there is a window.opener, then this window was spawned by another instance of the website as the auth app.
-    // In this case, we don't need to actually login, because that data will be passed to the parent window and done there instead.
-    if (!window.opener) {
+    // In some cases this component is mounted on a url that is part of the login process and doesn't need to make another 
+    // request, because that data will be passed to the parent window and done there instead.
+    if (!window.location.pathname.includes('auth/login/callback')) {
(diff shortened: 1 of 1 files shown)


src/js/login/containers/Main.jsx
```diff
@@ -40,9 +40,9 @@ class Main extends React.Component {
     });
     this.bindNavbarLinks();
 
-    // If there is a window.opener, then this window was spawned by another instance of the website as the auth app.
-    // In this case, we don't need to actually login, because that data will be passed to the parent window and done there instead.
-    if (!window.opener) {
+    // In some cases this component is mounted on a url that is part of the login process and doesn't need to make another 
+    // request, because that data will be passed to the parent window and done there instead.
+    if (!window.location.pathname.includes('auth/login/callback')) {
```
raquelromano-gov: Moving my comment from https://github.com/department-of-veterans-affairs/vets-website/pull/6936/  to here:

Hard to test this thoroughly, but it does make sense. Is the worst case that this is too restrictive, so we end up making a few unnecessary extra fetches for login URLs (which I believe we were doing before for a while), but no user-perceived effect?

jbalboni: I believe so. I tested the login flow with the check completely removed and I was still able to login fine.

bshyong (approved):
This change makes sense to me

raquelromano-gov (approved):

PR #6945, opened 2017-12-08
Separate SiP logic in Redux code

jbalboni:
This is less intimidating than it looks!

There's very little new code here. The changes are:

1. `reducers/index.js` is very simple now and mostly just figures out the right set of reducer functions depending on SiP being on or off. I am going to move this file to the `state` folder later, this PR is big enough as it is.
2. The previous code in the above file is now in `state/reducers`, `save-in-progress/reducers`, and `formState.js`. There's also a different format for writing reducers, which is to use a lookup object instead of a big switch statement.
3. I moved `save-load-actions` to just `actions` and `sip-api` to `api`.  We know they're SiP related because of the folder they're in.
4. I moved some Redux logic into `selectors` files. I'm not sold on this completely yet, but it does keep the code for getting data out of Redux in one place, instead of multiple container files.

So, now instead of mixing in the SiP logic and state, it's in separate files and merged together with the normal state when needed.


cehsu (approved):
LGTM. Had a minor organizational question.

src/js/common/schemaform/reducers/index.js
```diff
@@ -1,247 +1,30 @@
...
-import {
-  createFormPageList,
-  checkValidSchema
-} from '../helpers';
-
 import {
-  updateSchemaAndData,
-  replaceRefSchemas,
-  updateItemsSchema
+  createInitialState,
```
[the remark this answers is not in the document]
jbalboni: Yeah, I'm planning on moving some of these file around in the next PR. I was trying to limit the size of this one.

src/js/common/schemaform/save-in-progress/selectors.js
```diff
@@ -0,0 +1,24 @@
+import { createStructuredSelector } from 'reselect';
+import { getFormData } from '../state/selectors';
+
+export const getSaveInProgressState = createStructuredSelector({
```
annekainicUSDS: I see this replaced the previous code that just used `createSelector`. Reading the docs and am a bit confused what creating a selector does in general and why the change to the structured one here?

annekainicUSDS (approved):
LGTM!

src/js/common/schemaform/save-in-progress/selectors.js
jbalboni: The structured selector is a short hand version of what we were doing before. Instead of making you create an object yourself in the function passed as the last argument, it creates the object for you based on the object you pass in.

PR #6947, opened 2017-12-08
DW bugfixes

bshyong:
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/6619
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/6626
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/6565

### DIFF
--- a/src/js/discharge-wizard/components/InstructionsPage.jsx
+++ b/src/js/discharge-wizard/components/InstructionsPage.jsx
@@ -5,9 +5,7 @@ class InstructionsPage extends React.Component {
   constructor(props) {
     super(props);
 
-    this.state = {
-      activeFAQ: null,
-    };
+    this.state = {};
   }
 
   componentDidMount() {
@@ -17,7 +15,7 @@ class InstructionsPage extends React.Component {
   handleFAQToggle = (e) => {
     e.preventDefault();
     this.setState({
-      activeFAQ: this.state.activeFAQ ? null : e.target.name
+      [e.target.name]: !this.state[e.target.name],
     });
   }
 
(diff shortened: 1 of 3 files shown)


src/js/discharge-wizard/components/InstructionsPage.jsx
```diff
@@ -110,8 +119,8 @@ class InstructionsPage extends React.Component {
                             </div>
                           </li>
                           <li itemScope itemType="http://schema.org/Question">
-                            <button className="usa-button-unstyled usa-accordion-button" aria-controls="dbq5" itemProp="name">What if I have a DD-215 showing an upgraded discharge, but my DD-214 is still incorrect?</button>
-                            <div id="dbq5" className="usa-accordion-content" itemProp="acceptedAnswer" itemScope itemType="http://schema.org/Answer">
+                            <button className="usa-button-unstyled usa-accordion-button" aria-controls="dbq5" itemProp="name" aria-expanded={this.state.activeFAQ === 'q5'} onClick={this.handleFAQToggle} name="q5">What if I have a DD-215 showing an upgraded discharge, but my DD-214 is still incorrect?</button>
```
U-DON: Can only one accordion be expanded at a time? If we should be able to expand multiple, perhaps we should keep track of each question in state.

U-DON (approved):
LGTM, but it looks like you'd be only able to expand one accordion at a time. I think it would make more sense to be able to open multiple, but I don't know the desired behavior.

src/js/discharge-wizard/components/InstructionsPage.jsx
```diff
@@ -2,16 +2,25 @@ import React from 'react';
...
-    script.src = 'https://standards.usa.gov/assets/js/vendor/uswds.min.js';
-    script.id = 'uswds';
-    document.getElementsByTagName('head')[0].appendChild(script);
     window.scrollTo(0, 0);
   }
 
+  handleFAQToggle = (e) => {
+    e.preventDefault();
+    this.setState({
+      activeFAQ: this.state.activeFAQ ? null : e.target.name
```
U-DON: If I'm reading this right, it seems like if there's already a question expanded, clicking another question would only close the currently opened question and not open the new one?

bshyong: Do’h you’re right. Need to check whether the question is already open

src/js/discharge-wizard/components/InstructionsPage.jsx
bshyong: That makes sense. I’ll make this change. Ditched the USWDS script because it doesn’t make sense for this use case (and is not designed to reinitialize inside of React components )

PR #6949, opened 2017-12-09
Lodash get and set replacement

cvalarida:
Contains hand-written lodash function replacements. It may make sense to rewrite some of the places that use certain functions rather than rewriting the functions themselves, but that's on a case-by-case basis.

I don't know that this is the best place to put these replacement functions and I don't know that that's the best filename for it either. I'm open to suggestions there for sure.

Also, I think I covered most of the test cases, but if you can think of anything else we need to test, please mention it!


src/js/common/utils/lodash-replacements/deconstructPath.js
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * Takes a string and casts it into an array.
+ * Can take strings like a.b[4].c
+ *
+ * @param {string} path
+ * @return {Array}
+ */
+export default function deconstructPath(path) {
+  const arrayPath = path.split('.');
```
jbalboni: Since array/property indexes will coerce to the right type, I think you could also do `.split(/[\.\[\]]/)` and filter out empty strings.

src/js/common/utils/lodash-replacements/get.js
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * Gets a the value at the end of the path.
+ *
+ * @param {Object} object
+ * @param {Array|string} path
+ * @param {*} [defaultValue]
+ * @return {*}
+ */
+export default function get(path, object, defaultValue) {
```
jbalboni: I forgot that I had a naive version of this here: https://github.com/department-of-veterans-affairs/vets-website/blob/rjsf-refactor-formapp/src/js/common/schemaform/formState.js#L13

If that helps any.

src/js/common/utils/lodash-replacements/get.js
```diff
@@ -0,0 +1,26 @@
...
+
+  for (let i = 0; i < arrayPath.length; i++) {
+    if (typeof currentValue[arrayPath[i]] !== 'undefined') {
+      currentValue = currentValue[arrayPath[i]];
+    } else {
+      return defaultValue;
+    }
+  }
+
+  // Should this clone? the current value? It might use a different ref--not sure.
```
jbalboni: I think not cloning is fine.

src/js/common/utils/lodash-replacements/set.js
```diff
@@ -0,0 +1,70 @@
...
+  if (!arrayPath.length) {
+    // We're at the end of our path; time to assign
+    return value;
+  }
+
+  // Do some one-time prep work
+  let newObj = object;
+  if (level === 0) {
+    // Only clone the whole object once
+    newObj = cloneDeep(object);
```
jbalboni: I don't think you need to deep clone the whole object. `set` should end up creating shallow copies of each object in the path (including the root).

src/js/common/utils/lodash-replacements/set.js
```diff
@@ -0,0 +1,70 @@
...
+    // The following path segments are not allowed
+    arrayPath = arrayPath.filter(e => e !== undefined && e !== null);
```
jbalboni: Could we throw an error when this occurs instead? Bad path segments seems like a case of developer error.

src/js/common/utils/lodash-replacements/set.js
```diff
@@ -0,0 +1,70 @@
...
+ * Same as `set`, but uses the level param to determine when to clone and give a more helpful error message.
+ *
+ * @param {Array|Obect} object
+ * @param {Array|string} path
+ * @param {*} value
+ * @param {Number} level  How many times we've recursed
+ * @return {Object} A new object with the appropriate value set
+ */
+function baseSet(path, value, object, level = 0) {
+  let arrayPath = Array.isArray(path) ? path : deconstructPath(path);
```
jbalboni: I'm curious why this is done at each level vs once at the beginning.

src/js/common/schemaform/validation.js
```diff
@@ -296,7 +296,7 @@ export function validateAddress(errors, address, formData, schema) {
 
   // Adds error message for state if it is blank and one of the following countries:
   // USA, Canada, or Mexico
-  if (_.includes(address.country)(['USA', 'CAN', 'MEX'])
+  if (['USA', 'CAN', 'MEX'].includes(address.country)
```
jbalboni: Could probably make this a Set, too.

Feels weird to say that in a real PR and not a homework eval :-)

src/js/common/utils/lodash-replacements/cloneDeep.js
```diff
@@ -0,0 +1,34 @@
...
+  const cloneArrayElement = e => {
+    // Deep clones arrays and objects
+    if (typeof e === 'object' && e !== null) {
+      return cloneDeep(e);
+    }
+
+    return e;
+  };
+
+  for (const key of Object.keys(newObj)) {
```
jbalboni: Is there a reason for this vs `forEach`?

src/js/common/schemaform/actions.js
```diff
@@ -1,5 +1,5 @@
 import Raven from 'raven-js';
-import _ from 'lodash/fp';
+import _ from '../utils/lodash-replacements';
```
jbalboni: I kind of think of these as data utils, but maybe that's too generic?

jbalboni (commented):
Looks good so far, had a few comments.

src/js/common/schemaform/actions.js
cvalarida: I wasn't sure where to put it them or how we should import them (as `_`? in `{ get, set }`? from different files entirely?) What's your opinion?

src/js/common/schemaform/validation.js
cvalarida: Hehe, I almost did. 😁 

Will do!

src/js/common/utils/lodash-replacements/clone.js
```diff
@@ -0,0 +1,10 @@
+/**
+ * Clones an object. Currently just uses Object.assign(), but we can change that later if desired.
+ *
+ * @param {Object} object
+ * @return {Object}
+ */
+export default function clone(object) {
```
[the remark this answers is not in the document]
cvalarida: Oh goodness. I forgot about those. I'll give it a shot and maybe just make a comment if it breaks?

src/js/common/utils/lodash-replacements/cloneDeep.js
cvalarida: Mmm...I think this is just vestigial from when I was doing it differently? I'll change it to use `forEach()`. Good call.

src/js/common/utils/lodash-replacements/deconstructPath.js
```diff
@@ -0,0 +1,36 @@
...
+    // Skip the nubers we insert
+    if (typeof e === 'number') {
+      return;
+    }
+
+    const arrayIndexMatches = e.match(arrayIndexRE);
+    const numOnlyMatches = e.match(numOnlyRE);
+    if (arrayIndexMatches) {
+      arrayPath[i] = arrayIndexMatches[1];
+      arrayPath.splice(i + 1, 0, parseInt(arrayIndexMatches[2], 10));
```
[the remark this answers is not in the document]
cvalarida: I think I tried that and it didn't work. I'll try again, though; me memory is nae so good.

src/js/common/utils/lodash-replacements/deconstructPath.js
cvalarida: Ahhh, instead of doing a regex match later? Clever. I like it!

src/js/common/utils/lodash-replacements/get.js
cvalarida: Ahh, very nice. Yeah, you're way is cleaner; I'll adapt.
image

src/js/common/utils/lodash-replacements/set.js
cvalarida: Because I added the `level` bit later and forgot about this. I'll fix it. 👍

src/js/common/utils/lodash-replacements/set.js
cvalarida: Note to self: Test the following

```js
const obj = {
  sub1: {
    foo: 'bar'
  },
  sub2: {
    baz: 'asdf'
  }
}

const copy = set(['sub1', 'foo'], 'newVal', obj);

copy.sub2.baz = 'changed on copy';

console.log(obj.sub2.baz);
```

src/js/common/utils/lodash-replacements/set.js
cvalarida: Yeah, I should also pull that bit out into a helper function to be used in `get()` as well, methinks.

src/js/common/schemaform/actions.js
cvalarida: Ohh, unless your comment is directed at the name `lodash-replacements`. Hmm?

src/js/common/utils/lodash-replacements/deconstructPath.js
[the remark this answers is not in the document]
cvalarida: I lied. I'm sorry. I used your suggestion above and had to use `parseInt()`. :man_shrugging:

src/js/common/utils/lodash-replacements/get.js
cvalarida: Without cloning, modifying the result of `_.get()` modifies the original:
```js
const obj = {
  sub: {
    foo: 'bar'
  }
};

const sub = _.get(['sub'], obj);
sub.foo = 'baz';

console.log(obj.sub.foo); // output: baz
```
Our options are
1. `cloneDeep` the return value to make it safe
2. Put a comment in there to note that modifying the return value isn't safe

src/js/common/utils/lodash-replacements/set.js
cvalarida: Without cloning, the following test (and others) fail:
```js
it('should not modify original object when changing an existing property', () => {
  const o = {
    a: 'a',
    b: { c: 'c' },
    k: { a: { y: 'f' } },
    g: ['h', 'i', 'j']
  };

  // Perhaps using cloneDeep in here is in bad taste?
  const oCopy = _.cloneDeep(o);

  _.set('b.c', 'd', o);
  expect(o).to.eql(oCopy);
});
```
The test output is as follows:
```
  2) lodash replacements set should not modify original object when changing an existing property:

      AssertionError: expected { Object (a, b, ...) } to deeply equal { Object (a, b, ...) }
      + expected - actual

       {
         "a": "a"
         "b": {
      -    "c": "d"
      +    "c": "c"
         }
         "g": [
           "h"
           "i"

      at Context.<anonymous> (test/common/utils/lodash-replacement.unit.spec.js:207:20)
```
Am I missing something?

jbalboni: I think you still need to shallow clone at the top level, just not deep clone. Doing a deep clone defeats the purpose of a function/immutable version of `set`.

If I call `set('b.c', ...)` on the object in your test, the object containing `b` and the object containing `c` should not shallowly equal the previous objects at those levels. But `g`, for example, should be shallowly equal in both the new and the old version.

src/js/common/schemaform/actions.js
jbalboni: Yeah, I meant `lodash-replacements`.

test/common/utils/lodash-replacement.unit.spec.js
```diff
@@ -0,0 +1,290 @@
...
+    it('should set the value of an existing path', () => {
+      const o = {
+        a: 'a',
+        b: { c: 'c' },
+        k: { a: { y: 'f' } },
+        g: ['h', 'i', 'j']
+      };
+
+      const newObj = _.set('b.c', 'd', o);
+      expect(newObj.b.c).to.equal('d');
```
jbalboni: Should these set tests also include a check to make sure the rest of the object is deeply equal to what we expect?

test/common/utils/lodash-replacement.unit.spec.js
```diff
@@ -0,0 +1,290 @@
...
+
+    it('should return undefined if not found and no default is provided', () => {
+      expect(_.get('does.not.exist', o)).to.equal();
+    });
+  });
+
+
+  describe('set', () => {
+    it('should set the value of an existing path', () => {
+      const o = {
```
jbalboni: It looks like this object is the same for all the tests, so you could define it once in the describe block.

jbalboni (approved):
LGTM

test/common/utils/lodash-replacement.unit.spec.js
cvalarida: I was going to, but since we're also testing whether this is mutative on the original object, I thought it would be safer to create a new object per test.

test/common/utils/lodash-replacement.unit.spec.js
cvalarida: Yeah, that's probably wise. Will do. 👍

PR #6953, opened 2017-12-12
Add SiP check to form title in FormApp

jbalboni:
My previous SiP refactor removed the SiP state from forms with it disabled, and this one line didn't have a SiP check and was triggering the wrong condition.

The result was two titles on the pre-need intro page, on from IntroductionPage and the other from FormApp.

### DIFF
--- a/src/js/common/schemaform/FormApp.jsx
+++ b/src/js/common/schemaform/FormApp.jsx
@@ -207,7 +207,6 @@ class FormApp extends React.Component {
       );
     }
 
-
     return (
       <div>
         <div className="row">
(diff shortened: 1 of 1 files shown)


src/js/common/schemaform/FormApp.jsx
```diff
@@ -216,7 +215,7 @@ class FormApp extends React.Component {
             {
               formConfig.title &&
               // If we’re on the introduction page, show the title if we’re actually on the loading screen
-              (!isIntroductionPage || this.props.loadedStatus !== LOAD_STATUSES.notAttempted) &&
+              (!isIntroductionPage || (!formConfig.disableSave && this.props.loadedStatus !== LOAD_STATUSES.notAttempted)) &&
```
jbalboni: When SiP is disabled, this status is undefined.

annekainicUSDS (approved):
LGTM! I think this will also have to be pulled into my PR somehow that refactors the FormApp code, but we can figure that out later.

jbalboni:
@annekainicUSDS Possibly, I think your PR might remove the need for this check.

PR #6954, opened 2017-12-12
Set up custom select accessibility tests

cehsu:
These changes create additional accessibility tests beyond our automated tools, to ensure that custom components, such as the select used in the facility locator behave as expected. 

Behaviors identified by the [accessibility audit](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4698) have been focused on here.

Feedback on how these tests are set up and what is tested in terms of behaviors are both welcome.

Another idea is to make use of a domain specific language such as Gherkin via a tool such as [nightwatch-cucumber](http://mucsi96.github.io/nightwatch-cucumber/), this was not done here but feedback on that proposal is also appreciated.


test/facility-locator/01-accessibility.e2e.spec.js
```diff
@@ -0,0 +1,57 @@
...
+    client
+      .waitForElementVisible('body', Timeouts.normal)
+      .waitForElementVisible('.facility-locator', Timeouts.slow)
+      // do not run 'wcag2a' rules because of open aXe bug https://github.com/dequelabs/axe-core/issues/214
+      .axeCheck('.main', { rules: ['section508'] });
+
+    // Traverse form controls via keyboard input
+    client
+      .sendKeys('input[name="streetCityStateZip"]', client.Keys.TAB)
+      .execute(FacilityHelpers.checkActiveElement, ['div[class="facility-dropdown-wrapper"]'], result =>
```
jbalboni: Can you make this a custom command or assertion? I haven't tried that for assertions, would be nice if that were possible.

jbalboni (approved):
This is very cool. Just one question.

test/facility-locator/01-accessibility.e2e.spec.js
cehsu: That would be much tidier, great idea.

PR #6956, opened 2017-12-12
Facility locator bug fix

cehsu:
These changes fix some bugs on the facility locator page, including:
- overlap of map and nav menu
- conflicting classes on disabled select button


jbalboni (approved):
LGTM

PR #6960, opened 2017-12-13
Reorganize schemaform folder

jbalboni:
This rearranges the schemaform folder to hopefully make things easier to find. I'm not tied to any of this, so let me know if you think things should be in different places.

The diff is probably going to be hard to view, so I would try looking at the schemaform folder in this branch and looking through the different folders.


cehsu (approved):
LGTM.

PR #6961, opened 2017-12-14
Ignore LOA for health tools with MHV sign-in.

U-DON:
This is to support allowing non-premium MHV sign-in users to access their apps without identity proofing. This ultimately boils down to determining access based purely on the list of available services when accessing one of the health tools as an MHV sign-in user.

### DIFF
--- a/src/js/common/components/RequiredLoginView.jsx
+++ b/src/js/common/components/RequiredLoginView.jsx
@@ -104,9 +104,7 @@ RequiredLoginView.propTypes = {
     PropTypes.string,
     PropTypes.array,
   ]).isRequired,
-  userProfile: PropTypes.object.isRequired,
-  loginUrl: PropTypes.string,
-  verifyUrl: PropTypes.string
+  userProfile: PropTypes.object.isRequired
 };
 
 export default RequiredLoginView;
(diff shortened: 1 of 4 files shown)


bshyong (approved):
LGTM

src/js/common/components/RequiredLoginView.jsx
```diff
@@ -57,10 +65,10 @@ class RequiredLoginView extends React.Component {
           attemptingAppealsAccess = requiredApp === 'appeals-status';
         }
 
-        if (this.props.userProfile.status === 'SERVER_ERROR' && !attemptingAppealsAccess) {
+        if (userProfile.status === 'SERVER_ERROR' && !attemptingAppealsAccess) {
           // If va_profile is null, show a system down message.
           return <SystemDownView messageLine1="Sorry, our system is temporarily down while we fix a few things. Please try again later."/>;
-        } else if (this.props.userProfile.status === 'NOT_FOUND' && !attemptingAppealsAccess) {
+        } else if (userProfile.status === 'NOT_FOUND' && !attemptingAppealsAccess) {
```
saneshark: @U-DON could you confirm that this check doesn't interfere with Basic and Advanced users signing in via MHV? Keep in mind these users will NOT have MVI data. I think the comment above is wrong. It says "if va_profile is null, show a system down message." Thats no longer accurate.

If it results in a server error or not found, then that is one thing. But for MHV sign-in users who are not premium, we won't even fetch MVI on the backend so MVI va profile will ALWAYS be null (even though there isn't an error).

saneshark:
@U-DON might be worth having a unit spec of sorts around va_profile == "NOT AUTHORIZED" still working with MHV sign-in, something tells me someone could add an additional condition for that in the future and maybe break our logic and we wouldn't really have anyway knowing. Thoughts?

PR #6964, opened 2017-12-14
Rjsf review e2e tests

annekainicUSDS:


test/common/schemaform/review/00-complete-form.e2e.spec.js
```diff
@@ -0,0 +1,97 @@
...
+
+    client.expect.element('.review-row span').text.to.equal('Jacques');
+
+    // Edit form fields unsuccessfully
+    client.waitForElementVisible('.edit-btn', Timeouts.normal);
+    client
+      .click('.edit-btn');
+    client.expect.element('input[name="root_veteranFullName_first"]').to.be.visible;
+
+    client.clearElement('input[name="root_veteranFullName_first"]');
```
annekainicUSDS: Here's where I use `clearElement`

test/common/schemaform/review/00-complete-form.e2e.spec.js
```diff
@@ -0,0 +1,97 @@
...
+    client.expect.element('input[name="root_veteranFullName_first"]').text.to.equal('');
+    client
+      .pause(1200)
+      .click('input[name="root_veteranFullName_middle"]');
```
annekainicUSDS: Redirect to the next field to remove focus from the first name field

test/common/schemaform/review/00-complete-form.e2e.spec.js
```diff
@@ -0,0 +1,97 @@
...
+    client.expect.element('input[name="root_veteranFullName_first"]').text.to.equal('');
+
+    client
+      .click('.usa-button-primary')
```
annekainicUSDS: Once we click the "Update" button, the original value for the first name field (`Jacques`) reappears and the form successfully updates, when it should have failed.

test/common/schemaform/review/01-blank-form.e2e.spec.js
```diff
@@ -0,0 +1,59 @@
...
+
+    // Try to save form with validation errors
+    client
+      .waitForElementVisible('.edit-btn', Timeouts.normal)
+      .click('.edit-btn');
+    client.expect.element('input[name="root_veteranFullName_first"]').to.be.visible;
+
+    client
+      .waitForElementVisible('.form-review-panel-page .usa-button-primary', Timeouts.normal)
+      .click('.form-review-panel-page .usa-button-primary');
```
annekainicUSDS: This test is failing here; sometimes the button gets clicked, and sometimes it doesn't, causing the next line that checks for the error class to be present to fail.

annekainicUSDS:
Encountering 2 issues with these tests that are causing each to fail:

1. For the first test, 00-complete-form, the test is failing when trying to clear a field and test that saving that section fails due to a validation error with the blank field. When running the e2e test, I'm seeing that the field appears to clear when I use any combination of `removeText()`, `clearValue()`, `clearElement()`, or just `setValue(selector, '')`, but then when I navigate to the next field or click the "Update" button, the original value that I had just cleared reappears and the section successfully saves. I have no idea why this is happening, have done a ton of research and reading of the docs and can't find this issue documented anywhere, which makes me wonder if it's something specific about HCA?

2. For the second test, 01-blank-form, sometimes the test successfully clicks the "Update" button and and sometimes it doesn't (mostly doesn't). I can't figure out why; reading the test output, even with `--verbose`, isn't giving me any clues as to why it hangs up at that point. I left a comment in the test file where it fails.

test/common/schemaform/review/00-complete-form.e2e.spec.js
cehsu: Strange, could another field such as service start date be tested instead?

test/common/schemaform/review/01-blank-form.e2e.spec.js
cehsu: Could repeating the click help to ensure that the button gets clicked without disrupting the test?

jbalboni (approved):

PR #6968, opened 2017-12-15
Patch for login links disappearing while the request for user info is pending

ncksllvn:
Currently, there are no login links (Sign in or the user's name) visible when the page loads. This is because certain information is needed before displaying that information. This PR changes that so that those links are visible immediately, but faded out to show that they are disabled. It think this is an improvement over hiding the links altogether. 

Resolves department-of-veterans-affairs/vets.gov-team/issues/6077

### DIFF
--- a/src/js/common/helpers/login-helpers.js
+++ b/src/js/common/helpers/login-helpers.js
@@ -64,38 +64,34 @@ export function getUserData(dispatch) {
     error.status = response.status;
     throw error;
   }).then(json => {
-    if (json.data) {
-      const userData = json.data.attributes.profile;
-      // sessionStorage coerces everything into String. this if-statement
-      // is to prevent the firstname being set to the string 'Null'
-      if (userData.first_name) {
-        sessionStorage.setItem('userFirstName', userData.first_name);
-      }
-      dispatch(updateProfileFields({
-        savedForms: json.data.attributes.in_progress_forms,
-        prefillsAvailable: json.data.attributes.prefills_available,
-        accountType: userData.loa.current,
-        email: userData.email,
-        userFullName: {
-          first: userData.first_name,
-          middle: userData.middle_name,
-          last: userData.last_name,
-        },
-        authnContext: userData.authn_context,
-        loa: userData.loa,
-        multifactor: userData.multifactor,
-        gender: userData.gender,
-        dob: userData.birth_date,
-        status: json.data.attributes.va_profile.status,
-        veteranStatus: json.data.attributes.veteran_status.status,
-        isVeteran: json.data.attributes.veteran_status.is_veteran,
-        services: json.data.attributes.services,
-        healthTermsCurrent: json.data.attributes.health_terms_current,
-      }));
-      dispatch(updateLoggedInStatus(true));
-    } else {
-      dispatch(profileLoadingFinished());
+    const userData = json.data.attributes.profile;
+    // sessionStorage coerces everything into String. this if-statement
+    // is to prevent the firstname being set to the string 'Null'
+    if (userData.first_name) {
+      sessionStorage.setItem('userFirstName', userData.first_name);
     }
+    dispatch(updateProfileFields({
+      savedForms: json.data.attributes.in_progress_forms,
+      prefillsAvailable: json.data.attributes.prefills_available,
+      accountType: userData.loa.current,
+      email: userData.email,
+      userFullName: {
+        first: userData.first_name,
+        middle: userData.middle_name,
+        last: userData.last_name,
+      },
+      authnContext: userData.authn_context,
+      loa: userData.loa,
+      multifactor: userData.multifactor,
+      gender: userData.gender,
+      dob: userData.birth_date,
+      status: json.data.attributes.va_profile.status,
+      veteranStatus: json.data.attributes.veteran_status.status,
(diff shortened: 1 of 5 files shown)


jbalboni:
@ncksllvn I'm sort of on the fence about whether this is better or not, but I think most of the problem report in the connected ticket is that we had an issue where sign in links would never load if you had opened the site in a new tab from a link.

src/js/common/helpers/login-helpers.js
```diff
@@ -93,6 +96,13 @@ export function getUserData(dispatch) {
     } else {
       dispatch(profileLoadingFinished());
     }
+  }).catch(error => {
```
jbalboni: If we're going to add this `catch` block, you can probably remove the `json.data` check in the `then` above it, since that's the existing error handling (such as it is).

src/js/login/components/SearchHelpSignIn.jsx
```diff
@@ -24,28 +24,40 @@ class SearchHelpSignIn extends React.Component {
...
         clickHandler={() => {
           this.props.toggleSearchHelpUserMenu('account', !login.utilitiesMenuIsOpen.account);
         }}
         greeting={greeting}
         isOpen={login.utilitiesMenuIsOpen.account}
         onUserLogout={this.props.onUserLogout}/>);
-    } else if (this.props.profile.loading) {
-      content = null;
     } else {
+      const classList = [
```
jbalboni: Could use `classnames` for this.

src/js/login/components/SearchHelpSignIn.jsx
```diff
@@ -24,28 +24,40 @@ class SearchHelpSignIn extends React.Component {
...
+        isLoading ? 'disabled' : ''
+      ].join(' ');
+
       content = (<div>
-        <a href="#" onClick={this.handleSigninSignup}><span>Sign in</span><span className="signin-spacer">|</span><span>Sign up</span></a>
+        <a href="#" className={classList} onClick={this.handleSigninSignup}><span>Sign in</span><span className="signin-spacer">|</span><span>Sign up</span></a>
```
jbalboni: I wonder if we should just make this a button with the `va-button-link` class.

jbalboni (approved):
LGTM, but @bshyong should definitely take a look since he knows the login flow better than me.

src/js/common/helpers/login-helpers.js
ncksllvn: Ah gotcha, good catch. It makes sense to me to separate the error handling this way. I haven't checked, but wouldn't the previous way have errored out since `undefined` is being passed to `then` on line 66 if `response.ok` is false? Doesn't really matter if we're going this route anyway but just something I noticed.

src/js/login/components/SearchHelpSignIn.jsx
```diff
@@ -24,28 +25,36 @@ class SearchHelpSignIn extends React.Component {
...
         }}
         greeting={greeting}
         isOpen={login.utilitiesMenuIsOpen.account}
         onUserLogout={this.props.onUserLogout}/>);
-    } else if (this.props.profile.loading) {
-      content = null;
     } else {
       content = (<div>
-        <a href="#" onClick={this.handleSigninSignup}><span>Sign in</span><span className="signin-spacer">|</span><span>Sign up</span></a>
+        <a href="#" className={classNames({ disabled: isLoading })} onClick={this.handleSigninSignup}><span>Sign in</span><span className="signin-spacer">|</span><span>Sign up</span></a>
```
bshyong: nit: can we pull out the classNames invocation into a constant instead of having it inline?

bshyong (approved):

ncksllvn:
Thank you @jbalboni for the CR! From conversations it sounded like the links would sometimes take several seconds to show up and it would confuse users. The idea to disable the links while that login request is pending was mine though - do you think that this is bad UX? Those links always being visible on page load is an improvement IMO.

src/js/login/components/SearchHelpSignIn.jsx
ncksllvn: Definitely! I couldn't decide what to name that variable since I want to name is `classNames`

src/js/common/helpers/login-helpers.js
jbalboni: With the `response.ok` check that you added, yeah it would throw an `undefined` error. I meant that the way it was before, without the `response.ok` check, the `json.data` check is essentially the same error check (our errors have json payloads, but they don't have `data` in them).

I agree that the catch block is better.

jbalboni:
@ncksllvn I don't think it's worse than what we have now, but users are going to have a similar question. Right now that question is, "where is the sign in link?" With this change it's, "why is the sign in link disabled?" There's no indication that something is happening in the background.

ncksllvn:
@jbalboni Hm, good point. I was thinking that with this approach it's less likely that it will raise any concern to the user, because they probably notice those links missing more often than they actually need to click those links. I feel like a loading icon is too much. Maybe add a title attribute to the links that tells the user it's loading still?

src/js/common/helpers/login-helpers.js
ncksllvn: Ah that makes sense, good to know. Thanks for the explanation!

PR #6975, opened 2017-12-18
Post911 print sr mg

markgreenburg:
- Ensures screen readers don't read background behind print modal
- Adds print and back buttons to printable statement (these don't display once print button is clicked)
- Adds some print-specific styling
- Adds a few other small accessibility tweaks for print and status pages (focus elements once components are rendered)
- Accompanying tests

### DIFF
--- a/src/js/post-911-gib-status/containers/PrintPage.jsx
+++ b/src/js/post-911-gib-status/containers/PrintPage.jsx
@@ -18,6 +18,9 @@ class PrintPage extends React.Component {
     document.querySelector('nav').classList.remove('no-print-no-sr');
   }
 
+  backToStatement = () => this.props.router.push('/');
+  printWindow = () => window.print();
+
   render() {
     const enrollmentData = this.props.enrollmentData || {};
 
@@ -29,8 +32,18 @@ class PrintPage extends React.Component {
           <div className="print-screen">
             <img src="/img/design/logo/va-logo.png" alt="VA logo" width="300"/>
             <h1 className="section-header">Post-9/11 GI Bill<sup>&reg;</sup> Statement of Benefits</h1>
-            <p>The information in this letter is the Post-9/11 GI Bill Statement of Benefits for the beneficiary listed below as of {todayFormatted}. Any pending or recent changes to enrollment may affect remaining entitlement.</p>
+            <button
+              className="usa-button-primary"
+              onClick={this.printWindow}>Print This Page</button>
+            <p>
+              The information in this letter is the Post-9/11 GI Bill Statement of Benefits for
+              the beneficiary listed below as of {todayFormatted}. Any pending or recent changes to
+              enrollment may affect remaining entitlement.
+            </p>
             <UserInfoSection enrollmentData={enrollmentData}/>
+            <button
+              className="usa-button-secondary"
+              onClick={this.backToStatement}>Back to Statement Page</button>
           </div>
         </div>
       </div>
--- a/src/js/post-911-gib-status/containers/PrintPage.jsx
+++ b/src/js/post-911-gib-status/containers/PrintPage.jsx
@@ -3,10 +3,11 @@ import { connect } from 'react-redux';
 
 import UserInfoSection from '../components/UserInfoSection';
 
-import { formatDateLong } from '../../common/utils/helpers';
+import { formatDateLong, focusElement } from '../../common/utils/helpers';
 
 class PrintPage extends React.Component {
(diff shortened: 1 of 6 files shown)


src/js/post-911-gib-status/containers/PrintPage.jsx
```diff
@@ -3,9 +3,25 @@ import { connect } from 'react-redux';
...
+    focusElement('.print-screen');
+    document.querySelector('header').classList.add('no-print-no-sr');
+    document.querySelector('footer').classList.add('no-print-no-sr');
+    document.querySelector('.va-nav-breadcrumbs').classList.add('no-print-no-sr');
+  }
+
+  componentWillUnmount() {
+    document.querySelector('header').classList.remove('no-print-no-sr');
+    document.querySelector('footer').classList.remove('no-print-no-sr');
+    document.querySelector('.va-nav-breadcrumbs').classList.remove('no-print-no-sr');
```
cvalarida: 🎉

src/js/post-911-gib-status/containers/StatusPage.jsx
```diff
@@ -31,7 +36,7 @@ class StatusPage extends React.Component {
       printButton = (
         <div className="section">
           <button onClick={this.navigateToPrint} className="usa-button-primary" id="print-button">
-            Print Statement of Benefits
+            Get Printable Statement of Benefits
```
cvalarida: Good call.

src/sass/post-911-gib-status.scss
```diff
@@ -82,11 +82,27 @@
...
+
+      button {
+        display: none;
+      }
     }
   }
 }
 
+.no-print-no-sr {
+    display: none !important;
```
cvalarida: This feels like it'd probably be helpful outside of GIBS. What do you think of moving it to the main stylesheet?

test/post-911-gib-status/components/PrintPage.unit.spec.jsx
```diff
@@ -0,0 +1,56 @@
+import React from 'react';
+import { expect } from 'chai';
+import { shallow } from 'enzyme';
+import sinon from 'sinon';
+
+import { PrintPage } from '../../../src/js/post-911-gib-status/containers/PrintPage';
+
+describe('<PrintPage/>', () => {
+  const pushSpy = sinon.spy();
```
cvalarida: Just to be safe, what do you think of resetting this spy after each test?
```js
afterEach(() => pushSpy.reset());
```

test/post-911-gib-status/components/PrintPage.unit.spec.jsx
```diff
@@ -0,0 +1,56 @@
...
+  const defaultProps = {
+    router: { push: pushSpy },
+    enrollmentData: {}
+  };
+
+  it('should render', () => {
+    const wrapper = shallow(<PrintPage {...defaultProps}/>, { disableLifecycleMethods: true });
```
cvalarida: I think you told me the other day, but why disable the lifecycle methods on these?

src/sass/post-911-gib-status.scss
```diff
@@ -82,11 +82,27 @@
     }
 
     @media print {
+      font: $font-serif;
+      color: $color-gray-dark;
       padding: 0;
+      a {
+        text-decoration: none;
+        &:after {
+          content:" (https://vets.gov" attr(href) ")";
```
cvalarida: Heyyyy, clever. I like it. 💯

cvalarida (approved):
:shipit:

src/sass/post-911-gib-status.scss
markgreenburg: Yeah, I went back and forth on making it a mixin, but applying directly to the va stylesheet is probably more appropriate.

test/post-911-gib-status/components/PrintPage.unit.spec.jsx
markgreenburg: If they're not disabled, they try to do DOMy things, which breaks the tests, because the lifecycle method in question tries to select header & footer elements directly from the DOM. These don't exist in the context of the unit test, or really the React app at all, since they live in the megamenu.

test/post-911-gib-status/components/PrintPage.unit.spec.jsx
markgreenburg: Ah! Good call.

PR #6978, opened 2017-12-19
Edit VIC Error messages about military service history

ayaleloehr:
Previously the same error message was being returned when both eMIS was down and a user was not in eMIS. This makes it so more specific error messages are shown, matching the text approved in https://github.com/department-of-veterans-affairs/vets.gov-team/issues/6673#issuecomment-352493969.  

Closes https://github.com/department-of-veterans-affairs/vets.gov-team/issues/6673


patrickvinograd (approved):
LGTM.

PR #6982, opened 2017-12-19
Don't redirect to login if API requests fail in auth flow.

U-DON:
Resolves department-of-veterans-affairs/vets.gov-team#6600.

This solves an edge case where logging in might result in invalid data in the SAML callback (perhaps a wrong token) and consequently unauthorized requests to the API.


bshyong (approved):
LGTM

PR #6984, opened 2017-12-19
Add events to support login custom dimension

robertfairhead:
This adds two events to the login flow to help us track user logins.

One should fire whenever we set the user attribute of `loggedIn` to true. When picked up, we'll mark the whole session as logged in from that point forward.

The other fires events to show if the LOA level is 1 or 3. We'll track this for the session as well.

Combined these events will allow us to create custom dimensions to help answer questions about usage patterns while logged in.

Nick touched this most recently. I tagged jeff and ben more as a "cc" since they seem to have done some heavy lifting on this in the past.

Closes: department-of-veterans-affairs/vets.gov-team#4539

### DIFF
--- a/src/js/common/helpers/login-helpers.js
+++ b/src/js/common/helpers/login-helpers.js
@@ -70,7 +70,7 @@ export function getUserData(dispatch) {
     if (userData.first_name) {
       sessionStorage.setItem('userFirstName', userData.first_name);
     }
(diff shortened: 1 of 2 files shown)


ncksllvn (approved):
LGTM!

jbalboni (approved):
LGTM, as long as you're ok with these firing on every page load for static pages and only on the first page in a React app.

PR #6985, opened 2017-12-20
Update review tests to remove pauses

jbalboni:
I think the pauses are causing some instability, so this should help. The other SiP tests have already been updated.


cehsu (approved):
LGTM.

cvalarida (approved):

PR #6990, opened 2017-12-21
Add letters list test

cvalarida:
Had to do some fancy footwork to make sure all the checkboxes got unchecked, but it was pretty straightforward otherwise.

### DIFF
--- a/test/letters/01-authed.e2e.spec.js
+++ b/test/letters/01-authed.e2e.spec.js
@@ -83,6 +83,43 @@ module.exports = E2eHelpers.createE2eTest(
       .click('div.step-content > p:nth-child(3) > a') // link to go back to confirm-address
       .expect.element('.city-state').to.be.present.before(Timeouts.normal);
 
+    // -- Go to letters list -- //
+
+    client
+      .click('.view-letters-button')
+      .assert.urlContains('/letters/letter-list')
+      .waitForElementVisible('.step-content', Timeouts.normal)
+      .click('.step-content div.form-review-panel:nth-of-type(4) button') // open the bsl accordion
+      .waitForElementPresent('#militaryService', Timeouts.normal);
+
+    // poke all the checkboxes and expect them to all be unselected
+    client
+      .expect.element('#militaryService').to.have.attribute('checked').equals('true');
+    client
+      .click('#militaryService')
+      .expect.element('#militaryService').to.not.have.attribute('checked');
+    client.execute(() => {
+      return Array.from(document.querySelectorAll('#benefitInfoTable input[type="checkbox"]'))
+        .map(e => e.getAttribute('id'));
+    }, [], (obj) => {
+      const ids = obj.value;
+      ids.forEach(id => {
+        client.expect.element(`#${id}`).to.have.attribute('checked').equals('true');
+        client.click(`#${id}`);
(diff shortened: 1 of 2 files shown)


test/letters/01-authed.e2e.spec.js
```diff
@@ -83,6 +83,43 @@ module.exports = E2eHelpers.createE2eTest(
       .click('div.step-content > p:nth-child(3) > a') // link to go back to confirm-address
       .expect.element('.city-state').to.be.present.before(Timeouts.normal);
 
+    // -- Go to letters list -- //
+
+    client
+      .click('.view-letters-button')
+      .assert.urlContains('/letters/letter-list')
+      .waitForElementVisible('.step-content', Timeouts.normal)
+      .click('.step-content div.form-review-panel:nth-of-type(4) button') // open the bsl accordion
```
markgreenburg: 😬

test/letters/01-authed.e2e.spec.js
```diff
@@ -83,6 +83,43 @@ module.exports = E2eHelpers.createE2eTest(
       .click('div.step-content > p:nth-child(3) > a') // link to go back to confirm-address
       .expect.element('.city-state').to.be.present.before(Timeouts.normal);
 
+    // -- Go to letters list -- //
+
+    client
+      .click('.view-letters-button')
+      .assert.urlContains('/letters/letter-list')
```
markgreenburg: Is the new URL available instantly or does there need to be a wait somewhere in there? Not sure how that works with Router.

test/letters/01-authed.e2e.spec.js
```diff
@@ -83,6 +83,43 @@ module.exports = E2eHelpers.createE2eTest(
...
+    // poke all the checkboxes and expect them to all be unselected
+    client
+      .expect.element('#militaryService').to.have.attribute('checked').equals('true');
+    client
+      .click('#militaryService')
+      .expect.element('#militaryService').to.not.have.attribute('checked');
+    client.execute(() => {
+      return Array.from(document.querySelectorAll('#benefitInfoTable input[type="checkbox"]'))
+        .map(e => e.getAttribute('id'));
+    }, [], (obj) => {
```
markgreenburg: Interesting . . . this callback receives an object that includes the value returned from body param function.

markgreenburg (approved):
:shipit:

test/letters/01-authed.e2e.spec.js
cvalarida: I considered dynamically adding an id to these accordion buttons, but it seems like more work than it was worth. I pretty much hate this selector, but it works. Until we change things. Then we cry. 😢

test/letters/01-authed.e2e.spec.js
cvalarida: It should be available instantly. If this fails suspiciously later, we can reassess. :man_shrugging:

test/letters/01-authed.e2e.spec.js
cvalarida: It's a fun, undocumented feature. 😏 
'Course, I say that, but it probably is documented somewhere. I just had to `console.log` and experiment to figure out how to make it work.

PR #6991, opened 2017-12-21
discharge page focus

bshyong:
resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/6859

We still want the focus on the download form button, but have the window scrolled to the top of the page


U-DON (approved):
LGTM.

PR #6994, opened 2017-12-21
WIP - Session Timeout Modal

ncksllvn:
### How do we know when the session is going to timeout?
The token expires one hour after issued, but that is expiration is increased by one hour per each API request for a maximum of 12 hours. So, if the user is refreshing the token all day at the end of each hour, it can be used for 12 hours and 59 minutes before it becomes invalid.
    
### Pure UI Ideas
- Centralize API access:
        - Override window.fetch (bad idea)
        - Switch all API access methods to use the API helper in the common directory (probably the best)
- Create a custom middleware that checks for network access via promises (probably the easiest)
- A big long reducer that expands the session expiration when certain actions are dispatched (a very controlled way but would be some work to maintain)

### API Ideas
- Include a header in API responses about when the authorization expires
    - This at least seems more reliable than having the front-end do all of the calculation for session expiry.
- API Route for getting session expiry
    - This is a little tricky because the UI still has to determine when to hit that route, but this is way seems like the most reliable way to do it. The UI can probably check every 45 minutes, or if it detects that the website was left idle for an undetermined amount of time.

Resolves department-of-veterans-affairs/vets.gov-team/issues/6582
  
On hold until API is implemented and Frontend's API access is centralized -
department-of-veterans-affairs/vets.gov-team/issues/7053
department-of-veterans-affairs/vets.gov-team/issues/7102

### DIFF
--- a/src/js/login/actions/index.js
+++ b/src/js/login/actions/index.js
@@ -3,6 +3,7 @@ import { gaClientId } from '../../common/utils/helpers';
 export const LOG_OUT = 'LOG_OUT';
 export const TOGGLE_LOGIN_MODAL = 'TOGGLE_LOGIN_MODAL';
 export const UPDATE_LOGGEDIN_STATUS = 'UPDATE_LOGGEDIN_STATUS';
+export const UPDATE_SESSION_EXPIRES_SOON = 'UPDATE_SESSION_EXPIRES_SOON';
 export const UPDATE_LOGIN_URLS = 'UPDATE_LOGIN_URLS';
 export const UPDATE_LOGOUT_URL = 'UPDATE_LOGOUT_URL';
 export const UPDATE_MULTIFACTOR_URL = 'UPDATE_MULTIFACTOR_URL';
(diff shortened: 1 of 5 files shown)


ncksllvn:
@patrickvinograd @omgitsbillryan Do you guys have any ideas on how best to keep track of session timeout? In the frontend, we can probably centralize our API hits and can keep track of the session timeout that way, but maybe you know of a better way? I don't know if response headers can be used for that sort of thing or that's a security concern. Just looking for ideas.

patrickvinograd:
Off the top of my head, I don't see a security concern with returning a header like `X-Session-Expires` in response to any authenticated API call. This would centralize the expiry logic in the backend where it  can be provided in a consistent way with how vets-api is enforcing expiry.

omgitsbillryan:
> Include a header in API responses about when the authorization expires
:+1: 

> So, if the user is refreshing the token all day at the end of each hour, it can be used for 12 hours and 59 minutes before it becomes invalid.

You're right that the backend timeout is variable; anywhere between 12h1m to 12h59m.  This was done for backend convenience but when we start returning this as a response header it should be tightened up to hard-expire at 12 hours.  I've created 2 issues to track this work:
- https://github.com/department-of-veterans-affairs/vets.gov-team/issues/7054
- https://github.com/department-of-veterans-affairs/vets.gov-team/issues/7053

ncksllvn:
@omgitsbillryan That sounds like a great plan to me.

The frontend will still need to centralize API access in order to check that header after each response and adjust the session timeout accordingly, so there's still a solid amount of work, but that's a change we've been meaning to make anyway I think.

Related issues about making API access consistent -
department-of-veterans-affairs/vets.gov-team/issues/3537
department-of-veterans-affairs/vets.gov-team/issues/6983

ncksllvn:
Closing this because this code is too outdated now

PR #7000, opened 2017-12-22
Update pension and preneed content

cehsu:
These changes update the form control at the introduction page and the preneed reference to 90 days (which as been removed).

Issues
https://github.com/department-of-veterans-affairs/vets.gov-team/issues/6657
https://github.com/department-of-veterans-affairs/vets.gov-team/issues/6850

Pensions introduction page prefill message:
screen shot 2018-01-02 at 3 21 36 pm

Burial introduction page prefill message:
screen shot 2018-01-02 at 3 21 20 pm

Preneed introduction page
screen shot 2017-12-22 at 7 55 23 am

Preneed confirmation page
screen shot 2017-12-22 at 7 57 46 am


peggygannon (approved):
@cehsu Looks good to me. thanks!

src/js/common/schemaform/save-in-progress/SaveInProgressIntro.jsx
```diff
@@ -35,16 +35,29 @@ export default class SaveInProgressIntro extends React.Component {
...
-          <div>
-            <div className="usa-alert usa-alert-info schemaform-sip-alert">
-              <div className="usa-alert-body">
-                <strong>Note:</strong> Since you’re signed in to your account, we can prefill part of your application based on your account details. You can also save your form in progress, and come back later to finish filling it out.
+        if (this.props.formId === '21P-527EZ') {
+          alert = (
+            <div>
+              <div className="usa-alert usa-alert-info schemaform-sip-alert">
+                <div className="usa-alert-body">
+                  <strong>Note:</strong>  If you’re signed in to your account, we can prefill part of your application based on your account details. You can also save your form in progress, and come back later to finish filling it out. You have 60 days from the date you start your pension application to submit the form. After 60 days, the form won't be saved and you'll need to start over.
```
U-DON: Are there supposed to be two spaces after "Note:"? Other similar content seems to only have one.

U-DON: Also, why is this "If you're signed in..."? If the user is getting to this point, aren't they definitely signed in since `prefillAvailable` is truthy?

U-DON (approved):
Changes look good. Some questions about the actual content.

src/js/common/schemaform/save-in-progress/SaveInProgressIntro.jsx
cehsu: Thanks @U-DON, good catch with the spacing, this has been fixed. As for `prefillAvailable`, that refers to the form, rather than the user.

cehsu: Thanks @U-DON, good catches, both fixed.

src/js/common/utils/validations.js
```diff
@@ -174,7 +174,7 @@ function isValidPartialMonthYearInPast(month, year) {
   if (typeof month === 'object') {
     throw new Error('Pass a month and a year to function');
   }
-  const momentDate = moment({ year, month: month ? parseInt(month, 10) - 1 : null });
+  const momentDate = moment({ year, month: month ? parseInt(month, 10) : null });
```
cehsu: This is to handle January dates, which have a zero value.

src/js/common/utils/validations.js
```diff
@@ -153,7 +153,7 @@ function isValidPartialMonthYear(month, year) {
   if (typeof month === 'object') {
     throw new Error('Pass a month and a year to function');
   }
-  if (month && (Number(month) > 12 || Number(month) < 1)) {
+  if (month && (Number(month) > 12 || Number(month) < 0)) {
```
cehsu: This is to handle January dates, which have a zero value.

test/common/utils/validations.unit.spec.js
```diff
@@ -437,7 +437,7 @@ describe('Validations unit tests', () => {
       expect(isValidPartialMonthYearInPast('2', '2001')).to.be.true;
     });
     it('should validate month and year that is current', () => {
-      expect(isValidPartialMonthYearInPast(moment().month().toString(),
+      expect(isValidPartialMonthYearInPast(moment().add(1, 'month').month().toString(),
```
cehsu: This is to handle zero indexed moment months, as our validation is 1 indexed.

jbalboni:
1. I'm not sure 60 is right. We should check that, it might still be 30 right now. I think it is supposed to be 60, though.
2. Either way, I'm not sure this is worded accurately, since we remove applications after 60 days of inactivity. So you can start an application more than 60 days in the past, but you can't leave it unchanged for 60 days. @peggygannon 
3. This isn't pension specific, can we make this message generic? It's useful info for all forms.

peggygannon:
@jbalboni @cehsu this message looks good to me.

cehsu:
Thanks @jbalboni, it looks like 60 days is the [current expiration duration](https://github.com/department-of-veterans-affairs/vets-api/blob/1fd8c7fce32c2938f2a3cd9f61e734958d63b863/app/models/in_progress_form.rb#L11). I have remove "pension" from the message and applied it to all forms.

jbalboni (approved):

PR #7002, opened 2017-12-22
Letters error logging mg

markgreenburg:
- Adds `Raven.captureException` logging events for failed actions across the app
- Adds `Raven.captureException` log event to `LettersApp` component if it receives a false `isDataAvailable` prop from the `RequiredLoginView` wrapper. Not sure this is ideal since we may be hitting multiple renders and thus logging redundantly.
- Simplifies some view logic in `Main` component based on conversations with Content team
- Adds action for `INVALID_ADDRESS_PROPERTY` which has its own state in the reducer as well as its own view, but wasn't getting triggered anywhere in the app.

These logs should cover all instances of the UI showing an error message. Each captured error message should be specific enough to allow us to see where things went wrong but also general enough that we're not capturing any PII at all.

### DIFF
--- a/src/js/letters/actions/letters.js
+++ b/src/js/letters/actions/letters.js
@@ -60,7 +60,6 @@ export function getLetterList(dispatch) {
   );
 }
 
-
 export function getBenefitSummaryOptions(dispatch) {
   return apiRequest(
     '/v0/letters/beneficiary',
@@ -77,7 +76,6 @@ export function getBenefitSummaryOptions(dispatch) {
   );
 }
 
-
 // Call getLetterList then getBenefitSummaryOptions
 export function getLetterListAndBSLOptions() {
   return (dispatch) => {
@@ -86,7 +84,6 @@ export function getLetterListAndBSLOptions() {
       .catch((error) => Raven.captureException(error));
   };
 }
-
 
 export function getAddressFailure() {
   window.dataLayer.push({ event: 'letter-update-address-notfound' });
--- a/src/js/letters/actions/letters.js
+++ b/src/js/letters/actions/letters.js
@@ -106,7 +106,6 @@ export function getMailingAddress() {
     return apiRequest(
       '/v0/address',
       null,
-      // on fetch success
       (response) => {
         const responseCopy = Object.assign({}, response);
         // translate military address properties to generic properties for use in front end
@@ -116,8 +115,13 @@ export function getMailingAddress() {
           data: responseCopy
         });
       },
-      // catch errors in fetch or success handler
-      () => dispatch(getAddressFailure())
+      (response) => {
+        const status = (response.errors && response.errors.length)
+          ? response.errors[0].status
+          : 'unknown';
+        Raven.captureException(new Error(`vets_letters_error_getMailingAddress: ${status}`));
+        return dispatch(getAddressFailure());
+      }
     );
   };
 }
--- a/src/js/letters/actions/letters.js
+++ b/src/js/letters/actions/letters.js
@@ -24,49 +24,47 @@ import {
   UPDATE_BENFIT_SUMMARY_REQUEST_OPTION,
   SAVE_ADDRESS_PENDING,
   SAVE_ADDRESS_FAILURE,
-  SAVE_ADDRESS_SUCCESS
+  SAVE_ADDRESS_SUCCESS,
+  INVALID_ADDRESS_PROPERTY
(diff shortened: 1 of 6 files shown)


src/js/letters/containers/LettersApp.jsx
```diff
@@ -16,6 +17,7 @@ export function AppContent({ children, isDataAvailable }) {
         If you think you should be able to access this information, please call the Vets.gov Help Desk at <a href="tel:855-574-7286">1-855-574-7286</a>, TTY: <a href="tel:18008778339">1-800-877-8339</a>, Monday &#8211; Friday, 8:00 a.m. &#8211; 8:00 p.m. (ET).
       </h4>
     );
+    Raven.captureException(new Error('vets_letters_user_unregistered'));
```
markgreenburg: Would appreciate some 👀  and thought on this - I'm specifically concerned about potential to log multiple times as the component re-renders, but don't think this will be an issue because of the way the `isDataAvailable` prop is passed from `RequiredLoginView`

src/js/letters/containers/Main.jsx
```diff
@@ -56,36 +56,18 @@ export class Main extends React.Component {
...
-                  We weren’t able to retrieve your VA letters. Please call <a href="tel:855-574-7286">1-855-574-7286</a>,
-                  TTY: <a href="tel:18008778339">1-800-877-8339</a>, Monday &#8211; Friday, 8:00 a.m. &#8211; 8:00 p.m. (ET).
-                </p>
-              </div>
-            </div>
-            <br/>
-          </div>
-        );
-        break;
+      case unavailable: // fall-through to default
```
markgreenburg: Per chat with Peggy & Alex. Though as a follow-up we may want to tweak `SystemDown` error message globally to get it more in line with current content direction.

src/js/letters/actions/letters.js
```diff
@@ -37,67 +38,53 @@ export function getLetterList(dispatch) {
     }),
     (response) => {
       window.dataLayer.push({ event: 'letter-list-failure' });
-      if (typeof response.errors === 'undefined' || response.errors.length === 0) {
-        return Promise.reject(new Error('vets_letters_error_server_get: undefined error'));
+      const status = getStatus(response);
```
cvalarida: This whole block is cleaner. I like it! 💯

src/js/letters/actions/letters.js
```diff
@@ -37,67 +38,53 @@ export function getLetterList(dispatch) {
...
     null,
-    response => dispatch({
+    (response) => dispatch({
       type: GET_BENEFIT_SUMMARY_OPTIONS_SUCCESS,
       data: response,
     }),
-    () => {
+    (response) => {
       dispatch({ type: GET_BENEFIT_SUMMARY_OPTIONS_FAILURE });
-      // Is there a prize for the longest error name?
```
cvalarida: 😛

src/js/letters/actions/letters.js
```diff
@@ -37,67 +38,53 @@ export function getLetterList(dispatch) {
...
     }
   );
 }
 
-
 // Call getLetterList then getBenefitSummaryOptions
 export function getLetterListAndBSLOptions() {
   return (dispatch) => {
     return getLetterList(dispatch)
-      // Maybe shouldn't try to get BSO options if we get an error or we don't get a BSL...
```
cvalarida: Ahh, because of the less convoluted error handling, we don't fetch BSL options if `getLetterList` failed. Brilliant!

src/js/letters/actions/letters.js
```diff
@@ -118,8 +104,11 @@ export function getMailingAddress() {
           data: responseCopy
         });
       },
-      // catch errors in fetch or success handler
```
cvalarida: I guess the success handler doesn't throw any errors, does it? Good catch (pun unfortunately intended).

src/js/letters/containers/LettersApp.jsx
cvalarida: Agreed that this isn't an issue since it seems to only re-render upon page refresh.

cvalarida (approved):
Looks great! Solid refactor work. 👍 
Thanks for taking the time to clean this all up. I'm feeling a lot better about the overall error handling and logging in this app now. 😄

cvalarida (approved):
Changes look good. Maybe a bit overkill, but better safe than sorry, eh?

PR #7016, opened 2018-01-01
Landing page content accessibility fixes

cehsu:
These changes resolve two landing page content accessibility issues:
- ['Skip to Content' link lacks focus](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/6629)
screen shot 2018-01-01 at 3 10 43 pm

- [Feedback Link is Out of Order](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/6633)
In this case the link was removed. This link seemed to serve the purpose of allowing users to skip to later content, however, it was not provided with a visual expression (unlike the skip to content link at the top), so rather than providing it with a visual expression, as this seem like it would require design work, I have removed it here for now. If we would like to include a visual expression and skip to feedback link here, I will create a ticket for this work.


jbalboni:
@cehsu I'm not sure I understand how the skip to content link issue is fixed. It looks like we're adding padding and that's it?

src/sass/base/_va.scss
```diff
@@ -21,6 +22,10 @@ body {
 }
 
 // Skip link
+.show-on-focus-wrapper {
+  padding: 1em;
+}
```
ncksllvn: Looks like this padding is showing when at all times, instead of just when focused.

cehsu: Good catch @ncksllvn, fixed.

cehsu:
Nice catch, @ncksllvn, the content id has been added to the home layout in a manner consistent with that of the home-no-search layout.

cehsu:
Thanks @jbalboni, I've added a margin instead in light of your question.

jbalboni:
@cehsu Any idea why these links below the buttons aren't aligned anymore? It looks correct on staging/prod for me, but not on the review instance.

screen shot 2018-01-03 at 9 33 31 am

ncksllvn:
The homepage has a `va-homepage-benefits usa-content` class around those links. If you delete the `usa-content` class I think it aligns correctly.

image

cehsu:
Thanks for the fix @ncksllvn, kindly refer to this updated screenshot:
 screen shot 2018-01-03 at 11 33 21 am

ncksllvn (approved):
Perfect!

content/includes/footer.html
```diff
@@ -1,6 +1,5 @@
 <footer class="footer" role="contentinfo">
   <div class="footer-inner">
-    <a class="sr-only" href="#feedback-tool" name="feedback-tool">Give feedback on this page</a>
```
cehsu: @ncksllvn is this change consistent with your updates to the feedback tool and a reasonable way to resolve the link order issue that is referenced here?

jbalboni (approved):

content/includes/footer.html
ncksllvn: Yeah that looks good! That's consistent with the feedback tool. The only thing I had to watch out for was that the FAQ page contains this in its content -

` <a href="#feedback-tool" onclick="document.getElementById('feedback-tool').click()">Provide feedback on Vets.gov</a>.`

But this work isn't affected by that.

cehsu: Thanks for the heads up @ncksllvn.

ncksllvn: Thank you for your great work!

PR #7024, opened 2018-01-02
Wrap headings in legends, and options in fieldsets

cehsu:
These changes address the following issue: 
[Screenreader tabbing skips section title](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/6644)

Which indicates that both the fieldset labels (now located in the legends), and the option group labels (headings), are not captured by the screenreader. The first issue has now been address (see screenshot below), but I could not replicate the second.

image

Please note that the below screenshot (which depicts the updated version on the left and the original version on the right) is based on directly altering staging via the developer tools, as it is not possible to test this locally (in addition, as we occasionally notice discrepencies between the styling on staging and locally, once these are merged, the styling will be confirmed and adjusted as needed). Also, these changes alter the visual style slightly in an attempt to make the spacing beneath the section titles consistent.

image

Additional work that was identified while making these changes:
- Our pattern of including headings within legends was previously [not valid HTML](https://stackoverflow.com/questions/8005887/is-header-in-legend-valid-legendh1caption-h1-legend), even though it [has become valid](https://w3c.github.io/html/sec-forms.html#the-legend-element), it may be of interest to validate our HTML in our automated tests perhaps via a tool such as [valimate](https://github.com/jamesseanwright/valimate).
- An additional [instance](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/7116) identified is the Schemaform component's ObjectField component, which does not provide a legend or any headings for the fieldset, to wit there isn’t a section header in our forms in general. It might make sense to create one that reflects the form page information (e.g. “applicant information”), which is currently displayed in the FormNav component.


jbalboni (approved):
LGTM. I suspect we have more instances of this, too.

src/js/common/components/form-elements/ErrorableRadioButtons.jsx
```diff
@@ -127,10 +130,22 @@ class ErrorableRadioButtons extends React.Component {
       return output;
     });
 
+    const fieldsetClass = classNames({
+      'fieldset-input': true,
+      'usa-input-error': this.props.errorMessage,
+      [`${this.props.additionalFieldsetClass}`]: !!this.props.additionalFieldsetClass
+    });
```
ncksllvn: Ah, new JS features are awesome. Does this have to be written wrapped in a template string, or can it be `[this.props.additionalFieldsetClass]`?

ncksllvn (approved):
LGTM! Great work!!

src/js/common/components/form-elements/ErrorableRadioButtons.jsx
cehsu: Good catch, @ncksllvn, this has been simplified.

cehsu:
Thanks @jbalboni, I've scanned instances of components that include fieldsets and have created an [issue](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/7116) for another similar potential accessibility issue in the Schemaform component's ObjectField component.

PR #7041, opened 2018-01-04
Add role to validation errors

jbalboni:
This adds `role="alert"` to error messages so that they're read aloud. It also adds an "Error" prefix for screen readers to help with context.


cehsu (approved):
LGTM. There may be other instances, such as usages of "date-range-error".

cvalarida (approved):
How tedious. 😬 
Thanks for doing this! Looks great. 👍

jbalboni:
@cehsu It looks like date-range-error is used mostly in an aria-live region, and should be announced? I'm not really sure how to address it if not. That code seems odd.

cehsu:
Thanks @jbalboni, that's helpful context.

PR #7043, opened 2018-01-04
Wait for page to load before checking url

jbalboni:
I think this test is sometimes running the url assertion before the page actually changes. This changes makes it so we wait for an element on the next page to be visible before doing the assertion.

### DIFF
--- a/test/save-in-progress/01-sip-finish-later.e2e.spec.js
+++ b/test/save-in-progress/01-sip-finish-later.e2e.spec.js
@@ -17,12 +17,13 @@ module.exports = E2eHelpers.createE2eTest(
 
     client.axeCheck('.main');
 
+    E2eHelpers.overrideVetsGovApi(client);
+    E2eHelpers.overrideSmoothScrolling(client);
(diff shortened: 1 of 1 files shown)


ncksllvn (approved):
LGTM

cehsu (approved):

test/save-in-progress/01-sip-finish-later.e2e.spec.js
```diff
@@ -79,13 +79,15 @@ module.exports = E2eHelpers.createE2eTest(
       .click('.schemaform-sip-save-link');
     /* eslint-enable camelcase */
 
+    client.waitForElementVisible('.saved-form-metadata-container', Timeouts.normal);
     client.assert.urlContains('form-saved');
```
ncksllvn: Do you think it would it make any difference to change that to `Timeouts.slow`?

jbalboni: I can try

PR #7045, opened 2018-01-04
Remove vestigial mega-menu code

cvalarida:
So mega-menu.js isn't removed since it still has some nav-specific functionality--the nav menu behaves a little differently then "normal" menus when in a small window.

Specifically, it's still needed for:
- Hiding and revealing the nav menu on small windows
- Determining whether to show the menu / close button or the whole nav menu when the screen is resized

Since this is a pretty integral part of the site, we need to make sure this is tested thoroughly to make sure I didn't break anything. Please please please try to break it!

**Known weird functionality**
- When opening a menu by clicking on it, the focus goes to the first item in the newly-opened menu instead of staying on the menu button just clicked
  - If this is undesired, we can fix it pretty easily; it's just a bit of a code smell
- When a first-level menu is opened (Explore benefits, for example), hitting the up arrow when the focus is on "Explore benefits" opens the last sub-menu
  - This is fine and expected when the menu isn't already opened, and the up arrow is pressed, but maybe a bit awkward when the menu is already opened
- Cycling through the menu options in a first-level menu with the down arrow does _not_ focus on the opening menu button ("Explore benefits"), but cycling through with the up arrow _does_
  - Again, we can make this happen, but it's just a bit awkward

Find anything else? Make a comment!

### DIFF
--- a/src/js/common/utils/accessible-menus.js
+++ b/src/js/common/utils/accessible-menus.js
@@ -362,8 +362,7 @@ export default function addMenuListeners(menuElement, closeOnResize = false) {
   });
 
   // Close when the screen is resized
-  // Useful for when the menu might be hidden on smaller screens
-  // TODO: Actually use this
+  // Useful for when the menu might be hidden on smaller screens (nav menubar)
   if (closeOnResize) {
     window.addEventListener('resize', () => closeAll(menuElement));
   }
@@ -380,8 +379,6 @@ export default function addMenuListeners(menuElement, closeOnResize = false) {
     } else if (inMenu) {
       openMenu(targetLi);
     }
-
-    // TODO: Handle open and close menu button for small windows
   });
 
   // Handle clicking away from the menu
@@ -389,8 +386,6 @@ export default function addMenuListeners(menuElement, closeOnResize = false) {
     const target = event.target;
     if (!menuElement.contains(target)) {
       closeAll(menuElement);
-      // TODO: Handle the open / close menu button for small windows
-      // this.resetMenu();
     }
   });
 }
--- a/src/js/common/utils/accessible-menus.js
+++ b/src/js/common/utils/accessible-menus.js
@@ -150,9 +150,9 @@ function getMenuStructure(menuLi) {
  *
  * @param {HTMLLIElement} menuLI  The <li> containing the menubutton and menu
  */
-function closeMenu(menuLi) {
-  // If we're not dealing with a menu structure, abort
-  const struct = getMenuStructure(menuLi);
+function closeMenu(menuLiOrStruct) {
+  // Make sure we've got a menu
+  const struct = menuLiOrStruct.menuButton ? menuLiOrStruct : getMenuStructure(menuLiOrStruct);
   if (!struct) {
     return;
   }
@@ -173,9 +173,9 @@ function closeMenu(menuLi) {
  *                                 second-level menu in a wide screen, we want the third-level opened as well
  * @param {bool} stealFocus       Whether or not to focus on the first item in the opened menu
  */
-function openMenu(menuLi, openSubMenu = false, stealFocus = true) {
+function openMenu(menuLi, menuStructure = null, openSubMenu = false, stealFocus = true) {
   // If we're not dealing with a menu structure, abort
-  const struct = getMenuStructure(menuLi);
+  const struct = menuStructure || getMenuStructure(menuLi);
   if (!struct) {
     return;
   }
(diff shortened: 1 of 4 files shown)


cehsu:
It looks like there are a few behaviors mentioned in the Keyboard Support section [here](https://w3c.github.io/aria-practices/examples/menubar/menubar-1/menubar-1.html) that might be helpful to incorporate.

Also, should tabbing navigate across the nav menu rather than through it?

cehsu:
Is addressing [this](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/6630) issue of interest here?

cvalarida:
@cehsu Those are great points. I think it makes sense to address them in separate PRs, though. can you write up another ticket to look into the `menuitem` bit? I think I put those roles in there first, but the screen reader read it off weirdly...I could be misremembering things, though. Could be that's talking about the sub-menu menu items. :man_shrugging: 

At any rate, this PR is just focused on eliminating unnecessary code in `mega-menu.js` and transferring the functionality to `accessible-menus.js`, so I think those changes are out of scope for this this one.

cehsu:
Sounds good @cvalarida, ticket created [here](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/7141).

src/js/common/utils/accessible-menus.js
```diff
@@ -173,9 +173,9 @@ function closeMenu(menuLi) {
  *                                 second-level menu in a wide screen, we want the third-level opened as well
  * @param {bool} stealFocus       Whether or not to focus on the first item in the opened menu
  */
-function openMenu(menuLi, openSubMenu = false, stealFocus = true) {
+function openMenu(menuLi, menuStructure = null, openSubMenu = false, stealFocus = true) {
   // If we're not dealing with a menu structure, abort
-  const struct = getMenuStructure(menuLi);
+  const struct = menuStructure || getMenuStructure(menuLi);
```
cehsu: Is it of interest to make the logic and and on line 155 the same, they appear to do the same thing differently, or even to pull this logic out into a helper and use in both places?

src/js/common/utils/accessible-menus.js
```diff
@@ -340,10 +361,32 @@ export default function addMenuListeners(menuElement) {
...
+      toggleMenu(targetLi);
+    } else if (inMenu) {
+      openMenu(targetLi);
+    }
+  });
+
+  // Handle clicking away from the menu
+  document.addEventListener('click', event => {
+    const target = event.target;
+    if (!menuElement.contains(target)) {
```
cehsu: This line got me thinking about how we might want to allow users to open menus via hover, also if the user clicks on a part of the menu that is not a li, the menu closes, which is an edge case but which doesn't seem like expected behavior. Shall we include these in the enhancements ticket?

cehsu (approved):
LGTM. I had a few minor question/comments. Also, is line 186 an opportunity to make use of `closeAll`?

src/js/common/utils/accessible-menus.js
cvalarida: Yeah, I oscillated on this one, but decided to leave `closeMenu` (hopefully) simpler with just one param. I gave `openMenu` both `menuLi` and `menuStructure = null` because we actually use `menuLi` for more than just getting the structure of the menu down on line 186. In `closeMenu`, it just needs the structure, not the `menuLi`.

Hopefully that made sense, but if not, we I can try to clarify some more.

I'm having a hard time with words this morning. 😞

src/js/common/utils/accessible-menus.js
```diff
@@ -340,10 +361,32 @@ export default function addMenuListeners(menuElement) {
...
 
-  // TODO: Implement this and get rid of MegaMenu usage
-  // menuElement.addEventListener('click', (e) => {
-  //   // Handle opening menus
-  //   // Handle closing menus when clicked away
-  // });
+  // Close when the screen is resized
+  // Useful for when the menu might be hidden on smaller screens (nav menubar)
+  if (closeOnResize) {
+    window.addEventListener('resize', () => closeAll(menuElement));
```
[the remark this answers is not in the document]
cvalarida: Y'know, I'm not sure about accessibility considerations, actually. I did this just because that's the way it was working before and I wanted to replicate all the former behavior.

I think this functionality was for when the menu is resized to small screen (with the "Menu" button) where the menubar is hidden, then resized back up to where it's shown again. It's unexpected to have the menu stay open.

We can definitely revisit this later to make sure it's still a good idea, but for now, I was just replicating functionality that's already there. :man_shrugging:

src/js/common/utils/accessible-menus.js
cvalarida: I don't now if opening menus via hover is the desired functionality (enter UX researchers!), but if it is, we could give it a shot. Probably should write a ticket to do some discovery on that.

For clicking on a part of the menu that isn't an `li`, do you mean a menu item that isn't in an `li`, or like the space between / around / above / below the menu items but still "in" the menu?

PR #7049, opened 2018-01-04
Replace react-autosuggest with downshift

jbalboni:
This seems to work much better and you can actually choose options in VoiceOver now.

screen shot 2018-01-04 at 4 50 42 pm


cehsu (approved):

rlbaker (approved):

src/js/common/schemaform/fields/AutosuggestField.jsx
```diff
@@ -156,24 +148,46 @@ export default class AutosuggestField extends React.Component {
...
+          getInputProps,
+          getItemProps,
+          isOpen,
+          selectedItem,
+          highlightedIndex
+        }) => (
+          <div className="autosuggest-container">
+            <input {...getInputProps({ id, name: id, onKeyDown: this.handleKeyDown, onBlur: this.handleBlur })}/>
+            {isOpen && (
+              <div className="autosuggest-list">
```
cehsu: Running aXe coconut indicates that this div should be given the listbox role:
>Certain ARIA roles must contain particular children
Issue description
Ensures elements with an ARIA role that require child roles contain them

>Impact: critical
 Learn more
Element location
#root_application_veteran_serviceRecords_0_serviceBranch
Element source
<input role="combobox" aria-autocomplete="list" aria-expanded="false" autocomplete="off" value="army" id="root_application_veteran_serviceRecords_0_serviceBranch" name="root_application_veteran_serviceRecords_0_serviceBranch">

>To solve this violation, you need to:
Fix the following:
Required ARIA child role not present: listbox

src/sass/modules/_m-schemaform.scss
```diff
@@ -508,11 +508,11 @@ legend.schemaform-label.schemaform-file-label {
   margin-top: 0;
 }
 
-.react-autosuggest__container {
+.autosuggest-container {
```
cehsu: It looks like there might be a styling issue on mobile:
screen shot 2018-01-04 at 4 12 36 pm

jbalboni:
@cehsu Both those issues should be fixed now.

bshyong (approved):
nice!

PR #7051, opened 2018-01-05
Main nav keyboard navigation e2e tests

cvalarida:
It's written like unit tests, but is _actually_ a browser (e2e) test. I think it's a good phase 2 for these kinds of tests.

I wrote a little `.focusOn()` function that I used to help reset the state so I could write the tests more like unit tests. `.focusOn(selector)` in conjunction with `.keys()` instead of `.sendKeys(selector)` I think will also more closely follow what happens in the browser. For example, if the result of `.sendKeys()` moves the focus to a spot that the next `.sendKeys(selector)` didn't anticipate, that undesired behavior will just get passed over because `.sendKeys(selector)` doesn't care where the focus currently is (from my understanding).

Moving forward, I think it might make sense to encapsulate the individual functionality tests a bit more next time (e.g. hitting the down arrow in a certain context). Unfortunately while breaking them out into helper functions I think may succeed in increasing readability, it'll also open up for bugs where the state wasn't reset properly. Not sure what the best way to do that is (with the least overhead). I have some ideas, but they seem kinda clunky.

### DIFF
--- a/test/common/utils/accessible-menus.e2e.spec.js
+++ b/test/common/utils/accessible-menus.e2e.spec.js
@@ -1,0 +1,36 @@
+const E2eHelpers = require('../../e2e/e2e-helpers');
+const Timeouts = require('../../e2e/timeouts.js');
+
+const exploreButton = '#vetnav-menu button[aria-controls="vetnav-explore"]';
+
+module.exports = E2eHelpers.createE2eTest(
+  (client) => {
+    client
+      .url(`${E2eHelpers.baseUrl}/`);
+
+    E2eHelpers.overrideSmoothScrolling(client);
+
+    // landing page
+    client
+      .waitForElementVisible('body', Timeouts.normal)
+      .waitForElementVisible('#vetnav', Timeouts.slow)
+      // do not run 'wcag2a' rules because of open aXe bug https://github.com/dequelabs/axe-core/issues/214
+      .axeCheck('.main');
+
+
+    // --------------------- //
+    // --- Menubar tests --- //
+    // --------------------- //
+
+
+    // Down arrow should open menu
+    client
+      .focusOn(exploreButton)
+      .assert.isActiveElement(exploreButton);
+
+    // Up arrow should
+
+
+    client.end();
+  });
+
--- a/test/common/utils/accessible-menus.e2e.spec.js
+++ b/test/common/utils/accessible-menus.e2e.spec.js
@@ -2,9 +2,13 @@ const E2eHelpers = require('../../e2e/e2e-helpers');
 const Timeouts = require('../../e2e/timeouts.js');
 
 const exploreButton = '#vetnav-menu button[aria-controls="vetnav-explore"]';
+const benefitsButton = '#vetnav-menu button[aria-controls="vetnav-benefits"]';
+const facilitiesLink = '#vetnav-menu a[href="/facilities/"]';
 
 module.exports = E2eHelpers.createE2eTest(
   (client) => {
+    const { SPACE, ENTER, UP_ARROW, DOWN_ARROW, LEFT_ARROW, RIGHT_ARROW, ESCAPE } = client.Keys;
+
     client
       .url(`${E2eHelpers.baseUrl}/`);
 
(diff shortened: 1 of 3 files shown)


test/util/nightwatch-commands/selectDropdown.js
```diff
@@ -20,7 +20,7 @@ exports.command = function selectDropdown(name, value) {
     element.value = clientValue;
     element.dispatchEvent(evt);
     return element.value;
-    /* eslint-disable */
+    /* eslint-enable */
```
cvalarida: Has nothing to do with this PR, but I noticed it probably wasn't supposed to be disabled twice. :man_shrugging:

test/util/nightwatch-commands/focusOn.js
```diff
@@ -0,0 +1,14 @@
+/**
+ * Change select value and trigger change event programatically. This
+ * is necessary because long select boxes tend to render offscreen,
+ * causing Selenium to fail in unexpected ways.
+ * The first parameter is the field name, not the whole selector.
```
cehsu: Should this specify that the first parameter is the whole selector?

test/common/utils/accessible-menus.e2e.spec.js
```diff
@@ -0,0 +1,228 @@
...
+    // --------------------- //
+    // --- Menubar tests --- //
+    // --------------------- //
+
+    // Explore benefits menu items
+    const firstMenuItem = 'button[aria-controls="vetnav-disability"]';
+    const secondMenuItem = 'button[aria-controls="vetnav-healthcare"]';
+    const lastMenuItem = 'button[aria-controls="vetnav-burials"]';
+
+    const testOpenExploreMenu = (key, focusedOn, keyName) => {
```
cehsu: Would it make sense to move this to a helper file?

test/common/utils/accessible-menus.e2e.spec.js
```diff
@@ -0,0 +1,228 @@
...
+    // Space should follow links
+    // Enter should follow links
+
+
+    // --------------------------- //
+    // --- Miscellaneous tests --- //
+    // --------------------------- //
+
+    // Escape should close all menus
+    //  Should focus be doing anything special here? When hitting tab, it goes to the next menubar item
```
cehsu: It looks like focus should return to the toggle element/menubar item for the closed submenu.

test/common/utils/accessible-menus.e2e.spec.js
```diff
@@ -0,0 +1,228 @@
...
+      .assert.isActiveElement(hcFirstItem);
+
+    // Left arrow should move focus back to the opening menu button
+    client
+      .focusOn(hcFirstItem)
+      .keys(LEFT_ARROW)
+      .assert.isActiveElement(secondMenuItem);
+
+    // TODO: Write these; I'm not sure what to do with them
+    //  Do we just poke them and test the url? Seems like it'd get pretty slow
```
cehsu: It makes sense not to test this here, and to emphasize focus management and whether things are opening/closing as expected.

cehsu (approved):
LGTM so far. Had a few minor questions/comments.

By phase two, do you mean using e2e tests for unit test style accessibility testing? It seems reasonable/convenient to have these together if so.

test/common/utils/accessible-menus.e2e.spec.js
cvalarida: We could, but I think it'll only be used in this one test, so it seemed a bit extreme to pull it out into a helper file.

test/util/nightwatch-commands/focusOn.js
cvalarida: Heh, woops! That comment is from another command that I just copied and gutted. I need to rewrite the whole thing. Good catch!

test/common/utils/accessible-menus.e2e.spec.js
cvalarida: Ahh, okay. I'll write up another ticket to address that so I don't clutter up this PR. 👍

cvalarida: https://github.com/department-of-veterans-affairs/vets.gov-team/issues/7149

test/common/utils/accessible-menus.e2e.spec.js
cvalarida: 🤔  Yeah, I guess this is the same as clicking, which is outside the scope of keyboard navigation, isn't it?

What do y'all think?

cvalarida:
@cehsu Yeah, by phase 2, I mean the wannabe-unit-test-style test writing where we sorta reset the state between testing different functionalities.

(I meant to say this the other day but totally forgot to write it down. Sorry for the delay!)

PR #7055, opened 2018-01-05
Use minimum bounding box for facility searches

bshyong:
Some zip codes are extremely small, particularly in the DC area.

image

This PR sets a minimum bounding box so that even small zip codes will result in a search over a larger area.

Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4279


U-DON (approved):
LGTM.

PR #7056, opened 2018-01-05
Construct "Available Services" section in profile based on services list from API

U-DON:
Resolves department-of-veterans-affairs/vets.gov-team#6805.

We build a list of services that are accessible when serializing the object returned for `/v0/user`, and that should be the source of truth when deciding whether a service is available to the user.

LOA should not need to be checked on the front end and should already be accounted for when serializing the user object.

The specific ticket being fixed is a case where we still want to display Rx for a certain sign-in method that grants LOA1. Our previous condition based on LOA would say it requires identity proofing. But in this specific case, we already include `'rx'` in that `services` list built for this user, so it should be accessible without identity proofing based on that list.

> screen shot 2018-01-05 at 3 45 36 pm


src/js/user-profile/components/AuthApplicationSection.jsx
```diff
@@ -1,60 +1,103 @@
...
-          <p><a href="/health-care/apply/">Apply for health care</a></p>
-          <p><a href="/education/apply-for-education-benefits">Apply for education benefits</a></p>
-          <p><a href="/discharge-upgrade-instructions/">Apply for a discharge upgrade if needed</a></p>
-          <p><span className="label">You need to <a href="/verify?next=/profile">verify your identity</a> in order to:</span></p>
-          <p>Refill your prescription</p>
-          <p>Message your health care team</p>
-          <p>Get your VA health records</p>
-          <p>Check your claim and appeal status</p>
-          <p>Check your Post-9/11 GI Bill statement of benefits</p>
-          <p>Apply for a Veteran ID Card</p>
```
U-DON: @patrickvinograd: Since the logic for showing this in the list is a little different now, just wanted to check if you thought it makes sense to display "Apply for a Veteran ID Card" under the verify list when `id-card` is not in the list of services in the user object.

Before, it wouldn't show up in that list as long as the user is LOA3, but the test user I was using was LOA3 and didn't have `id-card` in the service list, so it appears as shown in the screenshot in the description.

src/js/user-profile/components/AuthApplicationSection.jsx
```diff
@@ -1,60 +1,103 @@
...
+            )
+          }
+          {availableServices}
+          {
+            !!unavailableServices.length && (
+              <p><span className="label">You need to <a href="/verify?next=/profile">verify your identity</a> in order to:</span></p>
+            )
+          }
+          {unavailableServices}
           <p><strong>Note:</strong> If you have a less than honorable discharge, you can apply for an upgrade. If your
```
U-DON: @ncksllvn || @bshyong: Since there have been a few changes to this page from Discharge Wizard, I wanted to check if I resolved the merge conflicts properly, and everything is in the right place.

U-DON:
Previously, we would display all of those services in that list as long as the user was LOA3, but it is possible to have an LOA3 user who does not have one of those services. The screenshot demonstrates the case where an LOA3 user (user+264) does not have access to VIC according to the `services` list.

With that being the case, I wonder if it's still accurate messaging to say that you need to verify your account to access a service that is unavailable to you even as an LOA3 user.

src/js/user-profile/components/AuthApplicationSection.jsx
```diff
@@ -1,60 +1,103 @@
...
-    let content;
-    const verifiedAccountType = 3;// verified ID.me accounts are type 3
-    const isVerifiedUser = this.props.userProfile.accountType === verifiedAccountType;
-    const idCardAvailable = this.props.userProfile.services.indexOf('id-card') !== -1;
-    if (isVerifiedUser) {
-      content = (
+    const { services } = this.props.userProfile;
+    const availableServices = [];
+    const unavailableServices = [];
+    const isAvailable = services.reduce((acc, service) => {
```
bshyong: Is there a better name that we can use for this so that it's easier to understand difference between availableServices vs. isAvailable?

bshyong (approved):

bshyong:
@U-DON regarding the messaging, we should bring that up so that it can be changed. It will be frustrating for users who Verify their account but still don't have access, resulting in more help desk tickets or calls.

src/js/user-profile/components/AuthApplicationSection.jsx
patrickvinograd: Right now the frontend doesn't have enough information to make the claim it's making - because LOA3 vs. not is not the only reason why `id-card` doesn't show up in the services list. It could also be because they are ineligible based on title38 status. 

This is true for other services as well - but less likely to happen because most of the conditions for other services are more straightforward - e.g. "loa3 and has ssn". 

I think it would be more accurate to say something like "Verifying your account may give you access to more services:" with a list, and maybe additionally a tool tip or something describing what the requirements are for each unavailable service. That would be a decision to be made by UX and/or PM though.

saneshark: I think the context of this view is meant to be precisely that. By identity proofing you could potentially use these additional features, but not that you absolutely can. If we need to change the language of the "Available Services" section to more accurately reflect that I think we should. 

But keeping the logic behind this on the front-end somewhat streamlined makes the most sense to me, especially since a lot of those other attributes are only ever available once we're able to query MVI (LOA3 only).

saneshark:
@melwoodard @DanielleSoCompany I wonder if either of you have any thoughts on how we could more effectively communicate what services "may or may not" be available after identity proofing. Without identity proofing, it's impossibly to convey to the user definitively what services will be available. I think there is some room to improve the language here.

DanielleSoCompany:
@saneshark that makes sense. Running out now, but will think on it a bit and post something for consideration on Monday. thanks!

DanielleSoCompany:
@saneshark @melwoodard Does this work? 

**When you verify your identity, you may have access to more services. These may include:**

An alternate might be: **Verify your identity to access more services you may be eligible for, like:**

Happy to tweak more based on feedback. Thanks!

Also, Patrick had mentioned including notes re: eligibility for the different services. Do we need language around that? If yes, can you let me know the specific services that need notes? Thanks.

melwoodard:
Thanks Danielle! I like your second option. I'm inclined to hold on surfacing eligibility info for now and keep this page as simple as we can. Our energy would be better spent on a deeper dive on a profile revamp, which is hopefully not so far away.

src/js/user-profile/components/AuthApplicationSection.jsx
```diff
@@ -1,60 +1,111 @@
...
```
jbalboni: This might be clearer if you convert it into a Set instead of an object:

`const isAvailable = new Set(services);`

Then below you can do `isAvailable.has(service)`. And like Ben says above, coming up with a different name than `isAvailable` would help, too.

src/js/user-profile/components/AuthApplicationSection.jsx
```diff
@@ -1,60 +1,111 @@
...
+
+    contentConditions.forEach(([content, requiredServices]) => {
+      // Services without any requirements are available to every user.
+      if (!requiredServices) {
+        availableServices.push(content);
+        return;
+      }
+
+      const accessible = requiredServices.length > 1 ?
+        requiredServices.reduce((acc, service) => acc && isAvailable[service], true) :
```
jbalboni: Might be clearer if this was written as `requiredServices.every(service => isAvailable[service])`

jbalboni (approved):
Had some very minor comments, but LGTM.

U-DON:
Updated copy.

screen shot 2018-01-08 at 8 33 06 pm

PR #7058, opened 2018-01-05
Letters progress updates mg

markgreenburg:
Tweaks a few bits of Letters styling:
- Aligns schemaform content with the `StepHeader` (progress bar description) on the left.
- Makes progress bar same width as content
- Standardizes right padding between progress bar, schemaform (address / letters list) content, and the intro text paragraph. This includes a media query for small screens
- Simplifies some of the markup by switching to semantic tags and removing unnecessary wrapping divs.

**Before:**
34581376-c9ee1eda-f155-11e7-8c4c-66478b9b9a25

**After:**
34628572-689e3eb4-f22a-11e7-9476-83f7a0c82ca6

### DIFF
--- a/src/sass/letters.scss
+++ b/src/sass/letters.scss
@@ -21,14 +21,6 @@
     }
   }
 
-  .step-content {
-    margin-left: 2em;
-  }
-
-  .letters-progress-bar {
-    width: 6em;
-  }
-
   .address-block {
     background-color: $color-gray-lightest;
     padding: 1em;
--- a/src/sass/letters.scss
+++ b/src/sass/letters.scss
@@ -1,7 +1,7 @@
-@import "shared-variables";
-@import "modules/m-form-process";
-@import "modules/m-progress-bar";
-@import "modules/m-schemaform";
+@import 'shared-variables';
+@import 'modules/m-form-process';
+@import 'modules/m-progress-bar';
+@import 'modules/m-schemaform';
 @mixin schemaform-padding {
   // query borrowed from _m-schemaform.scss
   @media(max-width: 40.063em) {
@@ -19,8 +19,8 @@
   }
 
   .feature {
-      min-height: 0;
-      margin-top: 1em;
+    margin-top: 1em;
+    min-height: 0;
 
     &.help-desk {
       margin-top: 2em;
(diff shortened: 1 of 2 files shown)


src/sass/letters.scss
```diff
@@ -1,7 +1,13 @@
-@import "shared-variables";
-@import "modules/m-form-process";
-@import "modules/m-progress-bar";
-@import "modules/m-schemaform";
+@import 'shared-variables';
+@import 'modules/m-form-process';
+@import 'modules/m-progress-bar';
+@import 'modules/m-schemaform';
+@mixin schemaform-padding {
```
cvalarida: Oh bother. We're not using schemaform in this app, so it probably doesn't make sense to import the css. I'll write up another ticket for that.

As far as this goes, it probably makes sense to rename this one to not include `schemaform-`.

src/sass/letters.scss
```diff
@@ -67,6 +65,15 @@
     }
   }
 
+  .va-introtext {
```
cvalarida: There's a lot of re-organizing and reformatting--is chunk all that's new in this PR? `.va-introtext` and `.progress-bar-segmented` with the use of `schemaform-padding`.

cvalarida (commented):
It looks alright, but can you add a screenshot? Before and after would be 💯 . Reading the diff on this one is kind of a pain.

src/sass/letters.scss
markgreenburg: Good catch. We are using the `.schemaform-chapter-progress` class from `_m-schemaform.scss`, though. This is the wrapper that holds most of the app content, from the `StepHeader` on down. I named it this way to make it obvious that this style is basically there to align with what's defined in `_m-schemaform.scss`, but am definitely open to better suggestions.

src/sass/letters.scss
markgreenburg: Er yeah, sorry. Linter didn't like the way things were organized so I just re-shuffled some stuff that should have probably just been done in a separate PR. The only substantive changes here are the removal of `.step-content`, `.letters-progress-bar`, and addition of the media-query mixin to `.progress-bar-segmented` as well as `.va-intro-text`

markgreenburg:
@cvalarida yeah, images were in the linked issue but I should have added them here as well. Updated the original description to include before / after.

src/sass/letters.scss
cvalarida: https://github.com/department-of-veterans-affairs/vets.gov-team/issues/7150

cvalarida (approved):
Looking at the screenshots and reading @goldenmeanie's input, I think it's great! (Pending Alex's approval on whether the screenshot looks right, of course.)

PR #7065, opened 2018-01-08
Attempting to fix timeout during E2E tests

ncksllvn:


test/discharge-wizard/00-required.e2e.spec.js
```diff
@@ -42,12 +42,11 @@ module.exports = E2eHelpers.createE2eTest(
...
 
-    client.pause(Timeouts.verySlow);
-
     // results page
     client
       .click('a.usa-button-primary')
-      .waitForElementVisible('.dw-guidance', Timeouts.verySlow)
+      .waitForElementVisible('.dw-guidance', Timeouts.slow)
+      .pause(Timeouts.verySlow)
       .axeCheck('.main');
```
ncksllvn: There's a PDF download occurring here to `'http://arba.army.pentagon.mil/documents/dd0293.pdf'`. It's too fast during local testing to register, but I think Jenkin's is stumbling on it.

bshyong (approved):
LGTM

test/discharge-wizard/00-required.e2e.spec.js
```diff
@@ -42,12 +42,9 @@ module.exports = E2eHelpers.createE2eTest(
       .axeCheck('.main')
       .click('a.usa-button-primary');
 
-    client.pause(Timeouts.verySlow);
-
     // results page
     client
-      .click('a.usa-button-primary')
-      .waitForElementVisible('.dw-guidance', Timeouts.verySlow)
+      .waitForElementVisible('.dw-guidance', Timeouts.slow)
```
jbalboni: It seems like this removes a button click. Is that intentional?

jbalboni (approved):
LGTM, had one question.

test/discharge-wizard/00-required.e2e.spec.js
ncksllvn: Yeah it does. That click was unintentionally in there twice and the second click was causing an error.

PR #7067, opened 2018-01-09
Nav accessibility small

cehsu:
These changes allow users to enter (via ENTER or SPACE), toggle (via ESCAPE), and exit (via TAB on last menu item, and SHIFT+TAB on the first menu item) the nav menu on small screens.

The idea was to make the change as small as possible as this is a critical issue. Eventually, it may make sense to provide `addMenuListeners` with access to the menu controls, and to relocate the accessibility logic in mega-menu.js to accessible-menus.js, as part of the additional enhancement work being planned in nightwatch issue 7141.
  
Also, in order to test this, it looks like we will need to replace electron with chromedriver, to overcome a potential bug with  resizeWindow (nightwatch issue-856 comment-203115764). This change will also involve updating the activeElement check/assert to make use of  .elementActive and .elementIdAttribute instead.


jbalboni:
@cehsu It seems confusing to close the menu when you tab after the last menu item. You have to be very careful to realize that you're on the last menu item and that if you tab past it, you can't go back unless you reverse tab all the way back to the menu button (through items that you didn't tab through in the first place).

The issue mentioned this acting a lot like a modal and in the modal we put focus back at the top when you tab to the end of the content in it.

cehsu:
Thanks @jbalboni changes have been added to trap focus in a manner consistent with the standard behavior for full screen menus, by sending focus back to the initial tabbable item after going through the menu.

jbalboni:
@cehsu That seems much better to me. I'm noticing that if I shift+tab from a menu item that I tab through the sign in/up links (though I don't hit them going forward).

config/nightwatch.js
```diff
@@ -38,7 +37,6 @@ module.exports = {
         acceptSslCerts: true,
         webStorageEnabled: true,
         chromeOptions: {
-          binary: electron,
```
jbalboni: Removing this makes the e2e tests start up Chrome instead of Electron.

src/js/legacy/mega-menu.js
```diff
@@ -30,6 +38,42 @@ class MegaMenu {
...
+    }
+  }
+
+  exitSmallMegaMenu(e) {
+    if (!isWideScreen() && isTab(e) && !e.shiftKey) {
+      this.lastTabbableElement.focus();
+    }
+  }
+
   showMegaMenu() {
```
jbalboni: Is this duplicate code now?

src/js/legacy/mega-menu.js
```diff
@@ -30,6 +38,42 @@ class MegaMenu {
...
+
+  toggleSmallMegaMenu(e) {
+    if (!isWideScreen() && isEscape(e)) {
+      this.hideMegaMenu();
+      this.openControl.focus();
+    }
+  }
+
+  enterSmallMegaMenu(e) {
+    if (!isWideScreen() && isTab(e) && !e.shiftKey) {
```
jbalboni: Does this work on a screenreader? Do they fire key events to change focus?

config/nightwatch.js
cehsu: Thanks @jbalboni, as mentioned in the PR description, there appears to be a bug when using electron related to resizeWindow, should chromedriver be installed/configured instead, or should these tests be put on the back burner for now, or is another solution preferred?

src/js/legacy/mega-menu.js
cehsu: These are both (showMegaMenu/hideMegaMenu) the result of a bad merge, and will be removed.

src/js/legacy/mega-menu.js
cehsu: Focus changes as expected using a screenreader, but there are a few discrepancies with Safari that need to be resolved as well (the final menu item doesn't get reached, and is skipped in the tab order), good catch.

config/nightwatch.js
cvalarida: Without doing any kind of digging into this, I think this is _probably_ fine, but I'd like a larger discussion before changing the browser we're running all e2e tests on.

cehsu: Sounds good @cvalarida, I'll remove the test related changes from the PR.

src/js/legacy/mega-menu.js
cehsu: It looks like the cross-browser issues encountered may be beyond the scope of these changes, more information available [here](https://developer.mozilla.org/en-US/docs/Learn/Tools_and_testing/Cross_browser_testing/Accessibility#Using_native_keyboard_accessibility).

src/js/legacy/mega-menu.js
```diff
@@ -50,8 +90,9 @@ function initNavMenu() {
   const menuElement = document.querySelector('#vetnav');
   const openMenuElement = document.querySelector('.vetnav-controller-open');
   const closeMenuElement = document.querySelector('.vetnav-controller-close');
+  const menuElements = document.querySelectorAll('#vetnav-menu li');
```
jbalboni: Instead of adding this here, you could query for it in `menuElement` in the constructor for MegaMenu.

src/js/legacy/mega-menu.js
```diff
@@ -42,6 +54,34 @@ class MegaMenu {
...
+      this.firstMenuElement.focus();
+    }
+  }
+
+  exitSmallMegaMenu(e) {
+    if(e.target === this.firstMenuElement){
+      if (!isWideScreen() && isReverseTab(e)) {
+        e.preventDefault();
+        this.closeControl.focus();
+      []}
```
jbalboni: Typo?

src/js/legacy/mega-menu.js
```diff
@@ -1,23 +1,32 @@
...
     this.addListeners();
   }
 
   addListeners() {
     this.closeControl.addEventListener('click', this.hideMegaMenu);
+    this.closeControl.addEventListener('keydown', this.enterSmallMegaMenu);
     this.openControl.addEventListener('click', this.showMegaMenu);
+    this.firstMenuElement.addEventListener('keydown', this.exitSmallMegaMenu);
+    this.lastMenuElement.addEventListener('keydown', this.exitSmallMegaMenu);
+    this.menuElements.forEach(item => item.addEventListener('keydown', this.toggleSmallMegaMenu));
```
jbalboni: What do you think about having one event listener on `#vetnav-menu` and checking that the target is an `li`?

jbalboni (commented):
Made a couple comments. Also, when I tab past the last element, I get put on the address bar of the browser. Is that what you intend to happen?

src/js/legacy/mega-menu.js
cehsu: Indeed, thanks for the catch, fixed.

src/js/legacy/mega-menu.js
cehsu: That's definitely a better idea, thanks @jbalboni, changed.

src/js/legacy/mega-menu.js
cehsu: Thanks @jbalboni, sounds good, changed.

cehsu:
Thanks @jbalboni, I've responded to your comments in line. As for the navigation question, this appears to be consistent with how focus trapping menus generally behave and is intended to match user expectations. But I invite @1Copenut and @stacyhlinn to weigh in on this as well.

src/js/common/utils/accessible-modal.js
```diff
@@ -0,0 +1,30 @@
+import { isTab, isReverseTab, getTabbableElements } from './accessibility-helpers';
+
+/*
+ * Creates function that captures Veterans Crisis Line modal focus.
+ */
+
+document.addEventListener('DOMContentLoaded', () => {
```
jbalboni: Can we change the name of this file to make it more clear that this is for the VCL?

jbalboni (approved):

src/js/common/utils/accessible-modal.js
cehsu: Good idea, changed.

PR #7068, opened 2018-01-09
Appeals Status V2 -- Need help sidebar

cvalarida:
Seems a little high on the wide-screen version, but I'll let @gnakm decide if we want to bring it down or not.

Also, I didn't follow the design exactly; I used the styling from the claims list page. If you want more what you had in the mockup, @gnakm, let me know and I can make it happen.

## Appeals status
**Widescreen**
image

**Small screen**
image

## Claims and appeals index
The changes also affected the sidebar on this page, so I fiddled around with it to make it work well in both places.

**Widescreen**
image

**Small screen**
image

### DIFF
--- a/src/js/claims-status/components/appeals-v2/AppealHelpSidebar.jsx
+++ b/src/js/claims-status/components/appeals-v2/AppealHelpSidebar.jsx
@@ -36,7 +36,8 @@ const boardVersion = (
  */
 export default function AppealHelpSidebar({ location, aoj }) {
   if (location === 'aoj') {
-    // If the location is 'aoj', we have to check which agency it came from
+    // If the location is 'aoj', we have to check which agency it came from to show
+    //  the appropriate information.
     switch (aoj) {
       case 'vba': return vbaVersion;
       case 'vha': return null; // vha version (coming soon to a sidebar near you!)
--- a/src/js/claims-status/components/appeals-v2/AppealHelpSidebar.jsx
+++ b/src/js/claims-status/components/appeals-v2/AppealHelpSidebar.jsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import Raven from 'raven-js';
 
 const vbaVersion = (
   <div>
@@ -44,12 +45,12 @@ export default function AppealHelpSidebar({ location, aoj }) {
     } else if (aoj === 'other') {
       return boardVersion;
     } else {
-      // Ah, bugger. Better log it.
+      Raven.captureMessage(`appeal-status-unexpected-aoj: ${aoj}`);
     }
   } else if (location === 'bva') {
     return boardVersion;
   } else {
-    // Well that's not good; log it!
+    Raven.captureMessage(`appeal-status-unexpected-location: ${location}`);
   }
 
   return null;
--- a/src/js/claims-status/components/appeals-v2/AppealHelpSidebar.jsx
+++ b/src/js/claims-status/components/appeals-v2/AppealHelpSidebar.jsx
@@ -6,6 +6,7 @@ const vbaVersion = (
     <p className="help-talk">Call the Veterans Affairs Benefits and Services</p>
     <p className="help-phone-number">
       <a className="help-phone-number-link" href="tel:1-800-827-1000">1-800-827-1000</a>
+      <br/>
       Monday - Friday, 8:00am - 9:00pm (ET)
     </p>
   </div>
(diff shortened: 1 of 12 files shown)


gnakm:
@cvalarida I'm adding more love to this to follow the vets.gov. I hope the specs aren't scary O.O

- 45px space between the grey box and Need Help. For reference, this is following the spacing for the H2 pattern. 
- Need help should now be in font bitter
- The spacing between the 3 lines should be 16px
- There should be about 60px space between the last line and top of the footer

image

gnakm:
RE: tabs. I wouldn't worry too much about it in this PR. We won't be adding anything in the detail section until later sprints so I say, lgtm for this ticket and I'll add more specific specs later :)

cvalarida:
I love specs! They make me irrationally happy. I'll work some magic to make it look as good as the design.

Should this also change the look of the sidebar in the index page? I assume so for consistency, but figured I should ask.

Current index page (sans actual claims):
image


Thanks, @gnakm!

gnakm:
@cvalarida Yes if other pages are doing it, we should definitely leave it but with the new specs :)

cvalarida:
Done! The screenshots have been updated in the original post (comment?).

## Old screenshots
For record-keeping purposes only.

**Wide screen**
image

**Mobile**
image

gnakm (approved):
Yassss 🎉 
LGTM

src/js/claims-status/components/appeals-v2/AppealHelpSidebar.jsx
```diff
@@ -0,0 +1,57 @@
...
+    if (aoj === 'vba') {
+      return vbaVersion;
+    } else if (aoj === 'vha') {
+      // vha version (coming soon to a sidebar near you!)
+    } else if (aoj === 'nca') {
+      // nca version (coming soon to a sidebar near you!)
+    } else if (aoj === 'other') {
+      return boardVersion;
+    } else {
+      // Ah, bugger. Better log it.
```
markgreenburg: Did you want to actually log these? Also on L#51

src/js/claims-status/components/appeals-v2/AppealHelpSidebar.jsx
```diff
@@ -0,0 +1,57 @@
...
+ * @param {String} location  The location of the appeal. Possible options:
+ *                            ['aoj', 'bva']
+ * @param {String} aoj       The Agency of Original Jurisdiction.
+ *                            Used if the location is 'aoj'
+ *                            Possible options:
+ *                            ['vba', 'vha', 'nca', 'other']
+ */
+export default function AppealHelpSidebar({ location, aoj }) {
+  if (location === 'aoj') {
+    if (aoj === 'vba') {
```
markgreenburg: Tinsy bit confusing having both `aoj` and `'aoj'` as separate nested conditions, but I've tried several ways of breaking this out and they're all bad so 🤷‍♂️

src/js/claims-status/containers/AdditionalEvidencePage.jsx
```diff
@@ -114,7 +114,7 @@ class AdditionalEvidencePage extends React.Component {
                   onDirtyFields={this.props.setFieldsDirty}/>
               </div>
             </div>
-            <div className="small-12 usa-width-one-third medium-4 columns">
+            <div className="small-12 usa-width-one-third medium-4 columns help-sidebar">
```
markgreenburg: out of curiosity, what's going on here? Are we using 2 separate grid systems?

src/js/claims-status/components/appeals-v2/AppealHelpSidebar.jsx
cvalarida: Yeah, s'pose I could do that now. I was going to wait, but I have no real compelling reason to not do it now.

src/js/claims-status/components/appeals-v2/AppealHelpSidebar.jsx
cvalarida: I'll add a clarifying comment and make it a `switch` for brevity.

src/js/claims-status/containers/AdditionalEvidencePage.jsx
cvalarida: I think so, yeah. `usa-width-one-third` is the grid from USWDS, but the problem with that is it's not responsive, so we're still still using `small-12 medium-4 columns` from Foundation.

As I understand it, the goal is to use `usa-width-*` where possible, then flexbox where it makes sense, and only using Foundation as a fallback when neither option is really great.

markgreenburg (approved):
Looks good! Might want to add some tests for the new component & add some assertions for `AppealInfo`?

cvalarida:
Oh, great point, @markgreenburg. Will do. :+1:

PR #7069, opened 2018-01-09
content structure for screen reader accessibility

bshyong:
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/5445

### DIFF
--- a/src/js/login/components/Verify.jsx
+++ b/src/js/login/components/Verify.jsx
@@ -75,23 +75,21 @@ class Verify extends React.Component {
         <div className="container">
           <div className="row">
             <div className="columns small-12">
-              <h1>Verify your identity</h1>
-            </div>
-          </div>
-          <div className="row">
-            <div className="columns small-12">
-              <AlertBox
-                content={`You signed in with ${signinMethod[this.props.profile.authnContext] || 'ID.me'}`}
-                isVisible
-                status="success"/>
-              <p>
-                We'll need to verify your identity so that you can securely access and manage your benefits.<br/>
-                <a href="/faq/#why-verify" target="_blank">Why does Vets.gov verify identity?</a>
-              </p>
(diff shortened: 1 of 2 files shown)


src/js/login/components/Verify.jsx
```diff
@@ -75,23 +75,21 @@ class Verify extends React.Component {
...
-                status="success"/>
-              <p>
-                We'll need to verify your identity so that you can securely access and manage your benefits.<br/>
-                <a href="/faq/#why-verify" target="_blank">Why does Vets.gov verify identity?</a>
-              </p>
-              <p>This one-time process will take <strong>5 - 10 minutes</strong> to complete.</p>
-              <button className="usa-button-primary va-button-primary" onClick={this.handleVerify}>
-                <img alt="ID.me" src="/img/signin/idme-icon-white.svg"/><strong> Verify with ID.me</strong>
-              </button>
+              <div>
```
U-DON: We don't really need this outer `div`, do we?

U-DON (approved):
LGTM.

src/js/login/containers/Main.jsx
```diff
@@ -214,18 +216,26 @@ class Main extends React.Component {
...
             login={this.props.login}
             profile={this.props.profile}
             handleLogin={this.handleLogin}/>);
+        break;
       default:
-        return null;
+        content = null;
     }
+
+    return (
```
U-DON: Is there a reason to do it this way now?

src/js/login/components/Verify.jsx
bshyong: Yes I'll get rid of this

src/js/login/containers/Main.jsx
U-DON: I actually would preferr returning from the switch since there's no common content between all three of those to be rendered in this chunk.

PR #7070, opened 2018-01-09
Refactor form review page to separate SiP code

jbalboni:
Similar to the other PRs, this pulls the save in progress code out into a separate component, instead of having it mixed in with SubmitButtons and ReviewPage.


src/js/common/schemaform/review/ReviewPage.jsx
```diff
@@ -224,10 +181,7 @@ const mapDispatchToProps = {
...
   setPrivacyAgreement,
   setData,
-  uploadFile,
-  saveAndRedirectToReturnUrl,
-  autoSaveForm,
-  toggleLoginModal
+  uploadFile
 };
 
 ReviewPage.propTypes = {
```
cvalarida: Should we add `contentAfterButtons`, `renderErrorMessage`, and `setData` in here?

src/js/common/schemaform/save-in-progress/RoutedSavableReviewPage.jsx
```diff
@@ -0,0 +1,136 @@
...
+import { toggleLoginModal } from '../../../login/actions';
+
+import { ReviewPage } from '../review/ReviewPage';
+import { setData, setPrivacyAgreement, setEditMode, setSubmission, submitForm, uploadFile } from '../actions';
+
+class RoutedSavableReviewPage extends React.Component {
+  constructor(props) {
+    super(props);
+    this.autoSave = this.autoSave.bind(this);
+    this.debouncedAutoSave = _.debounce(1000, this.autoSave);
```
cvalarida: For consistency, can we make `autoSave` an arrow function? Not sure if that'd mess up `debouncedAutoSave` or not.

src/js/common/schemaform/save-in-progress/RoutedSavableReviewPage.jsx
```diff
@@ -0,0 +1,136 @@
...
+          form={form}
+          user={user}
+          saveAndRedirectToReturnUrl={this.props.saveAndRedirectToReturnUrl}
+          toggleLoginModal={this.props.toggleLoginModal}/>
+      </div>
+    );
+
+    return (
+      <ReviewPage
+        {...this.props}
```
cvalarida: This will also pass all the props from `connect`, which `ReviewPage` won't need. `user` shouldn't cause any conflicts (just extra prop passing), but I'm not sure how `form` will interact with the`form` from `ReviewPage`'s `mapStateToProps`.

If there's no conflict by passing `form`, the worst I can think of that'll happen is unnecessary renders because `user` changed in the redux store, but I figured it's probably worth mentioning.

cvalarida (approved):
Had a few small suggestions / questions, but it looks good. 👍

src/js/common/schemaform/save-in-progress/RoutedSavableReviewPage.jsx
jbalboni: This page is not using the connected `ReviewPage`, but the regular component. So it should conflict. But you're right that it is passing some properties that are unnecessary. I went back and forth on deciding if it was less brittle to do this vs passing them individually (which is more than just form, it also includes the action functions).

cvalarida: Ohhh, I see what you did there. Clever! 😄 
Yeah, like I said, worst case scenario is it updates more than it should because props `ReviewPage` doesn't care about are being changed. `RoutedSavableReviewPage` is updating because of those props anyhow, so...kinda whatever, I guess.

PR #7071, opened 2018-01-09
Add applicant info page for VRE chapter 36

jbalboni:
screen shot 2018-01-09 at 2 45 01 pm


cehsu (approved):
LGTM. It looks like an extra file (src/js/vre/chapter36/config/.form.js.swm) snuck onto this PR.

PR #7073, opened 2018-01-09
Appealsv2 status overhaul mg

markgreenburg:
- Adds a header to the status page
- Split up CurrentStatus and Timeline components (changed html structure, render `CurrentStatus` in `AppealInfo` instead of in `Timeline`
- Match `CurrentStatus` and `Timeline` styling to mocks in linked issue
- Update tests and some proptypes / prop passing in components.

**New Screenshots:**
screen shot 2018-01-09 at 4 05 23 pm

screen shot 2018-01-09 at 4 05 13 pm

screen shot 2018-01-09 at 4 04 18 pm

screen shot 2018-01-09 at 4 04 04 pm

### DIFF
--- a/src/js/claims-status/components/appeals-v2/AppealsV2TabNav.jsx
+++ b/src/js/claims-status/components/appeals-v2/AppealsV2TabNav.jsx
@@ -3,10 +3,9 @@ import PropTypes from 'prop-types';
 
 import TabItem from '../TabItem';
 
-const AppealsV2TabNav = ({ appealId, claimHeading }) => {
+const AppealsV2TabNav = ({ appealId }) => {
   return (
     <div>
-      <h1>{claimHeading}</h1>
       <ul className="va-tabs claims-status-tabs small-12 medium-10" role="tablist">
         <TabItem shortcut={1} className="appeals-tabs-item" id="v2status" tabpath={`appeals-v2/${appealId}/status`} title="Status"/>
         <TabItem shortcut={2} className="appeals-tabs-item" id="v2detail" tabpath={`appeals-v2/${appealId}/detail`} title="Detail"/>
--- a/src/js/claims-status/components/appeals-v2/AppealsV2TabNav.jsx
+++ b/src/js/claims-status/components/appeals-v2/AppealsV2TabNav.jsx
@@ -3,12 +3,15 @@ import PropTypes from 'prop-types';
 
 import TabItem from '../TabItem';
 
-const AppealsV2TabNav = ({ appealId }) => {
+const AppealsV2TabNav = ({ appealId, claimHeading }) => {
   return (
-    <ul className="va-tabs claims-status-tabs small-12 medium-10" role="tablist">
-      <TabItem shortcut={1} className="appeals-tabs-item" id="v2status" tabpath={`appeals-v2/${appealId}/status`} title="Status"/>
-      <TabItem shortcut={2} className="appeals-tabs-item" id="v2detail" tabpath={`appeals-v2/${appealId}/detail`} title="Detail"/>
-    </ul>
+    <div>
+      <h1>{claimHeading}</h1>
+      <ul className="va-tabs claims-status-tabs small-12 medium-10" role="tablist">
+        <TabItem shortcut={1} className="appeals-tabs-item" id="v2status" tabpath={`appeals-v2/${appealId}/status`} title="Status"/>
+        <TabItem shortcut={2} className="appeals-tabs-item" id="v2detail" tabpath={`appeals-v2/${appealId}/detail`} title="Detail"/>
+      </ul>
+    </div>
   );
 };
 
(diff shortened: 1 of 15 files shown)


src/js/claims-status/components/appeals-v2/AppealsV2TabNav.jsx
```diff
@@ -3,12 +3,15 @@ import PropTypes from 'prop-types';
...
 
-const AppealsV2TabNav = ({ appealId }) => {
+const AppealsV2TabNav = ({ appealId, claimHeading }) => {
   return (
-    <ul className="va-tabs claims-status-tabs small-12 medium-10" role="tablist">
-      <TabItem shortcut={1} className="appeals-tabs-item" id="v2status" tabpath={`appeals-v2/${appealId}/status`} title="Status"/>
-      <TabItem shortcut={2} className="appeals-tabs-item" id="v2detail" tabpath={`appeals-v2/${appealId}/detail`} title="Detail"/>
-    </ul>
+    <div>
+      <h1>{claimHeading}</h1>
```
cvalarida: Just for separation of concerns, can we pull this header out and put it directly in `AppealInfo`?

src/js/claims-status/components/appeals-v2/CurrentStatus.jsx
```diff
@@ -2,17 +2,22 @@ import React from 'react';
...
+  </div>
 );
 
 CurrentStatus.PropTypes = {
-  key: PropTypes.number,
   title: PropTypes.string.isRequired,
-  description: PropTypes.string.isRequired,
+  description: PropTypes.oneOfType([
+    PropTypes.string,
+    PropTypes.element
```
cvalarida: Ahh, I had to hunt this down to find out that it already should take a string or element. It's not a documentation oversight in this PR, but can we fix the documentation in `src/js/claims-status/utils/appeals-v2-helpers.jsx` on line 30 for `getStatusContents`? It should read `@property {string|HTMLElement}`.

src/js/claims-status/components/appeals-v2/CurrentStatus.jsx
```diff
@@ -2,17 +2,22 @@ import React from 'react';
...
 const CurrentStatus = ({ title, description }) => (
-  <li className="process-step section-current">
+  <div className="current-status">
     <h2>Current Status</h2>
-    <h4>{title}</h4>
-    <div>{description}</div>
-  </li>
+    <div className="current-status-content">
+      <h3>{title}</h3>
+      <p>{description}</p>
```
cvalarida: If `getStatusContents` returns an element for the description (because the appeal type is `bvaDecision`, for instance), we'll have
```html
<p>
  <div></div>
</p>
```
and React will yell at us again for the invalid DOM nesting.

Maybe do something more like:
```jsx
const descElement = typeof description === 'string' ? (<p>{description}</p>) : description;
```

cvalarida (approved):
Looks great! I like the little refactors you threw in there too. Way to clean the place up a bit. _[Insert high five emoji here]_

I had a few suggestions / questions, but it looks good. 👍

gnakm:
@markgreenburg A couple of very tiny things that I think it's worth adding to this PR:

- [ ] Could we make the triangle exclamation point icon in the error box aligned with the status title and description text (see image below)
- [ ] Add a space between the title and description in the alert box
- [ ] Change the  ----- OR ------ in the What Happens next section to lowercase "or" bump it up to 22pt and make it font bitter?

image

cvalarida:
@gnakm This particular PR is only re-styling the Current Status section (and the overhead involved with that), so I don't know that it makes sense to do those tweaks you mentioned in here. I think we should make another ticket for them and address them separately.

gnakm:
@cvalarida Gotcha 👍 Opening another ticket :)

gnakm (approved):
lgtm!

markgreenburg:
@gnakm How's this?

screen shot 2018-01-10 at 3 48 43 pm

screen shot 2018-01-10 at 3 48 28 pm

gnakm:
lgtm!

src/js/claims-status/components/appeals-v2/AppealsV2TabNav.jsx
markgreenburg: That seems like a very reasonable idea.

markgreenburg: Also happens to introduce a visual regression in the help sidebar top alignment; I'll open a new issue for that.

src/js/claims-status/components/appeals-v2/CurrentStatus.jsx
markgreenburg: 🤔 yeah agreed, and I don't want to introduce any more of these bugs than there already are. Seems like the safest thing to do might be to change this to a div, and on the `getStatusContents` side have each description output a jsx snippet instead of a string or snippet. That would let us enforce styles on the wrapping div, but have a semantic nested element with the actual content. Most of the content is already jsx snippets anyway.

src/js/claims-status/components/appeals-v2/CurrentStatus.jsx
markgreenburg: Well, now it's just `HTMLElement` 👍

PR #7074, opened 2018-01-10
Disable Feedback Widget on VIC

ncksllvn:
We're receiving too much feedback for us to handle coming from VIC.

Resolves department-of-veterans-affairs/vets.gov-team/issues/7555


omgitsbillryan (approved):
LGTM

PR #7079, opened 2018-01-10
Gibs vet not found message mg

markgreenburg:
Replaces 'Record Not Found' error message in GIBS app (which happens when `vet_not_found` causes `vets-api` to return a 404 to the front end) with a 'System Down' error message.


cvalarida (approved):

markgreenburg:
Screengrab:

screen shot 2018-01-10 at 4 43 32 pm
This is just the vets.gov default system down error message.

PR #7086, opened 2018-01-11
Main Nav Fix -- IE

cvalarida:
IE didn't like the `forEach` on an array-like object, so I just made it an array.


cehsu (approved):

cehsu:
Does IE support forEach?

ncksllvn (approved):
LGTM

cvalarida:
@cehsu Yeah, since IE9: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach

Trouble was it was giving the following error: `Object doesn't support property or method 'forEach'`

Turns out a `NodeList` doesn't have `forEach` in IE. Because IE. :man_shrugging:

cehsu:
Thanks @cvalarida!

PR #7089, opened 2018-01-12
Add VIC scaffolding

jbalboni:
I chose `vic-v2` as the url and folder name, but I'm assuming we'll take `veteran-id-card` eventually for at least the folder name. Maybe for the url, too, depending on the static content IA.


src/js/vic-v2/routes.jsx
```diff
@@ -0,0 +1,12 @@
+import { createRoutes as createFormRoutes } from '../common/schemaform/helpers';
+import formConfig from './config/form';
+import Chapter36App from './VeteranIDCardApp';
```
cehsu: Can we please rename this?

cehsu (approved):
LGTM. It looks like a temp file snuck into the PR. (src/js/vic-v2/config/.form.js.swm)

src/js/vic-v2/routes.jsx
jbalboni: Doh, yes.

PR #7090, opened 2018-01-12
Scheduled Downtime Notifications

ncksllvn:
This PR implements scheduled downtime throughout our website for application that are all or nothing - if an external service that an application depends on is unavailable, then so is the application. These applications include:

- ~~The claim status tool~~
- The health records tool
- The secure messaging tool
- The Veteran ID Card application
- The prescription refill tool
- The facility locator
- The user profile page
- The Post-9/11 GI Bill benefits tracking tool
- ~~The  appeals status tool~~

Below are the different states of an application that implements Downtime Notifications.

# Status: OK
Application is rendered as normal.

---------------------
# Status: Downtime Approaching (< 1 hour)

image

---------------------
# Status: In Downtime

image

----------------------

Resolves department-of-veterans-affairs/vets.gov-team/issues/7113
[Spreadsheet about application dependencies](https://docs.google.com/spreadsheets/d/1QtHiKZW5DHpm8oZ8rtNFxCfgVNyIW7UNqaeBDdQg8u4/edit#gid=0)


src/js/common/helpers/api.js
```diff
@@ -15,10 +15,12 @@ export function apiRequest(resource, optionalSettings = {}, success, error) {
...
     : resource;
 
+  const Authorization = sessionStorage.userToken && `Token token=${sessionStorage.userToken}`;
+
   const defaultSettings = {
     method: 'GET',
     headers: {
-      Authorization: `Token token=${sessionStorage.userToken}`,
+      Authorization,
       'X-Key-Inflection': 'camel'
```
ncksllvn: I'm adding a check here because otherwise if sessionStorage.userToken is `undefined`, then the authorization header is set to `Token token=undefined`. Then our mock API registers the requests to that instead of `_global`.

test/common/components/RequiredTermsAcceptance.unit.spec.jsx
```diff
@@ -18,6 +18,9 @@ const defaultProps = {
 };
 
 describe('<RequiredTermsAcceptanceView>', () => {
+
+  global.window.dataLayer = [];
+
```
ncksllvn: I have no idea why this test was failing. I don't think it has anything to do with this PR because it also failed when I switched to master and reran it there locally...so I think another test was initializing `window.dataLayer` and this test would usually run after that.

ncksllvn:
#### Design Implementation
An API route provides an array of downtime windows pertaining to different services. A React components called `DowntimeNotifications` can then be used to consume that information by passing it a property called `dependencies`', which is a list of external services for which that application depends. The JSX returned from the `render` function can then be wrapped in a `<DowntimeNotifcation/>`, and if the API provides information about an external service going offline within the hour or is current offline, your application will respond accordingly.

So, the code for some app is something like below. In the case that MHV is going offline within the hour, the user is told that information but some app is still usable. If MHV is currently is offline, then the user is told that information and they are not shown some app.

```jsx
import React from 'react';
import DowntimeNotification, { services } from '../../common/containers/DowntimeNotification';

class SomeApp extends React.Component {
    render() {
        return (
            <DowntimeNotification dependencies={[services.mhv]}>
                <h1>Hey there, if you're reading this, then MHV must be good to go!</h1>
            </DowntimeNotification>
        )
    } 
}
```

ncksllvn:
To mock data for testing, you can see this [commit ](https://github.com/department-of-veterans-affairs/vets-website/blob/91834c9c09eeefe003eb336755ed845de4f06a1f/src/js/common/actions/index.js) where instead of an API hit the Redux action just supplies test data. Or you can modify your local API's maintenance_window_controller.rb to do something like below. Or actually throw some data in your local database and do it that way.

```ruby
   def index
      render json: { 
        data: [
        {
          id: '139',
          type: 'maintenance_windows',
          attributes: {
            external_service: 'appeals',
            description: 'Benefits services will be unavailable for 24 hours.',
            start_time: Time.zone.now.yesterday,
            end_time: Time.zone.now.tomorrow,
            description: 'maintenance from yesterday'
        }
       ]
     }
```

melwoodard (dismissed):
Hi @ncksllvn -- I think there was some miscommunication on what the UI should be for this first iteration. The UI should look the same whether authenticated or not, the only difference is whether downtime is _approaching_ vs. _in effect_ 

Intended presentation when the user lands on the affected application:

1. Downtime approaching = modal overlays the page

2. Downtime in effect = plain messaging replaces all content below the page title

It should just be a matter of swapping the two alert banners you show above to instead be an overlay (for upcoming) and plain text (for in effect).

ncksllvn:
Thanks @melwoodard, I guess I overlooked that change in requirements. I think we will need to add another ticket for when we want to add the functionality back in later because I didn't include it in Pt 2 or 3.

I'll make the changes asap.

melwoodard:
Thanks @ncksllvn - to be clear, the different experience for auth vs. non-auth is what's outlined in the Pt 3 ticket (https://github.com/department-of-veterans-affairs/vets.gov-team/issues/8124) -- showing an alert in the sign in modal when a non-authed user tries to access an auth-required app. So we should be good!

ncksllvn:
@melwoodard Ohhh, I was under the impression that that ticket was specifically for warnings in the Sign-in Modal because that requires the extra step of mapping the URL parameters to application dependencies. Are we planning on ever adding back the stuff that was originally in this PR that changed used the warning banners?

I just updated this PR with those updates btw.

melwoodard:
Nope, the warnings in the sign in modal cover the need, we don't need the warnings you initially had in this PR.

src/js/common/containers/DowntimeNotification.jsx
```diff
@@ -0,0 +1,248 @@
...
+    const inclusive = '[]';
+
+    if (!downtime) return serviceStatus.ok;
+    if (current.isBetween(downtime.startTime, downtime.endTime, inclusive)) return serviceStatus.down;
+    if (warning.isBetween(downtime.startTime, downtime.endTime, inclusive)) return serviceStatus.downtimeApproaching;
+
+    return serviceStatus.ok;
+  }
+
+  getDowntimeWindow(downtimes) {
```
cvalarida: So if we have something like
```js
downtimes = [
  { startTime: '10:00', endTime: '11:00' },
  { startTime: '12:00', endTime: '13:00' }
}
```
then `getDowntimeWindow(downtimes)` will return `{ startTime: '11:00', endTime: '13:00' }`, is that right?
(This took me entirely too long to figure out what the purpose was. :grimacing: )

src/js/common/containers/DowntimeNotification.jsx
```diff
@@ -0,0 +1,248 @@
...
+      const endTime = moment(downtime.endTime);
+      return {
+        startTime: result.startTime && result.startTime.isBefore(startTime) ? result.startTime : startTime,
+        endTime: result.endTime && result.endTime.isAfter(endTime) ? result.endTime : endTime,
+      };
+    }, {});
+  }
+
+  // Converts the array of dependencies/service names into key/value pairs, with service statuses as keys and a list of
+  // downtime information (each downtime corresponding to a dependency/service name) as the values.
```
cvalarida: Perhaps an example would help clear this description up? Could just be me, but I was having a hard time making sense of it.

src/js/common/helpers/api.js
```diff
@@ -18,11 +18,14 @@ export function apiRequest(resource, optionalSettings = {}, success, error) {
...
     headers: {
-      Authorization: `Token token=${sessionStorage.userToken}`,
       'X-Key-Inflection': 'camel'
     }
   };
 
+  if (sessionStorage.userToken) {
+    defaultSettings.headers.Authorization = `Token token=${sessionStorage.userToken}`;
+  }
+
```
ncksllvn: The way this was before we would see `Token token=undefined` and that didn't work well with the mock API

ncksllvn:
Thanks @melwoodard, sounds good! Sorry about the mixup there.

src/js/common/containers/DowntimeNotification.jsx
```diff
@@ -0,0 +1,248 @@
...
+
+    return serviceStatus.downtimeApproaching;
+  }
+
+  renderStatusDown({ endTime }) {
+    const title = `The ${this.props.appTitle} is down for maintenance.`;
+    let message = <p>We’re making some updates to {this.props.appTitle}. We’re sorry it’s not working right now. Please check back soon.</p>;
+    if (endTime) {
+      message = (
+        <p>We’re making some updates to {this.props.appTitle}. We’re sorry it’s not working right now, and we hope to be finished by {endTime.format('MMMM Mo, LT')}. Please check back soon.</p>
```
cvalarida: Ooh, didn't know `LT` was a thing. Very cool.
Just to make sure all our bases are covered, have you made sure this plays nice with time zones? I know Mark ran into some funny stuff a few days ago with time zones.

src/js/common/containers/DowntimeNotification.jsx
```diff
@@ -0,0 +1,248 @@
...
+    return (
+      <DowntimeNotificationWrapper status={serviceStatus.down}>
+        {downtimeNotification}
+      </DowntimeNotificationWrapper>
+    );
+  }
+
+  renderStatusDownApproaching({ startTime, endTime }) {
+    const title = `The ${this.props.appTitle} will be down for maintenance soon`;
+    const message = <p>We'll be doing some work on {this.props.appTitle} on {startTime.format('MMMM Mo')} between {startTime.format('LT')} and {endTime.format('LT')} . If you have trouble using this tool during that time, please check back soon.</p>;
```
cvalarida: Couple things:
1. Straight quote needs to be a curly one (I'm not sure how to do this on this linux machine, so no example, sorry!)
2. Looks like there's an extra space between `{endTime.format('LT')}` and the following `.`

src/js/claims-status/containers/ClaimsStatusApp.jsx
```diff
@@ -28,9 +29,11 @@ class ClaimsStatusApp extends React.Component {
         userProfile={this.props.profile}
         loginUrl={this.props.loginUrl}
         verifyUrl={this.props.verifyUrl}>
-        <AppContent>
-          {this.props.children}
-        </AppContent>
+        <DowntimeNotification appTitle="claims status tool" dependencies={[services.evss]}>
```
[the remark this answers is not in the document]
ncksllvn: The former was my thinking too, but they'll actually be asked to login before they land on the application anyhow because of the way the menu links works with the `?next=` redirect. We have some work planned for a third piece that will tell the user in the login window prior to logging in if the application is experiencing downtime but for now it only checks when landing on the application. 

If they navigate directly (via bookmark or something) order would matter though - do you think it's worth making the change?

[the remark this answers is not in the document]
ncksllvn: Originally we were planning on difference messaging based on authenticated or not authenticated, so the downtime notifications had to wait anyway.

src/js/common/containers/DowntimeNotification.jsx
```diff
@@ -0,0 +1,248 @@
...
+};
+
+export const serviceStatus = {
+  down: 'down',
+  downtimeApproaching: 'downtimeApproaching',
+  ok: 'ok'
+};
+
+function DowntimeNotificationWrapper({ status, children }) {
+  return <div id="downtime-notification" data-status={status} style={{ marginBottom: '1em' }} className="row-padded">{children}</div>;
```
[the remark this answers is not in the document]
ncksllvn: No great explanation other than that I really didn't want to add a rule for just that single margin. Think it's worth adding?

src/js/common/containers/DowntimeNotification.jsx
```diff
@@ -0,0 +1,248 @@
...
+}
+
+class DowntimeNotification extends React.Component {
+
+  static propTypes = {
+    appTitle: PropTypes.string,
+    children: PropTypes.node,
+    content: PropTypes.node,
+    dependencies: PropTypes.arrayOf(PropTypes.oneOf(objectValues(services))).isRequired,
```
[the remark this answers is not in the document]
ncksllvn: It did work but it failed on Jenkins somehow!

src/js/common/containers/DowntimeNotification.jsx
```diff
@@ -0,0 +1,248 @@
...
+  }
+
+  componentWillMount() {
+    this.props.getScheduledDowntime();
+  }
+
+  // This is here just for caching the result of calculateDowntime, so that it isn't run every time a prop changes.
+  // Currently, this component should only do its calculations once, because it would be weird if in the middle of filling
+  // out a form a modal or alert appears alerting them of downtime, or if it just shuts down the app because we entered
+  // scheduled downtime.
```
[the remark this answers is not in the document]
ncksllvn: Right now it calculates downtime based on landing and that's all. That's how we defined it in our requirements to keep things simple in this first iteration. At a later date we may figure out those "transitional" states.

src/js/common/containers/DowntimeNotification.jsx
```diff
@@ -0,0 +1,248 @@
...
+
+    return nextState.cache.status !== serviceStatus.down;
+  }
+
+  getStatusForDowntime(downtime, current = moment(), warning = moment().add(1, 'hour')) {
+    const inclusive = '[]';
+
+    if (!downtime) return serviceStatus.ok;
+    if (current.isBetween(downtime.startTime, downtime.endTime, inclusive)) return serviceStatus.down;
+    if (warning.isBetween(downtime.startTime, downtime.endTime, inclusive)) return serviceStatus.downtimeApproaching;
```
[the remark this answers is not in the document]
ncksllvn: I wrote this assuming that outages will be at least an hour long so I liked the explicitness of `isBetween` but otherwise I think you're right. @patrickvinograd can you confirm?

src/js/common/containers/DowntimeNotification.jsx
ncksllvn: Yeah exactly! It has the full downtime object as returned from the action [here](http://jenkins.vetsgov-internal/blue/organizations/jenkins/testing%2Fvets-website/detail/scheduled_downtime/13/pipeline), but it only needs that properties. This may have been overkill, but it's ultimately finding the earliest start time and the latest end time from the application's dependencies and returning that as the downtime window.

Is this too cryptic? I tried to write it clearly but it was one of those things were it doesn't seem like there was a super clear way to do it.

ncksllvn: Sort of - it's ultimately just trying to find the earliest start time and the latest end time. This is probably overkill but I figured it's also the most correct way to do it in case there's more than one outage. I might rewrite this with variables named more explicitly for `earliestStart`.

src/js/common/containers/DowntimeNotification.jsx
ncksllvn: Good idea, will do

src/js/common/containers/DowntimeNotification.jsx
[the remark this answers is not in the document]
cvalarida: Ohh, that's interesting...Failed in the unit test, which wouldn't have loaded up the `src/js/common/polyfill.js`...but that should be running in Node which I thought shouldn't have a problem with `Object.values`. Unless we're running an old version?

Well, I guess we can ignore it for now. That's so odd, though!

src/js/common/containers/DowntimeNotification.jsx
ncksllvn: I believe it should be adjusted for the [locale](https://momentjs.com/docs/#/displaying/)

src/js/common/containers/DowntimeNotification.jsx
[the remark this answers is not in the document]
cvalarida: Oh, I get it: [Mozilla](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/values) says it's ES2017, which isn't ES6 (ES2015)...which [isn't supported](http://node.green/#ES2017-features-Object-static-methods) in Node v6.11.1 (and _that's_ what Jenkins is running the unit tests in).

ES2017 !== ES6 :man_facepalming: 

My bad.

src/js/claims-status/containers/AppealStatusPage.jsx
```diff
@@ -105,50 +106,52 @@ class AppealStatusPage extends React.Component {
     const { status } = appealStatusDescriptions(lastEvent, previousHistory);
 
     return (
-      <div className="claims-container">
-        <div className="row">
-          <div>
-            <h1 className="claims-container-title">Your Compensation Appeal Status</h1>
-            <p>This information is accurate as of {moment().format('MMMM DD, YYYY')}.</p>
+      <DowntimeNotification appName="appeals status tool" dependencies={[services.appeals]}>
```
jbalboni: On this app, this is only going to show up for the downtime approaching case, correct? Because if appeals is down, they'll be stuck on the claim list page.

src/js/claims-status/containers/ClaimsStatusApp.jsx
jbalboni: This is going to block someone from viewing their appeals if evss is down. Is that ok?

src/js/common/containers/DowntimeNotification.jsx
```diff
@@ -0,0 +1,234 @@
...
+  appeals: 'appeals',
+  arcgis: 'arcgis',
+  dslogon: 'dslogon',
+  emis: 'emis',
+  es: 'es',
+  evss: 'evss',
+  idme: 'idme',
+  mvi: 'mvi',
+  mhv: 'mhv',
+  tims: 'tims',
```
jbalboni: I doubt this one will ever be used. For edu, we submit over SFTP and going down doesn't impact users submitting applications.

ncksllvn:
Thanks for the CR, @cvalarida! Will make some updates here soon.

src/js/common/containers/DowntimeNotification.jsx
```diff
@@ -0,0 +1,234 @@
...
+      };
+    }, {});
+  }
+
+  // Converts the array of dependencies/service names into key/value pairs, with service statuses as keys and a list of
+  // downtime information (each downtime corresponding to a dependency/service name) as the values.
+  calculateDowntime(dependencies, scheduledDowntime) {
+    const now = moment();
+    const nextHour = moment().add(1, 'hour');
+    const downtimeMap = {};
```
jbalboni: I think you could use the native `Map` for this.

src/js/common/containers/DowntimeNotification.jsx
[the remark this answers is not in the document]
cvalarida: Yeah, I'm not sure. Is it in a preexisting pattern that may get a style update later? If so, it might be worth it. If not, I wouldn't worry about it.

src/js/common/containers/DowntimeNotification.jsx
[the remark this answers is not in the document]
jbalboni: Yeah, I think in the case of the longer forms, the downtime pattern will have to be different. We would probably want to notice logged in users in place of the submit button, and notify non-logged in users some less invasive way. Definitely makes sense to leave for the second iteration.

src/js/claims-status/containers/ClaimsStatusApp.jsx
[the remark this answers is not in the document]
cvalarida: Ahh, yeah that makes sense. I wouldn't worry about it.

At first, I didn't think the redirects happened until after the page was loaded up, but then I found this little number in `src/js/login/containers/Main.jsx`

```js
  bindNavbarLinks() {
    [...document.querySelectorAll('.login-required')].forEach(el => {
      el.addEventListener('click', e => {
        e.preventDefault();
        const nextQuery = { next: el.getAttribute('href') };
        const nextPath = appendQuery('/', nextQuery);
        history.pushState({}, el.textContent, nextPath);
        this.props.toggleLoginModal(true);
      });
    });
  }
```

I don't like :point_up: because it's not obvious that this is happening, but it's there and out of scope for this PR, so...I would say don't worry about it. The way you've got it right now makes enough sense. Navigating directly to an app with a required login is probably not super common (though I would love to see analytics on that).

src/js/common/containers/DowntimeNotification.jsx
jbalboni: There's also a particular way that the content team has defined for showing AM and PM, that we've had to set directly: https://github.com/department-of-veterans-affairs/vets-website/blob/933acc5b11ab52db6269fcc0ae275cb6b33a06b7/src/js/common/schemaform/save-in-progress/RoutedSavableApp.jsx#L31. So may want that in here, too.

src/js/common/containers/DowntimeNotification.jsx
```diff
@@ -0,0 +1,234 @@
...
+    return (
+      <DowntimeNotificationWrapper status={serviceStatus.downtimeApproaching}>
+        {downtimeNotification}
+        {this.props.children || this.props.content}
+      </DowntimeNotificationWrapper>
+    );
+  }
+
+  render() {
+    if (!this.props.isReady) return this.props.loadingIndicator || <LoadingIndicator message={`Checking ${this.props.appTitle} status...`}/>;
```
jbalboni: This would be clearer, I think, if it weren't a one line if statement.

src/js/common/containers/DowntimeNotification.jsx
cvalarida: Do we know what we'd expect to get from the api if we had downtimes on the same day, but like 10 hours apart? Would it give us both? I'm wondering if this means it'll say we've got a downtime of 11 hours, when we really only have two 1-hour downtimes.

src/js/claims-status/containers/ClaimsStatusApp.jsx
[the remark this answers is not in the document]
ncksllvn: Yeah exactly! I definitely agree. That's something we might be working with later on.

src/js/common/containers/DowntimeNotification.jsx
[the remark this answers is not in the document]
ncksllvn: I think it's okay at the moment. If any more style gets added then it'll definitely make sense to move it over but I think it might be staying pretty simple. Hoping to get by using default components/tags.

src/js/common/containers/DowntimeNotification.jsx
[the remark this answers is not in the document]
ncksllvn: Haha no worries, this is deep into the web of JS frontend/backend and preprocessors. We may need to revisit this at some point.

src/js/common/containers/DowntimeNotification.jsx
ncksllvn: That is a really great point, nice find. I suppose the pattern it should follow is finding the earliest start point, and the end point based on continuous outage...or maybe that's overkill and it should just be earliest start point and just use the corresponding end point.

patrickvinograd: I vote for the latter - just display the next upcoming outage for a given service. I doubt we'll have windows defined in rapid succession or overlapping, but we might, say, populate multiple distinct windows coming up in the next week when we hear about them.

src/js/common/containers/DowntimeNotification.jsx
ncksllvn: Good suggestion, thanks!

src/js/claims-status/containers/AppealStatusPage.jsx
ncksllvn: It was implemented based purely off of the spreadsheet; I don't know enough about the application to know to do otherwise. Is this a special case? If it is we can hold off for later.

ncksllvn:
@jbalboni Definitely let us know if you have some insight into the way Appeals is structured with its dependencies. If it isn't as simple as suggested in the [spreadsheet](https://docs.google.com/spreadsheets/d/1QtHiKZW5DHpm8oZ8rtNFxCfgVNyIW7UNqaeBDdQg8u4/edit#gid=0) we'll hold off for a later date.

src/js/claims-status/containers/ClaimsStatusApp.jsx
ncksllvn: Not sure, is there a better place to add the Downtime Notification? I went purely based on the spreadsheet and what seemed to me as the best entry points but feel free to point me in the right direction if I was off on any of this.

src/js/common/containers/DowntimeNotification.jsx
ncksllvn: That sounds good to me!

src/js/common/containers/DowntimeNotification.jsx
ncksllvn: Yeah good call. At one point it was nice and clean as a one liner but not so much now.

src/js/claims-status/containers/ClaimsStatusApp.jsx
jbalboni: This app has two independent dependencies, so given the current design, I don't think there's a good place to put it that doesn't block use of the app if just one of the dependencies is down.

ncksllvn: The downtime notification doesn't have to wrap the whole thing - is it blocking a menu in this case? And that's why they wouldn't be able to view their appeals if evss is down?

jbalboni: In the claim status app the initial page of the app is a list of claims and appeals, mixed together. My understanding is that this would replace that whole page with the downtime message.

cvalarida: Yeah, this is one of those places that an all-or-nothing block isn't a good fit. @ncksllvn here's kind of how it works:

The "claims-status" app is a bit of a misnomer. It acts as a way for the Veteran to view their claims _and_ the appeals to those claims (which are often seen as part of the claims themselves, which is why we've kept the name). When we go to the app, it'll query for both claims and appeals and list them on the index page. After one is selected, it'll open up either the claims status page or the appeals status page, depending on the type.

The trouble here is that claims and appeals rely on two different services. We could block those status pages since they rely on one service each, but that doesn't give the user any warning that a maintenance window has been scheduled. Ideally, we'd be able to handle showing messages to indicate either claims or appeals are down, and block the whole app if they're both down.

But that's not MVP.

@djmunoz1 or @rroueche, does it make sense to have this downtime notification block the app if _either_ evss or the appeals service is down?

src/js/common/containers/DowntimeNotification.jsx
cvalarida: Idle curiosity: What's the benefit of using a native Map vs just an object?

src/js/claims-status/containers/AppealStatusPage.jsx
cvalarida: Like I mentioned in another comment, this page is basically only accessed from the claims & appeals index page. If the appeals service is down, the appeals just won't show up in the index page, so the user won't be able to navigate here. Still, if they have it bookmarked or something, this'll help. What we do here will depend on what we do with the `ClaimsStatusApp`.

src/js/common/containers/DowntimeNotification.jsx
jbalboni: Clarity, mainly. Map is the data type that is intented for this use case. It is also slightly faster in some uses cases, but I doubt enough to matter here.

ncksllvn:
@jbalboni should we consider moving that `moment.updateLocale('en', { ...` stuff that the content team prefers into a single file somewhere? I see it several times in our project.

ncksllvn:
@jbalboni  @cvalarida  We moved Appeals and Claims to iteration 2 to keep things simple. If you guys don't mind taking another look, I think all the fixes have been made. Thanks a ton for your awesome reviews, definitely appreciated. 

We're trying to finish QA'ing with the Pager Duty API tomorrow and go live if possible so please keep me posted if you have any questions/concerns. If you want to run locally and need help getting set up I'm happy to help.

@leannammiller for visibility

src/js/common/containers/DowntimeNotification.jsx
```diff
@@ -0,0 +1,244 @@
...
+      if (!currentSoonest) return downtime;
+
+      const isEarlier = moment(downtime.startTime).isBefore(currentSoonest.startTime);
+      return isEarlier ? downtime : currentSoonest;
+    }, null);
+
+    // Return the startTime/endTime as moment objects for the soonest downtime.
+    return {
+      startTime: soonestDowntime && moment(soonestDowntime.startTime),
+      endTime: soonestDowntime && moment(soonestDowntime.endTime)
```
cvalarida: Maybe splitting hairs here because this may never happen, but if we've got the following downtimes
```js
downtimes = [
  { startTime: '12:00', endTime: '13:00' },
  {  endTime: '15:00' }
]
```
I think this will return
```js
{
  startTime: moment(undefined), // resolves to the current time
  endTime: moment('15:00')
}
```
when the current time is < 12:00. This might be the intended behavior, but I wanted to point it out.

test/common/containers/DowntimeNotification.unit.spec.jsx
```diff
@@ -0,0 +1,112 @@
...
+    it('should render the children components when there are no dependencies', () => {
+      const wrapper = getComponent();
+      wrapper.setProps({ isReady: true, scheduledDowntime: [] });
+      expect(wrapper.text()).to.contain(innerText, 'The message was rendered.');
+    });
+
+    it('should render the children components when dependencies have no downtime', () => {
+      const wrapper = getComponent([services.mvi]);
+      wrapper.setProps({ isReady: true, scheduledDowntime: [] });
+      expect(wrapper.text()).to.contain(innerText, 'The message was rendered.');
```
cvalarida: Huh. Well, if this passes, I'm satisfied. When I was trying to figure out how `getDowntimeWindow` might break things, I found if `downtimes` was an empty array, it should return `{ startTime: false, endTime: false }` which would cause an error in `renderStatusDown` or `renderStatusDownApproaching`. It also seemed like `getDowntimeWindow` _could_ be called with an empty array from `componentWillRecieveProps`.

Turns out I must be missing something, though, if this passes. :man_shrugging:

cvalarida (approved):
Looks good from where I sit! I was concerned that it might break, but if the tests pass, I think it's golden.

src/js/common/containers/DowntimeNotification.jsx
ncksllvn: Actually it's only looking at the `startTime` now and grabbing the corresponding `endTime` off that. So that should return 

```
{
  startTime: moment('12:00'), // resolves to the current time
  endTime: moment('13:00')
}
```
But that brings up a good point, because the endTime can be omitted in which case we shouldn't pass it into moment. I just published a patch for that case.

ncksllvn: Oh I see what you mean. The `startTime` is required though, so that shouldn't ever happen. `endTime` though is optional, which isn't clear here. I just published a patch for this.

test/common/containers/DowntimeNotification.unit.spec.jsx
ncksllvn: It's confusing, but I just added a check to do the `getDowntimeWindow` only when the status isn't ok. There was already a check for that prior to this, but it wasn't clear enough. It was tough working with so many date operations.

DanielleSoCompany:
@leannammiller let me know if this works for the new cases:

**Please save your application and come back to submit it after [DOWNTIME DATE]

We’re making some updates to this application right now. You can still fill out the application. And, if you’re signed in to Vets.gov, you can save your information. But you won’t be able to submit your application until we’re finished with our updates. Please save your application and come back after [DOWNTIME DATE] to submit it.

**We won’t be able to accept applications from [DOWNTIME START DATE] to [DOWNTIME END DATE]**

We’re going to be making some updates to this application from [DOWNTIME START DATE] to [DOWNTIME END DATE].** During this time, you’ll still be able to fill out the application. And, if you’re signed in to Vets.gov, you’ll be able to save your information. But you won’t be able to submit your application until we’re finished with our updates. Please save your application and come back after [DOWNTIME DATE] to submit it.

jkassemi:
Aside: this would be great to tie into the future user notifications feature.

melwoodard:
@jkassemi read my mind -- I'm discussing with @Samara-Strauss to figure out how this fits in with asynchronous submission.

Samara-Strauss:
Noted! We're actually talking about how to handle downtime here https://github.com/department-of-veterans-affairs/vets.gov-team/issues/8573#issuecomment-364206744

PR #7095, opened 2018-01-15
VIC applicant info page

jbalboni:
Based on conversations with Stacy, this does not include the file number or the branch of service.

screen shot 2018-01-15 at 2 43 37 pm


cehsu (approved):
LGTM.

PR #7098, opened 2018-01-15
Remove the rest of the SiP code from schemaform code

jbalboni:
I think this is the rest of it. The main changes are:

1. The prefill message we show on different pages is now a regular component that checks `prefilled` in `formContext`.
2. `SchemaForm` exposes a `formContext` prop that gets merged with the object passed to `Form`. This replaces the `prefilled` prop. There's a reselect selector used to avoid changing the object more often than necessary.


src/js/hca/config/form.js
```diff
@@ -369,6 +366,7 @@ const formConfig = {
           path: 'military-service/service-information',
           title: 'Service periods',
           uiSchema: {
+            'ui:description': __BUILDTYPE__ !== 'production' ? MilitaryPrefillMessage : undefined,
```
cvalarida: Because we can't pre-fill military history outside of production?

cvalarida (approved):

src/js/hca/config/form.js
jbalboni: We actually can and do, but this message is just being shown in staging for user testing.

PR #7100, opened 2018-01-16
VIC v2 photo cropper

cehsu:
This implements the VIC v2 photo cropping tool. Although some work remains to be done,  including style refinement for transitioning between breakpoints, styles clean up, code reorganization into helpers, and tests, this version should be ready for testing purposes.

localhost_3001_vic-v2_documents_photo iphone 7

localhost_3001_vic-v2_documents_photo iphone 7 1

localhost_3001_vic-v2_documents_photo iphone 7 2

localhost_3001_vic-v2_documents_photo iphone 7 3

localhost_3001_vic-v2_documents_photo

localhost_3001_vic-v2_documents_photo 1

localhost_3001_vic-v2_documents_photo 2

localhost_3001_vic-v2_documents_photo 3


src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -1,9 +1,144 @@
...
+
+  onZoom = (e) => {
+    const newZoomValue = e.detail.ratio;
+    this.setState({ zoomValue: newZoomValue }, () => {
+      this.refs.slider.value = this.state.zoomValue;
+    });
+  }
+
+  moveUp = (e) => {
+    e.preventDefault();
```
jbalboni: If you put `type="button"` on the buttons, these prevent default calls shouldn't be necessary.

src/sass/style.scss
```diff
@@ -23,6 +23,8 @@
   font-style: normal;
 }
 
+@import "~cropperjs/dist/cropper.css";
```
jbalboni: We should move this to the vic scss file.

jbalboni (commented):
Looks good so far. Is it possible to fix the position of the crop box? Also, is it possible to have it start with the uploaded image fit horizontally in the canvas, rather than vertically?

src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -1,9 +1,144 @@
...
+          <Cropper
+            ref="cropper"
+            src={this.state.src}
+            style={{ height: 400, width: '100%' }}
+            aspectRatio={1 / 1}
+            guides={false}
+            viewMode={1}
+            zoom={this.onZoom}
+            crop={this.cropImage}/>
+          <input type="range"
```
jbalboni: This slider doesn't seem to work right now.

src/sass/style.scss
cehsu: Thanks @jbalboni, this import has been moved to the vic scss file.

src/js/vic-v2/components/PhotoField.jsx
cehsu: Thanks @jbalboni, change made.

src/js/vic-v2/components/PhotoField.jsx
cehsu: I'm not able to replicate this bug (using the buttons, the slider, and the mouse to zoom, all appear to keep the image zoom and the slider position in sync), could you please indicate what steps are necessary to cause it?

jbalboni: If I upload an image and move the slider, it immediately zooms in a bunch. Sliding it back and forth does zoom in and out after that, but it's really zoomed in because of the initial jump and I can't use the slider to set it back the way it was initially.

cehsu: Thanks @jbalboni, I'll investigate this.

cehsu:
Thanks @jbalboni for your comments and helpful feedback, I have provided some thoughts on your comments below in line:

>Looks good so far. Is it possible to fix the position of the crop box? 

Great idea. It doesn't look like this is available by default but I'll look into a custom solution.

>Also, is it possible to have it start with the uploaded image fit horizontally in the canvas, rather than vertically?

I invite @stacyhlinn and @lalexanderson-dsva to share their thoughts on how we should go about considering default fitting for certain orientations.

It looks like we'll want for the image to fill the crop size, so depending on the orientation (landscape, portrait, square), it's possible we would want to favor filling the canvas (which is landscape orientation) vertically, rather than horizontally, so that if a passport photo is uploaded, for example, the applicant, would be able to crop it according to its original dimensions, if we fill horizontally, it could result in something like the following:
screen shot 2018-01-16 at 12 19 33 pm

jbalboni:
@cehsu I think the `cropBoxMovable` option will work for the first issue.

lalexanderson-dsva:
@cehsu Your suggestion to have the default canvas fill be vertical instead of horizontal makes sense to me.

jbalboni:
@cehsu Looks awesome so far! I noticed that if I upload an image that's wider than it is tall, I can't move it far enough left or right to get the edges into the (now unmovable) crop box. Is that something we can configure somewhere?

cehsu:
Good catch! Thanks for spotting that bug, @jbalboni, we can fix that by setting `viewMode` to 1, changed.

src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -1,9 +1,257 @@
...
+      zoomValue: 2,
+      errorMessage: null
+    };
+  }
+
+  componentDidUpdate(prevProps, prevState) {
+    const newView = !prevState.src || prevState.done;
+    const cropper = this.refs.cropper;
+    if (newView && cropper) {
+      setTimeout(() => {
```
jbalboni: Curious why this is needed

src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -1,9 +1,257 @@
...
+                Move left
+              </button>
+            </div>
+          </div>
+          <button type="button" onClick={this.onDone} style={{ 'float': 'right' }}>
+            I'm done
+          </button>
+        </div>
+        }
+        {!this.state.src && !this.state.done && <Dropzone onDrop={this.onChange} accept="image/jpeg, image/png"/>}
```
jbalboni: This accept list is different from the allowed file types in other places.

src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -1,9 +1,257 @@
...
+    this.setState({ src: null, done: true });
+  }
+
+  onChange = (files) => {
+    if (files && files[0]) {
+      const file = files[0];
+      const reader = new FileReader();
+      reader.readAsDataURL(file);
+      reader.onload = () => {
+        isValidImage(file)
```
jbalboni: It seems like we're reading in the file twice. Once on line 97, then again in isValidImage.

jbalboni (approved):

src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -1,9 +1,307 @@
...
+  'jpeg',
+  'jpg',
+  'bmp'
+];
+const MIN_SIZE = 350;
+
+function detectDragAndDrop() {
+  const div = document.createElement('div');
+  const supportsDragAndDrop = ('draggable' in div) || ('ondragstart' in div && 'ondrop' in div);
+  const iOS = !!navigator.userAgent.match('iPhone OS') || !!navigator.userAgent.match('iPad');
```
rianfowler: Probably not widely used but iPads support drag and drop on iOS 11.

src/js/vic-v2/components/PhotoField.jsx
cehsu: Thanks @jbalboni, I've adjusted the slider configuration (namely the default value) to address this bug.

src/js/vic-v2/components/PhotoField.jsx
cehsu: Thanks @rianfowler, I'll include looking into this in the ticket for work after user testing.

src/js/vic-v2/components/PhotoField.jsx
cehsu: Good catch, updated.

src/js/vic-v2/components/PhotoField.jsx
cehsu: Good catch, it is not, it has been removed.

rianfowler (approved):

src/js/vic-v2/components/PhotoField.jsx
cehsu: Removing this breaks the cropbox style setting logic. I'll include looking into alternative solutions in the ticket for work after testing.
This approach was recommended in cropBoxData issue #59 in the react-cropper repo.

src/js/vic-v2/components/PhotoField.jsx
cehsu: Good catch, @jbalboni, this has been changed.

PR #7101, opened 2018-01-16
VIC v2: Military and contact information

jbalboni:
Updated based on Invision prototype. One difference is that the branch is a text field. We will need to change that to a dropdown conditionally when we have a list of branches to choose from.

screen shot 2018-01-16 at 12 29 06 pm
screen shot 2018-01-16 at 12 28 50 pm


cehsu (approved):
LGTM.

PR #7104, opened 2018-01-16
Prescription pagination

bshyong:
- Adds support for pagination on Active prescriptions


U-DON (approved):
LGTM.

U-DON:
Unsure of the intended UX, but this works. I wonder if there are few enough active prescriptions at a time that we could just show all of them without bothering with a pagination component.

bshyong:
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4764

PR #7106, opened 2018-01-17
WIP/January link check 

peggygannon:
Updated broken link from monthly link check

### DIFF
--- a/content/pages/employment/job-seekers/less-than-one-year.md
+++ b/content/pages/employment/job-seekers/less-than-one-year.md
@@ -53,7 +53,7 @@ relatedlinks:
 
 ##### 6 Months Before Leaving Active Duty
 
-- Complete the [Preseparation Counseling Checklist](https://www.esd.whs.mil/Directives/forms/dd2500_2999/) (Form DD2648).
+- Complete the Preseparation Counseling Checklist found on Form DD2648.
 - Compile your service-treatment records and military hospital reports if you plan on registering for VA medical resources.
 Both of these can be ordered through the [National Archives](https://www.archives.gov/veterans/military-service-records/). Treatment records are not required for registration or treament at VA.
 - Compile private medical records and hospital reports.
--- a/content/pages/employment/job-seekers/less-than-one-year.md
+++ b/content/pages/employment/job-seekers/less-than-one-year.md
@@ -53,7 +53,7 @@ relatedlinks:
 
 ##### 6 Months Before Leaving Active Duty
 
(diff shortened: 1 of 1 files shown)


peggygannon:
@bethpotts January link check PR is ready for your review

bethpotts:
@peggygannon I got this scary message with that new path, so let's see if there are any other options...

screen shot 2018-01-18 at 4 42 02 pm

bethpotts:
@peggygannon Sure, let's do that. The chances that someone will even be on the less-than-one-year page are slim as I don't think we link to this page anymore from higher up employment pages.

peggygannon:
@bethpotts OK, I updated the link. Ready for your review again.

bethpotts:
@peggygannon Hmmm, I got the same error message as I pasted in above. 
@robertfairhead Can I get your opinion on this? We wouldn't want Veterans to have this experience, right? See the screenshot above. This is another one of those instances where I wonder if we just leave out a link altogether.

robertfairhead:
That error is because their server is misconfigured with a bad certificate.

The link is fine, it's just their server is insecure.

The only way to make that error not show up on Chrome would be for whoever owns that server to fix, nothing we do on our end.

Not all browsers will show that error message (which is why some of you may see it and others may not). Chrome tends to be the most aggressive about throwing that error in front of users.

For the updated link, if you make it `http://www.esd.whs.mil/Directives/forms/dd2500_2999/` that is to `http` not `https` it won't show the message, but it won't be encrypted.

robertfairhead:
Looking closer at that page it says

> If the form number does not have a hyperlink, the form is not available electronically. To obtain hard copies of current forms not available in electronic format, please contact your own Military Service or DoD Component Forms Management Officer. Cancelled forms are not available in electronic formats. Here is a list of Forms Management POCs.

And the `DD2648` is grayed out (and not downloadable from there). Maybe just removing the link and adding something like their warning would be best?

peggygannon:
@robertfairhead thanks, bob!

@bethpotts what do you think we should do. Should we remove the link and rewrite that sentence?

bethpotts:
@peggygannon Yeah, would you remove the link, rewrite, and send back to me for a review?

peggygannon:
@bethpotts OK, this is ready for you again to review.

bethpotts (approved):

bethpotts:
You can merge this, @peggygannon

peggygannon:
@phiden This PR is ready for your review and merge! Thank you!

PR #7110, opened 2018-01-17
Improve messaging when user is removed from a health care team

bshyong:
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/5723

Separates out notice box from health care team warning.
Adds use of AlertBox for better visibility when health care team is changed.


U-DON (approved):
LGTM.

PR #7112, opened 2018-01-17
Appeals V2 -- Docket

cvalarida:
It was a little awkward to get the "You are here" text to not run off the screen but keep it aligned with the marker when it's in the middle of the bar, but other than that, it really wasn't bad. I feel like I leveled up my CSS skillz with this one.

**Note:** This does not address the accessibility concerns on this component. As it stands, this will be kind of an awkward reading for a screen reader, I think. We should address that in a separate ticket.

## Once upon a time
image
The marker starts out uncomfortably far to the left, dipping into the left gutter. I'm not a huge fan of that, but the alternatives aren't great.

## Then we move along a bit
image
Notice how the text stays to the left, aligned with the text above.

## Middle of the pack
image
The text is centered on the marker.

## The end is near!
image
That's as far to the right as the text will go--don't want no cliff jumpers here!

## End of the line
image
According to @cmgiven, :point_up: should never happen, so I'm not going to worry about the gold line overlapping the marker...Unless I should, then it's a simple `z-index: 1`.

### DIFF
--- a/src/js/claims-status/components/appeals-v2/Docket.jsx
+++ b/src/js/claims-status/components/appeals-v2/Docket.jsx
@@ -1,13 +1,15 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
+// TODO: Depending on when this gets rendered, we may want to return null if either
+//  `ahead` or `total` don't exist (or if `total` === 0 because of the divide by 0  error)
 function Docket({ ahead, total }) {
   const completedWidth = { width: `${((total - ahead) / total) * 100}%` };
 
   // TODO: Assess how accessible this is
   return (
     <div className="docket-container">
-      <span className="appeals-ahead">{ahead.toLocaleString()}</span>
+      <p className="appeals-ahead">{ahead.toLocaleString()}</p>
       <p>Appeals ahead of you</p>
 
       <div className="marker-container">
--- a/src/js/claims-status/components/appeals-v2/Docket.jsx
+++ b/src/js/claims-status/components/appeals-v2/Docket.jsx
@@ -2,7 +2,7 @@ import React from 'react';
 
 class Docket extends React.Component {
   render() {
-    const { ahead, ready } = this.props;
+    const { ahead, total } = this.props;
 
     // TODO: Assess how accessible this is
     return (
@@ -10,7 +10,7 @@ class Docket extends React.Component {
         <span className="appeals-ahead">{ahead.toLocaleString()}</span>
         <p>Appeals ahead of you</p>
         <div className="docket-bar">
-          <div className="completed" style={{ width: `${ahead / ready}%` }}/>
+          <div className="completed" style={{ width: `${(ahead / total) * 100}%` }}/>
         </div>
       </div>
     );
--- a/src/js/claims-status/components/appeals-v2/Docket.jsx
+++ b/src/js/claims-status/components/appeals-v2/Docket.jsx
@@ -1,42 +1,39 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-class Docket extends React.Component {
-  render() {
-    const { ahead, total } = this.props;
-    const completedWidth = { width: `${((total - ahead) / total) * 100}%` };
+function Docket({ ahead, total }) {
+  const completedWidth = { width: `${((total - ahead) / total) * 100}%` };
 
-    // TODO: Assess how accessible this is
-    return (
-      <div className="docket-container">
-        <span className="appeals-ahead">{ahead.toLocaleString()}</span>
-        <p>Appeals ahead of you</p>
+  // TODO: Assess how accessible this is
+  return (
+    <div className="docket-container">
+      <span className="appeals-ahead">{ahead.toLocaleString()}</span>
+      <p>Appeals ahead of you</p>
 
-        <div className="marker-container">
-          <div>
-            <div className="marker-text-spacer" style={completedWidth}/>
-            <span className="marker-text">You are here</span>
-          </div>
-          <div>
-            <div className="spacer" style={completedWidth}/>
-            <div className="marker">
-              <img src="/img/Docket-line-pin.svg" alt=""/>
-            </div>
-          </div>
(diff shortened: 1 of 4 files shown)


gnakm:
💯 !

gnakm (approved):
lgtm!

test/claims-status/components/appeals-v2/Docket.unit.spec.jsx
```diff
@@ -0,0 +1,34 @@
...
+
+  it('should show the total number of appeals on the docket', () => {
+    const wrapper = shallow(<Docket {...defaultProps}/>);
+    expect(wrapper.find('.front-of-docket-text + p strong').first().text()).to.equal('283,941');
+  });
+
+  it('should show a visual representaton of where the appellant is in the line', () => {
+    const wrapper = shallow(<Docket {...defaultProps}/>);
+    const { ahead, total } = defaultProps;
+    const computedWidth = ((total - ahead) / total) * 100;
```
cvalarida: I debated putting in the computed value in here rather than copying the logic in `Docket`. What say you, @markgreenburg?

src/js/claims-status/components/appeals-v2/Docket.jsx
```diff
@@ -1,10 +1,45 @@
...
-  render() {
-    return (null);
-  }
+function Docket({ ahead, total }) {
+  const completedWidth = { width: `${((total - ahead) / total) * 100}%` };
+
+  // TODO: Assess how accessible this is
+  return (
+    <div className="docket-container">
+      <span className="appeals-ahead">{ahead.toLocaleString()}</span>
```
markgreenburg: Might be more semantic to use a heading or paragraph tag, since this is block-level as far as I can tell

src/js/claims-status/components/appeals-v2/Docket.jsx
```diff
@@ -1,10 +1,45 @@
...
+
+      <div className="marker-container">
+        <div>
+          <div className="marker-text-spacer" style={completedWidth}/>
+          <span className="marker-text">You are here</span>
+        </div>
+        <div>
+          <div className="spacer" style={completedWidth}/>
+          <div className="marker">
+            <img src="/img/Docket-line-pin.svg" alt=""/>
```
markgreenburg: No alt is intentional?

src/js/claims-status/components/appeals-v2/Docket.jsx
```diff
@@ -1,10 +1,45 @@
 import React from 'react';
+import PropTypes from 'prop-types';
 
-class Docket extends React.Component {
-  render() {
-    return (null);
-  }
+function Docket({ ahead, total }) {
```
markgreenburg: A bit paranoid probably, but what happens if we're missing either of these? We may want to just not show the component / have it return `null` or something since it wouldn't make much sense to show anything here with a number missing. 🤷‍♂️

src/js/claims-status/containers/AppealsV2StatusPage.jsx
```diff
@@ -25,7 +25,7 @@ const AppealsV2StatusPage = ({ appeal }) => {
         description={currentStatus.description}/>
       <Alerts alerts={alerts}/>
       <WhatsNext nextEvents={nextEvents}/>
-      <Docket/>
+      <Docket {...docket}/>
```
markgreenburg: Is anything other than `total` and `ahead` passed along in the `docket` attribute?

cvalarida: Yes actually. There's also `front`, `ready`, and `eta` which may or may not end up being used.

src/js/claims-status/components/appeals-v2/Docket.jsx
cvalarida: Good call. We currently don't have any logic surrounding _when_ to show the docket, but when we get that, I'm guessing it won't be a problem (probably that logic will say something like `if the current status is one of [these], and there's information in the docket object, render <Docket/>`).

I can make a comment for paranoia's sake for sure, and if our logic surrounding when to render the document doesn't involve checking `appeal.attributes.docket`, then we can do that check in here.

src/js/claims-status/components/appeals-v2/Docket.jsx
cvalarida: Yup. It'll make more sense to a screen reader to ignore the whole progress bar and marker, methinks, but we'll address that in a separate ticket.

src/js/claims-status/components/appeals-v2/Docket.jsx
cvalarida: :thinking: Yeah, I think I used a `<p>` at first, but it had a bothersome margin, so I swapped it out for a `<span>`. Since I took away all `<p>`s' margins in here, I can move it back to a `<p>`. Makes sense to me. :+1:

src/sass/claims-status.scss
```diff
@@ -948,3 +948,98 @@ h1:focus {
...
+  // This is slightly darker than the cards above it
+  background-color: $color-gray-light-alt;
+  padding: 16px;
+  margin: 3em 0px;
+
+  p {
+    margin: 0px;
+  }
+
+  span.appeals-ahead {
```
markgreenburg: is this selector needed for specificity or would a purely class-based one work?

test/claims-status/components/appeals-v2/Docket.unit.spec.jsx
markgreenburg: While I'm not a huge fan of replicating component code in the test, I think this is fine, it makes the purpose of test a little less mysterious IMO.

markgreenburg (approved):
A few comments but looks great! Excited to get this up on the Appeal Status page.

src/sass/claims-status.scss
cvalarida: Nah, I was just being specific. :man_shrugging:

PR #7114, opened 2018-01-18
Appeals V2 -- Missing events alert

cvalarida:
image
If you squint hard enough, you can overlook the fact that there probably _shouldn't_ be an incomplete history before "VBA sent the original claim decision to you."


gnakm (commented):
Boo! Sorry. I added this is another comment but not in this one. We're taking out all the alert icons in the alert UI for appeals status.

cvalarida:
Good call, @gnakm.

@markgreenburg already took out the icon in #7139, so I'll wait on that to get merged into master before tackling it.

src/js/claims-status/components/appeals-v2/Expander.jsx
```diff
@@ -13,13 +23,15 @@ const Expander = ({ title, dateRange, onToggle, cssClass, hideSeparator }) => {
         <h3 style={{ color: 'inherit' }}>{title}</h3>
       </button>
       <div className="appeal-event-date">{dateRange}</div>
+      {expanded && alert}
```
markgreenburg: Took me a bit to realize that `alert` doesn't _always_ show when the timeline is expanded, because it's either content or `null`. Might be worth doing something like 
`const alert = (missingEvents && expanded) ? missingEventsAlert : null;`
`return ( <stuff...>{alert}</stuff>)`
to keep all of the logic in the same place. Or not, I can't really decide 🤷‍♂️

cvalarida: I'll go through and see if I can clarify it.

cvalarida:
Updated screenshots:
image

Didn't affect the alert list:
image

gnakm (approved):
Looks great!

markgreenburg (approved):
Great!

PR #7116, opened 2018-01-18
Refactor Custom Date Range selector for better accessibility

bshyong:
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/6628#event-1428864897

screen shot 2018-01-19 at 11 07 14 am


src/js/health-records/containers/Main.jsx
```diff
@@ -225,35 +226,9 @@ export class Main extends React.Component {
...
-                  placeholderText="MM/DD/YYYY"
-                  selected={endDate}
-                  disabled={datePickerDisabled}
-                  minDate={startDate}
-                  maxDate={moment()}
-                  tetherConstraints={[{ to: 'scrollParent', attachment: 'none' }]}
-                  className={!datePickerDisabled && this.state.endDateError ? 'date-range-error' : ''}/>
-              </div>
+              <span id="vets-custom-date-range">Custom date range</span>
+              {/* The span class="usa-sr-only" is invisible to sighted users, but will offer additional instruction for screen readers */}
```
U-DON: Is this comment necessary? I think it should be evident already from the class name, even if it needs to be looked up.

src/js/health-records/components/ErrorableRadioButtons.jsx
```diff
@@ -0,0 +1,188 @@
...
+ * `name` - String for the name attribute.
+ * `toolTipText` - String with help text for user.
+ * `tabIndex` - Number for keyboard tab order.
+ * `options` - Array of options to populate group.
+ * `required` - is this field required.
+ * `value` - string. Value of the select field.
+ * `onValueChange` - a function with this prototype: (newValue)
+ */
+
+// NOTE: This component has been modified specifically for 508 compliance on the health records app
```
U-DON: What was the modification? Is it that it uses `children` to render anything nested within the component? That should be made clearer.

U-DON (changes requested):
I don't think this change works as intended.

> screen shot 2018-01-19 at 11 02 09 am

src/js/health-records/components/ErrorableRadioButtons.jsx
bshyong: Specifically if additional content like those date boxes need to be displayed, it cannot be nested in the label and should be outside of it instead. I'll make some changes here to get it working in the way it was intended

bshyong:
Updated. Easier to add an attribute to the common radiobox component that lets the options have additional content.

bshyong:
@U-DON can you do a re-review of this?

U-DON (approved):
LGTM.

PR #7119, opened 2018-01-18
VIC DD 214 document upload

jbalboni:
Will still need to be connected to api to upload files

screen shot 2018-01-18 at 1 35 54 pm


src/js/vic-v2/config/form.js
```diff
@@ -168,6 +171,26 @@ const formConfig = {
...
+        dd214Upload: {
+          path: 'documents/dd214',
+          title: 'DD214 upload',
+          depends: form => !form.verified,
+          editModeOnReviewPage: true,
+          uiSchema: {
+            'ui:description': DD214Description,
+            dd214: fileUploadUI('Upload your DD214', {
+              endpoint: '/v0/vic/vic_attachments',
+              fileTypes: ['pdf']
```
cehsu: Should this also include png, jgp?

cehsu (approved):
LGTM. Had a minor question.

src/js/vic-v2/components/DD214Description.jsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+
+export default function DD214Description() {
+  return (
+    <div>
+      <p>Please upload a copy of your DD 214 to verify that you were honorably discharged.</p>
```
cehsu: It looks like this is also spelled DD214 elsewhere on the site, should we use that spelling here as well?

PR #7121, opened 2018-01-18
Main nav back button fix

cvalarida:
Fixed the back button along with some general cleanup and even added in some focus management--now when a menu is closed, focus goes back to where it should. This is true even when escape is hit, which I'm pretty pleased about.

### DIFF
--- a/content/includes/veteran-navigation.html
+++ b/content/includes/veteran-navigation.html
@@ -8,13 +8,13 @@
             aria-controls="vetnav-explore"
             role="button"
             aria-haspopup="true"
-            class="vetnav-level1 vetnav-trigger"
+            class="vetnav-level1"
             onClick="window.dataLayer.push({ event: 'nav-header-top-level' });">Explore benefits</button>
           <div id="vetnav-explore" class="vetnav-panel" role="none" hidden>
             <ul role="menu" aria-label="Explore benefits">
               <li>
                 <button
-                  class="vetnav-level2 vetnav-trigger"
+                  class="vetnav-level2"
                   role="button"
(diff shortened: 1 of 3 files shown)


content/includes/veteran-navigation.html
```diff
@@ -8,13 +8,13 @@
             aria-controls="vetnav-explore"
             role="button"
             aria-haspopup="true"
-            class="vetnav-level1 vetnav-trigger"
```
cvalarida: Didn't know what this `.vetnav-trigger` did, so I did a little search. Didn't come up in `src/` anywhere (javascript or sass), and this was the only file in `content/` that it was found in, so I think it's safe to delete.

src/js/common/utils/accessible-menus.js
```diff
@@ -371,7 +372,10 @@ export default function addMenuListeners(menuElement, closeOnResize = false) {
     const inMenu = targetLi.parentElement.getAttribute('role').toLowerCase() === 'menu';
 
     // Handle opening (and closing) menus
-    if (inMenubar) {
+    if (e.target.classList.contains('back-button')) {
```
cvalarida: I wasn't sure if this should be handled in here or `mega-menu.js` since all other logic for small screen menus is in `mega-menu`. In the end, I decided to put it here for a couple reasons:
1. The helper functions were already here; why recreate the same functionality in `mega-menu.js`?
1. Other menus might have a back button to navigate back to a previous menu level

If anybody has strong opinions about this, I'd be happy to hear them.

ncksllvn (approved):
LGTM, great job!

PR #7122, opened 2018-01-18
Hide top-level menu when submenu is open on small screen

cvalarida:
## Small screen sub menus hide menu underneath
image

## Large screen is unaffected
image


ncksllvn (approved):
LGTM

PR #7124, opened 2018-01-19
Appeals V2 -- Past events content

cvalarida:
image


src/js/claims-status/utils/appeals-v2-helpers.jsx
```diff
@@ -194,43 +194,44 @@ export function getEventContent(event) {
...
       return {
-        title: 'VBA received your Notice of Disagreement',
+        title: 'VA received your Notice of Disagreement',
         description: '',
         liClass: 'section-complete'
       };
     case EVENT_TYPES.droHearing:
       return {
-        title: 'Dro Hearing',
+        // TODO: Verify property names these are correct
```
cvalarida: This is in fact still relevant: When Chris gets back to me [these questions](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/7894#issuecomment-358822837), I'll delete these comments.

...and probably even make sure we're grabbing the information from the right place if we're lucky. :man_shrugging:

markgreenburg: Might almost be worth just [PLACEHOLDER]ing them instead of guessing the property names / paths but not sure it makes a real difference.

src/js/claims-status/utils/appeals-v2-helpers.jsx
```diff
@@ -266,37 +268,38 @@ export function getEventContent(event) {
       };
     case EVENT_TYPES.bvaRemand:
```
markgreenburg: Not really 100% related to this PR but I noticed that the value for the `bvaRemand` is misspelled in the EVENT_TYPES map.

src/js/claims-status/utils/appeals-v2-helpers.jsx
markgreenburg: Just saw that Chris G mentioned the events won't have any fill-in-the-blanks, so not sure if this comment still applies.

cvalarida: Yeah, Chris said the content on these are going to change to not include the dynamic bits. I can [PLACEHOLDER] them, but they'll get changed anyhow. :man_shrugging:

cvalarida:
@gnakm @cmgiven Does it still make sense to merge this if the markdown is [going to be updated](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/7894#issuecomment-358966635) later?

markgreenburg (approved):
LGTM! Looks like content might change again though, so not sure whether this PR is still relevant.

src/js/claims-status/utils/appeals-v2-helpers.jsx
cvalarida: Oh, good catch!

cvalarida: Fixed https://github.com/department-of-veterans-affairs/vets-website/pull/7155

cvalarida:
@cmgiven updated the [markdown file](https://github.com/department-of-veterans-affairs/vets.gov-team/blob/master/Products/Appeals%20Status/V2/Design%20files/Events.md) which I then used to polish up this PR.

PR #7125, opened 2018-01-19
Revert dev to deploy master

robertfairhead:
Per Slack convo, removing dev being pinned to `user_feedback_footer_form-I3445` because it's no longer needed.


patrickvinograd (approved):
LGTM.

PR #7126, opened 2018-01-19
Update VIC chapters based on new flow

jbalboni:
screen shot 2018-01-19 at 11 28 40 am
screen shot 2018-01-19 at 11 28 25 am
screen shot 2018-01-19 at 11 28 00 am


lalexanderson-dsva:
This looks good. Two minor things, @jbalboni 
1) I can't tell if I'm being too picky at this point but can we change page title to "Apply for Veteran ID Card" 
2) Have we decided that we are taking out gender, or are we waiting on an answer from Lihan et al?

jbalboni:
1. Not at all. I changed that in the intro page PR, which isn't merged yet. I'll add it here, too.
2. According to him, we still need it to do the MVI check.

cehsu (approved):
LGTM.

PR #7132, opened 2018-01-22
Disable shutdown notification banner

robertfairhead:
This will disable the warning banner once the shutdown ends.

Should stay on hold until the VA is fully back to work and VA.gov removes their banner.


robertfairhead:
Closes https://github.com/department-of-veterans-affairs/vets.gov-team/issues/8349

jbalboni (approved):

PR #7136, opened 2018-01-22
Remove schemaform style dependency from letters

cvalarida:
The generated `letters.css` when from 1551 lines to 984, meaning 567 lines were removed. I'm not sure what the actual file size change is since neither my local dev environment nor staging has the stylesheets minified, but I think we can estimate about a 30% reduction in minified file size.

There's only one small visual change: There's a slightly smaller space between the title and subtitle content:
image

I could have fixed that, but it seemed small enough of a change, I didn't worry about it. If there needs to be that extra little bit of space, let me know, @goldenmeanie.

### DIFF
--- a/src/js/letters/components/StepHeader.jsx
+++ b/src/js/letters/components/StepHeader.jsx
@@ -5,14 +5,14 @@ class StepHeader extends React.Component {
   render() {
     const { current, steps, name } = this.props;
     return (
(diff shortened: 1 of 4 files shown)


markgreenburg (approved):
Makes sense to me, no reason to tie into schemaform if we're not using it in the app.

PR #7137, opened 2018-01-23
Photo cropper user feedback changes 

cehsu:
These changes fix a few bugs and make minor adjustments based partially off findings from user testing. 

- [x] update cropbox on window resize
- [x] fix medium breakpoint bug (determine styles on large screen size rather than medium)
- [x]  move picture when dragging on crop box (restyle dual border using a pseudo element and the provided line elements, rather than two pseudo elements)
- [x] restrict zoom to (0.2 -1.7 range) and when zooming beyond 1.3, provide warning that "zooming in this close will make your ID picture less clear"
screen shot 2018-01-24 at 11 38 09 am


src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -150,15 +147,60 @@ export default class PhotoField extends React.Component {
 
   onZoom = (e) => {
     const newZoomValue = e.detail.ratio;
-    this.setState({ zoomValue: newZoomValue }, () => {
-      this.refs.slider.value = this.state.zoomValue;
-    });
+    if (newZoomValue < MAX_RATIO && newZoomValue > MIN_RATIO) {
+      let warningMessage;
+      if (this.state.zoomValue >= WARN_RATIO) {
+        warningMessage = 'Zooming in this close will make your ID picture less clear.';
```
cehsu: @peggygannon is this warning content alright?

src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -29,14 +30,14 @@ const FILE_TYPES = [
   'jpg',
   'bmp'
 ];
+
 const MIN_SIZE = 350;
+const MIN_RATIO = 0.2;
+const MAX_RATIO = 1.7;
+const WARN_RATIO = 1.3;
```
rianfowler: Could we test this with images created by selfie cameras on phones and laptop cameras? I wouldn't mind doing the testing on a phone if it's just a matter of running it locally.

rianfowler (approved):

src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -29,14 +30,14 @@ const FILE_TYPES = [
...
 
-function detectDragAndDrop() {
-  const div = document.createElement('div');
-  const supportsDragAndDrop = ('draggable' in div) || ('ondragstart' in div && 'ondrop' in div);
-  const iOS = !!navigator.userAgent.match('iPhone OS') || !!navigator.userAgent.match('iPad');
-  const dragAndDropDetected = supportsDragAndDrop && window.FileReader;
-  return dragAndDropDetected && !iOS;
+function isSmallScreen(width) {
+  return  width < 1201;
```
jbalboni: Could make a constant for this number, too.

src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -150,15 +147,60 @@ export default class PhotoField extends React.Component {
...
 
+  detectWidthAndDrag = () => {
+    const windowWidth = window.innerWidth;
+    const div = document.createElement('div');
+    const supportsDragAndDrop = ('draggable' in div) || ('ondragstart' in div && 'ondrop' in div);
+    const iOS = !!navigator.userAgent.match('iPhone OS') || !!navigator.userAgent.match('iPad');
+    const dragAndDropSupported = supportsDragAndDrop && window.FileReader;
+    const previousWidth = this.state.windowWidth;
+    this.setState({ windowWidth, dragAndDropSupported: dragAndDropSupported && !iOS }, () => {
+      const newWidth = this.state.windowWidth !== previousWidth;
```
jbalboni: I wonder if the two setState callbacks would be better implemented in `componentWillReceiveProps`.

src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -150,15 +147,60 @@ export default class PhotoField extends React.Component {
...
   }
 
+
   getErrorMessage = () => {
     return this.state.errorMessage;
   }
 
+  detectWidthAndDrag = () => {
+    const windowWidth = window.innerWidth;
+    const div = document.createElement('div');
```
jbalboni: I looks like we're doing this feature detection on every resize and component re-render, even though only the screen width would be changing. You might pull the drag and drop detection into the constructor and only set it once.

jbalboni (approved):

src/js/vic-v2/components/PhotoField.jsx
cehsu: Thanks @jbalboni, great point, changed.

src/js/vic-v2/components/PhotoField.jsx
cehsu: Thanks @jbalboni, great point, changed.

src/js/vic-v2/components/PhotoField.jsx
peggygannon: @cehsu Can you make this tweak to the copy:
If you zoom in this close, your ID photo will be less clear.

src/js/vic-v2/components/PhotoField.jsx
cehsu: Excellent idea, but since componentWillReceiveProps won't be fired on setState, we can use componentWillUpdate to consolidate this logic instead. Changed.

src/js/vic-v2/components/PhotoField.jsx
cehsu: Thanks @peggygannon, updated.

PR #7139, opened 2018-01-23
Appealsv2 alert content mg

markgreenburg:
- Updates alert styling and design to latest version
- Updates supported alert types & content
- Updates some unit tests

### DIFF
--- a/src/js/claims-status/components/appeals-v2/Alert.jsx
+++ b/src/js/claims-status/components/appeals-v2/Alert.jsx
@@ -23,7 +23,7 @@ const Alert = ({ title, description, displayType }) => {
 
 Alert.propTypes = {
   title: PropTypes.string.isRequired,
-  description: PropTypes.object.isRequired, // nullable JSX snippet
+  description: PropTypes.element,
   displayType: PropTypes.string.isRequired,
 };
 
--- a/src/js/claims-status/components/appeals-v2/Alert.jsx
+++ b/src/js/claims-status/components/appeals-v2/Alert.jsx
@@ -1,9 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { getAlertContent } from '../../utils/appeals-v2-helpers';
 
-const Alert = ({ alert }) => {
-  const { title, description, cssClass } = getAlertContent(alert);
+const Alert = ({ title, description, cssClass }) => {
   return (
     <li>
       <div className={`usa-alert ${cssClass}`}>
@@ -17,10 +15,9 @@ const Alert = ({ alert }) => {
 };
 
 Alert.propTypes = {
-  alert: PropTypes.shape({
-    type: PropTypes.string.isRequired,
-    details: PropTypes.object
-  })
+  title: PropTypes.string.isRequired,
+  description: PropTypes.object.isRequired, // nullable JSX snippet
+  cssClass: PropTypes.object.isRequired,
 };
 
 export default Alert;
--- a/src/js/claims-status/components/appeals-v2/Alert.jsx
+++ b/src/js/claims-status/components/appeals-v2/Alert.jsx
@@ -1,13 +1,20 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-const Alert = ({ title, description, cssClass }) => {
+const Alert = ({ title, description, displayType }) => {
+  let cssClass;
+  if (displayType === 'take_action') {
+    cssClass = 'usa-alert-warning';
+  } else if (displayType === 'info') {
+    cssClass = 'usa-alert-info';
+  }
+
   return (
     <li>
       <div className={`usa-alert ${cssClass}`}>
         <div className="usa-alert-body">
           <h4 className="usa-alert-heading">{title}</h4>
-          <p className="usa-alert-text">{description}</p>
+          <div className="usa-alert-text">{description}</div>
         </div>
       </div>
     </li>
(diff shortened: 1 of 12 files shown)


markgreenburg:
FYI @gnakm content will have [PLACEHOLDERS] instead of actual dynamic content (e.g., dates, locations, etc.) since the 'details' that the v2 api will return haven't been finalized yet.

src/js/claims-status/components/appeals-v2/Alert.jsx
```diff
@@ -1,26 +1,30 @@
...
   );
 };
 
 Alert.propTypes = {
-  alert: PropTypes.shape({
-    type: PropTypes.string.isRequired,
-    details: PropTypes.object
-  })
+  title: PropTypes.string.isRequired,
+  description: PropTypes.object.isRequired, // nullable JSX snippet
```
cvalarida: If it's nullable, should it still be required? Also, does it make sense to use `PropType.element`?

markgreenburg:
This one doesn't make much sense without screenshots, does it?

screen shot 2018-01-23 at 10 38 05 am

screen shot 2018-01-23 at 10 38 33 am

screen shot 2018-01-23 at 10 38 25 am

screen shot 2018-01-22 at 8 25 07 pm

src/js/claims-status/components/appeals-v2/Alerts.jsx
```diff
@@ -1,21 +1,40 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import Alert from './Alert';
+import { getAlertContent } from '../../utils/appeals-v2-helpers';
 
-const Alerts = (props) => {
-  if (typeof props.alerts === 'undefined' || props.alerts.length === 0) {
+const Alerts = ({ alerts }) => {
```
cvalarida: While I'm thinking about it, what do you think about renaming this to `AlertList` to avoid confusion between `Alerts` and `Alert`?

Doesn't have to be done in this PR. I'm just wondering if that'd be helpful.

src/js/claims-status/components/appeals-v2/Alerts.jsx
```diff
@@ -1,21 +1,40 @@
...
+        title={alert.title}
+        description={alert.description}
+        displayType={alert.displayType}/>);
+    });
 
   return (
-    <ul className="alerts-list">
-      {alertsList}
-    </ul>
+    <div>
```
cvalarida: Won't it be nice when we don't have to wrap this in an unnecessary `<div>` just because you add a title?

src/js/claims-status/utils/appeals-v2-helpers.jsx
```diff
@@ -428,27 +435,27 @@ export function getNextEvents(currentStatus) {
...
+ * @param {alert} type each alert can have one of several types as defined by ALERT_TYPES
+ * ------------------------------------------------------------------------------------------------
+ * @typedef {object} alertOutput
+ * @property {string} title Used for the alert header
+ * @property {HTMLElement} description Some descriptive text for the alert body
+ * @property {string} displayType Segments the alert into either 'take action' or 'info' buckets
+ * @property {string} type Pass-through for the input type. Should be one of ALERT_TYPES
+ * ------------------------------------------------------------------------------------------------
+ * @param {alertInput} alert has some properties we match against to generate an alert's content
+ * @returns {alertOutput} dynamically-generated title, description, and styling properties
```
cvalarida: Woo! Look at that JSDoc! It makes me happy.

test/claims-status/components/appeals-v2/Alert.unit.spec.jsx
```diff
@@ -3,18 +3,41 @@ import { shallow } from 'enzyme';
...
 
-const defaultProps = {
-  alert: {
-    type: 'hearing-scheduled',
-    details: {
-      date: 'March th, 2018'
-    }
-  }
-};
+const defaultProps = {};
```
cvalarida: Does it still make sense to have this here?

cvalarida (approved):
Looks great! I had a couple questions, but nothing blocking. :+1:

gnakm (changes requested):
The content looks 👍 
The alert box padding bottom looks a little smaller than the top, left and right. Could you make that adjustment so they're all the same size?

gnakm:
@markgreenburg Also, another comment. Take action should be left-aligned with the alert box. Could you also make sure that it is H3 instead of H2?

image

markgreenburg:
@gnakm made both changes:

screen shot 2018-01-24 at 12 21 22 pm

markgreenburg:
@gnakm oh also, 'take action' is already an h3 using standard sizing. If you'd like to use a different size just let me know.

src/js/claims-status/components/appeals-v2/Alert.jsx
markgreenburg: Yeah, after some investigation, I'm updating this to an optional element proptype. Not sure why I was having issues with this earlier.

test/claims-status/components/appeals-v2/Alert.unit.spec.jsx
markgreenburg: Perhaps not, but wanted to be explicit about what was going on

markgreenburg:
@gnakm Think this one's good for a re-review!

gnakm (approved):

PR #7140, opened 2018-01-23
Update VIC front end to work with prefill data

jbalboni:
This updates the VIC front end to prefill data and set the service branch list correctly.

Main change to the form library I made is to add the ability to set a transform function that runs on migrated, prefilled data before the form itself loads. I'm using this to take the `serviceBranches` data and set `serviceBranch` and the enum list in the schema.

screen shot 2018-01-23 at 10 04 03 am

Only one branch is possible with this user:
screen shot 2018-01-23 at 10 03 39 am


src/js/common/schemaform/save-in-progress/RoutedSavableApp.jsx
```diff
@@ -102,7 +102,7 @@ class RoutedSavableApp extends React.Component {
       newProps.setFetchFormStatus(LOAD_STATUSES.notAttempted);
     } else if (newProps.prefillStatus !== this.props.prefillStatus
       && newProps.prefillStatus === PREFILL_STATUSES.unfilled) {
-      newProps.router.push(newProps.routes[this.props.routes.length - 1].pageList[0].path);
+      newProps.router.push(newProps.routes[this.props.routes.length - 1].pageList[1].path);
```
jbalboni: This is a bug I noticed when there's an error prefilling. We actually want to start the form, which is the second page (intro page is first).

src/js/common/schemaform/save-in-progress/reducers.js
```diff
@@ -123,6 +124,7 @@ export function createSaveInProgressInitialState(formConfig, initialState) {
     prefillStatus: PREFILL_STATUSES.notAttempted,
     isStartingOver: false,
     migrations: formConfig.migrations,
+    prefillTransform: formConfig.prefillTransform,
```
jbalboni: Not thrilled about keeping this in the Redux state, but the prop wiring is hard to follow otherwise.

rianfowler: If this is a function, may want to rename it if it's going on the state e.g. `prefillTransformer`. Otherwise this should be fine as long as we don't plan on serializing the state.

jbalboni: That makes sense. And yeah, we're not planning on serializing the state. There are other functions in there already and we only currently serialize the `data` part (everything else can be rehydrated from the config file or isn't something we'd save).

src/js/vic-v2/helpers.js
```diff
@@ -0,0 +1,19 @@
+import _ from 'lodash/fp';
+
+export function prefillTransform(pages, formData, metadata) {
+  if (formData && Array.isArray(formData.serviceBranches) && formData.serviceBranches.length) {
+    const newData = _.unset('serviceBranches', formData);
+    newData.serviceBranch = formData.serviceBranches[0];
+    return {
+      metadata,
+      formData: newData,
+      pages: _.set('veteranInformation.schema.properties.serviceBranch.enum', formData.serviceBranches, pages)
```
cehsu: Also, does this restrict the applicant's choice of branch? Would it make sense to simply list those at the top of the full list, rather than restrict the options in case the pre-filled data is incomplete (is this a consideration?) and the the preferred branch is not listed?

cehsu (approved):
LGTM. Had a few minor questions about presentation and data issues.

src/js/vic-v2/helpers.js
jbalboni: It does, it only allows you to choose from the ones that we've found in emis. As far as I know, that's what we want to do, but we can change it if we want. @stacyhlinn @lalexanderson-dsva

src/js/vic-v2/helpers.js
```diff
@@ -0,0 +1,19 @@
+import _ from 'lodash/fp';
+
+export function prefillTransform(pages, formData, metadata) {
+  if (formData && Array.isArray(formData.serviceBranches) && formData.serviceBranches.length) {
+    const newData = _.unset('serviceBranches', formData);
+    newData.serviceBranch = formData.serviceBranches[0];
```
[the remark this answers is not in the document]
jbalboni: I'm not sure there's a reasonable way for us to choose between branches for people. Most recent might be some kind of reserve service that they don't consider their primary branch (though I could be wrong there). There is helper text saying you can pick between multiple ones if you have them.

Do you know what the order is, @lihanli?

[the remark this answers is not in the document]
lihanli: they're in random order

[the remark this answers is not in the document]
cehsu: Thanks @jbalboni, that makes sense. In that case, additional message regarding why a particular branch is set as default doesn't seem helpful. The helper text explaining that multiple branches have been detected and encouraging the applicant to choose their preferred branch is great.

src/js/vic-v2/helpers.js
jbalboni: I'm going to go ahead and create a ticket for this question and merge this PR. It seems like a question that requires more thought and research.

PR #7143, opened 2018-01-23
Add support for GTM development environment

robertfairhead:
Closes: department-of-veterans-affairs/vets.gov-team#7750

@ncksllvn I assigned you for review since this is a general Nebula/Platform type thing. Let me know if you have questions.

Basically, when the `buildtype` is development we will load an alternate GTM configuration. This will let us deploy/test/demo configuration changes in our `https://dev.vets.gov` site without impacting production configuration.

Main use case is for training but could be leveraged for testing and experimentation as well.

### DIFF
--- a/content/includes/footer.html
+++ b/content/includes/footer.html
@@ -101,7 +101,9 @@ visit <a href="https://www.veteranscrisisline.net/">VeteransCrisisLine.net</a> f
 {% if buildtype === 'development' %}
   <script>
     window.dataLayer = [];
+    window.dataLayer.push({ environment: '{{ buildtype }}' });
   </script>
+  <script async src="/js/google-analytics-dev.js"></script>
 {% else %}
   <!-- Google Tag Manager -->
   <script>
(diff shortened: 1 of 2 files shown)


content/includes/footer.html
```diff
@@ -101,7 +101,9 @@ <h3>Get help from Veterans Crisis Line</h3>
 {% if buildtype === 'development' %}
   <script>
     window.dataLayer = [];
+    window.dataLayer.push({ environment: '{{ buildtype }}' });
```
ncksllvn: Is there any chance that top line should be `window.dataLayer = window.dataLayer || [];`?  I thought I read once that it's possible that it could have been initialized somewhere else with data already pushed to it.

ncksllvn (approved):
LGTM

content/includes/footer.html
robertfairhead: yeah that's true. we do it that way in the production snippet below for that reason.

for dev we were just lazier since dropped data didn't matter that much. i'll update it though as the consistency is a good idea.

PR #7145, opened 2018-01-23
Change baseline vic rate limit

patrickvinograd:
It may change again but would like to avoid the risk of busy people forgetting to reset this.


jkassemi (approved):
LGTM

PR #7149, opened 2018-01-24
WIP - add schema form and uiform elements for applicant page

rianfowler:
image

### DIFF
--- a/src/js/vre/chapter36/config/form.js
+++ b/src/js/vre/chapter36/config/form.js
@@ -40,9 +40,7 @@ const formConfig = {
   title: 'Apply for vocational counseling',
   subTitle: 'Form 28-8832',
   defaultDefinitions: {
-    fullName,
-    gender,
-    ssn
+    fullName
   },
   chapters: {
     applicantInformation: {
--- a/src/js/vre/chapter36/config/form.js
+++ b/src/js/vre/chapter36/config/form.js
@@ -1,19 +1,19 @@
-import _ from 'lodash/fp';
-
-// import { transform } from '../helpers';
 import fullSchema36 from 'vets-json-schema/dist/28-8832-schema.json';
 
 import IntroductionPage from '../components/IntroductionPage';
 import ConfirmationPage from '../containers/ConfirmationPage';
 
-import fullNameUI from '../../../common/schemaform/definitions/fullName';
+import ssnUI from '../../../common/schemaform/definitions/ssn';
 
 const {
-  applicantFullName
+  seekingRestorativeTraining,
+  seekingVocationalTraining,
+  receivedPamphlet
 } = fullSchema36.properties;
 
 const {
-  fullName
+  ssn,
+  gender
 } = fullSchema36.definitions;
 
 const formConfig = {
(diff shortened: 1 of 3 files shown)


src/js/vre/chapter36/config/form.js
```diff
@@ -32,60 +32,46 @@ const formConfig = {
...
   subTitle: 'Form 28-8832',
   defaultDefinitions: {
-    fullName
   },
   chapters: {
     applicantInformation: {
       title: 'Applicant Information',
+      initialData: {
+        socialSecurityNumber: '424242424'
+      },
```
rianfowler: Got this from the cookbook but couldn't get it to work.

src/js/vre/chapter36/config/form.js
```diff
@@ -32,60 +32,46 @@ const formConfig = {
...
-                'enum': [
-                  'Spouse',
-                  'Surviving spouse',
-                  'Child',
-                  'Stepchild',
-                  'Adopted child'
-                ]
-              }
+              socialSecurityNumber: ssn,
+              gender
```
rianfowler: The hca form had a more complicated property for gender: https://github.com/department-of-veterans-affairs/vets-website/blob/a5bb293bbc8fe6fb4fa04938e5cf4a2f2892c3b6/src/js/hca/config/form.js#L272-L276

My understanding is that this is mapping the keys between the form UI and fullSchema36 and doing some other validation related stuff. What's different about the hca form's approach and just using the definition from the full schema?

src/js/vre/chapter36/config/form.js
```diff
@@ -32,60 +32,46 @@ const formConfig = {
...
-              }
-            }),
-            applicantRelationshipToVeteran: {
-              'ui:title': 'What is your relationship to the Servicemember or Veteran?',
-              'ui:widget': 'radio',
-              'ui:required': formData => formData['view:isVeteran'] === false,
-              'ui:options': {
-                expandUnder: 'view:isVeteran',
-                expandUnderCondition: false
-              }
```
rianfowler: this stuff isn't in the mock up so I removed it.

src/js/vre/chapter36/config/form.js
rianfowler: I figured this out. `gender` in the `fullSchema36` are not user friendly strings.

test/vre/chapter36/config/applicantInformation.unit.spec.jsx
```diff
@@ -50,8 +58,9 @@ describe('VRE chapter 36 applicant information', () => {
     expect(form.find('.usa-input-error').length).to.equal(1);
     expect(onSubmit.called).to.be.false;
   });
+  */
 
-  it('should not submit without required applicant info', () => {
+  it.only('should not submit without required applicant info', () => {
```
rianfowler: Need some guidance on what's being tested here and how to set it up in the form. Currently I can continue without any of the questions answered. Are required fields necessary before **Continue** will work or before **Submit** works on the review page?

rianfowler: Need some guidance on what's being tested here and how to set it up in the form. Is input from required fields necessary before **Continue** or before **Submit** on the review page? Currently I can continue from this page with none of the questions answered.

src/js/vre/chapter36/config/form.js
jbalboni: Ah, I think that's something I mixed up and isn't clear in the issue. The page you're working on is the second page int he applicantInformation chapter. So the existing config in here doesn't need to change. Sorry about that!

test/vre/chapter36/config/applicantInformation.unit.spec.jsx
jbalboni: They are necessary in both cases. I think the confusing here is partly what I mentioned a second ago, that you're building another page, not modifying the first one.

src/js/vre/chapter36/config/form.js
cehsu: initialData should be a property on the specific page. [Here](https://github.com/department-of-veterans-affairs/vets-website/blob/7b33713a457c0f47175f244fbb0a22f93d4dfe18/src/js/edu-benefits/1995/config/form.js#L88) is an example. The documentation for this is inconsistent, thanks for pointing it out.

test/vre/chapter36/config/applicantInformation.unit.spec.jsx
cehsu: According to the design, the questions on this page aren't required, I'll invite @jbalboni and @stacyhlinn to chime in as well.

rianfowler: That makes sense. Looks like the test files in hca are organized around these pages so I'll move these to another file and restore the deleted tests.

src/js/vre/chapter36/config/form.js
```diff
@@ -88,30 +105,75 @@ const formConfig = {
...
-    contactInformation: {
-      title: 'Contact Information',
-      pages: {
+        },
+        nonVeteranInformation: {
+          title: 'Applicant information',
+          path: 'non-veteran-information',
+          depends: {
+            'view:isVeteran': false
+          },
```
rianfowler: Ticket says this is for non veterans only. Not sure what I should call this page since it wasn't in the list. @jbalboni suggestions?

rianfowler:
What is the purpose of this object? https://github.com/department-of-veterans-affairs/vets-website/blob/b85f39439f418d898bac2926bb88911654b2cda3/src/js/vre/chapter36/config/form.js#L34-L36

I read the documentation but I'm not sure how `fullName` works here- it's not referenced in the `schema` object.

jbalboni:
@rianfowler JSON Schema has it's own built in method for reusing common schema definitions. If you look at the schema for VIC in vets-json-schema's dist folder, you'll see that there's a definitions object that has "common" definitions like fullName. Those are then used later in the file for specific fields (veteranFullName in this case) by setting a `$ref` property.

The form config file lets you add those common definitions into that `defaultDefinitions` object and then it will merge those definitions into each page's schema, so you don't have to create a `definitions` object on a particular page's schema.

src/js/vre/chapter36/config/form.js
```diff
@@ -41,6 +49,13 @@ const formConfig = {
         applicantInformation: {
           title: 'Applicant information',
           path: 'applicant-information',
+          initialData: {
+            applicantFullName: {
+              first: 'testFirst',
+              last: 'testLast'
+            },
+            applicantRelationshipToVeteran: 'Spouse'
```
rianfowler: I was able to get this to work for a simple field but not for these.

rianfowler:
@jbalboni I think I'm starting to understand this. Looked at [this](http://json-schema.org/example2.html) again. Properties reference definitions and definitions contain properties. 

On VIC I see the `fullName` definition referenced on the `veteranFullName` property on the VIC-schema.json. My understanding is this: Given `veteranFullName` property references the `fullName` definition in the VIC-schema.json, when `veteranFullName` is referenced as a property in a `schema` object in ui file, then the `fullName` definition must be supplied to the `defaultDefinitions` object for the validation of `veteranFullName`

src/js/vre/chapter36/config/form.js
jbalboni: I can't really think of a better name, so this seems alright to me. We can have @stacyhlinn check it when she has time.

PR #7150, opened 2018-01-24
Add error tracking

cvalarida:
It doesn't solve the problem, but hopefully we can figure out what's causing it this way. It seems to reoccur every couple of days, so we should watch for it. I'll set up a slack reminder for myself so I don't forget to loop back around.


markgreenburg (changes requested):
This needs changes.

markgreenburg (approved):

cvalarida:
Closing this since it won't actually be useful. When we figure out the right way to move forward, we'll make a new one.

PR #7152, opened 2018-01-24
Fix development GTM JS snippet

robertfairhead:
I forgot to remove the surrounding `<script>` tag when setting up the separate snippet for dev. This should fix things.

doh


ncksllvn (approved):
LGTM!

PR #7157, opened 2018-01-25
7136 applicant info page 2

rianfowler:
2018-01-25-12-31-localhost_3001

### DIFF
--- a/src/js/vre/chapter36/config/form.js
+++ b/src/js/vre/chapter36/config/form.js
@@ -40,9 +40,7 @@ const formConfig = {
   title: 'Apply for vocational counseling',
   subTitle: 'Form 28-8832',
   defaultDefinitions: {
-    fullName,
-    gender,
-    ssn
+    fullName
   },
   chapters: {
     applicantInformation: {
--- a/src/js/vre/chapter36/config/form.js
+++ b/src/js/vre/chapter36/config/form.js
@@ -1,19 +1,19 @@
-import _ from 'lodash/fp';
-
-// import { transform } from '../helpers';
 import fullSchema36 from 'vets-json-schema/dist/28-8832-schema.json';
 
 import IntroductionPage from '../components/IntroductionPage';
 import ConfirmationPage from '../containers/ConfirmationPage';
 
-import fullNameUI from '../../../common/schemaform/definitions/fullName';
+import ssnUI from '../../../common/schemaform/definitions/ssn';
 
 const {
-  applicantFullName
+  seekingRestorativeTraining,
+  seekingVocationalTraining,
+  receivedPamphlet
 } = fullSchema36.properties;
 
 const {
-  fullName
+  ssn,
+  gender
 } = fullSchema36.definitions;
 
 const formConfig = {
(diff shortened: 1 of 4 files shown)


rianfowler:
LMK if there are other types of tests I need to write for this and who else I should add as a reviewer.

src/js/vre/chapter36/config/form.js
```diff
@@ -88,30 +99,72 @@ const formConfig = {
               }
             }
           }
+        },
+        nonVeteranInformation: {
+          title: 'Applicant information',
+          path: 'non-veteran-information',
```
rianfowler: LMK if I need to update this path. I've seen other patterns follow a `chapter/page` format.

jbalboni: @stacyhlinn Do you have a good name for this page? We can use that come up with the url.

src/js/vre/chapter36/config/form.js
```diff
@@ -88,30 +99,72 @@ const formConfig = {
...
+            receivedPamphlet: {
+              'ui:title': 'Have you received a pamphlet explaining survivors\' and dependents\' educational assistance benefits?',
+              'ui:widget': 'yesNo'
+            }
+          },
+          schema: {
+            type: 'object',
+            properties: {
+              applicantSocialSecurityNumber,
+              applicantGender: {
```
jbalboni: We try to use the properties from the schema as much as possible. In this case, you could import `applicantGender` from the schema properties or `gender` from the schema definitions and use it here. If you do the former, you'll have to put the `gender` definition in `defaultDefinitions`. 

Then, to set the labels, you can put them in the uiSchema for this field:

```js
applicantGender: {
  'ui:title': 'Gender',
  'ui:widget': 'radio',
  'ui:options': {
    labels: genders
  }
}
```

Our select and radio widgets support getting labels both from here and in the `enumNames` property of the schema.

src/js/vre/chapter36/config/form.js
```diff
@@ -41,6 +50,8 @@ const formConfig = {
         applicantInformation: {
           title: 'Applicant information',
           path: 'applicant-information',
+          initialData: {},
```
jbalboni: You could remove this.

jbalboni (approved):
LGTM, just had a couple small comments.

src/js/vre/chapter36/config/form.js
```diff
@@ -88,30 +99,72 @@ const formConfig = {
...
+            seekingRestorativeTraining: {
+              'ui:title': 'Are you a child who is at least 14 years old, a spouse, or a surviving spouse with a disability and looking for special restorative training?',
+              'ui:widget': 'yesNo'
+            },
+            seekingVocationalTraining: {
+              'ui:title': 'Are you a child, a spouse, or a surviving spouse with a disability and looking for special vocational training',
+              'ui:widget': 'yesNo'
+            },
+            receivedPamphlet: {
+              'ui:title': 'Have you received a pamphlet explaining survivors\' and dependents\' educational assistance benefits?',
```
cehsu: We tend to use curly apostrophes (’) in general.

src/js/vre/chapter36/config/form.js
cehsu: If only dependents can claim this (besides the Veteran), could we call this dependent information?

cehsu (commented):
LGTM. Pending the decision regarding the path, and the recommended changes to the gender radio widget.

cehsu (approved):

src/js/vre/chapter36/config/form.js
rianfowler: @stacyhlinn is out but I like the idea to update it to dependent-information and will do that. We can update it later if we need to.

PR #7161, opened 2018-01-25
Add date to docket

cvalarida:
image


src/js/claims-status/components/appeals-v2/Docket.jsx
```diff
@@ -1,17 +1,18 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import moment from 'moment';
 
 // TODO: Depending on when this gets rendered, we may want to return null if either
 //  `ahead` or `total` don't exist (or if `total` === 0 because of the divide by 0  error)
-function Docket({ ahead, total }) {
+function Docket({ ahead, total, form9Date }) {
   const completedWidth = { width: `${((total - ahead) / total) * 100}%` };
+  const date = moment(form9Date, 'YYYY-MM-DD').format('MMMM YYYY');
```
markgreenburg: What happens if `form9Date` is falsey?

src/js/claims-status/containers/AppealsV2StatusPage.jsx
```diff
@@ -17,6 +17,11 @@ const AppealsV2StatusPage = ({ appeal }) => {
   const { type, details } = status;
   const currentStatus = getStatusContents(type, details);
   const nextEvents = getNextEvents(type);
+  const form9Event = events.find(e => e.type === EVENT_TYPES.form9, null);
```
markgreenburg: Is it possible to get a form 9 more than once? I believe that's the case for some of the other events, because of the fact that multiple things can be on one appeal, and each thing can ping pong from one status to the other repeatedly.

markgreenburg (commented):
Couple 'o comments re. business rules that I think it'd make sense to talk through before approving but otherwise 👍

src/js/claims-status/containers/AppealsV2StatusPage.jsx
cvalarida: Oh, great point. I have no idea. @cmgiven?

src/js/claims-status/components/appeals-v2/Docket.jsx
cmgiven: That's right, without a Form 9, the docket progress bar will not be shown. But check out https://github.com/department-of-veterans-affairs/vets.gov-team/issues/8454. I'm going to add the month into the `docket` object.

src/js/claims-status/containers/AppealsV2StatusPage.jsx
cmgiven: Multiple Form 9s are possible. It's generally the earliest. But check out department-of-veterans-affairs/vets.gov-team#8454.

cvalarida: Awesome! I just made a comment in the code. I think we can leave this for now and fix it up properly in that ticket.

markgreenburg (approved):
LGTM!

PR #7163, opened 2018-01-25
Fix appeals v1 bug

cvalarida:
Fix some errant copy pasta.


bshyong (approved):
LGTM thanks for catching this

PR #7173, opened 2018-01-26
Closed appeals styling

cvalarida:
## Closed appeal styling
image

## Open appeal styling
image
No changes here; just showing that it didn't break anything.

### DIFF
--- a/src/js/claims-status/components/appeals-v2/CurrentStatus.jsx
+++ b/src/js/claims-status/components/appeals-v2/CurrentStatus.jsx
@@ -14,7 +14,8 @@ const CurrentStatus = ({ title, description, isClosed }) => (
 
 CurrentStatus.PropTypes = {
   title: PropTypes.string.isRequired,
-  description: PropTypes.element.isRequired
+  description: PropTypes.element.isRequired,
+  isClosed: PropTypes.bool
 };
 
 export default CurrentStatus;
--- a/src/js/claims-status/components/appeals-v2/CurrentStatus.jsx
+++ b/src/js/claims-status/components/appeals-v2/CurrentStatus.jsx
@@ -1,14 +1,14 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-const CurrentStatus = ({ title, description }) => (
-  <div className="current-status">
+const CurrentStatus = ({ title, description, isClosed }) => (
+  <div className={`current-status ${isClosed ? 'closed' : ''}`}>
     <h2>Current Status</h2>
     <div className="current-status-content">
       <h3>{title}</h3>
       <div>{description}</div>
     </div>
-    <div className="down-arrow"/>
+    {!isClosed && <div className="down-arrow"/>}
   </div>
 );
 
--- a/src/js/claims-status/containers/AppealsV2StatusPage.jsx
+++ b/src/js/claims-status/containers/AppealsV2StatusPage.jsx
@@ -26,7 +26,8 @@ const AppealsV2StatusPage = ({ appeal }) => {
       <Timeline events={events}/>
       <CurrentStatus
         title={currentStatus.title}
-        description={currentStatus.description}/>
+        description={currentStatus.description}
+        isClosed={appealIsClosed}/>
       <AlertsList alerts={alerts}/>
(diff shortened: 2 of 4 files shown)


src/js/claims-status/components/appeals-v2/CurrentStatus.jsx
```diff
@@ -1,14 +1,14 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-const CurrentStatus = ({ title, description }) => (
-  <div className="current-status">
+const CurrentStatus = ({ title, description, isClosed }) => (
```
markgreenburg: Probably want a corresponding `isClosed` entry in `PropTypes`

src/js/claims-status/utils/appeals-v2-helpers.jsx
```diff
@@ -29,6 +29,17 @@ export const STATUS_TYPES = {
   otherClose: 'other_close',
 };
 
+export const CLOSED_STATUSES = [
```
markgreenburg: hah, constants depending on constants. You can't explain that.

src/sass/claims-status.scss
```diff
@@ -663,10 +665,21 @@ h1:focus {
 }
 
 .current-status {
+  &.closed {
+    .current-status-content {
+      border-left-color: transparent;
+
+      &::before {
+        background-color: $color-inactive;
```
markgreenburg: TIL we have a `$color-inactive`

src/js/claims-status/containers/AppealsV2StatusPage.jsx
```diff
@@ -17,15 +17,21 @@ const AppealsV2StatusPage = ({ appeal }) => {
...
       <CurrentStatus
         title={currentStatus.title}
-        description={currentStatus.description}/>
+        description={currentStatus.description}
+        isClosed={appealIsClosed}/>
       <AlertsList alerts={alerts}/>
-      <WhatsNext nextEvents={nextEvents}/>
-      <Docket {...docket}/>
+      {!appealIsClosed && <WhatsNext nextEvents={nextEvents}/>}
+      {!appealIsClosed && <Docket {...docket}/>}
```
markgreenburg: Might make sense to throw `WhatsNext` and `Docket` into one conditional since they both render under the same condition, but this is pretty explicit so 🤷‍♂️

src/sass/claims-status.scss
```diff
@@ -1048,3 +1061,13 @@ $marker-text-width: 9rem;
     }
   }
 }
+
+.closed-appeal-notice {
+  line-height: 40px;
+  text-align: center;
+  font-family:  "Source Sans Pro", sans;
```
markgreenburg: I don't know if this is specifically specced anywhere, but I've seen hierarchy used in some `/common` components:
`"Source Sans Pro", "Helvetica Neue", "Helvetica", "Roboto", "Arial", sans-serif;`

markgreenburg (approved):
LGTM! Would be great to get one little thingy added to PropTypes but that's about it.

src/sass/claims-status.scss
cvalarida: Nope. I made it a thing in this file only because I couldn't find that exact color code anywhere else. Not sure what the best practice here is. I figure we've got three ways to deal with it:
1. Change the design so the color matches an existing color
2. Add the color to our existing color variable collection in `_b-variables.scss` (or whatever it is)
3. Add it to just this file if it's not going to be used anywhere else

I just went with the last option. Dunno if that was right. :man_shrugging:

src/js/claims-status/containers/AppealsV2StatusPage.jsx
cvalarida: I was going to do that first, but I couldn't find an adequate variable name for them both, so I decided that might mean they shouldn't be coupled together so closely.

And like you said, this is pretty easy to figure out what's going on.

src/sass/claims-status.scss
cvalarida: Oh brilliant! I meant to revisit this to find what that hierarchy should be. Thanks!

PR #7175, opened 2018-01-29
Trigger uploadFile

cehsu:
These changes submit the photo via the preexisting `uploadFile` action.


src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -107,48 +100,60 @@ export default class PhotoField extends React.Component {
...
+    this.props.formContext.uploadFile(
+      this.state.cropResult.file,
+      filePath,
+      this.props.uiSchema['ui:options'],
+      this.updateProgress,
+      false
+    ).catch(() => {
+      // rather not use the promise here, but seems better than trying to pass
+      // a blur function
+      // this.props.onBlur(`${this.props.idSchema.$id}_${idx}`);
```
cehsu: Since we are restricting the file editing parameters, and the file types that can be uploaded for editing, file validation isn't really necessary here. Feedback on how best to communicate relevant errors to users is especially welcome.

src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -199,6 +204,13 @@ export default class PhotoField extends React.Component {
     return this.state.errorMessage;
   }
 
+  updateProgress = (progress) => {
+    if (progress === 100) {
```
cehsu: There isn't any progress indication in the designs, but we often provide some communication regarding this for applicants, is this only necessary for other types of files (and can be removed here)?

src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -247,12 +259,22 @@ export default class PhotoField extends React.Component {
...
   cropImage = () => {
-    const cropResult = this.refs.cropper.getCroppedCanvas().toDataURL();
-    this.setState({ cropResult });
+    const cropResult = {};
+    this.refs.cropper.getCroppedCanvas().toBlob(blob => {
+      const file = blob;
+      file.lastModifiedDate = new Date();
+      file.name = this.state.fileName;
+      cropResult.file = file;
+      cropResult.src = this.refs.cropper.getCroppedCanvas().toDataURL(blob.type);
```
cehsu: This is necessary to prevent the image from being resize per known issue of cropperjs library (as documented on their README).

src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -107,48 +100,60 @@ export default class PhotoField extends React.Component {
   }
 
   componentDidUpdate(prevProps, prevState) {
-    const newView = this.state.src && (prevState.src !== this.state.src || (prevState.done && !this.state.done));
+    const newView = this.state.src && (prevState.src !== this.state.src || ((this.state.progress === 100) && (this.state.progress < 100)));
```
rianfowler: `((this.state.progress === 100) && (this.state.progress < 100))` 
I think this will always evaluate to false.

src/js/vic-v2/components/PhotoField.jsx
rianfowler: This could also be rejected because of an network or backend error but neither of these rejections returns an error message. We could refactor this promise to return an error message and display it. e.g. "Check our network settings and try again" or go with something generic. Enabling retrying might require some rework, though. 

https://github.com/department-of-veterans-affairs/vets-website/blob/c73fce410fd64ad4d478cc087fa481a4fe4cdb6f/src/js/common/schemaform/actions.js#L178-L227

jbalboni: The normal way we handle errors with file fields is to set the error text in the file object in the form data (see the dispatch on line 204). Any error, networking or otherwise, will result in the component getting new props that include an error message.

src/js/vic-v2/components/PhotoField.jsx
jbalboni: I think we will want to have a progress bar if we need to do uploads and can't save this as base64 in the database. Should have an answer to that this week.

src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -107,48 +100,60 @@ export default class PhotoField extends React.Component {
...
 
   onEdit = () => {
-    this.setState({ done: false, warningMessage: null });
+    this.setState({ progress: 0, warningMessage: null });
   }
 
   onDone = () => {
-    this.setState({ done: true, warningMessage: null });
+    const filePath = 'data.photo';
+    this.props.formContext.uploadFile(
```
jbalboni: We don't know quite yet if we'll need to call this. Ryan B's working with Patrick to do load testing this week. If we can store images in the database as base64, then I think we'll want to remove this and call `onChange` with an object that's formatted like the other file fields. If not, we would keep this.

jbalboni (commented):
Think we'll have to wait a little bit on some outstanding questions for this PR.

src/js/vic-v2/components/PhotoField.jsx
cehsu: We have decided to present editing errors ("Zooming in this close will cause your ID photo to look unclear) using a warning style that is distinctive from our default error messaging. Should we make use of that pattern to communicate this kind of error as well?

src/js/vic-v2/components/PhotoField.jsx
cehsu: Thanks for the helpful context.

src/js/vic-v2/components/PhotoField.jsx
cehsu: Great catch, thanks @rianfowler .

src/js/vic-v2/components/PhotoField.jsx
cehsu: Thanks for the helpful context.

src/js/vic-v2/components/PhotoField.jsx
jbalboni: Probably, though you might want to confirm with Lauren and Stacy.

src/js/vic-v2/config/form.js
```diff
@@ -23,6 +24,7 @@ const {
   email,
   serviceBranch,
   dd214
+  // photo
```
cehsu: The photo schema needs to be updated to reflect the structure provided below. Should a file definition be created, and referenced by the photo schema?

src/js/vic-v2/components/PhotoField.jsx
cehsu: Here is a proposed progress presentation, if it turns out we decide that we want to communicate this to the applicant, @stacyhlinn @lalexanderson-dsva does this look alright?
screen shot 2018-01-29 at 10 04 08 pm
screen shot 2018-01-29 at 10 03 36 pm

rianfowler (approved):

src/js/vic-v2/config/form.js
jbalboni: There should be a file definition already (or we need to pull it out of the multiple files definition), we should update the schema to reference that, yeah.

src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -270,8 +288,24 @@ export default class PhotoField extends React.Component {
     }
   }
 
+  cropImage = () => {
+    const cropResult = {};
+    this.refs.cropper.getCroppedCanvas().toBlob(blob => {
+      const file = blob;
+      file.lastModifiedDate = new Date();
+      file.name = this.state.fileName;
```
cehsu: The file constructor isn't supported in IE or Edge, is this format acceptable, or do we need to use a [polyfill](https://github.com/moxiecode/moxie) to create the file to send back? @jbalboni @lihanli

src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -126,26 +119,48 @@ export default class PhotoField extends React.Component {
...
 
   onDone = () => {
-    const cropResult = this.refs.cropper.getCroppedCanvas().toDataURL(this.state.fileType);
-    this.setState({ cropResult, done: true, warningMessage: null });
+    const filePath = this.props.idSchema.$id.split('_').slice(1);
+    this.refs.cropper.getCroppedCanvas().toBlob(blob => {
+      const file = blob;
+      file.lastModifiedDate = new Date();
+      file.name = this.state.fileName;
+      this.props.formContext.uploadFile(
```
cehsu: The file constructor is not supported in IE or Edge, is this an appropriate format to send, or should we use a [polyfill](https://github.com/moxiecode/moxie) @lihanli @jbalboni ?

jbalboni: If that uploads successfully, then yeah, I think that's fine.

Judging by https://stackoverflow.com/questions/33821631/alternative-for-file-constructor-for-safari, what you're doing should work.

src/js/common/schemaform/validation.js
```diff
@@ -343,6 +343,32 @@ export function validateDateRange(errors, dateRange, formData, schema, errorMess
   }
 }
 
+export function validateFile(errors, file) {
```
jbalboni: Can we reuse this in validateFileField?

src/js/common/schemaform/actions.js
```diff
@@ -207,6 +208,7 @@ export function uploadFile(file, filePath, uiOptions, progressCallback) {
             errorMessage
           }, getState().form.data))
         );
+        if (errorCallback) errorCallback('Check our network settings and try again');
```
jbalboni: I don't think these changes are necessary. The error will be in the form data prop that the field component is passed.

cehsu: Thanks @jbalboni what is the best way to access that instead?

src/js/common/schemaform/validation.js
cehsu: The current function being used in the validateFileField map is quite close but slightly different:
```
export function validateFileField(errors, fileList) {
  fileList.forEach((file, index) => {
    let error;
    if (file.errorMessage) {
      error = file.errorMessage;
    } else if (file.uploading) {
      error = 'Uploading file...';
    } else if (!file.confirmationCode) {
      error = 'Something went wrong...';
    }

    if (error && !errors[index]) {
      /* eslint-disable no-param-reassign */
      errors[index] = {
        __errors: [],
        addError(msg) {
          this.__errors.push(msg);
        }
      };
      /* eslint-enable no-param-reassign */
    }

    if (error) {
      errors[index].addError(error);
    }
  });
}

```
In order to reuse this, would we want to write a wrapper function for this which would pass `errors` down, and update this function to accept an index argument, along with some associated conditional logic based on that as well?

src/js/vic-v2/config/form.js
```diff
@@ -160,17 +162,53 @@ const formConfig = {
...
+              endpoint: '/v0/vic/profile_photo_attachments',
+              fileTypes: [
+                'png',
+                'tiff',
+                'tif',
+                'jpeg',
+                'jpg',
+                'bmp'
+              ],
+              maxSize: 26214400,
```
cehsu: Should we store these magic numbers as constants somewhere?

rianfowler: Yes even if it's just naming them at the top of this file it makes their purpose / constraint / type clearer to maintainers.

src/js/common/schemaform/actions.js
jbalboni: The `formData` prop

src/js/common/schemaform/validation.js
jbalboni: In that case, maybe we should write a function for just the logic setting the error message.

src/js/vic-v2/config/form.js
cehsu: Thanks for the helpful idea @rianfowler, I've updated.

src/js/common/schemaform/actions.js
cehsu: Thanks for the reminder, updated.

src/js/common/schemaform/validation.js
cehsu: That is much cleaner, thanks, updated.

src/js/common/schemaform/validation.js
```diff
@@ -343,31 +343,39 @@ export function validateDateRange(errors, dateRange, formData, schema, errorMess
...
-    if (error) {
-      errors[index].addError(error);
-    }
+  if (error) {
+    errors[index].addError(error);
+  }
+}
+
+export function validateFile(errors, file) {
+  setError(file, 0, errors);
```
jbalboni: I don't think this will work, there's no need for an index if the field is a regular object. You would want to call `errors.addError` for this field to set the message correctly.

I was thinking that this could be the reusable function:

```js
export function getFileError(file) {
  if (file.errorMessage) {
    return file.errorMessage;
  } else if (file.uploading) {
    return 'Uploading file...';
  } else if (!file.confirmationCode) {
    return 'Something went wrong...';
  }

  return null;
}
```

Then you can leave validateFileField the way it was before, except for using that function to replace the above logic.

And for the validation for a single file field, I think that's specific to the photo field right now, so I would take it out of this file and put it in a validation.js file in the vic-v2 folder. You can have it call the getErrorMessage function that would be in here. The square image validation would go in that new file, too, but that can be left to the other a11y ticket, I think.

src/js/vic-v2/config/form.js
```diff
@@ -33,6 +35,8 @@ const {
   gender
 } = fullSchemaVIC.definitions;
 
+const twentyFiveMB = 26214400;
```
jbalboni: Might be better as TWENTY_FIVE_MB, to follow our common constant pattern.

cehsu: Thanks @jbalboni for the helpful guidance (updated).

src/js/common/schemaform/validation.js
cehsu: Thanks @jbalboni that's a great idea (updated).

src/js/common/schemaform/validation.js
```diff
@@ -343,16 +343,21 @@ export function validateDateRange(errors, dateRange, formData, schema, errorMess
   }
 }
 
+export function getFileError(file) {
+  if (file.errorMessage) {
+    return file.errorMessage;
+  } else if (file.uploading) {
+    return 'Uploading file...';
```
cehsu: It looks like we don't currently display this message, is it worth setting?

cehsu:
Thanks @jbalboni and @rianfowler, the recommended updates have been made.

PR #7177, opened 2018-01-29
Turn on sip for pre-need

jbalboni:
We never turned this on on the front end.

screen shot 2018-01-29 at 11 22 03 am
screen shot 2018-01-29 at 11 21 51 am
screen shot 2018-01-29 at 11 21 40 am


rlbaker (approved):

cehsu (approved):
LGTM.

PR #7178, opened 2018-01-29
Jenkins build performance - Dockerfile improvements

wyattwalter:
We aren't benefitting from having vets-website directory synced into the running container since this app is not distributed via a Docker image, and the resulting I/O from the copy/chown/chmod is maxing out disks plus adding extra layers for unionfs to resolve. This modifies the behavior of Jenkins to use a volume for all of `/application` during the build/test phase, which reduces run time by roughly 4 minutes on average.

Ref: https://github.com/department-of-veterans-affairs/vets.gov-team/issues/8407

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -60,8 +60,9 @@ RUN \
     echo "===> install Java"  && \
     echo debconf shared/accepted-oracle-license-v1-1 select true | debconf-set-selections  && \
     echo debconf shared/accepted-oracle-license-v1-1 seen true | debconf-set-selections  && \
-    DEBIAN_FRONTEND=noninteractive  apt-get install -y --force-yes libnss3 libgconf-2-4 libxss1 libxtst6 libgtk2.0-0 libasound2 xvfb netcat oracle-java8-installer oracle-java8-set-default
-
-RUN mkdir -p /application && chown jenkins:jenkins /application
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -60,28 +60,9 @@ RUN \
...
-# Install required npm dependencies
-
-COPY package.json .
-COPY yarn.lock .
-COPY .yarnrc .
-# skips all dev dependencies if NODE_ENV=production.. so..
-RUN yarn install --production=false
-
-# Copy application source to image
+    DEBIAN_FRONTEND=noninteractive  apt-get install -y --force-yes libnss3 libgconf-2-4 libxss1 libxtst6 libgtk2.0-0 libasound2 xvfb netcat oracle-java8-installer oracle-java8-set-default && \
```
edpaget: Setting this env var is actively discouraged by docker: https://docs.docker.com/engine/faq/ unless it's causing an error with the build I think we should drop it.

patrickvinograd (approved):
LGTM. Makes sense.

edpaget:
Also why do we download node instead of using a [node docker image](https://hub.docker.com/_/node/)?

Dockerfile
patrickvinograd: Since we're not exporting this as an image for any downstream use the concern in that FAQ does not really apply. I'm not opposed to removing it, or using a node docker image, but given how many vets-website builds happen I'm in favor of incremental improvements and maybe a separate ticket (if there's not one already) for overall build optimization.

wyattwalter:
@edpaget both great suggestions, thanks! I'm trying to time-box this work a bit, and am going to merge what I have so far as it's an improvement on build times already. The image creation stuff is up next! Happy to have you jump in if you'd like!

PR #7181, opened 2018-01-29
Schemaform field documentation

jbalboni:
Trying to add some more documentation for the vets.gov specific fields we have.

### DIFF
--- a/docs/schemaform/field-library.md
+++ b/docs/schemaform/field-library.md
@@ -1,8 +1,8 @@
 # Field library
 
-We have many common field and widgets that are available for use in forms. Stick with these common fields and definitions if at all possible.
+We have many common field and widgets that are available for use in forms. You should stick with these common fields and definitions if at all possible.
 
-There are generally two forms of common definitions: schema/uiSchema objects and functions that return schema/uiSchema objects. For the function versions, there will be documentation below for the parameters
+There are generally two forms of common definitions: schema/uiSchema objects and functions that return schema/uiSchema objects. For the function versions, there will be documentation in the fields for the parameters.
 
 ## Widgets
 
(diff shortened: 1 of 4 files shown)


docs/schemaform/field-library.md
```diff
@@ -0,0 +1,163 @@
...
+
+This is for our common EFT information field, where we collect account type (checking/savings), bank account number, and routing number.
+
+### Currency
+- File: src/js/common/schemaform/definitions/currency.js
+- uiSchema: Yes
+- schema: No
+- Function or object: Function
+
+### Current or past dates
```
cehsu: Is it beyond the scope to mention that we also validate by specific ranges?

cehsu (approved):
LGTM. This is really great. Would it make sense to list any libraries we use with this such as Downshift? Are strategies we've developed, such as prefilling certain fields/defaulting to zero, within the scope of this documentation?

rianfowler (approved):
This is useful.

docs/schemaform/field-library.md
jbalboni: I'm not sure what you mean

cehsu: For example, veterans need to be born at least 15 years prior to the current date for certain forms.

docs/schemaform/field-library.md
```diff
@@ -0,0 +1,163 @@
...
+
+Widget       | Default schema type or format
+------------ | -------------
+TextWidget | type: string
+SelectWidget | type: string with an enum property
+RadioWidget |
+CheckboxWidget | type: boolean
+yesNo |
+
+There are more widgets that just these, but we normally use definitions they're included in, rather than just the field. These are the ones you're likely to have to set at some point in a config file.
```
annekainicUSDS: Small typo: "There are more widgets **than** just these"

annekainicUSDS (approved):
LGTM!

docs/schemaform/field-library.md
jbalboni: I think those are mostly form specific (and we have a pattern in here for personId), so I think we can leave them out.

PR #7184, opened 2018-01-30
Hide top-level menu on iOS Safari

cvalarida:
@markgreenburg and I tested on macOS Safari and this bug wasn't present, but it is on iOS Safari. I was able to reproduce locally.

While testing, I noticed the sub-menu didn't cover everything underneath it, so I made a little tweak to use `100vh` instead of `100%`, and that somehow managed to fix what I thought was a transparency problem.

To be honest, I don't know _why_ this works, but my theory goes like this: In all other (sane) browsers, the `<ul>` has a height of the sum of its parts, but in iOS Safari, it doesn't. This means that the background color (applied to the `<ul>`) isn't covering up the menu underneath it.

This PR isn't a complete fix to everything, but it does fix the problem in the connected issue.

If the viewport height < first-level menu height, the first-level menu will show up below the second-level menu. This is because the second-level menu has a `min-height` of `100vh`, not `100%` of the `<ul>` above it in the DOM (great-grandparent). Changing this programmatically is painful because the breakpoints aren't on the javascript side (intentionally) and it behaves completely different on small screens vs large screens.

Too many words! More screenshots, please!
Okay, okay, I'll just show you!
image

### DIFF
--- a/content/includes/veteran-navigation.html
+++ b/content/includes/veteran-navigation.html
@@ -3,7 +3,7 @@
     <nav id="vetnav" role="navigation" hidden>
       <ul id="vetnav-menu" role="menubar">
         <li><a href="/" class="vetnav-level1" role="menuitem">Home</a></li>
-        <li role="none">
+        <li>
           <button
             aria-controls="vetnav-explore"
             role="button"
--- a/src/js/common/utils/accessible-menus.js
+++ b/src/js/common/utils/accessible-menus.js
@@ -200,9 +200,6 @@ function openMenu(menuLi, menuStructure = null, openSubMenu = false, stealFocus 
   const parentMenu = findNearestAncestor(menuButton, el => el.getAttribute('role') === 'menu');
   if (parentMenu) {
     parentMenu.classList.add('child-menu-opened');
-    console.log('parentMenu found:', parentMenu);
-  } else {
-    console.log('parentMenu not found');
   }
 
   if (stealFocus) {
--- a/src/js/common/utils/accessible-menus.js
+++ b/src/js/common/utils/accessible-menus.js
@@ -156,6 +156,13 @@ function closeMenu(menuLiOrStruct) {
 
   const { menuButton, menu } = struct;
 
+  // Tell the parent menu (if any) that this one is open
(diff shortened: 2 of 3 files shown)


cvalarida:
@mnorthuis Can you try out the main nav again to see if it's fixed in [this Heroku build](https://vetsgov-pr-7184.herokuapp.com/)?

(Note: If you view it on your desktop, the nav bar is going to look _waaaay_ too big. I fixed that, but it isn't being reflected in Heroku for some reason.)

mnorthuis:
The "transparent" background issue appears to be fixed, but you can still see the menu behind it if it isn't a long menu.  Is there a way to set the vertical height so the blue background always covers up the menu behind?  Find a Cemetery is the last option in the top menu in the screen shot...

img_1735

cvalarida:
@mnorthuis Thanks for testing! Glad to see the transparency thing is fixed!

That menu overlap bug is what I was talking about about in the original comment. I thought a lot about how to fix that, but I don't know that there's a good way to do it because it behaves completely differently on a large screen. Insert technobabble here, the underlying DOM structure isn't set up in a way that we can make it work well on both desktop and mobile.

If I remember right, there was some whispers about rethinking the whole nav menu. Do you know anything about that?

At any rate, I really tried to get the child menu to block the whole parent menu, but I wasn't able to make it happen. I'll write up another issue for this.

mnorthuis:
Top nav redesign is only chatter right now, but nothing officially on the books.  Do you want the "transparent background" issue tested on other devices? I can tag others in the other issue if you are ready for that...

cvalarida:
Yeah, that'd be awesome! I think it's ready for wider testing.

This part always makes me nervous.

ncksllvn (approved):
LGTM for the scope of this PR but something does seem a little wonky with the scrolling - if you open the first level and try scrolling down it seems to close prematurely. The height sometimes looks a little funny too. 

image

cvalarida:
Hmm...that looks like it doesn't get the `calc(100vh...)` for `$medium-screen`, perhaps? We should really enumerate all the bugs in this nav and make a separate ticket for them. It's so........touchy. Specific. Wonky.

PR #7186, opened 2018-01-30
Update FrontEndTests101.md

annekainicUSDS:
Updated docs to reflect additional conversational during front-end sync up.

### DIFF
--- a/docs/FrontEndTests101.md
+++ b/docs/FrontEndTests101.md
@@ -66,9 +66,7 @@ Write tests as you go! New functionality added in a PR should ideally be covered
 6. Mocks, spies, stubs
 - **sinon**: test spies, stubs, and mocks
 
-7. Code coverage reports
-
-8. Browser environment
+7. Browser environment
 - **chromedriver**: a server that links the browser (Chrome) to the test driver (Selenium), enabling you to execute actions in the browser through a test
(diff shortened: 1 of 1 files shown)


jbalboni:
@annekainicUSDS One thing we could mention is how to run tests with code coverage (`npm run test:coverage`). That is sometimes useful.

I think we're going to turn off code climate, but I think it would be great to have a tool like that focused only on code coverage, so we could see how well we were doing writing tests. Something like Coveralls.

annekainicUSDS:
@jbalboni thanks for the suggestion! added some notes on the code coverage thing

ncksllvn:
Great work @annekainicUSDS! Just an idea, but do you think we should put notes from these meeting in their own directory within `docs/`? Our docs are super useful but kind of hard to navigate at the moment.

annekainicUSDS:
@ncksllvn totally agree about needing to clean up the organization of our docs folder. I think that should be a separate task to review things in total and figure out a new organization, and then also rework the README and create a table of contents linking to things.

ncksllvn:
@annekainicUSDS Perfect, that's a great plan! Thanks!

annekainicUSDS:
@ncksllvn do you mind approving so I can merge? Thanks!

ncksllvn (approved):

ncksllvn:
@annekainicUSDS whoops sorry, thought I already did!

PR #7187, opened 2018-01-30
Appealsv2 issues mg

markgreenburg:
Adds an Issues component to the Details tab of an individual appeal
- Issues are segmented into either an 'open' or 'closed' issues accordion
- The 'open' accordion has two sublists: 'open' (there's no header for this one), and 'remand'
- The 'closed' accordion has two sublists: 'granted' and 'denied'
- Both the accordions and the sublists are conditionally rendered, so e.g. the 'closed' accordion will only render if there is a 'closed' issue, and the 'granted' sublists will only render if there is a 'granted' issue.

Still To Do:
- [x] More thorough render case unit tests
- [x] Style to match mocks in connected issue

We probably also want to look at modifying the standard accordion component to accept an extra prop or two to let us auto-expand it on render without the user having to click a button, but I'd like to segment that off into its own PR (tracking issue in issue [#8453](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/8453)

Lastly, we might want to give the 'Issues' header a little more space (esp. on the top margin), but I left it this way to line up with the margins on the 'Status' tab - we can change both of these in another PR if needed.

**Update**

The below screenshots are old in the sense that they don't have a separate 'withdrawn' section, but here's a new screenshot of the fact that the 'withdrawn' section is now broken out from granted / denied within the 'closed' accordion:
screen shot 2018-01-31 at 10 55 48 am

screen shot 2018-01-30 at 6 25 04 pm

screen shot 2018-01-30 at 6 03 42 pm

screen shot 2018-01-30 at 6 03 36 pm

screen shot 2018-01-30 at 6 03 17 pm

screen shot 2018-01-30 at 6 03 09 pm

screen shot 2018-01-30 at 6 02 58 pm


src/js/claims-status/utils/appeals-v2-helpers.jsx
```diff
@@ -43,6 +52,55 @@ export const ALERT_TYPES = {
...
+ */
+export function categorizeIssues(issues) {
+  return issues.map((issue) => {
+    let status = '';
+    switch (issue.lastAction) {
+      case ISSUE_STATUS.fieldGrant:
+        status = 'granted';
+        break;
+      case ISSUE_STATUS.withdrawn:
+        status = 'denied';
```
cmgiven: Actually, we need a withdrawn section, which should go underneath the granted and denied sections. The rest of this mapping is correct.

cmgiven:
Looks good.

* Occurs to me that if there is no open accordion, the closed accordion should default to expanded. Basically, the logic is that the first accordion should be expanded, whatever it is.
* I don't think we've given you a mapping for the "Board decision, May 2015" text. Here's what that should be:

`allowed` => "Board decision"
`denied` => "Board decision"
`remand` => "Board decision"
`cavc_remand` => "Court decision"
`field_grant` => "VBA decision"
`withdrawn` => "Withdrawn"

cmgiven (changes requested):

src/js/claims-status/utils/appeals-v2-helpers.jsx
markgreenburg: Ahh, I was kind of unsure about whether 'withdrawn' is a separate section. Makes sense 👍

markgreenburg: See new screenshot, I think the 'withdrawn' section is good to go now.

cmgiven: :+1:

cmgiven (approved):

src/js/claims-status/components/appeals-v2/Issues.jsx
```diff
@@ -0,0 +1,97 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import CollapsiblePanel from '../../../common/components/CollapsiblePanel';
+
+const Issues = ({ issues }) => {
+  // TO-DO: think about whether it's crazy to run through the issues list 4 times
```
cvalarida: Nah. Given the relatively short list of issues, I think it's a micro-optimization to make this O(n) instead of O(4n). We're not talking hundreds of thousands of iterations, so this is good like it is--even for low-powered mobile phones, it shouldn't be an issue.

src/js/claims-status/components/appeals-v2/Issues.jsx
```diff
@@ -0,0 +1,97 @@
...
+  const open = issues.filter(i => i.status === 'open');
+  const remand = issues.filter(i => i.status === 'remand');
+  const granted = issues.filter(i => i.status === 'granted');
+  const denied = issues.filter(i => i.status === 'denied');
+
+  const getListItems = (item, i) => {
+    return (
+      <li key={`${item}-${i}`}>
```
cvalarida: `item` will be an object here, right? Did you mean `item.status`?

src/js/claims-status/components/appeals-v2/Issues.jsx
```diff
@@ -0,0 +1,97 @@
...
+      </div>
+    );
+  }
+
+  let grantedSection = null;
+  if (grantedListItems.length) {
+    grantedSection = (
+      <div>
+        {/* USWDS $color-green */}
+        <h5 style={{ color: '#2e8540' }}>Granted</h5>
```
cvalarida: I hate to make unnecessary classes, but in the event that this color is updated in later versions of USWDS, it'd be better if we didn't hard-code this into the javascript style.

On the other hand, is it that big of a deal? Probably not.

src/js/claims-status/components/appeals-v2/Issues.jsx
```diff
@@ -0,0 +1,97 @@
...
+        <h5 style={{ color: '#e31c3d' }}>Denied</h5>
+        <ul>{deniedListItems}</ul>
+      </div>
+    );
+  }
+
+  let activeItems = null;
+  if (openListItems.length || remandListItems.length) {
+    activeItems = (
+      <CollapsiblePanel panelName={'Currently On Appeal'}>
```
cvalarida: Is using title case here consistent with the rest of the site, @peggygannon?

src/js/claims-status/containers/AppealsV2DetailPage.jsx
```diff
@@ -1,11 +1,28 @@
...
+import PropTypes from 'prop-types';
+import Issues from '../components/appeals-v2/Issues';
+import { categorizeIssues } from '../utils/appeals-v2-helpers';
 
-const AppealsV2DetailPage = () => {
+const AppealsV2DetailPage = ({ appeal }) => {
+  const issues = categorizeIssues(appeal.attributes.issues);
   return (
     <div>
-      Hey I'm the details page!
```
cvalarida: You shall be missed.

src/js/claims-status/utils/appeals-v2-helpers.jsx
```diff
@@ -43,6 +52,55 @@ export const ALERT_TYPES = {
...
+ * @property {string} diagnosticCode a codified version of the description
+ * @property {('field_grant'|'withdrawn'|'allowed'|'denied'|'remand'|'cavc_remand')} lastAction
+ * @property {string} date TO-DO: unsure of what this date siginifies
+ * ------------------------------------------------------------------------------------------------
+ * @typedef {Object} segmentedIssue issue with descriptor and status information
+ * @property {('granted'|'remand'|'allowed'|'denied')} status lifecycle stage of an issue
+ * @property {string} description pass-through for the description info of passed in issue object
+ * ------------------------------------------------------------------------------------------------
+ * @param {issue[]} issues all the individual issues that are attached to an appeal
+ * @returns {segmentedIssue[]} an array of issue objects with stuses and descriptions
```
cvalarida: Typo: `statuses`

src/js/claims-status/utils/appeals-v2-helpers.jsx
```diff
@@ -43,6 +52,55 @@ export const ALERT_TYPES = {
...
+ */
+export function categorizeIssues(issues) {
```
cvalarida: This kind of seems like it would be returning all the issues whole, but categorized like
```js
{
  granted: [ ... ],
  denied: [ ... ],
  withdrawn: [ ... ]
}
```

The implementation you've got is good, but I'm wondering if this could use a rename.

src/js/claims-status/utils/appeals-v2-helpers.jsx
cvalarida: The new screenshot looks good, but I'm not seeing where that's at in the code. Did you push it up?

test/claims-status/components/appeals-v2/Issues.unit.spec.jsx
```diff
@@ -0,0 +1,59 @@
...
+    expect(wrapper.type()).to.equal('div');
+  });
+
+  it('should render no panels when no issues passed in', () => {
+    // Note: this probably isn't possible in real-world usage
+    const wrapper = shallow(<Issues {...emptyIssues}/>);
+    expect(wrapper.find('CollapsiblePanel').length).to.equal(0);
+  });
+
+  it('should render one panel when only an open *OR* a closed issue is passed in', () => {
```
cvalarida: Should we test both panels separately rather than testing one and assuming the other works too?

test/claims-status/utils/helpers.unit.spec.js
```diff
@@ -556,4 +557,19 @@ describe('Disability benefits helpers: ', () => {
...
+    it('returns an array of same length as input array', () => {
+      const { issues } = mockData.data[1].attributes;
+      const formattedIssues = categorizeIssues(issues);
+      expect(formattedIssues.length).to.equal(issues.length);
+    });
+
+    it('returns an array of objects with status and description', () => {
+      const { issues } = mockData.data[1].attributes;
+      const formattedIssues = categorizeIssues(issues);
+      expect(formattedIssues[0].status).to.exist;
```
cvalarida: Would it be better to test all the issues with something like the following?
```js
expect(formattedIssues.every(i => i.status && i.description)).to.be.true;
```

src/js/claims-status/components/appeals-v2/Issues.jsx
markgreenburg: 😬 yep

src/js/claims-status/components/appeals-v2/Issues.jsx
markgreenburg: It's not consistent with the mock, so guessing the answer here is 'no.' I've updated.

src/js/claims-status/containers/AppealsV2DetailPage.jsx
markgreenburg: 😢

src/js/claims-status/utils/appeals-v2-helpers.jsx
markgreenburg: Yeah, fair point. And maybe that _should_ be the implementation, but it feels ok to keep it as a lean mapping of issue statuses to display statuses.

As for the name, how about `addStatusToIssue()`?

src/js/claims-status/utils/appeals-v2-helpers.jsx
markgreenburg: Not only did I not push it up, but I can't find it locally. Not sure what I did to lose it all but it's pushed up now.

test/claims-status/utils/helpers.unit.spec.js
markgreenburg: That seems like a good idea. Instead of using something like ^ I'll use exactly ^ 😛

src/js/claims-status/utils/appeals-v2-helpers.jsx
cvalarida: Naming is hard. Turns out I can't think of a better one. :man_shrugging:

test/claims-status/components/appeals-v2/Issues.unit.spec.jsx
```diff
@@ -21,8 +22,13 @@ describe('<Issues/>', () => {
     expect(wrapper.find('CollapsiblePanel').length).to.equal(0);
   });
 
-  it('should render one panel when only an open *OR* a closed issue is passed in', () => {
-    const wrapper = shallow(<Issues {...oneIssues}/>);
+  it('should render one panel when only an open issue is passed in', () => {
+    const wrapper = shallow(<Issues {...oneOpenIssue}/>);
+    expect(wrapper.find('CollapsiblePanel').length).to.equal(1);
+  });
```
cvalarida: I hate to have you keep iterating on this, but can we make sure it's the right panel? Something like
```js
const panels = wrapper.find('CollapsiblePanel');
expect(panels.length).to.equal(1);
expect(panels.first().text()).to.contain('Currently on appeal');
```

test/claims-status/components/appeals-v2/Issues.unit.spec.jsx
```diff
@@ -21,8 +22,13 @@ describe('<Issues/>', () => {
...
+
+  it('should render one panel when only a closed issue is passed in', () => {
+    const wrapper = shallow(<Issues {...oneClosedIssue}/>);
     expect(wrapper.find('CollapsiblePanel').length).to.equal(1);
```
cvalarida: Same here as above.

cvalarida (approved):
I had a couple of non-blocking suggestions, but otherwise it looks great!

markgreenburg:
@cvalarida Would be great to get another look at this PR - made a few changes per comments.

test/claims-status/components/appeals-v2/Issues.unit.spec.jsx
markgreenburg: Yeah that's fine; these don't have any special identifiers but I can dive into them and then inspect the panel names

src/sass/claims-status.scss
```diff
@@ -1086,4 +1086,26 @@ $marker-text-width: 9rem;
...
+    h5 {
+      color: $color-secondary;
+    }
+  }
 }
+
+.closed-appeal-notice {
+  line-height: 40px;
+  text-align: center;
+  font-family:  "Source Sans Pro", "Helvetica Neue", "Helvetica", "Roboto", "Arial", sans-serif;
```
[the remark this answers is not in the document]
markgreenburg: 👍 I had to dig for a bit, but it's in `uswds/dist/scss/_variables.scss`, and is logically called `$font-sans`

cvalarida (approved):
Looks great!

PR #7188, opened 2018-01-30
Connect supporting documents upload to api

cehsu:
These changes connect supporting documents upload to the API:
endpoint: /v0/vic/supporting_documentation_attachments
File field is: supporting_documentation_attachment[file_data]

### DIFF
--- a/src/js/vic-v2/config/form.js
+++ b/src/js/vic-v2/config/form.js
@@ -193,7 +193,22 @@ const formConfig = {
           uiSchema: {
             'ui:description': DD214Description,
             dd214: fileUploadUI('Upload your discharge document', {
-              endpoint: '/v0/vic/vic_attachments'
+              endpoint: '/v0/vic/supporting_documentation_attachments',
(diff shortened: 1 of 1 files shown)


rianfowler (approved):

src/js/vic-v2/config/form.js
```diff
@@ -193,7 +193,22 @@ const formConfig = {
           uiSchema: {
             'ui:description': DD214Description,
             dd214: fileUploadUI('Upload your discharge document', {
-              endpoint: '/v0/vic/vic_attachments'
+              endpoint: '/v0/vic/supporting_documentation_attachments',
+              fileTypes: ['pdf'],
```
jbalboni: We should be able to upload other file types.

jbalboni (approved):

src/js/vic-v2/config/form.js
cehsu: Thanks @jbalboni, changed to include the same types allowed for the ID photo as well as pdf.

PR #7194, opened 2018-01-30
Performance updates for photo cropper

jbalboni:
I found that moving the photo around on my Android phones (a Pixel and an old Moto G) was really slow. This ended up being two things:

1. The camera images are very large and can be resized (this is the guidance from the cropperjs readme)
2. We were generating a dataUrl on every crop event. This takes like 400ms on a slow phone and crop events are fired on every movement or zoom of the image. I changed the code to generate the data url only on when "I'm done" is clicked.
3. Dropzone says you should revoke preview urls, so I also made that change, but I didn't see any performance issues because of it.

The change for 1 also revealed an issue with the `setTimeout` call. When I loaded a dataUrl that was resized, I believe because the browser hadn't loaded and cached it already like with an image that has been read with a FileReader, it took longer for the cropper component to load the image and caused the crop box to be the wrong size. I changed the code to set the crop box on the `ready` event, which is fired when the cropper tool is done loading the image.

### DIFF
--- a/src/js/vic-v2/components/PhotoField.jsx
+++ b/src/js/vic-v2/components/PhotoField.jsx
@@ -36,6 +36,7 @@ const MIN_RATIO = 0.2;
 const MAX_RATIO = 1.7;
 const WARN_RATIO = 1.3;
 const LARGE_SCREEN = 1201;
+const MAX_DIMENSION = 1024;
 
 function isSmallScreen(width) {
   return  width < LARGE_SCREEN;
@@ -45,11 +46,17 @@ function isValidFileType(fileName) {
   return FILE_TYPES.some(type => fileName.toLowerCase().endsWith(type));
 }
 
+// data urls start with data:image/jpg; and we need to pull out the
+// image/jpg part so canvas doesn't change the file type
 function getMimeType(dataUrl) {
+  // limit to 2 so we don't try to split a giant data url string
   const prefix = dataUrl.split(';', 2)[0];
   return prefix.replace('data:', '');
 }
 
+// If any of the image dimensions are greater than the max specified, 
+// resize it down to that dimension while keeping the aspect ratio
+// intact
 function resizeToMaxDimension(img, mimeType, maxDimension) {
   const oc = document.createElement('canvas');
   const octx = oc.getContext('2d');
@@ -73,7 +80,6 @@ function resizeToMaxDimension(img, mimeType, maxDimension) {
   }
 
   return img.src;
-
 }
 
 function isValidImageSize(img) {
@@ -166,7 +172,7 @@ export default class PhotoField extends React.Component {
                 });
               } else {
                 this.setState({
-                  src: resizeToMaxDimension(img, getMimeType(reader.result), 1024),
+                  src: resizeToMaxDimension(img, getMimeType(reader.result), MAX_DIMENSION),
                   done: false,
                   cropResult: null,
                   errorMessage: null
@@ -178,7 +184,7 @@ export default class PhotoField extends React.Component {
                 src: null,
                 done: false,
                 cropResult: null,
-                errorMessage: 'Sorry, we weren\'t able to load the image you selected'
+                errorMessage: 'Sorry, we weren’t able to load the image you selected'
               });
             });
         };
--- a/src/js/vic-v2/components/PhotoField.jsx
+++ b/src/js/vic-v2/components/PhotoField.jsx
@@ -49,7 +49,7 @@ function isValidFileType(fileName) {
 // If any of the image dimensions are greater than the max specified, 
 // resize it down to that dimension while keeping the aspect ratio
 // intact
-function resizeToMaxDimension(img, mimeType, maxDimension) {
+function resizeIfAboveMaxDimension(img, mimeType, maxDimension) {
   const oc = document.createElement('canvas');
   const octx = oc.getContext('2d');
   let height = img.height;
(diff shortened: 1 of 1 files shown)


src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -296,16 +319,16 @@ export default class PhotoField extends React.Component {
           {!this.state.done && this.state.src && <div className="cropper-container-outer">
             <Cropper
               ref="cropper"
+              ready={this.setCropBox}
```
cehsu: Great improvement, @jbalboni, would it make sense to respond to react-cropper issue 59?

src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -45,30 +46,57 @@ function isValidFileType(fileName) {
...
+function getMimeType(dataUrl) {
+  // limit to 2 so we don't try to split a giant data url string
+  const prefix = dataUrl.split(';', 2)[0];
+  return prefix.replace('data:', '');
+}
+
+// If any of the image dimensions are greater than the max specified, 
+// resize it down to that dimension while keeping the aspect ratio
+// intact
+function resizeToMaxDimension(img, mimeType, maxDimension) {
```
cehsu: This is really helpful for the user. Should we default to the full canvas area or the cropping frame? Also, should we use "default" rather than "max", as users are still able to zoom in?

src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -296,16 +319,16 @@ export default class PhotoField extends React.Component {
...
               src={this.state.src}
-              aspectRatio={1 / 1}
+              aspectRatio={1}
               cropBoxMovable={false}
               toggleDragModeOnDblclick={false}
               dragMode="move"
               guides={false}
               viewMode={1}
-              zoom={this.onZoom}
-              crop={this.cropImage}/>
```
cehsu: Good idea removing this since we never crop by manually setting the crop box or by using autocrop, etc.

src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -106,62 +134,64 @@ export default class PhotoField extends React.Component {
...
-    }
-  }
-
   onEdit = () => {
     this.setState({ done: false, warningMessage: null });
   }
 
   onDone = () => {
-    this.setState({ done: true, warningMessage: null });
+    const cropResult = this.refs.cropper.getCroppedCanvas().toDataURL();
```
cehsu: To prevent the image from being resized and set to another type we can provide the original type here. This is done in the pending PR for triggering the file upload and is explained [here](https://github.com/fengyuanchen/cropperjs#known-issues).

cehsu (approved):
LGTM. Had a few minor questions. Also, your improvement on the setTimeout hack is really great and might be worth sharing on the react-cropper repo's issue 59.

src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -45,30 +46,57 @@ function isValidFileType(fileName) {
   return FILE_TYPES.some(type => fileName.toLowerCase().endsWith(type));
 }
 
-function isValidImageSize(result) {
+// data urls start with data:image/jpg; and we need to pull out the
+// image/jpg part so canvas doesn't change the file type
+function getMimeType(dataUrl) {
```
cehsu: Could we use file.type to get the media type? And in the validation helper as well? It seems like the downside of only being able to rely on this for standard types is not a problem here.

src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -106,62 +134,64 @@ export default class PhotoField extends React.Component {
...
-                done: false,
-                cropResult: null,
-                errorMessage: 'Please choose a file from one of the accepted types.'
-              });
-            } else if (isValidFileType(file.name)) {
+      if (file.preview) {
+        // dropzone recommendation
+        window.URL.revokeObjectURL(file.preview);
+      }
+      if (!isValidFileType(file.name)) {
```
cehsu: This `onChange` organization is much clearer, in fact, should we go ahead and remove the type validation as both our upload button and react-dropzone perform this automatically?

src/js/vic-v2/components/PhotoField.jsx
rianfowler: Might be helpful to call this `maybeResizeToMaxDimensions` to the name of this function since it doesn't upscale the image to the max size.

I also find using an object parameter makes calls to this function easier to read when there are more than 2 arguments  e.g. `({img, mimeType, maxDimension})`

rianfowler (approved):

src/js/vic-v2/components/PhotoField.jsx
jbalboni: Yep, I added a comment.

src/js/vic-v2/components/PhotoField.jsx
jbalboni: Yeah, I'll make the name clearer

src/js/vic-v2/components/PhotoField.jsx
jbalboni: Ha, I looked for something like that and totally missed it. Yes, that would be better.

src/js/vic-v2/components/PhotoField.jsx
jbalboni: I think we still need to do it here. The upload button's validation is more of a suggestion and can be easily avoided.

src/js/vic-v2/components/PhotoField.jsx
jbalboni: I'm not sure about using it for the validation helper. I think that would make the code more complicated, since we'd need to map the type to the file suffix (e.g. image/jpeg is for both .jpeg and .jpg).

cehsu: It looks like there may be some additional extensions that we may wish to account for, [bmp](https://tools.ietf.org/html/rfc7903)(dib), [jpeg](https://www.sitepoint.com/mime-types-complete-list/)(jpe).

PR #7195, opened 2018-01-31
Jenkins build performance - parallel lint, security, unit jobs

wyattwalter:
This cuts an additional minute+ off the job time by running lint and unit tests in parallel (security is only a few seconds, but may as well combine it while we're at it).

Ref: https://github.com/department-of-veterans-affairs/vets.gov-team/issues/8407


edpaget (approved):
LGTM

PR #7179, opened 2018-01-29
7137 vre veteran info page

rianfowler:
2018-01-29-10-26-localhost_3001

PR #7198, opened 2018-01-31
7137 update ssn va file number behavior

rianfowler:
2018-01-31-11-28-localhost_3001

### DIFF
--- a/src/js/vre/chapter36/config/form.js
+++ b/src/js/vre/chapter36/config/form.js
@@ -162,9 +162,13 @@ const formConfig = {
           uiSchema: {
             veteranFullName: fullNameUI,
             veteranDateOfBirth: currentOrPastDateUI('Date of birth'),
-            veteranSocialSecurityNumber: ssnUI,
+            veteranSocialSecurityNumber: _.assign(ssnUI, {
+              'ui:title': 'Social Security number (must have this or a VA file number)',
+              'ui:required': form => !form.veteranVaFileNumber,
+            }),
             veteranVaFileNumber: {
-              'ui:title': 'VA File Number',
+              'ui:title': 'VA file number (must have this or a Social Security number)',
+              'ui:required': form => !form.veteranSocialSecurityNumber,
               'ui:errorMessages': {
                 pattern: 'Your VA file number must be between 7 to 9 digits'
               }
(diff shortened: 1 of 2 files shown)


src/js/vre/chapter36/config/form.js
```diff
@@ -162,9 +162,13 @@ const formConfig = {
...
+            veteranSocialSecurityNumber: _.assign(ssnUI, {
+              'ui:title': 'Social Security number (must have this or a VA file number)',
+              'ui:required': form => !form.veteranVaFileNumber,
+            }),
             veteranVaFileNumber: {
-              'ui:title': 'VA File Number',
+              'ui:title': 'VA file number (must have this or a Social Security number)',
+              'ui:required': form => !form.veteranSocialSecurityNumber,
               'ui:errorMessages': {
                 pattern: 'Your VA file number must be between 7 to 9 digits'
```
rianfowler: Left this error message but it's different from the code that was mentioned [here](https://github.com/department-of-veterans-affairs/vets-website/pull/7179#discussion_r164576333).

cehsu (approved):
LGTM. Is this based on feedback from @stacyhlinn or other folks familiar with the design?

rianfowler:
@cehsu I found it [here](https://github.com/department-of-veterans-affairs/vets-website/blob/16cabaacc708f02a09230c1d8edec47139432f5e/src/js/common/schemaform/definitions/personId.js#L38). The default message shows a reg ex which I thought was incorrect. But definitely should have @stacyhlinn review it.

src/js/vre/chapter36/config/form.js
jbalboni: I think this message is correct.

jbalboni (approved):

src/js/vre/chapter36/config/form.js
cehsu: @rianfowler I was actually asking if this was conditionally required, since I asked about that on the previous PR. Just double checking on when what is required, since it seems like this is updated?

rianfowler: Stacy [clarified](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/7137#issuecomment-361964661) that at least one is required and it should behave like the pensions form. This doesn't perform exactly like the pensions form- I pulled the error message from the other another form since this is the default error message: 

screen shot 2018-01-31 at 16 17 39

The error message could be tweaked to be more accurate though e.g. "should start with a c, 0, or 1".

rianfowler: I should have mentioned this in the comments.

PR #7202, opened 2018-02-01
Remove `liClass`

cvalarida:
Uses a default instead of putting it in every event.

This needs to be merged in _after_ #7124 which I branched off of. I realize that isn't a great idea, but it already had all the updated event types.


cvalarida:
This suffered from the same weird merge issue as the parent branch (meh), so rather than try to fix it in this branch, I'm just going to start a new one. The changes were minor enough for that.

PR #7203, opened 2018-02-01
VIC v2 document, review, confirmation e2e tests

cehsu:
These changes set up the e2e tests for the second half of the VIC application.

The upload tests are somewhat complicated by the fact that our upload buttons are not visible, and therefore need to be styled for testing purposes in order for the browser's driver to be able to interact with them (as documented in issue 505 in Nightwatch's repo).


test/vic-v2/00-all-fields.e2e.spec.js
```diff
@@ -45,21 +45,63 @@ const runTest = E2eHelpers.createE2eTest(
...
+        .execute("document.getElementsByName('fileUpload')[0].style.display = 'block';");
+      // HACK: waitforElementVisible did not work but this does 
+      client.elementIdDisplayed('errorable-file-input-11');
+
+      // upload photo
+      client
+        .setValue('input[name="fileUpload"]', require('path').resolve(`${__dirname}/examplephoto.png`));
+
+      // crop photo
+      client.axeCheck('.main');
```
cehsu: Is visual regression testing the preferred way to test the cropper controls?

test/vic-v2/00-all-fields.e2e.spec.js
```diff
@@ -45,21 +47,84 @@ const runTest = E2eHelpers.createE2eTest(
...
       E2eHelpers.expectNavigateAwayFrom(client, '/contact-information');
 
+      // Photo page
+      client.assert.cssClassPresent('.progress-bar-segmented div.progress-segment:nth-child(3)', 'progress-segment-complete');
+      client.waitForElementVisible('label.usa-button.usa-button-secondary', Timeouts.normal);
+      client.axeCheck('.main');
+
+      // Disable upload button style to reveal input for test
+      /* HACK: style overridden so browser driver can find/interact with hidden inputs
+         (see https://github.com/nightwatchjs/nightwatch/issues/505) */
```
cehsu: Is this comment and/or the use of the HACK tag helpful?

test/vic-v2/00-all-fields.e2e.spec.js
```diff
@@ -45,21 +47,84 @@ const runTest = E2eHelpers.createE2eTest(
...
+      client.elementIdDisplayed('errorable-file-input-11');
+
+      // sighted path
+      // upload photo
+      client
+        .setValue('input[name="fileUpload"]', require('path').resolve(`${__dirname}/examplephoto.png`));
+
+      // crop photo
+      client.waitForElementVisible('.cropper-container-outer', Timeouts.normal);
+      client.click('.form-panel .usa-button-primary');
```
cehsu: Would visual regression tests be the best way to test the cropper's features?

test/vic-v2/00-all-fields.e2e.spec.js
rianfowler: I think so- It's a good idea to document anything that changes the system under test.

test/vic-v2/00-all-fields.e2e.spec.js
rianfowler: Could verify some of the button disabling behavior by rotating a wide image once (move up, move right, move left should all be disabled after this). That would validate that the canvas loaded the image and that the buttons can be disabled. 

Verifying the position of the image would require grabbing the image data from cropperjs.

rianfowler:
Looks good- was able to run locally (once I installed the schema form update 😸 )

rianfowler (approved):

test/vic-v2/00-all-fields.e2e.spec.js
jbalboni: Yeah, it's going to be tough to test a lot of this. Our visual regression testing set up isn't really at the point where we could use it for this yet.

jbalboni (approved):
LGTM, though I'm not sure where that example photo came from. As long as it's licensed in a way that it's ok for us to have it in the repo, that's fine.

test/vic-v2/00-all-fields.e2e.spec.js
cehsu: Thanks for the helpful guidance, that's a good way to determine whether to use these.

test/vic-v2/00-all-fields.e2e.spec.js
cehsu: Perhaps this testing should be done in tandem with the work to split the cropper component up? It sounds like we might not need to use nightwatch for some of these things. If so, I can create a ticket to describe this work.

cehsu:
Thanks @jbalboni for pointing out the photo issue, I've included a photo provided by @lalexanderson-dsva.

PR #7205, opened 2018-02-01
reorganize markup and tweak flexbox styling

rianfowler:
PR addresses: reorganize the move controls so that when the button sequence is Up Down Left Right on large screens as indicated in the design, rather than the current organization of Up Left Down Right.

LMK is this is this correct @cehsu 

screen shot 2018-02-01 at 08 54 28
screen shot 2018-02-01 at 08 54 19

### DIFF
--- a/src/js/vic-v2/components/PhotoField.jsx
+++ b/src/js/vic-v2/components/PhotoField.jsx
@@ -347,25 +347,27 @@ export default class PhotoField extends React.Component {
               </button>}
             </div>
             <div className="cropper-control-container">
-              <div className="cropper-control-column">
+              <div className="cropper-control-row">
                 {smallScreen && <button className="cropper-control cropper-control-label-container va-button va-button-link" type="button" onClick={this.zoomOut}>
                   <span className="cropper-control-label">Make smaller</span>
                 </button>}
+                {smallScreen && <button className="cropper-control cropper-control-label-container va-button va-button-link" type="button" onClick={this.zoomIn}>
+                  <span className="cropper-control-label">Make larger</span>
+                </button>}
+              </div>
+              <div className="cropper-control-row">
                 <button className={moveControlClass} type="button" onClick={this.moveUp}>
                   <span className="cropper-control-label">Move up<i className="fa fa-arrow-up"></i></span>
 
                 </button>
-                <button className={moveControlClass} type="button" onClick={this.moveLeft}>
-                  <span className="cropper-control-label">Move left<i className="fa fa-arrow-left"></i></span>
+                <button className={moveControlClass} type="button" onClick={this.moveDown}>
+                  <span className="cropper-control-label">Move down<i className="fa fa-arrow-down"></i></span>
 
                 </button>
               </div>
-              <div className="cropper-control-column">
-                {smallScreen && <button className="cropper-control cropper-control-label-container va-button va-button-link" type="button" onClick={this.zoomIn}>
-                  <span className="cropper-control-label">Make larger</span>
-                </button>}
-                <button className={moveControlClass} type="button" onClick={this.moveDown}>
-                  <span className="cropper-control-label">Move down<i className="fa fa-arrow-down"></i></span>
+              <div className="cropper-control-row">
+                <button className={moveControlClass} type="button" onClick={this.moveLeft}>
+                  <span className="cropper-control-label">Move left<i className="fa fa-arrow-left"></i></span>
 
                 </button>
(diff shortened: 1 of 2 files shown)


src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -347,25 +347,27 @@ export default class PhotoField extends React.Component {
               </button>}
             </div>
             <div className="cropper-control-container">
-              <div className="cropper-control-column">
+              <div className="cropper-control-row">
```
rianfowler: Removed the flex column and renamed this style.

src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -347,25 +347,27 @@ export default class PhotoField extends React.Component {
...
                 {smallScreen && <button className="cropper-control cropper-control-label-container va-button va-button-link" type="button" onClick={this.zoomOut}>
                   <span className="cropper-control-label">Make smaller</span>
                 </button>}
+                {smallScreen && <button className="cropper-control cropper-control-label-container va-button va-button-link" type="button" onClick={this.zoomIn}>
+                  <span className="cropper-control-label">Make larger</span>
+                </button>}
+              </div>
+              <div className="cropper-control-row">
```
rianfowler: Moved the smaller / larger buttons into their own row.

src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -347,25 +347,27 @@ export default class PhotoField extends React.Component {
...
                 <button className={moveControlClass} type="button" onClick={this.moveUp}>
                   <span className="cropper-control-label">Move up<i className="fa fa-arrow-up"></i></span>
 
                 </button>
-                <button className={moveControlClass} type="button" onClick={this.moveLeft}>
-                  <span className="cropper-control-label">Move left<i className="fa fa-arrow-left"></i></span>
+                <button className={moveControlClass} type="button" onClick={this.moveDown}>
```
rianfowler: New Rows: move up + move down, move left + move right

src/sass/vic-v2.scss
```diff
@@ -181,21 +181,26 @@ input[type=range] {
...
+  justify-content: space-around;
   @media (min-width: $large-screen) {
-    flex-direction: row;
+    width: unset;
   }
 }
 
 .cropper-control-label-container {
   margin: 10px;
+  min-width: 96px;
```
rianfowler: Not a fan of magic numbers but this was the approximate calculated width of sibling labels.

cehsu (approved):
LGTM. This looks great!

src/sass/vic-v2.scss
cehsu: It seems like a necessary evil in this case.

src/sass/vic-v2.scss
```diff
@@ -181,21 +181,26 @@ input[type=range] {
...
 
-.cropper-control-column {
+.cropper-control-row {
   display: flex;
-  flex-direction: column;
+  width: 100%;
+  justify-content: space-around;
   @media (min-width: $large-screen) {
-    flex-direction: row;
+    width: unset;
```
jbalboni: This isn't supported in IE11

(Also, I had no idea it existed and it seems useful).

jbalboni (approved):
One comment, otherwise LGTM.

src/sass/vic-v2.scss
rianfowler: Good catch. I will update.

src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -347,31 +347,42 @@ export default class PhotoField extends React.Component {
...
                 </button>}
-                <button className={moveControlClass} type="button" onClick={this.moveDown}>
-                  <span className="cropper-control-label">Move down<i className="fa fa-arrow-down"></i></span>
-
-                </button>
-                <button className={moveControlClass} type="button" onClick={this.moveRight}>
-                  <span className="cropper-control-label">Move right<i className="fa fa-arrow-right"></i></span>
-                </button>
               </div>
+              {[
```
rianfowler: Codeclimate gave a repeating code error.

jbalboni: @rianfowler You can feel free to ignore those in the future. We are planning to turn code climate off.

rianfowler: Yeah @cehsu let me know. I was confused by the UI and thought I had to resolve all of these issues. 😂

PR #7207, opened 2018-02-01
jenkinsfile bail on nextbuild

wyattwalter:
Have Jenkins abort a job if it detects a new build on the same branch that came after the current one. The evaluation only happens at the beginning of a stage, so it won't interrupt any of the longer-running tasks in the middle, but will bail as soon as it can cleanly to avoid extra work.

I specifically excluded master because the deploys have a timing-related condition where we could attempt to tag a build that didn't actually get built. I think it's worthwhile to continue to the end for each master commit.

ref: https://github.com/department-of-veterans-affairs/vets.gov-team/issues/8407

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -124,6 +124,9 @@ node('vetsgov-general-purpose') {
   // Perform a build for each build type
 
   stage('Build') {
+    if (currentBuild.nextBuild) {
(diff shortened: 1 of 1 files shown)


wunderhund (approved):
LGTM

PR #7213, opened 2018-02-02
[WIP] Photo Uploader Accessibility Updates

cehsu:
These changes address the following:

TODO:
- [x] provide non-sighted path to upload tool
- [ ] unit test support for cropper component and helpers
- [ ] additional accessibility fixes identified by axe
- [x] convert edit and upload buttons into link style buttons

screen shot 2018-02-06 at 4 14 51 pm
screen shot 2018-02-06 at 4 14 38 pm
screen shot 2018-02-06 at 4 14 28 pm
screen shot 2018-02-06 at 4 14 18 pm


src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -109,34 +91,6 @@ function makeMoveButtonsEnabledStates({ topBoundaryMet, bottomBoundaryMet, right
   };
 }
 
-// If any of the image dimensions are greater than the max specified,
-// resize it down to that dimension while keeping the aspect ratio
-// intact
-function resizeIfAboveMaxDimension(img, mimeType, maxDimension) {
```
jbalboni: I removed this because after testing on my Android phones some more, it seems like the data url serialization was the big performance issue, not image size. iPhones are faster than Android and shouldn't have much of an issue either. The resizing here was making it essentially impossible to work with large files, since they would get resized down to where you couldn't reasonable zoom in on them.

src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -152,17 +106,29 @@ function loadImage(dataUrl) {
...
       zoomValue: 0.4,
-      errorMessage: null,
       warningMessage: null,
-      progress: 0
+      progress: 0,
+      isCropping: false,
+      previewSrc
     };
+
+    this.screenReaderPath = false;
```
jbalboni: I really dislike having to track this, but it seems necessary.

src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -388,55 +413,81 @@ export default class PhotoField extends React.Component {
...
+    const errorMessage = file.errorMessage;
+    const screenReaderError = this.screenReaderPath && !!errorMessage;
+    const label = this.props.uiSchema['ui:title'];
     const smallScreen = isSmallScreen(this.state.windowWidth);
     const moveControlClass = ['cropper-control', 'cropper-control-label-container', 'va-button-link'];
     const fileTypes = this.props.uiSchema['ui:options'].fileTypes;
-    const isDone = !!file.confirmationCode;
     const progressBarContainerClass = classNames('schemaform-file-uploading', 'progress-bar-container');
+
+    let fieldView;
```
jbalboni: I'm hoping this is a clearer way of understanding the different visual states in the field. Let me know if it isn't and I can adjust it.

jbalboni:
@rianfowler @cehsu This is ready for an initial round of review on the things that have been checked off in the description.

src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -217,14 +211,50 @@ export default class PhotoField extends React.Component {
...
+    this.screenReaderPath = true;
+
+    const reader = new FileReader();
+    reader.readAsDataURL(file);
+    reader.onload = () => {
+      loadImage(reader.result)
+        .then((img) => {
+          if (!isSquareImage(img)) {
+            this.props.onChange({
+              errorMessage: 'The photo you uploaded is not a square photo. Please upload a new one that fits the requirements.'
```
rianfowler: We may not need to require loading a square image since the cropper crops the image into a square. Most images taken by phones, etc are not square and it might be a bad experience to require cropping before loading it.

rianfowler:
@jbalboni  LGTM so far. I don't anticipate any conflicts with the `PhotoField.jsx` changes I'm working on.

src/js/common/components/form-elements/ErrorableFileInput.jsx
```diff
@@ -54,7 +54,14 @@ class ErrorableFileInput extends React.Component {
             {requiredSpan}
           </label>
           {errorSpan}
-          <label role="button" tabIndex="0" htmlFor={this.inputId} className="usa-button usa-button-secondary">{this.props.buttonText}</label>
+          <label
+            role="button"
+            tabIndex="0"
+            htmlFor={this.inputId}
+            aria-describedby={this.props['aria-describedby']}
+            className={this.props.triggerClass || 'usa-button usa-button-secondary'}>
```
cehsu: Should we use classNames to manage the logic here if we have any classes we always plan to apply, or do we anticipate cases in which we would not some of the default classes applied?

src/js/common/schemaform/helpers.js
```diff
@@ -252,6 +252,11 @@ export function stringifyFormReplacer(key, value) {
     if (value.widget === 'autosuggest') {
       return value.id;
     }
+
+    // Exclude file data
+    if (value.confirmationCode && value.file) {
```
cehsu: Why are we using the confirmationCode to short circuit here?

src/sass/vic-v2.scss
```diff
@@ -397,7 +403,20 @@ input[type=range] {
...
+  font-weight: bold;
+  margin-bottom: .5em;
+}
+
+.photo-error-message {
+  outline: none;
+}
+
+.photo-preview-link {
+  margin-right: 10px;
```
cehsu: Should this be an `em` value such as 1em? Do we favor these in general?

test/vic-v2/config/photo.unit.spec.jsx
```diff
@@ -0,0 +1,85 @@
...
+        definitions={formConfig.defaultDefinitions}
+        schema={schema}
+        data={{}}
+        uiSchema={uiSchema}/>
+    );
+
+    expect(form.find('input').length).to.equal(3);
+  });
+
+  // Disabled pending PR 7175
```
cehsu: It looks like these comments are no longer necessary. I will remove them.

cehsu (commented):
LGTM. Should we split the additional tests and axe-related work into another PR, if this is otherwise pretty much ready to merge?

src/js/common/schemaform/helpers.js
jbalboni: I'm using that to make sure we're in the typical file object and not something else that happens to have a `file` property.

src/js/common/components/form-elements/ErrorableFileInput.jsx
jbalboni: I think this ok when we're just overriding with an option prop. If it gets more complicated, we should definitely use classNames.

jbalboni:
@cehsu Yeah, let's go ahead and get this merged and then have a separate branch for the tests and other minor aXe fixes. I'll also create one to incorporate Ryan's photo url changes when they're ready.

jbalboni (approved):
I'm only doing this because there are two other LGTMs on this PR already.

PR #7214, opened 2018-02-02
Change up the exit statuses on run-unit-test.sh

cvalarida:
Previously, the exit status was always going to be 0 for `npm run test:unit`, so...that wasn't good. My bad.


wyattwalter (approved):

script/run-unit-test.sh
```diff
@@ -11,8 +11,12 @@ fi
 # No test specified; run all of them
 if [ -z "$1" ]; then
   BABEL_ENV=test $MOCHA --recursive '$SCRIPT_DIR/../test/**/*.unit.spec.js?(x)' 
-  exit 0
+  exit $?
```
cvalarida: Oh bash. Why you gotta be like this? Of _course_  `$? ` means `the exit status of the previously executed command`.

:man_shrugging:

PR #7215, opened 2018-02-02
Appeals refactor -- Use default liClass

cvalarida:
Just a simple refactor. No visual changes.

I was having trouble with https://github.com/department-of-veterans-affairs/vets-website/pull/7202, so I just closed that out and re-did it here.


markgreenburg (approved):
LGTM

PR #7218, opened 2018-02-03
WIP: VIC review page document uploads tweaks

cehsu:
These changes update the review page in the following ways:

- [x] Change labels on review page from upload to review
- [x] Update the review display of the photo cropping tool to match the prototype
- [x] Validate that all other fields look correct

Some additional work is also included that begins to manage the image retrieval for preview. This was not in the original ticket, but feedback on whether that work should be included here, or on how it is being approached, is welcome.

screen shot 2018-02-07 at 12 02 32 pm

screen shot 2018-01-31 at 10 16 25 am

Please note that the "Edit" buttons are now round in order to be consistent with our general button styles, based on feedback from @stacyhlinn.


cehsu:
@stacyhlinn I wanted to check in with you as there are some slight differences from the prototype since edit buttons are typically placed next to the chapter title, as some chapters may contain multiple fields (for example, it is possible to upload multiple discharge documents).

src/js/vic-v2/config/form.js
```diff
@@ -202,6 +204,9 @@ const formConfig = {
               photo: {
                 type: 'object',
                 properties: {
+                  'view:dataURL': {
```
cehsu: This seems like a reasonable way to store the image data (to be stored on photo submission, and retrieved in `fetchInProgressForm`) that we need for the review page, but other ideas for handling this are welcome.

jbalboni: This is what @rlbaker is working on now. We can't store this data url in the database without performance issues, so he is working on getting us a way to request the image based on the guid associated with the image. For now, what I did in the screen reader PR is save the file object itself in the form data. The actual file won't be included when it's serialized to be saved in progress, so it doesn't cause performance issues, and we'll be able to show previews using it until Ryan's work is done.

src/js/vic-v2/config/form.js
```diff
@@ -235,7 +240,7 @@ const formConfig = {
                 'bmp'
               ],
               maxSize: TWENTY_FIVE_MB,
-              hideLabelText: true,
+              alternativeLabel: 'Upload your discharge document',
```
jbalboni: Is this used?

src/js/common/schemaform/review/ReviewFieldTemplate.jsx
```diff
@@ -7,21 +7,34 @@ import React from 'react';
...
+        <dd>{children}</dd>
+      </div>
+    );
+  }
+
+  if (uiSchema['ui:reviewDescription']) {
+    const ReviewDescriptionField = uiSchema['ui:reviewDescription'];
+    reviewDescription = (
+      <div className="va-growable-background">
+        <ReviewDescriptionField formData={children.props.formdData}/>
```
jbalboni: Typo in formData

src/js/vic-v2/config/form.js
```diff
@@ -190,6 +191,7 @@ const formConfig = {
               }
             }), {
               'ui:field': PhotoField,
+              'ui:reviewDescription': photoReviewDescription,
```
jbalboni: A simpler path than building this into the common schemaform code might be to detect if we're in review mode in the PhotoField component (it's in `this.props.formContext.reviewMode`), and return this description there. We've done that in some other custom fields.

src/js/vic-v2/config/form.js
```diff
@@ -159,10 +159,12 @@ const formConfig = {
     },
     documentUpload: {
       title: 'Document Upload',
+      reviewTitle: 'Document Review',
```
jbalboni: I think adding a reviewTitle is fine, but I wonder if it would be better in this case to just have "Documents" be the chapter title in both places?

src/js/vic-v2/config/form.js
cehsu: Thanks for the context. I'll follow your lead and save the file object itself as well.

src/js/common/schemaform/review/ReviewFieldTemplate.jsx
cehsu: Thanks, good catch!

src/js/vic-v2/config/form.js
cehsu: Good idea, @peggygannon is it preferred to have distinct titles for "Document upload" in the form and "Document review" on the review page, or could we simply use "Documents" as the chapter title in both places?

src/js/vic-v2/config/form.js
cehsu: This is intended to update the button text, I've updated the File field and used buttonText instead in an effort to make the intended use clearer.

src/js/common/schemaform/save-in-progress/actions.js
```diff
@@ -200,6 +200,7 @@ export function fetchInProgressForm(formId, migrations, prefill = false, prefill
   //  redux store, but form.migrations doesn’t exist (nor should it, really)
   return (dispatch, getState) => {
     const trackingPrefix = getState().form.trackingPrefix;
+    // const imageEndpoint = getState().form.imageEndpoint;
```
cehsu: The plan here is to create `fetchFormApi` in `./api` and use imageEndpoint to set up something like:
```
    const formUrl = `${environment.API_URL}/v0/in_progress_forms/${formId}`;
    if (imageEndpoint) {
      const imageUrl = `${environment.API_URL}${imageEndpoint}`;

      Promise.all([formUrl, imageUrl].map(
        url => fetchFormApi(url, responseHandler, errorHandler)
      ))
        .then(processResponseAsUsual)
        .catch(errorHandler);
    } else {
      fetchFormApi(formUrl, responseHandler, errorHandler)
        .then(processResponseAsUsual)
        .catch(errorHandler);
    }

```

jbalboni: I don't think I understand the purpose of that code.

src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -289,17 +295,31 @@ export default class PhotoField extends React.Component {
...
+      uiSchema,
+      formData,
+      formContext,
+    } = this.props;
+    const { file } = formData || {};
+    const onReview = formContext.reviewMode;
     const errorMessage = file.errorMessage || this.state.errorMessage;
-    const smallScreen = isSmallScreen(this.state.windowWidth);
-    const fileTypes = this.props.uiSchema['ui:options'].fileTypes;
+    const smallScreen = isSmallScreen(this.state.windowWidth) || onReviewPage(formContext.pageTitle);
```
jbalboni: Can you explain some of the changes in here, aside from the review page check on line 345? Some of these I think would conflict with changes I made in the screen reader PR. I'm not sure what changes we would need to make in here other than the review mode switch.

src/js/common/schemaform/save-in-progress/actions.js
cehsu: The purpose here is to retrieve the saved image file, as it is not being stored with the form data, correct? Or will we be able to use the same endpoint to retrieve this?

src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -81,6 +82,10 @@ function isSmallScreen(width) {
   return  width < LARGE_SCREEN;
 }
 
+function onReviewPage(pageTitle) {
+  return pageTitle === 'Review your photo';
```
jbalboni: You can tell you're on the review page by `this.props.formContext.reviewMode` being `true`.

src/sass/vic-v2.scss
```diff
@@ -401,3 +405,22 @@ input[type=range] {
   clear: both;
   margin: 0 0 1.5em 0;
 }
+/* USWDS overrides */
+.usa-accordion, .usa-accordion-bordered {
```
jbalboni: The accordion styles are such a pain to work with.

src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -481,9 +486,14 @@ export default class PhotoField extends React.Component {
   }
 
   render() {
+    const { formData, formContext } = this.props;
+    const onReview = formContext.reviewMode;
+
+    if (onReview) return <PhotoReviewDescription formData={formData}/>;
```
jbalboni: We might need to skip the resize event listener on review as well.

jbalboni (approved):
LGTM. Couple minor comments.

src/js/vic-v2/components/PhotoField.jsx
cehsu: It looks like reviewMode is false once edit is clicked:
screen shot 2018-02-07 at 12 34 10 pm

src/js/vic-v2/config/form.js
peggygannon: @cehsu I think we should keep the earlier chapter headers as "Document Upload", to keep with the style of other forms. And I think it's fine to change the header to "Documents" on the review title.

peggygannon (approved):

src/js/vic-v2/components/PhotoField.jsx
cehsu: Great idea, changed.

src/sass/vic-v2.scss
cehsu: Agreed.

src/js/vic-v2/config/form.js
cehsu: Thanks @peggygannon, updated.

src/js/vic-v2/components/PhotoField.jsx
jbalboni: I think that's fine. If you edit, our pattern is generally that you're on the regular version of the page and it should look just like what you saw filling out the form initially.

cehsu: Thanks @jbalboni so is checking the pageTitle the best way to ensure that we are on the review page? This is used to ensure that the small screen styles are applied, even if the screen size is large, to take into consideration the narrower review page styles.

jbalboni: I'm not really sure I understand why we need the small page styles if the review page is wider than the small page breakpoint we set. Should we change the small page breakpoint instead?

PR #7222, opened 2018-02-05
add additional proptypes to buttonText

rianfowler:
Fixes: 
screen shot 2018-02-05 at 14 38 00

### DIFF
--- a/src/js/common/components/form-elements/ErrorableFileInput.jsx
+++ b/src/js/common/components/form-elements/ErrorableFileInput.jsx
@@ -38,7 +38,7 @@ class ErrorableFileInput extends React.Component {
       labelErrorClass = 'usa-input-error-label';
     }
 
-    // Calculate required.
+    // Calculate required
     let requiredSpan = undefined;
     if (this.props.required) {
       requiredSpan = <span className="form-required-span">*</span>;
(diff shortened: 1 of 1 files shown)


src/js/common/components/form-elements/ErrorableFileInput.jsx
```diff
@@ -38,7 +38,7 @@ class ErrorableFileInput extends React.Component {
       labelErrorClass = 'usa-input-error-label';
     }
 
-    // Calculate required.
+    // Calculate required
```
rianfowler: my auto indenter didn't like this

cehsu (approved):

src/js/common/components/form-elements/ErrorableFileInput.jsx
```diff
@@ -71,7 +71,9 @@ class ErrorableFileInput extends React.Component {
 
 ErrorableFileInput.propTypes = {
   multiple: PropTypes.bool,
-  buttonText: PropTypes.string,
+  buttonText: PropTypes.oneOfType([
+    PropTypes.string,
+    PropTypes.object]),
```
jbalboni: I think there's a React element prop type we could use.

jbalboni (approved):

PR #7223, opened 2018-02-06
Appealsv2 api integration mg

markgreenburg:
- Replace hard coded mock data return with actually querying the `appeals_v2` endpoint for appeals data.
- Add unit tests.

### DIFF
--- a/src/js/claims-status/actions/index.jsx
+++ b/src/js/claims-status/actions/index.jsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import environment from '../../common/helpers/environment';
+import { apiRequest } from '../../common/helpers/api';
 import { makeAuthRequest, mockData } from '../utils/helpers';
 
 export const SET_CLAIMS = 'SET_CLAIMS';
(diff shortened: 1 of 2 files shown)


src/js/claims-status/actions/index.jsx
```diff
@@ -84,16 +86,18 @@ export function fetchAppealsSuccess(response) {
...
-    // Fake the fetch by just returning a resolved promice with the object shape we expect
-    //  to get from the api.
-    return setTimeout(() => Promise.resolve(mockData)
-      .then((response) => dispatch(fetchAppealsSuccess(response)))
-      .catch(() => dispatch({ type: SET_APPEALS_UNAVAILABLE })), 4000);
+    return apiRequest(
+      '/appeals_v2',
+      null,
+      (appeals) => dispatch(fetchAppealsSuccess(appeals)),
+      (error) => {
```
markgreenburg: Might want to segment errors better, but should probably do that in a separate PR since there's work that needs to be done to gather error states and map them out to messages in the UI.

markgreenburg: Tracking in https://github.com/department-of-veterans-affairs/vets.gov-team/issues/8567

src/js/claims-status/actions/index.jsx
```diff
@@ -84,16 +86,18 @@ export function fetchAppealsSuccess(response) {
...
+        Raven.captureException(`vets_appeals_v2_err_get_appeals ${error.message}`);
```
cvalarida: Should this be `Raven.captureMessage()`?

cvalarida (approved):
Looks great. Were you able to verify that the data structure received from the api is the same as we have in our mock data?

src/js/claims-status/actions/index.jsx
markgreenburg: 🤷‍♂️ wasn't sure what our standard usage is across the site; seems like we use both. I think you're right that captureMessage is probably more appropriate though, so updating.

PR #7224, opened 2018-02-06
Remove cssClass proptype, reorder remaining proptypes

markgreenburg:
Removed cssClass, and moved missingEvents around to line up with the order in which the props are destructured in the component's functional definition.

### DIFF
--- a/src/js/claims-status/components/appeals-v2/Expander.jsx
+++ b/src/js/claims-status/components/appeals-v2/Expander.jsx
@@ -31,10 +31,9 @@ const Expander = ({ expanded, dateRange, onToggle, missingEvents }) => {
 
 Expander.propTypes = {
   expanded: PropTypes.bool.isRequired,
-  missingEvents: PropTypes.bool.isRequired,
   dateRange: PropTypes.string.isRequired,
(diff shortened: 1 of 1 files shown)


src/js/claims-status/components/appeals-v2/Expander.jsx
```diff
@@ -31,10 +31,9 @@ const Expander = ({ expanded, dateRange, onToggle, missingEvents }) => {
 
 Expander.propTypes = {
   expanded: PropTypes.bool.isRequired,
-  missingEvents: PropTypes.bool.isRequired,
   dateRange: PropTypes.string.isRequired,
   onToggle: PropTypes.func.isRequired,
-  cssClass: PropTypes.string.isRequired,
+  missingEvents: PropTypes.bool.isRequired,
```
markgreenburg: ^ FYI missingEvents was already a proptype, but I re-ordered it to match the order in which props are destructured in the function definition. Only thing that's really changing is no more cssClass proptype.

cvalarida (approved):
Hmm...this is tough...I think we should ponder about it some more... :thinking: 

...aaaaaaand done! Looks good. :smile:

PR #7225, opened 2018-02-06
Fixed bugs in Health Records reducer

U-DON:
Resolves department-of-veterans-affairs/vets.gov-team#7923.

The modal stopped working because upon clicking the button to return to the Main component, it tried to map to props an undefined `errors` property from the refresh state object. At this point, refresh has already overwritten its state upon `REFRESH_POLL_SUCCESS`, clearing `errors`.

Also found that we were still using a snake_cased `last_updated` key, but we receive all keys in camelCase now, so I corrected it to `lastUpdated`. Probably was causing plenty of problems with error reporting we were unaware of.

### DIFF
--- a/src/js/health-records/actions/refresh.js
+++ b/src/js/health-records/actions/refresh.js
@@ -7,29 +7,19 @@ export function initialAppRefresh() {
     return apiRequest(
       '/health_records/refresh',
       null,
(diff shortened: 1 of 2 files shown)


bshyong (approved):
LGTM

PR #7227, opened 2018-02-07
Fixed missing submit and fill dates on active subscriptions

U-DON:
Resolves department-of-veterans-affairs/vets.gov-team#8578.

Also fixes the missing bottom border of the table on the history page:
> screen shot 2018-02-06 at 8 10 01 pm


bshyong (approved):
LGTM

PR #7228, opened 2018-02-07
Appealsv2 expand accordion mg

markgreenburg:
- Updates common accordion (`CollapsiblePanel`) component to optionally take another prop, `startOpen`, which (predictably) determines whether the accordion will render as collapsed or expanded. Once rendered, functionality remains the same.
- This change should not affect the codebase that already uses the component.
- Adds more logic to the `Issues` component to take advantage of this new prop - the new logic should work such that the first instance of an accordion on the Issues tab will render as expanded (regardless of whether both 'Currently on appeals' and 'Closed' are rendered, or just one of them).
- Updates tests for both components.

One closed / granted issue only:
screen shot 2018-02-07 at 10 29 20 am

One active issue only:
screen shot 2018-02-07 at 10 27 29 am

Both active and closed issues:
screen shot 2018-02-07 at 10 23 45 am

### DIFF
--- a/src/js/claims-status/components/appeals-v2/Issues.jsx
+++ b/src/js/claims-status/components/appeals-v2/Issues.jsx
@@ -67,8 +67,9 @@ const Issues = ({ issues }) => {
 
   let activeItems = null;
   if (openListItems.length || remandListItems.length) {
+    // Active panel should always render as expanded by default (when items present)
     activeItems = (
-      <CollapsiblePanel panelName={'Currently on appeal'}>
+      <CollapsiblePanel panelName={'Currently on appeal'} startOpen>
         {openSection}
         {remandSection}
       </CollapsiblePanel>
@@ -77,8 +78,9 @@ const Issues = ({ issues }) => {
 
   let closedItems = null;
   if (grantedListItems.length || deniedListItems.length || withdrawnListItems.length) {
+    // Closed panel should render as expanded by default only if no active panel present
     closedItems = (
-      <CollapsiblePanel panelName={'Closed'}>
+      <CollapsiblePanel panelName={'Closed'} startOpen={!activeItems}>
         {grantedSection}
         {deniedSection}
         {withdrawnSection}
--- a/test/common/components/CollapsiblePanel.unit.spec.jsx
+++ b/test/common/components/CollapsiblePanel.unit.spec.jsx
@@ -1,33 +1,53 @@
 import React from 'react';
-import SkinDeep from 'skin-deep';
+import { shallow, mount } from 'enzyme';
 import { expect } from 'chai';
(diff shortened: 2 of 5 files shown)


src/js/claims-status/components/appeals-v2/Issues.jsx
```diff
@@ -77,8 +78,9 @@ const Issues = ({ issues }) => {
 
   let closedItems = null;
   if (grantedListItems.length || deniedListItems.length || withdrawnListItems.length) {
+    // Closed panel should render as expanded by default only if no active panel present
     closedItems = (
-      <CollapsiblePanel panelName={'Closed'}>
+      <CollapsiblePanel panelName={'Closed'} startOpen={!activeItems}>
```
cmgiven: :+1:

gnakm:
Could you make the padding inside the closed accordion the same as the padding in currently on appeal?
image

test/common/components/CollapsiblePanel.unit.spec.jsx
```diff
@@ -1,33 +1,53 @@
...
+    expect(wrapper.find('.usa-accordion-content').length).to.equal(0);
+
+    toggleButton.simulate('click');
+    expect(wrapper.find('.usa-accordion-content').length).to.equal(1);
+  });
 
-    tree.getMountedInstance().toggleChapter();
+  it('should call scrollToTop on toggle open', () => {
+    const scrollSpy = sinon.spy();
+    CollapsiblePanel.prototype.scrollToTop = scrollSpy;
```
cvalarida: Oooh, clever! Should probably reset it at the end of the test, though. Or maybe do this in a hook in case it fails?

The impact is minimal, I understand, but best practices, y'know?

cvalarida (approved):
Nicely done! Bonus points for converting the `CollapsiblePanel` unit test to enzyme.

test/common/components/CollapsiblePanel.unit.spec.jsx
markgreenburg: ahhh, that's a good point. Thanks for catching that!

markgreenburg:
@gnakm good catch! Here's an updated screengrab with equal 15px padding for both open & closed appeals: 
screen shot 2018-02-08 at 10 04 13 am

gnakm:
@markgreenburg Could we make it 20px padding and also could the headers inside the accordion (withdrawn, remand...) be in source sans pro font?

markgreenburg:
@gnakm how about the font for the 'remand' header inside the open appeals section? source as well?

gnakm:
@markgreenburg yes basically anything inside the accordion should be SSP

gnakm (approved):
lgtm! 🎉

PR #7229, opened 2018-02-07
update zoom and move logic / repair layout issues

rianfowler:
This is a PR to repair the following issues: 

- cropbox shrinks when wide image is used; see [this](https://github.com/department-of-veterans-affairs/vets-website/pull/7229#pullrequestreview-94587190) and [this](https://github.com/department-of-veterans-affairs/vets-website/pull/7229#discussion_r166514321).
- cropbox is not centered after changing the window size
- zoom slider does not allow end to end adjustment 

screen shot 2018-02-06 at 20 38 51
screen shot 2018-02-06 at 20 38 21
screen shot 2018-02-06 at 20 37 27

A couple of factors played into these issues:
- changing the container height after cropper is mounted is not supported by cropperjs
- the default behavior is to display the entire image when it is initially displayed- this behavior ignores values set for the cropper size 

Final result:
wide image demo
narrow image demo

### DIFF
--- a/src/js/vic-v2/components/PhotoField.jsx
+++ b/src/js/vic-v2/components/PhotoField.jsx
@@ -335,7 +335,7 @@ export default class PhotoField extends React.Component {
     // at this point, the onZoom event has not resized the canvas-
     // this forces the canvas back into move bounds if the zoom pulls a canvas edge into the cropBox
     //   after the canvas has rendered with new width values
-    window.requestAnimationFrame(this.maybeMoveCanvasWithinBounds.bind(this, {}, zoomWarn));
+    window.requestAnimationFrame(() => this.maybeMoveCanvasWithinBounds({}, zoomWarn));
   }
 
   setCropBox = () => {
--- a/src/js/vic-v2/components/PhotoField.jsx
+++ b/src/js/vic-v2/components/PhotoField.jsx
@@ -411,8 +411,6 @@ export default class PhotoField extends React.Component {
   }
 
   updateBoundaryWarningAndButtonStates = (moveBoundariesMet, zoomWarn = false) => {
-    // const moveBoundariesMet = getMoveBoundariesMet(this.refs.cropper.getCanvasData(), this.refs.cropper.getCropBoxData());
-
     this.setState({
       warningMessage: makeCropBoundaryWarningMessage({ zoomWarn, ...moveBoundariesMet }),
       ...makeMoveButtonsEnabledStates(moveBoundariesMet)
--- a/src/js/vic-v2/components/PhotoField.jsx
+++ b/src/js/vic-v2/components/PhotoField.jsx
@@ -24,31 +24,55 @@ const PhotoDescription = (
   </div>);
 
 const MIN_SIZE = 350;
-const MIN_RATIO = 0.2;
-const MAX_RATIO = 1.7;
 const WARN_RATIO = 1.3;
 const LARGE_SCREEN = 1201;
 const MAX_DIMENSION = 1024;
 
-function getMoveBoundariesMet(canvasData, cropboxData) {
-  const { height: canvasHeight, width: canvasWidth, left: canvasLeft, top: canvasTop } = canvasData;
-  const { height: cropboxHeight, width: cropboxWidth, left: cropboxLeft, top: cropboxTop } = cropboxData;
+function makeMovedCanvasData({ canvasData, cropBoxData, moveX, moveY }) {
+  const { left: canvasLeft, top: canvasTop, width: canvasWidth, height: canvasHeight } =  canvasData;
+  const { left: cropBoxLeft, top: cropBoxTop, width: cropBoxWidth, height: cropBoxHeight } =  cropBoxData;
 
-  // origin is at center
-  // coodinates are for canvas's container
-  const canvasBoundaries = {
-    minLeft: Math.round(-canvasWidth + cropboxLeft + cropboxWidth, 2),
-    maxLeft: Math.round(cropboxLeft, 2),
-    minTop: Math.round(-canvasHeight + cropboxTop + cropboxHeight, 2),
-    maxTop: Math.round(cropboxTop, 2),
+  const boundaries = {
+    leftMin: cropBoxLeft + cropBoxWidth - canvasWidth,
+    leftMax: cropBoxLeft,
+    topMin: cropBoxTop + cropBoxHeight - canvasHeight,
+    topMax: cropBoxTop
   };
 
-  return {
-    bottomBoundaryMet: Math.round(canvasTop, 2) === canvasBoundaries.minTop,
-    topBoundaryMet: Math.round(canvasTop, 2) === canvasBoundaries.maxTop,
-    leftBoundaryMet: Math.round(canvasLeft, 2) === canvasBoundaries.maxLeft,
-    rightBoundaryMet: Math.round(canvasLeft, 2) === canvasBoundaries.minLeft
+  const newCanvasData = {
+    left: moveX ? moveX + canvasLeft : canvasLeft,
+    top: moveY ? moveY + canvasTop : canvasTop,
+    bottomBoundaryMet: false,
+    topBoundaryMet: false,
+    leftBoundaryMet: false,
+    rightBoundaryMet: false
   };
+
+  if (boundaries.topMin === boundaries.topMax) {
+    newCanvasData.top = boundaries.topMin;
+    newCanvasData.topBoundaryMet = true;
+    newCanvasData.bottomBoundaryMet = true;
+  } else if (newCanvasData.top > boundaries.topMax) {
+    newCanvasData.top = boundaries.topMax;
+    newCanvasData.bottomBoundaryMet = true;
+  } else if (newCanvasData.top  < boundaries.topMin) {
+    newCanvasData.top = boundaries.topMin;
+    newCanvasData.topBoundaryMet = true;
+  }
+
+  if (boundaries.leftMin === boundaries.leftMax) {
+    newCanvasData.left = boundaries.leftMin;
+    newCanvasData.leftBoundaryMet = true;
+    newCanvasData.rightBoundaryMet = true;
+  } else if (newCanvasData.left > boundaries.leftMax) {
+    newCanvasData.left = boundaries.leftMax;
+    newCanvasData.rightBoundaryMet = true;
+  } else if (newCanvasData.left < boundaries.leftMin) {
+    newCanvasData.left = boundaries.leftMin;
+    newCanvasData.leftBoundaryMet = true;
+  }
+
+  return newCanvasData;
 }
 
 function isSmallScreen(width) {
@@ -156,12 +180,14 @@ export default class PhotoField extends React.Component {
   constructor(props) {
     super(props);
     this.state = {
-      src: null,
       cropResult: null,
-      zoomValue: 0.4,
       errorMessage: null,
+      minRatio: 0.2,
+      maxRatio: 1.7,
+      progress: 0,
+      src: null,
       warningMessage: null,
-      progress: 0
+      zoomValue: 0.4
     };
   }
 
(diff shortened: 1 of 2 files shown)


src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -437,19 +510,23 @@ export default class PhotoField extends React.Component {
...
+                src={this.state.src}
+                aspectRatio={1}
+                cropBoxMovable={false}
+                cropend={this.handleCropend}
+                cropmove={this.handleCropMove}
+                minContainerHeight={smallScreen ? 240 : 300}
+                toggleDragModeOnDblclick={false}
+                dragMode="move"
+                guides={false}
+                viewMode={0}
```
rianfowler: The old `viewMode` value was 1 which
> `restrict the crop box to not exceed the size of the canvas.`
The side effect is that the cropper size is adjusted to meet the default behavior of displaying the full image when it is initially loaded and the cropper size cannot be manually set. 

`viewMode={0}` allows the cropBox size to be manually set but does not constrain the canvas movement to the bounds of the cropBox necessitating handling this manually. 
screen shot 2018-02-06 at 20 38 51

src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -437,19 +510,24 @@ export default class PhotoField extends React.Component {
...
+                ref="cropper"
+                key={smallScreen ? 'smallScreen' : 'largeScreen'}
+                ready={this.setCropBox}
+                responsive
+                src={this.state.src}
+                aspectRatio={1}
+                cropBoxMovable={false}
+                cropend={this.handleCropend}
+                cropmove={this.handleCropMove}
+                minContainerHeight={smallScreen ? 240 : 300}
```
rianfowler: For wide images, the container will use the `canvasHeight` that was calculated to display the whole image on initial load as its minimum height. Once the minimum height is set, it cannot be updated without unmounting the component. 

Forcing the component to 
- meet the 240 / 300 height values in the styling, and 
- swapping the component `key` when the layout changes 

resolves the following issues:

screen shot 2018-02-06 at 20 53 25
screen shot 2018-02-06 at 19 58 07

Unmounting the component is janky but it only happens when the threshold is crossed.

src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -284,40 +310,87 @@ export default class PhotoField extends React.Component {
...
-        warningMessage: makeCropBoundaryWarningMessage({ zoomWarn, ...moveBoundariesMet }),
-        ...makeMoveButtonsEnabledStates(moveBoundariesMet)
-      });
-    } else {
-      e.preventDefault();
+    // force zoom within zoom bounds
+    if (zoomToBoundaryRatio) {
+      this.refs.cropper.zoomTo(zoomToBoundaryRatio); // force zoom within constraints
+      e.preventDefault(); // prevents bad zoom attempt
+      return; // don't update state until the subsequent zoom attempt
```
rianfowler: The new zoom ratio is provided to `onZoom` before the `canvas` is updated and rendered and is not limited by any factors. Sometimes the event's `zoomValue` is out of bounds before the canvas has reached the min/max zoom. 

E.g. The minZoom is 0.5, the canvas is at 0.6, and the zoom increment is -0.2. The desired zoom must be set to the minimum 0.5 rather than remain at 0.6. 

This short circuits the zoom event and triggers a new zoom event with the minimum or maximum zoom if the above scenario occurs.

src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -284,40 +310,87 @@ export default class PhotoField extends React.Component {
...
     }
+
+    // zoom value is good- update the state / messaging
+    this.setState({ zoomValue });
+
+    const zoomWarn = zoomValue >= WARN_RATIO;
+    // at this point, the onZoom event has not resized the canvas-
+    // this forces the canvas back into move bounds if the zoom pulls a canvas edge into the cropBox
+    //   after the canvas has rendered with new width values
+    window.requestAnimationFrame(this.maybeMoveCanvasWithinBounds.bind(this, {}, zoomWarn));
```
rianfowler: Sometimes zooming can cause the canvas to move out of bounds. This forces the canvas back into move bounds. 

The canvas is updated after the `onZoom` event. Since the min / max values are dependent on the size of the canvas, the movement limiting function is called on the next frame.

boundary met while zooming

src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -24,31 +24,55 @@ const PhotoDescription = (
   </div>);
 
 const MIN_SIZE = 350;
-const MIN_RATIO = 0.2;
```
rianfowler: Since `viewMode={0}` does not constrain the movement or zooming of the canvas, the `minRatio` must be dynamic- it's used to maintain the zoom boundary. 

Both ratios were moved onto the state for consistency.

src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -439,16 +510,19 @@ export default class PhotoField extends React.Component {
...
               aspectRatio={1}
               cropBoxMovable={false}
               cropend={this.handleCropend}
+              cropmove={this.handleCropMove}
+              minContainerHeight={smallScreen ? 240 : 300}
               toggleDragModeOnDblclick={false}
               dragMode="move"
               guides={false}
-              viewMode={1}
+              viewMode={0}
```
rianfowler: The old `viewMode` value was 1 which
> `restrict the crop box to not exceed the size of the canvas.`

The side effect is that the cropper size is adjusted to meet the default behavior of displaying the full image when it is initially loaded and the cropper size cannot be manually set. 

`viewMode={0}` allows the cropBox size to be manually set but does not constrain the canvas movement to the bounds of the cropBox necessitating handling this manually. 
screen shot 2018-02-06 at 20 38 51

src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -439,16 +510,19 @@ export default class PhotoField extends React.Component {
...
               ref="cropper"
+              key={smallScreen ? 'smallScreen' : 'largeScreen'}
               ready={this.setCropBox}
               responsive
               src={this.state.src}
               aspectRatio={1}
               cropBoxMovable={false}
               cropend={this.handleCropend}
+              cropmove={this.handleCropMove}
+              minContainerHeight={smallScreen ? 240 : 300}
```
rianfowler: For wide images, the container will use the `canvasHeight` that displays the entire image at once as its height. Once this minimum height is set, it cannot be updated without unmounting the component. 

There's an attempt to override the height of the container with CSS which results in the behavior below i.e.
- the cropbox is the calculated container height
- the container is displayed with the overridden height

The cropbox cannot be set to a greater height than the calculated container height.

Forcing the component to 
- meet the minimum 240 / 300 height values in the styling, and 
- swapping the component `key` when the layout changes 

resolves the following issues:

screen shot 2018-02-06 at 20 53 25
screen shot 2018-02-06 at 19 58 07

src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -284,40 +310,87 @@ export default class PhotoField extends React.Component {
...
+    } else {
+      this.refs.cropper.setCanvasData({
+        height: cropBoxHeight,
+        top: 0,
+        left: (containerWidth - (cropBoxHeight / oldCanvasHeight * oldCanvasWidth)) / 2
+      });
+    }
+
+    // with the canvas resized, use its dimensions to determine the min zoom ratio
+    const { width: newCanvasWidth } = this.refs.cropper.getCanvasData();
```
rianfowler: The `minRatio` is determined after the canvas is zoomed to the height or width of the canvas i.e. the initial zoom is set so that the image is bound to the top + bottom or right + left edges of the cropBox.

src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -284,40 +310,87 @@ export default class PhotoField extends React.Component {
...
+    const minRatio = newCanvasWidth / naturalWidth;
+    const slider = this.refs.slider;
+    slider.value = minRatio;
```
rianfowler: With the `minRatio` set dynamically, the full slider width can be used for zooming.

rianfowler:
tl;dr
e8b0bcf855a0fde7399d9b9351ba1050

src/js/vic-v2/components/PhotoField.jsx
jbalboni: This seems to allow users to adjust the size of the crop box, which is something we wanted to prevent.

rianfowler:
Here are the test images I used to ensure centering and to make overall testing easier. I haven't tested the output of the cropper and am open on idea how to accomplish that. My assumption is that the cropper crops to the outer border instead of the inner boarder since there's no dimension information about the border available.

horizontal grid
vertical grid

jbalboni:
@rianfowler Yes, it is to the outer border. Which we may want to consider changing (if that's possible), since the border is thick enough that it can be surprising when you see the preview image.

src/js/vic-v2/components/PhotoField.jsx
rianfowler: Good catch. I just made a change that sets `cropBoxResizable` to false. It removes the grey border- which I guess was there to represent a draggable element.
screen shot 2018-02-07 at 08 40 14

rianfowler:
I think this is ready. I'm going to get some feedback from design on the snap to boundaries behavior on drag but if there's a desired change there I'll put it in another PR.

jbalboni:
@rianfowler There's an odd looking undimmed area around the crop box now. Is there a way to get rid of that?

screen shot 2018-02-07 at 10 20 51 am

rianfowler:
@jbalboni Before that border represented the draggable part of the cropBox and since disabling it, the component removed the border. I'll look into it- it might be just a styling thing I can force to display

src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -24,31 +24,55 @@ const PhotoDescription = (
...
-  const { height: cropboxHeight, width: cropboxWidth, left: cropboxLeft, top: cropboxTop } = cropboxData;
-
-  // origin is at center
-  // coodinates are for canvas's container
-  const canvasBoundaries = {
-    minLeft: Math.round(-canvasWidth + cropboxLeft + cropboxWidth, 2),
-    maxLeft: Math.round(cropboxLeft, 2),
-    minTop: Math.round(-canvasHeight + cropboxTop + cropboxHeight, 2),
-    maxTop: Math.round(cropboxTop, 2),
+function makeMovedCanvasData({ canvasData, cropBoxData, moveX, moveY }) {
```
jbalboni: What about `getCanvasDataAfterMove` or `getCanvasDataForMove`? `makeMovedCanvasData` is a little confusing to me.

src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -337,48 +410,55 @@ export default class PhotoField extends React.Component {
...
       ...makeMoveButtonsEnabledStates(moveBoundariesMet)
     });
   }
 
   handleCropend = () => { // casing matches Cropper argument
-    this.updateBoundaryWarningAndButtonStates();
+    this.maybeMoveCanvasWithinBounds();
+  }
+
+  handleCropstart = (e) => {
```
rianfowler: @jbalboni Alrighty I got the black border to display again. The component "disables" moving the cropbox edges by `display: none`ing the border with css rather than by limiting the actions in js. I re-enabled the draggable cropbox but block the drag action from propagating in `handleCropstart`.

src/js/vic-v2/components/PhotoField.jsx
jbalboni: Why did you choose to use `bind` over

```js
window.requestAnimationFrame(() => this.maybeMoveCanvasWithinBounds({}, zoomWarn))
```

I find that slightly easier to read than using `bind`, but don't have a strong opinion about it.

src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -284,40 +310,87 @@ export default class PhotoField extends React.Component {
...
-      left: cropBoxLeftOffset,
+    const left = smallScreen ? (containerWidth / 2) - 120 : (containerWidth / 2) - 150;
+    this.refs.cropper.setCropBoxData({
       top: 0,
-      width: cropBoxSize,
-      height: cropBoxSize
-    };
-    cropper.setCropBoxData(cropBoxData);
+      left,
+      height: smallScreen ? 240 : 300,
```
jbalboni: Can we make 240 and 300 constants? Looks like we're using them in several places

src/js/vic-v2/components/PhotoField.jsx
rianfowler: `getCanvasDataForMove` is better. Will update.

jbalboni:
@rianfowler The code looks good. I'm noticing some glitchy behavior when resizing sometimes. If I zoom in and then resize the window a small amount, sometimes I get some empty space on the edge:

screen shot 2018-02-07 at 11 00 07 am

It snaps back to the correct position if I adjust the image after that.

src/js/vic-v2/components/PhotoField.jsx
rianfowler: Updated. For some reason I thought `this` wouldn't be maintained with the right context without binding it.

src/js/vic-v2/components/PhotoField.jsx
rianfowler: Good catch. Forgot to do this.

rianfowler:
@jbalboni Yeah I'm not completely happy with the window sizing stuff. I noticed that sometimes `getContainerData()` doesn't return accurate values but I have not been able to consistently recreate the issue. 

I had added another `ref` container around the `Cropper` for a bit that gave me better results but after isolating / explaining my changes yesterday it didn't look like I needed it anymore. I may try that again and see if I can fix this.

src/js/vic-v2/components/PhotoField.jsx
jbalboni: We're using ES.next syntax for class methods that does the binding:

`someMethod = () => { }`

vs

`someMethod() { }`

I think it would actually work without the latter in this case, too, but this is why you don't see more `bind` calls in our components.

src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -284,40 +310,90 @@ export default class PhotoField extends React.Component {
...
-    const cropBoxSize = smallScreen ? 240 : 300;
-    const cropBoxLeftOffset = (containerWidth - cropBoxSize) / 2;
-    const cropBoxData = {
-      left: cropBoxLeftOffset,
-      top: 0,
-      width: cropBoxSize,
-      height: cropBoxSize
-    };
-    cropper.setCropBoxData(cropBoxData);
+    window.requestAnimationFrame(() => {
```
rianfowler: @jbalboni This smoothed out the window resizing for me. I guess all of the events from `Cropper` happen before the canvas data is updated. Moving this logic into the next animation frame (i.e. when any pending layouts are complete) seems to smooth this out significantly. LMK if this improves the jankiness. I may do the same thing to other containerData dependent events.

jbalboni: It definitely looks smoother, though I still see the issue with it leaving empty space in the crop box after resizing sometimes.

rianfowler: Ah alright. I'm able to consistently recreate this. I think I know how to fix it.

- select image and start in small screen layout
- zoom into image a couple of times
- move image all of the way up
- make window larger

src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -284,40 +310,99 @@ export default class PhotoField extends React.Component {
...
+
+      // use the cropbox dimensions to force canvas into default position
+      const { height: cropBoxHeight, width: cropBoxWidth, left: cropBoxLeft } = this.refs.cropper.getCropBoxData();
+      const { width: oldCanvasWidth, height: oldCanvasHeight, naturalHeight, naturalWidth } = this.refs.cropper.getCanvasData();
+      // tall images take the full width of the cropBox
+      if (naturalHeight > naturalWidth) {
+        this.refs.cropper.setCanvasData({
+          width: cropBoxWidth,
+          left: cropBoxLeft,
+          top: 0
```
rianfowler: @jbalboni This resets the top coordinate on narrow images after the window is resized. The canvas will now reset to its default position when the window is resized. This should fix the positioning issue you saw. 

Also noticed that error warning messages were not being reset when the window was resized- added that as well.

jbalboni (approved):

cehsu (approved):
LGTM. Is it possible that the move controls are reversed?

PR #7235, opened 2018-02-07
Use backend url for photo preview on VIC

jbalboni:
I'll put up a screenshot when it's working against the backend, but these are the front end changes as I understand them.

TODO
- [x] Update review page description with url

### DIFF
--- a/src/js/common/schemaform/actions.js
+++ b/src/js/common/schemaform/actions.js
@@ -183,7 +183,6 @@ export function uploadFile(file, filePath, uiOptions, progressCallback) {
         if (req.status >= 200 && req.status < 300) {
           const body = 'response' in req ? req.response : req.responseText;
           const fileData = uiOptions.parseResponse(JSON.parse(body), file);
-          fileData.file = file;
           dispatch(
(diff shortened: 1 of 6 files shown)


cehsu (approved):

src/js/common/schemaform/actions.js
```diff
@@ -183,7 +183,6 @@ export function uploadFile(file, filePath, uiOptions, progressCallback) {
         if (req.status >= 200 && req.status < 300) {
           const body = 'response' in req ? req.response : req.responseText;
           const fileData = uiOptions.parseResponse(JSON.parse(body), file);
-          fileData.file = file;
```
cehsu: It looks like a test set up for the [screen reader changes ](https://github.com/department-of-veterans-affairs/vets-website/commit/98e1324c2ad872e04eae2565c619047bad6a4a80)needs to be removed as well.

https://github.com/department-of-veterans-affairs/vets-website/blob/c089c56c6e4b6893df3c1df9febb0ab48e7b8963/test/common/schemaform/actions.unit.spec.js#L362-L365

PR #7236, opened 2018-02-07
Appealsv2 fix status mg

markgreenburg:
- Change default return of `makeDurationText` to an object with empty strings. This enables us to conditionally render the time duration cards for `NextEvent`s only when we have a duration to give.
- Remove property reference from past event text description of `hearingHeld` events. The property getting referenced in this text snippet doesn't exist in the api, and there's no clear spec for what it will be called, so just changed text to something generic. Will open another issue to ensure our event mapping & content are complete and final in a broader sense.
- Fixes some tests that were asserting against the default return values of the `makeDurationText` function.

### DIFF
--- a/test/claims-status/utils/helpers.unit.spec.js
+++ b/test/claims-status/utils/helpers.unit.spec.js
@@ -537,17 +537,20 @@ describe('Disability benefits helpers: ', () => {
 
     it('should return an object with header and description properties with nonsense input', () => {
       const testText = makeDurationText(inputs.nonsense);
-      expect(!!testText.header && !!testText.description).to.be.true;
+      expect(testText.header).to.equal('');
+      expect(testText.description).to.equal('');
     });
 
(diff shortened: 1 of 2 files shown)


test/claims-status/utils/helpers.unit.spec.js
```diff
@@ -537,17 +537,20 @@ describe('Disability benefits helpers: ', () => {
...
-      expect(!!testText.header && !!testText.description).to.be.true;
+      expect(testText.header).to.equal('');
+      expect(testText.description).to.equal('');
     });
 
     it('should return an object with header and description properties with no input', () => {
       const testText = makeDurationText();
-      expect(!!testText.header && !!testText.description).to.be.true;
+      expect(testText.header).to.equal('');
+      expect(testText.description).to.equal('');
```
cvalarida: Ooh, this is much more sensible. Thank you.

cvalarida (approved):

PR #7239, opened 2018-02-08
Change zoom warning trigger for cropper

jbalboni:
I was noticing that I wasn't seeing zoom warnings when I uploaded relatively small images and zoomed in a lot. The threshold for a fuzzy image should probably be different depending on the size of image, since large images can be zoomed in a lot, but small images can only be zoomed in a little before getting a cropped image under 350px.

So, this changes the threshold to be a pixel size for the cropped image, rather than just a zoom ratio.

I could be missing some of the background of this warning, so please correct me if this something we've talked about and didn't want to do.


rianfowler (approved):
LGTM.

cehsu (approved):

PR #7241, opened 2018-02-08
Add PartialDowntimeNotification and actions to set downtime status

edpaget:
Resolves: https://github.com/department-of-veterans-affairs/vets.gov-team/issues/8123 not sure if we actually want to go through with it given the comments at the end of that issue though. 

However, since this is my first vets-website contribution wouldn't mind people telling me what I did wrong...


edpaget:
Sounds like this is being moved to another team.

ncksllvn:
Sorry to hear, looked like great work though @edpaget!

PR #7244, opened 2018-02-09
WIP layout for ID card preview

rianfowler:
Tried to make all of the positioning percentage based. The background / mockup isn't exactly 3:2 but think I've got it mostly right.

Some questions:

- ~~I'm not sure where to store the static assets. I found the dev watch is being pulled from a build/development folder~~
- ~~I need a VA logo with a transparent background. I got an .eps from @stacyhlinn but I'm not sure how to export it correctly.~~
- There are some differences between design and implementation that may be hard to reconcile e.g. font weights, Myriad font. Will follow-up with designers.
- Other questions inline
screen shot 2018-02-12 at 20 13 07
screen shot 2018-02-12 at 20 12 27
screen shot 2018-02-12 at 20 14 40
screen shot 2018-02-12 at 20 14 26
screen shot 2018-02-12 at 20 14 08
screen shot 2018-02-12 at 20 13 52

Here's the original for comparison. 
vic-2 3 cropped


src/js/vic-v2/components/VeteranIDCard.jsx
```diff
@@ -1,7 +1,34 @@
...
+
 export default function VeteranIDCard() {
+  // <img className="example-photo" alt="placeholder" src="/img/air-force-coat-of-arms.png"/>
   return (
-    <div style={{ paddingTop: '95px', backgroundColor: 'gray', height: '200px', width: '300px' }}></div>
+    <div style={{ marginTop: '50px', backgroundColor: 'grey', height, width: height * WIDTH_HEIGHT_RATIO }}>
+      <svg width={'100%'} height={'100%'}>
+        <image width={'100%'} href={'/img/vic-preview-background.svg'} />
+        <image height={'28.5%'} href={'/img/va-seal.png'} x={'6.5%'} y={'7.3%'} />
+        <text x={'26.3%'} y={'24.3%'} stroke={'none'} fill={'white'} fontFamily={'Arial, Helvetica, sans-serif'} fontSize={23}>
```
rianfowler: Not sure what font family to use.

src/js/vic-v2/components/VeteranIDCard.jsx
```diff
@@ -1,7 +1,34 @@
...
+          Veteran Identification Card
+        </text>
+        <image height={'51.8%'} href={'/img/example-photo-2.png'} x={'62.1%'} y={'34.1%'} />
+        <image height={'21.1%'} href={'/img/air-force-coat-of-arms.png'} x={'7.3%'} y={'70%'} />
+        <text x={'7.3%'} y={'55.2%'} stroke={'none'} fill={'white'} fontFamily={'Arial, Helvetica, sans-serif'} fontSize={23}>
+          Firstname Lastname
```
rianfowler: Open to ideas on how to detect breaking this into two lines. Would like to stick to using `<text>` svg elements but that could get tricky to do- maybe measuring text width with a canvas.

src/js/vic-v2/components/VeteranIDCard.jsx
```diff
@@ -1,7 +1,34 @@
...
+        </text>
+        <text x={'22.8%'} y={'78.9%'} stroke={'none'} fill={'white'} fontFamily={'Arial, Helvetica, sans-serif'} fontSize={15}>
+          Air Force
+        </text>
+        <text x={'22.8%'} y={'85.2%'} stroke={'none'} fill={'white'} fontFamily={'Arial, Helvetica, sans-serif'} fontSize={15}>
+          Honorable Discharge
+        </text>
+        <text x={'62.1%'} y={'91.9%'} stroke={'none'} fill={'white'} fontFamily={'Arial, Helvetica, sans-serif'} fontSize={12}>
+          ID no: 05P3400000000pz
```
rianfowler: Do we have more examples of what this ID will be?

rianfowler:
The VA logo was not aligned with the name / branch seal. Want to make sure this is intentional. @stacyhlinn 

screen shot 2018-02-08 at 20 07 55

src/js/vic-v2/components/VeteranIDCard.jsx
cehsu: Could we use number of characters?

src/js/vic-v2/components/VeteranIDCard.jsx
```diff
@@ -1,7 +1,34 @@
...
+const WIDTH_HEIGHT_RATIO = 1.583;
+const height = 300;
+
 export default function VeteranIDCard() {
+  // <img className="example-photo" alt="placeholder" src="/img/air-force-coat-of-arms.png"/>
   return (
-    <div style={{ paddingTop: '95px', backgroundColor: 'gray', height: '200px', width: '300px' }}></div>
+    <div style={{ marginTop: '50px', backgroundColor: 'grey', height, width: height * WIDTH_HEIGHT_RATIO }}>
+      <svg width={'100%'} height={'100%'}>
+        <image width={'100%'} href={'/img/vic-preview-background.svg'} />
```
cehsu: Is the number of branches/discharge statuses small enough that it could be worthwhile to pre-generate all possible versions of the card image so that we only need to specify the name, id, and profile photo on the appropriate pre-generated card background?

cehsu (commented):
This looks really good so far, pending the questions you mentioned.

src/js/vic-v2/components/VeteranIDCard.jsx
rianfowler: We could with a monospaced font. There's a pretty big width difference between 
WWW and 
lll

I know we can render the text to a canvas and measure the width but it might be apples and oranges. There's an package for [word wrapping](https://www.npmjs.com/package/word-wrapper) in the wild or we could just make our own since this use case is very limited. 

We could go further: bump the font size down and measure recursively if the two line name doesn't fit the width constraint or support three line names if someone has a long hyphenated last name. 

Might be helpful to have some actual long last names to test with.

src/js/vic-v2/components/VeteranIDCard.jsx
rianfowler: That would allow the removal 3 text tags and 2 image tags from our code.

rianfowler:
Could we show a preview of this on the image cropping page?

src/js/vic-v2/components/VeteranIDCard.jsx
jbalboni: I _think_ it'll be a case number like `00027471`

jbalboni:
@rianfowler I think the designers talked about that and decided the confirmation page was the best place.

jbalboni:
@rianfowler Assets can be in the same place as the example photos on the upload/intro page.

src/js/vic-v2/components/VeteranIDCard.jsx
```diff
@@ -1,7 +1,106 @@
 import React from 'react';
 
-export default function VeteranIDCard() {
-  return (
-    <div style={{ paddingTop: '95px', backgroundColor: 'gray', height: '200px', width: '300px' }}></div>
-  );
+const FONT_FAMILY = 'Arial, Helvetica, sans-serif';
+// sets the aspect ratio of svg
+const VIEW_BOX = '0 0 2048 1284';
```
rianfowler: Measured the mock ups and based the svg view box off of that.

src/js/vic-v2/components/VeteranIDCard.jsx
```diff
@@ -1,7 +1,106 @@
...
+  veteranFirstName,
+  veteranLastName,
+  veteranBranch,
+  veteranID
+}) => (
+  // svg preserves aspect ratio
+  <svg
+    fill={'white'}
+    fontFamily={FONT_FAMILY}
+    preserveAspectRatio={'xMidYMid meet'}
```
rianfowler: SVG's container determines its size. Aspect ratio will be maintained.

src/js/vic-v2/components/VeteranIDCard.jsx
```diff
@@ -1,7 +1,106 @@
...
 
-export default function VeteranIDCard() {
-  return (
-    <div style={{ paddingTop: '95px', backgroundColor: 'gray', height: '200px', width: '300px' }}></div>
-  );
+const FONT_FAMILY = 'Arial, Helvetica, sans-serif';
+// sets the aspect ratio of svg
+const VIEW_BOX = '0 0 2048 1284';
+// height of font top to bottom bounding box in the above viewbox
+const LARGE_FONT_SIZE = 90;
```
rianfowler: Font height is relative to the coordinate system set by the viewbox.

src/js/vic-v2/components/VeteranIDCard.jsx
```diff
@@ -1,7 +1,106 @@
...
+};
+
+function getIsLongName(name) {
+  try {
+    return document
+      .createElement('canvas')
+      .getContext('2d')
+      .measureText(name)
+    // manually determined max width
+      .width > 120;
```
rianfowler: Didn't find a way to do this in the context of the svg's coordinate system so I tested a few widths and got this number. Will test in a few browsers to determine if the value is consistent.

Below represents the max width 1 line name. 

screen shot 2018-02-12 at 20 12 27
screen shot 2018-02-12 at 20 13 07

src/js/vic-v2/components/VeteranIDCard.jsx
```diff
@@ -1,7 +1,106 @@
...
+            <text fontSize={LARGE_FONT_SIZE} key={'lastName'} x={'6.5%'} y={'58.8%'}>
+              {veteranLastName}
+            </text>
+          )
+        ] : (
+          <text fontSize={LARGE_FONT_SIZE} x={'6.5%'} y={'55.2%'}>
+            {`${veteranFirstName} ${veteranLastName}`}
+          </text>
+        )
+    }
```
rianfowler: This spacing is kind of gross but wanted to make sure it was clear the first condition was returning an array of elements.

src/js/vic-v2/components/VeteranIDCard.jsx
```diff
@@ -1,7 +1,106 @@
...
+
+    <image height={'21.1%'} href={imagePaths[veteranBranch]} x={'6.5%'} y={'70%'}/>
+    <text fontSize={SMALL_FONT_SIZE} x={'22.8%'} y={'78.9%'}>
+      {veteranBranch}
+    </text>
+    <text fontSize={SMALL_FONT_SIZE} x={'22.8%'} y={'85.2%'}>
+      Honorable Discharge
+    </text>
+
+    <image height={'51.8%'} href={'/img/example-photo-1.png'} x={'62.1%'} y={'34.1%'}/>
```
rianfowler: @jbalboni Not sure how the cropped image is going to be passed to his component.

rianfowler:
@stacyhlinn I've added new screenshots / questions to this PR for review.

src/js/vic-v2/components/VeteranIDCard.jsx
jbalboni: You will either need to take `photo.serverName` and `photo.serverPath` as props and use the `getImageUrl` function from PhotoField to build the url in here, or you can build the url in ConfirmationPage and pass that in as a prop.

src/js/vic-v2/components/VeteranIDCard.jsx
```diff
@@ -1,7 +1,106 @@
...
+  'Coast Guard': '/img/vic-cg-emblem.png',
+  'Air Force': '/img/vic-air-force-coat-of-arms.png',
+  Army: '/img/vic-army-symbol.png',
+  Navy: '/img/vic-navy-emblem.png',
+  'Marine Corps': '/img/vic-usmc-emblem.png',
+  previewBackground: '/img/vic-preview-background.svg',
+  VASeal: '/img/vic-va-seal.png'
+};
+
+function getIsLongName(name) {
```
jbalboni: Maybe `isLongName`?

src/js/vic-v2/components/VeteranIDCard.jsx
```diff
@@ -1,7 +1,106 @@
...
+
+const VeteranIDCard = ({
+  veteranFirstName,
+  veteranLastName,
+  veteranBranch,
+  veteranID
+}) => (
+  // svg preserves aspect ratio
+  <svg
+    fill={'white'}
```
jbalboni: Curious why you do this vs `fill="white"` here and elsewhere for string props.

src/js/vic-v2/components/VeteranIDCard.jsx
```diff
@@ -1,7 +1,106 @@
...
+      y={'91.9%'}>
+      ID no: {veteranID}
+    </text>
+  </svg>
+);
+
+VeteranIDCard.propTypes = {
+  veteranFirstName: React.PropTypes.string.isRequired,
+  veteranLastName: React.PropTypes.string.isRequired,
+  veteranBranch: React.PropTypes.oneOf(['Air Force', 'Army', 'Coast Guard', 'Marine Corps', 'Navy']),
```
jbalboni: The form data will have the code for these branches, rather than the full description. You'll have to get the labels for those codes in ConfirmationPage or in here (they're in the form config file, but we can pull them into another file), whatever you think is cleaner.

src/js/vic-v2/components/VeteranIDCard.jsx
rianfowler: Habit. Will update.

src/js/vic-v2/components/VeteranIDCard.jsx
rianfowler: Would I be able to test this locally?

rianfowler:
@stacyhlinn Do we want to display middle name if provided? If so, how do we want to name to break if it's long e.g. middle name on first line if it fits, middle name on 2nd line if it doesn't fit, not middle name if it doesn't fit on either.

rianfowler:
@stacyhlinn Do we want to show the name in all caps? Do we want to allow uncapitalized names? e.g. rian fowler instead of Rian Fowler.

rianfowler:
Do we want to limit what characters veterans can use for their name? e.g. I don't think we stop them from using emoji or other unicode characters.

src/js/vic-v2/components/VeteranIDCard.jsx
cehsu: Perhaps a mock server could be used similar to how the uploads are tested in the e2e tests.

stacyhlinn:
@rianfowler here are what @lalexanderson-dsva and I discussed:

- We'd like to  propose that we keep printing names in one line until there is no more space then line break. so it could be

     Firstname Lastname

     Firstname Middlename Lastname

     Firstname Middlename
     Lastname

     Firstname
     Middlename Lastname

- Please take out "honorable discharge" 
- Add Veteran after the branch name, so for ex: Army Veteran, Coast Guard Veteran...etc 
- Add "TEMPORARY" on the top right corner of the card
- Is it possible to reduce the font size of ID no a bit, so the length doesn't exceed the width of the photo?

vic-3-03

src/js/vic-v2/components/VeteranIDCard.jsx
rianfowler: Good idea- I'll check that out.

stacyhlinn:
@rianfowler 
- Agreed that we should allow other letters, but no emojis.
- I'd like to enforce first letter cap - Firstname Middlename Lastname

stacyhlinn:
@rianfowler 

Sorry I take back what I just wrote in the previous comment. After talking to @lalexanderson-dsva we realized there are scenarios where some names don't always start with a capitalized letter. Let's go ahead and print names in ALL CAPS to get around this problem.

rianfowler:
@stacyhlinn Updates:

- Added TEMPORARY to the upper right corner 
- Removed "Honorable Discharge"
- Adjusted branch name position and added "Veteran"
- Names are displayed in all caps 
- Middle name and suffix are included 
- Each name part wraps independently 
- Reduced font size of ID 

screen shot 2018-02-15 at 11 49 48
screen shot 2018-02-15 at 11 51 52
screen shot 2018-02-15 at 11 52 04
screen shot 2018-02-15 at 11 52 15
screen shot 2018-02-15 at 11 52 26
screen shot 2018-02-15 at 11 52 50

Would like thoughts on how to handle long names. It would be nice if we could match the behavior to the how the printed version will handle it. We have a thirty character limit on each name part which can be pretty long if the name has a lot of wide characters.

screen shot 2018-02-15 at 12 39 47

src/js/vic-v2/containers/ConfirmationPage.jsx
```diff
@@ -41,9 +46,23 @@ class ConfirmationPage extends React.Component {
...
+    } = form.data.veteranFullName;
+
+    const { serviceBranch, verified } = form.data;
+
+    const photoUrl = getImageUrl(form.data.photo);
+
     const submittedAt = moment(form.submission.submittedAt);
 
+    const veteranFullNameStr =
+      `${firstName.toUpperCase()} ${middleName.toUpperCase()} ${lastName.toUpperCase()} ${suffix.toUpperCase()}` // upper case name
```
rianfowler: Opted for passing the whole name string in rather than the name parts even though the veteranID card breaks them back down into parts. The number of name parts vary (2 minimum up to 3 typically but can be higher with space separated double-barrel last names) so I thought it'd be simpler if the ID component could just assume a space splittable string since the which name part they are isn't relevant to it.

src/js/vic-v2/containers/ConfirmationPage.jsx
```diff
@@ -41,9 +46,23 @@ class ConfirmationPage extends React.Component {
     const response = form.submission.response
       ? form.submission.response.attributes
       : {};
-    const { veteranFullName, verified } = form.data;
+    const {
+      first: firstName = '',
+      middle: middleName = '',
+      last: lastName = '',
```
rianfowler: Setting some defaults here so I don't have to check if exists when calling `toUpperCase()` later.

src/js/vic-v2/containers/ConfirmationPage.jsx
```diff
@@ -36,10 +41,24 @@ class ConfirmationPage extends React.Component {
     const { form, userSignedIn } = this.props;
     // If someone refreshes this page after submitting a form and it loads
     // without an empty response object, we don't want to throw errors
+    const {
+      first: firstName = '',
+      middle: middleName = '',
+      last: lastName = '',
+      suffix = ''
+    } = form.data.veteranFullName;
```
rianfowler: Setting some defaults here so I don't have to check if exists when calling `toUpperCase()` later.

src/js/vic-v2/containers/ConfirmationPage.jsx
```diff
@@ -36,10 +41,24 @@ class ConfirmationPage extends React.Component {
...
+    const { serviceBranch, verified } = form.data;
+
+    const photoUrl = getImageUrl(form.data.photo);
+
     const response = form.submission.response || {};
-    const { veteranFullName, verified } = form.data;
     const submittedAt = moment();
 
+    const veteranFullNameStr =
+      `${firstName.toUpperCase()} ${middleName.toUpperCase()} ${lastName.toUpperCase()} ${suffix.toUpperCase()}` // upper case name
```
rianfowler: Opted for passing the whole name string in rather than the name parts even though the veteranID card breaks them back down into parts. The number of name parts vary (2 minimum up to 3 typically but can be higher with space separated double-barrel last names) so I thought it'd be simpler if the ID component could just assume a space splittable string since the which name part they are isn't relevant to it.

src/js/vic-v2/containers/ConfirmationPage.jsx
```diff
@@ -49,7 +68,11 @@ class ConfirmationPage extends React.Component {
         {verified && userSignedIn && <div>
           <p>We’ll send you emails updating you on the status of your application. You can also print this page for your records. You should receive your Veteran ID Card by mail in about 60 days.<br/>
             In the meantime, you can print a temporary digital Veteran ID Card.</p>
-          <VeteranIDCard/>
+          <VeteranIDCard
+            veteranFullName={veteranFullNameStr}
+            veteranBranchCode={serviceBranch}
+            caseId={this.props.caseId}
```
rianfowler: Assuming this is a prop for now.

jbalboni: It'll be `response.caseId`.

src/sass/vic-v2.scss
```diff
@@ -99,6 +99,11 @@ input[type=range] {
   font-size: 1.07em !important;
 }
 
+.id-card-preview {
+  width: 500px;
+  height: 350px;
```
rianfowler: IE 10/11 require a height and width to be set on an svg or its container. Percentages can be used but the image gets scaled to the larger dimension and an svg with a fixed aspect ratio does something like this. 
screen shot 2018-02-16 at 10 26 52

rianfowler:
@jbalboni @cehsu This should be ready to merge. Remaining items that'll be handled in future PR:

- handle long names, and
- meeting printing requirements

src/js/vic-v2/containers/ConfirmationPage.jsx
```diff
@@ -49,7 +68,13 @@ class ConfirmationPage extends React.Component {
         {verified && userSignedIn && <div>
           <p>We’ll send you emails updating you on the status of your application. You can also print this page for your records. You should receive your Veteran ID Card by mail in about 60 days.<br/>
             In the meantime, you can print a temporary digital Veteran ID Card.</p>
-          <VeteranIDCard/>
+          <div className="id-card-preview">
+            <VeteranIDCard
+              veteranFullName={veteranFullNameStr}
+              veteranBranchCode={serviceBranch}
+              caseId={this.props.caseId}
```
rianfowler: assuming this will come in as a prop since it's no on the schema.

jbalboni: Think my comment got lost. It'll be `response.caseId`.

src/js/vic-v2/components/VeteranIDCard.jsx
```diff
@@ -1,7 +1,135 @@
...
+const FONT_FAMILY = 'Arial, Helvetica, sans-serif';
+// sets the aspect ratio of svg
+const VIEW_BOX = '0 0 2048 1284';
+// height of font top to bottom bounding box in the above viewbox
+const X_LARGE_FONT_SIZE = 90;
+const LARGE_FONT_SIZE = 80;
+const SMALL_FONT_SIZE = 60;
+const X_SMALL_FONT_SIZE = 55;
+
+const branchCodeLabels = {
```
jbalboni: Can you put this in the vic-v2/labels.js file, and import it from there? You can also update the form config file to pull the labels from there.

jbalboni (approved):
One minor thing, otherwise LGTM.

cehsu (approved):

PR #7248, opened 2018-02-09
Improve Selenium test speed and reliability on jenkins

edpaget:
Run selenium in separate docker containers to make sure we're using an up-to-date version of chrome for tests.


edpaget:
Cool this passed 3x times in a row, so I feel fairly confident the kinks are worked out.

patrickvinograd:
Tagging @jbalboni @cvalarida @U-DON as a spectrum of front end engineers for any comments about the changes to actual test cases

Jenkinsfile.pr
```diff
@@ -133,15 +133,11 @@ node('docker') {
     try {
```
wyattwalter: I can't find any reference to this file anywhere else. We should remove it, can easily put it back if something breaks.

test/discharge-wizard/00-required.e2e.spec.js
```diff
@@ -24,16 +24,22 @@ module.exports = E2eHelpers.createE2eTest(
       .setValue('input[name="1_branchOfService"]', '1')
       .click('input[name="1_branchOfService"]')
       .selectDropdown('2_dischargeYear', '2016')
+      .waitForElementPresent('input[name="4_reason"]', Timeouts.normal)
```
cvalarida: I'm generally okay with adding these kinds of waits to the e2e tests.

@jbalboni I forget how this one works. Would it make more sense for it to be `.waitForElementVisible`?

test/e2e/hca-helpers.js
```diff
@@ -58,8 +58,8 @@ function completeMilitaryService(client, data) {
   client
     .selectDropdown('root_lastServiceBranch', data.lastServiceBranch)
     .fillDate('root_lastEntryDate', data.lastEntryDate)
-    .selectDropdown('root_dischargeType', data.dischargeType)
-    .fillDate('root_lastDischargeDate', data.lastDischargeDate);
+    .fillDate('root_lastDischargeDate', data.lastDischargeDate)
+    .selectDropdown('root_dischargeType', data.dischargeType);
```
cvalarida: Swapping these around like this is probably okay, but if it was timing out before, I don't know if this will help. :man_shrugging:

test/save-in-progress/01-sip-load-fail.e2e.spec.js
```diff
@@ -37,16 +37,19 @@ module.exports = E2eHelpers.createE2eTest(
...
     client
+      .url(url)
+      .waitForElementVisible('body', Timeouts.normal)
       .mockData({
         path: '/v0/in_progress_forms/1010ez',
         verb: 'get',
         value: {},
         status: 404
       }, token)
+      .pause(500)
```
cvalarida: Would it be better to `.waitForElementVisible` instead of just pausing?

test/save-in-progress/01-sip-load-fail.e2e.spec.js
```diff
@@ -60,9 +63,10 @@ module.exports = E2eHelpers.createE2eTest(
         value: {},
         status: 401
       }, token)
+      .pause(500)
       .click('.usa-button-primary');
 
-    client.waitForElementVisible('.usa-alert-error', Timeouts.slow);
+    client.waitForElementPresent('.usa-alert-error', Timeouts.slow);
```
cvalarida: Interesting. Why wait for it to be present instead of visible?

test/save-in-progress/01-sip-review.e2e.spec.js
```diff
@@ -33,17 +33,18 @@ module.exports = E2eHelpers.createE2eTest(
...
-      .waitForElementVisible('body', Timeouts.normal)
+      .waitForElementVisible('.schemaform-sip-save-link', Timeouts.normal)
       .mockData({
         path: '/v0/in_progress_forms/1010ez',
         verb: 'put',
         value: {},
         status: 500
       }, token)
-      .click('.schemaform-sip-save-link');
+      .pause(500)
```
cvalarida: Hmm... :thinking: Do we need to pause if we're already waiting for it to be visible?

cvalarida (commented):
No red flags here. Just a few questions. :+1:

test/discharge-wizard/00-required.e2e.spec.js
edpaget: For most of these the reason is because I tried both and it worked when I used Present. I think the specific problem is that sometimes these elements appear off screen in in the selenium window. Confusingly this only sometimes seems to change the results of the `.waitForElementVisibleCall`.

test/e2e/hca-helpers.js
edpaget: I think this is actually a bug in the code website's. If you fill out the discharge type dropdown first. Then type the discharge date in, after typing the first two digits of the year the discharge type dropdown is hidden then reappears when the final digits of the year are added while losing its set value. It seems like the show/hide shouldn't be toggled until the year is has four digits (doubt there are many vets born in 980...).

test/save-in-progress/01-sip-load-fail.e2e.spec.js
edpaget: `.waitForElementVisible` didn't work. There is some sort of timing issue maybe with how event handlers are being setup by react where clicking on the link doesn't do anything until a few ms after the element has rendered on the page.

test/save-in-progress/01-sip-review.e2e.spec.js
edpaget: Same as here: https://github.com/department-of-veterans-affairs/vets-website/pull/7248#discussion_r167944540

test/save-in-progress/01-sip-load-fail.e2e.spec.js
edpaget: I think this is the same issue as https://github.com/department-of-veterans-affairs/vets-website/pull/7248#discussion_r167942598 where an element can be present but whether Selenium sees it as visible depends on a variety of factors including where it is in relation to the chrome window's viewport.

test/e2e/hca-helpers.js
jbalboni: Yeah, we could probably adjust that. You can't continue if you enter a year before 1900, so it doesn't end up causing any problems, but it's a little janky.

test/save-in-progress/01-sip-load-fail.e2e.spec.js
jbalboni: If that's the case, it's surprising that it's only an issue here and not more widespread.

I think it would be safer to wait for the element to be visible/present, then pause, then click. When we've added pauses to wait for something to render, it's usually bitten us later.

cvalarida: Yeah, that's really odd. :confused:

edpaget: Sure I don't mind adding a `.waitFor..` method. I'll also try to look into more if this is something odd selenium is doing.

edpaget: I  think I have a fix for the `.pauses()`. It turns out this is an issue with the element to click being offscreen (not sure if this issue is with Selenium or Chrome). Instead of pausing, there's a nightwatch method called `.getLocationInView` that makes us scroll to the element, which we can then successfully click on.

edpaget:
It turned out the issue the `.pause` solved was that the page was scrolling, while Selenium was trying to click on the save form link, which caused Selenium to misclick. I think I have resolved it by adding a test helper that disables the scroll and can be used in places where the scrolling causes an issue.

wyattwalter (approved):
This is a great improvement on build times and reliability! Thank you for working through this!

Should probably have someone else with more FE experience also sign-off, but I'm 👍 on the Docker and Jenkins related stuff.

package.json
```diff
@@ -205,6 +207,7 @@
     "blob-polyfill": "^2.0.20171115",
     "camelcase-keys-recursive": "^0.8.2",
     "downshift": "^1.22.5",
+    "electron": "^1.7.12",
```
jbalboni: Why did this change package names?

jbalboni (approved):
Had one question, but front end changes LGTM.

package.json
edpaget: Electron-prebuilt is deprecated. The electron package now contains the most recent built version of electron with the same api as the prebuilt package.

PR #7250, opened 2018-02-12
Use max-width to allow slider to shrink on VIC cropper

jbalboni:
The slider width was keeping the cropper from shrinking to fit into the review page box on smaller phones.

localhost_3001_veteran-id-card_apply_review-and-submit iphone 7


rianfowler (approved):
verified locally LGTM

PR #7251, opened 2018-02-12
Use case number instead of case id

jbalboni:
I can't test this locally so I don't have a screenshot, but we're going to display the case number instead of the case id for requests.


rianfowler (approved):

cehsu (approved):
LGTM. Is there a corresponding PR on the backend for this?

jbalboni:
@cehsu Not sure, @lihanli and I talked about it, but I will create an issue for it.

PR #7257, opened 2018-02-13
Appealsv2 api errors mg

markgreenburg:
To-Do:
- Add tests
- Double check the error types vs. actual vets-api options
- Open an issue for final content.


src/js/claims-status/containers/AppealInfo.jsx
```diff
@@ -60,13 +78,22 @@ export class AppealInfo extends React.Component {
...
-      appealContent = <LoadingIndicator message="Please wait while we load your appeal..."/>;
-    } else {
-      // Appeals finished loading but appeal not found, so the ID passed in the params
-      // doesn't match any appeals in Redux appeals array (at least not for this user)
-      // TO-DO: Get input from content team
+    } else if (v2Availability === AVAILABLE && !appeal) {
+      // Yes, we have your appeals. No, the one you requested isn't one of them.
       appealContent = <AppealNotFound/>;
+    } else if (v2Availability === USER_FORBIDDEN) {
+      appealContent = <div>Thou shal not pass!</div>;
```
cvalarida: > Thou shalt not pass.

or
> You shall not pass.

Gotta be biblical or Gandalf. Can't have it both ways. :wink: 

Okay, okay, I know it doesn't matter, but I couldn't resist. :nerd_face:

src/js/claims-status/reducers/appeals.js
```diff
@@ -4,27 +4,65 @@ import {
   SET_APPEALS,
   SET_APPEALS_UNAVAILABLE,
   FETCH_APPEALS_PENDING,
-  FETCH_APPEALS_SUCCESS
+  FETCH_APPEALS_SUCCESS,
```
cvalarida: :yum: VS Code loves extra commas! Vim for the win. :wink:

src/js/claims-status/reducers/appeals.js
```diff
@@ -4,27 +4,65 @@ import {
...
+    case FETCH_APPEALS_SUCCESS: // Appeals v1 and v2
+      // appeals are added to Redux store via FETCH_APPEALS_SUCCESS
+      // reducer in claims-list.js
       return _.merge(state, {
         appealsLoading: false,
         available: true,
+        v2Availability: AVAILABLE // New and improved! More bits of info!
       });
-    case SET_APPEALS:
+    case SET_APPEALS: // Appeals v1
```
cvalarida: Is this actually from v1? I don't remember.

src/js/claims-status/reducers/appeals.js
```diff
@@ -4,27 +4,65 @@ import {
...
     case FETCH_APPEALS_PENDING:
       return _.set('appealsLoading', true, state);
-    case FETCH_APPEALS_SUCCESS:
+    case FETCH_APPEALS_SUCCESS: // Appeals v1 and v2
+      // appeals are added to Redux store via FETCH_APPEALS_SUCCESS
+      // reducer in claims-list.js
       return _.merge(state, {
         appealsLoading: false,
         available: true,
+        v2Availability: AVAILABLE // New and improved! More bits of info!
```
cvalarida: Oof. Can we add in a `TODO` to clean this up when we go full v2? This is...less than desirable, but I understand why you'd do it this way.

I'm really wondering if we should just make a completely separate reducer for v2, actually. :thinking: Separation of concerns and all that.

src/js/claims-status/reducers/appeals.js
```diff
@@ -4,27 +4,65 @@ import {
...
       return _.set('available', true, state);
-    case SET_APPEALS_UNAVAILABLE:
-      // Maybe should set appealsLoading to false here too
+    case SET_APPEALS_UNAVAILABLE: // Appeals v1
       return _.set('available', false, state);
+    // Following are reducers for Appeals v2 error states  
+    case USER_FORBIDDEN:
+      return _.merge(state, {
+        appealsLoading: false,
+        v2Availability: USER_FORBIDDEN,
```
cvalarida: If we keep the v1 and v2 reducer functionality in the same spot, does it make sense to set `available` to `false` here for these error states?

src/js/claims-status/utils/appeals-v2-helpers.jsx
```diff
@@ -50,6 +49,16 @@ export const CLOSED_STATUSES = [
   STATUS_TYPES.otherClose
 ];
 
+// Action Types & Availability statuses
+// Note: excludes FETCH_APPEALS_SUCCESS / UNAVAILABLE because there are defined in actions
+// and used in v1 as well
```
cvalarida: Ugh. This is getting messy, isn't it?

src/js/claims-status/utils/appeals-v2-helpers.jsx
```diff
@@ -50,6 +49,16 @@ export const CLOSED_STATUSES = [
...
+export const USER_FORBIDDEN = 'USER_FORBIDDEN';
+export const RECORD_NOT_FOUND = 'RECORD_NOT_FOUND';
```
cvalarida: What do you think of appending `_ERROR` to these for consistency?

src/js/claims-status/utils/appeals-v2-helpers.jsx
```diff
@@ -50,6 +49,16 @@ export const CLOSED_STATUSES = [
...
+export const VALIDATION_ERROR = 'VALIDATION_ERROR';
+export const BACKEND_SERVICE_ERROR = 'BACKEND_SERVICE_ERROR';
+export const FETCH_APPEALS_ERROR = 'FETCH_APPEALS_ERROR';
+export const AVAILABLE = 'AVAILABLE';
```
cvalarida: What do you think of `APPEALS_AVAILABLE` for this?

test/claims-status/utils/appealsV2Actions.unit.spec.js
```diff
@@ -57,15 +63,26 @@ describe('getAppealsV2', () => {
...
+  const appealsErrors = {
+    403: USER_FORBIDDEN,
+    404: RECORD_NOT_FOUND,
+    422: VALIDATION_ERROR,
+    502: BACKEND_SERVICE_ERROR,
+    504: APPEALS_FETCH_ERROR // works for any unspecified error code
+  };
 
+  Object.keys(appealsErrors).forEach((code) => {
+    it(`Dispatches ${appealsErrors[code]} when GET fails with ${code}`, (done) => {
```
cvalarida: Clever. I like it!

test/claims-status/utils/appealsV2Actions.unit.spec.js
```diff
@@ -57,15 +63,26 @@ describe('getAppealsV2', () => {
...
+      global.fetch.returns(Promise.reject({
+        errors: [{ status: `${code}` }],
+      }));
+      const thunk = getAppealsV2();
+      const dispatch = sinon.spy();
+      thunk(dispatch)
+        .then(() => {
+          const action = dispatch.secondCall.args[0];
+          expect(action.type).to.equal(appealsErrors[code]);
+        }).then(done, done);
```
cvalarida: Thinking this through...

If the `expect` fails, will it be caught by the `.then(done, done)`? Or would it only be caught if there were a `.catch()`? I always forget when the error callback on the `.then()` is called.

Anyhow, if the `.then()` error callback is called, it'll call `done()` with no argument, which will tell the test it succeeded when it actually failed, so maybe just verify that.

cvalarida (commented):
My biggest concern is whether we should keep the v1 and v2 reducers together. It's confusing and I think will make it harder to figure out what to extract later when we get rid of v1.

Other than that, it's looking good. :+1:

test/claims-status/utils/appealsV2Actions.unit.spec.js
markgreenburg: Confirmed, works 👍

src/js/claims-status/utils/appeals-v2-helpers.jsx
markgreenburg: Yes, to all the above comments. This needs to be broken out into own reducers for v2, but I think that should be in its own PR.

cvalarida: Agreed.

src/js/claims-status/actions/index.jsx
```diff
@@ -93,9 +101,28 @@ export function getAppealsV2() {
...
       (appeals) => dispatch(fetchAppealsSuccess(appeals)),
-      (error) => {
-        Raven.captureMessage(`vets_appeals_v2_err_get_appeals ${error.message}`);
-        return dispatch({ type: SET_APPEALS_UNAVAILABLE });
+      (response) => {
+        const status = getStatus(response);
+        const action = { type: '' };
+        switch (status) {
+          case '403':
+            action.type = USER_FORBIDDEN_ERROR;
```
cvalarida: Hmm...It's a little weird to use these as both action types and availability status constants...

I'm not really sure what best practice is, though, so I won't complain. :man_shrugging:

cvalarida (approved):
Looks good!

src/js/claims-status/actions/index.jsx
markgreenburg: Yeah, that's a fair point. These will likely end up getting renamed when we split out v1 and v2.

PR #7258, opened 2018-02-13
VIC slider bug fixes for IE

cehsu:
These changes fix the zoom slider bug in IE, and make the styles consistent across different browsers.

A bug related to fieldset width is also fixed.

win10_ie_11 0 1
win10_ie_11 0

### DIFF
--- a/src/js/vic-v2/components/PhotoField.jsx
+++ b/src/js/vic-v2/components/PhotoField.jsx
@@ -357,20 +357,20 @@ export default class PhotoField extends React.Component {
   }
 
   onZoomMouseMove = (e) => {
-    if (this.state.mouseDown) {
+    if (this.mouseDown) {
       this.onZoomSliderChange(e);
     }
   }
 
   onZoomMouseDown = (e) => {
-    this.setState({ mouseDown: true });
+    this.mouseDown = true;
     if (e.target.value !== this.state.zoomValue) {
       this.onZoomSliderChange(e);
     }
   }
 
   onZoomMouseUp = () => {
-    this.setState({ mouseDown: false });
+    this.mouseDown = false;
   }
 
   onZoom = (e) => {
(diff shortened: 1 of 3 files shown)


rianfowler (approved):
Verified locally. Looks great

src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -356,6 +356,23 @@ export default class PhotoField extends React.Component {
...
   }
 
+  onZoomMouseMove = (e) => {
+    if (this.state.mouseDown) {
+      this.onZoomSliderChange(e);
+    }
+  }
+
+  onZoomMouseDown = (e) => {
+    this.setState({ mouseDown: true });
```
jbalboni: I don't think you need to use `setState`, I think you can make it an instance property (`this.mouseDown = true`). Re-rendering on these events isn't necessary, as far as I can tell.

src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -665,6 +682,9 @@ export default class PhotoField extends React.Component {
                 aria-valuemin={this.state.minRatio}
                 aria-valuemax={this.state.maxRatio}
                 aria-valuenow={this.state.zoomValue}
+                onMouseDown={this.onZoomMouseDown}
+                onMouseUp={this.onZoomMouseUp}
+                onMouseMove={this.onZoomMouseMove}
                 onInput={this.onZoomSliderChange}/>
```
jbalboni: If this isn't firing in IE, does that mean that using the arrow keys to adjust the slider is also not working?

src/sass/modules/_m-schemaform.scss
```diff
@@ -9,6 +9,12 @@ legend {
   padding: 0 0 .5em 0;
 }
 
+// Override IE
+fieldset {
```
jbalboni: This is causing other pages in the app to have much smaller input boxes.

jbalboni:
The slider looks oddly small in IE11 on Windows 8:

screen shot 2018-02-13 at 9 35 13 am

src/js/vic-v2/components/PhotoField.jsx
cehsu: Good catch, will replace onInput with onChange, which will capture keyboard input.

src/js/vic-v2/components/PhotoField.jsx
cehsu: Thanks, that's better, changed.

cehsu:
@jbalboni good catch regarding the width issue, unset (unsupported by IE), has been replaced with auto.
screen shot 2018-02-13 at 10 25 26 am

src/sass/modules/_m-schemaform.scss
cehsu: Good catch, width has been set to 100%.
Large screen
bs_win10_ie_11 0 2
Small screen
bs_win10_ie_11 0 3

jbalboni (approved):

PR #7259, opened 2018-02-13
Fix nav bug in IE

cehsu:
This change fixes a bug in IE.

bs_win10_ie_11 0 1
bs_win10_ie_11 0

### DIFF
--- a/src/sass/modules/_m-vet-nav.scss
+++ b/src/sass/modules/_m-vet-nav.scss
@@ -459,6 +459,7 @@ body.va-pos-fixed {
 
 @include media($medium-screen) {
   .child-menu-opened {
+    height: auto;
     height: initial;
   }
 }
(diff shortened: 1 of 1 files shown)


src/sass/modules/_m-vet-nav.scss
```diff
@@ -459,6 +459,7 @@ body.va-pos-fixed {
 
 @include media($medium-screen) {
   .child-menu-opened {
+    height: auto;
```
cehsu: This allows us to use initial if it is supported, and auto otherwise. 
An issue created to address all such issues has been created [here](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/8672).

cvalarida: What do you think about getting rid of the `height: initial` below? If I remember right, the `height: auto` should do the same thing in this case, yeah?

cvalarida (approved):
Oh IE...Why you gotta be like that? :disappointed: 

Good catch! It's a super elegant fix. Nicely done. :smile:

src/sass/modules/_m-vet-nav.scss
cehsu: Sounds good, changed.

PR #7260, opened 2018-02-13
Use new labels, hide state when not in list on VIC

jbalboni:
We have to support more state lists for our addresses to validate against Salesforce. This adds a custom address definition to support that. We're also now hiding the state field when there's not a state list, since SF doesn't support sending a state then.

This is specific to VIC and SF, so didn't seem like something we should change the existing address definition to support.

screen shot 2018-02-13 at 2 18 13 pm
screen shot 2018-02-13 at 2 17 53 pm


src/js/vic-v2/definitions/address.js
```diff
@@ -0,0 +1,136 @@
+import _ from 'lodash/fp';
```
jbalboni: This file is copied from the original definition and modified.

rianfowler (approved):

cehsu (approved):

PR #7264, opened 2018-02-13
Copy claims & appeals index page

cvalarida:
Nothing fancy here. Only thing of note was I got rid of the buildtype conditional and always query for v1 appeals on the original page and v2 on the v2 page.


markgreenburg (approved):
LGTM, hurray for not testing in prod!

PR #7273, opened 2018-02-14
Reusable rate limiting code for VIC v2

jbalboni:
I took the basic logic for the rate limiting in VIC v1 and made it into a separate component that can wrap an application.

screen shot 2018-02-15 at 4 03 41 pm


rianfowler (approved):
What would this be used for e.g. what would be setting `rateLimitedAuthed`?

jbalboni:
We have a settings.js file in production that we can change on the fly. So we would be adjusting that number to throttle how many people could get to the actual form, and serve those who didn't make the threshold a page we create that says something like "The form is very busy right now, try again later." 

VIC v1 did this and showed an email capture page instead of the application at times, so that we could capture emails when there was too much load on the application.

src/js/common/components/RateLimiter.jsx
```diff
@@ -0,0 +1,72 @@
...
+ *     rateLimitAuthed: 1,
+ *     rateLimitUnauthed: 1
+ *   }
+ * }
+ */
+export class RateLimiter extends React.Component {
+  constructor(props) {
+    super(props);
+    // 0 is nothing gets through, 1 is everything gets through
+    const { rateLimitAuthed = 1, rateLimitUnauthed = 1 } = window.settings[props.id] || {};
```
cehsu: Could we make this more consistent by either replacing `{}` with `{ rateLimitAuthed: 1, rateLimitUnauthed: 1 }` or removing the default values?

src/js/common/components/RateLimiter.jsx
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  render() {
+    const { state, bypassLimiter, renderLimitedContent, waitForProfile, children } = this.props;
+
+    if (waitForProfile && state.user.profile.loading) {
+      return <LoadingIndicator message="Loading your profile information..."/>;
+    }
+
+    const rateLimited = state.user.login.currentlyLoggedIn
```
cehsu: Should we set up rateNotLimited instead, since we only seem to check if it is not true?

src/js/common/components/RateLimiter.jsx
```diff
@@ -0,0 +1,72 @@
...
+ */
+export class RateLimiter extends React.Component {
+  constructor(props) {
+    super(props);
+    // 0 is nothing gets through, 1 is everything gets through
+    const { rateLimitAuthed = 1, rateLimitUnauthed = 1 } = window.settings[props.id] || {};
+    const randomizer = Math.random();
+
+    this.state = {
+      disableRateLimit: window.sessionStorage.getItem(`${props.id}_disableRateLimit`),
```
cehsu: Might rateLimitDisabled be somewhat clearer?

cehsu (approved):
LGTM. I had a few minor questions.

src/js/common/components/RateLimiter.jsx
jbalboni: That's not quite the same thing. It would end up with undefined values when there's a vic object in the settings, but no rate limit properties. I don't want to assume that nothing else will go in settings objects created for particular apps.

peggygannon (approved):
@jbalboni I tweaked that last sentence since we removed the email stuff. Looks good. thank you

PR #7274, opened 2018-02-14
Appeals v2 index

cvalarida:
image
These are from the `/v0/appeals_v2` endpoint. They're not styled right and the status seems funny, but that's to be expected for now. https://github.com/department-of-veterans-affairs/vets.gov-team/issues/7060 will cover that.

Also....I don't know what this "Recently closed" bit is all about either. :grimacing: That can be addressed separately, though.


markgreenburg (approved):

cvalarida:
Oh, I should have mentioned that the last commit fixed the "Recently closed" warning, so that's no longer an issue.

PR #7275, opened 2018-02-14
Test against WCAG 2.0 AA

raquelromano-gov:
- update default rule set to include `wcag2aa`
- add a nightwatch command for testing against the `best-practice` ruleset, for future accessibility improvements: `npm run:best-practice`

### DIFF
--- a/config/nightwatch.js
+++ b/config/nightwatch.js
@@ -62,7 +62,7 @@ module.exports = {
     },
     wcag2a: {
       globals: {
-        rules: ['section508', 'wcag2aa']
+        rules: ['section508', 'wcag2a', 'wcag2aa']
       }
     }
   }
--- a/config/nightwatch.js
+++ b/config/nightwatch.js
@@ -62,7 +62,10 @@ module.exports = {
     },
     wcag2a: {
       globals: {
-        rules: ['section508', 'wcag2a', 'wcag2aa']
+        runOnly: {
+          type: 'tag',
+          values: ['section508', 'wcag2a', 'wcag2aa']
+        }
       }
     }
   }
--- a/config/nightwatch.js
+++ b/config/nightwatch.js
@@ -62,7 +62,7 @@ module.exports = {
     },
     wcag2a: {
       globals: {
-        rules: ['section508', 'wcag2a']
+        rules: ['section508', 'wcag2aa']
       }
     }
   }
--- a/config/nightwatch.js
+++ b/config/nightwatch.js
@@ -60,12 +60,9 @@ module.exports = {
     accessibility: {
       filter: './test/accessibility/*.spec.js'
     },
-    wcag2a: {
+    bestpractice: {
       globals: {
-        runOnly: {
-          type: 'tag',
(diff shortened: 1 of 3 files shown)


config/nightwatch.js
```diff
@@ -62,7 +62,7 @@ module.exports = {
     },
     wcag2a: {
       globals: {
-        rules: ['section508', 'wcag2a']
+        rules: ['section508', 'wcag2a', 'wcag2aa']
```
cehsu: Listing both is recommended in the second example provided [here](https://github.com/dequelabs/axe-core/blob/master/doc/API.md#options-parameter-examples).

test/pre-need/00-all-fields.e2e.spec.js
```diff
@@ -59,7 +59,7 @@ const runTest = E2eHelpers.createE2eTest(
       client.assert.cssClassPresent('.progress-bar-segmented div.progress-segment:nth-child(4)', 'progress-segment-complete');
       PageHelpers.completeBenefitSelection(client, testData.data.application);
       // do not run 'wcag2a' rules because of open aXe bug https://github.com/dequelabs/axe-core/issues/214
-      client.axeCheck('.main', { rules: ['section508'] })
+      client.axeCheck('.main', { rules: { 'aria-required-children': { enabled: false } } })
```
cehsu: I looks like we've been disabling entire rulesets rather than individual rules ([reference](https://www.axe-core.org/docs/#configure)).

raquelromano-gov: This is much better--thanks! There are other places we disable the whole set. Do you want to include finding and fixing them all in this PR or separately?

cehsu: Good idea, separate is better. I'll remove these changes.

raquelromano-gov: I'm not sure it's true we've been disabling rules in these tests. This `rules` array is being passed into our axe wrapper and supplied as part of the `run-only` configuration which seems consistent with the API docs:
https://github.com/department-of-veterans-affairs/vets-website/blob/master/test/util/nightwatch-commands/axeCheck.js#L26

So passing in `{ rules: ['section508'] }` would override the default `run-only` config, and not run the WCAG 2.0 rules.

I still think your idea to be more selective in what we disable is a great one, but maybe what should happen is we pass in an `exceptions` or `disable` array, and then in aXeCheck.js, add a `rules` value in addition to the `run-only`.

raquelromano-gov:
This is ready for review again.

I also added a command for testing against the `best-practice` ruleset but I'm not going to file a ticket for upgrading to that since a large number of the failures are that heading levels should only increase by one ([axeapi rule](https://dequeuniversity.com/rules/axe/1.1/heading-order)), and there are places all over our code where we throw in an `<h4>` for warning text.

Still, it could be helpful to be able to run `npm run test:best-practice` on a particular application and see what's there.

cehsu (approved):

PR #7280, opened 2018-02-15
Screen reader unit tests for PhotoField

jbalboni:
I attempted to write these with DefinitionTester, but never really got it to work correctly. So they're written as shallow rendered component tests to get around some of the mocking we'd have to do otherwise.


rianfowler (approved):

rianfowler:
Do we want to refactor `PhotoField` into smaller components before moving to VRE?

cehsu:
Thanks for the reminder @rianfowler, issue created [here](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/8805).

src/js/common/utils/helpers.js
```diff
@@ -153,10 +153,12 @@ export function scrollToFirstError() {
 }
 
 export function scrollAndFocus(errorEl) {
-  const currentPosition = window.pageYOffset || document.documentElement.scrollTop || document.body.scrollTop || 0;
-  const position = errorEl.getBoundingClientRect().top + currentPosition;
-  Scroll.animateScroll.scrollTo(position - 10, getScrollOptions());
-  focusElement(errorEl);
+  if (errorEl) {
```
cehsu: Should we be using refs to obtain this instead? If so, we could add this to the refactor ticket.

cehsu (approved):

jbalboni:
@rianfowler Let's do the VRE form we have in progress before the PhotoField refactoring.

src/js/common/utils/helpers.js
jbalboni: Yeah, we should probably use refs for the error message in PhotoField. I'll add it to the refactor ticket.

PR #7282, opened 2018-02-15
Appeals v2 index styling

cvalarida:
image
**Note:** The status needs to be updated and the appeal type needs to be run through a `programArea` -> `Human-friendly display` map. Also, that second appeal had a `programArea` of `null`, so I don't know what to do with that. I'm guessing that won't happen in production.

image


markgreenburg (approved):
LGTM

PR #7283, opened 2018-02-15
Fix VIC front end save in progress issues

jbalboni:
A couple issues here. The simpler one is that we hadn't added VIC to some of the profile config. Which sort of calls into question how useful those tests are. We had to ignore them initially, which meant that we could have forgotten to update them and gone into prod with it misconfigured.

screen shot 2018-02-15 at 2 55 40 pm

The second issue is that we weren't saving applications in progress after file uploads (we normally save after any change). This is because the save call is made in a component that intercepts onChange calls, and the file upload process dispatches change actions directly. I've updated this to pass an onChange function instead of a file path to the uploadFile function. 

I think this is a good path for the schema form code in general. We should keep Redux as a dependency of the top level components and leave the fields to be as purely React based as we can. Still some work to do there, but I think that lets us build a form library that has fewer and clearer dependencies.


test/user-profile/helpers.unit.spec.js
```diff
@@ -32,13 +33,13 @@ const schemaToConfigIds = {
...
   '40-10007': '40-10007',
+  VIC: 'VIC',
   definitions: 'N/A'
 };
 
 const excludedForms = new Set([
   '28-1900',
-  '28-8832',
-  'VIC'
+  '28-8832'
```
cehsu: Should we consider using the metalsmith config's ignorelist to keep this list current?

rianfowler (approved):

cvalarida (approved):

cehsu (approved):

src/js/common/schemaform/fields/FileField.jsx
```diff
@@ -35,10 +35,9 @@ export default class FileField extends React.Component {
       if (idx === null) {
         idx = files.length === 0 ? 0 : files.length;
       }
-      const filePath = this.props.idSchema.$id.split('_').slice(1).concat(idx);
       this.props.formContext.uploadFile(
         event.target.files[0],
-        filePath,
+        (file) => this.props.onChange(_.set(idx, file, this.props.formData || [])),
```
cehsu: Should `||[]` be moved outside the closing `)` for `_.set`?

test/user-profile/helpers.unit.spec.js
jbalboni: That assumes that we wouldn't launch a form without SiP, which is maybe true, but not sure I want to require us to do that.

PR #7284, opened 2018-02-15
Random tests

edpaget:
I noticed while investigating issues @U-DON was having with the mhv-controller-refactor branch that there are a fair number of tests that either depended on or changed and didn't clean up `window` or `global` variables, or mutated shared data between tests. Neither of these seem very good for testing in isolation, so I found a library [Choma](https://www.npmjs.com/package/choma) that randomizes the order the tests are run in and fixed any tests that could be affected by other tests.


edpaget:
Ah didn't realize the tests are run through a different mechanism in Jenkins :\

test/helper.js
```diff
@@ -0,0 +1,3 @@
+import setupJSDom from './util/mocha-setup';
+
+beforeEach(setupJSDom);
```
jbalboni: If you're doing this, should we remove the setupJSDom call that's in mocha-setup, and also remove mocha-setup from the mocha opts file?

Also, does it not work if you call beforeEach in the mocha-setup file?

jbalboni (approved):
I didn't look at the build failure, but the changes here look ok to me.

test/helper.js
edpaget: files that are required through mocha.opts don't have the mocha dsl setup yet. I founded I needed to do it this way (including keeping the setupJSDom call) because some of the api client code  references window.location on load and would fail without JSDom being setup.

edpaget:
it seems be somewhere between 20-50 if you run them randomly. Not sure what exactly is different about the coverage runner that it's not including the helper file's beforeEach.

cvalarida:
At the risk of stating what's already known, I'll record my discoveries here as I poke around.

I did a `console.log` in `setupJSDom` to see how often it was being called, and it looks like it's getting called before each test when running `npm run test:unit`, but not while using `nyc` in `npm run test:coverage`. Perhaps `nyc` (or istanbul?) doesn't run tests quite like we expect it to...

:thinking:

edpaget:
got it! Turned out the cli args needed to be in a slightly different order when running in nyc.

cvalarida:
:boom: I was wondering if it were something like that. Nicely done!

PR #7285, opened 2018-02-15
Add analytics to letters download

cvalarida:
~This space intentionally left blank.~

I think this is all the analytics events that need to get added, but I'm open to suggestions if anybody things we should add more.

### DIFF
--- a/src/js/letters/actions/letters.js
+++ b/src/js/letters/actions/letters.js
@@ -136,6 +136,10 @@ export function getLetterPdf(letterType, letterName, letterOptions) {
   }
 
   return (dispatch) => {
+    window.dataLayer.push({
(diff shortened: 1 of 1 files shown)


markgreenburg:
We have a failure event pushed for `getLetterList` but nothing for `getBenefitSummaryOptions`, which we might want to do since they're both used in `getLetterListAndBSLOptions` (could even move the code for both error types into the `getLetterListAndBSLOptions`, similar to how we're recording things to Sentry).

Also, not sure if this is needed, but the only other action creators that are left out are `getAddressCountries` and `getAddressStates` - might want to add events for both of those failure types as well, since they're logged to Sentry.

markgreenburg (approved):
Looks good!

PR #7286, opened 2018-02-16
Appealsv2 index notabs mg

markgreenburg:
- Removes tabbed interface from index page
- Breaks out appeals, claims actions and reducers into v2 file
- Re-enables sorting (by last updated date, desc. order) and pagination (10 rows per page)
- Consolidates appeals, claims, open / closed into one list

Needs:
- Tests
- Refactor to use reselect so that we don't recompute the view on every render

NOTE:
This PR is intended to Make Things Work for the purposes of usability testing. It needs a follow-up PR or two to straighten out some tests and refactor a bunch of things.


src/js/claims-status/containers/YourClaimsPageV2.jsx
```diff
@@ -105,27 +80,41 @@ class YourClaimsPageV2 extends React.Component {
...
+    // TO-DO: Figure out if this is still something we want
+    // if (canAccessClaims) {
+    //   if (claimsAvailable !== claimsAvailability.AVAILABLE) {
+    //     return <ClaimsUnavailable/>;
+    //   } else if (!claimsAuthorized) {
+    //     return <ClaimsUnauthorized/>;
+    //   }
+    // }
+
+    if (canAccessClaims && claimsAvailable !== claimsAvailability.AVAILABLE) {
```
cvalarida: We talked about this. The conclusion we came to is to check if the claims (or appeals) are still loading before rendering the error message.

src/js/claims-status/containers/YourClaimsPageV2.jsx
```diff
@@ -134,81 +123,36 @@ class YourClaimsPageV2 extends React.Component {
...
-      unfilteredAppeals,
-      unfilteredClaims,
       list,
       pages,
       page,
       claimsLoading,
       appealsLoading,
       show30DayNotice,
-      route,
-      synced
```
cvalarida: For feature parity, I think we keep this in for now until we figure out if we can remove it.

src/js/claims-status/containers/YourClaimsPageV2.jsx
```diff
@@ -134,81 +123,36 @@ class YourClaimsPageV2 extends React.Component {
...
-                      onValueChange={this.handleSort}/>
-                  </div>
-                  {innerContent}
-                </div>
-              }
-            </div>
-          ))}
-        </div>
-      );
+      content = (<div className="va-tab-content">{content}</div>);
```
cvalarida: Does it make sense to have `va-tab-content` without a va tab?

src/js/claims-status/utils/appeals-v2-helpers.jsx
```diff
@@ -50,14 +53,36 @@ export const ISSUE_STATUS = {
 };
 
 // Action Types & Availability statuses
-// Note: excludes FETCH_APPEALS_SUCCESS / UNAVAILABLE because there are defined in actions
+// TO-DO: Separate action types and availability statuses
+// Note: excludes FETCH_APPEALS_SUCCESS because there are defined in actions
```
cvalarida: > False

-- Dwight Schrute

src/js/claims-status/utils/appeals-v2-helpers.jsx
```diff
@@ -1138,3 +1163,69 @@ export const getStatus = (response) => {
...
+
+/**
+ * 
+ * @param {Object} item1
+ * @param {Object} item2
+ * @returns {-1|1|0}
+ */
+export function sortByLastUpdated(item1, item2) {
+  const lastUpdatedDate1 = getDate(item1);
+  const lastUpdatedDate2 = getDate(item2);
```
cvalarida: Might it make more sense to use `moment` up here to avoid parsing each date twice?

src/js/claims-status/utils/appeals-v2-helpers.jsx
```diff
@@ -1138,3 +1163,69 @@ export const getStatus = (response) => {
...
+};
+
+/**
+ * 
+ * @param {Object} item
+ * @returns {string}
+ */
+const getDate = (item) => {
+  if (!item.attributes) {
+    return '0';
```
cvalarida: Using this, it may end up at 1970-01-01. To ensure this ends up at the bottom of the pile, what do you think of returning the date that VA was founded?

src/js/claims-status/utils/appeals-v2-helpers.jsx
```diff
@@ -1138,3 +1163,69 @@ export const getStatus = (response) => {
...
+// Series of utility functions to sort claims and appeals by last updated date
+/**
+ * 
+ * @param {Object} appeal
+ * @returns {string}
+ */
+const getAppealDate = (appeal) => {
+  const { events } = appeal.attributes;
+  const dateString = (events && events.length)
+    ? events[events.length - 1].date
```
cvalarida: @cmgiven Are the events in `appeal.attributes.events` always sorted by date (where the last item in the array is the most recent)?

test/claims-status/components/YourClaimsPageV2.unit.spec.jsx
```diff
@@ -6,31 +6,25 @@ import sinon from 'sinon';
 import { YourClaimsPageV2 } from '../../../src/js/claims-status/containers/YourClaimsPageV2';
```
cvalarida: Might make sense to have a sort test... :thinking:
I hate to say this, but just to get this out the door, maybe we ignore it for now if it works with manual testing. We can write the test after everything needed for UAT gets merged.

cvalarida (approved):
Looks good as it is, but I think we should make some tickets to address some of these things after this gets merged in. #Deadline :disappointed:

markgreenburg:
@cvalarida noted - merged with prejudice :)
Will write up issues to get these issues taken care of

PR #7288, opened 2018-02-16
Display preview image from file data instead of url, handle failover

jbalboni:
In progress changes to use in memory image data for previews, and handle errors if we request an image from the backend that is still processing and fails.

screen shot 2018-02-19 at 2 01 38 pm


cehsu (approved):

src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -210,6 +216,17 @@ export default class PhotoField extends React.Component {
...
   }
 
+  componentWillReceiveProps(nextProps) {
+    const nextFormData = nextProps.formData || {};
+    const prevFormData = this.props.formData || {};
+    if (nextFormData.file instanceof Blob && nextFormData.file !== prevFormData.file) {
+      if (this.state.previewSrc) {
+        window.URL.revokeObjectURL(this.state.previewSrc);
+      }
+      this.setState({ previewSrc: window.URL.createObjectURL(nextFormData.file) });
```
cehsu: Is there an advantage of using a this over the file itself as the preview source?

rianfowler (approved):
LGTM.

src/js/vic-v2/components/PhotoField.jsx
jbalboni: As far as I could find, this is what you need to do to use the file itself as the preview source.

jbalboni:
@cehsu @rianfowler This could use another look.

src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -579,12 +612,29 @@ export default class PhotoField extends React.Component {
...
+    this.setState({ previewSrc: src });
+  }
+
   render() {
     const { formData, formContext } = this.props;
     const file = formData || {};
     const onReview = formContext.reviewMode;
 
-    if (onReview) return <PhotoReviewDescription url={getImageUrl(file)}/>;
+    if (onReview) {
```
rianfowler: What is `reviewMode`?

src/js/common/schemaform/helpers.js
```diff
@@ -254,8 +254,8 @@ export function stringifyFormReplacer(key, value) {
     }
 
     // Exclude file data
-    if (value.confirmationCode && value.serverPath) {
-      return _.omit(['serverPath', 'serverName'], value);
+    if (value.confirmationCode && (value.serverPath || value.file)) {
```
rianfowler: What does `stringifyFormReplacer` do?

src/js/common/schemaform/review/ArrayField.jsx
```diff
@@ -215,6 +216,7 @@ class ArrayField extends React.Component {
                         title={pageTitle}
                         hideTitle
                         name={fieldName}
+                        formContext={formContext}
```
rianfowler: Can you summarize how this is being passed down now i.e. the hierarchy of components that it's going through? I can see it getting passed through a few schemaform components but I don't see a new utilization of it.

src/js/vic-v2/components/PhotoPreview.jsx
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  render() {
+    const { src, id, className, isLoggedIn, processing } = this.props;
+
+    if (processing) {
+      return (
+        <div className="usa-alert usa-alert-warning vic-processing-warning">
+          <div className="usa-alert-body">
+            <h4 className="usa-alert-title">We're still processing your preview photo</h4>
```
rianfowler: My understanding is that we send the photo to our service, that service processes it, and we download the image and use that as the preview.

src/js/common/schemaform/helpers.js
jbalboni: This is used by `transformForSubmit` when stringifying the form data to be submitted to the backend. It's the second parameter for `JSON.stringify`, which lets you modify the data as it's being stringified.

rianfowler (approved):
LGTM- added some questions for my own understanding

src/js/common/schemaform/review/ArrayField.jsx
jbalboni: `RoutableSavableReviewPage > ReviewPage > ReviewCollapsibleChapter > SchemaForm` and `RoutableSavableReviewPage > ReviewPage > ReviewCollapsibleChapter > ArrayField > SchemaForm` 

Previously we had added to SchemaForm and passed it from `RoutedSavablePage > FormPage > SchemaForm`, so this is to do the wiring for the review page.

src/js/vic-v2/components/PhotoField.jsx
jbalboni: On the review page, we render every field in a read only view. Normally there are separate widgets for that (the SchemaForm component choose between them depending on a prop), but for custom fields it's easier to switch between them in the Field component.

src/js/vic-v2/components/PhotoPreview.jsx
jbalboni: That's what we were doing before, but now the image re-encoding that happens on the backend is done async, so it's possible that we won't be able to get the preview image immediately. Mostly this doesn't matter, because we still have image data in memory, but if you save your form, leave, and come back, it's possible that we'll still be processing.

PR #7289, opened 2018-02-16
enforce max crop size

rianfowler:


src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -254,7 +255,11 @@ export default class PhotoField extends React.Component {
   onDone = () => {
     this.setState({ isCropping: false, progress: 0, warningMessage: null });
 
-    this.refs.cropper.getCroppedCanvas().toBlob(blob => {
+    const croppedCanvasOptions = this.refs.cropper.getData().width > MAX_CROPPED_HEIGHT_WIDTH ?
+      { width: MAX_CROPPED_HEIGHT_WIDTH, height: MAX_CROPPED_HEIGHT_WIDTH } :
+      {};
+
+    this.refs.cropper.getCroppedCanvas(croppedCanvasOptions).toBlob(blob => {
```
rianfowler: `cropperjs` documents `maxWidth` and `maxHeight` options but my testing demonstrated it doesn't work. 

I split this blob out into an image tag and downloaded the image to verify that the image is never greater than 600. Using the maxes would not enforce any change.

There are [smoothing](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/imageSmoothingQuality) [options](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/imageSmoothingEnabled) available that I'm not using here. I can do some research or defer to someone else on this.

jbalboni: Might be worth trying out the smoothing quality, even if it is basically Chrome only. That's the main degradation in quality from resizing outside of the browser, at least when I compare them.

src/sass/vic-v2.scss
```diff
@@ -372,11 +372,10 @@ input[type=range] {
...
 .drag-active-text {
-  position: absolute;
-  width: 240px;
-  height: 240px;
-  padding-top: 110px;
-  text-align: center;
+  height: 100%;
+  display: flex;
+  align-items: center;
+  justify-content: center;
```
rianfowler: Noticed a minor layout issue caused by some bad assumptions on my part. Updated to use flex box.


screen shot 2018-02-19 at 08 29 09

jbalboni (approved):
LGTM

PR #7291, opened 2018-02-16
VRE Chapter 36: military history

jbalboni:
One note with this: there is a way to build this with one page instead of two, but it seemed like the review page worked slightly better this way. Either way, there's some extra work to account for the changes here in the review ticket, which I've noted there.

Veteran applicant:
screen shot 2018-02-16 at 4 11 41 pm

Other applicant:
screen shot 2018-02-16 at 4 12 57 pm

### DIFF
--- a/src/js/vre/chapter36/config/form.js
+++ b/src/js/vre/chapter36/config/form.js
@@ -8,6 +8,8 @@ import {
 import IntroductionPage from '../components/IntroductionPage';
 import ConfirmationPage from '../containers/ConfirmationPage';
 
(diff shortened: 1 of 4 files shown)


cehsu (approved):

stacyhlinn (approved):
I'm guessing your comment "there is a way to build this with one page instead of two" is for development of the version for Veterans vs version for spouses/dependents, and not related to design.

The screenshots look good to me.

jbalboni:
@stacyhlinn Yes, correct. Doesn't affect the design.

PR #7299, opened 2018-02-19
Remove login class from ch 33 benefits link

markgreenburg:
Removes the login required class from the menu link for ch. 33 benefits. This makes it so that a user doesn't have to login before seeing the maintenance message if they click on the ch 33 benefits link from the main nav.


cvalarida (approved):
:+1: Should we write a ticket to add this back once we take GIBS back down off maintenance mode? Not sure how we're capturing all the things that need done to revert that.

markgreenburg:
@djmunoz1 are we tracking the reversion of these changes for when GIBS is back up? I think this PR would be a good candidate to track as something we have to un-do once GIBS works.

djmunoz1:
good point @markgreenburg  -- just created #8844 for it

PR #7301, opened 2018-02-19
VRE review and confirmation page

jbalboni:
Pretty limited changes, since we don't have a backend api. I created another ticket for that.

I also left out the what happens after I apply link, since that doesn't exist yet, as far as I know.

screen shot 2018-02-19 at 3 32 42 pm


rianfowler (approved):

stacyhlinn (approved):
LGTM. I've noted the "what happens after I apply" as something we should get back to.

PR #7304, opened 2018-02-20
Remove prop types, fix enzyme warnings

jbalboni:
We've talked about this, but hadn't done it. This removes prop types when running in the test env.

Still a bunch of warnings, but it's better.


rianfowler (approved):

ncksllvn (approved):

cvalarida (approved):
Clever. :+1:

PR #7306, opened 2018-02-20
Add issues to appeal card

cvalarida:
image

Turns are there are no unit tests for this component, so rather than expanding the scope of this, I'll write another ticket to cover writing unit tests for the whole card. @markgreenburg and I will have to write up other tickets to cover the rest of the stuff we blazed through to get the functionality out the door in time for user testing sessions yesterday.

### DIFF
--- a/src/js/claims-status/components/AppealListItemV2.jsx
+++ b/src/js/claims-status/components/AppealListItemV2.jsx
@@ -50,6 +50,7 @@ export default function AppealListItem({ appeal }) {
         <div className={`status-circle ${appeal.attributes.active ? 'open' : 'closed'}`}/>
         <p><strong>Status:</strong> {getStatusContents(status.type, status.details).title}</p>
       </div>
+      {appeal.attributes.description && <span><strong>Issues on appeal:</strong> {appeal.attributes.description}</span>}
       <div className="communications">
         {renderNextAction(lastEvent, previousHistory)}
       </div>
--- a/src/js/claims-status/components/AppealListItemV2.jsx
+++ b/src/js/claims-status/components/AppealListItemV2.jsx
@@ -50,7 +50,8 @@ export default function AppealListItem({ appeal }) {
         <div className={`status-circle ${appeal.attributes.active ? 'open' : 'closed'}`}/>
         <p><strong>Status:</strong> {getStatusContents(status.type, status.details).title}</p>
       </div>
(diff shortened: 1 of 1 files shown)


src/js/claims-status/components/AppealListItemV2.jsx
```diff
@@ -50,6 +50,7 @@ export default function AppealListItem({ appeal }) {
         <div className={`status-circle ${appeal.attributes.active ? 'open' : 'closed'}`}/>
         <p><strong>Status:</strong> {getStatusContents(status.type, status.details).title}</p>
       </div>
+      {appeal.attributes.description && <span><strong>Issues on appeal:</strong> {appeal.attributes.description}</span>}
```
markgreenburg: Maybe not strictly necessary, but should this vary based on whether there is one issue or many issues? This could come from `appeal.attributes.issues.length`

markgreenburg (approved):
LGTM, but might want to investigate plural vs. singular per the one comment I dropped. 🤷‍♂️

src/js/claims-status/components/AppealListItemV2.jsx
cvalarida: Vary in what way? My understanding is the caseflow api (is that the one, @cmgiven?) will take care of constructing the description based on some logic, but it'll take into consideration of the number of issues on appeal.

markgreenburg: I.e., do we want the text to change from 'Issue on appeal' to 'Issues on appeal' based on how many issues there actually are?

cvalarida: Ahhhh, clever. Yeah, I think that makes sense. :+1:

Pinging @gnakm for visibility.

cmgiven:
@gnakm: Do you have views on whether we should change "issues on appeal" to "issue on appeal" for single-issue appeals?

It would be possible to do that based on the number of items in the `issues` array.

cmgiven:
(I.e. even closed issues are still included in the count.)

gnakm:
Yes we should make content specific to # of appeals. 

If there is only 1 issue, content should be "Issue on appeal"

If there are 2 or more issues, content should be "Issues on appeal"

gnakm:
And yes to closed appeals too.

PR #7307, opened 2018-02-20
[WIP] Feedback form changes: checkbox to radio button

phiden:


melwoodard (changes requested):
Looking good! One thing I see - error behavior for the radio buttons isn't quite right. They go into an error state when I start typing into the text field, and then if I do select one of the radio buttons, the error state doesn't go away.

feedback-form

phiden:
@melwoodard yup, I see what you mean. Will take a look!

ncksllvn:
@phiden there are E2E tests as well for the feedback form, but they've been temporarily disabled. The file is in `test/feedback/feedback.e2e.spec.js`. Just remove the last line, `module.exports['@disabled'] = true;`. Then run - 

```
npm run test:e2e -- test/feedback/feedback.e2e.spec.js
```

src/js/feedback/components/FeedbackForm.jsx
```diff
@@ -41,9 +42,18 @@ class FeedbackForm extends React.Component {
...
+  setResponse = ({ value: shouldSendResponse, dirty }) => {
+    //console.log('value: ', value, 'dirty:', dirty)
+    this.setState({ suppressResponseErrors: true })
+    this.props.setFormValues({ shouldSendResponse });
+  }
+
+  onSubmit = (event) => {
     event.preventDefault();
-    if (this.props.formIsSubmittable) this.props.sendFeedback(this.props.formValues);
+    if (this.props.formValues.shouldSendResponse) return this.props.sendFeedback();
```
cehsu: We should provide this.props.formValues here:
```
if (this.props.formValues.shouldSendResponse) return this.props.sendFeedback(this.props.formValues);
```

src/js/feedback/components/FeedbackForm.jsx
```diff
@@ -41,9 +42,18 @@ class FeedbackForm extends React.Component {
...
+    //Set the redux error message
```
cehsu: It looks like we can reuse this.setResponse here:
```
-    this.props.setFormValues({shouldSendResponse: null});
-    this.setState({ suppressResponseErrors: false });
+   this.setResponse({{value: null});
```
Then update setResponse:
```
+  setResponse = ({ value: shouldSendResponse}) => {
+    this.setState({ suppressResponseErrors: false })
+    this.props.setFormValues({ shouldSendResponse });
+  }
```

phiden: Doing this introduced a fun new error, but it works great otherwise, so I'm going to leave it as-is for the moment.

phiden:
@melwoodard https://vetsgov-pr-7307.herokuapp.com/ <-- behavior updated!

melwoodard (approved):
Awesome, looks good to me!

src/js/feedback/components/FeedbackForm.jsx
```diff
@@ -54,30 +64,43 @@ class FeedbackForm extends React.Component {
...
+              name="shouldSendResponse"
+              id="shouldSendResponse"
+              label="Would you like us to follow up with you about your ideas?"
+              options={[
+                { label: 'Yes', value: 'yes' },
+                { label: 'No', value: 'no' }
+              ]}
+              onValueChange={this.setResponse}
+              errorMessage={this.responseErrorMessage()}
+              value={{ value: shouldSendResponse, dirty: false }}
```
cehsu: Is dirty being used here?

src/js/feedback/components/FeedbackForm.jsx
```diff
@@ -54,30 +64,43 @@ class FeedbackForm extends React.Component {
...
+              required/>
```
cehsu: Should we include `ref` here since it is used on the other fields as well? Are these refs already being or planned to be used? @ncksllvn 
```
ref={component => { this.shouldSendResponseComp = component; }}
```

src/js/feedback/components/FeedbackForm.jsx
```diff
@@ -54,30 +64,43 @@ class FeedbackForm extends React.Component {
...
+
             <div className="usa-grid-full">
               <div className="usa-width-two-thirds">
-                {this.props.formValues.shouldSendResponse && (
+                {this.props.formValues.shouldSendResponse === 'yes' && (
```
cehsu: Since we have already destructured formValues on line 72, we can simply use `shouldSendResponse` here.

cehsu (commented):
Looks good so far. I had a few questions and minor comments. Please let me know if additional clarification regarding these would be helpful.

jkassemi:
Closing due to inactivity. @cehsu's checking for associated issues.

PR #7310, opened 2018-02-20
Set up additional info pages

cehsu:
These changes set up the additional information pages.

There is a discrepancy between how `previousBenefitsApplications` is defined here and in the schema, feedback on how to resolve this and any implications on the configuration here is welcome.

A new validation has also been set up for the checkboxes. Feedback on that is welcome as well (including whether we want to remove it from this PR, and revisit the idea later).

There was a question as to whether the divorce question affects eligibility posted on the original issue, and whether any warning might be warranted as well.

screen shot 2018-02-20 at 1 46 59 pm
screen shot 2018-02-20 at 2 06 38 pm
screen shot 2018-02-20 at 1 48 45 pm
screen shot 2018-02-20 at 1 43 26 pm
screen shot 2018-02-20 at 1 43 41 pm


src/js/vre/chapter36/config/form.js
```diff
@@ -247,6 +288,227 @@ const formConfig = {
...
+          depends: {
+            'view:isVeteran': true
+          },
+          uiSchema: {
+            previousBenefitApplications: previousBenefitApplicationsUI
+          },
+          schema: {
+            type: 'object',
+            properties: {
+              previousBenefitApplications: {
```
cehsu: This is defined differently in the schema. Should the configuration here (consistent with how we build checkbox fields elsewhere) be changed or the schema updated? The schema definition is:
```
    "previousBenefitApplications": {
      "type": "array",
      "items": {
        "type": "string"
      }
    },
```

src/js/vre/chapter36/config/form.js
```diff
@@ -52,6 +52,14 @@ const requiredDateRange = _.merge(dateRange, {
   required: ['to', 'from']
 });
 
+export function validateBooleanGroup(errors, userGroup, form, schema, errorMessages = {}) {
```
cehsu: This is a new validation, should this be included in these changes or do we not want to restrict the options in this way, or should we table this for now and revisit it later?

src/js/vre/chapter36/config/form.js
```diff
@@ -258,6 +289,226 @@ const formConfig = {
...
+          depends: {
+            'view:isVeteran': true
+          },
+          uiSchema: {
+            previousBenefitApplications: previousBenefitApplicationsUI
+          },
+          schema: {
+            type: 'object',
+            properties: {
+              previousBenefitApplications: {
```
rianfowler: Maybe an array of [enum](https://stackoverflow.com/questions/30924271/correct-way-to-define-array-of-enums-in-json-schema)?

src/js/vre/chapter36/config/form.js
rianfowler: Maybe an array of [enum](https://stackoverflow.com/questions/30924271/correct-way-to-define-array-of-enums-in-json-schema)?

src/js/vre/chapter36/config/form.js
rianfowler: Could be a two stage question. Have you applied for benefits and then disclose the additional options if they select yes.

src/js/vre/chapter36/config/form.js
rianfowler: I guess technically it would be oneOf either an enum of none or an enum of the other three options.

src/js/vre/chapter36/config/form.js
jbalboni: I think this validation makes sense. We should run it by @stacyhlinn, but I think we can keep it unless we change the question format, like Rian suggests.

src/js/vre/chapter36/config/form.js
jbalboni: I think we should update the schema. What you came up with is more consistent with how we've treated similar fields on other forms.

src/js/vre/chapter36/config/form.js
jbalboni: However, the name clashes with a similar validation function we have elsewhere, so maybe `validateNoneOrSomeChosen` or something like that?

src/js/vre/chapter36/config/form.js
```diff
@@ -258,6 +289,226 @@ const formConfig = {
...
+                none: {
+                  'ui:title': benefitsLabels.none
+                }
+              }
+            ),
+            previousVeteranBenefitsFullName: _.merge(fullNameUI, {
+              'ui:description': "Veteran's name under whom you've claimed benefits",
+
+              first: {
+                'ui:required': () => false
```
jbalboni: I don't think you need these if you're unset the required property in the regular schema.

src/js/vre/chapter36/config/form.js
```diff
@@ -258,6 +289,226 @@ const formConfig = {
...
+                other: {
+                  'ui:title': benefitsLabels.other
+                },
+                none: {
+                  'ui:title': benefitsLabels.none
+                }
+              }
+            ),
+            previousVeteranBenefitsFullName: _.merge(fullNameUI, {
+              'ui:description': "Veteran's name under whom you've claimed benefits",
```
jbalboni: Need curly apostrophes, I think.

src/sass/vre.scss
```diff
@@ -6,3 +6,11 @@
 @import "modules/m-modal";
 @import "modules/m-omb-info";
 @import "modules/m-form-confirmation";
+
+.schemaform-first-field .schemaform-label, .schemaform-field-template > .usa-input-error {
+  margin-top: 0;
+}
+
+.schemaform-field-container {
```
jbalboni: These changes are causing some spacing issues on other pages (like military history)

jbalboni:
@cehsu There's a lot of duplication in the config for these pages. Which causes a problem in some cases, like if you switch between applying as a Veteran or the relationship types; the previous benefit options that only appear for a particular page are still set in the form data, but can't be unset because they're not visible.

It seems like we could make additional information one page, and have the benefits option booleans use `hideIf` to conditionally show/hide them depending on veteran/relationship status.

src/js/vre/utils/labels.jsx
```diff
@@ -0,0 +1,18 @@
...
+  general: 'General',
+  other: 'Other Than Honorable',
+  'bad-conduct': 'Bad Conduct',
+  dishonorable: 'Dishonorable',
+  undesirable: 'Undesirable'
+};
+
+export const benefitsLabels = {
+  chapter31: <p>Vocational Rehabilitation Benefits (Chapter 31)</p>,
+  ownServiceBenefits: <p>Veterans' Education Assistance based on your own service</p>,
```
jbalboni: Curly apostrophe on this line and the next, I think.

stacyhlinn:
Here's a potential solution:
What is we take out the "none" checkbox? If nothing is selected, then we know it's none.

cehsu:
Thanks @stacyhlinn, @jbalboni @rianfowler does the idea of removing "none" work?

jbalboni:
@cehsu Works for me.

rianfowler:
What is an "Other" VA benefit? I think removing none is fine but the question may need to be reworked.

stacyhlinn:
@rianfowler that is a great question. I asked the stakeholder the question, and her response was "any VA benefits". They use this question to try to assess the applicant's situation and make sure they can give a holistic recommendation during couseling. We still need some content input from Peggy, but here's the latest design on this:

artboard

Question for all the FEs: Should I file the above change as a separate ticket?

src/js/vre/utils/labels.jsx
cehsu: Thanks, fixed.

src/sass/vre.scss
cehsu: Thanks, this has been removed, and the name field has been assigned the schemform-field-template class instead.

src/js/vre/chapter36/config/form.js
cehsu: Thanks, removed.

src/js/vre/chapter36/config/form.js
```diff
@@ -85,6 +86,61 @@ const applicantServiceHistory = _.merge(serviceHistory, {
...
 
+const previousBenefitApplicationsUI = {
+  'ui:title': 'Have you ever applied for any of the following VA benefits? (Check all that apply.)',
+  'ui:options': {
+    showFieldLabel: true
+  },
+  chapter31: {
+    'ui:title': benefitsLabels.chapter31,
+    'ui:options': {
+      hideIf: ({ 'view:isVeteran': isVeteran }) => !isVeteran
```
jbalboni: Could we make two functions, `isVeteran` and `isNotVeteran`, and use those everywhere? Feel like it would be clearer to be consistent in how we use those.

src/js/vre/chapter36/config/form.js
```diff
@@ -85,6 +86,61 @@ const applicantServiceHistory = _.merge(serviceHistory, {
...
+  },
+  otherExplanation: {
+    'ui:title': benefitsLabels.otherExplanation,
+    'ui:options': {
+      expandUnder: 'other'
+    }
+  }
+};
+
+const previousBenefitApplications = {
```
jbalboni: Is the plan to update the schema with this and then pull it from there?

src/sass/vre.scss
```diff
@@ -6,3 +6,7 @@
 @import "modules/m-modal";
 @import "modules/m-omb-info";
 @import "modules/m-form-confirmation";
+
+.schemaform-first-field .schemaform-label, .schemaform-field-template > .usa-input-error {
```
jbalboni: Do we need this? It looks like a duplicate of some styles in `_m-schemaform.scss`

jbalboni (approved):
Couple comments, but LGTM.

src/sass/vre.scss
cehsu: This is just an override, I'll add a comment clarifying this.

src/js/vre/chapter36/config/form.js
cehsu: Yes, thanks for the reminder. Will update.

src/js/vre/chapter36/config/form.js
cehsu: Thanks @jbalboni, good idea, changed.

src/sass/vre.scss
cehsu: This doesn't seem to be an issue anymore, it has been removed.

rianfowler (approved):

PR #7313, opened 2018-02-21
8823 vre ch31 ch3

rianfowler:
[Design](https://app.zenhub.com/workspace/o/department-of-veterans-affairs/vets.gov-team/issues/8823)

Todo:
- [x] jobDuties

- ~@stacyhlinn @jbalboni   The schema doesn't track if the applicant is working so they could technically select "Yes", fill out no employer information, and we would submit nothing related to work information.~
- ~@jbalboni Job Duties is listed as an array of strings in the schema. Should we update the schema to just capture a string or do we want to have something else in the UI?~

### DIFF
--- a/src/js/vre/chapter31/config/form.js
+++ b/src/js/vre/chapter31/config/form.js
@@ -96,6 +96,7 @@ const formConfig = {
             },
             employer: {
               'ui:title': 'Employer name',
+              'ui:required': (formData) => formData['view:isWorking'],
               ...expandIfWorking
             },
             jobDuties: {
--- a/src/js/vre/chapter31/config/form.js
+++ b/src/js/vre/chapter31/config/form.js
@@ -1,15 +1,29 @@
-// import _ from 'lodash/fp';
+import _ from 'lodash/fp';
 
 import fullSchema31 from 'vets-json-schema/dist/28-1900-schema.json';
+
+import * as address from '../../../common/schemaform/definitions/address';
+import currencyUI from '../../../common/schemaform/definitions/currency';
 
 import IntroductionPage from '../components/IntroductionPage';
 import ConfirmationPage from '../containers/ConfirmationPage';
 
-// const {} = fullSchema31.properties;
+const {
+  employer,
+  employerAddress,
+  jobDuties,
+  monthlyIncome
+} = fullSchema31.properties;
 
 const {
   fullName
 } = fullSchema31.definitions;
+
+const hideIfNotWorking = {
+  'ui:options': {
+    hideIf: (formData) => formData['view:isWorking'] === undefined || formData['view:isWorking'] === false
+  }
+};
 
 const formConfig = {
   urlPrefix: '/',
@@ -27,6 +41,7 @@ const formConfig = {
   title: '',
   subTitle: 'Form 28-1900',
   defaultDefinitions: {
+    address,
     fullName
   },
   chapters: {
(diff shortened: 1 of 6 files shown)


src/js/vre/chapter31/config/form.js
```diff
@@ -61,12 +76,39 @@ const formConfig = {
...
+          uiSchema: {
+            'view:isWorking': {
+              'ui:title': 'Are you working?',
+              'ui:widget': 'yesNo'
+            },
+            employer: _.merge(
+              {
+                'ui:title': 'Employer name'
+              },
+              hideIfNotWorking,
```
rianfowler: Not sure if this is an idiomatic approach.

test/vre/chapter31/config/workInformation.unit.spec.jsx
```diff
@@ -0,0 +1,72 @@
+import React from 'react';
+import { expect } from 'chai';
+import sinon from 'sinon';
+import { mount } from 'enzyme';
+
+import { DefinitionTester, selectRadio } from '../../../util/schemaform-utils.jsx';
+import formConfig from '../../../../src/js/vre/chapter31/config/form.js';
+
+describe.only('VRE chapter 31 work information', () => {
```
jbalboni: Remember to remove this

jbalboni:
@rianfowler We should probably require employer name at least when you choose Yes. Maybe others, depending on what @stacyhlinn says.

And yeah, let's change the job duties schema to a string.

src/js/vre/chapter31/config/form.js
jbalboni: What we probably want to do is use the `expandUnder` functionality, so that `employer` expands under `view:isWorking`.

stacyhlinn:
@rianfowler Yes let's make the employer's name required. Sorry for the confusion - we still haven't heard back from the stakeholder for all the required fields, but it makes sense to at least make the employer name required.

rianfowler:
This is ready for final review @stacyhlinn @jbalboni @cehsu 
2018-02-21-15-13-localhost_3001

src/js/vre/chapter31/config/form.js
```diff
@@ -76,9 +89,40 @@ const formConfig = {
...
+            employer: {
+              'ui:title': 'Employer name',
+              'ui:required': (formData) => formData['view:isWorking'],
+              ...expandIfWorking
+            },
+            jobDuties: {
+              'ui:title': 'Job duties',
+              ...expandIfWorking
+            },
+            monthlyIncome: _.merge(
```
rianfowler: Dunno why I used spread and `merge` here. Can update if there is a preference.

stacyhlinn (approved):
Looks good!

cehsu (approved):

test/user-profile/helpers.unit.spec.js
```diff
@@ -76,7 +76,7 @@ describe('profile helpers:', () => {
       const allMappedIds = Object.keys(schemaToConfigIds);
       const sipEnabledConfigs = configs.filter(config => !config.disableSave);
       const sipEnabledFormIds = sipEnabledConfigs.map(sipEnabledConfig => sipEnabledConfig.formId);
-      expect(allFormIds).to.deep.equal(allMappedIds);
+      expect(allFormIds).to.include.members(allMappedIds);
```
rianfowler: @jbalboni LMK what you think about this. I think there was a placeholder included in the `vets-json-schema` that's not implemented yet in `vets-website` and since I added a change after this commit things are out of sync. I think this check now doesn't fail if there's something in the `vets-json-schema` that isn't implemented in `vets-website`.

src/js/vre/chapter31/config/form.js
jbalboni: I think being consistent is good, but either approach seems fine to me.

test/user-profile/helpers.unit.spec.js
jbalboni: I think this is ok, but I would look and see if adding that net form to the ignored forms Set in the file this is testing helps. @cehsu may remember more about these tests than me.

rianfowler: Good idea- I'll update the unit tests

test/migrations.unit.spec.js
```diff
@@ -43,6 +43,7 @@ const configs = [
 
 const excludedForms = new Set([
   '28-1900',
+  '21-526EZ',
```
rianfowler: @cehsu LMK if this is the proper usage of this list. This form was added to the `vets-json-schema` recently but hasn't been added to the `vets-website` yet which is causing a couple of unit tests to fail.

cehsu: This is correct @rianfowler. I'll add documentation to that test, sorry about the lack of clarity.

cehsu (approved):

PR #7315, opened 2018-02-21
526 skeleton cv

cvalarida:
image

The intro page really needs love and attention, but the point is it loads the app at that URL.


config/webpack.config.js
```diff
@@ -21,6 +21,7 @@ const entryFiles = {
   '1995-edu-benefits': './src/js/edu-benefits/1995/edu-benefits-entry.jsx',
   '5490-edu-benefits': './src/js/edu-benefits/5490/edu-benefits-entry.jsx',
   '5495-edu-benefits': './src/js/edu-benefits/5495/edu-benefits-entry.jsx',
+  '526-claims-increase': './src/js/526/form-entry.jsx',
```
jbalboni: I think there a bunch of forms under the 526 umbrella, right? I wonder if we should put this form in a sub folder with the expectation that there will be more forms that live alongside it.

cvalarida: Yeah, to be honest, I don't know how this all works out. This is the 21-526EZ, which to my understanding isn't for a claims increase (that's a 526b, I think), but we're using it for a claims increase anyhow? I don't know what I'm doing. :confused:

You're right, though. I think I'll make a `src/js/disability-benefits/526EZ/` and move this all in there. Good call. :+1:

config/webpack.config.js
```diff
@@ -21,6 +21,7 @@ const entryFiles = {
   '1995-edu-benefits': './src/js/edu-benefits/1995/edu-benefits-entry.jsx',
   '5490-edu-benefits': './src/js/edu-benefits/5490/edu-benefits-entry.jsx',
   '5495-edu-benefits': './src/js/edu-benefits/5495/edu-benefits-entry.jsx',
+  '526EZ-claims-increase': './src/js/526EZ/form-entry.jsx',
```
jbalboni: I think this path is missing `disability-benefits`

src/js/disability-benefits/526EZ/components/IntroductionPage.jsx
```diff
@@ -0,0 +1,91 @@
...
+  render() {
+    return (
+      <div className="schemaform-intro">
+        <FormTitle title="This page needs attention"/>
+        <p>Equal to VA Form 21-526EZ.</p>
+        <SaveInProgressIntro
+          prefillEnabled={this.props.route.formConfig.prefillEnabled}
+          messages={this.props.route.formConfig.savedFormMessages}
+          pageList={this.props.route.pageList}
+          startText="Start the Education Application"
```
jbalboni: I assume you're leaving these edu reference in here to be addressed later, but pointing it out just in case.

src/js/disability-benefits/526EZ/form-entry.jsx
```diff
@@ -0,0 +1,33 @@
+import 'core-js';
+import '../../../sass/edu-benefits.scss';
```
jbalboni: Should change this.

jbalboni (approved):
Couple things, but LGTM.

src/js/disability-benefits/526EZ/components/IntroductionPage.jsx
cvalarida: Yup. Exactly that.

config/webpack.config.js
cvalarida: Woops....good catch!

src/js/disability-benefits/526EZ/form-entry.jsx
cvalarida: Mmm...yeah, probably... :grimacing:

PR #7321, opened 2018-02-22
Add gender to veteran applicant info page

cehsu:
These changes add gender to the veteran applicant information page, as indicated by @stacyhlinn.
screen shot 2018-02-21 at 10 36 14 pm


src/js/vre/chapter36/config/form.js
```diff
@@ -119,7 +119,7 @@ const formConfig = {
         applicantInformation: {
           title: 'Applicant information',
           path: 'applicant-information',
-          applicantRelationshipToVeteran: 'Spouse',
+          applicantRelationshipToVeteran: 'Spouse', // is this a typo?
```
jbalboni: Yeah, I think this can be removed.

src/js/vre/chapter36/config/form.js
```diff
@@ -221,10 +221,20 @@ const formConfig = {
...
               }
-            )
+            ),
+            applicantGender: {
+              'ui:title': 'Gender',
+              'ui:widget': 'radio',
+              'ui:required': formData => formData['view:isVeteran'] === true,
+              'ui:options': {
+                labels: genderLabels,
+                hideIf: ({ 'view:isVeteran': isVeteran }) => !isVeteran
```
jbalboni: Could we make this match the other hideIf function above (but negated)? Seems like it'll be easier to read if they're written the same way.

jbalboni (approved):
I think there are tests to change, but LGTM.

rianfowler (approved):

stacyhlinn:
Wanted to confirm that the gender question will only show up when the applicant is the Veteran. Looks good!

stacyhlinn (approved):

src/js/vre/chapter36/config/form.js
cehsu: Thanks, removed.

src/js/vre/chapter36/config/form.js
cehsu: Sounds good, changed.

PR #7328, opened 2018-02-22
Request photo before submit on VIC form

jbalboni:
The photo might be deleted by the time the confirmation page is displayed, so grab the image if we don't have it already before submitting, and display the blob url.

Can't really test this locally.


rianfowler (approved):

rlbaker (approved):
LGTM

cehsu (approved):

PR #7329, opened 2018-02-22
Claims list item v2 tests

cvalarida:
I moved `ClaimsListItemV2` up into `components/appeals-v2/` because it's only used in the v2 index page and it made more sense in there.

The tests are basically just copied from the v2 component tests, then converted to use `enzyme`. The last two tests are new since that's the only major functionality change in the v2 list item.


kreek (approved):
LGTM 🚀

PR #7331, opened 2018-02-22
Set up disability information page for VRE chapter 31

cehsu:
These changes set up the disability information page for VRE chapter 31, and relocate the disability record facility field from the veteran information page.
localhost_3001_employment_vocational-rehab-and-employment_application_chapter31_disability-information 2

localhost_3001_employment_vocational-rehab-and-employment_application_chapter31_disability-information 1


src/js/vre/chapter31/config/form.js
```diff
@@ -155,9 +146,52 @@ const formConfig = {
...
+                type: 'string'
+              },
+              vaRecordsOffice,
+              'view:inHospital': {
+                type: 'boolean'
+              },
+              hospital: {
+                type: 'object',
+                properties: {
+                  hospitalName: {
```
jbalboni: Do we need to add this to the schema?

src/js/vre/chapter31/config/form.js
```diff
@@ -155,9 +146,52 @@ const formConfig = {
...
             properties: {
+              disabilityRating,
+              disabilities: {
+                type: 'string'
+              },
+              vaRecordsOffice,
+              'view:inHospital': {
+                type: 'boolean'
+              },
+              hospital: {
```
jbalboni: Looks like this should be `view:hospital`.

test/vre/chapter31/config/disabilityInformation.unit.spec.jsx
```diff
@@ -0,0 +1,54 @@
...
+        schema={schema}
+        data={{}}
+        uiSchema={uiSchema}/>
+    );
+
+    form.find('form').simulate('submit');
+
+    expect(form.find('.usa-input-error').length).to.equal(0);
+    expect(onSubmit.called).to.be.true;
+  });
```
jbalboni: Could we add a test that tries to submit with the hospital fields expanded, so we can check that the number of required fields is correct?

jbalboni (approved):
LGTM, had a couple minor things.

stacyhlinn:
@cehsu 

- For the disability ratings, can we make that a dropdown, with options being multiples of 10: 0%, 10%, 20%.....100%?

- My bad for not marking these as required. Can we make all the fields required, including hospital name?

src/js/vre/chapter31/config/form.js
```diff
@@ -63,17 +64,7 @@ const formConfig = {
...
-            vaRecordsOffice: {
-              'ui:title': 'VA benefit office where your records are located',
-              'ui:help': facilityLocatorLink
-            }
-          },
-          schema: {
-            vaRecordsOffice
-          }
-        })
+        veteranInformation: createVeteranInfoPage(fullSchema31)
```
rianfowler: What's the threshold for creating something like this? e.g. if the identical page is used on more than one form?

rianfowler (approved):

src/js/vre/chapter31/config/form.js
jbalboni: We don't have a hard rule about it, but generally if the page has more than a couple fields, has some specific validation/display logic, and is used on more than one form it's a good candidate for being pulled out like this.

We've been more aggressive about it in the past, but it can lead to complicated page definitions that try to account for variation between forms.

PR #7333, opened 2018-02-23
Break out biggest pieces of options-for-select

jbalboni:
The options-for-select.js file is really large because it contains the list of VA medical centers and state/country labels. It's also pulled into every bundle. This change moves things around so that

- Only HCA pulls in the VAMC list
- Apps that need address labels and validation have to pull that in specifically, and not accidentally import it because they pulled in an unrelated validation function.

This reduces non-HCA bundles by about 16kb (gzipped). That's around a 10% reduction for VIC.

This change looks larger than it is, I'm only moving code around and changing imports.


src/js/vic-v2/config/form.js
```diff
@@ -4,6 +4,7 @@ import fullSchemaVIC from 'vets-json-schema/dist/VIC-schema.json';
 
 import IntroductionPage from '../components/IntroductionPage';
 import ConfirmationPage from '../containers/ConfirmationPage';
+// import PhotoField from '../components/PhotoField';
```
cehsu: Should this be removed? Or is this to promote visibility of what is going on elsewhere with the asyncLoader?

src/js/vic-v2/definitions/address.js
```diff
@@ -2,10 +2,34 @@ import _ from 'lodash/fp';
 import { createSelector } from 'reselect';
 
 import { countryLabels, countryValues, states } from '../labels';
-import { validatePostalCodes } from '../../common/schemaform/validation';
 
 const requiredFields = ['street', 'city', 'country', 'state', 'postalCode'];
 
+export function isValidUSZipCode(value) {
```
cehsu: It looks like this is a duplicate of the one in `src/js/common/utils/address.js`, is this another optimization?

cehsu (approved):
LGTM. Great work!

src/js/vic-v2/config/form.js
jbalboni: This is I think a bad merge. None of the changes in this file should be in here.

src/js/vic-v2/definitions/address.js
jbalboni: Yeah, I don't want to import that file because it has labels that this address field doesn't need. It felt more awkward to make another file just for these zip code functions.

cvalarida (approved):
Quick pass looks good to me. :+1:

PR #7334, opened 2018-02-23
Move sanitize so we don't import the world into every bundle

jbalboni:
The user profile code imports a `santizeForm` function from a helper file that imports all the schemaform code, which meant that the schemaform code was getting into every bundle.

Moving this function knocks 40kb (gzipped) off of non-form app bundles, which is around a 30% reduction, depending on the app.


cehsu (approved):
LGTM. Great improvement!

PR #7335, opened 2018-02-23
Don't call onBlur if list is open

jbalboni:
We were calling blur if you clicked on an option, which moved the list down and made your click not register. This uses Downshift's onOuterClick to handle blur if the list is open, and uses the input blur event otherwise.


cehsu (approved):
LGTM. This is great, and can probably also be used in other places where we implement similar logic, such as in the facility locator dropdowns. (I will note this in the relevant ticket.)

PR #7336, opened 2018-02-23
Fix VIC image type mismatch and update messages

jbalboni:
A few changes:

1. Overwrite the image file name when cropping. The output is always image/png and the file name doesn't really matter, so setting it to a static name.
2. Remove bmp from allowed types, but allow them into the cropper tool. The cropper tool will convert it to png, so no reason to restrict it there.
3. Add gif to allowed types. This was allowed on the backend, but not the front.
4. Update messages.

### DIFF
--- a/src/js/vic-v2/components/DD214Description.jsx
+++ b/src/js/vic-v2/components/DD214Description.jsx
@@ -7,7 +7,7 @@ export default function DD214Description() {
       <p>If you have more than one discharge document, please upload the one with the highest character of discharge.</p>
       <p>If you don’t have a digital copy of your discharge document, you can scan or take a photo of it and then upload it to your computer or phone. The digital file could be a .pdf file or other photo file format, like a .jpeg or .png.</p>
 
-      <div>File types you can upload: .pdf, .jpeg, or .png</div>
+      <div>File types you can upload: .pdf, .jpeg, .gif, .tif, or .png</div>
       <div>Maximum file size: 25 MB</div>
     </div>
   );
--- a/src/js/vic-v2/components/DD214Description.jsx
+++ b/src/js/vic-v2/components/DD214Description.jsx
@@ -7,7 +7,7 @@ export default function DD214Description() {
       <p>If you have more than one discharge document, please upload the one with the highest character of discharge.</p>
       <p>If you don’t have a digital copy of your discharge document, you can scan or take a photo of it and then upload it to your computer or phone. The digital file could be a .pdf file or other photo file format, like a .jpeg or .png.</p>
 
-      <div>File types you can upload: .pdf, .jpeg, .gif, .tif, or .png</div>
+      <div>File types you can upload: .pdf, .jpeg, .gif, .tiff, or .png</div>
       <div>Maximum file size: 25 MB</div>
     </div>
   );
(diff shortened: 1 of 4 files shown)


cehsu (approved):
LGTM.

src/js/vic-v2/components/DD214Description.jsx
```diff
@@ -7,7 +7,7 @@ export default function DD214Description() {
       <p>If you have more than one discharge document, please upload the one with the highest character of discharge.</p>
       <p>If you don’t have a digital copy of your discharge document, you can scan or take a photo of it and then upload it to your computer or phone. The digital file could be a .pdf file or other photo file format, like a .jpeg or .png.</p>
 
-      <div>File types you can upload: .pdf, .jpeg, or .png</div>
+      <div>File types you can upload: .pdf, .jpeg, .gif, .tif, or .png</div>
```
cehsu: Should we use .tiff here since we use .jpeg or the three letter extensions for each?

src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -306,7 +306,7 @@ export default class PhotoField extends React.Component {
     const fileTypes = this.props.uiSchema['ui:options'].fileTypes;
     if (!isValidFileType(file.name, fileTypes)) {
       this.props.onChange({
-        errorMessage: 'We weren’t able to upload your file. Please make sure the file you’re uploading is a jpeg, .png, .tiff,  or .bmp file and try again.'
+        errorMessage: 'We weren’t able to upload your file. Please make sure the file you’re uploading is a gif, jpeg, .png, or .tiff file and try again.'
```
cehsu: Are we missing a . in front of gif?

src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -354,7 +354,7 @@ export default class PhotoField extends React.Component {
       }
       if (!isValidFileType(fileName, fileTypes)) {
         this.props.onChange({
-          errorMessage: 'We weren’t able to upload your file. Please make sure the file you’re uploading is a jpeg, .png, .tiff,  or .bmp file and try again.'
+          errorMessage: 'We weren’t able to upload your file. Please make sure the file you’re uploading is a gif, jpeg, .png, .tiff, or .bmp file and try again.'
```
cehsu: Same as above.

test/vic-v2/components/PhotoField.unit.spec.jsx
```diff
@@ -182,7 +182,7 @@ describe('<PhotoField>', () => {
       }]);
 
       setTimeout(() => {
-        expect(onChange.firstCall.args[0].errorMessage).to.contain('make sure the file you’re uploading is a jpeg');
+        expect(onChange.firstCall.args[0].errorMessage).to.contain('make sure the file you’re uploading is a gif, jpeg');
```
cehsu: This test may need to be updated to include . as well.

src/js/vic-v2/components/PhotoField.jsx
jbalboni: I think the lack of one in front of jpeg is because saying "jpeg" implies both extensions. I'm going to use that for tiff, too, and left the `.` in front of the others.

PR #7341, opened 2018-02-26
Remove VIC form and e2e test from ignore list

jbalboni:
This will let the VIC v2 form go live behind a password. It does not include the content pages because we don't need those to do the initial prod testing.


cehsu (approved):
LGTM.

PR #7344, opened 2018-02-26
Scroll on sign in if has saved form

cehsu:
These changes scroll up if the user signs in and has a form saved for the current app.

### DIFF
--- a/src/js/vic-v2/components/IntroductionPage.jsx
+++ b/src/js/vic-v2/components/IntroductionPage.jsx
@@ -1,17 +1,21 @@
 import React from 'react';
 import { bindActionCreators } from 'redux';
 import { connect } from 'react-redux';
+import Scroll from 'react-scroll';
 
 import { focusElement } from '../../common/utils/helpers';
 import FormTitle from '../../common/schemaform/components/FormTitle';
 import SaveInProgressIntro, { introActions, introSelector } from '../../common/schemaform/save-in-progress/SaveInProgressIntro';
-import { hasSavedForm, scrollToElement, handleScrollEnd } from '../helpers';
+import { hasSavedForm } from '../helpers';
+
+const { animateScroll: scroll, Events } = Scroll;
 
 class IntroductionPage extends React.Component {
   componentDidMount() {
-    this.timer = null;
     focusElement('.va-nav-breadcrumbs-list');
-    window.addEventListener('scroll', this.onScroll);
+    Events.scrollEvent.register('end', () => {
+      focusElement('[id="2-continueButton"]', { preventScroll: true });
+    });
   }
 
   componentDidUpdate(prevProps) {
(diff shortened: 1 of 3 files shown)


src/js/vic-v2/components/IntroductionPage.jsx
```diff
@@ -5,12 +5,21 @@ import { connect } from 'react-redux';
...
 
 class IntroductionPage extends React.Component {
   componentDidMount() {
     focusElement('.va-nav-breadcrumbs-list');
   }
 
+  componentDidUpdate() {
+    const { savedForms } = this.props.saveInProgress.user.profile;
+    const { formId } = this.props.saveInProgress;
+    if (hasSavedForm(savedForms, formId)) {
```
jbalboni: I think we want to make sure we only scroll and focus when their login status changes. Otherwise we could steal focus at a weird time if the component happens to update.

src/js/vic-v2/components/IntroductionPage.jsx
```diff
@@ -5,12 +5,21 @@ import { connect } from 'react-redux';
...
+      focusElement('.va-nav-breadcrumbs-list');
```
jbalboni: I think this just focuses, but we also want to scroll.

src/js/vic-v2/components/IntroductionPage.jsx
cehsu: Good point, changed.

src/js/vic-v2/components/IntroductionPage.jsx
cehsu: Good point, changed.

src/js/vic-v2/helpers.jsx
```diff
@@ -163,3 +163,29 @@ export function submit(form, formConfig) {
   });
 }
 
+export function hasSavedForm(savedForms, formID) {
+  return savedForms.some(({ form }) => form === formID);
+}
+
+export function scrollToElement(selectorOrElement) {
```
jbalboni: We've been using react-scroll for this. If you search for `scrollToTop` in the codebase you can find examples of how it's used.

src/js/vic-v2/components/IntroductionPage.jsx
```diff
@@ -5,10 +5,36 @@ import { connect } from 'react-redux';
...
+      this.scrollingToTop = true;
+      scrollToElement('.va-nav-breadcrumbs-list');
+    }
+  }
+
+  componentWillUnmount() {
+    window.removeEventListener('scroll', this.onScroll);
+  }
+
+  onScroll = () => {
```
jbalboni: It looks like this is to focus after scrolling is finished? I think you can get away with scrolling and then focusing in `componentDidUpdate`. It shouldn't cause any problems to have the focus set before scrolling is finished.

cehsu: Thanks @jbalboni, great idea.

src/js/vic-v2/helpers.jsx
cehsu: Thanks for pointing that out, will update.

src/js/vic-v2/components/IntroductionPage.jsx
```diff
@@ -1,16 +1,34 @@
...
     focusElement('.va-nav-breadcrumbs-list');
   }
 
+  componentDidUpdate(prevProps) {
+    const { savedForms } = this.props.saveInProgress.user.profile;
+    const { formId } = this.props.saveInProgress;
+    const notPreviouslyLoggedIn = !prevProps.saveInProgress.user.login.currentlyLoggedIn;
+    if (notPreviouslyLoggedIn && hasSavedForm(savedForms, formId)) {
+      focusElement('[id="2-continueButton"]', { preventScroll: true });
+      scroll.scrollToTop(window.VetsGov.scroll || {
```
rianfowler: what is `window.VetsGov`?

rianfowler (approved):

src/js/vic-v2/components/IntroductionPage.jsx
jbalboni: It's a hackish way of letting us disable smooth scrolling in e2e tests. We set that global object in e2e tests and leave it empty in other environments. We just started disabling `window.scrollTo` entirely in e2e tests, so it may go away in the future.

jbalboni (approved):

PR #7351, opened 2018-02-28
Removed LOA check and manual redirect for identity proofing in auth flow.

U-DON:
Resolves department-of-veterans-affairs/vets.gov-team#8886.
Depends on department-of-veterans-affairs/vets-api#1708.

@omgitsbillryan: I tagged you since we're going to end up removing the `clientId` query param on a lot of these auth URLs since the redirects will be coming from the backend. I dug around to find out why we needed this and found this old issue (department-of-veterans-affairs/vets.gov-team#3440).

In order to maintain the query param, perhaps we can send up the GA `clientId` as part of the request header, and the API could append it to any ID.me URLs it ends up redirecting to?


saneshark (approved):
LGTM, much cleaner.

PR #7352, opened 2018-02-28
Fix focusElement bug

cehsu:
These changes update focusElement to make the tabIndex setting logic conditional, and apply only to non-focusable elements.

### DIFF
--- a/src/js/common/utils/helpers.js
+++ b/src/js/common/utils/helpers.js
@@ -125,7 +125,6 @@ export function focusElement(selectorOrElement, options) {
     : selectorOrElement;
 
   if (el) {
-    el.setAttribute('tabindex', '-1');
     el.focus(options);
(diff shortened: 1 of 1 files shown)


src/js/common/utils/helpers.js
```diff
@@ -125,7 +125,6 @@ export function focusElement(selectorOrElement, options) {
     : selectorOrElement;
 
   if (el) {
-    el.setAttribute('tabindex', '-1');
```
jbalboni: Removing this breaks the focus management on the forms when you move between pages. Maybe we can check for a tab index of 0 and not set it then?

cehsu: Thanks for the context @jbalboni. Updated.

jbalboni (approved):

PR #7353, opened 2018-02-28
Fix schemaform inputs shrinking at large screen sizes

jbalboni:
Current:
screen shot 2018-02-28 at 9 19 40 am

Fixed:
screen shot 2018-02-28 at 9 19 48 am

### DIFF
--- a/src/sass/modules/_m-schemaform.scss
+++ b/src/sass/modules/_m-schemaform.scss
@@ -14,9 +14,6 @@ fieldset {
   display: inline-block;
   width: 100%;
   max-width: 750px;
-  @media (min-width: $large-screen) {
-    width: auto;
-  }
   .form-review-panel & {
     width: 100%;
   }


src/sass/modules/_m-schemaform.scss
```diff
@@ -14,9 +14,6 @@ fieldset {
   display: inline-block;
   width: 100%;
   max-width: 750px;
-  @media (min-width: $large-screen) {
```
jbalboni: @cehsu I couldn't find anything that I broke by removing this, but do you remember what it was added for?

rianfowler (approved):

src/sass/modules/_m-schemaform.scss
cehsu: I'm not sure which specific screen size/browser/view scenario this addresses. The comment indicates this is to address an IE bug. It is part of a bug fix [PR](https://github.com/department-of-veterans-affairs/vets-website/pull/7338) for the photo tool page and its view within the review page. I'll test this out and report back.

jbalboni: I tested on IE and didn't notice any issues on the review page, at least.

jbalboni:
I'm going to merge this and get it released, since it's affecting all forms. If we find a place where it's causing problems in VIC, we can fix that later.

src/sass/modules/_m-schemaform.scss
cehsu: The issue is on the normal cropper page, not the review page:
screen shot 2018-03-09 at 3 03 03 pm
An issue has been created for this bug [here](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/9197).

PR #7355, opened 2018-02-28
separate cropper controls

rianfowler:
Moved Cropper and the controls into a `CropperController`. This handles all of the interactions between the buttons and the Cropper and provides a simplified interface for use by the `PhotoField` layout.


src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -704,179 +705,135 @@ export default class PhotoField extends React.Component {
...
-                onMouseDown={this.onZoomMouseDown}
-                onMouseUp={this.onZoomMouseUp}
-                onMouseMove={this.onZoomMouseMove}
-                onChange={this.onZoomSliderChange}/>
-              {smallScreen && <button className="cropper-control cropper-control-zoom cropper-control-zoom-in va-button va-button-link" type="button" onClick={() => this.handleZoomButtonClick('IN')}><i className="fa fa-search-plus"></i></button>}
-              {!smallScreen && <button className="cropper-control cropper-control-zoom cropper-control-zoom-in va-button va-button-link" type="button" onClick={() => this.handleZoomButtonClick('IN')}>
-                <span className="cropper-control-label">Make larger<i className="fa fa-search-plus"></i></span>
-              </button>}
-            </div>
-            <div className="cropper-control-container">
```
rianfowler: Moving this div into a separate Copmponent

src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -704,179 +705,135 @@ export default class PhotoField extends React.Component {
...
+                    onMouseDown={this.onZoomMouseDown}
+                    onMouseUp={this.onZoomMouseUp}
+                    onMouseMove={this.onZoomMouseMove}
+                    onChange={this.onZoomSliderChange}/>
+                  {smallScreen && <button className="cropper-control cropper-control-zoom cropper-control-zoom-in va-button va-button-link" type="button" onClick={() => this.handleZoomButtonClick('IN')}><i className="fa fa-search-plus"></i></button>}
+                  {!smallScreen && <button className="cropper-control cropper-control-zoom cropper-control-zoom-in va-button va-button-link" type="button" onClick={() => this.handleZoomButtonClick('IN')}>
+                    <span className="cropper-control-label">Make larger<i className="fa fa-search-plus"></i></span>
+                  </button>}
+                </div>
+                <CropperControls
```
rianfowler: new component

src/js/vic-v2/components/CropperControls.jsx
```diff
@@ -0,0 +1,77 @@
...
+      }, {
+        action: 'Rotate',
+        icon: 'rotate',
+        direction: 'right',
+        disabled: false
+      }]
+    ].map((row, index) => (
+      <div className="cropper-control-row" key={index}>
+        {row.map((button) => (
+          <button className={classNames(moveControlClass, { disabled: button.disabled })} type="button" onClick={() => move(`${button.action} ${button.direction}`)} key={button.direction}>
```
rianfowler: For now the move and rotate functions use the same callback. Will tweak to use different functions (even if `PhotoField` passes the same function twice) and simplify the values sent during the events.

jbalboni:
I think keeping them flat in the components folder is fine. It's not likely to be more than a few.

src/js/vic-v2/components/PhotoField.jsx
jbalboni: The name of this sort of implies to me that the range slider would be in it as well.

rianfowler: That's true- I'll rework the markup / classNames a bit to make that work.

src/js/vic-v2/components/CropperControls.jsx
```diff
@@ -0,0 +1,113 @@
...
+  onSliderMouseDown,
+  onSliderMouseMove,
+  onSliderMouseUp,
+  rotateLeft,
+  rotateRight,
+  sliderMaxValue,
+  sliderMinValue,
+  sliderValue,
+  zoomIn,
+  zoomOut
```
rianfowler: Next step: I'm going to move the CropperJS into this file and call this a `CropperController`- it will be responsible for taking UI actions, translating them into commands for the `CropperJS` canvas, and then sending those commands to the canvas. That should clean up `PhotoField` so that it's only concerned with layout and get rid of the need for this giant list of props.

src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -703,129 +360,28 @@ export default class PhotoField extends React.Component {
...
-            <div className="crop-button-container">
-              <button type="button" className="usa-button-primary" onClick={this.onDone}>
-                I’m Done
-              </button>
-            </div>
-          </div>
+          {fieldView === 'cropper' && <CropperController
+            narrowLayout={smallScreen}
+            onPhotoCropped={blob => this.uploadPhoto(blob)}
+            src={this.state.src}/>
```
rianfowler: Only need to provide the image, a callback, and a layout option.

src/js/vic-v2/components/CropperController.jsx
```diff
@@ -0,0 +1,482 @@
...
+            <button className="cropper-control cropper-control-label-container va-button va-button-link" type="button" onClick={this.onZoomOut}>
+              <span className="cropper-control-label">Make smaller</span>
+            </button>
+            <button className="cropper-control cropper-control-label-container va-button va-button-link" type="button" onClick={this.onZoomIn}>
+              <span className="cropper-control-label">Make larger</span>
+            </button>
+          </div>}
+          <div className="cropper-control-row">
+            <MoveRotateButton disabled={this.state.moveUpDisabled} onClick={this.onMoveUp} label="Move up" iconClassName="fa fa-arrow-up"/>
+            <MoveRotateButton disabled={this.state.moveDownDisabled} onClick={this.onMoveDown} label="Move down" iconClassName="fa fa-arrow-down"/>
```
rianfowler: Handler function for each button.

src/js/vic-v2/components/CropperController.jsx
```diff
@@ -0,0 +1,482 @@
...
+      // move photo within bounds
+      this.movePhoto();
+
+      this.updateWarningAndButtonState();
+    });
+  }
+
+  onMoveDown = () => {
+    this.movePhoto({ y: 5 });
+    window.requestAnimationFrame(() => this.updateWarningAndButtonState());
```
rianfowler: Split up a bunch of a behaviors so the button handlers can show the sequence of actions. Should be easier to understand what's going on now.

src/js/vic-v2/components/CropperController.jsx
```diff
@@ -0,0 +1,482 @@
...
+    const defaultPhotoPosition = getDefaultPhotoPosition({
+      photoData: this.refs.cropper.getCanvasData(),
+      cropBoxData: this.refs.cropper.getCropBoxData(),
+      containerWidth: this.refs.cropper.getContainerData().width
+    });
+    this.movePhotoToPosition(defaultPhotoPosition);
+  }
+
+  // relative positioning
+  movePhoto = ({ x = 0, y = 0 } = {}) => {
```
rianfowler: Ditched the canvas nomenclature since it's not relevant to our concerns.

src/js/vic-v2/components/CropperController.jsx
```diff
@@ -0,0 +1,480 @@
...
+          cropmove={() => true}
+          minContainerHeight={this.props.narrowLayout ? SMALL_CROP_BOX_SIZE : LARGE_CROP_BOX_SIZE}
+          toggleDragModeOnDblclick={false}
+          dragMode="move"
+          guides={false}
+          viewMode={0}
+          zoom={this.onZoom}/>
+        <div className="cropper-zoom-container">
+          <button className="cropper-control cropper-control-zoom cropper-control-zoom-in va-button va-button-link" type="button" onClick={this.onZoomOut}>
+            <span className="cropper-control-label">{this.props.narrowLayout || 'Make smaller'}<i className="fa fa-search-minus"></i></span>
```
rianfowler: This icon will always have a span around it now but it seems to work the same and made the code [simpler](https://github.com/department-of-veterans-affairs/vets-website/blob/5d6dc7f6ec81286911ace84f4a72c7c9ba78e3a1/src/js/vic-v2/components/PhotoField.jsx#L744).

jbalboni (approved):
LGTM. I think this is a good split of functionality. All the cropping code makes sense in one file. If we break it up further, we're likely going to make ourselves switch back and forth between the different components.

PR #7358, opened 2018-02-28
Depends callback mod cv

cvalarida:
Took a long time to figure out the right way to do it (thanks, @jbalboni!), but I think this is the least lift possible. We may have to revisit this for the review page stuff.

Just so we have an artifact of how this is all tied together, the I took the following notes while I was trying to figure out how to go about making this change. I don't guarantee it'll make any sense, but basically I was trying to create a call stack of all the possible ways to get to the `depends` callback.

---
# Refactoring the `depends` callback

- [x] src/js/common/utils/helpers.js -- isActivePage()
    - This is where the `depends` callback comes from
    - [ ] src/js/common/utils/helpers.js -- getActivePages()
        - [ ] src/js/common/components/NavButtons.jsx
            - I don't think this is even being used anywhere...
            - We're using ProgressButton, but not NavButtons, it looks like
            - grep -F NavButtons src/js/ -R
        - [ ] src/js/common/schemaform/helpers.js -- getPageKeys
            - [ ] src/js/common/schemaform/review/ReviewCollapsibleChapter.jsx
                - [ ] src/js/common/schemaform/review/ReviewPage.jsx
            - [ ] src/js/common/schemaform/review/ReviewPage.jsx
        - [ ] src/js/common/schemaform/review/ReviewCollapsibleChapter.jsx
            - [ ] src/js/common/schemaform/review/ReviewPage.jsx
        - [ ] src/js/common/schemaform/review/ReviewPage.jsx
        - [x] src/js/common/schemaform/routing.js -- getEligiblePages()
            - Modified to use the expanded page list, which has page.index
            - [ ] src/js/common/schemaform/routing.js -- getNextPagePath()
                - [ ] src/js/common/schemaform/containers/FormPage.jsx
            - [ ] src/js/common/schemaform/routing.js -- getPreviousPagePath()
                - [ ] src/js/common/schemaform/containers/FormPage.jsx
        - [x] src/js/common/schemaform/components/FormNav.jsx
            - Modified to use the expanded page list, which has page.index
            - [ ] src/js/common/schemaform/containers/FormApp.jsx
            - [ ] src/js/common/schemaform/save-in-progress/RoutedSavableApp.jsx
    - [ ] src/js/common/utils/helpers.js -- getInactivePages()
        - [ ] src/js/hca/helpers.jsx -- transform()
            - Searching for this one was a bit trickier because `transform` isn't exactly unique, but I couldn't find anywhere that imported this one specifically
            - vimgrep /\<transform\>/ src/js/**
              - Then search the quickfix list for `import`
        - [ ] src/js/common/schemaform/helpers.js -- transformForSubmit()
            - [ ] src/js/burials/helpers.jsx -- transform()
            - [ ] src/js/vic-v2/helpers.jsx -- submit()
            - [ ] src/js/disability-benefits/526EZ/helpers.jsx -- transform()
            - [ ] src/js/vre/chapter36/helpers.js -- transform()
            - [ ] src/js/common/schemaform/actions.js -- submitForm()
            - [ ] src/js/edu-benefits/1990/helpers.jsx -- transform()
            - [ ] src/js/edu-benefits/1990e/helpers.jsx -- transform()
            - [ ] src/js/edu-benefits/1995/helpers.jsx -- transform()
            - [ ] src/js/edu-benefits/5495/helpers.js -- transform()
            - [ ] src/js/edu-benefits/5490/helpers.jsx -- transform()
            - [ ] src/js/edu-benefits/1990n/helpers.js -- transform()
            - [ ] src/js/pensions/helpers.jsx -- transform()
    - [ ] src/js/common/schemaform/validation.js -- isValidForm()
        - [ ] src/js/common/schemaform/review/ReviewPage.jsx
            - [ ] src/js/common/schemaform/helpers.js -- createRoutes
                - I don't think we need to modify this...right?
            - [ ] src/js/common/schemaform/save-in-progress/RoutedSavableReviewPage.jsx
                - [ ] src/js/common/schemaform/helpers.js -- createRoutes
                    - I don't think we need to modify this...right?
    - [ ] src/js/common/components/Nav.jsx
        - I don't think this is being used...couldn't find any imports of this
        - vimgrep /\<Nav\>/ src/js/**
          - Then search the quickfix list for `import`

### DIFF
--- a/src/js/common/schemaform/routing.js
+++ b/src/js/common/schemaform/routing.js
@@ -11,11 +11,7 @@ function getEligiblePages(pageList, data, pathname) {
   // We update the `path` for each of those pages to replace `:index` with the current item index.
   const expandedPageList = expandArrayPages(pageList, data);
   const pageIndex = _.findIndex(item => item.path === pathname, expandedPageList);
-  const eligiblePageList = getActivePages(expandedPageList, data, expandedPageList[pageIndex].index);
-  console.log('getEligiblePages -- expandedPageList:', eligiblePageList);
-  console.log('  pathname:', pathname);
-  console.log('  pageIndex:', pageIndex);
-  console.log('  array page index:', expandedPageList[pageIndex].index);
+  const eligiblePageList = getActivePages(expandedPageList, data);
   return { pages: eligiblePageList, pageIndex };
 }
 
--- a/src/js/common/schemaform/routing.js
+++ b/src/js/common/schemaform/routing.js
@@ -7,12 +7,16 @@ import { expandArrayPages } from './helpers';
  * their dependencies and therefore should be skipped.
  */
 function getEligiblePages(pageList, data, pathname) {
-  const eligiblePageList = getActivePages(pageList, data);
   // Any `showPagePerItem` pages are expanded to create items for each array item.
   // We update the `path` for each of those pages to replace `:index` with the current item index.
-  const expandedPageList = expandArrayPages(eligiblePageList, data);
+  const expandedPageList = expandArrayPages(pageList, data);
   const pageIndex = _.findIndex(item => item.path === pathname, expandedPageList);
-  return { pages: expandedPageList, pageIndex };
+  const eligiblePageList = getActivePages(expandedPageList, data, expandedPageList[pageIndex].index);
+  console.log('getEligiblePages -- expandedPageList:', eligiblePageList);
+  console.log('  pathname:', pathname);
+  console.log('  pageIndex:', pageIndex);
+  console.log('  array page index:', expandedPageList[pageIndex].index);
+  return { pages: eligiblePageList, pageIndex };
 }
 
 export function getNextPagePath(pageList, data, pathname) {
--- a/src/js/common/schemaform/routing.js
+++ b/src/js/common/schemaform/routing.js
@@ -10,8 +10,8 @@ function getEligiblePages(pageList, data, pathname) {
   // Any `showPagePerItem` pages are expanded to create items for each array item.
   // We update the `path` for each of those pages to replace `:index` with the current item index.
   const expandedPageList = expandArrayPages(pageList, data);
-  const pageIndex = _.findIndex(item => item.path === pathname, expandedPageList);
   const eligiblePageList = getActivePages(expandedPageList, data);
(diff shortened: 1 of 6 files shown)


cvalarida:
I'm not sure where exactly to test this--I didn't touch `src/js/common/components/Nav.jsx`, but that seems to be where most of the `depends` callback testing is.

src/js/common/schemaform/routing.js
```diff
@@ -7,12 +7,12 @@ import { expandArrayPages } from './helpers';
...
  */
 function getEligiblePages(pageList, data, pathname) {
-  const eligiblePageList = getActivePages(pageList, data);
   // Any `showPagePerItem` pages are expanded to create items for each array item.
   // We update the `path` for each of those pages to replace `:index` with the current item index.
-  const expandedPageList = expandArrayPages(eligiblePageList, data);
+  const expandedPageList = expandArrayPages(pageList, data);
   const pageIndex = _.findIndex(item => item.path === pathname, expandedPageList);
-  return { pages: expandedPageList, pageIndex };
+  const eligiblePageList = getActivePages(expandedPageList, data);
```
jbalboni: I think this needs to be above the line that gets `pageIndex`. Otherwise, we might get the wrong index if some pages are filtered out in the this step?

src/js/common/utils/helpers.js
```diff
@@ -66,7 +66,8 @@ export function isActivePage(page, data) {
 }
 
 export function getActivePages(pages, data) {
-  return pages.filter(page => isActivePage(page, data));
+  const active =  pages.filter(page => isActivePage(page, data));
```
jbalboni: Does this change need to be here now?

jbalboni (commented):
Looks good, I think we need some tests for the change, though.

jbalboni:
@cvalarida There are tests for routing.js, so that's one place.

src/js/common/schemaform/routing.js
cvalarida: :thinking: Yeah, good catch.

src/js/common/utils/helpers.js
cvalarida: Ah, probably not. :+1:

jbalboni (approved):
LGTM

PR #7359, opened 2018-02-28
Don't allow addresses through with just street, city, and zip

jbalboni:
The way we handle partial addresses is that if you have city, street, and zip, we allow it through. For countries that would normally have a state, this is weird and can cause problems. Specifically, on edu forms it can break the routing of apps to the right processing center.

This adds a validation message if you supply city, state, and zip and not state for non-required addresses. We may in the future want to disallow all partial addresses, but that's a bigger change.

I also updated the tests for the address definition to use enzyme.

screen shot 2018-02-28 at 4 59 16 pm

### DIFF
--- a/src/js/common/schemaform/definitions/address.js
+++ b/src/js/common/schemaform/definitions/address.js
@@ -31,6 +31,16 @@ function validateAddress(errors, address, formData, currentSchema) {
     errors.state.addError('Please select a state or province');
   }
 
+  const hasAddressInfo = stateRequiredCountries.has(address.country)
+    && !currentSchema.required.length
+    && typeof address.street !== 'undefined'
+    && typeof address.city !== 'undefined'
+    && typeof address.postalCode !== 'undefined';
+
+  if (hasAddressInfo && typeof address.state === 'undefined') {
+    errors.state.addError('Please select a state or province or remove other address information');
+  }
+
   validatePostalCodes(errors, address);
 }
 
--- a/src/js/common/schemaform/definitions/address.js
+++ b/src/js/common/schemaform/definitions/address.js
@@ -38,7 +38,7 @@ function validateAddress(errors, address, formData, currentSchema) {
     && typeof address.postalCode !== 'undefined';
 
   if (hasAddressInfo && typeof address.state === 'undefined') {
-    errors.state.addError('Please select a state or province or remove other address information');
+    errors.state.addError('Please enter a state or province, or remove other address information.');
   }
 
   validatePostalCodes(errors, address);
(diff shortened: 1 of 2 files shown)


src/js/common/schemaform/definitions/address.js
```diff
@@ -31,6 +31,16 @@ function validateAddress(errors, address, formData, currentSchema) {
     errors.state.addError('Please select a state or province');
   }
 
+  const hasAddressInfo = stateRequiredCountries.has(address.country)
+    && !currentSchema.required.length
+    && typeof address.street !== 'undefined'
+    && typeof address.city !== 'undefined'
```
rianfowler: If I add a city and then delete it, it becomes undefined or an empty string?

test/common/schemaform/definitions/address.unit.spec.jsx
```diff
@@ -17,152 +16,127 @@ describe('Schemaform definition address', () => {
...
-
-    ReactTestUtils.Simulate.change(countryField, {
-      target: {
-        value: 'CAN'
-      }
-    });
+    fillData(form, 'select#root_country', 'CAN');
 
     // Check to see if the postal code and state updated
-    expect(stateLabel.textContent === 'Provice');
```
rianfowler: wonder how this ever passed

peggygannon (commented):
@jbalboni Thanks. Does this message replace the current: "Please enter a valid state/province." message? 

If so, then let's tweak to: Please enter a valid state or province, or remove other address information.

rianfowler (approved):
LGTM with content updates from @peggygannon

test/common/schemaform/definitions/address.unit.spec.jsx
jbalboni: it just resolves to expect(false), which isn't an assertion. I'm not sure how we missed that originally.

jbalboni:
@peggygannon No, it's a new message for the case where the address isn't required, but you've put in a street, city, and postal code and missed state. You either need to enter a state or remove the other address info. The other state message is for when the address is required and you have to enter a state.

It doesn't have to be different from the other message, but it's a slightly different scenario.

peggygannon:
@jbalboni Ok, thanks. so the below message looks good to me. 

Please enter a state or province, or remove other address information.

src/js/common/schemaform/definitions/address.js
jbalboni: It becomes undefined. Empty strings don't trigger required errors in JSON schema land, so all the form widgets removes data entirely instead of using an empty string.

PR #7366, opened 2018-03-02
Send VIC application as anonymous if fields differ

jbalboni:
screen shot 2018-03-06 at 1 08 31 pm

### DIFF
--- a/src/js/vic-v2/helpers.jsx
+++ b/src/js/vic-v2/helpers.jsx
@@ -23,7 +23,8 @@ export function prefillTransformer(pages, formData, metadata, state) {
     newData = _.set('processAsIdProofed', true, newData);
     newData.originalUser = {
       veteranSocialSecurityNumber: newData.veteranSocialSecurityNumber,
-      veteranFullName: newData.veteranFullName
+      veteranFullName: newData.veteranFullName,
+      veteranDateOfBirth: newData.veteranDateOfBirth
     };
   }
 
--- a/src/js/vic-v2/helpers.jsx
+++ b/src/js/vic-v2/helpers.jsx
@@ -3,33 +3,30 @@ import Raven from 'raven-js';
 import environment from '../common/helpers/environment.js';
 import { transformForSubmit } from '../common/schemaform/helpers';
 
-export function prefillTransformer(pages, formData, metadata) {
+export function prefillTransformer(pages, formData, metadata, state) {
+  let newPages = pages;
+  let newData = formData;
+
   if (formData && formData.serviceBranches) {
     // Mostly we'll be getting branch lists of one or two branches, creating a Set seems like overkill
     const allowedBranches = pages.veteranInformation.schema.properties.serviceBranch.enum;
     const validUserBranches = formData.serviceBranches.filter(branch => allowedBranches.includes(branch));
 
-    const newData = _.unset('serviceBranches', formData);
+    newData = _.unset('serviceBranches', newData);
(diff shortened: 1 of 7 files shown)


src/js/vic-v2/helpers.jsx
```diff
@@ -3,36 +3,67 @@ import Raven from 'raven-js';
...
+
+  // If we pulled their id info at the start and they haven't changed it, then we can submit on the
+  // backend with id info from MVI and discharge status from eMIS
+  // If they changed it, then we have to verify they're not trying to submit a fradulent
+  // request and process them as an anonymous request
+  if (form.data.processAsIdProofed && identityMatchesPrefill(form.data)) {
+    newData = _.omit([
+      'veteranFullName',
+      'veteranSocialSecurityNumber'
+    ], newData);
```
rianfowler: Why do we omit these?

src/js/vic-v2/helpers.jsx
```diff
@@ -3,36 +3,67 @@ import Raven from 'raven-js';
...
+    && formData.veteranDateOfBirth === originalUser.veteranDateOfBirth;
+}
+
+export function transform(form, formConfig) {
+  let newData = _.omit(['verified', 'originalUser', 'processAsIdProofed'], form.data);
+
+  // If we pulled their id info at the start and they haven't changed it, then we can submit on the
+  // backend with id info from MVI and discharge status from eMIS
+  // If they changed it, then we have to verify they're not trying to submit a fradulent
+  // request and process them as an anonymous request
```
rianfowler: What's the significance of processing as an anonymous request? This changes the way its handled by Salesforce?

jbalboni: Yes, it means that admins in SF will need to review their discharge documentation and verify their identity.

src/js/vic-v2/helpers.jsx
```diff
@@ -156,6 +187,9 @@ export function submit(form, formConfig) {
     }).then(resp => {
       const guid = resp.data.attributes.guid;
       pollStatus(guid, response => {
+        window.dataLayer.push({
+          event: `${formConfig.trackingPrefix}-submission-successful`,
```
rianfowler: Is this some sort of analytics thing?

rianfowler (approved):
Looks good to me- what makes this a WIP?

src/js/vic-v2/helpers.jsx
jbalboni: The backend is going to get these from MVI instead and throw an error if they're present in the form data (and processAsAnonymous is not present). It could ignore them in that case instead, but since it's not an expected situation, an error seems better.

src/js/vic-v2/helpers.jsx
jbalboni: Yeah, it's how we send events to Google Analytics.

jbalboni:
@rianfowler Nothing, I forgot to remove the WIP from the title.

peggygannon (commented):
@jbalboni Looks good.Only tweak is changing we will to we'll (sorry I should have fixed that before I sent it). I updated the file. Hope that's OK.

jbalboni:
@peggygannon Yep, looks like I actually took a screenshot before I updated it. It's correct in the code:

https://github.com/department-of-veterans-affairs/vets-website/pull/7366/files#diff-3e447372e27883809fdcd8a57dbfc640R10

PR #7367, opened 2018-03-02
Routing documentation

jbalboni:
Very late, but here's a cleaned up version of my notes on routing from the FE meeting.

### DIFF
--- a/docs/Routing101.md
+++ b/docs/Routing101.md
@@ -3,7 +3,7 @@
 ### Production
 The production deployment of the vet-website front end consists of static HTML, CSS, and JS assets. We have an nginx server that serves those static assets and does a little bit of extra route handling for our single page React apps.
 
-React applications have a single entry page in the `content/pages` folder and a special nginx [config entry](https://github.com/department-of-veterans-affairs/devops/blob/master/ansible/roles/revproxy-configure/defaults/main.yml#L91). For each of the urls listed in that section of the config, the nginx server routes anything that starts with that url to the static page at that url, instead of trying to find a content page at that spot in the `content/pages` folder structure.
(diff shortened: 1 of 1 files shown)


docs/Routing101.md
```diff
@@ -0,0 +1,28 @@
+## Routing for React apps
+
+### Production
+The vets-website front end deployed to production consists of static HTML, CSS, and JS assets. We have an nginx server that serves those static assets and does a little bit of extra route handling for our single page React apps.
```
rianfowler: Maybe
>Production deployment of the vet-website front end consists off..

or 

> The vets-website front end that's deployed...

docs/Routing101.md
```diff
@@ -0,0 +1,28 @@
...
+
+React applications have a single entry page in the `content/pages` folder and a special nginx [config entry](https://github.com/department-of-veterans-affairs/devops/blob/master/ansible/roles/revproxy-configure/defaults/main.yml#L91). For each of the urls listed in that section of the config, the nginx server routes anything that starts with that url to the static page at that url, instead of trying to find a content page at that spot in the `content/pages` folder structure.
```
rianfowler: Maybe provide an example
>... and a special nginx config entry. This nginx config alters the default routing behavior to `content/pages` by routing paths that match the beginning of an entry to a single static page. For example `https://www.vets.gov/letters/more/path/parts` would be routed to `https://www.vets.gov/letters`

I don't know enough about nginx to read this config- do all of the React apps exist in one singe page app? Might be useful to mention that if its the case.

rianfowler: I guess I don't really like the wording of my suggestions. Might be more useful to describe it as nginx and react router working in tandem to determine the correct page. Server side nginx provides several routes the same static page and front end react router renders the correct content.

rianfowler (approved):

docs/Routing101.md
```diff
@@ -0,0 +1,30 @@
+## Routing for React apps
+
+### Production
+The production deployment of the vet-website front end consists of static HTML, CSS, and JS assets. We have an nginx server that serves those static assets and does a little bit of extra route handling for our single page React apps.
+
```
ncksllvn: Is it worth noting at all anything about S3?

ncksllvn (approved):

docs/Routing101.md
jbalboni: Wouldn't hurt

PR #7369, opened 2018-03-02
Organize into personalization directory; do beta stuff

ncksllvn:
Some things to note -

- This gets a little more confusing because we can no longer reuse the "user" Redux store, since we have to coexist with the current user profile. It seems like that wouldn't be the case since it's either one or the other for each user, but the way the common store defines everything is by pulling from that user-profile/reducers/ directory and I wanted to leave that untouched.
- We'll need to update the dashboard module references when you move that into the personalization directory. This should be only the common and login imports.
- I committed with `--no-verify` because I'm a cheater and didn't want to fix the lint errors, which were that unused files we copied over from the `user-profile` app have incorrect references now that they are in a subdirectory.
- We'll need to update `SignInProfileMenu.jsx` when the API is configured with the beta information we need.


bshyong (approved):
LGTM

bshyong:
Merging into dev branch

PR #7371, opened 2018-03-02
Letters blank address fix cv

cvalarida:
When we get a blank address from EVSS, the user starts at this screen:
image
Not featured in the above screenshot is the disabled "View letters" button below the edit content.

If they cancel out of that, they'll go to the view state, which looks like this:
image

Valid addresses still show up as expected:
image

### DIFF
--- a/src/js/letters/containers/AddressSection.jsx
+++ b/src/js/letters/containers/AddressSection.jsx
@@ -16,7 +16,8 @@ import {
   isDomesticAddress,
   isInternationalAddress,
   isMilitaryAddress,
-  resetDisallowedAddressFields
+  resetDisallowedAddressFields,
+  isAddressEmpty
 } from '../utils/helpers';
 import {
   saveAddress,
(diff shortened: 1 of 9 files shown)


cvalarida:
Tests are on the way. I just wanted to get these screenshots out while I had it ready.

src/js/letters/containers/AddressSection.jsx
```diff
@@ -61,9 +70,14 @@ export class AddressSection extends React.Component {
     // we tell React that editableAddress has already been initialized with the
     // savedAddress values
     if (Object.keys(this.state.editableAddress).length > 0) {
```
kreek: Could everything inside the if statement be extracted into a self-documenting method? I think you can guard on address being not empty too to simply it e.g. 

```javascript
initializeSavedAddressValues() {
    this.setState({ hasLoadedAddress: true })
    if (!isAddressEmpty(this.state.editableAddress)) { return; }
    if (this.props.canUpdate) {
      this.props.editAddress();
    } else {
      Raven.captureMessage('letters_empty_address_update_not_allowed');
    }
  }
```

kreek (approved):
one minor comment but LGTM 🚀

src/js/letters/containers/AddressSection.jsx
cvalarida: :thinking: Yeah...could probably do that, but there are a couple of odd React things to watch out for.

First, calling `setState()` in a lifecycle method is generally frowned on, and the [official React docs](https://reactjs.org/docs/react-component.html#constructor) say "don’t try to call `setState()` from the constructor." We could get around this by just using `this.state.hasLoadedAddress = true;` again, but...then it's not usable outside of the constructor.

At that point, there really isn't much use for pulling it out of the constructor, right? :man_shrugging:

cvalarida:
Merging this so we can get it into production sooner rather than later. If we want to pull that functionality out into its own function, we can make another ticket for it.

src/js/letters/containers/AddressSection.jsx
kreek: If it does remain in the constructor I think you can still clean it up a bit. There’s a couple places guard clauses could simplify the code:

1. `Object.keys(this.state.editableAddress).length > 0` the function can return if that’s not the case, as a bonus you lose a level of nesting.
2. later if `!isAddressEmpty(this.state.editableAddress)` you can also exit early
3. we’re now only dealing with empty addresses and don’t need that check in the remaining if statement:
```js
if (this.props.canUpdate) {
  this.props.editAddress();
} else {
  Raven.captureMessage('letters_empty_address_update_not_allowed');
}
```

PR #7377, opened 2018-03-06
526 evidence type

cvalarida:
First off, there are a few things that don't match the prototype:
1. ~The title doesn't have dynamic content~
2. ~The title doesn't have the blue underline~
3. ~The description doesn't have dynamic content~
4. The spacing between the check boxes is off
    - I forget how this is done...@jbalboni?

@goldenmeanie If you notice anything else, please let me know.

## Collapsed
image

## Expanded
image


package.json
```diff
@@ -220,7 +220,7 @@
     "react-dropzone": "^4.2.5",
     "reselect": "^2.5.4",
     "url-search-params": "^0.10.0",
-    "vets-json-schema": "https://github.com/department-of-veterans-affairs/vets-json-schema.git#579f0fa2bf0641dd50012e09a98e891a426a54bf",
+    "vets-json-schema": "https://github.com/department-of-veterans-affairs/vets-json-schema.git#89739dddef46862cd10173e21712682f8f9cb375",
```
cvalarida: I guess we're still not using it, so this isn't super important. I thought we'd get around to it this PR, but realized after I started, these fields are just `view:`

src/js/disability-benefits/526EZ/config/form.js
```diff
@@ -85,8 +147,84 @@ const formConfig = {
...
+              }
+            }
+          }
+        },
+        vaMedicalRecordsIntro: {
+          title: '',
+          path: 'supporting-evidence/:index/va-medical-records',
+          showPagePerItem: true,
+          arrayPath: 'disabilities',
+          depends: (formData, index) => _.get(`disabilities.${index}.view:vaMedicalRecords`, formData),
```
cvalarida: This was just from experimenting with the array `depends` functionality.

src/js/disability-benefits/526EZ/helpers.jsx
```diff
@@ -14,3 +15,52 @@ export function transform(formConfig, form) {
...
 );
+
+// TODO: Change this into a function which returns the content with the appropriate [condition].
+export const evidenceTypesDescription = (
+  <p>What supporting evidence do you have that shows how your [condition] <strong>has worsened since VA rated your disability</strong>?</p>
+);
+
+
+// Shows or collapses the "Which should I choose?" link at the bottom of the evidence types page
+// TODO: Investigate whether this should just use `expandUnder`
```
cvalarida: This is a different pattern than other `expandUnder` items in that it's just a link, so I opted to just use a basic React component to handle the state.

src/sass/disability-benefits.scss
```diff
@@ -9,3 +9,12 @@
 // @import "modules/m-omb-info";
 @import "modules/m-form-confirmation";
 
+.dashed-underline {
+  text-decoration: none;
+  border-bottom: 1px dashed;
+
+  &:hover {
```
cvalarida: Because `.va-button-link` also has `&:hover`, this is needed.

goldenmeanie (approved):
Other than noted issues, only issue I see is the spacing in the helper text. The headlines seem very separate from their paragraphs. But this is a tendency across Vets.gov and should maybe be addressed as a holistic fix.

cvalarida:
Yeah, the spacing on basically all the things in this page bugs me... :grimacing: 
I'm just not sure how to fix it without (negatively) affecting all the other forms. :confused:

cvalarida:
**Update:** Good news! I was able to grab the data from the specific array item in the title and description without modifying the library. Many thanks to @jbalboni for this.

Updated screenshot:
image

src/js/disability-benefits/526EZ/config/form.js
```diff
@@ -85,8 +148,84 @@ const formConfig = {
             properties: {}
           }
         },
-        // pageTwo: {},
-        // pageThree: {},
+        evidenceType: {
+          title: (formData, { pagePerItemIndex }) => formData.disabilities[pagePerItemIndex].diagnosticText,
```
jbalboni: You might want to use `get` for this, since the array won't always be there.

src/js/disability-benefits/526EZ/helpers.jsx
```diff
@@ -14,3 +15,61 @@ export function transform(formConfig, form) {
...
+export const evidenceTypesDescription = ({ formData }) => {
+  return (
+    <p>What supporting evidence do you have that shows how your {formData.disability.diagnosticText} <strong>has worsened since VA rated your disability</strong>?</p>
+  );
+};
+
+
+// Shows or collapses the "Which should I choose?" link at the bottom of the evidence types page
+// TODO: Investigate whether this should just use `expandUnder`
+export class evidenceTypeHelp extends React.Component {
```
jbalboni: I think the name of this should be title cased. Components and class functions are normally title case.

src/sass/disability-benefits.scss
jbalboni: Is this something we should do everywhere?

jbalboni (approved):

src/sass/disability-benefits.scss
cvalarida: :man_shrugging: 
I can put it in `base/_m-utils.scss`.

src/js/disability-benefits/526EZ/config/form.js
cvalarida: Oh right. Good point.

PR #7387, opened 2018-03-07
Partially replaced use of LOA with new 'verified' flag.

U-DON:
Resolves department-of-veterans-affairs/vets.gov-team#9096.
Depends on department-of-veterans-affairs/vets-api#1752.

This makes use of a `verified` flag that is being added to the serialized user object instead of the `loa` object. This is the first part in phasing out usage of `loa` on the front-end. It will be completely phased out once I refactor the RequiredLoginView component, which ends up affecting other teams, so I wanted to make a separate PR for that.

--

Although this remove the explicit reference to LOA for API consumers, having a `verified` flag is still an implicit reference to LOA. It's a short term approach, and will eventually be replaced altogether.

The main purpose of LOA for the front-end right now is to determine whether the user should be prompted to "verify" (or identity proof) their account, which could possibly give them access to certain apps/services. But potentially, there are services that don't automatically become accessible once identity proofed. Being LOA 3 is not necessarily the only criterion for access, and is often only one of many, but the front-end doesn't really handle the nuances of why a user might not be able to access the service.

Ideally, with authZ in the works, we can eventually rely on an API endpoint that doesn't expose the nitty gritty details of an account like LOA. One idea we're currently entertaining is a services hash in the form of something like `{ eligible: false, authorized: false, eligibility_reason: "Not a VA patient" }` that the front-end can handle more granularly. Another idea is that we could introduce a services controller to query for availability along with any relevant data. In any case, the goal is for the API to provide the F/E a minimal and relevant set of data that's sufficient to inform the user flow.

CC: @jkassemi. Wanted to open up discussion in light of authZ refactors and discovery.

### DIFF
--- a/src/js/common/helpers/login-helpers.js
+++ b/src/js/common/helpers/login-helpers.js
@@ -85,6 +85,7 @@ export function getUserData(dispatch) {
       authnContext: userData.authn_context,
       loa: userData.loa,
       multifactor: userData.multifactor,
+      verified: userData.verified,
       gender: userData.gender,
       dob: userData.birth_date,
       status: json.data.attributes.va_profile.status,
--- a/src/js/common/schemaform/save-in-progress/FormSaved.jsx
+++ b/src/js/common/schemaform/save-in-progress/FormSaved.jsx
@@ -37,12 +37,10 @@ class FormSaved extends React.Component {
   }
 
   render() {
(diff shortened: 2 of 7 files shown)


jbalboni (approved):
The save in progress changes look good to me.

PR #7389, opened 2018-03-07
Implement beta wrapper component into personalization features

ncksllvn:
Super barebones wrapper component so that the user has to explicitly enroll in the beta to gain access. Doing it this way we shouldn't need to create a separate route for registering users - just share the `/profile-beta` link and make sure that landing component is wrapped in the `<BetaApp>`. For now, it's saving the beta information in the session storage but will need to be adjusted based on what comes out of department-of-veterans-affairs/vets.gov-team/issues/9155.

image

Resolves department-of-veterans-affairs/vets.gov-team/issues/9097

### DIFF
--- a/src/js/personalization/user-profile-beta/containers/UserProfileApp.jsx
+++ b/src/js/personalization/user-profile-beta/containers/UserProfileApp.jsx
@@ -14,6 +14,7 @@ import {
   openModal
 } from '../actions';
 
+import BetaApp, { features } from '../../../common/containers/BetaApp';
 import RequiredLoginView from '../../../common/components/RequiredLoginView';
 import DowntimeNotification, { services } from '../../../common/containers/DowntimeNotification';
 import ProfileView from '../components/ProfileView';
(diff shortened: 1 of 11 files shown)


ncksllvn:
I have the beta work in a separate [branch](https://github.com/department-of-veterans-affairs/vets-website/tree/beta-component), but I think it's probably way too soon to consider adding any of that work to master.

src/js/personalization/user-profile-beta/containers/UserProfileApp.jsx
```diff
@@ -34,23 +35,25 @@ class UserProfileApp extends React.Component {
...
-                  modal={{
-                    open: this.props.openModal,
-                    currentlyOpen: this.props.profile.modal,
-                    pendingSaves: this.props.profile.pendingSaves,
-                    errors: this.props.profile.errors
-                  }}
-                  updateActions={this.props.updateActions}
-                  fetchExtendedProfile={this.props.fetchExtendedProfile}/>
+          <BetaApp featureName={features.personalization}>
+            <DowntimeNotification appTitle="user profile page" dependencies={[services.mvi, services.emis]}>
```
ncksllvn: Here's where the beta stuff comes into play for our work.

bshyong:
@ncksllvn LGTM; note that we may not need the session storage work because beta access products will be included in the available services once the API work in https://github.com/department-of-veterans-affairs/vets.gov-team/issues/9155 is complete

ncksllvn:
@bshyong Definitely, gonna make adjustments here soon

PR #7391, opened 2018-03-08
526 va facilities

cvalarida:
**TODO**
- [x] Update the version of `vets-json-schema`, which will change the order. That'll put the facility name before the dates.
- [x] Write a proper view component
    - @goldenmeanie What should it look like after a facility is added?

image

### DIFF
--- a/src/js/common/utils/data-utils/omit.js
+++ b/src/js/common/utils/data-utils/omit.js
@@ -1,13 +1,15 @@
-import cloneDeep from './cloneDeep';
-
 /**
- * Returns a deep clone of he provided object without the fields specified.
+ * Returns a deep clone of the provided object without the specified fields.
  *
  * @param {Array} fields
  * @param {Object} object
  */
 export default function omit(fields, object) {
-  const newObj = cloneDeep(object);
-  fields.forEach(f => delete newObj[f]);
-  return newObj;
+  return Object.keys(object).reduce((newObj, k) => {
+    if (!fields.includes(k)) {
+      newObj[k] = object[k]; // eslint-disable-line
+    }
+
+    return newObj;
+  }, {});
 }
--- a/src/js/common/utils/data-utils/omit.js
+++ b/src/js/common/utils/data-utils/omit.js
@@ -1,0 +1,13 @@
+import cloneDeep from './cloneDeep';
+
+/**
+ * Returns a deep clone of he provided object without the fields specified.
+ *
+ * @param {Array} fields
+ * @param {Object} object
+ */
+export default function omit(fields, object) {
+  const newObj = cloneDeep(object);
+  fields.forEach(f => delete newObj[f]);
+  return newObj;
+}
--- a/src/js/common/utils/data-utils/set.js
+++ b/src/js/common/utils/data-utils/set.js
@@ -11,9 +11,9 @@ import checkValidPath from './checkValidPath';
  *
  * TODO: Link to some documentation which will be a better explanation than the above
  *
- * @param {Array|Obect} object
  * @param {Array|string} path
  * @param {*} value
+ * @param {Array|Obect} object
  * @param {Number} level  How many times we've recursed
  * @return {Object} A new object with the appropriate value set
  */
@@ -57,9 +57,9 @@ function baseSet(arrayPath, value, object, level = 0) {
  * Sets the value at the end of the path, creating the appropriate objects along the way if needed.
  * Separate from `baseSet` to not expose the level param.
  *
- * @param {Obect} object
  * @param {Array|string} path
  * @param {*} value
+ * @param {Obect} object
  * @return {Object} A new object with the appropriate value set
  */
 export default function set(path, value, object) {
--- a/src/js/common/utils/data-utils/set.js
+++ b/src/js/common/utils/data-utils/set.js
@@ -13,7 +13,7 @@ import checkValidPath from './checkValidPath';
  *
  * @param {Array|string} path
  * @param {*} value
- * @param {Array|Obect} object
+ * @param {Array|Object} object
  * @param {Number} level  How many times we've recursed
  * @return {Object} A new object with the appropriate value set
  */
@@ -59,7 +59,7 @@ function baseSet(arrayPath, value, object, level = 0) {
  *
  * @param {Array|string} path
  * @param {*} value
- * @param {Obect} object
+ * @param {Object} object
  * @return {Object} A new object with the appropriate value set
  */
 export default function set(path, value, object) {
--- a/src/js/disability-benefits/526EZ/config/form.js
+++ b/src/js/disability-benefits/526EZ/config/form.js
@@ -26,16 +26,16 @@ const {
 } = fullSchema526EZ.definitions;
 
 // We may add these back in after the typeahead, but for now...
-const treatmentsSchema = _.cloneDeep(treatments);
-treatmentsSchema.items.properties.treatment.properties = _.omit(
-  [
-    'treatmentCenterType',
-    'treatmentCenterCountry',
-    'treatmentCenterState',
-    'treatmentCenterCity'
-  ],
-  treatmentsSchema.items.properties.treatment.properties
-);
+const treatmentsSchema = _.set('items.properties.treatment.properties',
+  _.omit(
+    [
+      'treatmentCenterType',
+      'treatmentCenterCountry',
+      'treatmentCenterState',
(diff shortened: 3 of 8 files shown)


cvalarida:
Updated vets-json-schema:
image

cvalarida:
Added a proper view component. (Still needs verified by @goldenmeanie.)

image

src/js/disability-benefits/526EZ/config/form.js
```diff
@@ -12,9 +12,31 @@ import {
...
+  treatments
+} = fullSchema526EZ.properties;
+
+const {
+  date
+} = fullSchema526EZ.definitions;
+
+// We may add these back in after the typeahead, but for now...
+const treatmentsSchema = _.cloneDeep(treatments);
+treatmentsSchema.items.properties.treatment.properties = _.omit(
```
cvalarida: Looks like we'll need to use these for private records, but we can handle this in that PR.

cehsu: Would it make sense to move this logic to the backend schema as done with 22-5490 and 21-526EZ?

src/js/disability-benefits/526EZ/helpers.jsx
```diff
@@ -77,8 +77,27 @@ export const disabilityNameTitle = ({ formData }) => {
 };
 
 
+export const facilityDescription = ({ formData }) => {
+  return (
+    <p>Tell us about facilities where VA treated you for {formData.disability.diagnosticText}, <strong>after you got your disability rating</strong>.</p>
```
cehsu: Is this content finalized?

src/js/disability-benefits/526EZ/helpers.jsx
```diff
@@ -77,8 +77,27 @@ export const disabilityNameTitle = ({ formData }) => {
...
+  );
+};
+
+
+export const treatmentView = ({ formData }) => {
+  const { startTreatment, endTreatment, treatmentCenterName } = formData.treatment;
+  return (
+    <div>
+      <strong>{treatmentCenterName}</strong>
+      <p>{startTreatment} — {endTreatment}</p>
```
cehsu: Are neither of these required? If so, perhaps something like the conditional templating logic in EducationPeriodVew might also work here:
https://github.com/department-of-veterans-affairs/vets-website/blob/3166cadf727fbcff6fcfd181c04d435545bc4f96/src/js/vre/chapter31/components/EducationPeriodView.jsx#L3-L20

cehsu: Should this date be 12/12/2012 or 12-12-2012? The latter seems like a lot of dashes if used thus: 12-12-2012—12-12-2012

cehsu (approved):
LGTM. I had a few minor questions and comments.

src/js/common/utils/data-utils/omit.js
```diff
@@ -0,0 +1,13 @@
+import cloneDeep from './cloneDeep';
+
+/**
+ * Returns a deep clone of he provided object without the fields specified.
+ *
+ * @param {Array} fields
+ * @param {Object} object
+ */
+export default function omit(fields, object) {
+  const newObj = cloneDeep(object);
```
jbalboni: We shouldn't deep clone this, it will break some assumptions we make about immutable data. You can use reduce to build a new object from the list of keys from the old one and skip the fields in `fields`.

src/js/disability-benefits/526EZ/config/form.js
```diff
@@ -12,9 +12,31 @@ import {
...
+const {
+  treatments
+} = fullSchema526EZ.properties;
+
+const {
+  date
+} = fullSchema526EZ.definitions;
+
+// We may add these back in after the typeahead, but for now...
+const treatmentsSchema = _.cloneDeep(treatments);
```
jbalboni: This seems unnecessary. You could skip and use _.set instead of the assignment statement below.

src/js/common/utils/data-utils/omit.js
```diff
@@ -0,0 +1,13 @@
+import cloneDeep from './cloneDeep';
+
+/**
+ * Returns a deep clone of he provided object without the fields specified.
```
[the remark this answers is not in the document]
cvalarida: Works for me! :+1:

src/js/disability-benefits/526EZ/helpers.jsx
cvalarida: :man_shrugging:

src/js/disability-benefits/526EZ/helpers.jsx
cvalarida: Good call. I have zero clue about validation yet, though.

src/js/common/utils/data-utils/omit.js
cvalarida: Using reduce makes a lot of sense.

Is the assumption you're talking about the bit where we say "data is immutable; don't change it?" So we shouldn't _need_ to deep clone this? My concern was with nested objects. If we had something like so:

```js
const objA = { a: { b: 'b' }, c: 'c' };
const objB = _.omit(['c'], objA);
objB.a.b = 'c';
console.log(objA); // { a: { b: 'c' }, c: 'c' }
```

I was trying to ensure the return result was completely separate from the input, so modifying the return result _cannot_ modify the input.

src/js/disability-benefits/526EZ/config/form.js
cvalarida: Good point. I forgot you can set deeply nested properties like this. :man_facepalming:

src/js/common/utils/data-utils/omit.js
jbalboni: In general, we make the assumption that we're treating data as immutable and not changing it directly. There are two advantages to that. One is that an object you have isn't going to be changed by some far away function. The other benefit is that if you have two objects, you can compare them by doing `===` and `!==`, instead of having to do a deep equality check. You're trying to enforce the first benefit at the expense of the second, because now `objB.b` and `objA.b` are deeply equal, but `objB.b !== objA.b`. The lodash version of omit doesn't work this way.

We don't generally try to enforce the first advantage in this way because it means we're doing lots of cloning without gaining much. That second advantage I mentioned is one that redux-connect uses to make updates more efficient and one we use in React components with shouldComponentUpdate to make performance improvements when necessary. So losing it can have more of an impact.

If we feel the need to enforce the assumption that you can't change objects directly, then we would probably want to use a library like ImmutableJS or seamless-immutable, but those have tradeoffs.

src/js/disability-benefits/526EZ/config/form.js
cvalarida: Nah, we're going to have to change this anyhow. I'm not sure how quite yet, though...

src/js/disability-benefits/526EZ/helpers.jsx
cvalarida: Yeah, that's true. What's the convention elsewhere?

cehsu: screen shot 2018-03-09 at 4 52 58 pm

src/js/common/utils/data-utils/set.js
```diff
@@ -11,9 +11,9 @@ import checkValidPath from './checkValidPath';
  *
  * TODO: Link to some documentation which will be a better explanation than the above
  *
- * @param {Array|Obect} object
  * @param {Array|string} path
  * @param {*} value
+ * @param {Array|Obect} object
```
jbalboni: Typo

src/js/common/utils/data-utils/set.js
```diff
@@ -57,9 +57,9 @@ function baseSet(arrayPath, value, object, level = 0) {
  * Sets the value at the end of the path, creating the appropriate objects along the way if needed.
  * Separate from `baseSet` to not expose the level param.
  *
- * @param {Obect} object
  * @param {Array|string} path
  * @param {*} value
+ * @param {Obect} object
```
jbalboni: Another typo

jbalboni (approved):

src/js/common/utils/data-utils/set.js
cvalarida: Heh, these typos are just copy pasta from earlier (realized I changed the param order but never updated the documentation). :flushed:

Good eye! :male_detective:

PR #7393, opened 2018-03-08
ignore and block yalc with git

rianfowler:


hooks/pre-commit
```diff
@@ -1,3 +1,4 @@
 #!/bin/sh
 # Lint everything before commit, just in case our text editor isn't doing it for us
+yalc check
```
rianfowler: This errors the commit if there are yalc dependencies in package.json (this only happens if you try to commit package.json changes after using `yalc add` instead of `yalc link`- the design system instructions have been updated to prefer link).

jbalboni: This assumes that everyone has `yalc` installed globally, I think. I'm a little hesitant to make that required.

rianfowler: good point I'll remove it

jbalboni (approved):

PR #7395, opened 2018-03-08
Add private medical records intro

cvalarida:
image

### DIFF
--- a/src/js/disability-benefits/526EZ/config/form.js
+++ b/src/js/disability-benefits/526EZ/config/form.js
@@ -12,7 +12,8 @@ import {
   evidenceTypesDescription,
   EvidenceTypeHelp,
   disabilityNameTitle,
-  vaMedicalRecordsIntro
+  vaMedicalRecordsIntro,
+  privateMedicalRecordsIntro
 } from '../helpers';
 
 const initialData = {
@@ -203,7 +204,7 @@ const formConfig = {
         },
         vaMedicalRecordsIntro: {
           title: '',
-          path: 'supporting-evidence/:index/va-medical-records',
+          path: 'supporting-evidence/:index/va-medical-records-intro',
           showPagePerItem: true,
           arrayPath: 'disabilities',
           depends: (formData, index) => _.get(`disabilities.${index}.view:vaMedicalRecords`, formData),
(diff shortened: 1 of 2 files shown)


kreek (approved):
LGTM 🚀

src/js/disability-benefits/526EZ/config/form.js
```diff
@@ -229,7 +230,33 @@ const formConfig = {
           }
         },
         // pageFour: {},
-        // pageFive: {},
+        privateMedicalRecordsIntro: {
+          title: '',
+          path: 'supporting-evidence/:index/private-medical-records-intro',
+          showPagePerItem: true,
+          arrayPath: 'disabilities',
+          depends: (formData, index) => _.get(`disabilities.${index}.view:privateMedicalRecords`, formData),
```
cehsu: Should these properties be in `[]`?

src/js/disability-benefits/526EZ/helpers.jsx
```diff
@@ -82,3 +82,10 @@ export const vaMedicalRecordsIntro = ({ formData }) => {
     <p>Ok, first we’ll ask about your VA medical records related to your {formData.disability.diagnosticText}.</p>
   );
 };
+
+
+export const privateMedicalRecordsIntro = ({ formData }) => {
+  return (
+    <p>Ok, first we’ll ask about your private medical records related to your {formData.disability.diagnosticText}.</p>
```
cehsu: Is this content finalized or is this just temporary text?

src/js/disability-benefits/526EZ/config/form.js
```diff
@@ -229,7 +230,33 @@ const formConfig = {
...
+              items: {
+                'ui:title': disabilityNameTitle,
+                'ui:description': privateMedicalRecordsIntro,
+              }
+            }
+          },
+          schema: {
+            type: 'object',
+            properties: {
+              disabilities: {
```
cehsu: ~Should we rename one of the disabilities properties, either here or in the `privateRecordChoice` page?~
Should we create a component which like the photo cropper renders different views here rather than setting up objects without properties?

cehsu (commented):
Looks good so far. I had a few questions and minor comments.

cehsu (approved):

src/js/disability-benefits/526EZ/config/form.js
cvalarida: They could be, sure, but it works just fine like this. :man_shrugging:
I went with a string template over array for the slight readability win.

src/js/disability-benefits/526EZ/helpers.jsx
cvalarida: I don't know that the content is _finalized_ exactly, but it is what was provided in the mockup.
What I'm not sure about is whether we should be pulling from `diagnosticText` or some other property.

PR #7397, opened 2018-03-08
Move Dashboard into Personalization directory; Add redirect

ncksllvn:
I have to make adjustments for the Beta setup so for now you won't see dashboard in the dropdown; you'll have to navigate to it manually


src/js/user-profile/components/FormItem.jsx
```diff
@@ -0,0 +1,63 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import moment from 'moment';
+
+import { formTitles, formLinks, disabledForms } from '../helpers';
+
+class FormItem extends React.Component {
```
bshyong: Let's delete this; no need for saved apps on the profile page.

src/js/user-profile/components/FormList.jsx
```diff
@@ -0,0 +1,62 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+
+import ProgressButton from '../../common/components/form-elements/ProgressButton';
+import Modal from '../../common/components/Modal';
+
+import FormItem from './FormItem';
+import { isSIPEnabledForm } from '../helpers';
+
+class FormList extends React.Component {
```
bshyong: Let's delete this; no need for saved apps on the profile page.

src/js/common/components/DashboardRedirect.jsx
```diff
@@ -0,0 +1,59 @@
...
+
+class DashboardRedirect extends React.Component {
+
+  static defaultProps = {
+    services: []
+  };
+
+  componentWillMount() {
+    // Create a global event listener...
+    document.addEventListener('mousedown', event => {
```
bshyong: Can you help me understand why we need this global listener on all of the links?

If the links to the homepage are normal html and not triggered through react-router (full render), will the logic in `componentWillReceiveProps` suffice?  Are there other edge cases that we need to be aware of?

ncksllvn: The originally idea was to be able to access the Redux data that we need to determine whether or not to redirect, and do that in the most unobtrusive way. The edge case here is that if they click an index link before the user object is loaded then the global listener won't do anything, so they'll still see the regular homepage.

I just got another idea though, which is to drop this redirect on the index page itself, and cache a "beta dashboard" flag on the session to be able to do quick redirects. I'm gonna go ahead and do the merge here and make those adjustments later to keep things clean.

PR #7398, opened 2018-03-09
Remove resolve url loader to make the build faster

jbalboni:
The resolve url loader parses a source map and fixes relative imports inside a sass file. In our case, it's fixing the `../img` path for assets in uswds, since that doesn't work correctly when compiled outside of the uswds project itself. This loader appears to be very slow. The other recommended option from the sass-loader readme is to override the asset path that most sass libraries expose. Happily, uswds does this and we can set that path and remove resolve-url-loader.

This appears to take the build time on my machine from around ~130 seconds to ~40 seconds.

I found this when testing out Webpack 4, so I'm linking it to that issue.

### DIFF
--- a/src/sass/modules/_m-button.scss
+++ b/src/sass/modules/_m-button.scss
@@ -178,7 +178,7 @@ button.short {
 }
 
 .va-btn-close-icon {
-  background: transparent url(assets/img/icons/close.svg) center no-repeat;
+  background: transparent url(/img/icons/close.svg) center no-repeat;
   background-size: 1rem 1rem;
   height: 2.2rem;
   padding: 2.2rem;
--- a/config/webpack.config.js
+++ b/config/webpack.config.js
@@ -119,12 +119,7 @@ const configGenerator = (options) => {
             fallback: 'style-loader',
             use: [
               { loader: 'css-loader' },
-              {
-                loader: 'sass-loader',
-                options: {
-                  sourceMap: true,
-                }
-              }
+              { loader: 'sass-loader' }
             ],
           })
         },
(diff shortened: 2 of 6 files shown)


src/sass/modules/_m-button.scss
```diff
@@ -178,7 +178,7 @@ button.short {
 }
 
 .va-btn-close-icon {
-  background: transparent url(assets/img/icons/close.svg) center no-repeat;
+  background: transparent url(/img/icons/close.svg) center no-repeat;
```
ncksllvn: `resolve-url-loader` was fixing this path before? And this is the only case of that `asset` prefix? We were making a big sacrifice on the build time just for that lol.

ncksllvn (approved):
Awesome job! Love the optimizations you've been doing with our build.

rianfowler (approved):
Would like to know your process for profiling this- might be a good talk for the FE meeting.

src/sass/modules/_m-button.scss
jbalboni: It was mostly for the paths in the uswds sass code. It looks like this is the only place where our code needed it, though.

jbalboni:
Interestingly, this does not appear to have changed the jenkins build time much at all.

cehsu (approved):
LGTM. Great job!

cvalarida:
I'm not super familiar with all the _whys_ behind using `resolve-url-loader`, which means I'm not sure how to tell if this breaks anything. It looks good to me, but how can we tell if anything is broken? (Seems like it'd be obvious, but :man_shrugging: )

jbalboni:
@cvalarida It rewrites relative paths in sass files, and the build breaks if those paths aren't correct. So I believe it's enough to see that the build passed. I also tried to look around and make sure there were no broken images on the site, but that's not comprehensive, obviously.

PR #7399, opened 2018-03-09
remove querySelectors and use refs

rianfowler:
Think I got them all.


jbalboni:
Should we use the function syntax instead of the string method? https://reactjs.org/docs/refs-and-the-dom.html#legacy-api-string-refs

rianfowler:
We should use the callback- I hadn't seen the string ref before working on this project and thought it was new.

I usually store refs as an instance variable on the component. I think we could also store them on the state.

jbalboni:
Instance variable sound good to me.

rianfowler:
Updated to use ref callbacks.

cehsu (approved):
LGTM. Is there an issue that should be linked to this?

jbalboni (approved):

PR #7403, opened 2018-03-09
Fix security bug in the next parameter redirect

ncksllvn:
I have been looking at how that `?next=` parameter works and noticed that you can put any URL there and after login the user will be redirected there. For example -

https://dev.vets.gov/?next=https%3A%2F%2Fsecurity.stackexchange.com%2Fquestions%2F121643%2Fis-it-safe-to-redirect-to-url-parameter-without-filtering

I haven't tested this in production so maybe there is a patch there. But otherwise, if someone realized this they could drop a malicious URL as that parameter and send that link out. Then any veterans will be redirected there post-login, or immediately if they're already logged in.


jbalboni (approved):

bshyong:
LGTM!

cvalarida:
Yipes! Good catch, @ncksllvn! :male_detective:

PR #7404, opened 2018-03-09
Add URL to failed builds consolidate failure messages in one place

edpaget:
Use a post failure step instead of notifying in a try catch. Include a link to the Blue Ocean URL.


edpaget:
This at least includes a link to the failed build on Jenkins so we can see what happened. Still working on seeing if its possible to grab test failures.

edpaget:
Was able to add junit reporting of the unit tests, so can now do some fancier reports around failed tests in the future.

wyattwalter (approved):

Jenkinsfile
```diff
@@ -42,12 +42,13 @@ def buildDetails = { vars ->
   """.stripIndent()
 }
 
-def notify = { message, color='good' ->
+def notify = { ->
   if (env.BRANCH_NAME == devBranch ||
       env.BRANCH_NAME == stagingBranch ||
       env.BRANCH_NAME == prodBranch) {
+    message = "vets-website ${env.BRANCH_NAME} branch CI failed. |${env.RUN_DISPLAY_URL}".stripMargin()
```
wyattwalter: neato, did not know about RUN_DISPLAY_URL

edpaget: Yeah it links the Blue Ocean UI which I think is nicer for the vets-website build. Turns out the Jenkins docs around it are still not great.

PR #7405, opened 2018-03-09
Personalization dev beta cleanup

ncksllvn:
- Adjusts beta flow to work within `profile.services`
    - Once API is finished we will need to remove the session storage stuff
- Wraps dashboard in beta tag
- Moves dashboard redirect into personalization folder
- Establishes index page as a React entry point so that the dashboard can take precedence 
    - This could be further optimized. See render method of DashboardRedirect

Also, just a heads up that you'll have to manually navigate to `/dashboard-beta` in order to access it at first. It will save a flag in the session storage.


src/js/personalization/dashboard-beta/redirect-entry.jsx
```diff
@@ -0,0 +1,24 @@
...
+import initCommon from '../../common/init-common';
+
+const commonStore = initCommon();
+
+ReactDOM.render((
+  <Provider store={commonStore}>
+    <DashboardRedirect/>
+  </Provider>
+), document.getElementById('react-dashboard-renderer'));
+
```
ncksllvn: This whole thing could be substantially improved

ncksllvn:
I feel very meh about this code. I'm going to fix some things an reopen later.

PR #7407, opened 2018-03-12
Add rate limiting errors to forms

jbalboni:
screen shot 2018-03-12 at 5 35 33 pm
screen shot 2018-03-12 at 5 34 10 pm
screen shot 2018-03-12 at 5 33 33 pm

### DIFF
--- a/src/js/common/schemaform/actions.js
+++ b/src/js/common/schemaform/actions.js
@@ -1,7 +1,9 @@
 import Raven from 'raven-js';
+import moment from 'moment';
 import _ from '../utils/data-utils';
 import { transformForSubmit } from './helpers';
 import environment from '../helpers/environment.js';
+import { dateDiffDesc } from '../utils/helpers';
 
 export const SET_EDIT_MODE = 'SET_EDIT_MODE';
 export const SET_DATA = 'SET_DATA';
@@ -25,11 +27,12 @@ export function setEditMode(page, edit, index = null) {
   };
 }
 
-export function setSubmission(field, value) {
+export function setSubmission(field, value, extra = null) {
   return {
     type: SET_SUBMISSION,
     field,
-    value
+    value,
+    extra
   };
 }
 
@@ -61,7 +64,13 @@ function submitToUrl(body, submitUrl, trackingPrefix) {
         const results = JSON.parse(responseBody);
         resolve(results);
       } else {
-        const error = new Error(`vets_server_error: ${req.statusText}`);
+        let error;
+        if (req.status === 429) {
+          error = new Error(`vets_throttled_error: ${req.statusText}`);
+          error.extra = parseInt(req.getResponseHeader('x-ratelimit-reset'), 10);
+        } else {
+          error = new Error(`vets_server_error: ${req.statusText}`);
+        }
         error.statusText = req.statusText;
         reject(error);
       }
@@ -98,16 +107,16 @@ function submitToUrl(body, submitUrl, trackingPrefix) {
 }
 
 export function submitForm(formConfig, form) {
-  const captureError = (error, clientError) => {
+  const captureError = (error, errorType) => {
     Raven.captureException(error, {
       fingerprint: [formConfig.trackingPrefix, error.message],
       extra: {
-        clientError,
+        errorType,
         statusText: error.statusText
       }
     });
     window.dataLayer.push({
-      event: `${formConfig.trackingPrefix}-submission-failed${clientError ? '-client' : ''}`,
+      event: `${formConfig.trackingPrefix}-submission-failed${errorType.startsWith('client') ? '-client' : ''}`,
     });
   };
 
@@ -133,9 +142,14 @@ export function submitForm(formConfig, form) {
       .catch(error => {
         // overly cautious
         const errorMessage = _.get('message', error);
-        const clientError = errorMessage && !errorMessage.startsWith('vets_server_error');
-        captureError(error, clientError);
-        dispatch(setSubmission('status', clientError ? 'clientError' : 'error'));
+        let errorType = 'clientError';
+        if (errorMessage.startsWith('vets_throttled_error')) {
+          errorType = 'throttledError';
+        } else if (errorMessage.startsWith('vets_server_error')) {
+          errorType = 'error';
+        }
+        captureError(error, errorType);
+        dispatch(setSubmission('status', errorType, error.extra));
       });
   };
 }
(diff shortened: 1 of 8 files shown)


src/js/common/utils/helpers.js
```diff
@@ -210,7 +210,13 @@ export function dateDiffDesc(date, userFromDate = null) {
     return formatDiff(minuteDiff, 'minute');
   }
 
-  return 'less than a minute';
+  const secondDiff = date.diff(fromDate, 'seconds');
```
jbalboni: This function isn't used by anything at the moment so changing it doesn't break anything.

rianfowler: is this `dateDiffDescription` or `dateDiffDescending`?

src/js/common/schemaform/review/SubmitButtons.jsx
```diff
@@ -49,6 +51,21 @@ export default function SubmitButtons(props) {
         </div>
       </div>
     );
+  } else if (submission.status === 'throttledError') {
```
rianfowler: The `error.type` gets translated into `submission.status`. I guess `action.field` comes in as `status` in the 1SET_SUBMISSION` reducer?

src/js/common/schemaform/actions.js
```diff
@@ -25,11 +27,12 @@ export function setEditMode(page, edit, index = null) {
   };
 }
 
-export function setSubmission(field, value) {
+export function setSubmission(field, value, extra = null) {
```
cvalarida: This is where Flow would come in handy. What's `extra` for? Could you add some comments explaining why we might want `extra`?

I can see in this case we're using it to record the `x-ratelimit-reset` response header (right?), but it's potentially useful for anything we want to record from the api response, yeah?

src/js/common/schemaform/review/SubmitButtons.jsx
```diff
@@ -49,6 +51,21 @@ export default function SubmitButtons(props) {
...
+    submitButton = (
+      <ProgressButton
+        onButtonClick={onSubmit}
+        buttonText="Submit Application"
+        buttonClass="usa-button-primary"/>
```
cvalarida: Does it make sense to disable this button until they can submit again? I think we generally try to shy away from disabling buttons, but I figured I'd ask.

cvalarida (approved):

src/js/common/utils/helpers.js
jbalboni: I changed the name in the code, hopefully it's more clear now.

src/js/common/schemaform/review/SubmitButtons.jsx
jbalboni: Yeah, I'd like to refactor this. We have a Redux action that's written as a sort of generic setter action for data on the submission state. We should have actions that conceptually match what a user is doing.

src/js/common/schemaform/review/SubmitButtons.jsx
jbalboni: I'm hesitant to add that for something that will probably be seen by very few people. It would be nice, though.

src/js/common/schemaform/actions.js
jbalboni: Yeah, I needed a way to pass info from the submission error into the state. Not a big fan of this, but couldn't think of a better way of doing it.

rianfowler (approved):
Looks good

peggygannon (commented):
@jbalboni Thanks, Jeff. I have just one tweak: 
--For the digital photo message, I think we should remove: **We've run into a problem** . So that message will be consistent with the discharge document message.

src/js/common/schemaform/actions.js
```diff
@@ -133,9 +144,14 @@ export function submitForm(formConfig, form) {
...
         // overly cautious
         const errorMessage = _.get('message', error);
-        const clientError = errorMessage && !errorMessage.startsWith('vets_server_error');
-        captureError(error, clientError);
-        dispatch(setSubmission('status', clientError ? 'clientError' : 'error'));
+        let errorType = 'clientError';
+        if (errorMessage.startsWith('vets_throttled_error')) {
+          errorType = 'throttledError';
+        } else if (errorMessage.startsWith('vets_server_error')) {
+          errorType = 'error';
```
cehsu: Since we're specifying clientError and throttledError, would it make sense to do the same for error (i.e. serverError)?

src/js/common/schemaform/actions.js
```diff
@@ -190,9 +206,14 @@ export function uploadFile(file, uiOptions, onProgress, onChange, onError) {
         const fileData = uiOptions.parseResponse(JSON.parse(body), file);
         onChange(fileData);
       } else {
+        let errorMessage = req.statusText;
+        if (req.status === 429) {
+          errorMessage = `You’ve reached the limit for the number of submissions we can accept at this time. Please try again in ${timeFromNow(moment.unix(parseInt(req.getResponseHeader('x-ratelimit-reset'), 10)))}.`;
```
cehsu: Do we typically provide seconds when instructing users to repeat actions, do we expect them to measure the time they wait in seconds? Could minutes potentially be more practical for how they make use of the message? Looping in @peggygannon .

src/js/common/schemaform/review/SubmitButtons.jsx
```diff
@@ -49,6 +51,21 @@ export default function SubmitButtons(props) {
...
+    );
+    submitMessage = (
+      <div className="usa-alert usa-alert-error schemaform-failure-alert">
+        <div className="usa-alert-body">
+          <p className="schemaform-warning-header"><strong>We’ve run into a problem</strong></p>
+          <p>We’re sorry. Your submission didn’t go through because we received too many requests from you. Please wait {timeFromNow(moment.unix(submission.extra))} and submit your request again.</p>
```
cehsu: Should this match the message above?
"You’ve reached the limit for the number of submissions we can accept at this time. Please try again in..."
Or vice versa?

cehsu (approved):
LGTM. I had a few minor content questions.

jbalboni:
@peggygannon Currently that "We've run into a problem" language is prepended to all errors on the upload photo page. Should we change that?

src/js/common/schemaform/actions.js
jbalboni: The message will display the largest unit it can for the message. If it's over 60 seconds, it'll use minutes, if it's over 60 minutes it'll say hours, etc.

src/js/common/schemaform/actions.js
jbalboni: That's probably more clear now, yeah.

src/js/common/schemaform/review/SubmitButtons.jsx
jbalboni: They're different messages in the linked issue, though @peggygannon  can confirm.

peggygannon:
@jbalboni Ok, thanks for letting me know. Let's not change anything then and leave the We've run into a problem language.

PR #7408, opened 2018-03-12
Upgrade to Webpack 4

jbalboni:
This upgrades to Webpack 4 (from 2.x) and cleans up some build and polyfill configuration. There are not a lot of major differences. Bundles are very slightly smaller due to module concatenation. Dev builds do not appear to be much faster (maybe a few seconds). The production build appears to be 30-45 seconds faster on my machine, I think due to the cache and parallel options that are now used with Uglify. But it appears slower on Jenkins.

TODO
- [x] Figure out why export not found errors no longer show up
  - I'm just being dumb
- [x] Verify that Modernizr is not used anywhere else
  - It is used, so I added it back
- [x] Investigate why sourcemaps are now half as big
  - I can't find anything through Googling and they work, so I'm leaving it alone
- [x] Investigate why Jenkins build appears slower
  - Cache option makes it slower, but it's still slower without that. Parallel option speeds it up some.


config/webpack.config.js
```diff
@@ -65,6 +62,7 @@ const configGenerator = (options) => {
     'raven-js'
   ];
   const baseConfig = {
+    mode: 'development',
```
jbalboni: New in v4: this switches between lots of production/development options.

config/webpack.config.js
```diff
@@ -93,40 +90,31 @@ const configGenerator = (options) => {
...
             use: [
               { loader: 'css-loader' },
               { loader: 'sass-loader' }
-            ],
+            ]
           })
         },
         {
+          // if we want to minify these images, we could add img-loader
+          // but it currently only would apply to three images from uswds
```
jbalboni: There used to be an img-loader here, but it was never actually running. It doesn't seem necessary, given that it would only apply to 3 images if it did.

config/webpack.config.js
```diff
@@ -171,17 +153,38 @@ const configGenerator = (options) => {
       noParse: [/mapbox\/vendor\/promise.js$/],
     },
     resolve: {
-      alias: {
-        modernizr$: path.resolve(__dirname, './modernizrrc')
-      },
       extensions: ['.js', '.jsx']
     },
+    optimization: {
```
jbalboni: This is the v4 replacement for adding commons chunk and uglify to the plugins list.

config/webpack.config.js
```diff
@@ -171,17 +153,38 @@ const configGenerator = (options) => {
...
+          }
+        }
+      }
+    },
     plugins: [
       new webpack.DefinePlugin({
         __BUILDTYPE__: JSON.stringify(options.buildtype),
         __SAMPLE_ENABLED__: (process.env.SAMPLE_ENABLED === 'true'),
         'process.env': {
-          NODE_ENV: JSON.stringify(process.env.NODE_ENV || 'development'),
```
jbalboni: This is done automatically based on the mode now.

config/webpack.config.js
```diff
@@ -207,44 +204,17 @@ const configGenerator = (options) => {
     if (options.buildtype === 'production') {
       sourceMap = 'https://s3-us-gov-west-1.amazonaws.com/www.vets.gov';
     }
+
     baseConfig.plugins.push(new webpack.SourceMapDevToolPlugin({
       append: `\n//# sourceMappingURL=${sourceMap}/generated/[url]`,
       filename: '[file].map',
     }));
-    baseConfig.module.rules.push({
```
jbalboni: We don't use these modules anymore.

config/webpack.config.js
```diff
@@ -207,44 +204,17 @@ const configGenerator = (options) => {
...
-    });
-    baseConfig.module.rules.push({
-      test: /debug\/RoutesDropdown/,
-      use: {
-        loader: 'null-loader'
-      }
-    });
 
-    baseConfig.plugins.push(new WebpackMd5Hash());
+    baseConfig.plugins.push(new webpack.HashedModuleIdsPlugin());
```
jbalboni: This appears to be the recommended way of having consistent module hashes.

config/webpack.config.js
```diff
@@ -207,44 +204,17 @@ const configGenerator = (options) => {
...
     baseConfig.plugins.push(new ManifestPlugin({
       fileName: 'file-manifest.json'
     }));
-    baseConfig.plugins.push(new ChunkManifestPlugin({
```
jbalboni: This isn't necessary now since the runtime is not inserted into the vendor chunk.

content/includes/header.html
```diff
@@ -81,13 +81,7 @@
...
 <script src="/js/settings.js"></script>
 
-<script>
-//<![CDATA[
-window.webpackManifest = 'CHUNK_MANIFEST_PLACEHOLDER';
-//]]>
-</script>
-
-<script defer src="/generated/vendor.js"></script>
+<script defer src="/generated/vendor.entry.js"></script>
```
jbalboni: The file name changed to include `.entry`

src/js/common/polyfills.js
```diff
@@ -8,17 +8,9 @@
 import 'babel-polyfill';
 
 // Basic polyfills.
-// TODO(awong): These do NOT correctly conditionally load the polyfill.
-// The polyfill is always loaded. require.ensure() should be used instead but
-// then load ordering needs to be worked out. Fix later.
-import Modernizr from 'modernizr';
```
jbalboni: Modernizr is not necessary, as far as I can tell. I need to do a little more research to be sure.

src/js/common/polyfills.js
```diff
@@ -32,9 +24,6 @@ if (navigator.userAgent.includes('Edge/14')) {
 // above code
 require('whatwg-fetch');
 
-// This polyfill has its own test logic so no need to conditionally require.
-import 'polyfill-function-prototype-bind';
```
jbalboni: This was for phantomjs, but now all the environments we run in have `bind`.

config/webpack.config.js
```diff
@@ -93,40 +90,31 @@ const configGenerator = (options) => {
           use: {
             loader: 'babel-loader',
             options: {
-              presets: ['react'],
```
jbalboni: This is not used.

package.json
```diff
@@ -99,21 +98,18 @@
     "exports-loader": "^0.6.3",
     "express": "^4.14.0",
     "express-history-api-fallback": "^2.0.0",
-    "extract-text-webpack-plugin": "^2.1.0",
-    "file-loader": "^0.9.0",
+    "extract-text-webpack-plugin": "next",
```
jbalboni: Somewhat concerned about this, but the plugin appears to be working.

jbalboni:
FYI @wyattwalter @edpaget, I'm seeing this increase the build times some, though I'm not sure why. It's slightly faster for me locally.

script/build.js
```diff
@@ -638,7 +641,7 @@ if (options.buildtype !== 'development') {
     const manifest = {};
     Object.keys(originalManifest).forEach((originalManifestKey) => {
       const matchData = originalManifestKey.match(/(.*)\.js$/);
-      if (matchData !== null && matchData[1] !== 'vendor') {
```
cvalarida: Why remove the vendor check here?

script/build.js
```diff
@@ -652,7 +655,7 @@ if (options.buildtype !== 'development') {
     Object.keys(files).forEach((filename) => {
       if (filename.match(/\.(html|css)$/) !== null) {
         Object.keys(manifest).forEach((originalAssetFilename) => {
-          const newAssetFilename = manifest[originalAssetFilename];
+          const newAssetFilename = manifest[originalAssetFilename].replace('/generated/', '');
```
cvalarida: I wasn't ever super sure of what the deal was with `generated/`--why take it out of the file (path) name?

script/metalsmith-webpack.js
```diff
@@ -0,0 +1,99 @@
+/* eslint-disable */
```
cvalarida: Should we worry about linting this since it'll be living in our code base for...the foreseeable future?

cvalarida (commented):
I haven't run it yet to try it out, but it looks good.

Do we call it golden if it just builds and runs, or do we have stricter acceptance criteria on this upgrade?

script/build.js
jbalboni: Before the vendor file followed a different naming convention and lacked `.entry` in the file name.

script/build.js
jbalboni: It needs to be removed to match another name in the manifest (I think, I've already forgotten part of this, I need to write a comment explaining it)

script/metalsmith-webpack.js
jbalboni: Yeah, I probably should just update this section.

script/metalsmith-webpack.js
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  return process
+}
+
+module.exports = {
+  webpackPlugin,
+  webpackDevServerPlugin
+};
+
```
ncksllvn: This is what enables hot reload and ultimately plugs Metalsmith into the Webpack dev server?

package.json
```diff
@@ -140,13 +136,9 @@
     "metalsmith-redirect": "^2.1.0",
     "metalsmith-sitemap": "^1.0.0",
     "metalsmith-watch": "^1.0.3",
-    "metalsmith-webpack": "department-of-veterans-affairs/metalsmith-webpack",
-    "metalsmith-webpack-dev-server": "department-of-veterans-affairs/metalsmith-webpack-dev-server",
     "mkdirp": "^0.5.1",
```
ncksllvn: Actually never realized til now that we maintained the Metalsmith plugin for Webpack dev server.

ncksllvn (approved):
Awesome job. Running it locally and it's working great. 

```
[metalsmith-watch] ✓ Live reload server started on port: 35729
[metalsmith-webpack-dev-server]: Running webpack dev server at http://localhost:3001
[metalsmith-watch] ✔︎ Watching ../content/**/*
Metalsmith build finished!  Starting webpack-dev-server...
ℹ ｢wdm｣: Time: 25150ms
Built at: 3/15/2018 9:51:55 AM
ℹ ｢wdm｣: Compiled successfully.
```

package.json
jbalboni: Actually, I didn't either. @bshyong, do you think we should continue updating those repositories or move the code into vets-website like I did here?

bshyong: @jbalboni I believe we forked these because I made some changes to those libraries to work with vets-website (will have to diff to see what the changes were).

I think it’s fine to move them into our repo instead of using a fork.

jbalboni: It looks like the only change was to update the webpack dependency version.

script/metalsmith-webpack.js
jbalboni: Yep. This is the code that connects the metalsmith and webpack build processes.

PR #7409, opened 2018-03-12
Personalization - Implement Beta API route; homepage redirect to dashboard

ncksllvn:
- The beta stuff is now hooked up to the API
- Adds a component that handles redirects from the index page to the dashboard. It does this by intercepting links going to the index page. It is attached conditionally, meaning that it is only rendered for users who are logged in and enrolled in the beta. It is powered by the services property in the user object. This means that if you reload the page, then click a link before the user object is retrieved, then you will not be redirected. This is an acceptable edge case IMO since once we're out of beta we won't have to check for the beta flag.
- ~~For now, I switched to a hackish method of redirecting from the index page to the dashboard - see the bottom <script> element on that index.md. The thing is that we currently need an immediate method of checking whether or not a user is enrolled in the beta to redirect from the homepage to the dashboard, because sometimes that /v0/user route takes awhile, and we need a redirect to be fast. For testing purposes, I added a flag into localStorage that enables beta features for the site which we can use to check immediately~~


ncksllvn:
@bshyong I actually feel great about this current solution. Once we go from beta to live, we may need to revise the redirect handler, but otherwise I think it's in great shape. No sessionStorage or localStorage necessary. You just need to enroll the user in the beta by visiting `http://localhost:3001/dashboard-beta/` or the profile.

PR #7410, opened 2018-03-13
9175 refactor width layout

rianfowler:
Will need to get to the css changes but needed to wrap up the layout tracking first. Had a couple of questions below.

### DIFF
--- a/src/js/vic-v2/components/CropperController.jsx
+++ b/src/js/vic-v2/components/CropperController.jsx
@@ -126,15 +126,14 @@ export default class CropperControls extends React.Component {
   constructor(props) {
     super(props);
 
+    const windowWidth = window.innerWidth;
     this.state = {
+      smallCropBox: props.forceNarrowLayout || windowWidth < LARGE_SCREEN,
+      windowWidth,
       zoomMin: 0.2,
       zoomMax: 1.7,
       zoomValue: 0
     };
-  }
-
-  componentWillMount() {
-    this.detectWidth();
   }
 
   componentDidMount() {
@@ -345,7 +344,7 @@ export default class CropperControls extends React.Component {
     const windowWidth = window.innerWidth;
     this.setState({
       windowWidth,
-      smallCropBox: windowWidth < LARGE_SCREEN
+      smallCropBox: this.props.forceNarrowLayout || windowWidth < LARGE_SCREEN
     });
   }
 
@@ -421,11 +420,19 @@ export default class CropperControls extends React.Component {
   // Cropper doesn't support changing width after componentDidMount
   // use the windowWidth as a key to for the component to remount when the window width changes
   render() {
+
+    // narrow-layout class controls visiblity of elements based on window width
+    // if layout is forced narrow, remove this behavior
+    const narrowLayoutButtonsClassNames = classNames('cropper-control-row', { 'narrow-layout': !this.props.forceNarrowLayout });
+
+    // prevent the cropper from re-mounting when forcing narrow layout and width is in large screen range
+    const cropperKey = this.props.forceNarrowLayout && this.state.windowWidth >= LARGE_SCREEN ? LARGE_SCREEN : this.state.windowWidth;
+
     return (
       <div className="cropper-container-outer">
         <Cropper
           ref="cropper"
-          key={this.state.windowWidth}
+          key={cropperKey}
           ready={this.setCropBox}
           responsive
           src={this.props.src}
@@ -442,7 +449,11 @@ export default class CropperControls extends React.Component {
           zoom={this.onZoom}/>
         <div className="cropper-zoom-container">
           <button className="cropper-control cropper-control-zoom cropper-control-zoom-out va-button va-button-link" type="button" onClick={this.onZoomOut}>
-            <span className="cropper-control-label"><span className="normal-layout">Make smaller</span><i className="fa fa-search-minus"></i></span>
+            <span className="cropper-control-label">
+              {this.props.forceNarrowLayout || <span className="normal-layout">Make smaller</span>}
+              <i className="fa fa-search-minus">
+              </i>
+            </span>
           </button>
           <input type="range"
             className="cropper-zoom-slider"
@@ -458,11 +469,15 @@ export default class CropperControls extends React.Component {
             onChange={this.onSliderChange}
             value={this.state.zoomValue}/>
           <button className="cropper-control cropper-control-zoom cropper-control-zoom-in va-button va-button-link" type="button" onClick={this.onZoomIn}>
-            <span className="cropper-control-label"><span className="normal-layout">Make larger</span><i className="fa fa-search-plus"></i></span>
+            <span className="cropper-control-label">
+              {this.props.forceNarrowLayout || <span className="normal-layout">Make larger</span>}
+              <i className="fa fa-search-plus">
+              </i>
+            </span>
           </button>
         </div>
         <div className="cropper-control-container">
-          <div className="cropper-control-row narrow-layout">
+          <div className={narrowLayoutButtonsClassNames}>
             <button className="cropper-control cropper-control-label-container va-button va-button-link" type="button" onClick={this.onZoomOut}>
               <span className="cropper-control-label">Make smaller</span>
             </button>
@@ -497,6 +512,7 @@ export default class CropperControls extends React.Component {
 }
 
 CropperControls.PropTypes = {
+  forceNarrowLayout: React.PropTypes.bool,
   onPhotoCropped: React.PropTypes.func.isRequired,
   src: React.PropTypes.string.isRequired
 };
--- a/src/js/vic-v2/components/CropperController.jsx
+++ b/src/js/vic-v2/components/CropperController.jsx
@@ -3,10 +3,6 @@ import React from 'react';
 import Cropper from 'react-cropper';
 import classNames from 'classnames';
 
-/*
- * get rid of the small screen check on photofield
- *
- */
 const MIN_SIZE = 350;
 const SMALL_CROP_BOX_SIZE = 240;
 const LARGE_CROP_BOX_SIZE = 300;
--- a/src/js/vic-v2/components/CropperController.jsx
+++ b/src/js/vic-v2/components/CropperController.jsx
@@ -131,13 +131,30 @@ export default class CropperControls extends React.Component {
     };
   }
 
-  componentWillUpdate(nextProps) {
-    if (nextProps.windowWidth !== this.props.windowWidth) {
+  componentWillUpdate(nextProps, nextState) {
+    if (nextState.windowWidth !== this.state.windowWidth) {
       const cropper = this.refs.cropper;
       if (cropper) {
         this.setCropBox();
       }
     }
+  }
+
+  componentWillMount() {
+    this.detectWidth();
+  }
+
+  componentDidMount() {
+    window.addEventListener('resize', this.detectWidth);
+  }
+
+  componentWillUnmount() {
+    window.removeEventListener('resize', this.detectWidth);
+  }
+
+  detectWidth = () => {
+    const windowWidth = window.innerWidth;
+    this.setState({ windowWidth });
   }
 
   onCropstart = (e) => {
@@ -314,9 +331,11 @@ export default class CropperControls extends React.Component {
 
   // initialize cropbox
   setCropBox = () => {
+    console.log('setCropBox');
     // use container and cropbox constants to set cropbox size
+    const smallScreen = window.innerWidth < 1201;
     const containerWidth = this.refs.cropper.getContainerData().width;
-    const heightWidth = this.props.narrowLayout ? SMALL_CROP_BOX_SIZE : LARGE_CROP_BOX_SIZE;
+    const heightWidth = smallScreen ? SMALL_CROP_BOX_SIZE : LARGE_CROP_BOX_SIZE;
     const left = (containerWidth / 2) - (heightWidth / 2);
     this.refs.cropper.setCropBoxData({
       top: 0,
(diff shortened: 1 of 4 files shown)


src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -130,19 +136,24 @@ export default class PhotoField extends React.Component {
...
                 file,
                 this.props.uiSchema['ui:options'],
                 this.updateProgress,
-                (fileData) => {
-                  this.props.onChange(fileData);
-                },
+                (formData) => this.handleUploadComplete(formData, file),
                 () => {
                   this.props.onBlur(this.props.idSchema.$id);
                   this.uploadRequest = null;
```
rianfowler: What is this callback for? It's on both of the `onChange` handlers- I'm not sure how to update the layout state for this scenario.

src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -130,19 +136,24 @@ export default class PhotoField extends React.Component {
...
+                warningMessage: null
+              });
               this.uploadRequest = this.props.formContext.uploadFile(
                 file,
                 this.props.uiSchema['ui:options'],
                 this.updateProgress,
-                (fileData) => {
-                  this.props.onChange(fileData);
-                },
+                (formData) => this.handleUploadComplete(formData, file),
```
rianfowler: Screenreader handler for upload complete got out of sync with the `onChange` handler. Want to make sure that both of them should handle the upload complete the same way- I noticed an issue where the screen reader upload wasn't showing the preview image after uploading it.

src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -11,6 +11,14 @@ import CropperController from '../components/CropperController';
 const LARGE_SCREEN = 1201;
 const MIN_SIZE = 350;
 
+const layouts = {
+  choosePhoto: 1,
+  cropPhoto: 2,
+  watchUpload: 3,
+  previewPhoto: 4,
+  screenReaderError: 5
```
rianfowler: Except for the error, the layouts are named as actions from the user's perspective.

src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -289,27 +312,33 @@ export default class PhotoField extends React.Component {
...
+        fieldView = 'progress';
+        break;
+      case layouts.previewPhoto:
+        fieldView = 'preview';
+        break;
+      case layouts.screenReaderError:
+        fieldView = 'error';
+        break;
+      default:
+        fieldView = 'initial';
```
rianfowler: Will probably clean up the JSX a bit but kept using `fieldView` for now.

src/js/vic-v2/components/PhotoField.jsx
jbalboni: This for when an error happens. I don't think you will need to do anything in here.

src/js/vic-v2/components/PhotoField.jsx
jbalboni: Yeah, they should be the same.

src/sass/vic-v2.scss
```diff
@@ -287,6 +281,19 @@ input[type=range] {
   }
 }
 
+.narrow-layout {
+  @media (min-width: $large-screen) {
+    display: none;
+  }
+}
+
+.normal-layout {
```
rianfowler: Using these for determining if elements should be visible.

src/js/vic-v2/components/CropperController.jsx
```diff
@@ -131,13 +137,19 @@ export default class CropperControls extends React.Component {
...
-      }
-    }
+  componentWillMount() {
+    this.detectWidth();
+  }
+
+  componentDidMount() {
+    // updating the width state triggers Cropper remount- debounce event to reduce jank during active resize
+    this.debouncedDetectWidth = this.debouncedDetectWidth || _.debounce(this.detectWidth, 250);
+    window.addEventListener('resize', this.debouncedDetectWidth);
```
rianfowler: Couldn't completely get away from detecting the windowWidth but now it's _only_ used to manage remounting the Cropper- it doesn't support changing the width of the canvas once its mounted.

Cropper uses `windowWidth` as a key to force the component to remount and redraw the canvas when the width changes. The only use case for the window width changing is orientation change on a mobile device or if someone is actively dragging a desktop browser. This debounces the event so that typically happens just once- after the drag or orientation change is complete. 

Tried a few timings- anything less than 200ms was too bouncy.

src/js/vic-v2/components/CropperController.jsx
```diff
@@ -131,13 +137,19 @@ export default class CropperControls extends React.Component {
     };
   }
 
-  componentWillUpdate(nextProps) {
-    if (nextProps.windowWidth !== this.props.windowWidth) {
-      const cropper = this.refs.cropper;
-      if (cropper) {
-        this.setCropBox();
```
rianfowler: this is triggered as part of remounting the component so it's not longer necessary to trigger it here.

src/js/vic-v2/components/CropperController.jsx
```diff
@@ -416,15 +438,15 @@ export default class CropperControls extends React.Component {
           cropstart={this.onCropstart}
           cropend={this.onCropend}
           cropmove={() => true}
-          minContainerHeight={this.props.narrowLayout ? SMALL_CROP_BOX_SIZE : LARGE_CROP_BOX_SIZE}
+          minContainerHeight={this.state.smallCropBox ? SMALL_CROP_BOX_SIZE : LARGE_CROP_BOX_SIZE}
```
rianfowler: Since this is an aspect of the canvas, it has to be managed with js instead of css.

src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -289,66 +292,46 @@ export default class PhotoField extends React.Component {
...
+      case layouts.previewPhoto:
+        description = <div>Success! This photo will be printed on your Veteran ID Card.</div>;
+        uploadControlClass = classNames('photo-input-container', 'photo-input-container-left');
+        break;
+      case layouts.screenReaderError:
+        uploadControlClass = classNames('photo-input-container', 'photo-input-container-left');
+        break;
+      default:
+        currentLayout = layouts.choosePhoto;
+        description = <p>Drag and drop your image into the square or click the upload button.</p>;
```
rianfowler: This switch statement consolidates several instance variables and if statements.

rianfowler:
@cehsu @jbalboni I consolidated several things and used css for managing the visibility of elements. I think this refactor is good but I can make additional changes if requested.

src/sass/vic-v2.scss
cehsu: These names make sense. Also it's good to know that @media works this way, rather than according to window width.

src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -8,16 +8,15 @@ import { scrollAndFocus } from '../../common/utils/helpers';
...
 import CropperController from '../components/CropperController';
 
-const LARGE_SCREEN = 1201;
 const MIN_SIZE = 350;
 
-function isSmallScreen(width) {
-  return  width < LARGE_SCREEN;
-}
-
-function onReviewPage(pageTitle) {
```
cehsu: This was used to properly style the cropper on the review page, we will now want to make sure we detect the parent component's width (potentially via an initial empty render used to capture this information), rather than the window width.

src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -8,16 +8,15 @@ import { scrollAndFocus } from '../../common/utils/helpers';
...
-  return pageTitle === 'Photo review';
-}
+const layouts = {
+  choosePhoto: 1,
```
cehsu: It makes sense, given the sequential nature of this flow, that these should have numeric values, however is there any interest in using strings instead ('choose_photo') similar to our redux constants for consistency sake, if that applies here?

cehsu (approved):
Looks good so far. There are lots of great changes here. I had one comment about how we should handle the cropper sizing on the review page.

src/js/vic-v2/components/CropperController.jsx
```diff
@@ -131,13 +133,19 @@ export default class CropperControls extends React.Component {
...
   }
 
-  componentWillUpdate(nextProps) {
-    if (nextProps.windowWidth !== this.props.windowWidth) {
-      const cropper = this.refs.cropper;
-      if (cropper) {
-        this.setCropBox();
-      }
-    }
+  componentWillMount() {
```
jbalboni: It looks like using this method isn't recommended and we could put this in the constructor

https://reactjs.org/docs/react-component.html#componentwillmount

src/js/vic-v2/components/PhotoField.jsx
```diff
@@ -289,66 +292,46 @@ export default class PhotoField extends React.Component {
...
-    }
-
     let description;
-    if (fieldView === 'cropper') {
-      description = <p>Move and resize your photo, so your head and shoulders fit in the square frame below. Click and drag, or use the arrow and magnifying buttons to help.</p>;
-    } else if (fieldView === 'preview') {
-      description = <div>Success! This photo will be printed on your Veteran ID Card.</div>;
-    } else if (fieldView === 'initial' && this.state.dragAndDropSupported) {
-      description = <p>Drag and drop your image into the square or click the upload button.</p>;
+    let uploadControlClass = classNames('photo-input-container');
```
jbalboni: Do we need to use classNames here now?

src/js/vic-v2/components/CropperController.jsx
```diff
@@ -1,10 +1,12 @@
+import _ from 'lodash';
 import React from 'react';
 import Cropper from 'react-cropper';
 import classNames from 'classnames';
 
 const MIN_SIZE = 350;
 const SMALL_CROP_BOX_SIZE = 240;
 const LARGE_CROP_BOX_SIZE = 300;
+const LARGE_SCREEN = 1201;
```
jbalboni: I'm seeing an issue where resizing the screen above this size doesn't correctly set the crop box size.

src/js/vic-v2/components/CropperController.jsx
rianfowler: Good catch

src/js/vic-v2/components/PhotoField.jsx
rianfowler: Will remove

src/js/vic-v2/components/CropperController.jsx
rianfowler: resizing

This is what I'm seeing locally. Remounting the canvas is debounced so it won't be smooth as the CSS driven changes. I could force it to re-mount when it crosses that threshold.

jbalboni: What I was seeing was that if I resized while staying above 1200 (so like 1205->1300 and vice versa) it was not resetting correctly.

src/js/vic-v2/components/PhotoField.jsx
rianfowler: I understand now. I will have to pass a prop into the component to stop the `detectWidth` from firing and use that to control the markup if it's passed in.

src/js/vic-v2/components/CropperController.jsx
rianfowler: I recreated this issue on the review page. Looking into it.

src/js/vic-v2/components/PhotoField.jsx
rianfowler: Updated this.

src/js/vic-v2/components/CropperController.jsx
rianfowler: This has been updated.

src/js/vic-v2/components/PhotoField.jsx
rianfowler: This has been removed.

src/js/vic-v2/components/CropperController.jsx
rianfowler: This should be repaired now.

rianfowler:
This should be good now.

src/sass/vic-v2.scss
```diff
@@ -216,9 +216,11 @@ input[type=range] {
   }
 }
 
-.cropper-control-label-container {
-  margin: 10px;
-  min-width: 106px;
+.cropper-control-row {
+  .cropper-control-label-container {
```
rianfowler: This was being overridden on the review page which was removing the space between the buttons. Using a more specific query fixed it.

src/js/vic-v2/components/CropperController.jsx
```diff
@@ -402,12 +417,22 @@ export default class CropperControls extends React.Component {
...
+  // Cropper doesn't support changing width after componentDidMount
+  // use the windowWidth as a key to for the component to remount when the window width changes
   render() {
+
+    // narrow-layout class controls visiblity of elements based on window width
+    // if layout is forced narrow, remove this behavior
+    const narrowLayoutButtonsClassNames = classNames('cropper-control-row', { 'narrow-layout': !this.props.forceNarrowLayout });
+
+    // prevent the cropper from re-mounting when forcing narrow layout and width is in large screen range
+    const cropperKey = this.props.forceNarrowLayout && this.state.windowWidth >= LARGE_SCREEN ? LARGE_SCREEN : this.state.windowWidth;
```
rianfowler: This a little ugly but I wanted to prevent the cropper from remounting if the width of its container wasn't changing.

jbalboni (approved):

PR #7413, opened 2018-03-13
Set up tests for existing/in progress pages

cehsu:
These changes set up tests for existing/in progress pages of the disability benefits 526 form. ([Issue](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/9015))

### DIFF
--- a/test/disability-benefits/526EZ/config/vaFacilities.unit.spec.jsx
+++ b/test/disability-benefits/526EZ/config/vaFacilities.unit.spec.jsx
@@ -1,0 +1,120 @@
+import React from 'react';
+import { expect } from 'chai';
+import sinon from 'sinon';
+import { mount } from 'enzyme';
+
+import { DefinitionTester, fillData } from '../../../util/schemaform-utils.jsx';
+import formConfig from '../../../../src/js/disability-benefits/526EZ/config/form.js';
+
+const initialData = {
+  // For testing purposes only
+  disabilities: [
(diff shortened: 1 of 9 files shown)


test/disability-benefits/526EZ/config/vaFacilities.unit.spec.jsx
```diff
@@ -0,0 +1,120 @@
+import React from 'react';
+import { expect } from 'chai';
+import sinon from 'sinon';
+import { mount } from 'enzyme';
+
+import { DefinitionTester, fillData } from '../../../util/schemaform-utils.jsx';
+import formConfig from '../../../../src/js/disability-benefits/526EZ/config/form.js';
+
+const initialData = {
```
cvalarida: What do you think about pulling this `initialData` out into a separate file? We could then use it in all these test _and_ as the initial data in the `form.js` (until we're done building the form, of course).

cvalarida (approved):
If this isn't up-to-date with master, we'll probably want to ensure it is before merging it in to avoid it breaking the build on master. It all looks good, though. I just had one comment on reducing duplication.

test/disability-benefits/526EZ/config/vaFacilities.unit.spec.jsx
cehsu: Good idea, updated.

PR #7416, opened 2018-03-13
Fixed bug that allowed attempt at account creation even if MHV services were not accessible

U-DON:
Resolves department-of-veterans-affairs/vets.gov-team#9219.

The introduction of the `DowntimeNotification` component as an intermediate gating component failed to propagate the `isDataAvailable` flag that gets passed down from `RequiredLoginView`. Part of the problem is that I failed to check for that anyway first thing before doing anything in `MHVApp`, but foremost is that it wasn't propagating anymore.

This fix will prevent `MHVApp` from doing anything account creation related when the user is ineligible, which will result in showing the MHV access error with the three possible reasons. Longer term, I'd like to revamp `RequiredLoginView`, not only to stop using `loa`, but also to introduce a more reliable mechanism to handle inaccessible services.

I also shuffled order of wrapper components, which fixes visual margin problems when we we get account creation errors.
> screen shot 2018-03-13 at 4 00 40 pm

### DIFF
--- a/src/js/common/containers/MHVApp.jsx
+++ b/src/js/common/containers/MHVApp.jsx
@@ -44,7 +44,11 @@ export class MHVApp extends React.Component {
 
   isAccessible = () => ['existing', 'upgraded'].includes(this.props.account.state);
 
+  isIneligible = () => this.props.account.state === 'ineligible';
+
   handleAccountState = () => {
+    if (this.isIneligible()) { return; }
+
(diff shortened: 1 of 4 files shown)


jkassemi:
@U-DON can you please tie this back to an issue in vets.gov-team with the "bug" label?

saneshark (approved):
LGTM

PR #7417, opened 2018-03-14
Add **CSP_NONCE** to all inline script to support CSP nonce whitelisting

edpaget:
The `**CSP_NONCE**` will be replaced by a random string matching a nonce header by nginx so we can disable unsafe-inline in our Content-Security-Policy for vets.gov.


ncksllvn (approved):
LGTM

jbalboni (approved):

PR #7418, opened 2018-03-14
Remove Bob's verification file for search console

robertfairhead:
This is a retry of a PR to remove the verification file for my access to Google's Search Console for vets.gov

I'm no longer on the project so the access should be removed. Google won't do it while the file exists. We have a new `vetsdotgovanalytics` account that has a separate verification file to main continuity when mine is removed.


robertfairhead:
hmm, @phiden it looks like jenkins isn't picking up the PR because it's from my fork (I don't have direct access to the repo anymore). Closing this out.

PR #7422, opened 2018-03-14
Claims Index Page -- Appeals buttons on new lines

cvalarida:
image


gnakm (approved):
lgtm!

cehsu (approved):
LGTM. Rather than switching to `<p>` and adding inline styling, should we add a `<br/>` after the `<span>` instead?

cvalarida:
@cehsu I don't know that I'm super opinionated either way, but for a blurb of text that's supposed to be on its own line, it seems like a `<p>` is more semantic anyhow.

PR #7424, opened 2018-03-14
Autogenerate nonced script tags

edpaget:
Reverts #7417 to add a metalsmith step to automatically add the nonces at build time. `unsafe-inline` also disables javascript in `onlick` handlers, so the build step also automatically extracts `onclick` handlers into a new script tag.


edpaget:
Ugh uses too much memory :disappointed:

script/build.js
```diff
@@ -690,6 +691,40 @@ smith.use(redirect({
...
+/*
+Add nonce attribute with substition string to all inline script tags
+Convert onclick event handles into nonced script tags
+*/
+smith.use((files, metalsmith, done) => {
+  Object.keys(files).forEach((file) => {
+    if (path.extname(file) !== '.html') return;
+
+    const data = files[file];
+    const dom = new jsdom.JSDOM(data.contents.toString());
```
jbalboni: I wonder if something like cheerio would fit better here? It doesn't look like you need something as in depth as jsdom. Though I realize we already depend on it, so it's easier to pull in.

https://github.com/cheeriojs/cheerio

script/build.js
```diff
@@ -690,6 +691,40 @@ smith.use(redirect({
   '/education/apply-for-education-benefits/': '/education/apply/'
 }));
 
+/*
+Add nonce attribute with substition string to all inline script tags
+Convert onclick event handles into nonced script tags
+*/
+smith.use((files, metalsmith, done) => {
```
jbalboni: Could we put this in a separate file? I think we should start doing that more, this file is pretty big.

src/js/common/utils/accessible-menus.js
```diff
@@ -66,7 +75,7 @@ function isVisible(element) {
 
   // If the element has children, check them; otherwise, naively assume it's visible
   const childrenAreVisible = element.children.length ?
-    Array.from(element.children).some(e => !hiddenDisplays.includes(getComputedStyle(e).display)) :
+    Array.from(element.children).some(e => isScript(e) || !hiddenDisplays.includes(getComputedStyle(e).display)) :
```
jbalboni: Not sure who the right person to validate this change is. @cvalarida, maybe?

script/build.js
ncksllvn: @jbalboni Is this going to have conflicts with the work being done for Webpack 4?

src/js/common/utils/accessible-menus.js
ncksllvn: If I'm reading this right, do we potentially have a script tag in the middle of the HTML for our navigation?

script/build.js
ncksllvn: Yeah that file is a beast. I'm in favor breaking every plugin into its own module as tedious as that may be.

script/build.js
```diff
@@ -690,6 +691,40 @@ smith.use(redirect({
...
+        onclickEl.id = Math.random().toString(36).replace(/[^a-z]+/g, '').substr(0, 10);
+      }
+      const id = onclickEl.id;
+      const onclick = onclickEl.getAttribute('onclick');
+      const newScript = dom.window.document.createElement('script');
+      newScript.setAttribute('nonce', '**CSP_NONCE**');
+      newScript.textContent = `(function() { var e = document.getElementById('${id}'); e.addEventListener('click', function(ev) { ${onclick} }); })();`;
+      onclickEl.removeAttribute('onclick');
+      onclickEl.appendChild(newScript);
+    });
```
ncksllvn: Are there a lot of instances of defining the onclick handler in the element like this? This is cool code but I think it would be better to define a standard for not defining onclick handlers in the element, because this seems a little magical/cryptic for a build step.

ncksllvn:
Great idea to define this as a Metalsmith plugin @edpaget!

script/build.js
edpaget: I'm fine to move it. Is there a good place in the repo for this stuff to live?

script/build.js
edpaget: I totally agree that this is probably too magic. There are just 97 instances of onclick handlers (doesn't seem like there are any other inline event handlers being used). I can manually remove them all and turn this into a step that will error if elements have inline handlers instead. However, I think its possibly fine to keep this build step for now and gradually remove the handlers along with making it a rule to not add any more.

src/js/common/utils/accessible-menus.js
edpaget: Yeah the onclick transformer will append a script tag as a child of the element with an inline onclick to make sure it's in the dom before the handler gets step.

edpaget:
Thanks @ncksllvn  seemed if we wanted to restrict the content policy and still have inline script tags, we needed a more automatic way to handle it.

src/js/common/utils/accessible-menus.js
edpaget: Also noting that there was an e2e test that failed before this fix. I should probably make sure the unit tests exercise this code path.

script/build.js
ncksllvn: Yeah that sounds right. I think the navigation alone has a lot of instances for capturing analytics. But that makes sense to me. @jbalboni what do you think?

script/build.js
ncksllvn: I don't think we've really defined a place for it - maybe a subdirectory, `build/metalsmith-plugins`?

script/build.js
jbalboni: It may be hard to get away from them entirely in the static content, but it seems like we could at least handle the analytics use case differently.

script/build.js
jbalboni: @ncksllvn No, shouldn't have any conflicts with the Webpack changes.

src/js/common/utils/accessible-menus.js
cvalarida: Yeah, looks good to me. :+1:

script/metalsmith/nonceTransformer.js
```diff
@@ -0,0 +1,42 @@
...
+const path = require('path');
+
+function generateNewId(existingIds) {
+  const newId = Math.random().toString(36).replace(/[^a-z]+/g, '').substr(0, 10);
+  if (!existingIds.has(newId)) {
+    existingIds.add(newId);
+    return newId;
+  }
+  return generateNewId(existingIds);
+}
```
ncksllvn: Not sure if it's exactly what you're looking for but Lodash has the [`uniqueId`](https://lodash.com/docs/4.17.5#uniqueId) function.

ncksllvn (approved):
LGTM. Great work!

PR #7425, opened 2018-03-15
526 e2e -- The case of the missing page

cvalarida:
Due to a couple PRs getting merged into master at once, the e2e test was failing. This adds the missing page to the test which was causing the failure. It _doesn't_ add the page missing from master right now.

If you find your knowledge in this missing something, don't worry--you're not missing much. Let's just say I tried a thing, but it was a swing and a miss. I was close, but missed the mark.

Too much?


cehsu (approved):
LGTM. Thanks for being proactive on this!

PR #7429, opened 2018-03-15
Hide and log unknown events

cvalarida:
There's not much by way of screenshots, but I did test with mock data to ensure it's working properly.

### DIFF
--- a/src/js/claims-status/utils/appeals-v2-helpers.jsx
+++ b/src/js/claims-status/utils/appeals-v2-helpers.jsx
@@ -650,10 +650,12 @@ export function getEventContent(event) {
         description: '',
       };
     default:
-      return {
-        title: 'Unknown event',
-        description: '',
-      };
+      Raven.captureMessage('appeals-unknown-event', {
+        extra: {
+          eventType: event.type
+        }
+      });
+      return null;
   }
 }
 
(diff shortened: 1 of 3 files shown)


src/js/claims-status/utils/appeals-v2-helpers.jsx
```diff
@@ -650,10 +650,12 @@ export function getEventContent(event) {
         description: '',
       };
     default:
-      return {
-        title: 'Unknown event',
-        description: '',
-      };
+      Raven.captureMessage('appeals-unknown-event', {
+        extra: {
+          eventType: event.type
```
cehsu: Would any other information about the event be helpful to track?

cehsu (approved):
LGTM. Had a question about what else (if anything), might also be of interest to track when logging the error.

src/js/claims-status/utils/appeals-v2-helpers.jsx
cvalarida: :thinking: I don't think so, but @cmgiven might have an idea. What do you think, Chris?

cmgiven: I think that would do it. This helpfully flags a mismatch between the API and the front-end. Knowing is most of the battle, and we can quickly figure out the rest once we know there's an issue.

PR #7430, opened 2018-03-15
Fix IE layout and remove normal layout mode on cropper

jbalboni:
The original problem is that the cropper control container is set at a minimum width of 600px for large screens. This is wider than the normal form page width, and Chrome handles this by expanding the whole page. IE, though, only expands that particular div. Either way, we shouldn't be making the cropper tool wider than the grid column it's in, but the normal width of the page is too small for the "normal" layout. So I've pulled that out, done what's possible to make the width larger, and made the move/rotation buttons more responsive.

screen shot 2018-03-15 at 12 32 53 pm
screen shot 2018-03-15 at 12 32 33 pm


rianfowler:
This definitely makes things simpler in a lot of places.

rianfowler (approved):

PR #7431, opened 2018-03-15
Make the (+) clickable

cvalarida:
No visual changes here.

I did have to do some fancy footwork with [`pointer-events`](https://developer.mozilla.org/en-US/docs/Web/CSS/pointer-events) which I learned is a thing today. It's not the most elegant because we have to have a click event handler on an `<li>`, but that isn't the end of the world, I suppose. We keep it on the button for accessibility and cross-browser compatibility (for keyboard navigation, mostly).

To be fair, I don't know if we _really_ need it on both the parent `<li>` and child `<button>`, but I figured it'd be safer this way. We've got a `e.stopPropagation()` to avoid it from opening and closing in one fell swoop.

### DIFF
--- a/src/js/claims-status/components/appeals-v2/Timeline.jsx
+++ b/src/js/claims-status/components/appeals-v2/Timeline.jsx
@@ -23,7 +23,10 @@ class Timeline extends React.Component {
     return `${first} - ${last}`;
   };
 
-  toggleExpanded = () => this.setState((prevState) => ({ expanded: !prevState.expanded }));
+  toggleExpanded = (e) => {
+    e.stopPropagation();
+    this.setState((prevState) => ({ expanded: !prevState.expanded }));
+  }
 
   render() {
     const { events, missingEvents } = this.props;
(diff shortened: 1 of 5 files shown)


src/js/claims-status/components/appeals-v2/Timeline.jsx
```diff
@@ -23,7 +23,10 @@ class Timeline extends React.Component {
     return `${first} - ${last}`;
   };
 
-  toggleExpanded = () => this.setState((prevState) => ({ expanded: !prevState.expanded }));
+  toggleExpanded = (e) => {
+    e.stopPropagation();
+    this.setState((prevState) => ({ expanded: !prevState.expanded }));
```
cehsu: Interesting, what's the advantage of using `prevState` over `this.state`?

cehsu (approved):
LGTM.

src/js/claims-status/components/appeals-v2/Timeline.jsx
cvalarida: :thinking: Y'know, I don't remember....I'm sure I had a good reason!

PR #7432, opened 2018-03-15
Add merged status

cvalarida:
Index page:
image

Status page:
image

There definitely should be a space above the "Your appeal is now closed" section, but that probably should to be a different PR.

### DIFF
--- a/src/js/claims-status/utils/appeals-v2-helpers.jsx
+++ b/src/js/claims-status/utils/appeals-v2-helpers.jsx
@@ -2,6 +2,7 @@ import React from 'react';
 import moment from 'moment';
 import _ from 'lodash';
 import Raven from 'raven-js';
+import { Link } from 'react-router';
 
 // This literally determines how many rows are displayed per page on the v2 index page
 export const ROWS_PER_PAGE = 10;
@@ -41,6 +42,7 @@ export const STATUS_TYPES = {
   reconsideration: 'reconsideration',
   death: 'death',
   otherClose: 'other_close',
+  merged: 'merged'
 };
 
 export const ISSUE_STATUS = {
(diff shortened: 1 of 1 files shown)


gnakm:
@cvalarida Should I open a different ticket for the spacing? that looks very gnarly.

gnakm:
Also, if an appeal was merged, it should still show as active (green dot) next to status on the grey card. Thoughts there? @cvalarida

cvalarida:
@gnakm Yeah, a new ticket would be good. :thinking: Or what if it were in https://github.com/department-of-veterans-affairs/vets.gov-team/issues/9010? I know that one covers a different thing, but this could fall under a generic "padding tweaks" ticket. I'm okay with adding that extra complexity to that one if you are.

As far as the green circle goes, that's based on whether the appeal is active (so `appeal.attributes.active`), not on the status type. I just plugged in the status type to a closed appeal because I thought it made sense. If it'll be sent as an active appeal with a merged current status from caseflow, then it'll show up with a green dot.

I will change the placement of the status type in our list, though, since I put it under the "closed statuses" section. That doesn't _do_ anything, though--it's just to help us read the code.

cvalarida:
Actually, it looks like it's a bit more nuanced than that. We consider the appeal "active" if:
1. `appeal.attributes.active` is true
2. `appeal.attributes.status.type` is _not_ one of the following
    - `pending_soc`
    - pending_form_9
    - decision_in_progress
    - bva_development
3. `appeal.attributes.type` is _not_ one of the following
    - reconsideratione
    - cue

gnakm:
Yeah the appeal status type will always be active.

gnakm (approved):
lgtm!

cehsu (approved):
LGTM.

PR #7436, opened 2018-03-15
Personalization: Remove unused files

ncksllvn:
The diff was messed up the first time I opened this.


src/js/personalization/user-profile-beta/components/ProfileAccordion.jsx
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable import/no-unresolved */
```
bshyong: @ncksllvn do you know why this was added?

ncksllvn: Yeah, I don't want to delete this file because I want to keep it as a reference in case we move back to the accordion UI. But a linting error occurs keeping it around without any references to it.

bshyong: @ncksllvn sorry, I'm not understanding what the problem is with the `import/no-unresolved` rule. Can you explain?

That rule checks if any of the import paths are unresolved/non-existent, e.g. if you do `/path/foo.jsx` and `foo.jsx` does not exist

ncksllvn: Yeah exactly! That file is never used, and the linter picks up on that, then errors out with `import/no-unresolved`

ncksllvn: But I want to keep that file...

ncksllvn:
```
/Users/nicksullivan/Repos/vets-website/src/js/personalization/user-profile-beta/components/ProfileAccordion.jsx
  2:41  error  Unable to resolve path to module '../../common/components/Accordion'  import/no-unresolved
```

bshyong: I think you’re misinterpreting that rule. We should not have incorrect paths in the code. 

Maybe you are thinking about this one?
https://eslint.org/docs/rules/no-unused-vars

PR #7439, opened 2018-03-15
Redirect v1 ids

cvalarida:
This assumes v2 ids will never "override" v1 ids. That is, it assumes that a v2 id will never point to a _different_ v1 appeal. Is this assumption correct, @cmgiven? (If that's not clear, I can try to rephrase to make more sense...)

### DIFF
--- a/src/js/claims-status/actions/index.jsx
+++ b/src/js/claims-status/actions/index.jsx
@@ -94,10 +94,23 @@ export function getAppeals(filter) {
 }
 
 export function fetchAppealsSuccess(response) {
+  const appeals = response.data;
+  const v1ToV2IdMap = {};
+  appeals.forEach(appeal => {
+    // In case there are no v1 appeal ids
+    if (!appeal.attributes.appealIds) {
+      return;
+    }
+
+    appeal.attributes.appealIds.forEach(id => {
+      v1ToV2IdMap[id] = appeal.id;
+    });
+  });
+
   return {
     type: FETCH_APPEALS_SUCCESS,
-    // filter: filter, // No idea why this would be needed, but it's in the old version
-    appeals: response.data
+    appeals,
+    v1ToV2IdMap
   };
 }
 
--- a/src/js/claims-status/containers/AppealInfo.jsx
+++ b/src/js/claims-status/containers/AppealInfo.jsx
@@ -117,11 +117,12 @@ AppealInfo.propTypes = {
 
 function mapStateToProps(state, ownProps) {
   const { appealsLoading, appealsAvailability } = state.disability.status.claimsV2;
+  const { v1ToV2IdMap } = state.disability.status.appeals;
   return {
-    appeal: isolateAppeal(state, ownProps.params.id),
+    appeal: isolateAppeal(state, ownProps.params.id, v1ToV2IdMap),
     appealsLoading,
     appealsAvailability,
-    fullName: state.user.profile.userFullName
+    fullName: state.user.profile.userFullName,
   };
 }
 
--- a/src/js/claims-status/utils/appeals-v2-helpers.jsx
+++ b/src/js/claims-status/utils/appeals-v2-helpers.jsx
@@ -148,13 +148,15 @@ export function addStatusToIssues(issues) {
 }
 
 /**
- * Finds an appeal from the Redux store with ID matching arg ID
+ * Finds an appeal from the Redux store with ID matching arg ID.
+ * `id` may be a v1 id or a v2 id.
+ *
  * @param {object} state Full redux store state tree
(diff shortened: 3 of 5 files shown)


cmgiven:
If you mean that a v1 id will always point to a v2 id found in the `appeal.attributes.appealIds` array of the correct v2 appeal, that's correct. Is that what you mean by "override"?

cvalarida:
I was specifically asking about a v2 id accidentally pointing to a v1 record.

For example, if we have the following:

**V1 ID List**
```
1234: 'Appeal 1',
4321: 'Appeal 2'
```

**V2 ID List**
```
1233: 'Appeal 1',
1234: 'Appeal 2'
```

I have the url for `1234` bookmarked, so when I go there, I get 'Appeal 2' instead of 'Appeal 1' like I expected.

cmgiven:
Yeah, so these ids are drawn from the same source and reference the same data. The only difference is that in many cases what would have been discrete appeals under v1 are now combined into a single appeal under v2.

cvalarida:
Okay, perfect. I had a hunch, but wanted to verify. Thanks kindly!

src/js/claims-status/utils/appeals-v2-helpers.jsx
```diff
@@ -148,13 +148,15 @@ export function addStatusToIssues(issues) {
...
+ * `id` may be a v1 id or a v2 id.
+ *
  * @param {object} state Full redux store state tree
  * @param {string} id Appeal ID of the appeal to find
  * @returns {object} One appeal object or undefined if not found in the array
  */
-export function isolateAppeal(state, id) {
-  return _.find(state.disability.status.claimsV2.appeals, (a) => a.id === id);
+export function isolateAppeal(state, id, v1ToV2IdMap) {
+  return _.find(state.disability.status.claimsV2.appeals, (a) => a.id === id || v1ToV2IdMap[id]);
```
cehsu: This is great as is. Is if of interest to use the map to store the appeal as the value of the key above, so that we can have O(1) lookup here? If so, we could also store the v2 ids and appeals in the same structure as well (but probably with a new name).

src/js/claims-status/actions/index.jsx
```diff
@@ -94,10 +94,23 @@ export function getAppeals(filter) {
...
+  const appeals = response.data;
+  const v1ToV2IdMap = {};
+  appeals.forEach(appeal => {
+    // In case there are no v1 appeal ids
+    if (!appeal.attributes.appealIds) {
+      return;
+    }
+
+    appeal.attributes.appealIds.forEach(id => {
+      v1ToV2IdMap[id] = appeal.id;
```
cehsu: What is the plan to use this? Would it make sense to create an object with the v2 appeal id as the key, and the appeals as the values?

cehsu (approved):
LGTM. I had a few questions about how the appeal ids are used.

src/js/claims-status/containers/AppealInfo.jsx
```diff
@@ -117,11 +117,12 @@ AppealInfo.propTypes = {
...
 function mapStateToProps(state, ownProps) {
   const { appealsLoading, appealsAvailability } = state.disability.status.claimsV2;
+  const { v1ToV2IdMap } = state.disability.status.appeals;
   return {
-    appeal: isolateAppeal(state, ownProps.params.id),
+    appeal: isolateAppeal(state, ownProps.params.id, v1ToV2IdMap),
     appealsLoading,
     appealsAvailability,
-    fullName: state.user.profile.userFullName
+    fullName: state.user.profile.userFullName,
```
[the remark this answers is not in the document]
cvalarida: Nope. I probably added a thing, then removed it and forgot the comma.

src/js/claims-status/actions/index.jsx
```diff
@@ -94,10 +94,23 @@ export function getAppeals(filter) {
 }
 
 export function fetchAppealsSuccess(response) {
+  const appeals = response.data;
+  const v1ToV2IdMap = {};
+  appeals.forEach(appeal => {
+    // In case there are no v1 appeal ids
```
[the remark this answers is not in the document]
cvalarida: Right, exactly. v1 appeals may or may not be any number of v1 appeals all rolled up in one.

src/js/claims-status/utils/appeals-v2-helpers.jsx
cvalarida: Yeah, great point. We mostly copy-pasta'd this from v1 without trying to make it efficient. That would be a good candidate for refactor (which we will have a lot of if we can ever get around to addressing tech debt).

PR #7444, opened 2018-03-16
Switch between fieldset and div for object field in forms

jbalboni:
Switches between fieldset and div, so that we don't have fieldsets that don't have a legend in them. Also changes the code so that we always use legend, instead of switching between that and h5. I had done that for style reasons, but it's not a good practice for a11y.


annekainicUSDS (approved):
LGTM!

cvalarida (approved):

PR #7446, opened 2018-03-16
Personalization: Beta Enhancements

ncksllvn:
This is still very very early stages of the flow for an application in beta. Previously, there was a wrapper component that would ask the user to enroll in the beta on the same route as the beta app. It seems like a much more natural flow is to redirect the user to a standalone page for beta enrollment, and to provide an explanation of what the beta enrollment entails. It's also manageable in markdown in case the content team wants to reword it. 

Once a larger initiative of using AB Testing for beta applications has begun, this will definitely change, but I think this is still a valuable update to our current beta flow.


src/js/common/actions/index.js
```diff
@@ -1,2 +1,27 @@
...
+export function getScheduledDowntime() {
+  return (dispatch) => {
+    dispatch({ type: RETREIVE_SCHEDULED_DOWNTIME });
+    return apiRequest(
+      '/maintenance_windows/',
+      undefined,
+      (json) => { receiveScheduledDowntime(dispatch, json.data); },
+      () => { receiveScheduledDowntime(dispatch, []); });
+  };
+}
```
ncksllvn: This file I ultimately just restored to what it was prior to the beta actions, because those are now stored in beta-enrollment/actions.

src/js/common/containers/BetaApp.jsx
```diff
@@ -2,59 +2,35 @@ import React from 'react';
...
-import { registerBeta, isUserRegisteredForBeta } from '../actions';
-import AlertBox from '../components/AlertBox';
-
-export const features = {
-  healthAccount: 'health_account',
-  veteranIdCard: 'veteran_id_card',
-  profile: 'profile',
-  dashboard: 'dashboard'
-};
+import { isUserRegisteredForBeta } from '../../beta-enrollment/actions';
```
ncksllvn: This file is shortened dramatically, because now it ultimately just confirms that they are enrolled in the corresponding beta, and redirects them if not. The logic to register them in the beta has been moved into the beta-enrollment project.

content/pages/beta-enrollment/personalization.md
```diff
@@ -0,0 +1,27 @@
...
+You are attempting to access the Personalization features of Vets.gov. These features are currently in active development and are therefore prone to change.
+
+Enrolling in this beta will enable the following features:
+
+- Personalized Dashboard
+- Personalized Profile
+- Account Information
+
+<div id="react-root"></div>
+</div>
```
ncksllvn: The react-root here will ultimately just render a button for enrolling in a beta.

PR #7447, opened 2018-03-16
Add e2e test for chapter 36

jbalboni:


package.json
```diff
@@ -220,7 +220,7 @@
     "react-dropzone": "^4.2.5",
     "reselect": "^2.5.4",
     "url-search-params": "^0.10.0",
-    "vets-json-schema": "https://github.com/department-of-veterans-affairs/vets-json-schema.git#923c0ad5676bcdcad0c8fe08017feb05a5302da9",
+    "vets-json-schema": "https://github.com/department-of-veterans-affairs/vets-json-schema.git#b34bc2468dd4c88089cf824078431f1f9c3fc2ec",
```
jbalboni: Need to update this when corresponding PR is merged.

rianfowler (approved):
Does this represent what we generally do with e2e test- test all of the fields on the happy path?

jbalboni:
@rianfowler Yes, we usually go through and fill out all the fields to make sure moving between pages works and all the fields exist as we expect. We try to test more of the conditional and validation logic in the config unit tests.

PR #7449, opened 2018-03-19
Update Node and Yarn

jbalboni:
Updating us to reasonably recent version of yarn and Node.


ncksllvn:
Does `.nvmrc` need to be updated as well?

.nvmrc
```diff
@@ -1 +1 @@
-6.11.1
+8.10.0
```
rianfowler: `async await` 🎆

package.json
```diff
@@ -211,6 +211,7 @@
     "modernizr": "^3.6.0",
     "modernizr-loader": "^1.0.1",
     "nyc": "^11.1.0",
+    "prop-types": "^15.6.1",
```
rianfowler: No more deprecation warning 👍

rianfowler (approved):

package.json
jbalboni: Eh, I think that's still there until we update some other dependencies.

ncksllvn (approved):
LGTM! This is exciting!

edpaget (approved):
:rocket:

PR #7454, opened 2018-03-19
Add sidebar content

cvalarida:
image

### DIFF
--- a/src/js/claims-status/containers/YourClaimsPageV2.jsx
+++ b/src/js/claims-status/containers/YourClaimsPageV2.jsx
@@ -191,7 +191,7 @@ class YourClaimsPageV2 extends React.Component {
             <div>
               <h2 className="help-heading">Can’t find your appeal?</h2>
               <p>
-                If you submitted a Notice of Disagreement for an appeal in within the last 3 months,
+                If you submitted a Notice of Disagreement for an appeal within the last 3 months,
                 VA might still be processing your appeal. For more information, contact your Veterans
                 Service Organization or representative.
               </p>
--- a/src/js/claims-status/containers/YourClaimsPageV2.jsx
+++ b/src/js/claims-status/containers/YourClaimsPageV2.jsx
@@ -188,6 +188,14 @@ class YourClaimsPageV2 extends React.Component {
           <div className="small-12 usa-width-one-third medium-4 columns help-sidebar">
             <FeaturesWarning/>
             <AskVAQuestions/>
+            <div>
+              <h2 className="help-heading">Can’t find your appeal?</h2>
+              <p>
+                If you submitted a Notice of Disagreement for an appeal in within the last 3 months,
+                VA might still be processing your appeal. For more information, contact your Veterans
+                Service Organization or representative.
+              </p>
+            </div>
           </div>
         </div>
       </div>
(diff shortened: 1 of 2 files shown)


gnakm:
Could we tighten the spacing between the first and second line of "Can't find your appeal?"

cvalarida:
Reduced the line height from 1.5 to 1.2
image

cvalarida:
Added spacing per @gnakm's request
image

gnakm:
:shipit:

gnakm (approved):

src/js/claims-status/containers/YourClaimsPageV2.jsx
```diff
@@ -188,6 +188,14 @@ class YourClaimsPageV2 extends React.Component {
           <div className="small-12 usa-width-one-third medium-4 columns help-sidebar">
             <FeaturesWarning/>
             <AskVAQuestions/>
+            <div>
+              <h2 className="help-heading">Can’t find your appeal?</h2>
+              <p>
+                If you submitted a Notice of Disagreement for an appeal in within the last 3 months,
```
cehsu: Extra word here ("in within")?

src/js/claims-status/containers/YourClaimsPageV2.jsx
```diff
@@ -188,6 +188,14 @@ class YourClaimsPageV2 extends React.Component {
           <div className="small-12 usa-width-one-third medium-4 columns help-sidebar">
             <FeaturesWarning/>
             <AskVAQuestions/>
+            <div>
```
cehsu: Should we put this into a component similar to AskVAQuestions? Would that enable us to reuse some of that components styles and allow us to avoid creating new style rules?

cehsu (approved):
LGTM. Had a question about whether we might be able to reuse styles rather than create new rules.

src/js/claims-status/containers/YourClaimsPageV2.jsx
gnakm: I don't know what AskVAQuestions is but assuming that it means a forum where we can talk about site wide things, I think it is a great idea. My guess is also that after font Bitter was on vetsdotgov, a lot of the spacing shifted.

src/js/claims-status/containers/YourClaimsPageV2.jsx
gnakm: Good 👀 Could we take `in` out?

src/js/claims-status/containers/YourClaimsPageV2.jsx
cvalarida: @gnakm `AskVAQuestions` is a reusable component that displays the "Need help?" section of the sidebar on this page.

@cehsu I thought about it, but it didn't seem like a great idea. `AskVAQuestions` is used on a number of different pages in the claims status app, but this "Can't find your appeals?" bit is only used here. Abstracting it out for the two sections to share a common component seemed overkill, and creating a new component that's only used here seemed like more overhead than it was worth.

As far as the new style rule goes, that was needed for the spacing between the sidebar components which wasn't there in the first place, so we'd need to make them either way.

src/js/claims-status/containers/YourClaimsPageV2.jsx
cvalarida: :female_detective:

PR #7462, opened 2018-03-20
Appeals v2 launch cv

cvalarida:
No visual changes.

URLs changed, links changed, api endpoint changed...It all works.

To be clear, I tested before I changed the api endpoint because calling `/v0/appeals` wouldn't call the right one since the switch hasn't been merged in `vets-api` yet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ accidentally modify copies of upstream.
 
 
 
-The requirements for running this application are Node.js 8.10.0 and yarn 0.27.5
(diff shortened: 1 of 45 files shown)


kreek:
@cvalarida if you want to test it locally you can use the vets-api `replace-appeals-v1-with-v2` branch, but we'll let the auto deploy go and get this in after so we can check things out on staging.

cvalarida:
Blast, gotta fix up the old appeals e2e test...

kreek (approved):
LGTM once specs pass 🚀

PR #7463, opened 2018-03-20
Upgrade choma to get random test seed printed for reproducing failures

edpaget:
The new version prints its seed to the console so I can debug the random failures that are still occasionally occurring.


ncksllvn:
Does this need to be updated in `package.json` as well?

edpaget:
I don't think so `^1.1.0` allows minor version bumps so we should be fine with just the updated yarn.lock to get us to `1.2.0`. I don't think any other behaviour changed between releases besides printing the seed to console.

ncksllvn (approved):

PR #7465, opened 2018-03-20
Implemented on-demand session routes

U-DON:
Resolves department-of-veterans-affairs/vets.gov-team#8888.
Depends on department-of-veterans-affairs/vets-api#1781.

Removes all of the old code that fetched URLs upfront and instead only fetch when the relevant links are clicked. The responses would then be loaded in the pop-up window.


jbalboni:
@U-DON This looks awesome! I tried to checkout both this and the backend branch, but the front end just hangs on a loading spinner when I click on Sign In. Anything else I need to do to get this to run locally?

U-DON:
I'm just using this vets-website branch (`login_routes`) and the `redirect_url_json` branch for vets-api.

U-DON:
Oh, I see, there are still lingering mentions of `loginUrls`.

U-DON:
Resolved. There should no longer be a spinner.

U-DON:
Added tests for the new helpers.

Added some error handling:
- If URL failed to be fetched, we open the generic login error in the popup window. Kind of hijacked the failure case for the SAML auth flow for this...
- If failed to open a window at all, we report it to Sentry. We should probably show an explicit error, but unsure yet what that should look like.
- Reporting to Sentry addresses the idea to do so in department-of-veterans-affairs/vets.gov-team#8930. We've probably also eliminated those errors about `then` not being a property of `undefined`, which seem related to the upfront URL fetch requests.

src/js/login/utils/helpers.js
```diff
@@ -0,0 +1,65 @@
...
+  if (popupWindow) {
+    window.dataLayer.push({ event: openedEvent });
+    popupWindow.focus();
+
+    return apiRequest(
+      popupUrl,
+      null,
+      ({ url }) => { popupWindow.location = url; },
+      () => { popupWindow.location = `${environment.BASE_URL}/auth/login/callback`; }
+    ).then(() => Promise.resolve(popupWindow));
```
jbalboni: You should be able to return popupWindow here without the Promise.resolve, I think.

src/js/login/utils/helpers.js
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  Raven.captureMessage('Failed to open new window', {
+    extra: { url: popupUrl }
+  });
+
+  return Promise.reject('Failed to open new window');
```
jbalboni: In general, I think we should try to reject with Errors instead of strings.

jbalboni (approved):
This looks great. I tested it locally on the forms and didn't run into any issues. I had a couple minor comments, but this is so, so much better than what was there before.

dhall (approved):
This looks great! Thanks for all your work to simplify this as much as you have.

U-DON:
@jbalboni @saneshark @dhall Ready for review/test.

PR #7466, opened 2018-03-20
change gibsAlert to false

egarreau:


cvalarida (approved):
Nice! I didn't realize that was a thing. Nifty! :smile:

PR #7471, opened 2018-03-21
pointing Facility Locator at the new controller for cached data

dhall:
This will be a temporary change. In a subsequent vets-api change we will replace the v0 controller and delete the vbeta controller. This interim step of having both available in the api will allow us to rollback the change very quickly if there is a problem.


U-DON (approved):
LGTM.

PR #7472, opened 2018-03-21
Add cropper unit tests, plus some others for VIC

jbalboni:
I also removed the string ref from CropperController, because there's no way to fake them in shallow rendering tests.

These aren't ideal tests for the cropping tool. I'd prefer writing more behavior focused tests, but we don't have a great test environment for that.


script/run-unit-test.sh
```diff
@@ -9,12 +9,12 @@ if [ ! -f $MOCHA ]; then
 fi
 
 # No test specified; run all of them
-if [ -z "$1" ]; then
-  BABEL_ENV=test $MOCHA test/helper.js --recursive '$SCRIPT_DIR/../test/**/*.unit.spec.js?(x)' 
+if [[ -z "$1" ]] || [[ $1 == -* ]]; then
```
jbalboni: @cvalarida I made some changes here so that we can pass the debugging flags to the mocha command and debug tests in dev tools again.

rianfowler (approved):
I guess we could have made it easier to test by externalizing the math / state parts. Probably should have made a reducer for this. Then testing the component would mostly be stubbing out calls and validating call sequences.

script/run-unit-test.sh
cvalarida: Ohh, yeah, that's much nicer!

I wrote this before I knew the magic bash incantation `"$@"`. Looks great! :+1:

PR #7477, opened 2018-03-21
WIP - Personalization - Update account to settings

ncksllvn:
# Account Settings
Accessible via `/settings-beta`. After enrolling in the beta it becomes available in the user dropdown as Settings.

### No multifactor, not verified
image

### Multifactor, verified
image


ncksllvn:
@melwoodard @BBarone1 Can you all check the screenshots and see if things look okay? I couldn't test terms and conditions because there is additional setup to run locally.

BBarone1:
@ncksllvn LGTM. the only discrepancy I see is that the "add security step" button is open (clear) in your screenshots above, whereas it is closed (blue) in the design [here](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/8917)

@melwoodard can you comment on this? should the security button be opened or closed?

ncksllvn:
@BBarone1 Ah that was just an oversight! Will Fix. Thanks!

BBarone1:
@ncksllvn ah ok! i kinda liked it and noticed that it was previously documented this way as well:
LOA1 user:

screen_shot_2018-02-22_at_9_40_21_am

so thought i'd check with @melwoodard

ncksllvn:
@BBarone1 Oh haha. I'll leave it then until @melwoodard decides

melwoodard:
Looks great Nick! Couple things:

- Re ghost vs. solid button, let's make both buttons solid as we're trying to keep things consistent sitewide where solid buttons are the default unless there's a binary action
- Can you change the label in the dropdown menu to `Account`, as well as the URL to `/account-beta` so they're consistent with the page title?

ncksllvn:
> Can you change the label in the dropdown menu to Account, as well as the URL to /account-beta so they're consistent with the page title?

Ah I must misunderstood the conversation here, department-of-veterans-affairs/vets.gov-team/issues/8917, because that's how I had it before lol. Sure, I'll change it back.

> Re ghost vs. solid button, let's make both buttons solid as we're trying to keep things consistent sitewide where solid buttons are the default unless there's a binary action

Will do!

ncksllvn:
Updated! 

image

PR #7478, opened 2018-03-21
Show cavcOption alert after "What happens next?"

cvalarida:
Ignore the current status and docket line surrounding the "alert"--I just threw this alert into a random appeal to make sure it displayed where it was supposed to go and not where it wasn't supposed to go.
image

### DIFF
--- a/src/js/claims-status/components/appeals-v2/AlertsList.jsx
+++ b/src/js/claims-status/components/appeals-v2/AlertsList.jsx
@@ -10,15 +10,13 @@ const AlertsList = ({ alerts, appealIsActive }) => {
 
   const allAlertsContent = alerts.map((alert) => getAlertContent(alert, appealIsActive));
 
-  const takeActionAlerts = allAlertsContent.filter((alert) => (alert.displayType === 'take_action'));
-  const infoAlert = allAlertsContent.filter((alert) => (alert.displayType === 'info'));
+  const takeActionAlertsPresent = allAlertsContent.some((alert) => (alert.displayType === 'take_action'));
 
-  const takeActionHeader = (takeActionAlerts.length)
+  const takeActionHeader = takeActionAlertsPresent
     ? (<h3>Take Action</h3>)
     : null;
 
-  const alertsList = takeActionAlerts
-    .concat(infoAlert)
+  const alertsList = allAlertsContent
     .map((alert, index) => {
       const key = `${alert.type}-${index}`;
       return (<Alert
--- a/src/js/claims-status/containers/AppealsV2StatusPage.jsx
+++ b/src/js/claims-status/containers/AppealsV2StatusPage.jsx
@@ -78,9 +78,9 @@ const AppealsV2StatusPage = ({ appeal, fullName }) => {
         isClosed={!appealIsActive}/>
       <AlertsList alerts={filteredAlerts} appealIsActive/>
       {appealIsActive && <WhatsNext nextEvents={nextEvents}/>}
-      {afterNextAlerts}
       {shouldShowDocket && <Docket {...docket} aod={aod} form9Date={form9Date} appealType={appealType}/>}
       {!appealIsActive && <div className="closed-appeal-notice">This appeal is now closed</div>}
+      {afterNextAlerts}
     </div>
   );
 };
--- a/src/js/claims-status/containers/AppealsV2StatusPage.jsx
+++ b/src/js/claims-status/containers/AppealsV2StatusPage.jsx
@@ -1,7 +1,15 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-import { getStatusContents, getNextEvents, APPEAL_TYPES, EVENT_TYPES, STATUS_TYPES } from '../utils/appeals-v2-helpers';
+import {
+  getAlertContent,
+  getStatusContents,
+  getNextEvents,
+  ALERT_TYPES,
+  APPEAL_TYPES,
+  EVENT_TYPES,
+  STATUS_TYPES
+} from '../utils/appeals-v2-helpers';
 
 import Timeline from '../components/appeals-v2/Timeline';
 import CurrentStatus from '../components/appeals-v2/CurrentStatus';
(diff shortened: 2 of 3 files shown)


src/js/claims-status/containers/AppealsV2StatusPage.jsx
```diff
@@ -43,15 +51,34 @@ const AppealsV2StatusPage = ({ appeal, fullName }) => {
     !hideDocketStatusTypes.includes(status.type) &&
     !hideDocketAppealTypes.includes(appealType);
 
+  const filteredAlerts = alerts.filter(a => a.type !== ALERT_TYPES.cavcOption);
+  const afterNextAlerts = (
+    <div>
+      {alerts
+        .filter(a => a.type === ALERT_TYPES.cavcOption)
+        .map((a, i) => {
```
cvalarida: I decided to stick with the overkill method because
1. I had it all written before it occurred to me that we're only looking for one alert type and there will only be one _of_ that type
2. The claims and appeals world seems pretty nebulous, and we may want to add more alerts after the "What happens next?" section later
    - I talked with @gnakm and she agreed to leave it future-proofed because things may change

That said, it's totally unnecessary and isn't hard to do later should the need arise, so I'm okay with nixing it for now.

src/js/claims-status/components/appeals-v2/AlertsList.jsx
```diff
@@ -10,15 +10,13 @@ const AlertsList = ({ alerts, appealIsActive }) => {
 
   const allAlertsContent = alerts.map((alert) => getAlertContent(alert, appealIsActive));
 
-  const takeActionAlerts = allAlertsContent.filter((alert) => (alert.displayType === 'take_action'));
-  const infoAlert = allAlertsContent.filter((alert) => (alert.displayType === 'info'));
+  const takeActionAlertsPresent = allAlertsContent.some((alert) => (alert.displayType === 'take_action'));
```
cvalarida: This seemed to be more in line with what we were trying to accomplish.

src/js/claims-status/containers/AppealsV2StatusPage.jsx
cehsu: Thanks for the helpful context @cvalarida, sounds good.

src/js/claims-status/containers/AppealsV2StatusPage.jsx
```diff
@@ -43,15 +51,34 @@ const AppealsV2StatusPage = ({ appeal, fullName }) => {
...
       <Timeline events={events} missingEvents={incompleteHistory}/>
       <CurrentStatus
         title={currentStatus.title}
         description={currentStatus.description}
         isClosed={!appealIsActive}/>
-      <AlertsList alerts={alerts} appealIsActive/>
+      <AlertsList alerts={filteredAlerts} appealIsActive/>
       {appealIsActive && <WhatsNext nextEvents={nextEvents}/>}
+      {afterNextAlerts}
       {shouldShowDocket && <Docket {...docket} aod={aod} form9Date={form9Date} appealType={appealType}/>}
```
cehsu: This change was not made here, but would it be of interest to use destructuring consistently here?

src/js/claims-status/containers/AppealsV2StatusPage.jsx
```diff
@@ -43,15 +51,34 @@ const AppealsV2StatusPage = ({ appeal, fullName }) => {
     !hideDocketStatusTypes.includes(status.type) &&
     !hideDocketAppealTypes.includes(appealType);
 
+  const filteredAlerts = alerts.filter(a => a.type !== ALERT_TYPES.cavcOption);
+  const afterNextAlerts = (
+    <div>
+      {alerts
+        .filter(a => a.type === ALERT_TYPES.cavcOption)
```
cehsu: This is great as is. Should we also think about doing something like Ruby's `Enumerable#partition` so we only need to filter this once, as part of a refactor later on?

src/js/claims-status/containers/AppealsV2StatusPage.jsx
```diff
@@ -43,15 +51,34 @@ const AppealsV2StatusPage = ({ appeal, fullName }) => {
...
+        .map((a, i) => {
+          const alert = getAlertContent(a, appealIsActive);
```
cehsu: If we destructure this so we can simply use `description` and `title` below.

cehsu (approved):
LGTM. Had a few minor questions/comments.

src/js/claims-status/containers/AppealsV2StatusPage.jsx
cvalarida: How do you mean?

src/js/claims-status/containers/AppealsV2StatusPage.jsx
cvalarida: I'm not familiar with this...what do you mean?

src/js/claims-status/containers/AppealsV2StatusPage.jsx
cehsu: Never mind, I misread where these props are coming from, apologies! Looks good!

src/js/claims-status/containers/AppealsV2StatusPage.jsx
cehsu: It looks like there is a [lodash partition function](https://lodash.com/docs/4.17.5#partition) available as well, actually, in case that is of interest to use. Using these we could produce both of the arrays needed in one fell swoop.

gnakm:
@cvalarida What if I disagree with my decision should be the last thing on the page. The mock you posted has the progress bar at the bottom. In theory, those two things will never be on a page at the same time so maybe it doesn't matter?

src/js/claims-status/containers/AppealsV2StatusPage.jsx
cvalarida: Ahh, I see. Yeah, that would make sense. In this case, I think I'm going to avoid the lodash dependency. The performance gain is minimal since this array is only ever going to be like 5 items long at most (the number of saved iterations on `O(2n)` vs `O(n)` as `n` approaches 0 is tiny).

I'll definitely keep an eye out for use cases for this where we already use lodash, though.

TIL: `_.partition()` :+1:

cvalarida:
@gnakm Ah, okay. Yeah, it probably shouldn't matter, but I can move it to the end anyhow. That's easy enough. :+1:

gnakm (approved):
lgtm 🎉

PR #7480, opened 2018-03-22
Removed use of LOA in RequiredLoginView

U-DON:
Resolves department-of-veterans-affairs/vets.gov-team#2487.
Resolves department-of-veterans-affairs/vets.gov-team#8883.

Instead of using LOA, we now have a `verify` flag that implicitly checks if the user is currently LOA3. Still some work being done to make the services list a more reliable source of truth (department-of-veterans-affairs/vets-api#1765) that will be incorporated in a separate PR. This is just the work to clean up logic so that we don't do the work of comparing LOA current and highest on the F/E.


test/e2e/login-helpers.js
```diff
@@ -88,6 +88,7 @@ function logIn(token, client, url, level) {
 
   client
     .url(`${E2eHelpers.baseUrl}${url}`)
+    .pause(500)
```
U-DON: Had to introduce a short pause here to allow the client to redirect. Not sure why it's necessary now.

src/js/common/components/RequiredLoginView.jsx
```diff
@@ -8,111 +8,104 @@ import LoadingIndicator from '../../common/components/LoadingIndicator';
...
+        const props = typeof child.type === 'function'
+          ? { isDataAvailable: false }
+          : null;
+        return React.cloneElement(child, props);
+      });
     }
 
-    return false;
+    // If the user is not verified and doesn't have MHV access, they need to verify.
+    return <Main renderType="verifyPage"/>;
```
U-DON: I'm not sure why we have to render the verify page inline. I think it would be okay to just redirect, which is more consistent with how we redirect when the user needs to log in.

src/js/common/components/RequiredLoginView.jsx
```diff
@@ -8,111 +8,104 @@ import LoadingIndicator from '../../common/components/LoadingIndicator';
...
+    // Ignore required verification when accessing health tools with MHV sign-in.
+    const hasMhvAccess =
+      healthTools.includes(serviceRequired) &&
+      user.profile.authnContext === 'myhealthevet';
+
+    if (user.profile.verified || hasMhvAccess) {
+      // TODO: Delete the logic around attemptingAppealsAccess once we
+      // resolve the MVI/Appeals Users issues.
+      // if app we are trying to access includes appeals,
+      // bypass the checks for user profile status
```
U-DON: Not sure if this is still relevant and don't know who can confirm.

U-DON: If we no longer have a scenario where appeals should still be accessible even when MVI can't be accessed for the user, I'd like to remove this.

test/e2e/login-helpers.js
U-DON: Actually would really not like to have this if possible, since it's indicative of brittle tests. Without it, the test will hang for a long time, much longer the timeout specified in the wait, and the redirect doesn't happen in the meantime. Introducing the pause somehow allows the `window.location.replace` to occur, but why?

src/js/common/components/RequiredLoginView.jsx
U-DON: Going to do more refactoring to address this in separate PR. The sign-in nav and verify components really can be separated into their own components.

src/js/common/components/RequiredLoginView.jsx
```diff
@@ -8,111 +8,113 @@ import LoadingIndicator from '../../common/components/LoadingIndicator';
...
+      const attemptingAppealsAccess = Array.isArray(serviceRequired) ?
+        serviceRequired.includes('appeals-status') :
+        serviceRequired === 'appeals-status';
+
+      if (user.profile.status === 'SERVER_ERROR' && !attemptingAppealsAccess) {
```
U-DON: If the availability of certain services are predicated on being able to access MVI data, perhaps we shouldn't even be checking this here. Ideally, if the API response can be granular enough to inform us that service X is unavailable for Y reason, the F/E won't have to gather separate pieces of data and interpret why the user can't access this service (whether it's right now or ever). @saneshark

dhall (commented):
This LGTM but I'll defer to @saneshark for approval

src/js/common/components/RequiredLoginView.jsx
jbalboni: Maybe @cvalarida?

U-DON:
Removed the pause from the test and still works on Jenkins... Boggles the mind.

src/js/common/components/RequiredLoginView.jsx
cvalarida: :thinking: After looking around the code for a bit, I don't know why this was needed for the first iteration of appeals status, so I can't speak confidently to v2. I do know when we make a request to `/v0/appeals`, we send along the user token so when vets-api makes a request to Caseflow, it can send along the appropriate user information. To me, that sounds like we'd need the profile, but this is also how the first iteration worked, so... :man_shrugging: 

@bshyong Git blame says you made this check for the first version of appeals status? Do you have any insight on why it was originally necessary? (Or can you point us to who would have that insight?)

jbalboni (approved):
This looks good to me. Tested it out locally, too, without any issues.

src/js/common/components/RequiredLoginView.jsx
bshyong: @cvalarida I don't remember why this was originally needed; feel free to remove this check if it's irrelevant now

saneshark (approved):
LGTM

src/js/common/components/RequiredLoginView.jsx
saneshark: I agree. I'm not a huge fan of doing this in the login view component. Should be based on available services.

PR #7481, opened 2018-03-22
6568 button needs exit icon

dhall:


dhall:
I didn't checkout master before I branched.. I'll pull changes unrelated to this issue out this PR

U-DON (approved):
Looks good, but there's a more general solution for this by adding a CSS rule `usa-button[rel="external"], usa-button[href^="http"]` with the exit-icon-white `background-image` and playing around with the `background-position` and `padding-right`. I noticed VIC uses a similar pattern as you implemented, but it'd be nice to generalize the style.

U-DON:
As an example, I get something like this with these styles (quick test from dev tools in the browser):

```
background-image: url(/img/icons/exit-icon-white.png);
background-position: right 1.2em center;
padding-right: 2.9em;
```

> screen shot 2018-03-23 at 2 55 36 am

src/js/discharge-wizard/containers/GuidancePage.jsx
```diff
@@ -146,7 +146,7 @@ class GuidancePage extends React.Component {
             Important tips for completing Form {form.num}:
           </p>
           {this.props.formValues['4_reason'] === '8' ? dd214Tips : nonDd2014Tips}
-          <a target="_blank" href={form.link} className="usa-button-primary va-button" ref={(el) => { this.downloadFormBtn = el; }}>Download Form {form.num}</a>
+          <a target="_blank" href={form.link} className="usa-button-primary va-button" ref={(el) => { this.downloadFormBtn = el; }}>Download Form {form.num}<span className="exit-icon"></span></a>
```
U-DON: I noticed `va-button` class is used in a few places in this app. It's not a class that is defined anywhere, and I think it was meant to be `usa-button`.

dhall:
Thanks @U-DON I was looking through the app to see how it had been done in other places -- thanks for steering me to the more general approach to it, I like that better as well.

PR #7483, opened 2018-03-22
9174 526 transform treatments

cehsu:
These changes create a helper to nest and flatten 526 form data.


src/js/disability-benefits/526EZ/helpers.jsx
```diff
@@ -6,8 +6,41 @@ import { stateRequiredCountries } from '../../common/schemaform/definitions/addr
...
+        formData[sibling] = [];
+        formData[sibling][idx] = disability[sibling];
+        delete disability[sibling]; // eslint-disable-line no-param-reassign
+      }
+    });
+  });
+  return formData;
+}
+
+export function nest(data) {
```
cvalarida: What is this used for?

src/js/disability-benefits/526EZ/helpers.jsx
```diff
@@ -6,8 +6,41 @@ import { stateRequiredCountries } from '../../common/schemaform/definitions/addr
...
 
 
+const siblings = ['treatments', 'privateRecordReleases', 'privateRecords', 'additionalDocuments'];
+
+export function flatten(data) {
+  const formData = Object.assign({}, data);
+  formData.disabilities.forEach((disability, idx) => {
+    siblings.forEach(sibling => {
+      if (disability[sibling]) {
+        formData[sibling] = [];
```
cvalarida: For now, initializing them all to arrays is fine because they all _are_ arrays, but I don't know if that will always be true.
Maybe do something like `formData[sibling] = Array.isArray(disability[sibling]) ? [] : {}` or add a comment explaining that every sibling is an array.

src/js/disability-benefits/526EZ/helpers.jsx
```diff
@@ -6,8 +6,41 @@ import { stateRequiredCountries } from '../../common/schemaform/definitions/addr
 import { transformForSubmit } from '../../common/schemaform/helpers';
 
 
+const siblings = ['treatments', 'privateRecordReleases', 'privateRecords', 'additionalDocuments'];
+
+export function flatten(data) {
+  const formData = Object.assign({}, data);
```
cvalarida: This actually only does a shallow clone, so the references to all properties of `data` of type `Object` (including arrays because JavaScript) will be the same. This means when you modify `formData.disabilities`, it's also modifying `data.disabilities`, making this function "impure".

Not sure how much it _really_ matters, but if you want to make it pure, you can use the deep clone from `src/js/common/utils/data-utils/cloneDeep.js` (to avoid lodash dependency :wink: ). Alternatively, `set` and `omit` from `data-utils` will probably do the job as well, but...I don't know whether that's a better approach than just deep cloning the thing.

src/js/disability-benefits/526EZ/helpers.jsx
```diff
@@ -6,8 +6,41 @@ import { stateRequiredCountries } from '../../common/schemaform/definitions/addr
...
+      });
+    }
+  });
+  return formData;
+}
+
 export function transform(formConfig, form) {
-  const formData = transformForSubmit(formConfig, form);
+  let formData = transformForSubmit(formConfig, form);
+  formData = flatten(formData);
```
cvalarida: What do you think of `const formData = flatten(transformForSubmit(formConfig, form));`?
Don't make no nevermind to me, but I thought I'd ask. :man_shrugging:

cvalarida (approved):
I had a few non-blocking comments and questions, but it looks good. :+1:

src/js/disability-benefits/526EZ/helpers.jsx
cehsu: Good catch, removed.

src/js/disability-benefits/526EZ/helpers.jsx
cehsu: Sounds good, I've added a comment since this seems like it's specifically oriented towards managing arrays for now, if that changes in the future, the recommended change makes sense.

src/js/disability-benefits/526EZ/helpers.jsx
cehsu: Good point, will use reduce to build this instead.

src/js/disability-benefits/526EZ/helpers.jsx
cehsu: Good idea, combining these function calls is easier to read (I'll double check that there is precedent for this). Can you imagine splitting it up facilitating debugging?

cehsu: `\(\w*\(\w*\)\)+` indicates this is a very common pattern, thanks for the heads up, have combined onto one line.

src/js/disability-benefits/526EZ/helpers.jsx
cehsu: Good idea, will use the common data-utils instead.

PR #7485, opened 2018-03-22
Personalization: Integrate w/Service History API route; remove profile picture

ncksllvn:
# Service History
@bshyong log in as `vets.gov.user+42@gmail.com`

# Hero w/o profile picture
image


src/js/personalization/user-profile-beta/actions/index.js
```diff
@@ -53,7 +53,8 @@ async function sendProfileRequests() {
     ['email', '/profile/email'],
     ['primaryTelephone', '/profile/primary_phone'],
     ['alternateTelephone', '/profile/alternate_phone'],
-    ['mailingAddress', '/profile/mailing_address']
+    ['mailingAddress', '/profile/mailing_address'],
+    ['serviceHistory', '/profile/service_history']
```
ncksllvn: The main point here is that anytime you had a new element to that array, it will become part of the profile object. The first element of each array is the property name, and the second is the corresponding API route.

bshyong (approved):
LGTM. @ncksllvn are we at a good enough point to start writing tests for the profile functionality?

ncksllvn:
@bshyong We're really really close. Only thing is with updating - when you use the modal to update a field, it stores that only in state, not the Redux store, until the form is submitted. Outside of vets-website this is how I've done things, but the general flow of vets-website is to keep everything in the store. What do you think?

ncksllvn:
Also, gonna go ahead and merge this so that things are in sync on the dev branch for the meeting in a bit.

bshyong:
@ncksllvn Personally I would use redux stores to be consistent. This is a good question to raise in the FE channel to get other feedback and rationale.

PR #7486, opened 2018-03-23
Unit Test Watch

rianfowler:
- watches for unit tests 
- uses nyan reporter for unit tests (can switch to any of [these](https://mochajs.org/#reporters))
- runs unit tests and uses require cache to build src to unit test map (every src import is mapped to the unit tests that import it- if you update a src file the unit tests associated will run)
- unit tests are run on a forked process 
- changes taht occur during a unit test are queued up to run when the current tests are finished 

Todo

- [x] Update readme
- [x] Update npm script `test:watch` to use this file
- [x] clean up stuff (remove options)

screen shot 2018-03-27 at 13 05 28


script/watch.js
```diff
@@ -0,0 +1,103 @@
...
+    current.messages.forEach((message) => {
+      acc.push(chalk.cyan(`  line ${message.line} `));
+      acc.push(chalk.red(`${message.message}\n`));
+    });
+    return acc;
+  }, [])
+};
+
+async function getChangedFiles() {
+  const gitDiffIndex = await exec('git diff-index --name-only HEAD');
```
rianfowler: Tried a couple of different ways but this seemed to be the easiest to manage. Wanted to clear the console before each test so that the information is always fresh.

script/watch.js
```diff
@@ -0,0 +1,103 @@
...
+}
+
+async function runLints() {
+  console.log('runLints');
+  if (running) {
+    restart = true;
+    return;
+  } else {
+    running = true;
+    restart = false;
```
rianfowler: Will probably ditch this for debouncing. Haven't tested with changing branches yet- I don't want this to blow up with a bunch of recursive calls to git after a branch change.

script/watch.js
```diff
@@ -0,0 +1,103 @@
...
+const sasslinter = require('sass-lint');
+
+const chalk = require('chalk');
+const path = require('path');
+const srcWatcher = chokidar.watch(['./src/**/*.jsx', './src/**/*.js', './src/**/*.scss']);
+
+let running = false;
+let restart = false;
+
+function formatter(eslintResults) {
```
rianfowler: Only including errors and their file name, line number, and error message. Can include whatever and format it however. 
screen shot 2018-03-23 at 00 08 22

package.json
```diff
@@ -79,6 +79,7 @@
     "body-parser": "^1.15.2",
     "chai": "^4.1.1",
     "chai-as-promised": "^7.1.1",
+    "chokidar": "^2.0.2",
```
rianfowler: [Everything](https://github.com/paulmillr/chokidar) uses this for file watching.

jbalboni:
@rianfowler For some reason, this isn't detecting anything for me. It sits with "Change detected- running lints 10:40:41" at the top of the screen.

rianfowler:
Guess it would help if I don't remove the watcher callback.

test/pensions/config/childrenInformation.unit.spec.jsx
```diff
@@ -73,7 +73,7 @@ describe('Child information page', () => {
     formDOM.submitForm(form);
     const errors = formDOM.querySelectorAll('.usa-input-error-label');
 
-    errors.forEach(e => console.log(e.getAttribute('for'))); // eslint-disable-line no-console
+    // errors.forEach(e => console.log(e.getAttribute('for'))); // eslint-disable-line no-console
```
rianfowler: Dug into the mocha reporter and there was not an easy way to stop logs from interfering with the reporter without filtering the `stderr` or `stdout` streams. I divert the `stderr` stream while mocha is running- there were too many errors for me to fix easily right now. This was the only place where `stdout` was interfering with the reporter.

These logs don't feed into the reporter and failing tests have enough information on their own to debug most scenarios- we should probably remove this.

script/watch.js
```diff
@@ -0,0 +1,170 @@
...
+  // use gitDiffIndex to limit tests that are run to changed files
+  const gitDiffIndex = await exec('git diff-index --name-only HEAD');
+  return gitDiffIndex.stdout.split('\n').map(file => path.join(projectRoot, file));
+}
+
+async function runMochaTests(files) {
+  // keep errors from polluting the reporter by redirecting the error stream
+  let errorStream;
+  const consoleErr = process.stderr.write;
+  if (!options.showErrors) {
```
rianfowler: `showErrors` flag will prevent diverting `stderr` stream. Should probably see if we can catch these errors in the unit tests so this is unnecessary.

script/watch.js
```diff
@@ -0,0 +1,170 @@
...
+async function runTests() {
+  const groupedFiles = groupFilesByTest(await getChangedFiles());
+
+  if (options.unit) {
+    const unitTests = groupedFiles.mocha.length > 0 ?
+      groupedFiles.mocha :
+      await glob('test/**/*.unit.spec.js?(x)');
+    console.log(`Running Mocha with ${unitTests.length} Files\n`);
+    // mocha requires flushing the require cache of all js files (tests and imports)
+    groupedFiles.eslint.forEach(file => delete require.cache[file]);
```
rianfowler: This is something I don't fully understand. I think because `mocha` requires the test and SUT files, then these files have to be flushed out of the node `require` cache so that `mocha` can run again. Without doing this, `mocha` only runs once or doesn't run with updated files.

script/watch.js
```diff
@@ -0,0 +1,170 @@
...
+  if (options.sasslint) {
+    const sasslintTargets = groupedFiles.sasslint.length > 0 ?
+      groupedFiles.sasslint :
+      await glob('./**/*.scss');
+
+    console.log(`Running sassLint with ${sasslintTargets.length} Files\n`);
+    sasslintFiles(sasslintTargets);
+  }
+
+  console.log(chalk.cyan('Watching for changes'));
```
rianfowler: Decided clearing the terminal history was rude.

rianfowler:
@jbalboni added a bunch of updates. I tested it on a test and SUT file and it seems to work well.

script/watch.js
```diff
@@ -0,0 +1,170 @@
...
+  './src/**/*.jsx',
+  './src/**/*.js',
+  './test/**/*.unit.spec.*'
+];
+
+if (options.sassLint) {
+  watchArray.push('./src/**/*.scss');
+}
+
+const srcWatcher = chokidar.watch(watchArray);
```
jbalboni: I wonder if something like this is doable? https://github.com/mochajs/mocha/issues/2763

jbalboni: Though I suppose that wouldn't tell us which tests to re-run.

script/watch.js
```diff
@@ -0,0 +1,170 @@
...
+  ));
+  console.log(formatter(sasslintResults) || chalk.green(`${files.length} scss files lint free`));
+}
+
+// ensures errors that blow up test watcher make it to console reporter
+process.on('uncaughtException', function(err) {
+  console.error((err && err.stack) ? err.stack : err);
+});
+srcWatcher.on('change', runTests);
+runTests();
```
jbalboni: It seems like it's hanging with "Running Mocha with 486 Files" when it first starts up. Should it be running all the tests, or do I need to change a file first?

script/watch.js
rianfowler: This is refining the `delete` approach to only uncache the files that have changed. This is cleaner and I like it more- right now there's no direct link between what mocha requires and what gets cached- I'm relying on using similar glob patterns and executing these two tasks independently. This would explicitly track what mocha is requiring and only decache those files.

There's something in this example about doing it recursively that I don't understand. I was just deleting the file path key from the require cache. 👀

rianfowler: What is this syntax? it's not CoffeeScript. 🙄

script/watch.js
rianfowler: It should run all of the tests when it starts up. It does hang for a few seconds while mocha requires and babelifies all of the files. I can provide more feedback or put the initial run behind a flag- this is a personal preference for me. I like to see if there are any failing tests when I start the watch task.

script/watch.js
```diff
@@ -0,0 +1,129 @@
...
+    const requiredSourceFiles = require.cache[testFile]
+      .children
+      .forEach(childModule => {
+        if (!childModule.id.includes('src')) {
+          return;
+        }
+        if (acc[childModule.id]) {
+          acc[childModule.id].push(testFile);
+        } else {
+          acc[childModule.id] = [testFile];
```
rianfowler: each src file will contain an array of unit test files that required it.

script/watch.js
```diff
@@ -0,0 +1,129 @@
...
+    const watchedFiles = Object.keys(require.cache).filter(file => !unwatchedModules.includes(file) && !file.includes('node_modules'));
+    const unitTestsForSrc = getUnitTestsForSrc(watchedFiles);
+    watcher = chokidar.watch(watchedFiles).
+      on('change', file => {
+        if (file.includes('.unit.spec')) {
+          uncache(file);
+          runUnitTests([file]);
+        }
+
+        if (file.includes('src')) {
```
rianfowler: If a unit test is updated, just run that unit test. if a src file is updated, run all the unit tests associated with it.

This will probably go crazy if many files are saved at once but I figure if that's a common scenario we can update this to accommodate.

script/watch.js
```diff
@@ -0,0 +1,129 @@
...
+
+async function runUnitTests(tests) {
+  const unitTests = tests || await glob('test/**/*.unit.spec.js?(x)')
+  console.log(chalk.cyan(`Running ${unitTests.length} unit test files`));
+
+  await runMochaTests(unitTests);
+
+  console.log(chalk.red('Watching for changes'));
+
+  if (!watcher) {
```
rianfowler: Will have to restart the watcher if a new file is added.

rianfowler:
@jbalboni let me know if this works for you now. I removed the linting- just going to add that to my vim. Using the require cache is a much cleaner way of doing this.

jbalboni:
@rianfowler Yes, it does! I was able to change the CropperController component and see the two related tests re-run.

I ran into an issue where if I change `src/js/common/utils/address.js`, I get a chokidar error. Not sure why.

rianfowler:
I'll check it out.

rianfowler:
@jbalboni This should be fixed now- I wasn't checking for the existence of unit tests before trying to run them. Now it notifies when a change is detected but no unit tests were found.

screen shot 2018-03-26 at 18 20 49

jbalboni:
Works now. It does seem to catch just the first level of imports. Changing address.js can break a bunch of unit tests, but we don't have any tests that import it directly. I'm not sure if that's possible to catch (the require cache stuff is a bit beyond my Node understanding at the moment).

rianfowler:
Might be best if I just uncache all of the js files every time. I'll try a couple of things.

rianfowler:
iu-1
Well I found it's leaking memory and blows up after three runs. 🙄 Gonna try using `exec` = hopefully that's where the leak is.

rianfowler:
@jbalboni This should work now- it's a little slower than before. I couldn't figure out why mocha wasn't getting flushed out of memory- suspect it has something to do with `babel-register`. I went with the nuclear option and just forked the entire mocha process. Now when it's done I just kill the process altogether. 

Also tweaked the watch so if files are updated during a test run, a subsequent test run is queued up. Only thing I didn't work around was the initial run- it doesn't watch during the initial run since I need mocha to run at least once to know what src files are linked to what unit tests.

For files that don't have specific unit tests, the whole suite will rerun when they change. e.g. if you update that address utility, all tests will run. There is a way with the `require.cache` to build a dependency tree that could be used to only run the tests for the src files that import this utility but it would require some fancy data structuring.

test/pensions/config/childrenInformation.unit.spec.jsx
jbalboni: That seems reasonable to me.

jbalboni (approved):
This looks good to me. I'm able to make changes and things re-run as expected. I didn't get too in depth on the scripts, but they look fine, too. Should definitely make writing tests a nicer experience.

PR #7487, opened 2018-03-23
use relative paths for internal links so they do not show with exit-icon

dhall:
There is an overarching css rule that gives any link that starts with ^http an exit icon. Since these resources are relative to the project I turned them into relative links. I search the codebase and ensured that relative links are used in other parts of the application but if there is a reason not to take this approach, let me know and I'll do something different.


U-DON (approved):
LGTM. I think this is the best way to go too.

PR #7495, opened 2018-03-28
migrate design system components

rianfowler:
This is kind of a cheesy way to do this but we could just swap in all of the jean-pants components at their src files and test them all at once. Some of these are hard to get to- like this AlertBox. I'm not sure how to test this with actual data- I was able to force it to render by modifying some components. 

- [x] Add [ticket](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/9609) to update these components in the design system:
    - DropDown
    - ErrorableRadioButton
    - ErrorableSelect
    - LoadingIndicator
    - Modal
    - SearchMenu
    - Tooltip
    - ErrorableTextInput

- [x] Organize [this](https://docs.google.com/spreadsheets/d/1BrMkVgS8KKWwnEsH3BcDdpn82e4zAhWV3KoWNDFR4A0/edit#gid=0) into apps that need to be tested and document what should be validated in each app

Test
- [x] AlertBox
- [ ] ~LoadingIndicator~
- [x] OMBInfo
- [x] SegmentedProgressBar
- [x] SystemDownView
- [ ] ~ErrorableCheckbox~
- [x] ExpandingGroup
- [x] ProgressButton
- [x] IconHelp


rianfowler:
@jbalboni I could swap in all of the components and document all of the places they appear as per the code. Then we could come up with a test plan for verifying that they work? I would rather try to do it all at once than try to validate each component separately.

jbalboni:
Yeah, maybe that's the better path. We can get a few others to help us test things out.

We also need to verify that there haven't been updates to each component since moving it over.

rianfowler:
I'll put that together today and review the components for changes since v1.

rianfowler:
@jbalboni I updated the summary with components I was able to verify on at least one form. I'm blocked on verifying the components that are behind a login. I'm not sure how to log in with the dev version of vets-api. I was able to go through the ID.me flow but got an error (documented in the test document).

jbalboni:
@rianfowler For the rx one, what user did you try? It looks like vets.gov.user+264@gmail.com is the one you should use (based on https://github.com/department-of-veterans-affairs/vets-api-mockdata/blob/master/mock_data_table.md). Also you should make sure you have betamocks set up correctly.

The GIBCT I think has a separate data service that needs to be run, but I'm not sure how that works, exactly. We may need to ask in the FE or engineers channel and get some help there.

rianfowler:
I was able to test everything in at least one place but the system down view. I'm not sure how to make that visible even manually forcing it by changing the code.

rianfowler:
I'm breaking the parts up into separate commits. Will document on this comment: 

[AlertBox](https://github.com/department-of-veterans-affairs/vets-website/pull/7495/commits/f1e7bac967b5888d2b30ebffc8a4e6b428cb25ed)

rianfowler:
@jbalboni So what do I need to do if I want to test this staging? Can I deploy just this branch to an environment and test it?

jbalboni:
To test in staging, you will need to merge to master. If you haven't, I would test it out in the review instance for this PR first (there are View Deployment buttons in the timeline for this PR, pick the one that isn't Heroku). And if there are specific things you haven't been able to test locally, then let's find someone to check that before merging.

jbalboni (approved):

rianfowler:
Hadn't noticed the `View Deployment` link- that's handy.

PR #7505, opened 2018-03-30
Improve accuracy of progress bar for Health Records updates

U-DON:
Resolves department-of-veterans-affairs/vets.gov-team#9576.


dhall (commented):
This LGTM, but I'll defer to @saneshark for approval

saneshark (approved):
@U-DON since this is extremely difficult to test on staging could we sync for 5-10 minutes today and walk me through the logic for the reducer before merging? 

I looked it over and LGTM. Just want to double check before merging.

PR #7508, opened 2018-03-30
Change location of fields and reviewFields in SchemaForm

edpaget:
avoids a random test order error where 'fields' is modified causing `ArrayField`
to be set to undefined causing subsequent tests to error

One source of inexplicable build failures down!

### DIFF
--- a/script/run-unit-test.sh
+++ b/script/run-unit-test.sh
@@ -10,7 +10,7 @@ fi
 
 # No test specified; run all of them
 if [[ -z "$1" ]] || [[ $1 == -* ]]; then
-  BABEL_ENV=test $MOCHA test/helper.js --recursive '$SCRIPT_DIR/../test/**/*.unit.spec.js?(x)' "$@"
+    BABEL_ENV=test $MOCHA --recursive '$SCRIPT_DIR/../test/**/*.unit.spec.js?(x)' test/helper.js "$@"
   exit $?
 fi
 
--- a/src/js/common/schemaform/components/SchemaForm.jsx
+++ b/src/js/common/schemaform/components/SchemaForm.jsx
@@ -18,21 +18,6 @@ import TitleField from '../fields/TitleField';
 import ReviewObjectField from '../review/ObjectField';
 import { scrollToFirstError } from '../../utils/helpers';
 
-const fields = {
-  ObjectField,
-  ArrayField,
-  BasicArrayField,
-  TitleField
-};
-
-const reviewFields = {
-  ObjectField: ReviewObjectField,
-  ArrayField: ReadOnlyArrayField,
-  BasicArrayField,
-  address: ReviewObjectField,
-  StringField
-};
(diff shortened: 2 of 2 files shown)


src/js/common/schemaform/components/SchemaForm.jsx
```diff
@@ -136,6 +121,21 @@ class SchemaForm extends React.Component {
   }
 
   render() {
+    const fields = {
```
jbalboni: Is it possible to modify the tests instead of moving these? This component is re-rendered a lot when you use a form.

edpaget: I was wrong when I said fields was modified. The Issue here is actually the way node/babel resolves circular dependencies. 

The random test order means that when the `ArrayField` spec is run before a spec that uses `SchemaForm`, `ArrayField` is imported and `SchemaForm` is imported somewhere down that load chain. Since `ArrayField` hasn't finished loading it is `undefined` at the time node evaluates the `SchemaForm` file when the `fields` and `reviewFields` objects are created.

Moving the object definitions inside `render` means they aren't evaluated until the object is instantiated, so `ArrayField` will have finished loading by that point. 

I don't think there's anyway around this that leaves these object defined only once outside the class body. They don't need to be in the render method and so could be somewhere else if you think it'd be more faster.

cvalarida: Could we solve both problems (test failures and efficiency) by making these properties of the `SchemaForm` class?

jbalboni: Yeah, I think that's probably the way to go. We can create these objects in the constructor, then. We're not creating lots of SchemaForm components usually, so I don't imagine that will have much impact.

edpaget: Sounds like a plan I updated to have them assigned as `this.fields` and `this.reviewFields` in the constructor.

script/run-unit-test.sh
```diff
@@ -10,7 +10,7 @@ fi
 
 # No test specified; run all of them
 if [[ -z "$1" ]] || [[ $1 == -* ]]; then
-  BABEL_ENV=test $MOCHA test/helper.js --recursive '$SCRIPT_DIR/../test/**/*.unit.spec.js?(x)' "$@"
+    BABEL_ENV=test $MOCHA --recursive '$SCRIPT_DIR/../test/**/*.unit.spec.js?(x)' test/helper.js "$@"
```
jbalboni: What's this change for?

jbalboni (approved):

script/run-unit-test.sh
edpaget: File need to be loaded in the same order as they are in jenkins so when choma shuffles the files with the seed that failed on Jenkins the tests will actually get run in that order.

PR #7511, opened 2018-03-30
Personalization: Fix existing unit and E2E tests

ncksllvn:
Also, moves the logic from `<DashboardRedirect/>` into a more sensible location


ncksllvn:
This was branched off another feature branch, with a PR already open - department-of-veterans-affairs/vets-website/pull/7509

PR #7525, opened 2018-04-03
Refactored health tools to accommodate services list affected by new MHV auth policies

U-DON:
Connected to department-of-veterans-affairs/vets-api#1765.

Because the services list will now accurately represent the actual accessible services for a given MHV account level (e.g., Basic account can't access Rx and Messaging), we need to update our gating logic for MHV apps that shows an error message when the user shouldn't be able to access the service. Previously, we only cared about whether the user had an existing or upgraded account. We still care about that when it involves account creation, but when it comes to determining access, we are going to base that entirely on the services list.

Currently on production, all MHV services are still available to any MHV account level, so this F/E change won't affect that and can be merged independently of the associated B/E change.


src/js/common/containers/MHVApp.jsx
```diff
@@ -91,33 +93,28 @@ export class MHVApp extends React.Component {
...
-
-    const view = React.Children.map(this.props.children,
-      (child) => {
-        let props = null;
-        if (typeof child.type === 'function') {
-          props = { isDataAvailable: enabled };
-        }
-        return React.cloneElement(child, props);
-      }
-    );
```
U-DON: We've determined previously that this `isDataAvailable` prop isn't being propagated properly due to the `DowntimeNotification` component not propagating it one level above. It doesn't matter anyway because this was related to doing the check for the services within `RequiredLoginView`, but we are checking that and handling it in `MHVApp` for these apps.

src/js/common/containers/MHVApp.jsx
```diff
@@ -91,33 +93,28 @@ export class MHVApp extends React.Component {
       return <AcceptTermsPrompt terms={terms} cancelPath="/health-care/" onAccept={this.props.acceptTerms}/>;
     }
 
-    if (!this.isAccessible()) {
+    if (!this.hasService()) {
```
U-DON: Made a distinction here where `hasAccount` determines whether we should attempt account creation, and `hasService` determines whether the user has access.

src/js/health-records/containers/HealthRecordsApp.jsx
```diff
@@ -6,44 +6,32 @@ import DowntimeNotification, { services } from '../../common/containers/Downtime
...
       <RequiredLoginView
         verify
-        serviceRequired="health-records"
+        serviceRequired={SERVICE_REQUIRED}
         user={this.props.user}>
         <DowntimeNotification appTitle="health records tool" dependencies={[services.mhv]}>
           <AppContent>
             <Breadcrumbs location={this.props.location}/>
-            <MHVApp>
+            <MHVApp serviceRequired={SERVICE_REQUIRED}>
```
U-DON: If I end up implementing passing error handler components to `RequiredLoginView`, we could pass `MHVApp` as that prop and inherit the `serviceRequired` prop from it. At which point we would only care about any of the account creation rigamarole if and only if the required service was not available in the user's list of available services in the first place.

src/js/common/containers/MHVApp.jsx
```diff
@@ -42,7 +42,9 @@ export class MHVApp extends React.Component {
 
   needsTermsAcceptance = () => this.props.account.state === 'needs_terms_acceptance';
 
-  isAccessible = () => ['existing', 'upgraded'].includes(this.props.account.state);
+  hasAccount = () => ['existing', 'upgraded'].includes(this.props.account.state);
```
U-DON: If we get rid of the `mhv_account_state` stuff altogether, we could (1) base eligibility for account creation in `MHVApp` on the existence of `mhvac` in the services list and (2) determine existence of an account from the MHV account level (`nil` for no account, otherwise string value) if it gets exposed in the user object.

dhall (commented):
LGTM, but I'll defer to @saneshark for approval

src/js/common/containers/MHVApp.jsx
saneshark: Maybe mhv_account_state could eventually be equivalent to the account type, then we wouldn't need to track it elsewhere, seeing as we have a pretty reliable means of determining it now. But yeah, I agree and would prefer not to persist anything to postgres eventually that isn't just a log of what action vets.gov has taken, such as creating or upgrading.

saneshark (commented):
Before signing off on this. Has this been thoroughly tested locally?

1. do the appropriate errors show for various account levels?
2. does account creation appear to still work as expected.

U-DON:
Seems to work as expected from local testing.

1. With account state of `existing` and removing the MHV service being accessed from `services`, I get the generic MHV error.
2. With account state of `unknown` or `registered`, account creation is attempted.
3. With account state of `existing` and with the MHV services available in `services`, the app is accessible.

saneshark (approved):
LGTM, now would be a good time to merge possibly since there is no risk of this making it into production without further testing on staging.

saneshark:
@briangryth with respect to the QA / Regression testing PR. This would be a prime candidate to some regression testing around once merged. CC: @NatalieEMoore

PR #7527, opened 2018-04-04
Re-enable the unauthed flow test

cvalarida:
Re-enables the commented out test. Looks like the other test that was disabled has already been re-enabled.


markgreenburg:
Oh? Do we know when the other test was re-enabled, and if it went live before GIBS went back up, why it didn't break? Just trying to make sure this test (and the other test that's already live again) works as expected.

markgreenburg (approved):
LGTM, looks like `/status` points to the actual app

cehsu (approved):
LGTM.

PR #7529, opened 2018-04-04
9644 fix object keys null error

rianfowler:
Assumption: if the formData object is invalid then 
`newState.prefillStatus = PREFILL_STATUSES.unfilled;`

This error doesn't look like it's happening in production anymore.

screen shot 2018-04-04 at 12 10 47

### DIFF
--- a/src/js/common/schemaform/save-in-progress/reducers.js
+++ b/src/js/common/schemaform/save-in-progress/reducers.js
@@ -74,7 +74,7 @@ export const saveInProgressReducers = {
 
       // We get an empty object back when we attempt to prefill and there's
       // no information
-      if (Object.keys(action.data.formData).length > 0) {
+      if (_.keys(_.get('data.formData', action)).length > 0) {
         newState.prefillStatus = PREFILL_STATUSES.success;
       } else {
         newState.prefillStatus = PREFILL_STATUSES.unfilled;
--- a/src/js/common/schemaform/save-in-progress/reducers.js
+++ b/src/js/common/schemaform/save-in-progress/reducers.js
@@ -74,7 +74,7 @@ export const saveInProgressReducers = {
 
       // We get an empty object back when we attempt to prefill and there's
       // no information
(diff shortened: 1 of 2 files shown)


src/js/common/schemaform/save-in-progress/reducers.js
```diff
@@ -74,7 +74,7 @@ export const saveInProgressReducers = {
 
       // We get an empty object back when we attempt to prefill and there's
       // no information
-      if (Object.keys(action.data.formData).length > 0) {
+      if (_.keys(_.get('data.formData', action)).length > 0) {
```
jbalboni: Are there advantages to this compared to `Object.keys(action.data.formData || {}).length > 0`?

jbalboni:
@rianfowler That assumption looks correct to me.

jbalboni (approved):
One question, but LGTM.

src/js/common/schemaform/save-in-progress/reducers.js
rianfowler: `_.keys` won't blow up if it's not an object like `undefined`; It returns an empty array. But it has a side effect of co-ercing other types into objects.

rianfowler: I looked again and the specific error is `Object.keys` throw is `TypeError: Cannot convert undefined or null to object`. `_.keys` returns an empty array when called on `undefined` or `null`. 

This can set `formData` to an empty object by default further up the chain (before the action creator is called and after the prefill and migrate functions are called). 

Not sure what the best approach is- in the past I've handled incomplete or bad data from an action in the reducer.

jbalboni: I don't mind handling it here, I just want to make sure there's a good advantage to using lodash methods over the Object.keys. I don't think we necessarily need the `get`, since the error is that undefined or null is passed to Object.keys. `_.keys` does seem less brittle, so I think that's ok.

PR #7530, opened 2018-04-04
Personalization:  Implement routes for updating fields

ncksllvn:
Probably needs some cleaning up, but this -

1. Updates the containers that use the RequiredLoginView component to the new props
2. Removes the mock function for Profile POST routes, and adds the actual API route.
3. Adds error handling for failed updates and an action for clearing that error


ncksllvn:
Jenkins should check out - the reason it doesn't right now is because an old version of Node is cached for this branch.

PR #7531, opened 2018-04-04
add some error handling to the mocha process

rianfowler:
Mocha doesn't catch errors during the require stage of its setup i.e. if there's a problem with an import, this was failing silently. Now the child process sends an error message when something fails allowing the watcher to 
- handle the error and keep watching if the watcher is set up
- stop the process if the watcher isn't set up.

### DIFF
--- a/script/watch.js
+++ b/script/watch.js
@@ -27,11 +27,18 @@ async function runUnitTests(unitTests = allUnitTests) {
   // watchedFiles: all of the project files that mocha required (no node_modules)
   // list of source files and their associated unit tests
   busy = true;
-  const {
-    watchedFiles,
-    unitTestsForSrc
-  } = await runMochaTests(unitTests);
-   busy = false;
+  let watchedFiles, unitTestsForSrc;
+  try {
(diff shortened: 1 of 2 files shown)


script/watch.js
```diff
@@ -78,9 +85,12 @@ function runMochaTests(tests) {
...
       showErrors
     });
-    forked.on('message', (requiredProjectFiles) => {
+    forked.on('message', (message) => {
+      if (message.error) {
+        reject(message.error);
+      }
       // sent the watcher the files to watch
-      fulfill(requiredProjectFiles);
+      fulfill(message);
```
jbalboni: The comment seems confusing now, since this isn't the files to watch? Or is it still, if there's not an error?

jbalboni (approved):
One question, but LGTM.

script/watch.js
rianfowler: This is pretty opaque. I updated the variable named to `requiredFiles` and used destructuring assignment on the message object so that's clearer what's in it. 

I don't know if this is the idiomatic way to communicate errors from a forked process. The `error` seems to be for setup problems and not something I can fire from code.

PR #7534, opened 2018-04-04
Issues 9494 9645

lbai001:
#### Proposed changes:
Discharge endDate limit -> 730 when env is not production
Discharge endDate limit is still 180 when env is production
Removed vets_application_status_slow call from ApplicationStatus.jsx
#### Files changed: 
        modified:   src/js/common/schemaform/save-in-progress/ApplicationStatus.jsx
        modified:   src/js/hca/validation.js
        modified:   test/hca/validation.unit.spec.js
#### Screenshots
screen shot 2018-04-04 at 2 52 31 pm
screen shot 2018-04-04 at 2 53 07 pm

### DIFF
--- a/src/js/hca/validation.js
+++ b/src/js/hca/validation.js
@@ -3,20 +3,27 @@ import _ from 'lodash/fp';
 import { convertToDateField, validateCurrentOrPastDate } from '../common/schemaform/validation';
 import { isValidDateRange } from '../common/utils/validations';
 
-export function validateServiceDates(errors, { lastDischargeDate, lastEntryDate }, { veteranDateOfBirth }) {
+function calculateEndDate() {
   let endDateLimit;
   if (process.env.BUILDTYPE === 'production') {
     endDateLimit = 180;
   } else {
     endDateLimit = 730;
   }
+  return {
+    endDateLimit,
+    endDate: moment().endOf('day').add(endDateLimit, 'days')
+  };
+}
+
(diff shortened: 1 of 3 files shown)


src/js/hca/validation.js
```diff
@@ -3,14 +3,27 @@ import _ from 'lodash/fp';
 import { convertToDateField, validateCurrentOrPastDate } from '../common/schemaform/validation';
 import { isValidDateRange } from '../common/utils/validations';
 
+function calculateEndDate() {
+  let endDateLimit;
+  if (process.env.BUILDTYPE === 'production') {
```
jbalboni: We normally use `__BUILDTYPE__` instead of `process.env.BUILDTYPE`

jbalboni (commented):
One small comment about the build type flag, but otherwise the code looks good. I want to have Peggy approve the message before merging this.

src/js/hca/validation.js
lbai001: ok

rtluu:
@kellyskennedy @DanielleSoCompany, @peggygannon is OOO until next week and our team was wondering if either of you could give us some quick feedback regarding how we should describe 730 days to a veterans. We're thinking '2 years' or '24 months'. What do you think?

DanielleSoCompany:
Hi @rtluu sure, happy to help! We usually go with 2 years. Thanks!

jbalboni (approved):

PR #7533, opened 2018-04-04
Create platform directory and simplify application startup code

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets-contrib/issues/22

Notes:
- I created the platform directory in src because that will cause fewer path changes down the road. We can move it up a directory later (or leave it, depending on how people feel)
- I tried to move as little as possible, but it's still a large-ish PR
- Most of the entry files fit in the simpler startApp mode, some of them do different things with history or routing and drop down a level of abstraction and use the functions that startApp uses.

PR #7536, opened 2018-04-05
Add documentation for generator questions and set up.

jbalboni:
https://github.com/department-of-veterans-affairs/vets-contrib/issues/68


docs/GeneratorOptions.md
```diff
@@ -0,0 +1,70 @@
...
+
+This is the url your application will live at. In your React apps you will likely have multiple pages and the urls for those pages will have this as the base. This value also gets translated into the page for the content page, which lives in `content/pages`.
+
+### Is this a form app?
+
+If this is a form, there are some more questions after this step. If not, you're done!
+
+### What's your form number?
+
+This is the form number for the paper form you're converting. Normally it's something like 22-1990 or 21-22. This is used in a couple places in the UI and also as a key for the save in progress functionality of our forms.
```
melaniegershman: One suggestion here would be to clarify where you might find the form number on the paper form, or even include a screenshot, like so: 
screen shot 2018-04-05 at 4 40 27 pm

docs/GeneratorOptions.md
```diff
@@ -0,0 +1,70 @@
...
+
+### What's the Google Analytics event prefix you want to use?
+
+Our shared form code sends events to Google Analytics and we need a unique prefix to be able to categorize the events by form. It's normally a dash separated value like `hca-` or `edu-1990-` and is made up by developers.
+
+### What's the respondent burden of this form in minutes?
+
+Each paper form has OMB information that lives in the margins of each page (usually in the top right). This is the respondent burden value from that information. All the information from this question and the next two can be found in the form's introduction page, passed as props to the OMBInfo component, if you need to change it later.
```
melaniegershman: Is it more clear to provide the file path for the Intro page, like so?:

> passed as props to the OMBInfo component **`found in containers/IntroductionPage.jsx`**

melaniegershman (approved):
These are really helpful updates. I made a few suggestions for your consideration below. Also, will there eventually be a troubleshooting section for issues such as the one I had earlier today, with the extra `/` in `manifest.json`'s `rootUrl`?

docs/GeneratorOptions.md
melaniegershman: I see now that this has been handled further below!

jbalboni:
@melaniegershman Thanks, I've made some updates. A troubleshooting section probably makes sense, I'll have to think about what goes in there. For the url issue, we have validation in the generator that should prevent it, plus I added a check for it to another PR, so that an error shows up in the console: https://github.com/department-of-veterans-affairs/vets-website/pull/7533/files#diff-188da8af3e1e05db38e141cae9f62f25R33

melaniegershman:
Thanks @jbalboni that check & error will be super helpful!

PR #7537, opened 2018-04-05
Dragons Personalization work 

bshyong:
- includes Dashboard work
- includes Profile work
- includes Account work

All features are behind a beta registration gate will not be available in production pending completion of https://github.com/department-of-veterans-affairs/vets.gov-team/issues/9631


ncksllvn:
Just updated to the latest platform work - the app entry points have been updated

ncksllvn:
I also moved the `beta-enrollment` app into the personalization folder. Until the beta process is more refined, I think it makes sense to keep it out of the way of the root application directory. I also moved the `BetaApp` container out of the common directory and into that folder as well, so we don't have to worry about it being affected by changes with `jean-pants` and general restructuring of the common modules.

ncksllvn (approved):
🚀 🎸 🎊 🎉 🥇

PR #7539, opened 2018-04-06
9702 pension burial polling

rianfowler:


src/js/pensions/helpers.jsx
```diff
@@ -22,7 +23,60 @@ function replacer(key, value) {
...
+    .then(res => {
+      if (res.ok) {
+        return res.json();
+      }
+
+      return Promise.reject(res);
+    }).catch(res => {
+      if (res instanceof Error) {
+        Raven.captureException(res);
+        Raven.captureMessage('vets_pension_poll_client_error');
```
rianfowler: Raven error message type

rianfowler:
@jbalboni Figured I could get some review on some of the error strings now. 

I'm also not sure what to expect for local testing- currently it submits but the API returns not found with the ID that's used.

src/js/pensions/helpers.jsx
```diff
@@ -22,7 +23,60 @@ function replacer(key, value) {
...
+  setTimeout(() => {
+    checkStatus(guid)
+      .then(res => {
+        if (!res || res.data.attributes.state === 'pending') {
+          pollStatus(guid, onDone, onError);
+        } else if (res.data.attributes.state === 'success') {
+          onDone(res.data.attributes.response);
+        } else {
+          // needs to start with this string to get the right message on the form
+          throw new Error(`vets_server_error_pensions: status ${res.data.attributes.state}`);
```
rianfowler: Error type

src/js/pensions/helpers.jsx
```diff
@@ -33,6 +87,50 @@ export function transform(formConfig, form) {
...
+    pollStatus(guid, response => {
+      window.dataLayer.push({
+        event: `${formConfig.trackingPrefix}-submission-successful`,
+      });
+      return response;
+    }, error => Promise.reject(error));
+  }).catch(respOrError => {
+    if (respOrError instanceof Response) {
+      if (respOrError.status === 429) {
+        const error = new Error('vets_throttled_error_pensions');
```
rianfowler: throttled error type

jbalboni:
@rianfowler Those error types look good to me. You'll be able to test this locally once @lihanli  pushes a branch with his changes, so you might have to wait a little bit if he's still working.

rianfowler:
I'm getting `null` when I poll the id that's returned from initial `GET` to pensions.  The only thing I changed from the vic implementation is adding `mode: 'cors'` to the `fetch` call but let me know if I missed something else. @jbalboni @lihanli 

**initial response (client):** 
screen shot 2018-04-06 at 16 35 55

**initial response (server)**
screen shot 2018-04-06 at 16 36 23

**polling request (headers)**
screen shot 2018-04-06 at 16 39 23


**polling response (server)**
screen shot 2018-04-06 at 16 36 37

lihanli:
i forgot to return the uuid in the create route, when [this pr](https://github.com/department-of-veterans-affairs/vets-api/pull/1830) is merged the uuid should appear under `['data']['attributes']['guid']`

rianfowler:
Seem to be working now but I'm getting a `failed` submission every time. @lihanli 

**Create Response**
screen shot 2018-04-07 at 12 26 52

**Polling Request**
screen shot 2018-04-07 at 12 28 20

**Polling Response**
screen shot 2018-04-07 at 12 28 12

rianfowler:
Verified both forms on dev.

src/js/burials/helpers.jsx
```diff
@@ -1,8 +1,120 @@
...
+const POLLING_INTERVAL = 1000;
+
+function pollStatus(guid, onDone, onError) {
+  setTimeout(() => {
+    checkStatus(guid)
+      .then(res => {
+        if (!res || res.data.attributes.state === 'pending') {
+          pollStatus(guid, onDone, onError);
+        } else if (res.data.attributes.state === 'success') {
+          onDone(res);
```
rianfowler: This is different from VIC which queries down into the response. Both burials and pensions action want the top level of the response.

src/js/burials/helpers.jsx
```diff
@@ -1,8 +1,120 @@
...
+        } else {
+          // needs to start with this string to get the right message on the form
+          throw new Error(`vets_server_error_burial: status ${res.data.attributes.state}`);
```
rianfowler: burial error

src/js/burials/helpers.jsx
```diff
@@ -1,8 +1,120 @@
...
+          event: `${formConfig.trackingPrefix}-submission-successful`,
+        });
+        return resolve(response);
+      }, error => reject(error));
+
+    });
+  }).catch(respOrError => {
+    if (respOrError instanceof Response) {
+      if (respOrError.status === 429) {
+        const error = new Error('vets_throttled_error_burial');
```
rianfowler: burial error

src/js/burials/helpers.jsx
```diff
@@ -1,8 +1,120 @@
...
+    .then(res => {
+      if (res.ok) {
+        return res.json();
+      }
+
+      return Promise.reject(res);
+    }).catch(res => {
+      if (res instanceof Error) {
+        Raven.captureException(res);
+        Raven.captureMessage('vets_burial_poll_client_error');
```
rianfowler: burial error

rianfowler:
@jbalboni This is ready for review.

src/js/burials/helpers.jsx
jbalboni: It might be good to update the pension and burials code to be consistent with VIC. I think those changes would just be in the confirmation page?

jbalboni (commented):
Had one comment. Could we also copy over the VIC polling unit tests?

src/js/burials/helpers.jsx
rianfowler: Probably- I'll take a look at it.

rianfowler: All 3 forms fire the `schemaForm` action [`setSubmitted`](https://github.com/department-of-veterans-affairs/vets-website/blob/a879c361370236708f5123414cd0c30d70938838/src/js/common/schemaform/actions.js#L51)

This works fine for VIC which passes `res.data.attributes.response` to `setSubmitted`
screen shot 2018-04-09 at 16 53 30

But the pensions and burials polling endpoints don't have an `res.data.attributes.response` in dev. 
screen shot 2018-04-09 at 16 57 17

Theres a spot on the confirmation page for a case number so I think this is missing or we need to change the design of the confirmation page. This is the expected data based on the [ticket](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/9702#issuecomment-379082391) though.

jbalboni: @lihanli I think there's a confirmation number that the backend needs to pass back, right?

rianfowler:
Added the polling unit tests to burials and pensions.

src/js/burials/helpers.jsx
lihanli: the confirmation number is passed on the form create

src/js/burials/helpers.jsx
```diff
@@ -1,8 +1,120 @@
...
+      }
+
+      // if we get here, it's likely that we hit a server error
+      return Promise.reject(res);
+    });
+}
+
+const POLLING_INTERVAL = 1000;
+
+function pollStatus({ guid, confirmationNumber, regionalOffice }, onDone, onError) {
```
rianfowler: Opted to just pass the values from the previous calls to `pollStatus`. I think when we refactor this, I would like to have the the response built from the different calls in a single closure so it's easier to see that we're building a complete response from multiple calls.

src/js/pensions/config/form.js
```diff
@@ -176,9 +176,8 @@ function createSpouseLabelSelector(nameTemplate) {
 
 const formConfig = {
   urlPrefix: '/',
-  submitUrl: '/v0/pension_claims',
+  submit,
   trackingPrefix: 'pensions-527EZ-',
-  transformForSubmit: transform,
```
rianfowler: calling this from inside of the helper functions- it doesn't need to be passed to the form anymore

jbalboni (approved):

PR #7540, opened 2018-04-06
Implemented new Health Records refresh error messages.

U-DON:
Resolves department-of-veterans-affairs/vets.gov-team#6995.

### DIFF
--- a/src/js/health-records/containers/DownloadPage.jsx
+++ b/src/js/health-records/containers/DownloadPage.jsx
@@ -12,7 +12,48 @@ import DownloadLink from '../components/DownloadLink';
 import { openModal, closeModal } from '../actions/modal';
 
 export class DownloadPage extends React.Component {
-  renderMessageBanner() {
+  renderIncompleteMessage = () => {
+    const outdatedStatuses = this.props.refresh.statuses.incomplete;
+    let alertProps = {
+      headline: 'Parts of your health record may not be current.',
+      content: (<p>Most of your health record should be up to date. But we’re having trouble accessing some of your data right now. Please refresh this page or try again later. If you still don’t see your updated information, please call the Vets.gov Help Desk at 855-574-7286 (TTY: 800-829-4833). We’re here Monday–Friday, 8:00 a.m.–8:00 p.m. (ET). If you need this information right away, please call your VA health care provider.</p>)
+    };
+
+    if (outdatedStatuses.length === 1) {
+      const { extractType } = outdatedStatuses[0];
+      switch (extractType) {
+        case 'ImagingStudy':
+          alertProps = {
+            headline: 'Your radiology reports may not be current.',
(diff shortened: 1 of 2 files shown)


src/js/health-records/containers/DownloadPage.jsx
```diff
@@ -12,33 +12,74 @@ import DownloadLink from '../components/DownloadLink';
...
+        headline: 'Couldn’t generate your records',
+        content: (<p>Unfortunately, we weren’t able to generate your health records. Please try again later. You can also call the Vets.gov Help Desk at <a href="tel:855-574-7286">1-855-574-7286</a>, TTY: <a href="tel:18008778339">1-800-877-8339</a>, Monday &#8211; Friday, 8:00 a.m. &#8211; 8:00 p.m. (ET).</p>),
+        status: 'error'
+      };
     } else if (refresh && !isEmpty(refresh.statuses.incomplete)) {
-      alertProps.headline = (<h4>Your health records are not up to date</h4>);
-      alertProps.content = (<p>This older version of your health records may have outdated or missing information.</p>
-      );
-      alertProps.status = 'warning';
+      return this.renderIncompleteMessage();
```
U-DON: Unsure if we should only do this when one or more statuses are out of date (not refreshed within the past 24 hours), or if we should also do this for when statuses fail. Do statuses of "fail" mean that the records of that type won't be included in the report at all, even if there exists some older version of it?

saneshark: Not entirely sure on that, but this is such a huge step in the right direction I say lets merge and open an issue to follow-up.

Actually the next time you run into a fail. We could also always sign in to MHV with that account and should be able to see how they handle that situation.

src/js/health-records/containers/DownloadPage.jsx
```diff
@@ -12,33 +12,74 @@ import DownloadLink from '../components/DownloadLink';
...
     } else if (refresh && !isEmpty(refresh.statuses.failed)) {
-      alertProps.headline = (<h4>Couldn’t update your records</h4>);
-      alertProps.content = (<p>Unfortunately, we weren’t able to generate your most recent health records. You can try again in 24 hours or download an older version of your records below.</p>);
-      alertProps.status = 'warning';
+      alertProps = {
+        headline: 'Couldn’t update your records',
```
saneshark: "Your records couldn't be updated?" for consistency maybe? This seems like an incomplete sentence.

src/js/health-records/containers/DownloadPage.jsx
```diff
@@ -12,33 +12,74 @@ import DownloadLink from '../components/DownloadLink';
...
+  renderMessageBanner = () => {
+    const { form, refresh } = this.props;
+    let alertProps;
+
     if (!form.ready) {
-      alertProps.headline = (<h4>Couldn’t generate your records</h4>);
-      alertProps.content = (<p>Unfortunately, we weren’t able to generate your health records. Please try again later. You can also call the Vets.gov Help Desk at <a href="tel:855-574-7286">1-855-574-7286</a>, TTY: <a href="tel:18008778339">1-800-877-8339</a>, Monday &#8211; Friday, 8:00 a.m. &#8211; 8:00 p.m. (ET).</p>);
-      alertProps.status = 'error';
+      alertProps = {
+        headline: 'Couldn’t generate your records',
```
saneshark: "Your records couldn't be generated" - for consistency with the others.

saneshark (approved):
LGTM, but please see comments. And lets try to merge this only after the 12 pm deploy so I can try to verify on staging or have Danielle or Natalie verify.

dhall (approved):
LGTM

PR #7546, opened 2018-04-09
point facility locator page and specs to original path

dhall:
point facility locator page and specs to original path (now has the new changes)


U-DON (commented):
LGTM. Going to let @saneshark give the approval for merge.

saneshark (approved):
LGTM

PR #7549, opened 2018-04-09
Award amounts of 0.0 should show in the UI

markgreenburg:
Per discussion, a monthly allowance of 0 is a legitimate amount that should show (as compared to a `null` value, which indicates no monthly award and shouldn't be shown). We had been treating these the same and they're not.

### DIFF
--- a/src/js/letters/utils/helpers.jsx
+++ b/src/js/letters/utils/helpers.jsx
@@ -255,7 +255,8 @@ export function getBenefitOptionText(option, value, isVeteran, awardEffectiveDat
     valueString = value;
   }
 
-  const isAvailable = value && value !== AVAILABILITY_STATUSES.unavailable;
+  // NOTE: 0 is a legitimate return value
+  const isAvailable = (value === 0 || value) && value !== AVAILABILITY_STATUSES.unavailable;
   const availableOptions = new Set([BENEFIT_OPTIONS.awardEffectiveDate, BENEFIT_OPTIONS.monthlyAwardAmount, BENEFIT_OPTIONS.serviceConnectedPercentage]);
 
   if (!availableOptions.has(option)) {
--- a/src/js/letters/utils/helpers.jsx
+++ b/src/js/letters/utils/helpers.jsx
@@ -6,7 +6,6 @@ import { apiRequest as commonApiClient } from '../../common/helpers/api';
 import environment from '../../common/helpers/environment';
 import { formatDateShort } from '../../common/utils/helpers';
 import {
-  AVAILABILITY_STATUSES,
   BENEFIT_OPTIONS,
   STATE_CODE_TO_NAME,
   ADDRESS_TYPES,
--- a/src/js/letters/utils/helpers.jsx
+++ b/src/js/letters/utils/helpers.jsx
@@ -256,7 +256,7 @@ export function getBenefitOptionText(option, value, isVeteran, awardEffectiveDat
   }
 
   // NOTE: $0 award is a legitimate number for award amounts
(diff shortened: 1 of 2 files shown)


src/js/letters/utils/helpers.jsx
```diff
@@ -255,7 +255,8 @@ export function getBenefitOptionText(option, value, isVeteran, awardEffectiveDat
     valueString = value;
   }
 
-  const isAvailable = value && value !== AVAILABILITY_STATUSES.unavailable;
+  // NOTE: 0 is a legitimate return value
+  const isAvailable = (value === 0 || value) && value !== AVAILABILITY_STATUSES.unavailable;
```
cvalarida: Wait, so this is either a status string or a value...oi. A'ight...

cvalarida (approved):
It pains me, but it's a necessary evil.

src/js/letters/utils/helpers.jsx
markgreenburg: Yeah to be honest I'm not even sure if it _can_ be a status string.

cvalarida: Oh.
Hmm.

@kreek Do we have a swagger doc for what we can expect from `vets-api`?

markgreenburg: Well it's more about what we're doing with it after we get it (not sure where we might be transforming to `unavailable`). I do know that we can get back `null` or a number (including 0), and those two should obviously be treated differently.

markgreenburg: @cvalarida I'm considering getting rid of the `unavailable` condition. We don't modify any of the `benefitInformation` contents before sending them to the redux store, so getting a value of `unavailable` shouldn't be possible unless it's directly returned by `vets-api`. This doesn't seem to be the case looking at swagger - all of the values are booleans, nullable booleans, or numbers (also nullable, apparently, as in the case of `monthly_award_amount`, though this isn't documented). The only string type is for `award_effective_date` - not sure what the expected return values are there.

Here's how we put the response into redux: 
https://github.com/department-of-veterans-affairs/vets-website/blob/7c2ae4148ec865cb34bea6cd0007841376af87a5/src/js/letters/reducers/index.js#L95

kreek: @cvalarida to view the swagger documentation locally follow the [directions in vets-api](https://github.com/department-of-veterans-affairs/vets-api/tree/master/app/swagger) (we're in the process of getting is easier to view online)

src/js/letters/utils/helpers.jsx
```diff
@@ -255,7 +255,8 @@ export function getBenefitOptionText(option, value, isVeteran, awardEffectiveDat
     valueString = value;
   }
 
-  const isAvailable = value && value !== AVAILABILITY_STATUSES.unavailable;
```
markgreenburg: @kreek can you confirm that the `/beneficiary` endpoint doesn't return a value of `'unavailable'` for any of the `benefitInformation` options?

kreek: @markgreenburg the data services entry for letters is copied from their swagger (which I'm getting a 403 for on the VPN right now 😐) https://github.com/department-of-veterans-affairs/vets.gov-team/tree/master/Data/Data-Services/EVSS/Letters#get-lettersv1letterbeneficiary and it should either return the value or be `null` (but not `'unavailable'`)

markgreenburg: Thanks! Looks like taking out this extra condition is the way to go. @cvalarida - do you see any problems with this update that I might be missing?

PR #7550, opened 2018-04-10
6898 pension apply intro updates

rianfowler:
download-3


src/js/pensions/components/IntroductionPage.jsx
```diff
@@ -61,34 +67,37 @@ class IntroductionPage extends React.Component {
...
               <p><strong>What if I need help filling out my application?</strong> An accredited representative, like a Veterans Service Officer (VSO), can help you fill out your claim. <a href="/disability-benefits/apply/help/index.html">Get help filing your claim</a>.</p>
               <h6>Learn about Veterans pension rates</h6>
               <a href="/pension/rates/" target="_blank">Find out how we decide pension rates.</a>
             </li>
-            <li className="process-step list-two">
+            <li className="process-step list-three">
               <div><h5>Apply</h5></div>
-              <p>Complete an Application for Pension (VA Form 21P-527EZ). <br/><a href="https://www.vba.va.gov/pubs/forms/VBA-21P-527EZ-ARE.pdf">Download VA Form 21P-527EZ</a>.</p>
-              <p>Mail the application to the Pension Management Center (PMC) for your state. <br/><a href="/pension/pension-management-center/">Find your PMC</a>.</p>
+              <p>Complete this pension benefits form.</p>
```
jbalboni: We changed this step when we turned off the form, to point to the pdf. Could we swap between the two depending on the isProduction variable?

jbalboni (approved):
One comment about keeping some old comment, looks good otherwise

peggygannon (approved):
@rianfowler Looks good. thank you!

src/js/pensions/components/IntroductionPage.jsx
rianfowler: Added- `isProduction===false` will show the new content and `true` will show the old content.

jbalboni (approved):

PR #7552, opened 2018-04-10
add action / reducer for toggling review chapter view state

rianfowler:
Figured I'd get some eyes on this since it's my first change to a reducer.


src/js/common/schemaform/state/reducers.js
```diff
@@ -12,6 +15,23 @@ import {
 } from '../state/helpers';
 
 export default {
+  [OPEN_REVIEW_CHAPTER]: (state, action) => {
+    const openChapters = _.concat(
+      _.get('reviewPageView.openChapters', state) || [],
+      action.openedChapter
+    );
+
+    return _.set('reviewPageView.openChapters', openChapters, state);
```
rianfowler: Following the patterns from the other reducers but not mutating the state was engrained in me pretty early on in my redux training. Curious if there have been any re-rendering problems from doing this or if there's some other way this is mitigating problems that can happen from the simple object identity checking that `connect` does.

src/js/common/schemaform/state/reducers.js
```diff
@@ -12,6 +15,23 @@ import {
 } from '../state/helpers';
 
 export default {
+  [OPEN_REVIEW_CHAPTER]: (state, action) => {
+    const openChapters = _.concat(
```
rianfowler: Decided to go with an array for now. If we opt to isolate the accordion behavior, this could be changed into an object that stores more stuff (e.g. errorState, openState, etc).

test/common/schemaform/state/index.unit.spec.js
```diff
@@ -64,6 +67,48 @@ describe('schemaform createSchemaFormReducer', () => {
...
+      };
+
+      const action = {
+        type: OPEN_REVIEW_CHAPTER,
+        openedChapter: 'chapter2'
+      };
+
+      const testState = reducer(previousState, action);
+
+      expect(testState).to.deep.equal(expectedState);
```
rianfowler: stoked that i got to use kitty watcher to make this.

src/js/common/schemaform/state/reducers.js
jbalboni: Any lodash function you get from lodash/fp treats data as immutable, so we're not mutating the state here. Lodash does a shallow clone of the objects at each level in that path (and the root).

src/js/common/schemaform/state/reducers.js
```diff
@@ -12,6 +15,23 @@ import {
...
+      _.get('reviewPageView.openChapters', state) || [],
```
jbalboni: We could probably default this to [] in the initial state.

src/js/common/schemaform/state/reducers.js
rianfowler: Good to know- `object to modify` on the [lodash](https://gist.github.com/jfmengels/6b973b69c491375117dc#_setpath-value-object) docs let me to me think that it was.

src/js/common/schemaform/state/reducers.js
rianfowler: Good idea.

src/js/common/schemaform/save-in-progress/RoutedSavableReviewPage.jsx
```diff
@@ -8,10 +8,20 @@ import SaveFormLink from '../save-in-progress/SaveFormLink';
...
 
 import { saveAndRedirectToReturnUrl, autoSaveForm, saveErrors } from './actions';
+import { getReviewPageOpenChapters } from '../state/selectors';
 import { toggleLoginModal } from '../../../login/actions';
 
 import { ReviewPage } from '../review/ReviewPage';
-import { setData, setPrivacyAgreement, setEditMode, setSubmission, submitForm, uploadFile } from '../actions';
+import {
+  closeReviewChapter,
+  openReviewChapter,
```
rianfowler: This tripped me up for a little bit- couldn't figure out why the redux functions weren't firing on `ReviewPage`

src/js/common/schemaform/state/reducers.js
```diff
@@ -12,6 +15,22 @@ import {
...
 export default {
+  [OPEN_REVIEW_CHAPTER]: (state, action) => {
+    const openChapters = [
+      ...state.reviewPageView.openChapters,
+      action.openedChapter
+    ];
+
+    return _.set('reviewPageView.openChapters', openChapters, state);
+  },
+  [CLOSE_REVIEW_CHAPTER]: (state, action) => {
```
rianfowler: This is very satisfying to see.
redux

src/js/common/schemaform/review/ReviewPage.jsx
```diff
@@ -141,6 +159,8 @@ class ReviewPage extends React.Component {
               <ReviewCollapsibleChapter
                 key={chapter}
                 onEdit={this.handleEdit}
+                toggleChapter={() => this.handleToggleChapter(chapter)}
+                open={this.props.openChapters.includes(chapter)}
```
rianfowler: on track to making `ReviewCollapsibleChapter` a stateless functional component.

src/js/common/schemaform/state/selectors.js
```diff
@@ -1,3 +1,4 @@
 export const getFormData = state => state.form.data;
 export const getFormPages = state => state.form.pages;
 export const getSubmission = state => state.form.submission;
+export const getReviewPageOpenChapters = state => state.form.reviewPageView.openChapters;
```
rianfowler: Not happy with this being on `state.form`- conceptually it's not related to form data and might be better at the on the root or in something like `state.viewState`. We don't have a lot of viewState reducers so I dunno if it's worth it to create one just for this.

test/common/schemaform/review/ReviewPage.unit.spec.jsx
```diff
@@ -507,4 +515,69 @@ describe('Schemaform review: ReviewPage', () => {
...
+        openChapters={['chapter3']}
+        route={{ formConfig, pageList }}
+        setPrivacyAgreement={f => f}
+        location={location}/>
+    ).instance();
+
+    instance.handleToggleChapter('chapter1');
+    expect(openReviewChapter.calledWith('chapter1')).to.be.true;
+    instance.handleToggleChapter('chapter3');
+    expect(closeReviewChapter.calledWith('chapter3')).to.be.true;
```
rianfowler: Tests the `includes` check in the `handleToggleChapter`.

rianfowler:
@jbalboni open chapter view state is done. 


reviewpage

src/js/common/schemaform/review/ReviewPage.jsx
```diff
@@ -141,6 +159,8 @@ class ReviewPage extends React.Component {
               <ReviewCollapsibleChapter
                 key={chapter}
                 onEdit={this.handleEdit}
+                toggleChapter={() => this.handleToggleChapter(chapter)}
```
rianfowler: Have had a lot of debate about naming props on stateless functional components i.e. should the name convey an intent (toggleChapter) or should it describe the action (toggleButtonClicked).

jbalboni:
@rianfowler The styling in your gif up there looks wonky. Is that a result of a change you made or was it like that before?

src/js/common/schemaform/review/ReviewPage.jsx
jbalboni: I think we're all over the place with this. I think I generally prefer the latter.

rianfowler:
Shouldn't be any styling changes. None of the conditionals were removed- just updated their from state to props. I was viewing it with a very narrow layout or there may have been a regression introduced from some other update. There are some artifacts in the gif from Licecap on the open / scroll animation.

**Master**
master

**branch**
branch

jbalboni (approved):
Looks good to me.

PR #7555, opened 2018-04-10
Move site-wide components and rename no-react

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets-contrib/issues/51

Three main changes in here:

- Move the site-wide code into the platform directory
- Move feedback widget tests into feedback directory and update config so they're picked up
- Rename no-react to static-pages everywhere


rianfowler (approved):

ncksllvn (approved):
🎸 ⚡️

PR #7556, opened 2018-04-10
Add uat query param flag to pension and burial forms

jbalboni:
This will allow us to do UAT.

I think this conflicts with some other intro page changes, so I'll plan on merging after those are merged.


src/js/burials/components/IntroductionPage.jsx
```diff
@@ -15,20 +15,20 @@ class IntroductionPage extends React.Component {
     this.props.router.push(this.props.route.pageList[1].path);
   }
   render() {
-    const isProduction = __BUILDTYPE__ === 'production';
+    const isLive = __BUILDTYPE__ !== 'production' || window.location.href.includes('?uat');
```
rianfowler: Good call on `isLive`

rianfowler (approved):

PR #7557, opened 2018-04-10
PROTOTYPE:  Melanie/add 2122 form

melaniegershman:
***do not approve, this is just a prototype***
Here's my first pass at creating form 21-22. Concerns and blockers so far have been documented [here](https://docs.google.com/document/d/1wGggipkb-f2fiJBHtA33P2_KAa3parggGUy53o31-RE/edit#).

Please provide any feedback on the schema and the form-config, and @carolynwill , let me know if any changes need to be made to the form layout.

Thanks!

Screenshots are included below.
screen shot 2018-04-10 at 10 07 55 am
screen shot 2018-04-10 at 10 07 28 am
screen shot 2018-04-10 at 10 07 17 am
screen shot 2018-04-10 at 10 07 09 am
screen shot 2018-04-10 at 10 06 59 am
screen shot 2018-04-10 at 10 06 49 am
screen shot 2018-04-10 at 10 06 37 am
screen shot 2018-04-10 at 10 06 22 am


src/js/veteran-representative/2122-schema.json
```diff
@@ -0,0 +1,91 @@
...
+    "vaFileNumber": {
+      "type": "string"
+    },
+    "insuranceNumber": {
+      "type": "string"
+    },
+    "claimantFullName": {
+      "type": "string"
+    },
+    "claimantAddress": {
```
jbalboni: We have a common address schema that separates out the fields, so we should switch to using that. It's in vets-json-schema, for reference.

src/js/veteran-representative/2122-schema.json
```diff
@@ -0,0 +1,91 @@
...
+      "type": "string"
+    },
+    "claimantEmail": {
```
jbalboni: This should also have a `format: 'email'` property, so that we take advantage of the default JSON Schema email validation logic.

src/js/veteran-representative/2122-schema.json
```diff
@@ -0,0 +1,91 @@
...
+        },
+        "suffix": {
+          "type": "string"
+        }
+      }
+    }
+  },
+  "properties": {
+    "veteranFullName": {
+      "type": "string"
```
jbalboni: If you change this line to `"$ref": "#/definitions/fullName"` you'll pull in the common fullName definition from above.

src/js/veteran-representative/config/form.js
```diff
@@ -0,0 +1,274 @@
...
+          schema: {
+            type: "object",
+            required: [
+              'fullName',
+              'veteranSSN',
+              'vaFileNumber',
+              'insuranceNumber',
+            ],
+            properties: {
+              fullName,
```
jbalboni: I think you want this to be `veteranFullName`, since `fullName` is the name of the common definition but not the actual field in the schema.

src/js/veteran-representative/config/form.js
```diff
@@ -0,0 +1,274 @@
...
+    veteranInformation: {
+      title: "Veteran Information",
+      pages: {
+        veteranInformation: {
+          path: "veteran-information",
+          title: "Veteran information",
+          uiSchema: {
+            fullName: fullNameUI,
+            veteranSSN: ssnUI,
+            vaFileNumber: { "ui:title": "VA file number" },
```
jbalboni: I think we have a common definition for this, too.

src/js/veteran-representative/config/form.js
```diff
@@ -0,0 +1,274 @@
...
+                "ui:title": "Claimant’s last name",
+              },
+              middle: {
+                "ui:title": "Claimant’s middle name",
+              },
+              suffix: {
+                "ui:title": "Claimant’s suffix",
+              },
+            }),
+            // FIX: need a default field for dropdown
```
jbalboni: As long as the schema for this field has an `enum` property, it should turn into a drop down automatically.

src/js/veteran-representative/config/form.js
```diff
@@ -0,0 +1,274 @@
...
+              organizationRepresentativeName,
+              organizationRepresentativeTitle,
+            },
+          },
+        },
+      },
+    },
+    authorizationForRepresentativeAccessToRecords: {
+      title:
+        "Authorization for Representative's Access to Records Protected by Section 7332, TITLE 38, U.S.C.",
```
jbalboni: This is a very long title. More of a design issue, though.

src/js/veteran-representative/config/form.js
```diff
@@ -0,0 +1,274 @@
...
+          },
+        },
+      },
+    },
+    claimantInformation: {
+      title: "Claimant Information",
+      pages: {
+        claimantInformation: {
+          path: "claimant-information",
+          title: "Claimant Information",
```
jbalboni: Typically we use sentence case for page titles and title case for chapter titles.

jbalboni (commented):
This looks great so far. The majority of the comments I had are just around re-using common definitions. There's also a bunch of code copied over from VIC that can be removed, but I left that alone.

src/js/veteran-representative/config/form.js
melaniegershman: @carolynwill what do you suggest I shorten this to?

src/js/veteran-representative/config/form.js
melaniegershman: Thanks for this feedback, I'll go back and update.

src/js/veteran-representative/config/form.js
melaniegershman: Ahh ok, thank you! If it goes against the grain I won't add it, I wasn't sure what the standard was here.

src/js/veteran-representative/config/form.js
carolynwill: Ooh yeah this one is a doozy. How about we cut everything from "protected...." So it reads "Authorization for Representative's Access to Records"

src/js/veteran-representative/2122-schema.json
melaniegershman: Ah! I was on a roll but meant to return to this and use that-- thank you for the reminder!

src/js/veteran-representative/definitions/address.js
```diff
@@ -0,0 +1,160 @@
+import _ from 'lodash/fp';
```
jbalboni: I think this file can be deleted if we can move to the common address definition.

src/js/veteran-representative/helpers.jsx
```diff
@@ -0,0 +1,215 @@
+import _ from 'lodash/fp';
```
jbalboni: I think this file can be deleted.

src/js/veteran-representative/config/form.js
```diff
@@ -0,0 +1,282 @@
+import fullSchema from '../2122-schema.json';
+import _ from 'lodash/fp';
+import IntroductionPage from '../containers/IntroductionPage';
+import ConfirmationPage from '../containers/ConfirmationPage';
+
+import fullNameUI from '../../common/schemaform/definitions/fullName';
+import ssnUI from '../../common/schemaform/definitions/ssn';
+import * as addressUI from '../definitions/address';
```
jbalboni: This should probably use the address definition from `src/js/common/schemaform/definitions`

src/sass/veteran-representative.scss
```diff
@@ -0,0 +1,8 @@
+@import "shared-variables";
```
jbalboni: You'll have to update this file to use the paths that match the other forms, which have changed recently (e.g. https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/js/burials/sass/burials.scss)

jbalboni (approved):

PR #7559, opened 2018-04-10
Move monitoring and production support related code into platform/monitoring 

ncksllvn:
- I created a helper method called `recordEvent` as an alias for `window.dataLayer.push` in the platform/monitoring directory. 
  - I wanted to call it `captureEvent` but there is apparently a deprecated native function called `captureEvents` and I thought that could be confusing.
  - I also defined it in `assets/js` because plain content uses this as well, and I didn't want overlap between global code and Webpack code. I considered having one just point to the other, but the assets code isn't run during unit tests, and then it was just getting too complicated for a one-liner. The content analytics could be heavily cleaned up (footer.html) so I think it should be revisited then.
- I added JSDocs but lmk if they could be improved.
- common/actions and common/reducers contains information for DowntimeNotifications. Should that be moved to platform as well?

Resolves department-of-veterans-affairs/vets-contrib/issues/52


jbalboni:
@ncksllvn On your last point, yes, definitely. I want to get everything out of that common directory by the end of this. I think you should make a downtime notifications directory in monitoring and consolidate the code in there.

src/platform/monitoring/DowntimeNotification.jsx
```diff
@@ -52,18 +59,56 @@ function DowntimeNotificationWrapper({ status, children }) {
...
+
+/**
+ * React component used to conditionally render children components based on the status (down, down-approaching, or ok) of Vets.gov services.
+ * @module platorm/monitoring/DowntimeNotification
+ */
 class DowntimeNotification extends React.Component {
 
   static propTypes = {
+
+    /** The name of the consuming application, which will be displayed in downtime messaging. */
```
jbalboni: I think I would rather see these comments in the `@property` tags above, so they show up whenever we start generating docs from JSDoc comments.

src/platform/monitoring/RateLimiter.jsx
```diff
@@ -1,9 +1,14 @@
...
-import LoadingIndicator from './LoadingIndicator';
+import LoadingIndicator from '../../js/common/components/LoadingIndicator';
 
-/*
+/**
+ * React component for enabling or disabling access to content based on a randomizer and an integer representing the threshold.
+ * The integers representing the threshold are defined in the create-settings build process.
+ * For content to be enabled, a randomly-generated number must be greater than than that integer.
+ * For example, in theory a "1" would mean 100% of users will be denied access, while a "4" means only 40% of users would be denied.
+ *
```
jbalboni: There's an example tag for JSDoc I think we could use here.

src/platform/monitoring/record-event.js
```diff
@@ -0,0 +1,10 @@
+/**
+ * Helper function for reporting events to Google Analytics. An alias for window.dataLayer.push.
+ * @module platform/monitoring/record-event
+ * @see https://developers.google.com/tag-manager/devguide
+ * @param {object} data - The event data that will be sent to GA.
+ */
+
+export default function recordEvent(data) {
+  window.dataLayer.push(data);
```
jbalboni: I think I saw some code that was returning `window.dataLayer.push()`, so we may want to return here to be consistent. Though I'm not sure why anyone would need that.

jbalboni (commented):
Looks good sof

jbalboni:
@ncksllvn Looks good so far. As far as JSDoc comments, they look good to me. I think my plan is to try to write JSDoc comments for the things a developer might interact with when developing an app. So it would make sense to document the DowntimeNotification component and props like you did, but internal helpers or reducers/actions probably don't need them.

ncksllvn:
@jbalboni Thanks for the feedback! I believe I have made all of the corrections. Couple notes -

- I left the JSDoc comments in DowntimeNotification for methods that won't necessarily concern devs who are just using it for an app because I think they're really useful in the event we need to do some work on it, but I can convert them to regular comments if you like.
- I consolidated DowntimeNotification's actions and reducer into `platform/monitoring/DowntimeNotification`. I renamed the React component to `index.jsx` so that modules that import it without the extra directory like `platform/monitoring/downtime/DowntimeNotification`, But again, I don't know if this would set a bad precedent, set lmk if you'd like that to be changed as well.

ncksllvn:
Also, just another suggestion, I think we should try to use `@typedef` whenever we can. I think it's super useful for so many different data structures being passed around.

ncksllvn:
Also, just moved the DowntimeNotification tests into its directory as well. Very exciting 🐎

src/platform/site-wide/index.js
```diff
@@ -1,10 +1,10 @@
 /**
- * Module for site wide components  
+ * Module for site wide components
  * @module platform/site-wide
  */
 
 // This will move later, most likely
```
jbalboni: Can remove this comment now

jbalboni (approved):

PR #7560, opened 2018-04-11
genderlabels in hca form

lbai001:
#### Proposed Changes
* use genderlabels insdead of gender in hca form

#### Screen Shot
screen shot 2018-04-11 at 3 53 27 am

### DIFF
--- a/src/js/hca/config/form.js
+++ b/src/js/hca/config/form.js
@@ -4,13 +4,14 @@ import moment from 'moment';
 import fullSchemaHca from 'vets-json-schema/dist/10-10EZ-schema.json';
 
 import {
-  genders,
   maritalStatuses
 } from '../../common/utils/options-for-select';
 
 import {
   states
 } from '../../common/utils/address';
+
+import { genderLabels } from '../../common/utils/labels';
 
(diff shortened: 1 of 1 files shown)


lbai001:
@jbalboni I'm not sure if we need the uiSchema options
```
              'ui:options': {
                labels: genderLabels
              }
```
any advice on this would be great

src/js/hca/config/form.js
```diff
@@ -274,8 +278,8 @@ const formConfig = {
             properties: {
               gender: {
                 type: 'string',
-                'enum': genders.map(gender => gender.value),
-                enumNames: genders.map(gender => gender.label)
+                'enum': Object.keys(genderLabels),
```
jbalboni: If we import `gender` from fullSchemaHca.properties along with the other fields, we could change the code to look like:

```js
properties: {
  gender,
  ...
```
I think that would simply things a bit.

jbalboni (commented):
Looks good, I think we could simplify it a little more. The ui:options config is only needed if we don't use enunNames in schema itself. So right now it's not doing anything, but we need to keep it if we make the change suggested in my comment.

jbalboni (approved):

PR #7565, opened 2018-04-11
tweak focus logic

rianfowler:


src/js/common/utils/helpers.js
```diff
@@ -125,7 +125,9 @@ export function focusElement(selectorOrElement, options) {
     : selectorOrElement;
 
   if (el) {
-    if (el.tabIndex === -1) el.setAttribute('tabindex', '-1');
+    if (el.tabIndex <= 0) {
```
rianfowler: @jbalboni @1Copenut The `tabIndex` of non-tabbable elements on IE is 0. On everything else it's -1.

jbalboni:
@rianfowler That change looks good to me, maybe @1Copenut can help us test this out in JAWS?

rianfowler:
@1Copenut [This](http://863f91d6d319f832d887e6341ccb5326.review.vetsgov-internal/) should be the review instance.

jbalboni (approved):
I still think we need to test this with JAWS, but I think we  should merge this if Trevor doesn't have time today, and test it in staging/prod later.

1Copenut (approved):
Tested two multi-page forms in IE11 on a VM. Both displayed the expected behavior when I pressed the `Continue` button with errors on the page.

PR #7566, opened 2018-04-11
Add Error capture

markgreenburg:
Adds some Sentry logging for strange response errors in appeals.
Should be temporary until cause of current issues is identified.


cvalarida (approved):
@kreek We're reasonably confident that this won't contain any PII, but I'd like to get your input on if this is an appropriate solution.

kreek (approved):
LGTM 🚀

PR #7567, opened 2018-04-12
Rolls back PR 7549 (need to wait for sustainment)

markgreenburg:
Reverts https://github.com/department-of-veterans-affairs/vets-website/pull/7549

Sustainment needs to make changes to the letter generator to handle $0 award amounts. These changes are in the pipeline but the next deploy isn't til May 2nd. We need to wait to deploy the changes in 7549 until the Letter Generator is ready.


cvalarida (approved):
:man_shrugging:

PR #7568, opened 2018-04-12
Split Login and Verify components into separate apps

U-DON:
Resolves department-of-veterans-affairs/vets.gov-team#9613.


src/js/common/components/RequiredLoginView.jsx
```diff
@@ -5,22 +5,41 @@ import { intersection } from 'lodash';
...
   componentDidUpdate() {
     const isLoading = this.props.user.profile.loading;
-    if (!isLoading && this.shouldSignIn()) {
-      window.location.replace(signInUrl);
+
+    if (!isLoading) {
+      if (this.shouldSignIn()) {
+        window.location.replace(signInUrl);
+      } else if (this.shouldVerify()) {
+        window.location.replace(verifyUrl);
```
U-DON: Didn't think there was a good reason to render the verify app inline but still have redirect logic in the verify component. So just being consistent by always having it redirect, similar how we would if we needed the user to log in.

src/js/common/components/RequiredLoginView.jsx
```diff
@@ -36,53 +55,50 @@ class RequiredLoginView extends React.Component {
   }
 
   renderVerifiedContent = () => {
-    const { serviceRequired, user } = this.props;
+    if (this.shouldVerify()) {
+      return <LoadingIndicator setFocus message="Redirecting to verify..."/>;
```
U-DON: This and the "Redirecting to login..." loaders should never actually be shown since the user should be redirected before then. But a placeholder nonetheless if it's somehow not happening right away, and easy to catch if it hangs in tests.

src/js/login/components/SearchHelpSignIn.jsx
```diff
@@ -12,12 +12,15 @@ import SignInProfileMenu from './SignInProfileMenu';
...
 import { isUserRegisteredForBeta } from '../../personalization/beta-enrollment/actions';
 
-class SearchHelpSignIn extends React.Component {
-  componentDidMount() {
-    const nextParams = new URLSearchParams(window.location.search);
-    const nextPath = nextParams.get('next');
-    if (nextPath) {
-      this.props.toggleLoginModal(true);
+export class SearchHelpSignIn extends React.Component {
+  componentDidUpdate() {
```
U-DON: Changed to `componentDidUpdate` part of the lifecycle since `currentlyLoggedIn` only gets determined after the initial user data fetch, and we need to confirm that before we redirect for login.

jbalboni (approved):
Had one question, but it looks good to me. I tried out locally and didn't run into any issues.

src/js/login/components/SearchHelpSignIn.jsx
U-DON: Going to make a correction to the conditional; the way it is currently, it won't close the modal if you try to X out since it's updating the state, but `showModal` would be false. Needs to be sure that this only happens when the modal has remained closed.

dhall (approved):
LGTM

PR #7575, opened 2018-04-12
Use common styles from jean-pants

jbalboni:
This PR is moving styles out of vets-website in favor of the ones in jean-pants, which I updated yesterday. 

The intent is that common styles and component styles for the components in the design system live in jean-pants and we import from there.

One exception right now are the header and footer styles, which I'm leaving alone for now, but may also move over in the future, depending on exactly how we want to organize the code for that.


rianfowler (approved):
LGTM- maybe add a sentence or two to the module readme describing the necessity of the sass when using jean-pants components

src/sass/modules/_m-schemaform.scss
```diff
@@ -1,5 +1,28 @@
...
+
+.float-right {
+  float: right;
+}
+
+.float-left {
+  float: left;
+}
+
 legend {
```
ncksllvn: How come this had to be added?

ncksllvn (approved):
LGTM!

src/sass/modules/_m-schemaform.scss
jbalboni: We had one file with these "growable" classes and the float classes and the only place they're used is in the schemaform apps, so it seemed like a good idea to consolidate them.

src/sass/686-dependent-status.scss
```diff
@@ -1,8 +1,8 @@
-@import "shared-variables";
+@import "~@department-of-veterans-affairs/jean-pants/sass/shared-variables";
 @import "modules/m-process-list";
```
U-DON: Why are some modules left out? Like `m-form-process` was moved into the design system, but not `m-process-list`.

U-DON (approved):
LGTM.

src/sass/686-dependent-status.scss
jbalboni: That particular one was a mistake, I meant to pull that one from jean-pants, too.

I think all of the common modules will get moved over unless they're specific to particular feature (like the downtime notification file, for example). I left the header/footer related ones in here for now because I'm a little fuzzy on the best way to organize that in the design system.

PR #7576, opened 2018-04-12
Updated decisionIssues refs to issues per actual api response shape

markgreenburg:
Updates the properties within an appeal's details that we pull from to assemble content dynamically. The API differs from what we were using.


cvalarida (approved):
Do you want to change `ssocTimeliness` to `socTimeliness` when `status.type === 'pendingSoc'` in this PR or another?

markgreenburg:
@cvalarida planning on adding another PR for timeliness shortly

PR #7577, opened 2018-04-12
Update timeliness property name

markgreenburg:
Updates the property name that we pull from to generate `timeliness` ranges because the variable name we use now isn't what the api provides.


cvalarida (approved):
Assuming you combed through all instances of `ssocTimeliness` and this is all that came up for `pendingSoc` status, I'd say it looks beautiful!

PR #7579, opened 2018-04-12
[WIP] 6083-mobile-breadcrumb

1Copenut:
## Purpose
To provide contextual breadcrumbs based on `viewport` dimensions on page load, and resize. 

## Methodology
I created two breadcrumb helper JS files, one for static content, one for React components. Each makes use of similar functions to find our breadcrumb markup by ID, clone the list, and determine which one to show based on viewport width. The React version also makes use of lifecycle methods to properly attach and remove breadcrumb methods when component mounts or unmounts.

## Testing
I've tested the breadcrumbs for handling in Chrome, Safari, and Firefox for OSX, and IE11 and Firefox in Windows 7. This included testing with VoiceOver and NVDA screen readers to ensure the breadcrumbs were announcing properly.

## Inclusion in 526 Testing
@goldenmeanie is interested in including these breadcrumbs in the 526 testing that will be happening in the next 2-3 weeks. Please advise if I need to adjust or change things so they're able to be folded into the staging environment.


content/includes/breadcrumbs.html
```diff
@@ -1,39 +1,33 @@
...
-            </a></li>
-        {% endif %}
-      {% endfor %}
-  </ul>
-</nav>
+    <!-- Adding back in the last link as aria-current="page" -->
+    {% comment %} Each breadcrumb item has a child object. One of the properties of that object is `file` which loads the front
+    matter metadata from the corresponding Markdown input file. Below, we’ve assigned that object to a variable named 'crumb_child_meta'.
+    If these templates are still using tinyliquid, you can dump the properties of this object using {{ crumb | keys }}. {%
+    endcomment %} {% assign crumb_child_meta = crumb.children.0.file %}
```
cvalarida: The formatting of this block seems kinda wonky. Can we pull that `{% assign... %}` onto its own line and bring `endcomment %}` to the same line as its `{%`?

content/includes/breadcrumbs.html
```diff
@@ -1,39 +1,33 @@
...
+    <li>
+      {% if current_page.name != crumb.name %}
+      <a href="/{{ crumb.path }}">
+        {% if crumb_child_meta.display_title %} {{ crumb_child_meta.display_title }} {% elsif crumb_child_meta.title %} {{ crumb_child_meta.title
+        }} {% else %} {{ crumb.name | replace: '-', ' ' | capitalize }} {% endif %}
+      </a>
+      {% else %}
+      <a aria-current="page" href="/{{ crumb.path }}">
+        {% if crumb_child_meta.display_title %} {{ crumb_child_meta.display_title }} {% elsif crumb_child_meta.title %} {{ crumb_child_meta.title
```
cvalarida: Readability suffers here by pulling these onto the same line. Can we keep it split up like it was before?

cvalarida (commented):
I didn't make it very far before I had to bounce to something else, so this isn't a complete review.

What I see so far is looking pretty good. I haven't fired it up myself, but what I remember of it from our call the other day, I like where it's heading.

assets/js/breadcrumbs/mobile-breadcrumb.js
```diff
@@ -0,0 +1,139 @@
...
+}
+
+function sliceMobileLink(targetId) {
+  // Targeted unordered list and cloned list
+  var target = document.getElementById(targetId);
+  var clonedTarget = target.cloneNode(true);
+  
+  // Cloned list items and array conversion
+  var targetList = clonedTarget.children;
+  var listArr = Array.prototype.slice.call(targetList);
```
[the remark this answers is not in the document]
jbalboni: This file isn't going to be processed by Babel, so I think we'll have to stay away from ES6 (or transform this file in the build process).

script/metalsmith/breadcrumbTransformer.js
```diff
@@ -0,0 +1,22 @@
...
+module.exports = (files, metalsmith, done) => {
+  Object.keys(files).forEach(file => {
+    if (path.extname(file) !== '.html') return;
+
+    const data = files[file];
+    const $ = cheerio.load(data.contents.toString());
+
+    $('#va-breadcrumbs').each(() => {
+      $.root().append(
+        '<script type="text/javascript" src="/js/breadcrumbs/mobile-breadcrumb.js"></script>'
```
jbalboni: I wonder if it might be better to inline this script. It's a little long for that, but we are starting to stack up small external JS scripts, which is pushing back the time to first byte on our main bundles.

assets/js/breadcrumbs/mobile-breadcrumb.js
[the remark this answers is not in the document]
1Copenut: Yep, I had to go back and change out an ES6 string interpolation this morning. Will commit it in a bit.

assets/js/breadcrumbs/mobile-breadcrumb.js
```diff
@@ -0,0 +1,139 @@
+var config = {
```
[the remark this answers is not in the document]
1Copenut: I ran the linter manually a couple of times to be sure, and it didn't throw any errors. Will be sure to hook up the pre-commit hooks so all future runs lint without manual intervention.

content/includes/breadcrumbs.html
```diff
@@ -1,39 +1,33 @@
...
-        {% if current_page.name != crumb.name %}
-          <!-- removed to remove 'active' breadcrumb -->
-        
-          {% comment %}
-            Each breadcrumb item has a child object. One of the properties of
-            that object is `file` which loads the front matter metadata from
-            the corresponding Markdown input file.
+<nav aria-label="Breadcrumb" aria-live="polite" aria-relevant="additions text" class="va-nav-breadcrumbs js-visual {{gatePageClassName}}"
+  id="va-breadcrumbs">
+  <p class="usa-sr-only">Breadcrumb navigation will usually show all page links. It will adjust to show only the previous page when zoomed in, or
```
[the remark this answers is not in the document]
1Copenut: It borrows from a pattern I saw on an accessible modal dialog. I tried out the screen reader while resizing the viewport, and it was reading out the updated mobile-first breadcrumb, but offering no context. This was my attempt to orient screen reader users on first load. It is easily skipped by tabbing, and can be removed easily if the 508 office says it's not effective.

assets/js/breadcrumbs/mobile-breadcrumb.js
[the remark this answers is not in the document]
1Copenut: Per @jbalboni request, I inlined this script using UglifyJS. I'll remove it here if the inlined version passes muster.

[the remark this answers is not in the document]
jbalboni: Could we have the breadcrumb transformer run uglify on this script and inline the result? Having the uglified version as the only source is going to be hard to maintain.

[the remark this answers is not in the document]
jbalboni: Also, I believe the linter ignores files in this directory, since these files aren't run through babel and can't use ES6.

We may want to change this process.

jbalboni:
@1Copenut It looks like you need to update this to pull in the newer jean-pants package: `yarn upgrade @department-of-veterans-affairs/jean-pants --latest`

content/pages/burials-and-memorials/application/530.md
```diff
@@ -4,13 +4,15 @@ entryname: burials
...
+  <nav aria-label="Breadcrumb" aria-live="polite" aria-relevant="additions text" class="va-nav-breadcrumbs js-visual"
+  id="va-breadcrumbs">
+    <p class="usa-sr-only">Breadcrumb navigation will usually show all page links. It will adjust to show only the previous page when zoomed in, or viewed on a mobile device.</p>
+    <ol class="row va-nav-breadcrumbs-list columns" id="va-breadcrumbs-list">
       <li><a href="/">Home</a></li>
       <li><a href="/burials-and-memorials/"> Burials and Memorials </a></li>
       <li><a href="/burials-and-memorials/survivor-and-dependent-benefits/"> Survivors and Dependents </a></li>
-      <li><a href="/burials-and-memorials/survivor-and-dependent-benefits/burial-costs/">Handling Burial Costs (Allowances)</a></li>
-    </ul>
+      <li><a aria-current="page" href="/burials-and-memorials/survivor-and-dependent-benefits/burial-costs/">Handling Burial Costs (Allowances)</a></li>
```
jbalboni: This last link isn't the page for the form, it's a content page that's one step back in the breadcrumbs.

content/pages/burials-and-memorials/pre-need/form-10007-apply-for-eligibility.md
```diff
@@ -4,12 +4,14 @@ entryname: pre-need
...
-    <ul class="row va-nav-breadcrumbs-list columns" role="menubar" aria-label="Primary">
+  <nav aria-label="Breadcrumb" aria-live="polite" aria-relevant="additions text" class="va-nav-breadcrumbs js-visual"
+  id="va-breadcrumbs">
+    <p class="usa-sr-only">Breadcrumb navigation will usually show all page links. It will adjust to show only the previous page when zoomed in, or viewed on a mobile device.</p>
+    <ol class="row va-nav-breadcrumbs-list columns" id="va-breadcrumbs-list">
       <li><a href="/">Home</a></li>
       <li><a href="/burials-and-memorials/">Burials and Memorials</a></li>
-      <li><a href="/burials-and-memorials/pre-need/">Pre-need Eligibility Determination</a></li>
-    </ul>
+      <li><a aria-current="page" href="/burials-and-memorials/pre-need/">Pre-need Eligibility Determination</a></li>
```
jbalboni: Same as above, this isn't the path to the form page, it's one step back.

content/pages/education/apply-for-education-benefits/application/1990.md
```diff
@@ -5,12 +5,15 @@ layout: page-react.html
...
+  <nav aria-label="Breadcrumb" aria-live="polite" aria-relevant="additions text" class="va-nav-breadcrumbs js-visual"
+  id="va-breadcrumbs">
+    <p class="usa-sr-only">Breadcrumb navigation will usually show all page links. It will adjust to show only the previous page when zoomed in, or viewed on a mobile device.</p>
+    <ol class="row va-nav-breadcrumbs-list columns" id="va-breadcrumbs-list">
       <li><a href="/">Home</a></li>
       <li><a href="/education/">Education</a></li>
-      <li class="parent"><a href="/education/apply-for-education-benefits/">Apply for Education Benefits</a></li>
-    </ul>
+      <li><a href="/education/apply/">Application Process</a></li>
+      <li><a aria-current="page"  href="/education/apply-for-education-benefits/application/1990/introduction">Apply for Education Benefits</a></li>
```
jbalboni: As you go through the form, this link is going to be inaccurate, since the introduction page is just the first page of the form. I think our typical pattern in the site is for breadcrumbs to not include a link to the current page. Are we changing that? If we are, we may need to move the breadcrumbs into React component for all the React apps, so we can keep them up to date.

script/metalsmith/breadcrumbTransformer.js
```diff
@@ -0,0 +1,26 @@
...
+
+const breadcrumb = fs.readFileSync(path.join(__dirname, '../../assets/js/breadcrumbs/mobile-breadcrumb.js'), 'utf8');
+
+module.exports = (files, metalsmith, done) => {
+  Object.keys(files).forEach(file => {
+    if (path.extname(file) !== '.html') return;
+
+    const data = files[file];
+    const $ = cheerio.load(data.contents.toString());
+    const minifiedCode = UglifyJS.minify(breadcrumb);
```
jbalboni: Might be useful to minify this outside of the `forEach` loop, since the same minified code is used for each file.

src/js/claims-status/components/Breadcrumbs.jsx
```diff
@@ -1,14 +1,59 @@
...
+      buildMobileBreadcrumb.bind(this);
+    });
+
+    window.addEventListener('resize', () => {
+      debouncedToggleLinks('va-breadcrumbs-claims-list');
+      debouncedToggleLinks.bind(this);
+    });
+  }
+
+  componentDidUpdate() {
```
jbalboni: This method may run multiple times, should we be adding new event listeners each time? It seems like adding them in `componentDidMount` would be enough.

src/js/claims-status/components/Breadcrumbs.jsx
```diff
@@ -1,14 +1,59 @@
...
+    });
+
+    window.addEventListener('resize', () => {
+      debouncedToggleLinks('va-breadcrumbs-claims-list');
+      debouncedToggleLinks.bind(this);
+    });
+  }
+
+  componentWillUnmount() {
+    window.removeEventListener('DOMContentLoaded', () => {
```
jbalboni: I don't think this is removing an event listener, since you're passing in a new function that won't match any of the current event listeners. I think what we usually to is create a method on a component and then do `addEventListener('resize', this.toggleLinks)` and `removeEventListener('resize', this.toggleLinks)`

src/js/health-records/components/Breadcrumbs.jsx
```diff
@@ -12,14 +36,28 @@ class Breadcrumbs extends React.Component {
 
     if (pathname.match(/download\/?$/)) {
       crumbs.push(<Link to="/" key="main">Get Your VA Health Records</Link>);
-      crumbs.push(<span key="download"><strong>Download Your Health Records</strong></span>);
+      crumbs.push(<Link to="/" key="download">Download Your VA Health Records</Link>);
```
jbalboni: Is this an intentional heading change?

src/js/utils/breadcrumb-helper.js
```diff
@@ -0,0 +1,169 @@
+const config = {
```
jbalboni: This file should move to `src/platform/utilities/ui/`

jbalboni: I'm a little concerned that we're doing this much direct DOM manipulation in a React component. I realize you're reusing code that's also being used for the static pages that isn't in React, but given that we're already going to have maintain two slightly different versions of this code, I wonder if it might be better to have a React breadcrumbs component that is written in idiomatic React and use that in the React apps that need it. I'm pretty sure the React code for this would much simpler than the DOM version here.

src/sass/base/_b-utils.scss
```diff
@@ -0,0 +1,86 @@
+.full-width {
```
jbalboni: I think you can delete this file.

src/sass/base/_va.scss
```diff
@@ -0,0 +1,1497 @@
+html,
```
jbalboni: I think you can delete this file, too.

src/js/claims-status/containers/AdditionalEvidencePage.jsx
```diff
@@ -64,23 +64,33 @@ class AdditionalEvidencePage extends React.Component {
...
   render() {
+    const breadcrumbArr = [
+      { key: 'home', href: '/', text: 'Home' },
+      { key: 'disability-benefits', href: '/disability-benefits/', text: 'Disability Benefits' },
+      { key: 'your-claims', href: '/track-claims/your-claims/', text: 'Track Your Claims and Appeals' }
+    ];
+
+    const listItem = breadcrumbArr.map(item => {
+      return <li key={item.key}><a href={item.href}>{item.text}</a></li>;
+    });
```
ncksllvn: This is kinda picky, but `listItem` seems vague to me. Maybe just `breadcrumbItems` or something?

src/js/discharge-wizard/components/Breadcrumbs.jsx
```diff
@@ -0,0 +1,48 @@
...
+
+  componentWillUnmount() {
+    window.removeEventListener('DOMContentLoaded', () => {
+      buildMobileBreadcrumb.bind(this);
+    });
+
+    window.removeEventListener('resize', () => {
+      debouncedToggleLinks.bind(this);
+    });
+  }
```
ncksllvn: Does each application have its own breadcrumb component? Are there differences between them that can't be consolidated via props?

src/js/utils/breadcrumb-helper.js
ncksllvn: ☝️ ➕

src/js/claims-status/components/Breadcrumbs.jsx
ncksllvn: ☝️ ➕

ncksllvn (commented):
Left a few comments, but I'm a concerned that there is too much code redundancy and I think we could improve by consolidating breadcrumbs logic into a common React component. We could even pass the list items into the React root from static pages.

What are your all's thoughts on that? It just seems like solid use case for Jean Pants to me 👖

content/pages/education/apply-for-education-benefits/application/1990.md
1Copenut: @jbalboni Good call. The final breadcrumb link is being put back, but we're using CSS to remove `pointer-events`, and no, the breadcrumb should not change as you move through multi-step forms. I'll update this href to `/education/apply-for-education-benefits/application/1990`

1Copenut:
I changed this to WIP because we're testing this week or early next, and didn't know what the right environment was. I'll follow up with teams to ensure we get it in the right place(s).

The `<Breadcrumb/>` component has been refactored out to a single, and made more robust. I'll take it as a follow on item to move this component to the Design System after user research is done.

content/pages/burials-and-memorials/pre-need/form-10007-apply-for-eligibility.md
1Copenut: @jbalboni There doesn't seem to be a form tied to this one. The content and link goes to a PDF that has to be downloaded or filled out. I'll change this as needed.

jbalboni: Ah, this one's a little confusing since the form isn't launched. We can leave it alone and update it whenever it goes live (a couple months, probably).

src/platform/utilities/ui/Breadcrumbs.jsx
```diff
@@ -0,0 +1,118 @@
...
+
+class Breadcrumbs extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      mobileShow: false,
+    };
+
+    this.id = this.props.id || uniqid('va-breadcrumbs-');
```
jbalboni: This is a little risky in that props could change and the instance variables wouldn't change to reflect that. You could make these `const` variables inside the render method (or other methods where they're needed), I think.

src/platform/utilities/ui/Breadcrumbs.jsx
```diff
@@ -0,0 +1,118 @@
...
+    this.listId = this.props.listId || uniqid('va-breadcrumbs-list-');
+    this.mobileWidth = this.props.mobileWidth || 425;
+
+    this.debouncedToggleDisplay = this.debouncedToggleDisplay.bind(this);
+    this.renderBreadcrumbLinks = this.renderBreadcrumbLinks.bind(this);
```
jbalboni: We do this both ways in our codebase, but I think the current preference is to use the auto bind syntax:

```js
someMethod = () => {
  ...
}
```

src/platform/utilities/ui/Breadcrumbs.jsx
```diff
@@ -0,0 +1,118 @@
...
+    return React.Children.map(this.props.children, (child, i) => {
+      if (i === this.props.children.length - 1) {
+        return (
+          <li>{React.cloneElement(child, {
+            'aria-current': 'page',
+          })}</li>
+        );
+      }
+
+      return <li>{React.cloneElement(child)}</li>;
```
jbalboni: Is the cloneElement call necessary here?

src/platform/utilities/ui/Breadcrumbs.jsx
```diff
@@ -0,0 +1,118 @@
...
+        id={this.id}>
+        <p className="usa-sr-only">Breadcrumb navigation will usually show all page links. It will adjust to show only the previous page when zoomed in, or viewed on a mobile device.</p>
+        { shownList }
+      </nav>
+    );
+  }
+}
+
+Breadcrumbs.propTypes = {
+  crumbs: PropTypes.array,
```
jbalboni: Doesn't look like we use this?

src/platform/utilities/ui/Breadcrumbs.jsx
```diff
@@ -0,0 +1,118 @@
...
+        );
+      }
+
+      return null;
+    });
+  }
+
+  render() {
+    const {
+      ...breadcrumbProps
```
jbalboni: I'm not sure I understand why this is being done.

src/platform/utilities/ui/Breadcrumbs.jsx
1Copenut: No, I'll remove this momentarily. Legacy item.

src/platform/utilities/ui/Breadcrumbs.jsx
1Copenut: @jbalboni There's a few instances where custom props are being fed into the `<Breadcrumb>` component, like the Facility Locator. I added this `...breadcrumbProps` defensively to ensure those could be added as necessary. Will remove if it's not something we want here.

jbalboni:
@1Copenut What's the plan for how this would be tested, do you know? If it's merged it'll go to production, which is maybe not ideal, unless we can make the changes behind a flag somehow (that seems difficult with all the changes here). We can publish arbitrary branches to dev.vets.gov, so that might be your best option, but I'm not sure if we can keep them from getting overwritten.

src/platform/utilities/ui/Breadcrumbs.jsx
jbalboni: Ok. It looks like you could `...this.props` later in this method to get the same effect without this line, unless I'm missing something.

1Copenut: @jbalboni Some instances of the `<Breadcrumb>` have custom props passed down, like the Facility Locator. I'm not sure they're needed, but didn't want to remove them either. Added this as a defensive way to not break things. Can remove as needed.

content/pages/burials-and-memorials/pre-need/form-10007-apply-for-eligibility.md
1Copenut: 👍

src/platform/utilities/ui/Breadcrumbs.jsx
```diff
@@ -0,0 +1,116 @@
...
+    );
+  }
+}
+
+Breadcrumbs.propTypes = {
+  id: PropTypes.string,
+  listId: PropTypes.string,
+  mobileWidth: PropTypes.number,
+};
+
```
ncksllvn: Great work!! I wonder if there's any way we could use this on pages that also aren't registered as React applications, like we do with the navigation. It's all solid code but that way we wouldn't have to worry about maintaining a separate JS file that performs the same function minus React. @jbalboni do you think it would be okay to do something like that in our common startup code?

assets/js/breadcrumbs/mobile-breadcrumb.js
```diff
@@ -0,0 +1,129 @@
...
+  containerString: 'va-breadcrumbs',
+  listString: 'va-breadcrumbs-list',
+  jsHiddenClass: 'js-hide',
+  jsVisualClass: 'js-visual',
+  mobileClass: 'va-nav-breadcrumbs-list__mobile-link',
+  triggerDelay: 500,
+  triggerWidth: 425,
+};
```
ncksllvn: I wonder if we could have `triggerWidth` be derived from what CSS has set as the small device width. Probably just wishful thinking.

assets/js/breadcrumbs/mobile-breadcrumb.js
```diff
@@ -0,0 +1,139 @@
...
+  );
+
+  return breadcrumbList;
+}
+
+function toggleLinks(targetId) {
+  var breadcrumb = document.getElementById(targetId);
+  var clone = document.getElementById(targetId + '-clone');
+
+  if (window.innerWidth <= config.triggerWidth) {
```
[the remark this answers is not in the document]
ncksllvn: `platform/utilities/accessibility` detects screen size using `matchMedia`. Might be worth doing something with here too.

src/platform/utilities/ui/Breadcrumbs.jsx
```diff
@@ -0,0 +1,116 @@
...
+    super(props);
+
+    this.state = {
+      mobileShow: false,
+    };
+  }
+
+  componentDidMount() {
+    const mobileWidth = this.props.mobileWidth || 425;
+
```
ncksllvn: `425` would make sense as a `defaultProp`

ncksllvn (commented):
Solid improvements, well done! I'm curious about what the rollout process for this will be as well. Since it has such a wide impact I think it makes sense to have some standalone unit and E2E tests written for the Breadcrumb component. I think it would also be interesting to try to explore ideas around rendering breadcrumbs always as React component, even for those that don't have a dedicated React entry point, just so we don't have maintain two separate files of complex logic.

assets/js/breadcrumbs/mobile-breadcrumb.js
1Copenut: @ncksllvn I like this suggestion a lot. I've been looking through the CSS and found 481px was the small-screen SCSS variable. I'm going to make that my breakpoint width for these breadcrumbs for consistency.

1Copenut:
@jbalboni @ncksllvn 
I added in the feedback you proposed to solidify our breakpoint around the CSS media query, and to use `window.mediaMatch`. Wasn't familiar with that, so it was a good thing to line up with other code in our base.

Added a full unit test and 10 or 11 e2e tests specifically for the breadcrumbs on different React apps. 

Added a rollout plan MD file here: https://github.com/department-of-veterans-affairs/vets-website/pull/7579/files#diff-22661d7a4cd7454a743b7a592695bf88

1Copenut:
Closing because this work is superseded by https://github.com/department-of-veterans-affairs/vets-website/pull/8004

PR #7580, opened 2018-04-12
Personalization - Add Beta Invitation onto Account page

ncksllvn:
This does a check for the production flag in the Dashboard `manifest.json` so it won't render until our apps are enabled in production. There is also an additional flag in the Dashboard manifest to ensure that we want to show the invitation, even after going to production, so that it is not shown for when we are in production but still in UAT.

Resolves department-of-veterans-affairs/vets.gov-team/issues/9866


bshyong (approved):
LGTM

PR #7588, opened 2018-04-13
Update view logic precedence for Docket

markgreenburg:
Fixes the order in which decide which content to show for the Docket component


cmgiven (approved):
LGTM!

PR #7591, opened 2018-04-13
Fake submit and 'review' omission in title

cehsu:
These changes provide a fake submit for the review page and omit "review" from chapter/page titles (this means instances of reviewTitle should now be removed).

- [x] verify that all instances of reviewTitle match above pattern and if so, remove


src/js/common/schemaform/review/ReviewCollapsibleChapter.jsx
```diff
@@ -151,7 +151,7 @@ export default class ReviewCollapsibleChapter extends React.Component {
                 {pageSchema &&
                   <SchemaForm
                     name={page.pageKey}
-                    title={page.reviewTitle || page.title}
+                    title={page.title.replace('Review ', '')}
```
cehsu: `reviewTitles` have been provided to omit the word "review" from chapter and page titles that include it, this will obviate that.

jbalboni: This change seems very specific to the way the 526 form is using "Review," which might not be the same on other forms. Are there other ways we could handle this without disallowing the word review in any review page title?

cehsu: Good point Jeff, I meant to double check that reviewTitle was only being used to omit the word review, and it turns out it is being used in a variety of ways, so you're correct, this is far too narrow. Instead, title should accept a function as you proposed earlier.

src/js/common/schemaform/review/ReviewCollapsibleChapter.jsx
```diff
@@ -93,6 +93,9 @@ export default class ReviewCollapsibleChapter extends React.Component {
     const hasUnViewedPages = this.pageKeys.some(key => !viewedPages.has(key));
 
     const ChapterDescription = chapter.reviewDescription;
+    let chapterTitle = chapter.title;
+    if (typeof chapter.title === 'function') chapterTitle = chapter.title(true);
+    if (chapter.reviewTitle) chapterTitle = chapter.reviewTitle;
```
cehsu: @jbalboni is this what you had in mind? that title could be fed an argument that indicated whether it was being used on a review page?

jbalboni: Yep, that'll work, though I think having single line `if`s without braces is confusing to read.

jbalboni (approved):
One readability concern, but otherwise this looks good to me.

src/js/common/schemaform/review/ReviewCollapsibleChapter.jsx
cehsu: Good point, updated.

PR #7596, opened 2018-04-16
Reorganize utilities into platform folder

jbalboni:
This is largely moving files from common into platform/utilities, platform/static-data, and platform/forms. The only code changes are to divide up the helpers file into different category.


src/js/common/utils/helpers.js
```diff
@@ -1,274 +0,0 @@
...
-      if (value && suffixes.some(suffix => key.toLowerCase().endsWith(suffix.toLowerCase()))) {
-        return 'removed';
-      }
-
-      return value;
-    });
-  } catch (e) {
-    return null;
-  }
-}
```
ncksllvn: Thank you for separating out all of these helpers

ncksllvn (approved):
LGTM

PR #7599, opened 2018-04-17
526 rated disabilities

cvalarida:
image

image

There was some funny stuff I had to do to make this work, but it does, in fact, work!


src/js/disability-benefits/526EZ/config/form.js
```diff
@@ -121,17 +123,39 @@ const formConfig = {
...
-        pageOne: {
-          title: 'Page One',
-          path: 'chapter-three/page-one',
-          uiSchema: {},
+        ratedDisabilities: {
+          title: 'Your Rated Disabilities',
+          path: 'select-disabilities',
+          uiSchema: {
+            'view:selectDisabilitiesDescription': {
+              'ui:description': 'Please check all the disabilities that you’re filing an increase for because the condition has gotten worse.'
```
jbalboni: You could probably just make this the description on `uiSchema`, instead of nesting it in this empty field.

src/js/common/schemaform/widgets/SelectArrayItemsWidget.jsx
```diff
@@ -0,0 +1,61 @@
...
+  render() {
+    const {
+      value: items,
+      id,
+      options,
+      required
+    } = this.props;
+    const { label: Label, selectedPropName, disabled } = options;
+
+    // TODO: Figure out how to handle required fields
```
jbalboni: This should mostly be handled by StringField, if you mark the field as required as usual.

src/js/disability-benefits/526EZ/config/form.js
```diff
@@ -121,17 +123,39 @@ const formConfig = {
...
+            },
+            disabilities: {
+              // Using StringField because it doesn't do much and we just need to render the widget.
+              // If this becomes a common(ish) pattern, we should make a BasicField or something.
+              'ui:field': 'StringField',
+              'ui:widget': 'SelectArrayItemsWidget',
+              'ui:options': {
```
jbalboni: For validation, I think you can add a validation function in `ui:validations` here that adds an error if none of the items is selected.

src/js/disability-benefits/526EZ/config/form.js
cvalarida: Ohh, right. I tried putting it on the `disabilities` array, but that didn't work (for what are now face-palmy obvious reasons).

src/js/disability-benefits/526EZ/config/form.js
cvalarida: Ahh, okay. I'll give that a shot, thanks!

src/js/common/schemaform/widgets/SelectArrayItemsWidget.jsx
```diff
@@ -0,0 +1,61 @@
+import React from 'react';
+import classNames from 'classnames';
+
+import get from '../../utils/data-utils/get';
+import set from '../../utils/data-utils/set';
+
+// TODO: Safety checks for `selected` callback and `label` element
+
+export default class SelectArrayItemsWidget extends React.Component {
```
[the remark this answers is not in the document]
cvalarida: I can reference it by importing the component and using the variable directly in `formConfig`. I see! Yes, that's nicer. :smile:

peggygannon:
@cvalarida Hi, Chris, can we please make the following content edits: 

1. Please update first sentence to: 
Please choose the disability that you're filing a claim for increase because the condition has gotten worse. 

2. For PTSD I would like to make the sentence in the parenthetical a full sentence. 
Can we change it TO: 
(This could also be claimed as Insomnia.)

3. Invertertebral box
please add a period after L4-L5, L5-S1.)

4. Error message. Please update to: 
Please select at least one condition (or all that apply).

src/js/disability-benefits/526EZ/helpers.jsx
```diff
@@ -276,6 +277,29 @@ export const evidenceSummaryView = ({ formData }) => {
...
+ *
+ * @param {Disability} disability
+ */
+export const disabilityOption = ({ disability }) => {
+  const { diagnosticCode, diagnosticText, ratingPercentage } = disability;
+  // May need to throw an error to Sentry if any of these doesn't exist
+
+  return (
+    <div>
+      {diagnosticCode && <h4>{getDiagnosticCodeName(diagnosticCode)}</h4>}
```
cehsu: If we're not sure we'll receive all of this information, should we provide a default generic h4 text for when the info is not provided? (this can be done later if that's preferable)

cehsu (approved):
LGTM. I had one question about what kind of data availability we can expect, but this can be addressed later if that makes more sense.

src/js/disability-benefits/526EZ/helpers.jsx
cvalarida: Yeah, that's a great question. I'll write a ticket for it since I think there's a larger discussion to be had surrounding if it even makes sense to let them select it if there isn't this kind of information (since the `/submit` endpoint requires it).

src/js/disability-benefits/526EZ/config/form.js
```diff
@@ -183,6 +213,8 @@ const formConfig = {
           title: (formData, { pagePerItemIndex }) => _.get(`disabilities.${pagePerItemIndex}.diagnosticText`, formData),
           path: 'supporting-evidence/:index/evidence-type',
           showPagePerItem: true,
+          // TODO: Remove the nested `disability` when we flatten that
+          itemFilter: (item) => _.get('disability.view:selected', item),
```
markgreenburg: `disability` will get removed in the process of resolving merge conflicts, so will disappear before ratedDisabilities is merged to master?

cvalarida: Ah, right. I have to fix all those `TODO`s now. Forgot about that, thanks. :+1:

src/js/disability-benefits/526EZ/helpers.jsx
```diff
@@ -315,15 +342,18 @@ export const GetFormHelp = () => {
   );
 };
 
+
 export const ITFDescription = (
   <span><strong>Note:</strong> By clicking the button to start the disability application, you’ll declare your intent to file, and this will set the date you can start getting benefits. This intent to file will expire 1 year from the day you start your application.</span>
 );
 
+
```
cehsu: Are these lines accidental?

src/js/disability-benefits/526EZ/components/SelectArrayItemsWidget.jsx
```diff
@@ -0,0 +1,59 @@
...
+  render() {
+    const {
+      value: items,
+      id,
+      options,
+      required
+    } = this.props;
+    const { label: Label, selectedPropName, disabled } = options;
+
+    // Note: Much of this was stolen from CheckboxWidget
```
cehsu: Should this be named ErrorableCheckboxesWidget or CheckboxesWidget, similar to the radio widget naming convention?

cehsu (approved):
LGTM. I had two minor questions.

src/js/disability-benefits/526EZ/components/SelectArrayItemsWidget.jsx
cvalarida: Unlike other fields of the kind, this takes an array of objects and creates "boolean" inputs, so I wanted a name that conveyed that the kind of data it takes is an array of user data.

I'm game to rename, but I don't think `ErrorableCheckboxesWidget` or `CheckboxesWidget` quite fit the bill.

src/js/disability-benefits/526EZ/helpers.jsx
cvalarida: No. I just like two spaces between exports, so I figured while I was already there, I may as well take 10 seconds to add them. :man_shrugging:

src/js/disability-benefits/526EZ/components/SelectArrayItemsWidget.jsx
cehsu: Isn't this the default for checkboxes, that a set of checkboxes would use an array of objects?

cvalarida: I don't think so. I looked at the `CheckboxWidget` we have already and it just made a single input where `checked={typeof value === 'undefined' ? false : value}`. The new widget takes an array and creates a list of inputs where `checked` is based on the value of the `view:selected` field (or custom option) of the underlying object.

cehsu: Ok, I must've been thinking of the other Checkboxes widget used in the wizard, which seems fairly similar, but perhaps it's not the default use case/configuration for checkboxes.

PR #7582, opened 2018-04-13
526 ez primary address

cehsu:
These changes set up the primary address page. Linked issue: https://github.com/department-of-veterans-affairs/vets.gov-team/issues/9278

[Prototype link.](https://adhoc.invisionapp.com/share/9MFXXQTUGDK#/screens/282992019)

There are a number of questions regarding the BE schema made inline.

# Review
image
# Update
localhost_3001_disability-benefits_526_apply-for-increase_review-veteran-details_address-information

PR #7602, opened 2018-04-17
526 use updated schema mg

markgreenburg:
NOTE:
- Depends on https://github.com/department-of-veterans-affairs/vets-json-schema/pull/201
- Requires updating `vets-json-schema` dependency hash in `package.json` before merging (which itself depends on the above PR getting merged)

Updates 526 form config to make existing chapters work with the new common schema / definitions.


src/js/disability-benefits/526EZ/config/form.js
```diff
@@ -1,13 +1,16 @@
...
 import fileUploadUI from '../../../common/schemaform/definitions/file';
-
-import initialData from '../../../../../test/disability-benefits/526EZ/schema/initialData';
-
+import dateRangeUI from '../../../common/schemaform/definitions/dateRange';
 import IntroductionPage from '../components/IntroductionPage';
 import ConfirmationPage from '../containers/ConfirmationPage';
 
+// TODO: Load live user prefill data from network
+import initialData from '../../../../../test/disability-benefits/526EZ/schema/initialData';
```
cehsu: This comment is quite clear already, can we clarify it by adding something like that initialData will be commented out or removed from here for production, and that its presence here is for dev purposes only?

src/js/disability-benefits/526EZ/config/form.js
```diff
@@ -21,55 +24,43 @@ import {
...
-    type: 'number'
-  }
-});
+  return {
+    type,
+    maxItems,
+    items: {
+      type: items.type,
+      properties: _.omit(
+        ['treatmentCenterAddress', 'treatmentCenterType'],
```
cehsu: isn't treatmentCenterType required for vaTreatments?

src/js/disability-benefits/526EZ/config/form.js
```diff
@@ -21,55 +24,43 @@ import {
...
+        items.properties
+      )
+    }
+  };
+
+})(treatments);
```
cehsu: Could we use _.set here instead, if we're only planning to change one property?

src/js/disability-benefits/526EZ/config/form.js
```diff
@@ -240,48 +232,23 @@ const formConfig = {
...
+                    treatmentCenterName: {
+                      'ui:title': 'Name of VA medical facility'
+                    },
+                    treatmentDateRange: dateRangeUI(
+                      'Approximate date of first treatment',
+                      'Approximate date of last treatment',
+                      'Date of last treatment must be after date of first treatment'
+                    ),
+                    // TODO: Put these back as hidden in the UI once typeahead fills this out
+                    // treatmentCenterType: {},
```
cehsu: Should this be included for now or is it not required?

src/js/disability-benefits/526EZ/config/form.js
```diff
@@ -400,58 +365,62 @@ const formConfig = {
...
-                        'startTreatment', 'endTreatment',
-                        'treatmentCenterCountry', 'treatmentCenterStreet1',
-                        'treatmentCenterStreet2', 'treatmentCenterCity',
-                        'treatmentCenterState', 'treatmentCenterPostalCode'
+                        'country',
+                        'addressLine1',
+                        'addressLine2',
+                        'city',
+                        'state',
+                        'zipFirstFive'
```
cehsu: It zipLastFour not required? We will probably want to use the existing UI pattern for postal codes (namely, one field) as I'm not aware of any design work on developing a two field option, and we would probably want to work with design to create this, if we decide to go that route. An example of how this could be set up is in `src/js/disability-benefits/526EZ/pages/primaryAddress.js` in [the primary address PR](https://github.com/department-of-veterans-affairs/vets-website/pull/7582/files). The plan is to move `addressUISchema` and `addressSchema` into a separate address.js file that can be shared across the different chapters.

src/js/disability-benefits/526EZ/config/form.js
```diff
@@ -400,58 +365,62 @@ const formConfig = {
...
+                      city: {
                         'ui:title': 'City'
                       },
-                      treatmentCenterState: {
+                      state: {
                         'ui:title': 'State'
                       },
-                      treatmentCenterPostalCode: {
+                      zipFirstFive: {
                         'ui:title': 'Postal code',
```
cehsu: We will probably want to specify ZIP here, unless @peggygannon indicates postal code is best, and perhaps obvious if we only display this field for addresses that specify 'USA' as the country.

src/js/disability-benefits/526EZ/config/form.js
```diff
@@ -559,7 +515,7 @@ const formConfig = {
                               '1',
                               '2',
                               '3',
-                              // '4',
+                              // '4', // TODO: Confirm this should be taken out
```
cehsu: All of these doc types are temporary and borrowed from another form actually as we are working with @goldenmeanie to produce the relevant list.

src/js/disability-benefits/526EZ/helpers.jsx
```diff
@@ -66,59 +66,59 @@ export const evidenceTypeHelp = (
...
 
-  if (startTreatment && endTreatment) {
-    treatmentPeriod = `${startTreatment} — ${endTreatment}`;
-  } else if (startTreatment || endTreatment) {
-    treatmentPeriod = `${(startTreatment || endTreatment)}`;
+  let treatmentPeriod = '';
+  if (from && to) {
+    treatmentPeriod = `${from} — ${to}`;
+  } else if (from || to) {
+    treatmentPeriod = `${(from || to)}`;
```
cehsu: Are from and two both optional? OR does the date range validation now require both? If the later we can simplify this and remove the conditional logic.

src/js/disability-benefits/526EZ/helpers.jsx
```diff
@@ -66,59 +66,59 @@ export const evidenceTypeHelp = (
...
     <div>
-      <strong>{treatmentCenterName}</strong><br/>
+      <strong>{formData.treatmentCenterName}</strong><br/>
       {treatmentPeriod}
     </div>
   );
 };
 
+
 export const releaseView = ({ formData }) => {
```
cehsu: It looks like this is the same as treatmentView? Can we remove this and use that instead?

src/platform/forms/address.js
```diff
@@ -399,3 +399,223 @@ export function isBlankAddress(address) {
     && isBlank(address.postalCode.value);
 }
 
+export const statesOnlyInPCIU = [
```
cehsu: Is the plan to fetch this information in the future?

src/js/disability-benefits/526EZ/config/form.js
```diff
@@ -151,7 +143,7 @@ const formConfig = {
           }
         },
         evidenceType: {
-          title: (formData, { pagePerItemIndex }) => _.get(`disabilities.${pagePerItemIndex}.diagnosticText`, formData),
+          title: (formData, { pagePerItemIndex }) => _.get(`disabilities.${pagePerItemIndex}.name`, formData),
```
cehsu: @goldenmeanie has indicated that at least one kind of evidence is required. We could use a UI validation to check for this, i think.

test/disability-benefits/526EZ/config/privateRecordReleases.unit.spec.jsx
```diff
@@ -38,23 +42,27 @@ describe('Disability benefits 526EZ VA facility', () => {
...
     );
 
-    fillData(form, 'select#root_privateRecordReleases_0_privateRecordRelease_startTreatmentMonth', '1');
-    fillData(form, 'select#root_privateRecordReleases_0_privateRecordRelease_startTreatmentDay', '3');
-    fillData(form, 'input#root_privateRecordReleases_0_privateRecordRelease_startTreatmentYear', '1950');
-    fillData(form, 'select#root_privateRecordReleases_0_privateRecordRelease_endTreatmentMonth', '1');
-    fillData(form, 'select#root_privateRecordReleases_0_privateRecordRelease_endTreatmentDay', '3');
-    fillData(form, 'input#root_privateRecordReleases_0_privateRecordRelease_endTreatmentYear', '1955');
-    fillData(form, 'input#root_privateRecordReleases_0_privateRecordRelease_treatmentCenterName', 'Local facility');
+    fillData(form, 'select#root_privateRecordReleases_0_treatmentDateRange_fromMonth', '1');
```
cehsu: Now that we are using the standard date widget, we could use the fillDate helper here, if desirable.

cehsu (approved):
LGTM. I had a few questions, but some of these can be split off into tickets to be addressed later. Great job updating the comments with more comments, very helpful.

src/js/disability-benefits/526EZ/config/form.js
markgreenburg: Yes, but we're not going to be able to fill this in without typeahead, so I'm just excluding for now. I accidentally left it off the schema file's `treatments.items.required` property though, so nice catch!

src/js/disability-benefits/526EZ/config/form.js
markgreenburg: I think this is referring to an older version of the code? Lemme know if not.

src/js/disability-benefits/526EZ/config/form.js
markgreenburg: Seems like a good idea, assuming we can confirm this is the case. I'd like to push that off into another PR.

src/js/disability-benefits/526EZ/config/form.js
markgreenburg: Not required, for now. We'll have to make sure we make `treatmentCenterType` required again once we get typeahead working.

src/js/disability-benefits/526EZ/config/form.js
markgreenburg: Yeah, I definitely don't want to suggest using a UI with two postal code fields - that seems bad. The easiest thing to do would probably be to submit as one field and let `vets-api` split it into the two pieces needed by the `submit` endpoint. I'll confirm this with the BE devs and then change the validations / names / properties here (as well as in the `schema.js`).

src/js/disability-benefits/526EZ/config/form.js
markgreenburg: Just referenced the relevant page in the current mocks to get this display label. Happy to change if that's not what we want.

src/js/disability-benefits/526EZ/helpers.jsx
markgreenburg: Yeah it's actually 'from' that's required (but not 'to'). I think it's more clear (if slightly overkill) to leave it as-is so that someone down the line doesn't have to wonder about why we didn't handle the 'from is missing' case. I just updated this requirement in the schema.

src/js/disability-benefits/526EZ/helpers.jsx
markgreenburg: Yep, perfect!

src/js/disability-benefits/526EZ/helpers.jsx
cehsu: Thanks for the context @markgreenburg, very helpful.

src/platform/forms/address.js
markgreenburg: Yeah, we'd probably fetch the full state and country list. In the meantime, I compared what we fetch in Letters to the existing constants state list and the `statesOnlyInPCIU` is the difference between the two. The countries are totally different in shape and formatting so I just copied those over from the endpoint response.

test/disability-benefits/526EZ/config/privateRecordReleases.unit.spec.jsx
markgreenburg: Oh, see, I didn't even know that existed! Seems like a good idea.

src/js/disability-benefits/526EZ/config/form.js
```diff
@@ -269,48 +264,23 @@ const formConfig = {
...
-                    }
+                    treatmentCenterName: {
+                      'ui:title': 'Name of VA medical facility'
+                    },
+                    treatmentDateRange: dateRangeUI(
+                      'Approximate date of first treatment',
+                      'Approximate date of last treatment',
+                      'Date of last treatment must be after date of first treatment'
+                    ),
+                    // TODO: Put these back as hidden in the UI once typeahead fills this out
```
cvalarida: It's probably find to do this before the typeahead fills these fields out, but I'd say don't worry about it for now. Can you write a ticket to make sure this happens later?

src/js/disability-benefits/526EZ/config/form.js
```diff
@@ -324,7 +294,7 @@ const formConfig = {
                 items: {
                   type: 'object',
                   properties: {
-                    treatments: treatmentsSchema
+                    vaTreatments
```
cvalarida: I forget--does the new 526 `/submit` schema call it `vaTreatments`?

src/js/disability-benefits/526EZ/config/form.js
```diff
@@ -426,61 +394,65 @@ const formConfig = {
...
+                    },
+                    treatmentDateRange: dateRangeUI(
+                      'Approximate date of first treatment',
+                      'Approximate date of last treatment',
+                      'Date of last treatment must be after date of first treatment'
+                    ),
+                    privateMedicalRecordsReleaseRestricted: {
+                      'ui:title': 'I give my consent, or permission, to my doctor to only release records related to this condition'
+                    },
+                    'view:releaseRestrictedNotice': {
```
cvalarida: I just wrote a note to remind myself to write a ticket to do something with this... (It'll have to go back to the api to be put into the 4142 or 4142a, I think, but it's fine as a `view:` field for now.)

src/js/disability-benefits/526EZ/config/form.js
cvalarida: We may remove this document type altogether if it's not useful when we upload. (Sounded like only the name was because whomever reviews these doesn't trust the document type...but perhaps that was just for additional evidence?)

src/js/disability-benefits/526EZ/helpers.jsx
```diff
@@ -146,8 +127,9 @@ export const privateRecordsChoiceHelp = (
 
 
 export const privateMedicalRecordsIntro = ({ formData }) => {
+  const firstOrNext = formData['view:vaMedicalRecords'] ? 'next' : 'first';
```
cvalarida: Excellent! I was going to write a ticket for this. :smile:

test/disability-benefits/526EZ/config/vaFacilities.unit.spec.jsx
```diff
@@ -38,23 +38,23 @@ describe('Disability benefits 526EZ VA facility', () => {
...
     );
 
-    fillData(form, 'select#root_treatments_0_treatment_startTreatmentMonth', '1');
-    fillData(form, 'select#root_treatments_0_treatment_startTreatmentDay', '3');
-    fillData(form, 'input#root_treatments_0_treatment_startTreatmentYear', '1950');
-    fillData(form, 'select#root_treatments_0_treatment_endTreatmentMonth', '1');
-    fillData(form, 'select#root_treatments_0_treatment_endTreatmentDay', '3');
-    fillData(form, 'input#root_treatments_0_treatment_endTreatmentYear', '1955');
-    fillData(form, 'input#root_treatments_0_treatment_treatmentCenterName', 'Local facility');
+    fillData(form, 'select#root_vaTreatments_0_treatmentDateRange_fromMonth', '1');
```
cvalarida: +1 on using `.fillDate` in this file too.

cvalarida (approved):
Looks good! I had a bunch of comments, but I don't think anything is blocking. We just need to make sure we keep track of our TODO list.

src/js/disability-benefits/526EZ/config/form.js
```diff
@@ -21,56 +25,46 @@ import {
...
-    type: 'number'
-  }
-});
+  return {
+    type,
+    maxItems,
+    items: {
+      type: items.type,
+      // TODO: use standard required property once treatmentCenterType added
+      // back in schema (because it's required)
```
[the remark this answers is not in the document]
markgreenburg: Yes, it's a required field in `submit`, so we'll need to figure out how to get it in there before using real data.

src/js/disability-benefits/526EZ/config/form.js
markgreenburg: Still `treatments` in the schema  :)

PR #7604, opened 2018-04-17
content updates

DanielleSoCompany:


melwoodard:
@ncksllvn any chance you could add a wrapper around all of the text so it's all a consistent width? It looks like only the intro-text is subject to our standard max-width for content, everything else is too wide.

The cutoff should be about here:

try_our_new_beta_tools__vets_gov

ncksllvn:
@melwoodard no problem! Will do that shortly

melwoodard (approved):
Looks great! Thanks @DanielleSoCompany @ncksllvn!

ncksllvn:
@DanielleSoCompany I believe this branch was created off of a very old version of master branch so I recommend doing a `git pull`. I believe that's why the navigation and enroll button weren't loading.

PR #7605, opened 2018-04-17
remove legend focusing behavior

rianfowler:
I tested this on a couple forms- it changes the default behavior to focus on the progress bar test when changing pages.

This behavior was originally added as part of this [issue](https://github.com/department-of-veterans-affairs/vets-website/pull/4715). @jbalboni I want to make sure that there aren't specific test cases we should regression test on i.e. was the legend just an add-on or does it act as a replacement for the `.nav-header` in some places.

### DIFF
--- a/src/js/common/schemaform/containers/FormPage.jsx
+++ b/src/js/common/schemaform/containers/FormPage.jsx
@@ -14,12 +14,7 @@ import { getNextPagePath, getPreviousPagePath } from '../routing';
 import { focusElement } from '../../../../platform/utilities/ui';
 
 function focusForm() {
-  const legend = document.querySelector('.form-panel legend:not(.schemaform-label)');
-  if (legend && legend.getBoundingClientRect().height > 0) {
-    focusElement(legend);
-  } else {
-    focusElement('.nav-header');
-  }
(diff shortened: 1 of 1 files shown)


jbalboni (approved):
This looks fine to me, would like to get Trevor's opinion about this being the right element to set to focus to.

1Copenut (approved):
@rianfowler This behavior feels spot on for screenreaders and keyboard users. I'll start with keyboard:

For keyboard navigation, I want to avoid the nav bar and header if possible. By settling on the progress bar, I feel users will acclimate quickly to the expectation when they see the bars, it will be the first focus on the page. Filling out the form was intuitive and tab order felt good.

Screenreader users are benefitting from this as well, because the step number, and description is being read out. I tested this in VoiceOver + Safari + Mac, and NVDA + Firefox + Win7. I suspect JAWS will do the same thing, and can test it in Staging when we have a non-localhost URL. This is a big benefit to non-sighted users because now they know where they are in the process, and what they will be doing on this page.

This is a nice win for 508 and lots of users.

rianfowler:
Thank you @1Copenut

PR #7606, opened 2018-04-17
Personalization: Fix for redirect not occurring on higher environments

ncksllvn:
There is a redirect in place for beta users so that anyone who clicks a link to `/`, that link should be stopped, and then the user should be taken to `/dashboard` instead. The homepage should never be visible. However, something weird was going on and I think the page was reloading before the event handler could catch it. I'm switching to the `click` event instead, which is supposed to be more aggressive, although I'm not sure if that fires on mobile.

Resolves department-of-veterans-affairs/vets.gov-team/issues/9967


bshyong (approved):
LGTM. FYI click will catch right clicks as well

PR #7611, opened 2018-04-18
Cleaned up login widget components

U-DON:
- Created a `SignInModal` component so that its content isn't its own component (previously `SignIn.jsx`).
- Converted `SearchHelpSignIn` from container to component.
- Simplified, consolidated, and moved similar automatic modal and redirect-related logic from `SignInModal` and `SearchHelpSignIn` into `Main.jsx`.

For a follow-up PR, I will move the `login` folder to `platform`, similar to what was done with `feedback`.


src/js/login/containers/Main.jsx
```diff
@@ -111,12 +136,18 @@ const mapStateToProps = (state) => {
...
       dispatch(toggleLoginModal(update));
     },
+    toggleSearchHelpUserMenu: (menu, isOpen) => {
+      dispatch(toggleSearchHelpUserMenu(menu, isOpen));
+    },
+    updateLoggedInStatus: (update) => {
+      dispatch(updateLoggedInStatus(update));
+    },
     getUserData: () => {
       getUserData(dispatch);
```
U-DON: This is the reason this `mapDispatchToProps` is so verbose. `getUserData` is not currently an action, so it needs to receive `dispatch` as an argument, but it would be more appropriate to have it as an action under `user-profile`. Once that's done, this `mapDispatchToProps` can just be an object with direct mappings.

src/js/login/components/SearchHelpSignIn.jsx
```diff
@@ -1,103 +1,100 @@
...
-    const { currentlyLoggedIn, showModal } = this.props.login;
-    const isModalStillClosed = !prevProps.login.showModal && !showModal;
-    if (!currentlyLoggedIn && isModalStillClosed) {
-      const nextParams = new URLSearchParams(window.location.search);
-      const nextPath = nextParams.get('next');
-      if (nextPath) {
-        this.props.toggleLoginModal(true);
-      }
-    }
-  }
```
U-DON: This logic is now handled in the `componentDidUpdate` method of `Main.jsx`.

src/js/login/components/SearchHelpSignIn.jsx
```diff
@@ -1,103 +1,100 @@
...
   }
 
-  hasSession() {
+  handleMenuClick = (menu) => (() => {
+    this.props.toggleMenu(menu, !this.props.isMenuOpen[menu]);
+  });
+
+  handleSearchMenuClick = this.handleMenuClick('search');
+  handleHelpMenuClick = this.handleMenuClick('help');
+  handleAccountMenuClick = this.handleMenuClick('account');
```
U-DON: Split into specific handlers so we don't redefine on each render cycle.

src/js/login/containers/Main.jsx
```diff
@@ -32,14 +55,18 @@ export class Main extends React.Component {
...
+
   bindNavbarLinks = () => {
     [...document.querySelectorAll('.login-required')].forEach(el => {
       el.addEventListener('click', e => {
-        e.preventDefault();
-        const nextQuery = { next: el.getAttribute('href') };
-        const nextPath = appendQuery('/', nextQuery);
-        history.pushState({}, el.textContent, nextPath);
-        this.props.toggleLoginModal(true);
+        if (!this.props.login.currentlyLoggedIn) {
```
U-DON: If we know the user is logged in, no need to send to "?next=..." for `componentDidUpdate` to then redirect accordingly. Can just treat it as a normal link in that case.

dhall (commented):
LGTM, but I'll defer to others for approval

src/js/login/containers/Main.jsx
```diff
@@ -3,12 +3,17 @@ import { connect } from 'react-redux';
 import appendQuery from 'append-query';
 
 import recordEvent from '../../../platform/monitoring/record-event';
-import Modal from '@department-of-veterans-affairs/jean-pants/Modal';
 import { getUserData } from '../../common/helpers/login-helpers';
+import { isUserRegisteredForBeta } from '../../personalization/beta-enrollment/actions';
```
jbalboni: This function is not an action, but it's written as one. Not something you need to fix, but it was confusing when reading through the code.

src/js/login/components/SearchHelpSignIn.jsx
```diff
@@ -1,103 +1,100 @@
...
-
+class SearchHelpSignIn extends React.Component {
   handleSignInSignUp = (e) => {
     e.preventDefault();
     recordEvent({ event: 'login-link-clicked' });
     this.props.toggleLoginModal(true);
   }
 
-  hasSession() {
+  handleMenuClick = (menu) => (() => {
```
jbalboni: This is somewhat hard to read for me. I'm not entirely sure we need the auto-binding part of this (I think that's happening on lines 22-24) and could maybe do something like this:

```js
handleMenuClick(menu) => {
  return () => this.props.toggleMenu(menu, !this.props.isMenuOpen[menu]);
}
```

Not sure, though.

src/js/login/containers/Main.jsx
```diff
@@ -32,14 +55,18 @@ export class Main extends React.Component {
     if (event.data === sessionStorage.userToken) { this.props.getUserData(); }
   }
 
+  getRedirectUrl = () => (new URLSearchParams(window.location.search)).get('next');
```
jbalboni: URLSearchParams isn't imported in this file and was removed from another file. That might be breaking things in IE? We could add it to our polyfills file.

jbalboni (approved):
Couple minor comments, but LGTM. I didn't have any trouble running it locally.

src/js/login/containers/Main.jsx
U-DON: I agree. Doesn't end up dispatching anything and simply checks something in the current state. But looks like something the personalization team was working on.

src/js/login/containers/Main.jsx
U-DON: Forgot to include it in this file. But adding to polyfills makes sense.

src/js/login/components/SearchHelpSignIn.jsx
U-DON: I was going for a more functional style, like
```
someFunction = (arg) => (anotherFunction);
```
, but it is a little harder to read without the `return`, so I can rewrite it to be more readable.

I did the binding on the following lines because I didn't want to recreate those functions on each render cycle (like when a menu is toggled).

PR #7615, opened 2018-04-18
WIP - 3946 review page

rianfowler:
- Moved some stuff around and eliminated the need for `ReviewPage`
- added `viewedPages` state to redux
- moved state querying logic out of components' render and into `mapStateToProps`

TODO
- [ ] update unit tests 
- [ ] test rendering perf difference on inputs
- [ ] remove `ReviewPage`
- [ ] test scrolling behavior 

reviewpage


src/js/common/schemaform/save-in-progress/RoutedSavableReviewPage.jsx
```diff
@@ -86,56 +191,136 @@ class RoutedSavableReviewPage extends React.Component {
...
+    const showUnviewedPageWarning = pageKeys.some(key => !viewedPages.has(key));
+
+    chaptersAcc.push({
+      expandedPages,
+      formConfig: chapterFormConfig,
+      name: chapterName,
+      open,
+      pageKeys,
+      showUnviewedPageWarning
+    });
```
rianfowler: I like to avoid creating objects in the props but it's a good idea for arrays of views. This is probably the biggest part of the refactor because this data was in many places across three components. 

Consolidating this makes it easier to figure out what should go into the `viewState` and what can be memoized with selectors.

src/js/common/schemaform/save-in-progress/RoutedSavableReviewPage.jsx
```diff
@@ -86,56 +191,136 @@ class RoutedSavableReviewPage extends React.Component {
...
+
+  // from redux state
+  const form = state.form;
+  const formData = state.form.data;
+  const openChapters = getReviewPageOpenChapters(state);
+  const user = state.user;
+  const viewedPages = getViewedPages(state);
+
+  const chapterNames = getActiveChapters(formConfig, formData);
+  const disableSave = formConfig.disableSave;
```
rianfowler: Pulled this decision out of the router. There are only a couple of places where the behavior diverged between `ReviewPage` and `RoutedSavableReviewPage` and I thought it'd be easier to consolidate them into a single component.

src/js/common/schemaform/save-in-progress/RoutedSavableReviewPage.jsx
```diff
@@ -86,56 +191,136 @@ class RoutedSavableReviewPage extends React.Component {
...
+                key={chapter.name}
+                onEdit={this.handleEdit}
+                open={chapter.open}
+                pageKeys={chapter.pageKeys}
+                setData={this.setData}
+                setValid={setValid}
+                showUnviewedPageWarning={chapter.showUnviewedPageWarning}
+                toggleButtonClicked={() => this.handleToggleChapter(chapter)}
+                uploadFile={this.props.uploadFile}
+                viewedPages={viewedPages}/>
```
rianfowler: Was able to dumb down this component a bit by pulling the querying stuff into `mapStateToProps`. I think scroll behavior should be consolidated before turning this into a couple of nested stateless functional components (later PR). This clean up will make that task easier.

src/js/common/schemaform/review/ReviewCollapsibleChapter.jsx
```diff
@@ -211,7 +210,7 @@ export default class ReviewCollapsibleChapter extends React.Component {
                 onClick={this.props.toggleButtonClicked}>
                 {chapterTitle}
               </button>
-              {hasUnViewedPages && <span className="schemaform-review-chapter-warning-icon"/>}
+              {showUnviewedPageWarning && <span className="schemaform-review-chapter-warning-icon"/>}
```
rianfowler: More explicit naming here.

src/js/common/schemaform/review/ReviewCollapsibleChapter.jsx
```diff
@@ -86,20 +82,23 @@ export default class ReviewCollapsibleChapter extends React.Component {
   render() {
     let pageContent = null;
 
-    const { form, pages, viewedPages, chapter, formContext } = this.props;
-    const activePages = getActivePages(pages, form.data);
-    const expandedPages = expandArrayPages(activePages, form.data);
-
-    this.pageKeys = getPageKeys(pages, form.data);
-    const hasUnViewedPages = this.pageKeys.some(key => !viewedPages.has(key));
```
rianfowler: All of this was moved up into `mapStateToProps` and passed down as a prop.

src/js/common/schemaform/review/ReviewCollapsibleChapter.jsx
```diff
@@ -32,9 +31,6 @@ export default class ReviewCollapsibleChapter extends React.Component {
     if (!oldProps.open && this.props.open) {
       this.scrollToTop();
     }
-    if (oldProps.open && !this.props.open) {
-      this.props.setPagesViewed(this.pageKeys);
```
rianfowler: This now triggers off of the close action.

src/js/common/schemaform/save-in-progress/RoutedSavableReviewPage.jsx
rianfowler: I also think some of this naming could be tweaked. I would like it to be clearer what's going into the component renderer vs what is used for decisions. Right now it's just a

src/js/common/schemaform/save-in-progress/RoutedSavableReviewPage.jsx
```diff
@@ -46,6 +111,46 @@ class RoutedSavableReviewPage extends React.Component {
...
+      this.props.setSubmission('hasAttemptedSubmit', true);
+    }
+  }
+
+  handleToggleChapter({ name, open, pageKeys }) {
+    if (open) {
+      this.props.closeReviewChapter(name, pageKeys);
+    } else {
+      this.props.openReviewChapter(name);
+    }
```
rianfowler: was able to simplify this more by explicitly passing the `open` state on the props. Now this is just an `if` statement and should be easier to unit test.

jbalboni:
@rianfowler The reason both ReviewPage and RoutedSavableReviewPage exist is so that the forms code can work without depending on save in progress. SiP is very specific to vets.gov, but the forms code is going to be pulled out into a separate library soon.

So, we do need a version of ReviewPage that works without save in progress functionality. RoutedSavableReviewPage doesn't necessarily need to build on top of it, whatever way of making the basic review page functionality reusable is ok with me.

rianfowler:
@jbalboni This should work as a regular review page. [This](https://github.com/department-of-veterans-affairs/vets-website/pull/7615/files#r182599414) passes the flag that's used in the router as a prop and switches on the SIP behavior: 

screen shot 2018-04-19 at 09 38 27
screen shot 2018-04-19 at 09 41 42

I can slice it up a different way. I want to avoid duplicating the prop interface between the two components since I'm growing `mapStateToProps` and trying to drive everything with props.
- One approach could be to always use `ReviewPage` as a connected component with some optional `ownProps` the SIP could hook into. This would allow other SIP implementations to be easily used in the future. 
- Another approach could be to pull out the `ReviewPage` connect component into a `@decorator` that the review page and SIP use. Could be inline on the `ReviewPage` and imported by the SIP component and just passed down as `{...props}` (I think this is how it was originally done.

jbalboni:
@rianfowler We're going to have a situation soon where the review page components are going to live in the us-forms-system repo, but our save in progress code will stay in here. So we need to have a review page that doesn't import any save in progress code. The SiP functionality needs to wrap or proxy the standard review page behavior in some way.

src/js/common/schemaform/save-in-progress/RoutedSavableReviewPage.jsx
jbalboni: I definitely like having all this in the mapStateToProps function. It makes all the transformation that's going on much clearer. I think we could also group some of this into reselect selectors to make it more efficient, since it looks like some of this calculation is unnecessary for every state change.

PR #7616, opened 2018-04-19
Fix the lint:js:changed task

jbalboni:
It looked like some of the escaping was messed up, but this seems to work and I think is clearer about what the intent is. With this change I still see JS changes picked up, but no scss changes.


cvalarida (approved):
I'm gonna be honest, I don't know _why_ this works, but it works. :rocket:

PR #7524, opened 2018-04-03
Set up 526 introduction page

cehsu:
These changes update the 526 introduction page.

TODO:
- [x] wire up mock prefill/ITF submission
- [x] write tests

Unauthenticated
localhost_3001_disability-benefits_526_apply-for-increase_introduction 4

Authenticated
localhost_3001_disability-benefits_526_apply-for-increase_introduction_ 3

Verified
localhost_3001_disability-benefits_526_apply-for-increase_introduction 5

Loading
image

Error
image

PR #7620, opened 2018-04-19
Disable ITF mock

cehsu:
The 526 ITF mock (introduced in PR [7524](https://github.com/department-of-veterans-affairs/vets-website/pull/7524/files)) is buggy and needs to be fixed, these changes revert the FormStartControls to their previous state while that is underway.


markgreenburg (approved):
LGTM; also ran locally without any problems.

PR #7622, opened 2018-04-19
526 user testing fixes

cehsu:
These changes remove the dropdown from document uploads, and update some paths and page depends.

image

image


src/js/disability-benefits/526EZ/config/form.js
```diff
@@ -112,7 +112,8 @@ const formConfig = {
         // paymentInformation:
         specialCircumstances: {
           title: 'Special Circumstances',
-          path: 'special-circumstances',
+          path: 'review-veteran-details/special-circumstances',
+          depends: formData => formData.prefilled,
```
markgreenburg: Not sure I understand the implication of this - does this mean that we don't want to show the page at all if the user has any prefilled data? I'm probably just missing something since this page seems to work as expected when paging through the form.

cehsu: There are two versions of the first chapter, verified and unverified. These depends ensure that users are routed to the appropriate version.

markgreenburg (approved):
LGTM 🚀 
Seems like we'll need to figure out which actual vets-api endpoint to hook this up to later.

cvalarida (approved):
Looks good. It'll be a bit weird until we get the view field PR merged in, but it shouldn't break in the meantime.

PR #7623, opened 2018-04-19
migrate sortable table

rianfowler:
Tried testing on the review instance but none of the accounts I tried have messages and the messaging and prescription apps don't work. 
screen shot 2018-04-19 at 18 01 42

screen shot 2018-04-19 at 18 03 05


jbalboni:
@rianfowler +264 works with rx/messaging for me locally.

jbalboni (approved):

PR #7624, opened 2018-04-20
Remove outdated components

jbalboni:
This does a few (minor) things:

- Remove some unused components
- Deletes temporary code devoted to feature flags. We very rarely use these and I think explaining them in documentation is better than having dummy code
- Moves ToolTip into the GIBCT folder. No other app uses this component and it's not a current pattern in use by designers.
- Moves some files that were missed in the previous re-organization.


rianfowler (approved):

README.md
```diff
@@ -298,11 +298,6 @@ all builds regardless of the build type. The important distinction is that your
 feature is still active within the code base, but the UI is either enabled or
 disabled by the feature flag.
 
-To enable or disable the feature in a specific build type, toggle the feature
-in `src/platform/testing/unit/mocha-setup.js` and `config/webpack.config.js`. See
-[`SampleFeature`](src/js/common/components/SampleFeature.jsx) and the associated `__SAMPLE_FEATURE__` env variables for an
-example implementation.
-
```
ncksllvn: I think I recall reading this awhile back but yeah, never used 🤷‍♂️

ncksllvn (approved):
LGTM!

PR #7625, opened 2018-04-20
add some typecasts to error handler

rianfowler:
Jumped the gun a bit still need to 
- [x] update the unit tests

### DIFF
--- a/src/js/common/schemaform/actions.js
+++ b/src/js/common/schemaform/actions.js
@@ -161,7 +161,7 @@ export function submitForm(formConfig, form) {
       .catch(errorReceived => {
         // overly cautious
         const error = errorReceived instanceof Error ? errorReceived : new Error(errorReceived);
-        const errorMessage = String(_.get('message', error));
+        const errorMessage = String(error.message);
         let errorType = 'clientError';
         if (errorMessage.startsWith('vets_throttled_error')) {
           errorType = 'throttledError';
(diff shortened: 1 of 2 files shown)


src/js/common/schemaform/actions.js
```diff
@@ -158,9 +158,10 @@ export function submitForm(formConfig, form) {
 
     return promise
       .then(resp => dispatch(setSubmitted(resp)))
-      .catch(error => {
+      .catch(errorReceived => {
         // overly cautious
-        const errorMessage = _.get('message', error);
+        const error = errorReceived instanceof Error ? errorReceived : new Error(errorReceived);
```
rianfowler: Everything downstream expects this to be of type `Error`. If it's `undefined`, a `string`, or an `object` this will force it something that won't blow up.

src/js/common/schemaform/actions.js
```diff
@@ -158,9 +158,10 @@ export function submitForm(formConfig, form) {
...
+        const errorMessage = String(_.get('message', error));
```
rianfowler: Typecast the error message in case it's `undefined`

src/js/common/schemaform/actions.js
```diff
@@ -158,9 +158,10 @@ export function submitForm(formConfig, form) {
 
     return promise
       .then(resp => dispatch(setSubmitted(resp)))
-      .catch(error => {
+      .catch(errorReceived => {
         // overly cautious
```
rianfowler: now this comment is accurate.

jbalboni (approved):

PR #7626, opened 2018-04-20
Move form styles, clean up site wide styles

jbalboni:
I've noted in the PR why things are happening, but the gist is that I'm moving form styles to the right place and moving styles out of style.scss (used on every page) to static-pages.scss (used on static pages) where appropriate.


src/js/static-pages/sass/static-pages.scss
```diff
@@ -6,6 +6,11 @@
 @import "layouts/l-playbook";
 
 // ----- Modules that only affect non-React pages. ---- //
+@import "modules/m-card";
```
jbalboni: These are all playbook styles or for features that are only used in the static pages. I searched for the classes from each file and did not find other uses.

src/sass/modules/_m-nav-channel.scss
```diff
@@ -1,48 +0,0 @@
-//============================================
```
jbalboni: I could not find these styles in use anywhere.

src/sass/modules/_m-nav-sidebar.scss
```diff
@@ -1,95 +0,0 @@
-//====================================
```
jbalboni: Getting from jean-pants

src/sass/style.scss
```diff
@@ -19,12 +15,9 @@ $image-path: "~uswds/src/img";
 @import "modules/m-vet-nav";
 @import "~@department-of-veterans-affairs/jean-pants/sass/modules/m-print";
 @import "~@department-of-veterans-affairs/jean-pants/sass/modules/m-dropdown";
-@import "modules/m-nav-sidebar";
 @import "modules/m-overlay";
-@import "modules/m-education-wizard";
-@import "modules/m-applications";
 @import "~@department-of-veterans-affairs/jean-pants/sass/modules/m-additional-info";
-@import "modules/m-downtime-notification";
+@import "../platform/monitoring/DowntimeNotification/sass/m-downtime-notification";
```
jbalboni: Maybe this should be imported by React app styles? Not sure, so I left it alone.

rianfowler (approved):

src/sass/style.scss
ncksllvn: ^ Good call, I can add that into the DowntimeNotification directory. Should I do that in this branch or hold off?

ncksllvn (approved):

src/sass/style.scss
jbalboni: It's in the right directory now, but I was thinking that we could just import this in the sass files for each React app that uses it. But it's only a few lines, so I don't think it matters that much.

PR #7629, opened 2018-04-20
Personalization: Require login only after enrollment button clicked

ncksllvn:
Resolves department-of-veterans-affairs/vets.gov-team/issues/10107
Resolves department-of-veterans-affairs/vets.gov-team/issues/10105


bshyong (approved):
LGTM

PR #7633, opened 2018-04-20
Personalization: Better redirect for homepage/dashboard

ncksllvn:


src/js/login/components/BetaDropdown.jsx
```diff
@@ -23,18 +23,12 @@ class BetaDropdown extends React.Component {
...
-    if (event.target.tagName.toLowerCase() === 'a' && event.target.pathname === '/' && event.which === LEFT_CLICK) {
-      event.preventDefault();
-      event.stopPropagation();
-
-      this.redirectToDashboard();
+    const target = event.target;
+    if (target.tagName.toLowerCase() === 'a' && target.pathname === '/' && event.which === LEFT_CLICK) {
+      target.href = '/dashboard-beta';
     }
   }
```
ncksllvn: Even though it's mutating the event.target, this seems like a really solid way to do the redirect. What do you think?

ncksllvn: Changing to the `click` event solved the problem already (with redirects not working in higher environments) but I still think this is worth updating as well

bshyong (approved):

PR #7640, opened 2018-04-23
526 disability increase view bug fix

cehsu:
These changes are a temporary bug fix for the continue button not validating the verified review form before navigating to the next page, until the Verified Review components are refactored, which is planned [here](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/10132).

screen shot 2018-04-23 at 11 09 10 am


cvalarida (approved):
Looks good! I can probably re-use some of this in that `ReviewCardField` I'm working on. :+1:

markgreenburg:
@cehsu FWIW I get an infinite loop between `military-service-history` and `special-circumstances`

cehsu:
Thanks @markgreenburg, that's a known bug and the fix hasn't been moved from the user test review branch yet. The fix is should be in the open address and payment information PRs, or will be added to them if not.

PR #7642, opened 2018-04-23
526 disability increase array field refactor

cehsu:
These changes update the review mode logic so that it is based on data availability.

screen shot 2018-04-25 at 11 55 46 am

This includes:
 - prefilled items are shown in a condensed view
 - invalid or empty prefilled items are shown in an unfurled view
 - empty items are no longer able to be added (in cases where schemas have no required fields)

Therefore:

1. valid items are now collapsed by default (this specifically affects the last item, which was previously displayed in edit mode with the white background "add another item" styling by default)
image

2. any invalid/incomplete items (including non-last items) are now displayed in edit mode
image

3. prefilled empty items with schemas that do not contain any required fields are displayed unfurled. This scenario is encountered if a user "adds another" saves the form, and returns later to continue filling the form out.
screen shot 2018-05-30 at 5 56 57 pm

4. empty items can no longer be added (in cases where schemas do not contain required fields, and no validation logic already prevents this)
This helps to determine whether empty items with schemas that do not contain any required fields are displayed properly. This scenario is encountered if a user "adds another" saves the form, and returns later to continue filling the form out.
screen shot 2018-05-31 at 10 02 25 am


cehsu:
@cvalarida @markgreenburg is this consistent with what we discussed yesterday? Looping in @jbalboni for visibility.

src/js/common/schemaform/fields/ArrayField.jsx
```diff
@@ -192,7 +192,7 @@ export default class ArrayField extends React.Component {
     const DescriptionField = typeof description === 'function'
       ? uiSchema['ui:description']
       : null;
-    const isReviewMode = uiSchema['ui:options'].reviewMode;
+    const isReviewMode = formData && formData.length;
```
cvalarida: Can you remind me again what review mode does?

jbalboni: I'm also not sure I understand why this change is being made.

cehsu: The 526 form will display the review form of the array field if data is available (prefilled), rather than the standard form view initially. 
screen shot 2018-04-25 at 11 55 46 am
If we want to hold off on this for the rest of the forms for now, perhaps we should create a 526 specific array field?

jbalboni: Yeah, I think this will have some negative effects on other forms, so we should come up with a different way of achieving it.

Off the top of my head, we might be able to add an option to ArrayField that starts with all items in the read only state if some condition is met.

cehsu: Sounds good, thanks @jbalboni, we will need to pass the formConfig via routes in order to have access to that form config option, I'll push up those changes for your consideration.

cehsu: We currently have it set up to specify this in uiSchema, but array field needs to be passed the formConfig from routes rather than from form in order to access that information. I'll make the changes to provide that access, and then update the reviewMode logic to also require that config setting.

src/js/common/schemaform/fields/ArrayField.jsx
```diff
@@ -192,7 +192,9 @@ export default class ArrayField extends React.Component {
     const DescriptionField = typeof description === 'function'
       ? uiSchema['ui:description']
       : null;
-    const isReviewMode = uiSchema['ui:options'].reviewMode;
+
+    const editingState = this.props.uiSchema['ui:options'].reviewMode;
+    const isReviewMode = formData && formData.length && editingState;
```
cehsu: @jbalboni Arrrayfield already had the necessary access to uiSchema, so have simply included editingState here.

cvalarida:
When there are no items in the array and `reviewMode` is true, this is what happens:

image

This is because the `formData` passed to `ArrayField` looks like
```javascript
[{ dateRange: {} }]
```
I forget why this happens, to be honest, but it was for good reason, I'm sure. _(Waves hands.)_

Line 197 then returns true:
```javascript
const isReviewMode = formData && formData.length && editingState;
```

So we're passing a basically-empty-but-not-really-empty object to the `viewField` for the first service period.

jbalboni:
I don't think the right change here to change how the review mode flag works. I think the thing we want to tweak is how the last item is handled when data exists for the array.

So I would make a flag in `ui:options` called something like `showLastItemInViewMode` and use that to tweak the array field component logic where `isLast` is used to show the last item in view mode.

jbalboni:
`reviewMode` is the prop we use to determine what set of widgets to display, the review ones or the normal ones. I'd rather not rename that, since it lines up with the review page wording.

src/applications/common/schemaform/fields/ArrayField.jsx
```diff
@@ -192,14 +192,22 @@ export default class ArrayField extends React.Component {
...
+      showLastItemInViewMode: viewMode
+    } = uiOptions;
     const hasTitleOrDescription = (!!title && !hideTitle) || !!description;
 
     // if we have form data, use that, otherwise use an array with a single default object
     const items = (formData && formData.length)
       ? formData
       : [getDefaultFormState(schema, undefined, registry.definitions)];
-
+    const lastItem = items[items.length - 1];
```
cvalarida: For some reason, I'm having a hard time follow this added chunk. @cehsu Can you explain what this doing?

cvalarida (commented):
Hmm...I found a bug. Steps to reproduce:
1) Go to (military service history page)[http://localhost:3001/disability-benefits/526/apply-for-increase/review-veteran-details/military-service-history]
    - The page is pre-filled with a single service period
    - It has a validation error, so it starts in edit mode like it should
2) Enter the start date (which is required but is currently blank)
    - It immediately goes to the view mode as soon as I enter the year
3) Hit the edit button
    - I get sent to the next page (payment information)

cvalarida:
Looks like Jenkins passed, which means none of the tests caught the bug I found. We should fix that.

I think this is a better approach than re-using the `reviewMode`, so I'd say it's heading in the right direction! Just needs some more work is all.

src/applications/common/schemaform/fields/ArrayField.jsx
cehsu: This is checking that the last time (the item this change effects), is valid, to determine whether the showlastItemInViewMode configuration should be applied or not (if it is not valid).

cehsu: Would it make sense to add some comments here to clarify?

src/applications/common/schemaform/fields/ArrayField.jsx
```diff
@@ -44,6 +45,19 @@ export default class ArrayField extends React.Component {
...
+    const items = (formData && formData.length)
+      ? formData
+      : [getDefaultFormState(schema, undefined, registry.definitions)];
+    const lastItem = items[items.length - 1];
+    this.showInViewMode = !!(uiSchema['ui:options'].showLastItemInViewMode && lastItem);
+    const lastIndex = this.props.formData && this.props.formData.length - 1;
+    if (this.props.errorSchema[lastIndex] && !errorSchemaIsValid(this.props.errorSchema[lastIndex])) {
+      this.showInViewMode = false;
+    }
+  }
```
cehsu: @cvalarida this change fixes the bug (view changing once user input is valid) you found.

test/common/schemaform/fields/ArrayField.unit.spec.jsx
```diff
@@ -239,4 +241,181 @@ describe('Schemaform <ArrayField>', () => {
...
+        registry={registry}
+        onChange={onChange}
+        onBlur={onBlur}
+        formData={formData}
+        formContext={formContext}
+        requiredSchema={requiredSchema}/>
+    );
+    expect(tree.everySubTree('SchemaField').length).to.equal(1);
+  });
+  it('should render edit mode until submit if initial data incomplete or invalid', () => {
```
cehsu: @cvalarida this test has been set up for the bug you found.

src/applications/disability-benefits/526EZ/config/form.js
```diff
@@ -196,7 +198,18 @@ const formConfig = {
           schema: {
             type: 'object',
             properties: {
-              servicePeriods
+              servicePeriods: {
+                type: 'array',
+                items: {
+                  type: 'object',
+                  properties: {
+                    dateRange: _.set('required', ['from'], dateRange),
```
cehsu: This requires that dateRange has a 'from' property to be valid.

cehsu:
Sure thing @cvalarida, I've set up a ticket based on your comment above and connected to this PR.

src/applications/disability-benefits/526EZ/config/form.js
cvalarida: Seems like this should be set in the schema, but this is probably okay. :man_shrugging:

src/applications/disability-benefits/526EZ/tests/schema/initialData.js
```diff
@@ -32,8 +32,8 @@ export default {
     {
       serviceBranch: 'Army',
       dateRange: {
-        from: '1990-02-02',
-        to: '2010-03-04'
+        // from: '1990-02-02',
```
cvalarida: Can we put these back in for ease of testing later?

test/common/schemaform/fields/ArrayField.unit.spec.jsx
```diff
@@ -126,13 +129,13 @@ describe('Schemaform <ArrayField>', () => {
         {},
         {}
       ];
-      errorSchema = {};
+      fullErrorSchema = {};
```
cvalarida: Is there a problem with using `errorSchema` here?

test/common/schemaform/fields/ArrayField.unit.spec.jsx
```diff
@@ -239,4 +241,181 @@ describe('Schemaform <ArrayField>', () => {
 
     expect(tree.subTree('button').props.disabled).to.be.true;
   });
+  it('should render review mode when specified and has data', () => {
```
cvalarida: Does calling it `review mode` here make sense? Seems like that could be confusing. What do you think of `view mode`?

test/common/schemaform/fields/ArrayField.unit.spec.jsx
```diff
@@ -239,4 +241,181 @@ describe('Schemaform <ArrayField>', () => {
...
+        uiSchema={uiSchema}
+        idSchema={idSchema}
+        registry={registry}
+        onChange={onChange}
+        onBlur={onBlur}
+        formData={formData}
+        formContext={formContext}
+        requiredSchema={requiredSchema}/>
+    );
+    expect(tree.everySubTree('SchemaField').length).to.equal(0);
```
cvalarida: Might be worth making `viewField` a dummy component and also checking that it renders.

test/common/schemaform/fields/ArrayField.unit.spec.jsx
```diff
@@ -239,4 +241,181 @@ describe('Schemaform <ArrayField>', () => {
...
+  });
+  it('should render edit mode when specified and has no data', () => {
```
cvalarida: Instead of `when specified and has no data`, would it be clearer to say `when showLastItemInViewMode is true and no data is given`?

test/common/schemaform/fields/ArrayField.unit.spec.jsx
```diff
@@ -239,4 +241,181 @@ describe('Schemaform <ArrayField>', () => {
...
+        registry={registry}
+        onChange={onChange}
+        onBlur={onBlur}
+        formData={formData}
+        formContext={formContext}
+        requiredSchema={requiredSchema}/>
+    );
+    expect(tree.everySubTree('SchemaField').length).to.equal(1);
+  });
+  it('should render edit mode when specified and has invalid data', () => {
```
cvalarida: Same as above but with `invalid data is given`.

test/common/schemaform/fields/ArrayField.unit.spec.jsx
cvalarida: What do you think of the test description text `should not exit edit mode prematurely when validation errors are resolved`? It took me a couple tries to figure out what this test should be doing.

test/common/schemaform/fields/ArrayField.unit.spec.jsx
```diff
@@ -239,4 +241,181 @@ describe('Schemaform <ArrayField>', () => {
...
+    ];
+    const uiSchema = {
+      'ui:title': 'List of things',
+      'ui:options': {
+        showLastItemInViewMode: true,
+        viewField: f => f
+      }
+    };
+    const fullErrorSchema = {
+      0: { thing: { __errors: ['some error'] } },
```
cvalarida: `thing` isn't a property of the array items. It will probably be more comprehensive if we make `field` required and then fill it out. It should start in the edit mode, and when we fill it out, it should stay in edit mode.

src/applications/common/schemaform/fields/ArrayField.jsx
markgreenburg: Does this cover the case where hitting 'edit' under any circumstance advances to the next page? Cause I ran into that. Basically -> use existing, valid initialData -> hit 'Edit' next to the first (and only) item -> advances to payments page.

markgreenburg:
Here's some feedback based solely on me playing around with this PR locally:
Here's some stuff that happened when I used a servicePeriod that I believe to be valid:
```{
      serviceBranch: 'Army',
      dateRange: {
        from: '1990-02-02',
        to: '2010-03-04'
      }
    }
```

1. The 'Edit' button (next to the service period) action seems inconsistent. I haven't been able to dive into why this might be happening yet.
  - Sometimes nothing happens
  - Sometimes I get advanced to the next page in the form flow (payments)
2. The 'Add another service period' button automatically adds a blank service period in 'view' mode (and clicking on the edit button yields the same results as point 1. above). This is probably meant to automatically open in 'edit' mode instead, I'd think

service history

src/applications/disability-benefits/526EZ/config/form.js
markgreenburg: Schema sets both `from` and `to` as required, which matches `submit` swagger so should be used in the config. Also, just as FYI, I'm working on some changes that restrict the `serviceBranch` to an `enum` list per EVSS. But that's also getting added to the schema, so shouldn't require anything else here as long as the schema's version of the definition is used.

src/applications/common/schemaform/fields/ArrayField.jsx
```diff
@@ -228,7 +242,7 @@ export default class ArrayField extends React.Component {
             const isEditing = this.state.editing[index];
             const notLastOrMultipleRows = !isLast || items.length > 1;
 
-            if (isReviewMode ? isEditing : isLast || isEditing) {
+            if (isReviewMode ? isEditing : !this.showInViewMode && (isLast || isEditing)) {
```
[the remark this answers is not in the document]
cehsu: Good catch, will update.

src/applications/disability-benefits/526EZ/config/form.js
cehsu: Thanks @markgreenburg will update using the schema.

src/applications/disability-benefits/526EZ/tests/schema/initialData.js
cehsu: Definitely.

test/common/schemaform/fields/ArrayField.unit.spec.jsx
cehsu: That's fine too, review mode is used in the field itself, also, this is sort of the array equivalent of the review card field now, so that was probably why I used "review" here.

test/common/schemaform/fields/ArrayField.unit.spec.jsx
cehsu: It is declared in the upper scope.

test/common/schemaform/fields/ArrayField.unit.spec.jsx
cehsu: Yeah that's a better description, will update.

test/common/schemaform/fields/ArrayField.unit.spec.jsx
cehsu: True, will update similarly.

test/common/schemaform/fields/ArrayField.unit.spec.jsx
cehsu: How about `should not render review mode until valid data submitted` to clarify and try to keep the message short.

src/applications/common/schemaform/fields/ArrayField.jsx
```diff
@@ -251,7 +265,7 @@ export default class ArrayField extends React.Component {
                           disabled={disabled}
                           readonly={readonly}/>
                       </div>
-                      {notLastOrMultipleRows &&
+                      {(notLastOrMultipleRows || this.showInViewMode) &&
```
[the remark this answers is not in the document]
cehsu: Good eye, "New" can probably be replaced by "Additional" as it is applied to items only if they are the last of a list with a length longer than 1.
The update button is typically not available for the last item, but we can use local state to determine whether we should provide that (if it is accessed via the edit button rather than add another).
Remove shouldn't be available if there's only one item left, so that needs to be `removed` as well.

cehsu:
There are two updates to the requirements here:

1. @goldenmeanie has clarified that the default behavior for the array field should now be to collapse any prefilled items, and therefore not require any configuration in the form config (previously achieved using `reviewMode` and was planned to be replaced by `showLastItemInViewMode`).
2. @markgreenburg has also explained that support for incomplete prefilled items is required.

Therefore:

1. valid items are now collapsed by default (this specifically affects the last item, which was previously displayed in edit mode with the add item styling by default)
image

2. any invalid items (including non-last items) are now displayed in edit mode
image

test/common/schemaform/fields/ArrayField.unit.spec.jsx
cehsu: Good point, test updated.

test/common/schemaform/fields/ArrayField.unit.spec.jsx
cehsu: Great idea, added.

src/applications/common/schemaform/fields/ArrayField.jsx
```diff
@@ -17,6 +19,21 @@ import { errorSchemaIsValid } from '../validation';
...
+    return true;
+  }
+  if (typeof collection !== 'object') {
+    return false;
+  }
+  return _.every(isEmpty, Object.values(collection));
+};
+
+const isInvalid = (item, index, errorSchema) => {
+  return isEmpty(item) || !errorSchemaIsValid(errorSchema[index]);
```
jbalboni: What scenario did you hit that requires the `isEmpty` check? It seems like it would only be necessary in cases where an array had a `minItems` requirement, but none of the fields in the array items were required, which I don't believe we do anywhere (because it'd lead to bad data).

src/applications/common/schemaform/fields/ArrayField.jsx
```diff
@@ -136,9 +152,9 @@ export default class ArrayField extends React.Component {
           ? false
           : val;
       });
-      const editingState = this.props.uiSchema['ui:options'].reviewMode;
+      const editingState = this.props.uiSchema['ui:options'].reviewMode ? true : 'adding';
       const newState = _.assign(this.state, {
-        editing: newEditing.concat(!!editingState)
+        editing: newEditing.concat(editingState)
```
jbalboni: I think we might benefit from tweaking this state name and possible values with these changes. `true`, `false`, and `'adding'` seems confusing. Maybe we should call it `itemMode` and use `viewing`, `editing`, and `adding`?

jbalboni (commented):
Left a couple comments. Please make sure to test these changes on our other forms, both on regular pages and review pages.

src/applications/common/schemaform/fields/ArrayField.jsx
cvalarida: Yeah, thinking through this, I can't think of a time where this will do what we want it to. If we have an array of empty items, it should be fine unless the objects have required fields. Alternatively, if we had an empty array, that should be fine unless there's a `minItems` (in which case, the `errorSchema` won't be valid).

This here would catch if `minItems > 0` and those items were all blank and those items had no required fields.

src/applications/disability-benefits/526EZ/config/form.js
```diff
@@ -131,7 +132,8 @@ const formConfig = {
     disabilities: omitRequired(disabiltiesDefinition),
     specialIssues,
     servicePeriods,
-    privateTreatmentCenterAddress
+    privateTreatmentCenterAddress,
+    dateRangeFromRequired
```
cvalarida: Where is this being used?

src/applications/common/schemaform/fields/ArrayField.jsx
cehsu: Thanks @jbalboni, that is much more expressive, updated.

src/applications/common/schemaform/fields/ArrayField.jsx
cehsu: Thanks @cvalarida and @jbalboni, good point, `isEmpty` removed.

src/applications/disability-benefits/526EZ/config/form.js
cehsu: This was actually changed in master.

test/common/schemaform/fields/ArrayField.unit.spec.jsx
```diff
@@ -144,55 +229,55 @@ describe('Schemaform <ArrayField>', () => {
...
+      expect(tree.everySubTree('SchemaField').length).to.equal(1);
     });
     it('add', () => {
-      expect(tree.everySubTree('SchemaField').length).to.equal(1);
+      expect(tree.everySubTree('SchemaField').length).to.equal(0);
 
       tree.getMountedInstance().handleAdd();
 
       expect(onChange.firstCall.args[0].length).to.equal(3);
-      expect(tree.getMountedInstance().state.editing[2]).to.be.false;
```
cvalarida: The new `expect` looks good, but this one confuses me, and I find that concerning. In this old test, this is my understanding:
1) We have two items in the array
2) The first item is rendered as not editing (`viewField`)
3) The second item is rendered as editing (because it's the last item in the array)
4) When we hit the add button, it puts the second item into a view state (not editing) and adds a third
5) This newly-added third item should be rendered in the edit state

So...why did we expect the edit state of the third item to be false? @jbalboni?

cvalarida (commented):
It looks good to me, but I'd like to get some clarification on some things.

Can you outline what the expected behavior (acceptance criteria) is when the following cases are true?
- No data is pre-filled
- A single item is pre-filled, no validation errors
- Multiple items are pre-filled, no validation errors
- A single item is pre-filled with validation error
- Multiple items are pre-filled with validation error on the last item
- Multiple items are pre-filled with validation error on not the last item

src/applications/common/schemaform/fields/ArrayField.jsx
```diff
@@ -27,14 +31,13 @@ export default class ArrayField extends React.Component {
...
 
     /*
-     * We’re keeping the editing state in local state because it’s easier to manage and
+     * We’re keeping the item modes state in local state because it’s easier to manage and
      * doesn’t need to persist from page to page
      */
 
     this.state = {
-      editing: props.formData ? props.formData.map(() => false) : [true]
+      itemModes: props.formData ? props.formData.map((item, index) => (isInvalid(item, index, props.errorSchema) ? 'editing' : 'viewing')) : ['adding']
```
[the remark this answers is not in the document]
cehsu: Good point @cvalarida, that is a great idea.

test/common/schemaform/fields/ArrayField.unit.spec.jsx
cehsu: One thing that might be confusing here is that this no longer called `editing` but `itemModes`, and the new item's `itemMode` is now being tested as being `adding`, in case that helps to clarify what the change is here.

test/common/schemaform/fields/ArrayField.unit.spec.jsx
```diff
@@ -77,27 +115,74 @@ describe('Schemaform <ArrayField>', () => {
...
         idSchema={idSchema}
         registry={registry}
+        formContext={formContext}
         formData={formData}
+        onChange={f => f}
+        requiredSchema={requiredSchema}/>
+    );
+    expect(tree.everySubTree('.va-growable-background').length).to.equal(2);
+  });
+  it('should render invalid prefilled items', () => {
```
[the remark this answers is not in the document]
cehsu: Could you please clarify why we would want to test whether the invalid prefilled item is the last or not?

test/common/schemaform/fields/ArrayField.unit.spec.jsx
```diff
@@ -53,8 +55,44 @@ describe('Schemaform <ArrayField>', () => {
...
+        errorSchema={errorSchema}
+        idSchema={idSchema}
+        registry={registry}
+        formContext={formContext}
+        onChange={f => f}
+        requiredSchema={requiredSchema}/>
+    );
+    expect(tree.everySubTree('SchemaField').length).to.equal(1);
+  });
+  it('should render prefilled items', () => {
```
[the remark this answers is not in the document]
cehsu: Good idea, that's much clearer.

test/common/schemaform/fields/ArrayField.unit.spec.jsx
```diff
@@ -53,8 +55,44 @@ describe('Schemaform <ArrayField>', () => {
...
+      }
+    };
+    const uiSchema = {
+      'ui:title': 'List of things',
+      'ui:options': {
+        viewField: f => f
+      }
+    };
+    const tree = SkinDeep.shallowRender(
+      <ArrayField
```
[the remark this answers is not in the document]
cehsu: Interesting, form data is not being provided here, since it is testing an empty form. Would providing the formData prop improve the readability of this test?

test/common/schemaform/fields/ArrayField.unit.spec.jsx
```diff
@@ -53,8 +55,44 @@ describe('Schemaform <ArrayField>', () => {
 
     expect(tree.subTree('TitleField').props.title).to.equal(uiSchema['ui:title']);
     expect(tree.everySubTree('SchemaField')).not.to.be.empty;
+    expect(tree.everySubTree('.va-growable-background')).to.be.empty;
   });
   it('should render items', () => {
```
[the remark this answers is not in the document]
cehsu: Yes more description would be helpful here, perhaps, should render an empty form?

test/common/schemaform/fields/ArrayField.unit.spec.jsx
```diff
@@ -77,27 +115,74 @@ describe('Schemaform <ArrayField>', () => {
...
+    };
+    const ViewField = f => f;
+    const uiSchema = {
+      'ui:title': 'List of things',
+      'ui:options': {
+        viewField: ViewField
+      }
+    };
+    const fullErrorSchema = {
+      0: { field: { __errors: ['some error'] } },
```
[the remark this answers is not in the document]
cehsu: Good point, I'll revert this to the original: `{ __errors: ['Test error'] };`

src/applications/common/schemaform/fields/ArrayField.jsx
```diff
@@ -17,6 +19,20 @@ import { errorSchemaIsValid } from '../validation';
...
+  }
+  if (typeof collection !== 'object') {
+    return false;
+  }
+  return _.every(isEmpty, Object.values(collection));
+};
+
+const notYetAnswered = (item, index, errorSchema) => {
+  return isEmpty(item) || !errorSchemaIsValid(errorSchema);
+};
```
cehsu: This helps to determine whether empty items with schemas that do not contain any required fields are displayed properly. This scenario is encountered if a user "adds another" saves the form, and returns later to continue filling the form out.
screen shot 2018-05-30 at 5 56 57 pm

src/applications/common/schemaform/fields/ArrayField.jsx
```diff
@@ -289,7 +304,7 @@ export default class ArrayField extends React.Component {
                 'usa-button-disabled': !this.props.formData
               }
             )}
-            disabled={!this.props.formData}
+            disabled={!this.props.formData || isEmpty(this.props.formData[lastIndex])}
```
cehsu: This disables the "add another" button when no information has been provided, and prevents the creation of empty items (particularly when no fields are required by the schema/no validation prevents this from occuring).

cehsu:
Thanks @jbalboni @cvalarida @markgreenburg for your comments which have been addressed. Thanks for taking another look when you can and kindly refer to the updated description for a summary of the changes. Thanks for any additional feedback as well.

src/applications/common/schemaform/fields/ArrayField.jsx
jbalboni: Ok, that's fine for now. Can we make a ticket to go through the forms and update these cases to have at least one required field? I think it's a bad practice to have array fields where there are no required fields in the items. It encourages bad data being sent over. Once we do that, I think we could simplify how this works.

cehsu: Thanks @jbalboni, issue created https://github.com/department-of-veterans-affairs/vets.gov-team/issues/10732.

src/applications/common/schemaform/fields/ArrayField.jsx
```diff
@@ -17,6 +19,20 @@ import { errorSchemaIsValid } from '../validation';
 const Element = Scroll.Element;
 const scroller = Scroll.scroller;
 
+const isEmpty = (collection) => {
```
jbalboni: Normally we use the regular function syntax for these functions defined outside of React components. Curious if you had a reason for defining them as arrow functions.

jbalboni (approved):

src/applications/common/schemaform/fields/ArrayField.jsx
cehsu: Thanks @jbalboni, changed.

src/applications/common/schemaform/fields/ArrayField.jsx
```diff
@@ -27,14 +43,13 @@ export default class ArrayField extends React.Component {
...
 
     /*
-     * We’re keeping the editing state in local state because it’s easier to manage and
+     * We’re keeping the item modes state in local state because it’s easier to manage and
      * doesn’t need to persist from page to page
      */
 
     this.state = {
-      editing: props.formData ? props.formData.map(() => false) : [true]
+      itemModes: props.formData && props.formData.length ? props.formData.map((item, index) => (notYetAnswered(item, index, props.errorSchema[index]) ? 'editing' : 'viewing')) : ['adding']
```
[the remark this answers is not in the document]
cehsu: Good point, @jbalboni, a helper function `getItemModes` has been set up and is now used here instead.

cehsu:
@annekainicUSDS please let us know when this can be merged in/the forms library code freeze ends.

markgreenburg (changes requested):
Been a while since I've taken a look at this :)
My main concern is that there is no 'remove' button for new service periods, like there is for pre-filled service periods. This is a problem because if a user hits 'add new service period', they will not be able to continue with the 526 until the enter in some fake 'required' information. Happy to re-review once you're able to figure that one out!

cehsu:
We could remove the condition in front of the remove button @markgreenburg, is that what you recommend? It doesn't seem like this would require design approval?

cehsu:
@jbalboni @goldenmeanie is there any reason we don't provide a remove button for new array items? If not, we can remove the conditional check and always provide a remove button, as recommended by @markgreenburg.

jbalboni:
I would need to understand the page more, but typically we've tried not to have pages with array fields that have a minItems of 0 without putting a yes/no question in to hide the field. If you're looking to change the default behavior for array fields I'd want to see your use case, because I think adding that button will cause problems on other forms.

cehsu:
Good point @jbalboni, if I understand you correctly, we could use a condition to check whether the minItems (typically 1) requirement has been reached, if not, don't add the button.

jbalboni:
Looking at it again, I think what you want to do is show the remove button when you're adding and `reviewMode` is true. I think that will keep things working as expected on the other forms.

src/applications/common/schemaform/fields/ArrayField.jsx
```diff
@@ -251,11 +270,11 @@ export default class ArrayField extends React.Component {
                           disabled={disabled}
                           readonly={readonly}/>
                       </div>
-                      {notLastOrMultipleRows &&
+                      {!isAddingFirst &&
```
cehsu: @markgreenburg actually we do already provide the remove buttons to added items (but we did not previously provide the update button to the last item -- which needed to be updated as we now allow last items to be in edit mode), I've fixed the bug here. Nice catch!

cehsu:
Closing this and will open a new PR on the new us-forms-system repo.

PR #7644, opened 2018-04-24
Move more styles over, update header components

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets-contrib/issues/128

This moves the rest of the header widgets to using jean-pants and organizes the styles accordingly.


U-DON (approved):
LGTM. Was wondering if it made sense for the HelpMenu and SearchMenu to move to the design system, though. I'm in the middle of reorganizing the `login` folder into `platform`, and since those menus are specific to that widget, I was thinking of moving them to `login/components`. `DropDown` definitely seemed like a good candidate to move to jean-pants, though.

rianfowler (approved):

jbalboni:
@U-DON I went back and forth on that a bunch of times. Just considering how we current vets.gov devs would use the design system, I don't think those two components should be in there. But longer term, we may have to make more of the header styles/component reusable, depending on how devs are building on top of our platform. So I sort of punted and put them in the design system for now, since they're fairly simple.

That said, I'm not tied to this and if you prefer them to stay in here, we can do that.

ncksllvn (approved):

PR #7648, opened 2018-04-24
Moved login widget to platform

U-DON:
Resolves department-of-veterans-affairs/vets-contrib#24.


ncksllvn:
Sorry if this wasn't ready for review! But getting some linter errors from the test directory -

```
/Users/ncksllvn/Repos/vets-website/src/js/claims-status/tests/e2e/00.claims-list.e2e.spec.js
  4:30  error  Unable to resolve path to module '../../../../../test/e2e/login-helpers'  import/no-unresolved
```

A bunch more of those looking for that login-helpers file.

jbalboni:
@U-DON There are still some files in `src/js/common` that look login related, can we move those, too?

Also, I had originally envisioned this code going in `platform/login` (or `platform/user-authentication`), but it does also fit in `platform/site-wide`. I'm trying to avoid a lot of nesting, but maybe `platform/site-wide` is less confusing?

U-DON:
The lint problems should be resolved now. I was hoping to resolve those before anyone got to take a look, but just got to them...

U-DON:
@jbalboni: Regarding the remaining files in `src/js/common` (aside from `schemaform`, which I presume you will be handling):

`login-helpers.js`: I would like the one remaining function in this to be moved to `user-profile` as an action. 

`RequiredLoginView.jsx`, `MHVApp.jsx`: Don't know where these will live... These are not actually for logging in the user, but more so going through a bunch of checks on their authorization. Perhaps they can belong somewhere like `src/platform/authorization`.

`AcceptTermsPrompt.jsx`, `RequiredTermsAcceptanceView.jsx`: These will be removed or pushed into a standalone T&C app in light of new T&C work being done.

`SignInLink.jsx`: I don't know if this is particularly necessarily, but it seems limited to usage within the SiP form code. From a cursory inspection, I'm inclined to believe there might be a better way to do what this is being used for.

U-DON:
@jbalboni: I put the `login` into `platform/site-wide` since I found it to be pretty similar to `feedback`. Both are initialized on every page as part of `startSiteWideComponents`.

src/js/login/login-entry.jsx
```diff
@@ -1,16 +0,0 @@
...
-
-import Main from './containers/Main';
-
-export default function createLoginWidget(store) {
-  startReactApp((
-    <Provider store={store}>
-      <Main/>
-    </Provider>
-  ), document.getElementById('login-root'));
-}
```
U-DON: Renamed the module to `index.js` as a top level module that exports the initializer for the app. `feedback-entry` and `login-entry` were kind of misleading in that these are usually the names of entry points for metalsmith pages, and they didn't correspond to any.

dhall (commented):
this LGTM but I'll defer to other for approval

U-DON:
Unsure why `setUserToken` in `login/tests/login-helpers.js` is having problems now. I think that's causing the E2E tests to break; the token is unable to be set successfully, so navigating to the app after a mocked login seems to fail.

jbalboni:
@U-DON That error is because the login-helpers file is getting caught up in our code coverage tooling. I think the e2e tests use the same Babel env as the unit tests. If I add `"src/**/tests/**",` to the nyc `exclude` array in package.json, the login test passes.

jbalboni:
@U-DON That sounds good for login-helpers and SignInLink. We could move the latter to the SiP code until we find a better approach for it.

I agree that RequiredLoginView and MHVApp are authorization, not login. Similarly, the T&C components seem like authorization, even if they're going to get moved later.

Putting the login widget in `site-wide` seems reasonable. My larger hope is for applications to have a clear set of tools for authentication, authorization, and the user profile, located in a discoverable place. Like: here are the actions you can use, here are some selectors to get the state info you need, and here are some common components you might need in your app.

Maybe it would make sense to have the widget `platform/site-wide/login` and also have a `platform/user` directory, where we organized some of the global functionality for authentication, authorization, and the profile? Interested in what you think about this. We don't need to do it in this PR, necessarily, I mainly want to get the login code somewhere in platform for now and we can iterate on it.

bshyong (approved):

U-DON:
@jbalboni: It makes sense to have a `platform/user` directory. In that case, I think it would make sense to pull the parts of the `user-profile` and `login` actions and reducer that are commonly used across apps.

I've also been thinking the login widget under `platform/site-wide/login` now is kind of misleadingly named, since it's really just specifically the user navigation header (search, help, sign-in/profile dropdown). The way it's named sounds like it has to do with the more general state of a logged in user.

jbalboni:
@U-DON Cool, I agree with both of those points. Let me know how much of that you want to do in here and what we should leave for another ticket.

U-DON:
@jbalboni: I think that should be done separately. There's already a lot of stuff moving in here, and it's in a place where we can iterate on it.

jbalboni (approved):

PR #7649, opened 2018-04-25
Personalization: Additional Address work

ncksllvn:
- [x] Add action creators for getting states/countries
- [x] Add the `toGenericAddress` and `expandGenericAddress` functions
  - These are only used when passed to the update-modal
- [x] Check for the `canUpdate` flag from the API
- [x] Implement `isAddressEmpty`
- [x] Fix problems with Address presentation
- [x] Fix mysterious error when saving to server

Resolves department-of-veterans-affairs/vets.gov-team/issues/10118
Resolves department-of-veterans-affairs/vets.gov-team/issues/10074


src/js/personalization/va-profile-beta/components/AddressSection.jsx
```diff
@@ -51,21 +62,45 @@ class EditAddressModal extends React.Component {
...
+
+  return (
+    <div>
+      {street}<br/>
+      {cityStateZip}<br/>
+      {country}
+    </div>
+  );
+}
+
```
ncksllvn: @bshyong there is no consolidation with the address formatting here, fyi. That's only done when going into the update-modal. It was too confusing for too little return to try to consolidate it everywhere.

bshyong (approved):
LGTM; could we reuse the state/military code json files so that we don't have duplication?

ncksllvn:
@bshyong sure! Do you think it would be better to change the reference in Letters to point to the JSON files, or to remove our JSON files and reference their constants instead? I for some reason really like them stored in plain JSON.

bshyong (approved):
LGTM!

PR #7650, opened 2018-04-25
Use Chrome locally for e2e testing

jbalboni:
We already use Chrome for e2e tests on Jenkins, this switches us to using them locally.

@rianfowler figured out that we needed to update our version of chromedriver for the axe check to work correctly, which was the last real hurdle.


config/nightwatch.js
```diff
@@ -38,7 +37,6 @@ module.exports = {
         acceptSslCerts: true,
         webStorageEnabled: true,
         chromeOptions: {
-          binary: electron,
           args: ['--window-size=1024,768']
```
ncksllvn: Chrome was the default?

ncksllvn (approved):
Runs great!

config/nightwatch.js
jbalboni: Yeah, to use Electron you use the normal Chrome set up and driver, just with the Electron binary. So all we have to do is remove the binary to switch to normal Chrome.

U-DON:
I ran this locally on the latest pull from master and got the same two errors as @cvalarida.

jbalboni:
That's bizarre. That pre-need test works for me if I make Chrome the active window, but if I leave it in the background or run in headless mode, it fails.

jbalboni:
Alright, pre-need test is fixed, the other test is failing in master, so I don't think it needs to be addressed in this PR.

U-DON (approved):
LGTM. Ran again and now only failing on the one test that's already failing on master.

PR #7653, opened 2018-04-25
Update react routes link to devops repo

wunderhund:
The nginx config link for react routes pointed to the old, pre-BRD config. Updated it to the current location of the react routes config.


cehsu (approved):
LGTM.

PR #7654, opened 2018-04-25
[WIP] GIBCT Yellow Ribbon Program 

nagaborra:
Added two extra drop downs for degree level and division or school when YR recipient is selected


jkassemi:
@adi-pen, same here - can you work with @maneshkommidi on developing a remote schema for this so you can stub out the state and associated actions and reducers? I recently requested a sample of what this change would require on the API side, and I think that would be very helpful in getting this connected.

jkassemi:
https://github.com/department-of-veterans-affairs/vets.gov-team/issues/11141

PR #7655, opened 2018-04-25
Personalization bugfixes

bshyong:
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/10152
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/10151

### DIFF
--- a/src/js/personalization/dashboard-beta/containers/DashboardApp.jsx
+++ b/src/js/personalization/dashboard-beta/containers/DashboardApp.jsx
@@ -13,6 +13,9 @@ import BetaApp, { features } from '../../beta-enrollment/containers/BetaApp';
 import RequiredLoginView from '../../../common/components/RequiredLoginView';
 import DowntimeNotification, { services } from '../../../../platform/monitoring/DowntimeNotification';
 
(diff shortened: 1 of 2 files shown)


src/js/personalization/dashboard-beta/containers/DashboardApp.jsx
```diff
@@ -13,6 +14,18 @@ import BetaApp, { features } from '../../beta-enrollment/containers/BetaApp';
...
+
+const scroller = Scroll.animateScroll;
+const scrollToTop = () => {
+  scroller.scrollTo(0, {
+    duration: 500,
+    delay: 0,
+    smooth: true,
+  });
+};
+
```
ncksllvn: Do multiple apps do this? Seems like something we could convert to a general util at some point.

ncksllvn (approved):
LGTM

src/js/personalization/dashboard-beta/containers/DashboardApp.jsx
bshyong: Agree... not all apps do this, but many do. This is a common UX problem for us because of how browsers handle content being loaded in the middle of the page

PR #7658, opened 2018-04-26
Move user related code into platform/user

jbalboni:
Connects to https://github.com/department-of-veterans-affairs/vets-contrib/issues/24

This moves the global user related code into `platform/user`. 

Some of this might seem weird at first glance, since now the login widget in site-wide and the user profile app have no actions or reducers. However, all of that logic is global and available to every application. It's also used in various locations, which means it doesn't really belong in application or widget code and is a platform wide concern.

I don't want to overload this PR, but the next step is to write some selectors and document this code, so it's more obvious how to use it.

Changes to structure are:

```
|-- src
    |-- forms
    |   |-- components
    |   |   |-- SignInLink.jsx
    |-- monitoring
    |-- polyfills
    |-- site-wide
    |   |-- user-nav
    |       |-- index.js
    |       |-- components
    |       |   |-- BetaDropdown.jsx
    |       |   |-- HelpMenu.jsx
    |       |   |-- SearchHelpSignIn.jsx
    |       |   |-- SearchMenu.jsx
    |       |   |-- SignInModal.jsx
    |       |   |-- SignInProfileMenu.jsx
    |       |-- containers
    |       |   |-- Main.jsx
    |       |-- sass
    |       |   |-- user-nav.scss
    |       |-- tests
    |           |-- user-nav.e2e.spec.js
    |           |-- components
    |           |-- containers
    |-- startup
    |-- static-data
    |-- testing
    |   |-- e2e
    |   |   |-- account-creation-helpers.js
    |-- user
    |   |-- authorization
    |   |   |-- components
    |   |   |   |-- AcceptTermsPrompt.jsx
    |   |   |   |-- RequiredLoginView.jsx
    |   |   |-- containers
    |   |       |-- MHVApp.jsx
    |   |       |-- RequiredTermsAcceptanceView.jsx
    |   |-- login
    |   |   |-- utilities.js
    |   |   |-- actions
    |   |   |-- reducers
    |   |-- profile
    |   |   |-- actions
    |   |   |   |-- index.js
    |   |   |   |-- mhv.js
    |   |   |-- reducers
    |   |-- tests
    |       |-- unauthed-flows.e2e.spec.js
    |       |-- authorization
    |       |   |-- components
    |       |   |-- containers
    |       |-- profile
    |           |-- reducers
    |-- utilities
```


src/platform/user/profile/actions/index.js
```diff
@@ -30,6 +31,62 @@ export function profileLoadingFinished() {
...
+  }).then(json => {
+    const userData = json.data.attributes.profile;
+    // sessionStorage coerces everything into String. this if-statement
+    // is to prevent the firstname being set to the string 'Null'
+    if (userData.first_name) {
+      sessionStorage.setItem('userFirstName', userData.first_name);
+    }
+    // Report out the current level of assurance for the user
+    recordEvent({ event: `login-loa-current-${userData.loa.current}` });
+    dispatch(updateProfileFields({
```
rianfowler: I think this action creator only uses one argument.

rianfowler (approved):

src/platform/user/profile/actions/index.js
jbalboni: Not sure what you mean. I think we're only passing one argument to `updateProfileFields`

ncksllvn (approved):
LGTM

U-DON (commented):
LGTM. But I would've kept the login actions and reducers within user-nav, because most of those are specifically tied to the widget (toggling the modal, toggling the menu states, etc.). Specifically, the only thing that is more general to the site is `currentlyLoggedIn`. I would also change `platform/user/login` to `platform/user/authentication`, since the utilities deal with more than just login.

jbalboni:
@U-DON Toggling the modal is only way we have of letting someone log in in place, without changing pages, which is required for the forms. It seems like that's broader than just the login widget.

But maybe we should put the menu state in a different part of the Redux store and keep that in site-wide/site-nav?

jbalboni:
I don't think the query param method works for the forms since we don't want to redirect after signing in. They need to stay on the same page and just change the state.

I'm a little wary of creating circular dependencies between the top level `platform` directories, which I think is what we'd get by leaving the login action in `site-wide/user-nav` (since it also depends on code in `platform/user/profile`). Down the road that may cause us problems. So I really want to avoid `platform/user` depending on `platform/site-wide`, since the reverse has to be true.

I don't think we should think of particular components as owning parts of the Redux state, especially state that is shared across applications. The Redux state is the core of our application and we might have one, or several, interfaces for interacting with that core. 

But related code should live together as much as possible, so I see your point that it's a little awkward to separate the sign in components from the actions and reducer. There's also a bunch more authentication related code in the `user-nav` directory than I realized at first, which does feel strange separated from the Redux related code.

As you pointed out, the only thing general to all our apps is the `currentlyLoggedIn`. So maybe creating a separate Redux state slice for the menu/modal state is a good path, and we can try to refactor some of the `platform/site-wide/user-nav` code to pull just the login process related code into `platform/user/authentication`, and leave the navigation related code where it is. The `SignInLink` component could become a container that controls it's own `SignInModal` and does the login process outside of the navigation (but reusing the same components/actions).

I'll move the login actions/reducers back for now, since I think that's better than what I'm doing here. And I'll think some more about the way to do the separation mentioned above in a separate PR.

PR #7659, opened 2018-04-26
ReviewCardField

cvalarida:
### View mode
image

### Edit mode
image

### Validation messages
image


src/js/disability-benefits/526EZ/config/form.js
```diff
@@ -120,110 +128,63 @@ const formConfig = {
...
-    reviewVeteranDetails: {
-      title: (isReviewPage) => {
-        const baseString = 'Review Veteran Details';
-        if (isReviewPage) {
-          return baseString.replace('Review ', '');
-        }
-        return baseString;
-      },
+    veteranDetails: {
+      title: (isReviewPage) => `${isReviewPage ? 'Review ' : ''}Veteran Details`,
```
cvalarida: I'm not sure that this should say any more than `Veteran Details`.

src/js/disability-benefits/526EZ/config/form.js
```diff
@@ -120,110 +128,63 @@ const formConfig = {
...
       pages: {
-        veteranInformation: createVerifiedVeteranInfoPage(fullSchema526EZ),
-        specialCircumstances: { // TODO: create page file and reuse 
-          title: 'Special Circumstances',
-          path: 'review-veteran-details/special-circumstances',
-          depends: formData => formData.prefilled,
+        veteranInformation: {
+          title: 'Veteran Details', // TODO: Figure out if this is even necessary
```
cvalarida: Is this, in fact, necessary?

src/js/disability-benefits/526EZ/config/form.js
```diff
@@ -120,110 +128,63 @@ const formConfig = {
...
-        if (isReviewPage) {
-          return baseString.replace('Review ', '');
-        }
-        return baseString;
-      },
+    veteranDetails: {
+      title: (isReviewPage) => `${isReviewPage ? 'Review ' : ''}Veteran Details`,
       pages: {
-        veteranInformation: createVerifiedVeteranInfoPage(fullSchema526EZ),
-        specialCircumstances: { // TODO: create page file and reuse
```
cvalarida: This is a particularly ugly diff...Basically, there were "duplicate" pages in these two chapters, where one chapter had all the pages for verified and one had all the pages for unverified. I've only replaced the `veteranInformation` page for now, but have removed the other duplicate pages.

test/disability-benefits/526EZ/components/ReviewCardField.unit.spec.jsx
```diff
@@ -0,0 +1,136 @@
...
+    expect(wrapper.find('viewComponent').length).to.equal(0);
+    expect(wrapper.find('.input-section').length).to.equal(1);
+
+    // Try to go back to viewing
+    wrapper.find('.update-button').simulate('click');
+    expect(wrapper.find('.input-section').length).to.equal(1);
+    expect(wrapper.find('viewComponent').length).to.equal(0);
+
+    // Also check that the validation error is rendered while we're at it
+    expect(wrapper.text()).to.contain('Arbitrary error string here');
```
cvalarida: We could also check for `.usa-input-error-message`... :thinking:

Do you guys care which way we test for errors? I don't think I have an opinion.

test/disability-benefits/526EZ/config/veteranInformation.unit.spec.jsx
```diff
@@ -38,7 +38,7 @@ describe('526EZ veteran information', () => {
 
     form.find('form').simulate('submit');
 
-    expect(form.find('.usa-input-error-message').length).to.equal(6);
+    expect(form.find('.usa-input-error-message').length).to.equal(5);
```
cvalarida: I'm actually not sure how this was working before. It currently throws validation messages on the following fields: First name, last name, VA file number, gender, date of birth. It doesn't throw errors on: Middle name, SSN.

src/js/disability-benefits/526EZ/config/form.js
cehsu: Perhaps a question for @goldenmeanie as that is how the prototype is set up?

src/js/disability-benefits/526EZ/config/form.js
cehsu: Is the title "Veteran Information" here?

src/js/disability-benefits/526EZ/config/form.js
```diff
@@ -120,110 +128,63 @@ const formConfig = {
...
-      }
-    },
-    veteranDetails: {
-      title: 'Veteran Details',
-      pages: {
-        veteranInformation: createUnverifiedVeteranInfoPage(fullSchema526EZ),
         specialCircumstances: {
           title: 'Special Circumstances',
           path: 'special-circumstances',
+          depends: formData => formData.prefilled,
```
cehsu: Will these pages always be displayed now?

src/js/disability-benefits/526EZ/config/form.js
cvalarida: Oh yeah. That probably makes sense. :stuck_out_tongue_winking_eye:

src/js/disability-benefits/526EZ/config/form.js
cvalarida: Yeah, exactly. We'll have just one set of pages that are always shown.

cehsu: Ok, great, should we remove the depends?

markgreenburg:
@cvalarida I noticed that hitting 'update' isn't actually required to save data updates. Is this desired?

jbalboni:
It probably shouldn't be an "Update" button, but something more like "Done" or "Close". Only updating data when you click that button is not feasible given how our form code works. You could implement a Cancel button without much trouble, though.

cvalarida:
Yeah. Originally, there was a cancel button, but that was taken out of the design, so I figured since there's no way to cancel, it doesn't really make sense to _not_ update as the user types.

src/js/disability-benefits/526EZ/components/ReviewCardField.jsx
```diff
@@ -0,0 +1,189 @@
...
+  constructor(props) {
+    super(props);
+
+    // Throw an error if there’s no viewComponent (should be React component)
+    if (typeof get('ui:options.viewComponent', this.props.uiSchema) !== 'function') {
+      throw new Error(`No viewComponent found in uiSchema for ReviewCardField ${this.props.idSchema.$id}.`);
+    }
+
+    if (!['object', 'array'].includes(this.props.schema.type)) {
+      throw new Error(`Unknown schema type in ReviewCardField: ${this.props.schema.type}`);
```
markgreenburg: Might suggest hinting at the allowed schema types instead:
`Expected schema type of 'array' || 'object' but got ${this.props.schema.type} instead`

src/js/disability-benefits/526EZ/components/ReviewCardField.jsx
```diff
@@ -0,0 +1,189 @@
...
+    }
+
+    this.state = {
+      // Set initial state based on whether all the data is valid
+      editing: !errorSchemaIsValid(props.errorSchema)
+    };
+  }
+
+  onPropertyChange(name) {
```
markgreenburg: Looks like this is the same change handler as `ObjectField` - de-duplicating these might be a bad abstraction, but thought I'd at least point out that this handler already lives somewhere else. Thoughts?

src/js/disability-benefits/526EZ/components/ReviewCardField.jsx
```diff
@@ -0,0 +1,189 @@
...
+            );
+          })}
+          <button className="usa-button-primary update-button" onClick={this.update}>Update</button>
+        </div>
+      </div>
+    );
+  }
+
+
+  getTitle = () => {
```
markgreenburg: I hate myself for commenting on this, but would it make sense to move `getTitle` above the `const title = this.getTitle();` on L89?

src/js/disability-benefits/526EZ/components/ReviewCardField.jsx
```diff
@@ -0,0 +1,189 @@
...
+
+  render() {
+    if (this.state.editing) {
+      return this.getEditView();
+    }
+
+    return this.getReviewView();
+  }
+}
+
```
markgreenburg: This sucker could probably use some `propTypes` given all the props used, if you're up for it 😬

src/js/disability-benefits/526EZ/config/form.js
```diff
@@ -122,50 +130,56 @@ const formConfig = {
...
+            'ui:options': {
+              viewComponent: veteranInformationViewField
             },
-            'view:homeless': {
-              'ui:title': 'Homeless or at risk of becoming homeless?'
-            },
-            'view:extremeFinancialHardship': {
-              'ui:title': 'Suffering from extreme financial hardship?'
+            fullName: fullNameUI,
+            socialSecurityNumber: merge({}, ssnUI, {
```
markgreenburg: Uhhh, just now realizing that there's no place for us to submit this data in the Swagger. Do we actually need ssn? Also applies to a bunch of other veteran info.

src/js/disability-benefits/526EZ/components/ReviewCardField.jsx
```diff
@@ -0,0 +1,189 @@
...
+      throw new Error(`No viewComponent found in uiSchema for ReviewCardField ${this.props.idSchema.$id}.`);
+    }
+
+    if (!['object', 'array'].includes(this.props.schema.type)) {
+      throw new Error(`Unknown schema type in ReviewCardField: ${this.props.schema.type}`);
+    }
+
+    this.state = {
+      // Set initial state based on whether all the data is valid
+      editing: !errorSchemaIsValid(props.errorSchema)
```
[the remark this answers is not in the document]
cvalarida: I actually don't know what the ramifications of that are, if any. I figure `props` is probably safe and `this.props` is...possibly less safe if React does something unexpected?

I'm not opinionated on this. :man_shrugging:

src/js/disability-benefits/526EZ/components/ReviewCardField.jsx
cvalarida: Yeah, I just copied this straight over from `ObjectField`. Good catch. I don't think it makes sense to abstract this out. We'd then put ourselves in potentially a weird spot where two unrelated fields share the same helper functions. If we then have to update it for one...

I dunno. It seems small enough I wasn't going to worry about it, to be honest.

src/js/disability-benefits/526EZ/config/form.js
```diff
@@ -122,50 +130,56 @@ const formConfig = {
...
+      title: (isReviewPage) => `${isReviewPage ? 'Review ' : ''}Veteran Details`,
       pages: {
-        veteranInformation: createVerifiedVeteranInfoPage(fullSchema526EZ),
-        specialCircumstances: { // TODO: create page file and reuse 
-          title: 'Special Circumstances',
-          path: 'review-veteran-details/special-circumstances',
-          depends: formData => formData.prefilled,
+        veteranInformation: {
+          title: 'Veteran Details', // TODO: Figure out if this is even necessary
+          path: 'veteran-information',
```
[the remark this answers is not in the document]
cvalarida: Yeah, it would be good to take a pass at all the paths when the form is in less flux. Can you write a ticket for that?

src/js/disability-benefits/526EZ/components/ReviewCardField.jsx
cvalarida: Don't make no nevermind to me. Eslint is opinionated on how these functions get ordered sometimes, but I don't know that it is on this one.

src/js/disability-benefits/526EZ/components/ReviewCardField.jsx
cvalarida: Eh...yeah...I started with them, but removed 'em because they were incomplete. I'll take a stab at putting all the props we're using in.

...I probably won't take a stab at outlining their structure. :grimacing:

markgreenburg (approved):
LGTM. We may want to change the 'update' button text, and I had a few other random questions, but don't think there's anything blocking.

src/js/disability-benefits/526EZ/components/ReviewCardField.jsx
```diff
@@ -0,0 +1,189 @@
...
+    disabled: false,
+    readonly: false,
+  }
+
+
+  constructor(props) {
+    super(props);
+
+    // Throw an error if there’s no viewComponent (should be React component)
+    if (typeof get('ui:options.viewComponent', this.props.uiSchema) !== 'function') {
```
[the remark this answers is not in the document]
jbalboni: Honestly, we should probably change the other one to `viewComponent`.

src/js/disability-benefits/526EZ/components/ReviewCardField.jsx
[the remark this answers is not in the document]
jbalboni: I think it's slightly less error-prone to use `props` instead of `this.props` in a constructor. If you do `super()` instead of `super(props)` then `this.props` won't work. Just using `props` works either way.

src/js/disability-benefits/526EZ/components/ReviewCardField.jsx
```diff
@@ -0,0 +1,212 @@
...
+      throw new Error(`Unknown schema type in ReviewCardField: ${schema.type}`);
+    }
+
+    const title = this.getTitle();
+
+    return (
+      <div className="review-card">
+        <div className="review-card--body input-section va-growable-background">
+          <h4>{title}</h4>
+          {properties.map(propName => {
```
jbalboni: I don't think you need this mapping. I think you can call SchemaField with the props passed to this custom field. That would also let you get rid of `onPropertyChange`. By choosing to call SchemaField for each property in your object, you're forcing yourself into duplicating the `ObjectField` component.

src/js/disability-benefits/526EZ/config/form.js
```diff
@@ -122,58 +130,63 @@ const formConfig = {
...
-              'ui:title': 'Homeless or at risk of becoming homeless?'
+            'ui:field': ReviewCardField,
+            'ui:title': 'Veteran information',
+            'ui:options': {
+              viewComponent: veteranInformationViewField
             },
-            'view:extremeFinancialHardship': {
-              'ui:title': 'Suffering from extreme financial hardship?'
+            fullName: fullNameUI,
+            socialSecurityNumber: merge({}, ssnUI, {
```
jbalboni: You don't need this empty object.

src/js/disability-benefits/526EZ/config/form.js
```diff
@@ -122,58 +130,63 @@ const formConfig = {
...
-                type: 'boolean'
+              vaFileNumber: {
+                type: 'string',
+                pattern: '^[cC]{0,1}\\d{7,9}$'
               },
-              'view:extremeFinancialHardship': {
-                type: 'boolean'
+              gender: {
+                type: 'string',
+                'enum': ['F', 'M']
```
jbalboni: Are you planning on pulling these fields from the schema later?

src/js/disability-benefits/526EZ/sass/disability-benefits.scss
```diff
@@ -100,3 +100,40 @@
   list-style: disc;
 }
 
+
+// ReviewCardField styles
+.review-card {
+  .review-card--header {
```
jbalboni: It seems like this nesting is unnecessary.

src/js/disability-benefits/526EZ/sass/disability-benefits.scss
```diff
@@ -100,3 +100,40 @@
...
+    background-color: $color-gray-lightest;
+    padding: 0 1rem;
+    display: flex;
+    justify-content: space-between;
+    align-items: center;
+
+    h4 {
```
jbalboni: Making classes for this and the button would be better, I think.

src/js/disability-benefits/526EZ/config/form.js
cvalarida: Yeah, definitely. For now, I copied over what we were already using in `createVerifiedVeteranInfoPage()`.

I don't remember offhand if much of this information is even _in_ the schema, now that I think about it... :thinking: That's a larger conversation, though. :stuck_out_tongue:

PR #7664, opened 2018-04-26
Fixed broken terms and conditions link in profile.

U-DON:
Quick fix while T&C flow is being reworked.


jbalboni (approved):
I get a 404 trying to pull the terms, but otherwise this looks fine.

U-DON:
Yeah, if you try to run this locally, for some reason, the migration to add the T&C doesn't automatically work, so you would have to do a `bundle exec rake db:migrate redo` on it.

PR #7665, opened 2018-04-27
Get past rated disabilities selection

cvalarida:
We were having trouble getting past the rated disabilities selection page because of `initialData` that no longer meets the schema requirements. This basically just blasts the requirements as far as the front end is concerned.

The trouble is this:
1) We can't move to the next page until all the information on the current page is valid
2) We may get "invalid" data from the `ratedDisabilities` endpoint
3) We want to be able to continue after selecting only valid disabilities

The proposal I outlined in https://github.com/department-of-veterans-affairs/vets.gov-team/issues/10123#issuecomment-384729257 lets us continue while still only allowing us to select conditions that we know will pass validation.

This is the first part of that.

---
Related to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/10125

### DIFF
--- a/src/js/common/schemaform/helpers.js
+++ b/src/js/common/schemaform/helpers.js
@@ -540,3 +540,22 @@ export function getActiveChapters(formConfig, formData) {
 
   return _.uniq(expandedPageList.map(p => p.chapterKey).filter(key => !!key && key !== 'review'));
 }
+
+/**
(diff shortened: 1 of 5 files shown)


jbalboni (approved):

cehsu (approved):
LGTM.

PR #7666, opened 2018-04-27
686 form spouse marriage information

jbalboni:
localhost_3001_disability-benefits_686_dependent-status_spouse-info
localhost_3001_disability-benefits_686_dependent-status_spouse-info 1


src/js/disability-benefits/686/config/form.js
```diff
@@ -1,40 +1,258 @@
...
+  vaFileNumber
+} = fullSchema.definitions;
 
-// const { } = fullSchema.definitions;
+function isMarried(form = {}) {
+  return ['Married', 'Separated'].includes(form.maritalStatus);
+}
+
+function createSpouseLabelSelector(nameTemplate) {
+  return createSelector(form => {
```
rianfowler: I'm not 100% sure how reselect works but I think redefining the selector (first argument passed to `createSelector`) creates a new cache for each call to `createSpouseLabelSelector`. 

I think this could make this first argument its own selector (e.g. `spouseFullNameSelector`) by making its `resultFunc` an identity function. This would memoize the `spouseFullNameSelector` once (as long as the form is immutable) for the subsequent calls. 

Then `createSpouseLabelSelector` could pass `spouseFullNameSelector` as the first argument to its `createSelector` and the new `nameTemplate` callback for each of the different `updateSchema` `resultFunc`s.

rianfowler (approved):

lbai001 (approved):

src/js/disability-benefits/686/config/form.js
jbalboni: Oh, that's interesting. As far as I know, you're right about that being how reselect works, I was just focused on using reselect to keep data changes from forcing a new schema object because of the label each time. Reusing the caches across all these selectors seems like a good improvement.

peggygannon (approved):
@jbalboni Thanks, looks good to me

PR #7667, opened 2018-04-27
Personalization: Add a check for empty array in Service History 

ncksllvn:


bshyong (approved):
Nit: can do serviceHistory && serviceHistory.length > 0

LGTM

ncksllvn:
@bshyong good idea! That file needs some cleaning up soon

PR #7670, opened 2018-04-27
Personalization: Fix for edit link showing on fields that already have an add button

ncksllvn:
We weren't able to test this with Betamocks because empty fields would 500, but in Staging I saw that blank fields rendered both the "edit" link and the "please add your address information." The edit link doesn't make sense there.


bshyong (approved):
LGTM

PR #7672, opened 2018-04-27
526 disability increase use successful ITF to gate form

cehsu:
These changes set up scaffolding to route users to the loading page and then to either the error page or to the current form page from the intro,  form saved and user profile pages.

Some changes to the UI include:

- The loading screen and error screens are now handled as a separate page to be more consistent with the overall form flow:
- A prestart success message is now displayed above the form, once the user arrives, and disappears once they navigate to another page.

Success:

image

Loading:
image

Error:
image

Additional work related to these changes include:

- [ ] [Fix PrestartAlert spacing above non-arrayfields](https://github.com/department-of-veterans-affairs/vets-website/issues/7822)
- [ ] [Display ITF alert on veteran information page](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/10891)
- [ ] [Display ITF message on review page](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/10892)


jbalboni:
@cehsu I think I would suggest doing this slightly differently, after seeing the number of integration points (which seem necessary, after given it some more thought). I think it may be better to integrate the `beforeStartForm` call into `fetchInProgressForm`. We could change the name of the latter to something more generic like `startForm` potentially (we could also leave that for later).

I was hoping to avoid making this a feature of our form library as a whole and just make it an option on FormStartControls, but it seems like that's not enough to support the functionality you need.

cvalarida:
Should this be connected to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/10113?

cehsu:
Sure thing @cvalarida, I've coped the PR description to the now linked issue, and renamed it so it's clearer. Please let me know if more information would be helpful.

cvalarida (commented):
I'm not really sure what the problem is that this is trying to solve, so I'm sure how to review this one, I'm afraid.

src/applications/common/schemaform/save-in-progress/actions.js
```diff
@@ -196,23 +204,32 @@ export function saveAndRedirectToReturnUrl(...args) {
...
   // TODO: Migrations currently aren’t sent; they’re taken from `form` in the
   //  redux store, but form.migrations doesn’t exist (nor should it, really)
-  return (dispatch, getState) => {
+  return async (dispatch, getState) => {
+    if (prestartForm) {
+      try {
+        await prestartForm();
+      // If prestart error, fail safely
+      } catch (err) {
+        return;
```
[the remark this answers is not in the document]
cehsu: Thanks @cvalarida, whether to use `try/catch` or `if` here is an important question, feedback on this gating logic is especially welcome. The user will have an alert displayed if the ITF fails (`ITFErrorAlert`) and the user will be directed to the form error page, and an error will also be logged as soon as the ITF fetch fails in `fetchITF`.

src/applications/common/schemaform/save-in-progress/FormStartControls.jsx
```diff
@@ -18,22 +18,29 @@ class FormStartControls extends React.Component {
...
   }
 
   handleLoadPrefill = () => {
+    const { beforeStartForm } = this.props.form;
     if (this.props.prefillAvailable) {
-      this.props.fetchInProgressForm(this.props.formId, this.props.migrations, true, this.props.prefillTransformer);
-    } else {
-      this.goToBeginning();
+      const prestartForm = beforeStartForm ? this.prestartForm : null;
+      return this.props.fetchInProgressForm(this.props.formId, this.props.migrations, true, this.props.prefillTransformer, prestartForm);
```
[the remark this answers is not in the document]
rianfowler: What is providing this `fetchInProgressForm` prop? 

If all of the arguments provided to the `fetchInProgressForm` are available in the context where `fetchInProgressForm` given to this component, then I recommend giving this component simple callback instead.

e.g. 

instead of this: 
```
<Component
dataProp={dataProp}
functionProp={functionProp}/>
// and calling this inside of Component like 
this.props.functionProp(dataProp)
```

invert the control: 

```
<Component
callback={() => functionProp(dataProp)}/>
```

This simplifies the component interface and defines the behavior in the context where the data is retrieved. It's not _always_ better to do this- it depends on the context and the story you're trying to tell in it. It's important to group like complexity and behavior.

src/applications/common/schemaform/save-in-progress/FormStartControls.jsx
```diff
@@ -18,22 +18,29 @@ class FormStartControls extends React.Component {
...
     }
+    return this.goToBeginning();
   }
 
   handleLoadForm = () => {
+    const { beforeStartForm } = this.props.form;
     // If successful, this will set form.loadedData.metadata.returnUrl and will
     //  trickle down to this.props to be caught in componentWillReceiveProps
-    return this.props.fetchInProgressForm(this.props.formId, this.props.migrations);
+    const prestartForm = beforeStartForm ? this.prestartForm : null;
```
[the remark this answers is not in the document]
cehsu: Indeed, how about:
`this.props.prestartForm` -> `this.props.checkGate`
`this.props.beforeStartForm` -> `this.props.hasGateCheck`
Recommendations welcome also.

src/applications/common/schemaform/save-in-progress/actions.js
```diff
@@ -196,23 +204,32 @@ export function saveAndRedirectToReturnUrl(...args) {
...
+      }
+    }
     const trackingPrefix = getState().form.trackingPrefix;
     const userToken = sessionStorage.userToken;
     // If we don’t have a userToken, fail safely
     if (!userToken) {
       dispatch(setFetchFormStatus(LOAD_STATUSES.noAuth));
-      return Promise.resolve();
+      Promise.resolve();
```
rianfowler: Probably need to return the resolved promise or return nothing if you don't want this fetch function to be thennable. `Promise.resolve()` creates a value but doesn't implicitly `return`

src/applications/common/schemaform/save-in-progress/actions.js
```diff
@@ -196,23 +204,32 @@ export function saveAndRedirectToReturnUrl(...args) {
...
     }
 
     // Update UI while we’re waiting for the API
     dispatch(setFetchFormPending(prefill));
```
rianfowler: I believe `dispatch` is synchronous. I don't think this will cause any issues but it's probably better to finish the work in this function first (like the `fetch`) and `dispatch` at the end.

src/applications/common/schemaform/save-in-progress/reducers.js
```diff
@@ -97,6 +98,22 @@ export const saveInProgressReducers = {
...
     });
   },
+  SET_PRESTART_STATUS: (state, action) => {
+    const newState = _.set('prestartStatus', action.status, state);
+
+    if (action.status) {
+      newState.prestartExpirationDate = action.expirationDate;
+    }
+
+    if (action.errorMessage) {
```
rianfowler: Just personal preference, but I thing having a separate error action is more clear. In the redux dev tools the story is easier to read from the action types e.g. `PENDING` => `SUCCESS` or `PENDING` => `FAILURE`.

src/applications/common/schemaform/save-in-progress/actions.js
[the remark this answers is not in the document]
rianfowler: Does `prestartForm` throwing an error put the application in an invalid state? I'm not sure what it does.

src/applications/common/schemaform/save-in-progress/FormStartControls.jsx
```diff
@@ -18,22 +18,29 @@ class FormStartControls extends React.Component {
     }
   }
 
+  goToIntroduction = () => {
```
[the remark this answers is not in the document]
markgreenburg: ^

src/applications/common/schemaform/save-in-progress/SaveInProgressErrorPage.jsx
```diff
@@ -6,6 +6,7 @@ import { withRouter } from 'react-router';
 import {
   LOAD_STATUSES,
   PREFILL_STATUSES,
+  PRESTART_STATUSES,
```
cvalarida: It seems...not right for this to be in the `SaveInProgressErrorPage`. Like, we've got a thing that we want to do before we start the form. It has nothing to do with SiP. But oh noes! The thing went wrong and we need to show an error page. The SiP error page is convenient, but doesn't seem like it should have its purpose expanded to include errors that happen outside of SiP.

src/applications/common/schemaform/save-in-progress/actions.js
```diff
@@ -15,6 +15,14 @@ export const SET_IN_PROGRESS_FORM = 'SET_IN_PROGRESS_FORM';
 export const SET_START_OVER = 'SET_START_OVER';
 export const SET_PREFILL_UNFILLED = 'SET_PREFILL_UNFILLED';
 
+
+export const PRESTART_STATUSES = Object.freeze({
```
cvalarida: This also isn't really a SiP thing.

src/applications/common/schemaform/save-in-progress/actions.js
```diff
@@ -196,23 +204,32 @@ export function saveAndRedirectToReturnUrl(...args) {
...
  *                                is different from the current version.
  */
-export function fetchInProgressForm(formId, migrations, prefill = false, prefillTransformer = null) {
+export function fetchInProgressForm(formId, migrations, prefill = false, prefillTransformer = null, prestartForm = null) {
+
   // TODO: Migrations currently aren’t sent; they’re taken from `form` in the
   //  redux store, but form.migrations doesn’t exist (nor should it, really)
-  return (dispatch, getState) => {
+  return async (dispatch, getState) => {
+    if (prestartForm) {
```
cvalarida: Same here. Seems like we should `await prestartForm` before calling `fetchInProgressForm`.

src/applications/common/schemaform/save-in-progress/actions.js
```diff
@@ -196,23 +204,32 @@ export function saveAndRedirectToReturnUrl(...args) {
...
-      return Promise.resolve();
+      Promise.resolve();
     }
 
     // Update UI while we’re waiting for the API
     dispatch(setFetchFormPending(prefill));
 
     // Query the api and return a promise (for navigation / error handling afterward)
-    return fetch(`${environment.API_URL}/v0/in_progress_forms/${formId}`, {
+    fetch(`${environment.API_URL}/v0/in_progress_forms/${formId}`, {
```
cvalarida: Why remove `return` here?

src/applications/common/schemaform/save-in-progress/reducers.js
```diff
@@ -97,6 +98,22 @@ export const saveInProgressReducers = {
       loadedStatus: LOAD_STATUSES.pending
     });
   },
+  SET_PRESTART_STATUS: (state, action) => {
```
cvalarida: Do I sound like a broken record yet? Sorry... :disappointed:

src/applications/common/schemaform/save-in-progress/selectors.js
```diff
@@ -3,6 +3,7 @@ import { getFormData } from '../state/selectors';
 import { PREFILL_STATUSES } from './actions';
 
 export const getSaveInProgressState = createStructuredSelector({
+  prestartStatus: state => state.form.prestartStatus,
```
cvalarida: I'll point this one out too.

src/applications/user-profile/helpers.jsx
```diff
@@ -1,6 +1,7 @@
 import Raven from 'raven-js';
 
 export const formBenefits = {
+  '21-526EZ': 'disability compensation and related compensation benefits',
```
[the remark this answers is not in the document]
cvalarida: Yeah, we should probably scope these changes to a different PR. (This probably ties into the "should this be part of SiP?" theme of this review.)

test/common/schemaform/save-in-progress/actions.unit.spec.js
```diff
@@ -312,7 +339,7 @@ describe('Schemaform save / load actions:', () => {
         status: 404
       }));
 
-      return thunk(dispatch, getState).then(() => {
+      return thunk(dispatch, getState).catch(() => {
```
cvalarida: This seems to be kind of a big change for it being only one word. What are the ramifications? (I didn't quite catch what made this change necessary.)

cvalarida (commented):
I'd like to reconsider whether this function call before the form starts should be tied to SiP at all. Seems like we should keep them separate.

src/applications/common/schemaform/save-in-progress/FormStartControls.jsx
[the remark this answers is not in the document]
cehsu: Simplifying this would be nice! Looping @jbalboni for additional insights, while I look into how this would work.

src/applications/common/schemaform/save-in-progress/actions.js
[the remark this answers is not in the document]
cehsu: Throwing an error here routes the user to the form error page. Would it make more sense to set this up as an if else statement instead?

src/applications/common/schemaform/save-in-progress/reducers.js
cehsu: Thanks @rianfowler, great point, I will create a separate error action.

src/applications/common/schemaform/save-in-progress/SaveInProgressErrorPage.jsx
cehsu: That's an interesting point @cvalarida, @kreek has explained that whenever the form is started or resumed (both are handled by `fetchInProgressForm`, within which @jbalboni has advised handling these types of gating checks), we will want to check the ITF, and there are at least three types of scenarios to consider:
1. when the user is on the introduction page or on any of the static pages with the SIP Application Status card and tries to either start, or resume their application
2. Once a form is "saved" if the user is on the form saved page (there is a SIP Application Status card there which enables that as well)
3. if the user tries to resume the form from the user profile page (this page has it's own version of the Application Status card, which leverages the RoutedSavableApp to handle all of the routing/loading/error handling logic)
The idea was to have a consistent error experience for all of these entry points, and to make use of the existing error flow that is already in place for forms.

test/common/schemaform/save-in-progress/actions.unit.spec.js
cehsu: These are necessary as the logic they are testing is also within a `catch`, I'm not sure why this wasn't required before.

src/applications/user-profile/helpers.jsx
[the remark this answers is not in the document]
cehsu: These can now be removed, as they ended up being necessary for the wizard, and are already in master.

src/applications/common/schemaform/save-in-progress/actions.js
cehsu: @jbalboni recommended handling the prestart gating within `fetchInProgressForm`.

src/applications/common/schemaform/save-in-progress/SaveInProgressErrorPage.jsx
jbalboni: Is it possible to limit the ITF to just being sent at the start of the form, and assumed to be in place after that? That would make this much simpler, otherwise, like Claire says, you have to somehow integrate this check with all the entry points to the form.

cehsu: Thanks @jbalboni, @kreek indicated that we must check this every time they enter/re-enter the form.

src/applications/common/schemaform/save-in-progress/FormStartControls.jsx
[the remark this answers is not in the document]
cehsu: Thanks again for your feedback @rianfowler, a ticket has been created (using your comment) to simplify fetchInProgressForm [here](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/10754).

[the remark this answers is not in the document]
cehsu: @cvalarida good point about restoring the if/else format, updated.

src/applications/common/schemaform/save-in-progress/RoutedSavableApp.jsx
```diff
@@ -159,9 +162,11 @@ class RoutedSavableApp extends React.Component {
     if (props.isLoggedIn) {
       const currentForm = props.formConfig.formId;
       const isSaved = props.savedForms.some((savedForm) => savedForm.form === currentForm);
-      const hasPrefillData = props.prefillsAvailable.includes(currentForm);
+      const hasPrefillData = props.prefillsAvailable.includes(currentForm) || currentForm === '21-526EZ'; // TODO: remove once 526 added to prefills list
```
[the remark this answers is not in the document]
cehsu: Yes it can, updated.

src/applications/common/schemaform/save-in-progress/FormStartControls.jsx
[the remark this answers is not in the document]
cehsu: `this.prestart` Helper renamed to `this.getPrestartForm` and changed slightly to clarify things a bit.
`this.props.beforeStartForm` renamed to `this.props.prestartCheck`

src/applications/common/schemaform/save-in-progress/actions.js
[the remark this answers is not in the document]
cehsu: This now uses an if/else statement rather than a try/catch to handle the resolved (rather than rejected) prestart check promise value. Network errors are logged immediately in the `apiRequest` configuration, and this gate is primarily used to determine whether the form should be entered, or if the user should be blocked from entering the form. Feedback on the new approach appreciated.

src/applications/common/schemaform/save-in-progress/FormStartControls.jsx
```diff
@@ -90,21 +111,37 @@ class FormStartControls extends React.Component {
...
   prefillAvailable: PropTypes.bool.isRequired,
   startPage: PropTypes.string.isRequired,
   startText: PropTypes.string,
   resumeOnly: PropTypes.bool
 };
 
-export default withRouter(FormStartControls);
+function mapStateToProps(state) {
+  return {
+    form: state.form
```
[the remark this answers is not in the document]
cehsu: Good question @markgreenburg, in this case, the `form` actually refers to the `formConfig`, which is a required parameter for `beforeStartForm` (which has been renamed `prestartCheck`). Should I set up an issue to discuss renaming this prop so to avoid this kind of confusion in the future?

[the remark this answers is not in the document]
cehsu: I have also renamed it to `formConfig` in mapStateToProps in an effort to improve the clarity of this for now.

src/applications/common/schemaform/save-in-progress/actions.js
cehsu: Thanks @rianfowler, that's a great point, in this case, a user is not logged in, so we use Promise.resolve and exit the call without fetching the form.

cehsu: In fact, as async functions always return a promise, we should convert these resolve/rejects to returns and throws. Will update.

src/applications/common/schemaform/save-in-progress/actions.js
cehsu: Good idea, updated.

src/applications/common/schemaform/save-in-progress/FormStartControls.jsx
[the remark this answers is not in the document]
cehsu: Actually it looks like we can remove the formConfig as an argument to prestartForm, as that is likely a premature optimization (to allow for reuse of shared prestartChecks between different forms, such as is likely to occur on these disability forms). Good catch @markgreenburg. Will update.

src/applications/common/schemaform/save-in-progress/FormStartControls.jsx
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { withRouter } from 'react-router';
```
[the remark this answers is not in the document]
cehsu: Good point, it seemed like we often pass the router down rather than import it for connected components, but that's not always the case, reverted.

src/applications/common/schemaform/save-in-progress/FormStartControls.jsx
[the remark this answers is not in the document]
cehsu: Good catch, this was part of the previous implementation, removed.

src/applications/common/schemaform/save-in-progress/FormStartControls.jsx
```diff
@@ -42,11 +49,25 @@ class FormStartControls extends React.Component {
...
-  render() {
 
+  prestartForm = () => new Promise((resolve, reject) => {
+    const { formId, migrations, prefillTransformer } = this.props;
+    const formConfig = { formId, migrations, prefillTransformer };
+    const { form: { beforeStartForm }, prestartForm } = this.props;
+    return prestartForm(beforeStartForm, formConfig, resolve, reject);
+  })
+
+  authenticate = (e) => {
```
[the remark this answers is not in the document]
cehsu: Good catch, this was part of the previous implementation, removed.

src/applications/disability-benefits/526EZ/helpers.jsx
```diff
@@ -414,13 +417,50 @@ export const disabilityOption = ({ diagnosticCode, name, ratingPercentage }) =>
...
+
+// TODO: add expired data to redux
+function successMessage(status, data, expiredData) {
+  const messages = {
+    created: (expirationDateString) => `Thank you for submitting your Intent to File for disability compensation. Your Intent to File will expire on ${expirationDateString}.`,
+    retrieved: (expirationDateString) => `Our records show that you’ve already submitted an Intent to File for disability compensation. Your Intent to File will expire on ${expirationDateString}.`,
+    renewed: (expirationDateString, expiredDateString) => `Your existing Intent to File expired on ${expiredDateString}, so we’ve created a new one for you. This new Intent to File request will expire on ${expirationDateString}.`
+  };
+  return messages[status](moment(data).format('M/D/YYYY'), moment(expiredData).format('M/D/YYYY'));
+}
```
cehsu: @peggygannon could you please share your thoughts on these success alert messages?

src/applications/disability-benefits/526EZ/helpers.jsx
```diff
@@ -414,13 +417,50 @@ export const disabilityOption = ({ diagnosticCode, name, ratingPercentage }) =>
...
+  ...commonPrestartStatuses
+};
+
+export const prestartPendingStatuses = new Set([PRESTART_STATUSES.none, PRESTART_STATUSES.expired, PRESTART_STATUSES.pending]);
+
+const successHeading = {
+  created: 'We’ve verified your Intent to File',
+  retrieved: 'We’ve retrieved your existing Intent to File',
+  renewed: 'We’ve verified your Intent to File'
+};
```
cehsu: @peggygannon could you please share your thoughts on these success alert headings?

cehsu:
Thanks for your comments @cvalarida, @markgreenburg, @jbalboni and @rianfowler. Updates have been made, PTAL.
The success alert content is TBD and will be updated with @peggygannon's recommendations.

src/applications/common/schemaform/containers/FormPage.jsx
```diff
@@ -103,6 +104,7 @@ class FormPage extends React.Component {
 
     return (
       <div className={pageClasses}>
+        {contentBeforeForm}
```
cehsu: @jbalboni `contentBeforeForm` is based on `contentAfterButtons` does this seem like the most consistent/intuitive way to allow for prestart success messages to be displayed above the form per @goldenmeanie's recommendation?

jbalboni: The way we do prefill messages is to put the prefill status in `formContext` in `RoutedSavableFormPage` and `RoutedSavableReviewPage`, then display them in the form config for the pages where they should be shown. It might be better to stay consistent with that approach, since it's similar functionality.

cehsu: Thanks much @jbalboni, I'll update.

src/applications/common/schemaform/save-in-progress/FormStartControls.jsx
```diff
@@ -90,21 +101,37 @@ class FormStartControls extends React.Component {
...
-  router: PropTypes.object.isRequired,
   formSaved: PropTypes.bool.isRequired,
   prefillAvailable: PropTypes.bool.isRequired,
   startPage: PropTypes.string.isRequired,
   startText: PropTypes.string,
   resumeOnly: PropTypes.bool
 };
 
-export default withRouter(FormStartControls);
+function mapStateToProps(state) {
```
jbalboni: I worry a little about connecting this component, since it's not immediately obvious why you would do it for just this field and not others. The prestart state and actions are integrated into the SiP code, so I wonder if we should just add these to the intro selector functions that the parent of this component uses.

cehsu: Great point @jbalboni, will update.

cehsu:
Thanks for your input on the alert content and design @peggygannon and @goldenmeanie, kindly refer to these screenshots for your review in reference to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/10717:

Errors

Not retrieved (saved form)
screen shot 2018-06-06 at 2 09 35 pm

Not retrieved (new form)
screen shot 2018-06-06 at 2 13 13 pm

Not created (new application)
screen shot 2018-06-06 at 2 32 38 pm

Not renewed (saved form, expired ITF)
screen shot 2018-06-06 at 2 39 16 pm


Success messages

Retrieved (found existing ITF)
screen shot 2018-06-06 at 10 49 10 am

Created
screen shot 2018-06-06 at 10 49 48 am

Renewed (existing ITF expired)
screen shot 2018-06-06 at 10 50 29 am

peggygannon:
@cehsu Thanks, Claire.. just a few tweaks for the below messages: 

1. Not created (new application). please change "couldn't" to "can't"
**We can’t process your Intent to File request** 
We’re sorry. We can’t process your Intent to File request at this time. Please try applying again tomorrow.

2. Retrieved (found existing ITF).
 I realized we updated the header via slack, but didn't tweak the body! Can you update message to:  
**We found your existing Intent to File**
Our records show that you already have an Intent to File for disability compensation. Your Intent to File will expire on [Month, Date, Year]
*can we make the date format: June, 6, 2018?* 

3. also, I think the Intent to File expander text is a bit long: 
Can we delete: "evidence or" in the first sentence? 
So it would now read: ....", while you prepare your application and gather supporting documents for your disability claim."

cehsu:
Thanks @peggygannon, kindly refer to these updated screenshots:

Retrieved (+ updated expander text)
image

Not created
image

kreek:
@cehsu this PRs pretty spicy 🌶- 😃 is there a way we could break it up into smaller PRs? e.g. could the ITF API calls be one PR with just `checkITFRequest` and `submitITFRequest` + specs?

cehsu:
@kreek sounds good to me, it is currently set up so reviewers can test things out locally as part of their review process, reviewers please add a 👍 to vote to split this up.

cehsu:
@cvalarida @jbalboni thanks for your comments, changes made, PTAL if you'd like to review things while the tests are being written.

src/applications/disability-benefits/526EZ/components/PrestartWrapper.jsx
```diff
@@ -0,0 +1,100 @@
...
+      );
+    }
+    if (profileLoading) {
+      content = <LoadingIndicator message="Retrieving your profile information..."/>;
+    }
+    return content;
+  }
+}
+
+PrestartWrapper.propTypes = {
```
cehsu: These need to be fixed, will update.

src/applications/disability-benefits/526EZ/components/ReviewCardField.jsx
```diff
@@ -169,11 +169,17 @@ export default class ReviewCardField extends React.Component {
 
 
   render() {
+    let content = this.getReviewView();
+    const PrestartMessage = this.props.uiSchema['ui:description'];
```
cehsu: This should probably be renamed UIDescription instead, will update.

src/applications/disability-benefits/526EZ/config/form.js
```diff
@@ -174,6 +178,7 @@ const formConfig = {
           path: 'review-veteran-details/military-service-history',
           initialData,
           uiSchema: {
+            'ui:description': PrestartMessage,
```
cehsu: This should be combined with the uiDescription on line 185, will update.

peggygannon:
Thanks, @cehsu the updated screenshots look good. 

Do you have any thoughts on my earlier suggestion to render the dates in this format: August 17, 2017, rather than 8/17/2017?

src/applications/disability-benefits/526EZ/components/ReviewCardField.jsx
cehsu: Renamed `Description`.

src/applications/disability-benefits/526EZ/config/form.js
cehsu: Actually this is fine as is.

cehsu:
Thanks for the reminder @peggygannon, would we want to spell out the entire month name or just the first 3-4 letters?

cehsu:
Also, if we update the expiration date on the ITF message, is there any interest in updating any of the dates on the Save In Progress message as well @peggygannon?

image

cehsu:
@peggygannon kindly refer to this updated screenshot with the month name or abbreviation (if the name is longer than 4 letters). We seem to use a mix of abbreviated month names and full month names, so please let me know if we should use full month names here instead.
screen shot 2018-06-08 at 2 00 34 pm

peggygannon:
@cehsu Thank you. Yes, I think we should spell out months here. I do think it's OK to abbreviate the following months for space: 
Jan.
Feb.
Aug.
Sept.
Oct.
Nov. 
Dec.

peggygannon (approved):

peggygannon:
@cehsu sorry, i didn't see your earlier note about the SIP page. Yes, I think we should spell out months for all our notifications for consistency. Please see the list above for the months that we could abbreviate for space reasons.

cehsu:
Thanks @peggygannon I'll set up a ticket for the SIP expiration date, so that it is consistent.

Also the current list of month abbreviations is as follows:
```
    'Jan.',
    'Feb.',
    'Mar.',
    'Apr.',
    'May',
    'June',
    'July',
    'Aug.',
    'Sept.',
    'Oct.',
    'Nov.',
    'Dec.'
```

The list you kindly provided above does not abbreviate April or March (as we do currently), should I replace the abbreviations in our list for March and April with the full versions?

peggygannon:
@cehsu Yes, can I would like to spell out March and April going forward (as long as making this change doesn't break anything on the back end)?

cehsu:
Thanks again @peggygannon, this only affects how we display dates, so it is fine to update, these have been updated and will now spell out March and April wherever months are abbreviated.

src/applications/disability-benefits/526EZ/components/PrestartWrapper.jsx
cehsu: This is now updated.

PR #7673, opened 2018-04-27
Personalization: Fix for redirect on index page

ncksllvn:
I refactored the redirect logic and left a reference to a function that no longer existed. We could only tell that was happening when a beta user logs in on the index page and I hadn't caught it til now.


bshyong (approved):
LGTM

PR #7676, opened 2018-04-30
Feature/6082

accbjt:
on focus changes for this ticket. design-system pull request needs to be approved before all onfocus changes can take effect.

https://github.com/department-of-veterans-affairs/vets.gov-team/issues/6082


src/platform/site-wide/sass/overrides/uswds/focus.scss
```diff
@@ -0,0 +1,138 @@
+@mixin focus-gold-light-outline {
```
jbalboni: It seems like these mixins could go in the design-system, since it's likely that anything built with it would have a need for them.

src/platform/site-wide/sass/overrides/uswds/focus.scss
```diff
@@ -0,0 +1,138 @@
...
+  outline-offset: 2px;
+}
+
+@mixin focus-gold-lighter-highlight {
+  background: $color-gold-lighter;
+  outline: 2px solid $color-gold-lighter;
+  outline-offset: 0;
+}
+
+a:focus {
```
jbalboni: The rules for a, input, button, and select feel like they should be in the design system.

src/platform/site-wide/sass/overrides/uswds/focus.scss
```diff
@@ -0,0 +1,138 @@
...
+
+input:focus {
+  @include focus-gold-light-outline;
+}
+
+select:focus {
+  @include focus-gold-light-outline;
+}
+
+.progress-bar-segmented:focus {
```
jbalboni: It seems like most of the rules in here should live in (or near) the files where the other style rules for these classes are.

accbjt:
@jbalboni I made the changes you suggested. I would need design-systems to be updated and merged before this gets merged. Here is the pull request for that one.

https://github.com/department-of-veterans-affairs/design-system/pull/123

jkassemi:
Who has point on getting this out?

jbalboni:
@jkassemi I'm not sure this PR is still relevant.

PR #7680, opened 2018-05-01
Feature/9144

accbjt:


accbjt:
@jbalboni Let me know if I put my css file in the correct place. I did not see any examples of where we put css code for specific elements.

jbalboni:
@accbjt Can you include a screenshot of your change in here? We try to do that for most PRs.

accbjt:
@jbalboni here is a screenshot of what it should look like.
screen shot 2018-05-07 at 12 10 08 pm

src/platform/site-wide/sass/schedule-an-appointment.scss
```diff
@@ -0,0 +1,7 @@
+button#crisis-expander-link {
```
jbalboni: We have a `va-button-link` class that should turn a button into a correctly styled link. Will that work for your use case?

accbjt: Okay I will try using it.

accbjt (commented):
Made the change

jbalboni (approved):

PR #7685, opened 2018-05-02
9698 686 chapter 1

lbai001:
screen shot 2018-05-02 at 1 31 17 am

screencapture-localhost-3001-disability-benefits-686-dependent-status-veteran-information-2018-05-02-01_53_29


lbai001:
@jbalboni @rianfowler 
1. I need help on is the section title "Applicant Information" under relationship radios, how to insert that in the schema form
2. Any more tests need to be added
3. The better way to do conditional requires

src/js/disability-benefits/686/config/form.js
```diff
@@ -1,42 +1,140 @@
 // import fullSchema from 'vets-json-schema/dist/686-schema.json';
+import _ from 'lodash';
```
jbalboni: We try to use `lodash/fp` for the forms, generally, since it avoid mutating the arguments you pass.

src/js/disability-benefits/686/config/form.js
```diff
@@ -1,42 +1,140 @@
...
+                expandUnder: 'relationship',
+                expandUnderCondition: (field) => field === 'spouse' || field === 'child' || field === 'other'
+              }
+            },
           },
           schema: {
             type: 'object',
+            required: ['ssnOrVa', 'relationship'],
             properties: {
+              fullName,
```
jbalboni: I think you want this to be `veteranFullName`, since that's the name of the property in the schema: https://github.com/department-of-veterans-affairs/vets-json-schema/blob/master/dist/21-686C-schema.json#L384

src/js/disability-benefits/686/config/form.js
```diff
@@ -1,42 +1,140 @@
...
+              ssnOrVa: {
+                type: 'string',
+              },
+              relationship: {
```
jbalboni: Since this field isn't in the schema (or the paper form), I think we should call it `view:relationship`, so it doesn't get submitted to the backend.

src/js/disability-benefits/686/config/form.js
```diff
@@ -1,42 +1,140 @@
...
+                  'veteran',
+                  'spouse',
+                  'child',
+                  'other'
+                ]
+              },
+              'view:applicantInfo': {
+                type: 'object',
+                properties: {
+                  fullName: {
```
jbalboni: I think you want this to be `claimantFullName`, since that's the other name property in the schema: https://github.com/department-of-veterans-affairs/vets-json-schema/blob/master/dist/21-686C-schema.json#L381

You can import it from `fullSchema.properties` and you won't need to define all the first/middle/last fields in here.

src/js/disability-benefits/686/config/form.js
```diff
@@ -1,42 +1,140 @@
...
+                },
+                suffix: {
+                  'ui:title': 'Suffix',
+                  'ui:options': {
+                    widgetClassNames: 'form-select-medium'
+                  }
+                }
+              },
+              ssn: _.assign(ssnUI, {
+                'ui:title': 'Social Security number',
```
jbalboni: I think this may already be the title in ssnUI and you don't need to overwrite it.

src/js/disability-benefits/686/config/form.js
```diff
@@ -1,42 +1,140 @@
...
+              'ui:title': 'Relationship to Veteran',
+              'ui:widget': 'radio',
+              'ui:options': {
+                labels: relationshipLabels
+              }
+            },
+            'view:applicantInfo': {
+              fullName: {
+                'ui:title': 'Applicant Information',
+                first: {
```
jbalboni: You can do something similar here to what you did with `ssnOrVa`, and use `_.merge(fullNameUI, ...)`. That will let you add the `ui:required` fields without having to redefined all the titles and options.

jbalboni (commented):
This is a good start. I have some comments mostly around how the schema and common definitions are being used. I also want to confirm that having a single field for ssn and file number is what we want, but that's a design question.

src/js/disability-benefits/686/tests/config/applicantInfo.unit.spec.js
```diff
@@ -0,0 +1,104 @@
+import React from 'react';
+import { expect } from 'chai';
+import sinon from 'sinon';
+
+import { DefinitionTester, getFormDOM, submitForm } from '../../../../../platform/testing/unit/schemaform-utils.jsx';
+import formConfig from '../../config/form';
+import ReactTestUtils from 'react-dom/test-utils';
```
jbalboni: This is not a fun change, but we try to write new tests with Enzyme, instead of ReactTestUtils. It's pretty similar and the basic logic is the same, just some of the commands are different. You can look at any of the tests in `src/js/vre/tests/config` or `src/js/vic-v2/tests/config` to get examples.

src/js/disability-benefits/686/config/form.js
```diff
@@ -1,42 +1,113 @@
...
       pages: {
-        page1: {
-          path: 'first-page',
-          title: 'First Page',
+        veteranInformation: {
+          path: 'veteran-information',
+          title: 'Veteran Information',
           uiSchema: {
+            veteranFullName: fullNameUI,
+            ssnOrVa: _.merge({}, ssnOrVaUI, {
```
jbalboni: You don't need the `{}` here.

jbalboni (commented):
The config looks great now, thanks. The tests look correct as far as what they cover, but they're using an older style, which I should have warned you about before.

jbalboni:
@lbai001 This looks good to me. Once we get an answer on the ssn or file number field there'll be some more changes: to either split that field into two fields, or to move the definition into the 686 folder, since we don't know if it will be a common pattern yet.

I'll push to get an answer on that question today.

peggygannon (approved):
@lbai001 Looks good. thanks!

jbalboni (approved):

PR #7686, opened 2018-05-02
Dragons: Fix for "undefined" showing in Address formatting

ncksllvn:
This is tricky because there are a lot of values and different formats for different address types. but basically I ditched the template strings for a nice series of ugly conditionals to make sure the address strings are being constructed correctly. There wasn't really any benefits to using the template strings - too many conditions for just adding a single space or comma was confusing with template interpolation.

Resolves department-of-veterans-affairs/vets.gov-team/issues/10222


bshyong (approved):
LGTM we eventually might want to have common formatters for the 3 types of addresses - I had to do this formatting for facility locator addresses as well

PR #7689, opened 2018-05-02
remove `changePostalToZip` function for form 1010ez

jholton:
remove `changePostalToZip` function, formerly used when submitting 1010ez form


jbalboni (approved):
Looks good to me. Once this gets merged into master and released in staging, we should submit a form in staging to make sure it goes through.

PR #7692, opened 2018-05-03
Terms and Conditions for Medical Information

U-DON:
Resolves department-of-veterans-affairs/vets.gov-team#10051.


src/platform/user/authorization/containers/MHVApp.jsx
```diff
@@ -52,52 +49,71 @@ export class MHVApp extends React.Component {
     if (this.isIneligible()) { return; }
 
     if (this.needsTermsAcceptance()) {
-      this.props.fetchLatestTerms(TERMS_NAME);
+      const redirectQuery = { tc_redirect: window.location.pathname }; // eslint-disable-line camelcase
+      const termsConditionsUrl = appendQuery('/health-care/medical-information-terms-conditions', redirectQuery);
+      window.location = termsConditionsUrl;
```
U-DON: This goes to the T&C page with the current health tool's URL as a query param. That allows acceptance to redirect back here and use another query param as a flag to trigger a success message.

src/platform/user/authorization/containers/MHVApp.jsx
```diff
@@ -52,52 +49,71 @@ export class MHVApp extends React.Component {
     if (this.isIneligible()) { return; }
 
     if (this.needsTermsAcceptance()) {
-      this.props.fetchLatestTerms(TERMS_NAME);
+      const redirectQuery = { tc_redirect: window.location.pathname }; // eslint-disable-line camelcase
+      const termsConditionsUrl = appendQuery('/health-care/medical-information-terms-conditions', redirectQuery);
+      window.location.replace(termsConditionsUrl);
```
U-DON: This goes to the T&C page with the current health tool's URL as a query param. That allows acceptance to redirect back here and use another query param as a flag to trigger a success message.

bshyong (approved):

dhall (approved):
LGTM!

src/js/terms-and-conditions/containers/MhvTermsAndConditions.jsx
```diff
@@ -0,0 +1,216 @@
...
+
+  handleSubmit = (e) => {
+    e.preventDefault();
+    this.props.acceptTerms(TERMS_NAME);
+    this.setState({ showAcceptedMessage: true }, () => {
+      scroller.scrollTo('banner', getScrollOptions());
+      this.redirect();
+    });
+  }
+
```
ncksllvn: Would it be worth returning a promise from the `acceptTerms` action and using `setState` in a `.then`?  I ran it locally and the API endpoint wasn't configured so it returned a `404`, but it was kinda confusing how it still showed the success message.

src/js/terms-and-conditions/containers/MhvTermsAndConditions.jsx
```diff
@@ -0,0 +1,216 @@
...
+      bannerProps = unagreedBannerProps;
+    }
+
+    return bannerProps && (
+      <ScrollElement name="banner">
+        <AlertBox
+          {...bannerProps}
+          isVisible
+          status="success"
+          onCloseAlert={this.handleCloseBanner}/>
```
ncksllvn: Should we add anything here to account for errors stored in `this.props.errors`? I see that added into the reducer but don't think it's being rendered.

src/js/terms-and-conditions/actions/index.js
```diff
@@ -0,0 +1,68 @@
...
+    apiRequest(
+      TC_ACCEPTANCE_URI(termsName),
+      settings,
+      () => {
+        recordEvent({ event: 'terms-accepted' });
+        dispatch({ type: ACCEPT_LATEST_TERMS_SUCCESS });
+      },
+      errors => dispatch({ type: ACCEPT_LATEST_TERMS_FAILURE, errors })
+    );
+  };
```
ncksllvn: `apiRequest` does return a Promise in case you want to return it and use it as a quick method of showing success/failure state that aren't worth storing in the store.

ncksllvn (approved):
LGTM

src/js/terms-and-conditions/containers/MhvTermsAndConditions.jsx
U-DON: Good point. We probably don't want to show the success before the API returns with a success. Usually I would handle this in `componentDidUpdate`.

```
  componentDidUpdate() {
    if (!prevProps.accepted && this.props.accepted) {
      this.setState({ showAcceptedMessage: true }, () => {
        scroller.scrollTo('banner', getScrollOptions());
        this.redirect();
      });
    }
  }
```

src/js/terms-and-conditions/containers/MhvTermsAndConditions.jsx
U-DON: We probably should. Don't have design/copy for it yet, though.

src/js/terms-and-conditions/containers/MhvTermsAndConditions.jsx
ncksllvn: Ah nice! That's a better way to do it than my idea but I still miss the days where you just used Promises for situations like that.

BBarone1:
subscribing to this thread as this change will impact google analytics tracking for personalization once merged in

U-DON:
@melwoodard: A warning message would seem to make more sense as a response to Cancel. Redirecting to go back could work too, but I will defer to @ehuntdsva on the design decision.

ehuntdsva:
@U-DON Apologies - I was working too quickly on the designs and forgot to change the BG color on that banner. @melwoodard is correct - on the "after cancelling T&C" screen, it should be the yellow "warning" BG color. 

@melwoodard for reference, here are the user flows.
flows_v2

PR #7695, opened 2018-05-03
Use ReviewCardField

cvalarida:
It's not nearly as complicated as it looks from the diff.


cehsu (commented):
I'm not sure we want to completely remove the VerifiedReviewContainer and Page, won't we still need those when using custom fields? This harkens back to the question of the purpose of this abstraction, which seems more like a general purpose presentational page component/container than a form data-determined field component (although the latter implementation achieves some additional code reuse).

src/js/disability-benefits/526EZ/components/ReviewCardField.jsx
```diff
@@ -181,14 +181,18 @@ export default class ReviewCardField extends React.Component {
...
+        [PropTypes.element, PropTypes.func]
+      ).isRequired
     }).isRequired
   }).isRequired,
   schema: PropTypes.object.isRequired,
   errorSchema: PropTypes.object.isRequired,
   idSchema: PropTypes.object.isRequired,
   registry: PropTypes.shape({
-    SchemaField: PropTypes.element.isRequired,
+    fields: PropTypes.shape({
```
cvalarida: This was throwing a prop type warning because I forgot it was `registry.fields.SchemaField`.

src/js/disability-benefits/526EZ/helpers.jsx
```diff
@@ -517,7 +466,11 @@ const AddressViewField = ({ formData }) => {
   let zipString;
   let stateString;
   let cityString;
-  if (zipCode) zipString = `${zipCode.slice(0, 5)}-${zipCode.slice(5)}`;
+  if (zipCode) {
+    const firstFive = zipCode.slice(0, 5);
+    const lastChunk = zipCode.slice(5);
+    zipString = `${firstFive}${lastChunk ? `-${lastChunk}` : ''}`;
```
cvalarida: If we only had 5 digits in the zip, it was showing up as `12345-` previously.

src/js/disability-benefits/526EZ/pages/paymentInfo.js
```diff
@@ -57,81 +56,64 @@ const paymentInformationViewField = ({ formData }) => {
...
-          hideIf: formData => formData.directDeposit.accountType === 'NOBANK'
-        }
-      },
-      'view:noBank': {
-        'ui:description': (<p>The Department of Treasury requires all federal benefit payments be made by electronic funds transfer (EFT), also called direct deposit. If you don’t have a bank account, you must get your payment through Direct Express Debit MasterCard. To request a Direct Express Debit MasterCard you must apply at <a href="https://www.usdirectexpress.com">www.usdirectexpress.com</a> or by telephone at 1-800-333-1795. If you chose not to enroll, you must contact representatives handling waiver requests for the Department of Treasury at 1-888-224-2950. They’ll address any questions or concerns you may have and encourage your participation in EFT.</p>),
-        'ui:options': {
-          hideIf: formData => formData.directDeposit.accountType !== 'NOBANK'
-        }
-      },
+export const uiSchema = {
```
cvalarida: The only notable changes here (and the other pages) are:
- Exporting `uiSchema` and `schema` directly
- Using `fullSchema526EZ` for the definitions
- Specifying `ui:field` and giving it its necessary `ui:options`

src/js/disability-benefits/526EZ/tests/00-all-fields.e2e.spec.js
```diff
@@ -14,7 +14,9 @@ const runTest = E2eHelpers.createE2eTest(client => {
       .waitForElementVisible('body', Timeouts.normal)
       .assert.title('Apply for education benefits: Vets.gov')
       .waitForElementVisible('.schemaform-title', Timeouts.slow) // First render of React may be slow.
-      .click('.schemaform-start-button');
+      .click('.schemaform-intro .usa-button-primary');
+
+    // Gets hung up at the login screen
```
cvalarida: We aren't _actually_ running these right now, so we don't have to fix it in this PR. (I didn't break anything here!) I'll be writing up another ticket to address e2e test troubles.

cvalarida: The ticket for this work now lives at https://github.com/department-of-veterans-affairs/vets.gov-team/issues/10255.

cvalarida:
@cehsu 

> ...won't we still need those when using custom fields?

Why?

At this point, the problem we're trying to solve is whether or not the user has the data, not whether or not the user is verified. If they're not verified, they won't have the data. If they are verified, they may or may not have the data.

Later, we can add functionality to `ReviewCardField` to only show the edit state without the gray background if the user isn't verified to mimic what it would look like on other forms. (At that point, they would have no reason to review anything, so that makes sense.)

cvalarida:
@cehsu and I spoke offline and figured out the confusion. Basically, the concern surrounded nesting custom fields within custom fields, but that shouldn't be a problem because of schemaform magic.

I can go in more depth if necessary, but I think we've got the confusion squared away.

Huzzah for collaboration!

src/js/disability-benefits/526EZ/pages/paymentInfo.js
```diff
@@ -57,81 +56,64 @@ const paymentInformationViewField = ({ formData }) => {
...
+    },
+    bankName: {
+      'ui:title': 'Bank name',
       'ui:options': {
-        hideTitle: true
+        hideIf: formData => formData.directDeposit.accountType === 'NOBANK'
       }
+    },
+    'view:noBank': {
+      'ui:description': (<p>The Department of Treasury requires all federal benefit payments be made by electronic funds transfer (EFT), also called direct deposit. If you don’t have a bank account, you must get your payment through Direct Express Debit MasterCard. To request a Direct Express Debit MasterCard you must apply at <a href="https://www.usdirectexpress.com">www.usdirectexpress.com</a> or by telephone at 1-800-333-1795. If you chose not to enroll, you must contact representatives handling waiver requests for the Department of Treasury at 1-888-224-2950. They’ll address any questions or concerns you may have and encourage your participation in EFT.</p>),
```
cehsu: We may want to check with @peggygannon whether the phone numbers here should be hyperlinked, it's not clear that there is a standard convention for presenting phone numbers. If so, we can address in a later PR.

src/js/disability-benefits/526EZ/pages/primaryAddress.js
```diff
@@ -464,161 +462,142 @@ const addressUISchema = (addressName, title) => {
...
-        pattern: "([a-zA-Z0-9-'.,,&#]([a-zA-Z0-9-'.,,&# ])?)+$"
-      },
-      city: {
-        type: 'string',
-        maxLength: 35,
-        pattern: "([a-zA-Z0-9-'.#]([a-zA-Z0-9-'.# ])?)+$"
-      },
-      zipCode: {
-        type: 'string'
+const addressSchema = {
```
cehsu: Is the idea to use the PCIU address schema for this eventually?

src/js/disability-benefits/526EZ/pages/primaryAddress.js
```diff
@@ -464,161 +462,142 @@ const addressUISchema = (addressName, title) => {
...
+      'ui:options': {
+        widgetClassNames: 'va-input-medium-large'
+      }
+    },
+    secondaryPhone: {
+      'ui:title': 'Secondary telephone number',
+      'ui:widget': SSNWidget,
+      'ui:validations': [validatePhone],
+      'ui:errorMessages': {
+        pattern: 'Phone numbers must be at least 10 digits (dashes allowed)'
```
cehsu: Should this be 7 instead of 10? Or "Either 7 or 10 digits (dashes allowed)" perhaps a question for @peggygannon ?

src/js/disability-benefits/526EZ/pages/veteranInfo.js
```diff
@@ -1,80 +1,64 @@
...
+  socialSecurityNumber: merge(ssnUI, {
+    'ui:title': 'Social Security number (must have this or a VA file number)',
+    'ui:required': form => !form.vaFileNumber
+  }),
+  vaFileNumber: {
+    'ui:title': 'VA file number (must have this or a Social Security number)',
+    'ui:required': form => !form.socialSecurityNumber,
+    'ui:help': VAFileNumberDescription,
+    'ui:errorMessages': {
+      pattern: 'Your VA file number must be between 7 to 9 digits'
```
cehsu: Similar to above, should this be 7 or 9?

cehsu (approved):
LGTM. Some minor ideas/questions for consideration.

src/js/disability-benefits/526EZ/helpers.jsx
cvalarida: :thinking: Yeah, that might clear things up a bit. I wasn't super fond of a template string nested in a template string myself.

src/js/disability-benefits/526EZ/pages/primaryAddress.js
cvalarida: It actually should be `exactly 10 digits` because that's what EVSS expects. 3 for the `areaCode`, 7 for the `phoneNumber`. Should also check that the validation does actually accept dashes...I think we don't right now, so that's misleading...

At any rate, I don't know that it makes sense to change it in this particular PR because I only modified the whitespace here, not the actual content. What would be the best way to track this work? Should we have a ticket to go through all the validation and make sure the error messages are appropriate?

src/js/disability-benefits/526EZ/pages/primaryAddress.js
cvalarida: I think so, yes.

src/js/disability-benefits/526EZ/pages/paymentInfo.js
cvalarida: Yeah, it probably should be hyperlinked. Seems like we've been consistent doing it that way of late.

src/js/disability-benefits/526EZ/pages/paymentInfo.js
```diff
@@ -57,81 +56,64 @@ const paymentInformationViewField = ({ formData }) => {
...
+      'ui:widget': 'radio',
+      'ui:options': {
+        hideTitle: true,
+        labels: accountLabels
+      }
+    },
+    accountNumber: {
+      'ui:title': 'Account number',
+      'ui:options': {
+        hideIf: formData => formData.directDeposit.accountType === 'NOBANK'
```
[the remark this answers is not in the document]
cvalarida: I'm a fan of `const`s.

src/js/disability-benefits/526EZ/helpers.jsx
```diff
@@ -538,7 +491,7 @@ export const PrimaryAddressViewField = ({ formData }) => {
   const {
     mailingAddress, primaryPhone, secondaryPhone,
     emailAddress, forwardingAddress
-  } = formData.veteran;
```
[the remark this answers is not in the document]
cvalarida: The way we're using this now, the current `formData` is what `formData.veteran` was previously.

src/js/disability-benefits/526EZ/pages/veteranInfo.js
cvalarida: Yeah, looks like it. This also looks to be this way (wrong) everywhere else I've seen it, too. :grimacing:

PR #7697, opened 2018-05-05
526 require evidence mg

markgreenburg:
- Adds validation that ensures at least one type of evidence selected for each disability.
- Updates some helpers and tests.

The formatting for the checkboxes is a bit of a regression from when they were spaced nicely, but this seems to be the standard form pattern for groups of checkboxes and there's no way to show a function-based label / description without having things formatted this way. The label / description is needed in order for the validation error to have something to attach to.

screen shot 2018-05-07 at 1 54 00 pm

screen shot 2018-05-07 at 1 53 49 pm


src/js/common/schemaform/validation.js
```diff
@@ -87,17 +87,33 @@ export function transformErrors(errors, uiSchema) {
  * This pulls custom validations specified in the uiSchema and validates the formData
  * against them.
  *
- * Expects validations that look like:
+ * Expects validations that EITHER look like:
```
markgreenburg: I didn't end up needing to use a custom validator, but it's the approach I looked into first. I was somewhat frustrated by the fact that I didn't realize there were two possible ways of creating a custom validation, so decided to update the documentation even though I didn't end up using a custom validator.

src/js/disability-benefits/526EZ/tests/config/evidenceType.unit.spec.jsx
```diff
@@ -39,12 +39,32 @@ describe('Disability benefits 526EZ evidence type', () => {
         uiSchema={uiSchema}/>
     );
 
-    selectCheckbox(form, 'root_view:vaMedicalRecords');
-    selectCheckbox(form, 'root_view:privateMedicalRecords');
-    selectCheckbox(form, 'root_view:otherEvidence');
+    selectCheckbox(form, 'root_view:selectableEvidenceTypes_view:vaMedicalRecords', true);
```
markgreenburg: Not sure if this ever worked right, because `selectCheckbox` doesn't set a default value for the third argument. 🤷‍♂️

src/js/disability-benefits/526EZ/tests/config/evidenceType.unit.spec.jsx
```diff
@@ -39,12 +39,32 @@ describe('Disability benefits 526EZ evidence type', () => {
...
+    selectCheckbox(form, 'root_view:selectableEvidenceTypes_view:privateMedicalRecords', true);
+    selectCheckbox(form, 'root_view:selectableEvidenceTypes_view:otherEvidence', true);
 
     form.find('form').simulate('submit');
-    expect(form.find('.usa-input-error').length).to.equal(0);
+    expect(form.find('.usa-input-error-message').length).to.equal(0);
```
markgreenburg: ...and the class selector was partially missing. Confirmed on latest master that it's actually `.usa-input-error-message`

cvalarida: :grimacing: Good catch!

src/js/common/schemaform/validation.js
cvalarida: Did you see the documentation concerning validation in [the formConfig document](https://github.com/department-of-veterans-affairs/vets-website/blob/master/docs/schemaform/form-config.md#i-need-to-write-custom-validation)?

src/js/common/schemaform/validation.js
```diff
@@ -87,17 +87,33 @@ export function transformErrors(errors, uiSchema) {
...
  *   "ui:validations": [
  *     someValidationFunction,
  *   ]
  * }
  *
- * The function is passed errors, fieldData, pageData, formData, and otherData and
- * should call addError to add the error.
- *
+ * with each validation function having a signature of:
+ * validation(pathErrors, currentData, formData, schema, uiSchema['ui:errorMessages'], currentIndex)
```
cvalarida: These signatures are different...looks like the last change was over a year ago. Oh! By me. Yeah, it's definitely outdated by now. Thanks for updating this!

Can you actually make this consistent with what's in the form config doc I linked to above?

src/js/common/schemaform/validation.js
```diff
@@ -87,17 +87,33 @@ export function transformErrors(errors, uiSchema) {
...
+ *  "ui:validations": [
+ *    {
+ *      options: {},
+ *      validator: someValidationFunction
+ *    }
+ *  ]
+ * }
+ * 
+ * In which case the validator function is passed the options object along with everything else:
+ * validation.validator(pathErrors, currentData, formData, schema, uiSchema['ui:errorMessages'], validation.options, currentIndex)
```
cvalarida: Same thing here with for consistency.

src/js/disability-benefits/526EZ/tests/config/evidenceType.unit.spec.jsx
cvalarida: Was this breaking when you were running it locally?

cvalarida (approved):
Looks good!

src/js/common/schemaform/validation.js
markgreenburg: I...have never seen this documentation. Only the the [walkthrough](https://github.com/department-of-veterans-affairs/vets-website/blob/master/docs/schemaform/walkthrough.md), though I'm not sure why, since they're both in the same directory.

src/js/disability-benefits/526EZ/tests/config/evidenceType.unit.spec.jsx
markgreenburg: It was, after I made changes to the schema.

src/js/disability-benefits/526EZ/config/form.js
```diff
@@ -331,15 +332,25 @@ const formConfig = {
...
-                'ui:description': evidenceTypesDescription,
-                'view:vaMedicalRecords': {
-                  'ui:title': 'VA medical records'
-                },
-                'view:privateMedicalRecords': {
-                  'ui:title': 'Private medical records'
-                },
-                'view:otherEvidence': {
-                  'ui:title': 'Lay statements or other evidence'
+                'view:selectableEvidenceTypes': {
```
cehsu: Should we plan to create an errorable boolean group widget based off of this in the future?

src/js/disability-benefits/526EZ/config/form.js
```diff
@@ -331,15 +332,25 @@ const formConfig = {
...
+                  'ui:options': {
+                    // Only way to get access to the disability info like 'name' within this nested schema
+                    updateSchema: (form, schema, uiSchema, index) => ({ title: getEvidenceTypesDescription(form, index) }),
```
cehsu: Why do we need to know the disability name? The second parameter of 'ui:validations' is the field data, is that sufficient?

cehsu (approved):
LGTM. Had one minor question and another idea.

src/js/disability-benefits/526EZ/config/form.js
cvalarida: This is to show the description for these check boxes. `What supporting evidence do you have that shows how your {disabilityName} has worsened since VA rated your disability?` `ui:validations` isn't related to this particular change.

We need this `updateSchema` now because these `view:` check boxes are now nested inside another object (`view:selectableEvidenceTypes`), and if we kept the `ui:description` the same, it made for some really wonky formatting. Mark and I scratched our head over this one for a while before asking Jeff (who came up with this particular solution).

It's a weird one for sure!

PR #7704, opened 2018-05-08
Personalization fixes

bshyong:
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/10153
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/10161
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/10015


src/js/personalization/dashboard-beta/containers/PrescriptionsWidget.jsx
```diff
@@ -58,6 +59,8 @@ class PrescriptionsWidget extends React.Component {
 
 const mapStateToProps = (state) => {
   const rxState = state.health.rx;
+  const profileState = state.user.profile;
+  const canAccessRx = profileState.services.includes('rx');
```
ncksllvn: Hadn't thought of it until now, but it seems like we should begin storing the services values as constants somewhere.

ncksllvn (approved):
LGTM

src/js/personalization/dashboard-beta/containers/PrescriptionsWidget.jsx
bshyong: Yep; it's on a list of things we should refactor once the dust settles..

We just need to mirror what the backend is using. 

https://github.com/department-of-veterans-affairs/vets-api/blob/master/lib/backend_services.rb

PR #7706, opened 2018-05-08
add temporary nsp exceptions

omgitsbillryan:
This adds exceptions for the two `nsp` vulns.  It allows us to complete `vets-website` builds again.

### DIFF
--- a/.nsprc
+++ b/.nsprc
@@ -1,0 +1,6 @@
+{
+  "exceptions": [
+    "https://nodesecurity.io/advisories/577",
+    "https://nodesecurity.io/advisories/157"
+  ]
+}


wyattwalter (approved):
👍 Just need to make sure this stays temporary 😄

omgitsbillryan:
Upgrading `modernizr` (which uses an outdated version of `lodash`) is not an option since we are at the latest `3.6.0`.

The other path is a deeper: `vets-website@1.0.0 > nyc@11.7.1 > micromatch@2.3.11 > braces@1.8.5 > expand-range@1.8.2 > fill-range@2.2.3 > randomatic@1.1.7` and will need to be investigated.

omgitsbillryan:
@wyattwalter 
image

ncksllvn:
@omgitsbillryan haha that is perfect

omgitsbillryan:
Turns out we're also on the latest version of `nyc` https://www.npmjs.com/package/nyc . So the real solution is to either

1. stop using these libraries
2. Keep the `nsp` security exceptions in place until `nyc` & / or `modernizr` release an update that updates their use of the outdates packages (`randomatic` & `lodash` respectively)

PR #7707, opened 2018-05-08
Import babel-polyfill

cvalarida:
It feels like a bit of a sledgehammer approach, but it works and doesn't _seem_ to have any undesired consequences. If anybody can think of how to handle this with more finesse, I'd be happy to hear it.

### DIFF
--- a/src/platform/testing/unit/mocha-setup.js
+++ b/src/platform/testing/unit/mocha-setup.js
@@ -8,6 +8,7 @@ import chai from 'chai';
 import chaiAsPromised from 'chai-as-promised';
 import { JSDOM } from 'jsdom';
 import 'blob-polyfill';
+import 'babel-polyfill';
 // import sinon from 'sinon'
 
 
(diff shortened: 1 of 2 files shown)


src/platform/testing/unit/mocha-setup.js
```diff
@@ -8,6 +8,7 @@ import chai from 'chai';
 import chaiAsPromised from 'chai-as-promised';
 import { JSDOM } from 'jsdom';
 import 'blob-polyfill';
+import 'babel-polyfill';
 // import sinon from 'sinon'
```
ncksllvn: We could also move this into `platform/testing/unit/mocha.opts` as -
```
--require babel-polyfill
```
Not sure which way makes more sense.

ncksllvn (approved):
Nice find!

src/platform/testing/unit/mocha-setup.js
cvalarida: Yeah, it probably makes sense to keep all the side-effect-only imports in the mocha.opts file. I'll move this and the `blob-polyfill` up (assuming all the unit tests pass locally when I do so).

PR #7708, opened 2018-05-09
Dragons: Send all Profile requests simultaneously

ncksllvn:
This refactors the code for sending the 7 profile requests and constructing the profile object. It was previously sending the requests one after another, and now sends them simultaneously.  My goal here was to do this as functionally as possible, and without explicitly using a Promise.

Resolves department-of-veterans-affairs/vets.gov-team/issues/10217

### DIFF
--- a/src/js/personalization/va-profile-beta/actions/index.js
+++ b/src/js/personalization/va-profile-beta/actions/index.js
@@ -71,8 +71,7 @@ function saveFieldHandler(apiRoute, requestStartAction, requestSuccessAction, re
 }
 
 async function sendProfileRequests() {
-  const result = {};
-  const requests = [
+  return [
     ['email', '/profile/email'],
     ['primaryTelephone', '/profile/primary_phone'],
     ['alternateTelephone', '/profile/alternate_phone'],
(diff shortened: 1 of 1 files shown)


ncksllvn:
@wyattwalter added you as a CR for visibility

src/js/personalization/va-profile-beta/actions/index.js
```diff
@@ -71,28 +71,29 @@ function saveFieldHandler(apiRoute, requestStartAction, requestSuccessAction, re
 }
 
 async function sendProfileRequests() {
```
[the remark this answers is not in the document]
ncksllvn: Yeah it's just a helper function for `fetchVaProfile`. Should it be separated out better, in a different file or something?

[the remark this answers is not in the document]
jbalboni: Oh, I just wasn't paying enough attention and didn't see that it wasn't exported. Seems fine, then.

src/js/personalization/va-profile-beta/actions/index.js
```diff
@@ -71,28 +71,29 @@ function saveFieldHandler(apiRoute, requestStartAction, requestSuccessAction, re
...
-  const result = {};
-  const requests = [
+  return [
```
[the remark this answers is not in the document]
ncksllvn: Yeah that would be ideal but it isn't a single endpoint because that would probably only make sense for our use case and one of the API goals was to be flexible for others. Right now we're dreaming of GraphQL for this, but a batch request also a good idea

src/js/personalization/va-profile-beta/actions/index.js
```diff
@@ -71,28 +71,29 @@ function saveFieldHandler(apiRoute, requestStartAction, requestSuccessAction, re
...
-
-  return result;
+    return [propertyName, propertyValue];
+  }).reduce(async (resultPromise, pendingRequest) => {
+    const result = await resultPromise;
+    const [propertyName, propertyValue] = await pendingRequest;
+    result[propertyName] = propertyValue;
+    return result;
+  }, {});
 }
```
ncksllvn: I'm really curious about thoughts on this. The .map above already sent all of requests, and this .reduce just constructs the profile result. I can't decide if it's too cryptic, or if as await/async become more mainstream this will become a common pattern for async functional code.

jbalboni: Yeah, I was going back and forth. I don't think it's clearer than using `Promise.all`, but I'm not sure if that's just because I'm not used to async/await yet.

bshyong (approved):
LGTM

src/js/personalization/va-profile-beta/actions/index.js
bshyong: maybe you could add a one-line comment that clarifies what this map/reduce is doing in case people reading this are not familiar with async/await

bshyong (approved):

src/js/personalization/va-profile-beta/actions/index.js
ncksllvn: Yeah my feelings exactly. I'm gonna merge this as-is but will revisit it. I think this code is neat but may not be worth it if it's overly complex.

ncksllvn: @bshyong whoops missed your comment. Good idea though, will do!

PR #7711, opened 2018-05-09
9819 686 marriage info

rianfowler:
Would like this reviewed before I write the unit tests for it. Thanks @peggygannon 
download-4
download-5
download-6
download-7


rianfowler:
Should also note it only goes up to "tenth" marriage.

peggygannon (approved):
@rianfowler Looks good!

src/js/disability-benefits/686/config/form.js
```diff
@@ -168,9 +176,139 @@ const formConfig = {
...
+                'ui:options': {
+                  updateSchema: (form, schema, uiSchema, index) => {
+                    return {
+                      title: getMarriageTitleWithCurrent(form, index)
+                    };
+                  }
+                },
+                spouseFullName: {
+                  'ui:options': {
+                    updateSchema: (form, schema, uiSchema, index) => {
```
rianfowler: This is the main difference from the pensions form config.

src/js/disability-benefits/686/config/form.js
```diff
@@ -168,9 +176,139 @@ const formConfig = {
...
+                      const spouse = isCurrentMarriage(form, index) ? 'Spouse’s' : 'Former spouse’s';
+                      return _.merge(schema, {
```
jbalboni: One thing to note with this is that you're going to return a new object each time, so the schema is going to get updated on every change and cause this field to be re-rendered. It may not be noticeable, but in the past we've either used reselect or generated the different outputs outside of this function switched between them here.

jbalboni (approved):
LGTM

src/js/disability-benefits/686/config/form.js
```diff
@@ -168,9 +176,163 @@ const formConfig = {
...
+                    };
+                  }
+                },
+                spouseFullName: {
+                  'ui:options': {
+                    updateSchema: (function makeUpdateSchema() {
+
+                      let formerSpouseSchema;
+                      let currentSpouseSchema;
+
```
rianfowler: @jbalboni decided to just curry a cache for this. Reselect seemed too indirect and hardcoding the whole schema creates an extra place to change it if it needs to be updated.

PR #7712, opened 2018-05-10
Add downtime Notifications to dashboard widgets

bshyong:
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/10011
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/10290
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/10308

### DIFF
--- a/src/js/personalization/dashboard-beta/containers/ClaimsAppealsWidget.jsx
+++ b/src/js/personalization/dashboard-beta/containers/ClaimsAppealsWidget.jsx
@@ -22,6 +22,9 @@ import ClaimsUnavailable from '../../../claims-status/components/ClaimsUnavailab
 import AppealsUnavailable from '../../../claims-status/components/AppealsUnavailable';
 import ClaimsAppealsUnavailable from '../../../claims-status/components/ClaimsAppealsUnavailable';
 
+import DowntimeNotification, { services } from '../../../../platform/monitoring/DowntimeNotification';
+import AlertBox from '@department-of-veterans-affairs/jean-pants/AlertBox';
+
 import ClaimsListItem from '../components/ClaimsListItem';
 import AppealListItem from '../components/AppealsListItemV2';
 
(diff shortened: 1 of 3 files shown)


src/js/personalization/dashboard-beta/containers/ClaimsAppealsWidget.jsx
```diff
@@ -78,6 +81,28 @@ class ClaimsAppealsWidget extends React.Component {
...
+                isVisible
+                status="warning"/>
+            </div>
+          );
+        default:
+          return children;
+      }
+    };
+  }
+
```
ncksllvn: It makes me so happy to see the first implementation of the custom render function.

src/js/personalization/dashboard-beta/containers/DashboardApp.jsx
```diff
@@ -50,10 +52,67 @@ moment.updateLocale('en', {
...
+              visible={!this.state.modalDismissed}>
+              <p>We’ll be making updates to some tools and features on {downtimeWindow.startTime.format('MMMM Do')} between {downtimeWindow.startTime.format('LT')} and {downtimeWindow.endTime.format('LT')} If you have trouble using parts of the dashboard during that time, please check back soon.</p>
+              <button type="button" className="usa-button-secondary" onClick={this.dismissModal}>Continue</button>
+            </Modal>
+            {children}
+          </div>
+        );
+      default:
+        return children;
+    }
```
ncksllvn: Ah man, it's kinda unfortunate we have to have so much similar logic as the modal that works by default in DowntimeNotification. The modal content is so customized though that I don't think there would be a good way for the DowntimeNotification API to make it any easier.

ncksllvn (approved):
LGTM! Very excited to see the first application with so much render customization during downtimes. It could serve as a really good example for other engineers.

Were you able to spoof the downtime information locally somehow? Awhile back I was mocking the downtime API route and can probably set that up again if we want to be 1000% sure.

src/js/personalization/dashboard-beta/containers/DashboardApp.jsx
bshyong: @ncksllvn this PR is kind of a dirty one to get something up. I think I'd want to explore breaking out the calculation of the Downtime Notification time windows so that it is easier to use a custom notification without having to mess with render functions.

The existing Downtime Notification components can still use those calculation functions, but the expectation would be that if you want to do custom rendering you create whatever container/component you need and just use the calculation functions to get the correct windows and status.

ncksllvn: That's a great idea. I was originally writing those time windows functions as `static` so we could reuse them, but something, something, and it seemed cleaner not to do it that way in the end. Definitely game to revisit that now we have great use case.

PR #7716, opened 2018-05-10
Move VAProfile Address functions into a util file

ncksllvn:
While working on the VA Profile application, I was trying to really condense any Address logic into a set of utility functions. With so much going on with Addresses lately, it seems like it would be better to do this sooner rather than later and possibly put us a step closer towards centralizing and sharing Address logic.

Definitely feel free to move this if there's a better location.

Resolves department-of-veterans-affairs/vets.gov-team/issues/10380


bshyong (approved):
LGTM

cehsu (approved):
LGTM. Should any unit tests be written/updated?

ncksllvn:
@cehsu Yeah good idea, we were planning on dedicating a lot time to just tests across all of the Personalization work, but I think it makes sense to go ahead and write tests for this piece now. I'll go ahead and get some pushed.

ncksllvn:
@cehsu I added some test cases. Do you mind taking another look and seeing if they make sense?

src/js/personalization/va-profile-beta/utils/tests/address.unit.spec.js
```diff
@@ -0,0 +1,118 @@
...
+  it('converts a military address into a standard address format', () => {
+    const expectedResult = {
+      type: 'MILITARY',
+      countryName: 'USA',
+      addressOne: 'First',
+      addressTwo: 'Second',
+      addressThree: 'Third',
+      city: 'APO',
+      stateCode: 'AA',
+      zipCode: '12345'
```
cehsu: Do military addresses have ZIP codes?

src/js/personalization/va-profile-beta/utils/tests/address.unit.spec.js
```diff
@@ -0,0 +1,118 @@
...
+describe('getStateName', () => {
+
+  it('gets the full name of a state from its abbreviation', () => {
+    expect(addressUtils.getStateName('KY')).to.equal('Kentucky');
+    expect(addressUtils.getStateName('ky')).to.equal('Kentucky');
+  });
+
+});
+
+describe('isEmptyAddress', () => {
```
cehsu: isIncomplete might be more descriptive?

src/js/personalization/va-profile-beta/utils/address.js
```diff
@@ -38,7 +38,7 @@ const ADDRESS_TYPES = {
  * @param {Address} address
  * @returns {Address}
  */
-function consolidateAddress(address) {
+export function consolidateAddress(address) {
```
cehsu: For 'US' country code country name, I think 'United States' might be preferred, @markgreenburg pointed this out previously.

src/js/personalization/va-profile-beta/utils/address.js
```diff
@@ -70,7 +70,7 @@ function getInferredAddressType(address) {
  * @param {Address} address
  * @returns {Address}
  */
-function expandAddress(address) {
+export function expandAddress(address) {
```
cehsu: I'm a little unclear on how country is handled for military addresses, it is deleted here, and defaulted to US above. Perhaps some comments about this would be helpful, if possible.

cehsu (approved):
LGTM. I had a few questions but the test cases look comprehensive/effective.

src/js/personalization/va-profile-beta/utils/tests/address.unit.spec.js
ncksllvn: They do! But I'll update my example to match the [API's](https://github.com/department-of-veterans-affairs/vets-api/blob/1efd2c206859b1a261e762a50cdb44dc8b66462d/spec/factories/pciu_address.rb#L34)

src/js/personalization/va-profile-beta/utils/tests/address.unit.spec.js
ncksllvn: I used some of the naming conventions from Letters to decide that name but I agree that is better. I'll update that.

ncksllvn: Ah actually on second thought, there can be incomplete addresses that still contain street information or other properties, but aren't considered empty

ncksllvn:
@cehsu Made some updates that hopefully clears things up somewhat. For some reason, countryName is not part of the Military address format. I'm assuming this is because Military addresses are always in the USA. I tried to make that clearer in the test cases and code.

cehsu (commented):
If multiple addresses are displayed, is it common that they have different names or any other type of label that could be used to distinguish them? For example, mailing address, forwarding address, etc. If so, would optionally displaying those labels be of interest?

ncksllvn:
@cehsu That would be great but I don't believe there is any property on the model to help distinguish what the addresses are. I think it would have to be done at a higher level. Vet360 models may provide that information but that will require some rework already, because some property names change. Is that what you mean?

jbalboni:
Yeah, somewhere in platform/utilities or platform/forms would make sense to me.

ncksllvn:
@jbalboni I'm not sure if it all logically fits into `platform/forms/address` but I moved it there anyway to keep all of the address stuff contained in a single location. I renamed `forms/address.js` to `forms/address/index.js` so I didn't need to update the paths.

image

PR #7719, opened 2018-05-10
Autosuggest features

cvalarida:
The connected ticket is a bit ambiguous, so I'll explain more of what's going on here.

This work allows the type ahead field to accept a `getOptions` callback which can be called whenever input is updated. By supplying `setData` to the callback, we can debounce the function so it's not called _every_ keystroke.

We needed this for the facility names because the list is unreasonably long to store everything in the front end. Instead, we'll fetch the new list of options when the user is done typing.


src/js/disability-benefits/526EZ/config/form.js
```diff
@@ -1,4 +1,6 @@
 import _ from '../../../../platform/utilities/data';
+// TODO: Write a merge function
```
cvalarida: I didn't want to write two data util functions in one PR, so I stuck with using lodash for merge.

src/js/disability-benefits/526EZ/helpers.jsx
```diff
@@ -562,3 +562,17 @@ export const noFDCWarning = (
...
+  { id: 2, label: 'second' },
+  { id: 3, label: 'third' },
+  { id: 4, label: 'fourth' },
+];
+
+
+// Mimic querying the api for options
+export function queryForFacilities(input = '', setOptions) {
+  // Emulate a fast api call
+  return Promise.resolve(setOptions(options.filter(o => o.label.includes(input))));
```
cvalarida: I'll be writing another ticket to hook this up to the api. I was trying to keep this PR smaller. :man_shrugging:

test/common/schemaform/fields/AutosuggestField.unit.spec.jsx
```diff
@@ -273,4 +299,124 @@ describe('<AutosuggestField>', () => {
...
+    });
+
+    setTimeout(() => {
+      const fieldData = validationSpy.firstCall.args[1];
+      expect(fieldData.freeInput).to.be.true;
+    });
+  });
+
+
+  it.skip('should fill in other formData when a suggestion is selected', () => {});
```
cvalarida: Skipping this for now because it's arguably not necessary for MVP, but I want to keep it on our radar.

src/js/disability-benefits/526EZ/config/form.js
```diff
@@ -409,9 +424,22 @@ const formConfig = {
...
                   items: {
-                    treatmentCenterName: {
-                      'ui:title': 'Name of VA medical facility'
-                    },
+                    treatmentCenterName: autoSuggestUiSchema(
+                      'Name of VA medical facility',
+                      // Wait for 1 second before querying
+                      // It's debatable tha we want this to be _.throttle instead of _.debounce
+                      // The 1 second is super arbitrary
+                      _.debounce(1000, queryForFacilities),
```
jbalboni: I kinda feel like this should be in the field component, instead of leaving it to someone creating a form. It's generally a good practice to not fire a request on every key stroke.

cvalarida: Yeah, I left it out here because this callback isn't _necessarily_ an AJAX request. I can't think of any other legitimate use cases, but I opted for the most flexibility.

That said, I'm not opinionated. I can make the debounce wait time a configurable option. If it's then set to 0, it's _basically_ like it's not debounced at all, which solves that problem if anybody really needs it solved.

Good point. I'll move it. :+1:

src/applications/common/schemaform/fields/AutosuggestField.jsx
```diff
@@ -85,12 +100,23 @@ export default class AutosuggestField extends React.Component {
       return suggestion.id;
     }
 
-    return _.set('widget', 'autosuggest', suggestion);
+    const newSuggestion = _.set('widget', 'autosuggest', suggestion);
+    // When freeInput is true, we'll return the label to the api instead of the id
+    if (this.props.uiSchema['ui:options'].freeInput) {
```
jbalboni: What if we returned a string when `freeInput` was true? Is that too disruptive to the rest of the code in this component?

cvalarida: :thinking: Good question. I'm not sure. It seems like not a great idea because we'd be returning different data types, but I'd have to try it out to know for sure. (It's been long enough I'm not sure what the cascading effect would be.)

I'll tackle this on Monday.

src/applications/disability-benefits/526EZ/helpers.jsx
```diff
@@ -126,6 +126,7 @@ export const facilityDescription = ({ formData }) => {
 export const treatmentView = ({ formData }) => {
   const { from, to } = formData.treatmentDateRange;
 
+  const name = formData.treatmentCenterName.label || '';
```
cehsu: If no label is provided, is a value provided that can be used instead?

src/applications/disability-benefits/526EZ/helpers.jsx
```diff
@@ -557,6 +558,22 @@ export const noFDCWarning = (
   </div>
 );
 
+
+const options = [
+  { id: 1, label: 'first' },
+  { id: 2, label: 'second' },
+  { id: 3, label: 'third' },
+  { id: 4, label: 'fourth' },
```
cehsu: Do we have any guidance regarding how many of these can be added?

src/platform/utilities/data/debounce.js
```diff
@@ -0,0 +1,13 @@
...
+ * @param {function}
+ * @param {number}
+ */
+export default function debounce(wait, func) {
+  let timeoutId;
+
+  return function debouncedFunction(...args) {
+    const context = this;
+    clearTimeout(timeoutId);
+    timeoutId = setTimeout(() => func.apply(context, args), wait);
```
cehsu: Can we avoid using the `this` workaround here?

cehsu (commented):
Looks good so far. I had a few minor questions/comments.

src/applications/common/schemaform/fields/AutosuggestField.jsx
jbalboni: I think we already return a string if `enum` is used. I don't mind the data type switching too much, it just has to match the schema. It seems like a win to avoid using this component specific data structure whenever possible. But I don't want to make this more complicated to understand than it already is.

src/applications/common/schemaform/fields/AutosuggestField.jsx
```diff
@@ -56,21 +60,32 @@ export default class AutosuggestField extends React.Component {
...
     }
   }
 
   componentWillUnmount() {
     this.unmounted = true;
   }
 
+  getOptions = (inputValue) => {
+    const getOptions = this.props.uiSchema['ui:options'].getOptions;
+    if (getOptions) {
```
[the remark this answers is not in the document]
cvalarida: Yeah. The only time `getOptions` would be defined is if we were querying the api for a list of options. The alternative is specifying an `enum` in the schema.

src/applications/common/schemaform/fields/AutosuggestField.jsx
```diff
@@ -85,12 +100,23 @@ export default class AutosuggestField extends React.Component {
...
+    }
+
+    return newSuggestion;
   }
 
   handleInputValueChange = (inputValue) => {
     if (inputValue !== this.state.input) {
-      let item = { widget: 'autosuggest', label: inputValue };
+      const uiOptions = this.props.uiSchema['ui:options'];
+      if (uiOptions.queryForResults) {
```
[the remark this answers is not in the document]
cvalarida: Yes. The only time `queryForResults` will be provided / true is if we wanted to query the api for a new list of options every time the field input changes.

src/platform/utilities/data/debounce.js
```diff
@@ -0,0 +1,13 @@
+/**
+ * @param {function}
+ * @param {number}
```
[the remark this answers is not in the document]
cvalarida: Oh woops. Good catch. We should really consider typescript or flow at some point...

src/applications/disability-benefits/526EZ/helpers.jsx
cvalarida: I'm not sure I understand the question. These are just dummy values returned by our fake api call. I was using these for testing; we'll get rid of them when we hook it up to the api for real.

I'm not sure how many options the autosuggest field can handle, if that's your question. I think I saw an option somewhere to limit the suggestions shown.

src/applications/disability-benefits/526EZ/helpers.jsx
cvalarida: I don't think so.

src/platform/utilities/data/debounce.js
cvalarida: I don't think so. I took this pretty much straight from https://medium.com/@griffinmichl/implementing-debounce-in-javascript-eab51a12311e when I realized I was going to have some `this` context issues.

src/applications/disability-benefits/526EZ/helpers.jsx
cehsu: Thanks for explaining.

src/applications/common/schemaform/fields/AutosuggestField.jsx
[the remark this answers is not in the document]
cehsu: Thanks for explaining.

src/applications/common/schemaform/fields/AutosuggestField.jsx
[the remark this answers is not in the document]
cehsu: Thanks for explaining.

src/platform/utilities/data/debounce.js
cehsu: Thanks for the `context`, it's not clear to me what assigning this to context is accomplishing? Could you explain?

src/applications/common/schemaform/fields/AutosuggestField.jsx
cvalarida: Looks like it shouldn't be a problem to return a string instead. I'll give it a shot.

Thanks!

src/applications/common/schemaform/fields/AutosuggestField.jsx
```diff
@@ -47,6 +48,9 @@ export default class AutosuggestField extends React.Component {
       suggestions = this.getSuggestions(options, input);
     }
 
+    const debounceRate = uiOptions.debounceRate === undefined ? 1000 : uiOptions.debounceRate;
```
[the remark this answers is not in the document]
cvalarida: Oh! Duh. I remember why I did this. If we don't pass in a `debounceRate` in the `formConfig`, it'll be `undefined`, and that's when I want it to default to 1000. I _also_ want to be able to pass in 0 so it's essentially _not_ debounced.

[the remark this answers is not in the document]
cvalarida: Good news: When I made the change you proposed, it broke tests that it should have broken. Huzzah for testing! :tada:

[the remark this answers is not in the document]
cehsu: That makes sense, thanks for explaining. When would we want a debounce rate of 0? Is it possible to not debounce in that case?

src/platform/utilities/data/debounce.js
cvalarida: Yeah, absolutely! No guarantees it'll be a _good_ explanation, but here goes:

First, we start off by returning a [closure](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Closures) `debouncedFunction`. This will modify the `timeoutId`. Note the use of the `function` keyword here instead of `=>`. What this means is when an object calls the `debouncedFunction`, the `this` context will be the calling object and _not_ `debounce`. That's what we want.

The trouble comes in when we use `setTimeout`. When `setTimeout` calls the function it's passed,  it's a "free function invocation" (I don't know if that's a standard term, but the article I linked to used it), which means the lexical `this` binding is more or less tossed out the window. The function `setTimeout` calls has `this` attached to nothing.

To overcome that, we use [`function.prototype.apply`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/apply) to give it the `this` we want.

Finally, `setTimeout`'s first parameter takes a function, so we can't just call `setTimeout(func.apply(context, args), wait)`. We need to pass it a function. We use an arrow function here because the `this` binding of the function we pass it doesn't _really_ matter; when it's called, it'll run the function we want with the right context now.

src/applications/disability-benefits/526EZ/helpers.jsx
cehsu: Thanks for explaining, does this mean that if there is no label, it is because there is no treatmentCenterName to being with?

src/platform/utilities/data/debounce.js
cehsu: Thanks @cvalarida, using `apply` and a traditional anonymous function (non-arrow) solves the global scope issue with `setTimeout`, I tried logging `this` in the `setTimeout` call with the workaround (storing `this` in the `context `variable): 
```
timeoutId = setTimeout(() => {console.log('this', context); func.apply(context, args);}, wait);
```
and without:
```
timeoutId = setTimeout(() => {console.log('this', this); func.apply(this, args);}, wait);
```
and it printed `AutosuggestField` regardless (in fact, in this case, the method is already bound before we pass it into debounce, so we could pass `null` as `this` in `apply`, and it still works). 

We may also want to .cancel() the `debouncedGetOptions` , so that it does not attempt to change state after the component is unmounted.

src/applications/common/schemaform/fields/AutosuggestField.jsx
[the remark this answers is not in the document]
cehsu: It looks like [lodash](https://github.com/lodash/lodash/blob/master/debounce.js) checks for zero wait times and handles those in the next tick when there is no leading call, is that the anticipated use case here?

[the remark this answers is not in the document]
cvalarida: I don't know when we'd want a debounce of 0 except perhaps in tests, but it's added complexity in the form implementation to handle the debouncing there when we're probably always going to want it. Also, making an exception in here to check for 0 and call the non-debounced function seems kind of...misleading, I guess.

I think when `setTimeout` has a `0` wait time, it'll fire as soon as possible, which is what we want, methinks. I'm not sure what problem calling it on the next tick would solve.

src/applications/disability-benefits/526EZ/helpers.jsx
cvalarida: Yeah, exactly.

src/platform/utilities/data/debounce.js
cvalarida: Yeah, good point about `.cancel()`. I'll go ahead and add that.

src/applications/disability-benefits/526EZ/helpers.jsx
cehsu: Thanks again, does this throw a "cannot read label property of undefined" if no treatmentCenterName is provided?

cehsu (approved):
LGTM once @jbalboni approves, as he's more familiar with this component.

src/applications/common/schemaform/fields/AutosuggestField.jsx
```diff
@@ -85,11 +101,21 @@ export default class AutosuggestField extends React.Component {
...
+      return suggestion.label;
+    }
+
     return _.set('widget', 'autosuggest', suggestion);
   }
 
   handleInputValueChange = (inputValue) => {
     if (inputValue !== this.state.input) {
+      const uiOptions = this.props.uiSchema['ui:options'];
+      if (uiOptions.queryForResults) {
```
jbalboni: It feels like in this case, if we're calling a server for results, all of the matching logic would be offloaded there, and I'm not sure how much of the rest of the method makes sense in that case.

jbalboni (approved):
Had a question, but looks good overall to me.

src/applications/common/schemaform/fields/AutosuggestField.jsx
```diff
@@ -56,19 +60,31 @@ export default class AutosuggestField extends React.Component {
...
     this.unmounted = true;
+    this.debouncedGetOptions.cancel();
+  }
+
+  getOptions = (inputValue) => {
+    const getOptions = this.props.uiSchema['ui:options'].getOptions;
+    if (getOptions) {
+      // If getOptions returns a promise, when it resolves, set the options to the result.
+      // getOptions may also set the options directly. This is useful if it's a debounced function.
+      const returnedPromise = getOptions(inputValue, this.setOptions);
```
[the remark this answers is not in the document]
cvalarida: Right, exactly. I want to say previously, it was set up to only work for returning a promise, but I found that wasn't always going to work. It's been a while, so I forget the exact reasons, I'm afraid.

:thinking: Actually, now that I think about it, this might be a vestige of debouncing in the `formConfig` instead of in here. Maybe we don't need it now?

src/applications/common/schemaform/fields/AutosuggestField.jsx
cvalarida: This wasn't a deliberate decision when I wrote it, but I think the rest of the method is still useful for while we're waiting for `getOptions` to fire (because debounce). While we're still typing, it'll match (or not) using our current list. When we stop typing, it'll fetch the new list based on that input.

PR #7720, opened 2018-05-11
Add DowntimeNotifications to Letters

ncksllvn:
Related to department-of-veterans-affairs/vets.gov-team/issues/8123

### DIFF
--- a/src/js/letters/containers/LettersApp.jsx
+++ b/src/js/letters/containers/LettersApp.jsx
@@ -45,9 +45,7 @@ export class AppContent extends React.Component {
 
     return (
       <div className="usa-grid">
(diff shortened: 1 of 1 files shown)


kreek (approved):

ncksllvn:
@cvalarida Test are fixed, but is there a better point to add the downtime notification? Most apps do it at the entry point, but we have plenty of [flexibility](https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/js/personalization/dashboard-beta/containers/DashboardApp.jsx#L132).

ncksllvn:
Oh nvm, I understand now. It seems like it would be ideal to have a separate service registered for Letters then?

PR #7727, opened 2018-05-14
Set up errorable checkboxes on review page

cehsu:
These changes use `keepInPageOnReview` to enable the errorable checkbox array widget to be displayed on the review page, as well as update the chapter's review page title.
image


src/applications/common/schemaform/review/ReviewCollapsibleChapter.jsx
```diff
@@ -127,6 +127,9 @@ export default class ReviewCollapsibleChapter extends React.Component {
               pageData = form.data;
               fullPageKey = page.pageKey;
             }
+            if (page.hideOnReview) {
```
cehsu: There is already a hideOnReview property for properties, is it helpful or confusing to use the same name for pages? Would it be helpful to specify that this is for array pages with custom field/widgets by including the word "array"?

src/applications/disability-benefits/526EZ/config/form.js
```diff
@@ -261,11 +261,10 @@ const formConfig = {
         ratedDisabilities: {
           title: 'Your Rated Disabilities',
           path: 'select-disabilities',
+          hideOnReview: true,
           uiSchema: {
             'ui:description': 'Please choose the disability that you’re filing a claim for increase because the condition has gotten worse.',
             disabilities: {
-              // Using StringField because it doesn't do much and we just need to render the widget.
```
cehsu: Issue created to track and discuss this work [here.](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/10449)

peggygannon (commented):
@cehsu Looks fine for content

src/applications/common/schemaform/review/ReviewCollapsibleChapter.jsx
jbalboni: I think having the same name is fine if it's doing the same thing, but I'm having a hard time wrapping my head about what's happening here. If we're just trying to exclude the page from being displayed, would it be clearer to put a filter it out of `expandedPages` before calling `map`?

src/applications/disability-benefits/526EZ/components/SelectArrayItemsWidget.jsx
```diff
@@ -19,10 +19,16 @@ export default class SelectArrayItemsWidget extends React.Component {
...
       options,
-      required
+      required,
+      formContext
     } = this.props;
     const { label: Label, selectedPropName, disabled } = options;
 
+    if (formContext.reviewMode) {
+      return (
+        <ul>{items.filter(i => i['view:selected']).map((d, i) => <li key={i}>{d.name}</li>)}</ul>
```
rianfowler: Easier to read if this uses descriptive names like `item` or `data`. also easier to read if the 
`.dotFunctions()  // because you can comment`
`.areOnDifferentLines() // on each one`

rianfowler:
How does rendering the errorable checkbox array widget on a chapter page differ from the review page that prevents it from being rendered on the review page with the current code?

src/applications/common/schemaform/review/ReviewCollapsibleChapter.jsx
```diff
@@ -127,6 +127,9 @@ export default class ReviewCollapsibleChapter extends React.Component {
               pageData = form.data;
               fullPageKey = page.pageKey;
             }
+            if (page.hasCustomArrayField) {
```
jbalboni: Is this necessary if you use the `keepInPageOnReview` option?

cehsu: Thanks @jbalboni, I had forgotten about that option, which displays the following:
screen shot 2018-05-16 at 3 39 13 pm
image
Is it possible to configure a summary view? It seems like `reviewMode` is not available?
@goldenmeanie is a summary view of interest? Or is simply displaying the form the plan here?

src/applications/disability-benefits/526EZ/components/SelectArrayItemsWidget.jsx
cehsu: Thanks @rianfowler, great points, changed.

cehsu:
@rianfowler in answer to your question, this is how the page currently looks, without these changes (however @jbalboni has recommended the `keepInPageOnReview` configuration option, which seems very promising (screenshots of that have been provided above), pending @goldenmeanie's approval and decision regarding whether there should be some sort of summary view when not in edit mode).

image

Also, while we are considering this, we may want to also think about the evidence pages (since it is not currently possible to edit, remove or add new documents via the review page), a ticket has been set up for that conversation [here](https://github.com/department-of-veterans-affairs/vets.gov-team/blob/3ac9f662f02a5c875b792229be909cd00c584735/Products/Identity/MVI%20Integration/reference_documents/mvi_users_s1a.csv#boards?labels=unicorns,front%20end&repos=31788863).

image

src/applications/disability-benefits/526EZ/config/form.js
```diff
@@ -277,7 +280,8 @@ const formConfig = {
               'ui:options': {
                 showFieldLabel: 'label',
                 label: disabilityOption,
-                widgetClassNames: 'widget-outline'
+                widgetClassNames: 'widget-outline',
+                keepInPageOnReview: true
```
cehsu: @jbalboni `keepInPageOnReview` is now used here, per your helpful recommendation.

src/applications/disability-benefits/526EZ/config/form.js
```diff
@@ -256,16 +256,19 @@ const formConfig = {
       }
     },
     ratedDisabilities: {
-      title: 'Your Rated Disabilities',
+      title: (isReview) => {
+        if (isReview) {
+          return 'Rated Disabilities';
+        }
+        return 'Your Rated Disabilities';
```
cvalarida: Is this backwards? (Not sure when we'd want to say "Your"...)
image

cvalarida (approved):
The simplicity here is excellent!

image
We'll have to make another ticket that labels the conditional array pages based on the disability they're under, but it looks like updating the selection in the review page shows where new information is required, so that's awesome!

cehsu:
@cvalarida good point about the labels, should that be part of the AC in https://github.com/department-of-veterans-affairs/vets.gov-team/issues/10482?

src/applications/disability-benefits/526EZ/config/form.js
cehsu: @peggygannon recommend omitting "Your" on the review page.

cvalarida:
No, I think it's separate. My understanding is that ticket removes the "edit" button from the evidence summary page on the review page. I think we need another ticket to show the condition name for each array page on the review page.

cvalarida:
@jbalboni Seems like we might already have that functionality with how we've used array pages for marriages. Can you remember if we clearly show what information is under which marriage in the review page?

markgreenburg:
This probably needs its own issue, but the Edit button on the review page for this chapter disappears on click. We probably don't need it at all, given that the checkboxes work as expected.

markgreenburg (approved):
LGTM. The 'edit' button gets wonky on the review page for this chapter, but I don't think that has anything to do with these changes.

PR #7731, opened 2018-05-15
Update spouse info pages to use page per item pattern

jbalboni:
This updates the spouse info pages to use the page per item pattern, similar to the pension form.

screen shot 2018-05-15 at 11 32 28 am
screen shot 2018-05-15 at 11 32 47 am


rianfowler (approved):

src/applications/disability-benefits/686/components/SpouseMarriageTitle.jsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import { getSpouseMarriageTitle } from '../helpers';
```
rianfowler: might be cleaner to call this in the config and pass the string into the component.

jbalboni: It needs the index of the current page, so I don't think I can call it in the config.

rianfowler: I understand now. I was reading `SpouseMarriageTitle` and `getSpouseMarriageTitle` as the same thing.

PR #7732, opened 2018-05-15
Update confirmation page on HCA for async processing

jbalboni:
Updates confirmation page copy for HCA async processing. I put the title and email message change behind a prod flag, so that won't go live immediately. The removal of the confirmation number will, since that's currently not useful.

screen shot 2018-05-15 at 4 50 01 pm


peggygannon (commented):
@jbalboni Looks good. Only change is to please add a period after this sentence: 
**Your claim has been submitted.**

I looked, but wasn't exactly sure where in the files to place this period.

rianfowler (approved):

rianfowler:
LGTM except for the broken unit test

PR #7740, opened 2018-05-16
Letters maintenance

cvalarida:
Updated the copy per @peggygannon to reflect that the tool is currently down for maintenance, not that it _will_ be.

image


peggygannon (approved):
@cvalarida Looks good. thanks.

markgreenburg (approved):
LGTM

PR #7742, opened 2018-05-16
Revert "Letters maintenance (#7740)"

cvalarida:
This resets the error messaging for after the maintenance window.


kreek (approved):
LGTM 🚀

PR #7746, opened 2018-05-17
Dashboard: Filter prescriptions widget and show tracking link

ncksllvn:
I don't know if this needs to go into the beta for UAT, but if it does we'll need to merge the two commits from this branch into master. I think it's good to go though.

Resolves department-of-veterans-affairs/vets.gov-team/issues/10291


bshyong (approved):
LGTM

PR #7747, opened 2018-05-18
Fixed infinite loading in RequiredLoginView when logged in status is already confirmed before mount

U-DON:
Resolves department-of-veterans-affairs/vets.gov-team#10465.


markgreenburg (approved):
Pulled the branch down and gave it a spin for good measure; LGTM, thanks for fixing!

PR #7750, opened 2018-05-18
 Fixed broken style for user nav dropdown and sign-in links.

U-DON:
Also fixed the weird gray background behind the user's name when loading the page.


bshyong (approved):
LGTM

PR #7751, opened 2018-05-18
Extract footer content

annekainicUSDS:
Completes https://github.com/usds/us-forms-system/issues/10

The footer content that is rendered on every form is very specific to Vets.gov. Abstract this out so instead `<FormApp>` renders whatever is passed to the form config under the property `footerContent`. This will enable users to create their own custom footer content component to render in their forms. I pulled out what had existed in `<FormApp>` for the footer and created a new component for it under `/platform` that will specific to only Vets.gov. I think added the `footerContent` property to all form configs across Vets.gov.

Also moved around a few other files that didn't belong under the shared schemaform code directory.


src/platform/forms/tests/components/FormFooter.unit.spec.jsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import SkinDeep from 'skin-deep';
```
jbalboni: We've been trying to use Enzyme for new tests. It's not too different from skin deep: https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/applications/post-911-gib-status/tests/containers/IntroPage.unit.spec.jsx

src/applications/edu-benefits/1990/config/form.js
```diff
@@ -102,6 +103,7 @@ const formConfig = {
   },
   title: 'Apply for education benefits',
   subTitle: 'Form 22-1990',
+  footerContent: FormFooter,
   getHelp: GetFormHelp,
```
jbalboni: Seems like we could remove getHelp in these configs?

jbalboni (approved):
Couple comments, but overall LGTM.

src/platform/forms/tests/components/FormFooter.unit.spec.jsx
annekainicUSDS: I will convert these tests to Enzyme, thanks for catching that!

src/applications/edu-benefits/1990/config/form.js
annekainicUSDS: Hm, not sure I totally understand. The new `FormFooter` component checks if anything is specified for `getHelp` and if so renders what has been passed to it. Some forms don't include additional getHelp text in the config, so I think we still need to pass it to the config where we specifically want it.

jbalboni: Oh, right, I see. I think we'll want to change that later; I'd rather just have a different component for forms that use GetHelp vs those that don't. You don't need to change that in here, though, I can make a ticket to refactor that later.

annekainicUSDS: ah gotcha, that makes sense

PR #7753, opened 2018-05-18
Enable prefill message on edu forms

jbalboni:
Looks like this for all the edu forms:

screen shot 2018-05-18 at 10 30 30 am

### DIFF
--- a/src/applications/edu-benefits/1990/config/form.js
+++ b/src/applications/edu-benefits/1990/config/form.js
@@ -89,6 +89,7 @@ const formConfig = {
     notFound: 'Please start over to apply for education benefits.',
     noAuth: 'Please sign in again to resume your application for education benefits.'
   },
+  prefillEnabled: true,
   transformForSubmit: transform,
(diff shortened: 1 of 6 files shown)


rianfowler (approved):

rianfowler:
how does prefill work? is it a standard set of info that we try to retrieve? does it impact the form flow at all?

jbalboni:
Prefill is already on for these forms. It's largely controlled by the backend, where they set up a mapping from the different VA data sources into the right format for a particular form. Prefilled data is returned from the same endpoint as an in progress form, so it looks mostly just like a saved application to the front end. The front end can tell if a form has prefill enabled by looking at a property in the user's profile.

But for logged out users, we can't check a profile, so we don't know if prefill is enabled until you sign in. But we want to show some messaging encouraging users to sign in, so they can take advantage of prefill. So, we need a flag in the config to switch on the extra messaging, which we forgot to switch on when we turned on prefill for the edu forms.

PR #7754, opened 2018-05-18
Add downtime notification to forms

jbalboni:
Given HCA's regular issues and a lack of movement on phase 2 of the downtime notification project, I thought I would do a simple implementation of downtime notifications on SiP forms.

Notices are in two places: the intro page and the review page. The form config has a downtime object which can be used to configure the dependencies and the message displayed.

Customized HCA message:

screen shot 2018-05-22 at 8 01 52 am

Default message (taken from existing downtime message)

screen shot 2018-05-22 at 8 37 17 am


src/applications/common/schemaform/save-in-progress/RoutedSavableReviewPage.jsx
```diff
@@ -103,18 +117,35 @@ class RoutedSavableReviewPage extends React.Component {
...
+        pageList={pageList}
+        path={path}
+        renderErrorMessage={this.renderErrorMessage}/>
+    );
+
+    let submitContent = submitController;
+
+    if (formConfig.downtime) {
+      submitContent = (
+        <DowntimeNotification
```
jbalboni: I didn't think we'd want to put this in SubmitController, since it's very vets specific.

src/applications/common/schemaform/save-in-progress/SaveInProgressIntro.jsx
```diff
@@ -147,6 +163,19 @@ class SaveInProgressIntro extends React.Component {
         <br/>
       </div>
     );
+
+    if (this.props.downtime && !this.props.isLoggedIn) {
```
jbalboni: I'm only blocking starting a form if you're not logged in, since a logged in user could fill in the form and save it, even if they can't submit.

src/applications/common/schemaform/save-in-progress/RoutedSavableReviewPage.jsx
```diff
@@ -103,18 +117,35 @@ class RoutedSavableReviewPage extends React.Component {
...
+          appTitle="application"
+          render={this.renderDowntime}
+          dependencies={formConfig.downtime.dependencies}>
+          {submitController}
```
rianfowler: Why does this wrap the `submitController`?

rianfowler: So it hides everything `submitController` renders?

rianfowler: I think it's easier to read declarative code if the conditionals are grouped and inline. I'm trying to figure out if we can shuffle some stuff around to eliminate the need to put a partial render in a variable.

Fex rather than push a boolean into component manage the condition, just use the condition at this level to control the display. 

The back button adds some complexity to that, though. Could either have `submitController` manage that or graduate it to it's own connected component so that this layout can explicitly control when it's displayed. 

I like the latter idea because then I can see something like `showDowntimeNotification` prop and all of the pieces that are hidden / displayed because of it in one place. It's more work, though.

rianfowler (dismissed):

src/applications/common/schemaform/save-in-progress/RoutedSavableReviewPage.jsx
jbalboni: We could just always render the DowntimeNotification component. I'm swapping it just to avoid another component render for forms that don't need it.

jbalboni: I'm not going to mess with showing the back button. This is meant to be a simple implementation, if we want something more complicated, I think we'll need UX help.

src/applications/common/schemaform/save-in-progress/RoutedSavableReviewPage.jsx
```diff
@@ -103,18 +118,35 @@ class RoutedSavableReviewPage extends React.Component {
...
           formConfig={formConfig}
           formContext={formContext}
           pageList={pageList}
           onSetData={() => this.debouncedAutoSave()}/>
-        <SubmitController
-          formConfig={formConfig}
-          pageList={pageList}
-          path={path}
-          renderErrorMessage={this.renderErrorMessage}/>
+        <DowntimeNotification
```
jbalboni: @rianfowler This is the simpler approach, just with the downside of always rendering DowntimeNotification.

rianfowler (approved):

src/applications/burials/config/form.js
```diff
@@ -89,6 +90,9 @@ const formConfig = {
   formId: '21P-530',
   version: 0,
   prefillEnabled: true,
+  downtime: {
+    dependencies: [services.icmhs]
```
rianfowler: I like how this is configurable.

src/platform/monitoring/DowntimeNotification/index.jsx
```diff
@@ -23,7 +23,8 @@ export const services = {
   mvi: 'mvi',
   mhv: 'mhv',
   tims: 'tims',
-  vic: 'vic'
+  vic: 'vic',
```
rianfowler: maybe add comments with the full names.

src/applications/common/schemaform/save-in-progress/RoutedSavableReviewPage.jsx
```diff
@@ -103,18 +118,25 @@ class RoutedSavableReviewPage extends React.Component {
...
+          appTitle="application"
+          render={this.renderDowntime}
+          dependencies={downtimeDependencies}>
+          <SubmitController
+            formConfig={formConfig}
+            pageList={pageList}
+            path={path}
+            renderErrorMessage={this.renderErrorMessage}/>
+        </DowntimeNotification>
         <SaveStatus
```
ncksllvn: Does wrapping SubmitController disable just the submit button?

src/applications/common/schemaform/save-in-progress/RoutedSavableReviewPage.jsx
```diff
@@ -92,6 +95,18 @@ class RoutedSavableReviewPage extends React.Component {
...
+    if (downtime.status === serviceStatus.down) {
+      const Message = this.props.formConfig.downtime.message || DowntimeMessage;
+
+      return (
+        <Message downtime={downtime}/>
+      );
+    }
+
+    return children;
+  }
```
ncksllvn: We ended up always providing always providing a [downtime](https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/platform/monitoring/DowntimeNotification/containers/DowntimeNotification.jsx#L60) object as that first argument, it's just empty if there isn't any data.

src/platform/monitoring/DowntimeNotification/config/services.js
```diff
@@ -6,11 +6,19 @@ export default {
...
+  // MyHealtheVet
   mhv: 'mhv',
+  // The Image Management System (education forms)
   tims: 'tims',
-  vic: 'vic'
+  // Veteran ID Card v1
+  vic: 'vic',
+  // Intake, conversion, and mail handling services (central mail)
+  icmhs: 'icmhs'
 };
```
ncksllvn: Thanks, these comments were overdue

ncksllvn (approved):
LGTM!

src/applications/common/schemaform/save-in-progress/RoutedSavableReviewPage.jsx
jbalboni: That and a back button, but I'm not too concerned about that since you can edit all of your info on the review page (plus use the browser back button).

PR #7755, opened 2018-05-18
526 file doctype

cehsu:
These changes provider users with a list of document types to select when uploading evidence.
image
image

### DIFF
--- a/src/applications/common/schemaform/widgets/SelectWidget.jsx
+++ b/src/applications/common/schemaform/widgets/SelectWidget.jsx
@@ -1,5 +1,4 @@
 import React from 'react';
-import _ from 'lodash/fp';
 import { asNumber } from '@department-of-veterans-affairs/react-jsonschema-form/lib/utils';
 import onlyUpdateForKeys from 'recompose/onlyUpdateForKeys';
 
--- a/src/applications/common/schemaform/widgets/SelectWidget.jsx
+++ b/src/applications/common/schemaform/widgets/SelectWidget.jsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import _ from 'lodash/fp';
 import { asNumber } from '@department-of-veterans-affairs/react-jsonschema-form/lib/utils';
 import onlyUpdateForKeys from 'recompose/onlyUpdateForKeys';
 
@@ -15,6 +16,38 @@ function getValue(event) {
   return event.target.value;
 }
 
+function getOptions(enumOptions, labels) {
+  return enumOptions.map((option, i) => {
+    return <option key={i} value={option.value}>{labels[option.value] || option.label}</option>;
+  });
+}
+
+function getGroup(group, enumOptions, labels, groups) {
+  const groupedLabels = Object.values(labels).filter(label => groups[label.value] === group.name);
+  const groupedOptions = Object.values(enumOptions).filter(enumOption => groups[enumOption.value] === group.name);
+  const children = getOptions(groupedOptions, groupedLabels);
+
+  return (
+    <optgroup key={group.index} label={group.name}>
+      {children}
+    </optgroup>
+  );
+}
+
+function getOptionList(enumOptions, labels, groups) {
+  const uniqueGroups = Object.values(groups).reduce((acc, item) => {
+    if (acc.indexOf(item) === -1) {
+      acc.push(item);
+    }
+    return acc;
+  }, []);
+
+  if (groups) {
+    return uniqueGroups.map((group, index) => getGroup({ name: group, index }, enumOptions, labels, groups));
+  }
+  return getOptions(enumOptions, labels);
+}
+
 function SelectWidget({
   schema,
   id,
@@ -28,7 +61,7 @@ function SelectWidget({
   onBlur,
   placeholder
 }) {
-  const { enumOptions, labels = {} } = options;
+  const { enumOptions, labels = {}, groups = {} } = options;
   return (
     <select
       id={id}
@@ -48,10 +81,7 @@ function SelectWidget({
         onChange(processValue(schema, newValue));
       }}>
       {!schema.default && <option value="">{placeholder}</option>}
-      {enumOptions.map((option, i) => {
-        return <option key={i} value={option.value}>{labels[option.value] || option.label}</option>;
-      })
-      }</select>
+      {getOptionList(enumOptions, labels, groups)}</select>
   );
 }
 
--- a/src/applications/common/schemaform/widgets/SelectWidget.jsx
+++ b/src/applications/common/schemaform/widgets/SelectWidget.jsx
@@ -41,7 +41,7 @@ function getOptionList(enumOptions, labels, groups) {
     return acc;
   }, []);
 
-  if (groups) {
+  if (Object.values(groups).length) {
     return uniqueGroups.map((group, index) => getGroup({ name: group, index }, enumOptions, labels, groups));
   }
   return getOptions(enumOptions, labels);
--- a/src/applications/disability-benefits/526EZ/config/form.js
+++ b/src/applications/disability-benefits/526EZ/config/form.js
@@ -761,15 +761,15 @@ const formConfig = {
                           17: 'Temporary Total Disability - PTSD',
                           18: 'Temporary Total Disability - PTSD',
                           19: 'Temporary Total Disability - PTSD',
-                          20: 'Other',                          
+                          20: 'Other',
                           21: 'Other',
-                          22: 'Other',                          
+                          22: 'Other',
                           23: 'Other',
-                          24: 'Other',                          
+                          24: 'Other',
                           25: 'Other',
-                          26: 'Other',                          
+                          26: 'Other',
                           27: 'Other',
-                          28: 'Other',                          
+                          28: 'Other',
                           29: 'Other',
                           30: 'Other',
                           31: 'Claim Dependents',
@@ -777,7 +777,7 @@ const formConfig = {
                           33: 'DBQ Scar',
                           34: 'DBQ Scar',
                           35: 'DBQ Vision',
-                          36: 'DBQ Vision' 
+                          36: 'DBQ Vision'
                         }
                       }
                     },
--- a/src/applications/disability-benefits/526EZ/config/form.js
+++ b/src/applications/disability-benefits/526EZ/config/form.js
@@ -737,6 +737,50 @@ const formConfig = {
                         confirmationCode: response.data.attributes.guid
                       };
                     },
+                    attachmentSchema: {
+                      'ui:title': 'Document type',
+                      'ui:options': {
+                        groupBy: true,
+                        groups: {
+                          1: 'Automobile',
+                          2: 'Individual Unemploy',
+                          3: 'Specially Adapted',
+                          4: 'Spouse Aid Attend',
+                          5: 'Special Monthly Comp',
+                          6: 'Special Monthly Comp',
+                          7: 'Temporary Total Disability - PTSD',
+                          8: 'Temporary Total Disability - PTSD',
+                          9: 'Temporary Total Disability - PTSD',
+                          10: 'Temporary Total Disability - PTSD',
+                          11: 'Temporary Total Disability - PTSD',
+                          12: 'Temporary Total Disability - PTSD',
+                          13: 'Temporary Total Disability - PTSD',
+                          14: 'Temporary Total Disability - PTSD',
+                          15: 'Temporary Total Disability - PTSD',
+                          16: 'Temporary Total Disability - PTSD',
+                          17: 'Temporary Total Disability - PTSD',
+                          18: 'Temporary Total Disability - PTSD',
+                          19: 'Temporary Total Disability - PTSD',
+                          20: 'Other',                          
+                          21: 'Other',
+                          22: 'Other',                          
+                          23: 'Other',
+                          24: 'Other',                          
+                          25: 'Other',
+                          26: 'Other',                          
+                          27: 'Other',
+                          28: 'Other',                          
+                          29: 'Other',
+                          30: 'Other',
+                          31: 'Claim Dependents',
+                          32: 'DBQ',
+                          33: 'DBQ Scar',
+                          34: 'DBQ Scar',
+                          35: 'DBQ Vision',
+                          36: 'DBQ Vision' 
+                        }
+                      }
+                    },
                     attachmentName: {
                       'ui:title': 'Document name'
                     }
@@ -758,7 +802,7 @@ const formConfig = {
                       type: 'array',
                       items: {
                         type: 'object',
-                        required: ['name'],
+                        required: ['name', 'attachmentId'],
                         properties: {
                           name: {
                             type: 'string'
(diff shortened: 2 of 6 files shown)


src/applications/disability-benefits/526EZ/helpers.jsx
```diff
@@ -272,6 +272,124 @@ const listifyCenters = (center, idx, list) => {
   );
 };
 
+export const documentTypeEnums = [
+  '1',
```
cehsu: @egarreau it looks like this is how `attachmentId`'s are enumerated in other forms, please let me know if we should approach this differently.

cehsu: It looks like we'll want to use the `L049`-type codes that the backend is already set up for here to simplify the transformation on the backend (so it will simply involve changing attachmentId to attachmentType, plus the existing hashing)?

src/applications/disability-benefits/526EZ/helpers.jsx
```diff
@@ -272,6 +272,124 @@ const listifyCenters = (center, idx, list) => {
...
+  'Buddy/Lay Statement',
+  'VA Form 21-4142 - Authorization To Disclose Information',
+  'VA Form 21-4142a - General Release for Medical Provider Information',
+  'Other Correspondence',
+  'Application to Claim Dependents',
+  'Disability Benefits Questionnaire (DBQ) - Veteran Provided',
+  'Disability Benefits Questionnaire (DBQ) - Veteran Provided',
+  'Photographs',
+  'Disability Benefits Questionnaire (DBQ) - Veteran Provided',
+  'Goldmann Perimetry Chart/Field Of Vision Chart'
```
cehsu: @egarreau it looks like this list has more items than https://github.com/department-of-veterans-affairs/vets-api/blob/master/app/models/evss_claim_document.rb#L23-L45, should some of these be removed?

src/applications/disability-benefits/526EZ/helpers.jsx
```diff
@@ -272,6 +272,124 @@ const listifyCenters = (center, idx, list) => {
...
+  'Medical Evidence from Government Facility',
+  'Social Security Disability Documentation',
+  'Military Personnel Record',
+  'Buddy/Lay Statement',
+  'VA Form 21-4142 - Authorization To Disclose Information',
+  'VA Form 21-4142a - General Release for Medical Provider Information',
+  'Other Correspondence',
+  'Application to Claim Dependents',
+  'Disability Benefits Questionnaire (DBQ) - Veteran Provided',
+  'Disability Benefits Questionnaire (DBQ) - Veteran Provided',
```
cehsu: Lots of duplicates here, sorry, will deduplicate.

src/applications/disability-benefits/526EZ/helpers.jsx
egarreau: no, the back-end list needs to be expanded to include everything from this list https://github.com/department-of-veterans-affairs/vets.gov-team/blob/master/Products/Disability%20Compensation/engineering/EVSS/526DocTypes.xlsx

src/applications/disability-benefits/526EZ/helpers.jsx
```diff
@@ -291,8 +409,9 @@ export const evidenceSummaryView = ({ formData }) => {
         <li>We have received the additional evidence you uploaded:
           <ul>
             {additionalDocuments.map((document, id) => {
+              // attachmentId is 1 indexed
               return (<li className="dashed-bullet" key={id}>
-                <strong>{document.name}</strong>
+                <strong>{`${documentTypes[document.attachmentId - 1]} (${document.name})`}</strong>
```
cehsu: This should simply use the `id` as the key for documentTypes, will update.

src/applications/disability-benefits/526EZ/helpers.jsx
cehsu: Actually, none of these are duplicates. Please disregard previous comment.

src/applications/disability-benefits/526EZ/helpers.jsx
```diff
@@ -272,6 +272,124 @@ const listifyCenters = (center, idx, list) => {
...
+export const documentTypes = [
+  'Application for Automotive Adaptive Equipment',
+  'Application for Compensation for Unemployability',
+  'Application for Specially Adapted Housing',
+  'Examination for Housebound Status',
+  'Request for Nursing Home Information',
+  'Examination for Housebound Status',
+  'Temporary Total Disability for PTSD_21_0781 and PTSD_21_0781A',
+  'Service Connected PTSD Statement',
+  'Service Connected Personal Assault PTSD Statement',
```
cehsu: 'Service Connected PTSD Statement' and 'Service Connected Personal Assault PTSD Statement' are not part of the EVSS attachment list, should they be included here?

src/applications/disability-benefits/526EZ/helpers.jsx
```diff
@@ -272,6 +272,124 @@ const listifyCenters = (center, idx, list) => {
...
+  'Police Report',
+  'Military Performance Report',
+  'ER Medical Report',
+  'Private Medical Treatment Record',
```
cehsu: @egarreau @djmunoz1 it looks like some of these document types such as 'ER Medical Report' and 'Private Medical Treatment Record' actually have the same VBMS eDocument Document Type? since that's the case it seems like we will actually want to keep the current enum scheme so that these are mapped to the appropriate label?

egarreau: We'll need to ask EVSS about this; the backend isn't currently set up to handle two document types with the same VBMS eDocument Document Type. We'll need to gather more information on what the scenario is and how we should handle it.

src/applications/disability-benefits/526EZ/helpers.jsx
egarreau: @djmunoz1 do you have any insight into this? Should we ask EVSS directly?

cehsu:
Hi @peggygannon and @goldenmeanie, could you please share any thoughts on the content/presentation of this select? Grouping options is new for our forms but seems like it makes sense to do here, based on how the information is organized (list available [here](https://github.com/department-of-veterans-affairs/vets.gov-team/blob/master/Products/Disability%20Compensation/engineering/EVSS/526DocTypes.xlsx)).

goldenmeanie:
Are those particular grouping mandated? Or can we adjust?

cehsu:
Good question, @goldenmeanie, perhaps @djmunoz1 might be able to answer that?

peggygannon:
@cehsu @goldenmeanie thanks Claire, a couple of things from a content perspective: 

1. What if the user is uploading a document type not on this list? 

2.  Can we tweak the copy in this list for better readability and grouping? I know there might be some space constraints. 

3. We might want to provide some context to what this drop-down list shows users. For example, something simple like this:
Please choose the type of document you're uploading (*Required)

cehsu:
Thanks @peggygannon, 
1. I think users are only allowed to select from the list, as the document type ID is required for the submission, @egarreau is that correct?
2. As the description text is not used beyond assisting the user, we are probably able to edit to improve readability? @goldenmeanie asked a similar question, perhaps @djmunoz1 can help to answer.
3. Great idea, @peggygannon should that now be the default used for other forms such as preneed (which also features this kind of document type dropdown) as well? Or is this content update specific to this form?

peggygannon:
@cehsu 

Yes, I think we should make this change for all our forms, not just the 526. 
Please choose the type of document you're uploading (*Required)

cvalarida:
@goldenmeanie I think since EVSS only cares about the selection, we can change the grouping and descriptions however we want. (As long as it's not misleading, of course.)

I'm also curious what you think about this list. Is a drop-down the right solution here? I'm not sure what else would solve the problem with minimal complexity...Perhaps that's a thing we can readdress after MVP?

src/applications/disability-benefits/526EZ/helpers.jsx
cvalarida: Currently, we've got duplicate `documentTypes` with different ids. This list on the back end doesn't have duplicates.

What I'd like to see is something more like what the back end's doing:
```js
export const documentTypes = {
  'L029': 'Copy of a DD214',
  'L450': 'STR - Dental - Photocopy',
  'L451': 'STR - Medical - Photocopy',
  'L049': 'Medical Treatment Record - Non-Government Facility',
  'L034': 'Military Personnel Record',
  'L107': 'VA Form 21-4142 - Authorization To Disclose Information',
  'L827': 'VA Form 21-4142a - General Release for Medical Provider Information',
  'L229': 'VA Form 21-0781a - Statement in Support of Claim for PTSD Secondary to Personal Assault',
  'L228': 'VA Form 21-0781 - Statement in Support of Claim for PTSD',
  'L149': 'VA Form 21-8940 - Veterans Application for Increased Compensation Based on Un-employability',
  'L115': 'VA Form 21-4192 - Request for Employment Information in Connection with Claim for Disability',
  'L159': 'VA Form 26-4555 - Application in Acquiring Specially Adapted Housing or Special Home Adaptation Grant',
  'L117': 'VA Form 21-4502 - Application for Automobile or Other Conveyance and Adaptive Equipment Under 38 U.S.C. 3901-3904',
  'L139': 'VA Form 21-686c - Declaration of Status of Dependents',
  'L133': 'VA Form 21-674 - Request for Approval of School Attendance',
  'L102': 'VA Form 21-2680 - Examination for Housebound Status or Permanent Need for Regular Aid & Attendance',
  'L222': 'VA Form 21-0779 - Request for Nursing Home Information in Connection with Claim for Aid & Attendance',
  'L702': 'Disability Benefits Questionnaire (DBQ)',
  'L703': 'Goldmann Perimetry Chart/Field Of Vision Chart',
  'L070': 'Photographs',
  'L023': 'Other Correspondence'
}
```
Then, if we need just the ids, we can do `Object.keys(documentTypes)` and when we want just the values, we can do `Object.values(documentTypes)`.

For the groups, since we no longer have duplicates, we can do something like this:
```js
export const documentTypeGroups = {
  'Automobile': ['L117'],
  'Individual Unemploy': [...],
  'Specially Adapted': [...],
  'Spouse Aid Attend': [...],
  'Special Monthly Comp': [...],
  'Temporary Total Disability - PTSD': [...],
  'Other': [...],
  'Claim Dependents': [...],
  'DBQ': [...],
  'DBQ Scar': [...],
  'DBQ Vision': [...]
}
```
Then what do we do with options that aren't in a group? :thinking: I think I'd like to put them at the top of the list, under no group that way it doesn't _look_ like they're under a group, y'know? That's where it might get a little tricky. To find those options, we should be able to do some fancy footwork like:
```js
/**
 * Loop through all options in groups to find options that don't have a group.
 */
function getUngroupedOptions(options, groups) {
  const ungroupedOptions = options.slice();
  Object.keys(groups).forEach(groupName => groups[groupName].forEach(id => {
    const index = ungroupedOptions.indexOf(id);
    if (index !== -1) {
      ungroupedOptions.splice(index, 1);
    }
  }));
  return ungroupedOptions;
}
```

cvalarida:
Oh, I meant to leave a full-review comment: This is only not approved right now because I think our ids need to match what the back end has. Those blocks of code were just my attempt at making the setup of these groups more ergonomic.

src/applications/disability-benefits/526EZ/helpers.jsx
cehsu: Sorry @cvalarida that comment is actually out dated, we will be sticking with the 1,2,3 enums since there are multiple display names/labels for some document types. Any additional feedback based on that is welcome however, looping in @egarreau as well to confirm.

cehsu:
@goldenmeanie @peggygannon would we like to refine the document groupings as part of this PR? or would it be helpful to set up a separate ticket to discuss that work and approach it as a separate update PR?

src/applications/disability-benefits/526EZ/helpers.jsx
egarreau: I mean using the document types rather than the 1,2,3 enums would be ideal for the backend but since it's not a one-to-one relationship between doc types and display names/labels I understand that's a challenge. If that's not an option, the 1,2,3 enums is a suitable second choice.

cehsu: Yeah, there would need to be a transformation either way, it sounds like the backend may be better suited for that?

cehsu:
@cvalarida good question about the dropdown, looping in @goldenmeanie.

This is something we have [discussed ](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/8901) in the past with other dropdowns with lots of options (such as the VA facilities list).

TLDR: USWDS recommends 7-15 options for dropdowns, if there are more, "Let users type the same information into a text input that suggests possible options instead.". However, in the case of the facilities list, @stacyhlinn favored sticking with dropdowns when confronted with lengthy dropdowns, in particular using a pair of dropdowns to narrow the options by broader categories (such as states), since it was determined that it would be more convenient for users to recognize from a list of options, rather than to remember terms starting from a blank slate.

src/applications/disability-benefits/526EZ/config/form.js
```diff
@@ -667,6 +670,13 @@ const formConfig = {
                         confirmationCode: response.data.attributes.guid
                       };
                     },
+                    attachmentSchema: {
+                      'ui:title': 'Document type',
+                      'ui:options': {
+                        groupBy: true,
```
markgreenburg: How does the `groupBy` ui:option work?

cehsu: Good catch, this is zombie code and should be removed.

markgreenburg:
@cehsu I'm having trouble getting the UI for these changes (i.e., what you've got in your screenshots) to show using the latest from this branch. Is there anything I should be doing to make this work?

markgreenburg:
Well, I was trying to upload pngs, which say they're supported, but actually aren't. These changes seem to work with pdfs. Probably worth opening a separate issue to address that.

cehsu:
@peggygannon it looks like preneed uses: "What kind of document is this?" as the title of the document type dropdown. Should we use that or "Please choose the type of document you’re uploading"?

peggygannon:
@cehsu Yes, let's use the same label that Preneed uses. 

"What kind of document is this?" Thank you for following up on this!

markgreenburg (commented):
I'd like to see if we can strip out the enum mapping and send up the doc codes; left a comment with some ideas in the helpers file.

src/applications/common/schemaform/widgets/SelectWidget.jsx
```diff
@@ -15,6 +15,38 @@ function getValue(event) {
   return event.target.value;
 }
 
+function getOptions(enumOptions, labels) {
```
markgreenburg: The implementation seems specific to our use-case. Should we support optgroups in the common widget at all? If so, is this the right abstraction for doing so? These aren't rhetorical, I'm genuinely curious 😛

cehsu: Good question, @markgreenburg, @goldenmeanie do we know whether we want to use a select with groups for the UI here? multiple dropdowns? an autosuggest field?
If there is interest in using option groups in a dropdown/select, whether that belongs in the common widget is a good question, perhaps @jbalboni can shed some light on that?

markgreenburg: Another question: just because the EVSS spreadsheet categorizes these document types, do we have to? The categories don't really seem to be that helpful. If we just didn't categorize these issues at all (knowing that @goldenmeanie probably has some UX alternatives to using `select` he could cook up post-MVP), that would mean we wouldn't need common forms lib changes, and could strip out a ton of complexity. While the list is long, I don't know that we're helping the veteran by categorizing it. Thoughts? @cvalarida @cehsu

cehsu: Good question @markgreenburg, whether to use categories at all seems like a @peggygannon @goldenmeanie question.

jbalboni: I think making this an autosuggest field would be more useful that the categorization, but if we want to make optgroups a pattern we use on forms, we need to figure out a generic approach that will work for everyone.

markgreenburg: Seems like we should strip out the optgroup / common changes and just go with an uncategorized list for now (we can change to using autosuggest later, if needed). Might be worth opening up a separate issue for discovery on adding optgroups to common forms code, if that's something we want to pursue (which is something we could decide as part of discovery).

goldenmeanie: The groups are pretty exciting to me. This is big list for something less regular than states or other familiar groups. We could potentially do a conditional radio cluster instead, if the infrastructure to support `optgroup` seems fraught.

markgreenburg: Thanks @goldenmeanie, that makes sense. I think what I'm trying to get at is that it'd be easier to separate the implementation of the optgroups / whatever the final UX will be from this PR. It seems to me that this PR's focus is on making sure that we're able to attach the right document type (id?) to each upload. Support for option groups in common forms lib seems tangential to what we're trying to accomplish here, and might require more thought given that we'd want to settle on an approach that works well for all use-cases. Just my 2 cents.

cehsu:
Thanks @markgreenburg for your ideas to improve how the data is structured, one thing to keep in mind is that there are multiple labels for some document types, and labels are occasionally duplicated in different categories.

src/applications/common/schemaform/widgets/SelectWidget.jsx
goldenmeanie: Oh, apologies for muddying the waters there! Take my involvement solely as a vote for the ability to group these options somehow!

markgreenburg:
@cehsu understood. I don't think the fact that doc IDs can be used multiple times changes anything for the example approach I outlined, though. Not that there aren't many other ways to accomplish the same thing. Probably a moot point though, if we decide not to tackle the `optgroup`s in this PR at all.

src/applications/disability-benefits/526EZ/helpers.jsx
cvalarida: I like the general approach @markgreenburg outlines here, but because a single item may be in multiple groups, it might make sense to use `categories: ['PTSD', 'Something else']`. This would mean we couldn't use `_.groupBy()`, but building the list this way would make a lot of sense, and I think that's the part that matters when we're finished.

Provided we have good documentation and can reason about how it all works out when we look at the widget later, of course.

cehsu:
Thanks @kreek @markgreenburg @cvalarida for helping to determine that we will transform the enums to the document type required by the backend on the frontend, so that there isn't any work required to coordinate updates to the enum list (as we may want to change how things are categorized) between the frontend and the backend moving forward, looping @egarreau in for visibility.

cehsu:
@goldenmeanie has indicated that we should avoid using a select here, closing this PR.

PR #7757, opened 2018-05-21
Downtime Notifications refactor

ncksllvn:
Resolves department-of-veterans-affairs/vets.gov-team/issues/10427

Adjustments to the DowntimeNotification component now that we're seeing more common and advanced use cases outside of the standard application wrapper. 

### Changes
- Account for more than one downtime notification rendered on the page at once
    - Mapping the API data is done outside of the React component so that it's only done once rather than per instance of the component
    - Flag put in place to prevent API hit being issued more than once
- Separate out functions for deriving downtime status into utilties
    - In the event your application doesn't want to use the DowntimeNotification component, you should instead be able to leverage the functions in util/helpers.js
- Separate out standard down/downtime-approaching messaging into components
- Update function arguments in components that pass the render property to DowntimeNotification

### Concerns
I was trying to make looking up downtime information for a service as simple as possible, so that in the event an application can't leverage the component and instead wants to go straight to the Redux store, they can -
- Wait for the `state.scheduledDowntime.isReady` flag to flip to true (occurs after the API hit)
- Lookup the downtime information using:
    - `state.scheduledDowntime.serviceMap.get('evss') // => { externalService, status, startTime, endTime }`

I'm not sure if it's good or bad practice to use a Map as part of the Redux store, but I felt like this was a clean way of doing it.


ncksllvn:
cc @BBarone1

src/platform/monitoring/DowntimeNotification/actions/index.js
```diff
@@ -0,0 +1,23 @@
...
+
+    let response;
+    try {
+      response = await apiRequest('/maintenance_windows/');
+    } catch (err) {
+      // Probably in a test environment and the route isn't mocked.
+    } finally {
+      dispatch({
+        type: RECEIVE_SCHEDULED_DOWNTIME,
+        map: createServiceMap(response && response.data)
```
rianfowler: A maintainer is going to look at the reducer of this action to find its behavior. Using a helper is fine but it's probably easier to track down a behavior if this is composed in the reducer. Putting this in the action muddles the purpose of an action vs reducer- what criteria do we use to decide if this should be composed here or in the reducer?

src/platform/monitoring/DowntimeNotification/containers/DowntimeNotification.jsx
```diff
@@ -0,0 +1,89 @@
...
+    children: PropTypes.node,
+    content: PropTypes.node,
+    dependencies: PropTypes.arrayOf(PropTypes.oneOf(Object.values(services))).isRequired,
+    getScheduledDowntime: PropTypes.func.isRequired,
+    isReady: PropTypes.bool,
+    loadingIndicator: PropTypes.node,
+    render: PropTypes.func
+  };
+
+  componentWillMount() {
```
rianfowler: When possible, networks calls that depend on rendering should be in `componentDidMount`. `componentWillMount` will be deprecated.

rianfowler:
@ncksllvn Can you comment in the code the concerns mentioned in the summary? I want to make sure to look at what you're talking about.

src/platform/monitoring/DowntimeNotification/components/DowntimeApproaching.jsx
```diff
@@ -0,0 +1,50 @@
...
+      startTime,
+      endTime,
+      appTitle,
+      children,
+      content
+    } = this.props;
+
+    let downtimeNotification = null;
+    if (!this.state.modalDismissed) {
+      downtimeNotification = (
```
rianfowler: I think declarative code is easier to read if these partials are used inline in the jsx. This could be accomplished with a ternary that returns `null` on `false`.

src/platform/monitoring/DowntimeNotification/actions/index.js
ncksllvn: That's raises an interesting question, because I tend to use actions as the starting point for understanding how data flows throughout the store. Mapping raw API data is simplest to me conceptually within the context of the request issuer, but either way, the work is abstracted into the helper so I can don't really feel strongly.

@jbalboni @cvalarida Do you all have thoughts on @rianfowler's point?

src/platform/monitoring/DowntimeNotification/containers/DowntimeNotification.jsx
ncksllvn: Ah good call. I'll update that.

src/platform/monitoring/DowntimeNotification/components/DowntimeApproaching.jsx
ncksllvn: I sometimes use the ternary operator but unless it's a very short conditional I generally prefer this style.

src/platform/monitoring/DowntimeNotification/actions/index.js
jbalboni: I generally try to put as much logic as possible into reducers, rather than action creators.

src/applications/hca/containers/IntroductionPage.jsx
```diff
@@ -9,8 +9,8 @@ import SaveInProgressIntro, { introActions, introSelector } from '../../common/s
 import DowntimeNotification, { services, serviceStatus } from '../../../platform/monitoring/DowntimeNotification';
 import AlertBox from '@department-of-veterans-affairs/formation/AlertBox';
 
-function renderDowntime(status, downtimeWindow, downtimeMap, children) {
-  if (status === serviceStatus.down) {
+function renderDowntime(downtime, children) {
+  if (downtime.status === serviceStatus.down) {
     const message = (
```
ncksllvn: You can pass a `render` prop to `DowntimeNotification` for more controlled responses to downtime/downtime approaching. The downtime parameter contains the status (either down or downtimeApproaching), externalService, startTime (moment object), and endTime (optional moment object).

src/platform/monitoring/DowntimeNotification/config/services.js
```diff
@@ -0,0 +1,16 @@
...
+  dslogon: 'dslogon',
+  emis: 'emis',
+  es: 'es',
+  evss: 'evss',
+  idme: 'idme',
+  mvi: 'mvi',
+  mhv: 'mhv',
+  tims: 'tims',
+  vic: 'vic'
+};
```
ncksllvn: These are all of the external services with maintenance windows as defined by the API.

src/platform/monitoring/DowntimeNotification/containers/DowntimeNotification.jsx
```diff
@@ -0,0 +1,89 @@
...
+
+    const downtime = getSoonestDowntime(this.props.serviceMap, this.props.dependencies);
+    const children = this.props.children || this.props.content;
+
+    let content = children;
+
+    if (downtime) {
+      if (this.props.render) {
+        content = this.props.render(downtime, children);
+      } else if (downtime.status === serviceStatus.down) {
```
ncksllvn: `getSoonestDowntime` will return the soonest downtime with status down or downtime approaching from the list of service names as defined in `this.props.dependencies`. It will return null if nothing is found and the children components will just render as normal. 

If there is a render prop, would it be preferable to always call it even when there is no downtime? Or just allow the children to pass through as it does now? I was unsure about that.

src/platform/monitoring/DowntimeNotification/util/dismissedDowntimeNotifications.js
```diff
@@ -0,0 +1,22 @@
...
+  },
+  contains(appTitle) {
+    return this._dismissedDowntimeNotifications.find(_appTitle => _appTitle === appTitle);
+  },
+  push(appTitle) {
+    this._dismissedDowntimeNotifications.push(appTitle);
+    window.sessionStorage[this.key] = JSON.stringify(this._dismissedDowntimeNotifications);
+  }
+};
+
```
ncksllvn: I really don't like this code but it was a fine way to abstract away all of the tedious `sessionStorage` stuff.

src/platform/monitoring/DowntimeNotification/components/DowntimeApproaching.jsx
```diff
@@ -0,0 +1,50 @@
...
+    dismissedDowntimeNotifications.setup();
+    this.state = {
+      modalDismissed: dismissedDowntimeNotifications.contains(props.appTitle),
+      cache: {}
+    };
+  }
+
+  dismissModal = () => {
+    dismissedDowntimeNotifications.push(this.props.appTitle);
+    this.setState({ modalDismissed: true });
```
cehsu: Do we need to store this as a state property? Or could we create a modalDismissed variable in render instead? Perhaps that would clutter render.

src/platform/monitoring/DowntimeNotification/util/helpers.js
```diff
@@ -0,0 +1,76 @@
...
+export function getSoonestDowntime(serviceMap, serviceNames) {
+  return serviceNames
+    .map(serviceName => serviceMap.get(serviceName))
+    .filter(service => !!service)
+    .filter(service => service.status !== serviceStatus.ok)
+    .reduce((mostUrgentService, service) => {
+      if (!mostUrgentService) return service;
+      return mostUrgentService.startTime.isBefore(service.startTime) ? mostUrgentService : service;
+    }, null);
+}
```
ncksllvn: This logic is confusing, but I thought it was much cleaner to move this outside of the React component lifecycle.

cehsu (commented):
This looks good so far, I had a few minor questions/comments.

src/platform/monitoring/DowntimeNotification/util/dismissedDowntimeNotifications.js
cehsu: Could using a set for `this._dismissedDowntimeNotifications` be of interest?

src/platform/monitoring/DowntimeNotification/util/helpers.js
```diff
@@ -0,0 +1,76 @@
...
+    serviceMap.set(externalService, {
+      externalService,
+      status,
+      startTime,
+      endTime
+    });
+  }
+
+  return serviceMap;
+}
```
ncksllvn: This serviceMap will become available in the Redux store like -

```
  return {
    evssDowntime: state.scheduledDowntime.isReady && state.serviceMap.get('evss')
  }
}
```
`this.props.evssDowntime` would have the shape -

```
{
  externalService: string // 'evss'
  status: string // down, downtimeApproaching, or ok
  startTime: moment,
  endTime: moment  // may be undefined if indefinite downtime
}
```
It may also be null if there isn't any scheduled downtime.

src/platform/monitoring/DowntimeNotification/actions/index.js
```diff
@@ -0,0 +1,23 @@
+import { apiRequest } from '../../../utilities/api';
+import { createServiceMap } from '../util/helpers';
+
+export const RETREIVE_SCHEDULED_DOWNTIME = 'RETREIVE_SCHEDULED_DOWNTIME';
+export const RECEIVE_SCHEDULED_DOWNTIME = 'RECEIVE_SCHEDULED_DOWNTIME';
+
+export function getScheduledDowntime() {
+  return async (dispatch) => {
+    dispatch({ type: RETREIVE_SCHEDULED_DOWNTIME });
```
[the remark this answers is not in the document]
ncksllvn: That is embarrassing! Good find, thanks

src/platform/monitoring/DowntimeNotification/util/dismissedDowntimeNotifications.js
```diff
@@ -0,0 +1,22 @@
...
+ * that may be experiencing downtime will still have the warning.
+ */
+const dismissedDowntimeNotifications = {
+  key: 'downtime-notifications-dismissed',
+  setup() {
+    const rawData = window.sessionStorage[this.key];
+    this._dismissedDowntimeNotifications = rawData ? JSON.parse(rawData) : [];
+  },
+  contains(appTitle) {
+    return this._dismissedDowntimeNotifications.find(_appTitle => _appTitle === appTitle);
```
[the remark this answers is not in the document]
ncksllvn: Oh yeah I guess I was thinking truthy here versus an actual boolean. I'll update that.

src/platform/monitoring/DowntimeNotification/components/DowntimeApproaching.jsx
ncksllvn: I actually just realized that instead of storing a local variable in render we can just pass the `this.state.modalDismissed` property to the Modal component. That should clean things up some.

I'm not sure how a modalDismissed variable not stored in state though could be used though. Maybe by moving it into the parent scope?

src/platform/monitoring/DowntimeNotification/actions/index.js
ncksllvn: That's interesting, because it seems like we do a lot of property mapping directly after an API call, such as in the user `getProfile` action. I'll revisit this.

src/platform/monitoring/DowntimeNotification/components/DowntimeApproaching.jsx
cehsu: Good point, that's a better idea.

src/platform/monitoring/DowntimeNotification/actions/index.js
jbalboni: That's my opinion, we are not consistent about it at all. Would be good to have some general guidelines for us as a group.

jbalboni: Also, I mean "as possible" as in, within the bounds of Redux's guidelines, which is that reducers should be be pure functions that only depend on the previous state and action object and do not mutate data.

src/platform/monitoring/DowntimeNotification/components/DowntimeApproaching.jsx
jbalboni: Another option is to store the list in Redux, which is what you would do in a stricter Redux app (again, would be good to have some agreed upon guidelines)

src/platform/monitoring/DowntimeNotification/actions/index.js
ncksllvn: I don't think it's a big concern either way in this case since that action is dispatched at most one time and the logic is all contained in a helper, but I'm definitely open to moving to the reducer approach. I'll have to adjust my test cases though so I'll have to circle back around.

src/platform/monitoring/DowntimeNotification/containers/DowntimeNotification.jsx
jbalboni: I think it's a little confusing that it's not always being called. It seems like I need to check the service status if I provide a render prop, but this means there's another condition that I can't control.

src/platform/monitoring/DowntimeNotification/util/dismissedDowntimeNotifications.js
jbalboni: Yeah, I think a Set would make sense here.

src/platform/monitoring/DowntimeNotification/actions/index.js
rianfowler: The other approach I've seen is to categorize this kind of work- mapping one schema to another- as a `parser` and to run API responses through a `parser` before they enter redux if the shape of the data in the response does not match the client state. This refines the purpose of redux as strictly business logic and behavior.

src/platform/monitoring/DowntimeNotification/util/dismissedDowntimeNotifications.js
jbalboni: Also, I think this is pretty clear use case for Redux: this is application state and it's shared among different components.

src/platform/monitoring/DowntimeNotification/util/helpers.js
jbalboni: I agree, I think this makes sense as a separate function. It's also something that might make sense as a selector that's called from `mapStateToProps`.

src/platform/monitoring/DowntimeNotification/actions/index.js
rianfowler: The important thing to me is that the patterns of work are organized consistently. I don't really care where they go. 

We should have a larger discussion about domain language and the kind of composition we want to encourage in PRs. I usually prioritize reducing cognitive load for maintainers above everything and I know that's not the best opinion in all circumstances.

src/platform/monitoring/DowntimeNotification/containers/DowntimeNotification.jsx
ncksllvn: Yeah you're right. I couldn't really think of any use cases for calling the render prop when there isn't any downtime, but it might still be simpler to do that anyway.  Do you think it's okay then for the downtime parameter to be null?

src/platform/monitoring/DowntimeNotification/util/dismissedDowntimeNotifications.js
ncksllvn: I suppose that I could move the handler for closing the DowntimeApproaching modal into an action and have that also drive this data.

src/platform/monitoring/DowntimeNotification/containers/DowntimeNotification.jsx
rianfowler: Would be good to push these conditionals into `mapStateToProps` and simplify the component's props with explicit view behaviors. 

Could destructure `appTitle`, `startTime`, `endTime`, `showChildren` and `downtimeComponentType` from `this.props` and `downtimeComponentType ` could be an enum of something like `CURRENT`, `FUTURE`, and `PROVIDED`. 

There's no reason to implicitly control the showing of children with the downtime type- just make it an explicit prop and pass it from `mapStateToProps`. Then it becomes much clearer exactly what this component is doing just from the props. 

Would be good to flatten the arguments passed to `render` prop- i.e. instead of passing `downtime` which isn't well defined, pass `startTime`and `endTime` explicitly. If an object is preferable, then pass the arguments explicitly e.g. `render({ startTime, endTime }, children)`.

src/platform/monitoring/DowntimeNotification/containers/DowntimeNotification.jsx
```diff
@@ -0,0 +1,89 @@
...
+        );
+      }
+    }
+
+    return content;
+  }
+}
+
+const mapStateToProps = (state) => {
+  const shouldSendRequest = !state.scheduledDowntime.isReady && !state.scheduledDowntime.isPending;
```
rianfowler: Something like `shouldRequestScheduledDowntime` or `downtimeScheduleRequested` would be clearer.

src/platform/monitoring/DowntimeNotification/tests/DowntimeNotification.unit.spec.jsx
```diff
@@ -13,9 +15,9 @@ const endTime = moment().add(6, 'hour').toDate();
 let old;
 
 const innerText = 'This is the inner text';
-function getComponent(dependencies = [], getScheduledDowntime = () => {}) {
+function getComponent(dependencies = [], getScheduledDowntime = () => {}, otherProps = {}) {
   return enzyme.shallow(
-    <DowntimeNotification dependencies={dependencies} scheduledDowntime={null} getScheduledDowntime={getScheduledDowntime}>
+    <DowntimeNotification dependencies={dependencies} getScheduledDowntime={getScheduledDowntime} {...otherProps} shouldSendRequest>
```
rianfowler: Just a personal preference but indented props are easier to read and change. Github cuts this component's props off. 
screen shot 2018-05-21 at 16 52 21

src/platform/monitoring/DowntimeNotification/util/helpers.js
rianfowler: What does `isReady` mean?

src/platform/monitoring/DowntimeNotification/tests/DowntimeNotification.unit.spec.jsx
ncksllvn: I enjoy a nice horizontal scroll ➡️

src/platform/monitoring/DowntimeNotification/containers/DowntimeNotification.jsx
ncksllvn: I can see that but being in the DowntimeNotification component I feel like `shouldSendRequest` is pretty clear.

src/platform/monitoring/DowntimeNotification/util/helpers.js
ncksllvn: It means that the request for scheduled downtime is finished and the component/underlying data is ready for render.

src/platform/monitoring/DowntimeNotification/containers/DowntimeNotification.jsx
ncksllvn: Thanks for the suggestions. I'll see what I can do!

jbalboni: Yeah, that seems fine.

ncksllvn:
@jbalboni @cehsu @rianfowler Thanks for the feedback! Made some updates. PTAL when you get the chance!

jbalboni (approved):
I had a couple more minor comments, but LGTM overall.

src/platform/monitoring/DowntimeNotification/actions/index.js
```diff
@@ -0,0 +1,28 @@
...
+    let data;
+    try {
+      const response = await apiRequest('/maintenance_windows/');
+      data = response.data;
+    } catch (err) {
+      // Probably in a test environment and the route isn't mocked.
+    } finally {
+      dispatch({
+        type: RECEIVE_SCHEDULED_DOWNTIME,
+        data
```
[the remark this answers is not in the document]
ncksllvn: Yeah, there are default parameters that will handle that. Do you think that's too unclear?

src/platform/monitoring/DowntimeNotification/containers/DowntimeNotification.jsx
```diff
@@ -0,0 +1,101 @@
...
+
+    if (this.props.status === serviceStatus.down) {
+      return <Down {...this.props}/>;
+    }
+
+    return children;
+  }
+}
+
+const mapStateToProps = (state, ownProps) => {
```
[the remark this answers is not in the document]
ncksllvn: Yeah good call, will do.

src/platform/monitoring/DowntimeNotification/actions/index.js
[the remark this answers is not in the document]
jbalboni: Nope, that's fine, I just saw it as a change from the last action creator function.

src/platform/monitoring/DowntimeNotification/containers/DowntimeNotification.jsx
```diff
@@ -0,0 +1,101 @@
...
+  render() {
+    if (!this.props.isReady) {
+      return this.props.loadingIndicator || <LoadingIndicator message={`Checking the ${this.props.appTitle} status...`}/>;
+    }
+
+    const children = this.props.children || this.props.content;
+
+    if (this.props.render) {
+      return this.props.render({
+        externalServive: this.props.externalServive,
```
U-DON: Typo: `externalService`.

src/platform/monitoring/DowntimeNotification/containers/DowntimeNotification.jsx
```diff
@@ -0,0 +1,101 @@
...
+  };
+};
+
+const mapDispatchToProps = (dispatch, ownProps) => {
+  return {
+    getScheduledDowntime() {
+      return dispatch(getScheduledDowntime());
+    },
+    dismissDowntimeApproachingModal() {
+      return dispatch(dismissDowntimeApproachingModal(ownProps.appTitle));
```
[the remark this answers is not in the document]
ncksllvn: Yeah I could see that. My thinking here was binding directly to ownProps so that if the action is passed down to the children the parameter will still be there without worrying about the extra prop. But that could also be done just as a method, so either way.

src/platform/monitoring/DowntimeNotification/reducer/index.js
```diff
@@ -0,0 +1,51 @@
...
+
+    case RETRIEVE_SCHEDULED_DOWNTIME:
+      return {
+        ...state,
+        isPending: true
+      };
+
+    case DISMISS_DOWNTIME_APPROACHING_MODAL: {
+      const dismissedDowntimeApproachingModals = state.dismissedDowntimeApproachingModals.concat(action.appTitle);
+      window.sessionStorage[DISMISSED_DOWNTIME_WARNINGS] = JSON.stringify(dismissedDowntimeApproachingModals);
```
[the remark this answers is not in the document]
ncksllvn: Oh gotcha, that makes sense. I'll move that over. Thanks.

src/platform/monitoring/DowntimeNotification/containers/DowntimeNotification.jsx
ncksllvn: Oh no! Thanks 🙂

src/platform/monitoring/DowntimeNotification/reducer/index.js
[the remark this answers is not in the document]
ncksllvn: I've been working on this, and I actually think this is an instance where it is cleanest to have it in the reducer. The initial state needs some way of initializing out of the session and it makes sense to me to have all of the session stuff in one place, especially since it should always be in sync with the state. Does that make sense to you too?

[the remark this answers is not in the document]
jbalboni: I can see why you would want to do that, but it goes against how Redux defines what reducers do (https://redux.js.org/basics/reducers): 

> Given the same arguments, it should calculate the next state and return it. No surprises. No side effects. No API calls. No mutations. Just a calculation.

This is the main expectation of reducers, so I'm really, really hesitant to go against it.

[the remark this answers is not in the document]
rianfowler: Side effects can go in the action creators.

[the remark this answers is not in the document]
ncksllvn: @jbalboni Okay, if you feel strongly that's what I will do. 

Appreciate it @rianfowler but @jbalboni actually explained that in his initial comment on this thread.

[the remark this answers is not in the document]
rianfowler: We could also generalize the behavior into a middleware if we want to update session storage in many actions and keep side effects out of actions.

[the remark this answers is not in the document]
ncksllvn: @rianfowler That's a great idea, I was thinking something similar. I mentioned a little while back that it might be nice to see our session storage behavior centralized or cleaned up so that we know exactly what we are storing there.

bshyong (approved):

PR #7758, opened 2018-05-21
Update FDC Config

markgreenburg:
- Update fdc UI widgets to use view-only properties since we don't want them submitted with the form
- Update FDC yes / no option to use `standardClaim` schema from `vets-json-schema`


cehsu (approved):

markgreenburg:
@cvalarida I think it does. I closed it and added a link back to this PR.

PR #7762, opened 2018-05-22
add pattern error messages

rianfowler:
screen shot 2018-05-22 at 09 01 39
screen shot 2018-05-22 at 09 01 50

### DIFF
--- a/src/applications/hca/config/form.js
+++ b/src/applications/hca/config/form.js
@@ -138,7 +138,7 @@ const formConfig = {
   submitUrl: '/v0/health_care_applications',
   trackingPrefix: 'hca-',
   formId: '1010ez',
-  version: 4,
+  version: 5,
   migrations,
   prefillEnabled: true,
   savedFormMessages: {
--- a/src/applications/hca/config/form.js
+++ b/src/applications/hca/config/form.js
@@ -173,9 +173,16 @@ const formConfig = {
           uiSchema: {
             'ui:description': applicantDescription,
             veteranFullName: _.merge(fullNameUI, {
+              first: {
+                'ui:errorMessages': {
+                  minLength: 'Please provide a valid name. Must be at least 1 character.',
+                  pattern: 'Please provide a valid name. Must be at least 1 character.'
+                }
+              },
               last: {
                 'ui:errorMessages': {
-                  minLength: 'Please provide a valid name. Must be at least 2 characters.'
+                  minLength: 'Please provide a valid name. Must be at least 2 characters.',
+                  pattern: 'Please provide a valid name. Must be at least 2 characters.'
                 }
               }
             }),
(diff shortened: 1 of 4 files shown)


src/applications/hca/config/form.js
```diff
@@ -301,7 +308,18 @@ const formConfig = {
           title: 'Permanent address',
           initialData: {},
           uiSchema: {
-            veteranAddress: addressUI('Permanent address', true)
+            veteranAddress: _.merge(addressUI('Permanent address', true), {
+              street: {
+                'ui:errorMessages': {
+                  pattern: 'Please provide a valid street. Must be at least 1 character.'
```
rianfowler: adds `ui:errorMessages` to `veteranAddress`

jbalboni:
@rianfowler We should probably also write a migration that checks to see if any of the in progress forms has an all spaces name and fixes them (by removing the name and sending the user back to that page). There are already some migrations in this form, so you can look at those as an example, or https://department-of-veterans-affairs.github.io/va-digital-services-platform-docs/docs/vets-developer-docs/vets-website/forms/production.html.

src/applications/hca/config/migrations.js
```diff
@@ -115,5 +115,24 @@ export default [
...
+  ({ formData, metadata }) => {
+    let newFormData = formData;
+    const notBlankStringPattern = /^.*\S.*/;
+
+    ['veteranFullName.first',
+      'veteranFullName.last',
+      'veteranAddress.street',
+      'veteranAddress.city'].forEach(selector => {
+      if (!notBlankStringPattern.test(_.get(selector, newFormData))) {
+        newFormData = _.unset(selector, newFormData);
```
jbalboni: We probably also want to send them back to the page with the error, so it's easy for them to fix, which you can do with the `metadata`.

rianfowler:
@jbalboni I added the returnUrl and migration but I'm not sure how to test this beyond the unit tests.

jbalboni:
This is what I do:

1. Log in and start a form
2. Open the dev tools and click the "Finish this application later" link at the bottom of the form
3. Find the network PUT request and click Copy > Copy as cURL
4. Paste the result in the terminal, tweak the form content to what I want, and run it
5. Load the form again and verify the transform ran correctly.

src/applications/hca/config/migrations.js
```diff
@@ -115,5 +115,28 @@ export default [
...
+
+    [
+      { selector: 'veteranAddress.city', returnUrl: 'veteran-information/veteran-address' },
+      { selector: 'veteranAddress.street', returnUrl: 'veteran-information/veteran-address' },
+      { selector: 'veteranFullName.last', returnUrl: 'veteran-information/personal-information' },
+      { selector: 'veteranFullName.first', returnUrl: 'veteran-information/personal-information' }
+    ].forEach(({ selector, returnUrl }) => {
+      if (!notBlankStringPattern.test(_.get(selector, newFormData))) {
+        newFormData = _.unset(selector, newFormData);
+        newMetaData = _.set('returnUrl', returnUrl, newMetaData);
```
rianfowler: @jbalboni That's great. I tested a few scenarios between the address and name with the migration and it works as expected. This will overwrite anything in the `returnUrl` but the order of the array puts them further back in the form if both the name and address have blank strings.

jbalboni (approved):

PR #7764, opened 2018-05-23
8193 update to downshift

lbai001:
screen shot 2018-05-22 at 9 43 37 pm
screen shot 2018-05-22 at 9 45 54 pm


src/applications/gi/components/search/KeywordSearch.jsx
```diff
@@ -82,22 +65,42 @@ export class KeywordSearch extends React.Component {
...
-          shouldRenderSuggestions={this.shouldRenderSuggestions}
-          suggestions={suggestions}
-          inputProps={{
-            value: searchTerm,
-            onChange: this.handleChange,
-            onKeyUp: this.handleKeyUp,
-            'aria-labelledby': 'institution-search-label',
-            onBlur: this.handleInputBlur,
-          }}/>
+        <Downshift onChange={this.handleChange} onSelect={(item) => this.handleSuggestionSelected(item)}>
```
jbalboni: I don't think you need `onChange` here, I think `onSelect` covers it. Plus, I think this causes some sluggishness with `handleChange` also being used on the input element below.

src/applications/gi/components/search/KeywordSearch.jsx
```diff
@@ -82,22 +65,42 @@ export class KeywordSearch extends React.Component {
...
+                onBlur: this.handleInputBlur
+              })}/>
+              {isOpen && (<div className="suggestions-list">
+                {suggestions.map((item, index) => (
+                  <div
+                    key={index}
+                    className="suggestion"
+                    {...getItemProps({
+                      item: item.label,
+                      style: {
```
jbalboni: Ideally, I think we'd toggle a class on and off here. We use a module called `classnames` to make that easier.

src/applications/gi/components/search/KeywordSearch.jsx
```diff
@@ -82,22 +65,42 @@ export class KeywordSearch extends React.Component {
...
+          }) => (
+            <div>
+              <input {...getInputProps({
+                value: searchTerm || inputValue,
+                onChange: this.handleChange,
+                onKeyUp: this.handleKeyUp,
+                'aria-labelledby': 'institution-search-label',
+                onBlur: this.handleInputBlur
+              })}/>
+              {isOpen && (<div className="suggestions-list">
```
jbalboni: I think this div needs `role="listbox"`

src/applications/gi/components/search/KeywordSearch.jsx
```diff
@@ -82,22 +65,42 @@ export class KeywordSearch extends React.Component {
...
+                {suggestions.map((item, index) => (
+                  <div
```
jbalboni: I think this div needs a role of `option` and probably an `aria-selected` attribute (you can see this in the `AutosuggestField` component in the schemaform folder)

jbalboni (commented):
This is great so far! One other thing I think we could do is in `gi/reducers/autocomplete.js`, the `AUTOCOMPLETE_STARTED` reducer resets the list of suggestions, which makes the dropdown feel janky. We could remove that. The UI of this app is kinda bad in general, but I think that's an improvement.

src/applications/gi/components/search/KeywordSearch.jsx
lbai001: k

src/applications/gi/components/search/KeywordSearch.jsx
lbai001: k

src/applications/gi/components/search/KeywordSearch.jsx
lbai001: k

src/applications/gi/components/search/KeywordSearch.jsx
lbai001: k

src/applications/gi/components/search/KeywordSearch.jsx
```diff
@@ -82,22 +62,38 @@ export class KeywordSearch extends React.Component {
...
-          shouldRenderSuggestions={this.shouldRenderSuggestions}
-          suggestions={suggestions}
-          inputProps={{
-            value: searchTerm,
-            onChange: this.handleChange,
-            onKeyUp: this.handleKeyUp,
-            'aria-labelledby': 'institution-search-label',
-            onBlur: this.handleInputBlur,
-          }}/>
+        <Downshift inputValue={searchTerm} onSelect={(item) => this.handleSuggestionSelected(item)}>
```
jbalboni: Playing around with this and VoiceOver, I think we need to provide a function for the `itemToString` prop, so that it doesn't read "Object object" in some cases. This is what worked for me:

```js
function itemToString(item) {
  if (item && typeof item === 'object') {
    return item.label;
  }
   
  return item;
}
```

jbalboni (approved):
One more thing to add, but LGTM after that. Since Trevor's on vacation, I think we can go ahead get this into master and then have him take a look at it when he gets back. This is a clear improvement over what's there now.

src/applications/gi/components/search/KeywordSearch.jsx
lbai001: kk

PR #7770, opened 2018-05-23
Update documentation for chapter title

cehsu:
These changes update the schemaform documentation to describe that a function that can be used to specify a title, rather than a string, when a distinct review page title is needed.

### DIFF
--- a/docs/schemaform/form-config.md
+++ b/docs/schemaform/form-config.md
@@ -16,8 +16,7 @@ Forms are created by creating a page that uses FormApp from the schemaform folde
   // The prefix for Google Analytics events that are sent for different form actions
   trackingPrefix: '',
 
-  // The title of the form. Displayed on all pages.
-  // Accepts a string or a function that will be passed `onReview` (with a value of true) as a parameter on the review page.
+  // The title of the form. Displayed on all pages
   title: '',
 
   // The subtitle (e.g. form number) of the form. Displayed on all pages, if there's also a title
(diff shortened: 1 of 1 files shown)


jbalboni (approved):

docs/schemaform/form-config.md
```diff
@@ -16,7 +16,8 @@ Forms are created by creating a page that uses FormApp from the schemaform folde
   // The prefix for Google Analytics events that are sent for different form actions
   trackingPrefix: '',
 
-  // The title of the form. Displayed on all pages
+  // The title of the form. Displayed on all pages.
```
markgreenburg: Is this true? I thought `onReview` was only for chapter titles. Maybe you meant to put this there instead of on the form title documentation?

docs/schemaform/form-config.md
```diff
@@ -16,7 +16,8 @@ Forms are created by creating a page that uses FormApp from the schemaform folde
...
+  // Accepts a string or a function that will be passed `onReview` (with a value of true) as a parameter on the review page.
```
markgreenburg: Up to you, but I kinda like the format for the page title - i.e., listing both types of title like:
```
 // The title of the chapter
 title: '',
 // This can also be a function that receives an onReview=true parameter only when on the review page
 title: formData => `A title for ${formData.thing}`,
```

Your call 🤷‍♂️

markgreenburg: On further thought, this is probably an unnecessary / opinionated nit on my part.

markgreenburg (commented):
I think you might've meant to update the documentation for chapter `title` instead of form `title` - happy to approve if I'm wrong or if you can update.

markgreenburg (approved):

PR #7772, opened 2018-05-24
Remove environment variables from being hard-coded in the library code

annekainicUSDS:
Connects to https://github.com/usds/us-forms-system/issues/50

There were a couple of changes needed for this PR:
1. `submitToUrl()` in `actions.js` needs a url to submit the form. This was being included directly in the `actions.js` file by importing the `environment` file. Instead, change the `submitUrl` value in each form's config to being the entire url with the environment variable included as opposed to just the last part of the url (e.g., `submitUrl: '${environment.API_URL}/v0/api'` instead of `submitUrl: '/v0/api'`).
2. `uploadFile()` in `actions.js` also required a url to post the file too, and was being included in the same way as the above by importing the `environment` file directly. Instead, change the `fileUploadUI()` definition that is passed to the config to take a `fileUploadUrl` value. 
NOTE: Some instances of `fileUploadUI()` were being passed an `endpoint` value, which was also included as a default in `fileUploadUI()`. Changed this to not include the url as a default and renamed the value to `fileUploadUrl` to be consistent with the `submitUrl` name and for clarity.

### DIFF
--- a/src/applications/common/schemaform/actions.js
+++ b/src/applications/common/schemaform/actions.js
@@ -3,7 +3,6 @@ import moment from 'moment';
 import recordEvent from '../../../platform/monitoring/record-event';
 import _ from '../../../platform/utilities/data';
 import { transformForSubmit } from './helpers';
-// import environment from '../../../platform/utilities/environment';
 import { timeFromNow } from '../../../platform/utilities/date';
 
 export const SET_EDIT_MODE = 'SET_EDIT_MODE';
--- a/src/applications/common/schemaform/actions.js
+++ b/src/applications/common/schemaform/actions.js
@@ -81,7 +81,7 @@ export function setViewedPages(pageKeys) {
 function submitToUrl(body, submitUrl, trackingPrefix) {
   return new Promise((resolve, reject) => {
     const req = new XMLHttpRequest();
-    req.open('POST', `${submitUrl}`);
+    req.open('POST', submitUrl);
     req.addEventListener('load', () => {
       if (req.status >= 200 && req.status < 300) {
         recordEvent({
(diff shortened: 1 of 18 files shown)


src/applications/common/schemaform/actions.js
```diff
@@ -82,7 +81,7 @@ export function setViewedPages(pageKeys) {
 function submitToUrl(body, submitUrl, trackingPrefix) {
   return new Promise((resolve, reject) => {
     const req = new XMLHttpRequest();
-    req.open('POST', `${environment.API_URL}${submitUrl}`);
+    req.open('POST', `${submitUrl}`);
```
jbalboni: I think this can be `req.open('POST', submitURL);`

src/applications/disability-benefits/526EZ/config/form.js
```diff
@@ -655,7 +656,7 @@ const formConfig = {
                 privateRecords: Object.assign({},
                   fileUploadUI('Upload your private medical records', {
                     itemDescription: 'Adding additional evidence:',
-                    endpoint: '/v0/preneeds/preneed_attachments', // TODO: update this with correct endpoint (e.g. '/v0/21-526EZ/medical_records')
+                    fileUploadUrl: `${environment.API_URL}/v0/preneeds/preneed_attachments`, // TODO: update this with correct endpoint (e.g. '/v0/21-526EZ/medical_records')
```
jbalboni: There's another instance of `endpoint` in this file, looks like.

jbalboni (approved):
Couple minor comments, but looks good to me.

src/applications/common/schemaform/actions.js
annekainicUSDS: ah yes, thanks

src/applications/disability-benefits/526EZ/config/form.js
annekainicUSDS: good catch! will fix that

PR #7773, opened 2018-05-24
Connect 686 submission

jbalboni:
This connects the 686 form to the vets-api endpoint, which currently just validates against the schema.

### DIFF
--- a/src/applications/disability-benefits/686/config/form.js
+++ b/src/applications/disability-benefits/686/config/form.js
@@ -25,7 +25,8 @@ import {
   disableWarning,
   childRelationshipStatusLabels,
   getMarriageTitleWithCurrent,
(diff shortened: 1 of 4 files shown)


src/applications/disability-benefits/686/config/form.js
```diff
@@ -126,12 +128,12 @@ const formConfig = {
           title: 'Veteran Information',
           uiSchema: {
             veteranFullName: fullNameUI,
-            veteranSSN: _.merge(ssnUI, {
+            veteranSocialSecurityNumber: _.merge(ssnUI, {
```
jbalboni: This is the field name in the schema, so we need to use the same one

jholton (approved):
LGTM

PR #7777, opened 2018-05-24
Letters/DowntimeNotifications - Add banner component

ncksllvn:
This adds a DowntimeBanner component that can be used for really simple messaging about services being offline and implements it in Letters.

Resolves department-of-veterans-affairs/vets.gov-team/issues/10601

CC @wyattwalter (if you want to test), @melwoodard (if you want to change messaging)

Screenshot -
image

### DIFF
--- a/src/applications/letters/containers/LettersApp.jsx
+++ b/src/applications/letters/containers/LettersApp.jsx
@@ -3,7 +3,8 @@ import Raven from 'raven-js';
 import { connect } from 'react-redux';
 
 import RequiredLoginView from '../../../platform/user/authorization/components/RequiredLoginView';
-import DowntimeNotification, { services } from '../../../platform/monitoring/DowntimeNotification';
+import { services } from '../../../platform/monitoring/DowntimeNotification';
+import DowntimeBanner from '../../../platform/monitoring/DowntimeNotification/components/Banner';
 
 const UNREGISTERED_ERROR = 'vets_letters_user_unregistered';
 
@@ -58,13 +59,12 @@ export class LettersApp extends React.Component {
         verify
         serviceRequired="evss-claims"
         user={this.props.user}>
-        <DowntimeNotification appTitle="Letters Generator" dependencies={[services.evss]}>
-          <AppContent>
-            <div>
-              {this.props.children}
-            </div>
-          </AppContent>
-        </DowntimeNotification>
+        <AppContent>
+          <DowntimeBanner appTitle="Letters Generator" dependencies={[services.evss]}/>
+          <div>
+            {this.props.children}
+          </div>
+        </AppContent>
       </RequiredLoginView>
     );
   }
--- a/src/platform/monitoring/DowntimeNotification/components/Banner.jsx
+++ b/src/platform/monitoring/DowntimeNotification/components/Banner.jsx
@@ -1,4 +1,6 @@
 import React from 'react';
+import PropTypes from 'prop-types';
+
 import AlertBox from '@department-of-veterans-affairs/formation/AlertBox';
 
 import DowntimeNotification, { serviceStatus } from '../index';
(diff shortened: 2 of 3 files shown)


bshyong (approved):

src/applications/letters/containers/LettersApp.jsx
```diff
@@ -58,13 +59,12 @@ export class LettersApp extends React.Component {
...
         user={this.props.user}>
-        <DowntimeNotification appTitle="Letters Generator" dependencies={[services.evss]}>
-          <AppContent>
-            <div>
-              {this.props.children}
-            </div>
-          </AppContent>
-        </DowntimeNotification>
+        <AppContent>
+          <DowntimeBanner appTitle="Letters Generator" dependencies={[services.evss]}/>
```
cvalarida: Do we know what this `services.evss` is hooked up to? My understanding is EVSS provides a whole gamut of services. This definitely isn't letter-generator-specific, but I'm wondering if there's a case in which this would flag EVSS as being down while the letter generator is actually still up.

cc @kreek @bshyong @ncksllvn @jbalboni because I don't know who would know that answer.

src/platform/monitoring/DowntimeNotification/components/Banner.jsx
```diff
@@ -0,0 +1,38 @@
...
+        return <div/>;
+      }}
+      loadingIndicator={<div/>}
+      dependencies={dependencies}/>
+  );
+}
+
+DowntimeBanner.propTypes = {
+  appTitle: PropTypes.string,
+  dependencies: PropTypes.arrayOf(PropTypes.oneOf(Object.values(services))).isRequired,
```
cvalarida: This probably makes me a little happier than it should. :grinning:

cvalarida (commented):
So I'm a little unclear what problem this is solving. If `services.evss` is down but the letter generator is up, this would work like intended, I think, but if the letter generator is down but the service is up, this doesn't solve that problem.

When I first read the description, I was thinking this would be a way to quickly throw a banner up without going through the services. I'm not sure if that's what we want or not, which is why I'd like a clearer picture on what it is we're trying to solve.

@djmunoz1 Can you provide any clarity there? Looks like you were flagged in the connected ticket, so I'll ask you here. :smiley:

src/applications/letters/containers/LettersApp.jsx
ncksllvn: I don't think it's necessarily tied directly to any EVSS systems - when we become aware of an outage, we schedule it ourselves under that service name.

ncksllvn:
@cvalarida My understanding is that EVSS has occasional nightly deployments and we're on the line to confirm the changes work as intended. During that timeframe, we want to display a banner informing the user that things may not be working as expected. The messaging on the banner is generic based on the system being down and may need to be changed.

src/applications/letters/containers/LettersApp.jsx
jbalboni: I think the concern is that if we use this `evss` service for multiple apps (claim status also uses evss, for example), we might end up putting apps in maintenance that aren't actually down, depending on what's actually happening. I don't know enough about EVSS to say if it normally goes down all at once or if we can have some services up, but others be down.

cvalarida: Ahh, yeah. I share that same concern with @jbalboni. If that's the case, do you know what the level of effort is to make more granular services?

cvalarida:
@ncksllvn That sounds good. As long as it's just a warning ("Heads up: this might not work"), it's probably safe. The concern, however, is like @jbalboni mentioned: If we know the letter generator is being updated but that has no bearing on claims status, we shouldn't mark the `evss` service as down since they're unrelated. I'd like to get more granularity.

For now, this is probably okay to merge and iterate on if we don't misuse it in the meantime.

cvalarida:
Oh, also, congratulations for winning slots! PR #7777.

ncksllvn:
Thank you @cvalarida. I think this is to be used only after the EVSS team is doing a deployment. I don't think it's Letters specifically.

CC @djmunoz1 Planning on merging this tomorrow morning.

PR #7780, opened 2018-05-24
8479-button-link-synthesis

1Copenut:
Codifying findings and guidelines for future design and development of UI for links and buttons. This MD file is the result of accessibility research and peer reviewed use case for `<a>` tags and `<button>` tags in ongoing UI work.

### DIFF
--- a/docs/LinksAndButtons.md
+++ b/docs/LinksAndButtons.md
@@ -65,10 +65,11 @@ It was the consensus of our test users that call to action links can be styled a
     1. SUCCESS CRITERIA:
     2. Respond to ENTER keypress
     3. The browser URL changes
-    4. HTML5 history object changes
-    5. Avoid hash route URLs if possible. These interfere with in-page anchor links, and were common with early SPA libraries like Backbone.
+    4. HTML5 history object changes. (Not applicable if using React Router.)
+    5. Avoid hash route URLs if possible. These interfere with in-page anchor links, and were common with early SPA libraries like Backbone. (Not applicable if using React Router.)
     6. Append anchor links to URL when a link points to an in-page anchor ID (skip nav, tabs)
-    7. Do not add role=“button” to links. Screen readers announce the link as “button”, and changes user   * Recommended to underline links when in paragraphs or blocks of text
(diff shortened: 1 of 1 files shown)


docs/LinksAndButtons.md
```diff
@@ -0,0 +1,97 @@
...
+* Sighted users will interpret “Start your application” and have the heading, sub-heading, and content around the CTA to determine intent
+* Screen reader users will hear “Start your application for health coverage on the next page” read out, and also have context because of the aria-label
+
+## Three rules for link and button success <sup id="heading3">[3]</sup>
+
+1. Links go to to another page OR jump to an anchor ID in the current page
+    1. SUCCESS CRITERIA:
+    2. Respond to ENTER keypress
+    3. The browser URL changes
+    4. HTML5 history object changes
```
jbalboni: 4 and 5 are good general guidance, but React Router handles this correctly for us, so we won't have to do anything special. I wonder if that will confuse someone working on an app here.

docs/LinksAndButtons.md
```diff
@@ -0,0 +1,97 @@
...
+    5. Avoid hash route URLs if possible. These interfere with in-page anchor links, and were common with early SPA libraries like Backbone.
+    6. Append anchor links to URL when a link points to an in-page anchor ID (skip nav, tabs)
+    7. Do not add role=“button” to links. Screen readers announce the link as “button”, and changes user   * Recommended to underline links when in paragraphs or blocks of text
```
jbalboni: Is the asterisk in here intentional?

jbalboni (approved):
LGTM

docs/LinksAndButtons.md
1Copenut: @jbalboni Nope, just missed it in reformatting. I'll fix before merging.

docs/LinksAndButtons.md
1Copenut: @jbalboni I'll reframe the items and call out React Router specifically as solving these. I put them in thinking of the vanilla JS and static site scenarios. Can see how this could add confusion when working with React. Good call.

PR #7782, opened 2018-05-25
Fix autosuggest issues on preneed form

jbalboni:
This fixes some bugs in the autosuggest field for pre-need.

One is that we were duplicating the label on the review page if the field was a string field. That's fixed:

screen shot 2018-05-25 at 11 34 46 am

Also, you couldn't type on fields that used an enum, and there was also an errant accordion plus icon. Fixed now:

screen shot 2018-05-25 at 11 34 55 am

### DIFF
--- a/src/applications/common/schemaform/fields/AutosuggestField.jsx
+++ b/src/applications/common/schemaform/fields/AutosuggestField.jsx
@@ -143,16 +143,16 @@ export default class AutosuggestField extends React.Component {
     if (formContext.reviewMode) {
       const readOnlyData = <span>{getInput(formData, uiSchema, schema)}</span>;
 
-      // If we're using an enum, this is a string field and the label will
+      // If this is an non-object field then the label will
       // be included by ReviewFieldTemplate
-      if (this.useEnum) {
+      if (schema.type !== 'object') {
         return readOnlyData;
       }
 
       return (
         <div className="review-row">
           <dt>{this.props.uiSchema['ui:title']}</dt>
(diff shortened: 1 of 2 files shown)


rianfowler (approved):

src/platform/forms/sass/_m-schemaform.scss
```diff
@@ -553,6 +553,11 @@ legend.schemaform-label.schemaform-file-label {
   position: relative;
 }
 
+.autosuggest-input {
+  // override accordion styles, which are the bane of our existence
```
cvalarida: :joy: :rage1:

src/applications/common/schemaform/fields/AutosuggestField.jsx
```diff
@@ -141,10 +141,18 @@ export default class AutosuggestField extends React.Component {
...
+      // be included by ReviewFieldTemplate
+      if (this.useEnum) {
+        return readOnlyData;
+      }
+
       return (
         <div className="review-row">
           <dt>{this.props.uiSchema['ui:title']}</dt>
-          <dd><span>{getInput(formData, uiSchema, schema)}</span></dd>
+          <dd><span>{readOnlyData}</span></dd>
```
cvalarida: This will render
```jsx
<dd><span><span>{getInput(...)}</span></span></dd>
```
right? We can probably nix the extra `<span>` here.

cvalarida (approved):

PR #7786, opened 2018-05-25
fixed a typo

hamidzr:


jkassemi:
@hamidzr appreciate the fix! We don't run CI on external contributions at this time, so I've opened up a new PR with your change. Thanks, and apologies for the delay.

PR #7790, opened 2018-05-29
Personalization: Add Profile unit tests

ncksllvn:
Resolves department-of-veterans-affairs/vets.gov-team/issues/9629

### DIFF
--- a/src/applications/personalization/va-profile/components/EmailSection.jsx
+++ b/src/applications/personalization/va-profile/components/EmailSection.jsx
@@ -63,14 +63,14 @@ export default function EmailSection({ emailResponseData, field, error, clearErr
   let content = null;
   let modal = null;
 
-  if (emailResponseData) {
+  if (emailResponseData.error) {
+    content = fieldFailureMessage;
+  } else {
     if (emailResponseData.email) {
       content = emailResponseData.email;
     } else {
       content = <button type="button" onClick={onAdd} className="va-button-link va-profile-btn">Please add your email address</button>;
     }
-  } else {
-    content = fieldFailureMessage;
   }
 
   if (isEditing) {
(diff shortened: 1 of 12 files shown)


bshyong (approved):

ncksllvn:
@bshyong I think I'm gonna go ahead and call it a stopping point for the Profile unit tests until we make the switch to Vet360. Do you want to take another look when you get the chance?

src/applications/personalization/va-profile/components/EmailSection.jsx
```diff
@@ -63,14 +63,14 @@ export default function EmailSection({ emailResponseData, field, error, clearErr
   let content = null;
   let modal = null;
 
-  if (emailResponseData) {
+  if (emailResponseData.error) {
+    content = fieldFailureMessage;
+  } else {
```
ncksllvn: This change should have been made already. I'm afraid it might be a bug in prod.

bshyong: What is the bug here? Looks like checking for emailResponseData existence is correct?  Is there a situation where the entire object would be null?

I’m no clear around the motivation behind this change

ncksllvn: It's confusing...

But previously emailResponseData would remain null if the value errored or failed to load. However, we needed a clean method of passing errors on down the line, so I started applying an error property like it is here. So now if it's null, the request for contact information is still pending, and we'll know if it errored by that error property.

`util/sendAndMergeApiRequests.js` might provide some more context.

ncksllvn: The end result of the bug here is just that the user might see a link about adding their email address when they should instead see an error message.

PR #7791, opened 2018-05-29
Personalization: Enable in prod

ncksllvn:
Resolves department-of-veterans-affairs/vets.gov-team/issues/10657


bshyong (approved):

src/platform/testing/e2e/auth.js
```diff
@@ -90,6 +90,7 @@ function logIn(token, client, url, level) {
     .url(`${E2eHelpers.baseUrl}${url}`)
     .waitForElementVisible('body', Timeouts.normal);
 
+  E2eHelpers.disableAnnouncements(client);
   setUserToken(token, client);
```
ncksllvn: This has to be here so that the change to local storage is applied to the right domain

PR #7792, opened 2018-05-29
copy updates

rianfowler:
Removed several of the conditionals that were switching on `production`. 
screen shot 2018-05-29 at 11 14 56

### DIFF
--- a/src/applications/hca/containers/ConfirmationPage.jsx
+++ b/src/applications/hca/containers/ConfirmationPage.jsx
@@ -29,8 +29,7 @@ export class ConfirmationPage extends React.Component {
     let dateTitle = 'Date received';
     let emailMessage;
 
-    // if (__BUILDTYPE__ !== 'production') {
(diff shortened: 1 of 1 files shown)


peggygannon (approved):
@rianfowler looks good. thanks!

jbalboni:
@rianfowler I don't think we want to remove the conditionals yet. If we do, we can't merge this until the HCA changes go live.

rianfowler:
@jbalboni so this is live and we're adding an email notification. 
screen shot 2018-05-30 at 09 14 02

jbalboni:
Correct, the HCA form is live, but we're making the process async and adding a failure email, which entails some copy changes here, which you're getting the second round of. Unless something changed yesterday that I'm not aware of.

rianfowler:
Nope- I need to ask for more context in the future before making logic changes.

jbalboni (approved):

PR #7793, opened 2018-05-29
10326 update error messages

rianfowler:
Need to test against backend updates.


src/applications/veteran-id-card/config.jsx
```diff
@@ -0,0 +1,25 @@
...
+    VIC010: "We're sorry. We can't proceed with your request for a Veteran ID card because we can't confirm your military history right now. Please try again in a few minutes. If it still doesn't work, please call the Vets.gov Help Desk at 1-855-574-7286, TTY: 1-800-877-8339, Monday &#8211; Friday, 8:00 a.m. &#8211; 8:00 p.m. (ET).",
+    VIC011: "We're sorry. We can't proceed with your request for a Veteran ID card because we can't confirm your military history right now. Please try again in a few minutes. If it still doesn't work, please call the Vets.gov Help Desk at 1-855-574-7286, TTY: 1-800-877-8339, Monday &#8211; Friday, 8:00 a.m. &#8211; 8:00 p.m. (ET).",
+    'default': "We're sorry, but something went wrong. Please try again in a few moments.",
+    VICV5: "We're sorry. Something went wrong on our end. We can't process your request right now. Please go to [AccessVA] to request a Veteran ID Card.",
+    VICV2: (<p>We’re sorry. We can’t confirm your eligibility for a Veteran ID card right now. To qualify for a Veteran ID Card, both of these must be true:
+      <ul>
```
rianfowler: This content update required an unordered list. I (think) just changing this file to `.jsx` and importing `React` is all I need in order for this to support jsx partials.

src/applications/veteran-id-card/config.jsx
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
```
rianfowler: This content update required an unordered list. I think just changing this file to .jsx and importing React is all I need in order for this to support jsx partials.

src/applications/veteran-id-card/config.jsx
```diff
@@ -0,0 +1,25 @@
...
+
+module.exports = {
+
+  messages: {
+    VIC002: (<p>We're sorry. We're having trouble finding your records in our system right now. Please go to <a href="https://pint.access.va.gov/accessva/">AccessVA</a> to request a Veteran ID Card.</p>),
```
rianfowler: Not sure if this is the link we want to use for AccessVA.

jbalboni: I _think_ `pint` is a test environment and we don't need that part of it.

rianfowler: They need to update their `robots.txt` 
screen shot 2018-05-31 at 08 44 03

rianfowler:
@jbalboni I went to test this but it looks like `VeteranIDCard` component was used instead of `Main` and I don't know what this configuration means. 

Both components have similar error handling but one looks at the user props and the other at the error code returned from the API. The app is displaying `VeteranIDCard` component. 

screen shot 2018-06-05 at 16 59 59

rianfowler:
Should note that I'm testing locally with the VIC022 beta mocks user: https://app.zenhub.com/workspace/o/department-of-veterans-affairs/vets.gov-team/issues/10705

The current flow doesn't even hit `id_card/attributes` according to my network timeline. Not sure but there look like two completely different veteran ID card apps in this directory- one that uses Main and one that doesn't.

jbalboni:
@rianfowler That structure in the routes file means that VeteranIDCard will wrap all the route components, and that Main is used for the `/` (index) route. There's only one route, so honestly this app doesn't even need React Router. The component structure should look like:

```js
<VeteranIDCard>
  <Main/>
</VeteranIDCard>
```

rianfowler:
@jbalboni 
Hitting [this](https://github.com/department-of-veterans-affairs/vets-website/blob/9af841fe3d7120d6ec484b3db9fd0b829d5c71c5/src/applications/veteran-id-card/components/RequiredVeteranView.jsx#L11) error when I test beta mocks user ***vets.gov.user+0@gmail.com*** locally. 

`if (this.props.userProfile.veteranStatus === 'SERVER_ERROR') {`
screen shot 2018-06-06 at 15 28 17

The state suggests that I'm authenticated but `veteranStatus` is `SERVER_ERROR` which comes from a user profile [action](https://github.com/department-of-veterans-affairs/vets-website/blob/9af841fe3d7120d6ec484b3db9fd0b829d5c71c5/src/platform/user/profile/actions/index.js#L72) fetching **/v0/user**.

I tried this on the review instance and got the same result. Is this an environment issue and what does `veteranStatus` represent? The client also handles `veteranStatus` `NOT_FOUND` and `OK` in `RequiredVeteranView` which may overlap some of the error codes we're trying to handle with the update to **id_card/attributes**.

src/applications/veteran-id-card/veteran-id-card-entry.jsx
```diff
@@ -1,12 +1,16 @@
...
+import VeteranIDCard from './containers/VeteranIDCard';
+import Main from './containers/Main';
 
-import routes from './routes';
 import reducer from './reducers';
 
 startApp({
   reducer,
-  routes
+  component: (<VeteranIDCard>
```
rianfowler: @jbalboni not sure how I feel about this- I think I could safely just render `Main` in `VeteranIDCard`- doesn't look like a reusable component.

rianfowler:
Was able to test the updates to `vets-api` by circumventing the error checking in `RequiredVeteranView` for now.

rianfowler:
When bypassing the `RequiredVeteranView` I was able to [validate](https://app.zenhub.com/workspace/o/department-of-veterans-affairs/vets.gov-team/issues/10326#issuecomment-395247368) all but one of the use cases. Working on getting the last one resolved.

The next step is to remove the bypass and log in with each of these users to see how `RequiredVeteranView` behaves. I would like to use ***id_card/attributes*** as the client's source of truth on this and stop using the state we get from ***/v0/user*** but this has UX implications. 

The app retrieves state from ***/v0/user*** before the user clicks a button and currently the shows them an error before any content is displayed (it's just the error with no context around it displayed after they log in and navigate to `veteran-id-card`).

Errors that come from our updates that are handled by `containers/Main.jsx` are displayed with some context:
download

rianfowler:
@jbalboni @lihanli Testing this locally and the client is receiving a list of [`services`](https://github.com/department-of-veterans-affairs/vets-website/blob/01e4dbcfcdd0d403adf0ff3e71a113b3a9d026e4/src/platform/user/profile/actions/index.js#L72) from `/v0/user` that doesn't include `id-card`. 

How is this list managed and when would `id-card` be included in it? This is causing most of the new betamocks user to hit an error condition in the client app.

---
From the same endpoint the client is getting `veteranStatus` is `NOT_FOUND` for betamocks user `vets.gov.user+0@gmail.com` which is causing the client to hit an error condition. [Comment](https://github.com/department-of-veterans-affairs/vets-website/blob/9af841fe3d7120d6ec484b3db9fd0b829d5c71c5/src/applications/veteran-id-card/components/RequiredVeteranView.jsx#L15-L16) on this condition says: 

>       // If eMIS status is "not found", show message that we cannot find the user
>      // in our system.

Can we verify that this state from `/v0/user` matches the state provided by `id_card/attributes` which represents:

> **Veteran does not have an EDIPI or has an EDIPI but no military history associated with their profile**

I would rather use `id_card/attributes` as the source of truth but that would cause some UX changes.

rianfowler:
Spoke with @lalexanderson-dsva and decided that the error messages should show up after the user has seen the "Printed Veteran ID Card" page.

src/applications/veteran-id-card/components/RequiredVeteranView.jsx
```diff
@@ -6,23 +6,12 @@ import SystemDownView from '@department-of-veterans-affairs/formation/SystemDown
 class RequiredVeteranView extends React.Component {
   render() {
     let view;
-    const serviceAvailable = this.props.userProfile.services.indexOf('id-card') !== -1;
 
     if (this.props.userProfile.veteranStatus === 'SERVER_ERROR') {
       // If eMIS status is null, show a system down message.
       view = <SystemDownView messageLine1="We’re sorry. We can’t process your request for a Veteran ID Card right now because we can't access your records at the moment. Please try again in a few minutes." messageLine2="If it still doesn’t work, please call the Vets.gov Help Desk at 1-855-574-7286, TTY: 1-800-877-8339. We’re here Monday &#8211; Friday, 8:00 a.m. &#8211; 8:00 p.m. (ET)."/>;
-    } else if (this.props.userProfile.veteranStatus === 'NOT_FOUND') {
```
rianfowler: This bypasses the error handling that relied on the user state.

rianfowler:
@jbalboni This is ready for review. I validated these changes again locally: https://app.zenhub.com/workspace/o/department-of-veterans-affairs/vets.gov-team/issues/10326#issuecomment-395247368

src/applications/veteran-id-card/config.jsx
```diff
@@ -0,0 +1,26 @@
...
+    VIC010: "We're sorry. We can't proceed with your request for a Veteran ID card because we can't confirm your military history right now. Please try again in a few minutes. If it still doesn't work, please call the Vets.gov Help Desk at 1-855-574-7286, TTY: 1-800-877-8339, Monday &#8211; Friday, 8:00 a.m. &#8211; 8:00 p.m. (ET).",
+    VIC011: "We're sorry. We can't proceed with your request for a Veteran ID card because we can't confirm your military history right now. Please try again in a few minutes. If it still doesn't work, please call the Vets.gov Help Desk at 1-855-574-7286, TTY: 1-800-877-8339, Monday &#8211; Friday, 8:00 a.m. &#8211; 8:00 p.m. (ET).",
+    'default': "We're sorry, but something went wrong. Please try again in a few moments.",
+    VICV5: (<p>We're sorry. Something went wrong on our end. We can't process your request right now. Please go to <a href="https://access.va.gov/accessva/">AccessVA</a> to request a Veteran ID Card.</p>),
```
rianfowler: `VICV5, VICV2, VICV4` are new error codes that correspond with the title 38 status from eMIS.

src/applications/veteran-id-card/routes.jsx
```diff
@@ -1,11 +0,0 @@
-import VeteranIDCard from './containers/VeteranIDCard';
```
rianfowler: Ditched the routes file.

src/applications/veteran-id-card/containers/Main.jsx
```diff
@@ -71,12 +71,12 @@ class Main extends React.Component {
...
     const { errors } = this.props;
     const code = head(errors).code;
-    const detail = has(messages, code) ? messages[code] : messages.default;
+    const detail = has(config.messages, code) ? config.messages[code] : config.messages.default;
     const content = (
       <div>
         <h4>We can't process your request</h4>
         <div>
-          <p>{detail}</p>
+          {detail}
```
rianfowler: React was complaining about nesting `<p>`

src/applications/veteran-id-card/components/RequiredVeteranView.jsx
```diff
@@ -11,6 +11,8 @@ class RequiredVeteranView extends React.Component {
     if (this.props.userProfile.veteranStatus === 'SERVER_ERROR') {
       // If eMIS status is null, show a system down message.
       view = <SystemDownView messageLine1="We’re sorry. We can’t process your request for a Veteran ID Card right now because we can't access your records at the moment. Please try again in a few minutes." messageLine2="If it still doesn’t work, please call the Vets.gov Help Desk at 1-855-574-7286, TTY: 1-800-877-8339. We’re here Monday &#8211; Friday, 8:00 a.m. &#8211; 8:00 p.m. (ET)."/>;
+    } else if (__BUILDTYPE__ !== 'production') {
```
rianfowler: This bypasses the error messages that are based on the user state.

src/applications/veteran-id-card/config.jsx
```diff
@@ -0,0 +1,30 @@
...
+      <ul>
+        <li>You served on active duty, in the military reserves, or in the National Guard, and </li>
+        <li>You separated under honorable conditions or general (under honorable conditions)</li>
+      </ul>
+    </p>)
+  }
+};
+
+if (__BUILDTYPE__ === 'production') {
+  config.messages.VIC002 = "We can't proceed with your request for a Veteran ID card because we can't find your DoD ID number in our system. Please call the Vets.gov Help Desk at 1-855-574-7286, TTY: 1-800-877-8339, Monday &#8211; Friday, 8:00 a.m. &#8211; 8:00 p.m. (ET).";
```
rianfowler: This is the only content update to existing error messages. The new error states in this config won't be reachable in production.

rianfowler:
@jbalboni I added checks that preserve the production behavior. I left the structural changes i.e. 
- removing the routes file
- updating the `config.js` to `config.jsx`
- removing the `<p>` from around the partial render

jbalboni (approved):
LGTM

PR #7794, opened 2018-05-29
526 require all selected mg

markgreenburg:
Adds schema requirements to ensure that the user can't submit a form without any supporting evidence. Except for PMR -> My Doctor has my records. The intent with that flow in particular is to let the user not upload anything - they will submit a 4142 on their own (for the time being).

### DIFF
--- a/src/applications/disability-benefits/526EZ/config/form.js
+++ b/src/applications/disability-benefits/526EZ/config/form.js
@@ -545,6 +545,7 @@ const formConfig = {
                 type: 'array',
                 items: {
                   type: 'object',
+                  required: ['view:uploadPrivateRecords'],
                   properties: {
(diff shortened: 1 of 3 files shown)


src/applications/disability-benefits/526EZ/config/form.js
```diff
@@ -545,6 +545,7 @@ const formConfig = {
                 type: 'array',
                 items: {
                   type: 'object',
+                  required: ['view:uploadPrivateRecords'],
```
markgreenburg: Requires selecting either 'yes' or 'no' to the question 'Do you want to upload your Private Medical Records' - assuming they chose PMR on the previous page as a type of supporting evidence they wanted to provide.

cvalarida (approved):

PR #7797, opened 2018-05-30
526 remove vcl mg

markgreenburg:
- Turns homelessness into a yes / no question (not required)
- Removes other questions that we can't support for MVP due to API limitations.

screen shot 2018-05-30 at 6 17 19 pm

### DIFF
--- a/src/applications/disability-benefits/526EZ/config/form.js
+++ b/src/applications/disability-benefits/526EZ/config/form.js
@@ -233,12 +233,6 @@ const formConfig = {
             'view:homelessness': {
               'ui:title': 'Are you homeless or in danger of becoming homeless?',
               'ui:widget': 'yesNo',
-              'ui:options': {
-                labels: {
-                  y: 'Yes',
-                  n: 'No'
-                }
-              }
             },
           },
           schema: {
--- a/src/applications/disability-benefits/526EZ/config/form.js
+++ b/src/applications/disability-benefits/526EZ/config/form.js
@@ -231,7 +231,7 @@ const formConfig = {
           uiSchema: {
             'ui:title': 'Homelessness',
             'view:homelessness': {
-              'ui:title': 'Are you homeless or in danger of becoming homeless?',
+              'ui:title': 'Are you homeless or at risk of becoming homeless?',
               'ui:widget': 'yesNo',
             },
           },
--- a/src/applications/disability-benefits/526EZ/config/form.js
+++ b/src/applications/disability-benefits/526EZ/config/form.js
@@ -61,7 +61,6 @@ import {
   // releaseView, // Where was this used before?
   disabilityOption,
   GetFormHelp,
-  specialCircumstancesDescription,
   FDCDescription,
   FDCWarning,
   noFDCWarning,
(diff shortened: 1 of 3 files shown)


src/applications/disability-benefits/526EZ/config/form.js
```diff
@@ -230,35 +229,24 @@ const formConfig = {
...
-            'view:homeless': {
-              'ui:title': 'Homeless or at risk of becoming homeless?'
-            },
-            'view:extremeFinancialHardship': {
-              'ui:title': 'Suffering from extreme financial hardship?'
+            'ui:title': 'Homelessness',
+            'view:homelessness': {
+              'ui:title': 'Are you homeless or in danger of becoming homeless?',
+              'ui:widget': 'yesNo',
+              'ui:options': {
```
cehsu: Interesting, are these not default on the `yesNo` widget?

cehsu (approved):
LGTM.

src/applications/disability-benefits/526EZ/config/form.js
markgreenburg: Good question, I didn't look at the widget internals to find out. Probably worth double-checking, though.

markgreenburg: Yep, those are definitely the defaults 😛 
Removed the whole deal.

cehsu: Thanks for checking 👍

peggygannon (changes requested):
@markgreenburg Hi, Mark, can we please update the homeless question to the following to match how we treat this question on the static page:
Are you homeless or at risk of becoming homeless? 
Yes
No

And I have a few followup questions:
1. what happens if the user chooses yes? does another screen pop up? 
2. Is this question on its own screen or with other info? If it's on its own screen, I'm wondering if we should have a line of text before the question. 
3.  Are the financial hardship and blind/low vision questions moving to another screen? or are they being taken out of the MVP altogether?

markgreenburg:
@peggygannon updated content and screenshot per your recommendation.

> what happens if the user chooses yes? does another screen pop up?

@goldenmeanie 's proposed UI was to have the follow-up content info expand under these radio buttons if the veteran chooses 'yes.' This work has been offloaded to another issue because, even though simple to implement, it touches a lot of different parts of form code.

> Is this question on its own screen or with other info? If it's on its own screen, I'm wondering if we should have a line of text before the question.

Alex's proposal was to have homelessness on its own form page - with the additional name / contact info questions also on the same page, but only showing if the veteran selects 'yes.'

> Are the financial hardship and blind/low vision questions moving to another screen? or are they being taken out of the MVP altogether?

If I recall correctly, they're being taken out altogether, because there's no way to submit this info, based on the API that EVSS is providing. The only data we're able to send them is point of contact name & phone number for homeless veterans.

peggygannon (approved):

markgreenburg:
Sure thing, @peggygannon ! Once we have the other PR with the follow-up contact info fields ready to go, we'll be sure to share screenshots.

PR #7799, opened 2018-05-30
Test US Forms System integration

annekainicUSDS:
A few things were changed in order to test the integration:

1. Imported the library code
2. Changed all of the save-in-progress files to use the imported us-forms-system code instead of the vets-website version
3. Converted all references to the vets-website library code from within HCA and burials apps to instead reference the us-forms-system code

### DIFF
--- a/package.json
+++ b/package.json
@@ -228,6 +228,7 @@
     "text-loader": "^0.0.1",
     "url-search-params": "^0.10.0",
     "vets-json-schema": "https://github.com/department-of-veterans-affairs/vets-json-schema.git#86263a396bfe2b13153797bac78c424a913b450d",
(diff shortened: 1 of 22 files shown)


annekainicUSDS:
There's currently an issue running the unit tests. I'm getting an error about `import` statements: `SyntaxError: Unexpected token import`. This seems to be traced to the us-forms-system transpiled code. Not sure why the transpilation isn't converting `import` to `require`; also seems like this is an issue uniquely experienced by unit tests rather than application form code when running Vets.gov locally, so not sure how unit tests plays into this yet. Still debugging.

annekainicUSDS:
Closing in favor of a new approach, will open new PR

PR #7800, opened 2018-05-31
526 upload types mg

markgreenburg:
For PMR Evidence and Lay Evidence:
- Updates accepted file types
- Updates endpoints to actual one for 526 documents
- Updates response parsing to get the correct `GUID` for each file

Also resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/10681


cehsu (approved):
LGTM. Should any tests be written/updated?

markgreenburg:
@SMLuthi to confirm if `vets-api` response shape needs to change to let us parse the upload response with the same pattern that we use for other forms. I.e., `response.guid` might actually need to change to `response.data.attributes.guid`

PR #7801, opened 2018-05-31
10607 name not required

lbai001:


src/applications/pre-need/config/form.jsx
```diff
@@ -68,6 +68,12 @@ const {
 
 const nonRequiredFullName = _.omit('required', fullName);
 
+function currentlyBuriedPersonsMinItem() {
```
jbalboni: You could also do this with one `_.merge` statement, but this function is correct and I don't have a strong preference either way.

jbalboni (approved):
LGTM

PR #7802, opened 2018-05-31
Update Nightwatch's Babel

ncksllvn:
I found this today while trying to use async/await in an e2e test.

### DIFF
--- a/config/nightwatch.js
+++ b/config/nightwatch.js
@@ -4,7 +4,8 @@
 const chromedriver = require('chromedriver');
 const seleniumServer = require('selenium-server');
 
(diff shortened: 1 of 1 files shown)


config/nightwatch.js
```diff
@@ -4,7 +4,8 @@
 const chromedriver = require('chromedriver');
 const seleniumServer = require('selenium-server');
 
-require('babel-core/register');
+require('babel-register');
+require('babel-polyfill');
```
ncksllvn: I switched to babel-register for consistency with the mocha.opts.

jbalboni (approved):

PR #7806, opened 2018-06-01
AsyncDisplayWidget

cvalarida:
### Loading
image
Still need to get a copy edit on this. Shown is the default message, but it's customizable.
@peggygannon Can I get both a default message and what we should show on this page (if they're different)? The default message will be used whenever we use the `AsyncDisplayWidget` to fetch data (presumably about the veteran) in a form when they can't edit it themselves in the form but that information may be changed between sessions.

### Success
image
It doesn't look like much yet, but that'll change when we use it "for real" in https://github.com/department-of-veterans-affairs/vets.gov-team/issues/10440.

### Failure
image
So....yeah...there's a reason I'm not a copy writer.
@peggygannon Same thing as the loading screen. Can I get content for both the default and page-specific alert?

### DIFF
--- a/src/applications/disability-benefits/526EZ/components/AsyncDisplayWidget.jsx
+++ b/src/applications/disability-benefits/526EZ/components/AsyncDisplayWidget.jsx
@@ -86,7 +86,7 @@ export default class AsyncDisplayWidget extends React.Component {
             isVisible
             headline="We're sorry, we can't seem to find your information right now."
             content="That's a real bummer, we know. Maybe try again later."
-            className="fetch-fail-alert"/>
+            className="async-display-widget-alert-box"/>
         );
         break;
       }
--- a/src/applications/disability-benefits/526EZ/components/AsyncDisplayWidget.jsx
+++ b/src/applications/disability-benefits/526EZ/components/AsyncDisplayWidget.jsx
@@ -27,7 +27,7 @@ export default class AsyncDisplayWidget extends React.Component {
     }
 
     if (typeof uiOptions.viewComponent !== 'function') {
-      throw new Error('AsyncDisplayWidget requires viewComponent in ui:options to be a React element.');
+      throw new Error('AsyncDisplayWidget requires viewComponent in ui:options to be a React component.');
     }
 
     if (typeof uiOptions.callback !== 'function') {
--- a/src/applications/disability-benefits/526EZ/components/AsyncDisplayWidget.jsx
+++ b/src/applications/disability-benefits/526EZ/components/AsyncDisplayWidget.jsx
@@ -80,7 +80,7 @@ export default class AsyncDisplayWidget extends React.Component {
         // Show error message or error component passed in
         const CustomAlert = uiOptions.failureComponent;
         // TODO: Get generic headline and content
-        content = <CustomAlert/> || (
+        content = CustomAlert ? <CustomAlert/> : (
           <AlertBox
             status="error"
             isVisible
--- a/src/applications/disability-benefits/526EZ/components/AsyncDisplayWidget.jsx
+++ b/src/applications/disability-benefits/526EZ/components/AsyncDisplayWidget.jsx
@@ -80,7 +80,7 @@ export default class AsyncDisplayWidget extends React.Component {
         // Show error message or error component passed in
         const CustomAlert = uiOptions.failureComponent;
         // TODO: Get generic headline and content
-        content = CustomAlert || (
+        content = <CustomAlert/> || (
           <AlertBox
             status="error"
             isVisible
--- a/src/applications/disability-benefits/526EZ/components/AsyncDisplayWidget.jsx
+++ b/src/applications/disability-benefits/526EZ/components/AsyncDisplayWidget.jsx
@@ -30,6 +30,10 @@ export default class AsyncDisplayWidget extends React.Component {
       throw new Error('AsyncDisplayWidget requires viewComponent in ui:options to be a React component.');
     }
 
+    if (uiOptions.failureComponent && typeof uiOptions.failureComponent !== 'function') {
+      throw new Error('AsyncDisplayWidget requires the optional failureComponent in ui:options to be a React component.');
+    }
+
     if (typeof uiOptions.callback !== 'function') {
       throw new Error('AsyncDisplayWidget requires callback in ui:options to be a function.');
     }
--- a/src/applications/disability-benefits/526EZ/components/AsyncDisplayWidget.jsx
+++ b/src/applications/disability-benefits/526EZ/components/AsyncDisplayWidget.jsx
@@ -33,9 +33,12 @@ export default class AsyncDisplayWidget extends React.Component {
     if (typeof uiOptions.callback !== 'function') {
       throw new Error('AsyncDisplayWidget requires callback in ui:options to be a function.');
     }
+  }
 
+
+  componentDidMount() {
     // TODO: Don't call the callback _every_ time the component is mounted
-    const cbPromise = uiOptions.callback();
+    const cbPromise = this.props.options.callback();
     if (cbPromise instanceof Promise) {
       cbPromise.then((data) => {
         this.setState({
--- a/src/applications/disability-benefits/526EZ/components/AsyncDisplayWidget.jsx
+++ b/src/applications/disability-benefits/526EZ/components/AsyncDisplayWidget.jsx
@@ -1,0 +1,104 @@
+import React from 'react';
+
+import LoadingIndicator from '@department-of-veterans-affairs/formation/LoadingIndicator';
+import AlertBox from '@department-of-veterans-affairs/formation/AlertBox';
+
+const PENDING = 'PENDING';
+const RESOLVED = 'RESOLVED';
+const REJECTED = 'REJECTED';
+
+
+/**
+ * Handles the various display statuses when calling an asynchronous function.
+ */
+export default class AsyncDisplayWidget extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      data: null,
+      promiseState: PENDING
+    };
+
+    // Make sure we configured it correctly
+    // StringField passes 'ui:options' as the options prop
+    const uiOptions = props.options;
+    if (!uiOptions) {
+      throw new Error('No ui:options supplied to AsyncDisplayWidget.');
+    }
+
+    if (typeof uiOptions.viewComponent !== 'function') {
+      throw new Error('AsyncDisplayWidget requires viewComponent in ui:options to be a React element.');
+    }
+
+    if (typeof uiOptions.callback !== 'function') {
+      throw new Error('AsyncDisplayWidget requires callback in ui:options to be a function.');
+    }
+
+    // TODO: Don't call the callback _every_ time the component is mounted
+    const cbPromise = uiOptions.callback();
+    if (cbPromise instanceof Promise) {
+      cbPromise.then((data) => {
+        this.setState({
+          data,
+          promiseState: RESOLVED
+        });
+      }).catch((data) => {
+        if (data instanceof Error) {
+          throw data;
+        }
+
+        this.setState({
+          data,
+          promiseState: REJECTED
+        });
+      });
+    } else {
+      throw new Error(`AsyncDisplayWidget: Expected callback to return a Promise, but got ${typeof cbPromise}.`);
+    }
+  }
+
+
+  // Not sure if this will be useful yet
+  // componentDidUnmount() {
+  //   // Cancel the promise if it isn't already fulfilled
+  // }
+
+
+  render() {
+    const uiOptions = this.props.options;
+
+    // Depending on the state of the promise, we'll render different things
+    let content;
+    switch (this.state.promiseState) {
+      case RESOLVED: {
+        // Show view component
+        const ViewComponent = uiOptions.viewComponent;
+        content = (<ViewComponent {...this.state.data}/>);
+        break;
+      }
+      case REJECTED: {
+        // Show error message or error component passed in
+        const CustomAlert = uiOptions.failureComponent;
+        // TODO: Get generic headline and content
+        content = CustomAlert || (
+          <AlertBox
+            status="error"
+            isVisible
+            headline="We're sorry, we can't seem to find your information right now."
+            content="That's a real bummer, we know. Maybe try again later."
+            className="fetch-fail-alert"/>
+        );
+        break;
+      }
+      case PENDING:
+      default: {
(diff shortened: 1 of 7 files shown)


src/applications/disability-benefits/526EZ/pages/veteranInfo.js
```diff
@@ -1,64 +1,31 @@
...
-import {
-  veteranInformationViewField,
-  VAFileNumberDescription
-} from '../helpers';
+import { veteranInformationViewField } from '../helpers';
 
 export const uiSchema = {
-  'ui:field': ReviewCardField,
+  'ui:field': 'StringField',
+  'ui:widget': AsyncDisplayWidget,
```
jbalboni: This looks good to me, but it makes me think that maybe there's a use case for some kind of hook in between a fully custom page component and the customization options in RJSF (fields, widgets, etc). Or to just adjust how the custom page component works. It doesn't look like you actually need anything RJSF related in here, but making a custom page component the way it works now would mean too much duplication.

src/applications/disability-benefits/526EZ/components/AsyncDisplayWidget.jsx
```diff
@@ -0,0 +1,104 @@
...
+
+    // Make sure we configured it correctly
+    // StringField passes 'ui:options' as the options prop
+    const uiOptions = props.options;
+    if (!uiOptions) {
+      throw new Error('No ui:options supplied to AsyncDisplayWidget.');
+    }
+
+    if (typeof uiOptions.viewComponent !== 'function') {
+      throw new Error('AsyncDisplayWidget requires viewComponent in ui:options to be a React element.');
```
jbalboni: Probably should change this message to say React component.

src/applications/disability-benefits/526EZ/components/AsyncDisplayWidget.jsx
```diff
@@ -0,0 +1,104 @@
...
+          promiseState: REJECTED
+        });
+      });
+    } else {
+      throw new Error(`AsyncDisplayWidget: Expected callback to return a Promise, but got ${typeof cbPromise}.`);
+    }
+  }
+
+
+  // Not sure if this will be useful yet
```
jbalboni: You'll need to add something to make sure you don't call `setState` if the promise resolves after the component is unmounted. In the past, I've set an instance variable called `unmounted` in `componentDidUnmount` and checked to make sure that wasn't set before calling `setState` when the promise resolves.

jbalboni (approved):

src/applications/disability-benefits/526EZ/pages/veteranInfo.js
cvalarida: Right, exactly. I'm not sure what the solution to the duplication problem is, but a custom page seems like it might be a better fit. This doesn't feel _too_ wonky, though. :man_shrugging:

src/applications/disability-benefits/526EZ/components/AsyncDisplayWidget.jsx
cvalarida: Ahem...Yeahhhhh........Forgot they were different things until I asked for help on the unit tests issue I was having.

I'll just go ahead and update this... :flushed:

src/applications/disability-benefits/526EZ/components/AsyncDisplayWidget.jsx
```diff
@@ -0,0 +1,104 @@
...
+
+    // Make sure we configured it correctly
+    // StringField passes 'ui:options' as the options prop
+    const uiOptions = props.options;
+    if (!uiOptions) {
+      throw new Error('No ui:options supplied to AsyncDisplayWidget.');
+    }
+
+    if (typeof uiOptions.viewComponent !== 'function') {
+      throw new Error('AsyncDisplayWidget requires viewComponent in ui:options to be a React component.');
```
cehsu: This seems very helpful, how do we decide to leverage these errors vs. rely on proptypes? Also, what is the best approach to graceful failure in these cases?

src/applications/disability-benefits/526EZ/components/AsyncDisplayWidget.jsx
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    if (typeof uiOptions.callback !== 'function') {
+      throw new Error('AsyncDisplayWidget requires callback in ui:options to be a function.');
+    }
+
+    // TODO: Don't call the callback _every_ time the component is mounted
```
cehsu: Why not? Should we use redux to store the state to decide whether it is necessary?

src/applications/disability-benefits/526EZ/components/AsyncDisplayWidget.jsx
```diff
@@ -0,0 +1,104 @@
...
+    const cbPromise = uiOptions.callback();
```
cehsu: This is a tough one to name, can we change the name of `cbPromise` to something that describes what it does rather than what it is?

src/applications/disability-benefits/526EZ/components/AsyncDisplayWidget.jsx
```diff
@@ -0,0 +1,104 @@
...
+      case REJECTED: {
+        // Show error message or error component passed in
+        const CustomAlert = uiOptions.failureComponent;
+        // TODO: Get generic headline and content
+        content = CustomAlert ? <CustomAlert/> : (
+          <AlertBox
+            status="error"
+            isVisible
+            headline="We're sorry, we can't seem to find your information right now."
+            content="That's a real bummer, we know. Maybe try again later."
```
cehsu: This content is obviously going to be updated, is it worth requesting that we use curly apostrophes anyway?

src/applications/disability-benefits/526EZ/pages/veteranInfo.js
```diff
@@ -1,64 +1,31 @@
...
-      pattern: 'Your VA file number must be between 7 to 9 digits'
-    }
-  },
-  dateOfBirth: currentOrPastDateUI('Date of birth'),
-  gender: {
-    'ui:title': 'Gender',
-    'ui:options': {
-      labels: genderLabels
+    viewComponent: veteranInformationViewField,
+    loadingMessage: 'Fetching your information...',
```
cehsu: @peggygannon will probably recommend that we use "Loading your information" or something else here as I don't recall using "Fetch" in these messages before.

src/applications/disability-benefits/526EZ/tests/config/veteranInformation.unit.spec.jsx
```diff
@@ -22,39 +22,22 @@ describe('526EZ veteran information', () => {
       uiSchema={uiSchema}/>
     );
 
-    expect(form.find('select').length).to.equal(3);
-    expect(form.find('input').length).to.equal(6);
+    expect(form.find('AsyncDisplayWidget').length).to.equal(1);
   });
 
-  it('should not submit without required info', () => {
```
cehsu: Why are we removing this test?

src/applications/disability-benefits/526EZ/sass/disability-benefits.scss
```diff
@@ -138,14 +138,10 @@
...
-    .legend-label {
-      max-width: 100%;
-    }
-
-  }
+.disability-increase-wizard .fieldset-input .legend-label {
+  max-width: 100%;
+}
 
+.fetch-fail-alert {
```
cehsu: Should we use the component name here .async-display-widget or something, as well?

cehsu (approved):
LGTM. I had a few minor questions/comments. Also is @peggygannon a reviewer of this PR, or are the content changes planned for later?

src/applications/disability-benefits/526EZ/tests/config/veteranInformation.unit.spec.jsx
cehsu: Ah, is it because the page will be view only? Does that mean the test isn't necessary however? Is it possible for any of the required information to be missing? If so, do we have a ticket to discuss/explain how that would be handled? If not, never mind 😄

src/applications/disability-benefits/526EZ/components/AsyncDisplayWidget.jsx
cvalarida: I decided to throw errors for these cases because the widget is _going_ to throw another error if these things are wrong, but the error it'll throw will probably not be super helpful. I think we really shouldn't fail gracefully in these cases because it's a programmer error and needs to be fixed in the form config.

src/applications/disability-benefits/526EZ/components/AsyncDisplayWidget.jsx
cvalarida: Yeah, I oscillated back and forth on this. I think we can do a sort of cache of the result because if they go to the next page, accidentally hit the back button, and then go forward again, we probably shouldn't query the api. I think the right answer is using redux to store whether the query has happened or not, but that gets complicated because we'd then have to store the state, etc. I decided to not go down that route for now in favor of optimizing it later if we see a problem.

src/applications/disability-benefits/526EZ/components/AsyncDisplayWidget.jsx
cvalarida: Heh, I'm wide open to suggestions! I couldn't think of anything better, though. :disappointed:

src/applications/disability-benefits/526EZ/components/AsyncDisplayWidget.jsx
cvalarida: When we get the real content, I'll be sure to use curly apostrophes. :+1:

src/applications/disability-benefits/526EZ/components/AsyncDisplayWidget.jsx
cehsu: We seem to use `apiRequest` pretty frequently as a name for a similar thing. Perhaps `requestHandler` or some such? @jbalboni any other ideas?

src/applications/disability-benefits/526EZ/tests/config/veteranInformation.unit.spec.jsx
cvalarida: Right, exactly. They can't edit this information anymore. Arguably, this PR shouldn't touch this and the veteran information page changes should be handled in https://github.com/department-of-veterans-affairs/vets.gov-team/issues/10440, but I wanted to at least get a proof of concept out there and use that ticket as an iteration on it.

The missing information bit is a great question. I had a similar one for the payment information and didn't update the veteran information ticket. I'll add that as a question in the ticket I linked to above (I don't know what the answer is yet). Good catch. :+1:

src/applications/disability-benefits/526EZ/sass/disability-benefits.scss
cvalarida: Yeah, good idea. :+1:

src/applications/disability-benefits/526EZ/components/AsyncDisplayWidget.jsx
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    if (typeof uiOptions.callback !== 'function') {
+      throw new Error('AsyncDisplayWidget requires callback in ui:options to be a function.');
+    }
+
+    // TODO: Don't call the callback _every_ time the component is mounted
+    const cbPromise = uiOptions.callback();
+    if (cbPromise instanceof Promise) {
+      cbPromise.then((data) => {
```
markgreenburg: Honestly not sure what the implications of calling `setState` within a promise within the constructor are, but I do know that React generally discourages use of `setState` and side effects within the constructor. Might it make sense to check that the cb is a promise in the constructor, but offload the actual call to `componentDidMount`?

peggygannon:
@cvalarida Hi, there. Please see below for my suggested content edits. For the first one, the default and custom message could be the same. Do you need customizable messages for the other 526 detail screens? 

Default message: Please wait while we load your information. 
Custom message: Please wait while we load your information. 

Default message: We're sorry. We can't find your information in our system right now. Please try again later. 
Custom message for Veteran detail screen: 
We're sorry. We can't find your personal details in our system right now. Please try again later.

src/applications/disability-benefits/526EZ/components/AsyncDisplayWidget.jsx
cvalarida: Well, we can't check that the callback returns a promise without calling it, so it's either all in the constructor or all in `componentDidMount`. That said, I think `componentDidMount` is a great place to put this. :+1:

src/applications/disability-benefits/526EZ/components/AsyncDisplayWidget.jsx
```diff
@@ -0,0 +1,104 @@
...
+    switch (this.state.promiseState) {
+      case RESOLVED: {
+        // Show view component
+        const ViewComponent = uiOptions.viewComponent;
+        content = (<ViewComponent {...this.state.data}/>);
+        break;
+      }
+      case REJECTED: {
+        // Show error message or error component passed in
+        const CustomAlert = uiOptions.failureComponent;
```
markgreenburg: If we're checking `if (typeof uiOptions.viewComponent !== 'function')`, does it make sense to also test the same with `CustomAlert`? Granted, it's optional, but if it's specified then we may want to double check that the user passed in the right thing.

cvalarida:
@peggygannon Is there a way we could split up the headline (bold) and content for the error message?

peggygannon:
@cvalarida Hi, Chris, Ok, I tweaked the message a bit, so we have a header. Let me know what you think of this: 

**We can't find your information**
We're sorry. We can't find your information in our system right now. Please try again later.

**We can't find your information**
We're sorry. We can't find your personal details in our system right now. Please try again later.

cvalarida:
Perfect, thanks! :raised_hands:

markgreenburg (approved):
Adding to the chorus of approvals; dropped a couple of thoughts in the comments.

cvalarida:
## Updated screenshots

**Default error box**
image

**Custom error messaging**
image

**Default loading screen**
image

Custom loading screen is the same as the default in this case.

peggygannon:
thanks, @cvalarida Updated screenshots look good.

src/applications/disability-benefits/526EZ/components/AsyncDisplayWidget.jsx
cvalarida: Yeah, not a bad idea.

PR #7807, opened 2018-06-01
 Added Health Records E2E test.

U-DON:
Resolves department-of-veterans-affairs/vets.gov-team#3012.


dhall (approved):
there is a code climate issue, but otherwise -- LGTM!

PR #7811, opened 2018-06-01
Sign-in Modal/DSLogon Downtime - Remove loading indicator (checking the undefined status)

ncksllvn:
`monitoring/DowntimeNotifications/components/Banner.jsx` was designed for basic messaging about services being down and renders children components only if the dependencies are offline. This PR implements it into the Sign-in Modal to get rid of that "checking the undefined status" loading message.

Resolves department-of-veterans-affairs/vets.gov-team/issues/10760
More info - department-of-veterans-affairs/vets.gov-team/issues/9390

### DIFF
--- a/src/platform/monitoring/DowntimeNotification/components/Banner.jsx
+++ b/src/platform/monitoring/DowntimeNotification/components/Banner.jsx
@@ -17,11 +17,12 @@ export default function DowntimeBanner({ appTitle, dependencies, children }) {
       appTitle={appTitle}
       render={(downtime) => {
         if (downtime.status === serviceStatus.down) {
+          if (children) return children;
           return (
             <AlertBox
               status="info"
               isVisible
-              content={children || <DownMessaging appTitle={appTitle} {...downtime}/>}/>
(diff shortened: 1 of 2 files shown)


jbalboni (approved):

ncksllvn:
@jbalboni I really hate to propose this idea because it reminds me too much of ColdFusion, but do you think that at this point it's worth rendering per status like this? 

```
<Status.Down dependencies={[services.evss]}>
We are offline right now.
</Status.Down>
<Status.DowntimeApproaching dependencies={[services.evss]}>
We will soon be offline.
</Status.DowntimeApproaching>
<Status.OK dependencies={[services.evss]}>
We are online
</Status.OK>
```

jbalboni:
@ncksllvn I don't have a problem with how that code looks, but I'm not sure I understand where it would go in this PR.

ncksllvn:
@jbalboni Okay cool, thanks. Yeah it would be overkill in this PR. I feel like we keep finding different use cases for downtime notifications and I'm trying to find the right balance of abstraction and flexibility.

PR #7813, opened 2018-06-01
526 homelessness poc mg

markgreenburg:
- Adds homelessness POC data capture into form (name, phone number) as a conditional expander when a veteran indicates they are / are at risk of becoming homeless
- Requires that either both name & number be entered, or neither

(Still working out a couple of failing unit tests)

Note: Requires separate PR to migrate the added POC schema into `vets-json-schema`, where it should live long-term.

### Default UI:
screen shot 2018-06-04 at 4 42 15 pm
-----

### Expanded UI ['Yes' selected]:
screen shot 2018-06-04 at 4 42 47 pm
-----

### Name / phone fields conditionally required:
conditionally_required

### DIFF
--- a/src/applications/disability-benefits/526EZ/config/form.js
+++ b/src/applications/disability-benefits/526EZ/config/form.js
@@ -243,6 +243,13 @@ const formConfig = {
                   'ui:title': 'Contact full name',
                   'ui:errorMessages': {
                     pattern: "Full names can only contain letters, numbers, spaces, dashes ('-'), and forward slashes ('/')"
+                  },
+                  'ui:required': (formData) => {
+                    const { veteran } = formData;
+                    if (!veteran['view:homelessness'] || !veteran.homelessness) {
+                      return false;
+                    }
+                    return !!veteran.homelessness.primaryPhone;
                   }
                 },
                 primaryPhone: {
@@ -250,6 +257,13 @@ const formConfig = {
                   'ui:widget': SSNWidget, // probably not the best name for this widget...
                   'ui:errorMessages': {
                     pattern: 'Phone numbers must be 10 digits (dashes allowed)'
+                  },
+                  'ui:required': (formData) => {
+                    const { veteran } = formData;
+                    if (!veteran['view:homelessness'] || !veteran.homelessness) {
+                      return false;
+                    }
+                    return !!veteran.homelessness.pointOfContactName;
                   }
                 }
               }
--- a/src/applications/disability-benefits/526EZ/config/form.js
+++ b/src/applications/disability-benefits/526EZ/config/form.js
@@ -229,41 +229,43 @@ const formConfig = {
           uiSchema: {
             veteran: {
               'ui:title': 'Homelessness',
-              'view:homelessness': {
-                'ui:title': 'Are you homeless or at risk of becoming homeless?',
-                'ui:widget': 'yesNo',
-              },
               homelessness: {
-                'ui:description': 'Please provide the name and number of a person we should call if we need to get in touch with you.',
-                'ui:options': {
-                  expandUnder: 'view:homelessness',
-                  expandUnderCondition: (homelessOrAtRisk) => (homelessOrAtRisk === true)
+                isHomeless: {
+                  'ui:title': 'Are you homeless or at risk of becoming homeless?',
+                  'ui:widget': 'yesNo',
                 },
-                pointOfContactName: {
-                  'ui:title': 'Name of person we should contact',
-                  'ui:errorMessages': {
-                    pattern: "Full names can only contain letters, numbers, spaces, dashes ('-'), and forward slashes ('/')"
+                pointOfContact: {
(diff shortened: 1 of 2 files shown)


src/applications/disability-benefits/526EZ/config/form.js
```diff
@@ -72,12 +72,14 @@ import {
...
 } from '../validations';
 import { validateBooleanGroup } from '../../../common/schemaform/validation';
+import PhoneNumberWidget from '../../../common/schemaform/widgets/PhoneNumberWidget';
 
 const {
   treatments: treatmentsSchema,
   privateRecordReleases,
   serviceInformation,
   standardClaim,
+  // veteran
```
cvalarida: Is this needed as a placeholder for anything, or is it a vestige of testing?

src/applications/disability-benefits/526EZ/config/form.js
```diff
@@ -229,18 +231,76 @@ const formConfig = {
...
+                'ui:options': {
+                  expandUnder: 'view:homelessness',
+                  expandUnderCondition: (homelessOrAtRisk) => (homelessOrAtRisk === true)
+                },
+                pointOfContactName: {
+                  'ui:title': 'Name of person we should contact',
+                  'ui:errorMessages': {
+                    pattern: "Full names can only contain letters, numbers, spaces, dashes ('-'), and forward slashes ('/')"
+                  },
+                  'ui:required': (formData) => {
```
cvalarida: @jbalboni I'd like to get your insight as to whether or not this will work. Our goals are the following:

1. Veteran is not homeless nor at risk of being homeless
   - We don't send the `homelessness` property to EVSS
1. Veteran is homeless or at risk of being homeless, but has no point of contact
    - We send EVSS `homelessness: {}`
    - Ideally, we'd send a boolean, but that's not how it works, unfortunately
1. Veteran is homeless or at risk of being homeless and has a point of contact
    - We send `homelessness: { pointOfContactName: 'asdf', primaryPhone: '1234567890' }`

If the PoC name and number are blank, I believe we'll clear out the homelessness object before sending it to the api. Is that correct?

@markgreenburg If my understanding is correct, we might want to make `view:homelessness` not a `view:` property so we send it to the api. They can then determine whether to send no object, an empty object, or an object with the contact info (which we'd need to write another ticket for).

src/applications/disability-benefits/526EZ/config/form.js
```diff
@@ -229,18 +231,76 @@ const formConfig = {
...
-                type: 'boolean'
-              },
+              veteran: {
+                type: 'object',
+                required: ['view:homelessness'],
+                properties: {
+                  'view:homelessness': {
+                    type: 'boolean'
+                  },
+                  // TODO: Update to use 526 homelessness schema once
```
cvalarida: ...once it's in vets-json-schema?

cvalarida (approved):
Looks good, though there seems to be some clean up in order. My biggest concern is what we'll actually be sending vets-api if they select yes but don't put any contact information in.

peggygannon (approved):
@markgreenburg the content looks good. thanks!

src/applications/disability-benefits/526EZ/config/form.js
jbalboni: Yes, that's the case. The form code tries to avoid saving empty objects in the form data and will also attempt to strip them out when you submit.

src/applications/disability-benefits/526EZ/config/form.js
markgreenburg: It might be needed, but it's not like it's hard to put back if needed. Depends on how we structure and import the schema. Going to take it out for now.

src/applications/disability-benefits/526EZ/config/form.js
markgreenburg: Didn't know that, thanks for pointing it out! So the initial approach wouldn't have worked anyway. I just got rid of the needlessly complicated bits and made the schema way more explicit, which I think is much nicer.

src/applications/disability-benefits/526EZ/config/form.js
```diff
@@ -229,18 +230,83 @@ const formConfig = {
...
+              homelessness: {
+                isHomeless: {
+                  'ui:title': 'Are you homeless or at risk of becoming homeless?',
+                  'ui:widget': 'yesNo',
+                },
+                pointOfContact: {
+                  'ui:description': 'Please provide the name and number of a person we should call if we need to get in touch with you.',
+                  'ui:options': {
+                    expandUnder: 'isHomeless',
+                    expandUnderCondition: (homelessOrAtRisk) => (homelessOrAtRisk === true)
```
cehsu: Is `expandUnderCondition` necessary if the condition is true? It might not be needed here.

cehsu (approved):
LGTM. Had a minor question.

src/applications/disability-benefits/526EZ/config/form.js
markgreenburg: You're right! That's nice.

PR #7814, opened 2018-06-04
Migrate pension and burials form to new zip code format

jbalboni:
Adds new validation and migrations for ICMHS zip code handling, since there are some restrictions on the format for US addresses.

screen shot 2018-06-04 at 9 43 46 am

### DIFF
--- a/src/applications/common/schemaform/definitions/address.js
+++ b/src/applications/common/schemaform/definitions/address.js
@@ -72,6 +72,8 @@ const requiredFields = ['street', 'city', 'country', 'state', 'postalCode'];
  *
  * @param {object} schema - Schema for a full form, including address definition in definitions
  * @param {boolean} isRequired - If the address is required or not, defaults to false
+ * @param {string} addressProperty - The name of the address definition to use from the common
+ *   definitions in currentSchema
  */
 export function schema(currentSchema, isRequired = false, addressProperty = 'address') {
   const addressSchema = currentSchema.definitions[addressProperty];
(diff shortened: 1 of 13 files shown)


src/applications/common/schemaform/definitions/address.js
```diff
@@ -73,8 +73,9 @@ const requiredFields = ['street', 'city', 'country', 'state', 'postalCode'];
  * @param {object} schema - Schema for a full form, including address definition in definitions
  * @param {boolean} isRequired - If the address is required or not, defaults to false
```
rianfowler: what is `addressProperty`

src/applications/pensions/config/form.js
```diff
@@ -1544,7 +1548,11 @@ const formConfig = {
           path: 'additional-information/contact',
           uiSchema: {
             'ui:title': 'Contact information',
-            veteranAddress: address.uiSchema('Mailing address'),
+            veteranAddress: _.set(
+              'ui:validations[1]',
```
rianfowler: How is this used?

rianfowler (approved):

src/applications/pensions/config/form.js
jbalboni: This is adding the central mail zip code validation to the list of validation functions, which already has one function for the general address validation.

src/applications/common/schemaform/definitions/address.js
jbalboni: Added an additional JsDoc comment below.

lbai001 (approved):

PR #7817, opened 2018-06-04
Pre-need form address updates

jbalboni:
This makes a couple updates to the pre-need address fields. I did this by copying the common address  definition, since that seems more comprehensible than trying to extend it. The changes are:

- Remove Mexican state list dropdown
- Update State validation to 3 characters
- Hide State field for non US/CAN addresses
- Update to schema with updated field lengths

### DIFF
--- a/src/applications/pre-need/definitions/address.js
+++ b/src/applications/pre-need/definitions/address.js
@@ -25,7 +25,8 @@ export const countriesWithStateCodes = new Set(['USA', 'CAN']);
 function validateAddress(errors, address, formData, currentSchema) {
   // Adds error message for state if it is blank and one of the following countries:
   // USA, Canada, or Mexico
-  if (address.state === undefined
+  if (countriesWithStateCodes.has(address.country)
+    && address.state === undefined
     && currentSchema.required.length) {
     errors.state.addError('Please select a state or province');
   }
@@ -71,7 +72,7 @@ const requiredFields = ['street', 'city', 'country', 'state', 'postalCode'];
  * @param {boolean} isRequired - If the address is required or not, defaults to false
  */
 export function schema(currentSchema, isRequired = false) {
-  const addressSchema = currentSchema.definitions.address || currentSchema.definitions.addressWithRequiredZip;
+  const addressSchema = currentSchema.definitions.address;
   return {
     type: 'object',
     required: isRequired ? requiredFields : [],
@@ -89,7 +90,7 @@ export function schema(currentSchema, isRequired = false) {
       },
       postalCode: {
         type: 'string',
-        maxLength: 10
+        maxLength: 6
       }
     })
   };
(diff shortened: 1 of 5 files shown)


src/applications/pre-need/definitions/address.js
```diff
@@ -0,0 +1,228 @@
...
+import { createSelector } from 'reselect';
+
+import { countries, states, isValidUSZipCode, isValidCanPostalCode } from '../../../platform/forms/address';
+
+function validatePostalCodes(errors, address) {
+  let isValidPostalCode = true;
+
+  // Checks if postal code is valid
+  if (address.country === 'USA') {
+    isValidPostalCode = isValidPostalCode && isValidUSZipCode(address.postalCode);
```
rianfowler: Why check `isValidPostalCode` here if it's set to true on line 7?

src/applications/pre-need/definitions/address.js
```diff
@@ -0,0 +1,228 @@
...
+  if (address.postalCode && !isValidPostalCode) {
+    errors.postalCode.addError('Please provide a valid postal code');
+  }
+}
+
+export const countriesWithStateCodes = new Set(['USA', 'CAN']);
+
+function validateAddress(errors, address, formData, currentSchema) {
+  // Adds error message for state if it is blank and one of the following countries:
+  // USA, Canada, or Mexico
```
rianfowler: PR notes suggest this is just USA and Canada now.

src/applications/pre-need/definitions/address.js
```diff
@@ -0,0 +1,228 @@
...
+export function uiSchema(label = 'Address', useStreet3 = false, isRequired = null, ignoreRequired = false) {
+  let fieldOrder = ['country', 'street', 'street2', 'street3', 'city', 'state', 'postalCode'];
+  if (!useStreet3) {
+    fieldOrder = fieldOrder.filter(field => field !== 'street3');
+  }
+
+  const addressChangeSelector = createSelector(
+    ({ formData, path }) => _.get(path.concat('country'), formData),
+    ({ formData, path }) => _.get(path.concat('city'), formData),
+    _.get('addressSchema'),
```
rianfowler: how does this work with one argument?

rianfowler (approved):

src/applications/pre-need/definitions/address.js
jbalboni: Lodash/fp does auto-currying, so this returns a function that takes formData and gets the object at `addressSchema`. I've changed my mind about using that feature since this code was originally written, so I think I'll change it.

rianfowler: I thought that might be the case but it's not being used as a function in the `resultFunc` e.g. `addressSchema.properties` on line 121.

PR #7820, opened 2018-06-05
Fix failing keyword search test in master

jbalboni:
Looks like sometimes the debounce finishes and tries to run some code that relies on a prop being passed in that wasn't. I've added that prop.


wyattwalter (approved):
Thanks Jeff!

PR #7823, opened 2018-06-05
526 veteran info

cvalarida:
This doesn't cover the payment information like the mock up in the connected ticket. That work will be done in https://github.com/department-of-veterans-affairs/vets.gov-team/issues/10750.

image

### DIFF
--- a/src/applications/disability-benefits/526EZ/helpers.jsx
+++ b/src/applications/disability-benefits/526EZ/helpers.jsx
@@ -357,37 +357,21 @@ export const evidenceSummaryView = ({ formData }) => {
   );
 };
 
-const FullNameViewField = ({ formData }) => {
-  const { first, middle, last, suffix } = formData;
-  return <strong>{first} {middle} {last} {suffix}</strong>;
-};
-
-const SsnViewField = ({ formData }) => {
-  const ssn = formData.slice(5);
-  const mask = <span>•••-••-</span>;
-  return <p>Social Security number: {mask}{ssn}</p>;
-};
-
-const VAFileNumberViewField = ({ formData }) => {
-  const vaFileNumber = formData.slice(5);
-  const mask = <span>•••-••-</span>;
-  return <p>VA file number: {mask}{vaFileNumber}</p>;
-};
-
-const DateOfBirthViewField = ({ formData }) => {
-  return <p>Date of birth: <DateWidget value={formData} options={{ monthYear: false }}/></p>;
-};
-
-const GenderViewField = ({ formData }) => <p>Gender: {genderLabels[formData]}</p>;
-
-export const veteranInformationViewField = (formData) => {
+// TODO: For the masked numbers, have screen readers say "ending in" instead of reading out the bullets
+export const veteranInformationViewField = (data) => {
+  const { ssn, vaFileNumber, dateOfBirth, gender } = data;
+  const { first, middle, last, suffix } = data.fullName;
   return (
     <div>
-      <FullNameViewField formData={formData.fullName}/>
-      <SsnViewField formData={formData.socialSecurityNumber}/>
-      <VAFileNumberViewField formData={formData.vaFileNumber}/>
-      <GenderViewField formData={formData.gender}/>
-      <DateOfBirthViewField formData={formData.dateOfBirth}/>
+      <p>This is the personal information we have on file for you.</p>
+      <div className="blue-bar-block">
+        <strong>{first} {middle} {last} {suffix}</strong>
+        <p>Social Security number: ●●●–●●–{ssn.slice(5)}</p>
+        <p>VA file number: ●●●–●●–{vaFileNumber.slice(5)}</p>
+        <p>Date of birth: <DateWidget value={dateOfBirth} options={{ monthYear: false }}/></p>
+        <p>Gender: {genderLabels[gender]}</p>
+      </div>
+      <p><strong>Note:</strong> If something doesn’t look right and you need to update your details, please call Veterans Benefits Assistance at <a href="tel:1-800-827-1000">1-800-827-1000</a>, Monday through Friday, 8:00 a.m. to 9:00 p.m. (ET).</p>
     </div>
   );
 };
--- a/src/applications/disability-benefits/526EZ/helpers.jsx
+++ b/src/applications/disability-benefits/526EZ/helpers.jsx
@@ -357,17 +357,31 @@ export const evidenceSummaryView = ({ formData }) => {
   );
 };
 
-// TODO: For the masked numbers, have screen readers say "ending in" instead of reading out the bullets
+/**
+ * @param {ReactElement|ReactComponent|String} srIgnored -- Thing to be displayed visually,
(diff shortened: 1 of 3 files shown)


src/applications/disability-benefits/526EZ/helpers.jsx
```diff
@@ -357,37 +357,35 @@ export const evidenceSummaryView = ({ formData }) => {
...
-};
-
-const DateOfBirthViewField = ({ formData }) => {
-  return <p>Date of birth: <DateWidget value={formData} options={{ monthYear: false }}/></p>;
+/**
+ * @param {ReactElement|ReactComponent|String} srIgnored -- Thing to be displayed visually,
+ *                                                           but ignored by screen readers
+ * @param {String} substitutionText -- Text for screen readers to say instead of srIgnored
+ */
+const srSubstitute = (srIgnored, substitutionText) => {
```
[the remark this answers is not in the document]
cehsu: Have we already checked in with @1Copenut for advice on this?

src/applications/disability-benefits/526EZ/helpers.jsx
```diff
@@ -357,37 +357,35 @@ export const evidenceSummaryView = ({ formData }) => {
...
+  const mask = srSubstitute('●●●–●●–', 'ending with');
   return (
     <div>
-      <FullNameViewField formData={formData.fullName}/>
-      <SsnViewField formData={formData.socialSecurityNumber}/>
-      <VAFileNumberViewField formData={formData.vaFileNumber}/>
-      <GenderViewField formData={formData.gender}/>
-      <DateOfBirthViewField formData={formData.dateOfBirth}/>
+      <p>This is the personal information we have on file for you.</p>
+      <div className="blue-bar-block">
```
cehsu: I like how descriptive this is, but wonder whether it is consistent with our class naming conventions, which seem to typically favor describing usage over appearance.

src/applications/disability-benefits/526EZ/helpers.jsx
```diff
@@ -357,37 +357,35 @@ export const evidenceSummaryView = ({ formData }) => {
...
+        <strong>{first} {middle} {last} {suffix}</strong>
+        <p>Social Security number: {mask}{ssn.slice(5)}</p>
+        <p>VA file number: {mask}{vaFileNumber.slice(5)}</p>
+        <p>Date of birth: <DateWidget value={dateOfBirth} options={{ monthYear: false }}/></p>
+        <p>Gender: {genderLabels[gender]}</p>
+      </div>
+      <p><strong>Note:</strong> If something doesn’t look right and you need to update your details, please call Veterans Benefits Assistance at <a href="tel:1-800-827-1000">1-800-827-1000</a>, Monday through Friday, 8:00 a.m. to 9:00 p.m. (ET).</p>
```
cehsu: @peggygannon do we typically tend to use "Monday through Friday" or "Monday - Friday" in these cases?

src/applications/disability-benefits/526EZ/sass/disability-benefits.scss
```diff
@@ -145,3 +145,13 @@
 .async-display-widget-alert-box {
   margin-bottom: 1em;
 }
+
+// This seems duplicative
```
cehsu: of `va-address-block`? I believe that was the original recommendation to use here, perhaps there are some differences?

src/applications/disability-benefits/526EZ/helpers.jsx
peggygannon: @cehsu In these short messages, we should use "Monday - Friday", rather than Monday through Friday.

src/applications/disability-benefits/526EZ/helpers.jsx
cvalarida: Yeah, I'm not fond of this name, but never looped back to change it because I didn't know what else to call it that wasn't similarly generic. If you have suggestions, I'm all ~ears~ eyes. :smiley:

src/applications/disability-benefits/526EZ/helpers.jsx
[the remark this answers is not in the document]
cvalarida: Nope. I asked in Ad Hoc's accessibility channel and that's the advice I got, but it would be great to get feedback from the guru himself.

What're your thoughts, @1Copenut?

src/applications/disability-benefits/526EZ/sass/disability-benefits.scss
cvalarida: Yeah, I saw `va-address-block` as well, but we shouldn't reuse it here. The name is misleading. Also, that particular one indents the div quite a ways, which is undesired.

We could rename that one and do the indenting with another class, but that seemed a little odd. Not a bad solution, though. On the other hand, this was a pretty small addition, so I could go either way. :man_shrugging:

src/applications/disability-benefits/526EZ/helpers.jsx
```diff
@@ -385,7 +385,7 @@ export const veteranInformationViewField = (data) => {
         <p>Date of birth: <DateWidget value={dateOfBirth} options={{ monthYear: false }}/></p>
         <p>Gender: {genderLabels[gender]}</p>
       </div>
-      <p><strong>Note:</strong> If something doesn’t look right and you need to update your details, please call Veterans Benefits Assistance at <a href="tel:1-800-827-1000">1-800-827-1000</a>, Monday through Friday, 8:00 a.m. to 9:00 p.m. (ET).</p>
+      <p><strong>Note:</strong> If something doesn’t look right and you need to update your details, please call Veterans Benefits Assistance at <a href="tel:1-800-827-1000">1-800-827-1000</a>, Monday – Friday, 8:00 a.m. to 9:00 p.m. (ET).</p>
```
cehsu: It looks like we're not very consistent about how we write this actually, several different combination of spaces and dash types appear to be in use, is there a preference among these @peggygannon? If so, we could set up a ticket to make these consistent:
`Monday &#8211; Friday`
`Monday – Friday`
`Monday — Friday`

cvalarida: According to http://www.getitwriteonline.com/archive/091502enem.htm, it should be an `en dash` (the middle one).

src/applications/disability-benefits/526EZ/helpers.jsx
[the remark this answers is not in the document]
1Copenut: I don't see any issue with this pattern. You have a visible string for sighted users that screen readers will ignore, and a screen reader only string that sighted users will never see. Feel free to send me a URL and I can validate it against assistive devices as needed.

src/applications/disability-benefits/526EZ/helpers.jsx
cehsu: something along the lines of `address-block` seems helpful? it could be qualified as `disability-address-block` to make it form specific if that is of interest.

cehsu (approved):

markgreenburg (approved):
Looks good, tested locally as well.

PR #7824, opened 2018-06-05
Fixed blank greeting when user has null first name.

U-DON:
The default object for `userFullName` is something like `{ first: null, last: null, middle: null, suffix: null }`. and the selector only checks whether `userFullName` is falsey. Reproduced the issue of having nothing next to the dropdown arrow by signing in with an LOA1 user.

> screen shot 2018-06-05 at 6 02 13 pm


jbalboni (approved):
Doh, thanks for catching this.

PR #7825, opened 2018-06-05
Remove homelessness schema and use imported definition instead

markgreenburg:
Uses new vets-json-schema homelessness schema instead of the current one that's defined directly within the form config itself.

Still to do:
- [x] Update the vets-json-schema dependency commit hash once those changes are merged (note to self: don't forget to `yarn install`)
- [x] Double check tests


package.json
```diff
@@ -227,7 +227,7 @@
     "reselect": "^2.5.4",
     "text-loader": "^0.0.1",
     "url-search-params": "^0.10.0",
-    "vets-json-schema": "https://github.com/department-of-veterans-affairs/vets-json-schema.git#80bbe9ff577324df326eba42ff74144db90e39af",
+    "vets-json-schema": "https://github.com/department-of-veterans-affairs/vets-json-schema.git#51ede2dbd960c4035f93bcef8bac6827f8cb1abe",
```
cvalarida: Is this the most up-to-date hash? (You still have the to-do item in the description is why I ask.)

cvalarida (approved):
Looks good.

package.json
markgreenburg: Yep, it's the right hash, though not technically the _most recent_ hash.

PR #7827, opened 2018-06-06
adds applicantInfoConditional

lbai001:
This pr is for if the applicant is NOT veteran according to the issue description.
so far the titles of the fields are default titles without "Your", seems like that is still in discussion.

screencapture-localhost-3001-disability-benefits-apply-dependents-form-686c-dependents-applicant-information-part-2-2018-06-05-23_29_26

### DIFF
--- a/src/applications/disability-benefits/686/config/form.js
+++ b/src/applications/disability-benefits/686/config/form.js
@@ -30,7 +30,7 @@ import {
   spouseRelationshipDescription,
   childRelationshipDescription,
   otherRelationshipDescription,
-  isVeteran
+  isNotVeteran
 } from '../helpers.jsx';
 
 import { validateAfterMarriageDate } from '../validation';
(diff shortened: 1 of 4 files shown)


src/applications/disability-benefits/686/helpers.jsx
```diff
@@ -105,7 +105,7 @@ export const otherRelationshipDescription = (
 );
 
 export function isVeteran(item) {
```
jbalboni: Should we change the name of this function? It's now if you're not a veteran, not that you are a veteran.

src/applications/disability-benefits/686/config/form.js
```diff
@@ -173,6 +168,23 @@ const formConfig = {
               }
             }
           }
+        },
+        applicantInformationConditional: {
```
jbalboni: We could call this `claimantInformation`, and have the path below be `claimant-information`.

lbai001: k

src/applications/disability-benefits/686/helpers.jsx
lbai001: k

jbalboni (approved):

emilywaggoner:
Thanks Annie! We will be using the "Your" pronouns here, so it looks like it will be in good shape as soon as those are in place. Thanks!

PR #7831, opened 2018-06-06
Vet360 profile

bshyong:
Fixes for vet360 profile updates


src/applications/personalization/profile360/actions/updaters.js
```diff
@@ -148,7 +149,7 @@ export const saveField = {
   updateWorkPhone: updateVet360Field('/profile/telephones', VET360_CONSTANTS.FIELD_NAMES.WORK_PHONE, 'phone'),
   updateTemporaryPhone: updateVet360Field('/profile/telephones', VET360_CONSTANTS.FIELD_NAMES.TEMP_PHONE, 'phone'),
   updateFaxNumber: updateVet360Field('/profile/telephones', VET360_CONSTANTS.FIELD_NAMES.FAX_NUMBER, 'phone'),
-  updateEmailAddress: updateVet360Field('/profile/email', VET360_CONSTANTS.FIELD_NAMES.EMAIL, 'email'),
-  updateMailingAddress: updateVet360Field('/profile/mailing_address', VET360_CONSTANTS.FIELD_NAMES.MAILING_ADDRESS, 'address'),
-  updateResidentialAddress: updateVet360Field('/profile/mailing_address', VET360_CONSTANTS.FIELD_NAMES.RESIDENTIAL_ADDRESS, 'address')
+  updateEmailAddress: updateVet360Field('/profile/email_addresses', VET360_CONSTANTS.FIELD_NAMES.EMAIL, 'email'),
```
jbalboni: It seems odd that email addresses is plural here, but I don't know what the endpoint is, so just pointing it out.

jbalboni (approved):

src/applications/personalization/profile360/actions/updaters.js
bshyong: Thanks! yes looks a little odd but we may be able to support multiple email addresses in the future https://github.com/department-of-veterans-affairs/vets-api/blob/master/config/routes.rb#L181

src/applications/personalization/profile360/components/PhoneSection.jsx
```diff
@@ -60,8 +60,8 @@ class EditPhoneModal extends React.Component {
 
             <ErrorableTextInput
               label="Number"
-              field={{ value: field.value.phoneNumber, dirty: false }}
-              onValueChange={this.onChange('phoneNumber')}
+              field={{ value: field.value.inputPhoneNumber, dirty: false }}
```
bshyong: Added this to handle the full phone number that users enter. When submitting to vet360 this input is split into Area Code and Phone Number components

PR #7837, opened 2018-06-07
Vet360 profile

bshyong:
Handle international postal code


rlbaker (approved):
LGTM

cvalarida (approved):
Assuming we can't reuse the input for `zipCode` as `internationalPostalCode` depending on the type, it looks good to me.

Is there any reason we can't reuse that input? Validations maybe?

bshyong:
Yes we potentially can but would have to be careful to send the correct keyNames in the request. We will revisit/refactor as we start adding additional error handling and validations for all of these inputs

PR #7838, opened 2018-06-07
Profile copy updates

bshyong:
- Copy changes to resolve https://github.com/department-of-veterans-affairs/vets.gov-team/issues/10853#event-1668909096
- Resolves bug when switching from domestic to international address


rlbaker (approved):
LGTM

PR #7839, opened 2018-06-07
Facilty type ahead

cvalarida:
image

### DIFF
--- a/src/applications/disability-benefits/526EZ/helpers.jsx
+++ b/src/applications/disability-benefits/526EZ/helpers.jsx
@@ -1,5 +1,7 @@
 import React from 'react';
 import AdditionalInfo from '@department-of-veterans-affairs/formation/AdditionalInfo';
+import Raven from 'raven-js';
+import appendQuery from 'append-query';
 
 import { isValidUSZipCode, isValidCanPostalCode } from '../../../platform/forms/address';
 import { stateRequiredCountries } from '../../common/schemaform/definitions/address';
@@ -7,6 +9,7 @@ import { transformForSubmit } from '../../common/schemaform/helpers';
 import cloneDeep from '../../../platform/utilities/data/cloneDeep';
 import get from '../../../platform/utilities/data/get';
 import set from '../../../platform/utilities/data/set';
+import { apiRequest } from '../../../platform/utilities/api';
 import { genderLabels } from '../../../platform/static-data/labels';
 import { getDiagnosticCodeName, getDiagnosticText } from './reference-helpers';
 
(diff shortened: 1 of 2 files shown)


cehsu (approved):

src/applications/disability-benefits/526EZ/helpers.jsx
```diff
@@ -606,18 +609,26 @@ export const noFDCWarning = (
...
+    name_part: input // eslint-disable-line camelcase
+  });
 
-// Mimic querying the api for options
-export function queryForFacilities(input = '') {
-  // Emulate a fast api call
-  return Promise.resolve(options.filter(o => o.label.includes(input)));
+  return apiRequest(url, {},
+    (response) => {
+      return response.data.map(f => ({ id: f.id, label: f.attributes.name }));
```
cehsu: Can we use a full word here rather than a letter ("f")?

src/applications/disability-benefits/526EZ/helpers.jsx
```diff
@@ -606,18 +609,26 @@ export const noFDCWarning = (
...
+    },
+    (error) => {
+      Raven.captureMessage('Error querying for facilities', { input, error });
```
cehsu: Should we return a data structure in the event of an error?

cehsu (approved):

src/applications/disability-benefits/526EZ/helpers.jsx
cvalarida: Do you mean log something else to sentry or the `return []` on the next line?

cehsu: I'm mainly asking if maintainers will expect a data structure to be returned from a failed request, it looks we typically use these to dispatch actions/logs, but perhaps I'm not considering the full range of how we use error handlers.

cvalarida: Yeah, if we don't return anything from this function, when we get a failed query, it'll "pass" `undefined` to the Autosuggest field which I imagine will break things because it expects an array. I think we don't want to break the functionality entirely since the user can still enter free text, so I opted to not throw an error. Dispatching an error action is interesting, but I don't know how we'd use it right now.

We could use some discovery on how we'd want to let the user know that we're having difficulty querying for the list of facilities at the moment. Might be worth writing a ticket for the backlog.

Do you want to do that, or shall I?

cehsu: Sorry the motivation of my question wasn't clear, rather than passing an empty array, could an empty array be used as a default parameter in the argument that uses this? I'm not as familiar with the downstream portion of this, so perhaps that isn't practical.

cvalarida: Ohhhhhhh, okay. I think I'm following now.
```js
// AutosuggestField
const results = queryForResults(input) || [];
```
That's what you're getting at?

Yeah, that's probably a generally better approach. I think for now, I'll leave it like this is, but write up another ticket to make it so in the common library. (I'd make the change now, but with the form library code freeze, I'd have to do it again, would be likely to forget, etc.)

Good call. :+1:

PR #7841, opened 2018-06-07
Autosuggest safety valve

cvalarida:
image

### DIFF
--- a/src/applications/common/schemaform/fields/AutosuggestField.jsx
+++ b/src/applications/common/schemaform/fields/AutosuggestField.jsx
@@ -14,6 +14,13 @@ function getInput(input, uiSchema, schema) {
 
   if (typeof input !== 'object' && input) {
     const uiOptions = uiSchema['ui:options'];
+    // When using this field in an array item, editing the item will throw an error
+    //  if there uiOptions.label is undefined (as when we queryForResults), so we
+    //  have to have this safety valve
+    if (!uiOptions.labels) {
+      return input;
+    }
+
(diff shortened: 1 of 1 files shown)


src/applications/common/schemaform/fields/AutosuggestField.jsx
```diff
@@ -14,6 +14,13 @@ function getInput(input, uiSchema, schema) {
 
   if (typeof input !== 'object' && input) {
     const uiOptions = uiSchema['ui:options'];
+    // When using this field in an array item, editing the item will throw an error
+    //  if there uiOptions.label is undefined (as when we queryForResults), so we
+    //  have to have this safety valve
```
cvalarida: Not sure this comment block is needed, strictly speaking, but it seemed like a specific enough case to warrant explanation.

src/applications/common/schemaform/fields/AutosuggestField.jsx
```diff
@@ -14,6 +14,13 @@ function getInput(input, uiSchema, schema) {
...
+    if (!uiOptions.labels) {
+      return input;
```
cvalarida: @jbalboni I don't think this will break anything, but I'm not sure how to be _certain_ it won't.

jbalboni: Looks good to me.

jbalboni (approved):

cehsu (approved):

PR #7842, opened 2018-06-08
preview build support

wyattwalter:
Enabling builds for preview environment. Also started messing with getting URLs configured at runtime to start making a pass at what it would take to do that for all the things.

ref: https://github.com/department-of-veterans-affairs/vets.gov-team/issues/10772

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,6 +1,6 @@
 import org.kohsuke.github.GitHub
 
-def envNames = ['development', 'staging', 'production']
+def envNames = ['development', 'staging', 'production', 'devpreview', 'preview']
 
 def devBranch = 'master'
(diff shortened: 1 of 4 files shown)


wyattwalter:
Ready for round of review here as a possibility to support builds. Exposes a feature flag (Webpack DefinePlugin) called `mergedbuild` which can be used to include/exclude old and new styles of builds. Right now it adds ~5 minutes to the Jenkins build times, going to spend some time today to see if we can tune that a bit.

jkassemi (approved):
👏 

Looks good. Curious to hear results on the build times.

wyattwalter:
Added some more CPU and bumped up number of parallel processes for uglify and times are now down by ~3 minutes from previous.

PR #7844, opened 2018-06-08
Add submitUrl

cvalarida:
We may actually need to use a `submit` callback if we have to do any data transformation. We'll have to test that once our schema stops changing.

I got the URL for the submit endpoint from our [swagger docs](https://department-of-veterans-affairs.github.io/va-digital-services-platform-docs/api-reference/#/form_526/postSubmitForm).

### DIFF
--- a/src/applications/disability-benefits/526EZ/config/form.js
+++ b/src/applications/disability-benefits/526EZ/config/form.js
@@ -119,8 +119,8 @@ const initialData = {
 const formConfig = {
   urlPrefix: '/',
   intentToFileUrl: '/evss_claims/intent_to_file/compensation',
-  // submitUrl: `${environment.API_URL}/v0/21-526EZ`,
-  submit: () => Promise.resolve({ attributes: { confirmationNumber: '123123123' } }),
+  submitUrl: `${environment.API_URL}/v0/disability_compensation_form/submit`,
+  // submit: () => Promise.resolve({ attributes: { confirmationNumber: '123123123' } }),
   trackingPrefix: 'disability-526EZ-',
   formId: '21-526EZ',
   version: 1,


src/applications/disability-benefits/526EZ/config/form.js
```diff
@@ -119,8 +119,8 @@ const initialData = {
 const formConfig = {
   urlPrefix: '/',
   intentToFileUrl: '/evss_claims/intent_to_file/compensation',
-  // submitUrl: `${environment.API_URL}/v0/21-526EZ`,
-  submit: () => Promise.resolve({ attributes: { confirmationNumber: '123123123' } }),
+  submitUrl: `${environment.API_URL}/v0/disability_compensation_form/submit`,
+  // submit: () => Promise.resolve({ attributes: { confirmationNumber: '123123123' } }),
```
cehsu: We can remove this comment now if we're able to use the actual submit endpoint.

cehsu (approved):

src/applications/disability-benefits/526EZ/config/form.js
cvalarida: Yeah, I was going to leave it as a sort of reminder if we have to do some data translations. And now that I think about it, I'm sure we will since we've got basically all the things wrapped up in the `disabilities` array. :thinking:

cehsu: Sounds good, could you please set up a ticket with your thoughts for that?

cvalarida: New ticket for this can be found at https://github.com/department-of-veterans-affairs/vets.gov-team/issues/10875

PR #7845, opened 2018-06-08
Change the 526 URL

cvalarida:
I grepped in `content/` and `src/` for `/disability-benefits/526` and went through each item, changing the URLs to the updated one. I think I got them all, but double-checking is always welcome.

### DIFF
--- a/script/build.js
+++ b/script/build.js
@@ -615,7 +615,6 @@ if (!options.watch && !(process.env.CHECK_BROKEN_LINKS === 'no')) {
         '/health-care/apply/application',
         '/veteran-id-card/apply',
         '/veteran-id-card/how-to-get',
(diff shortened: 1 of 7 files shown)


cehsu (approved):

script/build.js
```diff
@@ -615,7 +615,6 @@ if (!options.watch && !(process.env.CHECK_BROKEN_LINKS === 'no')) {
         '/health-care/apply/application',
         '/veteran-id-card/apply',
         '/veteran-id-card/how-to-get',
-        '/disability-benefits/apply-for-increase',
```
markgreenburg: 🎉

markgreenburg (approved):
👌

PR #7846, opened 2018-06-08
526 address logic mg

markgreenburg:
- Updates mailingAddress and forwardingAddress fields to latest shape
- Adds conditional display logic for certain fields that should only be shown under certain conditions
- Adds conditional required logic for certain fields that should only be required under certain conditions
- Updates a bunch of tests and adds more tests for the new logic
- Cleans up the 'view' UI for the address card (a little)

Separate issues need to be created to:
- Update `vets-json-schema` to strip PCIU address common def and create a new, very basic schema for our addresses
- Update the schema in the form config to pull from vets-json-schema instead of creating the schema in the config directly (once the above is done)
- Clean up the address view card, especially WRT forwarding addresses
- Update / review content for all of the validation errors

### Domestic Address:
screen shot 2018-06-08 at 1 30 00 pm
-----
### International Address:
screen shot 2018-06-08 at 1 30 41 pm
-----
### Military Address - Bad State:
screen shot 2018-06-08 at 1 31 52 pm
-----
### Military Address - Bad City:
screen shot 2018-06-08 at 1 32 06 pm
-----
### Forwarding Address - Works the same as mailing address
screen shot 2018-06-08 at 1 32 45 pm

### DIFF
--- a/src/applications/disability-benefits/526EZ/helpers.jsx
+++ b/src/applications/disability-benefits/526EZ/helpers.jsx
@@ -15,6 +15,9 @@ const siblings = ['treatments', 'privateRecordReleases', 'privateRecords', 'addi
 import { PREFILL_STATUSES } from '../../common/schemaform/save-in-progress/actions';
 import { DateWidget } from '../../common/schemaform/review/widgets';
 
+export const USA = 'USA';
+export const MILITARY_STATES = ['AA', 'AE', 'AP'];
+export const MILITARY_CITIES = ['APO', 'DPO', 'FPO'];
 
 /*
  * Flatten nested array form data into sibling properties
@@ -508,47 +511,38 @@ const EffectiveDateViewField = ({ formData }) => {
 };
 
 const AddressViewField = ({ formData }) => {
-  const {
-    addressLine1, addressLine2, addressLine3, city, state,
-    zipCode, militaryStateCode, militaryPostOfficeTypeCode
-  } = formData;
+  const { addressLine1, addressLine2, addressLine3, city, state, country, zipCode } = formData;
   let zipString;
-  let stateString;
-  let cityString;
   if (zipCode) {
     const firstFive = zipCode.slice(0, 5);
     const lastChunk = zipCode.length > 5 ? `-${zipCode.slice(5)}` : '';
     zipString = `${firstFive}${lastChunk}`;
   }
-  if (city || militaryPostOfficeTypeCode) {
-    cityString = `${city},` || `${militaryPostOfficeTypeCode},`;
-  }
-  if (state || militaryStateCode) {
-    stateString = `${state}` || `${militaryStateCode}`;
+
+  let lastLine;
+  if (country === USA) {
+    lastLine = `${city}, ${state} ${zipString}`;
+  } else {
+    lastLine = `${city}, ${country}`;
   }
   return (
     <div>
       {addressLine1 && <p>{addressLine1}</p>}
       {addressLine2 && <p>{addressLine2}</p>}
       {addressLine3 && <p>{addressLine3}</p>}
-      <p>{cityString} {stateString} {zipString}</p>
+      <p>{lastLine}</p>
     </div>
   );
 };
 
 export const PrimaryAddressViewField = ({ formData }) => {
   const {
-    mailingAddress, primaryPhone, secondaryPhone,
-    emailAddress, forwardingAddress
-  } = formData;
+    mailingAddress, primaryPhone, emailAddress, forwardingAddress } = formData;
   return (
     <div>
       <AddressViewField formData={mailingAddress}/>
       {primaryPhone && (
         <PhoneViewField formData={primaryPhone} name="Primary phone"/>
-      )}
-      {secondaryPhone && (
-        <PhoneViewField formData={secondaryPhone} name="Secondary phone"/>
       )}
       {emailAddress && (
         <EmailViewField formData={emailAddress} name="Email address"/>
--- a/src/applications/disability-benefits/526EZ/pages/primaryAddress.js
+++ b/src/applications/disability-benefits/526EZ/pages/primaryAddress.js
@@ -3,11 +3,15 @@ import _ from 'lodash';
 import fullSchema526EZ from 'vets-json-schema/dist/21-526EZ-schema.json';
 
 import dateUI from '../../../common/schemaform/definitions/date';
-import SSNWidget from '../../../common/schemaform/widgets/SSNWidget';
+import PhoneNumberWidget from '../../../common/schemaform/widgets/PhoneNumberWidget';
 
 import ReviewCardField from '../components/ReviewCardField';
 
 import { PrimaryAddressViewField } from '../helpers';
+
+// Used in our validations
+const MILITARY_STATES = ['AA', 'AE', 'AP'];
+const MILITARY_CITIES = ['APO', 'DPO', 'FPO'];
 
 function isValidZIP(value) {
   if (value !== null) {
@@ -37,11 +41,21 @@ function validateZIP(errors, zip) {
   }
 }
 
-const typeLabels = {
-  DOMESTIC: 'Domestic',
-  MILITARY: 'Military',
-  INTERNATIONAL: 'International'
-};
+function validateMilitaryCity(errors, city, formData, schema, messages, options) {
+  const isMilitaryState = MILITARY_STATES.includes(formData.veteran[options.addressName].state);
+  const isMilitaryCity = MILITARY_CITIES.includes(city.toUpperCase());
+  if (isMilitaryState && !isMilitaryCity) {
+    errors.addError('City must match APO, DPO, or FPO when using a military state code');
+  }
+}
+
+function validateMilitaryState(errors, state, formData, schema, messages, options) {
+  const isMilitaryCity = MILITARY_CITIES.includes(formData.veteran[options.addressName].city.toUpperCase());
+  const isMilitaryState = MILITARY_STATES.includes(state);
+  if (isMilitaryCity && !isMilitaryState) {
+    errors.addError('State must be AA, AE, or AP when using a military city');
+  }
+}
 
 const states = [
   'AL',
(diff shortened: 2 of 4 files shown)


src/applications/disability-benefits/526EZ/pages/primaryAddress.js
```diff
@@ -558,13 +527,53 @@ export const uiSchema = {
...
+          'city',
+          'state',
+          'zipCode'
+        ],
+        effectiveDate: _.merge(
+          {},
+          dateUI('Effective date'),
+          { 'ui:required': ({ veteran }) => (hasForwardingAddress(veteran)) }
+        ),
         country: {
```
cehsu: Can we now use the common address uiSchema?

markgreenburg: Maybe? I hadn't really explored that option because the schemas are different (mainly due to regex differences on all fields), and the validations / required fields are different as well. There's also the matter of the state list eventually getting populated by the API, whereas the common UI schema does some fancy work to change the list depending on the country (USA / MEX / CAN). I'd be happy to chat about this later - my gut feeling is that I'd end up spending more time trying to get the common schema to work for us than if I just make a new uiSchema for our addresses. I could definitely be wrong though.

cehsu: IIRC @jbalboni's recommendation was to make use of the common schema logic if possible, perhaps we could use _.set to update whatever needs to be extended (validations, requirements, etc.)?

src/applications/disability-benefits/526EZ/tests/config/primaryAddress.unit.spec.jsx
```diff
@@ -9,7 +9,7 @@ import { DefinitionTester, // selectCheckbox
 import formConfig from '../../config/form.js';
 import initialData from '../schema/initialData.js';
 
-describe('Disability benefits 526EZ primary address', () => {
+describe.only('Disability benefits 526EZ primary address', () => {
```
cvalarida: Missed a spot.

cvalarida (commented):
Only two things really stick out to me here:
1) `street1` vs `addressLine`
2) The stray `.only()` in the test

I also only skimmed the tests because I was running out of time. For the most part, it looks good, though.

src/applications/disability-benefits/526EZ/pages/primaryAddress.js
cehsu: Thanks @cvalarida can you please indicate which fields would be required but hidden using the common schema? I think @jbalboni recommended using the common schema because it provides the basic UI functionality we need here (besides some of the common validations @markgreenburg mentioned, which we could possibly add via _.set or somesuch), as well as some additional assistance for the user (when they select Canada or Mexico), but I'll let him chime in if that's not what he had in mind.

src/applications/disability-benefits/526EZ/pages/primaryAddress.js
```diff
@@ -392,92 +390,72 @@ const countries = [
...
-        }
-      }
-    },
     addressLine1: {
-      'ui:title': 'Street'
+      'ui:title': 'Street address'
     },
     addressLine2: {
-      'ui:title': 'Line 2'
+      'ui:title': 'Street address (optional)'
```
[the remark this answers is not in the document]
markgreenburg: I stole this text from Letters, since that's our only PCIU precedent. I'm intending to have Peggy weigh in on content (along with validation content) in another PR - this is mostly just to make things work.

src/applications/disability-benefits/526EZ/tests/config/primaryAddress.unit.spec.jsx
markgreenburg: 😬

src/applications/disability-benefits/526EZ/helpers.jsx
```diff
@@ -508,48 +511,39 @@ const EffectiveDateViewField = ({ formData }) => {
 };
 
 const AddressViewField = ({ formData }) => {
-  const {
-    addressLine1, addressLine2, addressLine3, city, state,
-    zipCode, militaryStateCode, militaryPostOfficeTypeCode
-  } = formData;
+  const { addressLine1, addressLine2, addressLine3, city, state, country, zipCode } = formData;
```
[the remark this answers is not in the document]
markgreenburg: Yeah, that makes sense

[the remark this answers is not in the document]
markgreenburg: Oh wait, I know why I did this. They're `addressLine1` in EVSS's swagger. I suppose we could have vets-api rename them if we care.

src/applications/disability-benefits/526EZ/pages/primaryAddress.js
```diff
@@ -392,92 +390,72 @@ const countries = [
...
-    militaryStateCode: {
-      'ui:title': 'Military State Code',
+    state: {
+      'ui:title': 'State',
+      'ui:required': ({ veteran }) => (veteran.mailingAddress.country === USA),
       'ui:options': {
         labels: stateLabels,
-        hideIf: formData => _.get(formData, `veteran[${addressName}].type`) !== 'MILITARY'
-      }
+        hideIf: ({ veteran }) => (veteran.mailingAddress.country !== USA),
```
cehsu: State is not required if the address is US, but does that mean it should be hidden? Can states still be provided for international addresses, it seems like these would be helpful?

src/applications/disability-benefits/526EZ/pages/primaryAddress.js
```diff
@@ -3,11 +3,12 @@ import _ from 'lodash';
 import fullSchema526EZ from 'vets-json-schema/dist/21-526EZ-schema.json';
 
 import dateUI from '../../../common/schemaform/definitions/date';
-import SSNWidget from '../../../common/schemaform/widgets/SSNWidget';
+import PhoneNumberWidget from '../../../common/schemaform/widgets/PhoneNumberWidget';
```
cehsu: There is an additional option for implementing the phone field using the common uiSchema and a BE transformation listed in this issue: https://github.com/department-of-veterans-affairs/vets.gov-team/issues/10897

src/applications/disability-benefits/526EZ/pages/primaryAddress.js
```diff
@@ -392,92 +390,72 @@ const countries = [
...
+        validator: validateMilitaryState
+      }]
     },
     zipCode: {
       'ui:title': 'ZIP code',
       'ui:validations': [validateZIP],
+      'ui:required': ({ veteran }) => (veteran.mailingAddress.country === USA),
       'ui:errorMessages': {
-        pattern: 'Please enter a valid 9 digit ZIP code (dashes allowed)'
+        pattern: 'Please enter a valid 5- or 9- digit ZIP code (dashes allowed)'
```
cehsu: Should the error message on line 22 be updated as well to match this?

src/applications/disability-benefits/526EZ/pages/primaryAddress.js
```diff
@@ -578,10 +587,8 @@ export const primaryAddressSchema = {
       properties: {
         mailingAddress: addressSchema,
         primaryPhone: {
-          type: 'string'
-        },
-        secondaryPhone: {
-          type: 'string'
+          type: 'string',
+          pattern: '^\\d{10}$'
```
cehsu: Will this permit dashes?  Should we allow dashes on the front end and remove them in a backend transformation?

src/applications/disability-benefits/526EZ/pages/primaryAddress.js
cehsu: Thanks Mark, I've added a list of the discrepancies to the issue to facilitate discussion on this:  https://github.com/department-of-veterans-affairs/vets.gov-team/issues/10595#issuecomment-396294475

src/applications/disability-benefits/526EZ/pages/primaryAddress.js
markgreenburg: So actually the phone number widget uses local state and sanitizes it before sending the number to redux. This means that we get to use a simple pattern here, while still allowing parens, spaces, and dashes in the front end.

src/applications/disability-benefits/526EZ/pages/primaryAddress.js
markgreenburg: I hid state because that's how we built Letters (country !== USA: hide state field), which is just a so-so reason on its own. But I also learned today that vets360 considers 'state' input for international addresses as invalid, so it seems that we should probably leave it off here as well.

cehsu: Thanks Mark, if we include some sort of state field for international addresses, it looks like they will likely need to be remapped to a province or provinceCode as part of the tranformations and toggle the label on the FE as well. (Accordion to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/10878#issuecomment-396270145)

cehsu: Is including the optional international province field something to check with @goldenmeanie about or should we plan to follow how the common address uiSchema behaves (it includes province for international addresses)?

goldenmeanie: @cehsu, I think go with standard!

src/applications/disability-benefits/526EZ/helpers.jsx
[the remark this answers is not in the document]
cvalarida: Yup, we should have vets-api rename them because we want to stick to as close a replica of the common address schema as we can. I believe @SMLuthi is sending up `street1` and friends in pre-fill, and submit should also transform them.

src/applications/disability-benefits/526EZ/pages/primaryAddress.js
cvalarida: Um...Is that the ticket you meant to link?

cvalarida (approved):

src/applications/disability-benefits/526EZ/pages/primaryAddress.js
cehsu: Yep, although it's an unrelated issue, it may be of interest here in the event that we want to be more consistent with the other forms (which seem to use the common phone uiSchema -- which itself uses the `PhoneNumberWidget`), I think I set this up using the widget before we decided to start using the BE to perform transformations. Not sure what the preference is here so mentioning it in case it's helpful.

src/applications/disability-benefits/526EZ/helpers.jsx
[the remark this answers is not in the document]
SMLuthi: This has actually already been changed to `addressLine1`, `addressLine2` and `addressLine3`. I *can* change it to street again but if its just for convention sake, I'm not sure if its worth the time. But I can be persuaded.

src/applications/disability-benefits/526EZ/pages/primaryAddress.js
markgreenburg: Sorry, not sure I understand the difference between the two options? To back up for a sec - we never intended to support functionality (e.g., international provinces) that differs from PCIU.
1. We will likely need to populate the state & countries lists from an api at some point - this api will only contain a US states list.
2. Letters went through extensive FE & BE validation for PCIU addresses (note - Letters actually *does* update the user's PCIU address on file), and also loads state / country data from an endpoint. We don't currently support non-US state / province input there.
3. The new Personalization dashboard includes a widget that lets veterans update their address. The specs for this address updater match PCIU / what we did in Letters - only US addresses can have states / provinces.

I understand these things may change with the switch to vets360, but I don't think we should complicate our implementation right now.

cehsu: Thanks for clarifying @markgreenburg, I couldn't tell from the specs that international addresses shouldn't accept states/provinces, I was only aware that they would not be required in that case, also I didn't realize we had already made a design decision on this, so thanks for the context. Let's go with whatever is actually accepted.

cehsu (approved):

src/applications/disability-benefits/526EZ/pages/primaryAddress.js
markgreenburg: All good! It's not just you, I think we all struggled with the specs for this initially (Chris made really good specs that are attached to this ticket, but it took us a really long time to get there). Appreciate you taking the time to think critically about whether these are / are not the right changes to make - I'm generally a proponent of using as much common code as possible.

src/applications/disability-benefits/526EZ/pages/primaryAddress.js
cvalarida: Ohhhh, I gotcha. You weren't referencing the problem in the ticket, but the proposed solution therein. :ok_hand:
I'm on the same page now. Thanks!

PR #7852, opened 2018-06-11
Migrate HCA to use us-forms-system

annekainicUSDS:


cvalarida:
So the idea here is to use the us-forms-system for everything except the custom stuff that isn't in there? (Plus moving all the custom stuff to `platform/` to separate it from the library all the other forms are using in `applications/`?)

jbalboni (approved):

jbalboni:
@cvalarida Yes. Once we have all the forms using the code in src/platform/forms/save-in-progress, src/applications/common will be removed.

PR #7854, opened 2018-06-11
Migrate first batch of forms

jbalboni:
This migrates the VIC v2, VRE, and Veteran Representative forms to the external forms system. None of these forms are live, so it's not very risky.


annekainicUSDS (approved):
LGTM! Do we want to wait to merge this until the HCA changes are deployed and we can ensure that's working well? Or are you comfortable merging this now?

annekainicUSDS:
One other thought I had: do we also want to update the scroll functions for the e2e tests for these? I noticed your tests are all passing without that change, but the edu forms I migrated last night weren't. I wonder if this might be a problem in the future that we may as well fix now?

jbalboni:
Ah, I should definitely update those tests. These forms are all disabled in prod, so the e2e tests aren't running in the build.

I think we can wait until HCA looks good to merge this.

PR #7856, opened 2018-06-11
Profile: Refresh state for profile after Vet360 transaction completes

ncksllvn:
This moves the mapping function for the profile object from the action creator into its own utility, which is instead called in the reducer. It was refactored so that we could reuse the `getProfile` action creator with the fewest amount of side effects.

Resolves department-of-veterans-affairs/vets.gov-team/issues/10880

### DIFF
--- a/src/applications/personalization/profile360/actions/updaters.js
+++ b/src/applications/personalization/profile360/actions/updaters.js
@@ -1,4 +1,5 @@
 import { apiRequest } from '../../../../platform/utilities/api';
+import { getProfile } from '../../../../platform/user/profile/actions';
 // import recordEvent from '../../../../platform/monitoring/record-event';
 // import { kebabCase } from 'lodash';
 import { pickBy } from 'lodash';
@@ -30,7 +31,7 @@ export const VET360_TRANSACTION_FINISHED = 'VET360_TRANSACTION_FINISHED';
 // }
 
 export function getTransactionStatus(transaction, fieldName) {
-  return async (dispatch, getState) => {
+  return async (dispatch) => {
     try {
       const { transactionId } = transaction.data.attributes;
       const response = isVet360Configured() ? await apiRequest(`/profile/status/${transactionId}`) : localVet360.updateTransaction(transactionId);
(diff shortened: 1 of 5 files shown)


src/applications/personalization/profile360/actions/updaters.js
```diff
@@ -43,22 +44,14 @@ export function getTransactionStatus(transaction, fieldName) {
...
-            fieldName
-          });
-        }, 3000);
-
-        // Update the property on the FE
+
+        // Refresh the profile object
+        await dispatch(getProfile());
+
+        // Remove the transaction from the VA Profile
```
ncksllvn: This ultimately checks the status for a Vet360 transaction and dispatches the getUser action creator if the transaction is completed.

src/platform/site-wide/user-nav/containers/Main.jsx
```diff
@@ -84,28 +84,11 @@ export class Main extends React.Component {
...
-        recordEvent({ event: 'login-user-logged-in' });
-        this.props.updateLoggedInStatus(true);
-      }
-    } else {
+    if (!sessionStorage.userToken) {
       this.props.updateLoggedInStatus(false);
       if (this.getRedirectUrl()) { this.props.toggleLoginModal(true); }
+    } else {
+      this.props.getProfile();
     }
```
ncksllvn: This was leftover code from many months ago when we tried to add a session refresh modal before realizing it's more complex than expected. I had to change it because I added a return statement into `getProfile` so that we can wait for that to finish before removing a Vet360 transaction from state.

src/platform/user/profile/reducers/index.js
```diff
@@ -56,8 +57,10 @@ const initialState = {
 
 function profileInformation(state = initialState, action) {
   switch (action.type) {
-    case UPDATE_PROFILE_FIELDS:
-      return merge(state, action.newState);
+    case UPDATE_PROFILE_FIELDS: {
+      const newState = mapRawUserDataToState(action.payload);
+      return merge(state, newState);
+    }
```
ncksllvn: I moved the actual mapping into the reducer for consistency across vets-website.

src/platform/user/profile/utilities/index.js
```diff
@@ -0,0 +1,66 @@
...
+      last
+    },
+    verified,
+    vet360: isVet360Configured() ? camelCaseObjectKeys(vet360ContactInformation) : camelCaseObjectKeys(mockContactInformation),
+    veteranStatus: {
+      isVeteran,
+      veteranStatus
+    }
+  };
+}
```
ncksllvn: I'm not sure if it's clearer using destructuring to this extent versus how we were doing things previously, but I wanted to be explicit with how we are building the user state.

bshyong: Agree this should be explicit

U-DON (approved):
LGTM. We could use the `X-Key-Inflection` in the fetch request to camelCase all the keys. But then again, I believe there was some discussion about doing away with that option anyway.

PR #7859, opened 2018-06-12
Migrate edu forms 1

annekainicUSDS:
Migrate the following forms to using us-forms-system as opposed to the version of the forms library code within vets-website.


jbalboni (approved):
Looks good to me. We should wait a little to merge it to give HCA some more time in production.

PR #7861, opened 2018-06-12
Separate getProfile into refresh/restore

ncksllvn:
There are occasions where we want to refresh the user's profile (by hitting the /user API route) but the single `getProfile` action creator wasn't really designed for that - it contained analytics reporting, caching/tearing down session variables, and dispatching other actions that may cause unknown side effects.

This PR separate `getProfile` into `restoreProfile`, which is used for the first load of the user profile, and `refreshProfile`, used for subsequent loads.

### DIFF
--- a/src/platform/user/profile/actions/index.js
+++ b/src/platform/user/profile/actions/index.js
@@ -1,7 +1,8 @@
-import recordEvent from '../../../monitoring/record-event';
 import { removeFormApi } from '../../../../applications/common/schemaform/save-in-progress/api';
 import { updateLoggedInStatus } from '../../authentication/actions';
 import environment from '../../../utilities/environment';
+
+import { setupProfileSession, teardownProfileSession } from '../utilities';
 
 export const UPDATE_PROFILE_FIELDS = 'UPDATE_PROFILE_FIELDS';
 export const PROFILE_LOADING_FINISHED = 'PROFILE_LOADING_FINISHED';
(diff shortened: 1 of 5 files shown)


src/platform/user/profile/actions/index.js
```diff
@@ -25,37 +26,37 @@ export function profileLoadingFinished() {
...
-      // Report out the current level of assurance for the user
-      recordEvent({ event: `login-loa-current-${userData.loa.current}` });
-      dispatch(updateProfileFields(payload));
+    }
+
+    const payload = await response.json();
+    dispatch(updateProfileFields(payload));
+    return payload;
+  };
+}
```
ncksllvn: I was a unsure of whether there should be any error handling here. With `restoreProfile` it makes sense to check for a `401` and take action if necessary, but I'm not sure if that's appropriate when doing a refresh.

src/platform/user/profile/actions/index.js
```diff
@@ -25,37 +26,37 @@ export function profileLoadingFinished() {
...
+
+export function restoreProfile() {
```
bshyong: thoughts on renaming the actions to `setUpProfile` and `fetchProfile`? the distinction between refresh and restore is a little unclear to me

ncksllvn: I was considering `initializeProfile` instead of `restoreProfile`, but `setupProfile` works for me too. I can make that change if you'd like.

ncksllvn:
Original issue is resolved, but linking here for consistency - department-of-veterans-affairs/vets.gov-team/issues/10880

ncksllvn:
@bshyong changed `restoreProfile` to `initializeProfile` for clarity

bshyong (approved):

U-DON (approved):
LGTM. Also, I noticed an incorrect mapping in `mapRawUserDataToState` that I missed from the last PR. Could you update the `mhv` property in the returned object like so?

```
    mhv: {
      account: { state: mhvAccountState, },
      terms: { accepted: healthTermsCurrent }
    },
```

ncksllvn:
@U-DON good catch, thanks for the heads up. Updated the MHV mapping!

PR #7862, opened 2018-06-12
526 ITF redux

cehsu:
These changes set up the disability increase compensation ITF redux action creators and reducer and have been split out of https://github.com/department-of-veterans-affairs/vets-website/pull/7672 .

Original ticket is available at https://github.com/department-of-veterans-affairs/vets.gov-team/issues/10113


src/applications/disability-benefits/526EZ/actions/index.js
```diff
@@ -1,64 +1,138 @@
...
-        setTimeout(resolve, ms);
+export function checkITFRequest(dispatch, hasSavedForm) {
+  return apiRequest(
+    '/intent_to_file',
+    null,
+    ({ data }) => {
+      let status;
+      let expirationDate;
+      const itfList = data.attributes.intentToFile;
+      if (!itfList || (Array.isArray(itfList) && itfList.length === 0)) {
```
cehsu: @SMLuthi helped to determine what possible shapes the response data might be, any additional insights appreciated as well.

cehsu:
@peggygannon mentioned that the content of the ITF messages will be updated today, @peggygannon are the different cases/scenarios expected to change as well? If not, this work reflects the ITF cases determined by @goldenmeanie and @peggygannon.

src/applications/disability-benefits/526EZ/tests/itfData.js
```diff
@@ -0,0 +1,1243 @@
+// The spelling of these status types has been validated with the partner team
+
+export default {
+  id: '',
+  type: 'evss_intent_to_file_intent_to_files_responses',
+  attributes: {
+    intentToFile: [
```
markgreenburg: Do all these ITFs test unique scenarios? It seems many are duplicates, except for small differences in creation and expiration dates. We might be able to remove many of them without any loss in test coverage.

cehsu: This is our BE mock data, we can certainly pare this down if that's preferred.

src/applications/disability-benefits/526EZ/actions/index.js
```diff
@@ -1,64 +1,138 @@
...
+        status = PRESTART_STATUSES.none;
       } else {
-        const errorMessage = 'Network request failed';
-        onChange({
-          ITFStatus,
-          errorMessage
-        });
-        reject(errorMessage);
+        const activeList = itfList.filter(itf => itf.status === 'active');
+        if (activeList.length === 0) {
```
cvalarida: These logic gates seem like they're getting a little convoluted. My suspicion is we can't _really_ simplify this, but can you put some comments in here explaining why all these if statements are necessary? I think it'll make reading it later much easier.

src/applications/disability-benefits/526EZ/actions/index.js
```diff
@@ -1,64 +1,138 @@
...
+          status = PRESTART_STATUSES.retrieved;
+          expirationDate = activeList[0].expirationDate;
+        }
       }
-    });
+      dispatch(setPrestartStatus(status, expirationDate));
+      return status;
+    },
+    ({ errors }) => {
+      const errorMessage = 'Network request failed';
```
cvalarida: This is perhaps misleading. It could have failed for any number of reasons, I think. Looks like it's only used in the sentry log. That error message will be as useful without this, I think.

src/applications/disability-benefits/526EZ/actions/index.js
```diff
@@ -1,64 +1,138 @@
...
+      const status = hasSavedForm ? PRESTART_STATUSES.notRetrievedSaved : PRESTART_STATUSES.notRetrievedNew;
+      Raven.captureMessage(`vets_itf_error: ${errorMessage}`);
```
cvalarida: What do you think of `vets_itf_check_failure`?

src/applications/disability-benefits/526EZ/actions/index.js
```diff
@@ -1,64 +1,138 @@
...
-//       const errorMessage = 'Network request failed';
-//       onChange({
-//         ITFStatus,
-//         errorMessage
-//       });
-//       Raven.captureMessage(`vets_itf_error: ${errorMessage}`);
-//     }
-//   );
-// };
+// function fakeITFRequest(x, cb) { // remove once local testing complete
```
cvalarida: This can probably be removed now, right?

src/applications/disability-benefits/526EZ/actions/index.js
```diff
@@ -1,64 +1,138 @@
...
+  return apiRequest(
+    '/intent_to_file/compensation',
+    { method: 'POST' },
+    ({ data }) => {
+      dispatch(setPrestartStatus(successStatus, data.attributes.intentToFile.expirationDate));
+      return successStatus;
+    },
+    ({ errors }) => {
+      const errorMessage = 'Network request failed';
+      Raven.captureMessage(`vets_itf_error: ${errorMessage}`);
```
cvalarida: Same as above regarding the `errorMessage`. What do you think of `vets_itf_submission_failure`?

src/applications/disability-benefits/526EZ/reducer.js
```diff
@@ -1,6 +1,58 @@
...
+  data: undefined,
+  display: false
+};
+
+export const prestart = (state = initialState, action) => {
+  switch (action.type) {
+    case SET_PRESTART_STATUS: {
+      const newState = _.set('status', action.status, state);
+      if (!state.display) {
+        newState.display = true;
```
cvalarida: This will always set `newState.display` to truthy; either it's truthy to begin with or it's true now.

If there's no reason to pass a falsey `display` value, can we remove the check altogether?
If this is to set a default value, you might be looking for `if (state.display === undefined)`.

src/applications/disability-benefits/526EZ/reducer.js
```diff
@@ -1,6 +1,58 @@
...
+      }
+
+      if (action.data && state.data) {
+        const previousData = state.data;
+        newState.data = {};
+        newState.data.previous = previousData;
+        newState.data.current = action.data;
```
cvalarida: What problem is this block solving?

src/applications/disability-benefits/526EZ/reducer.js
```diff
@@ -1,6 +1,58 @@
...
+      return newState;
+    }
+    case RESET_PRESTART_STATUS: {
+      let newState = _.set('status', PRESTART_STATUSES.notAttempted, state);
+      newState = _.unset('data', newState);
+
+      return newState;
+    }
+    case RESET_PRESTART_DISPLAY: {
+      const newState = _.set('display', false, state);
```
cvalarida: It's unclear to me what `display` will be used for. Can you explain?

src/applications/disability-benefits/526EZ/actions/index.js
```diff
@@ -1,64 +1,138 @@
-/* eslint-disable no-unused-vars */
 import { apiRequest } from '../../../../platform/utilities/api';
-import { fetchInProgressForm } from '../../../common/schemaform/save-in-progress/actions';
+import Raven from 'raven-js';
 
-export const ITFStatuses = Object.freeze({
```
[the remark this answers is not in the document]
cehsu: Good point, I'll add `active` back in, as we use it to filter in `checkITFRequest`.

src/applications/disability-benefits/526EZ/actions/index.js
```diff
@@ -1,64 +1,138 @@
...
-  active: 'active',
-  claim_received: 'claim_received', // eslint-disable-line camelcase
-  duplicate: 'duplicate',
-  expired: 'expired',
+export const SET_PRESTART_STATUS = 'SET_PRESTART_STATUS';
```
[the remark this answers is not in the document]
cehsu: Thanks for the heads up, will reverse these.

src/applications/disability-benefits/526EZ/actions/index.js
```diff
@@ -1,64 +1,138 @@
...
+export const RESET_PRESTART_STATUS = 'RESET_PRESTART_STATUS';
+export const RESET_PRESTART_DISPLAY = 'RESET_PRESTART_DISPLAY';
+
+export const PRESTART_STATUSES = {
```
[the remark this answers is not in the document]
cehsu: I see your point, we could break this up into `prestart` and `itf` status.

src/applications/disability-benefits/526EZ/actions/index.js
```diff
@@ -1,64 +1,138 @@
...
+    type: RESET_PRESTART_DISPLAY
+  };
+}
 
-      // TODO: if the backend handles resubmission, this check can be removed
-      if (ITFStatus === ITFStatuses.active) {
-        setTimeout(resolve, ms);
+export function checkITFRequest(dispatch, hasSavedForm) {
+  return apiRequest(
+    '/intent_to_file',
```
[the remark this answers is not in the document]
cehsu: It is, @SMLuthi confirmed.

src/applications/disability-benefits/526EZ/actions/index.js
cehsu: Comments are a good idea, will add.

src/applications/disability-benefits/526EZ/actions/index.js
cehsu: Would changing this to `vets_itf_check_failure` or the error in `submitITFRequest` to `vets_itf_compensation_error` be clearer (basing the error on the endpoint or the calling function)?

src/applications/disability-benefits/526EZ/actions/index.js
cehsu: Sounds good, asked a question about preference above but will change these as recommended for now.

src/applications/disability-benefits/526EZ/actions/index.js
cehsu: These are for local testing/review purposes, they can be removed.

src/applications/disability-benefits/526EZ/reducer.js
cehsu: Because we've got subsequent requests being made (`submit` after `check`), it's not necessary to set display to true once it's already been set by an earlier call.

src/applications/disability-benefits/526EZ/reducer.js
cehsu: In the event that a user's ITF has expired, we display both the previous ITF's expiration date (when it expired), and the current expiration date (when the created ITF will expire).

src/applications/disability-benefits/526EZ/reducer.js
cehsu: We only want to display the success message on the first form page that the users arrives on (whether starting or resuming a form), once they navigate to another form page, we no longer want to display the success message according to @goldenmeanie. This setting indicates whether the prestart message should be displayed.

src/applications/disability-benefits/526EZ/actions/index.js
```diff
@@ -1,64 +1,138 @@
...
-        const errorMessage = 'Network request failed';
-        onChange({
-          ITFStatus,
-          errorMessage
-        });
-        reject(errorMessage);
+        const activeList = itfList.filter(itf => itf.status === 'active');
+        if (activeList.length === 0) {
+          status = PRESTART_STATUSES.expired;
+          expirationDate = itfList.sort((a, b) => a.expirationDate - b.expirationDate)[0].expirationDate;
```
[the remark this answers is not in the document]
cehsu: Nice catch, these need to be converted into dates, will update.

src/applications/disability-benefits/526EZ/actions/index.js
cehsu: This error messaging was modeled on that of `submitToUrl`, is it better to remove this message? Or is there a more descriptive model to follow than `submitToUrl`?

src/applications/disability-benefits/526EZ/actions/index.js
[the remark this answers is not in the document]
markgreenburg: Might want to cover this functionality with a unit test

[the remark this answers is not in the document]
cehsu: Good idea, I'll pull this out as a helper and add a unit test for it.

src/applications/disability-benefits/526EZ/actions/index.js
```diff
@@ -1,64 +1,132 @@
...
+  none: 'none',
+  expired: 'expired',
+  created: 'created',
+  retrieved: 'retrieved',
+  renewed: 'renewed',
+  notRetrievedSaved: 'not-retrieved-saved',
+  notRetrievedNew: 'not-retrieved-new',
+  notCreated: 'not-created',
+  notRenewed: 'not-renewed',
+};
```
markgreenburg: Swagger for `/v0/intent_to_file` shows five statuses possible for each ITF entry in the array:
`[active, claim_recieved, duplicate, expired, incomplete]`

Whereas `PRESTART_STATUSES` seems to like a combination of ITF statuses (`active`, `expired`) and possible states for a veteran's overall ITF status (`notRetrievedSaved`, etc.).

Would it make sense to break these up for clarity?

cehsu: Interesting idea, so it looks like we'll stick closer with the original configuration and keep these in `status` (since these reflect application state), and then in `data` store the following to reflect what's returned from the endpoint: `currentStatus`, `currentExpirationDate`, and if we renew the ITF, also store `previousExpirationDate`. How does that sounds @cvalarida @markgreenburg ?

cehsu: Also instances where there are no `expired` ITF's should also be included in the `none` state, will update.

src/applications/disability-benefits/526EZ/actions/index.js
cvalarida: `submitToUrl()` creates a new `XMLHttpRequest` and sets event listeners for the different ways it could fail. These event listeners reject the promise with an error message that describes why the request "failed."

In this `apiRequest` failure callback, we don't have any insight as to _why_ the request failed, so it's probably not universally true that the `Network request failed`.

....well......Kind of. I guess it's _technically_ true that it failed for _some_ reason, but it uses the same language as `submitToUrl` without the same level of detail, which can be misleading if we had to debug this later.

That said, since we don't have any other possibilities (that is, `errorMessage` isn't going to be overridden with any other error message), we can remove it entirely and be safe without it. All we know is a certain request failed, and the error we log to sentry should reflect this.

src/applications/disability-benefits/526EZ/actions/index.js
cvalarida: I'd like to call out that it's specifically the `check` call. This should be distinct from the `create` call. As long as they have different names so we can track them separately, it sounds good to me.

src/applications/disability-benefits/526EZ/actions/index.js
cvalarida: Same story here; as long as we call out that this is specifically the submission or creation call, I'm okay with pretty much whatever (within reason, of course :wink: ).

src/applications/disability-benefits/526EZ/actions/index.js
```diff
@@ -1,64 +1,132 @@
...
+export function checkITFRequest(dispatch, hasSavedForm) {
+  return apiRequest(
+    '/intent_to_file',
+    null,
+    ({ data }) => {
+      let status;
+      let expirationDate;
+      const itfList = data.attributes.intentToFile;
+      // If the user does not have any existing ITFs, set none status
+      if (!itfList || (Array.isArray(itfList) && itfList.length === 0)) {
```
markgreenburg: Is it possible for `itfList` to exist but not be an array?

src/applications/disability-benefits/526EZ/reducer.js
cvalarida: I think checking the old state will take as much time as always setting the new state, so I'm not sure there is a whole lot to be gained here. Conversely, I think this will make it a bit harder to reason about. It sounds like we always want the `newState.display` to be true, so we should always set it to `true`.

I think @kreek's allergy to conditionals is contagious. I think I've caught it.

src/applications/disability-benefits/526EZ/reducer.js
cvalarida: So the state at any one time can look like
```
state: { stuff }
```
or
```
state: {
  previous: { stuff }
  current: { stuff }
}
```
Is that right?

Can we move the previous data out of `newState.data` and into something like `newState.previousData` to keep the structure consistent?

src/applications/disability-benefits/526EZ/actions/index.js
cehsu: Actually we never use the ITF status itself outside of the response handler, so we can simply store `currentExpirationDate` and `previousExpirationDate` in `data`.

src/applications/disability-benefits/526EZ/reducer.js
kreek: join [anti-if](https://francescocirillo.com/pages/anti-if-campaign) ✊- in this case _if_ it always  can be set to true without any adverse effects I would lose the conditional check just to reduce lines a follow developer needs to process

src/applications/disability-benefits/526EZ/actions/index.js
cehsu: @SMLuthi indicated these are the possible shapes of the data, any additional insights would be appreciated as well.

src/applications/disability-benefits/526EZ/reducer.js
cehsu: Yes, this is consistent with the other comments proposing to store currentExpirationDate and previousExpirationDate in data.

src/applications/disability-benefits/526EZ/reducer.js
cehsu: Sounds good, will update, thanks for sharing the link and the insights.

src/applications/disability-benefits/526EZ/reducer.js
cvalarida: Ahh, okay. That makes sense. Thanks!

src/applications/disability-benefits/526EZ/actions/index.js
cvalarida: I'm not sure I understand 100%, but I'm interested to see where it goes. It might be easier to code than to explain (as much as I hate to say it since if we can't explain our code clearly, we're probably complicating things).

src/applications/disability-benefits/526EZ/actions/index.js
```diff
@@ -1,64 +1,132 @@
...
-// };
-// }
-/* eslint-enable */
+export function verifyIntentToFile(hasSavedForm) {
+  return async (dispatch) => {
+    let submitSuccessStatus; // eslint-disable-line no-unused-vars
+    let submitErrorStatus;
+    dispatch(setPrestartStatus(PRESTART_STATUSES.pending));
 
+    const existingITFStatus = await checkITFRequest(dispatch, hasSavedForm);
```
markgreenburg: We expect this to be a string (`return status;`), but if the error handler in `checkITFRequest` throws, wouldn't we get an error back instead?

src/applications/disability-benefits/526EZ/actions/index.js
```diff
@@ -1,64 +1,132 @@
...
+    } else if (existingITFStatus === PRESTART_STATUSES.expired) {
+      submitSuccessStatus = PRESTART_STATUSES.renewed;
+      submitErrorStatus = PRESTART_STATUSES.notRenewed;
+    }
+    const newITFStatus = await submitITFRequest(dispatch, submitSuccessStatus, submitErrorStatus);
+
+    if (prestartFailureStatuses.has(newITFStatus)) {
+      return false;
+    }
+    return true;
```
markgreenburg: So it seems that we:
1. Check ITF status
2. If an ITF needs to be submitted, we try to submit one
3. We return a boolean from `verifyIntentToFile`, whose value depends on whether we get a success or error response string back from `submitITFRequest`

What happens if `verifyIntentToFile` returns `false`? Is this taken as an indication in the front end that an error has occurred and the veteran can't proceed to the form?

cehsu: We can remove these return statements since this is no longer passed along to gating logic within `fetchInProgressForm` (which has already been run by the time this is executed). It is still useful for testing purposes? Or is the logic simple enough that we can simply take these out.

cehsu: Vestigial returns removed.

src/applications/disability-benefits/526EZ/actions/index.js
cehsu: Good point, should we log these to sentry instead?

cehsu: It looks like apiRequest already logs the error, so no need to do so here. I'll remove the logging portion of this.

cehsu:
Thanks for your comments @cvalarida @markgreenburg, the updates discussed above have been made, PTAL, and thanks for any additional feedback.

src/applications/disability-benefits/526EZ/actions/index.js
```diff
@@ -1,64 +1,157 @@
...
+  );
+}
+
+export function verifyIntentToFile(hasSavedForm) {
+  return async (dispatch) => {
+    let submitSuccessStatus;
+    let submitErrorStatus;
+    dispatch(setPrestartStatus(PRESTART_STATUSES.pending));
+
+    const existingITFStatus = await checkITFRequest(dispatch, hasSavedForm);
```
markgreenburg: `checkITFRequest` first dispatches `setPrestartStatus(status)`, then returns `status`. The returned `status` string is then used to derive a payload for `submitITFRequest` within the `verifyIntentToFile` action creator.

My concern with this setup is that it might be possible to have inconsistencies in state crop up, due to the fact that we're both dispatching actions and at the same time using values returned from those action creators in other action creators that themselves make async calls using the same data. I'd be curious to hear @cvalarida 's thoughts on this as well.

src/applications/disability-benefits/526EZ/actions/index.js
```diff
@@ -1,64 +1,157 @@
...
+    }
+
+    if (existingITFStatus === PRESTART_STATUSES.none) {
+      submitSuccessStatus = PRESTART_STATUSES.created;
+      submitErrorStatus = PRESTART_STATUSES.notCreated;
+    } else if (existingITFStatus === PRESTART_STATUSES.expired) {
+      submitSuccessStatus = PRESTART_STATUSES.renewed;
+      submitErrorStatus = PRESTART_STATUSES.notRenewed;
+    }
+    submitITFRequest(dispatch, submitSuccessStatus, submitErrorStatus);
```
markgreenburg: Do we always want to `POST` an ITF when we `verifyIntentToFile`?

markgreenburg (commented):
Coming together nicely! Left some comments.

src/applications/disability-benefits/526EZ/reducer.js
```diff
@@ -1,6 +1,48 @@
...
+  PRESTART_STATUSES
+} from './actions';
+
+const initialState = {
+  status: PRESTART_STATUSES.notAttempted,
+  data: {
+    currentExpirationDate: null,
+    previousExpirationDate: null
+  },
+  display: false
```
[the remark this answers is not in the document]
cvalarida: @cehsu explained it to me in another comment: The idea behind this is to use it to show the success banner or not since we only want to show it on the first page that the veteran sees when they load the form. Once they move to the next page, the banner doesn't need to stay up.

src/applications/disability-benefits/526EZ/reducer.js
```diff
@@ -1,6 +1,48 @@
...
+  PRESTART_RESET,
+  PRESTART_DISPLAY_RESET,
+  PRESTART_STATUSES
+} from './actions';
+
+const initialState = {
+  status: PRESTART_STATUSES.notAttempted,
+  data: {
+    currentExpirationDate: null,
+    previousExpirationDate: null
```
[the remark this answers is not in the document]
cvalarida: I think the idea is to show the previous expiration date when they had an expired one and we had to submit a new one.

src/applications/disability-benefits/526EZ/actions/index.js
cehsu: Sounds good, `setPrestartStatus` moved into `verifyIntentToFile`.

src/applications/disability-benefits/526EZ/actions/index.js
cehsu: We only reach this point if `checkITFRequest` returns one of the pending status types (`none` or `expired`), in which case, we create a new ITF.

src/applications/disability-benefits/526EZ/actions/index.js
```diff
@@ -1,64 +1,162 @@
...
-//   );
-// };
-// }
-/* eslint-enable */
+export const handleCheckSuccess = (data, dispatch) => {
+  let status;
+  let expirationDate;
+  const itfList = data.attributes.intentToFile;
+  // If the user does not have any existing ITFs, set none status
+  if (!itfList || (Array.isArray(itfList) && itfList.length === 0)) {
```
cvalarida: I've been thinking about this and I'm wondering if the following might be a cleaner approach:

```js
if (!itfList.length) {
  return ITF_STATUSES.none;
}

// Grab the latest ITF
// This assumes some things about how ITFs are handled by EVSS (or VA writ large?)
const latestITF = getLatestITF(itfList);

// We only show information for active or expired ITFs
if (['active', 'expired'].includes(latestITF.status)) {
  return ITF_STATUSES.none;
}

const whichDate = latestITF.status === 'active' ? 'currentExpirationDate' : 'previousExpirationDate';
dispatch(setPrestartData({ [whichDate]: latestITF.expirationDate }));
return latestITF.status; // should be 'active' or 'expired', which should match properties in ITF_STATUSES
```

I'm not sure what about the approach you've got already that makes me uncomfortable, but maybe it's the level of `if`s. Does the above solve all the problems?

cvalarida (commented):
I just skimmed over the tests again, but from what I saw, they look okay. My two biggest comments are about splitting up the statuses into more discreet groups and how we can reduce the complexity of `handleCheckSuccess`. Like I mentioned in the comment, I'm not sure what makes me uncomfortable about it, but it seems like we can accomplish the same task more cleanly.

src/applications/disability-benefits/526EZ/actions/index.js
cvalarida: Actually, now that I think about it, if we keep what you've got there, what happens if the latest ITF is `claim_received`, but there's an `expired` earlier in the history? I think `handleCheckSuccess` will return `expired` erroneously (since they successfully submitted a claim after that).

src/applications/disability-benefits/526EZ/actions/index.js
```diff
@@ -1,64 +1,156 @@
...
 
-export const ITFStatuses = Object.freeze({
+import existingITFData from '../tests/itfData';
+
+import { getLatestTimestamp } from '../helpers';
+
+export const PRESTART_STATUS_SET = 'PRESTART_STATUS_SET';
+export const PRESTART_DATA_SET = 'PRESTART_DATA_SET';
+export const PRESTART_RESET = 'PRESTART_STATUS_RESET';
+export const PRESTART_DISPLAY_RESET = 'PRESTART_DISPLAY_RESET';
```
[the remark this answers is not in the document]
cehsu: I considered using ITF here but decided against it because it's an acronym that might not be as easy to maintain/conceptualize down the road, also we are actually retrieving multiple ITF, an array of them if the user has any, and then later, potentially creating one. It seems helpful to use prestart here to express that it is the gating check that this refers to, and not one of the various individual retrieved or created ITFs.

src/applications/disability-benefits/526EZ/actions/index.js
```diff
@@ -1,64 +1,162 @@
...
+const delay = (ms) => new Promise(resolve => setTimeout(resolve, ms));
+
+const fakeITFRequest = async (url, options, success) => {
+  await delay(1000);
+  const response = new Response();
+  response.data = existingITFData;
+  return success(response);
+};
+
+export function mockCheckITFRequest(dispatch, hasSavedForm) {
```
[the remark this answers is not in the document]
cehsu: We can test this on staging, it returns the mock data provided by the BE, which has also been saved as test data.

src/applications/disability-benefits/526EZ/actions/index.js
```diff
@@ -1,64 +1,162 @@
...
+    if (expiredList.length === 0 && activeList.length === 0) {
+      status = PRESTART_STATUSES.none;
+      // If the user doesn't have any active ITFs, set expired status
+    } else if (activeList.length === 0) {
+      status = PRESTART_STATUSES.expired;
+      expirationDate = getLatestTimestamp(expiredList.map(itf => itf.expirationDate));
+      dispatch(setPrestartData({ previousExpirationDate: expirationDate }));
+      // If the user has an active ITF, set retrieved status
+    } else {
+      status = PRESTART_STATUSES.retrieved;
```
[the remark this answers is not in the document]
cehsu: "active" on its own is insufficiently specific for the purposes of the UI, as there are three distinct active UI's/success messages: retrieved, created, and renewed.

src/applications/disability-benefits/526EZ/actions/index.js
cehsu: Thanks @cvalarida, that's pretty clean, will update according to your recommendation.

src/applications/disability-benefits/526EZ/actions/index.js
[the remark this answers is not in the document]
cehsu: @djmunoz1 indicated that we would want a mock for testing, I will add a comment to explain that this should be removed once that is done.

src/applications/disability-benefits/526EZ/tests/actions/actions.unit.spec.js
```diff
@@ -0,0 +1,312 @@
...
+
+function setFetchResponse(data) {
+  const response = {};
+  response.ok = true;
+  response.data = data;
+  response.headers = {
+    get: () => 'application/json' // for use by isJson in apiRequest
+  };
+  response.json = () => Promise.resolve(data);
+  mockFetch(response);
```
cehsu: Thanks for pointing this helper out @cvalarida !

src/applications/disability-benefits/526EZ/actions/index.js
cehsu: @cvalarida, how about splitting status up into status and message?
status types: not_attempted, pending, succeeded, failed
message types: retrieved, created, etc.

cehsu:
Thanks for your reviews so far @cvalarida @markgreenburg. I've updated the PR to incorporate feedback on what response will be received when a user doesn't have existing ITFs. Please take a look, and thanks for any additional questions, comments, ideas.

src/applications/disability-benefits/526EZ/tests/actions/actions.unit.spec.js
```diff
@@ -0,0 +1,312 @@
...
+  handleCheckSuccess,
+  handleCheckFailure,
+  handleSubmitSuccess,
+  handleSubmitFailure,
+  checkITFRequest,
+  submitITFRequest,
+  verifyIntentToFile
+} from '../../actions';
+
+function setFetchResponse(data) {
```
cvalarida: Can we use `mockApiRequest()` here?

src/applications/disability-benefits/526EZ/actions/index.js
cvalarida: Seems like messages should be derived from the state of the application as a whole, so I think `message_type` is perhaps to prescriptive? Like we're trying to define behavior, not state, but behavior should be derived _from_ state, y'know?

cehsu: Interesting, I had considered this as a type of [UI state](https://tonyhb.gitbooks.io/redux-without-profanity/content/ui_state.html), since it's primarily used to determine which message is displayed without having any other implications for the form. But I haven't thought about it more deeply than that, so if it makes sense to calculate these later than we can do that instead.

src/applications/disability-benefits/526EZ/tests/actions/actions.unit.spec.js
cehsu: Good idea, thanks @cvalarida, replaced.

src/applications/disability-benefits/526EZ/actions/index.js
```diff
@@ -1,64 +1,150 @@
...
-        });
-        reject(errorMessage);
-      }
-    });
-
-    return delay(2000);
+  retrieved: 'retrieved',
+  created: 'created',
+  notRetrieved: 'notRetrieved',
+  notCreated: 'notCreated'
```
cehsu: We need to track how the ITF was verified/failed to be verified, is the preference to store this in status or as another property such as `verificationType`, e.g. `verificationType: retrieved/created` @cvalarida @markgreenburg ?

src/applications/disability-benefits/526EZ/actions/index.js
```diff
@@ -1,64 +1,158 @@
...
-    });
-
-    return delay(2000);
+  succeeded: 'succeeded',
+  failed: 'failed'
+};
+
+export const PRESTART_TYPES = {
+  retrieve: 'retrieve',
+  create: 'create'
```
cehsu: @cvalarida @markgreenburg I decided to split the create or retrieve information out of the status and store it along with the rest of the data that is used to generate the message for the user. Thanks for your thoughts on this.

src/applications/disability-benefits/526EZ/actions/index.js
```diff
@@ -1,64 +1,158 @@
...
-        reject(errorMessage);
-      }
-    });
-
-    return delay(2000);
+  succeeded: 'succeeded',
+  failed: 'failed'
+};
+
+export const PRESTART_VERIFICATION_TYPES = {
```
cehsu: Type has been qualified as verification type.

src/applications/disability-benefits/526EZ/actions/index.js
```diff
@@ -1,64 +1,158 @@
...
+}
+
+// Returns most recent timestamp from a list of timestamps
+export const getLatestTimestamp = (timestamps) => timestamps.sort((a, b) => new Date(b) - new Date(a))[0];
+export const isNotEmptyObject = (object) => !(Object.keys(object).length === 0 && object.constructor === Object);
+export const getITFsByStatus = (itfs, status) => itfs.filter(itf => itf.status === status);
+
+export const handleCheckSuccess = (data, dispatch) => {
+  // If no existing ITF(s) are found, we will receive an empty {} response
+  const itfList = isNotEmptyObject(data) && data.attributes.intentToFile;
```
cvalarida: `data` is deconstructed from the response in `checkITFRequest`. In order for this to work, the response would have to look like
```
{
  data: {}
}
```
When there are no ITFs, the response is going to look like
```
{}
```
This means `data`, when it gets here, will be undefined, not an empty object. We should be able to do `const itfList = get('attributes.intentToFile', data)`.

src/applications/disability-benefits/526EZ/actions/index.js
```diff
@@ -1,64 +1,158 @@
...
+const delay = (ms) => new Promise(resolve => setTimeout(resolve, ms));
+
+const fakeITFRequest = async (url, options, success) => {
+  await delay(1000);
+  const response = new Response();
+  response.data = existingITFData;
+  return success(response);
+};
+
+// TODO: remove this mock and its helpers once user testing is complete
```
cvalarida: Ahh, that's why it's still in here. Makes sense. Thanks for the clarification! :raised_hands:

src/applications/disability-benefits/526EZ/tests/actions/actions.unit.spec.js
```diff
@@ -0,0 +1,275 @@
...
+  handleSubmitSuccess,
+  handleSubmitFailure,
+  checkITFRequest,
+  submitITFRequest,
+  verifyIntentToFile
+} from '../../actions';
+
+import existingData from '../itfData';
+
+const noData = {};
```
cvalarida: Same as an above comment; since this represents what's being passed to the `handleCheckSuccess` function, `const noData = undefined;` is more accurate. If `get()` is used, it'll work either way, but may be a bit misleading if we test `data` as an empty object when it never will be, y'know?

src/applications/disability-benefits/526EZ/tests/actions/actions.unit.spec.js
```diff
@@ -0,0 +1,275 @@
...
+
+      handleSubmitFailure(dispatch);
+      expect(dispatch.calledWith(setPrestartStatus(PRESTART_STATUSES.failed))).to.be.true;
+    });
+  });
+  describe('checkITFRequest', () => {
+    afterEach(() => resetFetch());
+    it('should handle success', (done) => {
+      const dispatch = sinon.spy();
+      mockApiRequest({ data: existingData });
```
cvalarida: `mockApiRequest()` overrides the `global.fetch`, so we should probably grab `const originalFetch = global.fetch` before the first `describe()` and then put in an `after(() => global.fetch = originalFetch);` in each `describe()` that calls `mockApiRequest()` to avoid any problems in unrelated tests.

cvalarida (approved):
There are a few things left, but after they get fixed up, I think it's good to go. As a heads up, we may find we have to refactor once we actually implement this, but the bones look good.

src/applications/disability-benefits/526EZ/actions/index.js
cehsu: Thanks for double checking this @cvalarida, @SMLuthi could you please explain what we will expect data to be when the response body is an empty object? Will the BE serialize this as an empty object as well? Or as undefined?

cehsu: Ah since we're using apiResponse, it won't use the `data` attribute on the response (which will be the serialized body) in this case. I wonder if we would want to update apiRequest to return the serialized data if it is already present, since that seems fairly standard. I'll update as you recommend.

SMLuthi: This is what a call to `/v0/intent_to_file` will look like if the user has no itfs and the response is `{}`

```
{
    "data": {
        "attributes": {
            "intent_to_file": []
        },
        "id": "",
        "type": "evss_intent_to_file_intent_to_files_responses"
    }
}
```

SMLuthi: Any response from vets-api that is serialized will follow the above format.

cehsu: Thanks so much @SMLuthi, will update accordingly.

src/applications/disability-benefits/526EZ/tests/actions/actions.unit.spec.js
cehsu: Thanks @cvalarida, updated.

PR #7872, opened 2018-06-13
Profile 360 - Restructure 360 fields into containers; add 360-dedicated reducer

ncksllvn:
- Restructure Contact Information fields into Vet360 containers
  - There was too much duplicate and confusing logic between fields.
- Adds reducer for Vet360
  - transactions: Array<string>
  - fieldTransactionMap: object - this is a map of field names to transaction IDs, which will be used for looking up transaction information for a given field prior to a browser refresh. This way we can keep the transaction property pure and in line with the API, and easily recover when the page is refreshed and we no longer have that data.
- Adds messaging for finished transactions
    - Renders error/success messaging according to status
    - Clearing the message will clear the transaction from state
- Adds action for retrieving all transactions on initial render

Resolves department-of-veterans-affairs/vets.gov-team/issues/10979
Resolves department-of-veterans-affairs/vets.gov-team/issues/10928
Resolves department-of-veterans-affairs/vets.gov-team/issues/10862


src/applications/personalization/profile360/containers/Vet360ProfileField.jsx
```diff
@@ -0,0 +1,206 @@
...
+Vet360ProfileFieldContainer.propTypes = {
+  analyticsSectionName: PropTypes.string.isRequired,
+  fieldName: PropTypes.oneOf(Object.values(VET360.FIELD_NAMES)).isRequired,
+  renderContent: PropTypes.func.isRequired,
+  renderEditModal: PropTypes.func.isRequired,
+  title: PropTypes.string.isRequired
+};
+
+export default Vet360ProfileFieldContainer;
+export { Vet360ProfileField };
```
ncksllvn: This is the main point of all of the changes - consolidating all of the common 360 field behavior into a container.

ncksllvn:
cc @bshyong this isn't ready yet but wanted to get your eyes on it sooner rather than later. This will hopefully make things more manageable.

src/applications/personalization/profile360/actions/misc.js
```diff
@@ -3,21 +3,11 @@ import { MILITARY_STATES } from '../../../letters/utils/constants';
...
 }
 
-export function clearErrors() {
-  return { type: CLEAR_PROFILE_ERRORS };
-}
-
-export function clearMessage() {
-  return { type: CLEAR_MESSAGE };
-}
-
```
ncksllvn: Messaging is driven by transaction statuses, so we no longer need to manually dispatch messaging actions.

src/applications/personalization/profile360/util/transactions.js
```diff
@@ -0,0 +1,53 @@
...
+const SUCCESS_STATUSES = new Set([
+  VET360.TRANSACTION_STATUS.COMPLETED_SUCCESS,
+  VET360.TRANSACTION_STATUS.COMPLETED_NO_CHANGES_DETECTED
+]);
+
+const ERROR_STATUSES = new Set([
+  VET360.TRANSACTION_STATUS.COMPLETED_FAILURE,
+  VET360.TRANSACTION_STATUS.REJECTED
+]);
+
```
ncksllvn: I'm not totally sure where some of the statuses fall within these

bshyong (approved):

PR #7878, opened 2018-06-14
import from moment startup to fix unstable test

wyattwalter:
This fixes a test I've observed failing a few times on the format of the date string coming out of `moment.format()`. It's reproducible with `CHOMA_SEED=81fYlOF4gS npm --no-color run test:coverag`. I'll admit that I'm not 100% sure why this ever works except that it must be some sort of ordering issue where the short month name defaults are being set in another test or script.

We seem to include moment fairly inconsistently. Sometimes we use this method which injects some settings overrides, sometimes we inject these settings directly in the file. There may or may not be a good reason to keep the current method, I didn't address that part of the issue here.

### DIFF
--- a/src/applications/claims-status/components/ClaimsDecision.jsx
+++ b/src/applications/claims-status/components/ClaimsDecision.jsx
@@ -1,6 +1,6 @@
 import PropTypes from 'prop-types';
 import React from 'react';
-import moment from 'moment';
+import moment from '../../../platform/startup/moment-setup';
 
 class ClaimsDecision extends React.Component {
   render() {


jbalboni:
I'm a little hesitant to start sprinkling this in different places. It should really be included somewhere globally. What happens if we import moment-setup in mocha-setup.js?

wyattwalter:
@jbalboni I share your concern about the inconsistency, but am a bit out of my element. Wouldn't putting it into mocha-setup.js only affect tests but not the build process? That does let the test suite pass.

jbalboni:
Right, I think we'll also need to put it somewhere in the site wide start up code, but putting it in mocha-setup will resolve the ordering issue.

cvalarida (approved):
:shipit: Tested locally. Fails without this, works with it. Looks good!

wyattwalter:
Really don't want to punt on what Jeff said, but I haven't had time to look into that and this is causing failures almost every day. Going to merge this as-is and I dropped a ticket into backlog to make this more consistent soon. https://github.com/department-of-veterans-affairs/vets.gov-team/issues/11157

PR #7879, opened 2018-06-14
Fix url path in pension upload config

jbalboni:
Discovered this while troubleshooting e2e tests, confirmed that it's causing errors in production.


annekainicUSDS (approved):
LGTM!

rianfowler (approved):

PR #7885, opened 2018-06-15
Remove DOB field from 686

jbalboni:
Date of birth isn't collected by the 686 form for the applicant, so we don't need to include it.

It does make the veteran info page a little short when you're applying as the veteran.

screen shot 2018-06-15 at 9 20 11 am


emilywaggoner (approved):
That works - thanks Jeff!

lbai001 (approved):

PR #7888, opened 2018-06-15
VA Facilities Test Fix

cvalarida:
To be perfectly honest, I'm not sure _why_ this works. In fact, I'm not sure _how_ @jbalboni figured out which tests were the offenders in the first place. After I got that information, I had to ponder about this for a while.

It had to be some side effect that was causing completely unrelated tests to fail. The only side effects I could think of here were filling out the treatment center name; that would query the api. Since it's not mocked, it'd try to _actually_ query the api.

Another solution would be to set up a little mock server like we have for the e2e tests, but it seems like that's the wrong approach for unit tests (less isolation and all that).


cehsu (approved):
LGTM.

markgreenburg (approved):
LGTM!

PR #7895, opened 2018-06-18
Profile 360: Consolidate Update Modal behavior

ncksllvn:
This continues the work done in https://github.com/department-of-veterans-affairs/vets-website/pull/7872 by consolidating the edit-modal behavior into a Vet360EditModal component. 

The accessibility issue we saw - where the form fields weren't receiving browser focus - was a result of the modal opening before the form fields rendered. The form fields were waiting for the form values to finish initializing in the redux store, which was dispatched during `componentDidMount` but not before the initial render. The modal rendered anyway and finished rendering the form fields when they were ready, but that was causing them not to receive browser focus.

Resolves department-of-veterans-affairs/vets.gov-team/issues/10497


src/applications/personalization/profile360/components/AddressEditModal.jsx
```diff
@@ -1,24 +1,16 @@
 import React from 'react';
-import { kebabCase } from 'lodash';
```
ncksllvn: Do we need this behavior? We can pass it on down to the Vet360EditModal component as an ID and apply it there, but I didn't see what we were using it for.

src/applications/personalization/profile360/components/PhoneSection.jsx
```diff
@@ -4,36 +4,20 @@ import PhoneNumberWidget from 'us-forms-system/lib/js/review/PhoneNumberWidget';
...
-      onChange={onChange}
-      phoneData={phoneData}
-      onSubmit={onSubmit}
-      isLoading={transactionRequest && transactionRequest.isPending}
-      onCancel={onCancel}
-      onDelete={onDelete}/>
-  );
-}
-
 export default function Vet360Phone({ title, fieldName, analyticsSectionName }) {
```
ncksllvn: All edit-modals follow the same API as the corresponding content and Vet360ProfileField, so we don't need a middle function that maps props to the edit-modal. The only thing to know here is that the transactionRequest is now exposed to the edit-modal, and that phoneData/addressData/emailData are all now referred to as just data, `this.props.data`.

src/applications/personalization/profile360/components/Vet360EditModal.jsx
```diff
@@ -0,0 +1,87 @@
...
+      isEmpty,
+      isInitialized,
+      props: {
+        onCancel,
+        title,
+        clearErrors,
+        render,
+        onDelete,
+        transactionRequest
+      }
```
ncksllvn: props here is the same props as the higher-level Vet360ProfileField, but it only uses a subset.

src/applications/personalization/profile360/components/AddressEditModal.jsx
bshyong: This was for analytics event names; okay with removing here. We can add back in as needed

bshyong (approved):

PR #7896, opened 2018-06-19
526 use updated address mg

markgreenburg:
- Updates address schema refs
- Breaks out constants into own file
- Updates PMR release to be commented out pending 4142
- Adds fancier options list switching for military vs. non-military addresses


src/applications/disability-benefits/526EZ/helpers.jsx
```diff
@@ -13,17 +13,15 @@ import { apiRequest } from '../../../platform/utilities/api';
...
-
-export const USA = 'USA';
-export const MILITARY_STATES = ['AA', 'AE', 'AP'];
-export const MILITARY_CITIES = ['APO', 'DPO', 'FPO'];
+import {
+  USA,
+  VA_FORM4142_URL
+} from './constants';
 
+const siblings = ['treatments', 'privateRecordReleases', 'privateRecords', 'additionalDocuments'];
```
cehsu: Should we use something like "uploadItems" or "documentTypes" here instead?

src/applications/disability-benefits/526EZ/pages/primaryAddress.js
```diff
@@ -37,364 +45,30 @@ function validateMilitaryState(errors, state, formData, schema, messages, option
...
-  WV: 'West Virginia',
-  WI: 'Wisconsin',
-  WY: 'Wyoming',
-  PI: 'Rizal state',
-  UM: 'United States Minor Outlying Islands'
-};
+const addressUISchema = (addressName, title) => {
+  const updateStates = (form) => {
+    const currentCity = _.get(form, `veteran.${addressName}.city`, '').trim().toUpperCase();
+    if (MILITARY_CITIES.includes(currentCity)) {
```
cehsu: Should we check that a state value doesn't already exist? So if there's no state value, restrict the enums, but if there's already a state value, simply provide a validation message upon submit instead?

cehsu (commented):
Looks great so far. I had a question about how the UI should work when a military city is selected.

src/applications/disability-benefits/526EZ/helpers.jsx
markgreenburg: No idea 😛 This is actually existing code that just moved down a few lines. I have to admit this is kind of a messy diff 😞

cehsu: Thanks for pointing that out, I wonder whether we should keep in mind the work that @goldenmeanie is doing to plan the document upload UI, and the types of variable names we are likely to want to use for that. Right now, the name used in the common fileUploadUI is `documentId` (for document type), so maybe avoiding documentXYZ, in favor or recordType or recordXYZ might be helpful to distinguish the meaning here? Not sure, naming is hard.

src/applications/disability-benefits/526EZ/pages/primaryAddress.js
markgreenburg: Good question. So here are the things that happen:
- When no state is present and the user enters a military city code:
    1. The enum list is restricted
    2. The 'empty' state option remains selected
    3. No validation errors fire yet, because the state field hasn't been touched
- When state is a non-military state and the user enters a military city code
    1. The enum list is restricted
    2. The selected state option is cleared and set to the empty option
    3. No validation errors fire yet, because the state field hasn't been touched
    4. Once the state field has been touched and the veteran has tabbed over it (i.e., kept the 'empty' selection and kept going to the other fields in the form), a validation message is shown
- When state is a non-military state, the city is changed to a military city code, and the state field has already been touched
    1. The enum list is restricted
    2. The selected state option is cleared and set to the empty option
    3. The validation error fires immediately, prompting the user to select a valid state now that they've changed the city to a military city code

I think this behavior is actually what I'd want, as a user. Don't bug me before I've had a chance to fill out the field, but let me know that I've done something invalid once I've selected an option that invalidates a previous selection in another field.

src/applications/disability-benefits/526EZ/helpers.jsx
markgreenburg: Yeah, I'd probably vote to split out renaming this into whatever PR / issue we end up making for the rest of the document upload work.

src/applications/disability-benefits/526EZ/pages/primaryAddress.js
cehsu: Thanks for listing it out @markgreenburg, I'm mostly curious about clearing user input, since I'm not familiar with us doing that elsewhere for situations where the field is the same, but that doesn't mean it's necessarily something we have to avoid. As a user I might be slightly confused why a field I thought was filled is suddenly blank and has a validation error. Maybe just providing the validation error, once the military city is selected could be less jarring in that case?

src/applications/disability-benefits/526EZ/constants.jsx
```diff
@@ -0,0 +1,91 @@
...
+const STATE_VALUES = PCIU_STATES.map(state => state.value);
+
+const MILITARY_STATE_VALUES = ['AA', 'AE', 'AP'];
+const MILITARY_STATE_LABELS = [
+  'Armed Forces Americas (AA)',
+  'Armed Forces Europe (AE)',
+  'Armed Forces Pacific (AP)'
+];
+
+const MILITARY_CITIES = ['APO', 'DPO', 'FPO'];
```
cehsu: Do/should these have accompanying labels that are similarly formatted with the military states (with parenthetical abbreviations)?

src/applications/disability-benefits/526EZ/tests/config/primaryAddress.unit.spec.jsx
```diff
@@ -78,6 +79,50 @@ describe('Disability benefits 526EZ primary address', () => {
     expect(form.find('input').length).to.equal(7);
   });
 
+  it('restricts state options to military state codes when city is a military city code', () => {
```
cehsu: Should we test the different flows you mentioned in your comment above (immediate validation vs. validation on submit) if we aren't already?

cehsu (approved):
LGTM so far. I am curious what the precedent is for reseting a form field from a UX perspective, in case @goldenmeanie or anyone else is able to provide insight into that. Also added a comment about test coverage, in case that makes sense to do here.

src/applications/disability-benefits/526EZ/pages/primaryAddress.js
markgreenburg: I dunno; the behavior in the existing approach is basically what comes with Forms out of the box. The implementation matches the [Forms cookbook example](https://github.com/department-of-veterans-affairs/vets-website/blob/master/docs/schemaform/form-config.md#i-want-to-change-the-options-of-a-dropdown-based-on-some-other-field-data). Not sure if you've tried to run this branch locally, but if not I'd be curious to hear if you still think the behavior is not intuitive after giving it a spin.
I see where you're coming from, though - one thing to keep in mind is that the city input comes before the state input, so in typical usage, the thing you just entered wouldn't disappear from the form just because the next thing you entered invalidates it.

For what it's worth, the Profile address widget and Letters app both clear non-military states once you enter a military city as well. Now that I think about it, I'm not certain how you'd restrict the options dropdown without clearing the existing option - I think you'd have to add it to the list dynamically, which could be strange (e.g., the list could look like `['Armed Forces Americas', ... , 'Michigan']`).

cvalarida (approved):
Had just one minor question. Looks good!

PR #7897, opened 2018-06-19
Error messages for auth flow failures

U-DON:
Resolves department-of-veterans-affairs/vets.gov-team#9548.

> screen shot 2018-06-19 at 8 18 39 am


dhall (approved):
LGTM -- we may want to coordinate our deploys, though this seems like it would be backwards compatible with the default case in place

PR #7898, opened 2018-06-19
Profile360: Add code-specific error messaging for failed/rejected transactions 

ncksllvn:
This adds the utility functions for inspecting transaction error codes and determining their meaning, then implements those within the top-level transaction error banner and the inline field error message.

Resolves department-of-veterans-affairs/vets.gov-team/issues/10854
Resolves department-of-veterans-affairs/vets.gov-team/issues/10944
Resolves department-of-veterans-affairs/vets.gov-team/issues/11032

### DIFF
--- a/src/applications/personalization/profile360/components/Vet360TransactionErrorBanner.jsx
+++ b/src/applications/personalization/profile360/components/Vet360TransactionErrorBanner.jsx
@@ -3,7 +3,6 @@ import AlertBox from '@department-of-veterans-affairs/formation/AlertBox';
 
 import {
   hasGenericUpdateError,
-  hasLowCondidenceAddressError,
   hasMVIError,
   hasMVINotFoundError,
   hasUserIsDeceasedError
@@ -17,19 +16,6 @@ export function GenericUpdateError(props) {
         <div>
           <h4>Your recent profile update didn’t save</h4>
           <p>We’re sorry. Something went wrong on our end and we couldn’t save the recent updates you made to your profile. Please try again later.</p>
-        </div>
-      )}/>
-  );
-}
-
-export function LowConfidenceError(props) {
-  return (
-    <AlertBox {...props}
-      status="error"
-      content={(
-        <div>
-          <h4>Your address didn’t save</h4>
-          <p>We’re sorry. We looked up the address you entered, and we’re not sure mail can be delivered there. Please try entering your address again.</p>
         </div>
       )}/>
   );
@@ -90,10 +76,6 @@ export default function Vet360TransactionErrorBanner({ transaction, clearTransac
 
     case hasMVIError(transaction):
       TransactionError = MVIError;
-      break;
-
-    case hasLowCondidenceAddressError(transaction):
-      TransactionError = LowConfidenceError;
       break;
 
     case hasGenericUpdateError(transaction):
--- a/src/applications/personalization/profile360/components/Vet360TransactionErrorBanner.jsx
+++ b/src/applications/personalization/profile360/components/Vet360TransactionErrorBanner.jsx
@@ -1,0 +1,105 @@
+import React from 'react';
+import AlertBox from '@department-of-veterans-affairs/formation/AlertBox';
+
+import {
+  hasGenericUpdateError,
+  hasLowCondidenceError,
+  hasMVIError,
+  hasMVINotFoundError,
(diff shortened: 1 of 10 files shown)


src/applications/personalization/profile360/components/Vet360TransactionErrorBanner.jsx
```diff
@@ -0,0 +1,105 @@
...
+      TransactionError = LowConfidenceError;
+      break;
+
+    case hasGenericUpdateError(transaction):
+    default:
+      TransactionError = GenericUpdateError;
+  }
+
+  return <TransactionError onCloseAlert={clearTransaction} isVisible/>;
+}
```
ncksllvn: I felt this syntax was clearer than a series of if-else but I think I could see the argument of `switch(true)` being bad practice.

src/applications/personalization/profile360/components/Vet360TransactionPending.jsx
```diff
@@ -1,13 +1,25 @@
...
     window.clearInterval(this.interval);
   }
+
   render() {
-    return <div>We’re working on saving your new {this.props.title.toLowerCase()}. We’ll show it here once it’s saved.</div>;
+    return (
+      <div className="vet360-profile-field-transaction-pending">
+        We’re working on saving your new {this.props.title.toLowerCase()}. We’ll show it here once it’s saved.
+      </div>
+    );
```
ncksllvn: This just makes the messaging for a pending transaction appear in blue text.

src/applications/personalization/profile360/selectors.js
```diff
@@ -34,10 +33,8 @@ export function selectVet360SuccessfulTransactions(state) {
   return state.vet360.transactions.filter(isSuccessfulTransaction);
 }
 
-export function selectVet360ErroredTransactions(state) {
-  return state.vet360.transactions.filter(transaction => {
-    return isFailedTransaction(transaction) || isErroredTransaction(transaction);
-  });
+export function selectVet360FailedTransactions(state) {
+  return state.vet360.transactions.filter(isFailedTransaction);
 }
```
ncksllvn: This selector is the top-level filter for the transaction error banners. It makes the assumption that any transaction that requires an error alert banner to show must have a status of failed or rejected. I believe this is a safe assumption, but BE may have to weigh-in if not.

cc @rlbaker @hpjaj @emersonthis

src/applications/personalization/profile360/util/transactions.js
```diff
@@ -36,6 +36,26 @@ const GENERIC_ERROR_CODES = new Set([
...
+]);
+
+const LOW_CONFIDENCE_ADDRESS_ERROR_CODES = new Set([
+  'ADDR306'
+]);
+
+const DECEASED_ERROR_CODES = new Set([
+  'VET360_MVI300'
+]);
+
```
ncksllvn: I just made all of these Sets in case new error codes with similar meaning are added in the future.

src/applications/personalization/profile360/util/transactions.js
```diff
@@ -36,6 +36,26 @@ const GENERIC_ERROR_CODES = new Set([
...
+const MVI_ERROR_CODES = new Set([
+  'VET360_MVI100',
+  'VET360_MVI101',
+  'VET360_MVI200',
+  'VET360_MVI202',
+  'VET360_MVI203'
+]);
+
+const LOW_CONFIDENCE_ADDRESS_ERROR_CODES = new Set([
+  'ADDR306'
```
hpjaj: This one will actually be `'VET360_ADDR306'`.

ncksllvn:
Need to a bit of sprucing up as a result of the insight from https://github.com/department-of-veterans-affairs/vets.gov-team/issues/11032#issuecomment-398184984

ncksllvn:
@bshyong PTAL when you get the chance.

src/applications/personalization/profile360/components/Vet360TransactionErrorBanner.jsx
bshyong: I usually lean towards if/else because switch is more verbose but don't have a strong opinion

bshyong (approved):

bshyong (approved):

PR #7900, opened 2018-06-19
Centralize localStorage access into module

ncksllvn:
#### What does this do?
This PR creates a module, `localStorage`, which either exports `window.localStorage` or a data structure that follows the `localStorage` API but does something more complex internally depending on whether the browser has `localStorage` available.

#### What problem does this solve?
If the user has their browser configured for some heavy-duty privacy, `localStorage` will not be available, and will throw a `SecurityError` if we attempt to access it. Originally, we thought this was limited to IE 10 Protected Mode, but any browser can be configured this way. 

To replicate this issue in FireFox, navigate to `about:config`, search `dom.storage.enabled`, and set that property to `false`. You should see a `SecurityError` thrown when trying to access `localStorage`. 

An example of the Sentry error described above can be found here - http://sentry.vetsgov-internal/vets-gov/website-production/issues/39240/.

#### Related issues
Replicating this problem in Firefox also reveals that we see the same error when working with `sessionStorage`. We should centralize access to that data as well. To limit the scope of this PR, this issue was logged here - department-of-veterans-affairs/vets.gov-team/issues/11224.

Resolves department-of-veterans-affairs/vets.gov-team/issues/10807


ncksllvn:
Ugh, so this successfully checks for localStorage and attempts to polyfill, but localStrage is apparently still a read-only property so it errors out.

image

ncksllvn:
Any ideas? cc @cehsu @jbalboni

ncksllvn:
This can be duplicated for the most part by running Firefox with storage totally disabled. Same result though.

image

I'm thinking that we'll need to centralize localStorage/sessionStorage as a module.

src/platform/utilities/storage/localStorage.js
```diff
@@ -0,0 +1,17 @@
...
+  try {
+    const testItem = 'testItem';
+    const localStorage = window.localStorage;
+    localStorage.setItem(testItem, testItem);
+    localStorage.removeItem(testItem);
+    return localStorage;
+  } catch (err) {
+    return require('local-storage-fallback').storage;
+  }
+}
```
ncksllvn: I wasn't able to use Modernizr because I wanted our unit tests to follow the pattern of importing the localStorage module and I was receiving a requirejs error there.

content/includes/footer.html
```diff
@@ -129,7 +129,7 @@ <h3>Get help from Veterans Crisis Line</h3>
 -->
 <script type="text/javascript">
   function footerNavClick(elem) {
-    const eventLabel = 'nav-' + elem.parentElement.parentElement.id
+    var eventLabel = 'nav-' + elem.parentElement.parentElement.id;
     recordEvent({
```
ncksllvn: `const` causes a syntax error in IE.

package.json
```diff
@@ -218,6 +218,7 @@
     "eslint-stats": "^1.0.0",
     "fast-levenshtein": "^2.0.6",
     "file-loader": "^1.1.11",
+    "local-storage-fallback": "^4.1.1",
     "modernizr": "^3.6.0",
```
ncksllvn: This module inspects the browser and exports a data structure that always follows the localStorage API but may use a different storage method internally (such as cookies, or in-memory) depending on what features are available in the browser.

src/applications/discharge-wizard/containers/GuidancePage.jsx
```diff
@@ -2,6 +2,7 @@ import { connect } from 'react-redux';
 import React from 'react';
 import moment from 'moment';
 import recordEvent from '../../../platform/monitoring/record-event';
+import localStorage from '../../../platform/utilities/storage/localStorage';
 
 import AlertBox from '@department-of-veterans-affairs/formation/AlertBox';
```
ncksllvn: In most cases, we're accessing localStorage without the formal `window` prefix, so we can just override the local variable `localStorage` in a top-level import.

src/applications/personalization/dashboard/tests/containers/DashboardApp.unit.spec.jsx
```diff
@@ -1,22 +1,27 @@
...
     loa: {}
   }
 };
 
 describe('<DashboardApp>', () => {
-  it('should render', () => {
-    window.localStorage = {
-      getItem: () => {}
-    };
+  before(() => {
```
ncksllvn: I'm actually unsure of why this wasn't causing an error previously, as the former value of localStorage isn't being restored.

src/platform/site-wide/announcements/actions/index.js
```diff
@@ -1,10 +1,12 @@
...
+
 export const INIT_DISMISSED_ANNOUNCEMENTS = 'INIT_DISMISSED_ANNOUNCEMENTS';
 export const DISMISS_ANNOUNCEMENT = 'DISMISS_ANNOUNCEMENT';
 
 export const ANNOUNCEMENTS_LOCAL_STORAGE = 'DISMISSED_ANNOUNCEMENTS';
 
 function localAnnouncements(dismissedAnnouncementName) {
-  const fromLocalStorage = window.localStorage[ANNOUNCEMENTS_LOCAL_STORAGE];
+  const fromLocalStorage = localStorage.getItem(ANNOUNCEMENTS_LOCAL_STORAGE);
   let parsed = [];
```
ncksllvn: I always preferred to work with localStorage as a plain object instead of using the `getItem`/`setItem` functions, but as a module we need to use the latter because the data structure may not be the default `localStorage`.

ncksllvn:
Tested and confirmed to be fixed in Firefox. However, we see the same error occurring with `sessionStorage`. This will be addressed in a later PR.

image

src/platform/utilities/storage/localStorage.js
```diff
@@ -0,0 +1,29 @@
...
+function getLocalStorage() {
+  try {
+    const testItem = 'testItem';
+    const localStorage = window.localStorage;
+    localStorage.setItem(testItem, testItem);
+    localStorage.removeItem(testItem);
+    return localStorage;
+  } catch (err) {
+    // Probably a SecurityError
+    return require('local-storage-fallback').storage;
```
jbalboni: I think you may as well use `import` and always import this module. I believe Webpack is always going to include this module in the bundle anyway.

src/applications/discharge-wizard/containers/GuidancePage.jsx
jbalboni: We might want to add a ticket to look into enforcing this with eslint. It's going to be easy to forget to import this module instead of using the global. But I think we could either write a custom rule or remove localStorage from the allowed globals list in eslint.

jbalboni (approved):

src/platform/utilities/storage/localStorage.js
ncksllvn: Good idea, fixed!

src/applications/discharge-wizard/containers/GuidancePage.jsx
ncksllvn: Great idea, moved to this ticket - department-of-veterans-affairs/vets.gov-team/issues/11283

PR #7903, opened 2018-06-19
526 -- Change all imports to schemaform library

cvalarida:
Seems to work alright. I renamed `common/` to `redacted/` to test out what would happen if we removed that whole folder. It broke on some profile code, so I think we'll need to address that in a separate PR.

What I did for this was just `grep schemaform src/applications/disability-benefits/526EZ/ -R` and made my way through each hit. I changed the imports according to where I thought they'd be (form library or platform directory) and ran the through the form (pretty quickly to test basic functionality). Seems to work alright, but I could use some more in-depth QA.


annekainicUSDS (approved):
LGTM! Haven't tested it locally, but the changes look right.

cvalarida:
Ahh, good call, @jbalboni. I'll take a look. :+1:

PR #7904, opened 2018-06-20
Vet360 profile updates for address/phone

bshyong:
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/11014
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/10927
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/11048

Makes various changes to address and phone number updates


src/applications/personalization/profile360/components/PhoneEditModal.jsx
```diff
@@ -2,9 +2,27 @@ import React from 'react';
 import { merge } from 'lodash';
 
 import ErrorableTextInput from '@department-of-veterans-affairs/formation/ErrorableTextInput';
+import AlertBox from '@department-of-veterans-affairs/formation/AlertBox';
 
 import Vet360EditModal from './Vet360EditModal';
 
+class PhoneTextInput extends ErrorableTextInput {
+  componentDidMount() {
```
bshyong: this is obtuse but unfortunately the most straightforward way to make this styling change to the inputs without copying the ErrorableTextInput component out of the design system library

ncksllvn: Yeah, a good example of how we need more flexibility in our design system's form components. They could use a lot of sprucing up IMO. There should also be a more graceful way to apply validation.

src/applications/personalization/profile360/components/Address.jsx
```diff
@@ -6,7 +6,7 @@ import { focusElement } from '../../../../platform/utilities/ui';
 
 import ErrorableSelect from '../../../letters/components/ErrorableSelect';
 import ErrorableTextInput from '../../../letters/components/ErrorableTextInput';
-import { STATE_CODE_TO_NAME, MILITARY_CITIES } from '../../../letters/utils/constants';
+import { STATE_CODE_TO_NAME, MILITARY_CITIES, MILITARY_STATES } from '../../../letters/utils/constants';
 import { militaryStateNames } from '../../../letters/utils/helpers';
```
ncksllvn: I'm not sure if it's changed since, but I was trying to centralize the address work into `platform/forms/address`. Not sure if it has what you need in there.

src/applications/personalization/profile360/components/EmailSection.jsx
```diff
@@ -7,7 +7,7 @@ function EmailView({ data: emailData }) {
   return <span>{emailData.emailAddress}</span>;
 }
 
-export default function Vet360Email({ title = 'email address', fieldName = 'email', analyticsSectionName = 'email' }) {
+export default function Vet360Email({ title = 'Email address', fieldName = 'email', analyticsSectionName = 'email' }) {
   return (
```
ncksllvn: Not relevant to this PR, but once we start adding analytics I'm in favor of removing that `analyticsSectionName` and just using the `kebabCase` or some type of transformer on the field name.

ncksllvn (approved):

src/applications/personalization/profile360/components/EmailSection.jsx
bshyong: Yep I will check to see how the names line up

PR #7906, opened 2018-06-20
Profile 360: Tweaks

ncksllvn:
- Update default field value for email's edit modal
- Share the isEmpty method between the field container and the edit modal
- Lowercase field title in edit modal

### DIFF
--- a/src/applications/personalization/profile360/components/EmailEditModal.jsx
+++ b/src/applications/personalization/profile360/components/EmailEditModal.jsx
@@ -14,7 +14,7 @@ export default class EditEmailModal extends React.Component {
       return { ...this.props.data };
     }
     return {
-      email: ''
+      emailAddress: ''
     };
   }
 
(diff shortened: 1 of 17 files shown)


src/applications/personalization/profile360/components/EmailEditModal.jsx
```diff
@@ -14,7 +14,7 @@ export default class EditEmailModal extends React.Component {
       return { ...this.props.data };
     }
     return {
-      email: ''
+      emailAddress: ''
```
bshyong: This is the correct value based on the object we need to set to Vet360, but the name in the modal reducer is `email`; down the line we may want to make these consistent for clarity

bshyong (approved):

src/applications/personalization/profile360/components/EmailEditModal.jsx
ncksllvn: Yeah for sure. Any way we can consolidate the terminology sounds great to me. 

This is a tricky one because `email` comes from the profile's Vet360 contact information, and `email.emailAddress` is how the actual value is accessed. Confusing.

PR #7909, opened 2018-06-21
Remove all refs to schemaform

annekainicUSDS:
3 main changes in this PR:

1. There were a few straggling references to `common/schemaform`, so I fixed those before deleting the entire directory.
2. I also moved the unit tests for the SiP code into the `platform` directory so the entire `test/common` directory can be deleted.
3. Also removed the docs because those have been moved into us-forms-system.

Once these changes are confirmed I will deleted the `src/applications/common/` directory and the `test/` directory.


jbalboni:
@annekainicUSDS Looks good to me, I can approve it once you remove the directories.

annekainicUSDS:
@jbalboni I think this is ready for another look!

jbalboni (approved):

PR #7910, opened 2018-06-21
Define constants for backend services

ncksllvn:
Throughout our codebase we generally just use plain strings to reference backend services, which is done usually when we're doing lookups in the `profile.services` array, or by registering an application to the `RequiredLoginView`. This PR defines those possible values as a constant `backendServices` and updates the references. It should hopefully add some more context to where those values come from, enforce valid service names via React propTypes, and make things easier to adjust in case services change in the future.


src/platform/user/authorization/components/RequiredLoginView.jsx
```diff
@@ -128,10 +130,12 @@ class RequiredLoginView extends React.Component {
...
 
+const validService = PropTypes.oneOf(Object.values(backendServices));
+
 RequiredLoginView.propTypes = {
   serviceRequired: PropTypes.oneOfType([
-    PropTypes.string,
-    PropTypes.array,
+    validService,
+    PropTypes.arrayOf(validService)
   ]).isRequired,
```
ncksllvn: Here's where those values are enforced via React prop-types.

ncksllvn:
On the API, services available for Downtime Notifications have been referred to as external, so I may rename those services in the FE to `externalServices` to help draw the distinction between those and the regular `backendServices`. Although, if we do try to get more specific with downtime per-app, they won't really be technically external anymore, like how we have one just called "vic".

src/platform/user/authorization/containers/MHVApp.jsx
```diff
@@ -7,6 +7,7 @@ import appendQuery from 'append-query';
 import AlertBox from '@department-of-veterans-affairs/formation/AlertBox';
 import LoadingIndicator from '@department-of-veterans-affairs/formation/LoadingIndicator';
 import { mhvAccessError } from '../../../static-data/error-messages';
+import backendServices from '../../../../platform/user/profile/constants/backendServices';
```
U-DON: Might be going up a couple extra levels for this path; this is already in `platform/user`.

U-DON (approved):
LGTM.

src/platform/user/authorization/containers/MHVApp.jsx
ncksllvn: Ah, good call! Thanks!

PR #7911, opened 2018-06-21
Disability wizard content fix

cehsu:
These changes update the content for the disability wizard when the user has a new condition and a condition that has increased.

image


cvalarida (approved):
Looks good! I'm not super familiar with the wizard; are these the only instances of this (and similar) text?

markgreenburg (approved):
LGTM

cehsu:
Thanks @cvalarida, that's probably a question for @peggygannon, but AFAIK this is the only instance of this kind of messaging.

peggygannon (approved):
@cehsu Looks good. thank you for making the update!

PR #7912, opened 2018-06-21
Update gating logic to check that services includes evss-claims

cehsu:
These changes gate the form buttons based on evss_claims being in the services list.

Form Introduction page (blocks users from starting new forms and continuing saved forms)
image

Static pages (only blocks users from continuing saved forms)
image

Linked issue: https://github.com/department-of-veterans-affairs/vets.gov-team/issues/10960


src/applications/disability-benefits/526EZ/components/FormStartControls.jsx
```diff
@@ -7,23 +7,27 @@ import SaveInProgressIntro from '../../../../platform/forms/save-in-progress/Sav
 import { ITFErrorAlert, VerifiedAlert, UnauthenticatedAlert, UnverifiedAlert, ITFDescription } from '../helpers';
 
 export default function FormStartControls(props) {
-  const { user } = props;
+  const { user: { login: { currentlyLoggedIn }, profile: { verified, services } } } = props;
 
   const somethingWentWrong = props.ITFStatus && props.ITFStatus === 'expired' || // This may not be possible
                              props.errors && props.errors.length;
+  const hasEVSSClaimsService = !!(services.find((service) => service === 'evss-claims'));
```
markgreenburg: Is there a list of constants that we could pull from instead of using the  `'evss-claims'` string directly?

src/applications/disability-benefits/526EZ/components/FormStartControls.jsx
```diff
@@ -7,23 +7,27 @@ import SaveInProgressIntro from '../../../../platform/forms/save-in-progress/Sav
...
+  const gateUnauthenticatedUser = !currentlyLoggedIn;
```
markgreenburg: Do you think we need this, vs. just using `{!currentlyLoggedIn}` directly within the `return (...)`? To me, 'no logged in' is easier to reason about than the extra layer of abstraction. Curious to hear other opinions.

src/applications/disability-benefits/526EZ/components/FormStartControls.jsx
cehsu: It looks like using the string directly is how services are typically handled, though constants are a good idea. [Ticket](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/11203) set up to discuss replacing these string instances with constants instead.

src/applications/disability-benefits/526EZ/components/FormStartControls.jsx
cehsu: Sounds good, updated.

src/applications/disability-benefits/526EZ/components/FormStartControls.jsx
cvalarida: This was actually just changed today in https://github.com/department-of-veterans-affairs/vets-website/pull/7910. Should probably happen in this PR.

cvalarida (commented):
Something about all this logic in `FormStartControls` is making me uncomfortable. What will happen if they're logged in, don't have access to `evss_claims`, and directly navigate to a page in the middle of a form?

cehsu:
@cvalarida it sounds like that flow should be added onto the prestart wrapper, once it is merged in, as another check to perform before displaying the form?

src/applications/disability-benefits/526EZ/components/FormStartControls.jsx
cehsu: Sounds good updated.

cehsu:
Thanks for your feedback @cvalarida and @markgreenburg, PTAL as I've made a few changes and updated the description to reflect these, based on your comments.

cehsu:
@goldenmeanie has indicated that we will want to use an alert here instead, I will update shortly.

src/applications/disability-benefits/526EZ/components/FormStartControls.jsx
```diff
@@ -2,28 +2,37 @@ import React from 'react';
...
+      {!currentlyLoggedIn && <div>
         {UnauthenticatedAlert}
         <button className="usa-button-primary" onClick={props.authenticate}>Sign In and Verify Your Identity</button>
       </div>}
-      {!user.profile.verified && user.login.currentlyLoggedIn && <div>
+      {gateAuthenticatedUser && <div>
         {UnverifiedAlert}
         <a href={`/verify?next=${window.location.pathname}`} className="usa-button-primary verify-link">Verify Your Identity</a>
       </div>}
+      {gateVerifiedUser && <div className="usa-alert usa-alert-error no-background-image">Sorry, our system is temporarily down while we fix a few things. Please try again later.</div>}
```
cvalarida: So we'll display this alert if the user is logged in and LOA3, but does not have the `evss_claims` service, right?

@kreek Is not having `evss_claims` a temporary state, or will the user have to take some action to fix the problem?

src/applications/disability-benefits/526EZ/components/FormStartControls.jsx
```diff
@@ -2,28 +2,37 @@ import React from 'react';
...
   const somethingWentWrong = props.ITFStatus && props.ITFStatus === 'expired' || // This may not be possible
                              props.errors && props.errors.length;
+  const hasEVSSClaimsService = !!(services.find(service => service === EVSS_CLAIMS));
+  const gateAuthenticatedUser = !verified && currentlyLoggedIn;
+  const gateVerifiedUser = !hasEVSSClaimsService && verified;
+  const permitVerifiedUser = verified && hasEVSSClaimsService;
   return (
     <div>
-      {!user.login.currentlyLoggedIn && <div>
+      {!currentlyLoggedIn && <div>
```
cvalarida: What do you think of pulling this out of the return statement. Maybe in another function? Seems like if we're not careful with the block of conditionals above this, we could display two separate alerts (or an alert and a button) which is probably not what we're wanting.

src/applications/disability-benefits/526EZ/components/createDisabilityIncreaseApplicationStatus.jsx
```diff
@@ -2,11 +2,29 @@ import React from 'react';
...
+  const state = store.getState();
+  const { user: { profile: { savedForms, services } } } = state;
+  const hasEVSSClaimsService = !!(services.find(service => service === EVSS_CLAIMS));
+  const hasSavedForm = !!(savedForms.find(form => form === formId));
+  const shouldGateForm = !hasEVSSClaimsService && hasSavedForm;
+  if (shouldGateForm) {
+    ReactDOM.render((<div itemProp="steps" itemScope itemType="http://schema.org/HowToSection">
+      <h3 itemProp="name">Ready to apply?</h3>
+      <div itemProp="itemListElement">
+        <div className="usa-alert usa-alert-error no-background-image">Sorry, our system is temporarily down while we fix a few things. Please try again later.</div>
```
cvalarida: Same concern here as above; should we display the "system is temporarily down" message if the user doesn't have `evss_claims`?

src/applications/disability-benefits/526EZ/components/FormStartControls.jsx
cehsu: Good idea, will set `content` above and return content instead.

src/applications/disability-benefits/526EZ/components/createDisabilityIncreaseApplicationStatus.jsx
cehsu: Unless the user has saved forms and doesn't have evss_claims as a servce, the wizard is provided, I'm not aware that we want to block access to information, only to accessing the form.

peggygannon:
@cehsu @markgreenburg @cvalarida @goldenmeanie 
I've made some tweaks to the error message to conform with our style (ie., we try to avoid saying "temporarily")
Can you tell me if this message below conveys the right message? 

We're sorry. The disability application is down right now while we make some updates. Please try applying again later.

src/applications/disability-benefits/526EZ/components/createDisabilityIncreaseApplicationStatus.jsx
cvalarida: Right, I think the gating logic is solid, but the error message is misleading.

@kreek What are the possible reasons a user won't have the `evss_claims` service?

cehsu:
@cvalarida @peggygannon it looks like there are some questions about the messaging here, perhaps @goldenmeanie can clarify what we want to say here and why? As I understand it, the plan is to use a generic message and revisit this content after MVP.

cehsu:
Closing this for now, since it's not needed for user testing.

PR #7913, opened 2018-06-21
Profile 360: Check for vet360 in available services before rendering contact information

ncksllvn:
- Renders an info alert banner if the user does not have vet360 in their list of services
- Separates out the different states for ProfleView
    - Originally I read this as the whole profile being offline if Vet360 is unavailable, so this wasn't necessary, but still nice
- Changes the Contact Information downtime notification dependencies to vet360

Resolves department-of-veterans-affairs/vets.gov-team/issues/10768

Branched from https://github.com/department-of-veterans-affairs/vets-website/pull/7910


ncksllvn:
@rlbaker I updated my local API, but I don't see Vet360 in my list of services for user 264. Is this expected?

bshyong:
users will never have vet360 on local unless we do some sort of custom mocking because vet360 is not available in betamocks

bshyong (approved):
LGTM

PR #7914, opened 2018-06-21
Add primaryPhone and emailAddress as required

cvalarida:
This is the only thing I could find that didn't match. They're marked as required in the schema, but we're not using the schema for this part, so there's a disconnect.

I think we'll catch any remaining outliers like this when we do the data transformation and unit tests. :crossed_fingers:

### DIFF
--- a/src/applications/disability-benefits/526EZ/tests/config/primaryAddress.unit.spec.jsx
+++ b/src/applications/disability-benefits/526EZ/tests/config/primaryAddress.unit.spec.jsx
@@ -1,10 +1,10 @@
-import React from 'react';
+/*  */import React from 'react';
 import { expect } from 'chai';
 import sinon from 'sinon';
 import { mount } from 'enzyme';
 
 
-import { DefinitionTester, // selectCheckbox 
+import { DefinitionTester, // selectCheckbox
 } from '../../../../../platform/testing/unit/schemaform-utils.jsx';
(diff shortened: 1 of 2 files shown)


src/applications/disability-benefits/526EZ/tests/config/primaryAddress.unit.spec.jsx
```diff
@@ -1,10 +1,10 @@
-import React from 'react';
+/*  */import React from 'react';
```
cehsu: It looks like this comment can be removed.

cehsu (approved):
LGTM.

src/applications/disability-benefits/526EZ/tests/config/primaryAddress.unit.spec.jsx
cvalarida: Oh. Uhhhh....yup! No idea how that sneaked in there. :confused:

markgreenburg (approved):
looks good to me, but wanted to point out that we should probably offload all of this email and phone stuff to `vets-json-schema`, which it looks like we're not using at all. Up to you whether you want to shove that into a separate PR or not.

cvalarida:
Eh, for testing tomorrow, let's just merge this in. It really should be offloaded to vets-json-schema, though. You're right.

PR #7915, opened 2018-06-21
Profile360: Add flow for validations

ncksllvn:
Branched from https://github.com/department-of-veterans-affairs/vets-website/pull/7913

This adds a "validations" property for each 360 field, so the PhoneEditModal winds up with props like -

```
data: {
  inputPhoneNumber: 'valid number saved to vet360'
},
field: {
  value: {
    inputPhoneNumber: 'invalid number typed into the form'
  },
  validations: {
    inputPhoneNumber: 'Please enter a valid phone number'
  }
}
```
The validations property is cleared out when the "dirty" flag is set to false - this means that only when a user tabs does validation run, and once they begin typing, it removes any validation messages.


src/applications/personalization/profile360/actions/misc.js
```diff
@@ -109,7 +110,7 @@ function cleanAddressDataForUpdate(value) {
 
 function updateProfileFormField(field, validator, type) {
   return (value, dirty) => {
-    const errorMessage = validator && dirty ? validator(value) : '';
+    const validations = validator && dirty ? validator(value) : {};
```
bshyong: 💯

bshyong (approved):

PR #7916, opened 2018-06-21
11133 housing zip

rianfowler:
Simulated data: 

beneficiaryzip

Todo:

- [x] Add feature flag
- [x] upper case zip
- [x] adjust regex 
- [x] Hide question when eligibility is non chapter 33
- [x] Use institution bah when question hides input.


src/applications/gi/reducers/calculator.js
```diff
@@ -69,6 +76,91 @@ export default function (state = INITIAL_STATE, action) {
         ...newState
       };
     }
+    case FETCH_BAH_FAILED: {
+      const { beneficiaryZipFetched } = action;
+      const { error } = action.payload;
+
+      // response mismatch - do nothing
+      if (beneficiaryZipFetched !== state.beneficiaryZipFetched) {
+        return state;
```
rianfowler: App will just do nothing if the input doesn't match the zip used for the request.

src/applications/gi/reducers/calculator.js
```diff
@@ -69,6 +76,91 @@ export default function (state = INITIAL_STATE, action) {
...
+    }
+
+    case FETCH_BAH_STARTED: {
+      const { beneficiaryZipFetched } = action;
+
+      const newState = {
+        beneficiaryZipError: '',
+        beneficiaryZip: beneficiaryZipFetched,
+        beneficiaryZipFetched,
+        housingAllowanceCity: 'Loading...'
```
rianfowler: Waiting on feedback from UX on what to display during loading.

src/applications/gi/reducers/calculator.js
```diff
@@ -69,6 +76,91 @@ export default function (state = INITIAL_STATE, action) {
...
+      const { beneficiaryZipFetched } = action;
+      const { error } = action.payload;
+
+      // response mismatch - do nothing
+      if (beneficiaryZipFetched !== state.beneficiaryZipFetched) {
+        return state;
+      }
+
+      const newState = {
+        beneficiaryZipError: error,
```
rianfowler: Response error gets displayed as an input error. Not sure if this is idiomatic i.e. if response errors should be displayed on a different label.

src/applications/gi/selectors/calculator.js
```diff
@@ -443,6 +443,8 @@ const getDerivedValues = createSelector(
 
     const totalHousingAllowance = monthlyRateFinal * termLength;
 
+    const bah = inputs.beneficiaryLocationBah || institution.bah;
```
rianfowler: Need to confirm that this is actually the change to the formula needed. There's a lot of references to housing allowance in this that aren't derived from the Basic Allowance for Housing.

rianfowler: Basically whenever the input is in an invalid state, this value will be null and it will fall back to the institution bah.

src/applications/gi/reducers/calculator.js
jbalboni: That seems reasonable to me. It's an error with the zip you entered (or trying to use it), so it seems like it should be an input error.

src/applications/gi/reducers/calculator.js
```diff
@@ -69,6 +76,91 @@ export default function (state = INITIAL_STATE, action) {
...
+        ...newState
+      };
+    }
+    case BENEFICIARY_ZIP_CODE_CHANGED: {
+      const { beneficiaryZip } = action;
+
+      let beneficiaryZipError = state.beneficiaryZipError || '';
+
+      if (beneficiaryZip.length >= 5 && !beneficiaryZipRegExTester.exec(beneficiaryZip)) {
+        beneficiaryZipError = 'Zip Code must be a five digit number';
```
jbalboni: Nitpick: I think zip is all uppercase in other places.

src/applications/gi/reducers/calculator.js
```diff
@@ -69,6 +76,91 @@ export default function (state = INITIAL_STATE, action) {
...
+        ...state,
+        ...newState
+      };
+    }
+    case BENEFICIARY_ZIP_CODE_CHANGED: {
+      const { beneficiaryZip } = action;
+
+      let beneficiaryZipError = state.beneficiaryZipError || '';
+
+      if (beneficiaryZip.length >= 5 && !beneficiaryZipRegExTester.exec(beneficiaryZip)) {
```
jbalboni: Do you need both of these? Could you write the regex to require that it's 5 digits?

jbalboni (commented):
This looks pretty reasonable to me so far. Only minor comments.

src/applications/gi/tests/actions/index.unit.spec.js
```diff
@@ -0,0 +1,87 @@
...
+    const dispatch = sinon.spy();
+
+    beneficiaryZIPCodeChanged('12345')(dispatch);
+
+    expect(dispatch.firstCall.calledWith({
+      type: FETCH_BAH_STARTED,
+      beneficiaryZIPFetched: '12345'
+    })).to.be.true;
+
+    setTimeout(() => {
```
rianfowler: probably a better way to do this- just went for something straight forward for now.

src/applications/gi/selectors/calculator.js
```diff
@@ -795,10 +807,13 @@ export const getCalculatedBenefits = createSelector(
     const giBillChapter = +eligibility.giBillChapter;
     const institutionType = institution.type.toLowerCase();
     const isOJT = institutionType === 'ojt';
+    const beneficiaryLocationQuestion = giBillChapter === 33;
 
     calculatedBenefits.inputs = {
       inState: false,
       tuition: true,
+      // only necessay for chapter 33 recipients who are the only beneficiaries to receive a housing allowance (BAH)
```
rianfowler: trying to add some comments on the things I understand- we don't have any test scenarios from our stakeholders on this calculation.

src/applications/gi/selectors/calculator.js
```diff
@@ -451,6 +451,18 @@ const getDerivedValues = createSelector(
 
     const totalHousingAllowance = monthlyRateFinal * termLength;
 
+    let bah;
+    if (__BUILDTYPE__ !== 'production') {
+      if (inputs.beneficiaryLocationQuestion === 'no' &&
```
rianfowler: The old behavior uses the `bah` from the institution endpoint. This becomes the "default" behavior - if the zip input is not shown or if there is no value from the end point (`beneficiaryLocationBah`) then the calculation defaults to using the default value.

src/applications/gi/reducers/calculator.js
```diff
@@ -128,6 +134,97 @@ export default function (state = INITIAL_STATE, action) {
         ...newState
       };
     }
+    case FETCH_BAH_FAILED: {
+      const { beneficiaryZIPFetched } = action;
+      const error = _.get(action, 'payload.errors[0].title');
+
+      // institution and zipcode_rates endpoints both return this generic error
```
rianfowler: I don't like this approach to error handling but probably the best we can do right now. We should probably report other kinds of errors but I don't think this app has any error reporting.

src/applications/gi/tests/actions/index.unit.spec.js
```diff
@@ -0,0 +1,87 @@
+import { expect } from 'chai';
```
rianfowler: Added some unit tests for my action since it can dispatch 4 different action types.

rianfowler:
@jbalboni This is ready to review.

Can test with [this](http://localhost:3001/gi-bill-comparison-tool/profile/11901944) on dev.

rianfowler:
New gif

beneficiaryzip2

src/applications/gi/components/profile/CalculatorForm.jsx
```diff
@@ -495,6 +537,7 @@ class CalculatorForm extends React.Component {
         {this.renderEnrolled()}
         {this.renderCalendar()}
         {this.renderKicker()}
+        {__BUILDTYPE__ !== 'production' && this.renderBeneficiaryZIP()}
```
rianfowler: This new component kicks of most of the new behavior. Gating it here seems easiest.

src/applications/gi/actions/index.js
```diff
@@ -199,3 +203,46 @@ export function calculatorInputChange({ field, value }) {
...
+  }
+
+  const url = `${api.url}/zipcode_rates/${beneficiaryZIP}`;
+
+  return dispatch => {
+    fetch(url, api.settings)
+      .then(res => res.json())
+      .then(payload => {
+        if (payload.errors) {
+          dispatch({
```
rianfowler: This doesn't follow the pattern of the other actions in this file but I couldn't get the error callback to fire when I received a 404 from this end point. Had to check for an errors attribute to get it to work.

src/applications/gi/components/profile/CalculatorForm.jsx
```diff
@@ -31,6 +32,12 @@ class CalculatorForm extends React.Component {
     this.props.onInputChange({ field, value });
   }
 
+  handleBeneficiaryZIPCodeChanged = (event) => {
+    if (!event.dirty) {
```
rianfowler: I don't understand why this input component is firing an onchange event onblur but this dirty flag seems to be the way to differentiate between the two events.

src/applications/gi/actions/index.js
jbalboni: Doesn't a 404 put you into the `catch` block? I'm not sure I understand what you mean here.

src/applications/gi/components/profile/CalculatorForm.jsx
jbalboni: Those Errorable components treat the "value" as an object containing the input value and a dirty flag, so it calls onValueChange when either of those changes.

I would really like to get some standalone input components built based on us-forms-system, but for now we have these quirky ones.

src/applications/gi/actions/index.js
rianfowler: Yeah I don't get it.

screen shot 2018-06-28 at 10 53 32
screen shot 2018-06-28 at 10 53 20

rianfowler: Might be something in `api.settings`?

rianfowler: Should also note that I don't think this app  handles any error condition- the fetches have error catches but nothing in the _behavior_ recovers from them. an invalid institution will cause the app to hang :/

rianfowler: Alrighty just confirmed that the catch is just for network errors. To handle the 404 I need to move further up the chain and check for `res.ok` (duh) and throw an error if it's not

src/applications/gi/actions/index.js
```diff
@@ -199,3 +203,48 @@ export function calculatorInputChange({ field, value }) {
...
+  return dispatch => {
+    fetch(url, api.settings)
+      .then(res => {
+        if (res.ok) {
+          return res.json();
+        }
+
+        return res.json()
+          .then(({ errors }) => {
+            throw new Error(errors[0].title);
```
rianfowler: Just throwing an error here if the `res.ok` isn't true.

rianfowler:
@jbalboni This should be ready. Can be tested with [this](http://localhost:3001/gi-bill-comparison-tool/profile/11901944).

jbalboni (approved):
Worked for me locally, and LGTM.

PR #7920, opened 2018-06-22
Remove initial data from form config

cehsu:
These changes remove initial data from the form config.


cvalarida (approved):
Looks good. We might want to remove `initialData` in favor of `maximal-test.json` or something when we get that, but for now, this should do the trick.

Note: When this is merged in, the only way to go through the form is to be logged in and either have the mock data or be connected to EVSS's VPN.

@SMLuthi Pre-fill betamocks have rated disabilities in there, right?

PR #7921, opened 2018-06-22
Remove unnecessary 'note' at bottom of paymentInfo page

markgreenburg:
This fixes a little content bug where we were giving conflicting information about how to update your accounts above and below the Payment Info content:

screen shot 2018-06-22 at 2 15 34 pm

This is now:
screen shot 2018-06-22 at 2 17 14 pm

Per the content in issue:
https://github.com/department-of-veterans-affairs/vets.gov-team/issues/10403
(Can't link it because it's already connected to another PR)

### DIFF
--- a/src/applications/disability-benefits/526EZ/helpers.jsx
+++ b/src/applications/disability-benefits/526EZ/helpers.jsx
@@ -702,14 +702,10 @@ export const AddressDescription = () => (
 );
 
 export const PaymentDescription = () => (
-  <div>
-    <p>
-      This is the bank account information we have on file for you. We’ll pay your
-      disability benefit to this account. If you need to update your bank information,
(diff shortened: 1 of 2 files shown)


cvalarida:
@peggygannon Can you verify that we should be directing the user to eBenefits instead of the call center?My understanding is we were pointing to the call center for everything.

markgreenburg:
@cvalarida yep, you're right. I was using the second revision of the content instead of the third and final 🤦‍♂️ . Fixed the before / after screenshots as well as the content in the PR. This should now _actually_ match what [Peggy recommended](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/10403#issuecomment-397719930).

cvalarida (approved):

PR #7927, opened 2018-06-26
526 vet info connect cv

cvalarida:
image

**Note:** The SSN and VA file number are missing intentionally. Whether they're needed or not is tracked in https://github.com/department-of-veterans-affairs/vets.gov-team/issues/11239. Once we have an answer to that question, we can write up another ticket to put them in there (once we know how we'll be getting that information).

@goldenmeanie Is the all-caps name okay, or should we be putting some kind of transformation on it? I shied away from it because of `O'Reilly` and the like that have multiple capital letters.


goldenmeanie:
I'm assuming we are getting the name in all caps and so have no clue about that person’s preferred pattern. Does the capitalization factor into any validation?

I'm okay with this solution, but it might be nice to take another swing at this for All Claims

cvalarida:
:thinking: Yeah, I'm not sure where that comes from. This is what's loaded up from the profile, but how it gets there is mostly a mystery to me.

markgreenburg (approved):
LGTM

src/applications/disability-benefits/526EZ/tests/config/veteranInformation.unit.spec.jsx
```diff
@@ -5,26 +5,11 @@ import { mount } from 'enzyme';
 
 import { DefinitionTester } from '../../../../../platform/testing/unit/schemaform-utils.jsx';
 import formConfig from '../../config/form.js';
-import initialData from '../schema/initialData.js';
 
-initialData.prefilled = false;
-
-describe('526EZ veteran information', () => {
+describe.skip('526EZ veteran information', () => {
```
cvalarida: Because this page uses a `ui:description` that's a connected component, I think we'll have to mock out the redux store since we can't really test the page without that.

That said, I didn't want to do that in this PR because it's arguably not the right approach. We've stayed away from it to this point and I didn't want to break that pattern without a larger discussion. Mocking the redux store is more of an integration test which...may be what this actually is? Need more clarity around that before making it happen, anyhow.

cvalarida: Ticket to re-enable these tests can be found [here](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/11281).

PR #7928, opened 2018-06-26
Dashboard - add additional check for document hostname before capturing index route 

ncksllvn:
Resolves department-of-veterans-affairs/vets.gov-team/issues/11234


bshyong (approved):
LGTM

PR #7937, opened 2018-06-27
MHV Upgrade Launch

U-DON:
Reverts the revert and includes some minor cleanup.


src/platform/user/profile/utilities/index.js
```diff
@@ -42,9 +41,6 @@ export function mapRawUserDataToState(json) {
     gender,
     isVeteran,
     loa,
-    mhv: {
-      terms: { accepted: healthTermsCurrent }
-    },
```
U-DON: Removed because everything will be in the `mhvAccount` object, which gets populated by the `fetchMHVAccount` action wherever that data is needed.

dhall (approved):
LGTM

PR #7939, opened 2018-06-27
Profile360: Add reducer property for storing pending transaction requests

ncksllvn:
Transaction updates were piling up when the server was taking longer than the polling timeout (> 1s), which was causing our success/failure handlers to fire more than once.

Resolves department-of-veterans-affairs/vets.gov-team/issues/11303


bshyong (approved):

src/applications/personalization/profile360/actions/updaters.js
```diff
@@ -37,17 +39,37 @@ function recordProfileTransaction(event, fieldName) {
...
     try {
       const { transactionId } = transaction.data.attributes;
-      const transactionRefreshed = isVet360Configured() ? await apiRequest(`/profile/status/${transactionId}`) : localVet360.updateTransactionRandom(transactionId);
+      const state = getState();
+      const isAlreadyAwaitingUpdate = state.vet360.transactionsAwaitingUpdate.includes(transactionId);
+
+      if (isAlreadyAwaitingUpdate) {
+        return;
+      }
+
```
ncksllvn: The transaction component will call this function every second. However, the function exits early if there is still a pending request, to keep transactions from piling up in case we're polling more quickly than the API can respond.

PR #7941, opened 2018-06-27
9367 grandfathered bah

rianfowler:
grandfathered


src/applications/gi/actions/index.js
```diff
@@ -180,11 +184,34 @@ export function fetchProfile(facilityCode, version) {
...
+        return res.json()
+          .then((errors) => {
+            throw new Error(errors);
+          });
+      })
+      .then(payload => {
+        const institutionZIP = _.get(payload, 'data.attributes.zip');
+        const bahUrl = `${api.url}/zipcode_rates/${institutionZIP}`;
+
+        return fetch(bahUrl, api.settings)
```
rianfowler: - Grandfathered rate is a new concept returned by the zip code end point
- calculation currently uses the non-grandfathered rate from the institution end point 
- institution end point was not updated to have the grandfathered data (apparently the rate on the institution end point comes from a different data source than the rate from used on the zip end point)
- this messiness requires calling the zip end point to get the grandfathered rate for the institution zip

Decided the simplest approach would be to chain a call to the zip end point off of the result of the institution call. A better experience would have been to call this when the user clicked the relevant question but I wanted to avoid spreading this complexity around the app.

src/applications/gi/selectors/calculator.js
```diff
@@ -451,6 +451,28 @@ const getDerivedValues = createSelector(
 
     const totalHousingAllowance = monthlyRateFinal * termLength;
 
+    let bah;
+    // if beneficiary has indicated they are using a localized rate and beneficiaryLocationBah exists, then a localized rate has been fetched and should be used
+    const useBeneficiaryLocationRate = inputs.beneficiaryLocationQuestion === 'no' &&
+      inputs.beneficiaryLocationBah !== null;
+    // if beneficiary has indicated they are using the grandfathered rate, use it when available;
+    const useGrandfatheredBeneficiaryLocationRate = inputs.giBillBenefit === 'yes';
```
rianfowler: Added some comments and tried to use some meaningful names here. 

- if the user has indicated that they started using their GI Bill benefit before 1/1/18, then we use the grandfathered rate if it's available (sometimes it's not).

src/applications/gi/actions/index.js
```diff
@@ -184,11 +184,34 @@ export function fetchProfile(facilityCode, version) {
...
+          .then(res => res.json())
+          // if there's an error from the zipRatesPayload the reducer will just use the values from the institution end point.
```
rianfowler: Decided not to error here since there's no resolution for a missing zip rate for a zip provided by the institution end point.

src/applications/gi/reducers/profile.js
```diff
@@ -35,10 +36,15 @@ export default function (state = INITIAL_STATE, action) {
...
     case FETCH_PROFILE_SUCCEEDED:
       const camelPayload = camelCaseKeysRecursive(action.payload);
-      const attributes = normalizedAttributes({
+      const bahGrandfathered = _.get(action, 'zipRatesPayload.data.attributes.mhaRateGrandfathered');
+      const attributes = {
+        ...normalizedAttributes({
         ...camelPayload.data.attributes,
         ...camelPayload.data.links
-      });
+      }),
```
jbalboni: This indentation reads a little odd to me.

jbalboni (approved):
Looks like it works locally and I don't see any code issues other than one nitpicky thing.

src/applications/gi/reducers/profile.js
rianfowler: fixed

PR #7944, opened 2018-06-28
360 profile formatting for military address

bshyong:
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/11309
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/11313

### DIFF
--- a/src/applications/personalization/profile360/components/Vet360EditModal.jsx
+++ b/src/applications/personalization/profile360/components/Vet360EditModal.jsx
@@ -67,6 +67,7 @@ export default class Vet360EditModal extends React.Component {
         onDelete,
         transactionRequest,
         analyticsSectionName,
+        deleteDisabled
       }
     } = this;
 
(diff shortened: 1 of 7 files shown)


src/applications/personalization/profile360/components/Vet360EditModal.jsx
```diff
@@ -89,7 +90,7 @@ export default class Vet360EditModal extends React.Component {
             onDelete={onDelete}
             title={title}
             analyticsSectionName={analyticsSectionName}
-            deleteEnabled={!isEmpty()}>
+            deleteEnabled={!isEmpty() && !deleteDisabled}>
             <LoadingButton isLoading={isLoading}>Update</LoadingButton>
```
ncksllvn: I feel like this could be more flexible by having `deleteDisabled` as a function instead, something like `isDeleteDisabled` that defaults to just `!isEmpty()`.

ncksllvn (approved):
LGTM

src/applications/personalization/profile360/components/Vet360EditModal.jsx
bshyong: yep we can refactor if it's being used in 2 or more places

PR #7945, opened 2018-06-29
change order of imports on entry file

rianfowler:


src/applications/veteran-id-card/veteran-id-card-entry.jsx
```diff
@@ -1,5 +1,5 @@
-import React from 'react';
 import '../../platform/polyfills';
```
rianfowler: Change the order of imports

jbalboni (approved):

PR #7946, opened 2018-06-29
Profile360: Limit messaging to just one success, one error banner

ncksllvn:
Also modifies the `/v0/user` route to only add the cache-busting parameter when explicitly told. This way we can preserve caching except when we modify the data in Vet360.

Resolves department-of-veterans-affairs/vets.gov-team/issues/11297

### DIFF
--- a/src/applications/personalization/profile360/actions/updaters.js
+++ b/src/applications/personalization/profile360/actions/updaters.js
@@ -62,7 +62,8 @@ export function refreshTransaction(transaction, analyticsSectionName) {
       });
 
       if (isSuccessfulTransaction(transactionRefreshed)) {
-        dispatch(refreshProfile());
+        const forceCacheClear = true;
+        dispatch(refreshProfile(forceCacheClear));
       } else if (isFailedTransaction(transactionRefreshed) && analyticsSectionName) {
         recordEvent({
           event: 'profile-edit-failure',
--- a/src/applications/personalization/profile360/containers/Vet360TransactionReporter.jsx
+++ b/src/applications/personalization/profile360/containers/Vet360TransactionReporter.jsx
@@ -7,7 +7,9 @@ import scrollToTop from '../../../../platform/utilities/ui/scrollToTop';
 
 import {
   selectVet360SuccessfulTransactions,
-  selectVet360FailedTransactions
+  selectVet360FailedTransactions,
+  selectMostRecentSuccessfulTransaction,
+  selectMostRecentErroredTransaction
 } from '../selectors';
 
 import {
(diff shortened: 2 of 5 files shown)


src/applications/personalization/profile360/actions/updaters.js
```diff
@@ -62,7 +62,8 @@ export function refreshTransaction(transaction, analyticsSectionName) {
       });
 
       if (isSuccessfulTransaction(transactionRefreshed)) {
-        dispatch(refreshProfile());
+        const forceCacheClear = true;
+        dispatch(refreshProfile(forceCacheClear));
       } else if (isFailedTransaction(transactionRefreshed) && analyticsSectionName) {
```
ncksllvn: Here's where we set that flag for clearing the cache.

src/applications/personalization/profile360/containers/Vet360TransactionReporter.jsx
```diff
@@ -26,39 +28,43 @@ class Vet360TransactionReporter extends React.Component {
...
       </div>
     );
   }
 }
 
 const mapStateToProps = (state) => {
   return {
+    mostRecentSuccessfulTransaction: selectMostRecentSuccessfulTransaction(state),
+    mostRecentErroredTransaction: selectMostRecentErroredTransaction(state),
     successfulTransactions: selectVet360SuccessfulTransactions(state),
```
ncksllvn: Added new selectors for grabbing just the most recent successful/errored transaction. However, we still need to get the corresponding arrays, bc we need to clear all of them once the user clicks the close button.

src/applications/personalization/profile360/reducers/vet360.js
```diff
@@ -10,10 +10,19 @@ import {
...
+
 const initialState = {
   transactions: [],
   fieldTransactionMap: {},
-  transactionsAwaitingUpdate: []
+  transactionsAwaitingUpdate: [],
+  metadata: {
+    mostRecentSuccessfulTransactionId: '',
+    mostRecentErroredTransactionId: '',
+  }
```
ncksllvn: I'm not sure if `metadata` is the best term here, but I was trying to use a term broad enough that we could store other information as we need it.

src/applications/personalization/profile360/selectors.js
```diff
@@ -49,6 +49,40 @@ export function selectVet360FailedTransactions(state) {
...
+    }
+  } = state;
+
+  let transaction = null;
+  if (mostRecentErroredTransactionId) {
+    transaction = transactions.find(t => t.data.attributes.transactionId === mostRecentErroredTransactionId);
+  }
+  return transaction;
+}
+
```
ncksllvn: These are so verbose, but they get the point across.

bshyong (approved):
LGTM

PR #7952, opened 2018-07-02
Profile360: Show loading indicator during delete

ncksllvn:
- Changes name of `FormActionButtons` to `Vet360EditModalActionButtons`
- Passes `transactionRequest` down props to delete button
- Fixes bug in Vet360 reducer
    - Should have been using the spread operator as the first step of generating new object props, not the last.

Resolves department-of-veterans-affairs/vets.gov-team/issues/11353

### DIFF
--- a/src/applications/personalization/profile360/reducers/vet360.js
+++ b/src/applications/personalization/profile360/reducers/vet360.js
@@ -57,10 +57,10 @@ export default function vet360(state = initialState, action) {
         fieldTransactionMap: {
           ...state.fieldTransactionMap,
           [action.fieldName]: {
+            ...state.fieldTransactionMap[action.fieldName],
             isPending: false,
             isFailed: true,
-            error: action.error,
-            ...state.fieldTransactionMap[action.fieldName]
+            error: action.error
           }
         }
       };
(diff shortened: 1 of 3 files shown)


src/applications/personalization/profile360/reducers/vet360.js
```diff
@@ -72,9 +72,9 @@ export default function vet360(state = initialState, action) {
         fieldTransactionMap: {
           ...state.fieldTransactionMap,
           [action.fieldName]: {
+            ...state.fieldTransactionMap[action.fieldName],
             isPending: false,
-            transactionId: action.transaction.data.attributes.transactionId,
-            ...state.fieldTransactionMap[action.fieldName]
+            transactionId: action.transaction.data.attributes.transactionId
           }
```
ncksllvn: The order here had to be switched so that `isPending` is properly overwritten

BBarone1:
:eyes:

bshyong (approved):

src/applications/personalization/profile360/reducers/vet360.js
bshyong: we should add an explanatory comment for clarity in the next PR

PR #7957, opened 2018-07-03
adjust yellow ribbon amount input behavior

rianfowler:
Some adjustments based on stakeholder feedback. 

newyellowribbon


src/applications/gi/components/Dropdown.jsx
```diff
@@ -7,13 +7,15 @@ class Dropdown extends React.Component {
     if (!this.props.visible) {
       return null;
     }
+    const hideArrowsClass = this.props.hideArrows ? 'hide-arrows' : '';
```
rianfowler: Hide the arrows on the select when there is only one item available.

src/applications/gi/selectors/calculator.js
```diff
@@ -1023,9 +1023,7 @@ export const getCalculatedBenefits = createSelector(
       };
     }
 
-    if (institution.yr &&
-      derived.tier === 1.0 &&
-      institution.yellowRibbonPrograms.length > 0) {
```
rianfowler: The institution end point provides a boolean that indicates if an institution has a yellow ribbon program. Adjusted the drop down behavior so the app continues to use this attribute to determine when to display the yellow ribbon question.

src/applications/gi/reducers/calculator.js
```diff
@@ -65,35 +65,48 @@ export default function (state = INITIAL_STATE, action) {
...
-          newState = {
-            ...newState,
-            yellowRibbonAmount,
-            yellowRibbonDivisionOptions,
-            yellowRibbonDivision: yellowRibbonDivisionOptions[0],
-            yellowRibbonProgramIndex,
-            yellowRibbonMaxAmount: yellowRibbonAmount,
-            yellowRibbonMaxNumberOfStudents
-          };
+          if (value === 'Enter an amount') {
```
rianfowler: Stakeholder wanted something to make it obvious that the user doesn't have to select an amount from the drop down. Open to feedback on this approach (we don't have a UX to assist with this adjustment).

src/applications/gi/components/profile/CalculatorForm.jsx
```diff
@@ -174,16 +178,18 @@ class CalculatorForm extends React.Component {
                 label="Degree Level"
                 name="yellowRibbonDegreeLevel"
                 alt="Degree Level"
+                hideArrows={yellowRibbonDegreeLevelOptions.length <= 1}
                 options={yellowRibbonDegreeLevelOptions}
-                visible
+                visible={showYellowRibbonOptions}
```
rianfowler: Hide the drop downs when there is yellow ribbon support but no specific programs to select from.

jbalboni:
@rianfowler Can you post a screenshot/gif of when the arrows are removed on a select? That's an odd pattern that we don't have elsewhere.

rianfowler:
@jbalboni It's in the gif when undergraduate is selected. It is an odd pattern. The whole thing is an odd pattern.

rianfowler:
@jbalboni Here's what it looks like when it's clicked. 

newyellowribbon2

rianfowler:
This was the specific feedback:

>     If a degree level only has one option the drop down arrows are still appearing.

Probably shouldn't even show an input if it's just one item that they can't do anything with.

jbalboni:
Ok. We often try to push back on misguided design feedback like that from stakeholders, but if the better way of doing it is a more in depth reorganization of the inputs (which it sounds like is the case), then seems ok to stick with hiding the arrows.

src/applications/gi/reducers/calculator.js
jbalboni: I'm a little worried that we'll tweak this text and miss changing it here. Could we use a value like `customAmount` or something and separate it from the description the user sees?

jbalboni (approved):

src/applications/gi/reducers/calculator.js
rianfowler: Will do

PR #7958, opened 2018-07-03
add independent study and priority enrollment filters

rianfowler:
- You can validate the changes to the institution page by pointing `gids` in your `vets-api` `settings.yml` file to *https://staging.vets.gov/gids* and navigating to *http://localhost:3001/gi-bill-comparison-tool/profile/11800048* 
 - Since the data is not seeded in dev, validating the search page changes is a little trickier. Do a search and add the filters you want (priority enrollment and/or independent study) to the locally running app. Use the generated url on staging to see filtered results e.g. *https://staging.vets.gov/gi-bill-comparison-tool/search?independentStudy=true&name=virginia+university&priorityEnrollment=true* You should see a filtered list with 6 universities in west Virginia. The filtering behavior happens through the search call by adding additional parameters to the query string which is why this works on staging.


src/applications/gi/components/search/InstitutionFilterForm.jsx
```diff
@@ -118,6 +118,16 @@ class InstitutionFilterForm extends React.Component {
           name="stemOffered"
           label="STEM (Science, Technology, Engineering, and Math)"
           onChange={this.handleCheckboxChange}/>}
+        {__BUILDTYPE__ !== 'production' && <Checkbox
```
rianfowler: Blocking the UI element puts this behind a flag.

src/applications/gi/components/profile/Programs.jsx
```diff
@@ -53,6 +53,12 @@ export class Programs extends React.Component {
         text: 'Military Tuition Assistance (TA)',
         link: false
       },
+
+      priorityEnrollment: {
```
rianfowler: new attributes are returned to the institution end point `priority_enrollment` and `independent_study`. The keys are used directly here and [here](https://github.com/department-of-veterans-affairs/vets-website/blob/1178155e3340b103c17f09734d9302fa6dd0cabe/src/applications/gi/components/profile/AdditionalInformation.jsx#L74).

cehsu (approved):
LGTM.

PR #7959, opened 2018-07-03
add prefill transform for 1990 email and phone numbers

rianfowler:
This can be test locally against the `prefill-test` branch of `vets-api`. All logged in users will return the prefill data for contact info.


src/applications/edu-benefits/1990/helpers.jsx
```diff
@@ -2,6 +2,51 @@ import React from 'react';
...
 import { transformForSubmit } from 'us-forms-system/lib/js/helpers';
 
+export function prefillTransformer(pages, formData, metadata) {
+
+  let newContactInfo = {};
+
+  if (formData && formData.email) {
+    newContactInfo = {
+      email: formData.email,
+      'view:confirmEmail': formData.email
```
jbalboni: I wonder if we should make them type this?

src/applications/edu-benefits/1990/helpers.jsx
```diff
@@ -2,6 +2,51 @@ import React from 'react';
...
+    };
+  }
+
+  if (formData && formData.homePhone) {
+    newContactInfo = {
+      ...newContactInfo,
```
jbalboni: You could set `homePhone` on `newContactInfo` directly. You created it in this function, so mutating it won't cause any issues.

src/applications/edu-benefits/1990/helpers.jsx
```diff
@@ -2,6 +2,51 @@ import React from 'react';
...
+      homePhone: formData.homePhone
+    };
+  }
+
+  if (formData && formData.mobilePhone) {
```
jbalboni: You also should remove these contact fields from the top level of the form data object you return. I think they'll get overwritten on submit, but we shouldn't leave them hanging around.

src/applications/edu-benefits/1990/helpers.jsx
rianfowler: Good idea- I'll take this part out.

src/applications/edu-benefits/1990/helpers.jsx
```diff
@@ -1,7 +1,44 @@
...
+  };
+
+  if (formData) {
+    const {
+      email,
+      homePhone,
+      mobilePhone
+    } = formData;
+
+    delete newFormData.email;
```
rianfowler: Remove these from `formData`.

src/applications/edu-benefits/1990/helpers.jsx
```diff
@@ -1,7 +1,44 @@
...
+    delete newFormData.homePhone;
+    delete newFormData.mobilePhone;
+
+    const newContactInfo = {
+      ...(email && { email }),
```
rianfowler: consolidated these into a single statement. seems easier to read

src/applications/edu-benefits/1990/helpers.jsx
```diff
@@ -1,7 +1,43 @@
...
+      ...(homePhone && { homePhone }),
+      ...(mobilePhone && { mobilePhone })
+    };
+
+    if (!_.isEmpty(newContactInfo)) {
```
rianfowler: seemed clearer to use a lodash function here.

src/applications/edu-benefits/1990/helpers.jsx
rianfowler: JS should have a better way of doing this kind of thing.

rianfowler:
@jbalboni addressed comments- thanks for the feedback.

jbalboni (approved):

PR #7960, opened 2018-07-03
526 -- chapter 1 content edits

cvalarida:
image

image

image
(This still needs to be view-only, so don't get too distracted there.)

image


markgreenburg (approved):
LGTM - I think your first screenshot is old, though

cvalarida:
Oh blast. You're right, @markgreenburg.

@peggygannon The content for that first screenshot is the following:
```
This is the personal information we have on file for you.

[[ personal information here ]]

This is the contact information we have on file for you. We’ll send any important information about your disability claim to the address listed here. Any updates you make here to your contact information will only apply to this application.
```

peggygannon (changes requested):
@cvalarida Just two edits: 
1. The first screen copy should be: 
This is the personal information we have on file for you. 
[personal details]

**Note:** If you need to update your personal information, please call Veterans Benefits Assistance at 1-800-827-1000, Monday through Friday, 8:00 a.m. to 9:00 p.m. (ET).

2. for the second screen can you replace the & with "and"

cvalarida:
Turns out I just plum missed that part. Thanks, @peggygannon!
Updated screenshot:
image

peggygannon:
@cvalarida thank you.. Looks good.

markgreenburg:
Hi @peggygannon - could you please take another look and approve this content if all looks good to you? Seems like you might've meant to approve it per your last comment.

peggygannon (approved):

PR #7963, opened 2018-07-03
Patching an issue where additional text was appearing in mobile breadcrumb

1Copenut:
The mobile breadcrumb script is picking up some extra text from an analytics handler, and making the back by one link too long on the account, profile, and profile360 pages. This moves that handler to its own script, and removes the `aria-label` attribute from the back by one mobile breadcrumb.


ncksllvn (approved):
LGTM, but also brings up a couple points we could probably split into tickets  -

1. Use React component for breadcrumbs on Personalization apps
2. Ensure safety of `innerHTML`, possibly switch to that.

1Copenut:
@ncksllvn @bshyong 

@U-DON asked if a CSS only solution was possible, and I think the answer might be yes. I'll investigate that after the holiday, and see if we can simplify the script and avoid cloning elements.

bshyong (approved):

PR #7964, opened 2018-07-04
Profile360 adjustments

bshyong:
Requires merge of https://github.com/department-of-veterans-affairs/design-system/pull/130

Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/11309#issuecomment-402298079
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/10884#issuecomment-401402278

### DIFF
--- a/src/applications/personalization/profile360/actions/misc.js
+++ b/src/applications/personalization/profile360/actions/misc.js
@@ -25,7 +25,7 @@ function inferAddressType(countryName, stateCode) {
   if (countryName !== 'United States') {
     addressType = 'INTERNATIONAL';
   } else if (MILITARY_STATES.has(stateCode)) {
-    addressType = 'MILITARY OVERSEAS';
+    addressType = 'OVERSEAS MILITARY';
   }
 
   return addressType;
--- a/src/applications/personalization/profile360/components/Address.jsx
+++ b/src/applications/personalization/profile360/components/Address.jsx
@@ -7,7 +7,6 @@ import { focusElement } from '../../../../platform/utilities/ui';
 import ErrorableSelect from '../../../letters/components/ErrorableSelect';
 import ErrorableTextInput from '../../../letters/components/ErrorableTextInput';
 import { STATE_CODE_TO_NAME, MILITARY_CITIES, MILITARY_STATES } from '../../../letters/utils/constants';
-import { militaryStateNames } from '../../../letters/utils/helpers';
 
 /**
  * Input component for an address.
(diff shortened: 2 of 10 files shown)


src/applications/personalization/profile360/constants/vet360.js
```diff
@@ -32,3 +32,15 @@ export const PHONE_TYPE = {
...
+  primaryTelephone: 'primary-telephone',
+  alternateTelephone: 'alternative-telephone',
+  homePhone: 'home-telephone',
+  mobilePhone: 'mobile-telephone',
+  workPhone: 'work-telephone',
+  faxNumber: 'fax-telephone',
+  email: 'email',
+  mailingAddress: 'mailing-address',
+  residentialAddress: 'home-address'
+};
```
ncksllvn: Nice! Not that we need to worry about it right now, but I was wondering if a full-out mapping might be beneficial, like:

```
const FIELDS = {
  [FIELDNAMES.PRIMARY_PHONE]: {
    vet360PropertyName: 'primaryPhone',
    analyticsName: 'primary-telephone',
    label: 'Primary phone number'
  }
}
```

But I could also see that being overkill.

src/applications/personalization/profile360/components/Address.jsx
```diff
@@ -28,39 +27,13 @@ class Address extends React.Component {
...
-        return 0;
-      });
-    }
+    const statesList = [];
 
-    return statesList;
-  }
+    _.mapKeys(STATE_CODE_TO_NAME, (value, key) => {
+      statesList.push({ label: value, value: key });
+    });
```
ncksllvn: Would it possibly make sense to rename "key" to "stateAbbreviation" and "value" to "stateName"? I had to lookup what `STATE_CODE_TO_NAME` looked like to be sure.

Could also initialize `stateList` to `_.mapKeys` by returning `{ label: value, value: key }` instead of doing a `.push`.

src/applications/personalization/profile360/actions/misc.js
```diff
@@ -25,7 +25,7 @@ function inferAddressType(countryName, stateCode) {
   if (countryName !== 'United States') {
     addressType = 'INTERNATIONAL';
   } else if (MILITARY_STATES.has(stateCode)) {
-    addressType = 'MILITARY OVERSEAS';
+    addressType = 'OVERSEAS MILITARY';
   }
```
ncksllvn: After things get quiet I would like to get our Vet360 address logic more organized, getting these address-types stored in a constant.

ncksllvn (approved):
LGTM! Just left some ideas for later on down the road.

PR #7965, opened 2018-07-05
Use latest forms library

cvalarida:
Just had to update the forms library since I had already made the fix once before.


markgreenburg (approved):
Not latest anymore since there have been some content edits but lgtm

cvalarida:
:stuck_out_tongue:

PR #7968, opened 2018-07-06
Set up Opt Out form app scaffold

cehsu:
These changes set up the Opt Out form app scaffold.
image

Additional work related to this has been documented in the following tickets:
- [Remove review-and-submit page from routes](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/11429)

### DIFF
--- a/content/pages/education/opt-out.md
+++ b/content/pages/education/opt-out.md
@@ -1,0 +1,34 @@
+---
+layout: page-breadcrumbs.html
+template: detail-page
+title: Work and Learn
+concurrence: complete
+plainlanguage: 12-02-16 certified in compliance with the Plain Writing Act
+collection: education
+children: educationWorkLearn
+order: 7
+majorlinks:
+  - heading:
+    links:
+    - url: /education/work-learn/workstudy/
+      title: Work Study
+      description: Learn more about the VA work-study program, which lets students make money while going to a school of higher learning.
+    - url: /education/work-learn/job-and-apprenticeship/
+      title: On-the-Job Training and Apprenticeships
+      description: Find out how to get money to cover books, supplies, and housing while learning a trade through an apprenticeship.
+    - url: /education/work-learn/co-op-training/
+      title: Co-op Training
+      description: Find out if you qualify for educational benefits if you’re part of a university or college co-op program. 
+    - url: /education/work-learn/non-college-degree-program/
+      title: Non-College Degree Programs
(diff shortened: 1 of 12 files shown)


content/pages/education/opt-out.md
```diff
@@ -0,0 +1,34 @@
+---
```
cehsu: This is a placeholder page for the static landing page.

rianfowler: My understanding is that this static page isn't connected to the app yet. Do we want to use a widget similar to what's on the `education/apply` [page](https://github.com/department-of-veterans-affairs/vets-website/blame/master/content/pages/education/apply.md#L15-L19)?

src/applications/edu-benefits/opt-out/routes.jsx
```diff
@@ -0,0 +1,12 @@
+import { createRoutesWithSaveInProgress } from '../../../platform/forms/save-in-progress/helpers';
+import formConfig from './config/form';
+import OptOutApp from './containers/OptOutApp.jsx';
+
+const route = {
+  path: '/',
+  component: OptOutApp,
+  indexRoute: { onEnter: (nextState, replace) => replace('/introduction') },
```
rianfowler: Since we're not going to have an introduction page, do we know what we need to change with the routes to make this basically function as just a review and submit page.

content/pages/education/opt-out.md
cehsu: Using a widget similar to that which is on the `education/apply` page makes sense, if it seems like overkill to display a modal, we can go with a vanilla JS option for now as well. 

Some general steps on the react approach are described below:

Once a widget is set up and ready to use:

 1. Loading logic specifying the appropriate markdown page needs to be added to `static-pages-entry.js`
2. the markdown page needs to include the widget in its frontmatter under `widgets` and an anchor for the react component must be included in the markup

src/applications/edu-benefits/opt-out/routes.jsx
cehsu: [An issue](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/11429) has been created to discuss this.

rianfowler (approved):

src/applications/edu-benefits/opt-out/routes.jsx
cehsu: The introduction and review-and-submit routes are now filtered out of the childRoutes.

PR #7970, opened 2018-07-06
Set up complaint-tool scaffold

cehsu:
These changes set up the scaffold for the complaint tool.


rianfowler (approved):
was able to run locally - looks good

PR #7971, opened 2018-07-06
add query params to action creator

rianfowler:
The way I validated this before was incorrect. 

- Point your gids in your `settings.yml` in `vets-api` to staging. 
- after adding the priority enrollment or independent study filters to the [gi bill search](http://localhost:3001/gi-bill-comparison-tool/search?), the client should make an XHR request similar to this: http://localhost:3000/v0/gi/institutions/search?name=west&priority_enrollment=true&independent_study=true

### DIFF
--- a/src/applications/gi/containers/SearchPage.jsx
+++ b/src/applications/gi/containers/SearchPage.jsx
@@ -67,7 +67,9 @@ export class SearchPage extends React.Component {
       'yellowRibbonScholarship',
       'principlesOfExcellence',
       'eightKeysToVeteranSuccess',
-      'stemOffered'
+      'stemOffered',
+      'priorityEnrollment',
+      'independentStudy'
     ];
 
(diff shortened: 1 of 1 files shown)


src/applications/gi/containers/SearchPage.jsx
```diff
@@ -67,7 +67,9 @@ export class SearchPage extends React.Component {
       'yellowRibbonScholarship',
       'principlesOfExcellence',
       'eightKeysToVeteranSuccess',
-      'stemOffered'
+      'stemOffered',
+      'priorityEnrollment',
```
rianfowler: These new query params need to be in this array to build the correct XHR.

cehsu: Should we use constants and proxies here?

cehsu (approved):
LGTM.

PR #7978, opened 2018-07-09
Remove VA-Profile directory

ncksllvn:
I believe that something with the `contentPage` property in `va-profile/manifest.json` was causing the issue with the `profile360`'s production deploy.


src/applications/personalization/va-profile/manifest.json
```diff
@@ -1,8 +0,0 @@
-{
-  "appName": "VA Profile",
-  "entryFile": "./va-profile-entry.jsx",
-  "entryName": "va-profile",
-  "rootUrl": "/profile",
-  "production": false,
-  "contentPage": "profile-legacy/index.md"
-}
```
ncksllvn: Something with the hyphen here maybe?

bshyong (approved):

PR #7984, opened 2018-07-10
526 treatment address mg

markgreenburg:
## What this PR Does:
- Adds country, city, and state fields to each VA Treatment Center
- Makes country always required (per swagger)
- Makes city & state always optional (per swagger + discussion with CSRA)
- Validates military cities and military states (if military city -> must be military state & vice-versa)
- Conditionally expands state field when country is USA
- Defaults country to USA
- Adds tests for the above things

## What this PR Does NOT Do:
- Restrict state list to military states when city is military city (the full list is always displayed)
- Conditionally require state / city (which isn't a big deal since they're actually optional)
- hide the state field when country is not USA (instead, it conditionally expands the state field when country is USA)

### Why This PR Doesn't Do These Things:
the VA addresses are part of a treatment. A treatment lives inside an array of treatment objects. The treatment array lives inside a disability. And a disability lives within an array of disabilities. So the schema involves nested arrays, and in its simplest reduction looks like this:
```
disabilities = [
  {
    treatments: [
      { treatmentCenterAddress },
      { treatmentCenterAddress }
    ]
  },
  {
    treatments: [
      { treatmentCenterAddress },
      { treatmentCenterAddress }
    ]
  }
];
```
Many config properties (e.g., `ui:validations`) are passed the index of the data in its array (when the data is in an array). This is helpful to find the right treatment within a given treatments array. However, config properties differ in what they pass into these functions as the `formData`. Some properties pass the data's parent object (`expandUnder, `ui:validations`) while most properties pass the entire `formData` object (`hideIf`, `updateSchema`, `ui:required`, etc.). Passing the entire `formData` object poses a problem because only one index is supplied, so while we can keep track of which treatment within the treatments array is currently being processed, we have no way of keeping track / specifying which treatments array we're looking at in cases where there are multiple disabilities, each with its own treatments array. So, certain things just aren't possible. An issue to this effect will be submitted to the us-forms repo shortly.

## Screenshots:
Country, State, and City show by default (because country defaults to USA - otherwise state would be hidden)
screen shot 2018-07-10 at 7 31 19 pm
-----

US addresses work as expected:
screen shot 2018-07-10 at 7 32 04 pm
-----

International addresses drop the state:
screen shot 2018-07-10 at 7 32 26 pm
-----

City is validated against military states:
screen shot 2018-07-10 at 7 32 39 pm
-----

And state is validated against military cities:
screen shot 2018-07-10 at 7 32 56 pm
-----

The view cards are unaffected, though I'm happy to add address info later if there's a need for that:
screen shot 2018-07-10 at 7 33 43 pm

### DIFF
--- a/src/applications/disability-benefits/526EZ/pages/treatmentAddress.js
+++ b/src/applications/disability-benefits/526EZ/pages/treatmentAddress.js
@@ -1,21 +1,14 @@
-import _ from 'lodash';
-
-import fullSchema526EZ from 'vets-json-schema/dist/21-526EZ-schema.json';
-// import fullSchema526EZ from '/path/vets-json-schema/dist/21-526EZ-schema.json';
+import get from '../../../../platform/utilities/data/get';
 
 import  {
   MILITARY_CITIES,
-  MILITARY_STATE_LABELS,
   MILITARY_STATE_VALUES,
-  STATE_LABELS,
-  STATE_VALUES,
   USA,
-  ADDRESS_TYPES
 } from '../constants';
 
 function validateMilitaryCity(errors, city, formData, schema, messages, index) {
   const isMilitaryState = MILITARY_STATE_VALUES.includes(
-    _.get(formData, `treatments[${index}].treatmentCenterAddress.state`, '')
+    get(formData, `treatments[${index}].treatmentCenterAddress.state`, '')
   );
   const isMilitaryCity = MILITARY_CITIES.includes(city.trim().toUpperCase());
   if (isMilitaryState && !isMilitaryCity) {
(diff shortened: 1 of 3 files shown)


src/applications/disability-benefits/526EZ/pages/treatmentAddress.js
```diff
@@ -0,0 +1,48 @@
...
+
+import  {
+  MILITARY_CITIES,
+  MILITARY_STATE_VALUES,
+  USA,
+} from '../constants';
+
+function validateMilitaryCity(errors, city, formData, schema, messages, index) {
+  const isMilitaryState = MILITARY_STATE_VALUES.includes(
+    get(formData, `treatments[${index}].treatmentCenterAddress.state`, '')
```
cvalarida: Did you verify this `index` is for `treatments` and not for `disabilities`? Should be able to do that by making two treatment centers where one has a domestic address and another has a military address.

cvalarida (commented):
Looks good! Can you throw a screenshot or two in here too?

markgreenburg:
Ah, yeah, screenshots would be good wouldn't they?

src/applications/disability-benefits/526EZ/pages/treatmentAddress.js
markgreenburg: Yep, verified by doing exactly that. Also, by logging out the index property while adding multiple facilities to the same disability.

markgreenburg:
@cvalarida added some screenshots. Can also confirm that the index passed in is the correct one.

cvalarida (approved):
Beautiful! :shipit:

PR #7992, opened 2018-07-12
Fix global Moment-formatting issues

ncksllvn:
We have repetitive code throughout the repo to apply settings under `moment.updateLocale`, some of which are slightly different and causing the global Moment object to display differently during unit tests, occasionally causing random failures. This PR imports `moment-setup.js` at startup and removes any other calls to `moment.updateLocale`. Any imports of moment can then be done by importing moment as normal, instead of needing to reference the platform's `moment-setup` module.

- Moves `moment-setup.js` to the `platform/site-wide` and imports it in `site-wide/index.js`. This way the formatting rules applied under `updateLocale` are applied globally and future imports can just use `moment`.
- Imports `moment.setup.js` in `mocha-setup.js` to ensure unit tests are executed with the same global settings applied.
- Updates references to `moment-setup` within `applications/` to instead just import `moment`, or nothing at all if moment was only being imported to call `updateLocale`.
- Removes all other calls to `updateLocale`

Resolves department-of-veterans-affairs/vets.gov-team/issues/11259


cehsu (approved):
LGTM. Great job making this consistent.

bshyong (approved):
LGTM

ncksllvn:
Thank you @cehsu!

PR #7997, opened 2018-07-12
Re-enable DS Logon Banner

wunderhund:


jkassemi (approved):
this was what we had last time, so 👍. 

Working to get more confirmation on the messaging since it doesn't seem quite right.

phiden (approved):

PR #8001, opened 2018-07-13
DO NOT MERGE: added crosslinks to optout static landing page

marybyrnepark:
These changes add the cross links that we will want to launch along with the form. This should not be merged before then.


bethpotts (approved):
Approved. Looks good.

bethpotts:
@cehsu I approved, then realized that my commits aren't reflected in preview in heroku. Does that matter as far as what gets merged into staging? If my edits are in the file itself, they should get merged even though they aren't showing up in preview, yes? Take a look above. The View Deployment button didn't move down to meet the latest commits.

PR #8003, opened 2018-07-13
526 valid disability check

cvalarida:
## No Eligible Disabilities
image
This will be shown when there are no eligible disabilities to claim an increase on.

## Error Pre-filling Form
image
**Note:** The copy I wrote isn't strictly speaking accurate. It's not so much that we couldn't find any rated disabilities as it is that there was an error somewhere along the way and we couldn't pre-fill the form.


cvalarida:
@peggygannon Can you provide real content for both of those scenarios please?

kreek (approved):
LGTM

peggygannon (commented):
@cvalarida These messages still need to be approved by Dawn and Andrea, but you can drop in this copy for now. This is just placeholder copy until they have a chance to approve what I wrote. Thx! 

Message for no rated disabilities
We’re sorry. Our records show you don’t have any rated disabilities, so you can’t apply for a Claim for Increase at this time. If you think our records aren’t correct, please call Veterans Benefits Assistance at 1-800-827-1000, Monday – Friday, 8:00 a.m. to 9:00 p.m. (ET).

Error pre-filling form
We’re sorry. It looks like we’re missing the following information needed for your application:
-Your rated disabilities
To update this information for your application, please call Veterans Benefits Assistance at 1-800-827-1000, Monday – Friday, 8:00 a.m. to 9:00 p.m. (ET).

markgreenburg (commented):
Looks like the 'ineligible' disabilities still show on the disability page. Getting clarity on what we should do with these and whether or not it makes sense to deal with them in the same PR.

markgreenburg (approved):
Seems like we want to keep ineligible disabilities but 1. not let a user enter the form without any eligible disabilities and 2. give ineligible disabilities special UI treatment. 1 is covered by this PR and I think it makes sense to figure out 2 in a separate PR, since this approach here works for keeping ineligible disabilities.

PR #8006, opened 2018-07-16
Safety check for missing disabilities

markgreenburg:
- Updates disabilities prefill transformer to not explode when disabilities is missing
- Adds test cases for these scenarios.

### DIFF
--- a/src/applications/disability-benefits/526EZ/helpers.jsx
+++ b/src/applications/disability-benefits/526EZ/helpers.jsx
@@ -121,7 +121,7 @@ export function addPhoneEmailToCard(formData) {
 
 export function prefillTransformer(pages, formData, metadata) {
   const { disabilities } = formData;
-  if (!disabilities) {
+  if (!disabilities || !Array.isArray(disabilities)) {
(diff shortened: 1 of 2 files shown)


kreek (approved):
LGTM 🚀

src/applications/disability-benefits/526EZ/helpers.jsx
```diff
@@ -120,7 +120,12 @@ export function addPhoneEmailToCard(formData) {
 }
 
 export function prefillTransformer(pages, formData, metadata) {
-  const withDisabilityActionType = set('disabilities', transformDisabilities(formData.disabilities), formData);
+  const { disabilities } = formData;
+  if (!disabilities || !Array.isArray(disabilities)) {
```
cehsu: Should we check for array length is zero if the default parameter is an array/if there will be no disabilities in the array if `no-rated-disabilities-found`?

cehsu (approved):
LGTM pending feedback on what we'll expect in the event that no disabilities are found.

PR #8011, opened 2018-07-17
change default zoom level and search radius

dhall:
I'll set something up to walk through these changes with @U-DON  and @NatalieEMoore and see if we like the behavior


U-DON (approved):

r-mc2:
Hey @dhall saw this got updated recently and you changed some of the bounding box calculations (0.5 --> 0.75). Saw
```js
this.zoomOut = debounce(() => this.refs.map.leafletElement.zoomOut(0.5), 2500, {
  leading: true,
});
```
up in the constructor and wondered if that needed updating as well or is Leaflet's zoom is a different case?

PR #8012, opened 2018-07-17
Profile - Show field update after successful transaction, rather than success banner

ncksllvn:
Most of this PR is just removing code we no longer need, as we no longer need to show top-level success banners or keep track of the most recent successful transaction.

Resolves department-of-veterans-affairs/vets.gov-team/issues/11351


src/applications/personalization/profile360/actions/updaters.js
```diff
@@ -55,7 +69,8 @@ export function refreshTransaction(transaction, analyticsSectionName) {
 
       if (isSuccessfulTransaction(transactionRefreshed)) {
         const forceCacheClear = true;
-        dispatch(refreshProfile(forceCacheClear));
+        await dispatch(refreshProfile(forceCacheClear));
+        dispatch(clearTransaction(transactionRefreshed));
         recordEvent({ event: 'profile-saved' });
```
ncksllvn: This is the main logical change - once transaction is updated to successful and the profile then finishes updating, we clear the transaction from state.

bshyong (approved):
LGTM

PR #8015, opened 2018-07-17
User testing fixes: Enable form load/prefill for opt out form app, implement back button

cehsu:
These changes fix bugs discovered during user testing, including:

- The opt out form application should load/prefill the form if the user is logged in and they have routed to the opt out applicant's only form page (/claimant-information), or rather have not navigated to any of the form's non-input pages (confirmation, form-saved, etc.). The opt out form application currently only does so if the form is being "resumed" (path ends with /resume).

- The back button should navigate to the previous page in the browser's history (pending @emilywaggoner's [input](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/11331#issuecomment-405766654))


jholton (commented):
I'll let @rianfowler approve, but I can confirm that this does now allow prefill to work (manually tested locally).

rianfowler (approved):

PR #8016, opened 2018-07-18
WIP - 11412 autocomplete widget

rianfowler:
This widget attempts to combine functionality from three existing autocomplete widgets from vets-website, CMS, and the forms library. 

### vets-website
The GI Bill Comparison [tool](https://www.vets.gov/gi-bill-comparison-tool/) provides a search and select experience. Users can
- select from a list, or
- search any term by typing a term and pressing enter. 

Notes
- only partially accessibility compliant- it deviates from some of the idiomatic approaches documented in the [Downshift](https://github.com/paypal/downshift) library
- changes the layout of the pay to display the results i.e. the results window are not absolutely positioned. 
- does not use the forms library as a data source
gibillautofill

## CMS design sytem
The CMS Design system has an Autocomplete [component](https://design.cms.gov/components/autocomplete/) that provides a select experience. This component includes a
- loading state
- no results state
- results heading 
- clear search button 

Notes
- does not use the forms library as a data source
- externalizes the loading of data to a props functions i.e. it is a simply a view component 
- has hard dependencies on other components in the CMS design system e.g. it requires exactly one child input from a custom component type
- has a hard height limit but displays a scrollable list when the number of items exceeds that length 
cmsautocomplete

## US forms library
The US forms library has an auto suggest component that provides a select experience. Users can
- select from a list, or
- search any term by typing a term and pressing enter. 

Notes
- has existing schema and is integrated into `vets-website` on few forms 
- lacks a clear search function
- has a hard height limit but displays a scrollable list when the number of items exceeds that length 
usformsautocomplete

## Technical approach
- This approach uses the [AutosuggestField](https://github.com/usds/us-forms-system/blob/0fc545d47a10c530ee6b944a7040853e828c3d94/src/js/fields/AutosuggestField.jsx) to provide most of the functionality to connect to the current form generator.
- The render from the AutosuggestField was replaced with the [render](https://github.com/CMSgov/design-system/blob/8545784ebd078f784f287ec64e68ea0d19dcc081/packages/core/src/components/Autocomplete/Autocomplete.jsx#L110-L163) from the CMS system

complaintautocomplete

Figma [comments](http://click.figma.com/wf/click?upn=9VbdpUDlorGWg71nAPf92O2pWG3v5YSFKcYChjtVRKCP0TuIgeOroryIqhT7b1pTVFiFVEUUI4RvDz21fYLpKqt-2Fd5bzRmm5U63a97LuWKo-3D_Exu8NnyglQ-2BH0XbbqX2dDRYbD9VJ5qOpBv2M-2FYMy2mRZ8AxpWH9CHIpNbFPNjF4Vszr8DLu6Sugnbmh2QqJJctRORMgLyCK3qfxFbEG-2Bzp7rw7SOFj1Kj6eXOKVElPpNt-2F3sLrtM2Uf8KKE5j8G5PqwhiVtFZfgnec7KLGZFrf7d6VMLdVExWtkniMp6s9LTligGLcvWRb7BI2-2FBQA-2FD2hpgkub-2FXqiyYLH364nDdNJTIxtJIPEOv8W9ZI-2BJ9NU45rA6fZOuibhSJ1lIEQZtVA-3D-3D)

Todo:
- [ ] add tests
- [ ] rename classes to match vets-website
- [ ] fetch institution details callback that populates form data
- [ ] figure out how to use callback to clear all of the form data on the page

### DIFF
--- a/src/applications/edu-benefits/complaint-tool/config/form.js
+++ b/src/applications/edu-benefits/complaint-tool/config/form.js
@@ -41,24 +41,6 @@ const formConfig = {
               'Search for some schools',
               value => fetchAutocompleteSuggestions(value)
                 .then(transformAutocompletePayloadForAutosuggestField),
-              /*
-              (value) => {
-
-                return new Promise(resolve => {
-                  setTimeout(() => resolve([
-                    { label: 'option1', value: 'option1' },
-                    { label: 'option2', value: 'option1' },
-                    { label: 'option3', value: 'option1' },
-                    { label: 'option4', value: 'option1' },
-                    { label: 'option5', value: 'option1' },
-                    { label: 'option6', value: 'option1' },
-                    { label: 'option7', value: 'option1' },
-                    { label: 'option8', value: 'option1' },
-                    { label: 'option9', value: 'option1' }
-                  ]), 1000);
-                })
-              },
-              */
               {
                 'ui:options': {
                   queryForResults: true,
@@ -72,11 +54,9 @@ const formConfig = {
                   pattern: 'Please enter a valid name.'
                 }
               }
-            ),
-              {
-                'ui:field': AutosuggestField
-              }
-            )
+            ), {
+              'ui:field': AutosuggestField
+            })
           },
           schema: {
             type: 'object',
--- a/src/applications/edu-benefits/complaint-tool/config/form.js
+++ b/src/applications/edu-benefits/complaint-tool/config/form.js
@@ -38,7 +38,7 @@ const formConfig = {
           title: 'First Page',
           uiSchema: {
             name: _.merge(autoSuggestUiSchema(
-              'Label',
+              'Search for some schools',
               value => fetchAutocompleteSuggestions(value)
                 .then(transformAutocompletePayloadForAutosuggestField),
               /*
@@ -63,7 +63,8 @@ const formConfig = {
                 'ui:options': {
                   queryForResults: true,
                   freeInput: true,
-                  debounceRate: 250
+                  debounceRate: 250,
+                  listLabel: 'Select one of options or submit to search'
                 },
                 'ui:errorMessages': {
                   // If the maxLength changes, we'll want to update the message too
--- a/src/applications/edu-benefits/complaint-tool/config/form.js
+++ b/src/applications/edu-benefits/complaint-tool/config/form.js
@@ -1,9 +1,13 @@
 // import fullSchema from 'vets-json-schema/dist/686-schema.json';
+import _ from 'lodash/fp';
 import IntroductionPage from '../containers/IntroductionPage';
 import ConfirmationPage from '../containers/ConfirmationPage';
 import AutosuggestField from '../../components/AutosuggestField.jsx';
 import { uiSchema as autoSuggestUiSchema } from 'us-forms-system/lib/js/definitions/autosuggest';
-import _ from 'lodash/fp';
+import {
+  fetchAutocompleteSuggestions,
+  transformAutocompletePayloadForAutosuggestField
+} from '../helpers';
 
 // const { } = fullSchema.properties;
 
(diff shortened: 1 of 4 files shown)


src/applications/edu-benefits/complaint-tool/config/form.js
```diff
@@ -31,6 +37,26 @@ const formConfig = {
           path: 'form-page',
           title: 'First Page',
           uiSchema: {
+            name: _.merge(autoSuggestUiSchema(
```
rianfowler: This uses the existing `autoSuggestUiSchema` and replaces the `ui:field` that's used with the new version

src/applications/edu-benefits/complaint-tool/config/form.js
```diff
@@ -31,6 +37,26 @@ const formConfig = {
...
+              'Search for some schools',
+              value => fetchAutocompleteSuggestions(value)
+                .then(transformAutocompletePayloadForAutosuggestField),
+              {
+                'ui:options': {
+                  queryForResults: true,
+                  freeInput: true,
```
rianfowler: I'm not sure if this should be true or false- it's used in a [couple](https://github.com/usds/us-forms-system/blob/0fc545d47a10c530ee6b944a7040853e828c3d94/src/js/fields/AutosuggestField.jsx#L132) of places to determine if the data passed to the `onChange` prop needs to have a `widget` property. It causes the new component to throw this error when it's false:

screen shot 2018-07-19 at 08 53 18

src/applications/edu-benefits/complaint-tool/helpers.js
```diff
@@ -0,0 +1,30 @@
...
+    return {
+      options: payload
+      .data
+      .map(({ value, label }) => ({ value, label })),
+      searchTerm
+    };
+  }
+  return [];
+}
+export function fetchAutocompleteSuggestions(text, version) {
```
rianfowler: Pulling from the existing end point- this will be updated once the new autosuggest / search are available.

src/applications/edu-benefits/components/AutosuggestField.jsx
```diff
@@ -0,0 +1,296 @@
+import React from 'react';
```
rianfowler: Most of this file comes from [this](https://github.com/usds/us-forms-system/blob/0fc545d47a10c530ee6b944a7040853e828c3d94/src/js/fields/AutosuggestField.jsx).

src/applications/edu-benefits/components/AutosuggestField.jsx
```diff
@@ -0,0 +1,296 @@
...
+      this.setState({
+        input: inputValue,
+        suggestions: this.getSuggestions(this.state.options, inputValue)
+      });
+    }
+  }
+
+  handleChange = (selectedItem) => {
+    const value = this.getFormData(selectedItem);
+    this.props.onChange(value);
```
rianfowler: This component will need to call something to fetch the details about the institution during this event. Once those details are fetched, they need to populate the other form fields. 

My assumption is that the form config will pass a callback (e.g. on `ui:options`) to fetch the data and return an object that gets merged onto the existing `formData` _via_ this prop. I don't think there's away in the config to do this easily.

src/applications/edu-benefits/components/AutosuggestField.jsx
```diff
@@ -0,0 +1,296 @@
...
+                        </li>
+                      ))}
+                  </ul>
+                </div>
+              ) : null}
+
+              <button
+                aria-label={'clear all fields on this page'}
+                className="ds-u-float--right ds-u-padding-right--0"
+                onClick={e => this.handleClearSelection(e, clearSelection)}>
```
rianfowler: One thing I noticed in all three places we're using Downshift is that we don't make full use of the library. Fex Downshift provides a `clearSelection()` callback that clears the input and sets the expected focus.

src/applications/edu-benefits/sass/edu-benefits.scss
```diff
@@ -28,3 +28,77 @@
 .row .edu-intro-spacing {
   margin-bottom: 7rem;
 }
+
+// autocomplete component
+.ds-c-autocomplete__list {
```
rianfowler: Have a todo to rename these styles- for now I just used what was in CMS design system.

src/applications/edu-benefits/components/AutosuggestField.jsx
```diff
@@ -0,0 +1,296 @@
...
+  render() {
+    const { formContext, formData, uiSchema, schema } = this.props;
+
+    const {
+      loading,
+      options: items
+    } = this.state;
+    const listLabel = loading ? 'Loading...' : this.state.listLabel;
+
+    if (formContext.reviewMode) {
```
rianfowler: review and edit mode are pretty broken-  this `z-index` might cause some issues. 
screen shot 2018-07-19 at 09 08 53
screen shot 2018-07-19 at 09 10 12

src/applications/edu-benefits/components/AutosuggestField.jsx
```diff
@@ -0,0 +1,296 @@
...
+      input: '',
+      searchTerm: '',
+      loading: false
+    }));
+  }
+
+  handleInputValueChange = (inputValue) => {
+    if (inputValue !== this.state.input) {
+      const uiOptions = this.props.uiSchema['ui:options'];
+      if (uiOptions.queryForResults) {
```
rianfowler: One significant behavior change: this queries on every input change. The previous version would fetch the data once and then do a fuzzy find on it. Not sure how we'll reconcile this if we want to pass this as an update into the forms library.

src/applications/edu-benefits/sass/edu-benefits.scss
erikphansen: It would be nice to take advantage of Sass's `&` on these selectors. `.ds-c-autocomplete__list` could be at the top level, and nested in that could be `&-item {...}` and in that `&--item`. And so on. [This post](http://alwaystwisted.com/articles/2014-02-27-even-easier-bem-ing-with-sass-33) explains what I'm talking about if I'm not being clear.

rianfowler: Thanks- I suck at css so l appreciate the link.

rianfowler:
Closing this for now- need to take a different approach.

PR #8022, opened 2018-07-19
Address error formatting

bshyong:
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/11359


src/applications/personalization/profile360/components/Vet360EditModalErrorMessage.jsx
```diff
@@ -15,20 +15,20 @@ export default function Vet360EditModalErrorMessage({ error: { errors = [] }, cl
 
   switch (true) {
     case hasError(LOW_CONFIDENCE_ADDRESS_ERROR_CODES, errors):
-      content = <p>We’re sorry. We looked up the address you entered and we're not sure mail can be delivered there. Please try entering your address again.</p>;
+      content = <div className="columns"><p>We’re sorry. We looked up the address you entered and we're not sure mail can be delivered there. Please try entering your address again.</p></div>;
       break;
```
ncksllvn: Since all of these require the columns class, would it make sense to just move this into the content prop, like `content={<div className="columns">{content}</div>}`?

ncksllvn (approved):
LGTM

PR #8023, opened 2018-07-19
allow set-cookie to bubble up from the fetch

dhall:
The fetch protocol https://fetch.spec.whatwg.org/#fetch-api doesn't allow cookies to be passed along by default. In react we need to supply an additional argument to the api call to make this work. We need to verify that there are no unintended side effects of doing this... only a proof of concept for sso cookie work. Corresponding vets-api PR is here https://github.com/department-of-veterans-affairs/vets-api/pull/2140

some additional reading
https://stackoverflow.com/questions/34558264/fetch-api-with-cookie (note that same-origin no longer works so we are using include)


dhall:
@U-DON, @saneshark and @jkassemi Here is the front end changes that I needed to make for the cookie to work, just fyi

U-DON:
I think it would need to be added here as well: https://github.com/department-of-veterans-affairs/vets-website/blob/bd6f217671a00eb0f2084e2be25ba1b0a0fe3a5d/src/platform/utilities/api/index.js#L18

Although now I'm unsure we need to pass this cookie since we're still authenticating with our original token rather than the cookie.

dhall:
adding in utilities will open up all api calls to cookie setting.. we should just need whatever point we initially set it via sign in. It's setup to merge in the options to the defaults so I'm passing it as an option from the auth call. I could be wrong but locally that allow the cookie to be set, then it's resident in the browser between calls and only over rides the default in that one place. 

Regarding passing the cookie, it's the other way around.. the api passes the set-cookie header back from the auth call as a response cookie .. my change enables the response cookie to be passed to the browser.

Sorry if I'm missing your point, just let me know and we can chat :)

U-DON:
Missed that it's a shortcoming of fetch that it needs to set `credentials` to allow the cookie to be set from the response. Ideally, its only purpose would be to send the cookie with the request, but looks like we need it here for the former reason. Carry on.

saneshark:
@dhall it might be worth running through a quick manual regression, but I'm thinking since specs are not broken this should be ok. Thoughts @U-DON ? I'll defer to you for approval.

U-DON (approved):
Tests should be passing now!

dhall:
Thanks @U-DON for figuring out the tests .. you ROCK!

PR #8030, opened 2018-07-20
Set up goBack on opt out form page

cehsu:
These changes send the user back to the introduction page when pressing the back button on the opt out form per @emilywaggoner's [comment](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/11331#issuecomment-405998236).

### DIFF
--- a/src/applications/edu-benefits/0993/containers/SubmitController.jsx
+++ b/src/applications/edu-benefits/0993/containers/SubmitController.jsx
@@ -27,7 +27,11 @@ class SubmitController extends React.Component {
   }
 
   goBack = () => {
-    this.props.router.goBack();
+    const formPageUrl = window.location.pathname;
+    let introductionPageUrl = formPageUrl.split('/');
+    introductionPageUrl.splice(-2);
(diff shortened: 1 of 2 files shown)


src/applications/edu-benefits/0993/containers/SubmitController.jsx
```diff
@@ -27,7 +27,11 @@ class SubmitController extends React.Component {
   }
 
   goBack = () => {
-    this.props.router.goBack();
```
rianfowler: We could add a property to the manifest to accomplish this. It's weird case where our `rootUrl` doesn't redirect to the intro page.

rianfowler (approved):
lgtm

src/applications/edu-benefits/0993/containers/SubmitController.jsx
cehsu: Yes, good idea, and that could also potentially be used to add the static page to the ignorelist, if an additional production flag for the static page is provided. Will create a ticket for this.

PR #8033, opened 2018-07-20
Profile: Add functionality for copying mailing address into residential edit-modal

ncksllvn:
We badly need to consolidate our logic between `consolidateAddress`, `cleanAddressDataForUpdate`, and `createAddressObject`. I think there should only be two forms - the natural address shape from the server, and the consolidated form for rendering in the edit-modal. I think we tied in extra conversations when transitioning between PCIU to Vet360 format. This work is noted here, department-of-veterans-affairs/vets.gov-team/issues/11561.

##### Todo
- [x] Test cases

Resolves department-of-veterans-affairs/vets.gov-team/issues/11298

### DIFF
--- a/src/applications/personalization/profile360/components/AddressEditModal.jsx
+++ b/src/applications/personalization/profile360/components/AddressEditModal.jsx
@@ -1,11 +1,17 @@
 import React from 'react';
 
 import {
+  FIELD_NAMES
+} from '../constants/vet360';
+
+import {
   consolidateAddress,
   expandAddress
 } from '../../../../platform/forms/address/helpers';
 
 import Vet360EditModal from './Vet360EditModal';
+
+import CopyMailingAddress from '../containers/CopyMailingAddress';
 import Address from './Address';
 
 export default class AddressEditModal extends React.Component {
@@ -34,15 +40,22 @@ export default class AddressEditModal extends React.Component {
     };
   }
 
+  copyMailingAddress = (consolidatedMailingAddress) => {
+    this.props.onChange(consolidatedMailingAddress, null, true);
+  }
+
   renderForm = () => {
     return (
-      <Address
-        address={this.props.field.value}
-        onInput={this.onInput}
-        onBlur={this.onBlur}
-        errorMessages={this.props.field.validations}
-        states={this.props.addressConstants.states}
-        countries={this.props.addressConstants.countries}/>
+      <div>
+        {this.props.fieldName === FIELD_NAMES.RESIDENTIAL_ADDRESS && <CopyMailingAddress copyMailingAddress={this.copyMailingAddress}/>}
+        <Address
+          address={this.props.field.value}
+          onInput={this.onInput}
+          onBlur={this.onBlur}
+          errorMessages={this.props.field.validations}
+          states={this.props.addressConstants.states}
+          countries={this.props.addressConstants.countries}/>
+      </div>
     );
   }
 
--- a/src/applications/personalization/profile360/components/AddressEditModal.jsx
+++ b/src/applications/personalization/profile360/components/AddressEditModal.jsx
@@ -40,8 +40,8 @@ export default class AddressEditModal extends React.Component {
     };
   }
 
-  copyMailingAddress = (consolidatedMailingAddress) => {
-    this.props.onChange(consolidatedMailingAddress, null, true);
+  copyMailingAddress = (mailingAddress) => {
+    this.props.onChange(mailingAddress, null, true);
   }
 
   renderForm = () => {
--- a/src/applications/personalization/profile360/containers/CopyMailingAddress.jsx
+++ b/src/applications/personalization/profile360/containers/CopyMailingAddress.jsx
@@ -4,7 +4,6 @@ import isEqual from 'lodash/isEqual';
 import pick from 'lodash/pick';
 
 import {
-  consolidateAddress,
   isEmptyAddress
 } from '../../../../platform/forms/address/helpers';
 
@@ -25,9 +24,7 @@ class CopyMailingAddress extends React.Component {
   onChange = (event) => {
     event.stopPropagation();
     if (event.target.checked) {
-      this.props.copyMailingAddress(
-        consolidateAddress(this.props.mailingAddress)
-      );
+      this.props.copyMailingAddress(this.props.mailingAddress);
     }
   }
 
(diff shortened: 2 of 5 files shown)


src/applications/personalization/profile360/containers/CopyMailingAddress.jsx
```diff
@@ -0,0 +1,90 @@
...
+  ];
+
+  const mailingAddress = selectVet360Field(state, FIELD_NAMES.MAILING_ADDRESS);
+  const hasEmptyMailingAddress = !mailingAddress;
+
+  const residentialAddress = selectEditedFormField(state, FIELD_NAMES.RESIDENTIAL_ADDRESS).value;
+
+  const mailingAddressConsolidated = pick(cleanAddressDataForUpdate(consolidateAddress(mailingAddress)), addressProps);
+  const residentialAddressConsolidated = pick(consolidateAddress(residentialAddress), addressProps);
+
```
ncksllvn: This is a crazy amount of conversions. We need to revisit this when we clean up the address-formatting logic.

bshyong: Yep; the address format has 3 forms that we need to deal with
1. natural address shape from the server (stored as-is in the user object)
2. formatted address shape for rendering in the edit modal
3. stripped/cleaned up address shape for sending to Vet360 for an update. we will get rejections if there are any extraneous attributes/inputs that are unexpected

src/applications/personalization/profile360/containers/CopyMailingAddress.jsx
```diff
@@ -0,0 +1,90 @@
...
+import {
+  cleanAddressDataForUpdate
+} from '../actions/misc';
+
+class CopyMailingAddress extends React.Component {
+  onChange = (event) => {
+    event.stopPropagation();
+    if (event.target.checked) {
+      this.props.copyMailingAddress(
+        consolidateAddress(this.props.mailingAddress)
```
bshyong: @ncksllvn do we need consolidateAddress here? When we call `copyMailingAddress`, which in turn calls `onChange`, that hits the `updateProfileFormField` action. In that action, `cleanAddressDataForUpdate` is automatically called

src/applications/personalization/profile360/containers/CopyMailingAddress.jsx
```diff
@@ -0,0 +1,90 @@
...
+    'zipCode'
+  ];
+
+  const mailingAddress = selectVet360Field(state, FIELD_NAMES.MAILING_ADDRESS);
+  const hasEmptyMailingAddress = !mailingAddress;
+
+  const residentialAddress = selectEditedFormField(state, FIELD_NAMES.RESIDENTIAL_ADDRESS).value;
+
+  const mailingAddressConsolidated = pick(cleanAddressDataForUpdate(consolidateAddress(mailingAddress)), addressProps);
+  const residentialAddressConsolidated = pick(consolidateAddress(residentialAddress), addressProps);
```
bshyong: @ncksllvn I think `consolidateAddress` is a vestige of PICU functionality; when we submit to Vet360, attributes like `militaryPostOfficeTypeCode` and `militaryStateCode` are automatically stripped away already. Why is `cleanAddressDataForUpdate` only called for `mailingAddress` and not also for `residentialAddress`?

src/applications/personalization/profile360/components/AddressEditModal.jsx
```diff
@@ -1,11 +1,17 @@
 import React from 'react';
 
+import {
+  FIELD_NAMES
+} from '../constants/vet360';
+
 import {
   consolidateAddress,
   expandAddress
```
bshyong: We need to get rid of `expandAddress` here and in the `onSubmit`. This one is from the old Letters implementation; but we do our own address inference here: https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/applications/personalization/profile360/actions/misc.js#L99

The expandAddress in this form helper also uses the wrong type for military address; Vet360 expects `MILITARY OVERSEAS` instead of `MILITARY`

ncksllvn: Yeah that makes sense, I didn't know if there was still some need for that transformation. We transitioned from PCIU to Vet360 addresses in a blur, and I wasn't sure how well I followed the new logic. 

I feel like this is out of the scope of this PR though. Do you think it makes more sense to tackle this during the upcoming refactor?

src/applications/personalization/profile360/containers/CopyMailingAddress.jsx
ncksllvn: Following the transformations every step was important because the checkbox needed to stay in sync at all times (even if they copy over the mailing address, they could still change a value), and it was tricky trying to apply transformations on edited form values to compare with the API's value for the other address.

src/applications/personalization/profile360/containers/CopyMailingAddress.jsx
```diff
@@ -0,0 +1,88 @@
...
+    'countryName',
+    'internationalPostalCode',
+    'province',
+    'stateCode',
+    'zipCode'
+  ];
+
+  const mailingAddress = selectVet360Field(state, FIELD_NAMES.MAILING_ADDRESS);
+  const hasEmptyMailingAddress = isEmptyAddress(mailingAddress);
+
```
ncksllvn: @bshyong  is it correct to still be using `isEmptyAddress`, or will empty addresses just be `null` for Vet360?

src/applications/personalization/profile360/tests/containers/CopyMailingAddress.unit.spec.jsx
```diff
@@ -0,0 +1,95 @@
...
+          internationalPostalCode: null,
+          province: null,
+          stateCode: 'NY',
+          zipCode: '97062'
+        }
+      };
+
+      const result = mapStateToProps(state);
+      expect(result.checked).to.be.true;
+    });
```
ncksllvn: @bshyong this unit test should cover the ultimate address-consolidation process.

ncksllvn:
@bshyong Made some updates and removed `consolidateAddress`. I think some address stuff could wait for the full cleanup, but we can bring them in to this PR if you prefer.

bshyong (approved):

src/applications/personalization/profile360/components/AddressEditModal.jsx
bshyong: Agreed; I will add it to that issue

PR #8037, opened 2018-07-23
Opt Out copy changes, fixes #11221

erikphansen:
Please let me know the preferred way to link a PR to a Zenhub issue. Just putting the issue number in the PR headline works but it doesn't look like that's what other folks are doing.

Screenshots of the changes:

screen shot 2018-07-23 at 4 02 38 pm
screen shot 2018-07-23 at 4 04 48 pm

### DIFF
--- a/src/applications/claims-status/sass/claims-status.scss
+++ b/src/applications/claims-status/sass/claims-status.scss
@@ -307,6 +307,7 @@ $color-inactive: #9b9b9b;
   font-weight: bold;
   text-decoration: none;
   margin: 0.5em 0;
+  word-break: keep-all;
 }
 
 .clearfix:after {
--- a/src/applications/claims-status/sass/claims-status.scss
+++ b/src/applications/claims-status/sass/claims-status.scss
@@ -307,7 +307,6 @@ $color-inactive: #9b9b9b;
   font-weight: bold;
   text-decoration: none;
   margin: 0.5em 0;
-  word-break: keep-all;
 }
 
 .clearfix:after {
--- a/src/platform/forms/sass/_m-schemaform.scss
+++ b/src/platform/forms/sass/_m-schemaform.scss
@@ -204,9 +204,6 @@ label {
 .schemaform-array-row {
   outline: none;
 }
-.schemaform-array-row {
-  outline: none;
-}
 
 .schemaform-block-header {
   margin-bottom: 0.7em;
@@ -293,9 +290,10 @@ fieldset.schemaform-field-template {
 }
 
 .help-phone-number-link {
-    font-weight: bold;
-    text-decoration: none;
-    margin: 0.5em 0;
+  font-weight: bold;
+  text-decoration: none;
+  margin: 0.5em 0;
+  word-break: keep-all;
 }
 
 //duplicated/renamed from claims-status.scss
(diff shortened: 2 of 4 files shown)


src/platform/forms/sass/_m-schemaform.scss
```diff
@@ -293,9 +290,10 @@ fieldset.schemaform-field-template {
 }
 
 .help-phone-number-link {
-    font-weight: bold;
-    text-decoration: none;
-    margin: 0.5em 0;
+  font-weight: bold;
+  text-decoration: none;
+  margin: 0.5em 0;
+  word-break: keep-all;
```
erikphansen: This prevents the phone number from wrapping across two lines in the modal. This behavior is approved by @emilywaggoner.

src/applications/claims-status/sass/claims-status.scss
```diff
@@ -307,6 +307,7 @@ $color-inactive: #9b9b9b;
   font-weight: bold;
   text-decoration: none;
   margin: 0.5em 0;
+  word-break: keep-all;
```
erikphansen: This prevents the phone number from wrapping across two lines. This behavior is approved by @emilywaggoner.

cehsu (approved):
LGTM, pending @emilywaggoner and @peggygannon approval.

src/applications/claims-status/sass/claims-status.scss
cehsu: It looks like we should set up a ticket to better organize the styles, it seems like edu styles shouldn't be affected by claims styles? Also, does this change affect any claims behavior? If so, we should document that as well via screenshot in this PR for approval.

peggygannon (commented):
@erikphansen HI, Erik, I made a couple additional proofread tweaks. Otherwise, the content updates look  good.

src/applications/claims-status/sass/claims-status.scss
erikphansen: Yes, the Sass organization could use a lot of love, based on my brief time with it. That's a dangerous game to play if there aren't really good visual regression tests in place before starting the process. But it's something that I'd like to do. This change will probably affect other phone numbers on the site. But that is a good thing because a phone number probably shouldn't ever wrap to a new line.

erikphansen:
@peggygannon cool, your changes to that file are all that are needed so we should be good there. Unless you'd like me to update the screenshots to show your latest tweaks?

peggygannon:
@erikphansen Yes, that would be great. I'll give the updated screenshots a quick look before we merge this into staging.

src/applications/claims-status/sass/claims-status.scss
cehsu: Thanks @erikphansen could you please add a comment to this file that we should move these to a shared style file, and out of this application specific file?

erikphansen:
Updated screenshot showing changes from @peggygannon 

screen shot 2018-07-23 at 4 46 44 pm

peggygannon:
@erikphansen OK, thanks. Looks good.

peggygannon (approved):

emilywaggoner (approved):
Looks good to me. Thanks!

PR #8038, opened 2018-07-23
Add opt out form to profile page

cehsu:
These changes add the opt out form to the profile page.


erikphansen (commented):
Looks good, but I'd love to refactor `helpers.jsx` to stop reusing error-prone strings for keys on different dictionaries.

erikphansen (approved):

cehsu:
Indeed @erikphansen, it could be good to refactor to pull this information from the manifest.json files instead.

PR #8039, opened 2018-07-24
Use localStorage for session data (with feature toggle)

U-DON:
Resolves department-of-veterans-affairs/vets.gov-team#9737.

This change persists the logged in session across tabs.

To toggle on, run in your browser's dev tools console: `localStorage.setItem('storageType', 'localStorage')`.


saneshark:
Is this work done and ready to be merged or does it still need tests? Lets try to get this reviewed tomorrow if possible.

U-DON:
@ncksllvn: Tagging since you implemented the localStorage fallback logic. We're going to start storing session data in localStorage so that sessions can persist across tabs. There are some things that will still  live in sessionStorage, and it makes sense for most of them. But in particular, I noticed there was some rate limiting data that was stored in sessionStorage, and I wasn't sure how it would be affected by this type of session persistence or if it should be moved to localStorage as well. I think it will be okay to keep out of the scope of these changes for now but wanted to check with you first.

https://github.com/department-of-veterans-affairs/vets-website/blob/eb5dc1c552703ef2061f050d8f5b738dd9c957e1/src/platform/monitoring/RateLimiter.jsx#L51

dhall (approved):
LGTM!
minor: conditionalStorage is a bit confusing as a name.. but I don't have a better suggestion :)

U-DON:
@ncksllvn: This PR won't be centralizing sessionStorage in the same way as localStorage. Rather, it'll use the centralized localStorage (with a feature toggle) in places where we used to use window.sessionStorage for session data. However, I've come across some things that need to change that would have also been an obstacle for department-of-veterans-affairs/vets.gov-team#11224, like the use of window.sessionStorage in our static page widgets.

src/applications/static-pages/static-page-widgets.js
```diff
@@ -6,7 +6,7 @@ function mountWidgets(widgets) {
       var slowLoadingThreshold = (widget.slowLoadingThreshold || 6) * 1000;
       var slowMessage = widget.slowMessage || 'Sorry, this is taking longer than expected.';
 
-      if (!widget.showSpinnerUnauthed && sessionStorage.userToken) {
+      if (!widget.showSpinnerUnauthed) {
```
U-DON: I don't think the sessionStorage check here made a difference... I still saw a blink of the loading indicator when I went to the relevant pages in Prod. Probably safe to remove here.

@cehsu: I noticed you added the application status widget to some of the static pages, so just letting you know in case that wasn't expected behavior.

PR #8043, opened 2018-07-24
EVSS Gate

cvalarida:
This PR does 3 things:
1. Includes a `EVSSClaimsGate` which doesn't allow the user into the form unless the they have `evss-claims` in their services list
2. Wraps everything under `EVSSClaimsGate` in a `RequiredLoginView` so we're not trying to make an ITF call when we know it'll fail
    - We were still redirecting to the login before, but only after the ITF call
    - I don't know _how_ we were redirecting to the login screen before, so I wasn't able to remove that... :grimacing: 
3. Fixes a newly-introduced bug wherein the ITF would be fetched twice in the following scenario:
    1. We log in
    2. We go to the middle of a form
        - The ITF is fetched / created here
    3. We back out to the intro page using the form controls
    4. We start the form
        - The ITF was being fetched again here in `ITFWrapper`'s `componentDidMount`

### Screenshot
image

### DIFF
--- a/src/applications/disability-benefits/526EZ/Form526EZApp.jsx
+++ b/src/applications/disability-benefits/526EZ/Form526EZApp.jsx
@@ -5,15 +5,18 @@ import formConfig from './config/form';
 
 import ITFWrapper from './containers/ITFWrapper';
 import DisabilityGate from './containers/DisabilityGate';
+import EVSSClaimsGate from './containers/EVSSClaimsGate';
 
 export default function Form526Entry({ location, children }) {
   return (
-    <ITFWrapper location={location}>
-      <DisabilityGate>
-        <RoutedSavableApp formConfig={formConfig} currentLocation={location}>
-          {children}
-        </RoutedSavableApp>
-      </DisabilityGate>
-    </ITFWrapper>
+    <EVSSClaimsGate location={location}>
+      <ITFWrapper location={location}>
+        <DisabilityGate>
+          <RoutedSavableApp formConfig={formConfig} currentLocation={location}>
+            {children}
+          </RoutedSavableApp>
+        </DisabilityGate>
+      </ITFWrapper>
+    </EVSSClaimsGate>
   );
 }
--- a/src/applications/disability-benefits/526EZ/containers/EVSSClaimsGate.jsx
+++ b/src/applications/disability-benefits/526EZ/containers/EVSSClaimsGate.jsx
@@ -18,7 +18,7 @@ export function EVSSClaimsGate({ user, location, children }) {
         <AlertBox
           isVisible
           headline="We’re sorry. It looks like we’re missing some information needed for your application"
-          content="Please call Veterans Benefits Assistance at 1-800-827-1000, Monday – Friday, 8:00 a.m. to 9:00 p.m. (ET)."
+          content="For help with your application, please call Veterans Benefits Assistance at 1-800-827-1000, Monday – Friday, 8:00 a.m. to 9:00 p.m. (ET)."
           status="error"/>
       </div>
     );
--- a/src/applications/disability-benefits/526EZ/containers/EVSSClaimsGate.jsx
+++ b/src/applications/disability-benefits/526EZ/containers/EVSSClaimsGate.jsx
@@ -17,8 +17,8 @@ export function EVSSClaimsGate({ user, location, children }) {
       <div className="usa-grid full-page-alert">
         <AlertBox
           isVisible
-          headline="Nope"
-          content="Looks like we don't have all the information we need. Please call the call center."
+          headline="We’re sorry. It looks like we’re missing some information needed for your application"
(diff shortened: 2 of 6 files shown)


cvalarida:
@peggygannon This error message will be seen if:
1) The user is logged in
2) The user is missing at least one of:
  i. EDIPI (which is a military ID of some sort)
  ii. SSN
  iii. Participant ID (which means...something something they "haven't interfaced with a corporate database backed VA service before," according to @kreek)
3) They try to start the form

This happens before the ITF call. In fact, we _can't_ make the ITF call unless we've got all those IDs.

I _think_ the only path forward for them is to call the call center, but maybe @djmunoz1 knows more? :man_shrugging:

peggygannon:
@cvalarida @djmunoz1 see below for my suggested message for this scenario? 

Can you tell me what would be the next step for the user if they get this message? Do they try again, or should they call the call center? I put both in the message, but I want to be sure that we're giving users a path forward. 


**We’re sorry. It looks like we’re missing some information needed for your application**
Please try again later or call Veterans Benefits Assistance at 1-800-827-1000, Monday – Friday, 8:00 a.m. to 9:00 p.m. (ET).

cvalarida:
As far as we know, the next step is to call the call center. Trying again won't help, unfortunately.

peggygannon:
@cvalarida OK, thanks for the clarification. Will the message I suggest below work for this scenario? 
cc @djmunoz1 

**We’re sorry. It looks like we’re missing some information needed for your application**
For help with your application, please call Veterans Benefits Assistance at 1-800-827-1000, Monday – Friday, 8:00 a.m. to 9:00 p.m. (ET).

src/applications/disability-benefits/526EZ/containers/EVSSClaimsGate.jsx
```diff
@@ -0,0 +1,41 @@
...
+import RequiredLoginView from '../../../../platform/user/authorization/components/RequiredLoginView';
+import backendServices from '../../../../platform/user/profile/constants/backendServices';
+
+export function EVSSClaimsGate({ user, location, children }) {
+  // Short-circuit the check on the intro page
+  if (location.pathname === '/introduction') {
+    return children;
+  }
+
+  if (user.login.currentlyLoggedIn && !user.profile.services.includes(backendServices.EVSS_CLAIMS)) {
```
markgreenburg: Does this in a way duplicate what `RequiredLoginView` already does? If the required service isn't in the user's services array, RequiredLoginView renders its children and also passes `isDataAvailable = false` to them. So, this could live in the `ITFWrapper` if we were using the standard pattern. Not sure how much difference it makes at the end of the day.

src/applications/disability-benefits/526EZ/containers/ITFWrapper.jsx
```diff
@@ -42,7 +42,8 @@ export class ITFWrapper extends React.Component {
   // When we first enter the form...
   componentDidMount() {
     // ...fetch the ITF if needed
-    if (!noITFPages.includes(this.props.location.pathname)) {
+    if (!noITFPages.includes(this.props.location.pathname)
+      && this.props.itf.fetchCallState === requestStates.notCalled) {
```
markgreenburg: 💯

markgreenburg (approved):
LGTM. Had a question on the implementation of the 'unauthorized' error implementation but nothing major.

src/applications/disability-benefits/526EZ/Form526EZApp.jsx
```diff
@@ -5,15 +5,18 @@ import formConfig from './config/form';
...
-    </ITFWrapper>
+    <EVSSClaimsGate location={location}>
+      <ITFWrapper location={location}>
+        <DisabilityGate>
+          <RoutedSavableApp formConfig={formConfig} currentLocation={location}>
+            {children}
+          </RoutedSavableApp>
+        </DisabilityGate>
+      </ITFWrapper>
+    </EVSSClaimsGate>
```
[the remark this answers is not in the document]
cvalarida: I totally agree. The alternative is handling all these discreet tasks in one big wrapper, but I figured it would make for easier reasoning this way. Also, if we have to extract or modify something later, it should be pretty straightforward.

But yeah. I totally agree this is getting silly.

src/applications/disability-benefits/526EZ/containers/EVSSClaimsGate.jsx
cvalarida: Yeah, but here's where it gets hairy:
1) We want to render the intro page without requiring the veteran to log in yet
2) We want to handle the gating in this component instead of the child component(s)
3) Implementing _another_ component to put between `RequiredLoginView` and `children` to check for the `isDataAvailable` prop is possible, but I think would be more difficult to reason about than doing it this way

It _is_ duplicative, but it's for simplicity. :man_shrugging:

PR #8048, opened 2018-07-24
Remove delay before redirecting the user post-login

ncksllvn:
After a user logs in, we receive the token from the sign-in window via the `setToken` callback, where we then use that token to request the user's profile information. We then store that user profile information in state. Finally, React components respond to this new data, one of which is the `user-nav/containers/Main.jsx`, which will perform a redirect to a certain application based on the "next" query parameter. The user is redirected to `/dashboard/` if they're on the homepage, `/`.

This PR refactors the login logic so that the application-redirect is executed after the FE is handed a `userToken` but prior to the user-profile being initialized. This way, we only issue the request for the user-profile when we don't need to do a redirect, so that immediately post-login we save some resources and save the user that second of delay (while the request for the user-profile is pending) before being redirected.

There's some more context in the issue below. In summary, it's confusing that after logging in the user sees the regular homepage for just a second before the redirect to their dashboard.

Resolves department-of-veterans-affairs/vets.gov-team/issues/10659

### DIFF
--- a/src/platform/site-wide/user-nav/containers/Main.jsx
+++ b/src/platform/site-wide/user-nav/containers/Main.jsx
@@ -22,6 +22,10 @@ import {
 import SearchHelpSignIn from '../components/SearchHelpSignIn';
 import { selectUserGreeting } from '../selectors';
 
+import dashboardManifest from '../../../../applications/personalization/dashboard/manifest.json';
+
+const DASHBOARD_URL = dashboardManifest.rootUrl;
+
 // const SESSION_REFRESH_INTERVAL_MINUTES = 45;
 
 export class Main extends React.Component {
@@ -38,17 +42,9 @@ export class Main extends React.Component {
 
   componentDidUpdate() {
     const { currentlyLoggedIn, showLoginModal } = this.props;
-    const nextParam = this.getRedirectUrl();
-
-    const shouldRedirect =
-      currentlyLoggedIn && nextParam && !window.location.pathname.includes('verify');
-
-    if (shouldRedirect) {
-      const redirectPath = nextParam.startsWith('/') ? nextParam : `/${nextParam}`;
-      window.location.replace(redirectPath);
-    }
-
     const shouldCloseLoginModal = currentlyLoggedIn && showLoginModal;
+
+    if (currentlyLoggedIn) this.executeRedirect();
 
     if (shouldCloseLoginModal) { this.props.toggleLoginModal(false); }
   }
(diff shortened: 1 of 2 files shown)


src/platform/site-wide/user-nav/containers/Main.jsx
```diff
@@ -38,18 +42,10 @@ export class Main extends React.Component {
...
-    if (shouldRedirect) {
-      const redirectPath = nextParam.startsWith('/') ? nextParam : `/${nextParam}`;
-      window.location.replace(redirectPath);
-    }
-
     const shouldCloseLoginModal = currentlyLoggedIn && showLoginModal;
 
+    if (currentlyLoggedIn) this.executeRedirect();
+
     if (shouldCloseLoginModal) { this.props.toggleLoginModal(false); }
```
ncksllvn: The removed code was largely moved into `executeRedirect`, which redirects the user if the URL warrants it.

src/platform/site-wide/user-nav/containers/Main.jsx
```diff
@@ -58,10 +54,35 @@ export class Main extends React.Component {
   }
 
   setToken = (event) => {
-    if (event.data === sessionStorage.userToken) { this.props.initializeProfile(); }
+    if (event.data === sessionStorage.userToken) {
+      this.executeRedirect();
+      this.props.initializeProfile();
+    }
   }
```
ncksllvn: Here, `executeRedirect` is called again, this time immediately after receiving the token from the sign-in window and bypassing the profile-initialization logic if a redirect is performed.

src/platform/site-wide/user-nav/containers/Main.jsx
```diff
@@ -58,10 +54,35 @@ export class Main extends React.Component {
...
+    return false;
+  }
+
+  getRedirectUrl = () => {
+    const nextParam = this.getRedirectUrlParameter();
+    if (nextParam) return nextParam;
+
+    return window.location.pathname === '/' && DASHBOARD_URL;
+  };
+
```
ncksllvn: I separated out `getRedirectUrlParameter` from `getRedirectUrl` because `getRedirectUrlParameter` is used to determine whether we should automatically open the login modal. Using `getRedirectUrl` instead would also take the user's dashboard into consideration, which means the sign-in modal would automatically open on the homepage.

ncksllvn:
@U-DON Can I get a CR on this when you have time? I think it's safe and a nice speedup, but I'm not sure of there's any logic that's being bypassed by doing the redirect so early on.

U-DON (approved):
LGTM. Had some minor comments/questions.

src/platform/site-wide/user-nav/containers/Main.jsx
```diff
@@ -58,10 +54,35 @@ export class Main extends React.Component {
...
   setToken = (event) => {
-    if (event.data === sessionStorage.userToken) { this.props.initializeProfile(); }
+    if (event.data === sessionStorage.userToken) {
+      this.executeRedirect();
+      this.props.initializeProfile();
+    }
   }
 
-  getRedirectUrl = () => (new URLSearchParams(window.location.search)).get('next');
+  getRedirectUrlParameter() {
```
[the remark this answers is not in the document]
ncksllvn: Good idea, will update!

bshyong (approved):

PR #8053, opened 2018-07-25
Profile: Fetch transactions only after confirming user is in Vet360 

ncksllvn:
Resolves department-of-veterans-affairs/vets.gov-team/issues/11694

### DIFF
--- a/src/applications/personalization/profile360/components/ProfileView.jsx
+++ b/src/applications/personalization/profile360/components/ProfileView.jsx
@@ -20,6 +20,7 @@ class ProfileView extends React.Component {
     downtimeData: PropTypes.object,
     isVet360AvailableForUser: PropTypes.bool,
     fetchAddressConstants: PropTypes.func.isRequired,
+    fetchTransactions: PropTypes.func.isRequired,
     fetchMilitaryInformation: PropTypes.func.isRequired,
     fetchHero: PropTypes.func.isRequired,
     fetchPersonalInformation: PropTypes.func.isRequired,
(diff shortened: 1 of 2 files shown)


src/applications/personalization/profile360/components/ProfileView.jsx
```diff
@@ -15,6 +16,21 @@ import IdentityVerification from './IdentityVerification';
...
+    fetchHero: PropTypes.func.isRequired,
+    fetchPersonalInformation: PropTypes.func.isRequired,
+    profile: PropTypes.shape({
+      addressConstants: PropTypes.object,
+      hero: PropTypes.object,
+      personalInformation: PropTypes.object,
+      militaryInformation: PropTypes.object
+    }),
+    user: PropTypes.object
+  };
```
ncksllvn: This isn't necessary at all, but wanted to go ahead and define this.

src/applications/personalization/profile360/containers/VAProfileApp.jsx
```diff
@@ -21,22 +21,30 @@ import RequiredLoginView from '../../../../platform/user/authorization/component
...
+    const justLoggedIn = !oldProps.user.login.currentlyLoggedIn && this.props.user.login.currentlyLoggedIn;
+    if (justLoggedIn) {
+      if (this.props.isVet360AvailableForUser) {
+        this.props.fetchTransactions();
+      } else {
+        // this.props.initializeUserToVet360()
+      }
+    }
   }
+
```
ncksllvn: Rather than check for a `null` value for the `profile.vet360`, we use the selector `isVet360AvailableForUser` that checks for `vet360` in the `profile.services` list.

src/applications/personalization/profile360/containers/VAProfileApp.jsx
```diff
@@ -55,7 +63,7 @@ class VAProfileApp extends React.Component {
 
 const mapStateToProps = (state) => {
   return {
-    account: state.user,
+    user: state.user,
     profile: state.vaProfile,
```
ncksllvn: I know the original reason for this, but I went ahead and corrected this because this is too tricky.

bshyong (approved):

PR #8061, opened 2018-07-25
Profile: Reorganize and restructure Vet360 components, reducer for portability 

ncksllvn:
Documentation describing to other engineers how to use our Vet360 components will come in a later PR so that they aren't exposed to all of this reorganization/cleanup.

##### Todo
- [x] Confirm portability
    -  I confirmed portability by disabling the index (`vaProfile`) reducer, then disabling the `Hero`, `PersonalInformtion`, and `MilitaryInformation` components. This ensured that the `ContactInformation` and its children components were only reliant on the `vet360` reducer.
- [x] Update tests. Specifically, the reducers

##### Changes
These are the changes necessary to make the Vet360 portion of the Profile portable into other applications.
- Actions
    - Removes all references to the `addressConstants`, and instead directly imports the constants in the `AddressEditModal`.
- Components
    - Creates a Vet360 component for each field, rather than passing configuration to each field-type component and implement them all in `ContactInformation`
    - Derives the `analyticsSectionName` properties in the container, rather than passing it as a prop
- Reducers
    - Moves the `modal`, `formFields` properties from the `vaProfile` reducer into `vet360`
    - Moves the `vet360.js` reducer to `profile360/vet360/reducers/index.js`

##### Relevant issues
department-of-veterans-affairs/vets.gov-team/issues/11649
department-of-veterans-affairs/vets.gov-team/issues/11641
department-of-veterans-affairs/vets.gov-team/issues/11650

Resolves department-of-veterans-affairs/vets.gov-team/issues/11646

### DIFF
--- a/src/applications/personalization/profile360/components/ContactInformation.jsx
+++ b/src/applications/personalization/profile360/components/ContactInformation.jsx
@@ -4,15 +4,20 @@ import React from 'react';
 import DowntimeNotification, { externalServices } from '../../../../platform/monitoring/DowntimeNotification';
 import recordEvent from '../../../../platform/monitoring/record-event';
 import accountManifest from '../../account/manifest.json';
-import { FIELD_NAMES, TRANSACTION_CATEGORY_TYPES } from '../constants/vet360';
+import { TRANSACTION_CATEGORY_TYPES } from '../constants/vet360';
 import Vet360PendingTransactionCategory from '../containers/Vet360PendingTransactionCategory';
-import PhoneSection from './PhoneSection';
-import AddressSection from './AddressSection';
-import EmailSection from './EmailSection';
 import LoadFail from './LoadFail';
 import { handleDowntimeForSection } from './DowntimeBanner';
 import MissingVet360IDError from './MissingVet360IDError';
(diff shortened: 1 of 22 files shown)


ncksllvn:
@bshyong This is ready for review when you have time!

src/applications/personalization/profile360/components/ContactInformation.jsx
```diff
@@ -4,29 +4,29 @@ import React from 'react';
...
+import MailingAddress from '../vet360/components/MailingAddress';
+import ResidentialAddress from '../vet360/components/ResidentialAddress';
+import HomePhone from '../vet360/components/HomePhone';
+import MobilePhone from '../vet360/components/MobilePhone';
+import WorkPhone from '../vet360/components/WorkPhone';
+import FaxNumber from '../vet360/components/FaxNumber';
+import Email from '../vet360/components/Email';
 
-  componentDidMount() {
-    this.props.fetchAddressConstants();
```
bshyong: Noting that is included because an endpoint does exist for us to dynamically pull state and country names. Switching to a hardcoded file because we do not expect the states/countries to have changes in the near term.

bshyong (approved):

PR #8068, opened 2018-07-26
Switch to using an aria-label directly on the button element.

erikphansen:
This seems to be a more modern, standards-compliant way to add
screen reader hints and fixes department-of-veterans-affairs/vets.gov-team#11698


rianfowler (approved):

cehsu (approved):
LGTM, looping @1Copenut in for additional thoughts.

1Copenut (approved):
👍 LGTM

I like the use of `aria-label` as the standard solution for longer aural labels. It is well-supported across VoiceOver, NVDA, and JAWS, and doesn't require an additional `<span>` or other semi-semantic markup. 

I tested code locally on the global footer modal with all 3 screen readers + their preferred browsers. All passed with no issue.

PR #8071, opened 2018-07-27
add 546 form to ignore list

rianfowler:
This form shouldn't be live yet as per @bethpotts. 

screen shot 2018-07-27 at 13 47 06
---
Adding to ignore list removes the form link from production: 
screen shot 2018-07-27 at 13 43 58

### DIFF
--- a/script/build.js
+++ b/script/build.js
@@ -129,6 +129,7 @@ if (options.buildtype === 'production') {
   ignoreList.push('veteran-id-card/how-to-upload-photo.md');
   ignoreList.push('education/opt-out-information-sharing.md');
   ignoreList.push('education/complaint-tool.md');
+  ignoreList.push('disability-benefits/apply/form-526-disability-claim.md');
 }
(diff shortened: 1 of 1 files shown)


cehsu (approved):

erikphansen (approved):
Cool. What does this change actually do? :)

script/build.js
```diff
@@ -129,6 +129,7 @@ if (options.buildtype === 'production') {
   ignoreList.push('veteran-id-card/how-to-upload-photo.md');
   ignoreList.push('education/opt-out-information-sharing.md');
   ignoreList.push('education/complaint-tool.md');
+  ignoreList.push('disability-benefits/apply/form-526-disability-claim.md');
```
rianfowler: @erikphansen metal smith is the core driver of our build process. the static content and styling is built before webpack takes over for the js apps. Currently this array is the only way to exclude static content pages from the site.

PR #8075, opened 2018-07-30
using facility classification for va_cemetery type to distinguish between national and state entries

dhall:
Once we load state cemetery data we don't currently have a way to show any difference between state and national entries. This PR will change 'Cemetery' to either 'National Cemetery' or 'State Cemetery' by displaying the `classification` property when the facility type is va_cemetery.

Search results listing:
screen shot 2018-07-30 at 11 11 32 am

Detail listing:
screen shot 2018-07-30 at 11 11 44 am

### DIFF
--- a/src/applications/facility-locator/components/FacilityTypeDescription.jsx
+++ b/src/applications/facility-locator/components/FacilityTypeDescription.jsx
@@ -3,8 +3,7 @@ import { facilityTypes } from '../config';
 
 class FacilityTypeDescription extends Component {
   renderFacilityType() {
-    const { facility } = this.props;
-    const { facilityType, classification } = facility.attributes;
+    const { facilityType, classification } = this.props.facility.attributes;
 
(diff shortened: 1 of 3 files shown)


src/applications/facility-locator/components/FacilityTypeDescription.jsx
```diff
@@ -0,0 +1,21 @@
+import React, { Component } from 'react';
+import { facilityTypes } from '../config';
+
+class FacilityTypeDescription extends Component {
+  renderFacilityType() {
+    const { facility } = this.props;
+    const { facilityType, classification } = facility.attributes;
```
U-DON: Since you're not specifically using `facility`, you can destructure everything in one line:

```
const { facilityType, classification } = this.props.facility.attributes;
```

U-DON (approved):
LGTM. Nice refactor.

dhall:
Thanks for the feedback @U-DON, I made the suggested change

PR #8077, opened 2018-07-30
526 fix email validations mg

markgreenburg:
We were defining the email and phone schema for the 526 form in vets-website directly instead of using the definitions from vets-json-schema, which match swagger. This change fixes that.

### DIFF
--- a/src/applications/disability-benefits/526EZ/pages/primaryAddress.js
+++ b/src/applications/disability-benefits/526EZ/pages/primaryAddress.js
@@ -144,7 +144,12 @@ const addressUISchema = (addressType, title) => {
   };
 };
 
-const { mailingAddress, forwardingAddress } = fullSchema526EZ.properties.veteran.properties;
+const {
+  mailingAddress,
+  forwardingAddress,
+  emailAddress,
+  primaryPhone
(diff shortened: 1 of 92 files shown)


src/applications/disability-benefits/526EZ/pages/primaryAddress.js
```diff
@@ -168,7 +173,7 @@ export const uiSchema = {
       emailAddress: {
         'ui:title': 'Email address',
         'ui:errorMessages': {
-          pattern: 'Please put your email in this format x@x.xxx'
+          pattern: 'Email format must be at least x@x.xx'
```
cvalarida: This could use some revision.

cvalarida (commented):
Looks good minus a weird error message.

src/applications/disability-benefits/526EZ/pages/primaryAddress.js
markgreenburg: For reference, the actual validation is that the email must have a minimum of `x@x.xx`

markgreenburg: Sure, thanks @peggygannon !

cvalarida (approved):

PR #8081, opened 2018-07-31
526 landing page

cvalarida:
# Scope
This is basically everything we need, but there are still some things we can do to improve upon it including:
- Adding `evss-claims` gating analytics (@djmunoz1 requested this)
- Adding validation

I also have a few content questions for Peggy.

# Screenshots

## Full page
image

@peggygannon What should the title say?

## Not logged in
image

This is where the rest of the alert boxes will show up.

## No `evss-claims` service
image

## Submission success
image

@peggygannon @goldenmeanie What should this box say?

## Submission failure
image
@peggygannon What should this box say? (They should be able to try again later.)

### DIFF
--- a/content/pages/disability-benefits/apply/form-526-disability-claim.md
+++ b/content/pages/disability-benefits/apply/form-526-disability-claim.md
@@ -2,7 +2,8 @@
 title: Apply for increased disability benefits
 entryname: 526EZ-claims-increase
 layout: page-react.html
-description: Learn how to apply online for increased disability compensation. 
+description: Learn how to apply online for increased disability compensation.
+hideFromSidebar: true
 ---
 <div id="main">
   <nav aria-label="Breadcrumb" aria-live="polite" class="va-nav-breadcrumbs"
--- a/src/applications/disability-benefits/526EZ/components/create526EmailForm.jsx
+++ b/src/applications/disability-benefits/526EZ/components/create526EmailForm.jsx
@@ -56,7 +56,7 @@ class EmailForm extends React.Component {
           isVisible
           headline="To use our new tool, you’ll need to sign in with your premium My HealtheVet or DS Logon account"
           content="If you don’t have a verified account, you can create an ID.me account to complete the verification process."
-          status="error"/>
+          status="info"/>
       );
     }
 
@@ -64,7 +64,7 @@ class EmailForm extends React.Component {
       return (
         <AlertBox
           isVisible
-          headline="We’re sorry. It looks like we’re missing some information needed for you to apply online for increased disability compensation"
+          headline="We’re sorry. It looks like we’re missing some information needed for you to apply online for increased disability compensation."
           content="For help, please call Veterans Benefits Assistance at 1-800-827-1000, Monday through Friday, 8:00 a.m. to 9:00 p.m. (ET)."
           status="error"/>
       );
--- a/src/applications/disability-benefits/526EZ/components/create526EmailForm.jsx
+++ b/src/applications/disability-benefits/526EZ/components/create526EmailForm.jsx
@@ -65,7 +65,7 @@ class EmailForm extends React.Component {
         <AlertBox
           isVisible
           headline="We’re sorry. It looks like we’re missing some information needed for you to apply online for increased disability compensation"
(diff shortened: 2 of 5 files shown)


peggygannon (changes requested):
Hi, @cvalarida A couple of tweaks: 

--for the title, I'd like to suggest: 
Increased Disability Compensation Testing

--for the blue button, please made the button language:  
Submit Your Email

--for the EVSS message, can you remove the hyphen between Monday - Friday and add "through". 
It should say "Monday through Friday,"

--Submission message:  
Thank you. We received your email address. We'll contact you to schedule a time to try out the claim for increase tool. 

--Failure message: 
We're sorry. We didn't receive your email address. We're working to fix the problem. Please try again later.

AZSchneider:
@peggygannon Can we take "testing" out of the title and the nav? This is a real submission and I don't want people to think it would just be a test and not an actual submission

peggygannon:
@AZSchneider Yes, thanks for clarifying that (see my suggestion below). Also, this brings up another question, should this page nest under the Disability "How to Apply" page? 

@cvalarida Here's a new header. 
Apply for Increased Disability Compensation

content/pages/disability-benefits/apply/form-526-disability-claim.md
```diff
@@ -2,7 +2,8 @@
 title: Apply for increased disability benefits
 entryname: 526EZ-claims-increase
 layout: page-react.html
-description: Learn how to apply online for increased disability compensation. 
+description: Learn how to apply online for increased disability compensation.
+hideFromSidebar: true
```
markgreenburg: Nice!

src/applications/disability-benefits/526EZ/components/create526EmailForm.jsx
```diff
@@ -0,0 +1,130 @@
...
+    if (loading) {
+      return null;
+    }
+
+    if (!isLoggedIn) {
+      return (
+        <AlertBox
+          isVisible
+          headline="To use our new tool, you’ll need to sign in with your premium My HealtheVet or DS Logon account"
+          content="If you don’t have a verified account, you can create an ID.me account to complete the verification process."
```
markgreenburg: Should we actually link to idme here?

src/applications/disability-benefits/526EZ/components/create526EmailForm.jsx
```diff
@@ -0,0 +1,130 @@
...
+          status="error"/>
+      );
+    }
+
+    if (!hasEVSSService) {
```
markgreenburg: Between this and `isLoggedIn`, we may want to actually use `RequiredLoginView` and then just check for an `isDataAvailable` prop here. But...that's also more complicated, and I don't really see a problem with doing things this way, especially in the short-term.

markgreenburg (approved):
Arguable whether we're duplicating RequiredLoginView functionality here, but 🤷‍♂️ 
Also, didn't see any tests. We can probably throw those into a separate issue to get this out the door.

markgreenburg:
@cvalarida one other enhancement to consider is whether or not we should tell veterans that they've already submitted their email when they reload the page. Currently, since we're not checking anything ahead of time, we always show an empty email input field without confirming that we've previously gotten a successful request from them.

cvalarida:
Updated screenshots:

image
I updated the title and changed the not-logged-in alert to be an info alert instead of an error. I channeled my inner @goldenmeanie and thought that would trigger less of a stress response. (Or something like that, anyhow...

image
Updated the button content

image
Updated success message

image
Updated failure message

image
Updated EVSS message

src/applications/disability-benefits/526EZ/components/create526EmailForm.jsx
cvalarida: Ideally? Yeah. I don't know that it's a problem like this because it's a temporary page, though.

src/applications/disability-benefits/526EZ/components/create526EmailForm.jsx
cvalarida: I thought about it but decided against it for a few reasons:

1) This is super simple
2) `RequiredLoginView` doesn't gate based on the service list provided, so we'd still have to gate it manually
3) I don't know whether the `RequiredLoginView` component is already in the static page bundle, but if it isn't, I didn't want to add it

peggygannon (approved):

cvalarida:
image

PR #8082, opened 2018-07-31
Vet360 initialize

bshyong:
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/11692

### DIFF
--- a/src/applications/personalization/profile360/components/ProfileView.jsx
+++ b/src/applications/personalization/profile360/components/ProfileView.jsx
@@ -5,6 +5,12 @@ import DowntimeNotification, { externalServices, externalServiceStatus } from '.
 import DowntimeApproaching from '../../../../platform/monitoring/DowntimeNotification/components/DowntimeApproaching';
 import recordEvent from '../../../../platform/monitoring/record-event';
 
+import {
+  isPendingTransaction,
+  isSuccessfulTransaction,
+} from '../vet360/util/transactions';
+
+import Vet360TransactionPending from '../vet360/components/base/TransactionPending';
 import Vet360TransactionReporter from '../vet360/containers/TransactionReporter';
 
 import Hero from './Hero';
@@ -18,8 +24,10 @@ import MVIError from './MVIError';
 class ProfileView extends React.Component {
   static propTypes = {
     downtimeData: PropTypes.object,
+    vet360InitializationTransaction: PropTypes.object,
     isVet360AvailableForUser: PropTypes.bool,
     fetchTransactions: PropTypes.func.isRequired,
+    refreshTransaction: PropTypes.func.isRequired,
     fetchMilitaryInformation: PropTypes.func.isRequired,
     fetchHero: PropTypes.func.isRequired,
     fetchPersonalInformation: PropTypes.func.isRequired,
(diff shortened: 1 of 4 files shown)


src/applications/personalization/profile360/vet360/actions/transactions.js
```diff
@@ -29,6 +29,42 @@ function recordProfileTransaction(event, fieldName) {
...
+    } catch (error) {
+      dispatch({
+        type: VET360_TRANSACTION_REQUEST_FAILED,
+        error,
+        fieldName
+      });
+    }
+  };
+}
+
```
ncksllvn: I was debating how we should go about this. I think when we refactor we ultimately need a generic transaction function that we can leverage here, and during the update/delete action creators.

src/applications/personalization/profile360/components/ProfileView.jsx
```diff
@@ -35,7 +35,7 @@ class ProfileView extends React.Component {
     if (this.props.isVet360AvailableForUser) {
       this.props.fetchTransactions();
     } else {
-      // this.props.initializeUserToVet360()
+      this.props.initializeUserToVet360();
     }
```
ncksllvn: I think we'll also need to check `this.props.isVet360AvailableForUser` during render and if false return a `<TransactionPending/>` that we can use for messaging and to poll for updates. Without that, I think the application will proceed as if the user was initialized from the start and we might see some weird behavior.

ncksllvn (commented):
It's difficult to say without a local Vet360 instance, but I think we need some sort of loading indicator on the Profile during initialization to protect from unknown behavior.

src/applications/personalization/profile360/components/ProfileView.jsx
bshyong: We do have messaging in Contact information that renders `<MissingVet360IDError/>` if the `isVet360AvailableForUser` is false; is there some other way you are thinking of doing messaging or handling actions in the non-initialized state?

ncksllvn: Some messaging about the initialization happening might be nice, but my main concern is that since we're not polling for updates on the initialization transaction, the application won't "unlock" once the transaction completes and the user will think there's some indefinite error on the Profile

ncksllvn: Does that make sense?

bshyong: Ah you know what I am missing the polling functionality in this PR. I will move the issue back to in-progress and get the polling/auto-refresh in here

src/applications/personalization/profile360/components/ProfileView.jsx
```diff
@@ -32,10 +40,13 @@ class ProfileView extends React.Component {
...
   componentDidMount() {
+    const initializationTransaction = this.props.vet360InitializationTransaction;
     if (this.props.isVet360AvailableForUser) {
       this.props.fetchTransactions();
     } else {
-      // this.props.initializeUserToVet360()
+      if (!initializationTransaction || !isPendingTransaction(initializationTransaction) || !isSuccessfulTransaction(initializationTransaction)) {
+        this.props.initializeUserToVet360();
+      }
     }
```
ncksllvn: How come we need to check the `initializationTransaction` existence here? Asking because it's in  `componentDidMount` and it seems like it will always be `undefined` in the beginning.  

Not positive but I think we want to be working with the `vet360InitializationTransaction.transaction` property here instead of just `vet360InitializationTransaction` btw.

ncksllvn (commented):
Left a few comments, mostly just a few questions. I haven't run this branch locally so I could be wrong on things.

ncksllvn (approved):

PR #8083, opened 2018-07-31
Hook up Opt Out to Google Analytics

erikphansen:
- [x] Click on opt-out on /education/opt-out-information-sharing/ | When user clicks, fire this event: `dataLayer.push({event: 'edu-navigation', 'edu-action': 'opt-out-button'})`
- [ ] "Are You Sure You Want to Opt-Out?" modal appears | When modal appears, fire this event - `dataLayer.push({event: 'edu-opt-out-modal'})`
- [x] Click on "Yes I want to Opt Out" | When user clicks, fire this event: `dataLayer.push({event: 'edu-navigation', 'edu-action': 'confirm'})`
- [x] Click on Cancel in Modal | `dataLayer.push({event: 'edu-navigation', 'edu-action': 'cancel'})`
- [x] On confirmation page for successful submission | Push event: `dataLayer.push({event: 'edu-opt-out-submission-successful'})`

This one fires on page one of the form but needs to fire on page two:
- [ ] Click on Back on /education/opt-out-information-sharing/opt-out-form-0993/claimant-information | `dataLayer.push({event: 'edu-navigation', 'edu-action': 'back'})`

Need to figure out a good way to pass additional actions to the buttons on the final page of multi-page forms:
- [ ] Click on Submit on /education/opt-out-information-sharing/opt-out-form-0993/claimant-information | `dataLayer.push({event: 'edu-navigation', 'edu-action': 'submit'})`

@nedierecel confirmed that it would be nice to add GA integration with other forms so we should figure out a proper way to wire up the actions to the buttons on a form's final page.

### DIFF
--- a/src/applications/edu-benefits/0993/containers/ConfirmationPage.jsx
+++ b/src/applications/edu-benefits/0993/containers/ConfirmationPage.jsx
@@ -20,7 +20,9 @@ const scrollToTop = () => {
 
 export class ConfirmationPage extends React.Component {
   componentDidMount() {
-    recordEvent({ event: 'edu-opt-out-submission-successful' });
+    if (this.props.form.submission.response) {
+      recordEvent({ event: 'edu-opt-out-submission-successful' });
+    }
     focusElement('.schemaform-title > h1');
     scrollToTop();
   }
--- a/src/applications/edu-benefits/0993/containers/ConfirmationPage.jsx
+++ b/src/applications/edu-benefits/0993/containers/ConfirmationPage.jsx
@@ -5,6 +5,7 @@ import Scroll from 'react-scroll';
 
 import { focusElement } from '../../../../platform/utilities/ui';
 import AskVAQuestions from '../../../../platform/forms/components/AskVAQuestions';
+import recordEvent from '../../../../platform/monitoring/record-event';
 
 import GetFormHelp from '../../components/GetFormHelp';
 
@@ -19,6 +20,7 @@ const scrollToTop = () => {
 
 export class ConfirmationPage extends React.Component {
   componentDidMount() {
+    recordEvent({ event: 'edu-opt-out-submission-successful' });
     focusElement('.schemaform-title > h1');
     scrollToTop();
   }
--- a/src/applications/edu-benefits/components/OptOutWizard.jsx
+++ b/src/applications/edu-benefits/components/OptOutWizard.jsx
@@ -1,17 +1,25 @@
 import React from 'react';
 import Modal from '@department-of-veterans-affairs/formation/Modal';
 
+import recordEvent from '../../../platform/monitoring/record-event';
+
 export default class OptOutWizard extends React.Component {
   constructor(props) {
     super(props);
     this.state = { modalOpen: false };
   }
 
+  confirmOptOut = () => {
+    recordEvent({ event: 'edu-navigation', 'edu-action': 'confirm' });
+  }
+
   closeModal = () => {
+    recordEvent({ event: 'edu-navigation', 'edu-action': 'cancel' });
     this.setState({ modalOpen: false });
   }
 
   openModal = () => {
+    recordEvent({ event: 'edu-navigation', 'edu-action': 'opt-out-button' });
     this.setState({ modalOpen: true });
   }
 
@@ -41,6 +49,7 @@ export default class OptOutWizard extends React.Component {
           <p><strong>Please note:</strong> If you opt out and then change your mind, you’ll need to call the Education Call Center at <a className="help-phone-number-link" href="tel:+1-888-442-4551">1-888-442-4551</a> to opt back in.</p>
           <div>
             <a
+              onClick={this.confirmOptOut}
               href="/education/opt-out-information-sharing/opt-out-form-0993"
               className="usa-button-primary">
               Yes, I Want to Opt Out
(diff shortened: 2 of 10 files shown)


src/applications/edu-benefits/components/OptOutWizard.jsx
```diff
@@ -41,6 +49,7 @@ export default class OptOutWizard extends React.Component {
           <p><strong>Please note:</strong> If you opt out and then change your mind, you’ll need to call the Education Call Center at <a className="help-phone-number-link" href="tel:+1-888-442-4551">1-888-442-4551</a> to opt back in.</p>
           <div>
             <a
+              onClick={this.confirmOptOut}
```
rianfowler: Does this event fire before navigation?

erikphansen: Technically I believe it does. Is that a concern?

rianfowler: I'm concerned the async call to GA will not fire before browser navigation is initiated but it's been a while since I've worked on something like this.

erikphansen: so then the URLs that gets reported maybe won't match what it was when the button was actually clicked?

rianfowler: I just dunno which one of the things happens first: `onClick` or `href` navigation. The model in my head is that they both get queued up and `onClick` queues a fetch that is sequenced after the page navigation.

erikphansen: It certainly appears that the onClick fires first. And I think that makes sense. In a standard, non-SPA, if the location was updated first, via the `href`, that would trigger a page refresh, effectively killing all of the JS code on the first page. Thus the `onClick` would never fire because the JS no longer exists.

cehsu (approved):
LGTM, pending a question regarding the confirmation page.

src/applications/edu-benefits/0993/containers/ConfirmationPage.jsx
```diff
@@ -19,6 +20,7 @@ const scrollToTop = () => {
 
 export class ConfirmationPage extends React.Component {
   componentDidMount() {
+    recordEvent({ event: 'edu-opt-out-submission-successful' });
```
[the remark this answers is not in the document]
erikphansen: Good point, but I'm not sure I follow. Check for a response from what? From the call to `recordEvent()`?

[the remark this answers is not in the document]
cehsu: response as in this.props.form.response, similar to how it is checked for the conditional rendering of the confirmation number below.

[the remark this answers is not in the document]
erikphansen: oh, so something like:

```js
if (this.props.form.submission.response) {
  recordEvent({ event: 'edu-opt-out-submission-successful' });
}
```

src/applications/edu-benefits/components/OptOutWizard.jsx
erikphansen: okay we really need to talk with @nedierecel and/or Tony to get more info/background on why we seem to be tracking both clicks and the page views that those clicks trigger.

nedierecel:
@erikphansen For form submissions, we wanted to track the Submit button as a nav element (this is bucketed under an Interactions event category); when the form is successfully submitted, then it is tracked as a Transaction (which is tracked as a Goal in a larger category).

We're tracking similar button interactions on the dashboard, profile, and account elements as well, if that's helpful to compare.

src/applications/edu-benefits/components/OptOutWizard.jsx
cehsu: @nedierecel could you please confirm that the modal button needs to be tracked, as it is actually a link?

nedierecel:
@cehsu I think we can omit for now, unless @rtluu or Tony would still like this tracked. I'm able to use the onClick via Google to fire the modal.

rtluu:
@nedierecel that works for me!

nedierecel:
@cehsu I am able to prevent the historyChanges on /claimant-information from firing extra pageviews. This is in draft mode and not published to staging or production. 

In Google Tag Manager, here is the logic:

1. A GA Pageview will fire once the individual hits the /claimant-information page
2. When a replaceState occurs, GA will not fire an additional pageview (I see 3 historyChanges occur for:

- /education/opt-out-information-sharing/opt-out-form-0993/introduction
- /education/opt-out-information-sharing/opt-out-form-0993/
- /education/opt-out-information-sharing/opt-out-form-0993/claimant-information (this pageview is already triggered on page load, and it does not require the historyChange to trigger it)

3. When the pushState occurs, another pageview will trigger and record pageviews to /education/opt-out-information-sharing/opt-out-form-0993/review-and-submit

cehsu:
Thanks you @nedierecel, the approach for preventing the historyChanges on /claimant-information from firing extra pageviews sounds great.

rtluu:
@nedierecel are there any other changes that need to be made for this item? Just want to double check before closing out

nedierecel:
@rtluu I'll check in with @tonyadhoc tomorrow. Based on our call on Monday, I'd like to think through this particular implementation and if we want to change anything.

rtluu:
@nedierecel if we could start a new PR for the `historyChanges` work, that would be great!

PR #8089, opened 2018-08-01
Profile: Fix for stale data showing post-update

ncksllvn:
This PR updates the way updated transactions are processed - if the transaction is successful, we refresh the profile and then the transaction data is simply removed from state; otherwise, the updated transaction is made available in state. Previously, we were aways making the transaction data available, and since successful transactions won't trigger pending-update messaging, the old data was revealed while we were in the process of refreshing the profile. 

Resolves department-of-veterans-affairs/vets.gov-team/issues/11690

### DIFF
--- a/src/applications/personalization/profile360/vet360/actions/transactions.js
+++ b/src/applications/personalization/profile360/vet360/actions/transactions.js
@@ -82,22 +82,24 @@ export function refreshTransaction(transaction, analyticsSectionName) {
 
       const transactionRefreshed = isVet360Configured() ? await apiRequest(`/profile/status/${transactionId}`) : await localVet360.updateTransaction(transactionId);
 
-      dispatch({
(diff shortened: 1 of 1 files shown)


src/applications/personalization/profile360/vet360/actions/transactions.js
```diff
@@ -82,22 +82,24 @@ export function refreshTransaction(transaction, analyticsSectionName) {
 
       const transactionRefreshed = isVet360Configured() ? await apiRequest(`/profile/status/${transactionId}`) : await localVet360.updateTransaction(transactionId);
 
-      dispatch({
-        type: VET360_TRANSACTION_UPDATED,
-        transaction: transactionRefreshed
-      });
-
```
ncksllvn: The ultimate change here is that this only happens when the transaction is not successful.

bshyong (approved):

PR #8099, opened 2018-08-01
Copied and culled increase-only app

cvalarida:
I copied the increase only app and removed most (if not all) of the unnecessary stuff...and probably a little bit more, but we can copy from the increase only app when we need it.

It's not doing the SPA routing quite right; when refreshing on `/introduction`, it goes to the home page instead of serving the app and letting it handle the routing. That's worth investigating at some point, but I'm leaving it for now to unblock the other tickets.


markgreenburg (approved):
LGTM! Here we go...

PR #8101, opened 2018-08-02
Update average number of days to process a disability claim in July 

jstanek:
Monthly CURF update received through Jotform for average number of days to process disability claims.


peggygannon (approved):
@jstanek Hi, Joe.. updates look good. Thanks!

1Copenut (approved):
@jstanek Looks good to me. 👍

PR #8109, opened 2018-08-02
Intro page content update

markgreenburg:
Updates content per latest copy in attached ticket.
DOES NOT touch form start controls, so the buttons aren't dynamic and don't do anything. This will be addressed in a separate PR.

Screenshot 1:
screen shot 2018-08-03 at 2 00 14 pm

Screenshot 2:
screen shot 2018-08-03 at 2 00 26 pm

Screenshot 3:
screen shot 2018-08-02 at 5 58 51 pm

### DIFF
--- a/src/applications/disability-benefits/all-claims/sass/disability-benefits.scss
+++ b/src/applications/disability-benefits/all-claims/sass/disability-benefits.scss
@@ -12,3 +12,11 @@
 .full-page-alert {
   margin-bottom: 2em;
 }
+
+button {
+  margin-top: 1em;
+}
+
+.list-four {
+  padding-bottom: 0;
+}
--- a/src/applications/disability-benefits/all-claims/sass/disability-benefits.scss
+++ b/src/applications/disability-benefits/all-claims/sass/disability-benefits.scss
@@ -17,6 +17,17 @@ button.temporary-start {
   margin-top: 1em;
 }
 
+ul.original-disability-list {
+  li {
+    margin-bottom: 0;
+  }
+}
+
 .process-step:last-child {
   padding-bottom: 0;
+}
+
+.list-item-connector {
+  display: block;
+  margin: 0;
 }
--- a/src/applications/disability-benefits/all-claims/sass/disability-benefits.scss
+++ b/src/applications/disability-benefits/all-claims/sass/disability-benefits.scss
@@ -13,10 +13,10 @@
   margin-bottom: 2em;
 }
 
-button {
+button.temporary-start {
   margin-top: 1em;
 }
 
-.list-four {
+.process-step:last-child {
   padding-bottom: 0;
 }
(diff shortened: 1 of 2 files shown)


src/applications/disability-benefits/all-claims/sass/disability-benefits.scss
```diff
@@ -12,3 +12,11 @@
 .full-page-alert {
   margin-bottom: 2em;
 }
+
+button {
```
cvalarida: I think this is probably too broad a rule to put on the whole form. Is this a temporary fix?

src/applications/disability-benefits/all-claims/sass/disability-benefits.scss
```diff
@@ -12,3 +12,11 @@
...
+  margin-top: 1em;
+}
+
+.list-four {
```
cvalarida: Alternatively, `.process-step:last-child` should do the trick as well (and scale better, though that's not strictly speaking necessary here).

cvalarida (approved):
My only concern is applying a too-broad rule to all the buttons in the form. Other than that, it looks good to me if it looks good to @peggygannon.

src/applications/disability-benefits/all-claims/sass/disability-benefits.scss
markgreenburg: Yeah, this will need to change, but might be a good idea to make this more explicit.

src/applications/disability-benefits/all-claims/sass/disability-benefits.scss
markgreenburg: Oh hey, that's a thing!

peggygannon (changes requested):
@markgreenburg Hi, Mark, see below for some copy tweak to the intro page: 
close up the space above and below the "and" in the bullet list. So it should look like this: "and" is run in on that first bullet line and bolded. 
**You'll be submitting an original disability claim, if both of these are true:**
*You’re a Veteran or Servicemember who’s between 180 and 90 days from ending your military service, **and**
*This is the first time you’re filing a disability claim

this line should be a link:
 [You can order service medical records through the National Archives.](https://www.archives.gov/veterans/military-service-records)

Please update the copy in the Disability ratings blue box to this (adding in info about an exam): 
For each disability claim, we assign a severity rating from 0% to 100%. We base this rating on the supporting documents you turn in with your claim. In some cases we may also ask you to have an exam to help us rate your disability.

markgreenburg:
Thanks @peggygannon - here are a few updated screenshots for your review:

Updated ratings box:
screen shot 2018-08-03 at 5 21 58 pm

----------
Updated header and list item formatting:
screen shot 2018-08-03 at 5 21 51 pm

----------
Link to archives:
screen shot 2018-08-03 at 5 23 50 pm

peggygannon (approved):

PR #8111, opened 2018-08-03
Sip intro props mg

markgreenburg:
Changes SiP data retention period to a prop, instead of hard-coding 60 days.

### DIFF
--- a/src/platform/forms/save-in-progress/SaveInProgressIntro.jsx
+++ b/src/platform/forms/save-in-progress/SaveInProgressIntro.jsx
@@ -67,6 +67,7 @@ class SaveInProgressIntro extends React.Component {
     } else if (renderSignInMessage) {
       alert = renderSignInMessage(prefillEnabled);
     } else if (prefillEnabled && !verifyRequiredPrefill) {
+      const { retentionPeriod } = this.props;
       alert = (
         <div>
           <div className="usa-alert usa-alert-info schemaform-sip-alert">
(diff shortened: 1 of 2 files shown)


src/platform/forms/save-in-progress/SaveInProgressIntro.jsx
```diff
@@ -67,14 +67,15 @@ class SaveInProgressIntro extends React.Component {
...
+      const { retentionPeriod } = this.props;
       alert = (
         <div>
           <div className="usa-alert usa-alert-info schemaform-sip-alert">
             <div className="usa-alert-body">
               <strong>If you’re signed in to your account, your application process can go more smoothly. Here’s why:</strong><br/>
               <ul>
                 <li>We can prefill part of your application based on your account details.</li>
-                <li>You can save your form in progress, and come back later to finish filling it out. You have 60 days from the date you start or update your application to submit the form. After 60 days, the form won’t be saved, and you’ll need to start over.</li>
+                <li>You can save your form in progress, and come back later to finish filling it out. You have {retentionPeriod} from the date you start or update your application to submit the form. After {retentionPeriod}, the form won’t be saved, and you’ll need to start over.</li>
```
markgreenburg: I decided to only make the actual period a prop (instead of the whole string) because I feel that there's some value to standardizing the messaging as much as possible.

src/platform/forms/save-in-progress/SaveInProgressIntro.jsx
```diff
@@ -195,6 +196,7 @@ SaveInProgressIntro.propTypes = {
   saveInProgress: PropTypes.object.isRequired,
   fetchInProgressForm: PropTypes.func.isRequired,
   removeInProgressForm: PropTypes.func.isRequired,
+  retentionPeriod: PropTypes.string,
```
markgreenburg: Naming is hard. Open to suggestions.

src/platform/forms/save-in-progress/SaveInProgressIntro.jsx
```diff
@@ -205,6 +207,10 @@ SaveInProgressIntro.propTypes = {
   downtime: PropTypes.object
 };
 
+SaveInProgressIntro.defaultProps = {
+  retentionPeriod: '60 days'
```
cvalarida: Hey, nice use of default props! :raised_hands:
I keep forgetting this is a thing.

src/platform/forms/save-in-progress/SaveInProgressIntro.jsx
cvalarida: Sounds good to me. :man_shrugging:

cvalarida (approved):

ncksllvn (approved):
LGTM! It would be cool if we could go a step further and derive that formatted string value `retentionPeriod` by whatever numerical value controls application expiration, but I have no idea where that is or if the FE has access to it.

markgreenburg:
Yeah, I think that's set by vets-api, unfortunately.

PR #8112, opened 2018-08-03
Feature/11693 mergedbuild

accbjt:
Here is my solution for separating the vets.gov and merger pages. One problem is that we can not use the same paths for vets.gov and merger with having conflicts of code. Trying to manage through a feature flag with templating logic seems to be a big task and very complex. This solution separates the content files which gives us the ability to use the same named directories/pages. Vets.gov will live in the original content folder and the merger will live in the merger folder. On build it will use the merger folder if the `mergedbuild=true`. I also added a watch script to run locally `yarn watch:devpreview`. It seems to work okay. So if you want to run the vets.gov version you use `yarn watch` and for merger you use `yarn:devpreview`.


wyattwalter:
Adding @kreek

wyattwalter (commented):
The approach in build.js makes sense to me and seems quite simple. From a content perspective, this essentially forks the website which is likely a decent approach.

It seems like this is maybe branched off master so it has a bunch of changes that you wouldn't want merged into your `feature/11693` branch though. Or was this strictly as a check of approach?

accbjt:
@wyattwalter This is a branch off of my feature branch but I did a pull from master. I will be updating my feature branch and updating this branch so that these are inline with each other. I wanted to make sure it's updated with master before moving forward.

merger/includes/veteran-navigation.html
```diff
@@ -0,0 +1,187 @@
...
+          <div id="vetnav-about" class="vetnav-panel" role="none" hidden>
+            <ul role="menu" aria-label="About VA">
+              <li>
+                <button class="vetnav-level2 vetnav-trigger" role="button" aria-haspopup="true" aria-controls="vetnav-inside-va" onClick="window.dataLayer.push({ event: 'nav-header-second-level' });">Inside VA</button>
+
+                <!-- submenu panel -->
+                <ul class="vetnav-panel vetnav-panel--submenu" id="vetnav-about-va" role="menu" aria-label="Inside VA" hidden>
+                  <!-- submenu trigger -->
+                  <li>
+<<<<<<< HEAD
```
bshyong: Some merge conflicts here—can you do a global search on the branch for conflict tags like this to make sure these don't get inadvertently merged?

bshyong (changes requested):

merger/includes/veteran-navigation.html
accbjt: Thanks, I didn't see this.

jkassemi:
Is there anything that can be done to break down this change?

accbjt:
@jkassemi The main changes is that the contents for merger is located in a different folder called merger. It is basically a duplicate of the content folder but with all the merger changes. In the scripts/build.js there is some logic for using which content folder for which buildtype.

PR #8113, opened 2018-08-03
526 confirmation poll

cvalarida:
## Description
This PR creates a `ConfirmationPoll` and uses it as the `confirmationPage` in the form config. `ConfirmationPoll` renders `ConfirmationPage` so we can show a loading spinner over the top of the confirmation page content without mucking about with redux.

### Submitted (less than 30 seconds)
image

### Submitted (30 seconds or more)
image

### Retrying / exhausted / api call failure
image

### Success
Only the blue box content changes.
image

### Non-retryable error
image

### DIFF
--- a/src/applications/disability-benefits/526EZ/components/ConfirmationPoll.jsx
+++ b/src/applications/disability-benefits/526EZ/components/ConfirmationPoll.jsx
@@ -54,6 +54,7 @@ const errorMessage = (jobId) => (
     <p>We’re sorry. Something went wrong on our end when we tried to submit your application. For help, please call the Vets.gov Help Desk at <a href="tel:+18555747286">1-855-574-7286</a>, Monday – Friday, 8:00 a.m. – 9:00 a.m. (ET).</p>
     <strong>Confirmation number</strong>
     <div>{jobId}</div>
+    <br/>
   </div>
 );
 
@@ -98,8 +99,7 @@ export class ConfirmationPoll extends React.Component {
       return;
     }
 
-    // apiRequest(`/disability_compensation_form/submission_status/${this.props.jobId}`)
-    apiRequest('/disability_compensation_form/submission_status/1f67ef1799012b1972d3772c')
+    apiRequest(`/disability_compensation_form/submission_status/${this.props.jobId}`)
       .then((response) => {
         // Don't process the request once it comes back if the component is no longer mounted
         if (!this.__isMounted) {
--- a/src/applications/disability-benefits/526EZ/components/ConfirmationPoll.jsx
+++ b/src/applications/disability-benefits/526EZ/components/ConfirmationPoll.jsx
@@ -1,10 +1,9 @@
 import React from 'react';
 import Raven from 'raven-js';
-import sinon from 'sinon';
 
-// import { apiRequest } from '../../../../platform/utilities/api';
+import { apiRequest } from '../../../../platform/utilities/api';
 
-const statuses = {
+export const submissionStatuses = {
   // The status returned by the API
   pending: 'pending',
   retry: 'retry',
@@ -14,38 +13,41 @@ const statuses = {
   apiFailure: 'apiFailure'
 };
 
-const pendingResponse = Promise.resolve({ status: statuses.pending });
-// const successResponse = Promise.resolve({
-//   status: statuses.succeeded,
-//   confirmationNumber: '123abc'
-// });
-const failureResponse = Promise.resolve({ status: statuses.failed });
-
-const apiRequest = sinon.stub();
-apiRequest.onCall(0).returns(pendingResponse);
-apiRequest.onCall(1).returns(failureResponse);
-
 export default class ConfirmationPoll extends React.Component {
+  static defaultProps = {
+    pollRate: 5000
+  }
+
   constructor(props) {
     super(props);
 
     this.state = {
-      submissionStatus: statuses.pending,
+      submissionStatus: submissionStatuses.pending,
       confirmationNumber: null,
       failureCode: null,
     };
   }
 
   componentDidMount() {
+    this.__isMounted = true;
     this.poll();
   }
 
+  componentWillUnmount() {
+    this.__isMounted = false;
+  }
+
   poll = () => {
+    // Don't continue to request after the component is unmounted
+    if (!this.__isMounted) {
+      return;
+    }
+
     apiRequest('')
       .then((response) => {
         // Check status
         // TODO: Get where this actually comes from in the response
-        if (response.status !== statuses.pending) {
+        if (response.status !== submissionStatuses.pending) {
           this.setState({
             submissionStatus: response.status,
             // TODO: Get where this actually comes from in the response
@@ -53,13 +55,13 @@ export default class ConfirmationPoll extends React.Component {
           });
         } else {
           // Wait for 5 seconds and recurse
-          setTimeout(this.poll, 5000);
+          setTimeout(this.poll, this.props.pollRate);
         }
       })
       .catch((response) => {
         // The call to the API failed; show a message or something
         this.setState({
-          submissionStatus: statuses.apiFailure,
+          submissionStatus: submissionStatuses.apiFailure,
           // NOTE: I don't know that it'll always take this shape.
           failureCode: response.errors[0].status
         });
@@ -68,11 +70,11 @@ export default class ConfirmationPoll extends React.Component {
 
   render() {
     switch (this.state.submissionStatus) {
-      case statuses.retry: {
+      case submissionStatuses.retry: {
         // What should we do here?
         return <p><strong>This is taking a while.</strong> Please check on the Claims Status tool later.</p>;
       }
-      case statuses.succeeded: {
+      case submissionStatuses.succeeded: {
         return (
           <div>
             <strong>Confirmation number</strong>
@@ -81,11 +83,11 @@ export default class ConfirmationPoll extends React.Component {
           </div>
         );
       }
-      case statuses.failed: {
+      case submissionStatuses.failed: {
         // What should we do here?
         return null;
       }
-      case statuses.apiFailure: {
+      case submissionStatuses.apiFailure: {
         // What should we do here?
         Raven.captureMessage('526_submission_failure', {
           context: {
--- a/src/applications/disability-benefits/526EZ/components/ConfirmationPoll.jsx
+++ b/src/applications/disability-benefits/526EZ/components/ConfirmationPoll.jsx
@@ -1,6 +1,8 @@
 import React from 'react';
 import Raven from 'raven-js';
 
+// import AlertBox from '@department-of-veterans-affairs/formation/AlertBox';
+
 import get from '../../../../platform/utilities/data/get';
 
 import { apiRequest } from '../../../../platform/utilities/api';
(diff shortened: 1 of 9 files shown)


cvalarida:
When I went to connect this PR to the ticket, I found [this open polling ticket](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/9798). That lead me to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/9702 which says VIC v2 already does this. I haven't looked at it yet, but we might want to abstract the logic out and use a helper. Two birds, one stone, etc.

cvalarida:
There are still a few things left to do here:
1) Hook it up to the actual endpoint when that's available
2) Fill in the content for what to show the user for the various states
    - Need to be very careful about this one since it's pretty high-stakes
3) Finish writing the tests once 2) is done

cvalarida:
@goldenmeanie From our conversation, we identified a few questions to ask @kreek so we can figure out what to show the user. I got the answers:

> 1) Where does the claim “go” after a successful or submission to EVSS?

The submitted json is removed if it succeeds. He still needs to check on the in-progress form, but it should go away too (or be soft deleted/hidden).

> 2) Do we have any indication of how long it’ll take between a successful submission and showing up in the claims status tool?

That’s instant.

> 3) How long do we keep the confirmation number?

Forever. He still needs to follow up if that’s right since it should maybe go away at some point.

> 4) If we ultimately fail to submit a claim, what happens to the form data?

If it fails because retries are exhausted we’ll still have the in-progress form. This means the user can go back to the form and continue where they left off (at the review page) and try again.

goldenmeanie:
@peggygannon this is the one we were talking about in copy sync

peggygannon:
@cvalarida @goldenmeanie Hello, can I get a little more guidance and/or screenshots about the type of messages that need to be written for these scenarios?

src/applications/disability-benefits/526EZ/containers/ConfirmationPage.jsx
```diff
@@ -61,53 +60,49 @@ class ConfirmationPage extends React.Component {
...
+
         <div className="inset">
-          <h4>Disability Compensation Claim for Increase <span className="additional">(Form 21-526EZ)</span></h4>
+          <h4>Disability Compensation Claim <span className="additional">(Form 21-526EZ)</span></h4>
           <span>
             For {first} {middle} {last} {suffix}
           </span>
           <ul className="claim-list">
-            <strong>Conditions claimed</strong>
+            <strong>Conditions claimed for increase</strong>
```
markgreenburg: I think this probably shouldn't show unless there's been a successful submission. If there's a submission error, this heading still shows up, but the disabilities list under it will be empty and there'll be an error message pushed into `{this.props.submissionMessage}`, which would look kind of strange.

cvalarida: Do you mean it shouldn't show unless there's a successful submission to EVSS or the api? If we get to the confirmation page, we've successfully submitted to the api, we should have the list of conditions.

Have you run into this problem before?

src/applications/disability-benefits/526EZ/components/ConfirmationPoll.jsx
```diff
@@ -0,0 +1,192 @@
...
+const errorMessage = (jobId) => (
+  <div>
+    <p>We’re sorry. Something went wrong on our end when we tried to submit your application. For help, please call the Vets.gov Help Desk at <a href="tel:+18555747286">1-855-574-7286</a>, Monday – Friday, 8:00 a.m. – 9:00 a.m. (ET).</p>
+    <strong>Confirmation number</strong>
+    <div>{jobId}</div>
+    <br/>
+  </div>
+);
+
+// TODO: Make this a loading spinner
```
jbalboni: Should this comment be removed?

src/applications/disability-benefits/526EZ/components/ConfirmationPoll.jsx
```diff
@@ -0,0 +1,192 @@
...
+  const message = !longWait
+    ? 'Please wait while we submit your application and give you a confirmation number.'
+    : 'We’re sorry. It’s taking us longer than expected to submit your application. Thank you for your patience.';
+  return <LoadingIndicator message={message}/>;
+};
+
+export class ConfirmationPoll extends React.Component {
+  // Using it as a prop for easy testing
+  static defaultProps = {
+    pollRate: 5000
```
jbalboni: This seems pretty high to me. Do we have any idea how long their api will take?

src/applications/disability-benefits/526EZ/components/ConfirmationPoll.jsx
```diff
@@ -0,0 +1,192 @@
...
+    this.state = {
+      submissionStatus: submissionStatuses.pending,
+      claimId: null,
+      failureCode: null,
+      longWait: false
+    };
+  }
+
+  componentDidMount() {
+    this.__isMounted = true;
```
jbalboni: Do we follow this `__` pattern elsewhere?

src/applications/disability-benefits/526EZ/components/ConfirmationPoll.jsx
```diff
@@ -0,0 +1,192 @@
...
+  <div>
+    <p>We’re sorry. Something went wrong on our end when we tried to submit your application. For help, please call the Vets.gov Help Desk at <a href="tel:+18555747286">1-855-574-7286</a>, Monday – Friday, 8:00 a.m. – 9:00 a.m. (ET).</p>
+    <strong>Confirmation number</strong>
+    <div>{jobId}</div>
+    <br/>
+  </div>
+);
+
+// TODO: Make this a loading spinner
+const pendingMessage = (longWait) => {
```
markgreenburg: I mean, I guess these messages could be in `helpers` or something just to clean this component up a bit, but I absolutely don't have a strong opinion since it's not like they're used anywhere else.

src/applications/disability-benefits/526EZ/components/ConfirmationPoll.jsx
```diff
@@ -0,0 +1,192 @@
...
+  pending: 'submitted', // Submitted to EVSS, waiting response
+  retry: 'retrying',
+  succeeded: 'received', // Submitted to EVSS, received response
+  exhausted: 'exhausted', // EVSS is down or something; ran out of retries
+  failed: 'non_retryable_error', // EVSS responded with some error
+  // When the api serves a failure
+  apiFailure: 'apiFailure'
+};
+
+const terminalStatuses = [
```
jbalboni: This could be a Set

src/applications/disability-benefits/526EZ/components/ConfirmationPoll.jsx
```diff
@@ -0,0 +1,192 @@
...
+            claimId: get('data.attributes.metadata.claimId', response) || null
+          });
+        } else {
+          // Wait for a bit and recurse
+          const waitTime = status === submissionStatuses.pending
+            ? this.props.pollRate
+            : this.props.pollRate * 2; // Seems like we don't need to poll as frequently when we get here
+
+          // Force a re-render to update the pending message if necessary
+          if (Date.now() - this.__startTime >= 30000) {
```
jbalboni: If I were a user, I think I'd consider it a long wait after 5-7 seconds. 30 seconds is a lot of time for no extra feedback.

src/applications/disability-benefits/526EZ/components/ConfirmationPoll.jsx
cvalarida: Yeah, I was thinking about moving them to a content file. That can happen outside of this PR, though (for sake of time).

src/applications/disability-benefits/526EZ/components/ConfirmationPoll.jsx
```diff
@@ -0,0 +1,192 @@
...
+            this.setState({ longWait: true });
+          }
+          setTimeout(this.poll, waitTime);
+        }
+      })
+      .catch((response) => {
+        // The call to the API failed
+        Raven.captureMessage('526_submission_status_poll_failure', {
```
jbalboni: I'm not sure if you need to log this to Sentry. If you get here, it's because the api responded with an error status, which will be logged in vets-api Sentry.

src/applications/disability-benefits/526EZ/components/ConfirmationPoll.jsx
```diff
@@ -0,0 +1,192 @@
...
+};
+
+const terminalStatuses = [
+  submissionStatuses.succeeded,
+  submissionStatuses.exhausted,
+  submissionStatuses.retry,
+  submissionStatuses.failed
+];
+
+const successMessage = (claimId) => (
```
jbalboni: It feels a little weird to me to have the messages in here. This component is mostly focused on polling, but with these UI snippets thrown in. Might be cleaner to have all the UI code in ConfirmationPage and leave this component to do polling and pass whatever status you need to ConfirmationPage.

src/applications/disability-benefits/526EZ/components/ConfirmationPoll.jsx
cvalarida: I asked Silvio if it was, and he said no. I think he made a PR to change that, now that you mention it.

You're saying all non-okay vets-api responses are logged to Sentry by the api?

src/applications/disability-benefits/526EZ/components/ConfirmationPoll.jsx
cvalarida: I thought I remembered seeing this pattern used elsewhere, but doing a quick grep, the only one I'm seeing is `this.__errors` in `src/applications/veteran-representative/validation.js`. I'll nix the `__`.

src/applications/disability-benefits/526EZ/components/ConfirmationPoll.jsx
cvalarida: Potentially a very long time. The main reason we're going down the asynchronous submission route is because we were consistently timing out at the load balancer (I think) after 55 seconds. It does seem really slow, but talking with @goldenmeanie, he feels like veterans are more likely to stick around for this one because it's pretty high-profile.

src/applications/disability-benefits/526EZ/components/ConfirmationPoll.jsx
cvalarida: Good point.

@goldenmeanie what think ye?

src/applications/disability-benefits/526EZ/components/ConfirmationPoll.jsx
cvalarida: Good point again. This is left over from when I had the component hierarchy swapped around. Definitely makes more sense to have all the UI in the one component now. :+1:

src/applications/disability-benefits/526EZ/components/ConfirmationPoll.jsx
jbalboni: Based on your other comment, this seems more reasonable now. But if we could see up to 60 seconds, then maybe some initial messaging about how it's going to really slow is useful.

src/applications/disability-benefits/526EZ/components/ConfirmationPoll.jsx
cvalarida: :man_shrugging: Works for me.

src/applications/disability-benefits/526EZ/containers/ConfirmationPage.jsx
```diff
@@ -61,53 +84,49 @@ class ConfirmationPage extends React.Component {
...
             <br/>
-            <ul className="disability-list">
-              {disabilities.filter(item => item['view:selected']).map((disability, i) => {
+            {/* <ul className="disability-list">
```
jbalboni: What's the plan for this code?

jbalboni (approved):

src/applications/disability-benefits/526EZ/containers/ConfirmationPage.jsx
cvalarida: Y'know, that's a great question. Kinda looks like I commented this out, doesn't it? I was just asking Mark yesterday why it was commented and he said it wasn't in master. :stuck_out_tongue:

I think I commented it out so I could actually get to the disabilities screen for some reason during testing and I just never cleaned it up. Guess that means I can go ahead and uncomment it! :grimacing:

PR #8119, opened 2018-08-06
rearrange markup

rianfowler:
Addresses [this](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/11846#issuecomment-410369429).

Not sure how I feel about the position of the results count but I think we should review it tomorrow with UX when everything is done and in staging. 

screen shot 2018-08-06 at 13 06 53


cehsu (approved):
LGTM, pending @1Copenut's approval.

PR #8124, opened 2018-08-06
Change reference to EVSSClaimsGate and delete from increase-only app

cvalarida:
Just like it says on the tin.


markgreenburg (approved):
LGTM 🚀

PR #8126, opened 2018-08-06
Fix tests

bshyong:


src/applications/personalization/profile360/vet360/components/AddressField/index.jsx
```diff
@@ -16,6 +16,80 @@ import Vet360ProfileField from '../../containers/ProfileField';
...
+
+export const validateCleanData = ({ addressLine1, city, stateCode,  internationalPostalCode, zipCode, countryName }, property) => {
+  const isInternational = inferAddressType(countryName, stateCode) === ADDRESS_TYPES.INTERNATIONAL;
+  const validateAll = !property;
+
+  return {
+    addressLine1: (property === 'addressLine1' || validateAll) && !addressLine1 ? 'Street address is required' : '',
+    city: (property === 'city' || validateAll) && !city ? 'City is required' : '',
+    stateCode: (property === 'stateCode' || validateAll) && !isInternational && !stateCode ? 'State is required' : '',
+    zipCode: (property === 'zipCode' || validateAll) && !isInternational && !zipCode ? 'Zip code is required' : '',
```
cehsu: This should be 'ZIP code' instead of 'Zip code'.

cehsu: Should the ZIP format (5-9 digits) be validated?

src/applications/personalization/profile360/vet360/components/AddressField/index.jsx
```diff
@@ -16,6 +16,80 @@ import Vet360ProfileField from '../../containers/ProfileField';
...
+};
+
+export const validateCleanData = ({ addressLine1, city, stateCode,  internationalPostalCode, zipCode, countryName }, property) => {
+  const isInternational = inferAddressType(countryName, stateCode) === ADDRESS_TYPES.INTERNATIONAL;
+  const validateAll = !property;
+
+  return {
+    addressLine1: (property === 'addressLine1' || validateAll) && !addressLine1 ? 'Street address is required' : '',
+    city: (property === 'city' || validateAll) && !city ? 'City is required' : '',
+    stateCode: (property === 'stateCode' || validateAll) && !isInternational && !stateCode ? 'State is required' : '',
```
cehsu: Is state displayed as a select? If not, should the state code value be validated, i.e. checked from a list of valid options?

cehsu (approved):
LGTM. I had some minor questions/comments for your consideration.

src/applications/personalization/profile360/vet360/components/AddressField/index.jsx
bshyong: thanks! yes these are loaded from a constants file for a list of valid options in a dropdown select

cehsu: Thanks for explaining, it seems like the forms library could be helpful here eventually.

src/applications/personalization/profile360/vet360/components/AddressField/index.jsx
bshyong: Yep we will add in more validations in the next iteration—there is also an address validator that sits in front of submissions and will reject addresses that don't exist

PR #8127, opened 2018-08-07
Remove the 526 from the ignore list

cvalarida:
The form is live, but doesn't show up in the side nav (because of a change in a separate PR).

image


SMLuthi (approved):
LGTM

PR #8128, opened 2018-08-07
add street address

rianfowler:
---
screen shot 2018-08-06 at 19 04 38
screen shot 2018-08-06 at 19 05 15

---
screen shot 2018-08-06 at 18 55 18
---
screen shot 2018-08-06 at 18 54 48
---
screen shot 2018-08-06 at 18 54 37

### DIFF
--- a/src/applications/edu-benefits/complaint-tool/reducers/schoolSelect.js
+++ b/src/applications/edu-benefits/complaint-tool/reducers/schoolSelect.js
@@ -80,8 +80,16 @@ export default function schoolSearch(state = initialState, action) {
 
       const institutionQuery = action.institutionQuery;
       const institutions = data.map(({ attributes }) => {
-        const { city, country, facilityCode, name, state: institutionState, zip } = attributes;
-        return { city, country, facilityCode, name, state: institutionState, zip };
+        const { city, country, facilityCode, name, state: institutionState, street, zip } = attributes;
+        return {
+          city: city || '',
+          country: country || '',
+          facilityCode,
+          name: name || '',
+          state: institutionState || '',
+          street: street || '',
+          zip: zip || ''
+        };
       });
       const pagesCount = Math.ceil(searchResultsCount / 10);
       const showInstitutions = institutions.length > 0;
--- a/src/applications/edu-benefits/complaint-tool/reducers/schoolSelect.js
+++ b/src/applications/edu-benefits/complaint-tool/reducers/schoolSelect.js
@@ -6,6 +6,7 @@ import {
   SEARCH_INPUT_CHANGED,
   INSTITUTION_SELECTED
 } from '../actions/schoolSearch';
+import _ from 'lodash';
 
 const initialState = {
   currentPageNumber: 1,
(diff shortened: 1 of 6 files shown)


src/applications/edu-benefits/complaint-tool/reducers/schoolSelect.js
```diff
@@ -80,8 +80,16 @@ export default function schoolSearch(state = initialState, action) {
 
       const institutionQuery = action.institutionQuery;
       const institutions = data.map(({ attributes }) => {
-        const { city, country, facilityCode, name, state: institutionState, zip } = attributes;
-        return { city, country, facilityCode, name, state: institutionState, zip };
+        const { city, country, facilityCode, name, state: institutionState, street, zip } = attributes;
+        return {
```
rianfowler: I want to be able to inline some conditional logic in the `jsx` for the appearance of these strings. String templates will render `false` and `null` but not an empty string 🙄 and the service returns `null` when an attribute is not available.

src/applications/edu-benefits/components/SchoolSelectField.jsx
```diff
@@ -189,8 +189,10 @@ export class SchoolSelectField extends React.Component {
                     <label
                       id={`institution-${index}-label`}
                       htmlFor={`page-${currentPageNumber}-${index}`}>
-                      <span className="institution-name">{name}</span>
-                      <span className="institution-city-state">{`${city}, ${state}`}</span>
+                      {name && <span className="institution-name">{name}</span>}
+                      {street && <span className="institution-street">{street}</span>}
+                      {(city || state) && <span className="institution-city-state">{`${city && city}${city && state && ', '}${state && state}`}</span>}
```
rianfowler: Will do: 
- Salt Lake City, UT
- Salt Lake City
- UT
- USA

but will never do: 
- Salt Lake City,
- , UT
- Salt Lake City, UT USA

src/applications/edu-benefits/sass/edu-benefits.scss
```diff
@@ -75,16 +75,19 @@
...
-        padding: 10px 0 0 10px;
+        padding: 10px 0 0px 10px;
         text-decoration: underline;
       }
       &.institution-city-state {
-        font-size: 30;
-        font-weight: bold;
         padding: 0 0 10px 10px;
       }
+      &.institution-country {
```
rianfowler: Something gross I did here: I'm styling the background of the `span`s here rather than the background of the label. I styled each span separately to get the padding right. 
screen shot 2018-08-06 at 19 08 45

The label is kind of messy and big. It's not semantically correct to use block elements or nested elements in a `label` either. 
screen shot 2018-08-06 at 19 11 10

screen shot 2018-08-06 at 19 10 34

Open to suggestions to make this neater.

src/applications/edu-benefits/complaint-tool/reducers/schoolSelect.js
cehsu: How about using lodash reduce to build this?

src/applications/edu-benefits/components/SchoolSelectField.jsx
cehsu: Should we consider moving this logic into a helper function with its own unit tests, it seems like a very helpful utility to include in `platform` eventually.

src/applications/edu-benefits/sass/edu-benefits.scss
cehsu: Could we put all of these spans in a container that handles this padding?

cehsu (approved):
LGTM, some minor question/comments for your consideration.

src/applications/edu-benefits/complaint-tool/reducers/schoolSelect.js
rianfowler: Maybe- it's using the native array map to transform an array of objects. What would lodash's reduce help with?

src/applications/edu-benefits/components/SchoolSelectField.jsx
rianfowler: That's probably a good idea. I'll make a ticket and let's discuss how to make it discoverable.

src/applications/edu-benefits/complaint-tool/reducers/schoolSelect.js
cehsu: line 84 could be refactored as something like:
```
return _.reduce(attributes, (result, value, key) => {
  result[key] = value || '';
  return result;
}, {});
```

PR #8139, opened 2018-08-08
Emis error handling

bshyong:
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/11988


src/applications/personalization/profile360/components/MilitaryInformation.jsx
```diff
@@ -20,21 +23,57 @@ class MilitaryInformationContent extends React.Component {
...
+              <p>You can also request to be added to DEERS through our online customer help center.</p>
+              <a href="https://iris.custhelp.va.gov/app/answers/detail/a_id/3036/~/not-registered-in-deers%2C-or-received-and-error-message-while-trying-to">Get instructions from our help center</a>.
+            </div>}/>
+        );
+      } else if (some(error.errors, (e) => {
+        return ['500', '503'].includes(e.code);
+      })) {
+        return <LoadFail information="military"/>;
+      }
+    }
```
ncksllvn: I think the `else if` condition here might be clearer moved into a helper function or declared above the JSX. It's just a little confusing.

src/applications/personalization/profile360/components/MilitaryInformation.jsx
```diff
@@ -20,21 +23,57 @@ class MilitaryInformationContent extends React.Component {
...
-            <div key={index}>
-              <h3>{service.branchOfService}</h3>
-              <div>{moment(service.beginDate).format('MMM D, YYYY')} &ndash; {moment(service.endDate).format('MMM D, YYYY')}</div>
-            </div>
-          );
-        })}
-      </div>
-    );
+    if (this.props.veteranStatus.servedInMilitary) {
+      if (serviceHistory.length === 0) {
```
ncksllvn: Is it necessary to check `servedInMilitary` here? I see we're checking it in the higher-level component already, so it seems like it's guaranteed to be true by the time we reach this point.

ncksllvn (approved):
LGTM!

PR #8144, opened 2018-08-09
adding dental to list of health services

dhall:
fixes #12043 by adding Dental Services to the Health Services drop down list. We've had the data but the entry was missed in my initial creation of the list. This will allow users to search for Dental Services explicitly.


U-DON (approved):
LGTM.

PR #8148, opened 2018-08-09
12067 move manual entry checkbox

rianfowler:
This PR 
- moves the manual entry checkbox out of the form config and into the `SchoolSelectField`
- retains visibility of search field when doing manual address entry 
- changes the state of the manual entry checkbox to `false` if a search is performed while the manual entry is visible 
screen shot 2018-08-09 at 17 26 11
screen shot 2018-08-09 at 17 26 25
search1


src/applications/edu-benefits/components/SchoolSelectField.jsx
```diff
@@ -138,41 +174,63 @@ export class SchoolSelectField extends React.Component {
...
           </div>
+          <ErrorableCheckbox
+            checked={manualSchoolEntryChecked}
+            onValueChange={() => this.handleManualSchoolEntryToggled(manualSchoolEntryChecked)}
+            label={<span>I’d rather type in my school information</span>}/>
           <div
             aria-live="polite"
             aria-relevant="additions text">
-            {searchResultsCount > 0 && <span>
+            {showSearchResults && searchResultsCount > 0 && <span>
```
rianfowler: Went with using a single prop in several places to handle the hiding of the search results when the manual entry is visible. I wanted to retain the state behind the scenes to keep it friendly- you can check and uncheck the box without losing any of your search state.

src/applications/edu-benefits/components/SchoolSelectField.jsx
```diff
@@ -200,92 +258,106 @@ export class SchoolSelectField extends React.Component {
...
+  const errorMessages = selectFacilityCodeErrorMessages(ownProps);
+  const facilityCodeSelected = ownProps.formData ? ownProps.formData.facilityCode : '';
   const institutionQuery = selectInstitutionQuery(state);
   const institutions = selectInstitutions(state);
   const institutionSelected = selectInstitutionSelected(state);
+  const manualSchoolEntryChecked = selectManualSchoolEntryChecked(state);
   const pagesCount = selectPagesCount(state);
   const searchInputValue = selectSearchInputValue(state);
   const searchResultsCount = selectSearchResultsCount(state);
+  const showErrors = errorMessages.length > 0 && selectFormSubmitted(ownProps) && !manualSchoolEntryChecked;
```
rianfowler: Putting this kind of logic in `mapStateToProps` is kind of a bad pattern- haven't really figured out and idiomatic way to handle making it easy to extract `ownProps` + `state` stuff into easily testable functions. They're not "selectors."

src/applications/edu-benefits/components/SchoolSelectField.jsx
```diff
@@ -138,41 +174,63 @@ export class SchoolSelectField extends React.Component {
...
-              </button>
+          <div className={schoolSearchClassNames}>
+            <label
+              id="school-search-label"
+              className={labelClassNames}
+              htmlFor="school-search-input">
+              {'School Information'}
+              {!manualSchoolEntryChecked && <span className="schemaform-required-span">{'(*Required)'}</span>}
+            </label>
+            {showErrors && <span
```
rianfowler: I tried to hack the `formConfig` to handle the error messaging but it didn't seem possible now that this field is an object. The `ui:field` is in the top level `facilityCode` but the `ui:required`  _validation_ is in `facilityCode.facilityCode`. I was able to get around this by reimplementing some of the error display logic but I'm open to adjusting this to a more idiomatic approach.

I tried adding `manualSchoolEntryChecked` to the `schema` but it tried to render a field. I couldn't figure out how to override that in the `ui:schema` so I just omitted it.

src/applications/edu-benefits/tests/complaint-tool/config/schoolInfo.unit.spec.js
```diff
@@ -11,14 +11,17 @@ import formConfig from '../../../complaint-tool/config/form';
 
 describe('complaint tool school info', () => {
   const { schema, uiSchema } = formConfig.chapters.schoolInformation.pages.schoolInformation;
+  delete uiSchema.school.facilityCode;
```
rianfowler: Now that I'm not hiding the search field, I need to be able to render the page without setting up a redux store to test `formConfig`. Simplest approach was to just remove the part of config that's not being tested. 🤷‍♂️

src/applications/edu-benefits/components/SchoolSelectField.jsx
rianfowler: screen shot 2018-08-09 at 17 46 31

src/applications/edu-benefits/components/SchoolSelectField.jsx
```diff
@@ -101,22 +127,36 @@ export class SchoolSelectField extends React.Component {
...
+    const fieldsetClassNames = classNames('search-select-school-fieldset');
+    const schoolSearchClassNames = classNames('school-search', {
+      'usa-input-error': showErrors
+    });
+    const labelClassNames = classNames('school-search-label', {
+      'usa-input-error-label': showErrors
+    });
+
+    if (formContext.reviewMode && manualSchoolEntryChecked) {
+      return (<div></div>);
```
rianfowler: The form generator is rendering the `SchoolSelectField` on the review page so it has to hide itself if manual entry is selected.

cehsu (approved):
LGTM, had a few comments/questions for your consideration.

PR #8150, opened 2018-08-09
Add warning banner to gibs intro page

markgreenburg:
This adds a warning banner to the GIBS intro page when GIBS status is up (when GIBS status is down, all this content gets replaced by an error message already).

screen shot 2018-08-09 at 4 52 14 pm

### DIFF
--- a/src/applications/post-911-gib-status/tests/containers/IntroPage.unit.spec.jsx
+++ b/src/applications/post-911-gib-status/tests/containers/IntroPage.unit.spec.jsx
@@ -35,6 +35,6 @@ describe('<IntroPage/>', () => {
 
   it('should render an alert when GIBS status down', () => {
     const wrapper = shallow(<IntroPage {...defaultProps} serviceAvailability={SERVICE_AVAILABILITY_STATES.down}/>);
-    expect(wrapper.find('.usa-alert')).to.have.lengthOf(2);
+    expect(wrapper.find('.usa-alert')).to.have.lengthOf(1);
   });
 });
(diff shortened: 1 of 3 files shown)


src/applications/post-911-gib-status/tests/containers/IntroPage.unit.spec.jsx
```diff
@@ -33,8 +33,8 @@ describe('<IntroPage/>', () => {
     expect(wrapper.find('Link').first().props().to).to.equal('status');
   });
 
-  it('should render an alert', () => {
+  it('should render an alert when GIBS status down', () => {
```
cvalarida: Is it just when GIBS is down?

cvalarida (approved):

src/applications/post-911-gib-status/tests/containers/IntroPage.unit.spec.jsx
markgreenburg: Well, this particular unit test was made to ensure that an error alert (which already exists) shows when GIBS status is down. The alert we just set up will show no matter what the GIBS status is - so in the case of GIBS being down, users would see this 'warning' alert, followed by an actual error message that says that GIBS is down for real. Given this is a temporary solution, I didn't think it was worthwhile to mess around with restructuring the component to make this new warning only show for GIBS up status.

markgreenburg: Thought about this some more and I think it's best to just show the banner when GIBS is up (since they already see another error when it's down and don't need to be reminded that it might not work when it in fact doesn't work). I made the banner conditional in a way that should be easy to strip back out later.

PR #8153, opened 2018-08-10
Deleted the email-gathering page and its javascript

cvalarida:
If we want to put it back in later, we can revert the merge into master.

### Staging
image

### Local (this PR)
image


SMLuthi (approved):
LGTM!

PR #8155, opened 2018-08-10
Datestamp formats

bshyong:
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/10295#issuecomment-410795841


ncksllvn (approved):
LGTM. I wonder if it would make sense to leverage the platform date helpers here via the `formatDateLong` function.

PR #8157, opened 2018-08-10
Feature/va website mobile

phiden:


accbjt:
@phiden all the merger content code should live in the va-gov folder and not the content folder.

ncksllvn:
Closing, because the branch has since diverged too much from the base branch.

PR #8158, opened 2018-08-10
Use country names instead of abbreviations

cehsu:


src/applications/edu-benefits/complaint-tool/config/form.js
```diff
@@ -419,11 +419,13 @@ const formConfig = {
                     updateSchema: (formData) => {
                       if (formData.address && formData.address.country && formData.address.country !== 'USA') {
                         let newSchema = _.set('properties.country.enum', countries, internationalSchoolAddress);
+                        newSchema = _.unset('required', newSchema);
```
cehsu: The nested manual entry required property is preventing school search submit.

rianfowler (approved):

PR #8161, opened 2018-08-10
Refactor ITFWrapper

cvalarida:
There's no ticket corresponding to this.

I was trying to figure out how to get the form to _not_ re-render after the ITF comes through, so I went about refactoring the wrapper. This didn't do the trick, but I figured it's maybe a better structure anyhow, so I figured I'd push it up.

Definitely willing to just close it out if that's the path we want to take. I just had this work sitting here nice and neat, so I decided to push it up.


markgreenburg (commented):
I prefer the way this component works compared to the old one, but I'd like to understand the nested conditions a bit better. Let's chat through it at some point and then I'm happy to approve.

markgreenburg (approved):
LGTM 🦄

PR #8162, opened 2018-08-11
Block first SiP call

cvalarida:
After _much_ debugging, Mark and I finally figured out what the problem was and outlined it in the connected ticket. I then hit my head against a wall for a few days trying to block the second SiP call, but kept running into cascading issues.

In the end, I had something of an epiphany and ended up feeling like an idiot for not thinking about this route sooner.

When a user hit the start button, we'd call `fetchInProgressForm` twice.
1) The common SiP `FormStartControls` called it properly when the button was hit
2) After that comes through successfully, we're redirected to the first page of the form
3) The ITFWrapper then makes its service call and renders a loading screen while we wait
4) After it comes through, it renders its children again
5) This causes `RoutedSavableApp` to believe it's getting started in the middle of a form and fetches the pre-fill data again, but this time, without the `prefillTransformer`
6) The `SET_IN_PROGRESS_FORM` action would then be called again and the old state (with the transformed data) would be merged with the new state (with the non-transformed data) and we'd get to an awkward state wherein some disabilities would have a rating when they shouldn't and any disabilities that _should_ have been filtered out now no longer have a `disabilityActionType` which prevents us from getting past the rated disabilities screen

My epiphany was to not stop the second SiP call (which was proving to be hugely bothersome), but to stop the _first_ one from `FormStartControls`. I still had to pass the `prefillTransformer` from `RoutedSavableApp`, but since `fetchInProgressForm()` checks to see if it's been prefilled or not before applying the transformer, I thought this was safe.

### DIFF
--- a/src/applications/disability-benefits/526EZ/components/IntroductionPage.jsx
+++ b/src/applications/disability-benefits/526EZ/components/IntroductionPage.jsx
@@ -40,6 +40,7 @@ class IntroductionPage extends React.Component {
           pathname={this.props.location.pathname}
           user={user}
           authenticate={this.authenticate}
+          noPrefill
           {...this.props}/>
         <h4>Follow the steps below to apply for increased disability compensation.</h4>
         <div className="process schemaform-process">
@@ -86,6 +87,7 @@ class IntroductionPage extends React.Component {
           pathname={this.props.location.pathname}
           user={user}
           authenticate={this.authenticate}
+          noPrefill
           {...this.props}
           buttonOnly/>
         {/* TODO: Remove inline style after I figure out why .omb-info--container has a left padding */}
--- a/src/platform/forms/save-in-progress/FormStartControls.jsx
+++ b/src/platform/forms/save-in-progress/FormStartControls.jsx
@@ -24,7 +24,7 @@ class FormStartControls extends React.Component {
   }
 
   handleLoadPrefill = () => {
-    if (this.props.prefillAvailable && !this.props.noPrefill) {
+    if (this.props.prefillAvailable && !this.props.skipPrefill) {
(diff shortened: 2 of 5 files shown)


src/platform/forms/save-in-progress/FormStartControls.jsx
```diff
@@ -23,7 +23,7 @@ class FormStartControls extends React.Component {
   }
 
   handleLoadPrefill = () => {
-    if (this.props.prefillAvailable) {
+    if (this.props.prefillAvailable && !this.props.noPrefill) {
```
markgreenburg: Do you want to throw this in the proptypes list?

markgreenburg: Might be worth documenting the difference between `prefillAvailable` and `noPrefill` somewhere, since they read pretty close. Just one of those things we might not understand a few months down the road.

markgreenburg (approved):
Couple of little comments but nothing blocking. LGTM!

src/platform/forms/save-in-progress/FormStartControls.jsx
cvalarida: Ahh, great point. Will do. :+1:

src/applications/disability-benefits/526EZ/components/IntroductionPage.jsx
```diff
@@ -40,6 +40,7 @@ class IntroductionPage extends React.Component {
           pathname={this.props.location.pathname}
           user={user}
           authenticate={this.authenticate}
+          noPrefill
```
jbalboni: Is `skipPrefill` maybe a better name? Feels like it's a little more descriptive of what's going on.

jbalboni (approved):
LGTM

PR #8165, opened 2018-08-13
Dashboard: Update title; remove empty breadcrumbs

ncksllvn:
Resolves department-of-veterans-affairs/vets.gov-team/issues/10167
Resolves department-of-veterans-affairs/vets.gov-team/issues/10502

### DIFF
--- a/src/applications/personalization/dashboard/containers/DashboardApp.jsx
+++ b/src/applications/personalization/dashboard/containers/DashboardApp.jsx
@@ -196,7 +196,7 @@ class DashboardApp extends React.Component {
     const view = (
       <div className="row user-profile-row">
         <div className="usa-width-two-thirds medium-8 small-12 columns">
-          <h1>Your Homepage</h1>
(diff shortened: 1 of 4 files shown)


src/applications/personalization/dashboard/containers/DashboardApp.jsx
```diff
@@ -196,7 +196,7 @@ class DashboardApp extends React.Component {
     const view = (
       <div className="row user-profile-row">
         <div className="usa-width-two-thirds medium-8 small-12 columns">
-          <h1>Your Homepage</h1>
+          <h1 id="dashboard-title">Your Homepage</h1>
           <div className="va-introtext">
```
ncksllvn: Wasn't happy about adding an `id`, but somehow we have style selectors like`.usa-width-one-whole:first-child > :first-child` that make things really difficult to override.

bshyong (approved):

PR #8176, opened 2018-08-14
Double restart bugfix cv

cvalarida:
I figured out how to fix the double fetch bug by preventing `RoutedSavableApp` from calling `fetchInProgressForm` if we pass it the `skipPrefill` prop. The only time we should want this behavior is when we deliberately wrap the component in another one which will cause it to be mounted, unmouted, and mounted again on a different page, so this seemed like a safe way to do it.

That aside, the double restart bug was happening because `RoutedSavableApp` thought the form was still saved when it wasn't. The scenario went a little something like this:

1) Load the intro page with a saved form
2) Hit start over
3) This would trigger the form delete, which would return successfully from api
4) This would trigger a `fetchInProgressForm`
5) Upon success, we'd get sent to the first page in the form, `/veteran-information`
6) This would trigger the `ITFWrapper` to do its thing and unmount `RoutedSavableApp` to display a loading spinner
7) After `ITFWrapper` was done, it'd mount a new `RoutedSavableApp` which would then check to see if it needed to fetch the form since it was being mounted in the middle of the form
8) It would see the list of saved forms with the 526 in it because that was never updated when we removed it
9) It would try to fetch the saved form and get an error

The main change here is to call the profile action `REMOVING_SAVED_FORM_SUCCESS` with the form ID which removes it from the list of saved forms `RoutedSavableApp` is hooked up to.

### DIFF
--- a/src/applications/disability-benefits/526EZ/components/FormStartControls.jsx
+++ b/src/applications/disability-benefits/526EZ/components/FormStartControls.jsx
@@ -22,7 +22,6 @@ export default function FormStartControls(props) {
         {...props}
         buttonOnly={props.buttonOnly}
         prefillAvailable
-        noPrefill={props.noPrefill}
         verifiedPrefillAlert={VerifiedAlert}
         unverifiedPrefillAlert={UnauthenticatedAlert}
         verifyRequiredPrefill={props.route.formConfig.verifyRequiredPrefill}
--- a/src/platform/forms/save-in-progress/actions.js
+++ b/src/platform/forms/save-in-progress/actions.js
@@ -7,6 +7,7 @@ import environment from '../../utilities/environment';
 import conditionalStorage from '../../utilities/storage/conditionalStorage';
 import { sanitizeForm } from '../helpers';
 import { removeFormApi, saveFormApi } from './api';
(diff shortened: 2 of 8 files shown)


src/applications/disability-benefits/526EZ/components/FormStartControls.jsx
```diff
@@ -22,7 +22,6 @@ export default function FormStartControls(props) {
         {...props}
         buttonOnly={props.buttonOnly}
         prefillAvailable
-        noPrefill={props.noPrefill}
```
cvalarida: I forgot to rename these last time. :man_facepalming:
Guess it doesn't matter now, eh?

src/platform/forms/save-in-progress/actions.js
```diff
@@ -343,6 +345,8 @@ export function removeInProgressForm(formId, migrations, prefillTransformer) {
         recordEvent({
           event: `${trackingPrefix}sip-form-start-over`
         });
+        // This action removes the form from the profile list
+        dispatch({ type: REMOVING_SAVED_FORM_SUCCESS, formId });
```
cvalarida: This is what makes it work.

It's needed now because of the re-mounting of `RoutedSavableApp`.

jbalboni (approved):
LGTM

PR #8174, opened 2018-08-14
Update transform for submit

cehsu:
Transform should conform to:
```
{
  gi_bill_feedback: {
    form: // form as JSON string
  }
}
```

PR #8185, opened 2018-08-15
display required on required field

rianfowler:
screen shot 2018-08-15 at 14 59 04

### DIFF
--- a/src/applications/edu-benefits/complaint-tool/config/form.js
+++ b/src/applications/edu-benefits/complaint-tool/config/form.js
@@ -283,8 +283,7 @@ const formConfig = {
                 },
                 'ui:errorMessages': {
                   atLeastOne: 'Please select at least one'
-                },
-                'ui:required': () => true
+                }
               },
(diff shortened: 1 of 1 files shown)


src/applications/edu-benefits/complaint-tool/config/form.js
```diff
@@ -283,7 +283,8 @@ const formConfig = {
                 },
                 'ui:errorMessages': {
                   atLeastOne: 'Please select at least one'
-                }
+                },
+                'ui:required': () => true
```
cehsu: Since this is not conditional we can set `required:  ['programs']` on the educationDetails schema for this. 
This should have been carried over when this schema was nested but was missed https://github.com/department-of-veterans-affairs/vets-website/pull/8174/files#diff-e03d34052d066c0db8877c8048b7a1f3L304

cehsu (approved):
LGTM, had a suggestion for an alternative approach.

PR #8187, opened 2018-08-15
Profile: Vet360 Initialization tweaks

ncksllvn:
This PR does the following -

- ~~Applies a Staging feature-flag so that Production users aren't affected~~
- Updates the transaction-refresh to allow for a route parameter, bc the initialization transaction has its own special route for checking transaction status.
- Fixes an error with the transaction-refresh being called before we have transaction data.

### DIFF
--- a/src/applications/personalization/profile360/vet360/actions/transactions.js
+++ b/src/applications/personalization/profile360/vet360/actions/transactions.js
@@ -50,7 +50,7 @@ export function clearTransactionRequest(fieldName) {
   };
 }
 
-export function refreshTransaction(transaction, analyticsSectionName) {
+export function refreshTransaction(transaction, analyticsSectionName, _route = null) {
   return async (dispatch, getState) => {
     try {
       const { transactionId } = transaction.data.attributes;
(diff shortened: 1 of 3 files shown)


src/applications/personalization/profile360/vet360/actions/transactions.js
```diff
@@ -66,7 +66,8 @@ export function refreshTransaction(transaction, analyticsSectionName) {
         transaction
       });
 
-      const transactionRefreshed = isVet360Configured() ? await apiRequest(`/profile/status/${transactionId}`) : await localVet360.updateTransaction(transactionId);
+      const route = _route || `/profile/status/${transactionId}`;
+      const transactionRefreshed = isVet360Configured() ? await apiRequest(route) : await localVet360.updateTransaction(transactionId);
 
       if (isSuccessfulTransaction(transactionRefreshed)) {
```
ncksllvn: I didn't know this, but there is a [special route](https://dsva.slack.com/archives/C909ZG2BB/p1534514851000100?thread_ts=1534434974.000100&cid=C909ZG2BB) for checking the initialization transaction status.

bshyong (approved):

PR #8191, opened 2018-08-16
Heroku preview style builds

wyattwalter:
Adds support for Heroku to dynamically do build types based upon the branch name in a PR. Any property of a PR is fair game, but so far we're just using branch name.

This uses a new property called `HEROKU_APP_NAME` which gets set to `vetsgov-pr-8191` and calls back to GitHub API. I didn't really add any error handling except to just print a message in the log and fall back to the default `development`. If we think this is a decent approach, I can add support for authenticating back to GitHub API so we don't get rate-limited and start seeing failures from that.

I had to mess with the output destination of Webpack and Metalsmith in Heroku builds because the Procfile uses a static web server and the logic to decide which path to use wouldn't have been easy to contain.

ref: https://github.com/department-of-veterans-affairs/vets.gov-team/issues/12115

### DIFF
--- a/app.json
+++ b/app.json
@@ -8,9 +8,6 @@
     },
     "HEROKU_APP_NAME": {
       "required": true
-    },
-    "HEROKU_PARENT_APP_NAME": {
-      "required": true
     }
   },
   "addons":[],
--- a/app.json
+++ b/app.json
@@ -5,6 +5,12 @@
     "CHECK_BROKEN_LINKS": "no",
     "YARN_PRODUCTION":{
       "required": true
+    },
+    "HEROKU_APP_NAME": {
+      "required": true
+    },
+    "HEROKU_PARENT_APP_NAME": {
+      "required": true
     }
   },
   "addons":[],
--- a/script/build.js
+++ b/script/build.js
@@ -62,36 +62,14 @@ if (options.buildtype === undefined) {
   options.buildtype = 'development';
 }
 
-if (process.env.HEROKU_APP_NAME) {
-  const request = require('sync-request');
-  options.destination = path.resolve(__dirname, '../build/heroku');
-  try {
-    const pr = process.env.HEROKU_APP_NAME.match(/pr-(\d+)$/)[1];
-    const res = request(
-      'GET',
-      `https://api.github.com/repos/department-of-veterans-affairs/vets-website/pulls/${pr}`,
-      {
-        headers: {
-          'User-Agent': 'vets-website-builder'
-        },
-        json: true
-      }
-    );
-    const respObj = JSON.parse(res.getBody('utf8'));
+const isHerokuBuild = !!process.env.HEROKU_APP_NAME;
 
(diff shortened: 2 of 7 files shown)


jkassemi (approved):
seems solid given the unfortunate lack of source detail from heroku 😢

script/build.js
```diff
@@ -61,6 +62,38 @@ if (options.buildtype === undefined) {
...
+      `https://api.github.com/repos/department-of-veterans-affairs/vets-website/pulls/${pr}`,
+      {
+        headers: {
+          'User-Agent': 'vets-website-builder'
+        },
+        json: true
+      }
+    );
+    const respObj = JSON.parse(res.getBody('utf8'));
+
```
ncksllvn: Ah, it's a shame to see another HTTP module in vets-website, but since the operation has to be done synchronously I don't think there's currently an easy way to do that installed.

app.json
```diff
@@ -5,6 +5,12 @@
     "CHECK_BROKEN_LINKS": "no",
     "YARN_PRODUCTION":{
       "required": true
+    },
+    "HEROKU_APP_NAME": {
+      "required": true
+    },
+    "HEROKU_PARENT_APP_NAME": {
+      "required": true
     }
```
ncksllvn: What does `HEROKU_PARENT_APP_NAME` do?

ncksllvn (approved):
Left a few small suggestions for clarity but otherwise LGTM!

ncksllvn:
Sorry, another question -

> I had to mess with the output destination of Webpack and Metalsmith in Heroku builds because the Procfile uses a static web server and the logic to decide which path to use wouldn't have been easy to contain.

Could you explain this a little more? I see how the destination changes locally versus in a review instance, but not sure if I understand why it's necessary.

app.json
wyattwalter: Ah yeah, that was from an earlier approach. I don't actually need it!

script/build.js
wyattwalter: Yeah, I spent a bunch of time trying to avoid pulling in yet another library but didn't find any other options that would work.

wyattwalter:
The output path is a bit tricky.

We use that in Jenkins to do builds against all environments in parallel, so it drops a directory in `/build` for each.

The thing that we use to host the files in Heroku takes a set of hard-coded arguments, so I needed to override the above behavior so that the Heroku process could know which directory to serve from ahead of time (this changes the buildtype to affect behavior).

app.json
wyattwalter: To answer this more directly:

We're (ab)using a feature of Heroku called "review applications" which uses a base application (that we don't actually deploy), and then spins up a special type of application for each PR opened (these are the review applications). The parent app in this context is the name of the app that this is "previewing".

ncksllvn:
Updates looks great @wyattwalter! And thanks for the info about Heroku!

PR #8192, opened 2018-08-16
capitalizeEach fix

cvalarida:
There was a double space in a disability name, so when we did a `name.split(' ').map(capitalizeEach)`, it tried to call `.toUpperCase()` on `undefined` which, of course, threw in an error.

### DIFF
--- a/src/applications/disability-benefits/526EZ/helpers.jsx
+++ b/src/applications/disability-benefits/526EZ/helpers.jsx
@@ -255,6 +255,11 @@ export const evidenceTypeHelp = (
 );
 
 const capitalizeEach = (word) => {
+  // Prevent calling toUpperCase() on undefined because of a double space in the name
+  if (!word) {
+    return '';
+  }
+
   const capFirstLetter = word[0].toUpperCase();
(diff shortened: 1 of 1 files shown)


src/applications/disability-benefits/526EZ/helpers.jsx
```diff
@@ -255,6 +255,11 @@ export const evidenceTypeHelp = (
 );
 
 const capitalizeEach = (word) => {
+  // Prevent calling toUpperCase() on undefined because of a double space in the name
+  if (!word) {
```
jbalboni: Another option might be to do `split(/ +/)`

jbalboni (approved):
Made an alternate suggestion, but LGTM either way.

src/applications/disability-benefits/526EZ/helpers.jsx
cvalarida: Oh hey. Forgot that `split` can take a regex too. I like that better.

markgreenburg:
@djmunoz1 this fixes the issue we had in Wednesday morning's UAT.

PR #8195, opened 2018-08-16
Change div to span to avoid React error

jbalboni:
Taking on the complicated problems today.

I don't see any style differences based on changing this (it's only on the payment page, from what I can tell), but I also don't see any commentary about why this was initially created as a `div` but with `display: inline` instead of a `span`.


cvalarida (approved):
:joy: :shipit:

PR #8196, opened 2018-08-16
add polling to claims status

rianfowler:
[Ticket](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/12278). Will put appeals status polling in another PR.


src/applications/claims-status/actions/index.jsx
```diff
@@ -147,16 +148,51 @@ export function fetchClaimsSuccess(response) {
...
+        onSuccess(response);
+        return;
+      }
+
+      if (shouldFail(response)) {
+        onFailure(response);
+        return;
+      }
+
+      setTimeout(
```
rianfowler: Noticed a lot of our polling implementations do timeout-first polling. Set up a poll first approach here.

src/applications/claims-status/actions/index.jsx
```diff
@@ -147,16 +148,51 @@ export function fetchClaimsSuccess(response) {
...
 }
 
-export function getClaimsV2() {
+const POLLING_INTERVAL = 1000;
+
+export function pollClaimsStatus({
+  onFailure,
+  onSuccess,
+  pollingInterval,
+  request = apiRequest,
```
rianfowler: This obviates the need to `mockFetch` which is tedious and pointless when we have a fetch helper that's already unit tested.

src/applications/claims-status/tests/actions.unit.spec.js
```diff
@@ -247,6 +249,153 @@ describe('Actions', () => {
...
+        shouldFailStub.returns(false);
+
+        pollClaimsStatus({
+          pollingInterval: 10,
+          request: apiRequestSpy,
+          shouldFail: shouldFailStub,
+          shouldSucceed: shouldSucceedStub
+        });
+        apiRequestSpy.firstCall.args[2]();
+        setTimeoutStub.firstCall.args[0]();
```
rianfowler: Unit testing recursive polling is gross.

cehsu (approved):
LGTM, pending a response to the question regarding what to dispatch if anything on success.

src/applications/claims-status/actions/index.jsx
```diff
@@ -1,5 +1,6 @@
 import React from 'react';
 import Raven from 'raven-js';
+import _ from 'lodash';
```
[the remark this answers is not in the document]
rianfowler: should we be using the get from the platform helpers?

src/applications/claims-status/actions/index.jsx
```diff
@@ -147,16 +148,51 @@ export function fetchClaimsSuccess(response) {
...
-    return apiRequest(
-      '/evss_claims',
-      null,
-      (response) => dispatch(fetchClaimsSuccess(response)),
-      // TO-DO: parse out errors, log in Sentry
-      () => dispatch({ type: FETCH_CLAIMS_ERROR })
-    );
+
+    poll({
+      onFailure: () => dispatch({ type: FETCH_CLAIMS_ERROR }),
```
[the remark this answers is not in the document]
rianfowler: I will add this to the clean up ticket I am going to make to remove the old claims status. We need some more information on what's returned on a failure and how it should be displayed. 

This approach just wraps the existing [behavior](https://github.com/department-of-veterans-affairs/vets-website/blob/5db1378e4b27c0b9a7ed95250aab8d50dd630fab/src/applications/claims-status/actions/index.jsx#L157) with polling.

src/applications/claims-status/actions/index.jsx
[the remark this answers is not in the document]
cehsu: Good idea.

src/applications/claims-status/actions/index.jsx
```diff
@@ -147,16 +148,51 @@ export function fetchClaimsSuccess(response) {
...
+
+export function pollClaimsStatus({
+  onFailure,
+  onSuccess,
+  pollingInterval,
+  request = apiRequest,
+  shouldFail,
+  shouldSucceed
+}) {
+  return request(
```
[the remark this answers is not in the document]
rianfowler: Not really sure what the best composition is. The polling logic is just `setTimeout` with a recursive call but I was thinking it would be good to have the forking behavior grouped i.e. 
- does it poll again?
- does it succeed?
- does it fail? 

I could pass another predicate / callback pair in like `shouldPoll` and `onPoll` so it's more consistent. Then I wouldn't need to pass in the polling interval. 

screen shot 2018-08-17 at 12 11 23

src/applications/claims-status/actions/index.jsx
```diff
@@ -147,16 +148,51 @@ export function fetchClaimsSuccess(response) {
...
+
+export function getClaimsV2(poll = pollClaimsStatus) {
+
   return (dispatch) => {
     dispatch({ type: FETCH_CLAIMS_PENDING });
-    return apiRequest(
-      '/evss_claims',
-      null,
-      (response) => dispatch(fetchClaimsSuccess(response)),
-      // TO-DO: parse out errors, log in Sentry
```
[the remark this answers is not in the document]
rianfowler: Yes 
https://app.zenhub.com/workspace/o/department-of-veterans-affairs/vets.gov-team/issues/12469

src/applications/claims-status/actions/index.jsx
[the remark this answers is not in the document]
rianfowler: Okay I see what you mean- I'll refactor this a bit.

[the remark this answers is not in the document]
rianfowler: Are you talking about drying the `get`? 

```
shouldFail: response => _.get(response, 'meta.syncStatus') === 'FAILED',
shouldSucceed: response => _.get(response, 'meta.syncStatus') === 'SUCCESS'
```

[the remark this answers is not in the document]
cehsu: Precisely.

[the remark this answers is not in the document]
rianfowler: Refactored that get to use the platform and added a helper for that.

I tried a couple of different ways to reorganize this but decided to leave it mostly the same. Couple differences from the burials polling example:
- It takes a timeout first approach which means the initial request doesn't fire until the first polling timeout 
- It's not using the `apiRequest` helper which has our error reporting baked in and is not promisified

src/applications/claims-status/actions/index.jsx
```diff
@@ -147,16 +148,55 @@ export function fetchClaimsSuccess(response) {
   };
 }
 
-export function getClaimsV2() {
```
cehsu: Will we also be making use of the `show` (`/v0/evss_claims_async/:id`) endpoint mentioned in the linked issue in another PR?

src/applications/claims-status/actions/index.jsx
```diff
@@ -147,16 +148,55 @@ export function fetchClaimsSuccess(response) {
...
-      null,
-      (response) => dispatch(fetchClaimsSuccess(response)),
-      // TO-DO: parse out errors, log in Sentry
-      () => dispatch({ type: FETCH_CLAIMS_ERROR })
-    );
+
+    poll({
+      onError: () => dispatch({ type: FETCH_CLAIMS_ERROR }),
+      onSuccess: response => dispatch(fetchClaimsSuccess(response)),
+      pollingInterval: window.VetsGov.pollTimeout || POLLING_INTERVAL,
```
cehsu: Do we anticipate using this in a setting where we would want to be able to configure the pollingInterval, rather than accessing it as a variable in the outer scope?

cehsu (approved):
LGTM. I had a few minor organizational questions for your consideration.

src/applications/claims-status/actions/index.jsx
```diff
@@ -147,16 +148,55 @@ export function fetchClaimsSuccess(response) {
...
+  return request(
+    '/evss_claims_async',
+    null,
+    response => {
+      if (shouldSucceed(response)) {
+        onSuccess(response);
+        return;
+      }
+
+      if (shouldFail(response)) {
```
cehsu: Do we anticipate wanting to configure shouldFail and shouldSucceed? If not, could we pull this logic into a checkStatus helper to house these checks, as well as the call to pollClaimsStatus (Similar to the burial app)?

cehsu (approved):
LGTM. Meant to include this question in the previous review.

src/applications/claims-status/actions/index.jsx
rianfowler: I plan on reusing this for the appeals status. Could have gone with a switch statement- this is mostly a style decision on my part to invert the control so it's easier to tell what's going on without having to read the polling function. I figure a maintainer will get to this from the action name or the action creator call, and they can tell what this is doing just by reading the arguments. 

screen shot 2018-08-20 at 13 53 04

src/applications/claims-status/actions/index.jsx
rianfowler: Nope- I would rather use the hard coded value since it's not reused. I'll update this.

src/applications/claims-status/actions/index.jsx
cehsu: Thanks for the context, sounds good.

cehsu: Since we are planning to reuse this for appeals, will it have the same shouldFail, and shouldSucceed logic? If so, could that simply be specified, rather than require configuration?

rianfowler: I think the extra boilerplate is fine- I can read the `poll` function's arguments and tell what its doing without reading its implementation. If I push the predicates into the polling function then I know it's "polling" but I don't know when it's going to call success or failure without looking at its implementation. 

Dunno if this is a legitimate design approach, though- it's inversion of control to me.

PR #8197, opened 2018-08-16
Check for .then() instead of instanceof Promise

cvalarida:
Turns out Firefox handles Promise constructors [a bit differently](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise#Browser_compatibility). [`instanceof`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/instanceof) checks the prototype chain for the constructor prototype, but...turns out that doesn't work on Firefox. :man_shrugging:

### DIFF
--- a/src/applications/disability-benefits/526EZ/components/AsyncDisplayWidget.jsx
+++ b/src/applications/disability-benefits/526EZ/components/AsyncDisplayWidget.jsx
@@ -44,7 +44,7 @@ export default class AsyncDisplayWidget extends React.Component {
     // TODO: Don't call the callback _every_ time the component is mounted
     const cbPromise = this.props.options.callback();
     // instanceof Promise doesn't work in Firefox, so we just check for .then() and hope it's a promise
-    if (typeof cbPromise.then === 'function') {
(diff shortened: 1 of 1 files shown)


src/applications/disability-benefits/526EZ/components/AsyncDisplayWidget.jsx
```diff
@@ -43,7 +43,8 @@ export default class AsyncDisplayWidget extends React.Component {
   componentDidMount() {
     // TODO: Don't call the callback _every_ time the component is mounted
     const cbPromise = this.props.options.callback();
-    if (cbPromise instanceof Promise) {
+    // instanceof Promise doesn't work in Firefox, so we just check for .then() and hope it's a promise
+    if (typeof cbPromise.then === 'function') {
```
jbalboni: You might want to check that cbPromise is truthy, otherwise if that callback function returns undefined or something, you're going to throw a different error than the one in the else block.

jbalboni (approved):
One suggestion, LGTM

markgreenburg (approved):
Nothing to add to Jeff's review.

PR #8199, opened 2018-08-17
Split up feedback tool applicant information page

cehsu:
These changes fix a bug where the schemaform library was deleting hidden prefilled data, by reorganizing the form pages, and removing the problematic hiding logic.

Unit tests and e2e tests have been updated.

In addition, the backend schema has been used directly where possible. There is a [PR](https://github.com/department-of-veterans-affairs/vets-json-schema/pull/244) to update the backend schema to include a missing program type, once the updated schema is used here, the current unit test failure (`src/applications/edu-benefits/tests/complaint-tool/config/benefitsInfo.unit.spec.js:21:42`) will be resolved.


src/applications/edu-benefits/complaint-tool/config/form.js
```diff
@@ -1,22 +1,40 @@
...
+  phone,
+  serviceAffiliation,
+  serviceBranch,
+  serviceDateRange,
+} = fullSchema.properties;
 
-const { school } = educationDetails;
+const { assistance, programs, school } = educationDetails;
 
 const { schoolInformation } = school.oneOf[0];
```
cehsu: This will be replaced with `const schoolInformation = school.oneOf[0];` once the updated BE schema is used.

rianfowler (commented):
e2e tests run fine. not sure if this is related to your PR but here are a couple of issues I saw that aren't in staging: 
user facing labels are missing
screen shot 2018-08-17 at 09 35 58

Entered data missing from review page 
screen shot 2018-08-17 at 09 41 05

cehsu:
Thanks @rianfowler:
1. the labels are missing because the BE schema does not provide them, and the BE schema is now being used to generate the schema on the FE. 
2. I was not able to reproduce the review page bug you identified, could you please provide the steps to reproduce?

rianfowler:
@cehsu I wasn't able to reproduce- I might have refreshed the form before getting to the review page.

rianfowler (approved):

src/applications/edu-benefits/complaint-tool/config/form.js
```diff
@@ -387,40 +351,40 @@ const formConfig = {
           uiSchema: {
             educationDetails: {
               school: {
-                facilityCode: {
+                facilityCode: { // Can we unnest this?
```
cehsu: @rianfowler can we unnest facilityCode.facilityCode?

cehsu: It looks like the nesting is necessary for the custom field?

rianfowler: It only needs the string- it doesn't have to be submitted this way.

cehsu: Since we have `view:manualSchoolEntry` perhaps this should be `view:searchSchoolEntry`?

rianfowler: That works. Will need to update some selectors

cehsu: Thanks for the heads up, will update accordingly.

PR #8200, opened 2018-08-17
Switch to Formation ErrorableCheckboxGroup in wizard

jbalboni:
No visual change:

screen shot 2018-08-17 at 10 26 30 am


markgreenburg (approved):
Great, thanks for catching this! I bet we're using non-standard components in other places that don't need them as well.

PR #8205, opened 2018-08-17
All claims 1dot2 mg

markgreenburg:
## Description
Adds a page with a required yes/no question about whether the veteran has served with other names. Selecting 'yes' expands a names section that requires the veteran to fill out at least one previous name. 

## Testing done
- Tested locally in Chrome

## Screenshots
Base screen:
screen shot 2018-08-20 at 11 26 04 am

Selecting 'yes' expands a names form:
screen shot 2018-08-20 at 12 55 02 pm

A veteran can add up to a 100 names (which is probably not super useful, but this is the upper limit that the evss swagger dictates):
screen shot 2018-08-20 at 12 55 22 pm

#### Applies to all PRs

- [x] Appropriate logging


src/applications/disability-benefits/all-claims/config/form.js
```diff
@@ -34,11 +38,16 @@ const formConfig = {
       pages: {
         veteranInformation: {
           title: 'Veteran Information',
-          description: 'This is the personal information we have on file for you.',
```
markgreenburg: Didn't need to be a part of this PR, but this property doesn't do anything (needs a 'ui:description' property instead) so wanted to remove it to clean things up.

src/applications/disability-benefits/all-claims/pages/alternateNames.js
```diff
@@ -0,0 +1,66 @@
...
+      maxItems: 100,
+      items: {
+        type: 'object',
+        required: ['first', 'last'],
+        properties: {
+          first: {
+            type: 'string',
+            minLength: 1,
+            maxLength: 30,
+            pattern: "^([a-zA-Z0-9-/']+( ?))+$"
```
markgreenburg: pattern & length requirements taken from EVSS `submit` swagger

jbalboni: I think we would save ourselves a step later if we still wrote the pages as the schema was in a separate file in vets-json-schema, and just kept that file in vets-website. That's the approach that the initial version of the pre-need form took. It saves you from having to delete all of this code and replace it with the imported schema.

src/applications/disability-benefits/all-claims/pages/alternateNames.js
```diff
@@ -0,0 +1,66 @@
...
+  'view:hasAlternateName': {
+    'ui:title': 'Did you serve under another name?',
+    'ui:widget': 'yesNo'
+  },
+  alternateNames: {
+    'ui:description': 'What names did you serve under?',
+    'ui:options': {
+      viewField: FullNameField,
+      reviewTitle: 'Other names',
+      expandUnder: 'view:hasAlternateName'
```
jbalboni: You could also include `addAnotherLabel` and set it to `Add Another Name` to change the add another button label. That's usually something the content folks ask for in their review on other forms I've worked on.

jbalboni (commented):
Had a couple of comments, but the config itself looks good to me.

src/applications/disability-benefits/all-claims/pages/alternateNames.js
markgreenburg: Yeah, that seems like a good call.

src/applications/disability-benefits/all-claims/pages/alternateNames.js
markgreenburg: Ah! Thanks, didn't know that was a thing.

src/applications/disability-benefits/all-claims/pages/alternateNames.js
```diff
@@ -0,0 +1,42 @@
...
+    'ui:title': 'Did you serve under another name?',
+    'ui:widget': 'yesNo'
+  },
+  alternateNames: {
+    'ui:description': 'What names did you serve under?',
+    'ui:options': {
+      viewField: FullNameField,
+      reviewTitle: 'Other names',
+      expandUnder: 'view:hasAlternateName',
+      itemName: 'Name'
```
markgreenburg: Looks like `ArrayField` doesn't support `addAnotherLabel`, but it does support `itemName` instead, which basically amounts to the same thing:
https://github.com/usds/us-forms-system/blob/2410134ffda3767058ae7ef16e512e3fc5124632/src/js/fields/ArrayField.jsx#L294

jbalboni (approved):

src/applications/disability-benefits/all-claims/config/schema.js
```diff
@@ -0,0 +1,41 @@
+// This file to be moved to vets-json-schema repo before form hits production
```
cvalarida: Ahhhhh, I see. I like this.

cvalarida (approved):
Looks great!

peggygannon (approved):
@markgreenburg Hi, Mark. The screen looks good. I wanted to ask however, if the link at the bottom of the screen should say: 
[Save and finish this application later.]

src/applications/disability-benefits/all-claims/config/schema.js
markgreenburg: This was all Jeff's idea :)

PR #8206, opened 2018-08-17
Update e2e test config and README

erikphansen:


rianfowler (approved):

README.md
```diff
@@ -31,8 +31,8 @@ very secret.
 | run only unit tests | `npm run test:unit` |
 | run all unit tests and watch | `npm run test:watch` |
 | run only unit tests for a subset of tests | `npm run test:unit -- path/to/my/test.unit.spec.jsx` <br> or <br> `npm run test:unit -- --recursive 'path/to/my/**/*.unit.spec.js?(x)'` |
-| run only e2e tests                       | `npm run test:e2e`                       |
-| run only e2e tests for a subset of tests | `npm run test:e2e -- test/edu-benefits/1995/*.e2e.spec.js` (provide file paths) |
+| run only e2e tests | Make sure the site is running locally (`npm run watch`) and run the tests with `npm run test:e2e` |
```
cehsu: Is there any interest in noting that the mock server port should be made available? Or should that kind of thing be in a FAQ?

cehsu (approved):

README.md
erikphansen: As in make sure that the APIs aren't running locally? I believe the script warns you if the port isn't available. But I'll add it as a caveat here if others agree is belongs here.

ncksllvn (approved):

README.md
cehsu: Sounds good as is.

PR #8207, opened 2018-08-17
heroku github auth

wyattwalter:
## Description
Provides authentication to API call back to GitHub from the Heroku review application build process. Auth isn't necessary to read the data we need (and the token has zero privileges), but helps us avoid rate-limiting.

## Testing done
Only affects builds in Heroku, tested there

#### Unique to this PR
- [x] authenticates requests back to GitHub

#### Applies to all PRs

- [x] Appropriate logging

https://github.com/department-of-veterans-affairs/vets.gov-team/issues/12115

### DIFF
--- a/app.json
+++ b/app.json
@@ -8,6 +8,9 @@
     },
     "HEROKU_APP_NAME": {
       "required": true
+    },
+    "FOO": {
+      "required": true
     }
   },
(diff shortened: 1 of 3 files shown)


ncksllvn (approved):
LGTM!

PR #8212, opened 2018-08-18
Fetch user profile after MHV upgrade and before UI updates.

U-DON:
## Description
Fetch user profile immediately after invoking MHV upgrade in the Redux action. This fixes a bug where the UI flickers an access error right before fetching the user. Resolves department-of-veterans-affairs/vets.gov-team#11307.

## Testing done
Created an E2E test that will be pushed in a separate PR as it makes an important change to E2E utilities.

## Testing Plan
Tested with automated testing, although there is no delay in requests due to mocked responses. Can be tested manually if need be, but would require upgrading one of our test accounts.

User should not see an error flicker during a successful MHV upgrade.

#### Applies to all PRs

- [x] Appropriate logging


dhall (approved):
LGTM

PR #8214, opened 2018-08-20
Fix hanging spinner on disability benefits pages

jbalboni:
## Description
Update the static widgets code to allow you to set a buildtype, in order to exclude the widgets in certain environments.

## Testing done
Tested locally, updated unit tests

## Testing Plan
We should verify this still allows the widget to show up in dev and staging.

#### Unique to this PR
- [ ] Spinner no longer hangs in production
- [ ] Status widget still works in staging

#### Applies to all PRs

- [x] Appropriate logging

### DIFF
--- a/content/pages/disability-benefits/index.md
+++ b/content/pages/disability-benefits/index.md
@@ -9,7 +9,7 @@ template: level2-index
 widgets:
   - root: react-applicationStatus
     timeout: 20
-    buildtype: staging
+    production: false
     loadingMessage: Checking your application status.
     errorMessage: <strong>We’re sorry. Something went wrong when we tried to load your saved application.</strong><br/>Please try refreshing your browser in a few minutes.
 majorlinks:
--- a/content/pages/disability-benefits/index.md
+++ b/content/pages/disability-benefits/index.md
@@ -9,6 +9,7 @@ template: level2-index
 widgets:
   - root: react-applicationStatus
     timeout: 20
+    buildtype: staging
     loadingMessage: Checking your application status.
     errorMessage: <strong>We’re sorry. Something went wrong when we tried to load your saved application.</strong><br/>Please try refreshing your browser in a few minutes.
 majorlinks:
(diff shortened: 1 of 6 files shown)


content/pages/disability-benefits/index.md
```diff
@@ -9,6 +9,7 @@ template: level2-index
 widgets:
   - root: react-applicationStatus
     timeout: 20
+    buildtype: staging
```
jbalboni: So, I'm not sure this is a great way of doing this, since it's adding yet another way we can gate features. Open to other suggestions.

markgreenburg: As an alternative, can we comment out the widgets section here entirely until we're ready? Maybe add it to a launch checklist or something.

markgreenburg (approved):
LGTM. Had an alternative suggestion but really I'm not sure what the best path forward is so feel free to take it or leave it.

content/pages/disability-benefits/index.md
jbalboni: We could, but you wouldn't be able to test out the widgets in staging, which can be confusing during final content passes before launch. But I don't feel super strongly either way.

cvalarida:
I think what we've done before is put a `buildtype` gate in the widget itself. Not sure if that's a _better_ solution, but it follows a pattern we've already got established in other places.

jbalboni:
@cvalarida That exists now, but since there's a `widgets` entry in the content page, there's some JS set up code that is embedded directly into the page, so that we don't leave a spinner up forever if something goes wrong in the JS bundles (which is what's happening here, it spins for 20 seconds and then shows an error).

cvalarida:
Ahh, okay. I see. :thinking:

What if we did something more like what we do for showing or hiding apps in production?

```yaml
widgets:
  - root: react-applicationStatus
    timeout: 20
    production: false
```

Will we need the extra flexibility of showing it in dev but not staging?

At any rate, it looks good the way it is. No qualms here. I think I'm just trying to figure out something a little more mnemonic than `buildtype: staging` since that's not a pattern we currently use except within javascript.

jbalboni:
@cvalarida Yeah, I think that's better. I've updated it to use that format.

PR #8215, opened 2018-08-20
526 remove old requireds mg

markgreenburg:
## Description
- Removes some fanciness in modifying the treatment schema that's imported form vets-json-schema, because we cleaned that schema up and now don't need to modify it after the import
- Updates the vets-json-schema version that vets-website uses

Technically, this PR doesn't _need_ to be implemented, because the validation problems are purely a vets-api concern (what we submit doesn't change). We've already cleaned up vets-json-schema so that what we submit to vets-api doesn't get incorrectly flagged as invalid, so this PR is really just for cleaning up dead code and doing the due diligence of bumping the vets-json-schema commit hash to point to the new schema to alleviate future mystery bugs.

## Testing done
- Tested locally all the way through to submit; unit tests pass locally.

## Testing Plan
- [ ] Validate on staging
- [ ] Validate on production

## Screenshots
No UI changes

- [x] Able to submit form without getting back validation errors for missing `treatmentCenterType` or `servedInCombatZone` properties.

- [x] Appropriate logging

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -10177,9 +10177,9 @@ verror@1.3.6:
   dependencies:
     extsprintf "1.0.2"
 
-"vets-json-schema@https://github.com/department-of-veterans-affairs/vets-json-schema.git#f42bad3ddff042d2700dd93a2dccd646a755f905":
-  version "3.79.0"
-  resolved "https://github.com/department-of-veterans-affairs/vets-json-schema.git#f42bad3ddff042d2700dd93a2dccd646a755f905"
+"vets-json-schema@https://github.com/department-of-veterans-affairs/vets-json-schema.git#a87b73d66112a7f4279e92b4eafe8fff14e27fc2":
+  version "3.80.0"
+  resolved "https://github.com/department-of-veterans-affairs/vets-json-schema.git#a87b73d66112a7f4279e92b4eafe8fff14e27fc2"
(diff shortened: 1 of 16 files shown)


yarn.lock
```diff
@@ -10177,9 +10177,9 @@ verror@1.3.6:
   dependencies:
     extsprintf "1.0.2"
 
-"vets-json-schema@https://github.com/department-of-veterans-affairs/vets-json-schema.git#2b1e3463725279589728e666f478a756769bf580":
-  version "3.76.0"
-  resolved "https://github.com/department-of-veterans-affairs/vets-json-schema.git#2b1e3463725279589728e666f478a756769bf580"
+"vets-json-schema@https://github.com/department-of-veterans-affairs/vets-json-schema.git#f42bad3ddff042d2700dd93a2dccd646a755f905":
+  version "3.79.0"
```
jbalboni: This is a couple versions up, it looks like. Are the other changes in between just to the 526, or other forms?

markgreenburg: Fair question! These were all small 526-related changes that didn't have any impact on the form / config.
In any case, looks as though someone bumped the version to 3.80 in the meantime 😬

cvalarida (approved):
Assuming this is golden in the schema, it looks good.

PR #8224, opened 2018-08-22
Fixed redirect issues with MHV upgrade flow.

U-DON:
## Description
The flow actually works properly without these changes when I tested locally. It looks like the URL automatically appends the trailing slash between the URL and the query param. In contrast, when I tested on Staging, it looks like the URL dropped any query params when they were not separated from the main URL by the slash.

For example, on Staging, I could navigate to `/account/?query_param=testing`, which would keep the query param. But navigating to `/account?query_param=testing` would drop it and become just `/account`. This is clearly problematic when we store the redirect URL in the query param. As I explained above, when I tested locally, either URL would resolve to the one with the trailing slash.

I think this will work for now as a hack (and it's not really much of a hack anyway), but I think the root issue is something with how the routing is configured in other environments. I could've sworn the redirects were working on other environments before, so it's possible that it was a recent change.

Resolves department-of-veterans-affairs/vets.gov-team#12445.

## Testing done
Testing explained above. ^

## Testing Plan
Test verification flow in Staging. Navigate to Secure Messaging as an Advanced user, which should redirect to verify page. Confirm whether the URL retains a `?next=...` query parameter.

If it is retained, compare by removing the slash between the URL and the `?next` param to verify that the param is dropped when there is no trailing slash.

Verify and T&C acceptance should redirect the user back to the app (Secure Messaging) if they were redirected initially from the app (Secure Messaging).

#### Applies to all PRs

### DIFF
--- a/src/platform/user/authorization/containers/MHVApp.jsx
+++ b/src/platform/user/authorization/containers/MHVApp.jsx
@@ -85,7 +85,7 @@ export class MHVApp extends React.Component {
     // rendered, but if it hasn't for some reason, we will redirect now.
     if (accountState === 'needs_identity_verification') {
       const nextQuery = { next: window.location.pathname };
-      const verifyUrl = appendQuery('/verify', nextQuery);
+      const verifyUrl = appendQuery('/verify/', nextQuery);
       window.location.replace(verifyUrl);
       return;
     }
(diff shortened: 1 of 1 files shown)


dhall (approved):
LGTM

saneshark (approved):
@U-DON this LGTM, but I'm curious to understand what those other changes might have been to routing. This is a difficult thing to test so we want to make sure we get it right.

Let's only merge this if we are prepared to test it on staging.

@davidsbao

PR #8225, opened 2018-08-22
Add prefill messaging to GIBFT and track changes to On Behalf Of...

erikphansen:
## Description
1. Adds prefill messaging to appropriate pages of GIBFT
1. Records GA events when changing who the feedback is being filed on behalf of

## Testing done
- Tested locally to confirm that the prefill message function is called, but was unable to test with a prefill enabled user locally.
- Confirmed that the GA event recording function was called when needed.

## Testing Plan
Manual testing of a user with prefill enabled for the GIBFT. Confirm that the prefill message appears on the following pages:
- [ ] applicant information
- [ ] service information
- [ ] contact information
Also
- [ ] Confirm that the correct events are logged in GA

#### Applies to all PRs

- [ ] Appropriate logging N/A
- [ ] Documentation has been updated, if applicable N/A
- [ ] Provide which alerts would indicate a problem with this functionality (if applicable) N/A

### DIFF
--- a/src/applications/edu-benefits/sass/edu-benefits.scss
+++ b/src/applications/edu-benefits/sass/edu-benefits.scss
@@ -75,9 +75,6 @@
   }
 
   .radio-button {
-    input[type="radio"] {
-    }
-
     span {
       display: inline-block;
       width: 100%;
--- a/src/applications/edu-benefits/components/UserInteractionRecorder.jsx
+++ b/src/applications/edu-benefits/components/UserInteractionRecorder.jsx
@@ -1,0 +1,36 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+/**
+ * This renderless-component is used to register events with Google Analytics
+ * by tying into a US Form System's uiSchema['ui:description'] property. When
+ * the data on the form page changes, this component's `componentDidUpdate()`
+ * method fires and registers an event with Google Analytics (via `recordEvent`)
+ * if the property the component cares about (`selectedValue`) has changed.
+ *
+ * @export
+ * @class UserInteractionRecorder
+ * @extends {React.Component}
+ */
+export default class UserInteractionRecorder extends React.Component {
+  componentDidUpdate(prevProps) {
+    const { trackingEventMap, selectedValue, eventRecorder } = this.props;
+    const { selectedValue: oldSelectedValue } = prevProps;
+    if (selectedValue === oldSelectedValue) {
+      return;
+    }
+    if (trackingEventMap[selectedValue]) {
+      eventRecorder(trackingEventMap[selectedValue]);
+    }
+  }
+
+  render() {
+    return null;
+  }
+}
+
+UserInteractionRecorder.propTypes = {
+  trackingEventMap: PropTypes.objectOf(PropTypes.object),
+  selectedValue: PropTypes.string,
+  eventRecorder: PropTypes.func,
+};
--- a/src/applications/edu-benefits/feedback-tool/config/form.js
+++ b/src/applications/edu-benefits/feedback-tool/config/form.js
@@ -7,6 +7,7 @@ import { validateBooleanGroup, validateMatch } from 'us-forms-system/lib/js/vali
 
 import FormFooter from '../../../../platform/forms/components/FormFooter';
 import fullNameUI from '../../../../platform/forms/definitions/fullName';
+import PrefillMessage from '../../../../platform/forms/save-in-progress/PrefillMessage';
 import dataUtils from '../../../../platform/utilities/data/index';
 
 const { get, omit, set } = dataUtils;
@@ -160,6 +161,7 @@ const formConfig = {
           title: 'Applicant Information',
           depends: isNotAnonymous,
           uiSchema: {
+            'ui:description': PrefillMessage,
             fullName: _.merge(fullNameUI, {
               prefix: {
                 'ui:title': 'Prefix',
@@ -202,6 +204,7 @@ const formConfig = {
           title: 'Service Information',
           depends: isVeteranOrServiceMember,
           uiSchema: {
+            'ui:description': PrefillMessage,
             serviceBranch: {
               'ui:title': 'Branch of service',
             },
@@ -224,6 +227,7 @@ const formConfig = {
           title: 'Contact Information',
           depends: (formData) => formData.onBehalfOf !== anonymous,
           uiSchema: {
+            'ui:description': PrefillMessage,
             address: {
               street: {
                 'ui:title': 'Address line 1'
--- a/src/applications/edu-benefits/feedback-tool/config/form.js
+++ b/src/applications/edu-benefits/feedback-tool/config/form.js
@@ -17,7 +17,7 @@ import ConfirmationPage from '../containers/ConfirmationPage';
 import SchoolSelectField from '../../components/SchoolSelectField.jsx';
 import GetFormHelp from '../../components/GetFormHelp';
 
-import { transform, submit } from '../helpers';
+import { transform, submit, recordApplicantRelationship } from '../helpers';
 
 const {
   address: applicantAddress,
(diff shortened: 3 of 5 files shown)


src/applications/edu-benefits/complaint-tool/helpers.js
```diff
@@ -128,3 +129,26 @@ export function submit(form, formConfig) {
...
+};
+
+// This is a hack. This function is intended to be used as a uiSchema's
+// ui:description callback. We're simply taking advantage of the fact that
+// ui:description callbacks are fired whenever the form data changes so that
+// we can check the current state of the form data and register what option the
+// user selected for who they are filing feedback on behalf of.
+export function recordApplicantRelationship({ formData }) {
+  const onBehalfOf = _.get('onBehalfOf', formData);
+  if (applicantRelationshipEventMap[onBehalfOf]) {
```
cehsu: We will likely want to use a react lifecycle method such as [`componentDidUpdate`](https://reactjs.org/docs/react-component.html#componentdidupdate), to ensure that this event only fires when a change is made (by comparing prevProps with current props), and not every time an applicant relationship field value exists.

src/applications/edu-benefits/sass/edu-benefits.scss
```diff
@@ -75,9 +75,6 @@
   }
 
   .radio-button {
-    input[type="radio"] {
```
cehsu: Could you please provide context on this change?

src/applications/edu-benefits/complaint-tool/config/form.js
```diff
@@ -202,6 +205,7 @@ const formConfig = {
           title: 'Service Information',
           depends: isVeteranOrServiceMember,
           uiSchema: {
+            'ui:description': PrefillMessage,
```
cehsu: @jholton @lihanli will service information always be prefilled? How about address information? If these are not always prefilled we may want to conditionally display this prefill message. @erikphansen we could add a field 'view:serviceInformationPrefilled' and set it to true if service information, for example, is included in the prefilled information, via a prefill transformer (e.g. [0993](https://github.com/department-of-veterans-affairs/vets-website/blob/8e1427fa92037ebdd57f889373caebcd3eeb4d35/src/applications/edu-benefits/0993/helpers.js#L4-L17)), and then check for that value in determining whether to display the prefill message.

src/applications/edu-benefits/complaint-tool/helpers.js
```diff
@@ -1,3 +1,4 @@
+import _ from 'lodash/fp';
```
cehsu: We could use the platform data utils here instead, as is done in the config.
```
import dataUtils from '../../../../platform/utilities/data/index';
 const { get, omit, set } = dataUtils;
```

erikphansen: Sure. I swear I've seen `_.get` used a lot in recent PRs. Should we really be using our own `get` method moving forward?

cehsu: Yeah, it definitely depends on the situation. I believe lodash works for both objects and arrays, whereas the platform utils may not. The signature is also different. The platform data utils are intended to help keep bundle sizes small, when possible. If we have ideas for updating/improving the utils so that they are more useful, we can set up some tickets for those, and use lodash for the time being.

src/applications/edu-benefits/complaint-tool/helpers.js
```diff
@@ -128,3 +129,26 @@ export function submit(form, formConfig) {
...
+  Anonymous: 'anonymous',
+};
+
+// This is a hack. This function is intended to be used as a uiSchema's
+// ui:description callback. We're simply taking advantage of the fact that
+// ui:description callbacks are fired whenever the form data changes so that
+// we can check the current state of the form data and register what option the
+// user selected for who they are filing feedback on behalf of.
+export function recordApplicantRelationship({ formData }) {
+  const onBehalfOf = _.get('onBehalfOf', formData);
```
cehsu: We can also use destructuring here rather than a utility.

src/applications/edu-benefits/complaint-tool/helpers.js
erikphansen: That totally makes sense! Thanks.

src/applications/edu-benefits/complaint-tool/config/form.js
jholton: If a piece of prefill data is unavailable, it will not be in the prefill `formData` json at all (the key will not be present) that the backend returns.

cehsu: Thanks @jholton, in general is only name and email guaranteed to be prefiled for signed in users? Would address and service information only be prefilled for verified users (is not guaranteed to always be prefilled for signed in users)?

jholton: I honestly have no idea about the general availability of address and service information.
I'd say operating as if only email and name are guaranteed is a good idea.
@lihanli do you have any knowledge as to the reliability of an address or service information being available for signed-in users?

lihanli: neither of those are guaranteed for verified users

cehsu: Thanks @jholton and @lihanli, we'll ensure that those (address, service information) have been prefilled when displaying the message.

cehsu: @erikphansen I think this (checking that individual pages/fields were actually prefilled) is a new pattern, so feel free to create a new ticket if it's something that we decide we want to implement.

src/applications/edu-benefits/complaint-tool/config/form.js
```diff
@@ -160,6 +162,7 @@ const formConfig = {
           title: 'Applicant Information',
           depends: isNotAnonymous,
           uiSchema: {
+            'ui:description': PrefillMessage,
```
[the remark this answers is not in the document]
cehsu: If so, a new ticket could be set up for this work.

src/applications/edu-benefits/complaint-tool/config/form.js
erikphansen: Done [here](https://app.zenhub.com/workspace/o/department-of-veterans-affairs/vets.gov-team/issues/12607)

src/applications/edu-benefits/complaint-tool/config/form.js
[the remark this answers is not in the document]
emilywaggoner: I think if we can do that, we should. Thanks @cehsu!

jbalboni:
Are you trying to capture how many forms are submitted in each relationship type bucket, or are you trying to record how users switch between the relationship type options?

erikphansen:
The goal here is to track, in real time, the user's interaction with the On Behalf Of radio buttons on the GI Bill Feedback Tool. Part of the thinking here is so that a small bit of data can be captured for users who don't complete the form and fall out of the funnel.

jbalboni (approved):

erikphansen:
I 100% agree about this feeling wrong. I used to have a comment in my code pointing out that this was a hack.

1. Long term we are looking to modify the US Form System to allow adding GA hooks in the form config. Before we do that, however, we want a better idea of what sort of capabilities to add to the form system. So in the short term we are going to add these features on a one-off basis.
2. Wouldn't adding a config option also involve adding capabilities to the `us-forms-system`? I think maybe you're thinking along the same lines as my point 1 above. But please elaborate if you want to!
3. I'm not sure what this would involve as I'm still pretty new to using React outside of simple projects on my own and am *very* new to Redux.

The bottom line is that this is hopefully a short term hack. I probably went overboard making a flexible component we could easily reuse in a similarly hacky manner.

jbalboni:
Since we still have `RoutedSavablePage` in vets-website, you could add essentially the same hooks you're thinking of for 1 to it without modifying us-forms-system. But ideally they'd probably go in `FormPage` in us-forms-system.

3 would be a little more involved to set up. If this is meant to be short term, then I think I would leave it as is for now.

erikphansen:
Thanks for those final comments. They should come in handy as I look into more long term solutions for expanding analytics.

PR #8229, opened 2018-08-23
All claims 1.5 service info

jbalboni:
## Description
This adds four service related pages to the veteran details chapter. This is a WIP because I still have outstanding questions about what fields I need to be adding for these pages.

## Testing done
Local testing, unit testing

## Testing Plan
Unit tests, plus testing in staging

## Screenshots
screen shot 2018-08-23 at 11 40 18 am
screen shot 2018-08-23 at 11 40 05 am
screen shot 2018-08-23 at 2 27 44 pm
screen shot 2018-08-23 at 11 38 56 am

#### Unique to this PR

- [x] Correct pages are added
- [x] Correct fields are added
- [x] Verified with UX and backend that fields being added will be in the API and are necessary
- [x] Code that's been moved from the increase form hasn't broken that form

#### Applies to all PRs

- [x] Appropriate logging


src/applications/disability-benefits/all-claims/pages/reservesNationalGuardService.jsx
```diff
@@ -0,0 +1,163 @@
+import { get } from 'lodash';
+
+import fullSchema from '../config/schema';
+
+import dateRangeUI from 'us-forms-system/lib/js/definitions/dateRange';
+import {
+  ReservesGuardDescription,
+} from '../utils';
+
+import  {
```
jbalboni: There's an address in here, which maybe should be its own definition, but I figured we could leave it alone until we get an answer from EVSS on if they'll add unitAddress to their API.

src/applications/disability-benefits/all-claims/utils.jsx
```diff
@@ -17,3 +22,58 @@ const srSubstitute = (srIgnored, substitutionText) => {
 };
 
 export default srSubstitute;
+
+export const hasGuardOrReservePeriod = (formData) => {
+  const serviceHistory = formData.servicePeriods;
```
markgreenburg: I think this needs to actually look at `formData.serviceInformation.servicePeriods`

src/applications/disability-benefits/all-claims/pages/reservesNationalGuardService.jsx
```diff
@@ -0,0 +1,163 @@
...
+            pattern: 'Please enter a valid 5- or 9-digit ZIP code (dashes allowed)'
+          },
+          'ui:options': {
+            widgetClassNames: 'va-input-medium-large',
+            hideIf: formData => get(formData, 'serviceInformation.reservesNationalGuardService.unitAddress.country', USA) !== USA,
+          }
+        }
+      },
+      unitPhone: {
+        'ui:title': 'Unit phone number',
```
markgreenburg: I think in the past we've used `PhoneNumberWidget` to do fancy things like allow dashes.

markgreenburg (commented):
99% LGTM, found a little validation thing that would probably make sense to address before merging.

cvalarida (approved):
Like you called out in your AC list, it would be good to make sure moving around the helpers doesn't break the increase-only app. Other than that, it looks good. Huge, but straightforward.

src/applications/disability-benefits/all-claims/config/form.js
```diff
@@ -40,7 +66,9 @@ const formConfig = {
   confirmation: ConfirmationPage,
   // footerContent: FormFooter,
   // getHelp: GetFormHelp,
-  defaultDefinitions: {},
+  defaultDefinitions: {
+    ...fullSchema.definitions
```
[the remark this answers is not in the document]
jbalboni: Pretty sure I copied it from somewhere.

src/applications/disability-benefits/all-claims/utils.jsx
jbalboni: Good catch. For now, I changed the form config page to pass the service information object, rather than create a separate helper.

markgreenburg (approved):
🎉

PR #8247, opened 2018-08-24
Brand-Consolidation: Temporary fix for LeftRailNav leading-slash issue

ncksllvn:
Temporarily fixes an issue we were seeing with a slash being added before all links in the sidebar, making it impossible to add external links.

We had to import the compiled JSX code for now, until the LeftRailNav is published to GitHub so we can fix it in Formation instead.


jkassemi (approved):
cc @accbjt for awareness

PR #8252, opened 2018-08-25
Bc external link fix

accbjt:
re-order links for left rail nav and added va links to list.


ncksllvn (approved):
LGTM! 

@minafarzad Can we double check the left sidebar requirements for Cost of Care? It looks like Cost of Care is an external link, but it also appears to have child links of its own, but since Cost of Care is external those links are never rendered as a sidebar.

PR #8255, opened 2018-08-26
Implement feature flag and build command for brand-consolidation

ncksllvn:
## Description
We need a method of building the website as normal (Vets.gov), and another method of building the brand-consolidated site, VA.gov. This PR provides a basic method of doing so by passing a feature flag from the command line into Webpack and Metalsmith. This provides the first step in building a bridge between the `master` and `brand-consolidation` branches, and ultimately towards merging.

Work includes:
- Integrated the build command `npm run watch:brand-consolidation` for starting Webpack and Metalsmith with feature flags for the Brand Consolidation (VA.gov) build
    - A command flag, `--build-consolidation-enabled`, is passed to `build.js`. This flag is processed by Metalsmith into metadata, a boolean called `BRAND_CONSOLIDATION_ENABLED`. Webpack processes this into `process.env.BRAND_CONSLIDATION_ENABLED`. 
- Added helper function `platform/brand-consolidation/feature-flag.js` to discourage using the global environment variable in application code.

At this time, the feature flag is not used at all, and would have no effect whether true or false. I also don't think Metalsmith will ever need it, since the brand-consolidated Metalsmith code (on the `brand-consolidation` branch) is split into a separate directory, but I added it anyway just in case.

## Testing done
I manually tested the feature flag both in Webpack code and Metalsmith templates repeatedly. I also wrote a unit test for `platform/brand-consolidation/feature-flag.js` to confirm the environment variable. I don't think I can write an E2E test, however, because that involves starting the site, and the feature flag will always be false in that case.

## Screenshots
I made a couple updates in a Metalsmith page and the React navbar to confirm.  These changes were not checked in 🙂 

### npm run watch
image

image

The feature flag demonstrates that it's being properly generated during builds.


script/build.js
```diff
@@ -37,7 +37,7 @@ const smith = Metalsmith(__dirname); // eslint-disable-line new-cap
 
 const optionDefinitions = [
   { name: 'buildtype', type: String, defaultValue: 'development' },
-  { name: 'mergedbuild', type: Boolean, defaultValue: false },
+  { name: 'brand-consolidation-enabled', type: Boolean, defaultValue: false },
```
ncksllvn: I change the command variable to `brand-consolidation-enabled` for consistency with the naming conventions throughout the site, and hyphenated it to keep it explicit as the top-level command arg.

jkassemi (approved):

src/platform/brand-consolidation/feature-flag.js
```diff
@@ -0,0 +1,8 @@
+/**
+ * Returns whether the current Webpack build is executing as the brand-consolidation of Vets.gov/VA.gov.
+ * @returns {boolean}
+ * @module platform/brand-consolidation/feature-flag
+ */
+export default function isBrandConsolidationEnabled() {
+  return process.env.BRAND_CONSOLIDATION_ENABLED !== undefined;
```
jbalboni: It might be safer to force the variable to be true or false in the Webpack define plugin config, so that you don't have to do `!== undefined` in the application code and can treat it like a normal boolean.

src/platform/brand-consolidation/tests/feature-flag.jsx
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
```
jbalboni: I think you need to suffix this file `unit.spec.jsx`.

jbalboni (approved):
Couple minor comments, but LGTM.

ncksllvn:
@jbalboni I would rather it be a boolean, but it's confusing - a Node `process.env` can only contain string values, but the one created in the DefinePlugin in Webpack is just using the same name and isn't really a `process.env`, so it technically can. But this make it difficult to mock in Mocha, which does have the Node `process.env`, so I can't write a consistent unit test using the boolean value. 

More info here - https://webpack.js.org/plugins/environment-plugin/

Would love to use that `DotenvPlugin` sometime.

wyattwalter (approved):
👍

ncksllvn:
Some additional info for this -

Instead of using `process.env.BRAND_CONSOLIDATION_ENABLED` in Webpack, we're writing the value as a boolean to `window.settings.brandConsolidationEnabled`. The reason for this is that eventually, the content will be separate from the application code, and will build separately. The same application code could then extend for several build cycles of the content (which will build each time a content file is merged to master), so using the window.settings, we can easily pass build information from the content to the application code.

PR #8256, opened 2018-08-26
Brand-Consolidation: Fix "About VA Health Benefits" breadcrumbs

ncksllvn:


va-gov/pages/health-care/about-va-health-benefits.md
```diff
@@ -2,7 +2,6 @@
 layout: page-breadcrumbs.html
 template: detail-page
 title: About VA Health Benefits
-display_title: VA Health Care Coverage
```
ncksllvn: Omitting this metadata property will cause the breadcrumbs to default to the `title` property instead.

ncksllvn:
@bethpotts 

image

bethpotts (approved):

PR #8260, opened 2018-08-27
add medallia script to the footer includes [WIP DO NOT MERGE]

accbjt:
test for medallia.


content/includes/footer.html
```diff
@@ -145,6 +145,7 @@ <h3>Get help from Veterans Crisis Line</h3>
 })();
 </script>
 {% endif %}
+<script type="text/javascript" src="https://resources.digital-cloud-gov.voice.medallia.com/wdcgov/2/onsite/embed.js" async></script>
```
ncksllvn: @wyattwalter @patrickvinograd It looks like we'll need to whitelist this domain if it isn't already. This script has to be served from that domain.

ncksllvn (dismissed):
Added a condition to disable in production.

ncksllvn (approved):

PR #8264, opened 2018-08-27
[WIP] Search Beta

ncksllvn:


src/applications/search/manifest.json
```diff
@@ -0,0 +1,8 @@
+{
+  "appName": "Search",
+  "entryFile": "./search-entry.jsx",
+  "entryName": "search",
+  "rootUrl": "/search",
+  "production": true,
+  "contentPage": "search/index.md"
```
ncksllvn: @hpjaj @kfrz Forgot about this - we'll need to flip it to false to keep that page from being deployed to prod after being merged to master. It wouldn't hurt anything since it's a hidden page, but as a best practice.

PR #8267, opened 2018-08-27
Send the same school data regardless of school input method

erikphansen:
## Description
Change school search component so that when it is used the same data as sent to the backend as when the user manually enters the school data. [There is a corresponding change to the JSON schema will](https://github.com/department-of-veterans-affairs/vets-json-schema/pull/250).

## Testing done
Tested locally to confirm that the request body matches regardless of the school input method.

## Testing Plan
Manual testing

## Screenshots
N/A

#### Applies to all PRs

- [ ] Appropriate logging
- [ ] Provide which alerts would indicate a problem with this functionality (if applicable)

### DIFF
--- a/src/applications/edu-benefits/components/SchoolSelectField.jsx
+++ b/src/applications/edu-benefits/components/SchoolSelectField.jsx
@@ -46,7 +46,7 @@ export class SchoolSelectField extends React.Component {
 
   componentDidMount() {
     // hydrate search if restoring from SiP
-    // if there is a seach term stored in the form data
+    // if there is a search term stored in the form data
     // if the search term in the form data isn't already in the redux state and displayed
     const institutionSelected = this.props.formData['view:institutionSelected'];
     const searchTermToRestore = this.props.formData['view:institutionQuery'];
@@ -82,18 +82,19 @@ export class SchoolSelectField extends React.Component {
     });
   }
 
-  handleOptionClick = ({ address1, address2, address3, city, facilityCode, name, state }) => {
+  handleOptionClick = ({ address1, address2, address3, city, facilityCode, name, state, zip, country }) => {
     this.props.selectInstitution({ address1, address2, address3, city, facilityCode, name, state });
     this.props.onChange({
       ...this.props.formData,
-      facilityCode,
-      'view:institutionSelected': {
-        address1,
-        address2,
-        address3,
+      name,
+      'view:facilityCode': facilityCode,
+      address: {
+        country,
+        street: address1,
+        street2: address2,
         city,
-        name,
-        state
+        state,
+        postalCode: zip,
       }
     });
   }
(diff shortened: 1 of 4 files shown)


src/applications/edu-benefits/tests/components/SchoolSelectField.unit.spec.jsx
```diff
@@ -310,17 +312,26 @@ describe('<SchoolSelectField>', () => {
...
+      },
+      name: 'testName',
+      'view:facilityCode': 'test'
+    });
+    expect(selectInstitution.firstCall.args[0]).to.eql({
+      address1: 'testAddress1',
+      address2: 'testAddress2',
+      address3: 'testAddress3',
+      city: 'testcity',
+      facilityCode: 'test',
```
cehsu: Should this also be `view:facilityCode`?

src/applications/edu-benefits/components/SchoolSelectField.jsx
```diff
@@ -271,7 +272,7 @@ export class SchoolSelectField extends React.Component {
               {`${searchResultsCount} results for ${institutionQuery}`}
             </span>}
             {showSearchResults && showInstitutions && <div>
-              {institutions.map(({ address1, address2, address3, city, country, facilityCode, name, state }, index) => (
+              {institutions.map(({ address1, address2, address3, city, country, facilityCode, name, state, zip }, index) => (
```
cehsu: The BE schema expects "USA" to be "United States". Should we perform that transformation in the reducer on the frontend?

cehsu (approved):
LGTM. I had a few questions/comments for your consideration. Good idea to store the undisplayed data, I forgot we could do that with custom fields.

src/applications/edu-benefits/components/SchoolSelectField.jsx
erikphansen: Thanks for catching that. How did you catch that, actually? Yeah, I'll see what I can do to make that transform happen on the front end.

src/applications/edu-benefits/tests/components/SchoolSelectField.unit.spec.jsx
erikphansen: nope. `selectInstitution()` is passed the data exactly as it comes from the backend the populates the school search options.

src/applications/edu-benefits/components/SchoolSelectField.jsx
erikphansen: It looks like sending "USA" instead of "United States" isn't the only potential issue. The search tool returns schools from other countries and those countries come back in ALL CAPS while the backend expects Title Case. Is that an issue we also need to address? Perhaps @jholton could weigh in since this is a question of what the backend can handle.

src/applications/edu-benefits/tests/components/SchoolSelectField.unit.spec.jsx
```diff
@@ -290,7 +290,7 @@ describe('<SchoolSelectField>', () => {
       name: 'testName',
       state: 'testState',
       zip: '12345',
-      country: 'United States'
+      country: 'USA'
```
erikphansen: Changing this mock data to reflect what we'd actually get back from the BE that powers school search and confirm that the country will be converted to 'United States'.

PR #8270, opened 2018-08-27
Bc fix for broken styles

accbjt:
This is a fix for the broken styles. This happened because I published a version of formation that was specific to brand-consolidation which has not been merged to master yet. So when I did the MegaMenu update on formation it didn't have the left rail nav styles. I copy and pasted all of the styles that I need from the formation library. This should fix the issues.

### DIFF
--- a/src/platform/site-wide/sass/merger.scss
+++ b/src/platform/site-wide/sass/merger.scss
@@ -913,6 +913,106 @@
       z-index: 2;
     }
   }
+
+  .va-sidebarnav {
+    h4 {
+      display: inline-block;
+      width: 75%;
+    }
+
+    .left-side-nav-title {
+      padding-bottom: 35px;
+    }
+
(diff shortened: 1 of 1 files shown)


src/platform/site-wide/sass/merger.scss
```diff
@@ -913,6 +913,106 @@
       z-index: 2;
     }
   }
+
+  .va-sidebarnav {
+    h4 {
```
ncksllvn: It sounds like this will be removed later on, since the styles were copied and pasted for the time being. We should add a comment explaining that.

ncksllvn (approved):

src/platform/site-wide/sass/merger.scss
accbjt: Yes this will get removed once we update formation. I'll make sure to cross reference both to make sure we don't have repeat code.

PR #8271, opened 2018-08-28
fix for MegaMenu to hide the Sign in area and fix for left rail nav c…

accbjt:
This is a fix for the bug that the MegaMenu should hide the global utilities. I also fixed the weirdness you get when clicking on the close icon on the mobile left rail nav. When you try to click it the first time is moves then you have to click it again to close out the left rail nav. For some reason it changes when the element is on focus and it changes the `position: absolute` to `position: relative`


dhall (approved):
LGTM

ncksllvn (approved):

PR #8274, opened 2018-08-28
Build Settings - Integrate build-created settings into Redux; add brand-consolidation containers

ncksllvn:
## Description
This is a continuation of department-of-veterans-affairs/vets-website/pull/8255, which adds a feature flag for brand-consolidation - when the site is built with `npm run watch:brand-consolidation`, then the websites sees that as `window.settings.brandConsolidationEnabled: true`, and that same property is `false` if running the ordinary watch command. This is applied by `js/settings.js`.

This PR does the following -

1. Adds `window.settings` into the Redux store under `state.buildSettings`, so that build-settings can follow the same data patterns as the rest of the application code.
2. Adds two containers that consume `state.buildSettings`:
    - `brand-consolidation/containers/BrandConsolidation.jsx`: The renders its children only during a brand-consolidation build using the feature flags described above.
    - `brand-consolidation/containers/VetsDotGov.jsx`: This renders its children only during a regular website build.
3. Adds unit tests for the containers described above.

## Testing done
Unit tests, manual in-browser testing

### npm run watch
Demonstrates the build-settings, and that `brandConsolidationEnabled` is set to `false`.
image

Demonstrates that `brandConsolidationEnabled` is set to `true`.
image

A pattern is established for providing properties to application code post-build, and demonstrates this by consuming the brand-consolidation feature flag.


jbalboni (approved):
This LGTM, but I don't have a good sense of how those containers are going to be used, so it's hard to evaluate them thoroughly.

ncksllvn:
Thank you @jbalboni. They're ultimately going to be used as part of the integration between `master` and `brand-consolidation`. I'm trying to keep it as non-hackish and explicit as possible, because I'm anticipating that the UX of applications will change for VA.gov, while we have to ensure that Vets.gov remains unaffected.

PR #8279, opened 2018-08-29
Brand-Consolidation - Implement a route component to handle page lookups between Vets.gov and VA.gov

ncksllvn:
## Description
This PR continues the work for reconciling the `brand-consolidation` branch with `master`, based on the work from department-of-veterans-affairs/vets-website/pull/8274.

The major update of this PR is that it implements a link component that will update the HREF of a child element if we're viewing the brand-consolidation website and the Vets.gov page located at the HREF's path has moved to another location.

For example - On Vets.gov, to find an overview of applying for healthcare, you navigate to `health-care/apply/`. On the brand-consolidation VA.gov site, this page lives at `health-care/how-to-apply`. Currently on the `brand-consolidation` branch, that link is a hardcoded change. However, we can reconcile this with `master` by using this new component like -

```jsx
import ResolvedLink from '../../platform/brand-consolidation/containers/ResolvedLink';

const hybridLink = (
  <ResolvedLink>
    <a className="usa-button" href="/health-care/apply/">Apply for Health-Care</a>
  </ResolvedLink>
);
```

When the build is started with `npm run watch:brand-consolidation`, that HREF value will be updated to `/health-care/how-to-apply`. It does this by looking up the vets.gov link in a static JSON file and seeing that the page location has been updated. This data was obtained from a spreadsheet and will increase as we move closer to the launch.

Other changes:
- Changes the shape of the build-consolidation flag from `state.buildSettings.buildConsolidationEnabled` to `state.buildSettings.buildConsolidation.enabled`
- Implements a `TOGGLE_BRAND_CONSOLIDATION_ENABLED` action, which when dispatched via Redux dev tools, will toggle the `state.buildSettings.brandConsolidation.enabled` boolean, so you can see how your app will render on Vets.gov vs. VA.gov.

## Testing Plan
Unit tests

### npm run watch
Here you can see that `state.buildSettings.brandConsolidation.enabled` is `false`, and the Healthcare Apply link renders the correct Vets.gov path -
image

Now you can see that `state.buildSettings.brandConsolidation.enabled` is set to `true`, and the Healthcare link was adjusted to the VA.gov path -
image

Creates a method of reconciling content-location changes between Vets.gov and VA.gov.

#### Unique to this PR
Not consumed in the website at all at this point, and has no visible effect. I'm just doing these changes incrementally to avoid a giant PR.


wyattwalter (commented):
As someone very weak in front-end dev skills, this seems like a great approach to me. I'm curious about thoughts on the `redirects.json` (left a comment there).

jbalboni:
@ncksllvn Do we have a lot of these links from React pages? Is there another PR coming for handling this on static pages, or am I misunderstanding how this works?

ncksllvn:
Great question @jbalboni. There are a lot of links to content from within my first [pr](https://github.com/department-of-veterans-affairs/vets-website/pull/8226/files) to [brand-consolidation] to update the healthcare links on the `brand-consolidation` branch, before realizing we need a more sophisticated approach if that branch will ever make it to master. 

So a PR hopefully coming later today will implement this link component, and I'll then merge it both ways - into master and brand-consolidation - to hopefully reconcile the src directories in each branch.

There are still other problems to be solved, but I want to do this incrementally.

ncksllvn:
@jbalboni Forgot to answer one of your questions - A PR won't be coming that updates the content directories, because the VA.gov content is a separate directory on the `brand-consolidation` branch, so those changes can be safely hardcoded.

src/platform/brand-consolidation/constants/redirects.json
```diff
@@ -0,0 +1,126 @@
+[
```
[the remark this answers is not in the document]
ncksllvn: I thought for awhile about that, and at this point because we have to wrap link tags in the React component designed here, adding redirects will have to be coordinated between a content editor and the engineer. I do like the idea of having it regenerated at build-time so if front-matter adds some redirect data then we can see those changes appear in the diff, but then we also have to keep in mind that the content will eventually be separate from the app code, so then we'd have to do a handoff from the content to the app code through the build-settings. But even then, an engineer will need to implement the redirects by wrapping the updated links.

Really complicated, and I tried to keep it simple to avoid any of the unknowns that may come out of a bigger approach.

src/platform/brand-consolidation/containers/VetsDotGov.jsx
```diff
@@ -10,7 +10,7 @@ function VetsDotGov({ vetsDotGovEnabled, children }) {
 
 export function mapStateToProps(state) {
   return {
-    vetsDotGovEnabled: !state.buildSettings.brandConsolidationEnabled
+    vetsDotGovEnabled: !state.buildSettings.brandConsolidation.enabled
```
jbalboni: Might be worth making a standalone selector for getting this enabled flag, since you're checking it in a couple places.

src/platform/brand-consolidation/actions/dev.js
```diff
@@ -0,0 +1 @@
+export const TOGGLE_BRAND_CONSOLIDATION_ENABLED = 'TOGGLE_BRAND_CONSOLIDATION_ENABLED';
```
jbalboni: Why name this file `dev.js`?

jbalboni (approved):
I put some very minor comments, but LGTM.

ncksllvn:
There was merit to this approach, but we're going to go a different route involving frontmatter and Metalsmith middleware that I'll opening a PR for later today. Thank you @jbalboni and @wyattwalter for your feedback!

PR #8284, opened 2018-08-30
526 confirmation poll take 2

cvalarida:
## Description
This is the same as https://github.com/department-of-veterans-affairs/vets-website/pull/8113, but I changed the `apiRequest` helper to not conditionally chain--so it's like it is in master. Turns out that was the cause of a bunch of failed tests after it was merged into master. I still have no idea how it was able to make it into master originally.

Aside from that, I fixed up a `ConfirmationPage` test and added a few more.

Because of the revert commit in master, I had to do some kind of fancy footwork with git:
1) Merge master in (including the revert commit) and take all the changes
2) Make a revert commit of the revert commit
3) Cherry pick my previous fixes

It _should_ be okay, but please keep an eye out for any buffoonery that could have resulted from that.


src/applications/disability-benefits/526EZ/components/ConfirmationPoll.jsx
```diff
@@ -0,0 +1,119 @@
...
+      },
+      (response) => {
+        // Don't process the request once it comes back if the component is no longer mounted
+        if (!this.__isMounted) {
+          return;
+        }
+
+        this.setState({
+          submissionStatus: submissionStatuses.apiFailure,
+          // NOTE: I don't know that it'll always take this shape.
```
jbalboni: It won't take this shape if you get a client error, like being unable to connect.

jbalboni (approved):

PR #8287, opened 2018-08-30
Flow cv

cvalarida:
## Description
The idea behind this PR is to get a feel for how difficult it is to add flow to the project. I did the following:
1) Added flow to the build pipeline
    - All it took was adding the preset npm package and using it in the babel config--it was a breeze! :man_shrugging:
2) Enabled it for a few files and added some types
    - I definitely had to look up how to do some things, but it really wasn't bad
    - I've never used a JS typing system, so this was new, though I'd seen the syntax before
3) Used `flow-typed` to add the definition for moment since that was throwing an error
    - I had to add `flow-bin` to run `flow-typed` for some reason
4) Added a custom module definition for `us-forms-system`
    - This actually isn't awesome since it looks like we'll have to make a new definition for each component we import from `us-forms-system` as well as the design system

## Testing done
I used [`flow-for-emacs`](https://flow.org/en/docs/editors/emacs/) to run flow on the project as I was working. I wasn't able to get automagic errors inline, which was kind of a bummer, but that's just an editor thing. (The flow scene for emacs seems to be pretty sparse. :disappointed:) It's basically just like running `flow` from inside `vets-website/src/` (where I put the `.flowconfig`).

## The build
I didn't enable it on all that many files, but it didn't seem to slow the build down noticeably. Running `time flow` on my machine currently takes about 0.1s.

Flow errors currently don't stop the build, so if we enable it as-is, we don't have to fix all the problems we run into to benefit from it.

## Object property checking
Okay, so my favorite part is definitely the property checking. This lets us know when we do something like `constantObject.mispeledPropertie`.


src/applications/disability-benefits/all-claims/constants.js
```diff
@@ -78,6 +79,8 @@ export const PCIU_STATES = [
   { label: 'Wyoming', value: 'WY' }
 ];
 
+// Not sure what the deal is here, but I'm getting the following from flow for the next two lines:
+// Error: Missing type annotation for `U`.
```
cvalarida: Any thoughts? I removed the `U` from `PCIU` as well as all the label and values inside the array. No dice. I even got rid of the `_` just in case something funny was happening there too. Still the same error. It's really quite baffling. :confused:

src/applications/disability-benefits/all-claims/utils.jsx
```diff
@@ -1,18 +1,34 @@
...
+// @flow
+import * as React from 'react';
 import moment from 'moment';
 
 import get from '../../../platform/utilities/data/get';
 
 import { RESERVE_GUARD_TYPES } from './constants';
+
+
+// Would be good to pull these type definitions into a common file
```
cvalarida: Not necessarily in `src/custom-types/`, but an app-wide file or something, I'm thinking.

src/applications/disability-benefits/all-claims/utils.jsx
```diff
@@ -1,18 +1,34 @@
...
+};
+
+
 /**
  * Show one thing, have a screen reader say another.
- * NOTE: This will cause React to get angry if used in a <p> because the DOM is "invalid."
  *
- * @param {ReactElement|ReactComponent|String} srIgnored -- Thing to be displayed visually,
- *                                                           but ignored by screen readers
- * @param {String} substitutionText -- Text for screen readers to say instead of srIgnored
```
cvalarida: Maybe we don't want to get rid of the JSDoc type? :man_shrugging:

src/applications/disability-benefits/all-claims/utils.jsx
```diff
@@ -69,9 +85,9 @@ export const ReservesGuardDescription = ({ formData }) => {
   );
 };
 
-export const title10DatesRequired = (formData) => get('view:isTitle10Activated', formData, false);
+export const title10DatesRequired = (formData: FormDataType): boolean => get('view:isTitle10Activated', formData, false);
 
-export const isInFuture = (errors, fieldData) => {
+export const isInFuture = (errors: { addError: (string) => mixed }, fieldData: string): void => {
```
cvalarida: So this is an interesting thing. I was confused about how to define a function type for a bit, but after I read the documentation on it, it makes a ton of sense.

src/flow-typed/npm/moment_v2.3.x.js
```diff
@@ -0,0 +1,331 @@
+// flow-typed signature: 23b805356f90ad9384dd88489654e380
+// flow-typed version: e9374c5fe9/moment_v2.3.x/flow_>=v0.25.x
```
cvalarida: I got this by running `flow-typed install moment`. It looked at our dependency version and used the one that fit.

yarn.lock
```diff
@@ -276,6 +276,13 @@ ansi-wrap@0.1.0:
   version "0.1.0"
   resolved "https://registry.yarnpkg.com/ansi-wrap/-/ansi-wrap-0.1.0.tgz#a82250ddb0015e9a27ca82e82ea603bbfa45efaf"
 
+anymatch@^1.3.0:
```
cvalarida: Not super keen on all of these additions. :unamused:

But what're y'gonna do? :man_shrugging:

src/applications/disability-benefits/all-claims/constants.js
cvalarida: Looks like it's a thing: https://github.com/facebook/flow/issues/6151

PR #8289, opened 2018-08-31
Create "resources" middleware and assign resource_ids to content pages

ncksllvn:
## Description
This PR creates a data structure that can be used to lookup a page or asset location by a manually-created `resource_id` property. This data is accessible at `window.settings.resources`. It has the follow properties:

- `window.settings.resources.pages` - Generated by processing the `resource_id` frontmatter in each Markdown page. 
- `window.settings.resources[anything_else]` - Extended from an as-is config at `assets/resources.json` that contains static asset paths.

Ultimately, a page path can be looked up by `resource_id` in Metalsmith templates by `resources.pages.My_Resource_ID`. Example:

### Content Frontmatter Example
`/health-care/apply.md`
```
title: How to Apply for Health Care Benefits
resource_id: HealthCare_Apply
```

### Metalsmith Example
```html
<a href="{{ resources.pages.HealthCare_Apply }}">Apply for health care</a>
```

### JSX Example
There will be a helper method or component for abstracting the actual data structure, but if you wanted raw access, that would look like -
```jsx
<a href={window.resources.pages.HealthCare_Apply}>Apply for health care</a>
```

### Why this was warranted
This was developed as part of the work behind Brand Consolidation. Currently, the work there lives on a separate branch, and we need to reconcile its `src/` directory with master's if we ever want to merge the two. Since the `src/` directory contains links to content pages, many of which were moved as part of the brand-consolidation, we need a method of looking up page locations. The work here provides us with a framework for doing that.

## Screenshots
It's beautiful -
image

#### Unique to this PR
There is no visible impact here. It just creates a data structure without consuming it anywhere. That will come in a later PR.

Original approach - department-of-veterans-affairs/vets-website/pull/8279
Resolves department-of-veterans-affairs/vets.gov-team/issues/12804


script/create-resource-config.js
```diff
@@ -0,0 +1,40 @@
...
+    };
+
+    metalsmith.metadata({ resources });
+    options.resources = resources;
+
+    done();
+  };
+}
+
+module.exports = createResourceConfig;
```
ncksllvn: This is the actual Metalsmith middleware for processing the `resource_ids`.

script/build.js
```diff
@@ -71,6 +72,8 @@ if (isHerokuBuild) {
   applyHerokuOptions(options);
 }
 
+smith.use(createResourceConfig(options));
+
```
ncksllvn: I added the middleware registration here so that it's processed into the `metadata` in time for the pages to use if they choose to.

content/pages/health-care/about-va-health-care.md
```diff
@@ -7,6 +7,7 @@ concurrence: complete
 lastupdate: 2017-06-28
 collection: healthCare
 children: healthCareCoverage
+resource_id: HealthCare_About
```
ncksllvn: I user Proper_Case because I thought that would sell this best to the content editors if they choose to use this later on down the road.

PR #8290, opened 2018-08-31
All claims 2.2

jbalboni:
## Description
**Note**: This PR is giant only because of the conditions list.

This adds the questions for 2.2, which are new disabilities that you are claiming. All these fields appear to be new and don't fit the existing rated disabilities array, so I'm not sure how they map to the EVSS API. They might be all new, since they wouldn't have made any sense for an increase only form.

Also, this only has a simplified autosuggest approach, where we use the standard list of conditions and descriptions. I believe we want to do some kind of conceptual matching, but that requires a backend end point. We can do that in a follow up PR when the backend is ready.

## Testing done
Local testing, unit tests

## Screenshots
screen shot 2018-09-07 at 10 03 57 am

screen shot 2018-09-07 at 10 04 07 am

screen shot 2018-09-07 at 10 04 15 am

screen shot 2018-08-31 at 3 54 13 pm

screen shot 2018-09-07 at 2 56 38 pm

screen shot 2018-09-07 at 2 56 29 pm

#### Unique to this PR
- [x] You can add new conditions
- [x] You can answer follow up questions about those conditions
- [x] When you add a secondary disability you get to pick from the other disabilities in your list.

#### Applies to all PRs

- [x] Appropriate logging


src/applications/disability-benefits/all-claims/pages/newDisabilityFollowUp.js
```diff
@@ -0,0 +1,97 @@
...
+          }
+        }
+      },
+      primaryDisability: {
+        'ui:title': 'Which disability caused the disability you’re claiming here?',
+        'ui:required': (formData, index) => formData.newDisabilities[index].cause === 'SECONDARY',
+        'ui:options': {
+          expandUnder: 'cause',
+          expandUnderCondition: 'SECONDARY',
+          updateSchema: createSelector(
```
jbalboni: This is building the enum list based on the rated disabilities and the new ones that have been added (aside from the current one), based on Alex's guidance.

src/applications/disability-benefits/all-claims/content/disabilityLabels.js
```diff
@@ -0,0 +1,837 @@
+export default {
```
cvalarida: Is this list temporary? Will it go away once we've got the back end set up?

src/applications/disability-benefits/all-claims/content/disabilityLabels.js
```diff
@@ -0,0 +1,837 @@
...
+  1500: 'cerebral hemorrhage',
+  1510: 'cerebral vascular accident',
+  1520: 'cerebrovascular accident',
+  1530: 'cerebrovascular hemorrhage',
+  1540: 'cervical sprain',
+  1550: 'cervical strain',
+  1560: 'cervicitis',
+  1570: 'CHF',
+  1580: 'chest pain',
+  1590: 'Chloracne or other acneform disease consistent w/ chloracne',
```
cvalarida: Should that first `Chloracne` be capitalized? Also, the `w/` abbreviation is inconsistent, I think.

Not sure how much it matters, but while we're here... :man_shrugging:

src/applications/disability-benefits/all-claims/content/disabilityLabels.js
```diff
@@ -0,0 +1,837 @@
...
+  2065: 'diabetes, adult-onset',
+  2085: 'diabetes mellitus, type 2',
+  3035: 'Gulf War veteran with diagnosed illness as qualifying chronic disability',
+  530: 'ASHD',
+  8917: 'Miscellaneous',
+  8918: 'hip replacement',
+  8919: 'knee replacement',
+  109: '2010 Presumptive Herbicide Condition',
+  111: '2010 Presumptive Herbicide Condition',
+  8920: 'Adhesions - Digestive',
```
cvalarida: From here down, the conditions are capitalized, which is inconsistent with the above. If the list is temporary, we can ignore it, but if it's ever going to be in production, we probably want to make them consistent.

src/applications/disability-benefits/all-claims/pages/newDisabilityFollowUp.js
```diff
@@ -0,0 +1,97 @@
...
+export const uiSchema = {
+  newDisabilities: {
+    items: {
+      'ui:title': disabilityNameTitle,
+      cause: {
+        'ui:title': '',
+        'ui:widget': 'radio',
+        'ui:options': {
+          labels: {
+            NEW: 'My disability was caused by—or got worse because of—an injury or exposure during my service in the military.',
```
cvalarida: This is probably just my misunderstanding, but `NEW` is also for existing? I.e. not new?

cvalarida (approved):

src/applications/disability-benefits/all-claims/pages/newDisabilityFollowUp.js
```diff
@@ -0,0 +1,97 @@
...
+            formData => formData.ratedDisabilities,
+            formData => formData.newDisabilities,
+            (formData, primarySchema, primaryUISchema, index) => index,
+            (disabilities = [], newDisabilities, currentIndex) => {
+              const newDisabilitiesWithoutCurrent = newDisabilities.filter((item, index) => index !== currentIndex);
+              return {
+                'enum': disabilities.map(disability => disability.diagnosticCode)
+                  .concat(newDisabilitiesWithoutCurrent.map(disability => disability.diagnosticCode)),
+                enumNames: disabilities.map(disability => disability.name)
+                  .concat(newDisabilitiesWithoutCurrent.map(disability => getNewDisabilityName(disability.diagnosticCode))),
```
markgreenburg: I think this gets a little wonky if you try to add a second disability, because:
1. You click 'add another disability'
2. `updateSchema` fires
3. `newDisabilities` is set to `[{ diagnosticCode: xxx }]` in the selector
4. `newDisabilitiesWithoutCurrent` removes the new disability, so now we're back to `[]`
5. We do a bunch of `map` and `concat` against this empty array
6. `getNewDisabilityName` gets called with `undefined`
7. `getNewDisabilityName` calls `capitalizeEachWord` with `disabilityLabels[undefined]`
8. `capitalizeEachWord` tries to do `undefined.split`
9. explosions

Probably needs an extra check somewhere along the way. 🤷‍♂️

src/applications/disability-benefits/all-claims/pages/newDisabilityFollowUp.js
markgreenburg: Is the expectation that there will always be a `ratedDisability` in this situation?

src/applications/disability-benefits/all-claims/content/disabilityLabels.js
jbalboni: I think this list will persist even after the backend is set up, but I'm assuming it will need some copy edits.

src/applications/disability-benefits/all-claims/pages/newDisabilityFollowUp.js
jbalboni: I pulled this from some other status list that looked close to what we needed for this question. All of the disabilities are new in here in that the user is entering them for the first time. These questions don't show up for rated disabilities.

peggygannon (changes requested):
Hi, @jbalboni 
It looks like we're missing this first screen (below). (I made some slight adjustments just now to what was in the ticket)
1. 
**New disabilities**
Now we’ll ask you about your new service-connected disabilities or conditions.
[ Back | Continue ] { buttons }
Save and finish this application later.

2. please make this header bold:
**Add a new disability**

3. please add a period after this sentence for all applicable screens: 
Please tell us what caused your $disabilityname.

4. please update this question to:

Please choose the disability that caused the disability you’re claiming here.

5. please add a period after this sentence:
Please describe the VA mistreatment that caused your disability.

also, please note, I'm working on editing/adding laymen terms for the conditions list. that is being tracked in a separate issue.

src/applications/disability-benefits/all-claims/pages/newDisabilityFollowUp.js
jbalboni: Yeah, I think hiding it when there are no possible options is a good choice. My understanding is that this list should be all the rated disabilities, plus any new disabilities other than the current one.

src/applications/disability-benefits/all-claims/content/disabilityLabels.js
cvalarida: :thinking: I think this could get tricky if we have to maintain two separate sets of data. Especially when we throw in the conceptual matches as well.

I talked with Alastair the other day and we were thinking when we get back a conceptual match, the api would return something like "Post Traumatic Stress Disorder (**PTSD**)" if the query text is "PTSD", the condition name is "Post Traumatic Stress Disorder", and "PTSD" is a conceptual match. Meaning the api would return something like `${conditionName} (${conceptualMatch})`.

How would this work if we ran the result through this list and displayed what was in here?

src/applications/disability-benefits/all-claims/pages/newDisabilityFollowUp.js
jbalboni: This should be working better now.

jbalboni:
@peggygannon Updates made (and I added the screenshot of the first page that I forgot to upload originally).

markgreenburg (approved):
LGTM 🚀

markgreenburg:
@jbalboni one other thing actually - it's still possible to run into `undefined.split()` when:
1. Add a new disability
2. Hit button to add another new disability
3. Don't fill the disability out
4. Hit 'continue'

I think all this would need to resolve is to make the disability name a required field.

jbalboni:
@markgreenburg Ah, yeah, that's a good point. But it means we need a Yes/No question in front of the new disabilities array field. I'll work on that and ping Alex.

jbalboni:
@peggygannon Screenshots are updated. I'm a little confused on the follow up question content, not sure if I used the write source for that.

peggygannon:
@jbalboni Hi, Jeff. Below is the latest content update, based on the new outline for Follow-up questions. Let me know if you have any questions.  

#### 2 of 3 Follow up questions

| Follow up questions |
| ------------------ |
| Loop through the listed new conditions to collect the relevant details. Sort PTSD to last position in loop |


## X of Y List Disabilities

#### Disability details

{ for each *new* disability }

{ if not PTSD }

What caused this service-connected disability?  { radio }

- My disability was caused by—or got worse because of—an injury or exposure during my service in the military.
- My disability was caused by another disability (for example, I have a limp that caused lower-back problems).
- My disability was caused by VA mistreatment.

{ if secondary (opt 2) }

Please choose the secondary disability that caused the disability you’re claiming here. { select }

{ options include all disabilities but this one : append to condition name above }

{ endif secondary }

{ if directly connected (opt 1) }

Please briefly describe the {event|exposure} that caused your condition. { textarea }

{ endif }

Date this {event|exposure} happened (This date doesn't have to be exact.) { date }

{ endfor }

{ elseif }

If you're filing a claim for PTSD, you'll need to provide additional documents and details about the traumatic (or “stressor”) event that caused your PTSD. You may also need to provide treatment records and contact details for any relevant supporting statements.

If you need more time to prepare your evidence for your PTSD claim, you can save your application and return to it later when you have your evidence ready to upload.

{ endif not PTSD}

{ endif yes new disabilities }

[ Back | Continue ] { buttons }
[Save and finish this application later.]()

peggygannon:
@jbalboni let's update the event/exposure questions to the below: 

{ if directly connected (opt 1) }
Please briefly describe the injury or event that caused your disability. { textarea }

{ endif }

Date this injury or event happened (This date doesn't have to be exact.) { date }

jbalboni:
@peggygannon Thanks. There's also another issue, with "Please choose the secondary disability that caused the disability you’re claiming here." The user isn't choosing a secondary disability there. They're claiming a secondary disability and that field is to choose the disability that caused it (which may or not be another secondary disability).

peggygannon:
@jbalboni Ok, let's update that question to: 


 {if secondary (opt 2) }
Please choose the disability that caused the disability you’re claiming here. { select }

{ options include all disabilities but this one : append to condition name above }

jbalboni:
@peggygannon Updated.

peggygannon (approved):

peggygannon:
@jbalboni OK, thanks. Copy updates look good. I assume we're intentionally leaving out the PTSD copy, as that's being handled by BAH? If so, I'll make a note in the outline.

PR #8292, opened 2018-09-01
Fixed incorrect object path resulting in incorrect last updated date for facility services

U-DON:
Facility Locator uses the `X-Key-Inflection: camel` header in its requests, so the actual key is `lastUpdated`, not `last_updated`. The latter resulted in passing `undefined` to `moment`, which would default to the current date.


ncksllvn (approved):

dhall (approved):
LGTM -- great catch!

saneshark (approved):

PR #8293, opened 2018-09-03
Add Metalsmith middleware for processing "aliases" property into redirects

ncksllvn:
## Description
This PR adds Metalsmith middleware into the build so that when an "aliases" property exists in a page it is processed into redirect pages. Example -

**/some-path/index.md**
```
title: Some Path
aliases:
  - /some-old-path/
  - /another-old-path
```

During the build, `/some-old-path/index.html` and `/another-old-path/index.html` will be generated as an HTML page containing a meta property that will briefly tell the user that the page has moved before automatically navigating to `/some-page/`. 

This is valuable as part of the `brand-consolidation` work. Many pages are being moved around, and in the event we miss a link, this redirect will act as a fallback for the time being. This will also allow us to leave the `src/` directories untouched, because we can allow the old links to remain until we're further along.

#### Unique to this PR
There is no impact to master on this PR, as no pages contain an "aliases" property.

### DIFF
--- a/script/create-redirects.js
+++ b/script/create-redirects.js
@@ -1,0 +1,50 @@
+/* eslint-disable no-param-reassign, no-continue, no-console */
+const path = require('path');
+
+function getPagePath(fileName, fileData) {
+  const { permalink } = fileData;
+  return permalink ? permalink.replace('index.html', '') : `/${fileName.replace('.md', '')}/`;
+}
+
+function getRedirectPage(redirectToPath) {
+  return `
+    <!doctype html>
+    <head>
+      <meta http-equiv=refresh content="1; url=${redirectToPath}">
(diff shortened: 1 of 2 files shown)


script/create-redirects.js
```diff
@@ -0,0 +1,50 @@
+/* eslint-disable no-param-reassign, no-continue, no-console */
+const path = require('path');
+
+function getPagePath(fileName, fileData) {
+  const { permalink } = fileData;
+  return permalink ? permalink.replace('index.html', '') : `/${fileName.replace('.md', '')}/`;
```
jkassemi: Is the final path not reliably available in metalsmith at this point?

jkassemi (approved):

PR #8296, opened 2018-09-04
[WIP] header/footer proxy embed

jkassemi:
## Description

For https://github.com/department-of-veterans-affairs/vets.gov-team/issues/12783

Generates a script that can be embedded in existing VA.gov pages designed to remove non-compatible design elements and replace them with the latest from the brand consolidation effort. 

This is intended to replace the `<-- brand-consolidation-X -->` comment currently being placed into several sites on TeamSite. As of this time, I was unable to find any proxied resources that contained the comment. 

I was not yet able to configure several react components, which are commented out here.

Styles still appear to conflict badly despite the namespacing. The other option was an iframe, but we have content in menus that pops over any borders we may specify. I'm not personally aware of any options around this. My gut says we address the conflicts as one offs until this works for varying subsets of the TeamSite properties. Plan is to then move toward a better, more correct and maintainable integration.

## Testing done

Using a local vagrant box with the vetsgov-revproxy application, I disabled gzip encoding and enabled a substitution for the </head> element, placing a prefixed sass stylesheet and a local reference to the proxy rewrite "application":

    <style type="text/css">.brand-consolidation-deprecated { display: none !important; } </style> <link rel="stylesheet" href="http://localhost:3001/generated/styleConsolidated.css" /> <script type="text/javascript" src="http://localhost:3001/generated/vendor.entry.js"></script> <script type="text/javascript" src="http://localhost:3001/generated/proxy-rewrite.entry.js"></script> 

I removed SSL configuration and S3 file serving, and mapped http://dev-preview-kassemi.va.gov:8080/ on my localhost to this configuration. This provided a proxy to the VA.gov site at that URL. I was then able to make requests to http://dev-preview-kassemi.va.gov:8080/health/topics/index.asp, which would load the assets referenced here.

## Testing Plan

- [ ] Integration into preview development site / dev.va.gov
- [ ] Cross browser testing
- [ ] Performance testing / improvements
- [ ] Alternative approach research

## Screenshots

screen shot 2018-09-03 at 10 26 29 pm


src/platform/startup/react.js
```diff
@@ -39,8 +39,13 @@ export default function startReactApp(component, root = null) {
     mountElement = document.getElementById('react-root');
   }
   // eslint-disable-next-line scanjs-rules/call_addEventListener
-  document.addEventListener('DOMContentLoaded', () => {
+
+  if (document.readyState !== 'loading') {
```
jkassemi: at this point proxied requests have already triggered the DOMContentLoaded event, and we've added new elements to the page. We're doing a naive check here to see if we can just render the component immediately.

src/applications/proxy-rewrite/partials/header.js
```diff
@@ -0,0 +1,80 @@
+export default `
```
jkassemi: @ncksllvn the partial you provided in https://github.com/department-of-veterans-affairs/vets-website/pull/8295/ wasn't from the va gov content directory. Since those are using react I figured we'd give that integration a shot first and address performance issues as they pop up.

jkassemi: Also not quite sure how we can best get those into the static pages... this seemed like a good first pass. Just need to validate the approach (or start thinking _really_ hard about other options)...

src/applications/proxy-rewrite/manifest.json
```diff
@@ -0,0 +1,7 @@
+{
```
jkassemi: Wasn't sure if this was an `application`, but it needs to be consumable by an external party, could go through versions, has a name, etc...

If this works well enough, we're going to want to pass it off to teamsite properties that we don't have  a proxy in front of for manual integration.

ncksllvn: Yeah, it almost seems like something that belongs in its own repo, but we aren't there yet.

If it matters at all, since in this case we aren't worried about wiring up to content pages or routing, I think we only need to define the `entryFile` and `appName`

.gitignore
```diff
@@ -26,3 +26,5 @@ package-lock.json
 # Ignore user's editor/IDE prefs
 .vscode/
 .idea/
+.swp
+.swo
```
ncksllvn: It looks like a couple of `.swp` files snuck into a commit at some point

src/platform/startup/react.js
ncksllvn: I wonder if having a second optional parameter for each startup function (startMegaMenuWidget, etc.) to override `startReactApp` would makes things any easier, or just more work -

```jsx
export default function startSomeComponent(store, callback = startReactApp) {
    callback((
      <Provider store={store}>
        <Announcement/>
      </Provider>
    ), elementId);
}
```

ncksllvn (approved):
I know it's still a WIP, but LGTM for laying down the groundwork 💥

bshyong:
Below screenshot of local implementation with the following snippet inserted in `<head>`

```
<!-- JK --><style type="text/css">.brand-consolidation-deprecated { display: none !important; } </style>
<link rel="stylesheet" href="http://localhost:3001/generated/styleConsolidated.css" />
<link rel="stylesheet" href="http://localhost:3001/generated/static-pages.css">
<script type="text/javascript" src="http://localhost:3001/generated/vendor.entry.js"></script>
<script type="text/javascript" src="http://localhost:3001/generated/proxy-rewrite.entry.js"></script> <!-- /JK -->
```

screencapture-localhost-8000-va-health-html-2018-09-25-08_22_37

bshyong:
@ncksllvn I'm going to merge this branch into master; should have no impact to existing applications. Introduces an entry point for injection scripts

ncksllvn:
Sounds good @bshyong! I think it's at a great point to iterate on.

Also, sorry about messing up your commit log when we changed the base branch to `master` lol

PR #8302, opened 2018-09-04
Reconcile scripts directory with brand-consolidation

ncksllvn:
## Description
The work for the brand-consolidation of VA.gov is currently on its own branch, `brand-consolidation`, with a modified build script for building the FE project. There was a feature flag implementation there, but it wasn't always used when it should have been, and the build itself need evolution to make it more clear how the build changes between different environments.

This PR moves all code that modifies the `options` variable in `build.js` into its own file, `scripts/options.js`. This includes -

- Function for gathering command line flags
- Function for Heroku overrides
- Function for environment overrides
- Function for brand-consolidation overrides

The goal here is that after the `options` object is established, the rest of the build can become more straightforward and we don't need to check certain build flags (primarily the brand-consolidation build flag) before each step.

The flip-side of this - where the `brand-consolidation` branch is receiving the same changes - is here, department-of-veterans-affairs/vets-website/pull/8303.

## Testing done
I've run builds on my own machine for both `npm run watch` and `npm run watch:brand-consolidation`.

## Testing Plan
Open to suggestions if there's a great way to ensure each environment builds exactly how we expect it to.

Resolves department-of-veterans-affairs/vets.gov-team/issues/12827

### DIFF
--- a/script/build.js
+++ b/script/build.js
@@ -263,6 +263,7 @@ if (!BUILD_OPTIONS.watch && !(process.env.CHECK_BROKEN_LINKS === 'no')) {
         '/education/apply-for-education-benefits/application',
         '/pension/application/527EZ',
         '/burials-and-memorials/application/530',
+        '/health-care/apply/application',
         '/health-care/apply-for-health-care-form-10-10ez',
         '/veteran-id-card/apply',
         '/veteran-id-card/how-to-get',
--- a/script/build.js
+++ b/script/build.js
@@ -73,8 +73,6 @@ if (BUILD_OPTIONS.buildtype === 'production') {
   });
   ignoreList.push('veteran-id-card/how-to-get.md');
   ignoreList.push('veteran-id-card/how-to-upload-photo.md');
-  ignoreList.push('education/complaint-tool.md');
-  ignoreList.push('disability-benefits/apply/form-526-disability-claim.md');
 }
 smith.use(ignore(ignoreList));
 
--- a/script/build.js
+++ b/script/build.js
@@ -5,7 +5,6 @@ const archive = require('metalsmith-archive');
 const assets = require('metalsmith-assets');
 const blc = require('metalsmith-broken-link-checker');
 const collections = require('metalsmith-collections');
-const commandLineArgs = require('command-line-args');
 const dateInFilename = require('metalsmith-date-in-filename');
 const define = require('metalsmith-define');
 const filenames = require('metalsmith-filenames');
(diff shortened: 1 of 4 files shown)


U-DON (approved):
LGTM. Separating the options into a new module looks much cleaner too.

U-DON (approved):

script/build.js
```diff
@@ -31,113 +30,51 @@ const {
...
 
 const ignoreList = [];
-if (options.buildtype === 'production') {
+if (BUILD_OPTIONS.buildtype === 'production') {
   manifests.filter(m => !m.production).forEach(m => {
     ignoreList.push(m.contentPage);
   });
   ignoreList.push('veteran-id-card/how-to-get.md');
   ignoreList.push('veteran-id-card/how-to-upload-photo.md');
+  ignoreList.push('education/complaint-tool.md');
```
jkassemi: should this be added to the ignoreList in this PR?

script/build.js
```diff
@@ -31,113 +30,51 @@ const {
...
+  ignoreList.push('disability-benefits/apply/form-526-disability-claim.md');
```
jkassemi: also curious for an update here if it's relevant. I was under (a potentially outdated) impression that this was meant to go live behind a password, just hidden from navigation.

script/collections/brand-consolidation.json
```diff
@@ -0,0 +1,314 @@
...
+    }
+  },
+  "vreServiceDisabled": {
+    "pattern": "employment/vocational-rehab-and-employment/service-disabled/*md",
+    "sortBy": "order",
+    "metadata": {
+      "name": "Servicemember & Veteran Programs"
+    }
+  }
+}
```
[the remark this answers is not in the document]
jkassemi: 💯

jkassemi (approved):
Definitely more readable. 

I'm guessing that the timestamp-based cache busting in the filename is probably the most major contributor to nondeterministic builds in this case. Should probably reevaluate that at some point, but this seems pretty straightforward.

script/build.js
ncksllvn: Ah thank! Removing both of these from the ignore-list. Normally that list hardly ever changes, so I'm surprised there was any difference between them.

ncksllvn:
Would someone mind also approving the brand-consolidation side of things, department-of-veterans-affairs/vets-website/pull/8306?

jbalboni (approved):
This looks good to me, and seems like an improvement regardless of the brand consolidation work.

wyattwalter (approved):
A much-needed improvement. Looks great to me!

script/build.js
```diff
@@ -638,14 +261,14 @@ if (!options.watch && !(process.env.CHECK_BROKEN_LINKS === 'no')) {
         '/education/apply-for-education-benefits/application',
         '/pension/application/527EZ',
         '/burials-and-memorials/application/530',
-        '/health-care/apply/application',
+        '/health-care/apply-for-health-care-form-10-10ez',
```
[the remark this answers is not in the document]
ncksllvn: Ah we do! Good catch, appreciate it 🙂Adding it back!

PR #8314, opened 2018-09-05
Typescript cv

cvalarida:
## Description
Building off the babel-7-upgrade-cv branch, I gave implementing TypeScript a whirl. Setting it up wasn't _nearly_ as easy as it was with Flow, but since we only need to do this once, that's not a huge problem.

More important is the ease of use. There were a few things that bugged me as I went through the process:

1. When changing the file extension, it results in an error for the current watch, so a watch restart is required to pick up the new `.ts` file instead of the old `.js` one.
1. My editor is telling me that the import from `'./constants'` doesn't exist because of the implicit `.js` extension (I renamed the file to `.ts`), but I don't get a compile error. If I change it to `'../constants'`, it _does_ result in a compile error, so the build works fine. I just need to figure out how to set up my editor for that kind of shenanigans.
    - Worst case scenario, we disable the [import/no-unresolved] rule...But let's...not do that
1. Trying to get `srSubstitute`'s first param, `srIgnored` to be a generic thing that React can generate, analogous to Flow's React.Node. I'm not finding anything...
    - Mark found https://github.com/sw-yx/react-typescript-cheatsheet#useful-react-type-examples which has a reference to React.ReactNode, but...where's the documentation??

---
I definitely spent more time fiddling with TypeScript or hunting down how to do a thing than I did with Flow. (Should I use an `interface` or `type`? How do I say the type is anything React can render? Why is `tslint` telling me I have a shadowed variable name when I can't see another declaration of it in scope?)


package.json
```diff
@@ -11,9 +11,11 @@
     "heroku-serve": "http-server",
     "heroku-postbuild": "npm run build --  --entry static-pages,style",
     "lint": "npm run lint:js && npm run lint:sass",
-    "lint:changed": "npm run lint:js:changed && npm run lint:sass",
+    "lint:changed": "npm run lint:js:changed && npm run lint:ts:changed && npm run lint:sass",
     "lint:js": "eslint --quiet --ext .js --ext .jsx .",
     "lint:js:changed": "LIST=`git diff-index --name-only HEAD | grep \"\\.js[x]\\{0,1\\}$\"`; if [ \"$LIST\" ]; then eslint --quiet $LIST; fi",
+    "lint:ts": "tslint -c tslint.json -p tsconfig.json",
+    "lint:ts:changed": "LIST=`git diff-index --name-only HEAD | grep \"\\.ts[x]\\{0,1\\}$\"`; if [ \"$LIST\" ]; then tslint -c tslint.json $LIST; fi",
```
cvalarida: Having a separate file type means we need to run two linters, which maybe isn't great, but doesn't seem so bad. My biggest concern here is the potential differences between our eslint config and our tslint config.

package.json
```diff
@@ -200,9 +205,11 @@
     "semver": "^5.3.0",
     "sinon": "^3.2.1",
     "skin-deep": "^1.0.0",
+    "source-map-loader": "^0.2.4",
```
cvalarida: I didn't actually verify that sourcemaps still work.

src/applications/disability-benefits/all-claims/utils.tsx
```diff
@@ -4,15 +4,29 @@ import moment from 'moment';
...
  * NOTE: This will cause React to get angry if used in a <p> because the DOM is "invalid."
  *
- * @param {ReactElement|ReactComponent|String} srIgnored -- Thing to be displayed visually,
- *                                                           but ignored by screen readers
- * @param {String} substitutionText -- Text for screen readers to say instead of srIgnored
+ * @param srIgnored -- Thing to be displayed visually, but ignored by screen readers
+ * @param substitutionText -- Text for screen readers to say instead of srIgnored
  */
-const srSubstitute = (srIgnored, substitutionText) => {
+const srSubstitute = (srIgnored: React.ReactNode, substitutionText: string): React.ReactNode => {
```
cvalarida: If anybody can find documentation for the React types, that'd be awesome. Like I said in the PR description, Mark found a thing that happened to mention this type. I looked in `node_modules/@types/react/index.d.ts` and found it.

tsconfig.json
```diff
@@ -0,0 +1,13 @@
+{
+  "compilerOptions": {
+    "target": "es5",
+    "module": "commonjs",
+    "lib": ["es6"],
+    "sourceMap": true,
+    "outDir": "./dist/",
```
cvalarida: I _think_ this is right? :man_shrugging:

tslint.json
```diff
@@ -0,0 +1,20 @@
+{
+  "defaultSeverity": "error",
+  "extends": [
+    "tslint:recommended"
+  ],
+  "jsRules": {},
+  "rules": {
+    "quotemark": [true, "single", "jsx-double", "avoid-template"],
```
cvalarida: I put most of these rules in here to minimize the changes to the files I converted to TS.

cvalarida:
Hrm... :thinking:

I tried to break the compilation by giving a bad return type for a function (`boolan`), but `yarn watch` said it compiled successfully. Running `tslint` also didn't catch it. :grimacing:

cvalarida:
This is pretty out-of-date by now. We may end up circling back around to this at some point, but it doesn't sound like adding types is super helpful right now.

PR #8315, opened 2018-09-05
Fixed login functionality in user nav and consolidated render logic with use of feature flag

U-DON:
The separate va-gov-user-nav seemed to be a little out of date with user-nav. Now that we have the JS feature flag, we can account for the brand consolidated content differences more granularly, so that's what I did here.

From what I've tested locally, I've also gotten login to work again in the dev-preview build.


src/platform/site-wide/user-nav/components/HelpMenu.jsx
```diff
@@ -3,23 +3,37 @@ import React from 'react';
 
 import DropDownPanel from '@department-of-veterans-affairs/formation/DropDownPanel';
 import IconHelp from '@department-of-veterans-affairs/formation/IconHelp';
+import isBrandConsolidationEnabled from '../../../brand-consolidation/feature-flag';
 
 class HelpMenu extends React.Component {
   render() {
+    const buttonText = isBrandConsolidationEnabled ? 'Contact Us' : 'Help';
```
ncksllvn: Ah maybe this is misleading, but `isBrandConsolidationEnabled` is actually a function. If you feel that it would make more sense as a boolean though I'm onboard with that change.

ncksllvn (approved):
LGTM! 💯 💥 🎸

PR #8316, opened 2018-09-05
Brand consolidation feature flag for relevant user nav components

U-DON:
This is essentially the same set of changes as #8315 but for the `master` branch.


ncksllvn:
Ah, looks like we need to update the unit tests to backfill the `window.settings` data

```
TypeError: Cannot read property 'brandConsolidationEnabled' of undefined
```

ncksllvn (approved):
LGTM!

PR #8327, opened 2018-09-06
Change handling of school addresses from the search tool

erikphansen:
## Description
- [x] Update package.json to use the new [version of JSON schema](https://github.com/department-of-veterans-affairs/vets-json-schema/pull/254) when it's ready
- ~Combine the second and third address fields so no data is lost; this
is especially important for some international addresses that the search
tool returns as the many of those contain important address info in the
`address3` field.~ We have instead moved to adding a `street3` field to the address object we send to the backend.
- If the address from the search tool is domestic, transform the country
to the prop enum value, "United States".
- ~If the address is not domestic, send the country as-is and add the
`viaSearchTool=true` field on the `address` so the backend will use the
school address schema that accepts any string for the country value.~ We are instead simply adding a less stringent version of the international address schema that allows for any string for the country, rather than requiring the country to be selected from the defined list.

## Testing done
Updated unit tests to test handling of domestic and international addresses from the search tool.

## Testing Plan
The format of addresses that come from the search tool vary widely. This calls for manual testing, especially of international addresses, to assure that the correct address data gets passed to the backend.

## Screenshots
N/A

#### Unique to this PR
- [x] Use the stricter schema for domestic addresses, use the more forgiving address schema for international addresses.

#### Applies to all PRs

- [ ] Appropriate logging
- [ ] Provide which alerts would indicate a problem with this functionality (if applicable)

### DIFF
--- a/src/applications/edu-benefits/feedback-tool/config/form.js
+++ b/src/applications/edu-benefits/feedback-tool/config/form.js
@@ -37,8 +37,8 @@ const {
 
 const { assistance, programs, school } = educationDetails.properties;
 const { address: schoolAddress, name: schoolName } = school.properties;
-const domesticSchoolAddress = schoolAddress.oneOf[0];
-const internationalSchoolAddress = schoolAddress.oneOf[1];
+const domesticSchoolAddress = schoolAddress.anyOf[0];
+const internationalSchoolAddress = schoolAddress.anyOf[1];
 const countries = domesticSchoolAddress.properties.country.enum.concat(internationalSchoolAddress.properties.country.enum); // TODO access via default definition
 
 function configureSchoolAddressSchema(schema) {
--- a/src/applications/edu-benefits/feedback-tool/config/form.js
+++ b/src/applications/edu-benefits/feedback-tool/config/form.js
@@ -35,7 +35,7 @@ const {
   serviceDateRange,
 } = fullSchema.properties;
 
-const { assistance, programs, school } = educationDetails;
+const { assistance, programs, school } = educationDetails.properties;
 const { address: schoolAddress, name: schoolName } = school.properties;
 const domesticSchoolAddress = schoolAddress.oneOf[0];
 const internationalSchoolAddress = schoolAddress.oneOf[1];
@@ -382,6 +382,9 @@ const formConfig = {
                     },
                     street2: {
                       'ui:title': 'Address line 2'
+                    },
+                    street3: {
+                      'ui:title': 'Address line 3'
                     },
                     city: {
                       'ui:title': 'City',
--- a/src/applications/edu-benefits/feedback-tool/helpers.js
+++ b/src/applications/edu-benefits/feedback-tool/helpers.js
@@ -193,46 +193,22 @@ export function transformSearchToolAddress({ address1, address2, address3, city,
   if (isDomesticAddress) {
     return {
       country: 'United States',
-      street: typeof address1 === 'string'
-        ? address1.slice(0, domesticSchoolAddressFields.street.maxLength)
-        : address1,
-      street2: typeof address2 === 'string'
-        ? address2.slice(0, domesticSchoolAddressFields.street2.maxLength)
-        : address2,
-      street3: typeof address3 === 'string'
-        ? address3.slice(0, domesticSchoolAddressFields.street3.maxLength)
-        : address3,
-      city: typeof city === 'string'
-        ? city.slice(0, domesticSchoolAddressFields.city.maxLength)
-        : city,
-      state: typeof state === 'string'
-        ? state.slice(0, domesticSchoolAddressFields.state.maxLength)
-        : state,
-      postalCode: typeof zip === 'string'
-        ? zip.slice(0, domesticSchoolAddressFields.postalCode.maxLength)
-        : zip,
+      street: address1 && address1.slice(0, domesticSchoolAddressFields.street.maxLength),
+      street2: address2 && address2.slice(0, domesticSchoolAddressFields.street2.maxLength),
+      street3: address3 && address3.slice(0, domesticSchoolAddressFields.street3.maxLength),
+      city: city && city.slice(0, domesticSchoolAddressFields.city.maxLength),
+      state: state && state.slice(0, domesticSchoolAddressFields.state.maxLength),
+      postalCode: zip && zip.slice(0, domesticSchoolAddressFields.postalCode.maxLength),
     };
   }
   return {
     country,
-    street: typeof address1 === 'string'
-      ? address1.slice(0, searchToolSchoolAddressFields.street.maxLength)
-      : address1,
-    street2: typeof address2 === 'string'
-      ? address2.slice(0, searchToolSchoolAddressFields.street2.maxLength)
-      : address2,
-    street3: typeof address3 === 'string'
-      ? address3.slice(0, searchToolSchoolAddressFields.street3.maxLength)
-      : address3,
-    city: typeof city === 'string'
-      ? city.slice(0, searchToolSchoolAddressFields.city.maxLength)
-      : city,
-    state: typeof state === 'string'
-      ? state.slice(0, searchToolSchoolAddressFields.state.maxLength)
-      : state,
-    postalCode: typeof zip === 'string'
-      ? zip.slice(0, searchToolSchoolAddressFields.postalCode.maxLength)
-      : zip,
+    street: address1 && address1.slice(0, searchToolSchoolAddressFields.street.maxLength),
+    street2: address2 && address2.slice(0, searchToolSchoolAddressFields.street2.maxLength),
+    street3: address3 && address3.slice(0, searchToolSchoolAddressFields.street3.maxLength),
+    city: city && city.slice(0, searchToolSchoolAddressFields.city.maxLength),
+    state: state && state.slice(0, searchToolSchoolAddressFields.state.maxLength),
+    postalCode: zip && zip.slice(0, searchToolSchoolAddressFields.postalCode.maxLength),
   };
 
 }
--- a/src/applications/edu-benefits/feedback-tool/helpers.js
+++ b/src/applications/edu-benefits/feedback-tool/helpers.js
@@ -14,11 +14,11 @@ import UserInteractionRecorder from '../components/UserInteractionRecorder';
 
 const { get } = dataUtils;
 const domesticSchoolAddressFields = get(
-  'properties.educationDetails.properties.school.properties.address.oneOf[0].properties',
+  'properties.educationDetails.properties.school.properties.address.anyOf[0].properties',
   fullSchema
 );
 const searchToolSchoolAddressFields = get(
-  'properties.educationDetails.properties.school.properties.address.oneOf[2].properties',
+  'properties.educationDetails.properties.school.properties.address.anyOf[2].properties',
   fullSchema
 );
 
@@ -214,7 +214,6 @@ export function transformSearchToolAddress({ address1, address2, address3, city,
     };
   }
   return {
-    viaSearchTool: true,
     country,
     street: typeof address1 === 'string'
       ? address1.slice(0, searchToolSchoolAddressFields.street.maxLength)
--- a/src/applications/edu-benefits/feedback-tool/helpers.js
+++ b/src/applications/edu-benefits/feedback-tool/helpers.js
@@ -26,7 +26,7 @@ export function fetchInstitutions({ institutionQuery, page, onDone, onError }) {
 }
 
 export function transform(formConfig, form) {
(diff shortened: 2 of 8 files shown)


src/applications/edu-benefits/components/SchoolSelectField.jsx
```diff
@@ -84,21 +84,27 @@ export class SchoolSelectField extends React.Component {
...
       'view:facilityCode': facilityCode,
-      address: {
-        country: backendFriendlyCountries[country] || country,
-        street: address1,
-        street2: address2,
-        city,
-        state,
-        postalCode: zip,
-      }
+      address
```
rianfowler: Did we move away from submitting the facility code for these addresses?

erikphansen: Yes. Lihan requested that the first day both you and Claire were gone. Since the search tool provided all of the required address info to the frontend it seemed doable.

src/applications/edu-benefits/tests/components/SchoolSelectField.unit.spec.jsx
```diff
@@ -291,46 +291,90 @@ describe('<SchoolSelectField>', () => {
...
+    const internationalInstitution = {
       address1: 'testAddress1',
       address2: 'testAddress2',
       address3: 'testAddress3',
       city: 'testcity',
       facilityCode: 'test',
       name: 'testName',
-      state: 'testState'
+      state: 'testState',
+      zip: '12345',
```
cehsu: Do international institution addresses use the zip and state fields?

src/applications/edu-benefits/components/SchoolSelectField.jsx
```diff
@@ -84,21 +84,27 @@ export class SchoolSelectField extends React.Component {
 
   handleOptionClick = ({ address1, address2, address3, city, facilityCode, name, state, zip, country }) => {
     this.props.selectInstitution({ address1, address2, address3, city, facilityCode, name, state });
-    const backendFriendlyCountries = {
-      USA: 'United States'
+    const isDomesticAddress = country === 'USA';
+    let street2 = address2;
+    if (address3) {
+      street2 = `${street2} ${address3}`;
```
cehsu: We might want to have the backend handle this as they're already performing other schema related translations.

src/applications/edu-benefits/components/SchoolSelectField.jsx
rianfowler: We really shouldn't be doing these kinds of data transformations in the client. If we can't submit exactly what we get back from the API then something should change on the server or the data.

cehsu (commented):
Looks good so far. I had a few comments/question for your consideration.

src/applications/edu-benefits/components/SchoolSelectField.jsx
erikphansen: Absolutely, if we add a `street3` to the schema.

src/applications/edu-benefits/tests/components/SchoolSelectField.unit.spec.jsx
erikphansen: I'm not sure and given that the international address formats we get back from the search tool are a little weird I don't think we can count on them always missing them. @rianfowler likely knows more about the different address formats we get from the search tool.

cehsu: is nothing sacred?

src/applications/edu-benefits/components/SchoolSelectField.jsx
cehsu: Fascinating point @rianfowler, looping @jholton for any additional thoughts from the backend perspective on this idea for organizing logic.

src/applications/edu-benefits/components/SchoolSelectField.jsx
```diff
@@ -84,21 +84,24 @@ export class SchoolSelectField extends React.Component {
 
   handleOptionClick = ({ address1, address2, address3, city, facilityCode, name, state, zip, country }) => {
     this.props.selectInstitution({ address1, address2, address3, city, facilityCode, name, state });
-    const backendFriendlyCountries = {
-      USA: 'United States'
+    const isDomesticAddress = country === 'USA';
+    const address = {
```
rianfowler: Maybe put this transformation into a helper.

src/applications/edu-benefits/components/SchoolSelectField.jsx
```diff
@@ -84,21 +84,24 @@ export class SchoolSelectField extends React.Component {
 
   handleOptionClick = ({ address1, address2, address3, city, facilityCode, name, state, zip, country }) => {
```
rianfowler: Alphabetize this argument list.

src/applications/edu-benefits/components/SchoolSelectField.jsx
erikphansen: Nice. Will do.

src/applications/edu-benefits/components/SchoolSelectField.jsx
erikphansen: Ha, should I admit I was looking at the order of this list and thought it wasn't a logical way to order address info? Alphabetical didn't even enter my mind.

rianfowler (approved):

src/applications/edu-benefits/feedback-tool/helpers.js
```diff
@@ -26,7 +26,7 @@ export function fetchInstitutions({ institutionQuery, page, onDone, onError }) {
 }
 
 export function transform(formConfig, form) {
-  const formData = transformForSubmit(formConfig, form);
+  const formData = transformForSubmit(formConfig, form, null);
```
erikphansen: This will make the `transformForSubmit()` method no longer use the very aggression JSON stringify transformer that's removing international addresses that lack a zip/postal code (which included all int'l addresses added via the school search tool).

erikphansen: the transformForSubmit method is defined [here](https://github.com/usds/us-forms-system/blob/master/lib/js/helpers.js#L382)
the stringify transformer I'm avoiding using is defined [here](https://github.com/usds/us-forms-system/blob/master/lib/js/helpers.js#L336)

src/applications/edu-benefits/feedback-tool/config/form.js
```diff
@@ -383,6 +383,9 @@ const formConfig = {
                     street2: {
                       'ui:title': 'Address line 2'
                     },
+                    street3: {
```
cehsu: We may need to transform this on the FE until the backend has included this transformation in its transformation list.

cehsu (approved):
LGTM. IIRC salesforce doesn't support street3, so we may need to temporarily combine street2 and street3 on the frontend (as you previously had set up), until the backend includes this transformation in its list.

src/applications/edu-benefits/feedback-tool/config/form.js
cehsu: Nevermind, it looks like this work is already done. Way to go @jholton.
https://github.com/department-of-veterans-affairs/vets-api/blob/f367cbde99b9fc812d6bd29bab06b0b312a69ea2/app/models/gi_bill_feedback.rb

erikphansen: @cehsu I only see that transform happening [here](https://github.com/department-of-veterans-affairs/vets-api/blob/f367cbde99b9fc812d6bd29bab06b0b312a69ea2/app/models/gi_bill_feedback.rb#L34), which is no longer gonna run.

erikphansen: Since the backend needs to be updated before we can release this I don't think we should bandaid it on the frontend.

src/applications/edu-benefits/feedback-tool/config/form.js
```diff
@@ -35,10 +35,10 @@ const {
   serviceDateRange,
 } = fullSchema.properties;
 
-const { assistance, programs, school } = educationDetails;
+const { assistance, programs, school } = educationDetails.properties;
 const { address: schoolAddress, name: schoolName } = school.properties;
-const domesticSchoolAddress = schoolAddress.oneOf[0];
-const internationalSchoolAddress = schoolAddress.oneOf[1];
+const domesticSchoolAddress = schoolAddress.anyOf[0];
```
cehsu: Very well done 👏 , great catch using anyOf instead of oneOf.

src/applications/edu-benefits/feedback-tool/config/form.js
cehsu: Should we create a backend ticket to update the international school address transformation so that street3 is included when condensing international addresses onto only the street1 and street2 fields?

src/applications/edu-benefits/feedback-tool/helpers.js
cehsu: Would it make sense to create an issue on the us-forms-system repo to break the cleaning logic out of the stringify transformer?

src/applications/edu-benefits/feedback-tool/helpers.js
```diff
@@ -167,3 +178,61 @@ const applicantRelationshipEventMap = {
...
+ * A helper that takes data from the SchoolSelectField back end and transforms
+ * it to a valid format as specified by the FEEDBACK-TOOL's
+ * educationDetails.school.address schema.
+ *
+ * @param {*} _ - Object with address fields
+ * @returns {Object} An Object that passes the FEEDBACK-TOOL's educationDetails.
+ * school.address schema.
+ */
+export function transformSearchToolAddress({ address1, address2, address3, city, country, state, zip }) {
+  const isDomesticAddress = country === 'USA';
```
cehsu: It seems inconsistent to use an abbreviation for USA only in the school address country enum list, would it make sense to create a ticket to convert this enum to the full name? If so, we could use the country value below on line 195 rather than a magic string.

src/applications/edu-benefits/feedback-tool/helpers.js
```diff
@@ -167,3 +178,61 @@ const applicantRelationshipEventMap = {
...
+  if (isDomesticAddress) {
+    return {
+      country: 'United States',
+      street: typeof address1 === 'string'
```
cehsu: Do we anticipate that these might not be strings? In what situation would that be the case?

cehsu (approved):
LGTM. I had a few questions and comments for your consideration.

src/applications/edu-benefits/feedback-tool/helpers.js
erikphansen: yeah, I talked to Anne about this late yesterday and she agrees that the problematic JSON stringify transformer code is ooooold and the transform logic should be refactored. We're not sure the best approach, but we should document this.

src/applications/edu-benefits/feedback-tool/helpers.js
erikphansen: you might be misreading this (or I'm misunderstanding)? the search tool returns 'USA' and we're transforming it to 'United States'

src/applications/edu-benefits/feedback-tool/helpers.js
erikphansen: no. but they might be null or maybe undefined. and we don't want to `slice()` on something we can't slice.

src/applications/edu-benefits/feedback-tool/helpers.js
cehsu: Thanks for clarifying, that makes sense.

src/applications/edu-benefits/feedback-tool/helpers.js
cehsu: That makes sense, if it exists it will be a string, so we can simply could do something like:
```
street: address1 && address.slice(....)
```

But what you've got works great.

erikphansen: Ha, yours is totally better. Thanks for catching this!

src/applications/edu-benefits/feedback-tool/config/form.js
erikphansen: ticket: https://app.zenhub.com/workspace/o/department-of-veterans-affairs/vets.gov-team/issues/13074

src/applications/edu-benefits/feedback-tool/config/form.js
erikphansen: This caused so much confusion this morning. Then things still didn't work when I switch to `anyOf` because of a stupid copy/paste error I'd introduced in my code :(

PR #8333, opened 2018-09-06
Brand consolidation - Append success_relay query param into login-url API request

ncksllvn:
## Description
In order for the API to enable login on VA domains, we have to include an extra parameter in the request for login URL to indicate the callback domain - preview.va.gov, *.va.gov, and vets.gov.

We should also open a PR to master containing the same changes as contained in this PR.

Connects to department-of-veterans-affairs/vets.gov-team/issues/12757
Corresponding API info - Further context - department-of-veterans-affairs/vets-api/pull/2252


omgitsbillryan (commented):
This LGTM from the API perspective, but I'll wait for another FE dev for the actual approval.

U-DON (approved):
LGTM.

PR #8373, opened 2018-09-10
Brand Consolidation: Blind redirects for health tools

U-DON:
Resolves department-of-veterans-affairs/vets.gov-team#12758.

Tested locally with and without a signed in user.


src/applications/static-pages/createCallToActionWidget.js
```diff
@@ -0,0 +1,31 @@
...
+  '/health-care/schedule-view-va-appointments/': `${mhvDomain}/mhv-portal-web/web/myhealthevet/scheduling-a-va-appointment`,
+
+  '/health-care/view-test-and-lab-results/': `${mhvDomain}/mhv-portal-web/labs-tests`
+};
+
+export default function createCallToActionWidget() {
+  if (conditionalStorage().getItem('userToken')) {
+    const redirectUrl = urlMap[location.pathname];
+    window.open(redirectUrl, 'redirect-popup');
+  }
```
U-DON: Will move this into the CTA widget when it's created so that it can be based on the dynamic Redux state.

ncksllvn (approved):
LGTM. Will review more thoroughly after preview deploy.

PR #8375, opened 2018-09-10
526 add no evidence mg

markgreenburg:
## Description
- Adds a qualifying question to the evidence types screen to allow veterans to not submit any evidence with their 526 submission.
- Only requires an evidence type be selected if the user selected that they have evidence they would like to submit
- Skips the 'evidence summary' page when veteran indicates that they don't have any evidence on the preceding page

## Testing done
Went through all the flows locally

## Testing Plan
Test out on staging

## Screenshots
Default view when user hits this page:
screen shot 2018-09-10 at 2 29 18 pm

View if user selects 'no':
screen shot 2018-09-10 at 2 29 29 pm

#### Unique to this PR
This may change once we remove looping from the form flow.

### DIFF
--- a/src/applications/disability-benefits/526EZ/config/form.js
+++ b/src/applications/disability-benefits/526EZ/config/form.js
@@ -59,7 +59,8 @@ import {
   queryForFacilities,
   getEvidenceTypesDescription,
   veteranInfoDescription,
-  editNote
+  editNote,
+  validateBooleanIfEvidence
 } from '../helpers';
 
 import {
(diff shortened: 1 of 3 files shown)


peggygannon (approved):
@markgreenburg Hi, Mark, copy looks good, my only feedback/question is: 
--when the user chooses "No" is it possible to hide the link "which evidence type should I choose?"

kreek (approved):
LGTM

markgreenburg:
Oops, looks like I added Alastair as reviewer after he already approved; merging.

PR #8381, opened 2018-09-11
#13010: 404 for Education Benefits "apply now"  Eduwizard fixes

RobAWilkinson:
Updates EduWizard code to work with new BC urls

### DIFF
--- a/src/applications/edu-benefits/components/EducationWizard.jsx
+++ b/src/applications/edu-benefits/components/EducationWizard.jsx
@@ -2,7 +2,9 @@ import React from 'react';
 import _ from 'lodash/fp';
 import classNames from 'classnames';
 
+import isBrandConsolidationEnabled from '../../../platform/brand-consolidation/feature-flag';
 import ErrorableRadioButtons from '@department-of-veterans-affairs/formation/ErrorableRadioButtons';
+import { URLMap } from '../config/bc-urls';
 
 const levels = [
   ['newBenefit'],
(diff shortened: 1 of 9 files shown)


ncksllvn (approved):

src/applications/edu-benefits/components/EducationWizard.jsx
```diff
@@ -23,10 +25,11 @@ export default class EducationWizard extends React.Component {
   }
 
   getButton(form) {
+    const url = isBrandConsolidationEnabled() ? URLMap[form] : `/education/apply-for-education-benefits/application/${form}`;
```
jbalboni: Is the intent that the site is fully functional with brand consolidation disabled on `va-gov/qa-PI9-S3`? With the root url change elsewhere in this PR, I don't think it will, because all these forms use `rootUrl` as the base url for React Router in `startApp` and there's no switching there.

If the intent is that the `va-gov/qa-PI9-S3` branch doesn't work when brand consolidation is disabled, is this switch actually necessary?

RobAWilkinson: @jbalboni I believe the idea is to develop a strategy later on that allows for different rootUrls for BC and non-BC pages. 
You're correct, if we dont' need to worry about building vets.gov, then the ternary is not needed (although the URLmap still is).
My thought was this was one of those things that would get lost and forgotten if we ever went to build vets.gov and figured out the url problem

jbalboni: I'm fine with this either way as long as @ncksllvn is.

ncksllvn: Yeah, anything we change in `src` will be manually reconciled with master soon. W didn't have the foundation we needed to resolve changing `rootUrl`s while we were under strict time constraints to meet requirements for each PI.

PR #8382, opened 2018-09-11
Medallia/Foresee - Add Metalsmith logic for determining JavaScript snippets for survey tools

ncksllvn:
## Description
- Added the Foresee Production code 
    -  Both Foresee and Medallia are limited to Staging anyway, but I didn't want to lose the snippet.
- Added Metalsmith include for processing which survey to render based on Frontmatter `entryname` property
    - We are supposed to embed Foresee on LOA3 pages; Medallia on all others. The simplest way I could think to do that was checking the Webpack entryname.

## Testing done
Checked generated HTML on localhost to see if LOA3/non-LOA3 pages were generating Foresee or Medallia script links. 

Related ticket - department-of-veterans-affairs/vets.gov-team/issues/12914

### DIFF
--- a/content/includes/survey-tools.html
+++ b/content/includes/survey-tools.html
@@ -1,0 +1,8 @@
+{% if buildtype === 'development' %}
+  {% assign loa3_entrynames = 'burials, claims-status, letters, post-911-gib-status, hca, health-records, messaging, rx, pensions, profile-360, claims-status' %}
+  {% if loa3_entrynames contains entryname %}
+    <script type="text/javascript" src="/js/foresee/{{ buildtype }}.js"></script>
+  {% else %}
+    <script type="text/javascript" src="https://resources.digital-cloud-gov.voice.medallia.com/wdcgov/2/onsite/embed.js" async></script>
+  {% endif %}
+{% endif %}
--- a/content/includes/survey-tools.html
+++ b/content/includes/survey-tools.html
@@ -1,4 +1,4 @@
-{% if buildtype === 'staging' %}
+{% if buildtype == 'staging' %}
   {% assign auth_required_entrynames = 'claims-status, letters, health-records, messaging, rx, profile-360' | split: ', ' %}
   {% if auth_required_entrynames contains entryname  %}
     <script type="text/javascript" src="/js/foresee/{{ buildtype }}.js"></script>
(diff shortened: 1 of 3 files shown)


content/includes/survey-tools.html
```diff
@@ -0,0 +1,8 @@
+{% if buildtype === 'staging' %}
+  {% assign loa3_entrynames = 'burials, claims-status, letters, post-911-gib-status, hca, health-records, messaging, rx, pensions, profile-360, claims-status' %}
```
jkassemi: Can't assign arrays in liquid, but according to the docs, we can split a string into an assignment:

```
{% assign loa3_entrynames = "burials, claims-status, ..." | split: ", " %}
   {% if entryname in loa3_entrynames %}
    ...
```

so we don't end up with any potential issues with an loa1 entryname with a name that's a subset of an loa3 name. 

Minor note - claims-status is mentioned twice.

jkassemi (changes requested):
I like the approach. How'd you collect the list? I went through each entrypoint manually to confirm:

1. Burials: https://www.vets.gov/burials-and-memorials/application/530/introduction, form is available without LOA3
2. claims-status has two source pages, but just the one entrypoint. Not sure the best route here, I'd honestly be almost in favor of keeping the duplication since it makes it clear there's a wonky pattern there.
3. https://www.vets.gov/education/gi-bill/post-9-11/ch-33-benefit isn't LOA3 gated
4. https://www.vets.gov/health-care/apply/application/introduction isn't LOA3 gated
5. https://www.vets.gov/pension/application/527EZ/introduction isn't LOA3 gated

Not saying that they shouldn't have the foresee integration, just want to make sure this is what's required of #12914. (that ticket may need to be fleshed out).

content/includes/survey-tools.html
ncksllvn: Ah I must have missed the `in` operator! Updating now.

ncksllvn:
@jkassemi I put that list together really quickly anticipating that it would be changed a lot no matter how much time I put into it. Thanks for checking! I filtered the list down to these entry names but obviously it only takes a second to add/remove entries - `claims-status`, `letters`, `health-records`, `messaging`, `rx`, and `profile-360`.

I added the `split` filter for that list but I couldn't get the `in` operator to work. `contains` did work with the array as well though so the benefits are still there.

Thanks for reviewing! I may end up adding the new Foresee button requirements into this as well to get it all at once - department-of-veterans-affairs/vets.gov-team/issues/13047

jkassemi (approved):
This looks good as is. Iterate away. Please request an additional review if you include those foresee button requirements.

PR #8385, opened 2018-09-11
Port search.gov from vets.gov to preview.va.gov

kfrz:
## Description
We want to port over the search.gov search results page for vets.gov to preview.va.gov. This PR will add conditional logic that sends the search query to the preview.va.gov branded site. 

## Testing done
Manually tested locally that these changes work as expected. 

#### Unique to this PR

- [x] Search action sends user to search.vets.gov if on vets.gov, or search.usa.gov if on preview.va.gov
- [x] Update URL with as-of-yet undecided subdomain. 

#### Applies to all PRs

- [x] Appropriate logging

### DIFF
--- a/src/platform/site-wide/user-nav/components/SearchMenu.jsx
+++ b/src/platform/site-wide/user-nav/components/SearchMenu.jsx
@@ -12,8 +12,10 @@ class SearchMenu extends React.Component {
     super(props);
     this.makeForm = this.makeForm.bind(this);
     this.toggleSearchForm = this.toggleSearchForm.bind(this);
-    this.searchAction = isBrandConsolidationEnabled() ? 'https://search.usa.gov/search' : 'https://search.vets.gov/search';
-    this.searchAffiliate = isBrandConsolidationEnabled() ? 'va' : 'vets.gov_search';
+    this.state = {
+      searchAction: isBrandConsolidationEnabled() ? 'https://search.usa.gov/search' : 'https://search.vets.gov/search',
+      searchAffiliate: isBrandConsolidationEnabled() ? 'va' : 'vets.gov_search'
+    };
   }
 
   componentDidUpdate() {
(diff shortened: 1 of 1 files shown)


src/platform/site-wide/user-nav/components/SearchMenu.jsx
```diff
@@ -60,7 +65,7 @@ class SearchMenu extends React.Component {
         id="searchmenu"
         icon={icon}
         isOpen={this.props.isOpen}>
-        {this.makeForm()}
+        {this.makeForm(searchAction, searchAffiliate)}
```
kfrz: Is it better to inject as such from the top-most component, or is it okay to define these values within the `makeForm()` function?

rianfowler: Could put them on the component state in the constructor so this helper is not run every time the component renders. Not sure why `makeForm()` is a function and not just inlined jsx but you don't have to refactor that.

src/platform/site-wide/user-nav/components/SearchMenu.jsx
```diff
@@ -4,12 +4,16 @@ import classNames from 'classnames';
...
+import isBrandConsolidationEnabled from '../../../brand-consolidation/feature-flag';
+
 
 class SearchMenu extends React.Component {
   constructor(props) {
     super(props);
     this.makeForm = this.makeForm.bind(this);
     this.toggleSearchForm = this.toggleSearchForm.bind(this);
+    this.searchAction = isBrandConsolidationEnabled() ? 'https://search.usa.gov/search' : 'https://search.vets.gov/search';
+    this.searchAffiliate = isBrandConsolidationEnabled() ? 'va' : 'vets.gov_search';
```
rianfowler: This will work but I think it's more idiomatic to put data like this into `this.state` in the constructor rather than on the React component instance itself since it's used as part of the render update.

rianfowler (approved):

src/platform/site-wide/user-nav/components/SearchMenu.jsx
kfrz: Thanks @rianfowler  -- something like this?

```jsx
class SearchMenu extends React.Component {
  constructor(props) {
    super(props);
    this.makeForm = this.makeForm.bind(this);
    this.toggleSearchForm = this.toggleSearchForm.bind(this);
    this.state = {
      searchAction: isBrandConsolidationEnabled() ? 'https://search.usa.gov/search' : 'https://search.vets.gov/search',
      searchAffiliate: isBrandConsolidationEnabled() ? 'va' : 'vets.gov_search'
    };
  }

// ...

      <form
        acceptCharset="UTF-8"
        action={this.state.searchAction}
        id="search"
        method="get">
```

rianfowler: Exactly.

hpjaj (approved):
LGTM!

ncksllvn (approved):
LGTM 👍

PR #8395, opened 2018-09-12
Add loading state svg to vets-website

markgreenburg:
## Description
Adds the loading spinner svg to vets-website. This is the exact loading spinner found in Formation:
https://github.com/department-of-veterans-affairs/design-system/blob/85a0275275490770b8e72b17b863acf1690b59f2/public/img/loading-state.svg

I really don't know if this is the right way to fix this bug, because in my mind this file should be getting pulled in automatically from Formation, and this obviously isn't happening. So not sure what's going on there - happy to entertain ideas.

## Testing done
- Tested locally.

## Screenshots
screen shot 2018-09-12 at 3 04 52 pm

## Acceptance criteria
- [x] As a user, I want to know that the site isn't frozen when things take a while.

## Definition of done


jbalboni:
@markgreenburg This looks like the wrong spinner. There's one that says "Vets.gov" in the middle that I think is the correct one.

rianfowler:
@jbalboni Are we migrating to this svg spinner as part of the brand consolidation?

markgreenburg:
@jbalboni I know which one you're talking about, but there's only a `gif` version. My assumption was that the `LoadingIndicator` component was intentionally changed to use the new `svg` but this change hasn't made its way to vets-website yet.

jbalboni:
@ncksllvn Do you know anything about the spinner and the brand consolidation work?

ncksllvn:
As of a couple days ago, that's the same spinner currently in use on the `brand-consolidation` branch, but I think it was picked up for the same reasons as described here.

jbalboni:
Ok, I guess this is new spinner, then. It looks weird to me at that size.

jbalboni (approved):

cvalarida (approved):
:man_shrugging:

PR #8396, opened 2018-09-12
Isolate Metalsmith build from Webpack build

ncksllvn:
## Description
This works separates the Webpack and Metalsmith builds out of `script/build.js` with the objective of making the two portable enough for separate build cycles. Webpack should not need to access the Metalsmith directories at all, and Metalsmith should not need to know what Webpack is doing internally. 

Most of this was just code reorganization - Anything that had to do with Webpack was moved to `config/`. The module `config/webpack-metalsmith-connect/index.js` then exposes a build function and a watch function as Metalsmith middleware. I initially considered renaming this from "webpack" to something more generic (`config/appcode-content-connect`) so that in theory we could change our asset compiler without changing our naming conventions in the content build, but I decided to keep it explicit with Webpack in the module name.

The biggest complication was that the `manifest.json` accesses the Metalsmith content directory in order to enforce the `production` flag - if a manifest has `production` set to `false`, it in return uses the `rootUrl` to derive a `contentPage` property, which is then added to a list passed to `metalsmith-ignore`, removing the corresponding content file during production. For this change, Webpack can't reach into the Metalsmith files. 

The `production` flag is really useful though, so to fix this issue I added middleware called `create-environment-filter`. When an environment/buildtype name is added to frontmatter with the value `false`, it is removed during builds of that type. For example, this would build the file in all builds except prod:

```
---
title: All builds but prod
production: false
---
```

I should have kept the scope focused to just this, but it was hard to conceptualize this reorganization without doing some cleaning up along the way. I also made these changes -

- Moves `production: false` from applicable `manifest.json`'s into the corresponding content file
- Moves the buildtype floating-values into `script/constants/environments`
- Expands on the broken-link-checker to ignore links to pages that are React apps instead of using a list of hardcoded values
- Removes `metalsmith-define`
    - This was used during the build to generate a global Metalsmith variable `site`. This was never used, and the plugin itself is just a wrapper around the standard `smith.metadata()`.
- Removes `metalsmith-redirect`
    - This was only used with two very old links - one URL was added to the revproxy and the other didn't exist at all anymore.
- Moves Metalsmith middleware into their own files that were previously defined directly in the build

### Why this was warranted
The Webpack build has been abstracted from the Metalsmith/content, except for two endpoints for building the assets and watching during local dev. This does two things -

- Positions us closer to where `brand-consolidation` can be merged to `master`. Now that the Webpack build does not need access to the file-system, we can more safely enable dynamic `rootUrl`s as they change from Vets.gov to VA.gov.
- In the near future, we can separate the content from the assets compilation.

## Todo
- [ ] Add flag that contains a Git SHA that indicates not to execute Webpack, and instead just grab a file-manifest over the wire. 
- [ ] Allow .js extension for app manifests 

☝️ These will be done in later PRs.

## Testing done
This refactor could fortunately leverage the safety of all of our automated tests, so those passing is assurance that the build was (at least largely) unchanged. I also ran some local production builds to ensure the environment-filter was working.

## Definition of done

## More information
department-of-veterans-affairs/vets.gov-team/issues/12871


config/manifest-helpers.js
```diff
@@ -24,9 +10,6 @@ function getAppManifests(root) {
 
       manifest.filePath = file;
       manifest.entryFile = path.relative(root, path.join(path.dirname(file), manifest.entryFile));
-      if (manifest.rootUrl) {
-        manifest.contentPage = convertUrlToMarkdownPath(root, manifest.rootUrl);
-      }
```
ncksllvn: This was removed because Webpack can longer reach into Metalsmith directories.

config/webpack-metalsmith-connect/convert-paths-to-relative.js
```diff
@@ -0,0 +1,16 @@
...
+        delete files[file];
+        /* eslint-enable no-param-reassign */
+      }
+    });
+
+    done();
+  };
+}
+
+module.exports = convertAbsolutePathsToRelative;
```
ncksllvn: This was just moved from the build.js as part of the cleanup.

script/check-broken-links.js
```diff
@@ -0,0 +1,41 @@
...
+    const ignoreLinks = new RegExp(ignoreGlobs.join('|'));
+    const brokenLinkChecker = createBrokenLinkChecker({
+      allowRedirects: true,
+      warn: false,
+      allowRegex: ignoreLinks
+    });
+
+    brokenLinkChecker(files);
+    done();
+  };
```
ncksllvn: Previously we had a hardcoded array of paths to not evaluate as link-values during builds. We used this for React routers, which sometimes has child paths that otherwise don't exist in the file system so would otherwise cause an error. This was confusing with the `brand-consolidation` branch, which is renaming React applications. To account for this, this middleware generates the list by gathering all file paths for file's that contain an entry name for Webpack bundles.

script/constants/environments.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
+  DEVELOPMENT: 'development',
+  STAGING: 'staging',
+  PRODUCTION: 'production',
+  VAGOVDEV: 'vagovdev',
+  VAGOVSTAGING: 'vagovstaging',
+  PREVIEW: 'preview'
+};
```
ncksllvn: This was just part of the general cleanup. These values were just floating around otherwise.

script/create-environment-filter.js
```diff
@@ -0,0 +1,17 @@
...
+      const file = files[fileName];
+      if (file[environmentName] === false) {
+        delete files[fileName];
+      }
+    }
+    done();
+  };
+}
+
+module.exports = createEnvironmentFilter;
```
ncksllvn: The environment filter is very small. It just deletes the file from the build if it contains `false` as the build-type's property value.

script/options.js
```diff
@@ -41,32 +42,29 @@ function applyDefaultOptions(options) {
...
         }
       }
       break;
 
-    case 'vagovdev':
-    case 'vagovstaging':
-    case 'preview':
+    case environments.VAGOVDEV:
+    case environments.VAGOVSTAGING:
+    case environments.PREVIEW:
```
ncksllvn: There were no changes here besides importing the environments constants.

script/process-file-manifest.js
```diff
@@ -0,0 +1,50 @@
...
+          const regex = new RegExp(originalAssetFilename, 'g');
+          file.contents = new Buffer(contents.replace(regex, newAssetFilename));
+        });
+      }
+    });
+    done();
+  };
+}
+
+module.exports = processFileManifest;
```
ncksllvn: This was just copied from the build.js. I kept in the scripts directory instead of moving it into the config/Webpack directory because soon content builds will have to link to assets via a `file-manifest.json` downloaded over HTTPS and won't compile assets at all. Isolating it here makes it easy to add a step later on to skip compiling the assets at all, and just grab a `file-manifest.json` over the wire.

content/pages/disability-benefits/apply/dependents/form-686c-dependents.md
```diff
@@ -2,6 +2,7 @@
 title: Declaration of status of dependents
 entryname: 686-dependent-status
 layout: page-react.html
+production: false
```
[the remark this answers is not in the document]
jbalboni: I'm a little disappointed that we can't use the production flag in the manifests to turn e2e tests on and off anymore. I don't think there's a way to read that from the markdown files.

script/process-file-manifest.js
```diff
@@ -0,0 +1,50 @@
+function processFileManifest() {
```
jbalboni: Is there a better name for this? I think it's confusing given that we have our own manifest files, in addition to the Webpack manifest. Maybe `addAssetHashes`?

jbalboni (approved):
One question, but LGTM otherwise. Regardless of WBC, I'm glad we're cleaning these files up.

jbalboni:
@ncksllvn This isn't directly relevant to this PR, but it prompted me to think: now that we don't map a content file from `rootUrl`, one way to deal with changing React app links might be to let `rootUrl` in the manifests be an array and have `startApp` pick the one that matches `window.location`.

content/pages/disability-benefits/apply/dependents/form-686c-dependents.md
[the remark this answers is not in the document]
ncksllvn: Yeah...We could add an explicit "e2e_disabled" property or something similar there.

ncksllvn:
Thanks @jbalboni! Updated with that`'process-file-manifest.js` renamed to `configure-assets`, exporting a function `addAssetHashes`. I kept the file name generic because it's going to be modified a bit in an upcoming PR to link to assets from a previous build.

PR #8403, opened 2018-09-13
ReviewCardField -- Render data type field on review

cvalarida:
## Description
The `ReviewCardField` no longer renders the "regular" review view on review. (Oi, terminology...) Instead, it renders the appropriate field based on the base data type.

## Testing done
- [x] Tested manually
- [x] Unit test written

### Mid-form, not editing
image

### Mid-form, editing
image

### Review page, not editing
image

### Review page, editing
image

## Acceptance criteria
- [x] The `ReviewCardField` renders an editable card in the form flow like normal
- [x] The `ReviewCardField` renders an object or array field on the review page

## Definition of done


src/platform/forms/save-in-progress/RoutedSavableReviewPage.jsx
```diff
@@ -166,7 +167,7 @@ function mapStateToProps(state, ownProps) {
     user
   } = state;
 
-  const formContext = getFormContext({ form, user });
+  const formContext = set('onReviewPage', true, getFormContext({ form, user }));
```
jbalboni: I think it would be better to add a `onReviewPage` property to the `getFormContext` parameter for this, and return it in the selector. Otherwise, formContext is going to be a different object each time `mapStateToProps` runs, which will cause this component to re-render on every state change (which is probably happening already, but we should avoid forcing it to for no reason).

cvalarida: Yeah, I thought about it, but didn't know what else it would affect, so I decided against it.

:thinking: I suppose like any optional parameter, it wouldn't affect the instances where this is used and not passed in. Good point.

A'ight, change incoming!

jbalboni (approved):

PR #8415, opened 2018-09-14
Va gov/rw footer update

RobAWilkinson:
## Description
Refactored footer component to use a json file, should allow for changing link titles, and href easily and quickly

## Testing done
Local testing

## Acceptance criteria
- [ ] Footer works

## Definition of done

### DIFF
--- a/src/platform/static-data/footer-links.json
+++ b/src/platform/static-data/footer-links.json
@@ -4,21 +4,21 @@
     "href": "https://www.va.gov/homeless/",
     "order": "1",
     "target": "",
-    "title": "For Homeless Veterans"
+    "title": "Homeless Veterans"
   },
   {
     "column": "1",
     "href": "https://www.va.gov/womenvet/",
     "order": "2",
     "target": "",
-    "title": "For Women Veterans"
+    "title": "Women Veterans"
   },
   {
     "column": "1",
     "href": "https://www.va.gov/centerforminorityveterans/",
     "order": "3",
     "target": "",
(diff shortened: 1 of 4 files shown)


src/platform/static-data/footer-links.json
```diff
@@ -0,0 +1,310 @@
...
+    "order": "5",
+    "target": "",
+    "title": "Mental Health"
+  },
+  {
+    "column": "1",
+    "href": "https://www.va.gov/adaptivesports/",
+    "order": "6",
+    "target": "",
+    "title": "Adaptive Sports and Special Events"
```
ncksllvn: Instead of the column property, would it possibly be easier to use a structure with the column values as keys?

```
{
    "column_1": // array of values that previously had "1" as their "column" value
}
```

ncksllvn (approved):
LGTM! If you think that there is enough happening in the footer to warrant a React component, then we can probably move it into the components folder and remove the Redux from the [entry point](https://github.com/department-of-veterans-affairs/vets-website/blob/brand-consolidation/src/platform/site-wide/va-footer/index.js). But if you're thinking a regular Metalsmith include would be better, then we can just log that as a ticket and move forward for now.

src/platform/static-data/footer-links.json
RobAWilkinson: I'd rather keep it as an array right now, seems simpler if content people need to change or update a column. Happy to update if you think it works better as an obj though.

PR #8416, opened 2018-09-14
Feedback error messages

cehsu:
## Description
These changes make the feedback form contact information page error messages more specific.

## Screenshots
image

## Definition of done

### DIFF
--- a/src/applications/edu-benefits/feedback-tool/config/form.js
+++ b/src/applications/edu-benefits/feedback-tool/config/form.js
@@ -3,7 +3,7 @@ import React from 'react';
 import fullSchema from 'vets-json-schema/dist/FEEDBACK-TOOL-schema.json';
 import dateRangeUI from 'us-forms-system/lib/js/definitions/dateRange';
 import phoneUI from 'us-forms-system/lib/js/definitions/phone';
-import { validateBooleanGroup, validateMatch } from 'us-forms-system/lib/js/validation';
+import { validateBooleanGroup } from 'us-forms-system/lib/js/validation';
 
 import FormFooter from '../../../../platform/forms/components/FormFooter';
 import fullNameUI from '../../../../platform/forms/definitions/fullName';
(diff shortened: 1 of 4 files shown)


src/applications/edu-benefits/feedback-tool/config/form.js
```diff
@@ -18,20 +18,21 @@ import SchoolSelectField from '../components/SchoolSelectField.jsx';
...
+  recordApplicantRelationship,
+  recruitingLabel,
+  refundIssuesLabel,
+  studentLoansLabel,
+  submit,
   transcriptReleaseLabel,
-  creditTransferLabel,
-  refundIssuesLabel
+  transform,
+  validateMatch,
```
erikphansen: Hmmm... https://eslint.org/docs/rules/sort-imports

src/applications/edu-benefits/feedback-tool/helpers.js
```diff
@@ -253,3 +253,11 @@ export function transformSearchToolAddress({ address1, address2, address3, city,
...
 
 }
+
+export function validateMatch(field1, field2, fieldType) {
+  return function matchValidator(errors, formData) {
+    if (formData[field1] !== formData[field2]) {
+      errors[field2].addError(`Please ensure your ${fieldType} entries match`);
+    }
+  };
+}
```
erikphansen: Why is this being used now instead of the one from the form system?

erikphansen (approved):

src/applications/edu-benefits/feedback-tool/helpers.js
cehsu: We are making the error message more specific. Should we add a todo/ticket to recommend that they update their function as well?

cehsu: Many thanks @erikphansen.

peggygannon (approved):
@cehsu Looks good. thanks!

src/applications/edu-benefits/feedback-tool/config/form.js
cehsu: We don't have that rule enabled, likely because we tend to organize our import according to their path.

erikphansen: Yeah, I was just checking it out and didn't like it. I just want to be able to enforce the sort order of the named imports. And the sort order of deconstructed variables, too, which is a different thing.

PR #8422, opened 2018-09-14
Use the review PhoneNumberWidget in ui:reviewWidget

cvalarida:
## Description
The phone number fields were in the edit mode on the review page because nothing said they shouldn't be.

Until now.

## Testing done
Made sure it worked locally.

### Veteran information, review page, not editing
image

### Veteran information, review page, editing
image

### Homelessness contact, review page, not editing
image

### Homelessness contact, review page, editing
image

## Acceptance criteria
- [x] The phone number fields don't get rendered in edit mode on the review page until we edit

## Definition of done

### DIFF
--- a/src/applications/disability-benefits/526EZ/config/form.js
+++ b/src/applications/disability-benefits/526EZ/config/form.js
@@ -75,7 +75,7 @@ import {
 import { requireOneSelected } from '../validations';
 import { validateBooleanGroup } from 'us-forms-system/lib/js/validation';
 import PhoneNumberWidget from 'us-forms-system/lib/js/widgets/PhoneNumberWidget';
-import { TextWidget } from 'us-forms-system/lib/js/review/widgets';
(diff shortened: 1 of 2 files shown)


src/applications/disability-benefits/526EZ/config/form.js
```diff
@@ -244,6 +245,7 @@ const formConfig = {
                   primaryPhone: {
```
cvalarida: :thinking: @markgreenburg Do you remember why we weren't using the phone number definition from `us-forms-system`?

jbalboni (approved):

src/applications/disability-benefits/526EZ/config/form.js
markgreenburg: I don't actually remember...I think we switched to using the `usaPhone` definition in all-claims and using `phoneUI()`. That might be a good idea here as well 🤷‍♂️

PR #8435, opened 2018-09-17
Only show prefill message on pages that were prefilled

erikphansen:
## Description
Added a new `conditionalPrefillMessage` function to act as a wrapper around the `prefillMessage` component. The `prefillMessage` will now only be shown on a form page if fields on that page were prefilled. Previously the prefill message would be shown on all form pages regardless of if that particular page we prefilled.

**BONUS:** Added a new eslint rule that will prevent us from submitting code with Mocha `.only()` statements to address this issue: https://github.com/department-of-veterans-affairs/vets.gov-team/issues/11940

## Testing done
Added unit tests. Tested locally with mock prefill data.

## Acceptance criteria
- [ ] Prefill messages show up only on pages that were prefilled; do not show the prefill message universally on all form pages.

## Definition of done

### DIFF
--- a/src/applications/edu-benefits/feedback-tool/config/form.js
+++ b/src/applications/edu-benefits/feedback-tool/config/form.js
@@ -20,10 +20,13 @@ import GetFormHelp from '../../components/GetFormHelp';
 import {
   accreditationLabel,
   changeInDegreeLabel,
+  conditionalPrefillMessage,
   creditTransferLabel,
   financialIssuesLabel,
   gradePolicyLabel,
   jobOpportunitiesLabel,
+  PREFILL_FLAGS,
+  prefillTransformer,
   qualityLabel,
   recordApplicantRelationship,
   recruitingLabel,
@@ -121,6 +124,7 @@ const formConfig = {
   formId: 'FEEDBACK-TOOL',
   version: 0,
   prefillEnabled: true,
+  prefillTransformer,
   defaultDefinitions: {
     date,
     dateRange,
@@ -179,7 +183,7 @@ const formConfig = {
           title: 'Applicant Information',
           depends: isNotAnonymous,
           uiSchema: {
-            'ui:description': PrefillMessage,
+            'ui:description': data => conditionalPrefillMessage(PREFILL_FLAGS.APPLICANT_INFORMATION, data, PrefillMessage),
             fullName: _.merge(fullNameUI, {
               prefix: {
                 'ui:title': 'Prefix',
@@ -222,7 +226,7 @@ const formConfig = {
           title: 'Service Information',
           depends: isVeteranOrServiceMember,
           uiSchema: {
-            'ui:description': PrefillMessage,
+            'ui:description': data => conditionalPrefillMessage(PREFILL_FLAGS.SERVICE_INFORMATION, data, PrefillMessage),
             serviceBranch: {
               'ui:title': 'Branch of service',
             },
@@ -245,7 +249,7 @@ const formConfig = {
           title: 'Contact Information',
           depends: (formData) => formData.onBehalfOf !== anonymous,
           uiSchema: {
-            'ui:description': PrefillMessage,
+            'ui:description': data => conditionalPrefillMessage(PREFILL_FLAGS.CONTACT_INFORMATION, data, PrefillMessage),
             address: {
               street: {
                 'ui:title': 'Address line 1'
--- a/src/applications/edu-benefits/feedback-tool/config/form.js
+++ b/src/applications/edu-benefits/feedback-tool/config/form.js
@@ -20,7 +20,7 @@ import GetFormHelp from '../../components/GetFormHelp';
 import {
   accreditationLabel,
   changeInDegreeLabel,
-  conditionalPrefillMessage,
+  conditionallyShowPrefillMessage,
   creditTransferLabel,
   financialIssuesLabel,
   gradePolicyLabel,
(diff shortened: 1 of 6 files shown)


cehsu (approved):
LGTM.

lihanli:
these are the fields in the schema that could possibly be prefilled, is this what you were asking about?
https://github.com/department-of-veterans-affairs/vets-api/blob/master/config/form_profile_mappings/FEEDBACK-TOOL.yml

erikphansen:
Thanks @lihanli! My question is more about what an empty field will look like if it isn't prefilled. For example, if `fullName` is not available, but other prefill data is, will the value of `fullName` be an empty object? Or `null`? Or will it just not be set on the prefilled data object at all?

lihanli:
@erikphansen it will not be on the object at all

src/applications/edu-benefits/feedback-tool/config/form.js
```diff
@@ -179,7 +183,7 @@ const formConfig = {
           title: 'Applicant Information',
           depends: isNotAnonymous,
           uiSchema: {
-            'ui:description': PrefillMessage,
+            'ui:description': data => conditionalPrefillMessage(PREFILL_FLAGS.APPLICANT_INFORMATION, data, PrefillMessage),
```
rianfowler: is this `formData`?

src/applications/edu-benefits/feedback-tool/helpers.js
```diff
@@ -22,6 +22,27 @@ const searchToolSchoolAddressFields = get(
   fullSchema
 );
 
+// The flags to add to formData that will indicate if a page is prefilled or not
+// These flags will be used for each form page's call to
+// conditionalPrefillMessage()
+export const PREFILL_FLAGS = {
+  // if formData.fullName is set:
+  APPLICANT_INFORMATION: 'view:prefilledApplicantInformation',
```
rianfowler: I think `prefilledApplicantInformation` suggests that this is the prefill data where `applicationInformationPrefilled` suggests boolean.

src/applications/edu-benefits/feedback-tool/helpers.js
```diff
@@ -251,7 +272,41 @@ export function transformSearchToolAddress({ address1, address2, address3, city,
     state: state && state.slice(0, searchToolSchoolAddressFields.state.maxLength),
     postalCode: zip && zip.slice(0, searchToolSchoolAddressFields.postalCode.maxLength),
   };
+}
 
+function addPrefilledFlagsToFormData(formData) {
+  const newFormData = { ...formData };
+  Object.keys(prefillFlagsToFieldsMap).forEach(flag => {
+    prefillFlagsToFieldsMap[flag].forEach(field => {
```
rianfowler: `find()` would probably work better here. You could separate the predicate from the assignment and the function would only loop as many times as needed (i.e. this only needs to loop over the predicate- not the assignment).

src/applications/edu-benefits/feedback-tool/helpers.js
```diff
@@ -251,7 +272,41 @@ export function transformSearchToolAddress({ address1, address2, address3, city,
...
+      // if the value of this field is set on formData, add the flag to
+      // newFormData
+      if (get(field, formData)) {
```
rianfowler: wish we had an `exists()` helper

src/applications/edu-benefits/feedback-tool/helpers.js
```diff
@@ -251,7 +272,41 @@ export function transformSearchToolAddress({ address1, address2, address3, city,
...
+/**
+ * Checks if the `prefillFlag` is set on the `data.formData`. If it is, returns
+ * the React Component created by the `messageComponentCreator`
+ *
+ * @param {string} prefillFlag - The key to look for on the form data
+ * @param {Object} data - Data object from form config
+ * @param {React Component} messageComponent - The React component to render
+ * @returns {React Component|null}
+ */
+export function conditionalPrefillMessage(prefillFlag, data, messageComponent) {
```
rianfowler: maybe `maybeShowPrefillMessage` or something similar that expresses what this does.

rianfowler (approved):

src/applications/edu-benefits/feedback-tool/helpers.js
erikphansen: `conditionallyShowPrefillMessage`?

src/applications/edu-benefits/feedback-tool/helpers.js
erikphansen: Well now I'm really happy I put these on the `PREFILL_FLAGS` so it's easy to change :)

src/applications/edu-benefits/feedback-tool/helpers.js
rianfowler: maybe `some()` is better. https://stackoverflow.com/a/25926600

src/applications/edu-benefits/feedback-tool/config/form.js
erikphansen: I'm not sure what the better name for `data` would be. But `formData` is a property _on_ it https://github.com/department-of-veterans-affairs/vets-website/blob/gibft-conditional-prefill/src/applications/edu-benefits/feedback-tool/helpers.js#L306

src/applications/edu-benefits/feedback-tool/helpers.js
cehsu: how about applicantInformationPrefilled?

src/applications/edu-benefits/feedback-tool/helpers.js
erikphansen: I new this smelled wrong because I wanted to avoid the inability to exit out of a `forEach` early, but I don't think I even knew about `find` until now. Thanks for the tip.

PR #8441, opened 2018-09-17
Va gov/feature/13131

accbjt:
## Description
Update src code to use manifest.json files.


RobAWilkinson:
@accbjt should we also update `static-pages-entry.js` to include the proper links for disability and and burials and memorials?

accbjt:
@RobAWilkinson none of the links in `static-pages-entry.js` are from the manifest.json files, so I don't think we need to update these in static-pages-entry.js.

ncksllvn:
Hate to do this, but because of some complications that arose from our pattern for deriving `rootUrls` for compatibility between `master` and `brand-consolidation` along with the URL reverts in the URL mappings spreadsheet, this work will need to be revisited with a more limited scope. But thanks so much for your work on this. Sorry that we weren't able to see it merged.

PR #8451, opened 2018-09-18
Jest conversion

cehsu:
## Description
These changes convert the unit tests to jest as part of the [Jest-aXe RFC conversation](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/12874).
In addition to using jest-codemods, a few remaining replacements were made manually.

How to test this locally: run `yarn test:unit`

TODO
- rename mocha-setup
- ~debug 6 disabled tests~
- ~resolve linting errors~
- ~update run-unit-test.sh~

## Acceptance criteria
- [ ] Convert unit tests

## Definition of done


package.json
```diff
@@ -74,8 +74,9 @@
     "ajv": "^4.7.1",
     "append-query": "2.0.1",
     "axe-core": "^2.3.1",
-    "babel-core": "^6.24.1",
+    "babel-core": "^6.26.3",
```
cvalarida: This upgrade looks okay. Check out the [babel changelog](https://github.com/babel/babel/blob/6.x/CHANGELOG.md) for specifics.

package.json
```diff
@@ -102,6 +103,7 @@
     "enhanced-resolve": "^0.7.6",
     "enzyme": "^3.1.0",
     "enzyme-adapter-react-15": "^1.0.2",
+    "es6-promise": "^4.2.5",
```
cvalarida: I'm guessing Jest needs this?

src/applications/pre-need/tests/definitions/address.unit.spec.jsx
```diff
@@ -34,7 +34,7 @@ describe('Pre-need definition address', () => {
     // country is USA and there is no blank option
     expect(selects.first().props().value).to.equal('USA');
     expect(selects.first().find('option').everyWhere(n => !!n.props().value)).to.be.true;
-  }).timeout(4000);
```
cvalarida: Oh interesting. Didn't know this was a thing. Seems like removing this changes the way the test works, doesn't it?

src/applications/vic-v2/tests/components/CropperController.unit.spec.jsx
```diff
@@ -9,7 +9,9 @@ describe('<CropperController>', () => {
   beforeEach(() => {
     window.addEventListener = sinon.spy();
     window.removeEventListener = sinon.spy();
+    jest.spyOn(window, 'requestAnimationFrame').mockImplementation(cb => cb());
```
cvalarida: What's this do? (Or rather, why is it needed now?)

src/applications/pre-need/tests/definitions/address.unit.spec.jsx
```diff
@@ -34,7 +34,7 @@ describe('Pre-need definition address', () => {
...
+  });
 
   it('should have required inputs if required', () => {
```
cvalarida: Wait, why didn't these get changed to `test`?

cvalarida (commented):
Nice! Looks like there's a lot of workarounds and `TODO: Figure out why this test is broken now`s, but aside from those, do you have an idea of how long it took to get to this state, @cehsu?

cvalarida:
Looks like we can [continue to use `it()`](https://jestjs.io/docs/en/api.html#testname-fn-timeout) if we wanted to. It's aliased to `test()`.

cvalarida:
Running `time npm run test:unit` on the current master:
```
real	0m53.920s
user	0m52.993s
sys	0m2.935s
```
I was watching `htop` in another terminal tab and found the CPU usage fairly low except on one core, where it was consistently 70%+ (but which core was spiked would jump around).

Running the same command on this branch:
```
real	1m41.532s
user	11m15.838s
sys	0m18.540s
```
This was actually super surprising; Seems like something is misconfigured. I read something somewhere about the setup being run before every test (or test suite?) so, I imagine that's what's going on here, causing it to be so much slower.

Watching `htop`, I found all 8 cores (:thinking: threads? :grimacing: technology...) were basically maxed out the whole time.

Anecdotally, I like the output a lot more in the jest version.
image

It's got like a live feed always at the bottom of how many tests have been run or skipped, and how long the tests have been running.

jest.config.js
```diff
@@ -0,0 +1,206 @@
...
+  // coverageThreshold: null,
+
+  // Make calling deprecated APIs throw helpful error messages
+  // errorOnDeprecated: false,
+
+  // Force coverage collection from ignored files usin a array of glob patterns
+  // forceCoverageMatch: [],
+
+  // A path to a module which exports an async function that is triggered once before all test suites
+  // globalSetup: null,
```
cvalarida: I think if we set this to run the setup before any of the tests (rather than be fore _each individual test_), we can cut back on the run time drastically. I think. Still trying to find my source for that, though; I read it _somewhere_ the other day... :thinking:

r-mc2: Or perhaps this as well? I know I haven't needed to use [clearMocks](https://jestjs.io/docs/en/configuration.html#clearmocks-boolean) in the past. And since it sounds like it takes time to clear out any mocks between *each* test that could be adding a bit (even if only to check that there are no mocks (which there likely aren't as most mocks are still Sinon)).

PR #8462, opened 2018-09-19
[Brand Consolidation] Open sign-in modal with the mega-menu sign-in button

U-DON:
## Description
Bound the mega-menu login button to a handler that opens the login modal.
Resolves department-of-veterans-affairs/vets.gov-team#13072.

NOTE: This change will give any element with the class `.signin-signup-modal-trigger` the ability to open the login modal, even if the element is not in a React component (where it would have access to the Redux action).

## Testing done
Local testing.

## Acceptance criteria
- [x] Clicking the login button in the mega-menu should open the login modal.

## Definition of done


ncksllvn (approved):
I think we'll need a solution for rendering widgets or something like that as we move forward, but given the time constraints and changing requirements I think makes sense.

PR #8470, opened 2018-09-19
Provide file metadata (front matter) to app manifests to enable dynamic root URLs

ncksllvn:
## Description
This work continues the work from PR https://github.com/department-of-veterans-affairs/vets-website/pull/8396, which is also the reason the commit log is so long. That PR modularized the Webpack build from the Metalsmith build, hooking the two together only at a single point - watch and build. 

Although the PR above also removed the `rootUrl` property's involvement in the build, we still didn't have the ability to enable dynamic root URLs, which is one of the biggest obstacles of merging `brand-consolidation` into `master`. The first step would be to optionally allow the `js` extension over the `json` extension in manifest files. My next thought was then to use the current window path to select the correct `rootUrl` for a manifest based on an array of values, but sometimes other applications import each other's manifests in order to link to one another, so we couldn't use the document location to determine that.  The simplest solution I could think was for a manifest file to check for the browser environment, import the brand-consolidation feature flag if the latter is true, and use the brand-consolidation root URL if the brand-consolidation feature flag is enabled.

There were a few issues with that approach - the manifest files lose the feel of being solely configuration, boilerplate between files to repeatedly check for the feature flag, rewriting some modules into the Node-module syntax, etc - which was a lot of work for what would be torn down after the feature flag is expired and to ultimately define a single value which is already known to the content file containing the corresponding `entryname` property to the manifest.

As a more permanent solution and as one that supports the idea of build-data flowing from the content to the app-code, I modified the `create-build-settings.js` middleware to collect the front matter of content pages that contain an entry-name. `settings.js` will then provide access to that information like -

```js
window.settings = {
  brandConsolidatedEnabled: false,
  applications: {
    "profile-360": {
     "contentProps": [
      {
       "title": "Your Profile",
       "layout": "page-react.html",
       "entryname": "profile-360",
       "mode": "0644",
       "filename": "profile/index.md",
       "path": "profile",
       "nav_path": "profile/index.html"
      }
     ]
    } 
  }
}
```

The `contentProps` property is an array containing the front matter for each page of that entry-name. There is almost always only one, but the claims-status app appeared to have two root URLs, so I didn't want to make the assumption that every app had to have one.

A manifest file can then gain access to that file by defining `receiveContentProps` function, and can bind to that `path` property as the `rootUrl`. For example, here is the revised manifest for Health Care Apply -

```
module.exports = {
  appName: '1010ez Health Care Application form',
  entryFile: './hca-entry.jsx',
  entryName: 'hca',
  receiveContentProps({ path: rootUrl }) {
    this.rootUrl = `/${rootUrl}`;
  }
};

```

Webpack sees the `js` extension on the manifest and injects code to immediately execute it with the correct data. This is so that the derived properties are always calculated when another module imports it, without having to check for the browser environment. This is also provides us with a useful way for content to provide data to the React application. 

I created the image below to help model the overall flow of the frontend build -

image

## Testing done
This work leverages the safety of automated tests, and those passing are a good sign that nothing's too broken.

## Acceptance criteria
- [x] The `brand-consolidation` branch has explicit support for dynamic root URLs

## Definition of done

## Issues
Related to this ticket - department-of-veterans-affairs/vets.gov-team/issues/12871


config/manifest-loader.js
```diff
@@ -0,0 +1,18 @@
...
+    (() => {
+      if (module.exports.receiveContentProps) {
+        const applicationSettings = window.settings.applications[module.exports.entryName];
+        module.exports.receiveContentProps(...applicationSettings.contentProps);
+      }
+    })();
+  `;
+}
+
+module.exports = manifestLoader;
```
ncksllvn: This is the actual loader that Webpack uses to add code into a `manifest.js` that will automatically execute the `receiveContentProps` function with the front matter.

config/parse-app-settings.js
```diff
@@ -0,0 +1,51 @@
...
+      contentProps
+        .map(file => file.path)
+        .forEach(filePath => routes.push(`/${filePath}`));
+    }
+
+    return routes;
+  }
+};
+
+module.exports = parsedAppSettings;
```
ncksllvn: These are function used by Webpack Dev Server and the static web server for E2E tests to create the History API rewrites.

config/webpack-metalsmith-connect/index.js
```diff
@@ -27,29 +27,33 @@ function getEntryPoints(buildOptions) {
...
+    const apps = getEntryPoints(buildOptions);
+    const webpackConfig = generateWebpackConfig(buildOptions, apps);
+    const webpackDevServerConfig = generateWebpackDevConfig(buildOptions);
+    const webpackMiddleware = webpackDevServerPlugin(webpackConfig, webpackDevServerConfig);
+
+    webpackMiddleware(files, metalsmith, (err) => {
+      if (err) throw err;
+      convertPathsMiddleware(files, metalsmith, done);
+    });
+  };
```
ncksllvn: This file was rearranged so that the configurations are generated inside the middleware function as opposed to when the middleware is registered. The reason for this is that middleware that executes prior to it, `create-build-settings`, will bind the resultant `settings` to `buildOptions`, which can be used by the dev server.

script/build.js
```diff
@@ -119,36 +110,48 @@ smith.use(layouts({
...
+  const isDevBuild = [environments.DEVELOPMENT, environments.VAGOVDEV].includes(BUILD_OPTIONS.buildtype);
+  if (!isDevBuild) {
+    smith.use(addAssetHashes(BUILD_OPTIONS));
+  }
+
+  if (process.env.CHECK_BROKEN_LINKS !== 'no') {
+    smith.use(checkBrokenLinks(BUILD_OPTIONS));
+  }
+}
+
```
ncksllvn: I shifted around the middleware registrations around so that `createBuildSettings` was called before `watch`. This is so build-settings can be passed to Webpack Dev Server.

script/create-build-settings.js
```diff
@@ -13,17 +12,58 @@ function getBuildSettings(options) {
...
+      contents: `window.settings = ${JSON.stringify(settings, null, ' ')};`
+    };
+
+    options.settings = settings;
+
+    done();
+  };
+}
+
+module.exports = createBuildSettings;
```
ncksllvn: This is the logic where the front matter is collected into the data structure. For simplicity, I filtered the data to only allow primitives.

src/platform/testing/e2e/test-server.js
```diff
@@ -30,9 +29,12 @@ if (options.unexpected && options.unexpected.length !== 0) {
...
 const root = path.resolve(__dirname, `../../../../build/${options.buildtype}`);
+appSettings.parseFromBuildDir(root);
+const routes = appSettings.getAllApplicationRoutes();
+
 app.use(morgan('combined', { skip: (req, _res) => { return req.path.match(/(css|js|gif|jpg|png|svg)$/); } }));
 app.use(express.static(root));
-getRoutes(manifests).forEach(url => {
+routes.forEach(url => {
   app.use(url, fallback(`${url}/index.html`, { root }));
 });
```
ncksllvn: This is the server that serves files during E2E tests. I would honestly like to delete it and leverage Webpack as a static web server instead, but that wouldn't belong in this PR.

src/applications/burials/manifest.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
+  appName: '21P-530 Burials benefits form',
+  entryFile: './burials-entry.jsx',
+  entryName: 'burials',
+  receiveContentProps({ path: rootUrl }) {
+    this.rootUrl = `/${rootUrl}`;
+  }
+};
```
ncksllvn: Most of the changes are like this one - converting a `manifest.json` into a `manifest.js`, and then binding to the `path` property during `receiveContentProps`.

src/applications/disability-benefits/686/tests/00.686-dependent-status.e2e.spec.js
```diff
@@ -11,4 +11,4 @@ module.exports = E2eHelpers.createE2eTest((client) => {
   client.end();
 });
 
-module.exports['@disabled'] = !manifest.production;
+module.exports['@disabled'] = manifest.e2eTestsDisabled;
```
ncksllvn: An unfortunate consequence of decoupling the content directory from the manifests is that manifest's `production` property is only useful to Webpack code. To make that more clear, I renamed it to `e2eTestsDisabled`.

I think this value could be derived by checking the `settings.js` and seeing that there are no pages registered to the manifest's entry name, but our e2e tests in general will need to be revisited, as they sort of sit outside of both app-code and content.

jbalboni:
@ncksllvn I think the overall structure of this makes sense. But I'm a little hesitant about the Webpack loader. It makes the code a little harder to follow, because there's some extra magic that you don't see from reading an app entry or manifest file.

So, I wonder if we would lose anything by making a function that took the manifest as a function and merged it with the appropriate settings data:

```js
function getApplicationSettings(manifest) {
  const settings = Object.assign({}, manifest, settings.applications[manifest.entryName].contentProps);
  settings.rootUrl = `/${settings.path}`;
  return settings;
}
```

I suppose this would involve updating the entry files for every app, but otherwise it might be easier to follow? I could be missing something about how this all works, though.

jbalboni (approved):
This looks ok to me, given the question I asked earlier.

PR #8474, opened 2018-09-19
Brand consolidation - Change /facilities to /find-locations

ncksllvn:
## Description
- Used regex `\/facilities(\/)?` to update all links to the Facility Locator to `/find-locations`
    - Trailing slash is omitted because React roots can't end with a slash.
- Changes the rootUrl of the Facility Location manifest.json
- Renames the directory `va-gov/pages/facilities` to `va-gov/pages/find-locations`

Moves department-of-veterans-affairs/vets.gov-team/issues/13071 to validate
## Testing done
`npm run build -- --brand-consolidation-enabled`


RobAWilkinson (approved):
I ran the same regex locally as you and couldnt find any others that were missing LGTM

PR #8475, opened 2018-09-19
feature/11139-svg-integration

1Copenut:
## Description
IE11 + JAWS allows users to focus on links, and then focus again on the SVG, reading the same information out twice to assistive device users, which is not idea. I did some research, and added several attributes to the SVGs so they would not be focusable by default, and a new `IconBase` wrapper component as a starting point for all future SVGs used as links.

Original issue on Zenhub: https://app.zenhub.com/workspace/o/department-of-veterans-affairs/vets.gov-team/issues/11139

This PR also bumps Formation from 1.4.14 to 1.6.3. This history includes:
* Trevor bumped to 1.5.0, and subsequent bumps to 1.5.2 for the SVG wrapper icon and package.json double entry mistake
* Jeff, Mark, Nick bumped to 1.6, 1.6.1, and 1.6.2 respectively for some accordion fixes
* Trevor bumped to 1.6.3 for a dangling greater-than that was reading out strangely in JAWS

## Testing done
* Ran full suite of unit, e2e tests
* Screenreaders:
  * IE11 + JAWS, Win10
  * Chrome + JAWS, Win10
  * Firefox + NVDA, Win10
  * Chrome + NVDA, Win10
  * Edge + NVDA, Win10
  * Safari + VoiceOver, MacOS
  * Chrome + VoiceOver, MacOS

## Definition of done


src/platform/site-wide/user-nav/components/HelpMenu.jsx
```diff
@@ -8,7 +8,7 @@ import isBrandConsolidationEnabled from '../../../brand-consolidation/feature-fl
 class HelpMenu extends React.Component {
   render() {
     const buttonText = isBrandConsolidationEnabled() ? 'Contact Us' : 'Help';
-    const icon = <IconHelp color="#fff"/>;
+    const icon = <IconHelp color="#fff" role="presentation"/>;
```
ncksllvn: Should the `role` property be part of the `IconBase` class?

ncksllvn (approved):
LGTM 👍

src/platform/site-wide/user-nav/components/HelpMenu.jsx
cvalarida: Are icons always presented with accompanying text? I think it's safe to add the default `role="presentation"` to the `IconBase` if so. We can still override it for links and whatnot. I'm not sure when we'd want for the icon to be read out by a screen reader by default.

content/includes/top-nav.html
```diff
@@ -28,6 +28,7 @@
     <div role="banner" class="va-header-logo-wrapper">
       <a href="/" class="va-header-logo" onClick="recordEvent({ event: 'nav-header-logo' });">
       {% include "assets/img/svglogo.svg" %}
+      <span class="usa-sr-only">Vets dot gov from the U S Department of Veteran Affairs</span>
```
cvalarida: @ncksllvn Is there a good way to flag this for change / removal after the brand consolidation effort gets merged in?
Maybe like a
```
{% comment %} brand consolidation change needed {% endcomment %}
```
or something regular and searchable like that? My hope is that we can search for this one string through the whole project to find all the things we've already identified need to change.

src/platform/site-wide/user-nav/components/HelpMenu.jsx
1Copenut: @cvalarida Icons aren't always presented with text, no. The Search icon when expanded is a good example. The icon is the only thing present, so having it be role="presentation" would be wrong. Which means on further review, my code is also wrong on this PR in one instance. I'll fix it and comment after lunch, so I can still make a push to master this afternoon yet.

content/includes/top-nav.html
ncksllvn: Good idea! I'm up for defining that somehow. We can update the [doc](https://github.com/department-of-veterans-affairs/vets-website/tree/master/src/platform/brand-consolidation) to reflect that too.

ncksllvn: Follow-up on this - I don't think there's a special comment needed anymore, because VA.gov has its own Metalsmith directory (`va-gov`) that is now on master. If there's something going into Vets.gov content that you want to also see included in VA.gov content, then the change should go into the `va-gov` directory as well

ncksllvn (approved):
LGTM!

PR #8291, opened 2018-08-31
Babel 7 Upgrade

cvalarida:
## Description
Upgraded to Babel 7.

All the hard work was done by running `npx babel-upgrade --write`.

From there, I had some issues I had to resolve manually.
- Importing from `'babel-polyfill'` instead of the new `@babel/polyfill`
- Upgrading `babel-plugin-lodash` to [resolve an error](https://github.com/lodash/babel-plugin-lodash/issues/196)
  - Note: I didn't increase the version number in our package.json; I just ran `yarn upgrade babel-plugin-lodash` to install the latest version

The build on my machine said it took
```
real	0m46.417s
user	0m56.856s
sys	0m3.940s
```
Interpret that how you will.

We are getting this warning:
```
WARNING in ./src/platform/polyfills/index.js
12:5-14 "export 'default' (imported as 'Modernizr') was not found in 'modernizr'
 @ ./src/platform/polyfills/index.js
 @ multi ./src/platform/polyfills history react react-dom react-redux react-router redux redux-thunk raven-js
```
I haven't looked into it too much, but that _should_ "just work" with [`modernizr-loader`](https://www.npmjs.com/package/modernizr-loader), so I'm not sure what the story is there or how to check to see if it _really is_ working.

PR #8482, opened 2018-09-20
Clean up polyfills and vendor bundle to improve bundle sizes

jbalboni:
## Description
We're shipping more polyfill and vendor code than we really need to. There are a few changes I've made here to address that:

1. Split polyfills for older browsers into a separate file. I've also made that file conditional on browsers not understanding the `nomodule` attribute, which means that newer browsers will ignore the bundle. This means that we can ship some polyfills to older browsers and reduce the polyfills for the majority of users on vets.gov. You can read an article about this here: https://philipwalton.com/articles/deploying-es2015-code-in-production-today/.
2. I've removed Modernizr. We don't actually use this anywhere other than to gate some polyfills that we don't need. There are a couple class that use the touch and no-touch classes on html, but those are never used and it does not appear that design has a use case for them.
3. I've moved react-router and history out of the vendor bundle. These aren't necessary for static pages and moving them out of the vendor bundle gives us flexibility later to upgrade or change them.

The result of this is that for newer browsers, we've reduced the combined vendor and static pages bundle from 586kb to 487kb (before gzip). That's about half due to react-router/history and half due to the polyfills change.

The next step here would be compiling different bundles for older browsers, where we could turn off a lot of the Babel transforms and remove regenerator for modern browsers. But that's more complicated and will slow down prod builds.

## Testing done

Ran the site locally, opened some React apps in IE11 and looked for errors.

## Acceptance criteria
- [x] Site works as usual in IE11
- [x] Bundle size is reduce on newer browsers

## Definition of done


.babelrc
```diff
@@ -4,10 +4,20 @@
     ["env", {
       "targets": {
         "uglify": true,
-        "browsers": ["last 2 versions", "ie 11", "not ie 10"]
+        /* This is the full list of browsers we support, but we are not using it because we want babel-polyfill
+         * to include only polyfills for more modern browsers. 
+         * Older browser polyfills are in preModernBrowserPolyfills.js
+         */
+        /* "browsers": ["Chrome 60", "Firefox 57", "iOS 9", "Edge 14", "ChromeAndroid 64", "Safari 10", "ie 11"] */
```
jbalboni: I looked in GA and found all the browsers that have greater that 0.5% usage over the last year.

src/platform/polyfills/preModernBrowserPolyfills.js
```diff
@@ -0,0 +1,109 @@
+/*
+ * This file contains polyfills for browsers that don't support
+ * the "nomodule" script attribute. These are generally older browsers
+ * like IE 11 that need more polyfills than the more common, newer versions
+ * of evergreen browsers
+ *
+ */
+
+// This section is for custom polyfills that we need
+import 'classlist-polyfill'; // DOM element classList support.
```
jbalboni: IE 11 has some gaps in classList support, that's the only reason we still need it.

.babelrc
```diff
@@ -4,10 +4,20 @@
...
+        /*
+         * This is the list of browers we support that all also have support for the nomodule script
+         * attribute, meaning we can make them ignore a file that includes polyfills for older browsers
+         */
+        "browsers": ["Chrome 61", "Firefox 60", "iOS 11", "Edge 16", "ChromeAndroid 67", "Safari 11"]
```
jbalboni: I think this will become easier in Babel 7, where you can specifically target browsers that support ES modules.

src/platform/polyfills/index.js
```diff
@@ -5,18 +5,8 @@
...
-// Basic polyfills.
-import Modernizr from 'modernizr';
-
 import './canvas-toBlob';
 
-if (!Modernizr.classlist) {
-  require('classlist-polyfill'); // DOM element classList support.
-}
-if (!Modernizr.dataset) {
-  require('dataset');  // dataSet accessor support.
```
jbalboni: We don't need this polyfill, all our supported browsers have it.

src/platform/polyfills/preESModulesPolyfills.js
```diff
@@ -0,0 +1,109 @@
+/*
+ * This file contains polyfills for browsers that don't support
+ * the "nomodule" script attribute. These are generally older browsers
+ * like IE 11 that need more polyfills than the more common, newer versions
+ * of evergreen browsers
+ *
+ */
+
+// This section is for custom polyfills that we need
+import 'classlist-polyfill'; // DOM element classList support.
```
jbalboni: IE 11 has some gaps in classList support, that's the only reason we still need it.

.babelrc
```diff
@@ -4,10 +4,20 @@
     ["env", {
       "targets": {
         "uglify": true,
-        "browsers": ["last 2 versions", "ie 11", "not ie 10"]
+        /* This is the full list of browsers we support, but we are not using it because we want babel-polyfill
+         * to include only polyfills for more modern browsers. 
+         * Older browser polyfills are in preModernBrowserPolyfills.js
```
erikphansen: Should the referenced file actually be `preESModulesPolyfills.js`?

erikphansen (approved):

rianfowler (approved):

src/platform/polyfills/preESModulesPolyfills.js
```diff
@@ -0,0 +1,109 @@
...
+ * This section is all of the core-js polyfills we need.
+ *
+ * To recreate this list: 
+ *
+ * 1. Switch the list of browsers in babelrc to the full, older list
+ * 2. Set debug to true in babelrc
+ * 3. Run a build and get the list of polyfills included in babel-polyfill from the terminal output
+ * 4. Switch the browsers list back to the newer list and run another build
+ * 5. Remove the polyfills listed in the second list from the first
+ * 6. Now you have the full list and you can format them into import statements with some
```
cehsu: 👏 Great documentation!

cehsu (approved):
LGTM.

.babelrc
ncksllvn: GA is probably the most definitive, but it's interesting that the [USA analytics](https://analytics.usa.gov/veterans-affairs/) for all Veteran Affairs sites shows IE7 at 7.9%.

package.json
```diff
@@ -221,8 +221,6 @@
     "fast-levenshtein": "^2.0.6",
     "file-loader": "^1.1.11",
     "local-storage-fallback": "^4.1.1",
-    "modernizr": "^3.6.0",
-    "modernizr-loader": "^1.0.1",
```
ncksllvn: Farewell Modernizr 👋

ncksllvn (approved):
LGTM!

ncksllvn:
Unfortunately, I'm afraid VA.gov sees nearly half its traffic come from IE - https://github.com/department-of-veterans-affairs/vets.gov-team/issues/13539.

jbalboni:
@ncksllvn We already don't support anything below IE11. This doesn't change anything about what our current site supports.

jbalboni:
Actually, this does change things slightly in that I think there were gaps in what we said we were supporting and what we were polyfilling. If anything, the preESModulesPolyfills bundle will have a few more polyfills than we're using now.

package.json
cvalarida: This should actually fix the [babel 7 branch](https://github.com/department-of-veterans-affairs/vets-website/pull/8291) I have, I think. :crossed_fingers:

PR #8492, opened 2018-09-20
Va gov/megamenu mobile bugfix

RobAWilkinson:
## Description
Cleans up mega menu nav

## Screenshots
mobile-nav


ncksllvn (approved):
Could still use some cleaning up; we can just scroll right past it on mobile and it looks kinda funky. Either way though, I think this PR covers the scope of the bug ticket.

PR #8496, opened 2018-09-21
Va gov/accordion fix

RobAWilkinson:
## Description
Makes accordions on content pages match up with previous styles.
Also makes them multi expandable.

It seems the root cause of this was not the introduction of the accordion sidenav, but in fact the fact that multiple copies of `uswds.js` code were being load on some pages.

The accordion was in fact opening due to one of the js files running, then instantly being toggled back to closed.

## Testing done
Local testing

## Screenshots
accordions

@ncksllvn is there a place we can annotate to **not** include `uswds.min.js` in files as it is already being loaded globally?


ncksllvn (approved):
LGTM, had to do something similar with the USWDS imports in the health-care section.

PR #8503, opened 2018-09-21
Update USFS, fix 526 titles

jbalboni:
## Description

This pulls the new 1.1 branch that USFS created for us, which fixes the issue with some pages not being displayed on the review page. I've also updated the titles on some pages so that it's more clear on the review page what's being shown.

This fix has revealed another issue, which is that we need a way to hide pages with no form fields on the review page.

This also affects all forms, so tagging some Rainbows team members.

## Testing done
Tested locally.

## Screenshots
screen shot 2018-09-21 at 1 44 25 pm

## Acceptance criteria
- [x] All pages are showing up for 526 review page
- [x] Forms still work across the rest of the site

## Definition of done


src/platform/forms/preSubmitInfo.js
```diff
@@ -0,0 +1,3 @@
+export default {
+  error: 'You must accept the privacy policy before continuing.'
```
jbalboni: This should probably be done across applications, but I'm also wondering if we should just see if USFS will change it in their code (we just added a period).

src/applications/hca/helpers.jsx
```diff
@@ -6,15 +6,19 @@ import AdditionalInfo from '@department-of-veterans-affairs/formation/Additional
 import {
   stringifyFormReplacer,
   filterViewFields,
-  filterInactivePages,
```
jbalboni: This was removed in the new branch we're using.

jbalboni:
There's an error in the USFS branch, this PR will be updated when that's fixed.

cvalarida (approved):
It _looks_ good from where I sit, but I haven't tested it out locally or anything.

src/applications/hca/helpers.jsx
cehsu: If this is a common use case, would it make sense to create a `filterInactivePages` platform util to abstract this logic?

src/platform/forms/preSubmitInfo.js
cehsu: The latter sounds good, would it be helpful to set up a ticket on the USFS repo?

cehsu (approved):
LGTM.

src/applications/hca/helpers.jsx
jbalboni: So far we've used it in one form, so I would say it's not common.

PR #8532, opened 2018-09-24
add overriding style

rianfowler:
## Description
update the color of medallia survey link 

## Testing done
none- unable to generate this markup locally. will need to validate in staging. 

## Screenshots
It should look like this once the changes are applied: 
screen shot 2018-09-24 at 14 56 36

## Definition of done
~- [ ] Events are logged appropriately~
~- [ ] Documentation has been updated, if applicable~


src/platform/site-wide/sass/modules/_m-medallia.scss
```diff
@@ -11,3 +11,7 @@ button {
     }
   }
 }
+
+.kampyle_feedback-button.kampyle_bottom-right .kampyle_triangle {
+  border-bottom: 81px solid #fdb81e !important;
```
rianfowler: The original style is added directly to a `<style>` node which AFAIK requires an `!important` attribute for an imported stylesheet to override.

cehsu (approved):
LGTM. We may also want to add your comment in this stylesheet to indicate why important is being used.

src/platform/site-wide/sass/modules/_m-medallia.scss
erikphansen: Yeah, if that `<style>` is added by the 3rd party library there isn't much we can do here. I do wonder if it's possible to just do an override of just the `border-color` or `border-bottom-color` here.

erikphansen (approved):

erikphansen:
Probably out of scope for this, but it would be awesome if we could also get rid of that right margin of 0.5em, which actually comes from the containing button.

src/platform/site-wide/sass/modules/_m-medallia.scss
rianfowler: good idea- updated.

rianfowler:
@erikphansen I'll stick that margin fix in and see if someone says something.

PR #8533, opened 2018-09-24
Add frontend metrics tracking module

kfrz:
## Description
In #13291 we established an outline of work needed in order to implement pageload metrics. 

We need to implement a shared metrics setting that can be called on a given page in order to provide FE performance tracking data for that page.

This PR is linked to #13348 but represents the following issues:

- [#13348](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/13348) 
- [#13350](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/13350)
- [#13351](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/13351)
- [#13355](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/13355)

## Testing done
- [x] Manually set `metrics-sample-percentage` value in `options.js` run command, ensure debug messages do not appear, ensure data is sent to the backend properly
- [x] Manually tested browser compatibility handling by running requests in Firefox and ensuring `timeToPaint` comes back as `0`
- [x] Unit specs

## Acceptance criteria

- [x] All [concerns and considerations](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/13291#issuecomment-422518946) are addressed

#### Function skeleton/code organization  [#13348](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/13348)

- [x] Code exists that can be included on a given page that will contain metrics assessment functions
- [x] Code within this function/component will be loaded & ran after page load. 
- [x] Code can be limited to a subset of all users via [Rate Limiter](https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/platform/monitoring/RateLimiter.jsx). (Similarly used in [VIC code](https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/applications/veteran-id-card/containers/VeteranIDCard.jsx))
- [x] Code is only included when request matches a given whitelist

#### Add browser compatability checks & captureMetrics function [#13350](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/13350)
- [x] Leverage functions determined in #13349 to pull data from appropriate browser APIs
- [x] Handle browser incompatibilities

#### Build metrics payload [#13351](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/13351)
- [x] Include a page UUID or some other identifier that uniquely maps to a given page/url/resource
- [x] Include all key metrics as determined in #13349 and returned in #13350
- [x] Sync with BE to ensure object meets endpoint contract 
- [x] ~~Consider any [custom events](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/13291#issuecomment-422547266)~~ Code can be extended in future work effort to allow for custom event logging
- [x] ~~Conditionally add a tag if current page is a TeamSite page~~  Removed now that scope of MVP has shifted to smaller subset of whitelisted non-teamsite pages

#### Send metrics data to backend [#13355](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/13355)

- [x] Utilize [`navigator.sendBeacon`](https://developers.google.com/web/fundamentals/performance/navigation-and-resource-timing/#phoning_home) to async send data object built in #13351 to the endpoint built in #13352
- [x] Metrics payload is send on page `unload` event
- [x] Handle browser incompatibility 

## Definition of done


script/options.js
```diff
@@ -8,6 +8,7 @@ const environments = require('./constants/environments');
 const COMMAND_LINE_OPTIONS_DEFINITIONS = [
   { name: 'buildtype', type: String, defaultValue: environments.DEVELOPMENT },
   { name: 'brand-consolidation-enabled', type: Boolean, defaultValue: false },
+  { name: 'frontend-metrics-enabled', type: Boolean, defaultValue: false },
```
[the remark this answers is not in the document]
kfrz: Great question. I think for now limiting it to a passed argument is ok until we've done some more testing in a production-like environment.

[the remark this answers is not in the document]
jbalboni: I don't think we want the separate command line option here. Brand consolidation does that because we need to build parallel environments. In this case, I think we just want to keep it turned off in production? Typically we check `__BUILDTYPE__` in code to do that kind of gating.

src/platform/frontend-metrics/feature-flag.js
```diff
@@ -0,0 +1,8 @@
+/**
+ * Returns whether the current Webpack build is executing with the --frontend-metrics-enabled flag set.
+ * @returns {boolean}
+ * @module platform/frontend-metrics/feature-flag
+ */
+export default function isMetricsEnabled() {
+  return window.settings && window.settings.metrics.enabled;
+}
```
ncksllvn: `platform/monitoring` might be a good place for `frontend-metrics` to live

script/options.js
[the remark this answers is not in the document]
ncksllvn: Yeah agree with @jbalboni. This seems like a feature that should be toggled based on `__BUILDTYPE__`.

script/create-build-settings.js
```diff
@@ -9,6 +9,10 @@ function getBuildSettings(options) {
   return {
     type: options.buildtype,
     brandConsolidationEnabled: !!options['brand-consolidation-enabled'],
+    metrics: {
+      enabled: !!options['frontend-metrics-enabled'],
+      samplePercentage: options['metrics-sample-percentage'],
+    },
```
ncksllvn: I'm not against having these here, but unless they're possibly going to be changed outside of our regular build cycles, it might be easier to have them closer to the rest of the metrics code.

jbalboni: I could see us wanting to turn down the sample percentage outside of build cycle, but I'm not sure about turning it on or off.

src/platform/site-wide/index.js
```diff
@@ -18,13 +18,21 @@ import startVAFooter from './va-footer';
...
+import startMetrics from '../frontend-metrics/metrics';
+
 /**
  * Start up the site-wide components that live on every page, like
  * the login widget, the header menus, and the feedback widget.
  *
  * @param {Store} commonStore The Redux store being used by this application
  */
 export default function startSitewideComponents(commonStore) {
+  if (isMetricsEnabled() && withinMetricsSample()) {
```
jbalboni: I think I would favor putting a single function call in this file and doing this logic in code that lives with the metrics files. It's easier if we keep this file as simple as we can (I already dislike how much conditional logic is in here).

jbalboni (commented):
This looks good so far and I agree with Nick's comments as well. I can take another look when you make updates.

script/create-build-settings.js
kfrz: I removed the build flag and left the sampling percentage until after we have a sync this afternoon (9/26) to discuss our business case a bit more in-depth. Afterwords we'll have a better idea if this needs to be a configurable setting or hardcoded constant.

script/options.js
```diff
@@ -11,6 +11,7 @@ const defaultHost = 'localhost';
 const COMMAND_LINE_OPTIONS_DEFINITIONS = [
   { name: 'buildtype', type: String, defaultValue: environments.DEVELOPMENT },
   { name: 'brand-consolidation-enabled', type: Boolean, defaultValue: false },
+  { name: 'metrics-sample-percentage', type: Number, defaultValue: 0 },
```
jbalboni: How are you intending this option to get used? Just locally?

kfrz: The intention was for both local use as well as using this option to set sample percentage in `preview.va.gov`

Can we pull from the env to do something like `defaultValue: (ENV['sampleRate'] || 0)`? Or perhaps just decide on a value and hardcode it here (or close to the metrics module)?

Or do you have another suggestion altogether?

src/platform/monitoring/frontend-metrics/tests/metrics.unit.spec.js
```diff
@@ -0,0 +1,88 @@
...
+        expect(spy.callCount).to.be.equal(1);
+      });
+    });
+
+    describe('captureMetrics', () => {
+      xit('should instantiate a new PerformanceObserver', () => {
+        const mock = sinon.mock(PerformanceObserver());
+        window.PerformanceObserver = sinon.stub(PerformanceObserver.prototype, 'constructor').returns(mock);
+        // TODO: This fails, in captureMetrics() with PerformanceObserver is not defined.
+        // How can we test the constructor is called here?
```
bshyong: @kfrz will this work?

```

\\ PerformanceMock constructor is called in beforeEach: global.performance = new PerformanceMock();

const performanceObserverFunc = sinon.spy(window, 'PerformanceObserver');
expect(performanceObserverFunc.called).to.be.equal(true);
```

bshyong (approved):

script/options.js
jbalboni: Typically, if we wanted to vary a value like this, we would have a constant in the code and change it there, rather than a command line option, which requires changing the Jenkinsfile for any change (and restarting the watch task if you're running locally).

src/platform/monitoring/frontend-metrics/tests/metrics.unit.spec.js
kfrz: @jbalboni  Yes! Updated to use `global.PerformanceObserver = sinon.spy();` and it passes as expected. Thanks!

src/platform/monitoring/frontend-metrics/metrics.js
```diff
@@ -0,0 +1,166 @@
...
+export function captureMetrics() {
+  const observer = new PerformanceObserver(list => {
+    list.getEntriesByType('navigation').forEach(entry => {
+      const metricsPayload = buildMetricsPayload(entry);
+      window.addEventListener('unload', () => {
+        sendMetricsToBackend(metricsPayload);
+      });
+    });
+  });
+
```
ncksllvn: I think this was commented by @jbalboni before, but I've had weird experiences with the `unload` event, so I'm wondering if it would be more reliable to invert this by having only one listener for the `unload` event that iterates through all of the `entry` items, instead of creating a listener for each item.

src/platform/monitoring/frontend-metrics/metrics.js
```diff
@@ -0,0 +1,166 @@
...
+ * @see https://developer.mozilla.org/en-US/docs/Web/API/PerformanceObserver/observe
+ */
+function canCaptureMetrics() {
+  if (performance === undefined) {
+    return false;
+  } else if (!('PerformanceObserver' in window)) {
+    return false;
+  }
+  return true;
+}
```
ncksllvn: Could this possibly be reduced to `return window.performance && window.PerformanceObserver`?

src/platform/monitoring/frontend-metrics/tests/metrics.unit.spec.js
```diff
@@ -0,0 +1,99 @@
...
+
+    describe('captureMetrics', () => {
+      it('should instantiate a new PerformanceObserver', () => {
+        global.PerformanceObserver = sinon.spy();
+        metrics.captureMetrics();
+        expect(PerformanceObserver.calledWithNew()).to.be.equal(true);
+      });
+    });
+  });
+});
```
ncksllvn: 🔥

src/platform/monitoring/frontend-metrics/feature-flag.js
```diff
@@ -0,0 +1,12 @@
...
+ * Returns whether the current environment should have metrics enabled.
+ * @returns {boolean}
+ * @module platform/monitoring/frontend-metrics/feature-flag
+ */
+export default function isMetricsEnabled() {
+  const environments = [
+    'preview'
+  ];
+
+  return !!environments.includes(__BUILDTYPE__);
```
ncksllvn: The `preview` environment is considered a Production environment, so I think this should be `vagovstaging`, but I'm not sure what the launch plan is. It may also be worth adding dev, `vagovdev`.

Is there any plan to incorporate this into Vets.gov prior to launching VA.gov?

src/platform/monitoring/frontend-metrics/whitelisted-paths.js
```diff
@@ -0,0 +1,12 @@
...
+ * in the department-of-veterans-affairs/vets-api repository at:
+ * https://github.com/department-of-veterans-affairs/vets-api/blob/master/lib/benchmark/whitelist.rb
+ * Any changes made must be made to both.
+ */
+export const whitelistedPaths = [
+  '/',
+  '/disability',
+  '/facilities',
+  '/disability/how-to-file-claim'
+];
```
ncksllvn: This is confusing, but technically every page will end with a `/`, unless it's a React page. So, you could add the slashes to the URLs here (except facilities since that's a React app), and clean up [`pathIsWhitelisted`](https://github.com/department-of-veterans-affairs/vets-website/pull/8533/files#diff-ae86a2d3b3bef462fd5aa319654acb24R143) a bit.

ncksllvn (changes requested):
This is looking awesome and is super exciting! Only thing I see is adjusting the feature flag for the Staging environment.

src/platform/monitoring/frontend-metrics/tests/metrics.unit.spec.js
kfrz: :fire_engine::fire_engine::fire_engine:

src/platform/monitoring/frontend-metrics/whitelisted-paths.js
kfrz: It seems as if (locally) even React pages are loading with a trailing backslash. I updated this PR and the api whitelist to do this -- also updated the [devops ticket](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/13402) for accuracy.

src/platform/monitoring/frontend-metrics/feature-flag.js
kfrz: Reviewing the original [discovery](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/13291#issue-360974902) - looks like just dev and staging for VA.gov, and Vets.gov later.

ncksllvn (approved):
Awesome work!! Very excited to see this in action!

src/platform/monitoring/frontend-metrics/metrics.js
ncksllvn: Just making sure my suggestion didn't break anything here? Not sure if `getEntriesByType` needed to be outside of that `unload` event or not lol.

PR #8539, opened 2018-09-25
WIP 13351 build metrics payload

kfrz:
**CLOSED & SQUASHED INTO #8533**

---

## Description
Within the metrics FE code, we need to build an object consisting of page load metric data in a form that the backend can consume. 

## Acceptance criteria
- [x] Include a page UUID or some other identifier that uniquely maps to a given page/url/resource
- [x] Include all key metrics as determined in #13349 and returned in #13350
- [x] Sync with BE to ensure object meets endpoint contract 
- [ ] Consider any [custom events](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/13291#issuecomment-422547266) (see [W3C spec](https://dom.spec.whatwg.org/#interface-customevent))
- [ ] Conditionally add a tag if current page is a TeamSite page 

## Definition of done

### DIFF
--- a/src/platform/frontend-metrics/metrics.js
+++ b/src/platform/frontend-metrics/metrics.js
@@ -2,12 +2,65 @@ import Raven from 'raven-js';
 import isMetricsEnabled from './feature-flag';
 
 /**
+ * Returns 0 if Paint timing is not supported or the time of first-contentful-paint if present.
+ * @returns (number) The number retrieved from the paintEntries.startTime value, else 0.
+ */
+function contentfulPaintEntry() {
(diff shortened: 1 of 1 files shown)


src/platform/frontend-metrics/metrics.js
```diff
@@ -1,13 +1,66 @@
...
+    metricsArray.push({
+      metric,
+      duration: metrics[metric]
+    });
+  });
+
+  metricsData.append('metrics', JSON.stringify(metricsArray));
+  metricsData.append('page_id', pageUrl);
+
+  return metricsData;
```
kfrz: This should return a JSON object if at all possible so the backend doesn't have to parse it.

PR #8543, opened 2018-09-25
Add redirects for old va.gov pages that we are replacing

jbalboni:
## Description

In the URL mapping spreadsheet, there are current va.gov pages that we want to redirect to current vets.gov pages. This creates client-side redirects for each of those pages.

NOTE: This is not the ideal way of doing these redirects. I think this should be treated as a fallback and that we should try to create real 301 redirects in nginx as our primary solution.

This will also require an nginx config change anyway, to serve `.asp` urls as html.

The source list for these redirects was generated by this code: https://github.com/department-of-veterans-affairs/vets.gov-utility/pull/23/files.
## Testing done

Verified a few pages locally.

## Acceptance criteria
- [x] Pages listed as old va.gov pages redirect to the vets.gov page when opened

## Definition of done


script/options.js
```diff
@@ -76,7 +77,8 @@ function applyEnvironmentOverrides(options) {
 function applyBrandConsolidationOverrides(options) {
   Object.assign(options, {
     contentRoot: '../va-gov',
-    collections: require('./collections/brand-consolidation.json')
+    collections: require('./collections/brand-consolidation.json'),
+    redirects: require('../config/vagovRedirects.json')
```
ncksllvn: Since the redirects are set up in Metalsmith, can we store that JSON file in the `script` directory?

ncksllvn (approved):
Minor suggestion, but LGTM!

script/create-redirects.js
```diff
@@ -8,7 +8,7 @@ function getRedirectPage(redirectToPath) {
   return `
     <!doctype html>
     <head>
-      <meta http-equiv=refresh content="1; url=${absolutePath}">
+      <meta http-equiv=refresh content="0; url=${absolutePath}">
```
ncksllvn: I had that `1` there only to reveal momentarily that we had a link that should be updated, although that isn't an option for sites outside of ours.

script/options.js
jbalboni: Sure

script/create-redirects.js
jbalboni: I can revert that back. From what I read, shorter intervals there made it more likely that you wouldn't lose some of the link ranking in search engines. But given that this is a fallback, maybe that's not necessary.

jbalboni:
I've updated this PR to only cover the www.va.gov domain redirects. Redirects for other domains have to be handled in JS injected into the teamsite pages.

PR #8547, opened 2018-09-25
[FacilityLocator] Initial Changes from provloc-sprint1 Branch

r-mc2:
## Description
Initial work to break up the work already done in [provloc-sprint1](https://github.com/department-of-veterans-affairs/vets-website/compare/provloc-sprint1) branch. This pulls in the first two commits over there as well as adding in some initial logic for a feature flag to hide the new Community Care Provider features.

Those early commits essentially set up some groundwork for adding more later + some renamings. Also added descriptions to some methods and initial clean up of some React logic.

(Some additional help on how to better handle the feature flag would be appreciated or is setting it to trigger on prod vs. non-prod good enough? Also the e2e test isn't liking when I include the feature-flag function, `require` vs. `import` seems to confuse the engine)

CC: @ehuntdsva @mc-usds & @department-of-veterans-affairs/code-reviewers-for-external-teams-group For Code Review

## Testing done
Existing tests are sufficient for these modifications but some changes were made to the tests due to renaming (and early experimentation with redux state changes)

## Acceptance criteria
- [ ]

## Definition of done
- [x] Tests still pass with changes


src/applications/facility-locator/actions/index.js
```diff
@@ -38,24 +48,82 @@ export function fetchVAFacility(id, facility = null) {
...
+      console.log('Reverse geocoded address:', address); //eslint-disable-line
+
+      return fetchProviders(address, dispatch, serviceType, page); // eslint-disable-line
+    });
+
+    return null;
+  };
+};
+
+const fetchProviders = (address, dispatch, serviceType, page) => {
```
U-DON: I think you could avoid passing the `dispatch` if you wrap the contents of this function in `return (dispatch) => { ... }` and do `dispatch(fetchProviders(address, serviceType, page))` whenever you need to call it.

Or if you want to continue including it, maybe have `dispatch` before all of the other params to have a clearer separation between a helper function and the actual arguments of the function.

U-DON (commented):
I think this looks good, but maybe @dhall would have better insights.

src/applications/facility-locator/actions/index.js
r-mc2: This, too, gets cleaned up in a later commit when I refactor:
[actions/index.js#L79](https://github.com/department-of-veterans-affairs/vets-website/blob/4c586c2342f917601ceae2b0ff171659904ac577/src/applications/facility-locator/actions/index.js#L79) & [actions/index.js#L108](https://github.com/department-of-veterans-affairs/vets-website/blob/4c586c2342f917601ceae2b0ff171659904ac577/src/applications/facility-locator/actions/index.js#L108)

But that is a good point about wrapping the calls in `dispatch` initially, it felt weird passing it anyway 😄

src/applications/facility-locator/containers/VAMap.jsx
```diff
@@ -1,8 +1,14 @@
...
-import { updateSearchQuery, searchWithAddress, searchWithBounds, fetchVAFacility } from '../actions';
-import { map, find, compact, isEmpty, debounce } from 'lodash';
+import {
+  updateSearchQuery,
+  genBBoxFromAddress,
+  searchWithBounds,
+  searchProviders,
+  fetchVAFacility
+} from '../actions';
+import { map, /* find, */ compact, isEmpty, debounce } from 'lodash';
```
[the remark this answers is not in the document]
r-mc2: Yea, I left it in initially as I wasn't sure if my change was going to work out so I held on to it "just in case". It's gone in the latest version of the file.

src/applications/facility-locator/containers/VAMap.jsx
```diff
@@ -156,24 +178,29 @@ class VAMap extends Component {
...
     }, (err, res) => {
       const coordinates = res.features[0].center;
       const placeName = res.features[0].place_name;
-      const zipCode = find(res.features[0].context, (v) => {
-        return v.id.includes('postcode');
-      }).text;
+      const zipCode = res.features[0].context.find(v => v.id.includes('postcode')).text || '';
+      // const zipCode = find(res.features[0].context, (v) => {
+      //   return v.id.includes('postcode');
+      // }).text;
```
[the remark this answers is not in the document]
r-mc2: (same as below, I remove the dead code in a later commit)

src/applications/facility-locator/actions/index.js
```diff
@@ -38,24 +48,82 @@ export function fetchVAFacility(id, facility = null) {
...
+  const params = compact([
+    `address=${address}`,
+    'type=cc_provider',
+    serviceType ? `services[]=${serviceType}` : null,
+    `page=${page}`
+  ]).join('&');
+  const url = `${api.url}?${params}`;
```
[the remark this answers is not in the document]
r-mc2: I can certainly look into this. At the time I made these changes I was just trying to follow the patterns already established in the existing Locator code.

Also, to note, this file + the API calls themselves will change over time as later commits get added to subsequent PRs. (e.g. I refactored the API calls into their own class to make it easier to create a [MockLocatorAPI](https://github.com/department-of-veterans-affairs/vets-website/blob/cdcf3e9d7f09962ebf724552d409a76ebef0ed75/src/applications/facility-locator/api/MockLocatorApi.js) for local/unit testing)

r-mc2:
Any specific thoughts on the way I set up the feature flag? (i.e. better ways to do it/set it up?)

Also commenting out that chunk in the e2e test bothers me. Feel like there's got to be a way to keep it in without breaking the build.

U-DON:
Feature flagging based on environment seems good to me. I don't have any further context other than that it appears it should be disabled in Production, which makes sense.

It seems E2E tests use Node modules rather than the native ES2015 modules. If you simply wanted to base the feature flag on environment, a couple of E2E tests currently check `process.env.BUILDTYPE !== 'production'`. You could then add a `ccLocatorEnabled` function that does that to the `facility-helpers` module.

r-mc2:
Ahh, great @U-DON . I'll add that into the next PR just to keep this one simpler and hopefully move the process along a bit for all of us.

src/applications/facility-locator/containers/VAMap.jsx
```diff
@@ -128,25 +137,38 @@ class VAMap extends Component {
...
-  // pushes coordinates to URL so that map link is useful for sharing
-  // TODO (bshyong): try out existing query-string npm library
+  /**
+   * Regenerates the URL based on the given parameters so that
+   * the map link stays useful for sharing. 
+   * 
+   * @param {Object} params 
+   */
   updateUrlParams = (params) => {
+    // TODO (bshyong): try out existing query-string npm library
```
mc-usds: What is this todo for?

r-mc2: I would imagine we would have to ask 'bshyong'?

I'm honestly not sure, it was part of the codebase when I started on it. Although now that I think about it, it may be something along the lines of what @U-DON said above for API things:
> Maybe consider using the `append-query` library that we have available in our dependencies.

(Some idea to pull in a package and do a different/better job of handling URL params?)

r-mc2: Yea, this is more of the function:

```jsx
const queryParams = compact(map({
  ...location.query,
  zoomLevel: currentQuery.zoomLevel,
  page: currentQuery.currentPage,
  address: currentQuery.searchString,
  facilityType: currentQuery.facilityType,
  serviceType: currentQuery.serviceType,
  ...params,
}, (value, key) => {
  if (value) { return `${key}=${value}`; }
  return null;
})).join('&');
```

Handwritten logic to handle building out the URL query parameters. It's sounding like a "I should come back and do this less manually later with the library in the comment". But can't be 100% sure.

r-mc2:
> intended to make the PR review easier by breaking it up

My understanding was that it was this, yea. As for timeframe, given how the PRs are being built, the next PR will be ready as soon as the previous one is approved and merged to master (needs to be in master so I can make the next PR branch (if I branch off of the previous PR branch the new branch keeps the commit history and makes the next PR the history of both PRs thereby preventing keeping the PRs small)).

dhall:
I've talked with @r-mc2 via slack to better understand the overall goal and the process of getting us there, so I think I'm up to speed on where we are headed :)

Ideally small features and thus small PR's would be great, but given that breaking this up means incomplete features and code in a state that gets changed/improved in later PR's means that the feedback is premature (as we've seen in the feedback so far) and that we'd be merging code to production that's not our best work. That's not a knock on @r-mc2's code at all.. just the reality of the iterative nature of writing code.

With this in mind, I think it would be better from an engineering standpoint to put the time into reviewing the finished code in the branch mentioned in the beginning of this PR (https://github.com/department-of-veterans-affairs/vets-website/compare/provloc-sprint1) and merge it whole as opposed to the code churn that splitting it up into incomplete pieces would cause. Of course this is one engineers opinion, so I'd love to get other opinions about this approach.

cc @U-DON, @mc-usds, @jkassemi and anyone else on the thread that's interested :)

r-mc2:
Closing in favor of [PR#8651](https://github.com/department-of-veterans-affairs/vets-website/pull/8651) based on feedback above.

PR #8560, opened 2018-09-26
Update presubmit transformer

markgreenburg:
## Description
Updates the pre-submit transformer for the 526 Increase form to only include the `treatments` property when there's at least one treatment in the array.

## Testing done
- Tested locally
- Unit tests

## Screenshots
- No change

## Acceptance criteria
- [x] 526 is submitted without a `treatments` property when veteran does not select VA facility evidence.

## Definition of done

### DIFF
--- a/src/applications/disability-benefits/526EZ/helpers.jsx
+++ b/src/applications/disability-benefits/526EZ/helpers.jsx
@@ -118,6 +118,7 @@ export function transform(formConfig, form) {
 
   const additionalDocuments = aggregate(disabilities, 'additionalDocuments');
   const privateRecords = aggregate(disabilities, 'privateRecords');
+  const treatments = aggregate(disabilities, 'treatments');
 
   const transformedData = {
     disabilities: disabilities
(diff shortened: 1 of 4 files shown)


src/applications/disability-benefits/526EZ/helpers.jsx
```diff
@@ -118,19 +118,21 @@ export function transform(formConfig, form) {
...
     veteran: setPhoneEmailPaths(veteran),
-    // Extract treatments into one top-level array
-    treatments: aggregate(disabilities, 'treatments'),
     attachments: additionalDocuments.concat(privateRecords),
     privacyAgreementAccepted,
     serviceInformation,
     standardClaim,
+    // treatments has a minItems: 1 requirement so only include the property
+    // if there is at least one treatment to send
+    ...treatments.length && { treatments }
```
cvalarida: :thinking: How does this spread operator work with `.length`?

cvalarida (approved):

src/applications/disability-benefits/526EZ/helpers.jsx
markgreenburg: The whole condition gets evaluated and returns `treatments`, which then gets spread. It's kinda tricky, but I preferred this to a whole bunch of lines that conditionally set up different object shapes based on the `treatments` property.

BTW, `treatments` should always be an array in this case, because the `aggregate` function will return an empty array if it doesn't find any matching properties.

cvalarida: So this is more like a `...(treatments.length && { treatments })`? What happens if `!treatments.length`?

markgreenburg: evaluates to false, which doesn't spread anything. Which is convenient.

cvalarida: Cool deal. :+1:

PR #8562, opened 2018-09-26
Rewrite some va.gov urls to staging urls

jbalboni:
## Description

This searches for particular domain strings and replaces them with others, in non-production environments. We're doing this to support linking to va.gov staging domains for some links while on our own staging and development sites, so we can see the injected headers on those pages before we go live.

This is set up to work in the static pages and also in the mega menu and footer React components. I haven't created anything to handle this in React apps because we should have fairly few links to these pages in those apps, and because I don't yet know the exact domains we're going to be replacing. So I'm punting on that for now.

## Testing done

Tested locally that the links we should update are updated and site still works

## Screenshots
screen shot 2018-09-26 at 4 07 55 pm

## Acceptance criteria
- [x] Links are replaced in non-prod environments
- [x] Links aren't replaced in prod environments

## Definition of done


src/platform/utilities/environment/stagingDomains.js
```diff
@@ -0,0 +1,63 @@
+// This list also exists in script/options.js
+const domainReplacements = [
+  { from: 'www\\.cem\\.va\\.gov', to: 'staging.cem.va.gov' }
+];
```
ncksllvn: I think this data would have to be passed through the `create-build-settings.js` middleware and referenced through the window. We probably need some sort of module to make lookups to that data more centralized, versus global access.

ncksllvn: I'm going to explore some other ideas, but at the moment, that might be the method of passing data.

jbalboni:
There's going to be a later ticket to add more staging subdomains, once we get those back.

script/rewrite-va-domains.js
```diff
@@ -0,0 +1,29 @@
...
+
+          file.contents = new Buffer(contents);
+        });
+    }
+
+    done();
+  };
+}
+
+module.exports = createRedirects;
```
ncksllvn: I feel like we're getting pretty good at writing redirect middleware at this point 😆

src/platform/utilities/tests/environment/stagingDomains.unit.spec.js
```diff
@@ -0,0 +1,76 @@
...
+      expect(replaceDomainsInData(data)).to.deep.equal({
+        test: 'www.va.gov/testing',
+        other: {
+          href: `${currentEnv}.va.gov`,
+        }
+      });
+    });
+  });
+});
+
```
ncksllvn: 🔥

ncksllvn (approved):
LGTM!

PR #8563, opened 2018-09-26
Add Redux actions, reducers, selectors for 686 auth

erikphansen:
Hits new endpoint to check for 30% disability raiting

### DIFF
--- a/src/applications/disability-benefits/686/reducers/authorization.js
+++ b/src/applications/disability-benefits/686/reducers/authorization.js
@@ -1,0 +1,44 @@
+import { requestStates } from '../../../../platform/utilities/constants';
+import {
+  LOAD_DISABILITY_RATING_STARTED,
+  LOAD_DISABILITY_RATING_SUCCEEDED,
+  LOAD_DISABILITY_RATING_FAILED
+} from '../actions';
+
+const initialState = {
+  isLoading: false,
+  hasError: false,
+  disabilityRatingState: requestStates.notCalled,
+  payload: null,
+};
+
+function authorizationReducer(state = initialState, action) {
+  switch (action.type) {
+    case LOAD_DISABILITY_RATING_STARTED:
+      return {
+        ...state,
+        isLoading: true,
+        hasError: false,
+        disabilityRatingState: requestStates.pending,
+      };
+    case LOAD_DISABILITY_RATING_SUCCEEDED:
+      return {
+        ...state,
+        isLoading: false,
+        hasError: false,
+        disabilityRatingState: requestStates.succeeded,
+        payload: action.payload,
+      };
+    case LOAD_DISABILITY_RATING_FAILED:
+      return {
+        ...state,
+        isLoading: false,
+        hasError: true,
+        disabilityRatingState: requestStates.failed,
+        payload: action.error,
+      };
+    default:
+      return state;
+  }
+}
+export default authorizationReducer;
--- a/src/applications/disability-benefits/686/reducers/authorization.js
+++ b/src/applications/disability-benefits/686/reducers/authorization.js
@@ -1,40 +1,35 @@
-import { requestStates } from '../../../../platform/utilities/constants';
 import {
-  LOAD_DISABILITY_RATING_STARTED,
-  LOAD_DISABILITY_RATING_SUCCEEDED,
-  LOAD_DISABILITY_RATING_FAILED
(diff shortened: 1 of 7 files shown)


src/applications/disability-benefits/686/reducers/authorization.js
```diff
@@ -0,0 +1,44 @@
...
+} from '../actions';
+
+const initialState = {
+  isLoading: false,
+  hasError: false,
+  disabilityRatingState: requestStates.notCalled,
+  payload: null,
+};
+
+function authorizationReducer(state = initialState, action) {
```
cehsu: Should we rename this disabilityRating to distinguish from the authorization selector?

src/applications/disability-benefits/686/selectors/index.js
```diff
@@ -0,0 +1,9 @@
+// selectors for the `authorization686` state
+import { createStructuredSelector } from 'reselect';
+
+export const get686State = createStructuredSelector({
```
cehsu: Should we specify "authorization" in the name here?

cehsu (approved):
LGTM.

src/applications/disability-benefits/686/selectors/index.js
```diff
@@ -0,0 +1,9 @@
...
+  profileIsLoading: state => state.user.profile.loading,
+  isLoggedIn: state => state.user.login.currentlyLoggedIn,
+  isVerified: state => state.user.profile.verified,
```
cehsu: It looks like we might want to provide the user.profile.status as well to display appropriate error messages in the event that we cannot retrieve/find verified status, as done in RequireLoginView.

cehsu: We will also likely want to provide access to user.login.currentlyLoggedIn to determine where we route them in the error message (sign in or verify link)

erikphansen: `user.login.currentlyLoggedIn` is already in there, but we'll also want the get access to the status of the disability rating check since we'll want to show a spinner if it's still getting the status.

src/applications/disability-benefits/686/reducers/authorization.js
```diff
@@ -0,0 +1,44 @@
...
+import {
+  LOAD_DISABILITY_RATING_STARTED,
+  LOAD_DISABILITY_RATING_SUCCEEDED,
+  LOAD_DISABILITY_RATING_FAILED
+} from '../actions';
+
+const initialState = {
+  isLoading: false,
+  hasError: false,
+  disabilityRatingState: requestStates.notCalled,
```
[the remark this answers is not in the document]
erikphansen: Good call. I think my original thought was to lean toward giving more granular info even if it wasn't used. But this seems totally redundant with `isLoading` and `hasError`.

src/applications/disability-benefits/686/selectors/index.js
```diff
@@ -0,0 +1,16 @@
+// selectors for the `authorization686` state
+import { createStructuredSelector } from 'reselect';
+
+export const get686AuthorizationState = createStructuredSelector({
```
rianfowler: I'm not sure how the memoization will work here since the full state is being passed into every input selector. Been a while since I've used `reselect` but I think breaking these deep queries into smaller selectors might be a better approach. e.g. `state.user` could be set up as 

`memoizedUserSelector = createStructuredSelector({user: state.user})` 

and reused in 

`selectUserStatus = createSelector(memoizedUserSelector, user => profile.status)`
`selectIsVerified = createSelector(memoizedUserSelector, user => profile.verified`

Not sure about this, though- My understanding is that it memoizes off of the arguments and the root `state` should fail an identity check on every `mapStateToProps` call.

you could easily test this by adding a console log to selector passed into reselect and just calling the selector a couple of times. first with the state and then with `{...state}`. nothing would change identity but the outer state object.

rianfowler: Also recommend not spinning your wheels too much on this unless you're set on getting good at reselect- in case  it works and you want to move on 👍

rianfowler (approved):

src/applications/disability-benefits/686/selectors/index.js
erikphansen: Before I can get good at reselect I first have to get adequate at it :) I'm gonna merge this and save your notes for further review.

PR #8568, opened 2018-09-26
add hostnames files and use buildtype to determine hostname in sitemap

rianfowler:
## Description
add hostnames files and use buildtype to determine hostname in sitemap

## Testing done
verified locally (see screenshots)

## Screenshots
screen shot 2018-09-26 at 17 13 18
screen shot 2018-09-26 at 17 12 11
screen shot 2018-09-26 at 17 11 30
screen shot 2018-09-26 at 17 11 22
screen shot 2018-09-26 at 17 11 13
screen shot 2018-09-26 at 17 11 05

## Definition of done
~- [ ] Events are logged appropriately~
~- [ ] Documentation has been updated, if applicable~

### DIFF
--- a/script/build.js
+++ b/script/build.js
@@ -139,7 +139,6 @@ if (BUILD_OPTIONS.watch) {
   }
 }
 
-// TODO(awong): This URL needs to change based on target environment.
 smith.use(sitemap({
   hostname: BUILD_OPTIONS.hostname,
   omitIndex: true
--- a/script/build.js
+++ b/script/build.js
@@ -17,7 +17,6 @@ const watch = require('metalsmith-watch');
 
 const webpackMetalsmithConnect = require('../config/webpack-metalsmith-connect');
 const environments = require('./constants/environments');
-const hostnames = require('./constants/hostnames');
 const createBuildSettings = require('./create-build-settings');
 const createRedirects = require('./create-redirects');
 const checkBrokenLinks = require('./check-broken-links');
@@ -142,7 +141,7 @@ if (BUILD_OPTIONS.watch) {
 
 // TODO(awong): This URL needs to change based on target environment.
 smith.use(sitemap({
-  hostname: hostnames[BUILD_OPTIONS.buildtype] || 'https://www.vets.gov',
+  hostname: BUILD_OPTIONS.host,
   omitIndex: true
 }));
 
--- a/script/build.js
+++ b/script/build.js
@@ -141,7 +141,7 @@ if (BUILD_OPTIONS.watch) {
 
 // TODO(awong): This URL needs to change based on target environment.
 smith.use(sitemap({
-  hostname: BUILD_OPTIONS.host,
+  hostname: BUILD_OPTIONS.hostname,
   omitIndex: true
 }));
 
--- a/script/build.js
+++ b/script/build.js
@@ -140,7 +140,7 @@ if (BUILD_OPTIONS.watch) {
 }
 
 smith.use(sitemap({
-  hostname: BUILD_OPTIONS.hostname,
+  hostname: BUILD_OPTIONS.host === 'localhost' ? 'http://localhost' : BUILD_OPTIONS.host,
   omitIndex: true
 }));
 
--- a/script/build.js
+++ b/script/build.js
@@ -17,6 +17,7 @@ const watch = require('metalsmith-watch');
 
 const webpackMetalsmithConnect = require('../config/webpack-metalsmith-connect');
 const environments = require('./constants/environments');
+const hostnames = require('./constants/hostnames');
 const createBuildSettings = require('./create-build-settings');
 const createRedirects = require('./create-redirects');
 const checkBrokenLinks = require('./check-broken-links');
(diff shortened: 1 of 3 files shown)


script/build.js
```diff
@@ -141,7 +142,7 @@ if (BUILD_OPTIONS.watch) {
 
 // TODO(awong): This URL needs to change based on target environment.
 smith.use(sitemap({
-  hostname: 'https://www.vets.gov',
+  hostname: hostnames[BUILD_OPTIONS.buildtype] || 'https://www.vets.gov',
```
rianfowler: maybe this should just error out if the buildtype isn't present instead of failing silently

script/constants/hostnames.js
```diff
@@ -0,0 +1,10 @@
+const environments = require('./environments');
+
+module.exports = {
+  [environments.DEVELOPMENT]: 'https://dev.vets.gov',
+  [environments.PREVIEW]: 'https://preview.va.gov',
+  [environments.PRODUCTION]: 'https://www.vets.gov',
+  [environments.STAGING]: 'https://staging.vets.gov',
+  [environments.VAGOVDEV]: 'https://dev.va.gov',
+  [environments.VAGOVSTAGING]: 'https://staging.va.gov'
```
cehsu: Should existing references to hostname [magic strings](https://github.com/department-of-veterans-affairs/vets-website/blob/7fb24d3df71cef7a029d62b9abda0946e73f1915/src/applications/personalization/profile360/vet360/util/local-vet360.js#L6-L7)/[constants](https://github.com/department-of-veterans-affairs/vets-website/blob/2e282e7d54c6b3ee70fb2c9d16138250bc9f2346/src/platform/brand-consolidation/index.js#L3-L4) now make use of these instead?

cehsu (approved):

ncksllvn (approved):
LGTM!

script/build.js
ncksllvn: I just realized that since `host` is an optional build argument, it might make sense to derive this property in `options.js` and use its default

script/constants/hostnames.js
jbalboni: I asked @ncksllvn about this yesterday, I'm not sure if we have a good place to put constants that can be shared between the build scripts and the application code.

jbalboni (commented):
Looks good to me, but I agree with Nick's comment about incorporating options.js and the host name option that you can pass.

script/build.js
```diff
@@ -141,7 +141,7 @@ if (BUILD_OPTIONS.watch) {
 
 // TODO(awong): This URL needs to change based on target environment.
```
erikphansen: can this comment be removed now?

script/options.js
```diff
@@ -49,6 +51,10 @@ function applyDefaultOptions(options) {
 function applyEnvironmentOverrides(options) {
   const env = require('get-env')();
 
+  if (options.buildtype) {
+    options.hostname = hostnames[options.buildtype];
```
rianfowler: @ncksllvn @jbalboni is this what you're thinking?

script/options.js
```diff
@@ -14,6 +15,7 @@ const COMMAND_LINE_OPTIONS_DEFINITIONS = [
   { name: 'entry', type: String, defaultValue: null },
   { name: 'analyzer', type: Boolean, defaultValue: false },
   { name: 'host', type: String, defaultValue: 'localhost' },
+  { name: 'hostname', type: String, defaultValue: 'localhost' },
   { name: 'public', type: String, defaultValue: null },
```
ncksllvn: Actually I think we can make the change to the `host` property above and work with that.

It seems like if a `host` is explicitly passed, it should use that as the `hostname`. But if not, then it should derive it from the build-type.  If neither is passed, maybe assume `localhost`?

script/options.js
```diff
@@ -49,6 +52,15 @@ function applyDefaultOptions(options) {
 function applyEnvironmentOverrides(options) {
   const env = require('get-env')();
 
+  // priority order: command line option, watch task host (localhost), build type host, default host
+  if (!options.host) {
```
rianfowler: @ncksllvn the watch task gets a build type argument which causes it to try to run on `http://dev.vets.gov`. Added a check for `options.watch` that will force it to use `localhost`. 

Removed the default value from `options.host` to make it easy to determine if there's a command line argument set (didn't see anything in the docs that offered this and I thought filtering the array would obfuscate this). 

Not a fan of nested ifs so I'm open to another approach.

script/build.js
```diff
@@ -139,9 +139,8 @@ if (BUILD_OPTIONS.watch) {
   }
 }
 
-// TODO(awong): This URL needs to change based on target environment.
 smith.use(sitemap({
-  hostname: 'https://www.vets.gov',
+  hostname: BUILD_OPTIONS.host === 'localhost' ? 'http://localhost' : BUILD_OPTIONS.host,
```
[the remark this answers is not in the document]
rianfowler: Could leave this or just skip the site map generation for the `watch` task. Could see that being useful for local verification.

PR #8569, opened 2018-09-27
Final Webpack/Metalsmith separations

ncksllvn:
## Description
This PR finalizes the work for separating Webpack from Metalsmith, so that the content can be built in a separate build cycle without touching Webpack-compiled code, or any of the files that will belong to a future app-code repo.

Work includes -
- Moves `static-page-widgets.js` to Metalsmith assets. Because it's printed directly in the body of content pages, it would be required on every content build, so we need it in a Metalsmith-accessible directory.
- Update some Webpack references to use absolute paths instead of relative, which I found to be a necessary if they're going to be executed outside of the `vets-website` project root.

## Testing done
1. Added a clone of `vets-website` into a directory, `vagov-apps`, into a `vets-website` project.
2. From the parent, deleted the `src` and `config` as these pertain only to the app-code
3. In the parent's `script/build.js`, updated the [webpack-middleware import](https://github.com/department-of-veterans-affairs/vets-website/blob/master/script/build.js#L17) to reference the child's instead:
    - `const webpackMetalsmithConnect = require('../vagov-apps/config/webpack-metalsmith-connect');`
4. Run a build
5. 🤞 
6. Build works!

In the example above, the parent repo represents the content repo, while the child repo (`vagov-apps` directory) represents the app-code repo. The child repo is able to handle the Webpack compilation, while the parent handles all of the Metalsmith compilation. 

Side note - a lot of other files can be removed from the parent, while all of the content files and `build.js` can be removed from the child. This wasn't important in testing, but it does help illustrate the relationship of content vs. app-code.

This means that in our build script, we are freed up to implement something like -

```
# Check flag to see if we should build the app-code (more on this later)
# if flag is true:
  # const webpackMetalsmithConnect = require('../vagov-apps/config/webpack-metalsmith-connect');
  # compile the code, or do the watch task for local dev
# if false:
  # move on, without ever needing access to any Webpack code
  # use a git SHA or some other data to link up to precompiled app-code
```

## Acceptance criteria
- [ ] We can easily separate the app-code compilation from the content compilation

## Definition of done


content/includes/footer.html
```diff
@@ -140,7 +140,7 @@ <h3>Get help from Veterans Crisis Line</h3>
 (function() {
   var widgets = {{ widgets | json }};
   var module = {};
-  {% include "src/applications/static-pages/static-page-widgets.js" %}
+  {% include "assets/js/static-page-widgets.js" %}
   mountWidgets(widgets, '{{ buildtype }}' === 'production');
```
ncksllvn: I didn't want to have to move this, but it's a necessary step because after a full separation of the Metalsmith content from app-code, the Metalsmith build has to be able build content alone without accessing files within app-code.

config/webpack.config.js
```diff
@@ -10,13 +10,15 @@ require('babel-polyfill');
...
-  style: './src/platform/site-wide/sass/style.scss',
-  polyfills: './src/platform/polyfills/preESModulesPolyfills.js',
-  brandConsolidation: './src/platform/site-wide/sass/brand-consolidation.scss',
+  styleConsolidated: getAbsolutePath('src/platform/site-wide/sass/style-consolidated.scss'),
+  style: getAbsolutePath('src/platform/site-wide/sass/style.scss'),
+  polyfills: getAbsolutePath('src/platform/polyfills/preESModulesPolyfills.js'),
+  brandConsolidation: getAbsolutePath('src/platform/site-wide/sass/brand-consolidation.scss'),
   vendor: [
-    './src/platform/polyfills',
+    getAbsolutePath('src/platform/polyfills'),
```
ncksllvn: These changes are necessary so that Webpack can build the files even if it's not being executed from the `vets-website` project root.

config/manifest-helpers.js
```diff
@@ -9,7 +9,7 @@ function getAppManifests(root) {
       const manifest = require(file);
 
       manifest.filePath = file;
-      manifest.entryFile = path.relative(root, path.join(path.dirname(file), manifest.entryFile));
+      manifest.entryFile = path.resolve(root, path.join(path.dirname(file), manifest.entryFile));
```
ncksllvn: I found that we need an absolute path instead of relative if we're executing from outside of the project root.

src/platform/testing/e2e/test-server.js
```diff
@@ -28,7 +29,13 @@ if (options.unexpected && options.unexpected.length !== 0) {
...
 const app = express();
 
-const root = path.resolve(__dirname, `../../../../build/${options.buildtype}`);
+let root = path.resolve(__dirname, `../../../../build/${options.buildtype}`);
+if (!fs.existsSync(root)) {
+  // if there isn't a build directory here, then check the parent directory.
+  // This is a temporary adapation as we transition to vagov-content.
+  root = path.resolve(__dirname, `../../../../../build/${options.buildtype}`);
+}
+
```
ncksllvn: The test-server for E2E tests is a really weird case. It needs access to the generated files from a build in order to serve them, but those files should really only belong to content, so it's a weird relationship. I would like to see a way to more clearly define that relationship as we move forward, but it's not something that's priority at the moment.

I also wanted to leave the script files untouched between parent/child repos and not send any special args from the parent, because I felt that the child should have ownership of the script file because it's largely for testing app-code.

config/webpack.config.js
jbalboni: I'm not sure I understand this. I don't think we would separate the code from the build scripting required to build it, would we? Wouldn't the content build process grab assets that were already built if we were separating them?

ncksllvn: Before trying to explain, I want to preface this by saying that this has no effect on the current build. And we definitely need a spot for code that is safe for Metalsmith and Webpack to share.

But anyway, I hope I can answer your question and not just restate things you already knew. Let's definitely work together on this tomorrow if we can. I still have to create some sort of visual for this, but the mental model is something like this -

The content will grab assets that were already built during "CMS" deployments. But, during "full" deployments, it will need to execute the app-code's compilation as well, before following the rest of the CMS model as usual. In theory, we *could* have the app-code build totally separate from content build and then execute them individually with the content linking up after the app-code, but we also need a direct bridge between them for local development, and since they're developed directly for one another, it makes sense in my head to keep `script/build.js` as the central point of execution. However, even though they have that close relationship, the two separate deployment cycles - CMS and full - require a modular design.

There's also a some other concerns here -
- A CMS deployment is triggered any time a content-change is merged to master. What a "content-change" really means, and what effect intermediary code-merges could have, could be difficult to determine without a clear separation between files owned by the CMS versus the app-code.
- In version control, it would be valuable to preserve app-code versions in each commit from the content, so that we can easily checkout an earlier version of the site with the app-code exactly as it stood at that time. Kinda worried that information will only be visible on S3 logs, or by looking up an app-code commit by checking the dates.

No idea if I answered your question, but hope that helped in some way. This work is actually really close to being done. The next part is just incorporating the CMS deployment into the build, which I don't expect to be much.

jbalboni (approved):
These specific changes look good, though I think I need to get a better understanding of how we want to split app code and content.

PR #8571, opened 2018-09-27
Prettier ESLint rules

erikphansen:
## Description
This is a preview of a PR that:
- adds Prettier's ESLint rules (using [this method](https://prettier.io/docs/en/eslint.html#why-not-both))
- runs all the code through the new ESLint rules with `--fix` turned on

## The Plan
I'd like to get feedback on this over the next couple of days and barring any major objections or requests for changes to the rules, I'd like to create a new PR over the weekend that converts the current state of `master` to be compliant with the new rules. Perhaps that's a little optimistic.

### This PR is composed of two major changes

There are three commits in this PR to make it a little easier to review the code. The first two commits cover the changes to the ESLint rules, plus some new scripts added to `package.json`. The third commit is the massive "boil the ocean" reformatting PR.

#### Changes to the ESLint rules:
These changes involve setting up Prettier's ESLint rules, removing/updating existing rules that conflicted with Prettier, and reorganizing some existing rules. A [`.prettierrc` file](https://prettier.io/docs/en/configuration.html) has also been added with mostly default rules with some adjustments to account for our existing style. [The full list of options is here](https://prettier.io/docs/en/options.html).

#### Changes to the codebase to make the code compliant with the new ESLint rules:
99% of the code changes were done by running the code through ESLint with the `--fix` flag turned on. But that either created some new issues that needed to be resolved or was unable to resolve a few issues. The remaining issues were resolved manually. A big chunk of those changes were simply changing some `eslint-ignore-line` comments to `eslint-ignore-next-line` and repositioning them.

## Moving Forward
When these new rules are implemented there shouldn't be major changes to your workflow since we are already using ESLint. Simply continue using ESLint how you have been.

But...

If you are not currently using your editor's "fix auto-fixable ESLint errors" command — or some other method of auto-fixing ESLint errors — you might want to consider doing so. One of the rules introduced with Prettier is a line length limit. Dealing with this limit is not a problem at all if you allow the "error" to be fixed for you; it's wonderful not having to format your code! But it could be tedious and annoying to manually insert line breaks and fix indentation until ESLint is happy. Personally I have VSCode to auto-fix ESLint errors on save. This is almost entirely a wonderful experience. The one caveat to that is when I compulsively hit ⌘-S and ESLint/Prettier converts my half-written JSX tags to self-closing tags.

## Testing done
Unit tests and e2e tests


config/manifest-helpers.js
```diff
@@ -3,13 +3,17 @@ const find = require('find');
 const path = require('path');
 
 function getAppManifests(root) {
-  return find.fileSync(/manifest\.(json|js)$/, path.join(root, './src/applications'))
+  return find
+    .fileSync(/manifest\.(json|js)$/, path.join(root, './src/applications'))
```
rianfowler: I like this

config/nightwatch-globals-sauce.js
```diff
@@ -16,15 +16,19 @@ module.exports = {
       return item.failed === 0 && item.errors === 0;
     }
 
-    const testCaseData = Object.keys(browser.currentTest.results.testcases).map((key) => {
-      return browser.currentTest.results.testcases[key];
-    });
+    const testCaseData = Object.keys(browser.currentTest.results.testcases).map(
+      key => browser.currentTest.results.testcases[key],
```
rianfowler: This is good.

config/nightwatch-globals-sauce.js
```diff
@@ -16,15 +16,19 @@ module.exports = {
...
+    );
 
     const passed = testCaseData.every(noFailures);
 
-    saucelabs.updateJob(sessionid, {
-      passed,
-      name: jobName,
-    }, done);
+    saucelabs.updateJob(
+      sessionid,
```
rianfowler: This is a lot easier to edit now.

src/applications/auth/containers/AuthApp.jsx
```diff
@@ -133,26 +168,27 @@ export class AuthApp extends React.Component {
...
+      />
     );
-  }
+  };
 
   render() {
-    const view = this.state.error ?
-      this.renderError() :
-      <LoadingIndicator message="Signing in to Vets.gov..."/>;
+    const view = this.state.error ? (
```
rianfowler: I'm a fan of always enclosing (<jsx>)

src/applications/beta-enrollment/containers/BetaEnrollmentButton.jsx
```diff
@@ -21,36 +21,53 @@ class BetaEnrollmentButton extends React.Component {
...
   render() {
     if (this.props.isUserRegisteredForBeta) this.onRegistered();
 
     if (this.state.hasError) {
       return (
-        <AlertBox status="error"
+        <AlertBox
+          status="error"
           isVisible={this.state.hasError}
-          content={<div><h3>We can't turn on the beta tools</h3><p>We're sorry. Something went wrong on our end, and we can't turn on the beta tools for you. Please try again later.</p></div>}/>
```
rianfowler: very yes

src/applications/messaging/components/compose/MessageSubjectGroup.jsx
```diff
@@ -33,14 +32,16 @@ class MessageSubjectGroup extends React.Component {
...
+          />
           <MessageSubject
             charMax={this.props.charMax}
             cssClass="msg-subject"
             onValueChange={this.props.onSubjectChange}
             placeholder={this.props.subjectPlaceholder}
             required={this.props.subjectRequired}
-            subject={this.props.subject}/>
+            subject={this.props.subject}
+          />
```
rianfowler: yes

rianfowler (approved):
bring on the merge conflicts!

erikphansen:
Nope. Gonna go through the `eslint --fix` exercise again on a fresh branch off of `master` :)

.eslintrc
```diff
@@ -14,53 +18,35 @@
...
       2,
       { "args": "after-used", "argsIgnorePattern": "^_", "vars": "local" }
     ],
     "prefer-rest-params": 2,
-    "quote-props": [2, "as-needed", { "keywords": true }],
-    "space-unary-ops": 2,
-    "max-len": [0], // No agreed upon max length and react lines get > 100 chars easily.
-    "arrow-parens": 0, // We don't have an agreed upon pattern here
-    "no-underscore-dangle": 0, // We have build flags that use this
+    "quote-props": [2, "as-needed", { "keywords": false }],
```
cehsu: What's the thinking behind changing this? I've encountered the keyword error a few times. Thanks for any insights.

config/manifest-helpers.js
```diff
@@ -25,17 +29,16 @@ function getWebpackEntryPoints(manifests) {
...
 function displayApplications() {
   const root = path.join(__dirname, '../..');
-  getAppManifests(root)
-    .forEach(manifest => {
-      console.log(manifest.appName);
-      console.log(path.relative(root, manifest.filePath));
-      console.log(`Production: ${manifest.production || false}`);
-      console.log('');
-    });
+  getAppManifests(root).forEach(manifest => {
```
cehsu: Much clearer, does it enforce new lines for chained calls but not for individual calls? If so, that's a nice distinction.

config/webpack-metalsmith-connect/convert-paths-to-relative.js
```diff
@@ -1,6 +1,6 @@
 function convertAbsolutePathsToRelative(buildOptions) {
   return (files, metalsmith, done) => {
-    Object.keys(files).forEach((file) => {
+    Object.keys(files).forEach(file => {
```
cehsu: yay!

cehsu (approved):
LGTM. This seems like it will reduce a lot of subtle decision fatigue. Could you please set up another PR to update the [tooling documentation](https://github.com/department-of-veterans-affairs/vets.gov-team/blob/635f35ad555d0cdac50667ac48c92fef2b719c64/Work%20Practices/Engineering/Style%20Guides/Javascript/vets.gov%20Front%20End.md#existing-guides-and-tools) to include a mention of Prettier, this could simply be some of the description you've included for this PR.

package.json
```diff
@@ -155,6 +159,7 @@
     "node-sass": "^4.5.3",
     "null-loader": "^0.1.1",
     "polyfill-function-prototype-bind": "0.0.1",
+    "prettier": "^1.14.3",
```
cehsu: Could you please also include the yarn.lock updates?

ncksllvn:
I'm not necessarily against this, but I do like engineers having some level of flexibility in the way they style their code. I don't know; just a personal preference.

.eslintrc
erikphansen: For some time now it's been totally valid to use reserved words as unquoted property names. So no need to put `enum` in quotes, for example. https://stackoverflow.com/questions/4298594/javascript-keywords-in-hash-keys

config/manifest-helpers.js
erikphansen: I'm sure someone smarter than me could explain how it determines were to wrap lines and where to keep statements on the same line :( I just like not having to think about it.

package.json
erikphansen: I'm not sure why the yarn.lock didn't get updated. Weird. Will look into that.

jbalboni:
There's some research that suggests having a consistent coding style that reduces the flexibility programmers have to style code improves code readability: https://stackoverflow.com/questions/1325374/research-into-advantages-of-having-a-standard-coding-style

erikphansen:
How does that SO answer only have 14 upvotes?! Although familiar with the concept of chunking (and that specific chess-master example) I had never thought of consistent code formatting aiding that. That totally makes sense, though. Of course that means there's a learning curve if the style rules are a major departure from what you are used to. But I don't think that's the case here. I hope. Though I really did want to drop semicolons, to be honest :)

One thing that thread doesn't mention is the huge benefit of having a tool doing the formatting for you.

ncksllvn:
@erikphansen No,I just checked the diff and by the extent of changes it seemed strict to me. No specific complaints really, although TBH I don't really like that trailing comma that follows the last property when defining an object, but I might be alone in that lol.

@jbalboni Yeah that's a cool read. Really great points. Would probably also look good for anyone that's just perusing the repo.

jbalboni:
I don't like the trailing comma either, though it does make diffs easier to read sometimes. I think this is one of the few settings in Prettier, though, so maybe we can turn it off depending on what others think?

erikphansen:
I used to hate it because it looks so stupid, but I started using it in April and love love LOVE it now. I'd fight really hard to keep it, honestly. Not if I'm the only one, of course, but I've come to really appreciate its utility despite its ugliness. Now whenever I see a line in a diff and I see that the only change was adding or removing a comma I die inside a little bit.

End of sales pitch :)

config/webpack-metalsmith-connect/metalsmith-webpack.js
```diff
@@ -14,17 +14,18 @@ const path = require('path');
...
 }
 
 function normalizePath(fullPath) {
-  return fullPath.split(path.sep).filter((p) => {
-    return typeof p === 'string' && p.length > 0;
-  }).join('/');
+  return fullPath
+    .split(path.sep)
+    .filter(p => typeof p === 'string' && p.length > 0)
+    .join('/');
```
cvalarida: Yup, this is nice.

config/webpack-metalsmith-connect/metalsmith-webpack.js
```diff
@@ -88,7 +89,11 @@ function webpackDevServerPlugin(config, opts) {
     server.listen(options.port || 8081, options.host, () => {
       console.log(
         chalk.blue('[metalsmith-webpack-dev-server]: ') +
-        chalk.green(`Running webpack dev server at http://${options.host}:${options.port}`)
+          chalk.green(
+            `Running webpack dev server at http://${options.host}:${
+              options.port
+            }`,
```
cvalarida: I'm not a fan of this, though. Why did this get split to new lines? Line length limit?

script/options.js
```diff
@@ -57,7 +58,11 @@ function applyEnvironmentOverrides(options) {
     case environments.PRODUCTION:
       if (options['no-sanity-check-node-env'] === false) {
         if (env !== 'prod') {
-          throw new Error(`buildtype ${options.buildtype} expects NODE_ENV to be production, not '${process.env.NODE_ENV}'`);
+          throw new Error(
+            `buildtype ${
+              options.buildtype
+            } expects NODE_ENV to be production, not '${process.env.NODE_ENV}'`,
```
cvalarida: Is there any way to disable whatever it is that does this? It seems...wrong to split strings like this.

src/applications/auth/containers/AuthApp.jsx
```diff
@@ -75,43 +83,65 @@ export class AuthApp extends React.Component {
...
+                We’re sorry. It looks like you selected "Deny" on the last page
+                when asked for your permission to share information with
+                Vets.gov, so we couldn’t complete the process. To give you full
+                access to the tools on Vets.gov, we need to be able to share
+                your information with the site.
+              </p>
+              <p>
+                Please try again and click “Accept” on the final page. Or, you
+                can try signing in with your premium DS Logon or premium
+                MyHealtheVet account instead of identity proofing with ID.me.
```
cvalarida: Oh nice. Definitely a fan of this.

src/applications/beta-enrollment/containers/BetaEnrollmentButton.jsx
cvalarida: :point_up:

cvalarida (approved):
I definitely didn't go through all 56k additions and 35k subtractions, but it looks good. I think my only gripe is about splitting up the string templates on variable insertions to accommodate line length. I'm also not a _fan_ of the trailing comma, but I can live with it.

script/options.js
erikphansen: I'll look into this. It might be possible if we change the parser? But actually since we are using ESLint to parse, rather than standalone Prettier, that probably isn't a real option.

I'll dig a bit and see what I can find but I don't recall this being one of the Prettier options.

rianfowler:
I think trailing commas make things easier to alphabetize. Can prettier alphabetize deconstructed objects and arguments :trollface: ?

script/options.js
erikphansen: @cvalarida [You're not the only one who feels this way](https://github.com/prettier/prettier/issues/3368). But there's not prettier config setting to disable it. It won't break lines at [*all* expressions](https://github.com/prettier/prettier/issues/3368#issuecomment-379310998), though.

erikphansen: There's always `// eslint-disable-next-line prettier/prettier` (doesn't seem to be a way to disable just the `"printWidth"` rule.

cvalarida: Small price to pay, I suppose.

PR #8572, opened 2018-09-27
Update forms links, add feedback tool page

jbalboni:
## Description

This updates the profile form links to use manifest files, since some of them are in manifest.js files that could have different urls in WBC vs vets.gov. This isn't the case right now, but it'll give us some security if we do decide to change urls later.

I've also reverted the vic v2 manifest, since it has no content page in va-gov and it appears to be unlikely to start up again. That app will no longer be up when brand consolidation goes live.

Also, I found that the feedback tool form was missing a content page in va-gov, which is going to be necessary soon.

## Testing done
Checked dashboard locally, resumed a few saved applications

## Acceptance criteria
- [x] Forms can be resumed from dashboard
- [x] Feedback tool works with brand consolidation enabled

## Definition of done


src/applications/vic-v2/manifest.json
```diff
@@ -0,0 +1,6 @@
+{
+  "appName": "Veteran ID Card V2",
+  "entryFile": "./veteran-id-card-entry.jsx",
+  "entryName": "vic-v2",
+  "rootUrl": "/veteran-id-card/apply"
+}
```
ncksllvn: Isn't this one changing to [`/records/get-veteran-id-cards/`](https://docs.google.com/spreadsheets/d/1YucOtYihweNPI1gYtCgdsinKQmH4DMuSgPaSgJMfyYk/edit#gid=316089358)?

ncksllvn (approved):
LGTM!

src/applications/vic-v2/manifest.json
jbalboni: No, all of the content pages are for VIC v1, which is a different app than this (and isn't a form on our side).

jbalboni: This page doesn't exist in the WBC content and I'm not adding an equivalent because VIC v2 is on hiatus for the foreseeable future.

ncksllvn: Ah gotcha, thank you!

erikphansen (approved):

cehsu (approved):
LGTM. There is actually another issue that this closes as well: https://github.com/department-of-veterans-affairs/vets.gov-team/issues/12576

src/applications/personalization/dashboard/helpers.jsx
```diff
@@ -1,4 +1,18 @@
...
+import edu1990Manifest from '../../edu-benefits/1990/manifest.json';
+import edu1995Manifest from '../../edu-benefits/1995/manifest.json';
+import edu1990eManifest from '../../edu-benefits/1990e/manifest.json';
+import edu1990nManifest from '../../edu-benefits/1990n/manifest.json';
+import edu5490Manifest from '../../edu-benefits/5490/manifest.json';
+import edu5495Manifest from '../../edu-benefits/5495/manifest.json';
+import edu0993Manifest from '../../edu-benefits/0993/manifest.json';
+import preneedManifest from '../../pre-need/manifest.json';
+import pensionManifest from '../../pensions/manifest.json';
+import disability526Manifest from '../../disability-benefits/526EZ/manifest.json';
```
cehsu: Should we also plan to move the form benefits and form titles into the manifests as well, as part of a separate PR?

jbalboni: @cehsu Yeah, I think that's a good idea.

rianfowler (approved):

erikphansen (approved):

config/manifest-loader.js
```diff
@@ -7,8 +7,8 @@ function manifestLoader(source) {
   return `
     ${source}
     (function() {
-      if (module.exports.receiveContentProps) {
-        const applicationSettings = window.settings.applications[module.exports.entryName];
+      const applicationSettings = window.settings.applications[module.exports.entryName];
```
jbalboni: @ncksllvn Does this change make sense to you? The dashboard code imports manifests for some forms that aren't in production, which means they don't have settings values. This skips running `receiveContentProps` when that is the case. Should be pretty rare outside of applications with cross-cutting features like the dashboard.

ncksllvn: Ah that is a tricky case. That does make sense to me! Appreciate the heads up on that.

PR #8573, opened 2018-09-27
Fix for the watch-task crashing while reloading content

ncksllvn:
## Description
This PR fixes an issue with the webpack-metalsmith-middleware.  The problem occurred from the middleware attempting to reinitialize the WebpackDevServer plugin inside the middleware chain, when that should only be done once. 

## Testing done
1. `npm run watch`
2. Edit file in content
3. Save the content file
4. 🤞 
5. Doesn't crash

## Acceptance criteria
- [x] The watch task runs without crashing during a content change

## Definition of done

### DIFF
--- a/config/webpack-metalsmith-connect/index.js
+++ b/config/webpack-metalsmith-connect/index.js
@@ -28,13 +28,18 @@ function getEntryPoints(buildOptions) {
 }
 
 function compileAssets(buildOptions) {
-  return (files, metalsmith, done) => {
-    const convertPathsMiddleware = convertPathsToRelative(buildOptions);
-    const apps = getEntryPoints(buildOptions);
-    const webpackConfig = generateWebpackConfig(buildOptions, apps);
-    const webpackMiddleware = webpackPlugin(webpackConfig);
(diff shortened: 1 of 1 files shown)


config/webpack-metalsmith-connect/index.js
```diff
@@ -28,28 +28,38 @@ function getEntryPoints(buildOptions) {
...
+      convertPathsMiddleware = convertPathsToRelative(buildOptions);
+    }
 
-    webpackMiddleware(files, metalsmith, (err) => {
+    compileMiddleware(files, metalsmith, (err) => {
       if (err) throw err;
       convertPathsMiddleware(files, metalsmith, done);
     });
   };
 }
```
ncksllvn: The reason this is in here in the first place is that we pass information from Metalsmith to Webpack via `buildOptions`, and some of that information is applied during previous middleware. If we don't initialize inside of the middleware function, then certain properties won't exist yet.

cehsu (approved):

cvalarida (approved):

PR #8575, opened 2018-09-27
fix intermittently failing test

rianfowler:
## Description
test was intermittently failing: http://jenkins.vetsgov-internal/job/testing/job/vets-website/job/master/2581/testReport/

## Testing done
ran `CHOMA_SEED=SoKRhC4WEc npm run test:unit`
5 runs before change: 1 failure 
10 runs after change: 0 failures

### DIFF
--- a/src/applications/disability-benefits/all-claims/tests/utils.unit.spec.jsx
+++ b/src/applications/disability-benefits/all-claims/tests/utils.unit.spec.jsx
@@ -165,6 +165,7 @@ describe('526 helpers', () => {
     beforeEach(() => {
       // Replace fetch with a spy
       global.fetch = sinon.stub();
+      global.fetch.catch = sinon.stub();
       global.fetch.resolves({
         ok: true,
         headers: { get: () => 'application/json' },


src/applications/disability-benefits/all-claims/tests/utils.unit.spec.jsx
```diff
@@ -165,6 +165,7 @@ describe('526 helpers', () => {
     beforeEach(() => {
       // Replace fetch with a spy
       global.fetch = sinon.stub();
+      global.fetch.catch = sinon.stub();
```
rianfowler: I think as a practice we should inject `apiRequest` into our functions instead of calling it directly from an import e.g. `function aFetchingFunction(someArgument, doFetch = apiRequest) { ... }`. Makes this sort of thing easier to test by removing the need to resolve promises in the unit test.

erikphansen: I love this idea.

erikphansen (approved):

jbalboni (approved):

PR #8595, opened 2018-09-28
additional z index fixes

bshyong:
## Description
Adds z-index fix for modals on injected header

## Testing done
Local testing

## Acceptance criteria
- [x] Login modal, when open, should appear over existing in-page elements

## Definition of done
- [ ] A link has been provided to the originating GitHub issue (or connected to it via ZenHub)
- [x] No sensitive information (i.e. PII/credentials/internal URLs/etc.) is captured in logging, hardcoded, or specs

Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/13800

PR #8596, opened 2018-09-28
bump up z-indexes of overlays and dropdowns

rianfowler:
## Description
screen shot 2018-09-28 at 15 36 13

## Screenshots
screen shot 2018-09-28 at 15 35 53
screen shot 2018-09-28 at 15 35 47

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/src/platform/site-wide/sass/merger.scss
+++ b/src/platform/site-wide/sass/merger.scss
@@ -10,6 +10,10 @@
     .sign-in-nav {
       margin-left: 6px;
     }
+    z-index: 300;
+  }
+  .va-modal {
+    z-index: 301;
   }
 
   #dashboard-title {


rianfowler:
Should we name these z-index values and use that instead of these arbitrary numbers? I think we're going to be updating the z-indexes a bit to deal with the team site stuff and it might be easier if we abstract the it.

erikphansen (commented):
It's not possible to just move the offending nav bar down without also adjusting the modal?

ncksllvn:
@rianfowler that sounds like an awesome idea. There are z-index changes in this PR too, https://github.com/department-of-veterans-affairs/vets-website/pull/8595#pullrequestreview-160026670. Does this impact the work here?

ncksllvn (approved):

ncksllvn:
@bshyong @rianfowler @accbjt @RobAWilkinson @jbalboni  Looks like we have a few elements that may wind up fighting over the z-index. I really like @rianfowler's suggestion of naming these values. How we can get that pattern started?

rianfowler:
@erikphansen We're pushing our header onto a site which has styling we don't control- this is why we're using such high z-index values. For [example](https://dev.va.gov/health/). I think we'll endeavor to keep the values as low as we can.

cehsu (approved):
LGTM.

erikphansen (approved):

PR #8597, opened 2018-09-28
526 Messages

cvalarida:
## Description
A few messages needed changing.

@peggygannon I didn't update the last message in the ticket because it doesn't quite reflect what's happening. What we currently have is this:

**We’re sorry. It looks like we’re missing some information needed for your application**
For help with your application, please call Veterans Benefits Assistance at 1-800-827-1000, Monday – Friday, 8:00 a.m. to 9:00 p.m. (ET).

That doesn't have _what_ we're missing, though, so maybe that's important? At any rate, the main difference is that we'll block them if they don't have a ~~name, date of birth, SSN, and file number~~ EDIPI, SSN, and Participant ID (aka Corp ID). (I think; @kreek can you confirm? This is the evss_claims back end service being sent in the profile.)

Without this information, we won't be able to get their banking information or rated disabilities, but that's not why we stopped them with that message.

## Testing done
Manually tested what I could. I don't know how to test an unverified user, though, so I just changed the message that was already there and :crossed_fingers:. :grimacing:


kreek:
@cvalarida you need edipi, ssn, and participant id (also called corp id)

peggygannon (approved):
@cvalarida Looks good. Can I see screenshots when you have a moment? the content looks good, but it's also helpful for me to see the screenshots.

src/applications/disability-benefits/526EZ/containers/DisabilityGate.jsx
```diff
@@ -38,7 +38,7 @@ export const DisabilityGate = ({ prefillStatus, disabilities = [], children }) =
         <AlertBox
           isVisible
           headline="We're sorry"
-          content="We couldn't find any rated disabilites at this time."
+          content="Our records show you don’t have any rated disabilities, so you can’t apply for increased compensation benefits at this time. If you think our records aren’t correct, please call Veterans Benefits Assistance at 1-800-827-1000, Monday – Friday, 8:00 a.m. to 9:00 p.m. (ET)."
```
markgreenburg: Do you want to use a tel link for the vba number? I believe this is what we do elsewhere.

src/applications/disability-benefits/526EZ/tests/components/FormStartControls.unit.spec.jsx
```diff
@@ -47,7 +47,7 @@ describe('526 <FormStartControls>', () => {
         }}/>
     );
 
-    expect(tree.find('.usa-alert').text()).to.contain('To apply for a disability increase, you’ll need to verify your account');
+    expect(tree.find('.usa-alert').text()).to.contain('If you have a premium DS Logon or My HealtheVet account');
```
markgreenburg: Do the changes in DisabilityGate not produce any test errors? If that's the case, do we need to add any tests for those changes?

markgreenburg (approved):
Couple of suggestions but nothing blocking!

PR #8622, opened 2018-10-02
Add script to update the vets-json-schema

erikphansen:
- Updates package.json to use the latest version of the vets-json-schema
- Runs `yarn` to install the new version

### DIFF
--- a/script/update-json-schema.sh
+++ b/script/update-json-schema.sh
@@ -1,3 +1,5 @@
+#!/bin/bash
+
 NEW_HASH=$(git ls-remote git://github.com/department-of-veterans-affairs/vets-json-schema/ refs/heads/master | awk '{print $1}')
 sed -i -e 's/\(vets-json-schema\.git#\)\(.*\)\(",$\)/\1'$NEW_HASH'\3/' package.json
 rm package.json-e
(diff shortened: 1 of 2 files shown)


script/update-json-schema.sh
```diff
@@ -0,0 +1,4 @@
+NEW_HASH=$(git ls-remote git://github.com/department-of-veterans-affairs/vets-json-schema/ refs/heads/master | awk '{print $1}')
+sed -i -e 's/\(vets-json-schema\.git#\)\(.*\)\(",$\)/\1'$NEW_HASH'\3/' package.json
+rm package.json-e
```
erikphansen: The sed command creates a `package.json-e` file that then needs to be cleaned up. I'm not sure why this file gets created :-/ Anyone know how to prevent that so the cleanup isn't needed?

cehsu: That file isn't created for me locally, it seems like something set up in your environment is causing this.

cehsu (approved):
LGTM. Thanks for setting this up. The `rm package.json-e` line shouldn't be necessary.

script/update-json-schema.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+NEW_HASH=$(git ls-remote git://github.com/department-of-veterans-affairs/vets-json-schema/ refs/heads/master | awk '{print $1}')
+sed -i -e 's/\(vets-json-schema\.git#\)\(.*\)\(",$\)/\1'$NEW_HASH'\3/' package.json
+rm package.json-e
+yarn
```
cehsu: Is this meant to be `yarn install`?

rianfowler (approved):

script/update-json-schema.sh
erikphansen: Weird. Adding an empty string after the `-i` flag prevents it from happening on my machine. I'm not sure why it acts differently for us.

jbalboni (approved):

PR #8637, opened 2018-10-03
conditionally remove the facility code before submit

erikphansen:
## Description
Send the facilityCode to the backend. But only send it if the school was selected with the search tool.

TODO:
- [x] Add unit tests if this is a decent approach for this problem.

## Testing done
Tested locally. Confirmed the JSON payload when entering the school manually and when entering it with the search tool. Confirmed that after picking a school with the search tool and then switching to manual entry, the facility code was still not sent.

## Acceptance criteria
- [ ]

## Definition of done
~~- [ ] Events are logged appropriately~~
~~- [ ] Documentation has been updated, if applicable~~

### DIFF
--- a/src/applications/edu-benefits/feedback-tool/config/form.js
+++ b/src/applications/edu-benefits/feedback-tool/config/form.js
@@ -59,7 +59,11 @@ const {
 } = fullSchema.properties;
 
 const { assistance, programs, school } = educationDetails.properties;
-const { address: schoolAddress, name: schoolName } = school.properties;
+const {
+  address: schoolAddress,
+  name: schoolName,
+  facilityCode,
+} = school.properties;
 const domesticSchoolAddress = schoolAddress.anyOf[0];
 const internationalSchoolAddress = schoolAddress.anyOf[1];
 const countries = domesticSchoolAddress.properties.country.enum.concat(
@@ -464,7 +468,7 @@ const formConfig = {
             educationDetails: {
               school: {
                 'view:searchSchoolSelect': {
-                  'view:facilityCode': {
+                  facilityCode: {
                     'ui:required': manualSchoolEntryIsNotChecked,
                   },
                   'ui:field': SchoolSelectField,
@@ -535,9 +539,7 @@ const formConfig = {
                       'view:searchSchoolSelect': {
                         type: 'object',
                         properties: {
-                          'view:facilityCode': {
-                            type: 'string',
-                          },
+                          facilityCode,
                         },
                       },
                       'view:manualSchoolEntry': {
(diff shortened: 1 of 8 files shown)


src/applications/edu-benefits/feedback-tool/config/form.js
```diff
@@ -59,7 +59,11 @@ const {
 } = fullSchema.properties;
 
 const { assistance, programs, school } = educationDetails.properties;
-const { address: schoolAddress, name: schoolName } = school.properties;
+const {
+  address: schoolAddress,
+  name: schoolName,
+  facilityCode,
+} = school.properties;
```
erikphansen: okay, so I guess the drawback of Prettier is that is wrapped this block adding more noise to this particular diff :(

erikphansen:
To be honest, in the case where you have picked a school with the search tool and then decide to manually enter a school, I'm not entirely sure how the form processing on submit reconciles which school info to use. I'm a bit worried that it's a happy accident that it uses the manually entered info :-/ @rianfowler, do you have any insight here?

src/applications/edu-benefits/feedback-tool/helpers.js
```diff
@@ -58,8 +58,32 @@ export function fetchInstitutions({ institutionQuery, page, onDone, onError }) {
...
+// cleared out the (incorrect) facility code will be sent along with the
+// manually entered school address.
+function removeFacilityCodeIfManualEntry(form) {
+  if (
+    get(
+      'data.educationDetails.school.view:searchSchoolSelect.view:manualSchoolEntryChecked',
+      form,
+    )
+  ) {
+    return set(
```
cehsu: We could unset unset here instead.

src/applications/edu-benefits/feedback-tool/helpers.js
```diff
@@ -58,8 +58,32 @@ export function fetchInstitutions({ institutionQuery, page, onDone, onError }) {
   );
 }
 
+// Helper to clear out the facility code. Needed if the code was set via the
+// search tool and then manual address entry was selected. if this is not
+// cleared out the (incorrect) facility code will be sent along with the
+// manually entered school address.
+function removeFacilityCodeIfManualEntry(form) {
```
cehsu: Would it make sense to move this transformation to the BE with the rest of the transformations? @lihanli @jholton

cehsu (approved):
LGTM. I had a few questions/comments for your consideration.

src/applications/edu-benefits/feedback-tool/helpers.js
erikphansen: yessssss... thank you.

src/applications/edu-benefits/feedback-tool/helpers.js
erikphansen: Theoretically we could make the `view:manualSchoolEntryChecked` prop a non-`view:` prop so the BE could see it and then also rip that one off as part of the processing. But it makes just as much sense to just do the work here.

src/applications/edu-benefits/feedback-tool/config/form.js
rianfowler: net gain

rianfowler (approved):

PR #8640, opened 2018-10-03
Use regex for noITFPages

cvalarida:
## Description
There was a bug where if we went to the introduction page with a trailing slash, it would make ITF calls because the `pathname` didn't exactly match `/introduction` (since it was, in fact, `/introduction/`).

While I was here, I also opened it up to accept the list of `noITFPages` as a prop for when we get around to using this properly in all claims. (BDD claims won't file an ITF, so we won't want to start that on the first few pages until we're sure it isn't going to be BDD. I think that's how we'll want to handle that, anyhow.)

## Testing done
Manually tested; no ITF is made on the `/introduction/` or `/introduction` page(s), but _is_ made on the next one (increase only).

## Screenshots
N/A

## Acceptance criteria
- [x] No ITF call is made on either the `/introduction` or `/introduction/` page(s)
- [x] An ITF call is made on the next page

## Definition of done
    - No issue was filed for this; I just...fixed it...

### DIFF
--- a/src/applications/disability-benefits/all-claims/containers/ITFWrapper.jsx
+++ b/src/applications/disability-benefits/all-claims/containers/ITFWrapper.jsx
@@ -19,8 +19,12 @@ import {
 
 const fetchWaitingStates = [requestStates.notCalled, requestStates.pending];
 
-const noITFPages = ['/introduction', '/confirmation'];
 export class ITFWrapper extends React.Component {
+  static defaultProps = {
+    noITFPages: [/\/introduction/, /\/confirmation/],
+    /* noITFPages: ['/introduction', '/confirmation'], */
+  };
+
   constructor(props) {
     super(props);
     this.state = {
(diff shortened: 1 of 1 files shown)


src/applications/disability-benefits/all-claims/containers/ITFWrapper.jsx
```diff
@@ -19,8 +19,12 @@ import {
 
 const fetchWaitingStates = [requestStates.notCalled, requestStates.pending];
 
-const noITFPages = ['/introduction', '/confirmation'];
 export class ITFWrapper extends React.Component {
+  static defaultProps = {
+    noITFPages: [/\/introduction/, /\/confirmation/],
+    /* noITFPages: ['/introduction', '/confirmation'], */
```
markgreenburg: Might not need the strings anymore

markgreenburg (approved):
LGTM :shipit:

src/applications/disability-benefits/all-claims/containers/ITFWrapper.jsx
cvalarida: D'oh. Testing leftovers. Not too tasty cold.

Good call.

PR #8658, opened 2018-10-05
Modify build script to generate hashless assets for teamsite injection

bshyong:
## Description
This PR modifies the metalsmith build process to generate assets for injection that do not have cache-busting hashes. This is needed for injection into staging and production environments.

Injected assets for reference:
```
<style type="text/css">.brand-consolidation-deprecated { display: none !important; } </style>
<link rel="stylesheet" href="/generated/styleConsolidated.css" />
<link rel="stylesheet" href="/generated/static-pages.css">
<script type="text/javascript" src="/js/settings.js"></script>
<script type="text/javascript" src="/generated/vendor.entry.js"></script>
<script type="text/javascript" src="/generated/proxy-rewrite.entry.js"></script>
```

## Testing done
Local testing to check that builds are successful

## Acceptance criteria
- [x] Production build of vets-website should generate assets for teamsite injection that do not have cache-busting hashes

## Definition of done

Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/13947


script/configure-assets.js
```diff
@@ -45,6 +45,19 @@ function addAssetHashes() {
...
+      'proxy-rewrite.entry.js',
+      'styleConsolidated.css',
+      'static-pages.css',
+      'vendor.entry.js',
+    ].forEach(unhashedName => {
+      const hashedName = manifest[unhashedName];
+
+      files[`generated/${unhashedName}`] = files[hashedName.substr(1)]; // eslint-disable-line no-param-reassign
+    });
+
```
ncksllvn: Ah, based on how you had to check for `fileData` in the other middleware, I think it expects you to clone it like -

```js
const path = `generated/${unhashedName}`;
const copyFile = files[hashedName.substr(1)];
files[path] = {
   path,
   contents: new Buffer(copyFile.contents.toString())
};
```

My bad, I always forget that part. Clone like this should save us from having to touch more than one file.

ncksllvn (approved):
👍

script/configure-assets.js
bshyong: @ncksllvn I just tried removing it and the original script without the fileData check in `check-redirects`  results in a successful build so I think we are okay. I'll remove that extra safety check in the next PR if there are no side effects

PR #8668, opened 2018-10-05
Fix a11y issues on VCL: close button tabbing and focus state after closing

jbalboni:
## Description

This updates the VCL modal to remove the double close button tabbing issue in NVDA and also to correctly set the focus back to the triggering element when closing the modal. For some reason, we were only doing this when you used Escape to close the modal.

## Testing done

Tested locally in VoiceOver, bugged Trevor to do testing in other screen readers for me.

## Acceptance criteria
- [x] Close button doesn't read allow you to focus on close text
- [x] Focus is returned to correct element after closing modal

## Definition of done

Also linked to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/13949

### DIFF
--- a/src/platform/site-wide/accessible-VCL-modal.js
+++ b/src/platform/site-wide/accessible-VCL-modal.js
@@ -46,6 +46,9 @@ export function addFocusBehaviorToCrisisLineModal() {
     openControl.focus();
   }
 
+  // We're saving the element that triggered this modal
+  // in openControl, so that we can focus back on it later,
+  // when the modal is closed
   triggers.forEach(trigger => {
     trigger.addEventListener('click', () => {
       openControl = trigger;
--- a/src/platform/site-wide/accessible-VCL-modal.js
+++ b/src/platform/site-wide/accessible-VCL-modal.js
@@ -15,7 +15,9 @@ export function addFocusBehaviorToCrisisLineModal() {
   let openControl;
   const closeControl = tabbableElements[0];
   const lastTabbableElement = tabbableElements[tabbableElements.length - 1];
-  const triggers = document.querySelectorAll('[data-show="#modal-crisisline"]');
+  const triggers = Array.from(
+    document.querySelectorAll('[data-show="#modal-crisisline"]'),
+  );
 
   function captureFocus(e) {
     if (e.target === closeControl) {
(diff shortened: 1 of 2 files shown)


src/platform/site-wide/accessible-VCL-modal.js
```diff
@@ -45,8 +42,18 @@ export function addFocusBehaviorToCrisisLineModal() {
...
-  document.body.addEventListener('click', setOpenControl);
+  function resetFocus() {
+    openControl.focus();
+  }
+
+  triggers.forEach(trigger => {
+    trigger.addEventListener('click', () => {
+      openControl = trigger;
+    });
+  });
```
ncksllvn: Maybe it's just me, but it took me a moment to understand how this works with the focus. It might be worth adding a comment or a named function to help explain.

ncksllvn (approved):
LGTM!

cehsu (approved):
LGTM. Verified locally.

src/platform/site-wide/accessible-VCL-modal.js
```diff
@@ -15,7 +15,9 @@ export function addFocusBehaviorToCrisisLineModal() {
   let openControl;
   const closeControl = tabbableElements[0];
   const lastTabbableElement = tabbableElements[tabbableElements.length - 1];
-  const triggers = document.querySelectorAll('[data-show="#modal-crisisline"]');
+  const triggers = Array.from(
```
cehsu: Should we create a helper for this and add to the platform utils?

jbalboni: I think we probably need to refactor how this and the legacy/menu.js work together, and consolidate this code into something that works for any non-React modal we might create.

1Copenut (approved):
LGTM. I checked out the preview build to confirm `focus` was being restored to the trigger button, and it worked great. I have done preliminary testing of the `aria-label` locally, so have confidence in it as a solution for the big three screen readers.

src/platform/site-wide/accessible-VCL-modal.js
1Copenut: Does this script get run through Babel? `Array.from` will break in IE11, which is too bad because it's so useful.

jbalboni: Yep, any code in `src/` will get run through Babel.

PR #8698, opened 2018-10-09
CTA bug fixes

U-DON:
## Description
- Refactored helpers around URLs and redirects.
- VA Blue Button no longer has an automatic redirect. Related to department-of-veterans-affairs/vets.gov-team#13994.
- Updated copy about MHV alternate accounts. Related to department-of-veterans-affairs/vets.gov-team#13993.
- Fixed redirect buttons on MHV CTAs. Related to department-of-veterans-affairs/vets.gov-team#13991.

## Testing done
Local testing.

## Acceptance criteria
- [x] Get Medical Records should not automatically open a new tab.
- [x] CTAs with MHV redirects should have copy about alternate accounts.

## Definition of done


src/platform/site-wide/cta-widget/helpers.js
```diff
@@ -28,47 +38,94 @@ const mhvBaseUrl = () => {
...
   }
+
+  return null;
 };
 
-export const redirectUrl = contentUrl => {
-  switch (contentUrl) {
-    case '/health-care/secure-messaging/':
-      return `${mhvBaseUrl()}/mhv-portal-web/secure-messaging`;
+export const toolUrl = appId => {
```
U-DON: Consolidated `redirectUrl` and `continueUrl` into one helper to determine what URL the CTA should lead to.

saneshark: I could totally see us needing to add GA tags in the future, will that be easily supported by this change?

saneshark (commented):
Overall, looks good to me. But I'll defer to @ncksllvn

src/platform/site-wide/cta-widget/helpers.js
U-DON: Yup, any string manipulation shouldn't be an issue. Could just add another helper to append GA params.

src/platform/site-wide/cta-widget/helpers.js
```diff
@@ -28,47 +38,94 @@ export const mhvBaseUrl = () => {
...
+      return {
+        url: '/records/download-va-letters/letters',
+        redirect: false,
+      };
+
+    case frontendApps.VETERAN_ID_CARD:
+      return {
+        url: 'https://vicbdc.ppd.vba.va.gov/VIC',
+        redirect: false,
+      };
```
ncksllvn: It might help to show the relationship between values by creating a config map similar to [this](https://github.com/department-of-veterans-affairs/vets-website/blob/master/script/constants/hostnames.js), but that's probably just a preference.

ncksllvn (approved):
LGTM!

src/platform/site-wide/cta-widget/helpers.js
U-DON: I think an Object/Map would be more readable. I was thinking that a switch would use less memory, but that's probably moot given the small number of cases/members.

PR #8699, opened 2018-10-09
Va gov/bug sidenav open on hover

RobAWilkinson:
## Description
Cleans up sidenav javascript and allows for more complicated cycling with accordion buttons.

Previously, there were unticketed bugs related to cycling through elements with tab when opening the sidenav. Similar to the crisis line modal on the home page, tab was not cycling through the overarching element properly.

## Testing done
Local testing.

To verify @1Copenut I recommend checking out pages that have nesting such as `health-care/about-va-health-benefits/where-you-go-for-care/` as well as a top level page like `health-care/about-va-health-benefits/`


1Copenut:
@RobAWilkinson I found a few things, that are spread across several browsers and OS, so I'll add a comment for each one. I feel it's the best way to handle changes that aren't directly tied to code updated.

**1 of 4: Safari + Mac OS, 400% zoom**
The attached screenshot shows the `More in this section` menu in `:hover` pseudo-state, but the arrow is white. I'd expect it to be dark blue like the text.

---

screen shot 2018-10-09 at 2 59 21 pm

1Copenut:
**2 of 4: Firefox + Windows 10 + NVDA, 400% zoom**
This screenshot shows a strange stacking order, possibly `z-index` or `position: <SOMETHING>` in the related CSS.

---

more-in-this-section-z-index

1Copenut:
**3 of 4: Firefox + Windows 10 + NVDA, 400% zoom**
This screenshot shows a strange thing Firefox does with `text-indent: -9999em` text, in that it extends the focus bar off left. We had this same issue on the Veterans' Crisis Line modal, and handled it using this HTML:

```html
<button aria-label="Close this modal" class="va-modal-close va-overlay-close va-crisis-panel-close" type="button">
  <i class="fa fa-times-circle-o va-overlay-close" aria-hidden="true"></i>
</button>
```

The main key is the `aria-label` and the `aria-hidden` I element. That way we could control exactly what we wanted the button to read to assistive devices, and ensure Firefox + NVDA users could close it successfully with `SPACE` and `ENTER` keypresses.

---

firefox-focus-close-button

1Copenut:
**4 of 4: IE11 + JAWS, 400% zoom**
The menu wouldn't open because there's a non-ES5 use of the arrow function on the script being executed. My IE11 was running extremely slow, but I got a screenshot of the offending line and function.

---

ie11-arrow-bug

1Copenut (changes requested):
@RobAWilkinson I reviewed and added 4 comments inline. Not ideal; the adjustments I saw were visual and/or errors in console, so screenshots and notes felt the best way forward.

RobAWilkinson:
1. Fixed in updated PR
2. Fixed in PR
3. Fixed in PR
4. Discussed over slack, seems just a heroku issue

1Copenut:
Issue 2 of 4 I commented on is still happening. It has to do with an inordinately high z-index of 300 on `.merger .vet-toolbar`. Fixing it can be done by bumping z-index on the following element:

```css
.merger .va-btn-sidebarnav-trigger.fixed-trigger {
  /* Existing rules */
  z-index: 301
}
```
This a very unlikely use case, having the layout zoomed to 400% and the `How do you know` button expanded, but it seems an easy fix worth doing.

RobAWilkinson:
@1Copenut there is a separate issue to handle z-indexes across the site, are you ok merging this as is and just letting the work in there fix z-index-s everywhere?
Would hate to add to the crud by bumping by 1 and adding yet another new z-index value

1Copenut:
@RobAWilkinson Yes, let's handle z-indexes in one place. I'll make sure to comment on that item with my finding. Retesting all others now.

1Copenut (approved):

1Copenut:
Items 1, 3, and 4 from my previous comments have been fixed and verified. @1Copenut tested with different screen readers to ensure close button announced properly. Good to merge and close.

PR #8707, opened 2018-10-09
Fix accessibility issues in footer and refactor

jbalboni:
## Description

This PR is doing four things:

1. Fixes the markup so that links can be moved through by screen readers correctly (https://github.com/department-of-veterans-affairs/vets.gov-team/issues/13859)
2. Removes the flexbox based re-ordering in favor of separate markup for mobile/desktop versions (https://github.com/department-of-veterans-affairs/vets.gov-team/issues/13607)
3. Refactors the footer into multiple components and to limit React conditionals
4. Consolidates footer styles into a new footer sass file that replaces the existing footer file
   - There's still some refactoring here, I mostly stayed away from refactoring the styles

The refactor is focused on trying to reduce the React fanciness in the footer code because we probably want to move to a React-less solution in the future to reduce our JS payload. This pushes us in that direction by making the vast majority of the markup static.

The less than ideal thing about this refactoring is that we have separate markup for the desktop and mobile footers, both of which are in the DOM and just hidden or shown depending on screen size.

The reason for this is that to make the content understandable for screen readers, we need the markup to be different. We need to use normal accordion markup for mobile and a more basic groups of lists approach for the normal footer.

## Testing done
Tested locally at different breakpoints, tested in VoiceOver

## Screenshots

screen shot 2018-10-09 at 2 49 39 pm
screen shot 2018-10-09 at 2 49 28 pm

## Acceptance criteria
- [x] Footer still looks the same at different breakpoints
- [ ] Screen readers can read through the links in the footer correctly
- [x] Tab order follows the markup order in the mobile menu

## Definition of done


rianfowler (approved):

1Copenut:
@jbalboni There are some legacy dots and icons showing up on the mobile version of our footer links, but I have a [separate ticket (13688)](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/13688) in for those. Will append today's screenshot to it, and approve this item. JAWS, NVDA, and VoiceOver all recognize the links and pick them up in full readout.

1Copenut (approved):

src/platform/site-wide/sass/merger.scss
```diff
@@ -504,96 +405,6 @@
...
-        .usa-accordion-button[aria-expanded=true] {
-          background-image: url("/img/minus-white.png");
-          background-image: url("/img/minus-white.svg");
-        }
-      }
-      //.merger #footerNav .va-footer-linkgroup-title button
-
-
-    }
-
```
ncksllvn: Ah thank you for doing this!

src/platform/site-wide/sass/modules/_m-footer-brand-consolidation.scss
```diff
@@ -0,0 +1,195 @@
...
+        content: "|";
+        padding: 0 1rem;
+      }
+    }
+  }
+}
+
+.va-footer-content {
+  @include accessibly-hidden();
+}
```
ncksllvn: So much easier to work with!

src/platform/site-wide/va-footer/components/CrisisPanel.jsx
```diff
@@ -0,0 +1,84 @@
...
+            href="https://www.veteranscrisisline.net/"
+          >
+            VeteransCrisisLine.net
+          </a>
+          .
+        </div>
+      </div>
+    </div>
+  );
+}
```
ncksllvn: Was the Crisis Line modal in the header not working with the `data-show` trigger?

src/platform/site-wide/va-footer/index.js
```diff
@@ -6,20 +6,17 @@
...
-import Main from './containers/Main';
+import Footer from './components/Footer';
 
 /**
  * Sets up the login widget with the given store at login-root
  *
  * @param {Redux.Store} store The common store used on the site
  */
-export default function startVAFooter({ store, handleFooterDidMount }) {
+export default function startVAFooter(handleFooterDidMount) {
```
ncksllvn: Fixes this pesky error, department-of-veterans-affairs/vets.gov-team/issues/14104!

src/platform/site-wide/va-footer/components/Footer.jsx
```diff
@@ -0,0 +1,53 @@
...
+    if (this.props.handleFooterDidMount) {
+      this.props.handleFooterDidMount();
+    }
+  }
+  render() {
+    return (
+      <div>
+        <div className="footer-inner">
+          <DesktopLinks visible={!this.state.isMobile} links={this.linkObj} />
+          <MobileLinks visible={this.state.isMobile} links={this.linkObj} />
```
ncksllvn: If it's possible to wrap the Desktop/Mobile components in a class that shows only for those screen sizes, we might be able to get away with removing `isMobile` from state.

ncksllvn (approved):
LGTM 👍  I'm open to converting this React component into a Metalsmith component soon or even in this PR, if it seems like we're in good position to do it in your opinion. We'd just have to make sure the changes are coordinated with the Teamsite injection script.

ncksllvn:
@jbalboni Ticket created back in September for converting this to a non-React component, https://github.com/department-of-veterans-affairs/vets.gov-team/issues/13249.

src/platform/site-wide/va-footer/components/Footer.jsx
jbalboni: We need to dynamically set `aria-hidden` either way, so I figured it wasn't worth doing at the moment.

jbalboni:
@ncksllvn Converting to html and regular JS is trickier than I initially thought, because:

1. We need to render the footer on Teamsite pages, which means it needs to be bundled in JS files
2. We can't have Metalsmith rendered files in our JS bundles, because that ties them together and makes it so we can't deploy content independently.

We need to figure out a different approach for 1, so that we can avoid 2.

src/platform/site-wide/va-footer/components/Footer.jsx
ncksllvn: Ah yeah good call

ncksllvn:
@jbalboni There is this [partials](https://github.com/department-of-veterans-affairs/vets-website/tree/master/src/applications/proxy-rewrite/partials) folder that we could possibly leverage, but yeah, at the moment I'm not positive on how crossover between Metalsmith/React components will work and don't really want to get into it until the CMS deployment is set up.

PR #8720, opened 2018-10-10
Optimize images on homepage

jbalboni:
## Description

A few things are happening here:

1. I pulled new higher-res versions of the two va logos from the design folder, resized them to 512px, and then ran them through TinyPNG.
2. I ran the other homepage images through TinyPNG/TinyJPG
3. I added a lazy loader module and updated the homepage news story images to lazy load
4. I updated the VCL homepage button to use the same (smaller) svg that the VCL button at the top of the page uses

I didn't compare production to production, but this appears to take the initial image payload from 578kb to 125kb.

This should also address https://github.com/department-of-veterans-affairs/vets.gov-team/issues/14033, but I didn't see much fuzziness before (aside from when we had the header image too wide).

## Testing done
Tested locally, Browserstack for different browsers

## Acceptance criteria
- [x] Images are smaller
- [x] Images aren't fuzzy
- [x] Homepage news story images load in all browsers

## Definition of done


src/applications/static-pages/static-pages-entry.js
```diff
@@ -131,3 +132,9 @@ if (location.pathname === '/disability-benefits/increase-claims-testing/') {
 if (location.pathname === '/') {
   createMyVALoginWidget(store);
 }
+
+/* eslint-disable no-unused-vars,camelcase */
+const lazyLoad = new LazyLoad({
```
jbalboni: I'm on the fence about if this is really worth it. Curious about other thoughts.

rianfowler: We could profile it against a simulated slow connection using the timeline tool (with screenshots enabled). This branch vs master locally.

rianfowler (approved):

src/applications/static-pages/static-pages-entry.js
jbalboni: The results of testing this were that the JS isn't getting pushed back by the images, but lazily loading them appears to save about half a second (on simulated fast 3G) in when the page fully loads. Given that we also save users 150kb by not loading them if they just go to the homepage and open a menu, seems worth it.

script/check-broken-links.js
```diff
@@ -26,6 +28,17 @@ function checkBrokenLinks() {
...
+      // if we have lazy images, they don't have a src attr,
+      // so we need to add one for the blc
+      if (file.contents.includes('data-src=')) {
+        lazyImageFiles.push(fileName);
+        const doc = cheerio.load(file.contents);
+        doc('img[data-src]').each((i, el) => {
+          doc(el).attr('src', doc(el).attr('data-src'));
+        });
+        file.contents = doc.html();
+      }
```
ncksllvn: Could we possibly chunk this and the cleanup logic into their own function above `checkBrokenLinks`? Just worried that the comment could get lost or something.

src/platform/site-wide/sass/modules/_m-homepage.scss
```diff
@@ -167,3 +167,15 @@
...
+  width: 100%;
+  height: 0;
+  padding-bottom: 65%; /* You define this doing height / width * 100% */
+  position: relative;
+  > .lazy {
+    width: 100%;
+    /*height: auto;*/
+    position: absolute;
+  }
+}
```
ncksllvn: Are these style rules necessary for any elements that are lazy loaded, or just for those particular images on the homepage? Wondering if there's any way we can streamline this in case other pages end up wanting to use it. Probably only a major concern on the homepage either way.

ncksllvn (approved):
Pretty dang cool! Makes the homepage feel much lighter, and very cool seeing the images pop in when you scroll quickly. Great thinking with this.

src/platform/site-wide/sass/modules/_m-homepage.scss
jbalboni: These are necessary if you have images that are expanding to fit an area and you want to keep the aspect ratio. The padding-bottom value is image specific, so we can't make it completely generic. Though maybe a mixin would work? I think we could do that in a separate ticket and put it in the design system, once this is in production for a bit and is working alright.

jbalboni:
Also, there are two other images that we could theoretically lazy load: the veteran banner and the footer VA logo. But the footer VA logo is small and would require some React work. And the banner is different for mobile/desktop and is a background image, so it would take some investigation to see if it would work. The news images give us the best value without spending tons of time on this.

src/platform/site-wide/sass/modules/_m-homepage.scss
ncksllvn: Gotcha, I think that's an awesome idea! Also helps to make the strategy more discoverable as we add images.

PR #8726, opened 2018-10-10
All claims updatable payments mg

markgreenburg:
## Description
Swaps the view-only payments screen with a screen that allows editing of all payment details.

## Testing done
- [x] Tested locally
- [x] Added unit tests

## Screenshots
screen shot 2018-10-10 at 4 48 28 pm

## Acceptance criteria
- [x] Veteran must select between checking and savings account types
- [x] Veteran must enter account number
- [x] Veteran must enter routing number
- [x] Veteran must enter bank name
- [x] Validations for each field must match EVSS Swagger

## Definition of done


src/applications/disability-benefits/all-claims/config/schema.js
```diff
@@ -877,6 +877,23 @@ const schema = {
...
+      type: 'string',
+      enum: ['Checking', 'Savings'],
+    },
+    accountNumber: {
+      type: 'string',
+      minLength: 4,
+      maxLength: 17,
+    },
+    routingNumber: {
+      type: 'string',
```
cvalarida: I think I've hunted down this rabbit before, but do we have a `number` type we could use that tells the `StringField` the input should be `type="number"`?

src/applications/disability-benefits/all-claims/pages/paymentInformation.js
```diff
@@ -1,20 +1,49 @@
...
+  },
+  bankName: {
+    'ui:title': 'Bank name',
   },
 };
 
 export const schema = {
   type: 'object',
-  properties: {},
+  required: ['accountType', 'accountNumber', 'routingNumber', 'bankName'],
```
cvalarida: So this is where it gets tricky. These fields are only required _if_ they enter any information. They should be able to submit without any of this.

@goldenmeanie What's the best way to handle that? We've historically done a yes / no question that opens this section up.

If we went down that path again, we could default that question to `yes` (or whatever) to expand the bank account information fields if we've got pre-fill information for them.

cvalarida (commented):
Just had one concern about the required fields.

src/applications/disability-benefits/all-claims/pages/paymentInformation.js
markgreenburg: Ahh, right, I actually forgot this was meant to be optional. No expanding necessary. We can just do a ui:required that checks whether the other fields have been touched. We do this for homelessness - phone is required if there's an email, and vice-versa.

src/applications/disability-benefits/all-claims/config/schema.js
markgreenburg: Unfortunately switching to `type: 'number'` means we can't use `minLength`, `maxLength`, or `pattern` validations so this isn't really viable given the restrictions we need to place on the input.

markgreenburg:
@cvalarida Bank info is no longer required, but all the fields are conditionally required if any of them are touched. Could you please take a look and see what you think?

src/applications/disability-benefits/all-claims/utils.jsx
```diff
@@ -244,6 +230,27 @@ export const hasOtherEvidence = formData =>
...
 
+/**
+ * Inspects all given paths in the formData object for presence of values
+ * @param {object} formData  full formData for the form
+ * @param {array} fieldPaths full paths in formData for other fields that
+ *                           should be checked for input
+ * @returns {boolean} true if at least one path is not empty / false otherwise
+ */
+export const fieldsHaveInput = (formData, fieldPaths) =>
+  fieldPaths.reduce(
```
cvalarida: Alternatively:
```suggestion
  fieldPaths.some(path => !!_.get(path, formData, '');
```

Sidebar: GitHub suggestions are cool, but don't work super well for multi-line suggestions, I think. Unless I'm missing something. :man_shrugging:

cvalarida (approved):

PR #8728, opened 2018-10-10
Remove data-popup

ncksllvn:
## Description
Per @bethpotts 
> @ncksllvn For your awareness, contrary to the title of this issue, we will only use pop-ups for linking to pages off the site. We won't use pop-ups for pdfs (which we're considering "internal" pages after merger happens since they live on va.gov). So could you revert the above example file you added that code to? I'd rather you do it so I don't mess up the code!

https://github.com/department-of-veterans-affairs/vets.gov-team/issues/3384#issuecomment-428763474

## Acceptance criteria
- [x] Link does not open as a popup


bethpotts (approved):
Merge!

PR #8747, opened 2018-10-12
Render site-name conditionally in files shared between Vets.gov and VA.gov

ncksllvn:
## Description
In files that are shared between Vets.gov and VA.gov, we need to render the site name conditionally. 

### Tickets
Moves to validate - https://github.com/department-of-veterans-affairs/vets.gov-team/issues/14141

## Testing done
- Unit tests, many of which use the hardcoded "Vets.gov" string and would fail if "VA.gov" were rendered instead
- Checked the site manually

## Screenshots
Ignore the top title - that's an image and will have to be updated separately.

### Vets.gov
image

### VA.gov
image

## Acceptance criteria
- [x] "VA.gov" is rendered as the site name in shared assets when running under a VA.gov domain
- [x] "Vets.gov" is rendered as the site name in shared assets when running under a Vets.gov domain

## Definition of done

### DIFF
--- a/src/applications/auth/containers/AuthApp.jsx
+++ b/src/applications/auth/containers/AuthApp.jsx
@@ -178,7 +178,7 @@ export class AuthApp extends React.Component {
     const view = this.state.error ? (
       this.renderError()
     ) : (
-      <LoadingIndicator message="Signing in to {siteName}..." />
+      <LoadingIndicator message={`Signing in to ${siteName}...`} />
     );
 
     return (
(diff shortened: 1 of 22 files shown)


src/applications/auth/containers/AuthApp.jsx
```diff
@@ -177,7 +178,7 @@ export class AuthApp extends React.Component {
     const view = this.state.error ? (
       this.renderError()
     ) : (
-      <LoadingIndicator message="Signing in to Vets.gov..." />
+      <LoadingIndicator message="Signing in to {siteName}..." />
```
jbalboni: I think you need to make this a template string.

jbalboni (approved):
Found one thing, otherwise LGTM

src/applications/auth/containers/AuthApp.jsx
ncksllvn: Ah thank you!

PR #8751, opened 2018-10-12
Implement the new images for sharing to social media

ncksllvn:
## Description
This PR does a little more than add the new images, because we had to use absolute URLs to properly create the image links, and I was finding myself repeating code with detecting the protocol based on host and things around that. I figured it would make sense to derive that site-url property early on. 

This PR ultimately -
- Derives a `hostUrl` property on the `BUILD_OPTIONS`, that is ultimately the sum of the protocol, host, and port.
- Add that `hostUrl` as a Metalsmith metadata so that it's accessible in the header template
- Implements that `hostUrl` in the header template and in the sitemap

### Ticket
Moves to validate - https://github.com/department-of-veterans-affairs/vets.gov-team/issues/13730

## Testing done
Manually checked the `hostUrl` property in the console, sitemap, and HTML on localhost.

## Screenshots
These screenshots show the `hostUrl` is being derived correctly.

### localhost
image
image

### This PR's Heroku deployment
image

See the absolute paths of the image meta tags -
image

### vagovdev (did manual deploy)
image

image

## Acceptance criteria
- [x] Absolutely URLs are being generated corrected
- [x] Image preview on social media works

## Definition of done


script/options.js
```diff
@@ -46,23 +44,22 @@ function applyDefaultOptions(options) {
...
+    options.protocol = 'http';
+  } else {
+    options.host = hostnames[options.buildtype];
+    options.protocol = 'https';
   }
+
+  options.hostUrl = `${options.protocol}://${options.host}${
+    options.port ? `:${options.port}` : ''
+  }`;
 }
```
ncksllvn: In working through where to derive the `hostUrl` property and deciding what to name it, I realized that there is a discrepancy between what host/hostname means in the browser versus in Node, so the logic touched more than I wanted, but I think it makes sense. 

- No default value for `buildtype` from the command line. This is so we can detect a dev environment (dev.va.gov) versus localhost.
    - If no buildtype, we default to http, development, and use the host as passed from the command line.
    - If there is a buildtype, we use https and override the host to derive it from the buildtype.

script/constants/hostnames.js
```diff
@@ -1,10 +1,10 @@
...
-  [environments.STAGING]: 'https://staging.vets.gov',
-  [environments.VAGOVDEV]: 'https://dev.va.gov',
-  [environments.VAGOVSTAGING]: 'https://staging.va.gov',
+  [environments.DEVELOPMENT]: 'dev.vets.gov',
+  [environments.PREVIEW]: 'preview.va.gov',
+  [environments.PRODUCTION]: 'www.vets.gov',
+  [environments.STAGING]: 'staging.vets.gov',
+  [environments.VAGOVDEV]: 'dev.va.gov',
+  [environments.VAGOVSTAGING]: 'staging.va.gov',
 };
```
ncksllvn: This in particular was where I discovered the difference between host vs. hostname in Node vs. the browser. It makes sense to me to remove the protocol here and derive it as part of `hostUrl` instead, because it reconciles the format to match that of the `host` command line argument.

jbalboni:
I haven't checked to see how big they are, but could be worth putting the new images through tinypng.com.

rianfowler (approved):

rianfowler:
Good refinements!

ncksllvn:
@jbalboni saved us 4%!

script/options.js
ncksllvn: Update - add an override on the port as well

RobAWilkinson (approved):
Looks great, can we also use this url helper in the mega menu absolute links stuff?

ncksllvn:
@RobAWilkinson Definitely!!

PR #8753, opened 2018-10-12
Implement design updates to side nav on desktop

accbjt:
## Description
Implement design updates to side nav on desktop

- [x] add new styles and make sure the interactions are correct
- [x] Go through the spec in detail and make sure all requirements are met.
 
## Acceptance criteria
- [  ] Should meet design specs in issue.

## Definition of done


jbalboni:
Some things I noticed trying it out locally:

- The bolding of the link text on hover means that link text jumps around sometimes, because the text gets bigger and sometimes breaks into two lines. This feels weird to me as I mouse around in the menu. I see that's in the spec, but could be worth pointing out.
- Is this external link icon desired?

screen shot 2018-10-12 at 3 55 41 pm

- The font size for some of the child nav links also appears to be 17px, instead of 15px.
- On an iPad, some text is overflowing:
screen shot 2018-10-12 at 4 00 51 pm

src/platform/site-wide/sass/merger.scss
```diff
@@ -739,7 +775,43 @@
...
+    a {
+      &::before {
+        content: " \2039 ";
+        display: inline-block;
+        padding: 0 0.35em 0 0;
+      }
+    }
+  }
 }
+
```
ncksllvn: Ah man, do we still not have a dedicated sass file for the side bar? Hate to see this file keep growing...

ncksllvn: If you consider it to be risky, don't worry about it at all. We can wait til after launch.

jbalboni:
@accbjt Is the "More in this section" button intentionally a different color than before? Also, it'd be nice to fix the arrow position:

screen shot 2018-10-15 at 10 03 39 am

jbalboni:
The active state on the close button needs to be updated:

screen shot 2018-10-15 at 10 10 24 am

src/platform/site-wide/sass/merger.scss
```diff
@@ -549,12 +551,28 @@
...
         font-weight: bold;
         margin-left: 0;
+        padding-left: 25px;
       }
     }
 
     .usa-sidenav-sub_list{
       a {
-        width: 100% !important; // sorry this is temporary
+        width: 98% !important; // sorry this is temporary
```
jbalboni: Why did this go down to 98%?

accbjt: This was done because the focus state was getting cut off.
screen shot 2018-10-15 at 8 17 20 am

accbjt:
@jbalboni Fixed the `More in this section` arrow position and changed the close button on mobile to gray.

src/platform/site-wide/sass/merger.scss
accbjt: The border on focus was getting cut off on the right side, so I had to reduce it for it to show correctly.

jbalboni:
I get some horizontal scrolling on my phone now. Initially, it looks like focus outline is cutoff around the close icon, but you can scroll right:

screenshot_20181015-135501

jbalboni:
Actually, the horizontal scrolling wasn't caused by these changes, but I think it being there is why the close button looks cut off.

jbalboni:
@accbjt With the most recent changes, I can't scroll on my phone at all anymore (on any page). Maybe it's that overflow-y rule? We can take it out and wait for my PR to fix the scrolling if it's the culprit.

accbjt:
@jbalboni I removed the overflow-y.

jbalboni:
@accbjt There's a spacing change when you mouse back and forth between Get Benefits and Eligibility here on `/disability/eligibility/`:

screen shot 2018-10-16 at 12 57 43 pm

va-gov/includes/header.html
```diff
@@ -10,7 +10,8 @@
 <meta http-equiv="X-UA-Compatible" content="IE=edge" />
 <meta name="HandheldFriendly" content="True">
 <meta name="MobileOptimized" content="320">
-<meta name="viewport" content="width=device-width, initial-scale=1.0" />
+
+<meta name="viewport" content="width=device-width, initial-scale = 1.0, maximum-scale=1.0, user-scalable=no" />
```
jbalboni: I think we were going to take this back out?

accbjt: I thought I removed this but I guess I didn't. I'll revert it back to the old one.

accbjt: I also fixed that border problem but it must have gotten reintroduced when I pulled from masters.

jbalboni:
Similar/same issue above when collapsed (also the outline is cut off):

screen shot 2018-10-16 at 2 13 08 pm
screen shot 2018-10-16 at 2 13 04 pm

jbalboni:
The hover state background is not quite extending to the end of the top and bottom borders on the right:

screen shot 2018-10-16 at 2 16 00 pm

accbjt:
@jbalboni Good catch. Glad you saw this because this has probably been there for awhile now. Let me see if I can fix this.

accbjt:
@jbalboni `/va-payment-history` The page was using the wrong template. I am also going to see if there are any other pages that might have the same issue.

accbjt:
@jbalboni This should be good now.

jbalboni:
@accbjt Found a missing margin on iPad size:

screen shot 2018-10-16 at 4 36 49 pm

accbjt:
@jbalboni added some margin to this
screen shot 2018-10-16 at 2 00 14 pm

jbalboni (approved):
LGTM. It might be worth having Emily look through some screens or a review instance it to be sure, but I don't see any issues on my phone, ipad, or macbook.

PR #8759, opened 2018-10-13
526 analytics cv

cvalarida:
## Description
Adds some analytics events to the 526 MVP.

To do so more cleanly, I created a little redux middleware.

## Testing done
Manually tested that it all works.
I still need to write unit tests for the middleware itself.

## Screenshots
N/A

## Acceptance criteria
- [x] The events are fired at the appropriate times

## Definition of done


src/applications/disability-benefits/526EZ/components/IntroductionPage.jsx
```diff
@@ -17,11 +17,11 @@ import { toggleLoginModal } from '../../../../platform/site-wide/user-nav/action
 import { focusElement } from '../../../../platform/utilities/ui';
 
 import BetaGate from '../containers/BetaGate';
-
 import { VerifiedAlert } from '../helpers';
-
 import FormStartControls from './FormStartControls';
 
+const gaStartEventName = 'disability-526EZ-start';
```
cvalarida: Turns out there's already a `disability-526EZ-sip-form-loaded` event. @djmunoz1 Is that enough, or do we need a new one here?

src/platform/startup/store.js
```diff
@@ -37,18 +38,22 @@ export const commonReducer = {
  * sets up the Redux devtools in development and adds redux-thunk as middleware.
  *
  * @param {Object} [appReducer={}] An object with reducer functions as properties
+ * @param {Array} A list of analytics events to capture when redux actions are fired
```
markgreenburg: I think it's expecting that the first word after the types of your params is the argument name, e.g.
`* @param {Array} AnalaticsEvents A list of analytics....`

src/applications/disability-benefits/all-claims/analytics-events.js
```diff
@@ -0,0 +1,32 @@
...
+    action: ITF_CREATION_FAILED,
+    event: 'disability-526EZ-itf-not-creation',
+  },
+  {
+    action: ITF_FETCH_SUCCEEDED,
+    event: 'disability-526EZ-itf-retrieved',
+  },
+  {
+    action: ITF_FETCH_FAILED,
+    event: store => {
```
markgreenburg: Not sure where to put this, but what do you think about documenting the signature of an `event` function somewhere we're likely to see if next time we go to add events?

src/platform/forms/save-in-progress/FormStartControls.jsx
```diff
@@ -45,6 +50,7 @@ class FormStartControls extends React.Component {
   };
 
   startOver = () => {
+    this.captureAnalytics();
```
markgreenburg: Do we have requirements around tracking start and restart events separately?

src/platform/forms/save-in-progress/FormStartControls.jsx
```diff
@@ -22,7 +22,12 @@ class FormStartControls extends React.Component {
     this.props.router.push(this.props.startPage);
   };
 
+  captureAnalytics = () =>
+    this.props.gaStartEventName &&
```
markgreenburg: Could we use the `formId` for this instead (or like a `start-form-${formId}`)? Is there a reason we'd want the name to be different than the formID?

src/platform/forms/save-in-progress/FormStartControls.jsx
```diff
@@ -22,7 +22,12 @@ class FormStartControls extends React.Component {
     this.props.router.push(this.props.startPage);
   };
 
+  captureAnalytics = () =>
```
markgreenburg: Thinking about this some more, I wonder if we even need to log anything in `FormStartControls`? IIRC it's already covered by standard analytics event tracking somewhere higher up in Forms, but I could be wrong. Otherwise, we could also find the right action types and make entries for them in the nifty analytics list you just made for the middleware.

markgreenburg (approved):
❤️ 
Only concern is around dealing with FormStartControls and whether we need to log that at all (might already be getting logged?) but I love that the middleware approach is going to save us time down the road.

Also, tests.

src/applications/disability-benefits/all-claims/analytics-events.js
```diff
@@ -0,0 +1,32 @@
...
+      const state = store.getState();
+      return state.user.profile.savedForms.find(f => f.form === '21-526EZ')
+        ? 'disability-526EZ-itf-not-retrieved-existing-user'
```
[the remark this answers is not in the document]
cvalarida: Yeah, good point.

src/applications/disability-benefits/all-claims/analytics-events.js
cvalarida: I'm all for documenting, but I'd like to make sure it's discoverable. Any suggestions?

src/platform/forms/save-in-progress/FormStartControls.jsx
cvalarida: Not that I know of. @djmunoz1 Do you know?

src/platform/forms/save-in-progress/FormStartControls.jsx
cvalarida: I can't think of a reason that wouldn't work. I think I did this because I wasn't sure if I had access to the form ID, but if I recall, we do...

I'll look into it. :+1:

src/platform/forms/save-in-progress/FormStartControls.jsx
cvalarida: Yeah, I think these events are tracked in USFS, but I know they have plans on changing that. Something about opening up hooks for us to use.

It probably makes sense to rely on that logging for now and make the change to track it explicitly when we upgrade USFS to the version where they provide those hooks.

cvalarida: The events gathered when starting the form on this branch:
image

The `-start` event comes from the addition here. `-sip-form-loaded` comes from our SiP code. I think it's the same for both loading up a saved form and starting a new form (and getting the pre-fill information), so I think we will still need to track form starts after all.

@djmunoz1 to confirm.

cvalarida:
@nedierecel 
I changed up the naming of the events a bit to be hopefully clearer. This is what I came up with:
- `disability-526EZ-start`
- `disability-526EZ-itf-created`
- `disability-526EZ-itf-not-created`
- `disability-526EZ-itf-retrieved`
- `disability-526EZ-itf-not-retrieved-saved-form-found`
- `disability-526EZ-itf-not-retrieved-saved-form-not-found`

Does that work?

nedierecel:
@cvalarida Thanks for letting me know. This should not be an issue. I'll tweak some settings within Google Tag Manager using the new naming conventions.

cvalarida:
Awesome, thanks!

PR #8769, opened 2018-10-16
update jenkinsfile to pull down sibling repo

tskinn:
This turned out to be more changes than I anticipated. Maybe I am going about this the wrong way? 

Since the vets-website directory needs to have the vagov-content repo as a sibling directory we need to move vets-website somehow. So instead of the normal behavior of checking out the repo in the current directory, I have configured it to have two directories in the workspace: vets-website and vagov-content.


ncksllvn (approved):
Description makes perfect sense to me! Great approach.

PR #8792, opened 2018-10-18
Va gov/feature/13631: LeftRailNav style updates

accbjt:
## Description
A couple more updates to the LeftRailNav. These are the updates @emilywaggoner mention on this ticket.

https://github.com/department-of-veterans-affairs/vets.gov-team/issues/13631

## Testing done
Local Testing

## Screenshots
screen shot 2018-10-18 at 7 49 59 am
screen shot 2018-10-18 at 7 50 28 am

## Acceptance criteria
- [ ] Child links should be 15px
- [ ] There should be no border line between links
- [ ] There should be a border top and bottom on accordion
- [ ] on hover should not have a gap on the right side. It should be full width

## Definition of done


src/platform/site-wide/sass/merger.scss
```diff
@@ -557,9 +559,17 @@
             margin: 2px;
           }
 
+          &:focus:hover {
+            background-color: $mm-gray;
+            font-weight: bold;
+            margin: 2px;
+            width: 98%;
+          }
+
```
ncksllvn: Is `&:focus:hover` definitely what we want to be doing here?

src/platform/site-wide/sass/merger.scss
```diff
@@ -557,9 +559,17 @@
...
           &:hover {
             background-color: $mm-gray;
             font-weight: bold;
+            width: 100%;
```
ncksllvn: Would it make sense to apply `width: 100%` directly to the parent instead of only hover? Or would that break things?

jbalboni: Yeah, I think we might want to change this. It creates a jumpiness when you hover over and off again, because the you can see the right margin appear as the grey background fades.

accbjt: @ncksllvn Let mw try it out and see what happens. I know I was having issues with this.

src/platform/site-wide/sass/merger.scss
accbjt: Let me try to add the 100% to the parent and see if this fixes things.

emilywaggoner (approved):
These changes look great to me. Thanks everyone.

ncksllvn (approved):

PR #8793, opened 2018-10-18
Replaces PNG with SVG on 503.html page

hpjaj:
## Description
Changes 503.html to use SVG logo instead of inline PNG

## Testing done
Local testing

## Screenshots
image

## Acceptance criteria
- [x] Replaces PNG with SVG

## Definition of done


va-gov/pages/503.html
```diff
@@ -44,7 +44,7 @@
 <body>
   <div class="container">
     <a href="/">
-      <img width="250" alt="Go to VA.gov" src="{% include "va-gov/includes/logo-350px-base64.html" %}" />
+      <img width="250" alt="Go to VA.gov" src="/img/va-primary-horizontal-black-vector.svg" />
```
ncksllvn: This is gonna sound wild, but because the nature of this page is that the page itself is the only resource required, we have to import the SVG inline, like -

`src="{% include "assets/img/va-primary-horizontal-black-vector.svg" %}" />`

Hopefully the SVG isn't large to slow the page down.

assets/img/va-primary-horizontal-black-vector.svg
```diff
@@ -0,0 +1,9645 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg width="321px" height="74px" viewBox="0 0 321 74" focusable="false" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <!-- Generator: Sketch 43.2 (39069) - http://www.bohemiancoding.com/sketch -->
+    <title>va-primary-horizontal-black-vector</title>
```
1Copenut: The `<title>` and `<desc>` will need to be removed to make the SVG work with screen readers. VoiceOver wants to announce the title, which we want to prevent.

1Copenut (changes requested):

1Copenut:
@hpjaj @amycesal The SVG that was added to the project is rendering with some rough edges. I tested in Chrome, Safari, Firefox (MacOS) and got the same result. Screenshot is attached below.

This SVG is such a complicated shape, it might have to stay a JPG or PNG image, and use a different strategy for multiple sizes like `srcset`, with a best-fit for IE11.

* https://caniuse.com/#feat=srcset
* https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images#How_do_you_create_responsive_images

---
screen shot 2018-10-19 at 12 30 06 pm

hpjaj:
Closing was we favor sticking with the PNG, as the SVG does not render well.

PR #8802, opened 2018-10-19
Va gov/accordion margin fix

RobAWilkinson:
## Description
After digging into this, seemed like there were both issues with errant css rules, and also inconsistent markup throughout content pages.
Based on the [usds guidelines](https://designsystem.digital.gov/components/accordions/) the initial `ul` element needs the `usa-accordion` class. No need for a wrapping div.
Once I had this in place things started clicking

## Testing done
Visual regression testing

## Screenshots
screenshot_2018-10-19 types of veteran id cards

## Acceptance criteria
- [ ] Accordions have proper spacing


src/platform/site-wide/sass/merger.scss
```diff
@@ -528,6 +528,11 @@
 
   // TODO: add this into formation
   .va-sidebarnav {
+    ul.usa-accordion {
+      > li {
+        margin-bottom: -1px;
```
RobAWilkinson: need this so that way top level accordion buttons in sidenav have proper border lines

src/platform/site-wide/sass/modules/_m-footer-brand-consolidation.scss
```diff
@@ -97,6 +97,7 @@
   > .va-footer-links {
     > li {
       margin-bottom: 0.5em;
+      list-style: none;
```
RobAWilkinson: fixes regression where footer list items were getting a list-style

ncksllvn: Looks like this is being addressed here, https://github.com/department-of-veterans-affairs/vets-website/pull/8804, as well.

ncksllvn (approved):
Eliminates a lot of technical debt - awesome work! Should we increase the original estimate in the ticket based on the amount of work that was involved here?

src/platform/site-wide/sass/modules/_m-footer-brand-consolidation.scss
RobAWilkinson: ahh good catch @ncksllvn no conflicts though, let me pull master back in

PR #8809, opened 2018-10-19
526 service end date mg

markgreenburg:
## Description
Adds a validation for ensuring that service end dates are in the past

## Testing done
- Tested locally
- Unit tests for the validator

## Screenshots
screen shot 2018-10-19 at 10 56 20 am

## Acceptance criteria
- [x] Military service history must be in the past

## Definition of done

### DIFF
--- a/src/applications/disability-benefits/526EZ/config/form.js
+++ b/src/applications/disability-benefits/526EZ/config/form.js
@@ -79,7 +79,7 @@ import { FIFTY_MB } from '../../all-claims/constants';
 import { treatmentView } from '../../all-claims/content/vaMedicalRecords';
 import { evidenceTypeHelp } from '../../all-claims/content/evidenceTypes';
 import { additionalDocumentDescription } from '../../all-claims/content/additionalDocuments';
-import { requireOneSelected } from '../validations';
+import { requireOneSelected, isInPast } from '../validations';
 
 import { validateBooleanGroup } from 'us-forms-system/lib/js/validation';
 import PhoneNumberWidget from 'us-forms-system/lib/js/widgets/PhoneNumberWidget';
@@ -185,7 +185,7 @@ const formConfig = {
                 serviceBranch: {
                   'ui:title': 'Branch of service',
                 },
-                dateRange: _.merge(
+                dateRange: merge(
                   dateRangeUI(
                     'Service start date',
                     'Service end date',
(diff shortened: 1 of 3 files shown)


src/applications/disability-benefits/526EZ/config/form.js
```diff
@@ -1,4 +1,5 @@
 import _ from '../../../../platform/utilities/data';
+import merge from 'lodash/merge';
```
cvalarida: Eh, we really should write this thing.

src/applications/disability-benefits/526EZ/validations.js
```diff
@@ -19,3 +19,10 @@ export function requireOneSelected(
     );
   }
 }
+
+export function isInPast(errors, fieldData) {
+  const fieldDate = new Date(fieldData);
```
cvalarida: I think this is subject to some controversy. From [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date):
> Note: parsing of date strings with the Date constructor (and Date.parse, they are equivalent) is strongly discouraged due to browser differences and inconsistencies. Support for RFC 2822 format strings is by convention only. Support for ISO 8601 formats differs in that date-only strings (e.g. "1970-01-01") are treated as UTC, not local.

That said, we should be comparing user-clock UTC (from `new Date(fieldData)` -> `.getTime()`) to user-clock UTC (`Date.now()`), so it _should_ be alright unless the user's clock is way off. :crossed_fingers:

cvalarida (approved):
I spent a while thinking about the date comparison, but I think it'll be alright. I'd like to use `date-fns`, but I think we probably shouldn't add that dependency until we have a migration path to stop using `moment`.

src/applications/disability-benefits/526EZ/validations.js
markgreenburg: That was my thinking. Given that it's probably pretty unlikely that someone fills one of these forms out within hours of discharge, I think this is safe enough. If you think it'd be safer to just `moment` everything, I'm open to that as well. Not sure if we have a convention we follow in other places.

cvalarida: I think we generally use `moment` everywhere, but I'd like to get away from that if we can help it--if for nothing else, just because of the size of the library.

PR #8810, opened 2018-10-19
Remove "Community Care" option from Facility Locator in Brand Consolidation

ncksllvn:
## Description
Preview is a production environment, but a different build-type than `production`, because that refers to Vets.gov. We need an additional check for Preview to keep the "Community Care" option from appearing on https://preview.va.gov/facilities.

Moves to validate - https://github.com/department-of-veterans-affairs/vets.gov-team/issues/14318

## Acceptance criteria
- [x] Community Care option does not exist on https://preview.va.gov/facilities

## Definition of done

### DIFF
--- a/src/applications/facility-locator/config.js
+++ b/src/applications/facility-locator/config.js
@@ -20,7 +20,7 @@ export const api = {
  * existing Facility Locator App.
  */
 export const ccLocatorEnabled = () => {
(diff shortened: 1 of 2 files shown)


U-DON (approved):
LGTM. Though I think it would be better for @r-mc2 to confirm whether disabling this feature on preview is correct.

jbalboni (approved):

PR #8816, opened 2018-10-19
Include uswds in bundle, load conditionally

jbalboni:
## Description

We decided to start include the USWDS JS everywhere, as a separate script tag, which is confusing. Including it as a conditional import in our bundles is less confusing (still would be better not to have it on every page).

## Testing done
Opened locally, checked that accordions still work

## Acceptance criteria
- [x] Accordions still work outside of React apps

## Definition of done

### DIFF
--- a/src/platform/site-wide/index.js
+++ b/src/platform/site-wide/index.js
@@ -22,6 +22,10 @@ if (!brandConsolidation.isEnabled()) {
   require('./usa-banner-toggle');
 }
 
+if (brandConsolidation.isEnabled()) {
+  import(/* webpackChunkName: "uswds" */ 'uswds');
+}
+
 /**
  * Start up the site-wide components that live on every page, like
  * the login widget, the header menus, and the feedback widget.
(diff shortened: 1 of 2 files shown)


src/platform/site-wide/index.js
```diff
@@ -22,6 +22,10 @@ if (!brandConsolidation.isEnabled()) {
   require('./usa-banner-toggle');
 }
 
+if (brandConsolidation.isEnabled()) {
+  import(/* webpackChunkName: "uswds" */ 'uswds');
```
jbalboni: @rianfowler Is this going to break in Teamsite? Seems like it could for subdomain pages.

ncksllvn (approved):
LGTM!

rianfowler (approved):

src/platform/site-wide/index.js
rianfowler: I don't know if this will break team site.  I'll test it and see what happens

rianfowler: This seems fine. I built this branch and proxied the vendor and proxy-rewrite entry files on `http://www.cem.va.gov` and I was able to get the header to show up. 

Also verified that the header still works on preview.va.gov/health with those two scripts.

PR #8818, opened 2018-10-19
Remove dead feedback link

jstanek:
The feedback link that appears on the FAQ page doesn't lead anywhere. Remove the link until we decide on if and how to provide users with opportunities to provide feedback. 

screen shot 2018-10-19 at 3 39 07 pm


bethpotts:
@jstanek You'll need to remove that last sentence too (in both directories), since we're not inviting feedback
"We welcome your feedback too."

bethpotts (approved):
@jstanek Looks good! You can ask Nick to review and merge this.

PR #8820, opened 2018-10-19
14173 absolute image paths

rianfowler:
## Description
- team site implementation requires using absolute paths on markup generating components 

## Testing done
- tested preview build of vender-entry + proxy-rewrite against https://www.cem.va.gov and https://preview.va.gov/health via proxied scripts 
- tested brand consolidation watch task 

## Screenshots
screen shot 2018-10-19 at 14 16 54

## Acceptance criteria
- [ ]

## Definition of done


src/applications/proxy-rewrite/partials/header.js
```diff
@@ -1,3 +1,5 @@
+import { replaceWithStagingDomain } from '../../../platform/utilities/environment/stagingDomains';
```
rianfowler: Decided to leverage what we have laying around.

src/platform/utilities/environment/stagingDomains.js
```diff
@@ -9,17 +9,25 @@ if (__BUILDTYPE__ === 'preview') {
...
+  },
   { from: 'https://www\\.myhealth\\.va\\.gov', to: mhvBaseUrl() },
 ];
 
 // This doesn't include preview because we want to redirect to staging urls
 // in preview
 const prodEnvironments = new Set(['production']);
 
-function replaceWithStagingDomain(href) {
+export function replaceWithStagingDomain(href) {
```
rianfowler: Decided to reuse this- exported it and modified the reg ex behavior a little. It had a regex for processing www.va.gov links but it doesn't look like we were using that rule before this change.

src/platform/utilities/environment/stagingDomains.js
```diff
@@ -9,17 +9,25 @@ if (__BUILDTYPE__ === 'preview') {
...
+  currentEnv = 'www';
+}
+
 // This list also exists in script/options.js
 const domainReplacements = [
-  { from: 'www\\.va\\.gov', to: `${currentEnv}.va.gov` },
+  {
+    from: 'https://www\\.va\\.gov',
+    // use relative links on dev to accomodate localhost
+    to: currentEnv === 'dev' ? '' : `https://${currentEnv}.va.gov`,
```
rianfowler: Verified production, preview, and localhost builds.

src/platform/site-wide/mega-menu/mega-menu-link-data.json
```diff
@@ -10,19 +10,19 @@
                         "links": [
                             {
                                 "text": "About VA Health Benefits",
-                                "href": "/health-care/about-va-health-benefits/"
+                                "href": "https://www.va.gov/health-care/about-va-health-benefits/"
```
jbalboni: I think this is going to be weird on local machines, because I believe they will be re-written to dev.va.gov and not localhost. We might need to change that rewrite logic if it's going to affect links in the menus that we use more often.

rianfowler: I added some logic to remove the `http://www.va.gov` from dev builds so it should point to the relative path: localhost locally or `dev.va.gov` when deployed. I verified it against the watch task.

rianfowler: I really want to come up with a better way but I'd like to work for all of our needs vs creating yet another way to make urls (I think we have at least 2 right now).

src/applications/proxy-rewrite/partials/header.js
```diff
@@ -54,7 +62,9 @@ export default `
         <div class="va-crisis-line-inner">
           <span class="va-crisis-line-icon" aria-hidden="true"></span>
           <span class="va-crisis-line-text">Talk to a Veterans Crisis Line responder now</span>
-          <img class="va-crisis-line-arrow" src="/img/arrow-right-white.svg" aria-hidden="true"></img>
+          <img class="va-crisis-line-arrow" src="${replaceWithStagingDomain(
+            'https://www.va.gov/img/arrow-right-white.svg',
+          )}" aria-hidden="true"></img>
```
cehsu: Could this be a self-closing tag?

src/applications/proxy-rewrite/partials/header.js
```diff
@@ -17,7 +19,9 @@ export default `
         <div class="usa-accordion">
           <div class="usa-banner-header">
             <div class="usa-grid usa-banner-inner">
-            <img src="/img/tiny-usa-flag.png" alt="U.S. flag">
+            <img src="${replaceWithStagingDomain(
```
cehsu: Would it make sense to create a helper that handles the inclusion of `https://www.va.gov`, so that only the relative path needs to be passed?

cehsu (approved):
LGTM.

PR #8829, opened 2018-10-20
Create the Brand Consolidation announcement dialog

ncksllvn:
## Description
The Vets.gov -> VA.gov redirects are (obviously) not in place, but once they are, the user should see a modal explaining to them that Vets.gov is now part of VA.gov. This PR implements that announcement.

Moves to validate - https://github.com/department-of-veterans-affairs/vets.gov-team/issues/12060

### Todo
- [x] Add the "Vets.gov + VA.gov" graphic into the modal header
- [x] Implement `document.referrer`, or some other method of determining if they were redirect to VA.gov from Vets.gov
    - We'll likely need some sort of dummy Vets.gov page that always redirects to Preview in order to test this.
- [x] Design sign-off
- [x] Make sure Vets.gov redirect implementation is proceeding

## Testing done
- [x] Confirmed that this dialog shows on WBC, but not on Vets.gov
- [x] Shows only when redirected from a Vets.gov page

### Follow-up
The frequency of this modal being shown is being addressed here, https://github.com/department-of-veterans-affairs/vets.gov-team/issues/14556.

## Screenshots
image

## Acceptance criteria
- [x] Welcome modal is shown when the user is redirected from a Vets.gov page to its new VA.gov location

## Definition of done


ncksllvn:
@accbjt For reference, this is how I usually layout a work-in-progress PR. It's a tool mostly for my own tracking, because I can proceed with something else while this is awaiting some graphics.

accbjt:
@ncksllvn Thanks, I will try to start doing this.

content/pages/va-gov-redirect.html
```diff
@@ -0,0 +1,10 @@
+---
+production: false
+---
+<!doctype html>
+<html>
+  <head>
+      <meta http-equiv=refresh content="1; url=https://staging.va.gov">
+      <link rel=canonical href="/">
+  </head>
+</html>
```
ncksllvn: This will be used for validating that the onboarding modal is working, while the full Vets.gov-to-VA.gov redirects are being implemented.

content/pages/va-gov-onboarding-modal-check.html
```diff
@@ -0,0 +1,10 @@
+---
+production: false
+---
+<!doctype html>
+<html>
+  <head>
+      <meta http-equiv=refresh content="1; url=http://staging.va.gov">
+      <link rel=canonical href="/">
+  </head>
+</html>
```
ncksllvn: Because this modal is powered by `document.referrer`, I think we should have some sort of test for that modal appearing when redirected from Vets.gov to VA.gov. This will allow us to confirm that the onboarding modal appears when we navigate to `https://staging.vets.gov` and are redirected to `https://staging.va.gov`.

ncksllvn:
To test this, you can add `?onboarding-modal` to the end of a VA.gov URL.

ncksllvn:
PTAL when you can, @jbalboni @RobAWilkinson

src/platform/site-wide/announcements/components/VAPlusVetsModal.jsx
```diff
@@ -0,0 +1,59 @@
...
+
+    const referrer = document.referrer;
+    let wasRedirectedFromVets = !!referrer && referrer.includes('vets.gov');
+
+    // Allow an override on the URL to force the Onboarding Modal to appear for testing purposes.s
+    if (__BUILDTYPE__ !== 'preview') {
+      wasRedirectedFromVets = window.location.search.includes(
+        'onboarding-modal',
+      );
+    }
```
ncksllvn: This isn’t really the way the enabled property was intended to work for Announcements, but because this modal should take priority over all others and appear every time, it was a weird situation.

src/platform/site-wide/announcements/containers/Announcement.jsx
```diff
@@ -12,10 +12,14 @@ class Announcement extends React.Component {
...
-      announcement: { name: announcementName, relatedAnnouncements = [] },
+      announcement: {
+        name: announcementName,
+        showEverytime = false,
+        relatedAnnouncements = [],
+      },
     } = this.props;
 
-    this.props.dismissAnnouncement(announcementName);
+    this.props.dismissAnnouncement(announcementName, showEverytime);
```
ncksllvn: Announcements were designed to be dismissed once, and that dismissal to persist via localstorage. In this case, however, we're showing the dialog every time, so I had to add a `showEverytime` flag.

src/platform/site-wide/announcements/config/index.js
```diff
@@ -2,10 +2,18 @@ import DashboardIntro from '../components/DashboardIntro';
...
 
 const config = {
   announcements: [
+    {
+      name: 'brand-consolidation-va-plus-vets',
+      paths: /(.)/,
+      component: VAPlusVetsModal,
+      disabled: !VAPlusVetsModal.isEnabled(),
+      showEverytime: true,
+    },
```
ncksllvn: I moved this to the top of the list because it has highest priority. It does more than a typical Announcement - instead of just rendering based on the path, it needs to check the referrer - so I leveraged the `disabled` property to have it either render or seamlessly disable itself.

src/platform/site-wide/announcements/components/VAPlusVetsModal.jsx
```diff
@@ -0,0 +1,59 @@
...
+import brandConsolidation from '../../../brand-consolidation';
+
+export default class VAPlusVetsModal extends React.Component {
+  static isEnabled() {
+    if (!brandConsolidation.isEnabled()) return false;
+
+    const referrer = document.referrer;
+    let wasRedirectedFromVets = !!referrer && referrer.includes('vets.gov');
+
+    // Allow an override on the URL to force the Onboarding Modal to appear for testing purposes.s
```
jbalboni: Typo at the end

src/platform/site-wide/announcements/components/VAPlusVetsModal.jsx
```diff
@@ -0,0 +1,59 @@
...
+        </p>
+        <p>
+          With the new VA.gov, we're making it easier for you to access and
+          manage your VA benefits and health care services—all in one place.
+          Sign in with your My HealtheVet, DS Logon, or ID.me account to track
+          your claims, refill your prescriptions, and more.
+        </p>
+        <button
+          type="button"
+          aria-label="Dismiss this announcement"
```
jbalboni: Have you checked this in VoiceOver? This label text seems like it would be confusing combined with the text inside the element.

src/platform/site-wide/announcements/sass/style.scss
```diff
@@ -25,3 +25,12 @@
     max-width: 86px;
   }
 }
+
+.announcement-brand-consolidation {
```
jbalboni: Do you need this nesting?

src/platform/site-wide/announcements/config/index.js
jbalboni: Other announcement components I think return null if they shouldn't be rendered. Does that not work here? Or is this a better pattern?

src/platform/site-wide/announcements/actions/index.js
```diff
@@ -32,8 +32,9 @@ export function initDismissedAnnouncements() {
   };
 }
 
-export function dismissAnnouncement(announcement) {
-  localAnnouncements(announcement);
+export function dismissAnnouncement(announcement, showEverytime) {
+  if (!showEverytime) localAnnouncements(announcement);
```
jbalboni: Is there a better name for the `localAnnouncements` function? Or can we break it into two functions, since it seems to do two different things? Since we're separating the idea of dismissing the modal from whether it displays again, I think tweaking the function names to make things more clear is useful.

jbalboni (approved):
Some questions, but LGTM.

src/platform/site-wide/announcements/config/index.js
ncksllvn: Announcements were intended to be really simple, and just to be configured for a certain path. If it returns null in the render function, it'll eclipse the other announcements - https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/platform/site-wide/announcements/selectors.js#L13. Maybe it would be better to spruce up how Announcements work in general? This feels like a temporary edge case, but I'm not sure.

src/platform/site-wide/announcements/components/VAPlusVetsModal.jsx
ncksllvn: Ah you're right, I'll update that.

PR #8832, opened 2018-10-22
Search V2 - dev/staging only

bshyong:
## Description
Search V2 product for dev and staging.

Note: Search box in header needs to be modified once Search V2 product has been validated in staging and is ready for production

## Testing done
Local testing with mocked API responses

## Screenshots
screen shot 2018-10-22 at 8 55 53 am

## Acceptance criteria
- [x] Search V2 fetching results from search.gov API through vets-API

## Definition of done

Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/13440


src/applications/search/constants/index.js
```diff
@@ -0,0 +1,5 @@
+export const BASE_URI = 'https://search.usa.gov/api/v2/search';
+export const AFFILIATE = 'vets.gov_search';
```
kfrz: I think this should be 'va' - - right @hpjaj ?

kfrz (approved):
Looks great! One small question re: which affiliate we want to use in the `constants.js` file, other than that LGTM.

src/applications/search/constants/index.js
bshyong: @kfrz Are these deprecated because we are using vets-api for the endpoint now instead of hitting search.gov directly? I'm not using any of the constants here except for page size

src/applications/search/manifest.json
```diff
@@ -0,0 +1,8 @@
+{
+  "appName": "Search",
+  "entryFile": "./search-entry.jsx",
+  "entryName": "search",
+  "rootUrl": "/search",
+  "production": false,
+  "contentPage": "search/index.md"
+}
```
ncksllvn: Because of some the work with the build script, the `contentPage` and `production` are actually deprecated now and can be removed.

va-gov/pages/search.md
```diff
@@ -0,0 +1,6 @@
+---
+title: Search Results
+layout: page-react.html
+entryname: search
+includeBreadcrumbs: true
+---
```
ncksllvn: I think we need a `production: false` and possibly `preview: false` here to keep it out of Vets.gov and VA.gov.

src/applications/search/containers/SearchApp.jsx
```diff
@@ -0,0 +1,237 @@
...
+          href={result.url}
+          dangerouslySetInnerHTML={{
+            __html: formatResponseString(result.title),
+          }}
+        />
+        <p className="result-url">{result.url}</p>
+        <p
+          className="result-desc"
+          dangerouslySetInnerHTML={{
+            __html: formatResponseString(result.snippet),
```
ncksllvn: Is this for sure what we want to be doing here? Not sure if there are already HTML entities or something in there that would be escaped twice, but if it's just to apply the bolded text as done in the `formatResponseString` function, I'm wondering it we could use `string.replace` to iterate though and return JSX. Using `dangerouslySetInnerHTML` to render results from an outside service sounds scary to me 😨

src/applications/search/components/SimplePagination.jsx
```diff
@@ -0,0 +1,23 @@
...
+      </a>
+    )}
+    <span className="current-page">
+      Page {prevOffset ? prevOffset / PAGE_SIZE + 1 : 1}
+    </span>
+    {nextOffset && (
+      <a href="#" rel="next" onClick={handlePageChange(nextOffset)}>
+        Next »
+      </a>
+    )}
```
ncksllvn: Since the page changes, it seems like an `<a>` is the right element, but I'm wondering if that `href` value should be calculated or not. Probably low priority, but wondering if there's an accessibility hit.

src/applications/search/containers/SearchApp.jsx
```diff
@@ -0,0 +1,237 @@
...
+  };
+
+  /* eslint-disable arrow-body-style */
+  handlePageChange = offset => {
+    return e => {
+      e.preventDefault();
+      e.persist();
+      this.setState({ offset }, () => this.handleSearch(e));
+    };
+  };
```
ncksllvn: Never seen `event.persist()` before! How does it work here?

ncksllvn (commented):
Left a few comments, but looks good overall. Mostly concerned if there's anything else we can do besides `dangerouslySetInnerHTML`.

src/applications/search/components/SimplePagination.jsx
bshyong: I debated constructing a proper href here, but given that that results pages are transient and prone to change, used the simpler approach of handling the click and having the router handle the page change instead. Also has a better UX because this doesn't result in a full page reload

ncksllvn (approved):
LGTM, although I think we should consider revisiting `dangerouslySetInnerHTML` before going to production.

PR #8861, opened 2018-10-23
Homepage spacing adjustments

jbalboni:
## Description

Updates some element spacing on the homepage. Also moves some styles out to the homepage scss file instead of merger.scss.

## Testing done
Tested locally at different screen sizes.

## Screenshots
localhost_3001_ 7

## Acceptance criteria
- [x] Homepage updates look correct at desktop size
- [x] Homepage updates look correct at tablet size
- [x] Homepage updates look correct at phone size

## Definition of done


src/platform/site-wide/sass/modules/_m-homepage.scss
```diff
@@ -5,16 +5,75 @@
...
+  margin-bottom: 1em;
+
+  @include media($medium-large-screen) {
+    margin-bottom: 1.5em;
+  }
+}
+
+.hub-links {
+  &-row {
+    display: flex;
```
jbalboni: I changed these cards to use flexbox instead of the grid. The grid had gutters that are percentage based, which we can't match with the bottom margin. And making them a fixed size doesn't play all that well with the grid. The flexbox approach is fairly simple with a fixed margin.

va-gov/layouts/home.html
```diff
@@ -2,17 +2,16 @@
 <main>
 
   <!-- the hub -->
-  <section id="homepage-hub">
+  <section class="homepage-hub">
```
jbalboni: We should avoid styling ids.

RobAWilkinson (approved):
Thanks for pulling usa-grid

PR #8864, opened 2018-10-23
Copy MVP confirmation page and logic

cvalarida:
## Description
This copies over the logic and content from the MVP minus the wording for "increased" conditions. I also removed the 4142 content since we'll be integrating that form right into the middle of this one.

## Testing done
Manually tested and copied the unit tests.

### Scenario 1: Successfully submitted to EVSS
image
Because I couldn't get a successful submission yet, I had to mock this out. Act like there's a confirmation number where it says there should be. The code is there, I promise.

### Scenario 2: Check back later
image

### Scenario 3: Non-retryable error
image

## Acceptance criteria
- [x] All three scenarios show up under the right circumstances

## Definition of done

### DIFF
--- a/src/applications/disability-benefits/all-claims/constants.js
+++ b/src/applications/disability-benefits/all-claims/constants.js
@@ -141,3 +141,21 @@ export const PTSD = 'ptsd';
 export const NINE_ELEVEN = '2001-09-11';
 
 export const ERR_MSG_CSS_CLASS = '.usa-input-error-message';
+
+export const submissionStatuses = {
+  // Statuses returned by the API
+  pending: 'submitted', // Submitted to EVSS, waiting response
+  retry: 'retrying',
+  succeeded: 'received', // Submitted to EVSS, received response
+  exhausted: 'exhausted', // EVSS is down or something; ran out of retries
+  failed: 'non_retryable_error', // EVSS responded with some error
+  // When the api serves a failure
+  apiFailure: 'apiFailure',
+};
+
+export const terminalStatuses = new Set([
+  submissionStatuses.succeeded,
+  submissionStatuses.exhausted,
+  submissionStatuses.retry,
+  submissionStatuses.failed,
+]);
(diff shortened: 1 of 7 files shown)


src/applications/disability-benefits/all-claims/constants.js
```diff
@@ -141,3 +141,21 @@ export const PTSD = 'ptsd';
 export const NINE_ELEVEN = '2001-09-11';
 
 export const ERR_MSG_CSS_CLASS = '.usa-input-error-message';
+
+export const submissionStatuses = {
```
markgreenburg: ARE WE NOT GOING WITH CAPS LOCK FOR THESE?

src/applications/disability-benefits/all-claims/constants.js
```diff
@@ -141,3 +141,21 @@ export const PTSD = 'ptsd';
...
+  // Statuses returned by the API
```
markgreenburg: Does this duplicate the constants in Increase Only? Should we just point those refs to the All Claims constants file and remove those old ones?

src/applications/disability-benefits/all-claims/constants.js
```diff
@@ -141,3 +141,21 @@ export const PTSD = 'ptsd';
...
+  pending: 'submitted', // Submitted to EVSS, waiting response
+  retry: 'retrying',
+  succeeded: 'received', // Submitted to EVSS, received response
+  exhausted: 'exhausted', // EVSS is down or something; ran out of retries
+  failed: 'non_retryable_error', // EVSS responded with some error
+  // When the api serves a failure
+  apiFailure: 'apiFailure',
+};
+
+export const terminalStatuses = new Set([
```
markgreenburg: same question here

markgreenburg (approved):
LGTM. I wish we could reduce the duplication of the polling wrapper and `<ConfirmationPage/>` but I'm not sure there's a way to do this without unnecessary work, since each relies on the specific versions of its children that it imports from either `all-claims/` or `526EZ/`

src/applications/disability-benefits/all-claims/constants.js
cvalarida: Hehehe, I thought about it. I just didn't really care for `SUBMISSION_STATUSES.PENDING` or whatever. :man_shrugging:

I'm alright with changing them. I just don't like the way it looks.

PR #8873, opened 2018-10-24
Disable interactivity in messaging and Rx cards

kfrz:
## Description
Removed interactive elements from the MessagingWidget and PrescriptionCard components, as they are linking to MHV in an unexpected manner, and this work is out of scope for WBC. See [#14499](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/14499) and [#13839](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/13839) for more context, as well as [this DSVA slack thread](https://dsva.slack.com/archives/C909ZG2BB/p1540314756000100)

## Testing done
Manual testing

## Screenshots
screencast-localhost-3001-2018 10 24-07-18-06

## Acceptance criteria
- [x] Remove all clickable actions (including hiding buttons) on the prescriptions health widget other than leaving the "View all.." link as is
- [x] Remove all clickable actions on the secure messages health widget other than leaving the "View all.." link as is

## Definition of done

### DIFF
--- a/src/applications/personalization/dashboard/containers/MessagingWidget.jsx
+++ b/src/applications/personalization/dashboard/containers/MessagingWidget.jsx
@@ -38,8 +38,12 @@ class MessagingWidget extends React.Component {
       { label: 'Date', value: 'sentDate', nonSortable: true },
     ];
 
+    // eslint-disable-next-line
     const makeMessageLink = (content, id) => (
-      <Link href={`/health-care/messaging/inbox/${id}`}>{content}</Link>
+      // Messaging temporarily disabled.
+      // See: https://github.com/department-of-veterans-affairs/vets.gov-team/issues/14499
+      // <Link href={`/health-care/messaging/inbox/${id}`}>{content}</Link>
+      <Link>{content}</Link>
     );
 
     let { messages } = this.props;
--- a/src/applications/messaging/sass/partials/_messaging-folder.scss
+++ b/src/applications/messaging/sass/partials/_messaging-folder.scss
@@ -73,6 +73,8 @@
     &:visited {
       color: $color-black;
     }
+
+    cursor: default;
   }
 
   td {
--- a/src/applications/personalization/dashboard/components/PrescriptionCard.jsx
+++ b/src/applications/personalization/dashboard/components/PrescriptionCard.jsx
@@ -1,19 +1,21 @@
-import { Link } from 'react-router';
+// import { Link } from 'react-router';
 import PropTypes from 'prop-types';
 import React from 'react';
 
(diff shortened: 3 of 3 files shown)


BBarone1:
copying comment from slack for visibility:

This is fantastic. The only other piece we'd want here is to remove / hide those yellow highlighting boxes on messages so the user doesn't get confused and think they can click.

we'll want to restore those highlighting boxes and the buttons on Rx in the future, so we want to keep this flexible to reactivate at that time

kfrz:
cc: @1Copenut 

Do we need to do anything more specific here to keep 508-compliant? We're still rendering `<Link>` components for the messaging entries, but we want the link to be completely disabled, thus removing `href` attribute. The elements are no longer in the navigation tab order but content is in-flow. Thanks for the :eyes:

src/applications/personalization/dashboard/containers/MessagingWidget.jsx
```diff
@@ -38,8 +38,12 @@ class MessagingWidget extends React.Component {
       { label: 'Date', value: 'sentDate', nonSortable: true },
     ];
 
+    // eslint-disable-next-line
```
erikphansen: Why is this needed? To silent the warning about the unused `id` param?

erikphansen (commented):
Is this functionality going to be restored soon?

src/applications/personalization/dashboard/containers/MessagingWidget.jsx
kfrz: > Why is this needed? To silent the warning about the unused id param?

Yep


> Is this functionality going to be restored soon?
Functionality will be restored post WBC - we have work scoped out [here](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/13839) and other tickets.

erikphansen: thanks! As long as that `eslint-disable` is removed when functionality is restored and it's no longer needed it's cool with me.

erikphansen (approved):

1Copenut (approved):
I've reviewed the changes in `localhost` using VoiceOver. They didn't cause any errors with VO, and there were no axe-core errors.

I cannot check these using JAWS or NVDA yet, because of CORS boundaries on serving the site locally via IP. Please ping me when we're in Staging, and I will review with other browsers.

I'm leaving one comment about adding a CSS rule so the secure message links without `href` attributes do not show the click hand, but rather the default arrow. I'd recommend adding it for visual understanding if we can ensure the rule is silo'ed to this one view.

bshyong (approved):
For the Link, you can also replace with `<a></a>` as an alternative

PR #8874, opened 2018-10-24
Use session cookie in API requests instead of authorization headers.

U-DON:
## Description
Requests to the API will now use a cookie for authentication. The auth headers with the token are being dropped completely.

~~This should get merged at the same time as the vets-api PR that stops reading the auth headers.~~

Due to feature toggles on the vets-api changes, this can be merged after department-of-veterans-affairs/devops#3575

## Testing done
Local testing with the corresponding changes on the backend. Also got existing specs to pass.

Tested these permutations:
1. Vets.gov with FE auth headers and BE without session cookie auth.
2. Vets.gov without FE auth headers and BE with session cookie auth.
3. VA.gov with FE auth headers and BE without session cookie auth.
4. VA.gov without FE auth headers and BE with session cookie auth.

## Acceptance criteria
- [x] Logging in should be successful and create a session cookie.
- [x] Logging out should be successful and destroy the session cookie.
- [x] The current set of tools and services should be functional without using the auth headers.

## Definition of done


src/applications/auth/containers/AuthApp.jsx
```diff
@@ -1,82 +1,51 @@
...
-  setToken = () => {
+  handleAuthError = () => {
+    this.setState({ error: true });
+  };
+
+  handleAuthSuccess = payload => {
+    setupProfileSession(payload);
+
+    const parent = window.opener;
+    parent.postMessage('loggedIn', environment.BASE_URL);
```
ncksllvn: I was gonna ask if there's a way we can listen for a cookie-change event, but looks like we'd have a similar issue with [IE support](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/cookies/onChanged).

src/platform/user/profile/utilities/index.js
```diff
@@ -63,19 +65,36 @@ export function mapRawUserDataToState(json) {
   };
 }
 
+// Flag to indicate an active session for initial page loads.
+// It's distinct from the currentlyLoggedIn state, which
+// serves as confirmation that the user is logged in and
+// as a trigger to properly update any components that subscribe to it.
+export const hasSession = () => localStorage.getItem('hasSession');
```
ncksllvn: I wonder if sometime down the road this value could be part of the Redux store in `initialState`, and used in conjunction with `isLoggedIn` and maybe `sessionExpired` or something.

ncksllvn (approved):
LGTM.

src/applications/auth/containers/AuthApp.jsx
U-DON: Oh man, that would've been perfect...

src/platform/user/profile/utilities/index.js
U-DON: We need the flag in `localStorage` in order to to know at the beginning of a page load whether we should be making an API request at all (the initial fetch user). If we only have it in the Redux store, we would need it to be set to `true` in reaction to a successful API request, but we wouldn't know whether to make that request in the first place without some flag already being set.

PR #8875, opened 2018-10-24
All claims pow move mg

markgreenburg:
## Description
- Moves POW pages to new location
- Adds checkboxes for associating a disability (or disabilities) with POW status

Incidentally:
- Updates some forms test helpers

## Testing done
- [x] Tested locally
- [x] Unit tests

## Screenshots
Normal page:
screen shot 2018-10-23 at 4 48 11 pm

If veteran selects 'yes':
screen shot 2018-10-30 at 3 26 17 pm

If veteran selects 'yes' AND added some new disabilities earlier:
screen shot 2018-10-30 at 12 45 07 pm

## Acceptance criteria
- [x] If user POW, allow to select disabilities from list to associate with POW status
- [x] Only new disabilities should show in list
- [x] User does not need to select any disability from the list
- [x] Disabilities checkboxes & related question should not show if user has no new disabilities

## Definition of done


src/applications/disability-benefits/all-claims/utils.jsx
```diff
@@ -223,6 +219,19 @@ export const addCheckboxPerDisability = (form, pageSchema) => {
...
 
+export const addCheckboxPerNewDisability = createSelector(
+  form =>
+    form.newDisabilities && Array.isArray(form.newDisabilities)
+      ? form.newDisabilities
+      : [],
+  newDisabilities => ({
+    properties: newDisabilities.reduce((accum, disability) => {
+      const formattedName = getDisabilityName(disability.condition);
+      return _.set(`${formattedName}`, { type: 'boolean' }, accum);
```
cvalarida: Do we need the string template here?

peggygannon (changes requested):
@markgreenburg the screenshots posted don't match the final text in this ticket: 
https://app.zenhub.com/workspace/o/department-of-veterans-affairs/vets.gov-team/issues/14268

src/platform/testing/unit/schemaform-utils.jsx
```diff
@@ -315,14 +315,15 @@ export function selectRadio(form, fieldName, value) {
...
  * @param {string} dateString The date to fill in the input
  */
-export function fillDate(form, partialId, dateString) {
+export function fillDate(form, partialName, dateString) {
   const date = dateString.split('-');
-  const month = form.find(`select#${partialId}Month`);
-  const day = form.find(`select#${partialId}Day`);
-  const year = form.find(`input#${partialId}Year`);
+
+  const month = form.find(`select[name="${partialName}Month"]`);
```
cvalarida: What do you think of using an "ends with" selector for this? `[name$="${partialName}Month"`. This way we wouldn't have to pass in the whole name (minus `Month`, `Day`, `Year`).

src/applications/disability-benefits/all-claims/validations.js
```diff
@@ -126,6 +126,10 @@ export const validateIfHasEvidence = (
 
 export const hasNewPtsdDisability = formData =>
   some(_.get('newDisabilities', formData, []), item => {
+    if (!_.get('view:newDisabilities', formData, false)) {
```
cvalarida: If `view:newDisabilities` is false, will `newDisabilities` be more than an empty array? I'm not sure how this works if they enter some information only to go back and select no to the "do you have any new disabilities" question.

src/applications/disability-benefits/all-claims/utils.jsx
```diff
@@ -263,3 +272,9 @@ const post911Periods = createSelector(
 );
 
 export const servedAfter911 = formData => !!post911Periods(formData).length;
+
+export const getPOWDisabilities = createSelector(
```
cvalarida: Can we reuse this anywhere else, or is it specific to PoW?

src/applications/disability-benefits/all-claims/pages/prisonerOfWar.js
```diff
@@ -1,27 +1,39 @@
 import fullSchema from '../config/schema';
-
 import dateRangeUI from 'us-forms-system/lib/js/definitions/dateRange';
-
 import PeriodOfConfinement from '../components/PeriodOfConfinement';
+import { addCheckboxPerNewDisability } from '../utils';
 
 export const uiSchema = {
   'ui:title': 'Prisoner of War (POW)',
```
cvalarida: Interesting. I always thought it was `PoW` with a lowercase `o`. :man_shrugging:

cvalarida (approved):
Had lots of little comments, but nothing holding this back.

markgreenburg:
Ah sorry @peggygannon looks like I missed the content changes. Here's a new screenshot:
screen shot 2018-10-31 at 3 49 08 pm

peggygannon (approved):

src/applications/disability-benefits/all-claims/config/form.js
```diff
@@ -201,6 +195,7 @@ const formConfig = {
         },
         newDisabilityFollowUp: {
           title: formData => getDisabilityName(formData.condition),
+          depends: form => form['view:newDisabilities'] === true,
```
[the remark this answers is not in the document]
markgreenburg: Yeah, this was a problem before - if you add new disabilities but then switch the 'do you have new disabilities?' answer to 'no', they persist and you'd still see all the follow up for them.

src/applications/disability-benefits/all-claims/utils.jsx
markgreenburg: d'oh! Nope, definitely not.

src/platform/testing/unit/schemaform-utils.jsx
markgreenburg: Well, doing so doesn't break any existing tests, so there aren't any hard blockers here. But I think doing so goes against how these helpers typically work as a whole - all of them basically require a full ID or name to use, and I'd be worried about introducing inconsistent behavior in case someone down the road tries to use a partial that happens to match another field with a similar name. Thoughts?

src/applications/disability-benefits/all-claims/pages/prisonerOfWar.js
markgreenburg: I thought so as well, but the internet seems to disagree.

src/applications/disability-benefits/all-claims/validations.js
markgreenburg: Yeah, it'll persist, which is why this is necessary.

src/platform/testing/unit/schemaform-utils.jsx
cvalarida: :man_shrugging: I remember making some helper somewhere that took the end of a name. That's how I learned about the `$=` operator. :stuck_out_tongue:

It's a valid concern, though. Totally fine like this.

src/applications/disability-benefits/all-claims/validations.js
cvalarida: Makes sense. Good call. :+1:

PR #8894, opened 2018-10-25
Teamsite integration tests

cehsu:
## Description
These changes set up VRT/integrations tests for teamsite pages using Jest and Puppeteer.
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/14465

## Acceptance criteria
- [x] Set up integration tests for teamsite pages.

## Definition of done


rianfowler (approved):

src/applications/proxy-rewrite/tests/health.integration.spec.js
```diff
@@ -0,0 +1,44 @@
...
+describe('jest-image-snapshot usage with an image received from puppeteer', () => {
+  let browser;
+  jest.setTimeout(20000);
+  beforeAll(async () => {
+    browser = await puppeteer.launch();
+  });
+
+  it('compares screenshots of elements', async () => {
+    const page = await browser.newPage();
+    page.setViewport({ width: 1024, height: 800 });
```
erikphansen: maybe pull the viewport size objects up a level so that can be shared? Or ideally move them to a helper file since I'm sure we'd want to be able to reuse them across future tests.

src/applications/proxy-rewrite/tests/health.integration.spec.js
```diff
@@ -0,0 +1,44 @@
...
+    await page.goto('https://preview.va.gov/health/');
```
erikphansen: This URL should probably be moved to a const that all tests use.

src/applications/proxy-rewrite/tests/health.integration.spec.js
```diff
@@ -0,0 +1,44 @@
...
+    const image = await element.screenshot();
+    expect(image).toMatchImageSnapshot();
+  });
+  it('checks for expanded content', async () => {
+    const page = await browser.newPage();
+    page.setViewport({ width: 1024, height: 800 });
+    await page.goto('https://preview.va.gov/health/');
+    page.click('#usa-banner-toggle');
+    await page.waitForSelector('#gov-banner', {
+      visible: true,
```
erikphansen: does this `visible: true` serve as the actual test?

src/applications/proxy-rewrite/tests/health.integration.spec.js
```diff
@@ -0,0 +1,44 @@
...
+
+describe('jest-image-snapshot usage with an image received from puppeteer', () => {
+  let browser;
+  jest.setTimeout(20000);
+  beforeAll(async () => {
+    browser = await puppeteer.launch();
+  });
+
+  it('compares screenshots of elements', async () => {
+    const page = await browser.newPage();
```
erikphansen: maybe pull this up into a `beforeEach`?

package.json
```diff
@@ -49,6 +49,7 @@
     "test:visual:baseline": "./script/run-nightwatch.sh --visual --command=baseline",
     "test:visual:mobile": "./script/run-nightwatch.sh --visual --mobile",
     "test:visual:mobile:baseline": "./script/run-nightwatch.sh --visual --command=baseline --mobile",
+    "test:integration": "jest",
```
erikphansen: Just thinking ahead here, but is there a way to configure the `jest` command to only run the integration test? I realize that that's _all_ there are right now, but in the future we might have Jest tests for unit tests as well. You're more than welcome to veto this suggestion as it could be considered premature optimization!

erikphansen (approved):
Made a few suggestions with a "looking toward the future" theme. I'm pretty excited about this, though.

package.json
cehsu: Good point, we can specify and rename the config.

src/applications/proxy-rewrite/tests/health.integration.spec.js
cehsu: Good idea, updated.

src/applications/proxy-rewrite/tests/health.integration.spec.js
cehsu: Yes.

src/applications/proxy-rewrite/tests/health.integration.spec.js
cehsu: Good idea, updated.

src/applications/proxy-rewrite/tests/health.integration.spec.js
cehsu: Good idea, updated.

src/applications/proxy-rewrite/tests/health.integration.spec.js
```diff
@@ -0,0 +1,50 @@
...
+
+const fullScreen = { width: 1024, height: 800 };
+const mobileScreen = { width: 400, height: 700 };
+
+const VAGOV_HEALTH_URL = 'https://staging.va.gov/health/';
+
+const HEADER = 'header';
+const FOOTER = 'footer';
+
+describe('jest-image-snapshot usage with an image received from puppeteer', () => {
```
jbalboni: Probably should update this description

src/applications/proxy-rewrite/tests/health.integration.spec.js
```diff
@@ -0,0 +1,50 @@
...
+    await page.waitFor(1000);
+    const image = await element.screenshot();
+
+    expect(image).toMatchImageSnapshot();
+  };
+  beforeAll(async () => {
+    browser = await puppeteer.launch();
+  });
+  beforeEach(async () => {
+    page = await browser.newPage();
```
jbalboni: This might end up being a bottleneck if we write more tests for a particular page. We could probably do more than one test without reloading the page (and actually might benefit from some tests that do multiple things without reloading)

jbalboni (approved):
This is surprisingly simple. Looks good to me!

src/applications/proxy-rewrite/tests/health.integration.spec.js
erikphansen: That's a really good point. Hopefully when we get to that point the suite can be broken into logical pieces: one `describe` for a certain page at a certain size, for example.

src/applications/proxy-rewrite/tests/health.integration.spec.js
cehsu: Good idea, updated.

src/applications/proxy-rewrite/tests/health.integration.spec.js
cehsu: Good idea, we can use the screenshot comparison similar to how aXe checks are performed in the e2e tests. Reorganized.

src/applications/proxy-rewrite/tests/health.integration.spec.js
```diff
@@ -0,0 +1,57 @@
...
+    await page.goto(VAGOV_HEALTH_URL);
+  });
+
+  it('renders full screen size', async () => {
+    page.setViewport(fullScreen);
+
+    compareScreenshots(FOOTER);
+    compareScreenshots(HEADER);
+
+    page.click('#usa-banner-toggle');
```
ncksllvn: 🔥 🔥 🔥 

These interactions are long overdue!

ncksllvn (approved):
LGTM! This was an awesome idea, and a great implementation. Looking forward to seeing how this evolves as we move forward. 

I also wonder if there's any way we can consolidate logic/program flow with VRT? It seems like there may be some functional overlap. One of our ideas to iterate on VRT was to allow for applications to interact with the page instead of just stepping through the sitemap, and I think this opens that door. CC: @RobAWilkinson

PR #8906, opened 2018-10-26
Update external links to always open in new tab with noopener set

jbalboni:
## Description

This is slightly different from the attached issue, but fixing the links in the build is easier than spitting out a ton of errors and forcing people to fix them manually. Plus, this makes it easier on people creating links in Markdown.

On my machine this adds about 5 seconds to the build.

## Testing done

Tested out locally

## Acceptance criteria
- [x] External links are correctly updated
- [x] Internal links are not changed

## Definition of done


script/update-external-links.js
```diff
@@ -0,0 +1,60 @@
...
+          // external link that should always open in a new tab
+          // There is an escape hatch here, too
+          if (
+            !hrefAttr.includes('va.gov') &&
+            !hrefAttr.includes('vets.gov') &&
+            !link.attr('data-no-external')
+          ) {
+            if (targetAttr !== '_blank') {
+              linkUpdated = true;
+              link.attr('target', '_blank');
```
ncksllvn: Should there also be a check to make sure that the target attribute isn't being overridden?

script/update-external-links.js
```diff
@@ -0,0 +1,60 @@
...
+            }
+
+            if (!relAttr || !relAttr.includes('noopener')) {
+              linkUpdated = true;
+              link.attr('rel', `noopener ${relAttr || ''}`);
+            }
```
ncksllvn: Does this logic need to be here too? Looks pretty similar to line 24.

ncksllvn (approved):
A couple questions, but looks good. Looking forward to a potential performance boost!

script/update-external-links.js
jbalboni: Ah, good point. Yeah, I should check that.

script/update-external-links.js
jbalboni: It's the same logic, but this is run for links that should have `target="_blank"` but didn't, and on line 24 it's run on links that already had `target="_blank"`.

ncksllvn:
Just discovered this info about how we determine opening new tabs (thanks @bethpotts) -

> Nav items and links that link to pages that are EXTERNAL TO VA.GOV domain, should open in a new tab/window. Nav items and links that link to pages WITHIN THE VA.GOV domain, should open in the same tab/window. Exceptions to this rule can be made for VA.gov subdomains or pages that have a different design and user experience from the core VA.gov experience. Examples include, but not limited to accesstocare.va.gov, data.va.gov, oit.va.gov.

https://github.com/department-of-veterans-affairs/va.gov-team/blob/master/WBC%20Decisions.md

The work here could be super useful to standardizing this, in case there was interest in adding another JSON file containing a list of domains 😆

ncksllvn (approved):
LGTM!

PR #8910, opened 2018-10-26
Bah 4142 summary submission

aaronjlech:
## Description
This PR covers the Evidence Summary page and data transformation implementation for form 4142/4142a as part of the 526EZ flow. This page follows the Private Medical Records Release (form 4142). The data transformation takes place within the existing transform function for the 526EZ flow in the helpers.js file. 

## Testing done
Manual and Unit

## Screenshots
screen shot 2018-10-24 at 3 57 23 pm

## Acceptance criteria
- [x] List of entered providers displays on the evidence summary page
- [x] Data is successfully submitted to the API

## Definition of done


src/applications/disability-benefits/526EZ/helpers.jsx
```diff
@@ -137,6 +174,16 @@ export function transform(formConfig, form) {
 
   const attachments = additionalDocuments.concat(privateRecords);
 
+  const providerFacility = disabilities
```
markgreenburg: This is actually an array of (potentially) several provider facilities, right?

src/applications/disability-benefits/526EZ/helpers.jsx
```diff
@@ -102,6 +102,43 @@ export const getReservesGuardData = formData => {
   };
 };
 
+/**
+ * Converts the treatment date range into an array of objects from just an object
+ * @param {object} treatmentDateRange object containing from/to date range
+ * @returns {array} of treatmentDateRange's
+ */
+const transformDateRange = treatmentDateRange => [treatmentDateRange];
```
markgreenburg: What's the reason that this transformation is needed? Isn't each provider facility limited to exactly only one date range?

src/applications/disability-benefits/526EZ/helpers.jsx
```diff
@@ -102,6 +102,43 @@ export const getReservesGuardData = formData => {
...
+      facility,
+    ),
+  );
+  return newProviderFacility;
+};
+
+/**
+ * If any limited consent text is populated, collect it.
+ */
+export function gatherLimitedConsentText(disabilities) {
```
markgreenburg: Looking at the paper form, I see what you had to do here. Seems to me that having the limited consent field on the providers page may not be the best fit, since we could unintentionally force veterans to repeat their limited consent for each disability (when in fact there's only one field across all disabilities). To me, a solution for this could be to change the front end to ask the limited consent question on a follow-up page somewhere later in the flow, but maybe that's not ideal for other reasons.

src/applications/disability-benefits/526EZ/helpers.jsx
```diff
@@ -150,6 +197,10 @@ export function transform(formConfig, form) {
     // if there is at least one treatment to send
     ...(treatments.length && { treatments }),
     ...(attachments.length && { attachments }),
+    form4142: {
+      limitedConsent: gatherLimitedConsentText(disabilities),
```
markgreenburg: We probably don't need to send this if the string's empty, but not sure if this even matters or if the implementation is already agreed upon with vets-api.

markgreenburg (commented):
Looks good! I'd like to verify that the date range transformation is necessary, because if it's not, a lot of the transformations could be cut out entirely I'd think.

src/applications/disability-benefits/526EZ/helpers.jsx
aaronjlech: The api can handle a blank string/won't populate anything on the form. So there isn't any harm in returning an empty string, or the need to conditionally check for an empty string.

src/applications/disability-benefits/526EZ/helpers.jsx
aaronjlech: Correct

src/applications/disability-benefits/526EZ/helpers.jsx
aaronjlech: This was approved per requirements, they want to capture limited consent at an aggregate level.

src/applications/disability-benefits/526EZ/helpers.jsx
cvalarida: To avoid confusion, what do you think about using plurals for arrays? `providerFacilities` would then contain (potentially) many provider facilities.

src/applications/disability-benefits/526EZ/helpers.jsx
aaronjlech: They can have multiple date ranges per provider facility. It's easier for the backend to handle it as an array every time, regardless if there is only 1. Looks something along the lines of this:
```js
treatmentDateRange = [ { from: 'fromDate', to: 'toDate' }, { from: 'fromDate2', to: 'toDate2' } ... ]
```

src/applications/disability-benefits/526EZ/helpers.jsx
aaronjlech: @cvalarida I can definitely agree with that, the only problem is the Schema, and backend around it are `providerFacility` so changing the variable name would be the only place we could do it without a lot of overhead. I've made the change to the variable specifically. In the future I'll advise the team that we try to stick to this naming convention to avoid confusion.

src/applications/disability-benefits/526EZ/helpers.jsx
cvalarida: :thinking: Is it a business requirement to allow multiple date ranges per facility? I'm not sure how that would make sense to the veteran.

At any rate, it doesn't look like they can currently enter multiple date ranges. 
image

src/applications/disability-benefits/526EZ/helpers.jsx
cvalarida: :thinking: Thinking about this from the veteran's perspective, this is maybe a little misleading. It looks like the limitation applies to only a single condition, so I can see how the veteran would put in the same thing for each condition.

To gauge the impact of this, I have a few questions:
1) What impact will it have on the veteran if they think they're limiting consent on only a single condition?
    - I'm not sure of the kinds of things a veteran might put in here, so I'm having a hard time thinking of scenarios
2) What impact will it have on the reviewer (and thus the veteran) if they enter the same thing multiple times?
    - The only thing I can think of is it going to the overflow page unnecessarily, but I'd like to make sure that's all

I think I may not be tracking on the "approved per requirements" bit. What do you mean?

src/applications/disability-benefits/526EZ/helpers.jsx
```diff
@@ -102,6 +102,43 @@ export const getReservesGuardData = formData => {
...
+
+/**
+ * If any limited consent text is populated, collect it.
+ */
+export function gatherLimitedConsentText(disabilities) {
+  const fullLimitedConsent = disabilities
+    .filter(disability => disability['view:limitedConsent'])
+    .reduce((accumulator, disability) => {
+      let string = accumulator;
+      string = `${string} ${disability.limitedConsent}`;
```
cvalarida: I mentioned this in the large PR, but I think a space may not be the right delimiter if we are bundling all these strings up to put into a box on a paper form for a person to read. What do you think of a newline character instead?

src/applications/disability-benefits/526EZ/helpers.jsx
```diff
@@ -150,6 +197,10 @@ export function transform(formConfig, form) {
     // if there is at least one treatment to send
     ...(treatments.length && { treatments }),
     ...(attachments.length && { attachments }),
+    form4142: {
+      limitedConsent: gatherLimitedConsentText(disabilities),
+      providerFacility: transformProviderFacilities(providerFacilities),
```
cvalarida: Can you change this to `providerFacilities` to reflect that it's an array?

src/applications/disability-benefits/526EZ/helpers.jsx
```diff
@@ -460,7 +510,7 @@ const listDocuments = documents => (
 export const evidenceSummaryView = ({ formContext, formData }) => {
   const {
     treatments,
-    privateRecordReleases,
+    providerFacility,
```
cvalarida: Same naming concern here.

src/applications/disability-benefits/526EZ/helpers.jsx
aaronjlech: So I probably wasn't very clear here, and your concern with the array of a single dateRange object is valid. The main reasoning behind this is because the API treats it as an array of potential multiple dateRanges. The UI currently does not handle it as such, and as part of MVP it was decided(by our business analysts @Jennywalter ) to only allow them to add one treatmentDateRange as of now, this could change in the future, and the API is setup to support it.

src/applications/disability-benefits/526EZ/helpers.jsx
aaronjlech: Makes sense, but the business decision was made to keep them all on one line of the PDF if possible.

src/applications/disability-benefits/526EZ/helpers.jsx
aaronjlech: So I can't change the key here. The schema was approved a while back for 4142, the `vets-api` and `vets-json-schema` both have this as `providerFacility`. I will keep in mind the concern of plural variable/property names for lists/arrays, and advise our team.

src/applications/disability-benefits/526EZ/helpers.jsx
aaronjlech: image
See 4142 ^^

src/applications/disability-benefits/526EZ/helpers.jsx
Jennywalter: hi @cvalarida, I am the business analyst that solicited the requirements for the limited consent field functionality (and Form 4142 in its entirety) from Paul Shute (our business contact at Comp Service) and Andrea Schneider (VA Digital Service).  There is a process for how we get our requirements approved before development begins, so that the business agrees with the work that we are doing for a given feature/piece of functionality- that is what Aaron meant by 'approved per requirements.'
With that said, it was agreed that for 4142 MVP, the user will go through contention by contention and have the ability to indicate if he/she wants the VA to get his/her private medical records for each contention.  By doing so, it is understood that the user **could** limit consent more than once.  Also, it is important to note that consent can be limited for a variety of things- not just a condition.  That is why we expose a text box for the user to input a description of what he/she would like to limit consent for, if limited consent is selected.
If the Veteran puts in the same limitation for each condition, then that's fine and whatever they put will go in Field 12 of the 4142 form when it is generated.  If information is entered within the limited consent description text box that exceeds the field limit within the actual form, then it will go to the overflow pages.  This will be tested further during internal UAT on staging, once it is available, to ensure it is functioning like expected.
It's important to remember that this is MVP and going through condition by condition for the _526 Claim for Increase_ flow will only be in place until the _526 All Claims_ goes live in the January timeframe.

src/applications/disability-benefits/526EZ/helpers.jsx
markgreenburg: If you wanted to keep a link between disabilities and limited consent, you could add the disability name to the string, so that each string could look something like `${name} - ${limitedConsent}`
That way if someone types something like "only get ___ about this one", the rater would know which disability they're talking about.

Seems like saving space is a concern, though, and I don't know whether that extra information matters.

Jennywalter: hi, me again!  so there is not a business need to have the link between the contention/disability and the limited consent.  That is the reason the description field becomes available in the UI when the user selects to limit consent.  When a user limits consent, it is not always for just a contention/disability.  I recommend that we keep it as is for now.  Thanks!

src/applications/disability-benefits/526EZ/helpers.jsx
cvalarida: Thanks for the clarification, @Jennywalter. :+1:

src/applications/disability-benefits/526EZ/helpers.jsx
cvalarida: Is there some way we can track if this causes problems for reviewers? I'm not sure what that would look like, but my intuition tells me when we concat a bunch of arbitrary strings together and have a human read it, something's bound to get lost in translation eventually.

Seeing as this is pretty high-stakes for veterans, I'd prefer to be cautious, y'know?

Jennywalter: Yes, I can make a note to track it.  Our developer that worked on the 4142 pdf has the overflow working in a way that is very intuitive to the end user (in this case the reviewer of the claim).  So it shouldn't appear as a bunch of arbitrary stings.

src/applications/disability-benefits/526EZ/helpers.jsx
cvalarida: I guess my main concern is this is unnecessarily complicated if we decide to stick with the single date range. If it's on the radar to clean this up if necessary after the decision is made, that's good by me.

src/applications/disability-benefits/526EZ/helpers.jsx
cvalarida: :thinking: As it stand right now, it won't be a bunch of arbitrary strings as far as the back end sees. It'll be a single, consolidated string, that's comprised of a number of arbitrary strings separated by a space. If the veteran doesn't use any punctuation (which will definitely happen) or capitalization (which will very probably happen) on multiple limited consent, there will be no way to tell where one stops an the other begins unless the context is clear, which we can't guarantee.

Without examples, I'm having a hard time saying what I'm trying to say. Sorry about that.

Does that make sense?

If we had another delimiter, it'd be easier to tell what the veteran intended when they entered their limited consents. Maybe `; ` or ` | ` if not a new line?

src/applications/disability-benefits/526EZ/helpers.jsx
aaronjlech: @cvalarida At this point we can't redesign the API to account for taking in a single object, just for the sake of not wrapping it in an array. We're aware that you currently cannot enter in more than one treatmentDateRange, but as this is subject to change, it makes sense to not have to redo *both* the front end and the backend once we need this change implemented.

src/applications/disability-benefits/526EZ/helpers.jsx
Jennywalter: Thanks for the feedback.  At this juncture, we will be keeping the space.  Keep in mind that Limited Consent is not something that the Veteran will typically indicate in the 4142.  It tends to increase the time it takes to get their Private Medical Records, which is not ideal.  We still want to account for handling it, but we do not have time to implement any enhancements at this time.  As I mentioned previously, we will be tracking it. 
Are there any other questions/concerns preventing this PR from being approved?

cvalarida (approved):

PR #8942, opened 2018-10-30
Fix a bug on single-entry production build

cvalarida:
## Description
The build was breaking when I was trying to do a single-entry production build so I could run our bundle analyzer. After debugging a bit, I found this was because when we specify the `--entry` for the build, it doesn't grab the `proxy-rewrite.entry.js` (at least), which threw an error when we tried to run `.substr()` on `undefined`.

I admit this is just fixing symptoms. I don't know why it doesn't grab that file for the build--I also don't know what this proxy-rewrite is for, now that I think about it. :thinking: But it's new as of today, so...there's that, right?

### DIFF
--- a/script/add-asset-hashes.js
+++ b/script/add-asset-hashes.js
@@ -56,7 +56,9 @@ function addAssetHashes() {
     ].forEach(unhashedName => {
       const hashedName = manifest[unhashedName];
 
-      files[`generated/${unhashedName}`] = files[hashedName.substr(1)]; // eslint-disable-line no-param-reassign
+      if (hashedName) {
+        files[`generated/${unhashedName}`] = files[hashedName.substr(1)]; // eslint-disable-line no-param-reassign
+      }
     });
 
     done();
--- a/script/add-asset-hashes.js
+++ b/script/add-asset-hashes.js
@@ -56,6 +56,7 @@ function addAssetHashes() {
     ].forEach(unhashedName => {
       const hashedName = manifest[unhashedName];
 
+      // When an --entry is specified that isn't proxy-rewrite, these files won't be here
       if (hashedName) {
         files[`generated/${unhashedName}`] = files[hashedName.substr(1)]; // eslint-disable-line no-param-reassign
       }


jbalboni (approved):
I think this is the right fix. The proxy-rewrite app is for teamsite pages and it has some special processing that we can skip for builds where you're specifying the apps to build.

script/add-asset-hashes.js
```diff
@@ -56,7 +56,9 @@ function addAssetHashes() {
     ].forEach(unhashedName => {
       const hashedName = manifest[unhashedName];
 
-      files[`generated/${unhashedName}`] = files[hashedName.substr(1)]; // eslint-disable-line no-param-reassign
+      if (hashedName) {
+        files[`generated/${unhashedName}`] = files[hashedName.substr(1)]; // eslint-disable-line no-param-reassign
+      }
```
ncksllvn:
```
[
       'proxy-rewrite.entry.js',
       'styleConsolidated.css',
       'static-pages.css',
       'vendor.entry.js',
       'polyfills.entry.js',
    ]
```
Some context for these; apologies if you already know this - These bundles are all required for the Teamsite injection work (applying our header/footer to VA sites outside of our source code), but there was a problem encountered with the hashed bundle names (cache-busting in higher environments). So, this little piece reverses that hashing. It should probably be assigned to a variable for clarity.

Down the road I think I would wonder at what point is `hashedName` not there. Since this seems to be a situation with the `--entry` flag, I wonder if it would be more clear if we used that argument to apply a filter to that array?

ncksllvn (approved):

cvalarida:
Ahh, that makes a lot of sense, @jbalboni & @ncksllvn. Thanks for the clarification!

@ncksllvn I'm not sure I follow. Are you suggesting rather than checking if `hashedName` is truthy, check instead for the presence `--entry` flag?

script/add-asset-hashes.js
cvalarida: I'm not sure I follow. Are you suggesting rather than checking if `hashedName` is truthy, check instead for the presence `--entry` flag?

ncksllvn: I think that if there is an `--entry` argument that isn't `proxy-rewrite`, then we can skip that entire `forEach` loop. 

But also a comment there would probably suffice 😆

PR #8947, opened 2018-10-31
Fix More in this section cutoff

jbalboni:
## Description

The "More in this section" button was being cutoff on some pages when stuck to the top of the page. This is apparently because on certain pages we have a hidden wizard with radio and checkboxes. These are absolutely positioned by USWDS with a negative margin, but the margin offset is not enough when inside our gray box. This widens the page and makes it horizontally scrollable, which for some reason throws off the fixed position of the button. No idea why that is, but fixing the wizard fixes the button.

## Testing done
Tested locally

## Screenshots
screen shot 2018-10-31 at 10 35 07 am

## Acceptance criteria
- [x] Scrolling doesn't cut off more in this section button on mobile when fixed to top

## Definition of done


ncksllvn (approved):
LGTM!

PR #8959, opened 2018-10-31
10684-breadcrumb-wbc

1Copenut:
## Purpose of the PR

This PR is asking to move all React apps in production to use the new Formation `Breadcrumb` component, instead of one-off components. It seeks to wrap existing breadcrumb components in our Formation `<Breadcrumb />` component, to take advantage of the mobile "back-by-one" handling and maximize the accessible features of the breadcrumb. 

This is work that completes the move to a full breadcrumb started in June for our static content.

## Added

* `<Breadcrumb>` component from Formation 1.2.0
* `renderBreadcrumb()` method to several React containers. Always returns `crumbs` array.
* e2e test for breadcrumb functionality in Track Claims and Appeals app
* e2e test for deep Claim Estimate Date breadcrumb evaluation

## Updated

* 2 unit tests to pass in mocked `location` object for GI Bill Comparison Tool
* All instances of app containers using breadcrumb components
* All breadcrumbs maintain the existing local breadcrumb components, and use the Formation Breadcrumb inside them for formatting and styles.
* Carefully added routes and updated text for app breadcrumb "You are here" pages.
* The URL matching pattern for GI Bill Comparison Tool, per discussions with @mnorthuis
* Renamed two `Breadcrumbs.jsx` files to `ClaimsBreadcrumbs.jsx` and `GiBillBreadcrumbs.jsx` respectively in their apps, for clarity and so import statements match file name. 

## Did Not Modify

* Health Records because it appears it's being managed by another application outside our code
* RxRefills, for the same reason


src/applications/claims-status/components/Breadcrumbs.jsx
```diff
@@ -1,29 +1,38 @@
...
+class ClaimsBreadcrumbs extends React.Component {
+  renderBreadcrumbs = childNodes => {
+    const crumbs = [
+      <a href="/" key="home">
+        Home
+      </a>,
+      <a href="/disability-benefits/" key="disability-benefits">
+        Disability Benefits
+      </a>,
+      <Link to="/" key="claims-home">
```
1Copenut: Please advise if it'd be better to change this to `<Link to="your-claims" key="claims-home">`

The `routes.jsx` file seems to redirect there anyway, but I'm all about fault tolerance.

src/applications/claims-status/components/Breadcrumbs.jsx
```diff
@@ -1,29 +1,38 @@
...
+import { Link } from 'react-router';
+import Breadcrumbs from '@department-of-veterans-affairs/formation/Breadcrumbs';
+
+class ClaimsBreadcrumbs extends React.Component {
+  renderBreadcrumbs = childNodes => {
+    const crumbs = [
+      <a href="/" key="home">
+        Home
+      </a>,
+      <a href="/disability-benefits/" key="disability-benefits">
```
1Copenut: The new link here is actually `/disability/`, but I'm not sure it should be changed b/c it'll break on Vets.gov, at least for the next few days. Thoughts?

src/applications/claims-status/components/AppealLayout.jsx
```diff
@@ -1,16 +1,16 @@
 import React from 'react';
 import { Link } from 'react-router';
-import Breadcrumbs from './Breadcrumbs';
+import ClaimsBreadcrumbs from './Breadcrumbs';
```
jbalboni: I would change the name of the file, too, to avoid confusion.

src/applications/claims-status/components/Breadcrumbs.jsx
jbalboni: There's a redirect in place that will make this work correctly in both places for now if you leave it set to the vets.gov path. The other option is to use the brand consolidation flag and switch between them, but we haven't been doing that for links generally.

src/applications/claims-status/components/Breadcrumbs.jsx
jbalboni: Either way is fine, as long as it's not causing a full page reload in the browser when you click on it.

jbalboni: I'm not sure I'm understanding what's happening here. Isn't this link a duplicate of the link passed as a child to `ClaimsBreadcrumbs` in `AppealLayout`?

src/applications/claims-status/containers/ClaimEstimationPage.jsx
```diff
@@ -13,21 +12,18 @@ class ClaimEstimationPage extends React.Component {
...
-              </li>
-              <li>
-                <Link to={`your-claims/${this.props.params.id}/status`}>
-                  Your {claimType} Claim
-                </Link>
-              </li>
-            </Breadcrumbs>
+            <ClaimsBreadcrumbs>
+              <Link to={`your-claims/${this.props.params.id}/status`}>
+                Status Details
```
jbalboni: Was it intentional to change the breadcrumb text for this page? The old description seems better to me, but I don't know the rationale for the change.

src/applications/gi/components/heading/Breadcrumbs.jsx
```diff
@@ -1,31 +1,34 @@
...
       </a>,
       <a href="/education/" key="education">
-        Education Benefits
-      </a>,
-      <a href="/education/gi-bill/" key="gi-bill">
-        GI Bill
+        Education and Training
       </a>,
+      <Link to={root} key="main">
+        GI Bill Comparison Tool
```
jbalboni: Looks we lost the trademark symbol

jbalboni (commented):
Looks good so far. I left some comments and I'm going to try it out locally before approving.

src/applications/claims-status/containers/ClaimEstimationPage.jsx
1Copenut: @jbalboni Yes, this was a conscious choice, based on several conversations with @mnorthuis and @bethpotts regarding link text.

src/applications/claims-status/components/Breadcrumbs.jsx
1Copenut: @jbalboni I had to change this one to get a unit test to pass. This is a duplicate, but this particular file doesn't seem to be used. It's commented out in the `routes.jsx` logic, and I can't find any other uses of.

This was more a defensive move, to ensure things weren't breaking. I am able to see Appeals data on `localhost` reliably, and haven't noticed any unusual links popping up.

src/applications/claims-status/components/Breadcrumbs.jsx
1Copenut: @jbalboni I will follow suit and let the re-direct handle it.

src/applications/claims-status/containers/ClaimEstimationPage.jsx
1Copenut: @jbalboni Yes, this was a conscious choice, based on several conversations with @mnorthuis and @bethpotts regarding link text.

src/applications/claims-status/components/Breadcrumbs.jsx
jbalboni: Ok, sounds good.

I was trying things out in the GIBCT app and based on that, I think you want to change the link here to be `to="your-claims"` to avoid a page refresh.

src/applications/claims-status/containers/AppealInfo.jsx
```diff
@@ -103,14 +103,11 @@ export class AppealInfo extends React.Component {
...
-            <Breadcrumbs>
-              <li>
-                <Link to="your-claims">Track Your Claims and Appeals</Link>
-              </li>
-              <li>
-                <strong>{claimHeading}</strong>
-              </li>
-            </Breadcrumbs>
+            <ClaimsBreadcrumbs>
+              <Link to={`/appeals/${appeal.id}`} key="claims-appeal">
```
jbalboni: I think this should be a relative url (`appeals/${appeal.id}` instead of `/appeals/${appeal.id}`)

src/applications/gi/components/heading/Breadcrumbs.jsx
```diff
@@ -38,26 +41,19 @@ class Breadcrumbs extends React.Component {
           </a>,
         );
       }
+      crumbs.push(
+        <Link to={`/profile/${facilityCode}`} key="result-detail">
```
jbalboni: I don't think this is ever an active link, but it should be relative (i.e. not having the leading `/`)

src/applications/gi/components/heading/Breadcrumbs.jsx
```diff
@@ -1,31 +1,34 @@
...
+        Education and Training
       </a>,
+      <Link to={root} key="main">
+        GI Bill Comparison Tool
+      </Link>,
     ];
 
-    if (pathname.match(/search|profile/)) {
-      const root = { pathname: '/', query: version ? { version } : {} };
+    if (pathname.match(/search/)) {
```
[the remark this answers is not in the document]
1Copenut: Per our discussion, I reviewed search result breadcrumb and functionality with @mnorthuis for correct behavior. We do want search results included, and to remember that search when possible. If users copy and paste a profile URL, it will drop out the search results and just have 4 breadcrumb links.

src/applications/claims-status/components/AppealLayout.jsx
```diff
@@ -1,16 +1,16 @@
...
+import ClaimsBreadcrumbs from './ClaimsBreadcrumbs';
 
 export default function AppealLayout({ children }) {
   return (
     <div>
       <div className="row">
-        <Breadcrumbs>
+        <ClaimsBreadcrumbs>
           <li>
             <Link to="your-claims">Track Your Claims and Appeals</Link>
```
1Copenut: Is there a strong need to add a `key` to these `<Link />` elements? The previous code didn't have them, but I know React warns on.

jbalboni (approved):

src/applications/claims-status/components/AppealLayout.jsx
jbalboni: It's good practice generally, if the elements are in an array. It keeps React from having to recreate DOM nodes unnecessarily. It shouldn't break anything, though.

1Copenut: The code I wrote on line 19 of `ClaimsBreadcrumbs.jsx` looks for new child props, and adds them to the `crumbs` array before rendering, so there could be benefit to adding. But no, nothing breaks, having tested and verified manually and automatically.

PR #8975, opened 2018-11-01
[Mobile] [Signed in] Global account nav: Long names and emails go off screen #14392

accbjt:
## Description
[Mobile] [Signed in] Global account nav: Long names and emails go off screen #14392

- [x] Need to test on different mobile devices

How to Test Locally:
Since we don't have mock users to handle the different scenarios, I had to re-assign values to the user before it gets return in Redux.

```
// src/platform/sitewide/user-nav/selector.js

export const selectUserGreeting = createSelector(
  state => selectProfile(state).userFullName,
  state => selectProfile(state).email,
  () => conditionalStorage().getItem('userFirstName'),
  (name, email, sessionFirstName) => {
    name.first = ''; <----- re-assign these
    sessionFirstName = ''; <----- re-assign these

    if (name.first || sessionFirstName) {
      return startCase(toLower(name.first || sessionFirstName));
    }

    return 'My Account';
  },
);
```

## Testing done
Local Tests

## Screenshots
Small mobile screen
screen shot 2018-11-01 at 2 15 07 pm

Larger Mobile Screen
screen shot 2018-11-01 at 2 50 46 pm

Name with no first name
screen shot 2018-11-01 at 2 55 56 pm

Name with 11 letters
screen shot 2018-11-01 at 2 52 07 pm

## Acceptance criteria
- [ ] User with no first name should display `My Account`
- [ ] Should display first name if of User in user drop down if the do have a first name
- [ ] Should truncate to 9 letters if first name is longer then 11 letters
- [ ] On devices with 320 width or lower the name should truncate to 5 letters.

## Definition of done


src/platform/site-wide/sass/merger.scss
```diff
@@ -245,7 +242,19 @@
...
+        justify-content: space-between;
+      }
+    }
+
     .vet-toolbar {
+      margin-left: 0;
+      [data-reactroot] {
+        align-items: baseline;
+        display: inherit;
+      }
```
ncksllvn: Seems like we could use a more descriptive selector here than `[data-reactroot]`.

src/platform/site-wide/sass/merger.scss
```diff
@@ -259,6 +268,21 @@
...
+            max-width: 102px;
+            white-space: nowrap;
+            overflow: hidden;
+            text-overflow: ellipsis;
+            margin-bottom: -6px;
+
+            @media (max-width: $xsmall-screen) {
+              max-width: 60px;
+            }
+          }
```
ncksllvn: Ugh, hate seeing this merger file continue to grow. Was it not possible for us to write using mobile-first CSS for this?

src/platform/site-wide/user-nav/selectors.js
```diff
@@ -13,6 +13,6 @@ export const selectUserGreeting = createSelector(
       return startCase(toLower(name.first || sessionFirstName));
     }
 
-    return email;
+    return 'My Account';
```
ncksllvn: If we're never going to use `email`, then we can probably remove the email selector.

accbjt:
@ncksllvn I've made the changes you suggested. I didn't know what to do about the extra CSS. Can you give me some guidance on what you think I can do about that.

src/platform/site-wide/sass/merger.scss
```diff
@@ -175,6 +172,23 @@
...
+        margin-bottom: -6px;
+        max-width: 102px;
+
+        @media (max-width: $xsmall-screen) {
+          max-width: 60px;
+        }
+      }
+    }
+  }
+
```
accbjt: I made this more mobile-first, but I ran into issues on iPhone 5. It seems like iPhone 5 is still within the 320 width so the user name still goes of the page when the css is within the `@media (min-width: $xsmall-screen) and (max-width: $medium-screen - 1)` media query. It looks like this.

screen shot 2018-11-05 at 5 30 00 pm

I needed to add the `@media (max-width: $xsmall-screen)` to target this for iPhone 5 sizes and below. This is what it looks like with the `@media (max-width: $xsmall-screen)` media query.

screen shot 2018-11-05 at 6 38 14 pm

@ncksllvn @jbalboni if you can think of another way are targeting iPhone 5 please let me know.

jbalboni:
There's a gap below the button:
screen shot 2018-11-06 at 9 33 33 am

The hover state looks odd:
screen shot 2018-11-06 at 9 33 12 am

jbalboni:
Looks like we're no longer stacking the buttons on the left, we're filling the width of the screen:

screen shot 2018-11-06 at 9 36 00 am

Also looks like we're missing the ellipsis (and icon?) on larger screens:
screen shot 2018-11-06 at 9 39 05 am

src/platform/site-wide/sass/merger.scss
jbalboni: I think part of the problem is more that 102 pixels is too wide of a max width for screens just above 320px. It looks fine once you get to 375px, but in between there it cuts off like your screenshot above. I would do something like this:

```scss
  max-width: 60px
  @include media(@medium-phone-screen) {
    max-width: 102px;
  }
```

And set `@medium-phone-screen` to `375px` in this file. What you have now doesn't fit our existing breakpoints (each variable is intended to be a min width, not a max width) either, and I think maintaining the mobile first structure of the media queries will keep us from having problems later.

accbjt:
@jbalboni can you take a look at this again. Added the fixes.

jbalboni:
Looks like this is an issue at iPad width:
screen shot 2018-11-09 at 2 16 59 pm

This is very nitpicky, but the name seems to have shifted left compared to the production site. If this isn't easy to fix, I think we can leave it until we refactor the header:
screen shot 2018-11-09 at 2 16 14 pm

Looks like there's a line height issue or something here that pushes the baseline of the user name up:
screen shot 2018-11-09 at 2 14 28 pm

Is there a way to set the min width of the dropdown to the button width? Looks weird:
screen shot 2018-11-09 at 2 22 27 pm

rianfowler:
These changes should be verified on [TeamSite pages](https://github.com/department-of-veterans-affairs/vets.gov-team/blob/f9e309647d2d825c1fca229b9bf8a91c9378d12d/Work%20Practices/Engineering/Teamsite.md): 
www.vets.gov/health
www.benefits.va.gov
www.cem.va.gov 

These three pages have different sets of styling.

rianfowler:
I opened a PR for adding instructions for testing against the TeamSite subdomain pages here: https://github.com/department-of-veterans-affairs/vets.gov-team/pull/15083

We have an ongoing conversation on how we want to make this easier moving forward. As it stands, changing anything on the header is going to require testing across several sites. https://github.com/department-of-veterans-affairs/vets.gov-team/issues/14464

accbjt:
@jbalboni Can you take another look at this. I added in your suggestions and I also checked the subdomains @rianfowler mentioned above.

jbalboni:
I found one more nitpicky thing:

screen shot 2018-11-26 at 4 58 22 pm

There's no padding to the left of the name now, which looks a little odd.

I think what I'd suggest is to look into this for a short amount of time and if the solution isn't simple, then create a ticket for it and merge this.

jbalboni (approved):

accbjt:
Updated the padding issue.
screen shot 2018-11-27 at 12 17 00 am

PR #8986, opened 2018-11-02
add vagovprod environment configs

wyattwalter:
## Description
Adds new environment for va.gov production.

## Testing done
Built locally and tested. The intent of this is to kick off the process of being able to see it in a prod-like deployment.

## Definition of done

ref: https://github.com/department-of-veterans-affairs/vets.gov-team/issues/14786

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -21,7 +21,7 @@ def getEnvNames(isCMSDeploy) {
     'development', 'staging', 'production'
   ]
   def vagovEnvNames = [
-    'preview', 'vagovdev', 'vagovstaging'
+    'preview', 'vagovdev', 'vagovstaging', 'vagovprod'
   ]
 
   if (isCMSDeploy) {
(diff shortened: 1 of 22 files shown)


ncksllvn (approved):
LGTM! Will this have any effect on anything at the moment? We check for the "preview" build-type in the front matter of content files and occasionally in app code via `__BUILDTYPE__`, so we'll need to update those too. That work will probably be here, https://github.com/department-of-veterans-affairs/vets.gov-team/issues/14822.

wyattwalter:
Thanks @ncksllvn . My intent at this point is just to have a _working_ build that I can use. I've got most of the pieces in place to make this loadable from www.va.gov if a cookie is set (otherwise falls back to the old style page). My plan is to get this going and out the door today so that we can start seeing what the pages look like on the domain and start cleaning up the stuff like you mentioned.

I also noticed that it still says "you're viewing a preview" and couldn't figure out in a few minutes of searching how I could disable that on these builds.

ncksllvn:
@wyattwalter Can I use this PR to do that FE work?

PR #8996, opened 2018-11-02
Fix the Foresee 404 error on vagovdev or development builds 

accbjt:
## Description
Fix the Foresee 404 error on vagovdev or development builds 

## Testing done
Local Test

## Acceptance criteria
- [ ] The Foresee script should only render on the 'preview' or 'vagovstaging' environments.

## Definition of done


content/includes/survey-tools.html
```diff
@@ -1,6 +1,8 @@
 {% assign auth_required_entrynames = 'claims-status, letters, health-records, messaging, rx, profile-360' | split: ', ' %}
 {% if auth_required_entrynames contains entryname  %}
-  <script type="text/javascript" src="/js/foresee/{{ buildtype }}.js"></script>
+  {% if buildtype == 'vagovstaging' or buildtype == 'preview' %}
+    <script type="text/javascript" src="/js/foresee/{{ buildtype }}.js"></script>
+  {% endif %}
```
ncksllvn: The `content` directory is only for Vets.gov, so I don't think we have to touch this file.

va-gov/includes/survey-tools.html
```diff
@@ -1 +1,3 @@
-<script type="text/javascript" src="/js/foresee/{{ buildtype }}.js"></script>
+{% if buildtype == 'vagovstaging' or buildtype == 'preview' %}
+  <script type="text/javascript" src="/js/foresee/{{ buildtype }}.js"></script>
+{% endif %}
```
ncksllvn: 👍This also reminds me that I think we need Foresee script, `vagovprod.js`, to configure Foresee on `www.va.gov`.

CC: @BBarone1

BBarone1: thanks @ncksllvn is that something that we can add here or do we need a separate implementation ticket for a new PR?

accbjt: @ncksllvn Do we want to do this with this ticket or create a new ticket. If we are doing this in this ticket who is the one that is going to implement it. I am not familiar with the Foresee service so I would need some documentation if I were to implement it.

jbalboni: I think we just need to make a copy of the preview one named vagovprod.js. The current preview one has the production, va.gov changes in it.

va-gov/includes/survey-tools.html
```diff
@@ -1 +1,3 @@
-<script type="text/javascript" src="/js/foresee/{{ buildtype }}.js"></script>
+{% if buildtype == 'vagovstaging' or buildtype == 'preview' %}
```
jbalboni: I think this also needs to include `vagovprod`

jbalboni (approved):

PR #9008, opened 2018-11-03
Provide sign in link on VA subdomain pages

cehsu:
## Description
These changes address https://github.com/department-of-veterans-affairs/vets.gov-team/issues/14785

## Testing done
Unit tests created/updated

## Screenshots
image
image

## Acceptance criteria
- [x] Provide sign in link on VA subdomain pages

## Definition of done

### DIFF
--- a/src/platform/brand-consolidation/va-subdomain.js
+++ b/src/platform/brand-consolidation/va-subdomain.js
@@ -5,7 +5,7 @@
  */
 export default function isVASubdomain() {
   return !!window.location.hostname.match(
-    /www.(cem|benefits|telehealth|volunteer|ptsd|nutrition)+/,
+    /(cem|benefits|telehealth|volunteer|ptsd|nutrition).va/,
     'g',
   );
 }
--- a/src/platform/brand-consolidation/va-subdomain.js
+++ b/src/platform/brand-consolidation/va-subdomain.js
@@ -3,7 +3,7 @@
  * @returns {boolean}
  * @module platform/brand-consolidation/va-subdomain
  */
-export default function isVASubdomain() {
+export default function isVATeamSiteSubdomain() {
   return !!window.location.hostname.match(
     /(cem|benefits|telehealth|volunteer|ptsd|nutrition).va/gi,
   );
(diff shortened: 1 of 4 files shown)


rianfowler (approved):

src/platform/brand-consolidation/va-subdomain.js
```diff
@@ -0,0 +1,11 @@
+/**
+ * Returns whether the current hostname is a VA subdomain
+ * @returns {boolean}
+ * @module platform/brand-consolidation/is-subdomain
```
erikphansen: tiny typo: `is-subdomain` -> `va-subdomain`

src/platform/brand-consolidation/va-subdomain.js
```diff
@@ -0,0 +1,11 @@
...
+ */
+export default function isVASubdomain() {
+  return !!window.location.hostname.match(
+    /(cem|benefits|telehealth|volunteer|ptsd|nutrition).va/,
+    'g',
```
erikphansen: Two thoughts on this regex: first, should we make it case insensitive, just to be extra safe? Second, maybe it should work in the opposite way; that is, check to see if they are *not* on a subdomain, so that we don't have to maintain a list of valid subdomains to check against.

erikphansen (approved):

src/platform/brand-consolidation/va-subdomain.js
rianfowler: We never got a definitive list of subdomains that should have it and no one seems to know how many sites use this Team Site template.

erikphansen: Okay, so there *is* a list of subdomains that we want to match against; it's not just a matter of checking if the host is *a* VA subdomain. In that case I'd suggest renaming this to something a little clearer, like `isVATeamSiteDomain()`.

src/platform/brand-consolidation/va-subdomain.js
cehsu: Thanks, fixed.

src/platform/brand-consolidation/va-subdomain.js
cehsu: Good idea, updated.

PR #9012, opened 2018-11-05
Vets-to-VA Redirect - Clean destination before build

ncksllvn:
## Description
After a full deployment, I discovered that the catch-all redirect wasn't working and instead, the regular Vets.gov page was being served. I believe this is because we weren't clearing the directory before building the redirect-HTML. This PR fixes that by clearing out the directory.

## Testing done
Local builds

## Acceptance criteria
- [x] Global catch-all redirect to VA.gov works

## Definition of done

### DIFF
--- a/src/platform/testing/visual-regression/util/get-file-names.js
+++ b/src/platform/testing/visual-regression/util/get-file-names.js
@@ -1,5 +1,5 @@
 const path = require('path');
-const mkdirp = require('mkdirp');
+const fs = require('fs-extra');
 const { baseUrl } = require('../../e2e-puppeteer/helpers');
 const commandLineArgs = require('command-line-args');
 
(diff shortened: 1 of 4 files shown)


src/platform/testing/visual-regression/util/get-file-names.js
```diff
@@ -38,7 +38,7 @@ function getFileNames(route) {
 async function createDirectoryIfNotExist(filePath) {
   const directory = path.dirname(filePath);
   return new Promise((resolve, reject) => {
-    mkdirp(directory, err => (err ? reject(err) : resolve()));
+    fs.ensureDir(directory, err => (err ? reject(err) : resolve()));
```
ncksllvn: `fs-extra` offers what we were using `mkdirp` for, so I swapped out the module here and uninstalled it.

CC: @RobAWilkinson

jbalboni (approved):
Looks good to me.

PR #9013, opened 2018-11-05
Fix for local development after vets.gov-to-va.gov redirect implementation

ncksllvn:
## Description
Same logic as here, https://github.com/department-of-veterans-affairs/vets-website/pull/9014.

The watch task applies the behavior of `DEVELOPMENT`, but that is now a deprecated Vets.gov environment, causing some conflict during local dev.

## Testing done
`npm run watch`

## Acceptance criteria
- [ ] Watch works again

## Definition of done

### DIFF
--- a/script/options.js
+++ b/script/options.js
@@ -39,6 +39,16 @@ function gatherFromCommandLine() {
   return options;
 }
 
+function applyDeprecatedBuildtypes(options) {
+  const deprecatedEnvironments = [environments.DEVELOPMENT];
+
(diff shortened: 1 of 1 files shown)


script/options.js
```diff
@@ -132,6 +140,7 @@ function applyBrandConsolidationOverrides(options) {
 function getOptions() {
   const options = gatherFromCommandLine();
 
+  applyDeprecatedBuildtypes(options);
```
ncksllvn: I think just adding another step here is the clearest way to apply the redirect behavior.

jbalboni (approved):

PR #9019, opened 2018-11-05
Add GA layer event for best bets

kfrz:
## Description
This PR adds a `recordEvent()` call to best bet results on search app.

## Testing done
Verified that a best-bet result has the onClick event fired. 

## Acceptance criteria
- [x] Best bet results have proper GA tag with correct title

## Definition of done

### DIFF
--- a/src/applications/search/containers/SearchApp.jsx
+++ b/src/applications/search/containers/SearchApp.jsx
@@ -145,7 +145,7 @@ class SearchApp extends React.Component {
           <h4>Our Top Recommendations for You</h4>
           <ul className="results-list">
             {recommendedResults.map(r =>
-              this.renderWebResult(r, 'description'),
+              this.renderWebResult(r, 'description', true),
             )}
           </ul>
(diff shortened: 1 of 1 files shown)


narin (commented):
@kfrz Just a suggestion: if the only difference between the rendered link HTML is the `onClick`, to make it a little cleaner, maybe you could put the condition within the `onClick` so you can avoid duplicate html.

Something like this:

```
<a
  className="result-title"
  href={result.url}
  onClick={isBestBet ? 
    () => recordEvent({
      event: 'nav-searchresults',
      'nav-path': `Recommended Results -> ${result.title}`,
    })
    : null
  }
>
```

kfrz:
@narin awesome - I had actually attempted that and had some syntax errors but your version is perfect. 

Updated to use the simplified version, will leave the extraction to a component to future body of work ( if we decide to go that route ).

narin (approved):
LGTM

PR #9024, opened 2018-11-05
Show Welcome modal: only on homepage, only once

erikphansen:
> I'm still mentally walking through this to make sure this will work as expected in all cases.

## Description
Add a Welcome To The New VA modal that only appears on the homepage and only appears once.

## Testing done
Local testing

### Updated screenshot
image

### Old screenshot
image

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/src/platform/site-wide/announcements/components/WelcomeToNewVAModal.jsx
+++ b/src/platform/site-wide/announcements/components/WelcomeToNewVAModal.jsx
@@ -1,0 +1,46 @@
+import React from 'react';
+import Modal from '@department-of-veterans-affairs/formation/Modal';
+import brandConsolidation from '../../../brand-consolidation';
+
+export default class WelcomeToNewVAModal extends React.Component {
+  static isEnabled() {
+    if (!brandConsolidation.isEnabled()) return false;
+
+    if (__BUILDTYPE__ === 'preview') return false;
+
+    return true;
+  }
+
+  render() {
+    const { dismiss } = this.props;
+
+    return (
+      <Modal
+        cssClass="va-modal announcement-brand-consolidation"
+        visible
+        onClose={dismiss}
+        id="modal-announcement"
+      >
+        <div className="announcement-heading-brand-consolidation">
+          <img
+            className="announcement-brand-consolidation-logo"
+            src="/img/design/logo/va-plus-vets.png"
+            alt="VA.gov plus Vets.gov"
+          />
+        </div>
+        <h3 className="announcement-title">
+          Welcome to the New VA.gov — Built with Veterans, for Veterans
+        </h3>
+        <p>
+          Our new site offers one place to access all VA benefits and health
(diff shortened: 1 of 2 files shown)


DanielleSoCompany:
Hi @erikphansen. I think this looks great. My only question is whether this one is supposed to have the "+ Vets.gov" in the header, if this is being shown to every viewer and not just those being redirected from Vets.gov. I may be totally off here, so if this is correct, please feel free to ignore this comment. Just wanted to raised as a question in case. Thanks!

erikphansen:
Thanks @DanielleSoCompany. Good point. Do we have an existing VA.gov asset that's suitable for this modal?

DanielleSoCompany:
I'm not sure. @amycesal @emilywaggoner, would you be able to help answer Erik's question above?

src/platform/site-wide/announcements/components/WelcomeToNewVAModal.jsx
```diff
@@ -0,0 +1,46 @@
+import React from 'react';
+import Modal from '@department-of-veterans-affairs/formation/Modal';
+import brandConsolidation from '../../../brand-consolidation';
+
+export default class WelcomeToNewVAModal extends React.Component {
+  static isEnabled() {
+    if (!brandConsolidation.isEnabled()) return false;
+
+    if (__BUILDTYPE__ === 'preview') return false;
```
jbalboni: Why don't we want this to run in preview?

erikphansen: Good call. Will fix.

amycesal:
So this VA.gov + Vets.gov lock-up was intended for people who are coming from vets.gov. 
See this issue: https://github.com/department-of-veterans-affairs/vets.gov-team/issues/12060

Will the modal appear for everyone coming to the new va site, not just from vets? It might make sense to use the VA logo + seal. Or maybe we don't an image/logo.

erikphansen:
@amycesal this modal will be show the only once to visitors who navigate directly to the va.gov homepage. You will not see it if you are redirected from vets.gov. If we already have an excisting image asset to use I can probably just drop it in place of what is shown in the screenshot. Otherwise I'll need a new asset to use. Someone in design or product needs to make that call. Thank you!

erikphansen:
I am also in favor of calling all modals "lock ups" from here on out.

ncksllvn (approved):
This seems like a case we could just use the VA logo. But either way, LGTM! After merging, you can test this by going to https://dev.vets.gov and making sure that the Vets-to-VA modal shows after the redirect, and then again by navigating directly.

amycesal:
Yes, makes sense to use the VA logo here. Let me know if you need the image file, I can find one for you. 

The lockup I was referring to is the combination of va + vets.gov. 2 parts of a logo, like an icon and the wordmark, or the main company name and the subtitle are often called lockups. <- design speak

PR #9032, opened 2018-11-05
Update mistreatment content and date schema

cvalarida:
## Description
Updates the content and input type per the connected ticket.

## Testing done
Manually tested that it shows up as expected.
Updated unit test to not expect the month and day dropdowns.

## Screenshots
image

## Acceptance criteria
- [x] The content is updated
- [x] The date input is a text input, not a date selector

## Definition of done

### DIFF
--- a/src/applications/disability-benefits/all-claims/config/schema.js
+++ b/src/applications/disability-benefits/all-claims/config/schema.js
@@ -884,7 +884,6 @@ const schema = {
           },
           VAMistreatmentDate: {
             type: 'string',
-            format: 'date',
           },
         },
       },
--- a/src/applications/disability-benefits/all-claims/pages/newDisabilityFollowUp.js
+++ b/src/applications/disability-benefits/all-claims/pages/newDisabilityFollowUp.js
@@ -1,7 +1,5 @@
 import React from 'react';
 import { createSelector } from 'reselect';
-import dateUI from 'us-forms-system/lib/js/definitions/date';
-import merge from 'lodash/merge';
 
 import { getDisabilityName } from '../utils';
 import disabilityLabels from '../content/disabilityLabels';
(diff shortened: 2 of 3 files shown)


markgreenburg:
Am concerned that this field might line up with `approximateBeginDate` in the swagger, which expects a day, month, and year. If that's the case, we either need evss to change the field or we won't be able to implement this update.

cvalarida:
@djmunoz1 [said](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/14660#issuecomment-435370105) in the connected ticket
> and actually what will need to happen here is for all the of the inputs on this question to be translated into a single string in the Disability object for:
serviceRelevance (string, optional)stringMin. Length:1Max. Length:500,

:thinking: It does sound more like a `approximateBeginDate` instead of a `serviceRelevance`, doesn't it? :grimacing: 

@djmunoz1 How can we figure out what this field lines up with in terms of what EVSS expects?

djmunoz1:
@cvalarida yes it is like an Approximate Date but Alex wanted it designed as a string as opposed to a date input to prevent veterans from having to recall specifics. So these inputs collect when and where the VA interaction occurred (e.g. "Spring '02" or "right after discharge") not when the condition started, which could conceivably be different. So the entire story of what happened at the VA would get rolled into serviceRelevance, which EVSS displays on the overflow page to satisfy the "Explain" column on the right below.
screen shot 2018-11-06 at 12 08 26 pm

Does that make sense? If we want to rework that I'd entertain it as a corrective enhancement but I'd like to gather a little data first on what people are putting into that field. EVSS is aware of how this will get interpreted. They recommended a format for the string to make clear what the inputs are so {Event: `input`; Location:`input`, Timeframe:`input`}

cvalarida:
I think that makes sense. :+1:

peggygannon (approved):
@cvalarida Looks good to me, from a content perspective.

markgreenburg:
@djmunoz1 to double check, you're saying that the intent is for us to send EVSS one string field, with an organized content structure, such as

djmunoz1:
you got it @markgreenburg!

markgreenburg (approved):
🚀

PR #9038, opened 2018-11-06
fix hideFromSidebar

RobAWilkinson:
## Description
Siblings of an open link with `hideFromSidebar: true` weren't being hidden. "Cousin" links were, however. This fixes that.

## Testing done
local testing, VRT


accbjt (approved):
LGTM

PR #9069, opened 2018-11-07
Manage tab focus properly on claims and appeals tracker

kfrz:
## Description
The Track Claims and Appeals application is doing client-side routing, and is not actively managing focus when new views mount. This leaves keyboard and screen reader users to orient themselves, without a clear understanding of which element has focus on first "load".  See [attached issue](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/14488)

**This PR fixes a bug that would reset tab focus to the main `<body>` tag after claims component updated**

## Testing done
- Tested with keyboard that tab focus order works as expected, focusing on the breadcrumb menu. 
- Tested with ChromeVox that content is read to me when loaded. 

## Screenshots
Here's a cast of me navigating via the keyboard

screencast-localhost-3001-2018 11 07-13-45-23

## Acceptance criteria
- [x] As a keyboard user, I want to have a clear understanding of where focus is when the Claims and Appeals pages first load. I want to press TAB once and start tabbing through the breadcrumb menu.
- [x] I want the focus to be set consistently when the application first loads, and when I move through pages, whether that's by clicking links, pressing the Back button, or clicking on the breadcrumb menu links.
- [x] As a screenreader user, I want the same things as the keyboard user, and I want the content to be read to me when it is loaded on the page. This is my primary cue for new content on screen.

## Definition of done

### DIFF
--- a/src/applications/claims-status/containers/YourClaimsPageV2.jsx
+++ b/src/applications/claims-status/containers/YourClaimsPageV2.jsx
@@ -68,6 +68,7 @@ class YourClaimsPageV2 extends React.Component {
     if (!this.props.loading && prevProps.loading) {
       setPageFocus();
     }
+    setPageFocus();
   }
 
   changePage(page) {
@@ -184,7 +185,6 @@ class YourClaimsPageV2 extends React.Component {
       } else if (!this.props.canAccessClaims && bothRequestsLoaded) {
         content = <NoClaims />;
       }
-
       content = <div className="va-tab-content">{content}</div>;
     }
 
--- a/src/applications/claims-status/containers/YourClaimsPageV2.jsx
+++ b/src/applications/claims-status/containers/YourClaimsPageV2.jsx
@@ -64,10 +64,7 @@ class YourClaimsPageV2 extends React.Component {
   // an initial sort needs to happen in componentDidMount
   // }
 
-  componentDidUpdate(prevProps) {
-    if (!this.props.loading && prevProps.loading) {
-      setPageFocus();
-    }
+  componentDidUpdate() {
     setPageFocus();
   }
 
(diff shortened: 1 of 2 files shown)


src/applications/claims-status/containers/YourClaimsPageV2.jsx
```diff
@@ -68,6 +68,7 @@ class YourClaimsPageV2 extends React.Component {
     if (!this.props.loading && prevProps.loading) {
       setPageFocus();
     }
+    setPageFocus();
```
ncksllvn: Can this be reduced to just `setPageFocus();`, without the condition on line 68?

ncksllvn (approved):

src/applications/claims-status/containers/YourClaimsPageV2.jsx
1Copenut: Would this conditional logic prevent a case where the focus was being reset outside of a loading situation? I'm thinking specifically about the `Find out why we sometimes combine claims` link. It opens a modal window, and should re-take focus when the modal window is closed. Maybe the two are unrelated?

1Copenut (approved):
LGTM. I was able to test this locally in MacOS Mojave + Safari + VoiceOver. The breadcrumbs read out perfectly and gave me a good "grounding" of where I was in the app. I'll test further with NVDA and JAWS when we're in staging. Can't test them locally because of the way we handle logins on localhost.

src/applications/claims-status/containers/YourClaimsPageV2.jsx
1Copenut: Okay, so that's a separate issue. I will see if it has been logged previously and if not, will add a new issue.

bshyong (approved):
LGTM

PR #9073, opened 2018-11-07
526 submission bug fix cv

cvalarida:
## Description
Uses a custom form submission function to send the cookies along with the 526 submission request.

## Testing done
Manually tested.

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/src/applications/disability-benefits/526EZ/config/submitForm.js
+++ b/src/applications/disability-benefits/526EZ/config/submitForm.js
@@ -15,11 +15,9 @@ export default function submitForm(form, formConfig) {
     req.open('POST', formConfig.submitUrl);
 
     req.withCredentials = true;
-    if (conditionalStorage().getItem('userToken')) {
-      req.setRequestHeader(
-        'Authorization',
-        `Token token=${conditionalStorage().getItem('userToken')}`,
-      );
+    const userToken = conditionalStorage().getItem('userToken');
+    if (userToken) {
+      req.setRequestHeader('Authorization', `Token token=${userToken}`);
     }
 
     req.addEventListener('load', () => {
--- a/src/applications/disability-benefits/526EZ/config/submitForm.js
+++ b/src/applications/disability-benefits/526EZ/config/submitForm.js
@@ -1,0 +1,67 @@
+import { transformForSubmit } from 'us-forms-system/lib/js/helpers';
+
+export default function submitForm(form, formConfig) {
+  console.log('custom submitForm called');
+  const body = formConfig.transformForSubmit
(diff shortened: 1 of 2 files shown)


markgreenburg (approved):
LGTM. Tested locally as well.

src/applications/disability-benefits/526EZ/config/submitForm.js
```diff
@@ -0,0 +1,73 @@
...
+    const userToken = conditionalStorage().getItem('userToken');
+    if (userToken) {
+      req.setRequestHeader('Authorization', `Token token=${userToken}`);
+    }
+
+    req.addEventListener('load', () => {
+      if (req.status >= 200 && req.status < 300) {
+        // Using GA instead
+        // recordEvent({ event: 'form-submit-successful' });
+        window.dataLayer.push({ event: 'form-submit-successful' });
```
jbalboni: Shouldn't we be using recordEvent?

jbalboni (approved):
Looks okay to me, if it doesn't work for you locally still, are we still confident this will fix the issue?

markgreenburg:
I think Chris was having issues getting this to work locally for unrelated docker / sidekiq reasons. I was able to reproduce the 401 locally and then confirm the problem goes away using this branch. FWIW, this branch wasn't working until Chris's latest commit, which added the `'X-Key-Inflection': 'camel'` header

src/applications/disability-benefits/526EZ/config/submitForm.js
```diff
@@ -0,0 +1,73 @@
...
+      error.statusText = req.statusText;
+      reject(error);
+    });
+
+    req.setRequestHeader('X-Key-Inflection', 'camel');
+    req.setRequestHeader('Content-Type', 'application/json');
+
+    req.send(body);
+  });
+}
```
ncksllvn: Just curious - did this have to be done via `xhr`, instead of sort of replicating it with `fetch`? Or is this a temporary solution until we can address it at USFS?

ncksllvn (approved):

src/applications/disability-benefits/526EZ/config/submitForm.js
cvalarida: Yeah, I think the idea is to address this in USFS. I could have used `apiRequest`, but I wasn't confident that the difference in behavior wouldn't come around to bite us later.

'course, now there's a possible difference in sending the auth headers and cookies, but this seemed like the better solution. I _did_ try it with `apiRequest`, but when the call failed due to a 500 response, there was no error message shown to the user, so I went back to this. :man_shrugging:

PR #9078, opened 2018-11-08
BAH Private Medical Records Form Page for 4142

robfelder:
## Description
This PR covers the migration of the 4142 form page from the 526EZ flow to the All Claims flow. When a user selects the option for the VA to get their PMR's, and they accept the Patient Acknowledgment, they will be taken to this form page.

## Testing done
Manual & Unit

## Screenshots
screen shot 2018-11-08 at 10 11 51 am

## Acceptance criteria
- [x] Upon choosing the option to have the VA get the Veterans private medical records, the user will be presented with a form page.
- [x] The user is not allowed to proceed if the the required fields are not completed
- [x] The user is allowed to enter multiple facilities

## Definition of done


src/applications/disability-benefits/all-claims/components/PrivateProviderTreatmentView.jsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import { formatReviewDate } from 'us-forms-system/lib/js/helpers';
+
+export default function PrivateProviderTreatmentView({ formData }) {
+  let from = '';
+  let to = '';
+
+  if (formData.treatmentDateRange) {
```
rianfowler: This does a check for `treatmentDateRange` here (suggesting it could be missing) but the component will render ` - ` if this doesn't exist

rianfowler (approved):

src/applications/disability-benefits/all-claims/pages/privateMedicalRecordsRelease.js
```diff
@@ -0,0 +1,151 @@
+import _ from '../../../../platform/utilities/data';
```
erikphansen: Is this a pattern we've used before? This is the first time I've seen it. Can't decide if it's cool or misleading.

src/applications/disability-benefits/all-claims/pages/privateMedicalRecordsRelease.js
```diff
@@ -0,0 +1,151 @@
...
+
+import { validateZIP } from '../validations';
+import PrivateProviderTreatmentView from '../components/PrivateProviderTreatmentView';
+import { validateDate } from 'us-forms-system/lib/js/validation';
+
+const { address } = fullSchema526EZ.definitions;
+
+export const uiSchema = {
+  'ui:description': recordReleaseDescription,
+  'ui:title': disabilityNameTitle,
```
markgreenburg: `getDisabilityName` requires a `name` but in this case gets `formData.name`, which I don't think exists.
On that note, though - the PMR question pages don't seem to get iterated per disability, so it seems as though this title may need to be something more generic than a specific disability's name?

src/applications/disability-benefits/all-claims/pages/privateMedicalRecordsRelease.js
```diff
@@ -0,0 +1,151 @@
...
+      providerFacilityAddress: merge(addressUI('', false), {
+        street2: {
+          'ui:title': 'Street 2',
+        },
+        postalCode: {
+          'ui:title': 'Postal Code',
+          'ui:options': {
+            widgetClassNames: 'usa-input-medium',
+          },
+          'ui:validations': [
```
markgreenburg: Using a pattern on the schema would remove the need to run a custom validator. Here's one for the 526 MVP's primary address zip:
`pattern: '^\\d{5}(?:([-\\s]?)\\d{4})?$'`

src/applications/disability-benefits/all-claims/pages/privateMedicalRecordsRelease.js
```diff
@@ -0,0 +1,151 @@
...
+      itemName: 'Provider',
+      viewField: PrivateProviderTreatmentView,
+      hideTitle: true,
+    },
+    items: {
+      providerFacilityName: {
+        'ui:title': 'Name of private provider or hospital',
+      },
+      'ui:validations': [validateDate],
+      treatmentDateRange: dateRangeUI(
```
markgreenburg: Note that treatment start dates must be after the earliest service period start date.

src/applications/disability-benefits/all-claims/pages/privateMedicalRecordsRelease.js
markgreenburg: I've definitely seen this used in other places. Not sure how common.

src/applications/disability-benefits/all-claims/pages/privateMedicalRecordsRelease.js
robfelder: @markgreenburg - Actually, after taking a look at it, we use the address definition in the schema, which uses the baseAddressDef, which has that exact pattern on it. This means that this validation portion shouldn't be necessary lol.

src/applications/disability-benefits/all-claims/pages/privateMedicalRecordsRelease.js
robfelder: Hey, @markgreenburg , is that something we need to add to the validation? If so, does that cause the user to have to go back and remember what that earliest service period start date?

src/applications/disability-benefits/all-claims/pages/privateMedicalRecordsRelease.js
robfelder: @markgreenburg I'm going to check in with UX about this one. See what they want to do about it. Seems like the trend has been to just remove the title?

src/applications/disability-benefits/all-claims/pages/privateMedicalRecordsRelease.js
markgreenburg: This is a problem for us on 526 MVP (but for VA treatment facilities) - I know BGS rejects the facility if the treatment start date precedes the service start date. I'm assuming that the same business logic should probably apply here. So, I do think this should probably be added to the validations.

I think you could do something helpful with the error message, though, depending on what formData you have access to in the validator. For example, if you have access to the full `formData`, you could find the earliest service start date, format it for display, and insert it into the error message. I think, anyway. Haven't tried this out 😛

src/applications/disability-benefits/all-claims/pages/privateMedicalRecordsRelease.js
markgreenburg: 🎉

src/applications/disability-benefits/all-claims/pages/privateMedicalRecordsRelease.js
```diff
@@ -0,0 +1,151 @@
...
+      }),
+    },
+  },
+};
+
+export const schema = {
+  type: 'object',
+  required: ['providerFacility'],
+  properties: {
+    providerFacility: {
```
markgreenburg: Feels like we should put this in the `config/schema.js` file and then import from there. Not sure what guidance Chris gave you, though.

markgreenburg (approved):
Looking good! Left a few comments / ideas but nothing blocking.

src/applications/disability-benefits/all-claims/pages/privateMedicalRecordsRelease.js
robfelder: I like the idea of having more in that schema for sure. I'm going to try to enhance it a bit for the next PR. Test a few things out, and see if I can clean it up further. Thanks Mark!

PR #9082, opened 2018-11-08
All claims remove unit details mg

markgreenburg:
## Description
Removes unit address & unit phone from reserves All Claims form page.

## Testing done
- Tested locally
- Updated unit tests

## Screenshots
screen shot 2018-11-08 at 10 17 18 am

## Acceptance criteria
- [x] Phone and address removed from reserves page
- [x] obligation dates and unit name required

## Definition of done

### DIFF
--- a/src/applications/disability-benefits/all-claims/pages/reservesNationalGuardService.jsx
+++ b/src/applications/disability-benefits/all-claims/pages/reservesNationalGuardService.jsx
@@ -3,9 +3,9 @@ import dateRangeUI from 'us-forms-system/lib/js/definitions/dateRange';
 import { ReservesGuardDescription } from '../utils';
 
 const {
-  unitName,
-  obligationTermOfServiceDateRange,
-} = fullSchema.properties.serviceInformation.properties.reservesNationalGuardService.properties;
+  properties: { unitName, obligationTermOfServiceDateRange },
+  required,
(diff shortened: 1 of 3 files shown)


src/applications/disability-benefits/all-claims/pages/reservesNationalGuardService.jsx
```diff
@@ -181,11 +32,10 @@ export const schema = {
       properties: {
         reservesNationalGuardService: {
           type: 'object',
+          required: ['obligationTermOfServiceDateRange', 'unitName'],
```
cvalarida: :thinking: Should this be in the schema defined for the front end only, or in the schema shared with the back end?

cvalarida (approved):
Make like a Cyberman and DELETE!

Only question was about where the required fields should be defined.

src/applications/disability-benefits/all-claims/pages/reservesNationalGuardService.jsx
markgreenburg: 🤔 you're right, probably the schema

markgreenburg: Actually, I'm not really sure. We're just cherry picking properties from the schema, so the `required` doesn't get used. Maybe they should be in both places?

PR #9084, opened 2018-11-08
Refactor claim timeline phase header to be a11y friendly

kfrz:
## Description
The Status tab on Your Compensation Claims view has several expand/collapse accordions that are not keyboard focusable. The trigger elements are `<h5>`, instead of `<button>` and cannot receive keyboard focus as a result. See [issue](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/14495) for screenshots/more details.

## Testing done
Manually verified navigation through the page works. 

Verified screenreader (ChromeVox on Chrome/Debian Linux) reads as expected. 

## Screenshots
This is a cast of me navigating through the status timeline page with keyboard, and using both `space` and `enter` keys to toggle the phase sections.

screencast-localhost-3001-2018 11 08-12-01-02

## Acceptance criteria
- [x] As a keyboard user, I want the expand/collapse + sign to be able to receive keyboard focus. This means it should be a <button> element.
- [x] I also want the trigger element to show the yellow :focus halo when it is the active element. A quick console document.activeElement should suffice to test for validity.
- [x] The trigger should toggle open and closed when I press SPACE or ENTER on the keyboard

## Definition of done

### DIFF
--- a/src/applications/claims-status/sass/claims-status.scss
+++ b/src/applications/claims-status/sass/claims-status.scss
@@ -569,7 +569,6 @@ h1:focus {
     margin: inherit;
     padding: inherit;
     z-index: inherit;
-
   }
 
   .section-header-button:focus {
@@ -603,7 +602,6 @@ h1:focus {
   > .section-complete:not(.last) {
     cursor: pointer;
   }
-
 }
 
 .usa-alert {
--- a/src/applications/claims-status/sass/claims-status.scss
+++ b/src/applications/claims-status/sass/claims-status.scss
@@ -556,7 +556,7 @@ h1:focus {
 .claim-timeline {
   padding-top: 0;
 
-  h5 button {
+  .section-header-button {
     background: inherit;
     border: inherit;
     color: inherit;
--- a/src/applications/claims-status/sass/claims-status.scss
+++ b/src/applications/claims-status/sass/claims-status.scss
@@ -569,9 +569,10 @@ h1:focus {
     margin: inherit;
     padding: inherit;
     z-index: inherit;
+
   }
 
-  h5 button:focus {
+  .section-header-button:focus {
     outline: 2px solid;
     outline-color: $color-gold-light;
     outline-offset: 0;
--- a/src/applications/claims-status/sass/claims-status.scss
+++ b/src/applications/claims-status/sass/claims-status.scss
@@ -557,7 +557,18 @@ h1:focus {
   padding-top: 0;
 
   h5 button {
-    all: inherit;
+    background: inherit;
+    border: inherit;
+    color: inherit;
+    font-family: inherit;
+    font-size: inherit;
+    font-size-adjust: inherit;
+    font-style: inherit;
+    font-weight: inherit;
+    line-height: inherit;
+    margin: inherit;
+    padding: inherit;
+    z-index: inherit;
   }
 
   h5 button:focus {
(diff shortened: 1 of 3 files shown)


1Copenut (approved):
LGTM. Pulled feature branch and tested with MacOS Mojave + Safari + VoiceOver. Worked as expected.

erikphansen (approved):

src/applications/claims-status/sass/claims-status.scss
```diff
@@ -556,6 +556,16 @@ h1:focus {
 .claim-timeline {
   padding-top: 0;
 
+  h5 button {
+    all: inherit;
```
jbalboni: This isn't supported in IE or Edge, looks like: https://caniuse.com/#feat=css-all

kfrz: I don't see any `reset` helpers/mixins in formation -- @jbalboni are you aware of any such helper class we can leverage here? Also a bit of Googling shows no clear results to polyfill this particular feature. . Else I'll rewire the css to be more explicit. Good catch!

jbalboni: I made a mixin for the va-button-link class that might work? As a starting point, at least.

src/applications/claims-status/sass/claims-status.scss
```diff
@@ -556,6 +556,27 @@ h1:focus {
 .claim-timeline {
   padding-top: 0;
 
+  h5 button {
```
kfrz: @jbalboni @1Copenut This should cover the IE/Edge inconsistency, I am, however, having issues testing authenticated stuff via my Windows VM. Do either of you have a Windows machine handy to test the review instance? 

Also, should this live as a mixin - are we going to reuse on the Appeals timeline (or elsewhere)? 

If so, that should be changed/added to the `design-systems` repo correct?

1Copenut: @kfrz I can test this, but it won't be until end of day Central Time. Will let you know then any issues or if things line out as expected on IE11 and Edge.

jbalboni: I can check it in Browserstack, but it looks like the build failed, so I don't think the review instances is up to date yet.

I'm not sure this should be a mixin. I think we can wait on that until we come across more instances of it.

Also, I think you could change the selector here. Putting a class on the button and using that would be best (e.g. `.section-header-button` or something like that), but you could do `.section-header > button` if the single class one doesn't override everything you need.

1Copenut:
@kfrz I tested with Safari + VoiceOver. Won't be able to test others until Staging, but I'm confident it'll be good. All clear on the MacOS 508 front.

src/applications/claims-status/sass/claims-status.scss
```diff
@@ -556,6 +556,27 @@ h1:focus {
...
+    font-size-adjust: inherit;
+    font-style: inherit;
+    font-weight: inherit;
+    line-height: inherit;
+    margin: inherit;
+    padding: inherit;
+    z-index: inherit;
+  }
+
+  h5 button:focus {
```
jbalboni: Should update this selector with `section-header-button`, too.

jbalboni (approved):
I left one more comment, but otherwise this looks good. We can fully test it in IE on staging, like Trevor mentioned.

PR #9086, opened 2018-11-08
Let redirects count as links for blc

jbalboni:
## Description
We currently do our broken link checking before we create redirects, but this means that we might flag links that are missing in our content directory but are being redirected through aliases. Updating this will let us remove some pages that shouldn't exist in vagov-content in favor of aliases and let me continue making updates with the goal of getting the accesssibility checker back up.

## Testing done
Checked some redirects locally

## Acceptance criteria
- [x] Redirects still work and there are no broken links flagged

## Definition of done

### DIFF
--- a/script/create-redirects.js
+++ b/script/create-redirects.js
@@ -52,13 +52,16 @@ function createRedirects(options) {
       };
 
       for (const alias of aliases) {
-        let absolutePath = path.join(options.destination, alias);
-        if (!path.extname(absolutePath))
-          absolutePath = path.join(absolutePath, 'index.html');
+        let finalPath;
+        if (alias.startsWith('/')) {
+          finalPath = alias.slice(1);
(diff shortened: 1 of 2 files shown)


script/create-redirects.js
```diff
@@ -52,13 +52,16 @@ function createRedirects(options) {
       };
 
       for (const alias of aliases) {
-        let absolutePath = path.join(options.destination, alias);
-        if (!path.extname(absolutePath))
-          absolutePath = path.join(absolutePath, 'index.html');
+        let finalPath;
```
jbalboni: Maybe this will break on Jenkins, but the blc expects Metalsmith pages to have a relative path, not an absolute path. Locally, this seems to work for me.

ncksllvn (approved):
LGTM!

PR #9089, opened 2018-11-08
Detect invalid collections references from YAML front matter

accbjt:
## Description
Checking `children` and `collection` references against collection names from the `script/collections` directory provides protection against invalid naming. This can be addressed via custom metalsmith middleware.

Example: https://github.com/department-of-veterans-affairs/vets.gov-team/issues/13011

## Testing done
Local Testing

## Screenshots
None

## Acceptance criteria
- [ ] Should raise an error if a collection or child property in the front-matter does not exist in the collections.

## Definition of done


script/check-collections.js
```diff
@@ -0,0 +1,53 @@
...
+          }" ] does not exist in the collection: ${key}`,
+        );
+      }
+    }
+
+    return accum;
+  }, []);
+
+  if (brokenCollections.length > 0) {
+    console.warn(
```
accbjt: Not sure if this was the best way to log out the errors. If you guys know a better way definitely share.

jbalboni: Seems reasonable to me

script/build.js
```diff
@@ -148,6 +149,8 @@ function defaultBuild(BUILD_OPTIONS) {
   smith.use(createBuildSettings(BUILD_OPTIONS));
 
   smith.use(updateExternalLinks(BUILD_OPTIONS));
+  smith.use(checkBrokenLinks(BUILD_OPTIONS));
```
jbalboni: I think you can remove this line (it's on 159 now).

script/check-collections.js
```diff
@@ -0,0 +1,53 @@
+/* eslint-disable no-console */
+
+const defaultCollections = require('../script/collections/default.json');
+const brandConsolidationCollections = require('../script/collections/brand-consolidation.json');
+
+const checkCollections = buildOptions => (files, metalsmith, done) => {
```
jbalboni: Could you add some comments with a basic explanation of what things this code is checking? Collections are pretty confusing to me, I think this would help anyone going forward to understand this check better.

script/check-collections.js
```diff
@@ -0,0 +1,53 @@
...
+  const collection = buildOptions['brand-consolidation-enabled']
+    ? Object.keys(brandConsolidationCollections)
+    : Object.keys(defaultCollections);
+
+  const brokenCollections = Object.keys(files).reduce((accum, key) => {
```
ncksllvn: What made you decide to use `.reduce`? This seems like an instance where it may be simpler to `.map` the files array into an array of errors messages, then `.filter` out any `null` values.

script/check-collections.js
```diff
@@ -0,0 +1,53 @@
...
+    const file = files[key];
+
+    if (key.includes('.html')) {
```
ncksllvn: You can probably could just `.filter` down to `.html` files right out of the gate. Also, the Path module is considered a more meaningful way of checking the extension.

script/check-collections.js
```diff
@@ -0,0 +1,53 @@
...
+      if (file.collection.length > 0) {
+        file.collection.forEach(collectionName => {
+          if (!collection.includes(collectionName)) {
+            accum.push(
+              `Error: Your [ collection "${collectionName}" ] does not exist in the collection: ${key}`,
+            );
+          }
+        });
+      }
```
ncksllvn: Is `file.collection` an array? I always thought it was just a single value.

script/check-collections.js
```diff
@@ -0,0 +1,53 @@
+/* eslint-disable no-console */
+
+const defaultCollections = require('../script/collections/default.json');
+const brandConsolidationCollections = require('../script/collections/brand-consolidation.json');
+
+const checkCollections = buildOptions => (files, metalsmith, done) => {
+  const collection = buildOptions['brand-consolidation-enabled']
+    ? Object.keys(brandConsolidationCollections)
+    : Object.keys(defaultCollections);
```
ncksllvn: Is there a clearer name we could use for `collection`? It seems like it could be clearer to initialize it above the `(files, metalsmith, done) => ` function as `VALID_COLLECTIONS`, unless I'm reading it wrong.

ncksllvn (commented):
I see why this is being written, but it would have been a lot more useful when we were configuring files in bulk. It seems unlikely it'll catch anything now that things are slow and we're adding files one at a time.

I would ask that we consider whether we can envision a more straightforward way of configuring these navigations. Right now, this logic is very scattered - throughout the front matter in different property names; in the collections JSON file; in the `left-rail-nav-reset-levels.js` middleware; probably somewhere else, too. Ideally, we would have this configured in a single location where a content editor can easily understand and make changes.

script/check-collections.js
accbjt: @ncksllvn That is what I am doing with the reduce. I am creating a map and filter in the reduce. I thought it would be simpler to do it in 1 operation instead of 2.

script/check-collections.js
accbjt: Thanks, that actually would read better.

script/check-collections.js
accbjt: @ncksllvn Yes it's actually an array. It's because pages can have multiple collection items associated with them. I've seen it used on a couple pages.

script/check-collections.js
accbjt: Yes I can do that.

accbjt:
@ncksllvn I do agree we need to rethink these things. Everything is scattered and I think implementing a page and adding it to the LeftRailNav is brittle. It's not very straight forward. I think we might need to add some conventions so that it's easier to implement without having issues. Also it should be easier for content creators to create new pages with minimal developer's assistance.

jbalboni:
I also agree that the left nav configuration is confusing, however, I think a lot of it is going to have to change as a result of integrating with the CMS. So we probably don't want to spend much time right now re-thinking it. I'm not sure how that impacts this PR, but wanted to throw it out there.

ncksllvn:
Yeah, totally agree @jbalboni.

src/site/stages/build/plugins/tests/check-collections.spec.js
```diff
@@ -0,0 +1,24 @@
...
+          collection: 'healthCostOfCare',
+        },
+      },
+      {},
+      () => {
+        expect('My Account').to.equal('My Account');
+      },
+    );
+  });
+});
```
ncksllvn: I think your file extension should be `.unit.spec.js`.

script/check-collections.js
```diff
@@ -0,0 +1,67 @@
...
+      done(
+        new Error(`You have ${brokenCollections.length} broken collections \n`),
+      );
+    }
+
+    done();
+  };
+};
+
+module.exports = checkCollections;
```
ncksllvn: I think we can remove this now, since it's moved into `src/site/`.

package.json
```diff
@@ -44,6 +44,7 @@
     "test:unit:rx": "BABEL_ENV=test mocha test/helper.js --opts src/platform/testing/unit/mocha.opts --recursive 'test/rx/**/*.unit.spec.js?(x)' ",
     "test:unit:gi": "BABEL_ENV=test mocha test/helper.js --opts src/platform/testing/unit/mocha.opts --recursive 'test/gibct/**/*.unit.spec.js?(x)' ",
     "test:unit:login": "BABEL_ENV=test mocha --opts src/platform/testing/unit/mocha.opts --recursive 'test/login/**/*.unit.spec.js?(x)' ",
+    "test:unit:plugins": "BABEL_ENV=test mocha --inspect --opts src/platform/testing/unit/mocha.opts --recursive 'src/site/stages/build/plugins/**/*.spec.js?(x)' ",
```
ncksllvn: `plugins` seems unclear to me. How about something like `test:unit:build` instead?

src/site/stages/build/plugins/check-collections.js
```diff
@@ -0,0 +1,76 @@
+/* eslint-disable no-console */
+const path = require('path');
+const collections = require('../data/collections.json');
+
+/**
+ * This checks whether there are broken collections in a front-matter .md file.
+ * It checks whether a collection or children is available in the MetalSmith collections list
+ * @param {object} buildOptions
+ */
+
```
ncksllvn: At the top of this file, can we give a complete example of how to configure a left nav similar to how we do with the [fragments](https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/site/stages/build/plugins/apply-fragments.js) middleware? I'd like to make this a standard for every build step.

src/site/stages/build/plugins/check-collections.js
```diff
@@ -0,0 +1,76 @@
...
+      if (file.children && !VALID_COLLECTIONS.includes(file.children)) {
+        accum.push(
+          `Error: Your [ children: "${
+            file.children
+          }" ] does not exist in the collection: ${key}`,
+        );
+      }
+
+      return accum;
+    }, []);
```
ncksllvn: I still don't think `.reduce` is the right choice of function here. When I think of `.reduce`, I think of iterating through an array to produce a single value, not another array. `.map` is what I think of for producing an array out of another.  You could also reduce some duplicate code by making sure `file.collection` is always the same data structure. To be a bit more expressive, you could use a `Set` for `VALID_COLLECTIONS`.

This is untested, but something like -

```js
.filter(
  fileName =>
    path.extname(fileName) === '.html' || path.extname(fileName) === '.md',
)

// Map the file names into an array of error messages, or null if there's no error.
.map(fileName => {
  let errorMessage = '';
  const file = files[fileName];

  if (file.collection) {
    const collectionIsString = file.collection.constructor.name === 'String';
    const collections  = collectionIsString ? [file.collection] : file.collection;
    
    collections.forEach(collectionName => {
      if (!VALID_COLLECTIONS.has(collectionName)) {
        errorMessage += `Error: Your [ collection "${collectionName}" ] does not exist in the collection: ${key}\n`;
      }
    });
  }

  // Checks if a page children is valid. A child collection is always going to be a string
  if (file.children && !VALID_COLLECTIONS.has(file.children)) {
    errorMessage += `Error: Your [ children: "${file.children}" ] does not exist in the collection: ${key}`;
  }

  return errorMessage;
})

// Remove null error messages
.filter(error => !!error);
```

src/site/stages/build/plugins/check-collections.js
```diff
@@ -0,0 +1,76 @@
...
+    .filter(
+      fileName =>
+        path.extname(fileName) === '.html' || path.extname(fileName) === '.md',
+    )
+    // returns an Array of Error messages for pages that have collection errors
+    .reduce((accum, key) => {
+      const file = files[key];
+
+      // Checks whether a page has a collection which is an Array. A page can have more then 1 collection
+      console.log(`Checking ${key}`);
```
ncksllvn: I don't think we need this `console.log`.

src/site/stages/build/index.js
```diff
@@ -52,6 +52,7 @@ function defaultBuild(BUILD_OPTIONS) {
   smith.use(filenames());
 
   smith.use(applyFragments(BUILD_OPTIONS));
+  smith.use(checkCollections(BUILD_OPTIONS));
```
ncksllvn: Just wondering, is there a specific reason you chose to register this plugin at this point in the plugin chain?

src/site/stages/build/plugins/tests/check-collections.spec.js
```diff
@@ -0,0 +1,24 @@
+import { expect } from 'chai';
+import checkCollections from '../check-collections';
+import ware from 'ware';
```
ncksllvn: Is `ware` installed in our project? 

Also, curious about your choice with using `ware`, versus working with the plugin directly.

rtluu:
@ncksllvn & @accbjt , is there valid work here in this PR now that we've made a determination on the CMS approach? If not could we close this issue?

accbjt:
@ncksllvn @jbalboni I put this PR on pause while everything settles with the CMS stuff. Should we wait for the CMS to be integrated before we continue on? I think this should be revisited so we can talk further about it.

jbalboni:
I want to leave this up to @ncksllvn. The full CMS integration will take a while, so there is a potential use case for it for the next couple months, but @ncksllvn is following the vagov-content changes that are happening most closely and I think has the best context on if this will help us catch issues until everything is in the CMS.

ncksllvn:
Yeah, make sense to me. We're going to add a few more hubs probably before the CMS is rolled out, so this may save us some headache. Can we reduce this PR just to just these two files -

- `src/site/stages/build/plugins/check-collections.js`
- `src/site/stages/build/index.js`

And remove the `console.log` per this [comment](https://github.com/department-of-veterans-affairs/vets-website/pull/9089/files#r239267283), then we can merge this.

I like your idea for adding unit tests (long overdue), but I'm not sure about using `ware` and I'd like to revisit that idea as a whole later on.

src/site/stages/build/plugins/check-collections.js
accbjt: @ncksllvn can we create a separate ticket for this. I think we need to consolidate all the left rail nav stuff and then add the example in there. Right now let's just get this working so we can get some checks in while we are doing the CMS integration. I do agree that we need this somewhere close tot the code.

ncksllvn (approved):
LGTM

src/site/stages/build/plugins/check-collections.js
ncksllvn: Nah, no worries on the separate ticket. This is good as-is.

accbjt:
@ncksllvn Can you take a look at this now. I made the updates you requested. The reason I used ware for the tests was I wanted to simulate using middleware without having to do a full on build. Since MetalSmith uses ware for it's middleware library I thought it would be appropriate to use it for our tests.

PR #9093, opened 2018-11-09
All claims split training pay mg

markgreenburg:
## Description
Splits training pay questions onto two pages:
- First asks if veteran will receive training pay
- Second, conditional page asks whether or not they want to waive training pay to receive VA pay

## Testing done
- Tested locally
- Added unit tests

## Screenshots
First Page:
screen shot 2018-11-09 at 1 43 43 pm

-----

Follow-up page:
screen shot 2018-11-09 at 1 46 29 pm

## Acceptance criteria
- [x] Waive training pay question is on its own page
- [x] Content for questions and descriptions matches latest in ticket
- [x] Veteran must answer waiver question if they indicated they will receive training pay

## Definition of done

### DIFF
--- a/src/applications/disability-benefits/all-claims/config/schema.js
+++ b/src/applications/disability-benefits/all-claims/config/schema.js
@@ -786,6 +786,9 @@ const schema = {
       type: 'string',
       enum: serviceBranches,
     },
+    hasTrainingPay: {
+      type: 'boolean',
+    },
     waiveTrainingPay: {
       type: 'boolean',
     },
--- a/src/applications/disability-benefits/all-claims/pages/trainingPayWaiver.js
+++ b/src/applications/disability-benefits/all-claims/pages/trainingPayWaiver.js
@@ -22,6 +22,7 @@ export const uiSchema = {
 
 export const schema = {
   type: 'object',
+  required: ['waiveTrainingPay'],
   properties: {
     waiveTrainingPay,
   },
--- a/src/applications/disability-benefits/all-claims/pages/trainingPayWaiver.js
+++ b/src/applications/disability-benefits/all-claims/pages/trainingPayWaiver.js
@@ -4,8 +4,9 @@ import { waiveTrainingPayDescription } from '../content/trainingPayWaiver';
 const { waiveTrainingPay } = fullSchema.properties;
 
 export const uiSchema = {
+  'ui:title': 'Training pay waiver',
   waiveTrainingPay: {
-    'ui:title': 'Training pay waiver',
+    'ui:title': ' ',
     'ui:description': waiveTrainingPayDescription,
     'ui:widget': 'yesNo',
     'ui:options': {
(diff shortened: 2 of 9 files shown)


peggygannon (approved):
@markgreenburg Looks good to me. Thanks!

src/applications/disability-benefits/all-claims/config/schema.js
```diff
@@ -786,6 +786,9 @@ const schema = {
       type: 'string',
       enum: serviceBranches,
     },
+    hasTrainingPay: {
```
cvalarida: Ahh, okay. So this is not intended to be sent to the back end? :+1: 

Should this be added to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/15004?

src/applications/disability-benefits/all-claims/pages/trainingPayWaiver.js
```diff
@@ -0,0 +1,29 @@
...
+const { waiveTrainingPay } = fullSchema.properties;
+
+export const uiSchema = {
+  'ui:title': 'Training pay waiver',
+  waiveTrainingPay: {
+    'ui:title': ' ',
+    'ui:description': waiveTrainingPayDescription,
+    'ui:widget': 'yesNo',
+    'ui:options': {
+      yesNoReverse: true,
```
cvalarida: Oh hey. Didn't know this was a thing. What's it do exactly?

cvalarida (approved):
Looks good. Had a couple questions, but they're mostly for my own edification.

src/applications/disability-benefits/all-claims/config/schema.js
markgreenburg: Yes, and yes :)

It doesn't strictly _need_ to be sent up (since it can be deduced), but it's a lot simpler to just send the value up than to have the backend deduce it with code.

markgreenburg: [Added to that ticket]

src/applications/disability-benefits/all-claims/pages/trainingPayWaiver.js
markgreenburg: Yeah, this is awful. It literally reverses answers, so `true` would make its way into the data as `false`. The reason for this here is that the question is worded in such a way that 'yes means no' and vice-versa. Given:
```

Comment (cvalarida) [reply src/applications/disability-benefits/all-claims/pages/trainingPayWaiver.js]: :thinking: So if they select `I want to get...`, it'll get sent up to the api as `waiveTraningPay: true`. That makes sense.

Could we just reverse the question labels, though?

```js
labels: {
  N: 'I don’t want to get VA compensation pay for the days I receive training pay.',
  Y: 'I want to get VA compensation pay instead of training pay.',
},
```

As far as I'm concerned, it's six of one, half-dozen of the other, but I'm curious.

PR #9095, opened 2018-11-12
Tear down Brand Consolidation feature flag from build script

ncksllvn:
## Description
This pull request flattens out the build to only build Brand Consolidation, since the website has launched now and `--brand-consolidation-enabled` is now always `true` 🎉 Additionally, it updates references to Vets.gov buildtypes to instead use the VA.gov equivalent or localhost.

This PR has a wider scope than I was hoping, but as I was trying to convert the site to build under either `vagovdev` or `localhost`, I realized that would leave the `development` build type still floating around the client-side, which is confusing and deprecated, so I decided to address it all at once.

This PR does the following -

- Flattens out conditions/steps in the build script and removes redundant logic, since Brand Consolidation is now always true.
- Reconfigures the watch task and dev server to build VA.gov under the `localhost` buildtype.
- Separates out the Vets.gov-to-VA.gov redirect build into its own stage in the Jenkinsfile, since those are separate buildtypes with a unique build system (generating redirects.) 
- Updates references to the `development` buildtype to instead use the `vagovdev` or `localhost` buildtype.
- Updates our version of Yarn in the package.json engines property
    - This has nothing to do with this PR, but apparently this was overdue, because when I ran `yarn install` with version 1.5.1 I discovered that the `integrity` fields were being stripped.
    - I don't mind reverting this change at all. I would rather use NPM anyway.

## Testing done
- [x] Ensure redirects are generated okay
- [x] Ensure VA build-types generates okay
- [x] Ensure localhost builds okay

## Acceptance criteria
- [x] The FE build is more refined, as the condition for `brand-consolidation-enabled` is now always `true`.

## Definition of done

## Ticket
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/15065


Jenkinsfile
```diff
@@ -79,37 +61,16 @@ def buildDetails = { vars ->
...
-  def repo = github.getRepository('department-of-veterans-affairs/vets-website')
-  def pr = repo.queryPullRequests().head("department-of-veterans-affairs:${env.BRANCH_NAME}").list().asList().get(0)
-
-
-  // Post our comment with broken links formatted as a Markdown table
-  pr.comment("### Broken links found by Jenkins\n\n|File| Link URL to be fixed|\n|--|--|\n" +
-             broken_links.replaceAll(/\[production\] |>>> href: |,/,"|") +
-             "\n\n _Note: Long file names or URLs may be cut-off_")
-}
-
```
ncksllvn: This could go, because page are edited over at `vagov-content` so the link-checker belongs there instead.

Jenkinsfile
```diff
@@ -189,6 +147,29 @@ node('vetsgov-general-purpose') {
...
+        }
+      }
+
+      parallel builds
+    } catch (error) {
+      notify()
+      throw error
+    }
+  }
+
```
ncksllvn: I separated out the entry point for building Vets-to-VA redirects from the rest of the build script, because it's two completely different operations for two different sets of build-types.

Jenkinsfile
```diff
@@ -197,17 +178,11 @@ node('vetsgov-general-purpose') {
...
-
+      for (int i=0; i<VAGOV_BUILDTYPES.size(); i++) {
+        def envName = VAGOV_BUILDTYPES.get(i)
         builds[envName] = {
           dockerImage.inside(args) {
-            sh "cd /application && npm --no-color run build -- ${cmsDeployArg} --buildtype=${envName}"
+            sh "cd /application && npm --no-color run build -- --buildtype=${envName}"
             sh "cd /application && echo \"${buildDetails('buildtype': envName, 'ref': ref)}\" > build/${envName}/BUILD.txt"
           }
         }
```
ncksllvn: I also removed the CMS argument altogether, since we need a direct-to-prod build-type for it to mean anything. That work will be tracked here, https://github.com/department-of-veterans-affairs/vets.gov-team/issues/14855.

Jenkinsfile
```diff
@@ -1,71 +1,55 @@
...
   // abort the job if we're not on deployable branch (usually master) and there's a newer build going now
-  env.BRANCH_NAME != devBranch &&
-  env.BRANCH_NAME != stagingBranch &&
-  env.BRANCH_NAME != prodBranch &&
+  !IS_DEV_BRANCH &&
+  !IS_STAGING_BRANCH &&
+  !IS_PROD_BRANCH &&
   !env.CHANGE_TARGET &&
   currentBuild.nextBuild
 }
```
ncksllvn: I did some cleanup with the `env.BRANCH_NAME != devBranch` conditions.

package.json
```diff
@@ -218,7 +215,7 @@
   },
   "engines": {
     "node": "8.10.0",
-    "yarn": "1.5.1"
+    "yarn": "1.12.3"
```
ncksllvn: There was an `integrity` property in our `yarn.lock` that was overwritten when I ran `yarn install` with `1.5.1`, so I did some searching to find that we must have checked-in a  `yarn.lock` version generated by a new version of Yarn, without upgrading this property as well.

script/build.js
```diff
@@ -52,18 +49,12 @@ function defaultBuild(BUILD_OPTIONS) {
...
-
+  smith.use(applyFragments(BUILD_OPTIONS));
   smith.use(collections(BUILD_OPTIONS.collections));
   smith.use(leftRailNavResetLevels());
   smith.use(dateInFilename(true));
-
   smith.use(assets(BUILD_OPTIONS.appAssets));
-  if (BUILD_OPTIONS.contentAssets) {
-    smith.use(assets(BUILD_OPTIONS.contentAssets));
-  }
```
ncksllvn: These conditions will always evaluate to true.

ncksllvn:
In the client-side code, we sometimes check to see if we're in the "development" build-type. Historically, this means localhost or the Dev environment. However, we now have the ability to distinguish between the two, and it seems to me that the instances where we do check for the "development" environment we really mean "localhost" so I made that change by switching instances of `__BUILDTYPE__ === 'development'` to `__BUILDTYPE__ === 'localhost'`.

script/options.js
```diff
@@ -2,14 +2,11 @@
...
 const path = require('path');
 const commandLineArgs = require('command-line-args');
-const applyHerokuOptions = require('./heroku-helper');
 const environments = require('./constants/environments');
 const hostnames = require('./constants/hostnames');
 
 const COMMAND_LINE_OPTIONS_DEFINITIONS = [
   { name: 'buildtype', type: String, defaultValue: environments.LOCALHOST },
-  { name: 'brand-consolidation-enabled', type: Boolean, defaultValue: false },
-  { name: 'no-sanity-check-node-env', type: Boolean, defaultValue: false },
```
ncksllvn: I removed this `no-sanity-check-node-env` because it never made much sense to me.

src/platform/utilities/environment/index.js
```diff
@@ -41,7 +29,7 @@ function getEnvironment() {
   let platform;
 
   if (location.port === '3001') {
-    platform = 'local';
+    platform = 'localhost';
   } else if (location.port === `${process.env.WEB_PORT || 3333}`) {
     platform = 'e2e';
   } else {
```
ncksllvn: This file could use some serious sprucing up. The `location.hostname` and and the `process.env.BASE_URL` pieces are particularly confusing. Apparently, our Nightwatch Docker runs under `http://vets-website`, so we have some dynamic pieces like that.

ncksllvn:
This is finally ready for review! PTAL when you get the chance.

Jenkinsfile
jbalboni: Do we have links in layout files? We might miss a broken link if that's the case.

src/applications/facility-locator/api/index.js
```diff
@@ -3,4 +3,4 @@ import MockApi from './MockLocatorApi';
 import LiveApi from './LocatorApi';
 
 /* global __BUILDTYPE__ */
-export default (__BUILDTYPE__ === 'development') ? MockApi : LiveApi;
+export default (__BUILDTYPE__ === 'localhost') ? MockApi : LiveApi;
```
jbalboni: They might be intending to run their mock api in dev.va.gov. I vaguely remember that being the case.

jbalboni (approved):

jbalboni:
It's nice to see a bunch of this branching logic cleaned up!

src/applications/facility-locator/api/index.js
ncksllvn: Left a comment over here, https://github.com/department-of-veterans-affairs/vets-website/pull/8651#issuecomment-438314848.

r-mc2: My [feedback](https://github.com/department-of-veterans-affairs/vets-website/pull/8651#issuecomment-438377371) on this.

Jenkinsfile
```diff
@@ -1,71 +1,55 @@
...
+  'production'
+]
 
-def envNames = [
-  // Vets.gov envs
-  'development', 'staging', 'production',
-  // VA.gov envs
-  'preview', 'vagovdev', 'vagovstaging', 'vagovprod'
+VAGOV_BUILDTYPES = [
+  'preview',
```
wyattwalter: Don't need this!

ncksllvn: Actually have that captured here, https://github.com/department-of-veterans-affairs/vets.gov-team/issues/15080!

script/options.js
```diff
@@ -39,122 +35,67 @@ function gatherFromCommandLine() {
...
+  if (options.buildtype !== environments.LOCALHOST || isHerokuBuild) {
     options.port = 80;
     options.protocol = 'https';
-    options.host = hostnames[options.buildtype];
+
+    if (isHerokuBuild) {
+      options.host = `${process.env.HEROKU_APP_NAME}.herokuapp.com`;
+    } else {
+      options.host = hostnames[options.buildtype];
+    }
```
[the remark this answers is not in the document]
wyattwalter: Agree on both fronts.

PR #9108, opened 2018-11-14
All claims transform cv

cvalarida:
## Description
This was a bit more in-depth than I anticipated. I ended up needing to write a new data utils function. There's a comment on the line that I use it stating why it was necessary.

## Testing done
It was basically all writing unit tests, so...unit tests?

### Methodology
I ran through the form a few times and entered in both as much and as little as I could. There are probably still other paths we need to take to make sure all the possible routes are tested. The data that made it into the `.json` files came directly from the redux store after I made it to the review page.

## Screenshots
N/A

## Acceptance criteria
- [x] The form data gets run through the transform function and matches the schema

## Definition of done

### DIFF
--- a/src/applications/disability-benefits/all-claims/config/schema.js
+++ b/src/applications/disability-benefits/all-claims/config/schema.js
@@ -845,7 +845,7 @@ const schema = {
       type: 'array',
       items: {
         type: 'object',
-        required: ['condition', 'cause', 'disabilityStartDate'],
+        required: ['condition', 'cause'],
         properties: {
           condition: {
             type: 'string',
--- a/src/applications/disability-benefits/all-claims/tests/schema/schema.unit.spec.js
+++ b/src/applications/disability-benefits/all-claims/tests/schema/schema.unit.spec.js
@@ -3,41 +3,8 @@ import fs from 'fs';
 import path from 'path';
 import { Validator } from 'jsonschema';
 
-import get from '../../../../../platform/utilities/data/get';
-
 import formConfig from '../../config/form';
 import fullSchema from '../../config/schema';
-
-const conditionListPaths = [
-  'powDisabilities',
-  'vaTreatmentFacilities.relatedDisabilities',
-];
-
-/**
- * Returns a list of conditions that don't have a matching claimed condition.
- * @param {Object} data - The form data after it's been transformed
- * @return {Array<String>} - An array of condition names that aren't being claimed
- */
-function getMismatchedConditions(data) {
-  const claimedConditions = data.ratedDisabilities
-    ? data.ratedDisabilities.map(d => d.name)
-    : [];
-  if (data.newDisabilities) {
-    data.newDisabilities.forEach(d => claimedConditions.push(d.condition));
-  }
-
-  const unmatchedConditions = new Set();
-  conditionListPaths.forEach(dataPath => {
-    const relatedConditions = get(dataPath, data, []);
-    relatedConditions.forEach(name => {
-      if (!claimedConditions.includes(name)) {
-        unmatchedConditions.add(name);
-      }
-    });
-  });
-
-  return Array.from(unmatchedConditions);
-}
 
 describe('526 all claims schema tests', () => {
   const v = new Validator();
@@ -52,40 +19,11 @@ describe('526 all claims schema tests', () => {
       ).form526;
       const result = v.validate(JSON.parse(submitData), fullSchema);
 
-      const mismatchedConditions = getMismatchedConditions(submitData);
-
-      if (mismatchedConditions.length) {
-        // eslint-disable-next-line
-        console.log(
-          "Related conditions found that aren't being claimed:",
-          mismatchedConditions,
-        );
-      }
-
       if (!result.valid) {
         console.log(`Validation errors found in ${file}`); // eslint-disable-line
         console.log(JSON.stringify(result.errors, null, 2)); // eslint-disable-line
       }
       expect(result.valid).to.be.true;
-      expect(mismatchedConditions).to.eql([]);
     });
-  });
-});
-
-describe('checkConditionNames', () => {
-  it('should return a list of unmatched conditions', () => {
-    const formData = {
-      powDisabilities: ['Something', 'Ninja condition'],
-      vaTreatmentFacilities: {
-        relatedDisabilities: ['Something else', 'Ghost condition'],
-      },
-      ratedDisabilities: [{ name: 'Something' }],
-      newDisabilities: [{ condition: 'Something else' }],
-    };
-
-    expect(getMismatchedConditions(formData)).to.eql([
-      'Ninja condition',
-      'Ghost condition',
-    ]);
   });
 });
--- a/src/applications/disability-benefits/all-claims/tests/schema/schema.unit.spec.js
+++ b/src/applications/disability-benefits/all-claims/tests/schema/schema.unit.spec.js
@@ -20,7 +20,7 @@ describe('526 all claims schema tests', () => {
       const result = v.validate(JSON.parse(submitData), fullSchema);
 
       if (!result.valid) {
-        console.log(result.errors); // eslint-disable-line
+        console.log(JSON.stringify(result.errors, null, 2)); // eslint-disable-line
       }
       expect(result.valid).to.be.true;
     });
--- a/src/applications/disability-benefits/all-claims/tests/schema/schema.unit.spec.js
+++ b/src/applications/disability-benefits/all-claims/tests/schema/schema.unit.spec.js
@@ -20,6 +20,7 @@ describe('526 all claims schema tests', () => {
       const result = v.validate(JSON.parse(submitData), fullSchema);
 
       if (!result.valid) {
+        console.log(`Validation errors found in ${file}`); // eslint-disable-line
         console.log(JSON.stringify(result.errors, null, 2)); // eslint-disable-line
       }
       expect(result.valid).to.be.true;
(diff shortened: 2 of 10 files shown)


src/applications/disability-benefits/all-claims/config/schema.js
```diff
@@ -845,7 +845,7 @@ const schema = {
       type: 'array',
       items: {
         type: 'object',
-        required: ['condition', 'cause', 'disabilityStartDate'],
```
cvalarida: Turns out this isn't even a schema property. :man_facepalming:

src/platform/utilities/data/omit.js
```diff
@@ -5,11 +5,17 @@
  * @param {Object} object
  */
 export default function omit(fields, object) {
-  return Object.keys(object).reduce((newObj, k) => {
+  let fieldOmitted = false;
```
cvalarida: Found a bug in `omit` where if nothing was actually omitted, we'd still get a different reference to the root object. I discovered it because this was making a test for `removeDeeplyEmptyObjects` fail.

src/platform/utilities/data/removeDeeplyEmptyObjects.js
```diff
@@ -0,0 +1,41 @@
...
+}
+
+/**
+ * Remove any deeply empty objects.
+ * Note: This function is non-mutative; it'll return a new object without the empty properties.
+ *
+ * @param {Object} root - The object we're checking
+ * @param {Function} isEmpty - A callback to determine if a value should be considered empty
+ */
+export default function removeDeeplyEmptyObjects(root, checkEmpty = isEmpty) {
```
cvalarida: Definitely open to suggestions on how to make this function clearer!

src/platform/utilities/tests/data.unit.spec.js
```diff
@@ -54,7 +55,10 @@ describe('data utils (lodash replacements)', () => {
       },
       string: 'string!',
       array: ['0', 1, null, { s: 'thing', o: { k: 'I am nested' } }],
-      func: function() { return this.int; }, // eslint-disable-line
+      // eslint-disable-next-line
```
cvalarida: Thanks, Prettier.

src/platform/utilities/tests/data.unit.spec.js
```diff
@@ -463,4 +473,36 @@ describe('data utils (lodash replacements)', () => {
       });
     });
   });
+
+  describe('removeDeeplyEmptyObjects', () => {
```
cvalarida: If there are any more tests that need to be added for this, please let me know.

src/applications/disability-benefits/all-claims/utils.jsx
```diff
@@ -158,6 +160,45 @@ export function prefillTransformer(pages, formData, metadata) {
...
+ * Transforms the related disabilities object into an array of strings
+ * @param {Object} object - The object with dynamically generated property names
+ * @return {Array} - An array of the property names with truthy values
+ */
+export function transformRelatedDisabilities(object) {
+  return Object.keys(object).filter(key => object[key]);
+}
+
+export function transform(formConfig, form) {
+  // Have to do this first or it messes up the results from transformRelatedDisabilities for some reason.
```
markgreenburg: hmm, do you know why?

src/applications/disability-benefits/all-claims/utils.jsx
```diff
@@ -158,6 +160,45 @@ export function prefillTransformer(pages, formData, metadata) {
...
+      _.set(
+        'relatedDisabilities',
+        transformRelatedDisabilities(facility.relatedDisabilities),
+        facility,
+      ),
+    );
+    clonedData.vaTreatmentFacilities = newVAFacilities;
+  }
+  if (clonedData['view:isPOW']) {
+    clonedData['view:isPOW'] = transformRelatedDisabilities(
```
markgreenburg: Nice, at least there are two places where this is useful!

markgreenburg: Will this ever be true? `clonedData` has already been run through `transformForSubmit`, which removes `view:` properties, right?

If I do a test submit with a POW disability, I get:
```{"confinements":[{"from":"1980-05-05","to":"1999-01-01"}],"powDisabilities":{"ASHD":true}```

Comment (markgreenburg) [reply src/applications/disability-benefits/all-claims/utils.jsx]: Also, I'm not sure if we need to include `powDisabilities` as a child of confinements (i.e. if a `powDisabilities` list has to be associated with its given confinement period) - but I'd think so. @smluthi ^

File: src/applications/disability-benefits/all-claims/utils.jsx [hunk 8]
```diff
@@ -158,6 +160,45 @@ export function prefillTransformer(pages, formData, metadata) {
...
+      _.get('view:isPOW.powDisabilities', clonedData),
+    );
+  }
+
+  return JSON.stringify({
+    form526: JSON.stringify(clonedData),
```
markgreenburg: Would be nice if we didn't have to stringify, parse, and then re-stringify the form data, but I guess that depends on what the deal is with the `transformRelatedDisabilities` bit you commented on earlier.

src/platform/utilities/data/omit.js
markgreenburg: This makes sense to me, but I'm not sure what the expected behavior is. If you're using `omit` from the data utils, do you expect to always get a new object? Or do you only expect to get a new object when a field is actually omitted? Maybe a silly question but I don't know the answer.

src/applications/disability-benefits/all-claims/tests/utils.unit.spec.jsx
```diff
@@ -435,4 +436,18 @@ describe('526 helpers', () => {
...
+        'Some condition name': true,
+        'Another condition name': true,
+        'This condition is falsey!': false,
+      };
+      expect(transformRelatedDisabilities(relatedDisabilities)).to.eql([
+        'Some condition name',
+        'Another condition name',
+      ]);
+    });
+  });
```
markgreenburg: Do you think we should add some tests for the `transform` function in general?

markgreenburg: Unless...maybe the schema tests are sufficient?

src/applications/disability-benefits/all-claims/tests/schema/schema.unit.spec.js
```diff
@@ -0,0 +1,29 @@
...
+  const files = fs.readdirSync(__dirname);
+  files.filter(file => file.endsWith('json')).forEach(file => {
+    it(`should validate ${file}`, () => {
+      const contents = JSON.parse(
+        fs.readFileSync(path.join(__dirname, file), 'utf8'),
+      );
+      const submitData = JSON.parse(
+        formConfig.transformForSubmit(formConfig, contents),
+      ).form526;
+      const result = v.validate(JSON.parse(submitData), fullSchema);
```
markgreenburg: I don't think this will catch the dynamic schema bits relating to object with boolean disability keys -> array of disability key strings.

markgreenburg:
Looking good! Just a couple of small issues with the submit transformer.

src/platform/utilities/data/omit.js
cvalarida: I think in general, it should only return a new reference if something changed. The idea is so we can compare equality like `props.foo === newProps.foo` when `newProps.foo` is run through `omit()`.

cvalarida: this is also a bug in `set()`, I believe, but I bypassed it in `removeDeeplyEmptyObjects` by just checking for equality first.

src/applications/disability-benefits/all-claims/utils.jsx
cvalarida: Why? No :disappointed:. It was changing `relatedDisabilities: ['Stress Disorder']` to
```
"relatedDisabilities":[{"0":"S","1":"t","2":"r","3":"e","4":"s","5":"s","6":" ","7":"D","8":"i","9":"s","10":"o","11":"r","12":"d","13":"e","14":"r"}]}]
```

src/applications/disability-benefits/all-claims/utils.jsx
cvalarida: Oh, good point about removing the `view:` field. I had this check before I moved the `transformForSubmit` up to the top. Lemme look into it...

The latest EVSS swagger doc has 
image

:thinking: I'm...not actually sure what the `powDisabilities` matches up to in that doc... :confused:

src/applications/disability-benefits/all-claims/utils.jsx
cvalarida: Yup. Exactly. So having the `transformForSubmit` at the end like I originally had it was messing up the `relatedDisabilities` array, so I ended up putting it at the beginning of this function. `transformForSubmit` returns a JSON-stringified object, so in order to continue massaging the data, I had to parse it...only to stringify it again right at the end. :disappointed:

src/applications/disability-benefits/all-claims/tests/utils.unit.spec.jsx
cvalarida: Yeah, I was gonna stick with the schema tests. I think testing the `transform` function itself might be redundant.

src/applications/disability-benefits/all-claims/tests/schema/schema.unit.spec.js
cvalarida: Ah, that's true. I can add a check to make sure all the dynamically generated properties (array of strings by this point) match condition names. Good call. :+1:

src/platform/utilities/data/omit.js
markgreenburg: Ahh right, right. That explanation makes sense, thanks!

markgreenburg (approved):
LGTM 🚀 Only had to make a new data util and fix a bug in another to get the transformer to work 😛

PR #9119, opened 2018-11-14
use physical address when present on search and detailed views

rianfowler:
## Description
display physical city when available 

## Testing done
verified locally- couldn't find an example where these attributes differed but I verified with some test strings this was the place the value is normalized.

## Screenshots
screen shot 2018-11-14 at 12 24 32
screen shot 2018-11-14 at 12 24 44

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/src/applications/gi/reducers/profile.js
+++ b/src/applications/gi/reducers/profile.js
@@ -17,9 +17,8 @@ function normalizedAttributes(attributes) {
   const name = attributes.name
     ? attributes.name.toUpperCase()
     : attributes.name;
-  const city = attributes.city
-    ? attributes.city.toUpperCase()
-    : attributes.city;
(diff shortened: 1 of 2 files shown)


src/applications/gi/reducers/profile.js
```diff
@@ -17,9 +17,8 @@ function normalizedAttributes(attributes) {
   const name = attributes.name
     ? attributes.name.toUpperCase()
     : attributes.name;
-  const city = attributes.city
-    ? attributes.city.toUpperCase()
-    : attributes.city;
+  let city = attributes.physicalCity || attributes.city;
```
rianfowler: not clear to me that physical city is always available so using this as a fallback.

erikphansen (approved):

PR #9121, opened 2018-11-15
Catch for null website attribute

bshyong:
## Description

Correctly catches null website attributes in facility locator

Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/14995

## Testing done
Local testing

## Acceptance criteria
- [x] Facilities with null website attributes should not show invalid website link

## Definition of done

### DIFF
--- a/src/applications/facility-locator/containers/FacilityDetail.jsx
+++ b/src/applications/facility-locator/containers/FacilityDetail.jsx
@@ -34,7 +34,7 @@ class FacilityDetail extends Component {
         <div>
           <LocationPhoneLink location={facility} />
         </div>
-        { website &&
+        { website && website !== 'NULL' &&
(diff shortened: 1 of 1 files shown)


ncksllvn (approved):
It sounds like there's a legitimate record in the database for a facility called `NULL` so this doesn't address the root, but either way, LGTM!

bshyong:
@ncksllvn there is a facility called 'NULL'?

ncksllvn:
@bshyong Yeah that's I mean...shouldn't the API itself be returning "facility not found" or something?

bshyong:
@ncksllvn no, the facility exists here; only the website attribute does not. The API returns a string 'NULL' instead of an attribute `null`

PR #9127, opened 2018-11-15
Move most of va-gov's contents to src/site

ncksllvn:
## Description
This pull request moves our Metalsmith files into the `src` directory, which I think will make it less detached from the rest of the website. This is part of a post-launch clean-up effort. Ultimately, this PR -

- Moves the `va-gov/*` into `src/site/`
- Updates references from `va-gov/` to `src/site/`
    - I spent an irresponsible amount of time trying to reconfigure this to not need the `src/site` prefix, but the way Tiny Liquid references includes makes it really really difficult. It's evidently relative to where the build script is executed, rather than relative to a Metalsmith path and I couldn't find any way to change that behavior 😠 
- Restored some files to `va-gov/` that are directly included from `vagov-content`, and would otherwise cause the build to fail. After this, I'll have to change the reference over there, then open another PR here that deletes the duplicates.

I also feel like these files being in the `src` directory may make them more available for TeamSite, where we could use the Tiny Liquid template engine directly and inject the components onto the page.

Ticket - https://github.com/department-of-veterans-affairs/vets.gov-team/issues/15130

## Testing done
Ran the site

## Acceptance criteria
- [x] Repo is (getting) more organized

## Definition of done


script/options.js
```diff
@@ -36,19 +36,23 @@ function gatherFromCommandLine() {
 }
 
 function applyDefaultOptions(options) {
+  const contentPagesRoot = options['content-directory'];
+  const contentRoot = path.join(contentPagesRoot, '../');
```
ncksllvn: That command arg should probably be updated to reference the whole repo instead of just the pages directory.

jbalboni (approved):

PR #9130, opened 2018-11-15
Create src/site/stages

ncksllvn:
## Description
A continuation of https://github.com/department-of-veterans-affairs/vets-website/pull/9127.

The `scripts` directory is getting really cluttered, and based on how much our Metalsmith stack has grown for Brand Consolidation I think we need to reorganize. I was also trying to define a pattern for us to be able to share code between Metalsmith and Webpack. I'm most interested in sharing the `environments` constants.

This PR ultimately -

- Moves `build.js`, Metalsmith middleware, and associated files out of the `script` directory and into `src/site/stages/build`
- Moves `scripts/pre-archive` to `src/site/stages/prearchive` (apparently, "prearchive" is one word)
- Moves `scripts/vets-gov-to-va-gov` to `src/site/stages/vets-gov-to-va-gov`
- Modifies each entry point in `src/stages/` to instead export a `main` function, and executes that function from a module in `script/(build|prearchive|vets-gov-to-va-gov)`

Ticket - https://github.com/department-of-veterans-affairs/vets.gov-team/issues/15130

## Testing done
If this passes in Jenkins, then the testing is 99% done, although I will need to double confirm everything is okay at `vagov-content`.

## Acceptance criteria
- [x] Repo is more organized

## Definition of done


src/site/stages/build/index.js
```diff
@@ -0,0 +1,166 @@
...
+const inPlace = require('metalsmith-in-place');
+const layouts = require('metalsmith-layouts');
+const liquid = require('tinyliquid');
+const markdown = require('metalsmith-markdownit');
+const moment = require('moment');
+const navigation = require('metalsmith-navigation');
+const permalinks = require('metalsmith-permalinks');
+
+const getOptions = require('./options');
+const createBuildSettings = require('./middleware/create-build-settings');
```
jbalboni: Is middleware the right word here? These are technically Metalsmith plugins.

src/site/stages/build/options.js
```diff
@@ -3,12 +3,12 @@
 const path = require('path');
```
jbalboni: We're kind of burying this file in the src folder. I get the way you've laid things out, but I wonder if the extra stages directory buys us a lot. There's a lot of nesting, might be good to see if we can reduce that a little and still be better organized.

I say this partly because I find the deeply nested code in `src/platform` pretty annoying at times (even though it was me that put it that way).

jbalboni (approved):
I have a couple of questions, but this is much better than what we have now no matter what.

src/site/stages/build/index.js
ncksllvn: Ah yep, `'plugin` is the right word. I've been saying "middleware" for ages 😆

src/site/stages/build/options.js
ncksllvn: I really like that idea, but I'm really struggling to think of a good location to contain what is currently in `stages` 🤔

PR #9143, opened 2018-11-17
Add property check test

cvalarida:
## Description
I stumbled into a bug wherein the root property name was being used in two different pages. After a bit, I thought to myself, "What if we could check for that?"

So I wrote a test.

This build is currently failing because of the bug I found, which is to be expected.

## Testing done
This is the testing!

## Screenshots
image

## Acceptance criteria
- [x] The test fails if duplicate property names are used accidentally

## Definition of done


src/applications/disability-benefits/all-claims/tests/property-names.unit.spec.js
```diff
@@ -0,0 +1,53 @@
...
+      // Loop through the properties on the schema for a page
+      Object.keys(get(`${pageName}.schema.properties`, pages, {})).forEach(
+        propName => {
+          // Compile a list of duplicate property names
+          if (
+            properties.has(propName) &&
+            !get(propName, ignoreList, []).includes(pageName)
+          ) {
+            if (duplicatedProperties[propName]) {
+              duplicatedProperties[propName].push(pageName);
```
markgreenburg: So the goal here is to produce a list of every instance of a duplicated property, along with the name of its page for easier referencing?

src/applications/disability-benefits/all-claims/tests/property-names.unit.spec.js
```diff
@@ -0,0 +1,53 @@
...
+            } else {
+              duplicatedProperties[propName] = [
```
markgreenburg: Took me a while to understand what's going on here, probably because it's still Monday morning 😛
Just so I'm clear, you're pulling the page name of the property in the existing `properties` map, and also the `pageName` of the existing page being iterated, and adding both page names to the array of the new duplicated property, so that both instances of the duplication are identified.

markgreenburg (approved):
Cool, I like it. Should probably stay inside the All Claims app because the test might be somewhat confusing without any context (e.g., it's not always a bad thing to have the same property show up on different pages).

src/applications/disability-benefits/all-claims/tests/property-names.unit.spec.js
cvalarida: Yup! That's exactly it. :smile:

PR #9159, opened 2018-11-23
[FacilityLocator] Defect Remediation 

r-mc2:
## Description
- Fixed a typo/bug 'Location' --> 'LocationType'
- Fixed an issue of capitalization of JSON keys for Provider Services (due to Ruby-lib auto-boxing)
  - Updated all internal mock data accordingly
- Updating CCInfo info bubble to be more 508/Accessible (HTML & CSS changes)

### Fixes
- [[#15268] Service Types - The type ahead is not working correctly](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/15268)
- [[#15269]Subsequent "All Facilities" searches (after a CC Provider Search) are not returning results](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/15269)

Also addresses some initial 508 changes.

## Definition of done
- [x] Type-ahead for Provider Services functional

(@'ing for visibility @bshyong & @U-DON )


brettdt:
@ehuntdsva @rtluu

src/applications/facility-locator/config.js
```diff
@@ -2,6 +2,11 @@
 import environment from '../../platform/utilities/environment';
 import isProduction from '../../platform/utilities/environment/isProduction';
 
+// TODO: Remove me when done bug fixing
```
erikphansen: can this be removed now?

src/applications/facility-locator/sass/facility-locator.scss
```diff
@@ -637,6 +643,12 @@
   position: relative;
 }
 
+.ccInfoCloseBtn {
```
erikphansen: rename to `.cc-info-close-btn` for consistency

src/applications/facility-locator/sass/facility-locator.scss
```diff
@@ -637,6 +643,12 @@
...
+  background-color: rgba(0,0,0,0);
+  color: black;
```
erikphansen: use `$color-black` for consistency?

src/applications/facility-locator/sass/facility-locator.scss
```diff
@@ -620,10 +620,16 @@
   display: inline-block;
   font-size: 1.2rem;
   background-image: none;
+  background-color: rgba(0,0,0,0);
   text-align: right;
   text-decoration: none;
   text-align: right;
   padding: 0 !important;
+  margin-bottom: 0px;
```
erikphansen: nit: I like not using units when the value is `0`. So `0px` -> `0`

src/applications/facility-locator/sass/facility-locator.scss
```diff
@@ -620,10 +620,16 @@
   display: inline-block;
   font-size: 1.2rem;
   background-image: none;
+  background-color: rgba(0,0,0,0);
```
erikphansen: hmm... do we already have a var for this value? or a standard way to make a transparent background?

src/applications/facility-locator/config.js
r-mc2: Not as of yet, no. I had originally intended to check in a few more fixes to this branch before opening the PR, but in the interest of speeding things up on our end I opened this up as-is. I'll be opening another PR for later fixes but left that line in to remind me for when I don't need to connect to somewhere else for API reasons (e.g. running UI locally but connecting to a RI API).

src/applications/facility-locator/components/FacilityTypeDropdown.jsx
```diff
@@ -80,26 +80,34 @@ class FacilityTypeDropdown extends Component {
...
                   </span>
                 )}
                 {this.state.showBubble && (
                   <div id="infoBubble">
-                    <span onClick={this.toggleCCInfo}>X</span>
+                    <button
+                      className="ccInfoCloseBtn"
+                      onClick={this.toggleCCInfo}
+                    >
+                      X
```
jbalboni: On a screen reader, this will read out the X, which is confusing. This is the markup for a close button we have on our regular modals:

```js
<button
  className="va-modal-close"
  type="button"
  aria-label="Close this modal"
  onClick={this.handleClose}>
  <i className="fa fa-close"></i>
</button>
```
If you don't want to use an icon there, I would wrap the `X` in a span and add `aria-hidden="true"` to it, so screen readers ignore it. And either way, you should add and `aria-label` to the button so that screen readers read out what the button will actually do.

src/applications/facility-locator/components/ServiceTypeAhead.jsx
```diff
@@ -45,20 +45,19 @@ class ServiceTypeAhead extends Component {
   )
 
   shouldShow = (input, svc) => {
+    /* eslint-disable prettier/prettier */
```
jbalboni: What's the reason for not letting Prettier format this block of code?

src/applications/facility-locator/config.js
```diff
@@ -2,6 +2,11 @@
 import environment from '../../platform/utilities/environment';
 import isProduction from '../../platform/utilities/environment/isProduction';
```
jbalboni: There is an outstanding PR that would require a change here (https://github.com/department-of-veterans-affairs/vets-website/pull/9162), so heads up.

jbalboni (commented):
Added a couple comments

r-mc2:
Just commited some changes based on your recommendations @erikphansen
Let me know if there is a better option for the transparency and I'll update that as well.

src/applications/facility-locator/components/ServiceTypeAhead.jsx
r-mc2: At the time there was some weirdness where it wanted to split the line:
```js
  svc
    .name
      .trim()
      .toLowerCase()
      .includes(...)
```
where before it was fine leaving `svc.name` on the same line when it had the capital "N", which seemed overly excessive on newlines and clarity of the logic.

src/applications/facility-locator/config.js
r-mc2: Ooo, I like.

src/applications/facility-locator/components/FacilityTypeDropdown.jsx
erikphansen: Wow, saving this tidbit in my notes. Thanks, @jbalboni

src/applications/facility-locator/sass/facility-locator.scss
erikphansen: Based on existing code we normally just to `transparent`. But maybe there's a reason to do it as an `rgba` value that I'm not aware of?

src/applications/facility-locator/components/ServiceTypeAhead.jsx
erikphansen: Now I want to weigh in with how _I_ would manually format this bit of code which reminds me why just using Prettier as much as possible is so nice. Honestly, looking at the old version of this, which was run through Prettier, I kinda prefer it ¯\_(ツ)_/¯

erikphansen (approved):
Basically one comment about `transparent` but otherwise looks okay to me. I'd also prefer to just let Prettier do its thing.

src/applications/facility-locator/sass/facility-locator.scss
r-mc2: No, just my lack of knowledge with CSS shorthands 😄 
(was the only thing that came to mind to set alpha levels)

src/applications/facility-locator/components/ServiceTypeAhead.jsx
jbalboni: I prefer the Prettier formatting, but I can also see the appeal of this formatting. Except in cases where Prettier is doing something truly bizarre, we should leave it on and keep the benefits of having a consistently formatting codebase.

@r-mc2 Can you re-enable Prettier here?

jbalboni (approved):

PR #9162, opened 2018-11-26
Centralize access to environment name

ncksllvn:
## Description
This pull request expands on `src/platform/utilities/environment/index` to consume the environment names as defined in `src/site/constants/environments` and to implement helper functions for determine what type of environment code is executing.

For example, we currently use a global constant called `__BUILDTYPE__` to check the name of the environment, which is usually just a floating string value. As we discovered during WBC, this makes it difficult to ensure feature flags are working correctly (at one point we had three production environments and had to check for each throughout the codebase if a feature was not to be introduced into prod), and is not very meaningful especially to those new to our codebase, since `__BUILDTYPE__` is a concept unique to our project.

Ultimately, this PR -

- Moves `src/site/stages/constants` up to the parent directory to be more discoverable
- Implements env constants in the Webpack config
- Implement env constants and helper methods into `platform/utilities/environment`
    - Exported-module now includes `BUILDTYPE`, `BASE_URL`, `API_URL`, `isProduction()`, `isStaging()`, `isDev()`, `isLocalhost()`
- Replaces conditions based on `__BUILDTYPE__` with the equivalent helper method through the codebase
- Replaces all magic strings indicating a build-type with the corresponding constant

### Example
This -

```js
function someFeatureFlag() {
  return __BUILDTYPE__ !== 'vagovprod'
}
```

becomes -

```js
import environment from '../../platform/utilities/environment';

function someFeatureFlag() {
  return !environment.isProduction();
}
```

### TODO
- [x] Add documentation (somewhere)
    - Going to add into README as part of this ticket, https://github.com/department-of-veterans-affairs/vets.gov-team/issues/15144.

### Why this is warranted
With the size of our codebase and now with environment-level feature flags becoming a standard pattern for teams to test a feature prior to production, we need centralized access to environment configuration to increase clarity, discoverability, and to ultimately be easier to maintain as our project continues to evolve, especially with the impending CMS-integration coming next year where our concept of build-types may change entirely.

## Testing done
- Local builds of the environments
- Unit tests provide some safety for this

## Acceptance criteria
- [x] A maintainable pattern for accessing environment data (and thereby defining feature flags) has been established

## Definition of done
    - Todo!

## Ticket
Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/15149


src/platform/testing/unit/mocha-setup.js
```diff
@@ -8,12 +8,11 @@ import chai from 'chai';
...
 import { JSDOM } from 'jsdom';
 import '../../site-wide/moment-setup';
+import ENVIRONMENTS from '../../../site/constants/environments';
+
 // import sinon from 'sinon'
 
-global.__BUILDTYPE__ = process.env.BUILDTYPE || 'vagovdev';
-global.__ALL_CLAIMS_ENABLED__ =
-  global.__BUILDTYPE__ === 'vagovdev' ||
-  process.env.ALL_CLAIMS_ENABLED === 'true';
```
ncksllvn: I searched, and I didn't see any instances of code checking for `__ALL_CLAIMS_ENABLED__`.

src/platform/utilities/environment/index.js
```diff
@@ -1,43 +1,85 @@
...
+  /** Determines whether the current environment is a dev environment. */
+  isDev() {
+    return this.BUILDTYPE === ENVIRONMENTS.VAGOVDEV;
+  },
+
+  /** Determines whether the current environment is a local environment. */
+  isLocalhost() {
+    return this.BUILDTYPE === ENVIRONMENTS.LOCALHOST;
+  },
+};
```
ncksllvn: I'm hoping that this helper functions will be able to serve as feature flags without needing to check the actual build-type, which have names that aren't especially meaningful at first.

src/platform/utilities/environment/stagingDomains.js
```diff
@@ -19,10 +20,6 @@ const domainReplacements = [
   { from: 'https://www\\.myhealth\\.va\\.gov', to: mhvBaseUrl() },
 ];
 
-// This doesn't include preview because we want to redirect to staging urls
-// in preview
-const prodEnvironments = new Set(['production']);
-
```
ncksllvn: This was still set to a Vets.gov environment, so I'm not sure what affect it has updating it to function correctly. I appreciate any advise on this one.

.eslintrc
```diff
@@ -20,7 +20,6 @@
     "jest": true,
   },
   "globals": {
-    "__BUILDTYPE__": true,
```
ncksllvn: I removed this to discourage use of the global `__BUILDTYPE__`.

config/webpack.config.js
```diff
@@ -227,7 +214,7 @@ const configGenerator = (options, apps) => {
     baseConfig.devtool = '#eval-source-map';
   }
 
-  if (options.analyzer) {
+  if (buildOptions.analyzer) {
     baseConfig.plugins.push(
       new BundleAnalyzerPlugin({
         analyzerMode: 'disabled',
```
ncksllvn: There were several variations of the same condition here that I ultimately consolidated into `isOptimizedBuild`.

config/webpack.config.js
```diff
@@ -6,6 +6,8 @@ const BundleAnalyzerPlugin = require('webpack-bundle-analyzer')
 const UglifyJSPlugin = require('uglifyjs-webpack-plugin');
 const webpack = require('webpack');
 const path = require('path');
+const ENVIRONMENTS = require('../src/site/constants/environments');
+const BUCKETS = require('../src/site/constants/buckets');
```
ncksllvn: I was hesitant about importing from `src` into `config`, but I preferred the explicitness over something like deriving  build arguments from where our Metalsmith build calls `configGenerator`.

src/platform/forms/tests/save-in-progress/RoutedSavableApp.unit.spec.jsx
```diff
@@ -432,6 +435,6 @@ describe('Schemaform <RoutedSavableApp>', () => {
 
     expect(fetchInProgressForm.called).to.be.false;
     expect(router.replace.calledWith('/first-in-form-page')).to.be.true;
-    __BUILDTYPE__ = buildType;
+    environment.BUILDTYPE = buildType;
```
ncksllvn: I didn't want to remove it but I don't think `BUILDTYPE` here really matters, because unit tests run under `vagovdev` and the corresponding component only has unique behavior during `localhost` builds.

src/platform/utilities/environment/index.js
```diff
@@ -1,43 +1,85 @@
...
+
+import ENVIRONMENTS from '../../../site/constants/environments';
+
+// eslint-disable-next-line no-undef
+const BUILDTYPE = __BUILDTYPE__;
+
+const {
+  WEB_PORT: RESERVED_E2E_PORT = 3333,
+  API_PORT: E2E_API_PORT = 3000,
+} = process.env;
```
ncksllvn: It took me a while to understand what these environment variables were used for. I think there may have been some confusion at some point about how environment variables are made available throughout Webpack-compiled code, versus being passed into Mocha/E2E/other Node applications. I'd like to address the env variables/port numbers sometime down the road, since `3333` is not just an arbitrary number, and is instead a contract with the Nightwatch config to indicate that we're running E2E tests.

src/platform/utilities/environment/index.js
```diff
@@ -1,43 +1,85 @@
...
+    BUILDTYPE: ENVIRONMENTS.VAGOVPROD,
     BASE_URL: 'https://www.va.gov',
+    API_URL: 'https://api.va.gov',
   },
-  vagovstaging: {
-    API_URL: 'https://staging-api.va.gov',
+
+  [ENVIRONMENTS.VAGOVSTAGING]: {
+    BUILDTYPE: ENVIRONMENTS.VAGOVSTAGING,
     BASE_URL: 'https://staging.va.gov',
```
ncksllvn: We could derive the `BASE_URL` by importing `site/constants/hostnames` or even by passing `BUILD_OPTIONS.hostUrl` from Metalsmith into Webpack, but I preferred the explicitness of having hardcoded values on the Webpack side of things.

src/site/stages/build/options.js
```diff
@@ -3,8 +3,8 @@
 const path = require('path');
 const commandLineArgs = require('command-line-args');
 
-const ENVIRONMENTS = require('../constants/environments');
-const HOSTNAMES = require('../constants/hostnames');
+const ENVIRONMENTS = require('../../constants/environments');
+const HOSTNAMES = require('../../constants/hostnames');
```
ncksllvn: I moved the constants up to the parent directory to be more discoverable.

src/applications/personalization/dashboard/isPersonalizationEnabled.js
```diff
@@ -1,6 +1,3 @@
-import dashboardManifest from './manifest';
-
 export default function isPersonalizationEnabled() {
-  if (__BUILDTYPE__ !== 'production') return true;
-  return dashboardManifest.production;
+  return true;
 }
```
ncksllvn: This should ultimately be torn down entirely, but that's outside the scope of this PR.

src/platform/monitoring/frontend-metrics/feature-flag.js
```diff
@@ -4,7 +4,5 @@
  * @module platform/monitoring/frontend-metrics/feature-flag
  */
 export default function isMetricsEnabled() {
-  const environments = ['vagovdev', 'vagovstaging', 'vagovprod'];
-
-  return !!environments.includes(__BUILDTYPE__);
+  return true;
 }
```
ncksllvn: This should also be torn down entirely, unless for some reason we want the feature flag to solely act as a method of excluding local builds.

script/run-docker-nightwatch.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-node src/platform/testing/e2e/test-server.js --buildtype=vagovprod --host=0.0.0.0 --port=3001 &
+node src/platform/testing/e2e/test-server.js --buildtype=vagovprod &
```
ncksllvn: There's a lot of discrepancy between port numbers and that data is passed all over our code base. By removing this, E2E tests will execute consistently on port `3333`.

config/webpack.config.js
```diff
@@ -175,21 +179,11 @@ const configGenerator = (options, apps) => {
...
-        'process.env': {
-          API_PORT: process.env.API_PORT || 3000,
-          WEB_PORT: process.env.WEB_PORT || 3333,
-          API_URL: process.env.API_URL
-            ? JSON.stringify(process.env.API_URL)
-            : null,
-          BASE_URL: process.env.BASE_URL
-            ? JSON.stringify(process.env.BASE_URL)
-            : null,
-        },
```
ncksllvn: These can all for different reasons -

`process.env.API_PORT` and `process.env.WEB_PORT` - were only available for use potentially in our [E2E config](https://github.com/department-of-veterans-affairs/vets-website/blob/9fad755e49b0a3d8d271ea0dc710f17ef2426ef5/src/platform/utilities/environment/index.js#L22). However, they weren't actually useful at all, because Nightwatch-Docker tests (which execute on Jenkins) were executing under port `3001`, which caused this [condition](https://github.com/department-of-veterans-affairs/vets-website/blob/9fad755e49b0a3d8d271ea0dc710f17ef2426ef5/src/platform/utilities/environment/index.js#L32) to take effect instead. So in theory we could have been using these environment variables locally, but it seems like needless complexity.

`process.env.BASE_URL` and `process.env.API_URL` are never processed anywhere in the Webpack or Metalsmith code.

src/platform/utilities/environment/index.js
```diff
@@ -1,43 +1,80 @@
...
-  vagovprod: {
-    API_URL: 'https://api.va.gov',
+/*
+ * Configuration for the current environment.
+ * @module platform/utilities/environment
+ */
+
+import ENVIRONMENTS from '../../../site/constants/environments';
+
+// eslint-disable-next-line no-undef
```
jbalboni: Could we put a note here about why you're doing this?

src/platform/utilities/environment/index.js
```diff
@@ -1,43 +1,80 @@
...
+
+  /** The address of the FE website configured for this environment. */
+  BASE_URL: environment.BASE_URL,
+
+  /** The address of the API configured for this environment. */
+  API_URL: environment.API_URL,
+
+  /** Determines whether the current environment is a production environment. */
+  isProduction() {
+    return this.BUILDTYPE === ENVIRONMENTS.VAGOVPROD;
```
jbalboni: Instead of using `this`, could we do `environment.BUILDTYPE`? I know it's more repetitive, but it's a bit safer because if someone destructures `environment` and pulls a function out, it'll still work correctly, whereas this version won't.

src/platform/utilities/environment/stagingDomains.js
jbalboni: This should be fine now. We were trying to skip domain replacement in vets.gov, but do it in preview.va.gov. We don't need to do any replacement in production now.

jbalboni (approved):
Couple minor comments, but LGTM

src/platform/utilities/environment/index.js
ncksllvn: Makes perfect sense! I was using `this` only because of this test, https://github.com/department-of-veterans-affairs/vets-website/pull/9162/files#diff-c2e1d1dcef76f34ec88b742ad38d04ffR317, which changes the `BUILDTYPE` property to test in different environments, but I think I can visit that test and see if it can be spruced up.

src/platform/forms/tests/save-in-progress/RoutedSavableApp.unit.spec.jsx
```diff
@@ -432,6 +421,5 @@ describe('Schemaform <RoutedSavableApp>', () => {
 
     expect(fetchInProgressForm.called).to.be.false;
     expect(router.replace.calledWith('/first-in-form-page')).to.be.true;
-    __BUILDTYPE__ = buildType;
```
ncksllvn: These `BUILDTYPE` mutations were no longer necessary, because the module being tested only has special behavior on the `LOCALHOST` environment, and unit tests run under `VAGOVDEV`.

PR #9174, opened 2018-11-29
Disable Foresee during e2e tests and harden specific tests

jbalboni:
## Description
We've been seeing a lot more e2e test failures recently. This appears to be because the Foresee modal is popping up randomly in the middle of some tests. You can see this by running the tests locally with `vagovprod`.

The fix for this I chose is to create a `openUrl` wrapper for the normal `url` command. I did this because the script is triggered just being being in a script tag, so we can't conditionally turn it off at runtime. And we don't have a separate build type for e2e tests. I could have implemented this as a separate command and added to to every test, but that seemed more tedious than this approach.

## Testing done
Ran locally

## Acceptance criteria
- [x] Tests pass without failures (check the logs to look for failures that happened but went away after a retry)

## Definition of done


src/platform/testing/e2e/nightwatch-commands/openUrl.js
```diff
@@ -0,0 +1,14 @@
+exports.command = function openUrl(url, disableForesee = true) {
+  this.url(url);
+
+  if (disableForesee) {
+    this.execute(() => {
+      const css = document.createElement('style');
+      css.type = 'text/css';
+      css.innerHTML = '.__acs { display: none !important; }';
```
jbalboni: Foresee's API has a few methods of suppressing the modal, but none of them worked reliably for me. This does, though.

cvalarida: Clever.

cvalarida (approved):
Looking forward to less flaky e2e tests! :tada:

ncksllvn (approved):
Well done 🔥

PR #9176, opened 2018-11-29
All claims submit 500 mg

markgreenburg:
## Description
Fixes several issues that are preventing successful submission of all claims forms.

## Testing done
- [x] Local testing
- [x] Unit tests

## Acceptance criteria
- [x] All Claims form successfully submits (receives 200 response from vets-api) under minimal data case.
- [x] Confirmation page renders
- [x] Presubmit transformer is unit tested

## Definition of done


src/applications/disability-benefits/all-claims/tests/utils.unit.spec.jsx
```diff
@@ -482,6 +488,235 @@ describe('526 helpers', () => {
     });
   });
 });
+describe('transform', () => {
+  const transformedMinimalData = {
```
markgreenburg: @SMLuthi is going to take a look at these to make sure they pass validation.

src/applications/disability-benefits/all-claims/tests/schema/minimal-test.json
```diff
@@ -16,47 +16,24 @@
       "zipCode": "19234"
     },
     "view:contactInfoDescription": {},
-    "view:vaMedicalRecordsIntro": {},
-    "vaTreatmentFacilities": [
```
cvalarida: Should this be removed? Going through the form flow, they have to choose _some_ type of evidence, right?

src/applications/disability-benefits/all-claims/tests/schema/minimal-test.json
```diff
@@ -74,62 +51,8 @@
...
-        "ratedDisabilityId": "0",
-        "ratingDecisionId": "63655",
-        "diagnosticCode": 5238,
-        "decisionCode": "SVCCONNCTED",
-        "decisionText": "Service Connected",
-        "ratingPercentage": 100,
-        "disabilityActionType": "INCREASE"
-      },
-      {
-        "name": "Second Condition",
```
cvalarida: Good call. I just deleted a comment I wrote thinking this was the maximal test. :stuck_out_tongue:

src/applications/disability-benefits/all-claims/tests/schema/minimal-test.json
```diff
@@ -16,47 +16,24 @@
...
-      "view:selectableEvidenceTypes": {
-        "view:hasVAMedicalRecords": true
-      },
-      "view:evidenceTypeHelp": {}
-    },
     "view:powStatus": false,
-    "view:newDisabilities": true,
-    "newDisabilities": [
+    "view:unemployabilityStatus": false,
+    "view:newDisabilities": false,
```
cvalarida: :ok_hand:

src/applications/disability-benefits/all-claims/tests/utils.unit.spec.jsx
cvalarida: Like validation in the schema?

src/applications/disability-benefits/all-claims/tests/utils.unit.spec.jsx
```diff
@@ -482,6 +488,235 @@ describe('526 helpers', () => {
     });
   });
 });
+describe('transform', () => {
```
cvalarida: I went back and forth trying to figure out if this was solving the same problem as our `schema.unit.test.js`. Where that one checks if the submission passes the schema validations, this one checks that it doesn't, for example, remove anything it shouldn't remove, which the schema validation wouldn't necessarily catch (if the property wasn't required).

Good thinking. I like it. :+1:

Might be a pain to maintain, but we'll see, eh?

src/applications/disability-benefits/all-claims/tests/utils.unit.spec.jsx
```diff
@@ -482,6 +488,235 @@ describe('526 helpers', () => {
...
+              to: '1901-01-01',
+            },
+          },
+        ],
+      },
+      standardClaim: false,
+    },
+  };
+
+  const transformedMaximalData = {
```
cvalarida: This file is already getting pretty long. What do you think about putting these expected outputs into their own file? It'd save 220+ LoC in here.

src/applications/disability-benefits/all-claims/utils.jsx
```diff
@@ -190,6 +190,46 @@ export function transformProviderFacilities(providerFacilities) {
   }));
 }
 
+/**
+ * This is mostly copied from us-forms' own stringifyFormReplacer, but with
+ * the incomplete / empty address check removed, since we don't need this
+ * for any of the 3 addresses (mailing, forwarding, treatment facility) in our
+ * form. Leaving it in breaks treatment facility addresses because by design
+ * they don't have street / line 1 addresses, so would get incorrectly filtered
+ * out. Trivia: this check is also gone in the latest us-forms replacer.
```
cvalarida: Trivia. :joy:

src/applications/disability-benefits/all-claims/utils.jsx
```diff
@@ -190,6 +190,46 @@ export function transformProviderFacilities(providerFacilities) {
...
+ */
+export function customReplacer(key, value) {
+  // clean up empty objects, which we have no reason to send
+  if (typeof value === 'object') {
```
cvalarida: So this is a shallow checker. If we have something like
```js
{
  parent: {
    child: {}
  }
}
```
what we'd be left with at the end would be
```js
{
  parent: {}
}
```
To solve this, I wrote a `removeDeeplyEmptyObjects()` which we now use in the `transform` function. We could pull that out of there to put it in here in place of this shallow remover.

src/applications/disability-benefits/all-claims/utils.jsx
```diff
@@ -264,9 +304,7 @@ export function transform(formConfig, form) {
     delete clonedData.providerFacility;
   }
 
-  return JSON.stringify({
-    form526: JSON.stringify(clonedData),
-  });
+  return JSON.stringify({ form526: clonedData });
```
cvalarida: We talked about this in Slack, but just for posterity, are we confident that `clonedData` shouldn't be stringified?

@SMLuthi, can you confirm that `form526` should be a regular ol' object and not a JSON string that needs to be parsed?

cvalarida (approved):
Lots of comments, but only a couple questions. Overall, it's probably good to ship as-is, but I'd like confirmation on a couple things (whether we should remove the `vaTreatmentFacilities` from the `minimal-test.json` and if we shouldn't stringify `form526`'s value upon submit).

src/applications/disability-benefits/all-claims/tests/utils.unit.spec.jsx
markgreenburg: Yeah, on the vets-api side.

src/applications/disability-benefits/all-claims/tests/schema/minimal-test.json
markgreenburg: Nope, they can elect to choose no evidence.

src/applications/disability-benefits/all-claims/tests/utils.unit.spec.jsx
markgreenburg: Fair enough.

src/applications/disability-benefits/all-claims/utils.jsx
markgreenburg: I like that idea generally, but am concerned with deviating from the common replacer more than necessary. I think the helper you've got might be a nice addition to our version of `us-forms-system`, since it would seem to be a good idea across all forms. Thoughts?

src/applications/disability-benefits/all-claims/utils.jsx
markgreenburg: Well, works locally at least. 🤷‍♂️

PR #9042, opened 2018-11-06
Replace headers with session cookie for authentication.

U-DON:
## Description
Authentication with the cookie has been preliminarily tested on Dev and Staging. It was rolled back because we wanted more exhaustive testing, and to avoid any risk of impacting the VA.gov. This PR is to re-introduce the changes to support auth by cookie again.

## Acceptance criteria
- [ ]

## Definition of done
- [ ] Events are logged appropriately
- [ ] Documentation has been updated, if applicable
- [ ] A link has been provided to the originating GitHub issue (or connected to it via ZenHub)
- [ ] No sensitive information (i.e. PII/credentials/internal URLs/etc.) is captured in logging, hardcoded, or specs

PR #9179, opened 2018-11-30
Revert "Replace headers with session cookie for authentication. (#9042)"

U-DON:
Rolling back session cookie due to caching problems. Old FE code was still expecting the token in the query param. Need to figure out why old code was being served.


saneshark (approved):
ship it

PR #9183, opened 2018-12-01
Fix issue with december wrapping to jan

ncksllvn:
## Description
This fixes an issue with a unit tests -

```
Validations unit tests isValidPartialMonthYearInPast should validate month and year that is current – should validate month and year that is current
<1s
Stacktrace
AssertionError: expected false to be true
at Context.<anonymous> (src/platform/forms/tests/validations.unit.spec.js:482:7)
```

This was failing because we have a date function expecting the month expressed as an integer indexed starting at 1. The unit test was adding 1 to account for this index change, but we are now in December, so adding 1 to the current month circles back around to January. This broke the test.

## Acceptance criteria
- [x] We can deploy again

## Definition of done

### DIFF
--- a/src/platform/forms/tests/validations.unit.spec.js
+++ b/src/platform/forms/tests/validations.unit.spec.js
@@ -479,12 +479,11 @@ describe('Validations unit tests', () => {
       expect(isValidPartialMonthYearInPast('2', '2001')).to.be.true;
     });
     it('should validate month and year that is current', () => {
+      const currentMonthIndexedAtOne = moment().month() + 1;
+
       expect(
         isValidPartialMonthYearInPast(
-          moment()
-            .add(1, 'month')
-            .month()
-            .toString(),
+          currentMonthIndexedAtOne.toString(),
           moment()
             .year()
             .toString(),


src/platform/forms/tests/validations.unit.spec.js
```diff
@@ -479,12 +479,11 @@ describe('Validations unit tests', () => {
       expect(isValidPartialMonthYearInPast('2', '2001')).to.be.true;
     });
     it('should validate month and year that is current', () => {
+      const currentMonthIndexedAtOne = moment().month() + 1;
```
bshyong: Is possible for this to become > 12? If so, does it make sense to have values > 12 here? Or does the Valid Partial Month year automagically take into account month values > 12?

ncksllvn: Good question, but I don't know. My impression was that indexing starts at 1 for that helper.

ncksllvn: Based on how it subtracts a `1` from the month before passing it to the moment helper, I don't think it allows values greater than 12.

```
month: month ? parseInt(month, 10) - 1 : null,
```

bshyong (approved):
Great

PR #9184, opened 2018-12-01
[FacilityLocator] Defect Remediation 2: Electric Boogaloo

r-mc2:
## Description
Additional fixes for bugs found during testing and to go on top of the previous #9159 PR.

- Fixes Provider Service selection not being included with search requests to the API
- Fixes 'All Locations' searches not including the address needed to locate Providers
- Fixes Provider Service not being sent as the ID of the service selected
- Fixes some wonky-ness in how the lifecycle methods were operating in VAMap
- Adds some notes to VAMap for future reference/research/help
- Adds some new functions to assist in object comparisons

### Notes
- We are being instructed to try and get the Provider Locator changes up and live in Prod on Monday, so the more ASAP this can be looked at 🙏 
- 508 changes/updates based on Trevor's review will be in a separate PR as these fixes were more breaking/pressing

(A bit of an aside, but I could certainly use some help in understanding some of the design choices that went into how the logic flows through VAMap. In its current state there are issues with logic firing off multiple times when it should only happen once. Some [[notes on logic flow](https://gist.github.com/r-mc2/27e309840cda3d8a37735f6c2025bf08#order-of-operations)] I took while working through the debugger last week.)

(@'ing people for visibility @ehuntdsva @rtluu @jbalboni @U-DON @bshyong @erikphansen )


brettdt:
@

src/applications/facility-locator/utils/helpers.js
```diff
@@ -80,6 +80,42 @@ export const areGeocodeEqual = (pos1, pos2) => {
...
+ * Compares two geographic bounding boxes to determine if they are equal.
+ *
+ * A bounding box is expected to be of the shape
+ *   [lat1, long1, lat2, long2]
+ *
+ * @param {number[]} box1 The first bounding box's coords
+ * @param {number[]} box2 The second bounding box's coords
+ */
+export const areBoundsEqual = (box1, box2) => {
+  if (!box1 || !box2 || box1.length !== 4 || box2.length !== 4) {
```
jbalboni: What's the `4` mean here?

jbalboni (approved):
I had some minor questions, but otherwise this looks fine.

src/applications/facility-locator/utils/helpers.js
bshyong: It looks like the length 4 is checking that box1 and box2 have the expected 4 coordinates in the bounding box; if not 4 the coordinates are invalid and the method returns false

r-mc2: Yep, exactly as Ben said

src/applications/facility-locator/components/SearchControls.jsx
```diff
@@ -18,8 +18,8 @@ class SearchControls extends Component {
   // eslint-disable-next-line prettier/prettier
   handleFacilityTypeChange = (option) => {
     // eslint-disable-next-line prettier/prettier
-    const facilityType = (option === LocationType.ALL) ? null : option;
-    this.props.onChange({ facilityType, serviceType: null });
+    // const facilityType = (option === LocationType.ALL) ? null : option;
```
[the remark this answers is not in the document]
r-mc2: Yep, thought I already had. Will be gone!

r-mc2:
Is anyone able to understand what failed the unit test step of the pipeline?
I'm really only seeing Warnings and a lot of `target Element not found`.
Can't figure out if it's related to these changes I made or unrelated to this PR specifically.

src/applications/facility-locator/containers/VAMap.jsx
```diff
@@ -108,6 +111,38 @@ class VAMap extends Component {
...
+          (could have used inProgress but it gets tripped by other Actions)
+        - 
+      
+      The boundary checking of the current code below doesn't actually work.
+      Array equality isn't something that should be done with the operator,
+      and using the new method below causes `searchWithBounds` to never fire.
+      Goes in line with needing clearer ideas of what state of the app ==
+      when to fire off a new search, zoom out, or even just do nothing.
+
+      Near as I can tell this.zoomOut.cancel() does nothing.
```
cvalarida: Should the points in these notes be made tickets instead? It's probably totally appropriate to leave comments pointing to the tickets, but I think these won't be addressed until they're put into tickets to be prioritized.

src/applications/facility-locator/containers/VAMap.jsx
```diff
@@ -127,39 +162,40 @@ class VAMap extends Component {
...
-    const { currentQuery } = prevProps;
-    const newQuery = this.props.currentQuery;
+    const { currentQuery: prevQuery } = prevProps;
+    const updatedQuery = this.props.currentQuery;
 
     /* eslint-disable prettier/prettier */
-    // A long block of logic AND'ing is easier to read when seen wrapped in a "block"
-    // Yes, it's equivalent without the paren, but understanding is faster with 'em
-    const shouldUpdateSearchQuery = (
+    const shouldZoomOut = ( // ToTriggerNewSearch
```
cvalarida: I'm not familiar with how the app works. If this is `true`, does it always result in zooming out, or will it have some other effect given some edge case somewhere?

src/applications/facility-locator/containers/VAMap.jsx
```diff
@@ -169,6 +205,25 @@ class VAMap extends Component {
...
+   * Helper method to compare search parameters between
+   * component updates/renders.
+   *
+   * Currently compares search string, location type,
+   * service type, and map bounding box.
+   *
+   * @param {object} previous Previous component props
+   * @param {object} current Current componet props
+   */
+  didParamsChange = (previous, current) => {
```
cvalarida: Looks like the only time this is used is commented out for now. What do you think of also commenting this helper function as well?

Not sure how much it matters either way, but if that commented-out usage is removed without being implemented, then we're left with a function that doesn't ever get called, y'know? :man_shrugging:

src/applications/facility-locator/utils/helpers.js
cvalarida: That might be good, but I think it might be more helpful later if there were a comment saying why it needed to be 4 exactly. Something as simple as `// Bounding boxes need 4 coordinates to be valid` should do the trick.

src/applications/facility-locator/utils/helpers.js
```diff
@@ -80,6 +80,42 @@ export const areGeocodeEqual = (pos1, pos2) => {
...
+/**
+ * Compares two geographic bounding boxes to determine if they are equal.
+ *
+ * A bounding box is expected to be of the shape
+ *   [lat1, long1, lat2, long2]
+ *
+ * @param {number[]} box1 The first bounding box's coords
+ * @param {number[]} box2 The second bounding box's coords
+ */
+export const areBoundsEqual = (box1, box2) => {
```
cvalarida: Could we get a unit test for this too?

It looks like it's not being used right now, but as long as it's "live" (not commented out), it should probably be tested.

src/applications/facility-locator/utils/helpers.js
bshyong: A comment about how bounding boxes need 4 coordinates to be considered valid probably will suffice; IMO I don't see a need to use a variable because we aren't doing anything else with that number

bshyong (approved):

cvalarida (commented):
I think my biggest question is what the plan is for the `Future testing to fix excessive searches being fired` bit. There are bits of code that aren't commented out but only used in that commented code.

To avoid putting the code base in sort of an awkward place with regard to unused code, what do you think of either uncommenting the excessive searches fix or commenting out all the code that's only used by it?

src/applications/facility-locator/containers/VAMap.jsx
```diff
@@ -108,6 +111,38 @@ class VAMap extends Component {
...
+      whether or not the data in the fields that were updated represents a valid
+      state for triggering a new search is ambiguous at best nor should we simply
+      fire off a new search each time something changes in Redux.
+
+      New Flag Ideas:
+        - geocodeInProgress
+        - revGeocodeInProgress - should be a separate flag as both operations happen
+        - searchRequested - To track that the user clicked the search button
+          (could have used inProgress but it gets tripped by other Actions)
+        -
```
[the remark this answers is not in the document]
r-mc2: Ah, personal habit. Leave the next bullet open so I don't have to type it once I have my next item to add. So, yea, intentionally left blank.

src/applications/facility-locator/containers/VAMap.jsx
r-mc2: Yes, there should probably be some bugs/tickets opened for some of the behavior of the app, although I've been having trouble isolating exactly where it's coming from so I've got these notes in for now + that linked gist with my notes on a debugging session.
Any help with it would be appreciated. Would it be faster to open new bugs to do that?

src/applications/facility-locator/utils/helpers.js
r-mc2: Sounds good, I'll be sure to add the comment in 👍

src/applications/facility-locator/containers/VAMap.jsx
r-mc2: As far as I can tell, yes. It's the only thing that really happens in this lifecycle method.

It either fires `this.zoomOut()` or does the manual adjustment of the map's bounding box to effectively zoom out if on mobile resolutions.

There are a couple locations where some checks are done to see if zooming out should be canceled, however, as noted in the block above canceling appears to have no effect nor is anything currently done to undo the manual "zoom out" for mobile (lines 159 & 195).

src/applications/facility-locator/utils/helpers.js
r-mc2: Yep, should be easy enough to add that in. Pretty sure `helpers.js` already has a unit test suite.

src/applications/facility-locator/containers/VAMap.jsx
r-mc2: I can comment it out, sure. I left it uncommented as it's still a useful function and does work. Just not used yet based on the commented out section you mention.

r-mc2:
As to @cvalarida your questions about the commented out "Future fix" stuff. I know that's where I want to be going with that code, however, as the app is now, that code effectively causes searches to never be fired due to some of the other issues I pointed out above in the comment block.

Given the time crunch we're in and as side-tracked as I was getting with it, I figured it was better/safer to just comment out that bit of logic and push out the fixes for the bugs I know I had already fixed. The app isn't *perfect* atm, but searches will fire and results will come back.

(In retrospect, maybe should have pulled it off into a stash 🤔)

annaswims:
We're seeing a high number of error in sentry that look like they might be related to this pr. http://sentry.vetsgov-internal/vets-gov/platform-api-production/issues/51638/

PR #9185, opened 2018-12-02
Update FTUX modal for user dashboard

erikphansen:
Also disable the auto-redirect to the preferences/ route

image

### DIFF
--- a/src/platform/site-wide/announcements/config/index.js
+++ b/src/platform/site-wide/announcements/config/index.js
@@ -1,10 +1,9 @@
-import DashboardIntro from '../components/DashboardIntro';
+import RecommendationsIntro from '../components/RecommendationsIntro';
 import Profile360Intro from '../components/Profile360Intro';
 import PersonalizationBanner from '../components/PersonalizationBanner';
 import ClaimIncreaseBanner from '../components/ClaimIncreaseBanner';
 import VAPlusVetsModal from '../components/VAPlusVetsModal';
 import WelcomeToNewVAModal from '../components/WelcomeToNewVAModal';
-import isBrandConsolidationEnabled from '../../../brand-consolidation/feature-flag';
 
 const config = {
   announcements: [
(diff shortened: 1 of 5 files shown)


Samara-Strauss:
Screenshot above looks good! I'm not seeing it on staging yet — is it safe to assume that is expected based on where you guys are in build?

erikphansen:
@Samara-Strauss I totally change the label on the DISMISS button since the style of it changed and it's much larger than a simple text-only button. Also, the width of the modal is hard coded across all of these alerts so that's different as well.

Samara-Strauss:
All sounds good. Let me know when it's on staging!

erikphansen:
I'm apparently slow to wake up this morning. It won't go to staging until it's merged into `master`.

Samara-Strauss:
> I'm apparently slow to wake up this morning. It won't go to staging until it's merged into master.

Got it — is there an automatic update when that happens? If not, just let me know when this is up.

src/site/assets/img/dashboard-form.svg
```diff
@@ -0,0 +1,36 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg width="69px" height="80px" viewBox="0 0 69 80" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <!-- Generator: Sketch 51.3 (57544) - http://www.bohemiancoding.com/sketch -->
+    <title>Form illustration</title>
+    <desc>Created with Sketch.</desc>
+    <defs></defs>
+    <g id="Final-MVP-Mockups" stroke="none" stroke-width="1" fill="none" fill-rule="evenodd">
```
cehsu: Should we update this `id`? Maybe something like, Recommendations-Intro?

src/platform/site-wide/announcements/config/index.js
```diff
@@ -1,10 +1,9 @@
-import DashboardIntro from '../components/DashboardIntro';
+import RecommendationsIntro from '../components/RecommendationsIntro';
```
cehsu: Should this be named something "preferences"-related?

cehsu (approved):
This looks great, nice work! Should we add a unit test for the modal?

src/platform/site-wide/announcements/config/index.js
erikphansen: Preferences seems pretty general for a side-wide modal. But maybe something like FindVABenefitsIntro would be better.

erikphansen:
@cehsu It doesn't look like we have tests for these site-wide modals since they are really just setting up configuration to be used by our announcement framework.

PR #9193, opened 2018-12-04
All claims payment info -- Part 1

cvalarida:
## Description
I've decided to break this one up into two parts. In this episode, you'll see changes to the `ReviewCardField` that allows the user to cancel an update. Technically, they're not so much _cancelling_ as they are _resetting_ their changes. Staging the changes to commit them upon "save" was a lot more work.

The consideration here is that if a user starts updating then navigates away, their information _will_ be updated even though they didn't hit "save." This may be...not ideal. They can't, however, continue if they didn't enter valid information. (If they try to go forward, they'll get a validation error, if they go backward and _then_ forward, the `ReviewCardField` will open up in edit mode and not let them continue past the page.) I didn't think it was a big deal, but @goldenmeanie let me know if that's not the case.

The only UX problem I could think of is if they:
1) Hit the "New" button
2) Edited the information
3) Went back
4) Went forward to the payment information screen
5) They no longer have a way to get their old information back
   - If what they just entered was valid, they'd have to hit "New account" and cancelling would clear out the new edits but keep the previous changes
   - If the data was invalid they can't cancel because the `ReviewCardWidget` was initially given invalid information

## Testing done
Unit & manual.

### Started with no information
image

### Saved some information
image

**Note:** This doesn't have the warning that this is a different account than what's currently on file. That'll come in the next PR.

Also note the lack of an edit button up top. We now only have `New account` down below.

### Adding a new account
image

Note the addition of a cancel button.

## Acceptance criteria
- [x] The payment info screen uses a `ReviewCardField` with `volatileData: true`
- [x] A `ReviewCardField` with `volatileData: true`:
  - [x] Removes the edit button
  - [x] Adds a `New` button
  - [x] Handles "cancelling" updates

## Definition of done


src/applications/disability-benefits/all-claims/pages/paymentInformation.js
```diff
@@ -9,45 +11,64 @@ const {
...
+      'We’re currently paying your compensation to this account',
+    'ui:field': ReviewCardField,
     'ui:options': {
-      widgetClassNames: 'va-select-medium-large',
+      viewComponent: PaymentView,
+      reviewTitle: 'Payment information',
+      editTitle: 'Add new bank account',
+      itemName: 'account',
+      startInEdit: formData =>
+        Object.keys(formData).every(key => !formData[key]),
```
markgreenburg: This seems like it will only start in edit mode if none of the fields are filled, is that the intent? What about incomplete (invalid) account info coming from prefill?

src/applications/disability-benefits/all-claims/pages/paymentInformation.js
```diff
@@ -9,45 +11,64 @@ const {
...
-  bankName: {
-    'ui:title': 'Bank name',
-    'ui:required': bankFieldsHaveInput,
   },
 };
 
 export const schema = {
   type: 'object',
+  // I think uiRequired might be what we're looking for?
+  // required: ??
```
markgreenburg: Is the intent to allow users to submit a 526 without adding banking info, assuming they don't have any populated via prefill? Or should banking info always be required?

src/applications/disability-benefits/all-claims/pages/paymentInformation.js
```diff
@@ -9,45 +11,64 @@ const {
...
-  bankAccountNumber: {
-    'ui:title': 'Account number',
-    'ui:options': {
-      widgetClassNames: 'va-input-medium-large',
+    bankAccountType: {
+      'ui:title': 'Account type',
+      'ui:options': {
+        widgetClassNames: 'va-select-medium-large',
+      },
+      'ui:required': bankFieldsHaveInput,
```
markgreenburg: To make this work properly, `bankFieldsHaveInput` will need to be updated with the new, nested path for the bank info (`formData[view:bankAccount]`)

src/applications/disability-benefits/all-claims/components/ReviewCardField.jsx
```diff
@@ -142,14 +174,23 @@ export default class ReviewCardField extends React.Component {
             className="usa-button-primary update-button"
             onClick={this.update}
           >
-            Done
+            {volatileData ? 'Save' : 'Done'}
```
markgreenburg: I was thinking this might be better the other way around, because of the interaction you cited in the PR description wherein the data is updated without a user needing to actually hit 'save'. If that's the case, 'save' might be misleading since the data's already saved, and 'done' might be a better fit. Thoughts?

src/applications/disability-benefits/all-claims/pages/paymentInformation.js
cvalarida: In that case, it'll start in edit mode because of the logic that was already in the `ReviewCardField`.

src/applications/disability-benefits/all-claims/pages/paymentInformation.js
cvalarida: Oh, hey. This must be why the above happened. :bulb:

Good catch.

src/applications/disability-benefits/all-claims/components/ReviewCardField.jsx
cvalarida: Yeah, I mean in either case (`volatileData` or no), hitting the button is more of a "done."

@goldenmeanie What do you think? (Context for this can be found in the original PR comment up top.)

markgreenburg (commented):
LGTM, except a little thing with the `ui:required` properties not working as they are.

markgreenburg (approved):
Lookin good!

PR #9194, opened 2018-12-04
fix for not displaying email on desktop

accbjt:
## Description
[Mobile] [Signed in] Global account nav: Long names and emails go off screen. This is a fix for some bugs that came up.

## Testing done
Manual Tests

## Screenshots
screen shot 2018-12-04 at 11 30 26 pm

## Acceptance criteria
- [ ] Desktop should show email if user does not have a first name
- [ ] Name that have letters like `g` or `y` should not be cut off on the bottm
- [ ] Mobile should still show My Account if a user does not have a first name

## Definition of done


src/platform/site-wide/user-nav/selectors.js
```diff
@@ -4,14 +4,21 @@ import { startCase, toLower } from 'lodash';
...
+  (name, email, sessionFirstName) => {
     if (name.first || sessionFirstName) {
       return startCase(toLower(name.first || sessionFirstName));
     }
 
+    if (desktopMediaQuery.matches) {
+      return email;
+    }
+
     return 'My Account';
```
ncksllvn: I think the CSS responsive utility classes may be a better option here, because with the CSS approach you'll be able to resize the browser window and see the username field respond automatically, whereas to do that in JS, you'll have to set an event listener for a consistent display.

I don't know the responsive utility classes off the top of my head unfortunately, but assuming this function can return JSX without any issues, the markup should be something like -

```jsx
return (
  <div>
    <span className="desktop-only">{email}</span>
    <span className="desktop-hidden">My account</span>
  </div>
);
```

jbalboni: Yeah, I agree with this.

accbjt: @jbalboni @ncksllvn I didn't even know we had this available. I will make the updates.

accbjt: @ncksllvn @jbalboni I made the suggested updates to this. It works a lot better then using JavaScript. Thanks for the suggestion.

src/platform/site-wide/user-nav/selectors.js
```diff
@@ -13,6 +14,16 @@ export const selectUserGreeting = createSelector(
       return startCase(toLower(name.first || sessionFirstName));
     }
```
accbjt: @ncksllvn Made the changes you suggested. Heer is my solution for it. I had to use a span because it worked correctly with the foundation responsive utility classes, without having to fiddle with the css too much.

src/platform/site-wide/user-nav/selectors.js
```diff
@@ -13,6 +14,16 @@ export const selectUserGreeting = createSelector(
...
+      <span key="show-for-small-only" className="show-for-small-only">
+        My Account
+      </span>,
+      <span
+        key="show-for-medium-up"
+        className="user-dropdown-email show-for-medium-up"
+      >
+        {email}
+      </span>,
+    ];
```
accbjt: @ncksllvn Made the changes you suggested. Heer is my solution for it. I had to use a span because it worked correctly with the foundation responsive utility classes, without having to fiddle with the css too much.

accbjt: @ncksllvn @jbalboni can I gets some eye's on this so I can close this one out.

jbalboni:
@accbjt I'm seeing this happen with long emails now (doesn't happen in staging):

screen shot 2019-01-08 at 9 57 20 am

src/platform/site-wide/sass/merger.scss
```diff
@@ -187,13 +187,22 @@ $medium-phone-screen: 375px;
...
       @include media($medium-screen) {
         span {
-          display: inline-block;
           padding-left: 0;
+          max-width: 300px;
+          height: 20px;
+
+          &.user-dropdown-email {
+            margin-top: 5px;
+          }
```
ncksllvn: It seems like we just changed the text based on screen size - how come we needed some new style rules?

accbjt: @ncksllvn because we are now using the mobile-responsive utility, it sets the `display: initial !important`. This causes alignment issues with the top user-nav area in medium-up. This compensate for that misalignment.

accbjt:
@jbalboni @ncksllvn This should fix the issue with the long name on desktop and I also fixed the tests.

accbjt:
@jbalboni I made those updates. Let me know if it looks good.

jbalboni:
@accbjt I think we had this problem before, the text for the user name sits higher than the rest of that row at small screen sizes:

screen shot 2019-01-10 at 9 53 33 am

I'm going to approve this because the email sizing looks good, but can you look into that for a little bit and see if it's a quick fix?

jbalboni (approved):

accbjt:
@jbalboni Yes it was a quick fix. I just needed to update the margin. The span was causing the misalignment.

ncksllvn:
@U-DON @saneshark Making sure you all have eyes on this and are aware of the change.

PR #9198, opened 2018-12-04
short circuit debounced fetch

rianfowler:
## Description
- unit tests were failing intermittently because of a debounce behavior in the auto suggest component
- this fix removes the function that calls fetch from the config before this component is rendered for testing by the `<DefinitionTester />`

## Testing done
ran the unit tests locally ~25 times- usually have at least a couple of failures every 10 runs. 

## Screenshots
Expected behavior: 
screen shot 2018-12-03 at 16 26 40
Intermittent behavior: 
screen shot 2018-12-03 at 16 26 45
Wasn't able to figure out why the debouncing functions were called twice. 

## Acceptance criteria
- [ ]

## Definition of done


src/applications/disability-benefits/526EZ/tests/config/vaFacilities.unit.spec.jsx
```diff
@@ -28,6 +28,9 @@ describe('Disability benefits 526EZ VA facility', () => {
     uiSchema,
     arrayPath,
   } = formConfig.chapters.supportingEvidence.pages.vaFacilities;
+  uiSchema.disabilities.items.treatments.items.treatmentCenterName[
+    'ui:options'
+  ].getOptions = () => Promise.resolve([]);
```
rianfowler: I'm going to try to update the `AutoSuggest` component to _not_ debounce this call and maybe do an RFC that recommends how debouncing should be handled. TL;DR debouncing should be wrapped into the call and not something a component does. 

Even with that change, I think it would be necessary to alter this `getOptions` before testing it. Not sure the best way to handle this- if it's typical to edit the form config this way or if there's an easier way we could disable debouncing behavior. This fix in particular is pretty imperative.

jbalboni: I can't think of a better way off the top of my head. We probably want to unit test the `getOptions` function separately if we're overriding it, though.

jbalboni (approved):
One comment, but otherwise looks good.

src/applications/disability-benefits/526EZ/tests/config/vaFacilities.unit.spec.jsx
rianfowler: Looks like we have tests for `queryForFacilities`. `getOptions` is what the second argument of `autoSuggestUiSchema` get assigned to. https://github.com/department-of-veterans-affairs/vets-website/blob/daa11a751bf30802e3924b0487bd4346542fc726/src/applications/disability-benefits/all-claims/tests/utils.unit.spec.jsx#L204

PR #9203, opened 2018-12-04
Reintroduce session cookie authentication

U-DON:
## Description
We previously rolled back due to concerns around caching policies that would have resulted in a negative and confusing user experience. After some effort to reproduce the issue locally, I feel confident that Staging and Production won't experience the caching issues due to the filename hashes for the scripts.

I have also included simultaneous logout for multiple tabs. When one tab logs out, other tabs will also update their logged in states.

## Testing done
Local testing to reproduce the caching problem:
1. Created `vagovdev` build on this branch and renamed the build `dev_cookie`.
2. Created `vagovdev` build on master.
3. Ran the test server for `vagovdev` on localhost:3001 (so now the build from master is being served).
4. Disabled session cookie in the locally running API.
5. Logged in and logged out in localhost:3001.
5. Renamed `vagovdev` to `dev_master`.
6. Renamed `dev_cookie` to `vagovdev` to serve this branch's build.
7. Soft refreshed localhost:3001.
8. Attempt login and observe failure.

This failure is due to cached JS that expects the token query param in order to store in `sessionStorage` and pass in the auth header. The API is sending a cookie in the response, but the redirect URL doesn't include the token in the query param. The workaround in this case is to do a full refresh (Cmd-Shift-R).

## Acceptance criteria
- [x] Authenticated requests should be successful without sending authorization headers.
- [x] Signed in users should be automatically be signed in when opening VA.gov pages in other tabs.

## Definition of done


saneshark (approved):
LGTM, but would prefer a FE more qualified than me take a look at the latest commits.

src/platform/site-wide/user-nav/containers/Main.jsx
```diff
@@ -96,6 +97,15 @@ export class Main extends React.Component {
...
 
+  handleSessionChange = event => {
+    if (!this.props.currentlyLoggedIn) return;
+
+    const { key, newValue } = event;
+    if (!key || (key === 'hasSession' && !newValue)) {
+      window.location.reload();
+    }
+  };
+
```
saneshark: Someone more qualified on the FE should scrutinize this for edge cases and potential logic flaws. 

@cvalarida @ncksllvn if either of you have time could you take a look please?

PR #9207, opened 2018-12-05
unmount all mounts

rianfowler:
## Description
unmounts components on all unit tests that mount
Merging contingent on [RFC](https://github.com/department-of-veterans-affairs/vets.gov-team/pull/15504#pullrequestreview-181859471)

## Testing done
verified locally

## Acceptance criteria
- [ ]

## Definition of done
- [ ] Events are logged appropriately
- [ ] Documentation has been updated, if applicable
- [ ] A link has been provided to the originating GitHub issue (or connected to it via ZenHub)
- [x] No sensitive information (i.e. PII/credentials/internal URLs/etc.) is captured in logging, hardcoded, or specs

PR #9210, opened 2018-12-06
Add eslint plugin to cause error when `mount` or `shallow` is used without `unmount` in `it()` functions

rianfowler:
## Description
POC eslint plugin for [RFC](https://github.com/department-of-veterans-affairs/vets.gov-team/pull/15504#pullrequestreview-181859471)

Planning on merging this in. I was able to run it against this [PR](https://github.com/department-of-veterans-affairs/vets-website/pull/9207) with the errors in the screenshot below. Those errors will be unavoidable because 
- the components do prop validation in the constructor which throws errors- it's not possible to get a reference to the ReactWrapper from `mount` or `shallow` when a component throws an error there and unit test the error 
- the `<PrintPage />` unit testsdisables the lifecycle methods in its tests making `unmounting` unnecessary

This approach is rudimentary but it catches most of our scenarios: mounting in `it` callbacks. I will make an enhancement ticket to address edge cases:

Other scenarios to address:
- mounting in `beforeEach` and unmounting in `it` (prohibit- suggest using `afterEach`)
- mounting in `beforeEach` and unmounting in `afterEach` (error if not in `afterEach`)
- mounting in `describe` (prohibit)
- mounting multiple components at once in `it` (probably can't use simple identifiers to track this)
- assigning result of mount / shallow 

## Testing done
ran against this [PR](https://github.com/department-of-veterans-affairs/vets-website/pull/9207)

## Screenshots
screen shot 2018-12-06 at 10 07 24

## Acceptance criteria
- [ ]

## Definition of done


rianfowler:
Other scenarios to address:
- mounting in `beforeEach` and unmounting in `it` (prohibit- suggest using `afterEach`)
- mounting in `beforeEach` and unmounting in `afterEach` (error if not in `afterEach`)
- mounting in `describe` (prohibit)
- mounting multiple components at once in `it` (probably can't use simple identifiers to track this)
- assigning result of mount / shallow

rianfowler:
Any feedback is helpful- especially on naming and location of the plugin.

script/custom-eslint/enzyme-always-unmount/package.json
```diff
@@ -0,0 +1,11 @@
+{
+	"name": "eslint-plugin-enzyme-always-unmount",
+        "version": "0.0.1",
+	"main": "index.js",
+	"devDependencies": {
+            "eslint": "~2.6.0"
```
rianfowler: should probably make this a peer dependency

script/custom-eslint/enzyme-always-unmount/package.json
```diff
@@ -0,0 +1,11 @@
+{
+	"name": "eslint-plugin-enzyme-always-unmount",
+        "version": "0.0.1",
```
rianfowler: I don't know if versioning really matters on this since we're not publishing it but eslint requires that plugins are imported as modules.

.eslintrc
```diff
@@ -24,6 +24,7 @@
   },
   "rules": {
     // Override airbnb style.
+    "enzyme-always-unmount/no-matching-unmount": 2,
```
rianfowler: need to alphabetize

cehsu (approved):

cehsu:
The linting errors on this PR are a helpful demo.

package.json
```diff
@@ -114,6 +114,7 @@
     "eslint-config-airbnb": "^15.1.0",
     "eslint-config-prettier": "^3.1.0",
     "eslint-import-resolver-babel-module": "^4.0.0",
+    "eslint-plugin-enzyme-always-unmount": "./script/custom-eslint/enzyme-always-unmount",
```
ncksllvn: Is this a thing specific to eslint? It looks cool, but I haven't seen this before.

ncksllvn (approved):

ncksllvn:
@cehsu Yeah they really drive the point - `��� 887 problems (887 errors, 0 warnings)`

All of them - ` 238:21  error  shallow() has no matching ReactWrapper.unmount()  enzyme-always-unmount/no-matching-unmount`

package.json
erikphansen: Looks like it's described here, as the very last option for setting the source of a dependency https://docs.npmjs.com/files/package.json#dependencies

script/custom-eslint/enzyme-always-unmount/package.json
erikphansen: Like you said, this probably doesn't matter too much, but I'd still like to see it start at 1.0.0 as a best practice.

erikphansen (approved):

package.json
ncksllvn: I see local paths, but they're prefixed by `file:`. Maybe I'm reading it wrong 🤷‍♂️

erikphansen: Oh yeah... you're right. Weird... But this seems to work? 🤔

script/custom-eslint/enzyme-always-unmount/package.json
cvalarida: Mostly because I'm just curious: @erikphansen Why 1.0.0 to start?

package.json
cvalarida: :point_up:

rianfowler: yarn did that so maybe it's wrong. 
screen shot 2018-12-06 at 15 02 46

script/custom-eslint/enzyme-always-unmount/package.json
```diff
@@ -0,0 +1,11 @@
+{
+	"name": "eslint-plugin-enzyme-always-unmount",
```
rianfowler: update to better match eslint rule naming conventions: https://eslint.org/docs/developer-guide/working-with-rules-new#rule-naming-conventions

rianfowler (commented):
one more thing

script/custom-eslint/enzyme-always-unmount/index.js
```diff
@@ -0,0 +1,27 @@
+module.exports.rules = {
+  'no-matching-unmount': context => {
+    let shallowMountNode = null;
+    let unmountNode = null;
+    return {
+      "CallExpression[callee.name='it'] > ArrowFunctionExpression Identifier[name=/shallow|mount|unmount/]": node => {
```
cvalarida: Hey, cool! This doesn't match `foo.it()`.

Only improvement I can think of is to use `:function` instead of `ArrowFunctionExpression` since that'll also match `function foo() { ... }`. Not that it really matters _too_ much since we have another linting rule that calls `function` definitions an error. :man_shrugging: 

What this _won't_ do is catch it if the callback is defined outside of the `it()` call. (`it('should do a thing', testFunction)`, for example.) We can probably catch that in a later iteration if we care, though.

cvalarida (approved):

src/applications/disability-benefits/526EZ/tests/components/AsyncDisplayWidget.unit.spec.jsx
```diff
@@ -9,6 +9,7 @@ const viewComponent = () => <div id="view-component">View component</div>;
 
 describe('AsyncDisplayWidget', () => {
   it('should throw an error if ui:options are not present', () => {
+    // eslint-disable-next-line va-enzyme/unmount
```
rianfowler: ignoring some of these

src/applications/disability-benefits/all-claims/tests/components/ReviewCardField.unit.spec.jsx
```diff
@@ -226,19 +235,22 @@ describe('Schemaform: ReviewCardField', () => {
...
       const tree = shallow(<ReviewCardField {...props} />);
       expect(tree.find('.edit-button').text()).to.equal('New Doodad');
+      tree.unmount();
     });
 
     it('should not allow canceling if starting in edit mode', () => {
       const props = set('uiSchema.ui:options.startInEdit', true, defaultProps);
       const tree = shallow(<ReviewCardField {...props} />);
       expect(tree.find('.cancel-button').length).to.equal(0);
+      tree.unmount();
```
rianfowler: will do a final clean up of remaining `unmount` on Monday

src/applications/disability-benefits/all-claims/tests/components/ReviewCardField.unit.spec.jsx
```diff
@@ -56,12 +56,14 @@ describe('Schemaform: ReviewCardField', () => {
 
   it('should throw an error if no viewComponent is found', () => {
     expect(() => {
+      // eslint-disable-next-line enzyme-always-unmount/no-matching-unmount
       shallow(<ReviewCardField {...defaultProps} uiSchema={{}} />);
```
cvalarida: :thinking: Should we ignore this? That is, will this potentially put our testing in the same weird state that we're trying to prevent with this plugin?

If not, we can avoid having to throw an ignore line in here by adding a `VariableDeclaration` to the AST selector so it ignores these calls to `shallow` and `mount` if they're not assigned to a variable.

cvalarida: If it is possible to still get into whatever weird state we're trying to prevent, could we tack on a `.unmount()`?
```js
expect(() => {
  shallow(...).unmount();
}).to.throw(...);
```

rianfowler: I think the linter should catch when components are mounted but not assigned. It's not possible to do in this component because it throws an error in the constructor- enzyme only returns an instance _after_ the constructor and render are called. 

I think it's weird that this component props validation in the `constructor`. I would rather refactor the component because
- props validation should happen on every render- (putting it in the constructor only validates the props when the component is initialized), and
- components should render an error state and leave error handling to non-React js.

rianfowler: The second point is debatable- we don't really have any good directives on error handling. Might do an RFC for this later- I think it's cleaner to separate error handling concern from view rendering and to make views as dumb as possible.

rianfowler: We're unmounting components as a defensive measure. I think in the case it's safe to ignore because- there's no async behavior in this component that might call js later. so there should be no reference to this component after this unit test is complete. 

Ideally we don't ignore and let the linter work but I figured we could refactor the component later. We're doing this blanket `unmount` ing to make it easier to reason though- also to catch problems by async behavior that might be added during maintenance. Since these bugs are so hard to track down, I figured it would be best just to enforce unmounting always even if it's not necessary (it's recommended by enzyme maintainers).

cvalarida: Yeah, we could probably make better use of PropTypes. I had this idea that throwing an error in the constructor with a very specific message when the developer clearly messed up the `formConfig` would be more helpful than just relying on PropTypes. Now that I think about it again, I think they'd accomplish the same thing.

That said, there may be more of that kind of validation that makes sense to happen in the constructor and we should account for that. If the prop comes only from being hard coded, not any user input, and had validation that PropTypes won't catch...

It's pretty niche, though. PropTypes validations can be pretty thorough. :man_shrugging:

bshyong (approved):

src/applications/disability-benefits/all-claims/tests/components/ReviewCardField.unit.spec.jsx
rianfowler: We could add this behavior to a static method so it doesn't need to be instantiated / mounted to be unit tested. Then you could have two unit tests:

- on that mocks the method to validate it's called from the `constructor` and
- one that calls the static method on the definition to verify that it throws an error.

cvalarida: :thinking: Probably getting into the weeds a bit here, but would this be just so testing is easier?

rianfowler: Yes- I also think it's good design to just call methods/functions from life cycle methods instead of having logic in them. Static methods are a good place for helper functions conceptually since they don't make any changes to the instance.

cvalarida: Cool.

Anywho, I think I agree with you. We can ignore these lines and refactor the component later. If we end up writing components where we have to add more ignore lines, it's probably a sign that the component should be written differently.

rianfowler:
Added [ticket](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/15621) to track potential enhancements

PR #9232, opened 2018-12-10
[Facility Locator] 508 Accessibility Review Changes

r-mc2:
## Description
Changes to the Locator based on review feedback from @1Copenut in our recent [508 review](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/15212).

Specific issues targeted/fixed with this PR:
- [#15421](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/15421) - Community Care info button will be more usable if added to source order after Community Care button
- [#15428](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/15428) - Manage focus on the More Information Icon

Specific issues for the AV-CommCare group:
- [#15546](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/15546) - Targets #15421 above
- [#15545](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/15545) - Targets #15428 above

There is one more outstanding item from the review [#15418](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/15418) that is taking a bit more effort and will be included in its own PR once solved.

(@'ing for visibility @ehuntdsva @rtluu )

### DIFF
--- a/src/applications/facility-locator/components/FacilityTypeDropdown.jsx
+++ b/src/applications/facility-locator/components/FacilityTypeDropdown.jsx
@@ -22,11 +22,45 @@ class FacilityTypeDropdown extends Component {
     this.state = {
       showBubble: false,
     };
+
+    // eslint-disable-next-line prettier/prettier
+    this.setCCInfoRef = (element) => {
+      this.ccInfoIcon = element;
+    };
+    this.focusCCInfoIcon = () => {
+      if (this.ccInfoIcon) {
+        if (this.ccInfoCloseBtn) {
+          this.ccInfoCloseBtn.blur();
+        }
+        this.ccInfoIcon.focus();
+      }
+    };
+
+    // eslint-disable-next-line prettier/prettier
+    this.setCCInfoCloseRef = (element) => {
+      this.ccInfoCloseBtn = element;
+    };
+    this.focusCCInfoCloseBtn = () => {
+      if (this.ccInfoCloseBtn) {
+        this.ccInfoIcon.blur();
+        this.ccInfoCloseBtn.focus();
+      }
+    };
+  }
+
+  componentDidUpdate() {
+    // Set focus on the correct element (508)
+    if (this.state.showBubble) {
+      this.focusCCInfoCloseBtn();
+    } else {
+      this.focusCCInfoIcon();
+    }
   }
 
   // eslint-disable-next-line prettier/prettier
   toggleCCInfo = (e) => {
     e.preventDefault();
+
     this.setState({
       showBubble: !this.state.showBubble,
     });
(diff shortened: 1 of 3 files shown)


src/applications/facility-locator/components/FacilityTypeDropdown.jsx
```diff
@@ -22,11 +22,45 @@ class FacilityTypeDropdown extends Component {
...
+        this.ccInfoCloseBtn.focus();
+      }
+    };
+  }
+
+  componentDidUpdate() {
+    // Set focus on the correct element (508)
+    if (this.state.showBubble) {
+      this.focusCCInfoCloseBtn();
+    } else {
```
1Copenut: I've been doing some testing locally, and think we can remove this `else` logic. It's causing the "i" button to be focused on first render, which isn't quite what we want. It should be re-focused when the close button is clicked.

src/applications/facility-locator/components/FacilityTypeDropdown.jsx
```diff
@@ -22,11 +22,45 @@ class FacilityTypeDropdown extends Component {
...
+    }
   }
 
   // eslint-disable-next-line prettier/prettier
   toggleCCInfo = (e) => {
     e.preventDefault();
+
     this.setState({
       showBubble: !this.state.showBubble,
     });
```
1Copenut: I tried out adding the following function call, and it focused the "i" button again as expected:

```javascript
  // Refocus the "i" button
  this.focusCCInfoIcon();
```

src/applications/facility-locator/components/FacilityTypeDropdown.jsx
```diff
@@ -79,22 +113,48 @@ class FacilityTypeDropdown extends Component {
...
-                aria-relevant="additions"
-                htmlFor="facility-dropdown-toggle"
-                id="facility-dropdown-label"
-              >
-                Search for
+              <div className="row">
+                <div className="columns medium-4">
+                  <label
+                    aria-live="polite"
+                    aria-relevant="additions"
```
1Copenut: We will want to move these `aria-live` and `aria-relevant` attributes down to `<div id="infoBubble">` after having a chance to test with a screen reader. It's the containing element, and causes screen readers to read the entire popup.

The item in question is on line 164, approximately. I changed its markup to look like the following and verified with VoiceOver:

```html
<div
  aria-live="polite"
  aria-relevant="additions"
  id="infoBubble"
>
```

1Copenut (changes requested):
HI @r-mc2. I reviewed your change set by pulling the branch locally and trying it out with Safari + VoiceOver. I made a few requests for change, and will retest when ready. Thank you.

1Copenut:
@r-mc2 The changes I requested were all related to #15428 (focus management). The placement in source order for #15421 is right on.

src/applications/facility-locator/components/FacilityTypeDropdown.jsx
1Copenut: Yes, add the call to `focusCCInfoIcon` after the `setState` declaration. So actually on line 67 or 68 if you leave a space.

That gives us the right state for the next interaction, and then sets focus correctly as a second discrete action.

src/applications/facility-locator/components/FacilityTypeDropdown.jsx
r-mc2: Yea, I was worried that might be the case. Was considering adding some other flag to the component to track where we were in terms of onLoad vs. having clicked the info button to try and prevent something like that.
I know we do need to track the focus setting in a lifecycle method as the way I manage whether or not to show the info bubble relies on React's internal re-rendering (by way of updating the `state` variable).

If removing this `else` fixes it? 🎉

1Copenut: I had success removing the `else` block here and adding the `focusCCInfoIcon()` method in my comment below, so I think this should work as we want.

I like to try out code swaps locally before commenting, trying to offer good suggestions the first time instead of the third or fourth.

1Copenut:
@r-mc2 These changes are spot on. I tested on `localhost` after pulling code with MacOS + Safari + VoiceOver. Source order feels right, and focus management is good. I'll test again when it's on staging to ensure compliance in NVDA and JAWS.

1Copenut (approved):

src/applications/facility-locator/components/FacilityTypeDropdown.jsx
bshyong: Recommend adding the call to focusCC in the callback param for the setState above, to guarantee it only runs after setState completes.

bshyong (approved):

r-mc2:
I actually just tried that out locally @bshyong 
```jsx
toggleCCInfo = (e) => {
  e.preventDefault();

  this.setState(
    {
      showBubble: !this.state.showBubble,
    },
    // Refocus the "i" info button
    this.focusCCInfoIcon,
  );
};
```
and having it execute inside the callback is actually causing it to not focus the `X` in the popup when you click the info "`i`" now.

bshyong:
Strange; ok feel free to leave the call as is

1Copenut:
@r-mc2 I smoke tested this just now in staging with JAWS + IE11 and NVDA + Firefox. Specifically I was evaluating it for adherence to issues https://github.com/department-of-veterans-affairs/vets.gov-team/issues/15421 and https://github.com/department-of-veterans-affairs/vets.gov-team/issues/15428. 

Both worked as expected, no qualifications. Thank you.

PR #9234, opened 2018-12-11
Use semver for vets-json-schema dependency

cvalarida:
## Description
I made a little test to see if we could use semver to keep track of our `vets-json-schema` dependency. This is part of an [open RFC](https://github.com/department-of-veterans-affairs/vets.gov-team/pull/15646).

## Testing done
I made sure it works by breaking it:
1) Changed the dependency to a [GitHub URL](https://docs.npmjs.com/files/package.json#github-urls) with a valid semver `<commit-ish>`
2) Tested that it work
3) Changed the semver to one that isn't a release
4) Tested that it didn't work

## Screenshots
N/A

## Acceptance criteria
- [x] The dependency is fetched properly

## Definition of done

### DIFF
--- a/package.json
+++ b/package.json
@@ -247,7 +247,7 @@
     "url-search-params": "^0.10.0",
     "us-forms-system": "https://github.com/usds/us-forms-system.git#f476ed0ff038403d04e531bbbae41f72290e23b5",
     "vanilla-lazyload": "^8.17.0",
-    "vets-json-schema": "https://github.com/department-of-veterans-affairs/vets-json-schema.git#92816fb3a3d5166dba1404e14c2e12739ba3b3b2",
+    "vets-json-schema": "department-of-veterans-affairs/vets-json-schema.git#v3.119.0",
     "webpack-bundle-analyzer": "^2.11.1"
   }
 }
--- a/package.json
+++ b/package.json
@@ -247,7 +247,7 @@
     "url-search-params": "^0.10.0",
     "us-forms-system": "https://github.com/usds/us-forms-system.git#f476ed0ff038403d04e531bbbae41f72290e23b5",
     "vanilla-lazyload": "^8.17.0",
-    "vets-json-schema": "department-of-veterans-affairs/vets-json-schema.git#v3.119.0",
+    "vets-json-schema": "department-of-veterans-affairs/vets-json-schema#v3.119.0",
     "webpack-bundle-analyzer": "^2.11.1"
   }
 }
--- a/script/update-json-schema.sh
+++ b/script/update-json-schema.sh
@@ -2,7 +2,7 @@
 
 # Find the versions
 NEW_VERSION=$(git ls-remote --tags git://github.com/department-of-veterans-affairs/vets-json-schema/ | awk -F / '{ print $3 }' | sort -V -r | head -n 1)
-OLD_VERSION=$(grep -oP '(?<=department-of-veterans-affairs/vets-json-schema#)(.*)(?=",$)' package.json)
+OLD_VERSION=$(grep -o -e 'department-of-veterans-affairs\/vets-json-schema#\(.*\)",\?$' package.json | awk -F '[#"]' '{ print $2 }')
 
 # Set up color output
 GREEN='\033[0;32m'
--- a/script/update-json-schema.sh
+++ b/script/update-json-schema.sh
@@ -1,5 +1,26 @@
 #!/bin/bash
 
-NEW_HASH=$(git ls-remote git://github.com/department-of-veterans-affairs/vets-json-schema/ refs/heads/master | awk '{print $1}')
-sed -i '' -e 's/\(vets-json-schema\.git#\)\(.*\)\(",$\)/\1'$NEW_HASH'\3/' package.json
+# Find the versions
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -247,7 +247,7 @@
     "url-search-params": "^0.10.0",
     "us-forms-system": "https://github.com/usds/us-forms-system.git#f476ed0ff038403d04e531bbbae41f72290e23b5",
     "vanilla-lazyload": "^8.17.0",
-    "vets-json-schema": "https://github.com/department-of-veterans-affairs/vets-json-schema.git#92816fb3a3d5166dba1404e14c2e12739ba3b3b2",
+    "vets-json-schema": "department-of-veterans-affairs/vets-json-schema.git#v3.119.0",
```
cvalarida: If `vets-json-schema` gets updated before this gets merged in, I'll update this to match the new version.

script/update-json-schema.sh
```diff
@@ -1,5 +1,26 @@
...
+    echo -e "Newest version found on GitHub matches the version currently in package.json ($NEW_VER_OUT)"
+    exit 0
+fi
+
+echo -e "Updating vets-json-schema from $OLD_VER_OUT to $NEW_VER_OUT"
+
+REPLACE_COMMAND='s/\(department-of-veterans-affairs\/vets-json-schema#\)\(.*\)\(",$\)/\1'$NEW_VERSION'\3/'
+
+# `sed -i ''` works for BSD sed, but not GNU; fall back to GNU
+sed -i '' -e "$REPLACE_COMMAND" package.json 2>/dev/null || sed -i -e "$REPLACE_COMMAND" package.json
```
cvalarida: Only problem I can see with this is if `sed` fails for some reason that doesn't stem from the BSD / GNU mismatch. That might just be if `package.json` isn't there (in which case, the second command will likewise fail).

It _should_ be okay, though. :crossed_fingers:

script/update-json-schema.sh
```diff
@@ -1,5 +1,26 @@
 #!/bin/bash
 
-NEW_HASH=$(git ls-remote git://github.com/department-of-veterans-affairs/vets-json-schema/ refs/heads/master | awk '{print $1}')
-sed -i '' -e 's/\(vets-json-schema\.git#\)\(.*\)\(",$\)/\1'$NEW_HASH'\3/' package.json
+# Find the versions
+NEW_VERSION=$(git ls-remote --tags git://github.com/department-of-veterans-affairs/vets-json-schema/ | awk -F / '{ print $3 }' | sort -V -r | head -n 1)
```
cvalarida: I feel like this might require a bit of an explanation:
1) `git ls-remote --tags ...`: Fetch all the tags
2) `awk -F / ...`: Set the separator to `/` and print the third item--always the tag since the format is always `hash refs/tags/tagName`
3) `sort -V -r`: Sort by version in reverse order, so the latest is at the top
4) `head -n 1`: Take only the first entry (the latest version since we sorted in reverse order)

Only thing I'm not sure of there is if BSD `sort` has `-V`. Can somebody verify for me?

erikphansen: BSD `sort` has the `-V` flag:

```
-V, --version-sort
             Sort version numbers.  The input lines are treated as file names in form PREFIX VERSION SUFFIX, where SUFFIX matches the regular expression "(.([A-Za-z~][A-Za-
             z0-9~]*)?)*".  The files are compared by their prefixes and versions (leading zeros are ignored in version numbers, see example below).  If an input string does not match
             the pattern, then it is compared using the byte compare function.  All string comparisons are performed in C locale, the locale environment setting is ignored.
```

script/update-json-schema.sh
```diff
@@ -1,5 +1,26 @@
...
+OLD_VERSION=$(grep -oP '(?<=department-of-veterans-affairs/vets-json-schema#)(.*)(?=",$)' package.json)
```
erikphansen: the `vets-json-schema` entry is dangerously close to the end of the list of dependencies. Maybe to be extra super duper safe make that `,` before the end of the line optional?

erikphansen:
Seems like the grep usage on Mac is also different (there is no uppercase `-P` flag):

```
yarn update:schema
yarn run v1.5.1
$ ./script/update-json-schema.sh
usage: grep [-abcDEFGHhIiJLlmnOoqRSsUVvwxZ] [-A num] [-B num] [-C[num]]
	[-e pattern] [-f file] [--binary-files=value] [--color=when]
	[--context[=num]] [--directories=action] [--label] [--line-buffered]
	[--null] [pattern] [file ...]
Updating vets-json-schema from  to v3.119.0
```

script/update-json-schema.sh
cvalarida: Ahh, yeah. Good call.

cvalarida:
Blast. Okay. I can probably get away without using Perl regexes. :thinking:

cvalarida:
Aaand done! One of these days, I may even get semi-proficient at bash. :stuck_out_tongue_winking_eye:

script/update-json-schema.sh
cvalarida: Excellent! Thanks. :smile:

erikphansen (commented):
tested locally on mac os 10.14.1 👍

PR #9245, opened 2018-12-12
Upgrade yarn and add security check based on yarn audit

jbalboni:
## Description

This PR upgrades Yarn so that we can use `yarn audit` and it uses the output of that command to find valid issues for us, similar to what NSP was doing previously.

## Testing done
Ran locally

## Acceptance criteria
- [x] Security check works
- [x] Build fails when security check fails

## Definition of done

### DIFF
--- a/.nsprc
+++ b/.nsprc
@@ -1,6 +1,6 @@
 {
   "exceptions": [
-    "https://nodesecurity.io/advisories/577",
-    "https://nodesecurity.io/advisories/157"
+    "https://npmjs.com/advisories/577",
+    "https://npmjs.com/advisories/157"
   ]
 }
--- a/.nsprc
+++ b/.nsprc
@@ -1,6 +1,0 @@
-{
-  "exceptions": [
-    "https://npmjs.com/advisories/577",
-    "https://npmjs.com/advisories/157"
-  ]
-}
--- a/script/security-check.js
+++ b/script/security-check.js
@@ -1,4 +1,9 @@
 /* eslint-disable no-console */
+/*
+ * This script parses the yarn audit JSON results to find security advisories
+ * that affect modules in our dependencies list that are moderate or higher
+ * and aren't in our exceptions list
+ */
 const fs = require('fs');
 const path = require('path');
 const { spawn } = require('child_process');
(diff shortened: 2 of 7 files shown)


script/security-check.js
```diff
@@ -0,0 +1,86 @@
...
+    .map(a => JSON.parse(a));
+
+  const validAdvisories = advisories.filter(adv => {
+    if (adv.type === 'auditAdvisory') {
+      const advisoryData = adv.data.advisory;
+      const affectedModule = getAffectedModule(adv.data);
+      return (
+        !exceptionSet.has(advisoryData.url) &&
+        severitySet.has(advisoryData.severity) &&
+        dependencySet.has(affectedModule)
```
ncksllvn: Does this mean that we're only checking direct dependencies, instead of the dependencies required of our direct dependencies?

ncksllvn (approved):
This is awesome work. Sometime soon, it might be worth running a `yarn upgrade` to get our `yarn.lock` updated with the new version of Yarn. I thought I saw a performance boost after doing that while running the audit command locally.

script/security-check.js
jbalboni: No, the audit goes through and checks sub-dependencies and gives you a "path" to the dependency with the issue. The root of that path would be a dependency that's in our package.json, so this is restricting it to just ones in `dependencies` vs `devDependencies`.

.nsprc
```diff
@@ -1,6 +1,6 @@
 {
   "exceptions": [
-    "https://nodesecurity.io/advisories/577",
-    "https://nodesecurity.io/advisories/157"
+    "https://npmjs.com/advisories/577",
+    "https://npmjs.com/advisories/157"
   ]
-}
\ No newline at end of file
+}
```
ncksllvn: Ah last thing - would it make sense to rename this file to something else now?

jbalboni: Yeah, probably. Does it even need to be its own file?

ncksllvn: Probably not, unless we're going to add security exceptions all the time 😆

rianfowler (approved):

PR #9250, opened 2018-12-12
[FacilityLocator] Provider Locator Address & Map Fixes

r-mc2:
These should remedy the the two big items we discussed this afternoon and will pair with one of @SimonSchama 's PRs going in for the API tonight.
Two items being:
- [Static Map on Details page not generating](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/15273)
- `null`s appearing in the address fields for Providers that have bad/missing/malformed data.
- (Also snuck in a couple other fixes we found along the way or were small enough to include)

## Description
Updates/Tweaks/Changes for Address & Details Map
* New mock data for testing empty address fields
* Moved LocationMap's `pinNames` out to `constants/index.js`
* Changed how to check which pin png to use in the Mapbox API call for static maps
* Additional small changes to other files to handle empty address fields, returning appropriate data to indicate to parent components, and formatting text to display to the user
* Fixed tiny visual glitch where Details page had an extra space before "Directions"
* Fixed tiny warning from ProviderDetail where it would warn about `location` not existing while being required
  * Fix was removing the `isRequired` as `location` is something mapped off of props out of the Redux store and that value isn't loaded until a service call is made in the component's `componentWillMount`

(@'s for visibility to @rtluu @ehuntdsva & @brettdt )

### DIFF
--- a/src/applications/facility-locator/components/search-results/LocationDirectionsLink.jsx
+++ b/src/applications/facility-locator/components/search-results/LocationDirectionsLink.jsx
@@ -13,7 +13,8 @@ class LocationDirectionsLink extends Component {
           href={`https://maps.google.com?saddr=Current+Location&daddr=${address}`}
           target="_blank"
         >
-          <i className="fa fa-road" /> Directions
+          <i className="fa fa-road" />
+          Directions
         </a>
       </span>
     );
--- a/src/applications/facility-locator/components/search-results/LocationDirectionsLink.jsx
+++ b/src/applications/facility-locator/components/search-results/LocationDirectionsLink.jsx
@@ -19,6 +19,7 @@ class LocationDirectionsLink extends Component {
       <span>
         <a
           href={`https://maps.google.com?saddr=Current+Location&daddr=${address}`}
+          rel="noopener noreferrer"
           target="_blank"
         >
           <i className="fa fa-road" />
--- a/src/applications/facility-locator/components/search-results/LocationDirectionsLink.jsx
+++ b/src/applications/facility-locator/components/search-results/LocationDirectionsLink.jsx
@@ -5,7 +5,15 @@ import { buildAddressArray } from '../../utils/facilityAddress';
 class LocationDirectionsLink extends Component {
   render() {
     const { location } = this.props;
-    const address = buildAddressArray(location).join(', ');
+    let address = buildAddressArray(location);
+
+    if (address.length !== 0) {
+      address = address.join(', ');
+    } else {
+      // If we don't have an address fallback on coords
+      const { lat, long } = location.attributes;
+      address = `${lat},${long}`;
+    }
 
     return (
       <span>
--- a/src/applications/facility-locator/constants/index.js
+++ b/src/applications/facility-locator/constants/index.js
@@ -1,4 +1,3 @@
-/* eslint-disable camelcase */
 import { ccLocatorEnabled } from '../config';
 
 /**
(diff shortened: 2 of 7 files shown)


src/applications/facility-locator/components/search-results/LocationDirectionsLink.jsx
```diff
@@ -5,15 +5,24 @@ import { buildAddressArray } from '../../utils/facilityAddress';
...
+      address = address.join(', ');
+    } else {
+      // If we don't have an address fallback on coords
+      const { lat, long } = location.attributes;
+      address = `${lat},${long}`;
+    }
 
     return (
       <span>
         <a
```
jbalboni: Not directly related to your changes: can we add `rel="noopener"` here?

src/applications/facility-locator/constants/index.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable camelcase */
```
jbalboni: Can we add this comment just surrounding the block with camel case properties?

src/applications/facility-locator/constants/index.js
```diff
@@ -31,6 +32,17 @@ export const FacilityType = {
   VET_CENTER: 'vet_center',
 };
 
+/**
+ *
```
jbalboni: Should we remove this empty comment?

src/applications/facility-locator/utils/facilityAddress.js
```diff
@@ -1,15 +1,20 @@
-import { compact } from 'lodash';
+import compact from 'lodash/compact';
+import isEmpty from 'lodash/isEmpty';
```
jbalboni: For what it's worth, you can use the destructuring approach or the normal `import _ from 'lodash'` and the Lodash Babel plugin will convert the imports to look like this, so it doesn't bloat the bundle more.

jbalboni (approved):
Some minor comments, but LGTM

src/applications/facility-locator/components/search-results/LocationDirectionsLink.jsx
r-mc2: Good call. I'd been adding that to other anchors across other portions of the Locator, too.

src/applications/facility-locator/constants/index.js
r-mc2: Hah! I meant to actually leave a comment about this block, missed it in my rush. Thanks 👍

src/applications/facility-locator/utils/facilityAddress.js
r-mc2: Ahh, didn't know we had the babel plugin added. Last I paid attention we did not so I got in the habit of doing the direct imports (even mentioned it in my [overall review](https://gist.github.com/r-mc2/a2e269bc65eb77291de5cb5c20cef39f) of the app that I passed along months ago). 
I can definitely put it back to the destructure format if that's the case.

src/applications/facility-locator/constants/index.js
r-mc2: Moved 👍

rianfowler (approved):

src/applications/facility-locator/utils/facilityAddress.js
jbalboni: I don't think we have a prescribed way of importing them, so it's up to you, I just wanted to mention that they're all equivalent from a bundle size standpoint (unless you start using `_.chain`, anyway).

PR #9252, opened 2018-12-13
Preview content server

jbalboni:
## Description
This PR adds a Node server that can preview content from GH by running it through Metalsmith on demand.

It currently only works locally, but it can be deployed on Heroku if we need to.

## Testing done
Checked locally

## Acceptance criteria
- [x] Runs locally, doesn't break Heroku

## Definition of done


rianfowler (approved):
What's the difference between this and what we currently run on heroku? I can see it doesn't do somethings like build a sitemap.

preview/index.js
```diff
@@ -0,0 +1,78 @@
...
+      // eslint-disable-next-line no-console
+      console.log(err);
+      res.sendStatus(500);
+    });
+});
+
+app.listen(process.env.PORT || options.port, () => {
+  // eslint-disable-next-line no-console
+  console.log('Content preview server running');
+});
```
ncksllvn: Would it make sense for this to also live in `src/site/preview` or `script`?, or possibly `src/platform`? I'd like to keep the root of the repo pretty slim if possible.

src/site/stages/preview/index.js
```diff
@@ -0,0 +1,138 @@
...
+  smith.use(createBuildSettings(BUILD_OPTIONS));
+
+  smith.use(updateExternalLinks(BUILD_OPTIONS));
+
+  // smith.use(addAssetHashes(BUILD_OPTIONS));
+
+  return smith;
+}
+
+module.exports = createPipeline;
```
ncksllvn: Is it possible to share this pipeline/function with `src/stages/build/index.js`? At a glance it looks like the pipelines are pretty similar.

ncksllvn (approved):
Couple questions around organization but the server is really cool.

src/site/stages/preview/index.js
jbalboni: I looked into that a little bit and it would take some reorganization to make that work. There's not one single block that is the same as the build pipeline. I definitely want to make them share as much code as possible, but it felt like it was enough work to be a separate ticket.

preview/index.js
jbalboni: Yeah, I think I can put it back in script. It felt weird to put it in `src/`, since it's pretty separate from most of our source code.

jbalboni:
@rianfowler This server assumes there's already a build done in the background. The Heroku instances are a generic http server serving static content, but this server runs content through our Metalsmith pipeline on demand when you go to `/preview` and pass in a path to a markdown file in a query param.

PR #9253, opened 2018-12-13
Next Button Bug Fix

imeecuison:
## Description
Fix for bug where user cannot select NEXT with TAB button on intro page.
On the introduction page click "Find your disability claim form", hit the TAB button until you see the first bullet highlighted.
Press the DOWN arrow on your keyboard to select the second bullet.
Press the TAB key to go to the next selection which should be the NEXT button
--> Notice the highlight jumped to the link below the Next button completely skipping the button.
User has no way to continue. Same thing happens on the next page. After the user tabs from the "Back" button, it should go to "Next" but instead goes to the links below it.

http://localhost:3001/disability-benefits/apply/form-526-all-claims/introduction

## Testing done
No testing done.

## Screenshots
screen shot 2018-12-13 at 12 04 16 pm
screen shot 2018-12-13 at 12 03 44 pm

## Acceptance criteria
- [ ] User able to select NEXT with TAB Button and pressing ENTER.

## Definition of done

### DIFF
--- a/src/applications/disability-benefits/526EZ/components/ButtonContainer.jsx
+++ b/src/applications/disability-benefits/526EZ/components/ButtonContainer.jsx
@@ -67,7 +67,7 @@ export default function ButtonContainer(props) {
         </a>
       )}
       {!atGuidance() && (
-        <div
+        <a
           className="usa-button-primary"
           role="button"
           tabIndex="0"
@@ -76,7 +76,7 @@ export default function ButtonContainer(props) {
         >
           Next
           <span className="button-icon"> »</span>
-        </div>
+        </a>
       )}
     </div>
   );
--- a/src/applications/disability-benefits/526EZ/components/ButtonContainer.jsx
+++ b/src/applications/disability-benefits/526EZ/components/ButtonContainer.jsx
@@ -67,16 +67,15 @@ export default function ButtonContainer(props) {
         </a>
       )}
       {!atGuidance() && (
-        <a
+        <button
           className="usa-button-primary"
-          role="button"
           tabIndex="0"
           onClick={goForward}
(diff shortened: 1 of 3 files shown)


aaronjlech (approved):

src/applications/disability-benefits/526EZ/components/ButtonContainer.jsx
```diff
@@ -66,7 +67,13 @@ export default function ButtonContainer(props) {
         </a>
       )}
       {!atGuidance() && (
-        <a className="usa-button-primary" onClick={goForward}>
+        <a
```
cvalarida: I'm not sure why this was an anchor in the first place. Since it doesn't have an `href` and is _trying_ to act like a button, we should use the semantic `<button>` element.

I tested the following and it worked as expected.
```jsx
        <button className="usa-button-primary" onClick={goForward}>
          Next
          <span className="button-icon"> »</span>
        </button>
```

cvalarida (dismissed):

src/applications/disability-benefits/526EZ/components/ButtonContainer.jsx
jbalboni: Could we change this from an `a` to a `button`? That seems like a simpler approach.

jbalboni: I agree with this ^

imeecuison: Changed to button. Thank you!

jbalboni (approved):
One minor cleanup comment, but otherwise looks good.

PR #9259, opened 2018-12-14
Facility Locator - Fix for map locking up while dragging

ncksllvn:
## Description
Previously, when you would drag the map around on the Facility Locator, the page would sometimes lock up. This was due to a bug with the `onMoveEnd` property of the Map component provided by `react-leaflet`.

This PR upgrades our module `react-leaflet` to version `1.9.1` to fix this issue. I don't know what version actually patched this, but as of upgrading I can't replicate that error again.

## Testing done
- Local testing use the Dev API

## Screenshots
Todo

## Acceptance criteria
- [x] The Facility Locator no longer locks up while using

## Definition of done

### DIFF
--- a/src/applications/facility-locator/containers/VAMap.jsx
+++ b/src/applications/facility-locator/containers/VAMap.jsx
@@ -126,8 +126,8 @@ class VAMap extends Component {
         - revGeocodeInProgress - should be a separate flag as both operations happen
         - searchRequested - To track that the user clicked the search button
           (could have used inProgress but it gets tripped by other Actions)
-        - 
-      
+        -
+
       The boundary checking of the current code below doesn't actually work.
       Array equality isn't something that should be done with the operator,
       and using the new method below causes `searchWithBounds` to never fire.
@@ -451,6 +451,7 @@ class VAMap extends Component {
     const coords = this.props.currentQuery.position;
     const position = [coords.latitude, coords.longitude];
     const { currentQuery, results, pagination, selectedResult } = this.props;
+    const facilityLocatorMarkers = this.renderFacilityMarkers();
 
     return (
       /* eslint-disable prettier/prettier */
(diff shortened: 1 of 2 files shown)


src/applications/facility-locator/containers/VAMap.jsx
```diff
@@ -531,9 +536,12 @@ class VAMap extends Component {
...
                   <a href="http://creativecommons.org/licenses/by-sa/2.0/">CC-BY-SA</a>, \
                   Imagery © <a href="http://mapbox.com">Mapbox</a>' />
-              <FeatureGroup ref="facilityMarkers">
-                {this.renderFacilityMarkers()}
-              </FeatureGroup>
+              {facilityLocatorMarkers.length > 0 &&
+                <FeatureGroup
+                  ref="facilityMarkers">
+                  {facilityLocatorMarkers}
+                </FeatureGroup>
```
ncksllvn: An error was being thrown in `FeatureGroup ` if `renderFacilityMarkers` returned an empty array.

r-mc2: So moving the building of the markers array up above to the variable was the fix for this? Weird it would throw.

r-mc2 (commented):
You said you updated the `react-leaflet` package and I do see the `yarn.lock` file, but don't we need the `package.json` update as well?

jbalboni:
It looks like the current range for that module in package.json is >=1.2.0 and <2.0, so 1.9 doesn't force an update.

jbalboni (approved):

src/applications/facility-locator/containers/VAMap.jsx
ncksllvn: The fix was checking the length of the markers array prior to rendering the `FeatureGroup`, but I moved the variable above so `renderFacilityMarkers` only needed to be called once.

PR #9262, opened 2018-12-14
[All Claims] Form0781 transform

devin-mccurdy:
## Description

Transform 781 form data to match 781 schema.

This is related to [Story 15153](https://app.zenhub.com/workspaces/vft-59c95ae5fda7577a9b3184f8/issues/department-of-veterans-affairs/vets.gov-team/15153)

## Testing done
Updated test "maximal" form data and transformed schema validated 

## Screenshots
N/A

## Acceptance criteria
- [x] Transform has been updated to add all 0781 data to 526 All Claims
- [x] Transform has been updated to add all 0781a data to 526 All Claims

## Definition of done


src/applications/disability-benefits/all-claims/utils.jsx
```diff
@@ -267,6 +267,29 @@ export function transformProviderFacilities(providerFacilities) {
...
+    .join(', ');
+}
+
+export function getPtsdChangeFieldTitles(changeFields, formConfig) {
+  const titles = [];
+  Object.keys(changeFields)
+    .filter(
+      key => key !== 'other' && key !== 'otherExplanation' && formConfig[key],
+    )
+    .forEach(key => {
```
aaronjlech: Is there any way this could be a map instead, and just `return Object.keys(fields).filter(key => .....).map(key => .....)`. Or maybe even a `reduce()` ?

src/applications/disability-benefits/all-claims/utils.jsx
```diff
@@ -390,6 +413,75 @@ export function transform(formConfig, form) {
...
+    'secondaryIncident0',
+    'secondaryIncident1',
+    'secondaryIncident2',
+  ];
+
+  const incidents = incidentKeys
+    .filter(incidentKey => clonedData[incidentKey])
+    .map(incidentKey => ({
+      ...clonedData[incidentKey],
+      personalAssault: incidentKey.includes('secondary'),
```
aaronjlech: 👍 Gotta love when you unintentionally setup code for success in the future!

src/applications/disability-benefits/all-claims/utils.jsx
```diff
@@ -390,6 +413,75 @@ export function transform(formConfig, form) {
     delete clonedData.providerFacility;
   }
 
+  const incidentKeys = [
```
aaronjlech: I wonder if it would buy us anything to make this dynamic based off of the formIteration constant?

aaronjlech (commented):
Looks good! Just a few questions/suggestions

src/applications/disability-benefits/all-claims/utils.jsx
dmccurdybah: The source data is hardcoded to 6 incidents, so I think that would actually make it more fragile.

aaronjlech (approved):

dmccurdybah:
@department-of-veterans-affairs/code-reviewers-for-external-teams-group - Could I get a code review on this at your earliest convenience? Thanks!

src/applications/disability-benefits/all-claims/config/schema.js
```diff
@@ -681,7 +681,7 @@ const schema = {
     secondaryPtsdIncident: {
       type: 'object',
       properties: {
-        authorities: {
+        source: {
```
markgreenburg: We should aim to decrease reliance on this schema file, updating pages to point at the actual schema in `vets-json-schema` instead. The goal is to eventually get rid of this file entirely. Also, I think this property is called `sources` in the vets-json-schema (https://github.com/department-of-veterans-affairs/vets-json-schema/blob/master/src/schemas/21-526EZ-allclaims/schema.js#L179)

dmccurdybah: Yes, and I don't think we're referencing at anywhere at this point.  I think I'd updated this to match the schema changes and then didn't follow up when it changed further.  I can update this to match the schema if you'd like.

src/applications/disability-benefits/all-claims/pages/individualsInvolvedFollowUp.js
```diff
@@ -51,8 +51,8 @@ export const uiSchema = index => ({
             labels: {
               killedInAction: 'Killed in action',
               woundedInAction: 'Wounded in action',
-              killedInNonBattle: 'Killed non-battle',
-              injuredInNonBattle: 'Injured non-battle',
+              killedNonBattle: 'Killed non-battle',
```
markgreenburg: would it make sense to map this out in `/constants` since we're using it in multiple places?

src/applications/disability-benefits/all-claims/pages/individualsInvolvedFollowUp.js
```diff
@@ -80,7 +80,7 @@ export const schema = index => ({
     [`incident${index}`]: {
       type: 'object',
       properties: {
-        personInvolved: {
+        personsInvolved: {
```
markgreenburg: This whole schema seems to match what's in `vets-json-schema`, aside from the added `view` property and the fact that the order is different. You can import that whole `personsInvolved` schema from there and use it here, and merge the `view` property into it. If you have trouble with the display ordering, you can use the `ui:order` uiSchema property

src/applications/disability-benefits/all-claims/tests/schema/schema.unit.spec.js
```diff
@@ -4,7 +4,7 @@ import path from 'path';
 import { Validator } from 'jsonschema';
 
 import formConfig from '../../config/form';
-import fullSchema from '../../config/schema';
+import fullSchema from 'vets-json-schema/dist/21-526EZ-ALLCLAIMS-schema.json';
```
markgreenburg: 🎉

src/applications/disability-benefits/all-claims/tests/utils.unit.spec.jsx
```diff
@@ -180,6 +180,15 @@ describe('526 helpers', () => {
     });
   });
 
+  describe('transform', () => {
+    it('should transform each incident', () => {
+      expect(
```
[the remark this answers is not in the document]
markgreenburg: Reviewing further, I really like that y'all have broken out some of the transformations into their own functions, but I think these should all have test cases (i.e., `concatIncidentLocationString`, `getFlatIncidentKeys`, `getPtsdChangeFieldTitles`) - it seems as though the `length` test you already have wouldn't be sufficient to test the output of these helpers.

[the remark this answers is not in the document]
dmccurdybah: That test was from when I started testing the transform but before moving it into the existing transform method, so I don't think it adds anything.

The length test is only really testing the number of incidents in the maximal test data, so isn't really providing any value.  

I can definitely add tests for the functions if you'd like.  I do think that they're functionally covered by the actual transform.

src/applications/disability-benefits/all-claims/utils.jsx
```diff
@@ -267,6 +268,38 @@ export function transformProviderFacilities(providerFacilities) {
...
+  ]
+    .filter(locationField => locationField && locationField.length > 0)
+    .join(', ');
+}
+
+export function getFlatIncidentKeys() {
+  const incidentKeys = [];
+
+  for (let i = 0; i < PTSD_INCIDENT_ITERATION; i++) {
+    incidentKeys.push(`incident${i}`);
```
markgreenburg: 🤔 would it make sense to pull out the `'incident'` string into `constants` as well? What I'm afraid of is changing the key at a later time (say, it gets changed to `ptsdIncidentX`) and we miss changing it here. Could be overkill, but on the other hand you could have all the pages with that schema pull from that constant, as well as any place this ref is needed in the transformers. Thoughts?

src/applications/disability-benefits/all-claims/pages/individualsInvolvedFollowUp.js
dmccurdybah: Are we using it in multiple places?  I think these are just the display values and aren't needed anywhere outside of the UI schema.

src/applications/disability-benefits/all-claims/pages/incidentDescription.js
```diff
@@ -31,7 +31,7 @@ export const schema = index => ({
     [`incident${index}`]: {
       type: 'object',
       properties: {
-        description,
+        incidentDescription: description,
```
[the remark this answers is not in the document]
dmccurdybah: I don't really understand what we're gaining by referencing ptsdIncident in the schema in the first place. What's the advantage of doing that for a string field?

src/applications/disability-benefits/all-claims/pages/individualsInvolvedFollowUp.js
dmccurdybah: It looks like the reason that we have this in the UI is because of the personsInvolved.name using the `fullName` definition, which we don't want.  I just noticed that this is an issue in the schema as well.  I'll update this in `vets-json-schema` so that this is possible.

src/applications/disability-benefits/all-claims/utils.jsx
dmccurdybah: I understand the concern, and I'd like to address that in a separate PR.

src/applications/disability-benefits/all-claims/utils.jsx
```diff
@@ -398,6 +431,68 @@ export function transform(formConfig, form) {
...
+  });
+
+  if (incidents.length > 0) {
+    clonedData.form0781 = {
+      incidents,
+      remarks: clonedData.additionalRemarks781,
+      additionalIncidentText: clonedData.additionalIncidentText,
+      additionalSecondaryIncidentText:
+        clonedData.additionalSecondaryIncidentText,
+      otherInformation: [
```
markgreenburg: I'm not sure I'm clear on what's happening here; are we passing an array of `ui:title` values for the following properties to `vets-api`? If that's the case, I think we could add a piece of hidden formdata that does essentially the same thing, rather than depend on the uiSchema. @cvalarida any thoughts?

markgreenburg (commented):
Mainly I think there are more unit tests needed, but I'm also not sure about the reliance on the `uiSchema` in the transformer. Would like @cvalarida to take a pass at this as well. I think there's also an opportunity to import `personsInvolved` from vets-json-schema instead of redefining it in the page schema.

src/applications/disability-benefits/all-claims/utils.jsx
dmccurdybah: Yes, incident `otherInformation` are used as lines in the 781a otherInformation textbox.  We're sending the UI display values instead of the enum values because we want what is actually displayed to the user to make it onto the PDF.

src/applications/disability-benefits/all-claims/config/schema.js
markgreenburg: I was just thinking about this myself - would make it easier to unwind later.

src/applications/disability-benefits/all-claims/pages/individualsInvolvedFollowUp.js
dmccurdybah: I've updated the personsInvolved name in vets-json-schema to make this do-able.

src/applications/disability-benefits/all-claims/tests/utils.unit.spec.jsx
```diff
@@ -180,6 +180,15 @@ describe('526 helpers', () => {
     });
   });
 
+  describe('transform', () => {
```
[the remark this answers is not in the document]
dmccurdybah: Removed!

src/applications/disability-benefits/all-claims/tests/utils.unit.spec.jsx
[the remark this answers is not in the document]
dmccurdybah: I've added unit tests for `concatIncidentLocationString`, `getFlatIncidentKeys` and `getPtsdChangeFieldTitles`

src/applications/disability-benefits/all-claims/utils.jsx
```diff
@@ -267,6 +268,38 @@ export function transformProviderFacilities(providerFacilities) {
...
+
+  return incidentKeys;
+}
+
+export function getPtsdChangeFieldTitles(changeFields, formConfig) {
+  return Object.keys(changeFields)
+    .filter(
+      key => key !== 'other' && key !== 'otherExplanation' && formConfig[key],
+    )
+    .map(key => formConfig[key]['ui:title']);
```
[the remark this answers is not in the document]
dmccurdybah: I feel like a react component or snippet wouldn't work as intended for the 781 form so I added a check to only support string `ui:title` values.

src/applications/disability-benefits/all-claims/pages/individualsInvolvedFollowUp.js
cvalarida: What we've typically done is use the version in the shared schema and call out specific updates that need to happen. Something like:
```js
const personsInvolvedSchema = set('items.properties.rank.ui:collapsed', true, personsInvolved); // Pulled from vets-json-schema
personsInvolvedSchema = set('items.properties.unitAssigned.ui:collapsed', true, personsInvolvedSchema);
// etc.
```

I'm dreaming up writing a little `setMany` function to help with this, but it doesn't exist yet.

I see what you mean for the `name`. You don't want the suffix, right? That should be taken care of in the shared schema. It's fine if this overrides what's in the schema until you can get that fixed up, though. :+1:

:thinking: Although it looks like you removed the `name` schema override so it'll use the `fullName` definition. Was that intentional?

src/applications/disability-benefits/all-claims/pages/secondaryIncidentDescription.js
```diff
@@ -29,7 +29,7 @@ export const schema = index => ({
     [`secondaryIncident${index}`]: {
       type: 'object',
       properties: {
-        description,
+        incidentDescription: description,
```
cvalarida: Should this property still be called [`description`](https://github.com/department-of-veterans-affairs/vets-json-schema/blob/master/dist/21-526EZ-ALLCLAIMS-schema.json#L1538)?

src/applications/disability-benefits/all-claims/tests/schema/schema.unit.spec.js
cvalarida: :100:

src/applications/disability-benefits/all-claims/tests/schema/transformedData.js
```diff
@@ -187,5 +187,141 @@ export const transformedMaximalData = {
...
+                last: 'Doe',
+              },
+            },
+          ],
+        },
+        {
+          personalAssault: true,
+          incidentDate: '1994-04-03',
+          incidentLocation:
+            'City Name, AZ, USA, First 781a PTSD incident location additional details',
```
cvalarida: :thinking: How will this be used in the form submitted to...VBMS? (Not sure where the 781(a) goes.)

I'm trying to figure out if adding the description as part of what looks like a formatted address makes sense.

src/applications/disability-benefits/all-claims/utils.jsx
```diff
@@ -267,6 +268,51 @@ export function transformProviderFacilities(providerFacilities) {
...
+ * @returns {String} incident location string
+ */
+export function concatIncidentLocationString(incidentLocation) {
+  return [
+    incidentLocation.city,
+    incidentLocation.state,
+    incidentLocation.country,
+    incidentLocation.additionalDetails,
+  ]
+    .filter(locationField => locationField && locationField.length > 0)
```
cvalarida: If `locationField === ''`  then `locationField` is falsey. I don't think you need that second condition.
image

src/applications/disability-benefits/all-claims/utils.jsx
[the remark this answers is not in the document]
cvalarida: I'm also uncomfortable with relying on the `uiSchema`. What if we made a map that can be used both in the `uiSchema` and in this function?

Do we have to send the api the exact title of the boxes they've selected? This might be something best left to the PDF generator or the api if it affects what gets put on the PDF. Separation of concerns and all that, y'know?

src/applications/disability-benefits/all-claims/pages/secondaryIncidentDescription.js
dmccurdybah: Hmm...

`incidentDescription` is correct.  I updated the property on the `ptsdIncident` to be `incidentDescription` but missed the `ptsdSecondayincident` description property.

src/applications/disability-benefits/all-claims/tests/schema/transformedData.js
dmccurdybah: Yes, this is effectively one text field on the 781 form (and anything that overflows the available space is appended as additional pages)

src/applications/disability-benefits/all-claims/tests/utils.unit.spec.jsx
```diff
@@ -715,4 +722,73 @@ describe('isAnswering781aQuestions', () => {
...
+
+  describe('getPtsdChangeFieldTitles', () => {
+    const uiSchema = {
+      field1: {
+        'ui:title': 'field1',
+      },
+      field2: {
+        'ui:title': 'field2',
+      },
+      field3: <div>test</div>,
```
markgreenburg: I think you want to nest this in a `ui:title` property like you've done for the other two field properties.

src/applications/disability-benefits/all-claims/utils.jsx
```diff
@@ -267,6 +268,51 @@ export function transformProviderFacilities(providerFacilities) {
...
+}
+
+export function getPtsdChangeFieldTitles(changeFields, formConfig) {
+  return Object.keys(changeFields)
+    .filter(
+      key =>
+        key !== 'other' &&
+        key !== 'otherExplanation' &&
+        formConfig[key] &&
+        typeof formConfig[key]['ui:title'] === 'string',
```
cvalarida: :thinking: This actually has me pretty concerned.

What happens when we forget this little nugget (that all of these check boxes need string titles) and we have to embolden a word later? We'd change the `ui:title` to a react component and add a `<strong>` tag. This will work without failing anywhere down the line, but when we go to send the data to the api, if a user selects that option with the emboldened word, we'll filter it out here.

This means the option they selected won't make it to the PDF and we'll never be notified.

Grabbing the text from the `ui:title` to send to the PDF makes me a bit uneasy, but it's _mostly_ okay. We just need to protect our future selves for when we forget these all need to be strings.

Can you make a unit test that ensures all the fields that run through this function have a string `ui:title`? If we have to change this up later to embolden a word or something, we can refactor this bit, but I want to make sure we catch it in the first place.

cvalarida (changes requested):

src/applications/disability-benefits/all-claims/utils.jsx
dmccurdybah: I moved the titles to `constants` so it can be referenced by the transform and the ui, but the ui would be able to override the value in the UI if it ever came up. I also added a test to validate that the mapped change labels are strings.

src/applications/disability-benefits/all-claims/tests/utils.unit.spec.jsx
```diff
@@ -715,4 +722,53 @@ describe('isAnswering781aQuestions', () => {
...
+  describe('getFlatIncidentKeys', () => {
+    it('should return correct amount of incident keys', () => {
+      expect(getFlatIncidentKeys().length).to.eql(PTSD_INCIDENT_ITERATION * 2);
+    });
+  });
+
+  describe('getPtsdChangeText', () => {
+    it('should have valid labels', () => {
+      Object.keys(PTSD_CHANGE_LABELS).forEach(key => {
+        expect(PTSD_CHANGE_LABELS[key]).to.be.a('string');
```
cvalarida: Ahh, this is great! :tada: I think the only scenario we're not covering now is if the key isn't in the constant. That'll still do the wrong thing silently.

I hate to keep asking for more tests, but could you make another one to iterate through all the places you're using this constant and ensure that the all the properties are found in it? I'd hate to get caught on something silly like renaming a property, y'know?

src/applications/disability-benefits/all-claims/utils.jsx
cvalarida: With a recent update, we're no longer pulling directly from the `ui:title`. The strings we add in here now just happen to be the same as the titles, which is a great improvement. :tada:

src/applications/disability-benefits/all-claims/pages/secondaryIncidentDescription.js
cvalarida: I talked with @dmccurdybah "offline" (Slack) and he clarified: These `secondaryIncident${index}` properties get changed into items in the `form0781.incidents` array. These items do indeed have `incidentDescription`, not `description`.

Thanks for the clarification, @dmccurdybah! :raised_hands:

src/applications/disability-benefits/all-claims/tests/utils.unit.spec.jsx
```diff
@@ -715,4 +722,98 @@ describe('isAnswering781aQuestions', () => {
...
+      });
+    });
+
+    const ignoredFields = ['other', 'otherExplanation', 'noneApply'];
+    it('should have mappings for all workBehaviorChanges schema fields', () => {
```
cvalarida: :tada:

cvalarida (approved):
Thanks for making those changes!

PR #9270, opened 2018-12-17
Update Find VA Benefits to use BE

erikphansen:
## Description
This PR updates the FE to use the new endpoints to fetch and update user preferences, which are currently only used to allow the user to select benefits they'd like to learn more about.

## Testing done
- added unit tests especially for redux work
- local testing

## Screenshots
The UI is mostly unchanged from the latest version that was using local storage to fake persisting data

## Acceptance criteria
- [ ]

## Definition of done


src/applications/personalization/preferences/actions/index.js
```diff
@@ -1,33 +1,138 @@
...
+export const SAVED_DASHBOARD_PREFERENCES = 'SAVED_DASHBOARD_PREFERENCES';
 
-  return {
-    type: DASHBOARD_PREFERENCES_FETCHED,
-    data: (savedPrefs && JSON.parse(savedPrefs)) || {},
+// load the benefits the user has picked to learn more about
+export function fetchUserSelectedBenefits() {
+  return dispatch => {
+    dispatch({
+      type: SET_USER_PREFERENCE_REQUEST_STATUS,
```
rianfowler: Idiomatic redux forks the behavior with different action types i.e. it uses on variable on the action instead of many. Couple of issues with this: 
- the tooling is built around actions- if you use redux dev tools with this you will see a log of the same action twice and will need to dig into it to see if the app is behaving the way you expect
- I don't see any reducers for this action but I assume this will fork the behavior with a switch and an if vs reducers usually forking with a single switch statement. 

We have an open issue from @jbalboni about this: https://github.com/department-of-veterans-affairs/vets.gov-team/issues/10845

src/applications/personalization/preferences/actions/index.js
```diff
@@ -1,33 +1,138 @@
...
+      status: LOADING_STATES.pending,
+    });
+    return apiRequest(
```
rianfowler: I think this is fine but wanted to point out that `apiRequest` also returns a promise. `async` `await` or `then` `catch` might be cleaner here.

src/applications/personalization/preferences/actions/index.js
```diff
@@ -1,33 +1,138 @@
...
+      '/user/preferences',
+      null,
+      response => {
```
rianfowler: Recommend moving everything but the dispatch into the reducer so it can be unit tested. I think actions should just dispatch- not mutate, select, query, etc.

src/applications/personalization/preferences/containers/PreferencesWidget.jsx
```diff
@@ -48,50 +61,108 @@ class PreferencesWidget extends React.Component {
...
-      item => !!dashboard[item.slug],
+      item => !!dashboard[item.code],
     );
     const hasSelectedBenefits = !!selectedBenefits.length;
     let selectedBenefitAlerts = selectedBenefits
       .filter(item => !!item.alert)
       .map(item => item.alert);
     selectedBenefitAlerts = deduplicate(selectedBenefitAlerts);
+
+    if (loadingStatus === LOADING_STATES.pending) {
```
rianfowler: Might be more declarative to use ternaries in a single return block e.g. 
`{saveState === LOADING_STATES.error && <Component>}`

even better would be to move this logic into the `mapStateToProps` or the reducer so that these states can be named e.g. 
`showSaveError` could be a prop and the reducer could create that by checking the save status.

rianfowler (approved):
Added some comments- none of this stuff is recommended in documentation. Do plan on making an RFC for this eventually. This is a diagram my friend created couple of years ago. Generally we want to move logic to the right- reducers should encapsulate most of the behavior. Because they're pure functions they're easiest to test and the action types name the behavior. 

push_logic_right

src/applications/personalization/preferences/actions/index.js
erikphansen: That action type is handled by the first reducer. But I like the approach you are advocating of having a single action type for each event.

src/applications/personalization/preferences/actions/index.js
erikphansen: I'm not totally clear on how to refactor this with async/await...

src/applications/personalization/preferences/containers/PreferencesWidget.jsx
erikphansen: You're probably right and this isn't the most idiomatic React, but it's so easy for my brain the reason about :)

src/applications/personalization/preferences/actions/index.js
rianfowler: return an async arrow function: 
```
return async () => {
  // do something
}
```
then do something like 
`const response = await apiRequest`

then dispatch your actions and `try catch` the whole thing`

rianfowler: eh maybe- I don't know if thunks work with promises

src/applications/personalization/preferences/helpers.jsx
```diff
@@ -383,7 +383,7 @@ export const benefitChoices = [
     title: 'Careers and Employment',
     description:
       'Find a job, build skills, or get support for my own business.',
-    slug: 'careers',
+    code: 'careers-employment', // TODO: update rest
```
cehsu: We can probably remove these TODO comments, as the content has been updated.

src/applications/personalization/preferences/tests/containers/SetPreferences.unit.spec.jsx
```diff
@@ -43,25 +53,26 @@ describe('<SetPreferences>', () => {
...
 
+    const component = mount(<SetPreferences {...props} />);
     component
       .find('Checkbox')
       .first()
-      .simulate('click');
-    expect(setPreference.args[0][0]).to.equal('healthcare');
-    expect(setPreference.args[0][1]).to.equal(true);
+      .simulate('click'); // TODO: update test
+    // expect(setPreference.args[0][0]).to.equal('healthcare');
```
cehsu: Do we want to hold off on updating this until a later PR?

cehsu (approved):
Looks good.

src/applications/personalization/preferences/helpers.jsx
erikphansen: I'll go ahead and remove these now.

PR #9273, opened 2018-12-18
Move runtime dependencies into dependencies list

jbalboni:
## Description
This PR moves our dependencies around so that runtime deps that need security checks are in `dependencies` and build or test dependencies are in `devDependencies`. This lines up with our security checking better.

I've also reinstalled uswds and metalsmith-collections (without updating their versions) in order to pick up changed sub-dependencies and avoid some security issues.

## Testing done
Checked site locally

## Acceptance criteria
- [x] Build still works, site looks ok

## Definition of done


rianfowler (approved):

ncksllvn (approved):
LGTM! It would be awesome if there was a place we could enforce this (maybe just with console messaging during an install hook?) or maybe set the default install to `devDependencies`. Or maybe just a rule in the readme for installing new modules?

jbalboni:
@ncksllvn I don't think there's a good way to enforce it, but I did update the readme to reflect what we should be doing. I think the default of adding modules to `dependencies` is ok; we're erring on the side of auditing modules that we might not need to, rather than missing modules that we should check.

PR #9275, opened 2018-12-18
Add local TeamSite proxy server; add "last updated" field to TeamSite footer

ncksllvn:
## Description
This pull request does two things -

1. Creates a local proxy for VA.gov pages when the `--local-proxy-rewrite` flag is passed. When that flag is detected, extra middleware is added to the Webpack Dev Server that allows us to load TeamSite pages and inject our local bundle into the page.
    - Command: `npm run watch -- --local-proxy-rewrite`
    - To preview a TeamSite page, the URL of that page should be passed via a `target` query parameter, `http://localhost:3001/?target=https://www.va.gov/health/`.

2. Adds the "last updated" field back into the footer as an extra panel that sits on top of the React footer.
    - I chose to keep this out of the React footer because I really didn't want to intertwine TeamSite-specific logic into the rest of the footer.

## Testing done
Local testing

## Screenshots
See the "last updated" panel just above the footer.

`http://localhost:3001/?target=https://www.va.gov/health/`

image

## Acceptance criteria
- [x] The "last updated" date is visible again

## Definition of done


src/site/stages/build/options.js
```diff
@@ -22,6 +22,7 @@ const COMMAND_LINE_OPTIONS_DEFINITIONS = [
   { name: 'destination', type: String, defaultValue: null },
   { name: 'content-deployment', type: Boolean, defaultValue: false },
   { name: 'content-directory', type: String, defaultValue: defaultContentDir },
+  { name: 'local-proxy-rewrite', type: Boolean, defaultValue: false },
```
ncksllvn: The local proxy server only starts when this argument is passed, because I didn't want us to hit the production site (which is necessary for TeamSite dev) unless we explicitly tell it that's what we want it to do.

rianfowler (approved):
looks good- thanks for putting this together. 

if you make documentation for this, make sure to test it on `www.benefits.va.gov`- I think they're using an iframe or or some other js to inject the stuff into the head after the document is loaded.

cehsu (approved):

ncksllvn:
Thanks @rianfowler! Testing `https://benefits.va.gov/` just redirects me to `https://benefits.va.gov/benefits/`, but the latter works okay. I think this is okay though because navigating directly to `https://benefits.va.gov/` does the same redirect.

PR #9279, opened 2018-12-18
use empty object instead of empty array

erikphansen:
## Description
Revert a stupid change Erik had made in a previous PR. Defaulting to an empty array instead of empty object made it impossible for users to select more than one benefit they were interested in.

## Testing done
Local

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/src/applications/personalization/preferences/actions/index.js
+++ b/src/applications/personalization/preferences/actions/index.js
@@ -25,7 +25,7 @@ export function fetchUserSelectedBenefits() {
       '/user/preferences',
       null,
       response => {
-        // We just want to get an array of Benefits preferences);
+        // We just want to get an array of Benefits preferences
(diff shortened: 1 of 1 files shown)


cehsu (approved):

src/applications/personalization/preferences/actions/index.js
```diff
@@ -25,7 +25,7 @@ export function fetchUserSelectedBenefits() {
       '/user/preferences',
       null,
       response => {
-        // We just want to get an array of Benefits preferences);
+        // We just want to get an array of Benefits preferences
         let selectedBenefits = response.data.attributes.userPreferences;
```
cehsu: We might rename this user preferences here and selected benefits below, in case that helps to clarify the shape the data should be.

PR #9294, opened 2018-12-20
[DON'T MERGE until vets-json-schema updated] All claims split new disabilities mg

markgreenburg:
## Description
Splits `newDisabilities` into 2 arrays for `vets-api` so that primary & secondary disabilities can be associated correctly on the backend.

Also needs an update to `vets-json-schema` before this is merged in, so that we can update that dependency in `vets-website`: https://github.com/department-of-veterans-affairs/vets-json-schema/pull/309

## Testing done
- Tested locally
- Updated maximal test cases

## Acceptance criteria
- [x] Nothing breaks
- [x] Instead of sending `newDisabilities`, we send `newPrimaryDisabilities` (`cause !== 'SECONDARY'`) and `newSecondaryDisabilities` (`cause === 'SECONDARY'`)

## Definition of done


src/applications/disability-benefits/all-claims/utils.jsx
```diff
@@ -346,6 +333,59 @@ export function transform(formConfig, form) {
...
+        return d;
+      });
+
+      delete clonedData.powDisabilities;
+    }
+
+    // Add 'cause' of 'NEW' to new ptsd disabilities since form does not ask
+    const withPtsdCause = clonedData.newDisabilities.map(
+      disability =>
+        disability.condition.toLowerCase().includes('ptsd')
```
cvalarida: Probably outside the scope of this PR since it was already in master, but we should make a function for this since we also do the same kind of check to determine whether we should enter the 781(a) flow. Would be good to consolidate that logic.

cvalarida (approved):
Looks good. As long as it passes the schema tests (to verify that the transform function works as intended),  I think it's golden.

src/applications/disability-benefits/all-claims/utils.jsx
```diff
@@ -346,6 +333,59 @@ export function transform(formConfig, form) {
...
+
+    // Split newDisabilities into primary and secondary arrays for backend
+    const newPrimaryDisabilities = withPtsdCause.filter(
+      disability => disability.cause !== causeTypes.SECONDARY,
+    );
+    const newSecondaryDisabilities = withPtsdCause.filter(
+      disability => disability.cause === causeTypes.SECONDARY,
+    );
+
+    if (newPrimaryDisabilities.length) {
```
bshyong: Will `newPrimaryDisabilities` ever be null/undefined or otherwise not have a length?  Otherwise since `0` is falsey this is a clean way to check 👍

bshyong (approved):

src/applications/disability-benefits/all-claims/utils.jsx
cvalarida: If a veteran was only claiming an increase on an existing disability, this would be empty, yeah.

markgreenburg: In this context, no. This whole chunk only gets run `if (clonedData.newDisabilities)`, where:
1. `const withPtsdCause = clonedData.newDisabilities.map`...
2. `const newPrimaryDisabilities = withPtsdCause.filter(`...

So it could be an empty array, but not undefined/null. This is definitely not awesome because this will break if we rearrange the transformer, but I think refactoring the transformer itself is something that should be done in a separate ticket.

src/applications/disability-benefits/all-claims/utils.jsx
markgreenburg: Agree, would rather not do so here just because there's already a lot of stuff getting moved around.

PR #9300, opened 2018-12-21
keep a backup of the user-selected preferences

erikphansen:
## Description
Keep a backup version of the user-selected benefits that reflects the state of things as pulled from the server. This way it can be compared to the working-copy version to see if the user has made changes and we can also restore the working-copy to the last saved version when needed.

This change also disables the SAVE button on the Find VA Benefits page if the user has not actually made any changes to their preferences.

## Testing done
Local, added/updated unit tests

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/src/applications/personalization/preferences/containers/SetPreferences.jsx
+++ b/src/applications/personalization/preferences/containers/SetPreferences.jsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import { Link, withRouter } from 'react-router';
 import { connect } from 'react-redux';
+import { isEqual } from 'lodash';
 
 import LoadingIndicator from '@department-of-veterans-affairs/formation/LoadingIndicator';
 
@@ -71,6 +72,15 @@ class SetPreferences extends React.Component {
     this.props.setPreference(code, !this.props.preferences.dashboard[code]);
   };
 
+  // checks to see if the current state of the dashboard (ie preferences
+  // selected by the user) is different from how they were when they were pulled
+  // from the server
+  userHasNotMadeChange = () =>
+    isEqual(
+      this.props.preferences.dashboard,
+      this.props.preferences.dashboardBackup,
+    );
+
   // hydrate benefit options from the backend with data from the benefitChoices
   // helper array. We are storing user-facing info in the benefitChoices array
   // so that user-facing info can be updated by the frontend devs rather than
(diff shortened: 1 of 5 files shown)


src/applications/personalization/preferences/containers/SetPreferences.jsx
```diff
@@ -71,6 +72,15 @@ class SetPreferences extends React.Component {
     this.props.setPreference(code, !this.props.preferences.dashboard[code]);
   };
 
+  // checks to see if the current state of the dashboard (ie preferences
+  // selected by the user) is different from how they were when they were pulled
+  // from the server
+  userHasNotMadeChange = () =>
+    isEqual(
+      this.props.preferences.dashboard,
+      this.props.preferences.dashboardBackup,
```
cehsu: Should this be `dashboardArchive` or `savedDashboard`, backup doesn't quite communicate what is being backed up? We can address this later if we can come up with something that seems more descriptive. perhaps this could be `savedBenefits` and `stagedBenefits` since there will be other preferences in the future, and as dashboard is actually a little generic?

src/applications/personalization/preferences/reducers/index.js
```diff
@@ -47,12 +47,19 @@ export default function preferences(state = initialState, action) {
...
+        dashboardBackup: { ...selectedBenefits },
         userBenefitsLoadingStatus: LOADING_STATES.loaded,
       };
     }
     case SET_DASHBOARD_PREFERENCE: {
-      return _.set(`dashboard.${action.code}`, action.value, state);
+      const newState = { ...state };
+      if (action.value) {
+        newState.dashboard[action.code] = true;
+      } else {
```
cehsu: What's the reasoning for this change?

cehsu (approved):
Looks good.

src/applications/personalization/preferences/containers/SetPreferences.jsx
erikphansen: Those aren't great names, either. Bottom line is I want to get the naming in better order and more consistent :)

src/applications/personalization/preferences/reducers/index.js
erikphansen: Interesting... I saw `isEqualWith` was a function but didn't look at how it works...

cehsu: I mainly mention it to avoid potential merge conflicts, so nbd.

